PR #3697, opened 2016-09-12
Worker interface objects

zcorpan:
- Rewrite shared worker interface-object tests
- Update which interface objects to expect in workers


annevk:
From IRC:

> it's not clear to me why some events have exposed=Worker and some others do not. should all events be exposed? or only those that get fired on objects that are exposed in workers?

Only those that fire in workers.

annevk:
LGTM (assuming you ran this)

PR #3700, opened 2016-09-13
Referrer Policy: Test image loads inside srcdoc frames.

mikewest:
This patch teaches the 'img-tag'-style tests to load an image both
in the top-level document, and inside a 'srcdoc' frame. The referrer
should be the same in both places.

Addresses w3c/web-platform-tests#2851.

### DIFF
--- a/referrer-policy/generic/common.js
+++ b/referrer-policy/generic/common.js
@@ -120,10 +120,8 @@ function queryImage(url, callback, referrer_policy) {
   i.onload = function () {
     loadImageInWindow(url, function (img) {
       var srcdocData = decodeImageData(extractImageData(img));
-      console.log(srcdocData);
       loadImageInWindow(url, function (img) {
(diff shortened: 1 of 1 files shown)


mikewest:
@bzbarsky: Are you happy with this? If so, I'll merge it.

bzbarsky:
Yes, this seems reasonable in general.

For dynamic referrer policy changes (e.g. dynamic meta tag insertion?), it would be nice to test the situation in which the srcdoc iframe is created and loaded _before_ the dynamic mutation happens.  But we can push that off into a separate issue; it seems like a good bit of work to do.

estark37:
@mikewest ping -- is this ready to merge? It addresses https://github.com/w3c/webappsec-referrer-policy/issues/57.

mikewest:
Thanks for the ping, Emily!

PR #3717, opened 2016-09-14
[eme] Fix drm-mp4-playback-persistent-license-events configuration

ddorwin:
Also fixes comments and other minor issues.


mwatson2:
Looks good to me. Thanks for spotting / fixing this.

PR #3719, opened 2016-09-14
[eme] Add a README with tips for running tests the first time

ddorwin:


ddorwin:
Recording some things I had to do to run the *-persistent-license-* tests. We can add other notes here too.

mwatson2:
Looks good. These are the only unusual things I am aware of.

PR #3731, opened 2016-09-15
Test the MediaStreamTrackEvent constructor

foolip:


mediacapture-streams/MediaStreamTrackEvent-constructor.https.html
```diff
@@ -0,0 +1,43 @@
...
+    new MediaStreamTrackEvent("type", {});
+  });
+  assert_throws(new TypeError, function() {
+    new MediaStreamTrackEvent("type", { track: null });
+  });
+  assert_throws(new TypeError, function() {
+    new MediaStreamTrackEvent("type", { track: undefined });
+  });
+}, "The eventInitDict's track member is required.");
+
```
dontcallmedom: wouldn't all the above be covered by a idlharness test?

foolip: Unfortunately not, "TODO: test constructors if defined."
https://github.com/w3c/testharness.js/blob/484fd841a6fcec5dd00efbc7d420bbd9e7dcbd89/idlharness.js#L799

dontcallmedom: Darn :(

OK, LGTM then :)

mediacapture-streams/MediaStreamTrackEvent-constructor.https.html
```diff
@@ -0,0 +1,43 @@
...
+    new MediaStreamTrackEvent("type", undefined);
+  });
+}, "The eventInitDict argument is required");
+
+test(function() {
+  assert_throws(new TypeError, function() {
+    new MediaStreamTrackEvent("type", {});
+  });
+  assert_throws(new TypeError, function() {
+    new MediaStreamTrackEvent("type", { track: null });
```
foolip: This line is the main reason for writing the test, it revealed that Chromium has an IDL compiler bug that I'm looking into. I intend to import this test back to Blink once landed and it also reveals some things that are wrong in Gecko.

foolip:
Thanks for the review!

PR #3744, opened 2016-09-16
Test appending an iframe to its own document

zcorpan:
See https://gist.github.com/zcorpan/26eeb298ec252f19f1bda6ee881f9d43


zcorpan:
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/9184253/

PR #3746, opened 2016-09-16
Fail unexpected keys keystatuseschange

ddorwin:
Previously, the event would have been ignored.


ddorwin:
It seems we would want the events to be as consistent as possible and that there is nothing in this test that should cause other keystatuseschange events.

This passes for Clear Key on both Chrome and Firefox, and it does not change the failure for Widevine on Chrome. However, Widevine on Firefox produces consecutive events with all keys usable (`'allkeysusable' `in the test). @cpearce, is this expected?

cpearce:
@ddorwin As a quirk of our implementation, Firefox doesn't aggregate multiple keys changing at the same time into a single status change event. i.e. for each key that changes status, we'll dispatch one key status changed event. 

I assume that there are multiple keys in the resource here? The current spec says that we should aggregate the status changed events, so we'll fix that.

ddorwin:
Thanks, @cpearce. It sounds like we should indeed make this change to catch such inconsistencies.

mwatson2 (approved):

mwatson2:
According to the specification, it is allowed for keys to transition through `status-pending`, so there may be multiple `keystatuschange` events, the first with some of all of the keys marked `status-pending` and zero or more further events, so long as the `keyStatuses` field differs from the last event each time.

So, I think it is correct to ignore any `keystatuseschange` events with one or more `status-pending` keys, but to record `allkeysusable` for every event where all keys are usable, rather than just the first such event.

In that way, if there are multiple events with all the keys usuable, the events list will be incorrect.

ddorwin:
@mwatson2, was your last comment just explaining the logic of this change? I believe that is what it does, but I want to make sure you don't have any concerns before merging.

mwatson2:
@ddorwin Actually, I did have a concern, but I'll raise a separate issue and we can discuss.

PR #3748, opened 2016-09-18
Expand on the Touch constructor tests

foolip:
Chrome currently fails the new Touch({ identifier: 0, target: null })
test, while Safari fails even more, allowing even new Touch().

### DIFF
--- a/touch-events/touch-touchevent-constructor.html
+++ b/touch-events/touch-touchevent-constructor.html
@@ -10,9 +10,38 @@
 <div id="target0"></div>
 <script>
 test(function() {
-    assert_throws(new TypeError(), function() {new Touch();}, "Touch constructor requires initialize dictionary");
-    assert_throws(new TypeError(), function() {new Touch({});}, "Touch constructor requires identifier and target");
-}, "Create a Touch object with insufficient properties");
+    var testIdentifier = 0;
+    var testTarget = document.getElementById('target0');
+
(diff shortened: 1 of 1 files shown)


foolip:
@RByers for review, FYI @cdumez for incoming WebKit failure.

It's not a given to me that this is the best path, but this is what the spec currently says. There's still `document.createTouch` where all arguments are effectively optional and nullable, so just it's still a lie that `Touch`'s `target` is non-nullable.

Given the behavior of Safari on iOS it might be less overall work to just make `TouchInit` not have any required members and just (continue to) handle a null target internally. WDYT?

foolip:
I did an HTTP Archive search for 'new Touch' and based on that I'm pretty sure the constructor won't be a problem, just 251 hits, not a single instance of 'new Touch({' and none of the remaining looks like actual usage of the Touch constructor but other things using the same name.

Still, the `document.createTouch` problem remains, and there are so many hits for '.createTouch(' in HTTP Archive that the query fails. Not worth the time to analyze IMHO, will file a spec bug instead :)

foolip:
I found https://github.com/w3c/touch-events/issues/58 and was about to file a new issue to make the Touch constructor maximally permissive, but then I hesitated. For events I'd argue that it's sad to make things nullable if it's actually not required for web compat, so I'll at least measure a bit about `document.createTouch` first.

RByers (approved):
LGTM, thatks!  Sorry for the delay, missed this while at TPAC.

PR #3749, opened 2016-09-18
Test the BlobEvent constructor

foolip:
This is for the new BlobEvent("type", { data: null }) test, which fails
in Chromium, see https://crbug.com/647693

### DIFF
--- a/mediacapture-record/BlobEvent-constructor.html
+++ b/mediacapture-record/BlobEvent-constructor.html
@@ -1,0 +1,38 @@
+<!doctype html>
+<title>BlobEvent constructor</title>
+<link rel="help" href="https://w3c.github.io/mediacapture-record/MediaRecorder.html#blob-event">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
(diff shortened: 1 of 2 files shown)


foolip:
@miguelao, can you review and own the tests for MediaStream Recording? This is redundant with parts of Blink's LayoutTests/fast/mediarecorder/BlobEvent-basic.html, and when I fix the bug I'll try to remember to delete the overlap.

foolip:
Note that the test fails in Firefox too, they have the init dict as optional.

yell0wd0g (approved):
This test as-is and the Ownership LGTM.

I wonder if we should add a test a la [1], where a Blob is initialized out of a Uint8Array, then the BlobEvent.data is read and compared?

[1] https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/fast/mediarecorder/BlobEvent-basic.html

foolip:
I believe a test of that nature is redundant with checking that the blob object itself is the same. One could of course check that the constructor doesn't *modify* the blob object, but it seems a bit far fetched, and doesn't look like the real intent of the test.

PR #3754, opened 2016-09-19
Use temporary licenses

mwatson2:


encrypted-media/util/drm-messagehandler.js
```diff
@@ -135,7 +135,7 @@ MessageHandler.prototype.messagehandler = function messagehandler( messageType,
             getCustomHeaders : function( drmconfig, sessionType, messageType ) {
                 var customHeader = {    merchant: drmconfig.merchant,
                                         userId: "purchase",
-                                        sessionId: "a0d0f0p" + ( sessionType === 'persistent-license' ) ? "1" : "0" };
```
ddorwin: This resulted in `sessionId: "1"`, right? Does that result in a persistent session with the correct keys? That seems unexpected for the server side.

ddorwin (commented):
This change LGTM, but I'm concerned about the apparent server behavior.

PR #3755, opened 2016-09-19
Request audio instead of video for MediaStreamTrackEvent test

foolip:
It turns out that when testing in BrowserStack, requesting video fails
in Edge and Firefox, but requesting audio works. It doesn't matter what
kind of MediaStreamTrack is used, so go with audio.

Follow-up to 37153a1fe0517e79074e70e47d4067f879b9b2d4.

### DIFF
--- a/mediacapture-streams/MediaStreamTrackEvent-constructor.https.html
+++ b/mediacapture-streams/MediaStreamTrackEvent-constructor.https.html
@@ -31,7 +31,7 @@ test(function() {
 
 // a MediaStreamTrack instance is needed to test, any instance will do.
 promise_test(function() {
-  return navigator.mediaDevices.getUserMedia({ video: true })
+  return navigator.mediaDevices.getUserMedia({ audio: true })
     .then(function(stream) {
       var track = stream.getTracks()[0];
       assert_true(track instanceof MediaStreamTrack);


alvestrand:
Do you know why BrowserStack has this problem? Do they connect audio input to some null-thing and don't provide a video device, or is it something else?

I'm very hesitant to change tests to work around bugs in specific implementations.
The reason why tests generally work with "video: true" is that if you connect a stream to some output tag, using audio in loopback will strongly annoy my office-mates, while using video in loopback just works.

One alternative is to generate a stream from a canvas, which should be guaranteed to work no matter what the device situation is.

dontcallmedom:
another alternative is to generate a `MediaStream` from the Web Audio API.

foolip:
I guess there's just no camera in these VMs, but there is an audio device, which is true of my physical machine as well.

I could use `canvas.captureStream()`, but that'll give a `CanvasCaptureMediaStreamTrack` object. I had just figured out what @dontcallmedom already knew:
```JavaScript
var context = new AudioContext();
var dest = context.createMediaStreamDestination();
var stream = dest.stream;
```

foolip:
I'll file a new PR using Web Audio.

foolip:
https://github.com/w3c/web-platform-tests/pull/3780

PR #3761, opened 2016-09-20
[eme] Fix #3435: Expect TypeError instead of InvalidAccessError

### LINKED ISSUE #3435: [eme] generate-request-disallowed-input.js assumes InvalidAccessError instead of TypeError on initData that fails sanitization
The spec for [MediaKeySession.generateRequest()](https://w3c.github.io/encrypted-media/#dom-mediakeysession-generaterequest) says that initData which is invalid or which fails sanitization should result in the promise returned by MediaKeySession.generateRequest() being rejected with a TypeError, but test [Google/encrypted-media-generate-request-disallowed-input.html](https://w3c-test.org/encrypted-media/Google/encrypted-media-generate-request-disallowed-input.html) is testing that we're throwing an InvalidAccessError.

The test should be checking for TypeError, not InvalidAccessError.

ddorwin:


ddorwin:
The first commit is PR #3759. Just look at the second commit.

mwatson2:
LGTM

PR #3773, opened 2016-09-20
[eme] Add polyfill / patch for Edge keystatuses bugs

mwatson2:


encrypted-media/polyfill/edge-keystatuses.js
```diff
@@ -0,0 +1,144 @@
...
+        // The following function is the core of this JS patch. The rest of this file is infrastructure
+        // required to enable this function
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        function _proxyKeyStatusesChange( event ) {
+            this._keyStatuses.clear();
+            var keyStatuses = [];
+            this._session.keyStatuses.forEach( function( keyId, status ) {
+                var newKeyId = new Uint8Array( keyId );
+
+                function swap( arr, a, b ) { var t = arr[a]; arr[a] = arr[b]; arr[b] = t; }
```
jdsmith3000: This section reorders the returned keys.  Makes sense.

encrypted-media/polyfill/edge-keystatuses.js
```diff
@@ -0,0 +1,144 @@
...
+                swap( newKeyId, 0, 3 );
+                swap( newKeyId, 1, 2 );
+                swap( newKeyId, 4, 5 );
+                swap( newKeyId, 6, 7 );
+
+                keyStatuses.push( { key: newKeyId, status: status, ord: arrayBufferAsString( newKeyId ) } );
+            });
+
+            function lexicographical( a, b ) { return a < b ? -1 : a === b ? 0 : +1; }
```
jdsmith3000: Can you briefly explain what this section is intended to do?

encrypted-media/polyfill/edge-keystatuses.js
mwatson2: Actually, the `swap` function swaps bytes within one of the returned key ids.

encrypted-media/polyfill/edge-keystatuses.js
mwatson2: The whole section, or a specific part ?

The whole section iterates through the key ids exposed in the native keyStatuses map. For each key id, we make a copy, fix byte order within the key id and add the new key id and its status for the local `keyStatuses` var.

Then we sort that array in lexicographical order of (fixed) key id and add the keys to the `_keyStatuses` map on our proxy object. The iteration order of a JS Map is the insertion order, so this ensures we get the correct iteration order.

jdsmith3000: Thanks.  This seems clear to me.

jdsmith3000:
LGTM

encrypted-media/polyfill/edge-keystatuses.js
```diff
@@ -0,0 +1,144 @@
+(function() {
+
+    // This polyfill fixes the following problems with Edge browser
+    // (1) Various maplike methods for keystatuses are not supported or suported incorrectly
```
AlexeyKor: nit: suported -> supported

encrypted-media/scripts/keystatuses.js
```diff
@@ -49,10 +49,11 @@ function runTest(config,qualifier)
...
             function lexicographicalkey( a, b ) { return lexicographical( a.key, b.key ); }
-            var expected = [{ key: key1String, value: 'usable'}, { key: key2String, value: 'usable'}].sort( lexicographicalkey );
-            assert_equals(JSON.stringify(result),
-                          JSON.stringify(expected),
-                                 "keystatuses should have the two expected keys with keystatus 'usable'");
+            var expected1 = [{ key: key1String, value: 'usable'}, { key: key2String, value: 'usable'}].sort( lexicographicalkey );
+            var expected2 = [{ key: key1String, value: 'status-pending'}, { key: key2String, value: 'status-pending'}].sort( lexicographicalkey );
+            assert_in_array(    JSON.stringify(result),
+                                [ JSON.stringify(expected1),JSON.stringify(expected2) ],
+                                 "keystatuses should have the two expected keys with keystatus 'usable' or 'status-pending'");
```
AlexeyKor: Is there a way to understand if status should be exactly 'usable' or exactly 'status-pending'?

AlexeyKor (commented):
All items noted.

encrypted-media/scripts/keystatuses.js
mwatson2: Not exactly. The specification constraint is that keystatuseschanged should be fired when one or more keystatuses do change, that the keyStatus values eventually become something other than "status-pending" and that they do not return to "status-pending" once they have moved to some other status.

So, the test (which is not the subject of the PR, btw), _could_ be more exacting in checking these constraints.

sideshowbarker:
Does this polylfill cause Edge to appear to pass spec-conformance tests that it’s otherwise failing?

jgraham:
Please revert this, polyfills should not be checked in to the testsuite.

jdsmith3000:
Using test suite polyfills is an ongoing discussion in the HTML Media Extensions WG.  They are being used short term to assess the impact of known implementation gaps on meeting specific test requirements, and are not planned to be a permanent part of the EME test suite.

mwatson2:
@sideshowbarker No, these polyfills do not affect results for the basic test suite. We can run a separate copy of the entire test suite using the polyfills, which allows us to demonstrate what the results would be if the bugs fixed by the polyfills were fixed by implementations instead.

Polyfill might not be the correct term. For the most part these work around minor bugs, rather that being complete implementations of the feature.

PR #3782, opened 2016-09-21
Upstream a test for CustomElementRegistry interface from WebKit

rniwa:


rniwa:
@kojiishi

custom-elements/CustomElementRegistry.html
```diff
@@ -0,0 +1,580 @@
...
+        'customElements.define must throw a TypeError when the element interface is a string');
+    assert_throws({'name': 'TypeError'}, function () { customElements.define('badname', {}); },
+        'customElements.define must throw a TypeError when the element interface is an object');
+    assert_throws({'name': 'TypeError'}, function () { customElements.define('badname', []); },
+        'customElements.define must throw a TypeError when the element interface is an array');
+}, 'customElements.define must throw when the element interface is not a constructor');
+
+test(function () {
+    customElements.define('custom-html-element', HTMLElement);
+}, 'customElements.define must not throw the constructor is HTMLElement');
```
dominiccooney: Grammar: throw the constructor -> throw if the constructor

custom-elements/CustomElementRegistry.html
```diff
@@ -0,0 +1,580 @@
...
+test(function () {
+    class SomeCustomElement extends HTMLElement {};
+
+    var calls = [];
+    var OtherCustomElement = new Proxy(class extends HTMLElement {}, {
+        get: function (target, name) {
+            calls.push(name);
+            return target[name];
+        }
+    })
```
dominiccooney: Maybe trailing semi here to not rely on automatic insertion?

custom-elements/CustomElementRegistry.html
```diff
@@ -0,0 +1,580 @@
...
+
+    assert_false(unresolvedElement instanceof ConstructorCallsDefine);
+    assert_false(elementUpgradedDuringUpgrade instanceof DefinedDuringUpgrade);
+
+    customElements.define('constructor-calls-define', ConstructorCallsDefine);
+}, 'customElements.define unset the element definition is running flag before upgrading custom elements');
+
+(function () {
+    var testCase = async_test('customElements.define must not throw'
+        +' when defining another custom element in a different global object during Get(constructor, "prototype")', {timeout: 100});
```
dominiccooney: Style: maybe space between + and '

domenic:
It's worth noting we have some tests already in https://github.com/w3c/web-platform-tests/blob/master/custom-elements/custom-element-registry/define.html. I'm fine with duplicate coverage (more tests better!), but we might want to consider the folder structure a bit more closely. E.g. maybe this should be split up into define-2.html, whenDefined.html, etc.? It's more important to land though so don't let that stop you.

dominiccooney:
We have evolved defined.html downstream so it might be easier to upstream them as-is and do the merging upstream.

sideshowbarker:
> We have evolved defined.html downstream so it might be easier to upstream them as-is and do the merging upstream.

OK, will merge this to master as-is then

PR #3783, opened 2016-09-21
Import a test for HTMLElement constructor from WebKit

rniwa:


rniwa:
@dominiccooney @kojiishi

PR #3784, opened 2016-09-21
Test that the manifest script works.

Ms2ger:


Ms2ger:
Unfortunately, it doesn't work. @gsnedders, if you feel like figuring out a fix... :)

gsnedders:
This is https://github.com/w3c/web-platform-tests/issues/3775 for which https://github.com/w3c/wpt-tools/pull/112 is the fix.

PR #3786, opened 2016-09-21
Add a test for MouseEvent.prototype.offsetX and offsetY.

rniwa:
They should return the relative offset against the padding edge of the relative target
after the event retargetting algorithm had run.

Both Chrome and WebKit passes this test.

### DIFF
--- a/shadow-dom/MouseEvent-prototype-offsetX-offsetY.html
+++ b/shadow-dom/MouseEvent-prototype-offsetX-offsetY.html
@@ -114,7 +114,7 @@ test(function () {
     var container = shadowRoot.querySelector('#container');
 
     var eventLogs = attachLoggers([target, container, shadowRoot, host, document.body]);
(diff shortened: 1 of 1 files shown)


shadow-dom/MouseEvent-prototype-offsetX-offsetY.html
```diff
@@ -0,0 +1,154 @@
...
+
+test(function () {
+    shadowRoot.innerHTML = '<style>' + shadowStyle + '</style><div id="container"><slot></slot></div>';
+    host.innerHTML = '<style>' + shadowStyle + '</style><div id="target">Click here</div>';
+
+    var target = host.querySelector('#target');
+    var container = shadowRoot.querySelector('#container');
+
+    var eventLogs = attachLoggers([target, container, shadowRoot, host, document.body]);
+    var mouseEvent = new MouseEvent('mousedown', {clientX: 51, clientY: 37, scoped: false, bubbles: true});
```
TakayoshiKochi: scoped is now "composed" and the meaning is negated.
(you changed other places, forgot this one?)

PR #3788, opened 2016-09-21
Add a test for custom events in a document without a browsing context

matthewp:
This adds a test for dispatching custom events on documents created with
`document.implementation.createHTMLDocument`. This test passes in Chrome
53, Firefox 49, and fails in Safari 9.1.2.


Ms2ger:
Thanks!

PR #3791, opened 2016-09-21
Fix html/semantics/forms/the-meter-element/meter.html

cdumez:
Fix html/semantics/forms/the-meter-element/meter.html as it expected that
assigning "foobar" to meter.value (and other attributes) would be identical to
assigning to 0.

meter attributes are of type double and toNumber("foobar") will return NaN.
Therefore, we should throw a TypeError.

I verified that Firefox and Chrome throw a TypeError.


cdumez:
Oh, I'll add tests for low and max as well.

cdumez:
Ok, I updated the test to cover other attributes as well, not just 'value'.

cdumez:
@sideshowbarker @Ms2ger ping review?

zcorpan (approved):

PR #3794, opened 2016-09-21
Assert pressure is 0 during pointerup

appsforartists:
@NavidZ, can you please help me test that these are correct?  I tried using PEP, but kept getting bit by https://github.com/jquery/PEP/issues/314

### DIFF
--- a/pointerevents/pointerevent_constructor.html
+++ b/pointerevents/pointerevent_constructor.html
@@ -30,6 +30,8 @@
             var testHeight = 5;
             var testTiltX = -45;
             var testTiltY = 30;
+            var testButton = 0;
+            var testButtons = 1;
             var testPressure = 0.4;
             var testIsPrimary = true;
 
@@ -80,6 +82,8 @@
                 clientY: testClientY,
                 tiltX: testTiltX,
                 tiltY: testTiltY,
+                button: testButton,
+                buttons: testButtons,
                 pressure: testPressure,
                 isPrimary: testIsPrimary
                 });
--- a/pointerevents/pointerevent_constructor.html
+++ b/pointerevents/pointerevent_constructor.html
@@ -42,6 +42,8 @@
                     ["custom cancelable", event.cancelable, testCancelable],
                     ["custom pointerId", event.pointerId, testPointerId],
                     ["custom pointerType", event.pointerType, testPointerType],
+                    ["custom button", event.button, testButton],
+                    ["custom buttons", event.buttons, testButtons],
                     ["custom width", event.width, testWidth],
                     ["custom height", event.height, testHeight],
                     ["custom clientX", event.clientX, testClientX],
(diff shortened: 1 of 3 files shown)


NavidZ:
I don't think jquery polyfil is passing all the tests at the moment. @scottgonzalez might be able to help you with that. But looking at the code it does seem alright overall. Is there something particular you want me to test or something?

pointerevents/pointerevent_constructor.html
```diff
@@ -30,6 +30,8 @@
             var testHeight = 5;
             var testTiltX = -45;
             var testTiltY = 30;
+            var testButton = 0;
+            var testButtons = 1;
```
NavidZ: Although we have not setting this before and that is fine to set them this way. But there should be no expectation that when js itself creates a PointerEvent object it conforms with the spec. Meaning that it can have buttons=25 and pressure=2 as long as they fit in the data type.

appsforartists: The [PR I've opened](https://github.com/jquery/PEP/pull/313/commits/f0970758abc7be5817fb1a94ca67afb15a12b924) to support https://github.com/w3c/pointerevents/issues/146 in PEP sets pressure to 0 when buttons is 0.  Thus, if someone tried to create a `PointerEvent` with `pressure` set but not `buttons`, pressure would be 0.

This seems reasonable.    `pressure` without `buttons` would be nonsense, and allowing a constructor implementation to validate those parameters makes it really easy to assert correctness.  (Look at how simple [the PEP PR](https://github.com/jquery/PEP/pull/313/commits/f0970758abc7be5817fb1a94ca67afb15a12b924) is).

If allowing `PointerEvent` arguments to be nonsensical is a requirement, that should be its own test.

appsforartists:
You mentioned you'd be down to be a reviewer, so I'd mostly like advice for changes and/or and LGTM and a merge.  😃

pointerevents/pointerevent_constructor.html
NavidZ: From my understanding and from the discussions we had so far the answer is yes. In other words, any combination of the field values are possible when js creates an untrusted event. But I'd like someone else to confirm this as well. Maybe @mustaqahmed @RByers @scottgonzalez can chime in here too.

scottgonzalez: I don't have a strong opinion either way, but developers shouldn't be creating events with values that can't exist for real.

appsforartists: I also think there's a difference between "they technically can because an implementation doesn't validate" and "the spec requires that nonsense be passed-through unmolested."  

I'm not asking that the spec require validation in the constructor, but I do think this PR should be merged.  If we wanna specify that `pressure` is always passed-through in the constructor, even if `buttons` is unset, that should be a separate conversation (and a separate test).

NavidZ: LGTM. I do feel the same about this PR that it can go in without further changes.
My first comment about this part was just something to bring our attentions to this subject.

RByers: Yes, constructors let you specify any value permitted by the interface even if such events aren't created by the browser in practice.

RByers:
LGTM

PR #3796, opened 2016-09-22
Fix html/semantics/forms/the-input-element/checkbox.html

cdumez:
The test was complaining about 'e' being undefined.


jdm:
Thanks!

PR #3798, opened 2016-09-22
[eme] Add test for expiration attribute value

mwatson2:
This is a breaking change, in that it is necessary to switch the DRMToday server to "Upfront Authentication Mode" before this test will execute (this change takes several minutes to take effect).

As a result, when we merge this change we need to synchronize all users of the repo and change the server configuration.


encrypted-media/drm-expiration.html
```diff
@@ -0,0 +1,72 @@
...
+
+  </head>
+  <body>
+    <div id='log'></div>
+
+    <script>
+        var keysystem = getSupportedKeySystem();
+
+        getSupportedContentAndInitDataTypes( keysystem )
+        .then( function( supported )
```
AlexeyKor: nit: Should it be on the same line with getSupportedContentAndInitDataTypes?

AlexeyKor (approved):
Everything except spacing inconsistency LGTM.

jdsmith3000:
Should be ready to merge.

PR #3800, opened 2016-09-22
Add tests for historical SVG features that should be removed

foolip:


foolip:
These are all from reading https://svgwg.org/svg2-draft/changes.html with the exception of `requiredFeatures` and `hasExtension` which aren't mentioned but I happened to know about. In other words, there are likely more things missing.

svg/historical.html
```diff
@@ -0,0 +1,74 @@
...
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+var removedInterfaces = [
+  "SVGAnimatedPathData",
+  "SVGColor",
+  "SVGDocument",
+  "SVGElementInstanceList",
+  "SVGExternalResourcesRequired",
+  "SVGFilterPrimitiveStandardAttributes",
```
heycam: This should be in `mixinInterfaces`: https://drafts.fxtf.org/filters/#InterfaceSVGFilterPrimitiveStandardAttributes

svg/historical.html
```diff
@@ -0,0 +1,74 @@
...
+  "SVGICCColor",
+  "SVGLangSpace",
+  "SVGPaint",
+  "SVGPathSeg",
+  "SVGStylable",
```
heycam: Add SVGLocatable too?

svg/historical.html
```diff
@@ -0,0 +1,74 @@
...
+
+var removedMembers = {
+  "SVGElement": [
+    "getPresentationAttribute",
+    "xmlbase",
+    "xmllang",
+    "xmlspace"
+  ],
+  "SVGGraphicsElement": [
+    "getStrokeBBox",
```
heycam: Is it worth including this one?  It wasn't in SVG 1.1, just an earlier draft of SVG 2.

svg/historical.html
```diff
@@ -0,0 +1,74 @@
...
+    "getTransformToElement",
+    "hasExtension",
+    "requiredFeatures"
+  ],
+  "SVGPathElement": [
+    "getPointAtLength",
+    "getTotalLength",
+    "pathLength"
```
heycam: These three got moved up to `SVGGeometryElement`, so we shouldn't test that these have been removed from `SVGPathElement` unless you change the test to use `Object.hasOwnProperty` instead of `in`.

heycam (changes requested):

svg/historical.html
foolip: Thanks, this isn't the first time I've wondered where something went and the answer was that spec :)

svg/historical.html
foolip: Ah, "Turned SVGLocatable and SVGTransformable into SVGGraphicsElement, which directly inherits from SVGElement" not found when searching for "remove" :) I can't find in SVGTransformable https://www.w3.org/TR/SVG/single-page.html so I guess I'll omit that.

svg/historical.html
foolip: It was briefly in Blink, but not more, so testing for it won't be of any value.

svg/historical.html
foolip: Oh, didn't pay attention to which direction in the prototype chain they moved. Split into separate tests.

svg/historical.html
```diff
@@ -9,8 +9,8 @@
   "SVGDocument",
   "SVGElementInstanceList",
   "SVGExternalResourcesRequired",
-  "SVGFilterPrimitiveStandardAttributes",
   "SVGICCColor",
+  "SVGLocatable",
```
heycam: Nit: one line lower to keep the list sorted.

heycam (approved):
r=me with that, thanks.

PR #3802, opened 2016-09-22
DOM selector tests wrongly assume '>>' is invalid in selectors

cdumez:
DOM selector tests wrongly assume '>>' is invalid in selectors. However,
as per Selectors Level 4, this is a valid descendant combinator,
identical in behavior to ' ' combinator:
- https://drafts.csswg.org/selectors-4/#descendant-combinators

### DIFF
--- a/dom/nodes/selectors.js
+++ b/dom/nodes/selectors.js
@@ -27,7 +27,6 @@ var invalidSelectors = [
   {name: "Invalid class",                selector: ".foo..quux"},
   {name: "Invalid class",                selector: ".bar."},
   {name: "Invalid combinator",           selector: "div & address, p"},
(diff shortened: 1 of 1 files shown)


cdumez:
@rniwa

Ms2ger:
Is there consensus to add this seemingly redundant feature?

cdumez:
@Ms2ger I am not very familiar with the Selector specification. What I do know is that the '>>' is part of the specification at the moment and has been shipping in Safari for a while.

I think at the very least, we should drop the test that checks that we throw as this hurts browsers like Safari that implement the latest specification. I thought of doing this at first but reviewers usually do not like reducing test coverage.

zcorpan (approved):
LGTM. Please discuss the be or not to be about the >> combinator with the csswg.

PR #3803, opened 2016-09-22
Fix XPathEvaluator constructor test

cdumez:
Fix XPathEvaluator constructor test as it was expecting that the constructor
can be called without 'new'. Firefox / Chrome and Safari all do not allow
this.


cdumez:
@Ms2ger

Ms2ger:
Please instead change the test to expect an exception, rather than removing coverage altogether.

cdumez:
@Ms2ger I updated the PR as suggested.

PR #3807, opened 2016-09-23
added test that verifies proper basic auth caching behavior

Ms2ger:


Ms2ger:
Code reviewed upstream.

negibokken:
Hi @Ms2ger @gilbertw1, could you tell me the upstream of this PR and the related specs, if you remember?

I'm trying to fix Chromium behavior for this WPT (current: https://github.com/web-platform-tests/wpt/blob/ee42dc176d02bc8b6bc8d583ed9fb2495db42a4b/fetch/http-cache/basic-auth-cache-test.html ). 

This test

1. requests img with userinfo (`testuser:password`)
2. requests img **without userinfo** and check if the browser can use cached img

But in my understanding, userinfo is an authenticated request (though userinfo is deprecated in [RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-deprecation-of-userinfo-in-)) and it may not be used a cached response according to the [section 3.5 in RFC9111](https://www.rfc-editor.org/rfc/rfc9111.html#name-storing-responses-to-authen). Although section 3.5 in RFC9111 says only about the `Authorization` header, I think it's almost the same.

So I want to know the related specs when the test was written. And I want to judge which should be fixed the Chromium behavior or this WPT.

Thanks!

negibokken:
@Ms2ger Thank you for the reference! Now, I understand why the WPT is added by reading https://github.com/servo/servo/pull/13281 and https://github.com/servo/servo/issues/12095.

If not cached, many basic auth prompts are shown when the website contains the same resources. Hmm...

I'll think about this more. I appreciate your kindness!

PR #3808, opened 2016-09-23
removed race condition possibility from auth cache test & fixed up test errors

Ms2ger:


Ms2ger:
Code reviewed upstream.

PR #3811, opened 2016-09-23
Add crossorigin attribute and implement step 14 of prepare a script

Ms2ger:
Add WPT test for HTMLScriptElement crossOrigin IDL attribute


Ms2ger:
Code reviewed upstream.

PR #3820, opened 2016-09-24
Change html/syntax/serializing-html-fragments/serializing.html for <p…

zcorpan:
…re>\n

See
https://github.com/whatwg/html/pull/1815
https://github.com/w3c/web-platform-tests/pull/3814

Also change .replace() to avoid non-standard "g" as third argument.


cdumez:
r=me

PR #3830, opened 2016-09-27
Test that requestFullscreen() works for svg:svg but not svg:rect

foolip:
https://fullscreen.spec.whatwg.org/#dom-element-requestfullscreen

### DIFF
--- a/fullscreen/api/element-request-fullscreen-svg-svg-manual.html
+++ b/fullscreen/api/element-request-fullscreen-svg-svg-manual.html
@@ -1,0 +1,17 @@
+<!DOCTYPE html>
+<title>Element.requestFullscreen() for SVG svg element</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="../trusted-click.js"></script>
+<div id="log"></div>
+<svg></svg>
+<script>
+async_test(function(t)
+{
+    var svg = document.querySelector("svg");
+    assert_true(svg instanceof SVGSVGElement);
+    trusted_request(svg, document.body);
+    document.onfullscreenchange = t.step_func_done();
+    document.onfullscreenerror = t.unreached_func("fullscreenerror event");
+});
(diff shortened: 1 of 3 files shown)


foolip:
@upsuper PTAL. element-request-fullscreen-svg-rect-manual.html is failing in Firefox and in Chrome, so you might want to file a bug for that.

upsuper:
Thanks. Filed [bug 1305928](https://bugzilla.mozilla.org/show_bug.cgi?id=1305928) for that.

foolip:
Taking that as review and merging, please shout if you didn't actually review and find something wrong after the fact.

upsuper (commented):
It overall looks fine to me.

fullscreen/api/element-request-fullscreen-svg-svg-manual.html
```diff
@@ -0,0 +1,17 @@
...
+<script src="../trusted-click.js"></script>
+<div id="log"></div>
+<svg></svg>
+<script>
+async_test(function(t)
+{
+    var svg = document.querySelector("svg");
+    assert_true(svg instanceof SVGSVGElement);
+    trusted_request(svg, document.body);
+    document.onfullscreenchange = t.step_func_done();
```
foolip: I think this test can only call `t.step_func_done()` once, where would the second time be?

In testharness.js, nothing that happens after the first failure or done event matters, so in general it's fine if there are two ways the test can pass and if both happen.

upsuper: The second time would be when exiting fullscreen, I think. If anything after first done doesn't count, then probably fine.

foolip: Ah, the test actually never exits fullscreen, which I guess is annoying when testing manually. If it did it'd still not be necessary to clean up this event handler to avoid the extra call to `t.step_func_done()` though.

PR #3832, opened 2016-09-27
Fix #3706: Fail tests with TIMEOUT when DRM severs fail

### LINKED ISSUE #3706: [eme] Explicitly surface DRM test server errors
Problems with the DRM test server caused many tests to fail in a non-obvious way ([example](https://lists.w3.org/Archives/Public/public-html-media/2016Sep/0065.html)).

We should try to make such failures obvious in some way, at least by including it in the test page, but preferably in some unmistakable way in the reporting.

mwatson2:
Also fix a number of minor regressions and whitespace consistency.

### DIFF
--- a/encrypted-media/clearkey-events.html
+++ b/encrypted-media/clearkey-events.html
@@ -31,10 +31,11 @@
         getSupportedContent( 'org.w3.clearkey' )
         .then( function( contents )
         {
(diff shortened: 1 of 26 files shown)


jdsmith3000:
There's a lot here, but it's mostly formatting and white space changes.  The change for license server errors looks good.

LGTM.

PR #3835, opened 2016-09-27
Import tests for custom element reactions from WebKit.

rniwa:
Both WebKit and Chrome implementations pass the tests except Chrome fails some tests because it doesn't support `new Document`.

### DIFF
--- a/custom-elements/adopted-callback.html
+++ b/custom-elements/adopted-callback.html
@@ -1,0 +1,135 @@
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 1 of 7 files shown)


TakayoshiKochi:
Let's get this merged.
`new Document` is being discussed for Blink, but not a blocker for merging this.

TakayoshiKochi:
FYI Blink pulls WPT in layout test, and runs them as local files - which means XHR doesn't work.
A plan seems to exist to run them under http server, though - so we will have to skip this test in layout test framework and run manually.

PR #3836, opened 2016-09-28
Add remaining tests of CEReactions for the DOM specification.

rniwa:
Now that we have the basic testing infrastructure for CEReactions, add tests for the remaining interfaces in https://dom.spec.whatwg.org/

Most of tests pass on Chrome, and we intend to fix the behavior in WebKit.

### DIFF
--- a/custom-elements/reactions/DOMTokenList.html
+++ b/custom-elements/reactions/DOMTokenList.html
@@ -1,0 +1,223 @@
+<!DOCTYPE html>
+<html>
+<head>
+<title>Custom Elements: CEReactions on DOMTokenList interface</title>
+<meta name="author" title="Ryosuke Niwa" href="mailto:rniwa@webkit.org">
+<meta name="assert" content="add, remove, toggle, replace, and the stringifier of DOMTokenList interface must have CEReactions">
+<meta name="help" content="https://dom.spec.whatwg.org/#node">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="./resources/reactions.js"></script>
+</head>
+<body>
+<div id="log"></div>
+<script>
+
+test(function () {
+    var element = defineNewCustomElement(['class']);
+    var instance = document.createElement(element.name);
+    assert_array_equals(element.log().types(), ['constructed']);
+    instance.classList.add('foo');
+    var logEntries = element.log();
+    assert_array_equals(logEntries.types(), ['attributeChanged']);
+    assert_equals(logEntries.log().name, 'class');
+    assert_equals(logEntries.log().oldValue, null);
+    assert_equals(logEntries.log().newValue, 'foo');
+    assert_equals(logEntries.log().namespace, null);
+}, 'add on DOMTokenList must enqueue a attributeChanged reaction when adding an attribute');
+
+test(function () {
+    var element = defineNewCustomElement([]);
+    var instance = document.createElement(element.name);
+    assert_array_equals(element.log().types(), ['constructed']);
+    instance.classList.add('foo');
+    assert_array_equals(element.log().types(), []);
+}, 'add on DOMTokenList must not enqueue a attributeChanged reaction when adding an unobserved attribute');
+
+test(function () {
+    var element = defineNewCustomElement(['class']);
+    var instance = document.createElement(element.name);
+    instance.setAttribute('class', 'hello');
+    assert_array_equals(element.log().types(), ['constructed', 'attributeChanged']);
+    instance.classList.add('world');
+    var logEntries = element.log();
+    assert_array_equals(logEntries.types(), ['attributeChanged']);
+    assert_equals(logEntries.log().name, 'class');
+    assert_equals(logEntries.log().oldValue, 'hello');
+    assert_equals(logEntries.log().newValue, 'hello world');
+    assert_equals(logEntries.log().namespace, null);
+}, 'add on DOMTokenList must enqueue a attributeChanged reaction when adding a value to an existing attribute');
+
+test(function () {
+    var element = defineNewCustomElement([]);
+    var instance = document.createElement(element.name);
+    instance.setAttribute('class', 'hello');
+    assert_array_equals(element.log().types(), ['constructed']);
+    instance.classList.add('world');
+    assert_array_equals(element.log().types(), []);
+}, 'add on DOMTokenList must enqueue a attributeChanged reaction when adding a value to an existing unobserved attribute');
+
+test(function () {
+    var element = defineNewCustomElement(['class']);
+    var instance = document.createElement(element.name);
+    instance.setAttribute('class', 'hello world');
+    assert_array_equals(element.log().types(), ['constructed', 'attributeChanged']);
+    instance.classList.remove('world');
+    var logEntries = element.log();
+    assert_array_equals(logEntries.types(), ['attributeChanged']);
+    assert_equals(logEntries.log().name, 'class');
+    assert_equals(logEntries.log().oldValue, 'hello world');
+    assert_equals(logEntries.log().newValue, 'hello');
+    assert_equals(logEntries.log().namespace, null);
+}, 'remove on DOMTokenList must enqueue a attributeChanged reaction when removing a value from an attribute');
+
+test(function () {
+    var element = defineNewCustomElement(['class']);
+    var instance = document.createElement(element.name);
+    instance.setAttribute('class', 'hello world');
+    assert_array_equals(element.log().types(), ['constructed', 'attributeChanged']);
+    instance.classList.remove('foo');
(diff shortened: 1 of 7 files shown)


custom-elements/reactions/DOMTokenList.html
```diff
@@ -0,0 +1,223 @@
...
+<title>Custom Elements: CEReactions on DOMTokenList interface</title>
+<meta name="author" title="Ryosuke Niwa" href="mailto:rniwa@webkit.org">
+<meta name="assert" content="add, remove, toggle, replace, and the stringifier of DOMTokenList interface must have CEReactions">
+<meta name="help" content="https://dom.spec.whatwg.org/#node">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="./resources/reactions.js"></script>
+</head>
+<body>
+<div id="log"></div>
```
dominiccooney: Is this div necessary?

custom-elements/reactions/DOMTokenList.html
```diff
@@ -0,0 +1,223 @@
...
+    var instance = document.createElement(element.name);
+    assert_array_equals(element.log().types(), ['constructed']);
+    instance.classList.add('foo');
+    var logEntries = element.log();
+    assert_array_equals(logEntries.types(), ['attributeChanged']);
+    assert_equals(logEntries.log().name, 'class');
+    assert_equals(logEntries.log().oldValue, null);
+    assert_equals(logEntries.log().newValue, 'foo');
+    assert_equals(logEntries.log().namespace, null);
+}, 'add on DOMTokenList must enqueue a attributeChanged reaction when adding an attribute');
```
dominiccooney: Minor grammar point, but maybe "a attributeChanged" &rarr; "an attributeChanged"; below too

custom-elements/reactions/DOMTokenList.html
```diff
@@ -0,0 +1,223 @@
...
+}, 'add on DOMTokenList must enqueue a attributeChanged reaction when adding a value to an existing attribute');
+
+test(function () {
+    var element = defineNewCustomElement([]);
+    var instance = document.createElement(element.name);
+    instance.setAttribute('class', 'hello');
+    assert_array_equals(element.log().types(), ['constructed']);
+    instance.classList.add('world');
+    assert_array_equals(element.log().types(), []);
+}, 'add on DOMTokenList must enqueue a attributeChanged reaction when adding a value to an existing unobserved attribute');
```
dominiccooney: It must *not* enqueue an attributeChanged reaction. The test looks right but the description is confusing.

custom-elements/reactions/DOMTokenList.html
```diff
@@ -0,0 +1,223 @@
...
+}, 'remove on DOMTokenList must not enqueue a attributeChanged reaction when removing a non-existent value from an attribute');
+
+test(function () {
+    var element = defineNewCustomElement([]);
+    var instance = document.createElement(element.name);
+    instance.setAttribute('class', 'hello world');
+    assert_array_equals(element.log().types(), ['constructed']);
+    instance.classList.remove('world');
+    assert_array_equals(element.log().types(), []);
+}, 'remove on DOMTokenList not must enqueue a attributeChanged reaction when removing a value from an unobserved attribute');
```
dominiccooney: not must &rarr; must not

There's one other instance of this below, too.

custom-elements/reactions/DOMTokenList.html
```diff
@@ -0,0 +1,223 @@
...
+    instance.classList = 'hello';
+    var logEntries = element.log();
+    assert_array_equals(logEntries.types(), ['attributeChanged']);
+    assert_equals(logEntries.log().name, 'class');
+    assert_equals(logEntries.log().oldValue, 'hello');
+    assert_equals(logEntries.log().newValue, 'hello');
+    assert_equals(logEntries.log().namespace, null);
+}, 'the stringifier of DOMTokenList must enqueue a attributeChanged reaction when the setter is called with the original value of the attribute');
+
+</script>
```
dominiccooney: I wonder if we also need tests for eg add('foo', 'bar') that check we only run "update steps" once and hence only one attributeChanged callback?

Another thing to maybe test is an element with a mix of observed and unobserved attributes to ensure the implementation is making the decision about firing attributeChanged on a per-attribute and not per-element basis.

custom-elements/reactions/resources/reactions.js
```diff
@@ -78,6 +78,38 @@ function testNodeDisconnector(testFunction, name) {
...
+        container.appendChild(instance);
+
+        instance.setAttribute('id', 'foo');
+        assert_array_equals(element.log().types(), ['constructed', 'connected', 'attributeChanged']);
+        var newInstance = testFunction(instance);
+        var logEntries = element.log();
+        assert_array_equals(logEntries.types(), ['constructed', 'attributeChanged']);
+        assert_equals(logEntries.log().name, 'id');
+        assert_equals(logEntries.log().oldValue, null);
+        assert_equals(logEntries.log().newValue, 'foo');
```
dominiccooney: Maybe test the namespace parameter here, too?

custom-elements/reactions/DOMTokenList.html
```diff
@@ -0,0 +1,223 @@
...
+<script>
+
+test(function () {
+    var element = defineNewCustomElement(['class']);
+    var instance = document.createElement(element.name);
+    assert_array_equals(element.log().types(), ['constructed']);
+    instance.classList.add('foo');
+    var logEntries = element.log();
+    assert_array_equals(logEntries.types(), ['attributeChanged']);
+    assert_equals(logEntries.log().name, 'class');
```
dominiccooney: This pattern of four assertions is repetitive; I wonder if the tests would be more readable if you add a helper which takes logEntries.log() and four parameters and have the helper do the four assertions?

custom-elements/reactions/resources/reactions.js
```diff
@@ -78,6 +78,38 @@ function testNodeDisconnector(testFunction, name) {
...
+    test(function () {
+        var element = defineNewCustomElement(['id']);
+        var instance = document.createElement(element.name);
+        container.appendChild(instance);
+
+        instance.setAttribute('lang', 'en');
+        assert_array_equals(element.log().types(), ['constructed', 'connected']);
+        var newInstance = testFunction(instance);
+        assert_array_equals(element.log().types(), ['constructed']);
+    }, name + ' must not enqueue a attributeChanged reaction when cloning an element with an unobserved attribute');
```
dominiccooney: Is there a way to make this "an element with an unobserved attribute" more accurate but equally succinct? It must not queue an attribute changed callback for *unobserved attributes*; but a custom element must queue an attributeChanged callback for an observed attribute, even if there happen to be other unobserved attributes. Same goes for the earlier file.

dominiccooney (commented):
Thorough tests! Comments inline.

custom-elements/reactions/DOMTokenList.html
rniwa: Removed

custom-elements/reactions/DOMTokenList.html
rniwa: Alright, added assert_attribute_log_entry to reactions.js.

custom-elements/reactions/DOMTokenList.html
rniwa: Oops, fixed that.

custom-elements/reactions/DOMTokenList.html
rniwa: Oops, fixed

custom-elements/reactions/DOMTokenList.html
rniwa: Fixed

custom-elements/reactions/DOMTokenList.html
rniwa: Okay, I think we can address that by using non-empty attribute list.

rniwa: And added a test case for adding multiple tokens and removing multiple tokens.

custom-elements/reactions/resources/reactions.js
rniwa: Sure, I added a new test case. But I think in general, those edge cases are better tested in https://github.com/w3c/web-platform-tests/pull/3853 which targets each reaction type more throughly.  These individual CEReaction tests are just sanity check for each IDL attribute so I don't think we necessary have to exhaustively test each edge case.

custom-elements/reactions/resources/reactions.js
rniwa: Sure, now that I'm using a helper, that's done.

dominiccooney (commented):
LGTM

PR #3839, opened 2016-09-28
Revert "[html] Do not specify step attribute values multiplied by ste…

tkent-google:
…p scale factor. (#3838)"

This reverts commit f0cb4eb16a0fa8ddbbd4c4ccd068457aca579ca6.
It contained unexpected change to "tools".


tkent-google:
Skipped review because this is a revert.

PR #3847, opened 2016-09-29
Mark traverse_the_session_history_unload_prompt_1.html as manual

cdumez:
Mark traverse_the_session_history_unload_prompt_1.html as manual since it
requires the user to click on a dialog. The test is actually in a
non-automated folder but it is still run by the test runner because it
does not have the "-manual.html" suffix.


jdm:
Thanks!

PR #3753, opened 2016-09-19
Restore iteration order test for indexed properties

zcorpan:
See https://github.com/w3c/web-platform-tests/pull/3742#issuecomment-248016252

PR #3848, opened 2016-09-29
Fix html/dom/documents/dom-tree-accessors/document.forms.html

cdumez:
The test uses array.slice() with a wrong deleteCount parameter and then
did not take into account the fact that array.slice() modified the
input array.


cdumez:
Regression from PR #3753 I believe.

Ms2ger:
Doh, thanks.

PR #3849, opened 2016-09-29
CSP font-src reftests

hillbrad:


mikewest:
I wonder if we could turn these into automagical tests by examining box sizes or something.

This LGTM for the short term, regardless. Merging. Thanks!

hillbrad:
I'm sure that's possible, was thinking about it as I brushed my teeth this
morning, but not a big priority right now.  :)  Thx for reviewing.

PR #3851, opened 2016-09-30
Now does not execute optional subtests that would fail

halindrome:
If a .test file indicates that certain assertionTypes should be
skipped on fail, then those will not have their tests "run" so they
will show up on the results in a more sensible way.


BigBlueHat (approved):
This looks great and will help us cover more ground with these tests. Thanks @halindrome!

BigBlueHat:
Oh...@halidrome apparently there's some lint error...should be a quick fix, I hope.

PR #3083, opened 2016-05-28
Update generated tests; add shell script to avoid people working out how

gsnedders:
At the moment, most of our generated tests are badly documented as to how to generate them. One solution is to just provide a Makefile that does all the heavy lifting (and ultimately, hopefully, have a Makefile at the root which just regenerates all tests with a generation step, and ideally we can then run that on Travis CI to ensure they all stay up to date). This does that, and regenerates all the tests. And fixes the breakage caused by people not updating the source files properly.

PR #3852, opened 2016-09-30
Modify some tests under 2dcontext

xidachen:
There are some tests under 2dcontext/ that doesn't reflect the specs.
1. 2d.composite.operation.clear.html
'clear' is a valid globalCompositeOperation, so assign that should change ctx.globalCompositeOperation to be 'clear'.
1. 2d.drawImage.zerocanvas.html
   The spec here: https://html.spec.whatwg.org/multipage/scripting.html#drawing-images, says in step 1, that when source canvas size is 0, "returns bad, then abort these steps without drawing anything", so no exception should be thrown.
2. 2d.drawImage.zerosource.html
   The spec link is the same as the above one. In step 3, it states "If one of the sw or sh arguments is zero, abort these steps. Nothing is painted.", so no exception should be thrown.

Notice that Firefox and Edge is likely to fail under this change.

Please review.


jdm:
These tests are apparently automatically generated: https://github.com/xidachen/web-platform-tests/blob/a2b5cc58e8708363cbb7b6172b3bfd526a73212e/2dcontext/compositing/2d.composite.operation.clear.html#L2
We should be updating the generation script as well.

sideshowbarker:
> These tests are apparently automatically generated…
> We should be updating the generation script as well.

Yes, the scripts and data files for generating the tests are here:

https://github.com/w3c/web-platform-tests/tree/master/2dcontext/tools

xidachen:
In the new commit, I have updated the script that generates the test, could you let me know how should I run the test script locally? I run 2dcontext/gentest.py, and I got:

  File "2dcontext/tools/gentest.py", line 131, in <module>
    templates = yaml.load(open('templates.yaml', "r").read())
IOError: [Errno 2] No such file or directory: 'templates.yaml'

But I can see 'templates.yaml' is under the same dir as the gentest.py.

jdm:
You probably need to run gentest.py with `2dcontext/tools` as the working directory.

sideshowbarker:
> You probably need to run gentest.py with `2dcontext/tools` as the working directory.

Yes you do (we should fix it at some point so that you don’t have to be in the tools directory to call it but for now that is what’s required)

xidachen:
I ran the 2dcontext/tools/gentest.py locally, and it is making lots of changes that I didn't expect, and the one that I expect it to change doesn't seem to appear. Could you please take a look at the last commit and see what's going on. Thanks.

jgraham:
Making changes you don't expect isn't too unlikely, people either forget to run that tool, or it generates slightly different PNGs on their system compared to other systems. I will have to look more closely to understand why it might not generate the changes you expect.

xidachen:
Close this since I have a new pull request here:
https://github.com/w3c/web-platform-tests/pull/4234

gsnedders:
#3083 fixes the generated test issue, but is awaiting review.

PR #3855, opened 2016-10-02
fix type checking for InvalidAccessError

schien:


tidoust (changes requested):
I think there is a problem with that code: the test actually passes if the promise is resolved. It should not: the test needs to fail if the app can start a presentation without user gesture!

This problem actually existed before.

The whole test should rather be written using `promise_rejects`, as in:

``` javascript
promise_test(function (t) {
  return promise_rejects(t, 'InvalidAccessError', request.start());
});
```

schien:
Patch is updated according to @tidoust 's comment

tidoust (approved):

PR #3859, opened 2016-10-03
Fix WebDriver resize tests

andreastt:
This change disables one test which will hang because the conditional
wait statements are just plain wrong.

### DIFF
--- a/webdriver/contexts.py
+++ b/webdriver/contexts.py
@@ -1,34 +1,62 @@
+import json
 import pytest
 import webdriver
 
 
-def test_resize(session):
+@pytest.mark.xfail(raises=webdriver.UnsupportedOperationException)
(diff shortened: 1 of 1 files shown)


webdriver/contexts.py
```diff
@@ -1,27 +1,54 @@
...
+
+
+def test_window_resize(session):
     # setting the window size by webdriver is synchronous
     # so we should see the results immediately
 
-    session.window.size = (200, 100)
-    assert session.window.size == {"width": 100, "height": 200}
+    session.window.size = (400, 500)
+    assert session.window.size == (400, 500)
```
jgraham: This can also return unsupported operation

andreastt: Fixed

PR #3860, opened 2016-10-03
Add Get Window Position type test

andreastt:


webdriver/contexts.py
```diff
@@ -25,3 +25,18 @@ def test_resize_by_script(session):
     while size1 == size2:
         size2 = session.window.size
     assert size2 == {"width": 200, "height": 100}
+
+def test_window_position_types(http, session):
+    session.start()
+    with http.get("/session/%s/window/position" % session.session_id) as resp:
```
jgraham: Can return unsupported operation.

andreastt: Good catch. Addressed with an [`@pytest.mark.xfail`](http://doc.pytest.org/en/latest/skipping.html) decorator, which if I understand it correctly, will allow the test to fail with an exception if using the `raises=webdriver.UnsupportedOperationException` kwarg, but not cause any fuss if it runs fine because the `strict` kwarg isn’t used.

PR #3865, opened 2016-10-04
Update Shadow DOM event dispatch tests for synthetic events

hayatoito:
Synthetic events must be dispatched at least until the root of the tree
even when target and relatedTarget are identical.

### DIFF
--- a/shadow-dom/event-composed-path-with-related-target.html
+++ b/shadow-dom/event-composed-path-with-related-target.html
@@ -27,7 +27,7 @@ test(() => {
   let path = ['target', 'test1'];
   assert_event_path_equals(log, [['target', 'target', 'target', path],
                                  ['test1', 'target', 'target', path]]);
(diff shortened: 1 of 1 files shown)


TakayoshiKochi:
Code change LGTM.
Could you have a reference to the bug or spec clarification discussion in the description?

shadow-dom/event-composed-path-with-related-target.html
```diff
@@ -134,7 +136,9 @@
 test(() => {
   let n = createTestTree(test4);
   let log = dispatchEventWithLog(n, n.host1, new FocusEvent('my-focus', { bubbles: true, composed: true, relatedTarget: n.target }));
-  assert_event_path_equals(log, []);
```
rniwa: This change needs to reverted if we change the spec with your proposal to trim the event path to `[]` when the related target is inside the target's shadow tree but we can land this as is for now.

rniwa (approved):
r=me.

PR #3871, opened 2016-10-04
Add tests for DataTransfer.prototype.types as a frozen array

domenic:
These test the changes in https://github.com/whatwg/html/pull/1860.

These tests are unfortunately manual, but they do work (tested locally).


zcorpan:
LGTM

PR #3872, opened 2016-10-04
fix Resource Timing test for transferSize

igrigorik:
transferSize will vary from browser to browser depending on the set of
default headers and other variables. Instead of hardcoding a specific
value we'll check if the reported value is greater than the encoded body
payload (for uncached resources).

Also, moved nextHopProtocol test into separate test entry.

@toddreifsteck PTAL. Re: https://github.com/w3c/web-platform-tests/pull/2282.


toddreifsteck:
LGTM. Merging

PR #3873, opened 2016-10-04
Fix html/webappapis/scripting/events/event-handler-attributes-body-wi…

cdumez:
…ndow.html

Fix html/webappapis/scripting/events/event-handler-attributes-body-window.html so
that it no longer requires user interaction before moving on to the next test.
The test is registering a number of event listeners on the body / window, which
caused a confirmation dialog to be shown on unload.

So that the test can be run in an automated fashion, we now unregister the event
listeners at the end of the test. We now also checks that the event listeners
have properly been unregistered.


jdm:
Thanks!

PR #3875, opened 2016-10-05
Do not include a, area, or frameset in the window's named properties

domenic:
Aligns the tests with the spec as of https://github.com/whatwg/html/pull/1869. See discussion in https://github.com/whatwg/html/issues/1810.


cdumez:
glam

PR #3876, opened 2016-10-05
Fix custom-elements/htmlconstructor/newtarget.html

rniwa:
The document was a missing body element even though some tests accessed document.body.
In the first test case, a single variable was referred as either `afterDefinition` and `beforeDefinition`.

All test cases relied on a bound function to override its prototype but this doesn't work because `[[Construct]]` of a bound function uses its target as `new.target`:
https://tc39.github.io/ecma262/#sec-bound-function-exotic-objects-construct-argumentslist-newtarget

Replaced them with a regular function wrapped in a Proxy since "prototype" is not configurable.

The last two test cases were erroneously calling customElement.define with the same name multiple times to test multiple non-object values. The inverted the loop and the call to test_with_window so that each define call happens in a different window object.

Finally, the test case in which `Get(new.target, "prototype")` threw inside the `HTMLElement` constructor had to delay the throwing until we've actually reached the `HTMLElement` constructor since the kind of a regular function is "base" and `[[Construct]]` gets its prototype before invoking the constructor:
https://tc39.github.io/ecma262/#sec-ecmascript-function-objects-construct-argumentslist-newtarget

### DIFF
--- a/custom-elements/htmlconstructor/newtarget.html
+++ b/custom-elements/htmlconstructor/newtarget.html
@@ -5,100 +5,100 @@
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <script src="../resources/custom-elements-helpers.js"></script>
-
+<body>
 <script>
 "use strict";
 
 test_with_window(w => {
-  let afterDefinition = false;
+  let beforeDefinition = true;
   const proto1 = { "proto": "number one" };
   const proto2 = { "proto": "number two" };
 
-  const TestElement = (function () {
-    assert_throws({ name: "prototype throws" }, () => {
-      const o = Reflect.construct(w.HTMLElement, [], new.target);
+  function TestElement() {
+    const o = Reflect.construct(w.HTMLElement, [], new.target);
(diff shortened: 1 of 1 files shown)


custom-elements/htmlconstructor/newtarget.html
```diff
@@ -5,100 +5,100 @@
...
 
-  throws = true;
-  new TestElement();
+  new ElementWithDynamicPrototype();
 
 }, "Rethrow any exceptions thrown while getting the prototype");
 
-test_with_window(w => {
-  for (const notAnObject of [null, undefined, 5, "string"]) {
+[null, undefined, 5, "string"].forEach(function (notAnObject) {
```
domenic: This creates duplicate test names; you'll need to change the test name to include the thing being tested.

domenic (approved):
Thanks so much for doing this! Sorry for writing such a bad test. One minor issue but LGTM otherwise.

custom-elements/htmlconstructor/newtarget.html
rniwa: Oh, good point. Fixed that.

PR #3879, opened 2016-10-06
Use correct MIME type for multipart form data.

eehakkin:
Objects implementing the Body mixin return a new FormData object only if
a MIME type is a correct one and throw a TypeError otherwise. Thus use
a correct MIME type for formData tests.

### DIFF
--- a/fetch/api/response/response-consume.html
+++ b/fetch/api/response/response-consume.html
@@ -12,17 +12,49 @@
   </head>
   <body>
     <script>
-    function checkBodyText(response, expectedBody) {
+    function responsePromise(body, responseInit) {
+      return new Promise(function(resolve, reject) {
+        resolve(new Response(body, responseInit));
+      });
+    }
+
+    function responseStringToMultipartFormTextData(response, name, value) {
+        assert_true(response.headers.has("Content-Type"), "Response contains Content-Type header");
+        var boundaryMatches = response.headers.get("Content-Type").match(/;\s*boundary=("?)([^";\s]*)\1/);
+        assert_true(!!boundaryMatches, "Response contains boundary parameter");
+        return stringToMultipartFormTextData(boundaryMatches[2], name, value);
+    }
+
+    function streamResponsePromise(streamData, responseInit) {
+      return new Promise(function(resolve, reject) {
+        var stream = new ReadableStream({
+          start: function(controller) {
+            controller.enqueue(stringToArray(streamData));
+            controller.close();
+          }
+        });
+        resolve(new Response(stream, responseInit));
+      });
+    }
+
+    function stringToMultipartFormTextData(multipartBoundary, name, value) {
+      return ('--' + multipartBoundary + '\r\n' +
+              'Content-Disposition: form-data;name="' + name + '"\r\n' +
+              '\r\n' +
+              value + '\r\n' +
+              '--' + multipartBoundary + '--\r\n');
+    }
+
+    function checkBodyText(test, response, expectedBody) {
       return response.text().then( function(bodyAsText) {
         assert_equals(bodyAsText, expectedBody, "Retrieve and verify response's body");
         assert_true(response.bodyUsed, "body as text: bodyUsed turned true");
       });
     }
 
-    function checkBodyBlob(response, expectedBody, checkContentType) {
+    function checkBodyBlob(test, response, expectedBody, expectedType) {
       return response.blob().then(function(bodyAsBlob) {
-        if (checkContentType)
-          assert_equals(bodyAsBlob.type, "text/plain", "Blob body type should be computed from the response Content-Type");
+        assert_equals(bodyAsBlob.type, expectedType || "text/plain", "Blob body type should be computed from the response Content-Type");
 
         var promise = new Promise( function (resolve, reject) {
           var reader = new FileReader();
@@ -41,14 +73,14 @@
       });
     }
 
-    function checkBodyArrayBuffer(response, expectedBody) {
+    function checkBodyArrayBuffer(test, response, expectedBody) {
       return response.arrayBuffer().then( function(bodyAsArrayBuffer) {
         validateBufferFromString(bodyAsArrayBuffer, expectedBody, "Retrieve and verify response's body");
         assert_true(response.bodyUsed, "body as arrayBuffer: bodyUsed turned true");
       });
     }
 
-    function checkBodyJSON(response, expectedBody) {
+    function checkBodyJSON(test, response, expectedBody) {
       return response.json().then(function(bodyAsJSON) {
         var strBody = JSON.stringify(bodyAsJSON)
         assert_equals(strBody, expectedBody, "Retrieve and verify response's body");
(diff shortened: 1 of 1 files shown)


fetch/api/response/response-consume.html
```diff
@@ -21,7 +29,7 @@
 
     function checkBodyBlob(response, expectedBody, checkContentType) {
       return response.blob().then(function(bodyAsBlob) {
-        if (checkContentType)
+        if (checkContentType !== false)
```
youennf: This change is ok, but why should it be done? Is it the expected wpt coding style?

fetch/api/response/response-consume.html
```diff
@@ -58,61 +66,68 @@
 
     function checkBodyFormData(response, expectedBody) {
       return response.formData().then(function(bodyAsFormData) {
-        assert_true(bodyAsFormData instanceof FormData, "Should receive a FormData");
```
youennf: This assert is still true, why removing it?

fetch/api/response/response-consume.html
```diff
@@ -58,61 +66,68 @@
...
         assert_false(response.bodyUsed, "bodyUsed is false at init");
         return checkFunction(response, body);
       }, "Consume response's body as " + bodyType);
     }
 
-    var formData = new FormData();
-    formData.append("name", "value");
     var textData = JSON.stringify("This is response's body");
-    var blob = new Blob([textData], { "type" : "text/plain" });
+    var blob = new Blob([textData], { "type": "application/octet-stream" });
```
youennf: This is ok but is there a particular reason to change the type here?

fetch/api/response/response-consume.html
```diff
@@ -58,61 +66,68 @@
...
+    var blobResponseInit = { "headers": [["Content-Type", "text/PLAIN"]] };
+    var multipartBoundary = "boundary-" + Math.random();
+    var formResponseInit = { "headers": [["Content-Type", 'multipart/FORM-data; boundary="' + multipartBoundary + '"']] };
```
youennf: Would it make sense to add a test with a bad header value?

youennf (commented):
Sounds good to me.
A couple of questions and it would be great to add an error test for bad header values

fetch/api/response/response-consume.html
eehakkin: Yes, it would make sense. Done.

eehakkin:
I rebased on top of master, resolved conflicts caused by #3948 and improved based on that commit. Please take a look.

eehakkin:
@youennf, can I do something to get this merged?

youennf:
I forgot about this one... sorry about that.
I will check it today.

fetch/api/response/response-consume.html
```diff
@@ -12,17 +12,31 @@
...
+    function responseStringToMultipartFormTextData(response, name, value) {
+        assert_true(response.headers.has("Content-Type"), "Response contains Content-Type header");
+        var boundaryMatches = response.headers.get("Content-Type").match(/;\s*boundary=("?)([^";\s]*)\1/);
+        assert_true(!!boundaryMatches, "Response contains boundary parameter");
+        return stringToMultipartFormTextData(boundaryMatches[2], name, value);
+    }
+
+    function stringToMultipartFormTextData(multipartBoundary, name, value) {
+      return ('--' + multipartBoundary + '\r\n' +
+              'Content-Disposition: form-data; name="' + name + '"\r\n' +
```
youennf: I guess this is ok for now.
I wonder whether we will not want something like:
Content-Disposition: form-data;name...

fetch/api/response/response-consume.html
```diff
@@ -56,80 +70,134 @@
...
+        var entryName = "name";
+        var strBody = entryName + "=" + bodyAsFormData.get(entryName);
+        assert_equals(strBody, expectedBody, "Retrieve and verify response's body");
         assert_true(response.bodyUsed, "body as formData: bodyUsed turned true");
      });
     }
 
-    function checkResponseBody(body, bodyType, bodyConsumingType, checkFunction) {
+    function checkBodyFormDataError(response, expectedBody) {
+      return response.formData().then(
```
youennf: Can you use promise_rejects instead?

fetch/api/response/response-consume.html
```diff
@@ -56,80 +70,134 @@
...
+        function(bodyAsFormData) {
+          assert_unreached("Form data parsing should fail");
+        },
+        function(e) {
+          assert_throws(new TypeError(), function() { throw e });
+          assert_true(response.bodyUsed, "body as formData: bodyUsed turned true");
+        });
+    }
+
+    function checkResponseBody(body, expectedBody, bodyType, bodyConsumingType, checkFunction, responseInit) {
```
youennf: That makes a lot of parameters, making it a bit difficult to understand.
How about moving bodyType and bodyConsumeType at the end.
And maybe combining them in something like "from XX to YY", the response name being 
"Consume response body: from XX to YY"

Also we use have two parameters body and responseInit. Might want to group them together and put them at the start. Or directly pass a response if that improves readability.

fetch/api/response/response-consume.html
```diff
@@ -56,80 +70,134 @@
...
+    checkResponseBody(blob, textData, "blob", "blob", checkBodyBlob, textResponseInit);
+    checkResponseBody(blob, textData, "blob", "text", checkBodyText);
+    checkResponseBody(blob, textData, "blob", "arrayBuffer", checkBodyArrayBuffer);
+    checkResponseBody(blob, textData, "blob", "json", checkBodyJSON);
+    checkResponseBody(formBlob, formTextData, "blob", "formData with correct multipart type", checkBodyFormDataMultipart, formTextResponseInit);
+    checkResponseBody(formBlob, undefined, "blob", "formData without correct multipart type", checkBodyFormDataError, textResponseInit);
+    checkResponseBody(urlSearchParamsBlob, urlSearchParamsData, "blob", "formData with correct urlencoded type", checkBodyFormDataUrlencoded, urlSearchParamsResponseInit);
+    checkResponseBody(urlSearchParamsBlob, undefined, "blob", "formData without correct urlencoded type", checkBodyFormDataError, textResponseInit);
+
+    function checkFormDataResponseBody(body, expectedName, expectedValue, bodyType, bodyConsumingType, checkFunction, responseInit) {
```
youennf: Ditto here

fetch/api/response/response-consume.html
```diff
@@ -56,80 +70,134 @@
...
           assert_false(response.bodyUsed, "bodyUsed is false at init");
           return checkFunction(response, "top");
         });
-      }, "Consume fetched response's body as " + bodyType);
+      }, "Consume fetched response's body as " + bodyConsumingType);
     }
     checkFetchedResponseBody("blob", checkBodyBlob);
     checkFetchedResponseBody("text", checkBodyText);
     checkFetchedResponseBody("arrayBuffer", checkBodyArrayBuffer);
+    checkFetchedResponseBody("formData without correct type", checkBodyFormDataError);
```
youennf: It might not be very clear for readers that some tests are error cases (meaning we will not be able to get a body) and some are regular cases. Maybe we should group all the error tests together or give them a name suffix like (error case).

youennf:
Sounds good in general.
I am just wondering how we could improve a bit the readability. See comments above.

fetch/api/response/response-consume.html
eehakkin: Done. I was trying to avoid passing the test object to check function.

fetch/api/response/response-consume.html
eehakkin: > How about moving bodyType and bodyConsumeType at the end.
> And maybe combining them in something like "from XX to YY", the response name being
> "Consume response body: from XX to YY"

Done.

> Also we use have two parameters body and responseInit. Might want to group them together and put them at the start. Or directly pass a response if that improves readability.

I'd rather not pass a response directly because that would change initialization errors from test failures to global errors. However, I created helper functions which create response promises so that the two parameters are grouped together and named logically.

fetch/api/response/response-consume.html
eehakkin: I added the name suffix.

eehakkin:
With combined parameters and named helper functions, I think that the readability has improved.

youennf:
Sounds great, thanks!

Issue #3527, opened 2016-08-22
Test expects AbortError, but should rather get NotAllowedError


tidoust:
@louaybassbouss, @mariuswsl,

Presentation API test that checks display selection abortion expects to receive an `AbortError`:
https://github.com/w3c/web-platform-tests/blob/master/presentation-api/controlling-ua/startNewPresentation_error-manual.html#L33

That seems wrong to me. Aborting the selection of a display should match step 11 of the [starting a connection](http://w3c.github.io/presentation-api/#starting-a-presentation) algorithm and should thus result in a `NotAllowedError` exception.

Also, note that [exceptions messages](https://heycam.github.io/webidl/#dfn-exception-message) are implementation specific so the test should not assume that `err.message` will equal `Dialog closed.`.

louaybassbouss:
agree with @tidoust, @mariuswsl can you fix this issue?

PR #3880, opened 2016-10-06
Adding tests for Starting a presentation in Presentation API

tomoyukilabs:
This PR would add tests for [Starting a presentation](https://w3c.github.io/presentation-api/#starting-a-presentation):
- OperationError when there is already an unsettled Promise (step 5)
- NotFoundError when there is no available presentation display (step 10)
- NotAllowedError when the user denied permission (step 11)


louaybassbouss:
Thx @tomoyukilabs  for your contribution I will review this PR and come back to you ASAP

tidoust:
Hi @tomoyukilabs,

Code seems good. Thanks! 3 comments:

1. The test suite already contains a similar test for step 11 under the name `startNewPresentation_error-manual.html`. This test has a bug though (see #3527). Your version fixes that bug and also gives the file a more explicit name. That's all good! Could your PR also delete `startNewPresentation_error-manual.html`?
2. I created a PR on testharness.js to include `NotAllowedError` in the list of DOMException (see https://github.com/w3c/testharness.js/pull/216). Once merged, you should be able to replace `new DOMException('', 'NotAllowedError')` by `'NotAllowedError'`.
3. For step 10, the _will remain so_ condition in the spec leaves the door open for implementers to simply wait forever and never fail with a `NotFoundError`. I don't have really better suggestions, I think we should include this test and see how browsers behave.

tomoyukilabs:
Thanks a lot, @tidoust!
As your PR on testharness.js (https://github.com/w3c/testharness.js/pull/216) was merged, I updated my PR to reflect 1. and 2.

> I think we should include this test and see how browsers behave.

+1. As far as I have tested, Chrome shows "Device missing?" message in the dialog box after failing to search a device, and then throws `NotFoundError` when the dialog is closed. Of course, I look forward to seeing how other browsers behave.

PR #3882, opened 2016-10-06
Fix #3878: Reduce number of playbacks and playback duration for encrypted / clear test

### LINKED ISSUE #3878: [eme] drm-mp4-playback-temporary-encrypted-clear-sources.html needs a longer timeout
https://www.w3c-test.org/encrypted-media/drm-mp4-playback-temporary-encrypted-clear-sources.html is timing out on all three browsers. In Chrome, it plays three or four videos then reports TIMEOUT. When it played three, the fourth played after the TIMEOUT was reported. I'm guessing it needs more time.

The Clear Key version may just barely pass because it does not need to fetch licenses.

Another option would be to break this test up into two tests that play two videos each. It would also be nice if the description explained what the four videos are.

mwatson2:


encrypted-media/clearkey-mp4-playback-temporary-encrypted-clear-sources.html
```diff
@@ -2,7 +2,7 @@
 <html>
   <head>
     <meta charset=utf-8>
-    <title>Encrypted Media Extensions: Successful Playback, Temporary session with Clear Key, mp4, Encrypted and Clear sources</title>
+    <title>Encrypted Media Extensions: Successful Playback, alternate, separate, Encrypted and Clear sources, Temporary, mp4, Clear Key</title>
```
ddorwin: Is "alternate, separate, Encrypted and Clear sources" one thing? These blend in with other comma-separated items.

ddorwin (approved):
LGTM with comment

ddorwin (approved):

PR #3885, opened 2016-10-06
Respect type-in state when caching inline styles;

jgraham:
MozReview-Commit-ID: OZCUbNUvmz

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1250805


jgraham:
Code reviewed upstream.

PR #3887, opened 2016-10-06
The error event in a worker (firing at the worker's global) should have the original exception value in its .error property.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=986459


jgraham:
Code reviewed upstream.

PR #3891, opened 2016-10-06
Fix setting location.hash for documents with about:something URIs to actually work.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1298977


jgraham:
Code reviewed upstream.

PR #3896, opened 2016-10-06
Make sure the parser insertion point is defined when firing the load event for external <scripts> or firing the error event on a failed external script load (but not other cases, like bogus script URL).

jgraham:
If we have a creator parser, then we were a parser-inserted script and should
presumably be able to set a valid insertion point when we run or fire our
load/error events.  For the error event case, we do this in
nsScriptElement::ScriptAvailable, so that async error events due to things like
bogus script URLs do not end up with a valid insertion point.  For the load
event case, we just do this in ScriptEvaluated directly.

ScriptEvaluated is called while the scriptloader has our script set as the
current parser-inserted script.  But for the error event case we need to
maintain that state around the ScriptAvailable call that will fire the event.

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1297125


jgraham:
Code reviewed upstream.

PR #3897, opened 2016-10-06
Remove the existing Location hardcoding in js::SetPrototype now that Location handles that itself.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1299306


jgraham:
Code reviewed upstream.

PR #3903, opened 2016-10-06
Use valid and spec-compliant MP4 file for all tests. r=gerald,jgraham

jgraham:
The MP4 used in the tests was invalid:
1- It contains out of order and overlapping frames
2- The first frame for each of the first 9 media segments were marked as keyframe, even though they weren't

The new file was encoded such that:
- The total bitrate remains the same as in the previous files
- The total duration is similar
- Each media segment starts with a valid keyframe
- Frames are properly ordered
- GOP is made of 24 frames
- The video track doesn't start at 0, which helps testing the proper behaviour regarding buffered range calculation.
- Like in the previous video, the audio track is present but made of silence (which is puzzling for a video initially designed to test A/V sync)

The webm file presents the same characteristics.

MozReview-Commit-ID: 6AKKmInaenE

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1300296


jgraham:
Code reviewed upstream.

PR #3905, opened 2016-10-06
Set the one allowed sandbox navigator even when popups are allowed to escape the sandbox, so the new popup can be navigated by the sandboxed document.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1301138


jgraham:
Code reviewed upstream.

PR #3906, opened 2016-10-06
Fire error event for images with empty string src value;

jgraham:
MozReview-Commit-ID: AhXiTpgG9q3

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=599975


jgraham:
Code reviewed upstream.

PR #3907, opened 2016-10-06
LenientFloat handling should not prevent all arguments being processed; instead it should just cause an early return right after argument processing.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1303099


jgraham:
Code reviewed upstream.

PR #3909, opened 2016-10-06
Follow-up: allow more fuzz in the test for clock-skew.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1283720


jgraham:
Code reviewed upstream.

PR #3916, opened 2016-10-06
Fix #3915: Provide duration for multikey-sequential drm tests

### LINKED ISSUE #3915: [eme] drm-mp4-playback-temporary-multikey-sequential.html times out due to missing duration
drm-mp4-playback-temporary-multikey-sequential.html is timing out on all three browsers.

We found that `config.duration` is `undefined` in `onTimeupdate()`. Two issues:
 1. drm-mp4-playback-temporary-multikey-sequential.html does not specify `duration` like its peer, clearkey-mp4-playback-temporary-multikey-sequential.html.
 1. playback-temporary-multikey-sequential.js does not have `|| 0.5` like, for example, playback-temporary-encrypted-clear-sources.js.

More generally, why do any of the existing HTML files specify a duration? None of them are doing anything that requires an override.

mwatson2:


ddorwin (approved):
Thanks.

PR #3918, opened 2016-10-07
Fix #3917: Make the multikey-sequential test more stable

### LINKED ISSUE #3917: [eme] Multikey-sequential test is unstabled
As noted in this comment the [multikey-sequential](https://github.com/w3c/web-platform-tests/issues/3915#issuecomment-252115479) test is unstable on Chrome. This may be due to race conditions between the processing of the media and the retrieval of the license.

mwatson2:
Avoid race conditions between the processing of the media in the media source and the retrieval of the license. Check that playback does not continue whilst waiting for a key.

### DIFF
--- a/encrypted-media/scripts/playback-temporary-multikey-sequential.js
+++ b/encrypted-media/scripts/playback-temporary-multikey-sequential.js
@@ -38,20 +38,20 @@ function runTest(config,qualifier) {
             config.messagehandler(event.messageType, event.message, {variantId: event.target.variantId}).then(function(response) {
                 return event.target.update(response);
             }).then(function(){
-                if (!_waitingForKey) {
+                if (!video.src) {
                     _video.src = URL.createObjectURL(_mediaSource);
                     _video.play();
-                } else {
+                } else if (event.target.keyStatuses.size > 0){
                     _waitingForKey = false;
                 }
             }).catch(onFailure);
         }
 
         function onWaitingForKey(event) {
+            _waitingForKey = true;
             if (config.checkReadyState) {
                 assert_equals(_video.readyState, _video.HAVE_METADATA, "Video readyState should be HAVE_METADATA on watingforkey event");
             }
-            _waitingForKey = true;
             startNewSession();
         }
 
@@ -85,7 +85,6 @@ function runTest(config,qualifier) {
             return testmediasource(config);
         }).then(function(source) {
             _mediaSource = source;
-            
             startNewSession();
         }).catch(onFailure);
     }, testname);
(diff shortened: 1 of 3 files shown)


encrypted-media/scripts/playback-temporary-multikey-sequential.js
```diff
@@ -37,13 +37,21 @@ function runTest(config,qualifier) {
         function onMessage(event) {
             config.messagehandler(event.messageType, event.message, {variantId: event.target.variantId}).then(function(response) {
                 return event.target.update(response);
+            }).then(function(){
+                if (!_waitingForKey) {
+                    _video.src = URL.createObjectURL(_mediaSource);
```
ddorwin: It seems we could get here multiple times since _waitingForKey gets reset below. Even if that's not expected, it seems fragile. You could just check `!video.src` to avoid this.

encrypted-media/scripts/playback-temporary-multikey-sequential.js
```diff
@@ -37,13 +37,21 @@ function runTest(config,qualifier) {
...
+                    _video.play();
+                } else {
+                    _waitingForKey = false;
```
ddorwin: This is not guaranteed to be true. For example, multiple rounds might be necessary. (In practice, this probably only applies to the first message, which would fall into the if block instead.)

encrypted-media/scripts/playback-temporary-multikey-sequential.js
```diff
@@ -37,13 +37,21 @@ function runTest(config,qualifier) {
...
+                }
             }).catch(onFailure);
         }
 
         function onWaitingForKey(event) {
             if (config.checkReadyState) {
                 assert_equals(_video.readyState, _video.HAVE_METADATA, "Video readyState should be HAVE_METADATA on watingforkey event");
             }
+            _waitingForKey = true;
```
ddorwin: nit: Make this the first statement. We are waiting as soon as this is called.

ddorwin (approved):
LGTM with comments.

encrypted-media/scripts/playback-temporary-multikey-sequential.js
mwatson2: Ack

encrypted-media/scripts/playback-temporary-multikey-sequential.js
mwatson2: I could check `keyStatuses.size`.

encrypted-media/scripts/playback-temporary-multikey-sequential.js
mwatson2: ack

ddorwin (approved):

PR #3919, opened 2016-10-07
Fix step_func_done() usage in pointerlock_shadow-manual.html

TakayoshiKochi:
The test misused step_func_done(), and it could succeed when early
return happened, which should be ignored.
Use step_func() and test.done() explicitly so that we only catch
the correct condition.

In addition, this removes unexpected TABs.


TakayoshiKochi:
@scheib, could you take a look?

scheib (approved):

PR #3922, opened 2016-10-07
Fix our named property DOM proxy code to handle deletion correctly when expandos are involved.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1297304


jgraham:
Code reviewed upstream.

PR #3925, opened 2016-10-07
Backout newlines lost in .innerHTML serialization of <pre> when first character is a newline.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=838954


jgraham:
Code reviewed upstream.

gsnedders:
Should we not be changing the test, rather than deleting it?

jdm:
This was an upstream of a complete revert of the commit that added the test, unfortunately.

PR #3926, opened 2016-10-07
Use Symbol.iterator instead of values() to get Iterator prototype.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1299593


jgraham:
Code reviewed upstream.

PR #3927, opened 2016-10-07
A mochitest to verify that when text-decoration-color is currentColor, the value of it should be the same with color prop.

jgraham:
MozReview-Commit-ID: LBcHLWPawB7

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1267312


jgraham:
Code reviewed upstream.

PR #3933, opened 2016-10-07
Make update-the-finished-state.html not depend on frame timing;

jgraham:
I suspect we're hitting trouble when the refresh driver changes timer and we
end up with less time between frames.

MozReview-Commit-ID: I2dProiJTfh

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1297285


jgraham:
Code reviewed upstream.

PR #3936, opened 2016-10-07
[eme] Fix #3826: Test license expiration using playDuration

### LINKED ISSUE #3826: [eme] Test failure during playback when a license expires
When a license expires during playback, playback should fail. A license with a specific expiry time can be obtained with the code in PR #3798.

This test may be tricky to make reliable, since we only have 10 seconds of content and will need to set the expiration and start playback such that expiration occurs during playback. Likely the expiration will need to be set 10-15s in the future, to give time for the license exchange content loading etc. to complete, then playback could be started, then paused to wait until a few seconds before license expiration, then started again.

mwatson2:

### DIFF
--- a/encrypted-media/drm-mp4-playback-temporary-playduration-keystatus.html
+++ b/encrypted-media/drm-mp4-playback-temporary-playduration-keystatus.html
@@ -1,0 +1,53 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset=utf-8>
+    <title>Encrypted Media Extensions: Successful Playback, Temporary session limited playduration, check keystatus, DRM, mp4</title>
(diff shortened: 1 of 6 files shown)


paulbrucecotton:
@cpearce and @ddorwin :  Please look at this proposed new test ASAP so that we can finalize the EME test suite.

joeyparrish (approved):

gsnedders:
Is there any reason why this hasn't landed?

ddorwin:
There was some discussion of whether it would pass on the browsers in the issue (#3826). It looks like @joeyparrish reviewed it, though, so we can always land it and iterate or file browser bugs if there are issues. /cc @mwatson2 @jdsmith3000

PR #3942, opened 2016-10-10
SharedWorker no longer throws URLMismatchError

annevk:
See https://github.com/whatwg/html/pull/175 for details.

### DIFF
--- a/workers/constructors/SharedWorker/URLMismatchError.htm
+++ b/workers/constructors/SharedWorker/URLMismatchError.htm
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <meta charset="utf-8">
-<title>Web Workers: SharedWorker - throw URLMismatchError</title>
+<title>Web Workers: SharedWorker - same name, different URL</title>
 <link rel="author" title="Intel" href="http://www.intel.com">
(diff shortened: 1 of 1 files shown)


Ms2ger:
Do we want to check that they both run?

annevk:
Sure, took a little longer to make that actually work, but done.

domenic:
LGTM. Firefox passes this; Chrome fails per https://bugs.chromium.org/p/chromium/issues/detail?id=538914

For future reference here's a version using EventWatcher and promises to avoid the manual bookkeeping (untested):

```js
promise_test(t => {
  const worker1 = new SharedWorker("shared-worker.js", "name");
  worker1.port.postMessage("trigger a response");
  const eventWatcher1 = new EventWatcher(worker1, ["message"]);

  const worker2 = new SharedWorker("1", "name");
  const eventWatcher2 = new EventWatcher(worker2, ["message"]);
  
  return Promise.all([
    eventWatcher1.wait_for("message").then(e => {
      assert_equals(e.data, "ping");
    }),
    eventWatcher2.wait_for("message").then(e => {
      assert_array_equals(e.data, ["1", "name"]);
    })
  ]);
});
```

zcorpan (approved):

PR #3943, opened 2016-10-10
Workers and data URLs

annevk:
Provide basic tests for https://github.com/whatwg/html/pull/1782.

### DIFF
--- a/workers/data-url-shared.html
+++ b/workers/data-url-shared.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<title>data URL shared worker</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div id=log></div>
+<script>
+var worker = new SharedWorker("data:,onconnect = (e) => { fetch('/').then(() => e.ports[0].postMessage('fail'), () => e.ports[0].postMessage('pass')) }") // not same-origin
+worker.port.onmessage = (e) => {
+  assert_equals(e.data, "pass")
+  done()
+}
+</script>
--- a/workers/data-url-shared.html
+++ b/workers/data-url-shared.html
@@ -4,9 +4,10 @@
 <script src="/resources/testharnessreport.js"></script>
 <div id=log></div>
 <script>
-var worker = new SharedWorker("data:,onconnect = (e) => { fetch('/').then(() => e.ports[0].postMessage('fail'), () => e.ports[0].postMessage('pass')) }") // not same-origin
-worker.port.onmessage = (e) => {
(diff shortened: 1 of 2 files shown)


workers/data-url-shared.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<title>data URL shared worker</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div id=log></div>
+<script>
+var worker = new SharedWorker("data:,onconnect = (e) => { fetch('/').then(() => e.ports[0].postMessage('fail'), () => e.ports[0].postMessage('pass')) }") // not same-origin
```
foolip: https://fetch.spec.whatwg.org/ says that HTML "prevents creation of shared workers from such URLs." but that doesn't seem to match https://html.spec.whatwg.org/#dom-sharedworker

If they shouldn't be created at all, then presumably the constructor should throw?

workers/data-url.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<title>data URL worker</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div id=log></div>
+<script>
+var worker = new Worker("data:,fetch('/').then(() => self.postMessage('fail'), () => self.postMessage('pass'))") // not same-origin
+worker.onmessage = (e) => {
+  assert_equals(e.data, "pass")
```
foolip: testharness.js does implicitly create an `async_test` here, but for some reason a failure is still a harness error. @jgraham, how do these implicitly created tests work?

foolip (changes requested):
I've tested to run these:
https://w3c-test.org/submissions/3943/workers/data-url.html
https://w3c-test.org/submissions/3943/workers/data-url-shared.html

Only Firefox passes, the rest fail with a harness error.

workers/data-url-shared.html
annevk: Fetch needs to be updated, since we decided to change how shared workers should work.

annevk: I updated Fetch.

foolip (approved):

foolip:
Other now fail in the right way, merging.

zcorpan:
These already tested data: URLs

workers/constructors/SharedWorker/same-origin.html
workers/constructors/Worker/same-origin.htm

annevk:
Hmm, I remember running a grep looking for "data:". Weird. But those tests are not testing the origin of the worker itself, so I guess it was still useful for me to write those tests.

PR #3945, opened 2016-10-10
Basic <script> and data URL test

annevk:
See https://github.com/whatwg/html/pull/1782 for pointers to more
context.

### DIFF
--- a/html/semantics/scripting-1/the-script-element/data-url.html
+++ b/html/semantics/scripting-1/the-script-element/data-url.html
@@ -5,16 +5,19 @@
 <script src="/resources/testharnessreport.js"></script>
 <div id=log></div>
 <script>
-  setup({allow_uncaught_exception:true,explicit_done:true})
-  var counter = 0
-  window.onerror = (message, x, xx, xxx, e) => {
-    assert_equals(typeof e, "number")
-    assert_not_equals(message, "Script error.")
-    if(counter == 2) {
-      done()
-    }
-    counter++
-  }
+  setup({allow_uncaught_exception:true})
+  var t = async_test()
+  t.step(() => {
+    var counter = 0
+    window.onerror = t.step_func((message, x, xx, xxx, e) => {
+      assert_equals(typeof e, "number")
+      assert_not_equals(message, "Script error.")
+      if(counter == 2) {
+        t.done()
+      }
+      counter++
+    })
+  })
 </script>
 <script src="data:,throw 1"></script>
 <script src="data:,throw 2" crossorigin></script>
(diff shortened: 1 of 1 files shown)


annevk:
I can't quite get the `setup()` thing to work. Anyone any idea what I did wrong?

annevk:
Should be all good now.

html/semantics/scripting-1/the-script-element/data-url.html
```diff
@@ -0,0 +1,25 @@
+<!doctype html>
+<meta charset="utf-8">
+<title>data URL and scripts</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div id=log></div>
+<script>
+  setup({allow_uncaught_exception:true})
+  var t = async_test()
+  t.step(() => {
```
foolip: `async_test(t => { ... })` saves a line and one need not look to see if `t` is referenced elsewhere.

html/semantics/scripting-1/the-script-element/data-url.html
```diff
@@ -0,0 +1,25 @@
...
+    var counter = 1
+    window.onerror = t.step_func((message, x, xx, xxx, e) => {
+      assert_equals(typeof e, "number")
+      assert_equals(e, counter)
+      assert_not_equals(message, "Script error.") // Cannot be "muted" as data URLs are same-origin
+      if(counter == 3) {
```
foolip: More whitespace

foolip (changes requested):

foolip (approved):

foolip:
Hmm, https://w3c-test.org/submissions/3945/html/semantics/scripting-1/the-script-element/data-url.html actually fails on all but Firefox.

annevk:
Yeah, only Fx treats it as same-origin like it should.

foolip:
But the failures are because of other differences first. Edge only gives 4 arguments on the `onerror` callback, while Chrome and Safari make the 5th argument an object instead of number.

The spec is at https://html.spec.whatwg.org/#the-event-handler-processing-algorithm and Edge is wrong about the number of arguments, not sure about the type though, that probably affects `window.addEventListener('error', ...)` too.

annevk:
Pretty sure that's because Chrome and Safari treat these as cross-origin and therefore create the Error object with "Script error." as a message.

youennf:
I would think so as well.
WebKit might change that behavior soon so these tests are most welcome

foolip:
Ah, yes, the "Script error." check fails if those before it are commented out. Will change the order to make it obvious that it's a real failure to someone investigating the failure.

foolip:
@mikewest if you want to check this after the fact that'd be nice, but I've looked at it enough to be comfortable merging now.

PR #3946, opened 2016-10-10
Adding some more UTF-8 handling tests

youennf:


fetch/api/resources/utils.js
```diff
@@ -44,6 +44,11 @@ function checkRequest(request, ExpectedValuesDict) {
   }
 }
 
+function encode_utf8(str)
+{
+    return unescape(encodeURIComponent(str));
```
domenic: Although I hear that this works, I would feel more comfortable with `new TextEncoder.encode(str)` and comparing the resulting Uint8Arrays

domenic: Maybe it would be better to hard-code the values then as byte arrays.

youennf: Or use TextEncoder if available and fallback otherwise.
That is what encrypted-media/util/utf8.js is doing

fetch/api/resources/utils.js
```diff
@@ -46,6 +46,8 @@ function checkRequest(request, ExpectedValuesDict) {
 
 function encode_utf8(str)
 {
+    if (TextEncoder)
+        return TextEncoder.encode(str);
```
domenic: That doesn't work; you want `(new TextEncoder()).encode(str)`

youennf: Thanks, I fixed this accordingly!

PR #3947, opened 2016-10-10
Fixing redirect.py script so that it outputs correct parameters for the location URL

youennf:


jdm:
What's the difference between the old and new code?

youennf:
You can see the difference here: https://bugs.webkit.org/attachment.cgi?id=285414&action=review
Without the patch, we are encoding parameter values enclosed with [' and  '].

We should also probably replace:
    url += "&count=" + str(stashed_data['count']) 
with:
    url_parameters['count'] = str(stashed_data['count'])

But it seems the tests are passing without it

PR #1381, opened 2014-11-11
Test current Realm (still to be defined in more detail)

annevk:
Create a test based on http://web.mit.edu/bzbarsky/www/testcases/global-object-association/createImageData.html to see what the global object is for various things.

PR #3950, opened 2016-10-10
Adding more empty body tests

youennf:


domenic:
In Chrome the FormData test fails because Chrome serializes the body as

```
"------WebKitFormBoundary9G7XPy41qGToYQW7--
"
```

instead of as the empty string. Can you or @annevk confirm for me that the empty string is correct?

(Chrome also fails one test since `request.formData()` is missing for now. Maybe that is related? But I assume not.)

/cc @yutakahirano

annevk:
It's not defined, see https://github.com/whatwg/fetch/issues/392. We don't have a good definition of that format yet.

domenic:
Ah OK, probably better to not test that then. Glad I checked!

youennf:
I would prefer keeping that test if possible.
It runs within WebKit and I would prefer keeping it for non-regression purposes.

I am fine adding a comment noting that this test might need further validation.
Or moving it to an unstable/contribution test section.

domenic:
I don't know what the policies are generally, but I think it's not appropriate for WPT to test things that are not in specs and that are really regression tests for one specific vendor. You can always keep it within WebKit without keeping in in the web-platform-tests repository.

annevk:
Well, we do need tests for it since `formData()` is part of the standard, despite it not being defined in as much detail as it should be. Having it as a distinct test with some comment about multipart/form-data seems good enough to me. Perhaps it'll inspire someone to work on the standard for it.

domenic:
I don't understand though. Why would we include a test for something when we're not even sure what the expected values would be? That is, why is empty string valid and not the Chrome string?

annevk:
Until browsers are interoperable, we never really know what the expected values are. So I guess I don't see a problem picking one or the other here for now and keeping track of it in an issue.

We've been doing that elsewhere in WPT too as far as I've heard. Although admittedly a similar PR I wrote (#1381) is also in limbo at the moment.

youennf:
Since that particular test does not require any python script, I don't mind removing it.
Though having it in WPT may act as a reminder to get interoperability for that particular case. 
I will add a comment in the test for now.

sideshowbarker:
@domenic would you still rather this test be removed, or are you OK with it getting merged with the comment included?

domenic:
I guess it is OK for now, although I think we should reserve the right for @yutakahirano to reopen the discussion at a later date.

yutakahirano:
Oops, I overlooked this, sorry.

What is the expected Content-Type for such a Request / Response? According to the current spec it should be "multipart/form-data; boundary=...". And an empty string is not a valid body for "multipart/form-data" according to https://tools.ietf.org/html/rfc2046#section-5.1.

The test fails on Firefox 49 similarly. @wanderview, what do you think?

eehakkin:
It seems that most implementations have ended up encoding an empty FormData to "--boundary-string--\r\n" with a content type "multipart/form-data; boundary="boundary-string"" (with varying boundary strings, of course). This includes at least Chromium, Edge and Firefox.

See also https://github.com/w3c/web-platform-tests/pull/5906

wanderview:
It looks to me like chrome, edge, and FF all set content-type as @eehakkin says.  They only differ in the boundary string.

Chrome, Edge, and FF also all serialize `new Response(new FormData()).text()` to a single line of the boundary string.

Its difficult to say what safari would do here as it doesn't support FormData as a body yet.  (https://bugs.webkit.org/show_bug.cgi?id=161190)

Given that chrome/edge/ff are mostly in agreement, it would be nice to make the test check that `.text()` returns the boundary string as defined in `r.headers.get('content-type').split('=')[1]`.  I guess edge and firefox also append two additional dashes after that value where chrome does not.

In any case, empty string for text() seems incorrect based on current behavior.

eehakkin:
@wanderview wrote:
> Given that chrome/edge/ff are mostly in agreement, it would be nice to make the test check that .text() returns the boundary string as defined in r.headers.get('content-type').split('=')[1].

I did just that in https://github.com/w3c/web-platform-tests/pull/5906 (except that it also accepts quotes around the boundary parameter value). 

> I guess edge and firefox also append two additional dashes after that value where chrome does not.

No, they all encode empty FormData to "--boundary-string--\r\n" i.e. to "--" followed by the boundary string followed by "--" followed by "\r\n". It would be quite odd not to add the two additional dashes after the boundary string as that would form a delimiter-without-preceding "\r\n" instead of forming a close-delimiter-without-preceding-"\r\n". A close-delimiter denotes the end of a part whereas a delimiter denotes a beginning of a part.

> In any case, empty string for text() seems incorrect based on current behavior.

Yes, it seems.

PR #3951, opened 2016-10-10
Adding an error test where the error happens after the fetch promise resolution

youennf:


fetch/api/basic/error-after-response.html
```diff
@@ -0,0 +1,34 @@
...
+  <head>
+    <title>Fetch: network timeout after receiving the HTTP response headers</title>
+    <script src="/resources/testharness.js"></script>
+    <script src="/resources/testharnessreport.js"></script>
+    <script src="../resources/utils.js"></script>
+  </head>
+  <body>
+    <div id="log"></div>
+    <script>
+function checkReader(test, reader, checkRead)
```
domenic: I think this would be a lot clearer if it took a lambda instead of a boolean, so it was `checkReader(test, response.body.getReader(), r => r.read())` and `r => r.closed` at the call sites.

domenic:
This test does not pass in Chrome (http://w3c-test.org/submissions/3951/fetch/api/basic/error-after-response.html) which doesn't mean the test is wrong, but does mean @yutakahirano would be a good person to review.

Chrome stable seems to have an unhandled rejection, whereas Chrome Canary uses a NetworkError; the spec says TypeError.

youennf:
This test might also depend on the actual network stack.
Maybe there is a better way to test the error-after-resolving case.

domenic:
Hmm. I think it is a valid test, and Chrome just has its error types wrong. I would like @yutakahirano to confirm though if he has time. But if he does not have time by, say, 2016-10-19, we should merge this and Chrome can do any follow-up later when we investigate the test failures more thoroughly. I do not like letting tests sit waiting for review; WPT has too many of those already :)

yutakahirano:
Sorry for the delayed response.

The test fails on Chrome from two reasons:

1. Chrome applies mime sniffing. That means Chrome buffers first some bytes to determine/verify the mime type and then dispatch a responseReceived message. 10 bytes is small enough to be buffered.
2. Chrome errors a stream by DOMException(NetworkError) not by TypeError.

I resolved the former issue recently: https://codereview.chromium.org/2323623002/. After a few months the stable release will include the change. I don't know why we are using DOMException instead of TypeError. I will fix it shortly.

Thanks,

yutakahirano:
https://bugs.chromium.org/p/chromium/issues/detail?id=656601

yutakahirano:
Anyway, thank you for writing this test!

PR #3954, opened 2016-10-10
Add test for unwrapping a non-extractable key

mwatson2:
WebCrypto can unwrap wrapped JWK's marked as non-extractable. Such wrapped keys cannot be created with the `wrapKey` method, since wrapping is a form of extraction, but that can be created "by hand" by encrypting the appropriate JWK using `encryptKey`.

### DIFF
--- a/WebCryptoAPI/wrapKey_unwrapKey/wrapKey_unwrapKey.js
+++ b/WebCryptoAPI/wrapKey_unwrapKey/wrapKey_unwrapKey.js
@@ -321,13 +321,16 @@ function run_test() {
 
     // Can we compare key values by using them
     function canCompareNonExtractableKeys(key){
-        if (["AES-CTR", "AES-CBC", "AES-GCM", "RSA-OAEP", "RSASSA-PKCS1-v1_5", "RSA-PSS", "ECDSA", "HMAC"].indexOf(key.algorithm.name) === -1) {
+        if (["AES-CTR", "AES-CBC", "AES-GCM", "RSA-OAEP", "RSASSA-PKCS1-v1_5", "RSA-PSS", "ECDSA", "HMAC", "AES-KW"].indexOf(key.algorithm.name) === -1) {
             return false;
         }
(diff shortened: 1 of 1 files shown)


jimsch:
Why not use the WebCrypto function to do the AES KW since it will not unwrap correctly if it is not supported?

mwatson2:
WebCrypto will not wrap a non-extractable key, because wrapping is a form of extraction. So we have to do the wrapping ourselves in JS. We could use a "pure-JS" crypto library, but we can avoid bringing in one of those by (ab)using AES-CBC to do the ECB mode.

jimsch:
I am trying to figure out exactly what the test case you are looking for is and am not sure.  Part of the problem is that I think you need to verify that the ext value in the JWK is what you want it to be, but that may be an assumption on my part that it will not always be set on export.  There is a matrix of six cases that are possible here JWK.ext=<true, false, absent> and export=<true,false>.  It might make more sense to just explicitly test all cases as a single test rather than doing some the cases here and not doing the other cases.

mwatson2:
The tests I added are all for the case where we unwrap a key with JWK.ext=false and the extractable parameter also set to false.

We could also test that such unwrap calls fail when extractable is set to true.

The pre-existing tests cover the case where JWK.ext is set to true and where extractable is true. We could also test that unwrapKey with extractable = false still succeeds in this case.

We do not have tests for the case where JWK.ext is missing. The result should be the same as JWK.ext = true.

I didn't add any failure cases, though, since the test doesn't address any failires

jimsch:
There is still something about this test hat is bothering me, however I cannot put my finger on it so it does not make sense to hold up approval anymore.

mwatson2:
@jimsh Thanks. Is it just that is it not symmetrical ? Our API can consume wrapped non-extractable keys but cannot create them.

jimsch:
@mwatson2 no, it has more to do with questions about if the ext field is going to be set correctly in the case of ext:true by all exporters or if it would just be absent.  Defaults vs explicits.

PR #3955, opened 2016-10-10
Fix #3953: Minor problems with setMediaKeys test

### LINKED ISSUE #3953: [eme] *-mp4-setmediakeys-multiple-times-with-different-mediakeys.html have syntax error
```
setmediakeys-multiple-times-with-different-mediakeys.js:56 Uncaught SyntaxError: Unexpected token )
```

There appears to be an extra `)` after a `then(` without a `test.step_func(`. I tried fixing it but ran into a couple other problems, so someone familiar with this test should look at it.

Also, is there a reason that not all of the resolve functions use `test.step_func()`?

mwatson2:


encrypted-media/scripts/setmediakeys-multiple-times-with-different-mediakeys.js
```diff
@@ -53,34 +53,35 @@ function runTest(config, qualifier) {
             // Set mediaKeys2 on video element (switching MediaKeys) need not
```
ddorwin: This could probably be clearer. For example, "It is up the the implementation whether to allow the MediaKeys to be changed after setting the source. The following step handles both cases."

encrypted-media/scripts/setmediakeys-multiple-times-with-different-mediakeys.js
```diff
@@ -53,34 +53,35 @@ function runTest(config, qualifier) {
...
             // fail after src attribute is set.
             return _video.setMediaKeys(_mediaKeys2);
-        })).then(test.step_func(function() {
+        }).then(test.step_func(function() {
             // Switching setMediaKeys after setting src attribute on video element
```
ddorwin: This seems to say the same thing as above. Instead, it could say "Switching MediaKeys succeeded..."

encrypted-media/scripts/setmediakeys-multiple-times-with-different-mediakeys.js
```diff
@@ -53,34 +53,35 @@ function runTest(config, qualifier) {
...
             // is not required to fail.
-            assert_equals(_video2.mediaKeys, _mediaKeys2);
             fail = false;
+            assert_equals(_video.mediaKeys, _mediaKeys2);
             return Promise.resolve();
-        }, test.step_func(function(error) {
+        }), test.step_func(function(error) {
             fail = true;
```
ddorwin: And a similar comment here.

encrypted-media/scripts/setmediakeys-multiple-times-with-different-mediakeys.js
```diff
@@ -53,34 +53,35 @@ function runTest(config, qualifier) {
...
             assert_equals(_video.mediaKeys, _mediaKeys1);
             assert_in_array(error.name, ['InvalidStateError','NotSupportedError']);
```
ddorwin: Isn't there only one correct exception? I'm fine with having this for the purposes of making progress, but we should have an open issue and reference it here.

encrypted-media/scripts/setmediakeys-multiple-times-with-different-mediakeys.js
```diff
@@ -53,34 +53,35 @@ function runTest(config, qualifier) {
...
             // fail after src attribute is set.
             return _video.setMediaKeys(_mediaKeys2);
-        })).then(test.step_func(function() {
+        }).then(test.step_func(function() {
             // Switching setMediaKeys after setting src attribute on video element
             // is not required to fail.
-            assert_equals(_video2.mediaKeys, _mediaKeys2);
             fail = false;
+            assert_equals(_video.mediaKeys, _mediaKeys2);
             return Promise.resolve();
```
ddorwin: // Return something so the promise resolves properly.

encrypted-media/scripts/setmediakeys-multiple-times-with-different-mediakeys.js
```diff
@@ -53,34 +53,35 @@ function runTest(config, qualifier) {
...
+            assert_equals(_video.mediaKeys, null);
+            test.done();
         }), test.step_func(function(error) {
             if(fail) {
                 assert_equals(_video.mediaKeys, _mediaKeys1);
             } else {
                 assert_equals(_video.mediaKeys, _mediaKeys2);
             }
-            assert_is_array(error.name, ['InvalidStateError','ReferenceError']);
+            assert_in_array(error.name, ['InvalidStateError','ReferenceError']);
```
ddorwin: Ditto, though I'm not sure why `ReferenceError` would occur.

encrypted-media/scripts/setmediakeys-multiple-times-with-different-mediakeys.js
```diff
@@ -53,34 +53,35 @@ function runTest(config, qualifier) {
...
             fail = true;
             assert_equals(_video.mediaKeys, _mediaKeys1);
             assert_in_array(error.name, ['InvalidStateError','NotSupportedError']);
             assert_not_equals(error.message, '');
             // Return something so the promise resolves properly.
             return Promise.resolve();
         })).then(function() {
-            // Set null mediaKeys on video (clearing MediaKeys) not
-            // supported after src attribute is set.
+            // Set null mediaKeys on video (clearing MediaKeys) is not required to fail
```
ddorwin: ditto

encrypted-media/scripts/setmediakeys-multiple-times-with-different-mediakeys.js
```diff
@@ -53,34 +53,35 @@ function runTest(config, qualifier) {
...
+            // after src attribute is set.
             return _video.setMediaKeys(null);
         }).then(test.step_func(function() {
-            assert_unreached('Clearing mediaKeys after setting src should have failed.');
+            assert_equals(_video.mediaKeys, null);
+            test.done();
         }), test.step_func(function(error) {
             if(fail) {
```
ddorwin: `fail` is very specific. I think it should have a more specific name reflecting what exactly failed (or passed). Otherwise, it's not obvious why this `.then()` is the way it is.

ddorwin (changes requested):
Thanks for fixing this. Most of my comments are about the existing test.

resources
```diff
@@ -1 +1 @@
-Subproject commit e810600306d0612c609c8daab606229a11061627
+Subproject commit dee0f92b957973785097af3e1b1a5ded9b1e7a94
```
ddorwin: I'm assuming this and the next are unintentional.

encrypted-media/scripts/setmediakeys-multiple-times-with-different-mediakeys.js
```diff
@@ -50,36 +50,47 @@ function runTest(config, qualifier) {
...
+            // Return something so the promise resolves properly.
             return Promise.resolve();
         }), test.step_func(function(error) {
-            fail = true;
+            // Changing the Media Keys object failed
+            _usingMediaKeys2 = false;
             assert_equals(_video.mediaKeys, _mediaKeys1);
+            // The specification allows wither NotSupportedError or InvalidStateError depending on
+            // whether the failure was because changing Media Keys object is not supported
+            // or just not allowed at this time, respectively.
```
ddorwin: Thank you!

encrypted-media/scripts/setmediakeys-multiple-times-with-different-mediakeys.js
```diff
@@ -50,36 +50,47 @@ function runTest(config, qualifier) {
...
+            _usingMediaKeys2 = true;
+            assert_equals(_video2.mediaKeys, _mediaKeys2);
+            // Return something so the promise resolves properly.
             return Promise.resolve();
         }), test.step_func(function(error) {
-            fail = true;
+            // Changing the Media Keys object failed
+            _usingMediaKeys2 = false;
             assert_equals(_video.mediaKeys, _mediaKeys1);
+            // The specification allows wither NotSupportedError or InvalidStateError depending on
```
ddorwin: **e**ither
Same below

ddorwin (approved):
LGTM with typo and reverting subresource change (I assume).

PR #3957, opened 2016-10-10
[eme] Fix assert expectation in setmediakeys.js

ddorwin:


ddorwin:
#3959 is a better fix.

PR #3958, opened 2016-10-10
Add test coverage for the MessageEvent constructor

cdumez:
Add test coverage for the MessageEvent constructor and check that MessageEvent's
ports attribute is not nullable as per:
- https://github.com/whatwg/html/pull/1883
- https://github.com/whatwg/html/issues/1882

Also cover MessageEvent's initMessageEvent operation.

### DIFF
--- a/html/dom/interfaces.html
+++ b/html/dom/interfaces.html
@@ -2628,9 +2628,9 @@ interface MessageEvent : Event {
   readonly attribute DOMString origin;
   readonly attribute DOMString lastEventId;
   readonly attribute (WindowProxy or MessagePort)? source;
-  readonly attribute MessagePort[]? ports;
+  readonly attribute FrozenArray<MessagePort> ports;
 
-  void initMessageEvent(DOMString typeArg, boolean canBubbleArg, boolean cancelableArg, any dataArg, DOMString originArg, DOMString lastEventIdArg, (WindowProxy or MessagePort) sourceArg, MessagePort[]? portsArg);
+  void initMessageEvent(DOMString type, boolean bubbles, boolean cancelable, any data, DOMString origin, DOMString lastEventId, (WindowProxy or MessagePort) source, sequence<MessagePort> ports);
 };
 
 dictionary MessageEventInit : EventInit {
--- a/html/webappapis/scripting/events/messageevent-constructor.html
+++ b/html/webappapis/scripting/events/messageevent-constructor.html
@@ -1,0 +1,78 @@
+<!DOCTYPE html>
+<title>MessageEvent constructor</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+test(function() {
+  var ev = new MessageEvent("test")
+  assert_equals(ev.type, "test", "type attribute")
+  assert_equals(ev.target, null, "target attribute")
+  assert_equals(ev.currentTarget, null, "currentTarget attribute")
+  assert_equals(ev.eventPhase, Event.NONE, "eventPhase attribute")
+  assert_equals(ev.bubbles, false, "bubbles attribute")
+  assert_equals(ev.cancelable, false, "cancelable attribute")
+  assert_equals(ev.defaultPrevented, false, "defaultPrevented attribute")
+  assert_equals(ev.isTrusted, false, "isTrusted attribute")
+  assert_true(ev.timeStamp > 0, "timeStamp attribute")
+  assert_true("initMessageEvent" in ev, "initMessageEvent operation")
+  assert_equals(ev.data, null, "data attribute")
+  assert_equals(ev.origin, "", "origin attribute")
+  assert_equals(ev.lastEventId, "", "lastEventId attribute")
+  assert_equals(ev.source, null, "source attribute")
+  assert_array_equals(ev.ports, [], "ports attribute")
+}, "Default event values")
+
+test(function() {
+  var channel = new MessageChannel()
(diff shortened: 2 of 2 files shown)


cdumez:
@domenic @foolip

cdumez:
Will add coverage for initMessageEvent() as suggested by @foolip

domenic:
LGTM. It might be nice to also test the frozenarray semantics, e.g. test that `msgEvent.ports === msgEvent.ports`, that `Object.isFrozen(msgEvent.ports)` is true, and that `Array.isArray(msgEvent.ports)` is true.

cdumez:
@domenic Could you explain why msgEvent.ports === msgEvent.ports should be true? Since we're not using [SameObject], I thought it'd be fine to return a different object.

cdumez:
I integrated @foolip and @domenic 's comment except for the ev.ports === ev.ports check (for now).

cdumez:
FYI, all parameters to initMessageEvent() are mandatory in Firefox (I tested) and in Edge (according to @foolip), which matches the specification. However, they are all optional in WebKit and Blink.

domenic:
The spec for `ports` says "The ports attribute must return the value it was initialized to." Even though this is apparently using a rather unorthodox definition of "initialized" that also includes being re-set via initMessageEvent, it should still keep returning the same object when initMessageEvent is not called.

In general, the introduction of FrozenArray to the platform was done in order to get this guarantee (that `x.y === x.y`). It gives us something which stays stable (by virtue of being frozen) until it changes (and starts returning a new value).

cdumez:
Ok, I added the ev.ports === ev.ports. Sadly this fails in WebKit even though we are using a FrozenArray because we don't guarantee it returns the same object. We only make sure the returned array is frozen.

domenic:
Awesome. LGTM, will give @foolip a day to sign off as well, or will merge tomorrow otherwise.

html/dom/interfaces.html
```diff
@@ -2628,9 +2628,9 @@
   readonly attribute DOMString origin;
   readonly attribute DOMString lastEventId;
   readonly attribute (WindowProxy or MessagePort)? source;
-  readonly attribute MessagePort[]? ports;
+  readonly attribute FrozenArray<MessagePort> ports;
 
-  void initMessageEvent(DOMString typeArg, boolean canBubbleArg, boolean cancelableArg, any dataArg, DOMString originArg, DOMString lastEventIdArg, (WindowProxy or MessagePort) sourceArg, MessagePort[]? portsArg);
+  void initMessageEvent(DOMString typeArg, boolean canBubbleArg, boolean cancelableArg, any dataArg, DOMString originArg, DOMString lastEventIdArg, (WindowProxy or MessagePort) sourceArg, sequence<MessagePort> portsArg);
```
foolip: Nit: can you change the argument names to match the spec?

html/webappapis/scripting/events/messageevent-constructor.html
```diff
@@ -0,0 +1,78 @@
+<!DOCTYPE html>
+<title>MessageEvent constructor</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<body>
```
foolip: Don't need a body.

html/webappapis/scripting/events/messageevent-constructor.html
```diff
@@ -0,0 +1,78 @@
...
+  assert_equals(ev.data, "testData", "data attribute")
+  assert_equals(ev.origin, "testOrigin", "origin attribute")
+  assert_equals(ev.lastEventId, "testId", "lastEventId attribute")
+  assert_equals(ev.source, window, "source attribute")
+  assert_array_equals(ev.ports, [], "ports attribute")
+}, "MessageEventInit dictionary")
+
+test(function() {
+  assert_throws(new TypeError(), function() {
+    var ev = new MessageEvent("test", { ports: null })
```
foolip: Don't need `var ev =`.

html/webappapis/scripting/events/messageevent-constructor.html
```diff
@@ -0,0 +1,78 @@
...
+}, "Default event values")
+
+test(function() {
+  var ev = new MessageEvent("test", { data: "testData", origin: "testOrigin", lastEventId: "testId", source: window, ports: [] })
+  assert_equals(ev.type, "test", "type attribute")
+  assert_equals(ev.data, "testData", "data attribute")
+  assert_equals(ev.origin, "testOrigin", "origin attribute")
+  assert_equals(ev.lastEventId, "testId", "lastEventId attribute")
+  assert_equals(ev.source, window, "source attribute")
+  assert_array_equals(ev.ports, [], "ports attribute")
```
foolip: Using something other than an empty array would be good, as this would pass by ignoring ports.

html/webappapis/scripting/events/messageevent-constructor.html
```diff
@@ -0,0 +1,78 @@
...
+test(function() {
+  var ev = new MessageEvent("test", { ports: [] })
+  assert_true(Array.isArray(ev.ports), "Array.isArray() should return true")
+  assert_true(Object.isFrozen(ev.ports), "Object.isFrozen() should return true")
+  assert_true(ev.ports === ev.ports, "ev.ports should return the same object")
+}, "ports attribute should be a FrozenArray")
+
+test(function() {
+  var ev = document.createEvent("messageevent");
+  ev.initMessageEvent("test", true, false, "testData", "testOrigin", "testId", window, [])
```
foolip: Here too a non-empty ports array would be nice.

foolip (commented):
LGTM % nits

cdumez:
Included @foolip 's review comments.

foolip:
Thanks @cdumez!

PR #3961, opened 2016-10-11
Fix #3956: Fail gracefully on media source errors

### LINKED ISSUE #3956: [eme] Four tests failing on Chrome with MSE-related uncaught InvalidStateError:
http://www.w3c-test.org/encrypted-media/clearkey-mp4-playback-temporary-clear-encrypted.html,
http://www.w3c-test.org/encrypted-media/clearkey-mp4-playback-temporary-encrypted-clear.html, and the `drm-` equivalents fail on Chrome 53 and 56 with:
`Uncaught InvalidStateError: Failed to execute 'endOfStream' on 'MediaSource': The MediaSource's readyState is not 'open'.`

There is another exception before that in the console that may be related:
```
clearkey-mp4-playback-temporary-encrypted-clear.html:1 Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause().
testmediasource.js:33 Uncaught InvalidStateError: Failed to execute 'endOfStream' on 'MediaSource': The MediaSource's readyState is not 'open'.endOfStream @ testmediasource.js:33
```

Firefox 49 passes the **first** test, but it's not clear that the test is correct.

/cc @mwatson2 @wolenetz @jdsmith3000

mwatson2:
Pass media source errors as rejections in the outer promise flow.

The Media Source does not open until it is attached to a media element. At this point, we can create source buffers and append media. This is when an error could occur. Attaching to a media element is handled by the test flow, not the `testmediasource.js` utility.

I don't see an especially elegant way to handle this, so I am attaching a promise to the MediaSource object that will resolve or reject when appending is done. The test flows then need to be adjusted to include this promise.

### DIFF
--- a/encrypted-media/scripts/onencrypted.js
+++ b/encrypted-media/scripts/onencrypted.js
@@ -8,7 +8,7 @@ function runTest(config) {
     var currentData;
 
     async_test(function (test) {
-          var video = config.video,
+        var video = config.video,
             mediaSource,
             onEncrypted = function (event) {
                 currentData = new Uint8Array(event.initData);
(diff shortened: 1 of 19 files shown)


mwatson2:
Note that this has a side-effect of making the tests more deterministic, since we will wait until appending is complete before calling `video.play()`.

jdsmith3000:
LGTM

PR #3963, opened 2016-10-11
Clean whitespace from incoming URLs

BigBlueHat:
We received reports that tests were breaking due to at least a `container_url` being passed in with untrimmed white space (i.e. " http://..."). This fixes that.

/cc @halindrome @azaroth42 @iherman


halindrome (approved):
Yeah - it is possible the user will input the stuff wrong.  Might as well clean it.

PR #3964, opened 2016-10-11
Change drm-mp4-syntax-mediakeysystemaccess.html to expect TypeError for

cpearce:
things that don't convert to the expeced type.

This matches the current draft EME spec that expects TypeError when calling
navigator.requestMediaKeySystemAccess() with an empty keysystem or empty
MediaKeySystemConfiguration.

We also should expect implementations to throw a TypeError when they can't
convert arguments to the expected types, so adjust the test to expect that.


mwatson2 (approved):

jdsmith3000:
LGTM.  Proceeding with merge.

PR #3967, opened 2016-10-12
Add tests for CEReactions IDL attribute on core HTML objects.

rniwa:
These tests cover: DOMParser, DOMStringMap (dataset), Document, Element, ElementContentEditable, HTMLElement, Range, and Selection interfaces defined in:
https://w3c.github.io/DOM-Parsing/
https://dom.spec.whatwg.org/
https://html.spec.whatwg.org/

Fixed a bug in the two test cases for NamedNodeMap that were trying to remove an attribute that doesn't exist.


domenic:
We should figure out https://github.com/w3c/DOM-Parsing/issues/22 first; did you see my response there @rniwa?

rniwa:
@domenic : Yes, I got rid of the test case for now since testing that the element doesn't get instantiated isn't really CEReactions test, and I didn't want to include it in this PR.

custom-elements/reactions/DOMStringMap.html
```diff
@@ -0,0 +1,96 @@
...
+test(function () {
+    var element = define_new_custom_element(['data-foo']);
+    var instance = document.createElement(element.name);
+    instance.dataset.foo = 'bar';
+    assert_array_equals(element.takeLog().types(), ['constructed', 'attributeChanged']);
+    instance.dataset.foo = 'bar';
+    var logEntries = element.takeLog();
+    assert_array_equals(logEntries.types(), ['attributeChanged']);
+    assert_attribute_log_entry(logEntries.last(), {name: 'data-foo', oldValue: 'bar', newValue: 'bar', namespace: null});
+}, 'setter on DOMStringMap must enqueue an attributeChanged reaction when mutating the value of an observed data attribute to the same value');
```
domenic: Good test. Chrome fails this. The test is correct per spec, and IMO the spec is correct too; anything else would be overcomplicated.

custom-elements/reactions/HTMLElement.html
```diff
@@ -0,0 +1,42 @@
...
+testReflectAttribute('tabIndex', 'tabindex', '0', '1', 'tabIndex on HTMLElement');
+testReflectAttribute('accessKey', 'accesskey', 'a', 'b', 'accessKey on HTMLElement');
+testReflectAttributeWithContentValues('draggable', 'draggable', true, 'true', false, 'false', 'draggable on HTMLElement');
+testReflectAttribute('dropzone', 'dropzone', 'copy', 'move', 'dropzone on HTMLElement');
+testReflectAttribute('contextMenu', 'contextmenu', 'menu1', 'menu2', 'contextMenu on HTMLElement');
+testReflectAttributeWithContentValues('spellcheck', 'spellcheck', true, 'true', false, 'false', 'spellcheck on HTMLElement');
+
+testNodeDisconnector(function (customElement) {
+    customElement.parentNode.innerText = '';
+}, 'innerText on HTMLElement');
```
domenic: /cc @dominiccooney Chrome fails this, I couldn't imagine why. Same for outerText below.

custom-elements/reactions/Range.html
```diff
@@ -42,7 +42,12 @@
     var range = document.createRange();
     range.selectNodeContents(container);
     range.surroundContents(customElement);
-}, 'insertNode on Range');
+}, 'surroundContents on Range');
+
+testParsingMarkup(function (document, markup) {
+    var range = document.createRange();
+    return range.createContextualFragment(markup);
```
domenic: Another interesting Chrome failure @dominiccooney

custom-elements/reactions/Selection.html
```diff
@@ -0,0 +1,31 @@
+<!DOCTYPE html>
+<html>
+<head>
+<title>Custom Elements: CEReactions on Selection interface</title>
+<meta name="author" title="Ryosuke Niwa" href="mailto:rniwa@webkit.org">
+<meta name="assert" content="deleteFromDocument of Selection interface must have CEReactions">
+<meta name="help" content="http://w3c.github.io/selection-api/#selection-interface">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="./resources/reactions.js"></script>
```
domenic: Missing custom-element-helpers.js

domenic (approved):
Extra .DS_Store file; you probably want to add that to your global gitignore.

Overall LGTM after fixing the missing `<script>` include. Ideally @dominiccooney would take a look and get a chance to comment on Chrome's failures. But in general I'm really happy you took the time to do this. It feels like validation of our decision to add [CEReactions]; this level of coordination and testing on such tricky behavior is top-notch for a web spec.

custom-elements/reactions/Selection.html
rniwa: Fixed.

PR #3974, opened 2016-10-13
Synthetic click events and event dispatch

annevk:
This is testing some of the scenarios discussed in
https://github.com/whatwg/dom/issues/325 and specified in
https://github.com/whatwg/dom/pull/342.

### DIFF
--- a/dom/events/Event-dispatch-click.html
+++ b/dom/events/Event-dispatch-click.html
@@ -1,0 +1,119 @@
+<!doctype html>
+<title>Synthetic click event "magic"</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div id=log></div>
+<div id=dump style=display:none></div>
+<script>
+var dump = document.getElementById("dump")
+
+async_test(function(t) {
+  var input = document.createElement("input")
+  input.type = "checkbox"
+  dump.appendChild(input)
+  input.onclick = t.step_func_done(function() {
+    assert_true(input.checked)
+  })
+  input.click()
+}, "basic with click()")
+
+async_test(function(t) {
+  var input = document.createElement("input")
+  input.type = "checkbox"
+  dump.appendChild(input)
+  input.onclick = t.step_func_done(function() {
+    assert_true(input.checked)
+  })
+  input.dispatchEvent(new MouseEvent("click", {bubbles:true})) // equivalent to the above
+}, "basic with dispatchEvent()")
+
+async_test(function(t) {
+  var input = document.createElement("input")
+  input.type = "checkbox"
+  dump.appendChild(input)
+  input.onclick = t.step_func_done(function() {
+    assert_false(input.checked)
+  })
+  input.dispatchEvent(new Event("click", {bubbles:true})) // no MouseEvent
+}, "basic with wrong event class")
+
+async_test(function(t) {
+  var input = document.createElement("input")
+  input.type = "checkbox"
+  dump.appendChild(input)
+  var child = input.appendChild(new Text("does not matter"))
+  child.dispatchEvent(new MouseEvent("click")) // does not bubble
+  assert_false(input.checked)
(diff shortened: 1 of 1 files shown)


foolip:
@dtapuska can you review, and perhaps contribute tests from Blink in this area if there are any?

dtapuska:
Additional tests that we have that kind of related are:

https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/fast/events/stopPropagation-checkbox.html?q=stopPropagation-check&sq=package:chromium&l=1
https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/fast/events/simulated-click-disabled.html?q=simulated-click-dis&sq=package:chromium&l=1
https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/fast/events/programmatic-check-no-change-event.html?q=programmatic-chec&sq=package:chromium&l=1

foolip:
As a general point, I wonder if these tests should all be sync and more explicitly verify that the behavior indeed happens synchronously. Some of the tests have `t.done()` at the end of the outermost function, and nothing that happens after that point affects the results, so it's kind of implied (but far from obvious) that the `t.step_func`'s run before that point.

annevk:
They cannot use `test()` due to event handlers using "report the exception". Would result in uncaught exceptions all over. (Actually did when I first wrote these.)

foolip:
Oh, so exceptions thrown in the event handler wouldn't fail the right test, I see.

dtapuska:
annevk@ the blink ones test a disabled control and that stopPropagation doesn't change activation behaviour.   Can you add those to your test?

foolip:
I've filed https://bugs.chromium.org/p/chromium/issues/detail?id=656976 to get those tests into web-platform-tests. If it's not done by a Blink engineer, we'll probably end up with duplicated tests. It's still too hard to work with web-platfom-tests in Blink, but when it's easier in Blink and elsewhere, I think it'll be reasonable to ask implementers to volunteer to write the web-platform-tests changes more often than not.

dom/events/Event-dispatch-click.html
```diff
@@ -0,0 +1,94 @@
...
+}, "basic with wrong event class")
+async_test((t) => {
+  var input = document.createElement("input")
+  input.type = "checkbox"
+  dump.appendChild(input)
+  var child = input.appendChild(new Text("does not matter"))
+  child.dispatchEvent(new MouseEvent("click")) // does not bubble
+  assert_false(input.checked)
+  t.done()
+}, "only look at parents when event bubbles")
```
zcorpan: This test name is confusing to me; it reads like the event target should be ignored when the event bubbles. How about "ignore parent when event does not bubble"?

dom/events/Event-dispatch-click.html
```diff
@@ -0,0 +1,94 @@
...
+<div id=dump style=display:none></div>
+<script>
+var dump = document.getElementById("dump")
+async_test((t) => {
+  var input = document.createElement("input")
+  input.type = "checkbox"
+  dump.appendChild(input)
+  input.onclick = t.step_func_done(() => assert_true(input.checked))
+  input.click()
+}, "basic with click()")
```
zcorpan: Please add a blank line between tests

zcorpan (changes requested):

dom/events/Event-dispatch-click.html
zcorpan: I mean I read it as the "only" refers to the parents, rather than the bubbles. (In this test the event doesn't bubble.) Anyway, not a critical issue :-)

annevk: I could move "only" to just before "when" I suppose.

annevk:
@dtapuska I assume @foolip took care of your comment by filing that follow up issue.

zcorpan (approved):

PR #3975, opened 2016-10-13
Add a travis job for testing PR stability

jgraham:
The purpose of the job is to identify poorly-written, unstable, tests as
early in the cycle as possible. To this end we use wptrunner to run all
the changed tests in the PR in Firefox. If 10 runs of the tests don't
all give a consisent result then the job fails.

Adding runs in Chrome should be quite straightforward, adding Edge and
Safari may be more difficult because of their strict platform dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,48 @@
-sudo: false # cause Travis to start builds much faster
+dist: trusty
 language: python
-python:
-  - "2.7"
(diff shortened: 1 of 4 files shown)


jgraham:
@Ms2ger r?

Ms2ger:
Go ahead and merge.

PR #3976, opened 2016-10-13
Import one script test for fractions from the "MathML in HTML5" tests

fred-wang:
@jgraham: Following up on our discussion, this is a first commit for https://github.com/MathML/MathMLinHTML5-tests/issues/1

The tests are under a compatible license:
https://github.com/MathML/MathMLinHTML5-tests/blob/master/LICENSE

I'm starting with the frac-parameters-1.html test, since it is a good way to decide the locations of the various files (tools, fonts, mathml tests etc). This test checks the parameters described in http://www.mathml-association.org/MathMLinHTML5/S3.html#SS3.SSS2

FYI, this test passes on Gecko, WebKit as well as Igalia's Chromium MathML branch. It has also already been imported into WebKit's test suite and verified by WebKit reviewers: https://trac.webkit.org/browser/trunk/LayoutTests/imported/mathml-in-html5/mathml/presentation-markup/fractions/frac-parameters-1.html (Gecko has a similar test, but I don't remember whether we imported that one yet)

Note however that WebKit has issues with the Javascript web fonts API, so the imported test has some hack to delay a bit the execution of the tests: https://trac.webkit.org/browser/trunk/LayoutTests/imported/mathml-in-html5/mathml/presentation-markup/fractions/frac-parameters-1.html#L64 ; this hack is not present in the original test.

Any comment is welcome!


fred-wang:
I forgot to mention that I verified execution with the local test runner and that seems to work.

jgraham:
Seems good to me.

PR #3977, opened 2016-10-14
Fix #3521: changing the expected presentation connection state into "connecting"

### LINKED ISSUE #3521: `PresentationRequest.start` returns a presentation in a `connecting` state
@louaybassbouss, @mariuswsl, @mfoltzgoogle

While it seems to pass on Chromium, the basic Presentation API test that checks the call to `start` assumes that the `PresentationConnection` instance will be a `connected` state when the `Promise` is resolved, which I believe is incorrect:
https://github.com/w3c/web-platform-tests/blob/master/presentation-api/controlling-ua/startNewPresentation_success-manual.html#L35

If I'm reading the spec correctly, the instance will rather be in a `connecting` state:

1. The `Promise` is resolved in step 19 of the [start a presentation](http://w3c.github.io/presentation-api/#starting-a-presentation) algorithm when the presentation connection state is `connecting`. Resolving a `Promise` typically means *queueing a task* (t1) to run the `then` function.
2. The [establishment of the connection](http://w3c.github.io/presentation-api/#dfn-establish-a-presentation-connection) occurs after that and *queues another task* (t2) to change the state to `connected`.
3. t1 is queued before t2, and thus runs before as well.

In short, I think the line should be changed to:
```javascript
assert_equals(connection.state, "connecting", "The presentation is in a connecting state.");
```

Alternatively, the test could wait for the `connect` event on the `PresentationConnection` to check that the connection gets established.

tomoyukilabs:
This PR fixes #3521. Both controlling-ua/startNewPresentation_success-manual.html and controlling-ua/defaultRequest_success-manual.html are modified so that they examine if the initial state of presentation connection is `connecting` or not.

This PR also contains the following additional modifications:
- A test for `PresentationRequest.onconnectionavailable` is added.
- `defaultRequest` test will start via a user gesture on browser's user interface (i.e. "Cast..." menu in the case of Google Chrome).

### DIFF
--- a/presentation-api/controlling-ua/PresentationRequest_onconnectionavailable-manual.html
+++ b/presentation-api/controlling-ua/PresentationRequest_onconnectionavailable-manual.html
@@ -7,10 +7,13 @@
 <script src="/resources/testharnessreport.js"></script>
 
 <p>Click the button below and select the available presentation display, to start the manual test.</p>
-<button id="presentBtn" onclick="startPresentation()">Start Presentation Test</button>
+<button id="presentBtn">Start Presentation Test</button>
 
 
 <script>
+    // disable the timeout function for the tests
+    setup({explicit_timeout: true});
+
     // ----------
     // DOM Object
     // ----------
@@ -27,10 +30,21 @@
     // Start New PresentationRequest.onconnectionavailable Test (success) - begin
     // --------------------------------------------------------------------------
     var startPresentation = function () {
-        async_test(function (t) {
-            request.onconnectionavailable = t.step_func_done(function (evt) {
-                var connection = evt.connection;
-
+        promise_test(function (t) {
+            // Note: During starting a presentation, the connectionavailable event is fired (step 20)
+            //       after the promise P is resolved (step 19).
+            return new Promise(function(resolve, reject) {
+                request.onconnectionavailable = function (evt) {
+                    resolve(evt.connection);
+                };
+                request.start().then(function() {
+                    // This test fails if request.onconnectionavailable is not invoked regardless of starting the presentation
+                    t.step_timeout(function() { reject('The connectionavailable was not fired.') }, 5000);
+                }, function() {
+                    // This test fails if the presentation is not started successfully
+                    reject('The presentation failed to be started.');
+                });
+            }).then(function(connection) {
                 // Check the initial state of a presentation connection
                 assert_equals(connection.state, 'connecting', 'The initial state of the presentation connection is "connecting".');
 
@@ -43,10 +57,9 @@
                 // Check the instance of the connection
                 assert_true(connection instanceof PresentationConnection, 'The connection is an instance of PresentationConnection.');
             });
-
-            request.start();
-        }, 'The presentation was started successfully.');
+        }, 'The connectionavailable event was fired successfully.');
     }
+    presentBtn.onclick = startPresentation;
     // ------------------------------------------------------------------------
     // Start New PresentationRequest.onconnectionavailable Test (success) - end
     // ------------------------------------------------------------------------
--- a/presentation-api/controlling-ua/PresentationRequest_onconnectionavailable-manual.html
+++ b/presentation-api/controlling-ua/PresentationRequest_onconnectionavailable-manual.html
@@ -37,13 +37,11 @@
                 request.onconnectionavailable = function (evt) {
                     resolve(evt.connection);
                 };
+                // This test fails if request.onconnectionavailable is not invoked although the presentation is started successfully
+                // or the presentation fails to be started
                 request.start().then(function() {
-                    // This test fails if request.onconnectionavailable is not invoked although the presentation is started successfully
-                    t.step_timeout(function() { reject('The connectionavailable was not fired.') }, 5000);
-                }, function() {
-                    // This test fails if the presentation is not started successfully
-                    reject('The presentation failed to be started.');
-                });
+                    t.step_timeout(function() { assert_unreached('The connectionavailable event was not fired.'); }, 5000);
+                }, reject);
             }).then(function(connection) {
                 // Check the initial state of a presentation connection
                 assert_equals(connection.state, 'connecting', 'The initial state of the presentation connection is "connecting".');
--- a/presentation-api/controlling-ua/PresentationRequest_onconnectionavailable-manual.html
+++ b/presentation-api/controlling-ua/PresentationRequest_onconnectionavailable-manual.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <meta charset="utf-8">
-<title>Presentation.onconnectionavailable (manual test)</title>
+<title>PresentationRequest.onconnectionavailable (manual test)</title>
 <link rel="author" title="Tomoyuki Shimizu" href="https://github.com/tomoyukilabs">
 <link rel="help" href="https://w3c.github.io/presentation-api/#starting-a-presentation">
 <script src="/resources/testharness.js"></script>
--- a/presentation-api/controlling-ua/PresentationRequest_onconnectionavailable-manual.html
+++ b/presentation-api/controlling-ua/PresentationRequest_onconnectionavailable-manual.html
@@ -32,24 +32,16 @@
                 var connection = evt.connection;
 
                 // Check the initial state of a presentation connection
-                test(function() {
-                    assert_equals(connection.state, 'connecting');
-                }, 'The initial state of the presentation connection is "connecting".');
+                assert_equals(connection.state, 'connecting', 'The initial state of the presentation connection is "connecting".');
 
                 // Check, if the connection ID is set
-                test(function () {
-                    assert_true(!!connection.id);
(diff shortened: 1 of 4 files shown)


presentation-api/controlling-ua/PresentationRequest_onconnectionavailable-manual.html
```diff
@@ -0,0 +1,62 @@
...
+    // --------------------------------------------------------------------------
+    // Start New PresentationRequest.onconnectionavailable Test (success) - begin
+    // --------------------------------------------------------------------------
+    var startPresentation = function () {
+        async_test(function (t) {
+            request.onconnectionavailable = t.step_func_done(function (evt) {
+                var connection = evt.connection;
+
+                // Check the initial state of a presentation connection
+                test(function() {
```
tidoust: I haven't checked, but I'm not sure it's good practice to nest calls to `async_test` and `test`, especially because it means the number of test outcomes that appear in the final result depends on the context. For instance, here, if a timeout occurs before the `connectionavailable` event is fired, only the async test line will appear as `FAIL`. The other tests will simply not show up.

I would drop the `test` lines here and rather complete the `assert_xxx` statements with the failure reason as in:
`assert_equals(connection.state, 'connecting', 'The initial state of the presentation connection should be "connecting".');`

presentation-api/controlling-ua/defaultRequest_success-manual.html
```diff
@@ -2,12 +2,13 @@
...
 <link rel="author" title="Marius Wessel" href="http://www.fokus.fraunhofer.de">
+<link rel="author" title="Tomoyuki Shimizu" href="https://github.com/tomoyukilabs">
 <link rel="help" href="http://w3c.github.io/presentation-api/#dfn-controlling-user-agent">
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 
-<p>Click the button below and select the available casting device, to start the manual test.</p>
-<button onclick="startPresentation()">Start Presentation Test</button>
+<p>Click the button or the menu item for presentation on your browser (for example, "Cast"), to start the manual test.</p>
+<p>If your browser does not support <code>defaultRequest</code>, this test will not begin.</p>
```
tidoust: The test outcome should get reported to the test runner no matter what. There is no way to represent an optional test for the time being, I think. Last time I checked, the preferred outcome was to have the test FAIL. We could probably also prefix the test's label with `[Optional]` so that it's easy to spot that the failure may simply indicate lack of support for the feature.

To have the test fail, what should the user do? I suspect that if she just closes the window, she will have to state whether the test succeeded or failed to the test runner. It may be better to add a button that makes the test fail and instruct the user to press that button if the browser does not support `defaultRequest`.

That's open for discussion. I'm not saying this is the right approach. It's certainly not perfect :)

presentation-api/controlling-ua/defaultRequest_success-manual.html
```diff
@@ -22,29 +23,31 @@
...
-
-        }, "The presentation was started successfully.");
-
-        done();
-    };
+    async_test(function(t) {
+        navigator.presentation.defaultRequest.onconnectionavailable = t.step_func_done(function (evt) {
+            var connection = evt.connection;
+
+            test(function() {
```
tidoust: Same comment on nested calls to test as above.

presentation-api/controlling-ua/startNewPresentation_success-manual.html
```diff
@@ -31,8 +32,10 @@
             return request.start()
                     .then(function (connection) {
 
-                        // assert case for the promise_test
-                        assert_equals(connection.state, "connected", "The presentation has an connected state.");
+                        // Check the initial state of a presentation connection
+                        test(function() {
```
tidoust: Similarly, I would rewrite this test with an enclosing `async_test` call and no `test` calls to ensure consistency between a PASS and a FAIL outcome.

tidoust (commented):
See suggested changes. I realize that some of them (nesting of test calls typically) applied to the initial tests as well, we could address them in a later PR if needed.

tomoyukilabs:
@tidoust Thanks for your review!

> I would drop the `test` lines here and rather complete the `assert_xxx` statements with the failure reason

Okay, I removed nested `test()` and moved failure reason to `assert_xxx`.

> We could probably also prefix the test's label with `[Optional]` so that it's easy to spot that the failure may simply indicate lack of support for the feature.

I added `[Optional]` prefix to the test name.

> It may be better to add a button that makes the test fail and instruct the user to press that button if the browser does not support `defaultRequest`.

For now, I added the button to make the test fail. Since the current spec does not provide any hint about availability of `defaultRequest` for web apps, I cannot imagine any solution but using such a button...

tidoust:
Thanks. That all looks good to me. I'll let @louaybassbouss take a final look and merge this.

presentation-api/controlling-ua/PresentationRequest_onconnectionavailable-manual.html
```diff
@@ -0,0 +1,54 @@
...
+<link rel="help" href="https://w3c.github.io/presentation-api/#starting-a-presentation">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<p>Click the button below and select the available presentation display, to start the manual test.</p>
+<button id="presentBtn" onclick="startPresentation()">Start Presentation Test</button>
+
+
+<script>
+    // ----------
```
tidoust: Since this is a manual test, there needs to be a call somewhere to:
`setup({explicit_timeout: true});`

The test harness will time out after a few seconds otherwise. Note that one could argue that the test harness should automatically disable the timeout for tests that have "-manual" in their name. That's note the case today, though, so we need to do it explicitly in the code.

(I note I missed that in previous tests too... We'll fix other tests later on, no big deal)

presentation-api/controlling-ua/PresentationRequest_onconnectionavailable-manual.html
```diff
@@ -0,0 +1,54 @@
...
+                assert_true(!!connection.id, 'The connection ID is set.');
+
+                // Check the type of the connection.id
+                assert_true(typeof connection.id === 'string', 'The connection ID is a string.');
+
+                // Check the instance of the connection
+                assert_true(connection instanceof PresentationConnection, 'The connection is an instance of PresentationConnection.');
+            });
+
+            request.start();
```
tidoust: Another problem here is that there's no code to handle the case where the call to `start` fails. The Promise will be rejected but that will go unnoticed. To fix this, `promise_test` should be used instead of `async_test`, and the Promise should be returned as in:

```js
promise_test(function (t) {
  request.onconnectionavailable = [...];
  return request.start();
});
```

tidoust (commented):
@tomoyukilabs Reviewing the tests for other reasons (*), I realize that I missed a couple of things in my previous review. See inline.

(*) I'm preparing the migration of presentation URLs to a shared config file to avoid hardcoding the Cast app ID in each and every file.

tomoyukilabs:
Thanks, @tidoust

> Since this is a manual test, there needs to be a call somewhere to:
`setup({explicit_timeout: true});`

Sorry, I forgot adding this code. I added this to the tests.

> Another problem here is that there's no code to handle the case where the call to `start` fails.

I changed `async_test` into `promise_test`, and added codes to handle the two cases:
* the case where the call to `start` fails
* the case where `connectionavailable` event is not fired even though the call to `start` does not fail

presentation-api/controlling-ua/PresentationRequest_onconnectionavailable-manual.html
```diff
@@ -0,0 +1,67 @@
...
+            // Note: During starting a presentation, the connectionavailable event is fired (step 20)
+            //       after the promise P is resolved (step 19).
+            return new Promise(function(resolve, reject) {
+                request.onconnectionavailable = function (evt) {
+                    resolve(evt.connection);
+                };
+                request.start().then(function() {
+                    // This test fails if request.onconnectionavailable is not invoked although the presentation is started successfully
+                    t.step_timeout(function() { reject('The connectionavailable was not fired.') }, 5000);
+                }, function() {
```
tidoust: I may be nitpicking here, but wouldn't it be better to replace `function() { reject(...); }` by "reject"?

```js
request.start().then(function() { ... }, reject);
```

This would expose the actual reason for failure in the test result in particular (`NotAllowedError`, `NotFoundError`, etc.) which would help with debugging.

tidoust (commented):
Thanks, @tomoyukilabs 

> the case where connectionavailable event is not fired although the call to start does not fail

Good catch. I had not thought about that either :)

One last comment on my side.

tidoust:
OK, I think we've done enough back and forth on this one ;)
@louaybassbouss I'll merge it now so that I can work on the latest version of the test suite for the Cast URL update. Let's use another PR if further fixes are needed.

tomoyukilabs:
Thanks for reviewing, @tidoust!

PR #3981, opened 2016-10-15
Update javascript: URL return value tests

domenic:
See https://github.com/whatwg/html/issues/1896.

@bzbarsky to review. Spec patch incoming.


domenic:
This should not be merged until we figure out whatwg/html#1896, actually. It's potentially possible that we won't go the same route Blink/WebKit do.

Ms2ger:
Please ping me when this is ready to land

bzbarsky:
@ayg Why did you close this?

PR #3984, opened 2016-10-16
Avoid leak of global variable in dir attribute tests.

cscott:
Undeclared global `x` causes issues when running these tests under mocha.

### DIFF
--- a/html/dom/elements/global-attributes/dir_auto-contained-script-R-ref.html
+++ b/html/dom/elements/global-attributes/dir_auto-contained-script-R-ref.html
@@ -28,6 +28,7 @@
         display: none;
       }
     </style>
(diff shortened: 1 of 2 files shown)


domenic:
-1. These tests are not meant to run under Mocha and it is not a goal of this project to support such scenarios.

sideshowbarker:
Closing per the rationale given in https://github.com/w3c/web-platform-tests/pull/3984#issuecomment-254092663. It’s imaginable we have other tests in the suite that’ll have issues when trying to run them under Mocha. It’s not clear what purpose there is to running the tests under Mocha, but if there’s some good argument for supporting it, then it would be better raised as separate issue and discussed there.

PR #3991, opened 2016-10-17
Change ol.start reflection default

domenic:
This updates the tests for ol.start to match https://github.com/whatwg/html/pull/1916. The relevant tests pass in Firefox and, mostly, in Edge. (Edge has issues parsing "7e2" and does not support the .reversed IDL attribute, but it passes the tests for .start.)

Let's not merge this until https://github.com/whatwg/html/pull/1916 is merged, just in case people do not like this approach.

### DIFF
--- a/html/semantics/grouping-content/the-ol-element/grouping-ol.html
+++ b/html/semantics/grouping-content/the-ol-element/grouping-ol.html
@@ -247,32 +247,32 @@
 
         // basicRevGoodName - default should be 3
         test(function() {
-            assert_equals(document.getElementById("basicRevGoodName").start, 3, "no start attribute provided -> 3");
-        }, "Default start value (if none provided) for reversed list = child li elements.");
+            assert_equals(document.getElementById("basicRevGoodName").start, 1, "no start attribute provided -> 3");
+        }, "Default start value (if none provided) for reversed list = 1.");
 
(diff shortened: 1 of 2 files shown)


html/semantics/grouping-content/the-ol-element/grouping-ol.html
```diff
@@ -247,32 +247,32 @@
 
         // basicRevGoodName - default should be 3
         test(function() {
-            assert_equals(document.getElementById("basicRevGoodName").start, 3, "no start attribute provided -> 3");
-        }, "Default start value (if none provided) for reversed list = child li elements.");
+            assert_equals(document.getElementById("basicRevGoodName").start, 1, "no start attribute provided -> 3");
```
zcorpan: Update or remove the message argument (here and below)

html/semantics/grouping-content/the-ol-element/grouping-ol.html
```diff
@@ -247,32 +247,32 @@
 
         // basicRevGoodName - default should be 3
```
zcorpan: Update or remove the "should be 3" (here and below)

zcorpan (changes requested):

zcorpan (approved):

PR #3997, opened 2016-10-17
Fix #3993: Remove unused script playback-temporary-multikey-multisess…

### LINKED ISSUE #3993: [eme] Why does playback-temporary-multikey-multisession.js wait for pause, waiting, and stalled?
`playback-temporary-multikey-multisession.js` currently has the following `playing` event handler:
```
        function onPlaying(event) {
            consoleWrite("playing");
            waitForEventAndRunStep('pause', _video, onStopped, test);
            waitForEventAndRunStep('waiting', _video, onStopped, test);
            waitForEventAndRunStep('stalled', _video, onStopped, test);
        }
```
`onStopped()` generates a new session if necessary.

Is this supposed to handle the case where playback is blocked waiting for a key? If so:
 * Of those three events, it appears that only `waiting` could possibly be fired as a result (see [Event Summary](https://html.spec.whatwg.org/multipage/embedded-content.html#mediaevents)).
 * Shouldn't this be triggered by the `waitingforkey` event?

mwatson2:
…ion.js


ddorwin (approved):
Thanks :)

PR #3998, opened 2016-10-18
Import tests for parsing and ugprading custom elements from Webkit

rniwa:
These tests have been reviewed and landed into WebKit: https://trac.webkit.org/browser/trunk/LayoutTests/fast/custom-elements
Almost all test casess pass on both WebKit and Blink.

Both test cases in parser-uses-constructed-element.html fails on Blink because Blink erroneously falls back to creating HTMLUnknownElement when a custom element constructor returns an element different from the one returned by the first call to HTMLElement constructor.

The last test case in parser-uses-registry-of-owner-document.html fails on Blink because Blink erroneously creates a non-HTMLElement Element inside a document created by document.implementation.createDocument with HTML namespace.

### DIFF
--- a/custom-elements/parser/parser-constructs-custom-element-synchronously.html
+++ b/custom-elements/parser/parser-constructs-custom-element-synchronously.html
@@ -15,14 +15,14 @@
 
 var childElementCountInConstructor;
 var containerChildNodesInConstructor = [];
-var containerNextSilbingInConstructor;
+var containerNextSiblingInConstructor;
 class MyCustomElement extends HTMLElement {
     constructor() {
         super();
         var container = document.getElementById('custom-element-container');
         for (var i = 0; i < container.childNodes.length; i++)
             containerChildNodesInConstructor.push(container.childNodes[i]);
-        containerNextSilbingInConstructor = container.nextSibling;
+        containerNextSiblingInConstructor = container.nextSibling;
     }
 };
 customElements.define('my-custom-element', MyCustomElement);
@@ -42,7 +42,7 @@ test(function () {
     assert_equals(containerChildNodesInConstructor[0], instance.parentNode.firstChild);
     assert_equals(containerChildNodesInConstructor[1], document.getElementById('custom-element-previous-element'));
     assert_equals(containerChildNodesInConstructor[2], instance.previousSibling);
-    assert_equals(containerNextSilbingInConstructor, null);
+    assert_equals(containerNextSiblingInConstructor, null);
 
 }, 'HTML parser must only append nodes that appear before a custom element before instantiating the custom element');
 
--- a/custom-elements/parser/parser-constructs-custom-element-synchronously.html
+++ b/custom-elements/parser/parser-constructs-custom-element-synchronously.html
@@ -1,0 +1,51 @@
+<!DOCTYPE html>
+<html>
+<head>
+<title>Custom Elements: Changes to the HTML parser</title>
+<meta name="author" title="Ryosuke Niwa" href="mailto:rniwa@webkit.org">
+<meta name="assert" content="HTML parser must construct a custom element synchronously">
+<link rel="help" href="https://html.spec.whatwg.org/#create-an-element-for-the-token">
+<link rel="help" href="https://dom.spec.whatwg.org/#concept-create-element">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+</head>
+<body>
+<div id="log"></div>
+<script>
+
+var childElementCountInConstructor;
+var containerChildNodesInConstructor = [];
+var containerNextSilbingInConstructor;
+class MyCustomElement extends HTMLElement {
+    constructor() {
+        super();
+        var container = document.getElementById('custom-element-container');
+        for (var i = 0; i < container.childNodes.length; i++)
+            containerChildNodesInConstructor.push(container.childNodes[i]);
+        containerNextSilbingInConstructor = container.nextSibling;
+    }
+};
+customElements.define('my-custom-element', MyCustomElement);
+
+</script>
+<div id="custom-element-container">
+    <span id="custom-element-previous-element"></span>
+    <my-custom-element></my-custom-element>
+    <div id="custom-element-next-element"></div>
+</div>
+<script>
+
+test(function () {
+    var instance = document.querySelector('my-custom-element');
+
+    assert_equals(containerChildNodesInConstructor.length, 3);
+    assert_equals(containerChildNodesInConstructor[0], instance.parentNode.firstChild);
+    assert_equals(containerChildNodesInConstructor[1], document.getElementById('custom-element-previous-element'));
+    assert_equals(containerChildNodesInConstructor[2], instance.previousSibling);
+    assert_equals(containerNextSilbingInConstructor, null);
+
+}, 'HTML parser must only append nodes that appear before a custom element before instantiating the custom element');
+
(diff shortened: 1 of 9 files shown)


custom-elements/parser/parser-sets-attributes-and-children.html
```diff
@@ -0,0 +1,64 @@
...
+test(function () {
+    assert_equals(customElement.getAttribute('id'), 'custom-element-id', 'HTML parser must preserve the id attribute');
+    assert_equals(customElement.id, 'custom-element-id', 'HTML parser must preserve the semantics of reflect for the id attribute');
+    assert_equals(customElement.getAttribute('class'), 'class1 class2', 'HTML parser must preserve the class attribute');
+    assert_equals(customElement.classList.length, 2, 'HTML parser must initialize classList on custom elements');
+    assert_equals(customElement.classList[0], 'class1', 'HTML parser must initialize classList on custom elements');
+    assert_equals(customElement.classList[1], 'class2', 'HTML parser must initialize classList on custom elements');
+
+    assert_equals(customElement.childNodes.length, 2, 'HTML parser must append child nodes');
+    assert_equals(customElement.classList[0], 'class1', 'HTML parser must initialize classList on custom elements');
```
domenic: Copypasta? Can probably just be removed?

custom-elements/parser/parser-uses-constructed-element.html
```diff
@@ -0,0 +1,75 @@
+<!DOCTYPE html>
+<html>
+<head>
+<title>Custom Elements: Changes to the HTML parser</title>
```
domenic: Not sure why the thing being tested is hidden in `<meta name="assert">`, while multiple test files have the same `<title>`... I'd prefer to use `<title>` for this. But metadata nits do not really matter.

custom-elements/parser/parser-uses-registry-of-owner-document.html
```diff
@@ -0,0 +1,98 @@
...
+<script src="/resources/testharnessreport.js"></script>
+</head>
+<body>
+<div id="log"></div>
+<script>
+
+class MyCustomElement extends HTMLElement { };
+customElements.define('my-custom-element', MyCustomElement);
+
+document.write('<template><my-custom-element></my-custom-element></template>');
```
domenic: Any reason why this test file uses document.write, instead of the style in the other files where you intersperse script and non-script? Just curious. It's probably good coverage to have it this way.

custom-elements/parser/parser-uses-registry-of-owner-document.html
```diff
@@ -0,0 +1,98 @@
...
+test(function () {
+    var windowlessDocument = document.implementation.createDocument ('http://www.w3.org/1999/xhtml', 'html', null);
+    windowlessDocument.documentElement.innerHTML = '<my-custom-element></my-custom-element>';
+
+    var instance = windowlessDocument.querySelector('my-custom-element');
+    assert_true(instance instanceof HTMLElement);
+    assert_false(instance instanceof MyCustomElement);
+
+}, 'HTML parser must use the registry of window.document in a document created by document.implementation.createXHTMLDocument()');
+
```
domenic: Maybe add a test for `new Document()` as well? I know you used that in other test cases though so maybe it's covered implicitly by them.

domenic: Maybe add a test case for XHR responseType = document documents?

custom-elements/upgrading/Node-cloneNode.html
```diff
@@ -0,0 +1,205 @@
...
+<meta name="assert" content="Node.prototype.cloneNode should upgrade a custom element">
+<link rel="help" href="https://html.spec.whatwg.org/#upgrades">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+</head>
+<body>
+<div id="log"></div>
+<script>
+
+function withNewDocumentWithABrowsingContext(callback) {
```
domenic: This seems a little redundant with @dominiccooney's helper that he uses (I think it's called window_test or something?). It might be nice to reuse that.

custom-elements/upgrading/Node-cloneNode.html
```diff
@@ -0,0 +1,205 @@
...
+        var containerClone = container.cloneNode(true);
+
+        assert_equals(parentNodeInConstructor, containerClone,
+            'An upgraded element must have its parentNode set before the custom element constructor is called');
+        assert_equals(previousSiblingInConstructor, containerClone.firstChild,
+            'An upgraded element must have its previousSibling set before the custom element constructor is called');
+        assert_equals(nextSiblingInConstructor, containerClone.lastChild,
+            'An upgraded element must have its nextSibling set before the custom element constructor is called');
+    });
+}, 'Node.prototype.cloneNode(true) must be able to clone a descendent custom element');
```
domenic: Duplicate test name

custom-elements/parser/parser-constructs-custom-element-synchronously.html
```diff
@@ -0,0 +1,51 @@
...
+var childElementCountInConstructor;
+var containerChildNodesInConstructor = [];
+var containerNextSilbingInConstructor;
+class MyCustomElement extends HTMLElement {
+    constructor() {
+        super();
+        var container = document.getElementById('custom-element-container');
+        for (var i = 0; i < container.childNodes.length; i++)
+            containerChildNodesInConstructor.push(container.childNodes[i]);
+        containerNextSilbingInConstructor = container.nextSibling;
```
domenic: typo "sibling"

domenic (approved):
Pretty much all nits. I've confirmed the failing-in-Blink tests are correct per spec (i.e. that it's a Blink bug).

custom-elements/parser/parser-uses-registry-of-owner-document.html
rniwa: Yeah, I'm just trying to use different styles of writing tests here so that document.write is also adequately tested.

custom-elements/parser/parser-uses-registry-of-owner-document.html
rniwa: Added a test case (obviously Chrome fails this newly added test case).

rniwa: Done that.

custom-elements/upgrading/Node-cloneNode.html
rniwa: Unfortunately, this variant removes the iframe it added before ending the test so the semantics is quite different. I moved this to custom-elements-helpers.js so that we can reuse it elsewhere.

rniwa: Actually, it DOES add a cleanup to remove it so I should be able to reuse that.

custom-elements/parser/parser-uses-registry-of-owner-document.html
```diff
@@ -93,6 +93,34 @@
...
+    windowlessDocument.documentElement.innerHTML = '<my-custom-element></my-custom-element>';
+
+    var instance = windowlessDocument.querySelector('my-custom-element');
+
+    assert_true(instance instanceof Element);
+    assert_false(instance instanceof MyCustomElement);
+
+}, 'HTML parser must use the registry of window.document in a document created by new Document');
+
+test(function () {
```
rniwa: Oops, fixed.

PR #3999, opened 2016-10-18
Fix TrackEvent constuctor test

cdumez:
Fix TrackEvent constuctor test so that we pass a valid Track object in the dictionary.
We would previous pass {} which would throw in all browsers because they expect the
type: "(VideoTrack or AudioTrack or TextTrack)?".

### DIFF
--- a/html/semantics/embedded-content/media-elements/interfaces/TrackEvent/constructor.html
+++ b/html/semantics/embedded-content/media-elements/interfaces/TrackEvent/constructor.html
@@ -13,12 +13,13 @@ test(function(){
     assert_equals(ev.track, null, 'ev.track after assignment');
 }, document.title+', one arg');
 test(function(){
-    var obj = {};
-    var ev = new TrackEvent('foo', {track:obj});
+    var video = document.createElement('video');
+    var testTrack = video.addTextTrack('subtitles', 'foo', 'foo');
+    var ev = new TrackEvent('foo', {track: testTrack});
     assert_true(ev instanceof TrackEvent, 'ev instanceof TrackEvent');
     assert_true(ev instanceof Event, 'ev instanceof Event');
-    assert_equals(ev.track, obj, 'ev.track');
+    assert_equals(ev.track, testTrack, 'ev.track');
     ev.track = {};
-    assert_equals(ev.track, obj, 'ev.track after assignment');
+    assert_equals(ev.track, testTrack, 'ev.track after assignment');
 }, document.title+', two args');
 </script>


sideshowbarker (commented):
Reviewed against [spec/WebIDL requirements](https://html.spec.whatwg.org/multipage/embedded-content.html#trackevent) & tested in Firefox, Chromium, & Safari; as noted, without this fix they all throw, & with the fix, they pass

sideshowbarker (approved):

sideshowbarker:
cc @foolip 
Looking back at this test, the fix is still right but I think at this point this and probably all the other interface tests in the [html/semantics/embedded-content/media-elements/interfaces](https://github.com/w3c/web-platform-tests/tree/master/html/semantics/embedded-content/media-elements/interfaces subtree) are just duplicating the [idlharness.js](https://github.com/w3c/testharness.js/blob/master/idlharness.js)-based tests we have in the [html/dom/interfaces.html](https://github.com/w3c/web-platform-tests/blob/master/html/dom/interfaces.html) file.

Given that the preferred way to test WebIDL conformance for implementations of interfaces is using [idlharness.js](https://github.com/w3c/testharness.js/blob/master/idlharness.js)-based tests, somebody should audit the tests in [html/semantics/embedded-content/media-elements/interfaces](https://github.com/w3c/web-platform-tests/tree/master/html/semantics/embedded-content/media-elements/interfaces subtree) to identify and remove any that are just duplicating what’s already covered in [html/dom/interfaces.html](https://github.com/w3c/web-platform-tests/blob/master/html/dom/interfaces.html).

html/semantics/embedded-content/media-elements/interfaces/TrackEvent/constructor.html
```diff
@@ -13,12 +13,13 @@
     assert_equals(ev.track, null, 'ev.track after assignment');
 }, document.title+', one arg');
 test(function(){
-    var obj = {};
-    var ev = new TrackEvent('foo', {track:obj});
+    var video = document.createElement('video');
+    var testTrack = video.addTextTrack('subtitles', 'foo', 'foo');
```
foolip: 'foo' arguments not needed as label and language are optional

html/semantics/embedded-content/media-elements/interfaces/TrackEvent/constructor.html
```diff
@@ -13,12 +13,13 @@
...
+    var ev = new TrackEvent('foo', {track: testTrack});
```
foolip: Can you add a test for `{track:null}` and `{track:undefined}` and another for a case that should throw TypeError, like `{track:window}` or `{track:{}}`?

foolip:
@sideshowbarker can you file an issue for that? html/dom/interfaces.html is not imported and run for Blink because it is too big and causes timeouts, so removing redundant things could end up reducing test coverage. @cdumez are you also having trouble with some idlharness.js tests?

sideshowbarker:
Filed https://github.com/w3c/web-platform-tests/issues/4000

> `html/dom/interfaces.html` is not imported and run for Blink because it is too big and causes timeouts

If that’s the case then it seems like we can solve that just be splitting up `html/dom/interfaces.html` into separate files; e.g., we could make an `html/dom/mediaelements-interfaces.html` file, etc.

foolip:
Yeah, splitting would help.

sideshowbarker:
> > > html/dom/interfaces.html is not imported and run for Blink because it is too big and causes timeouts

> > If that’s the case then it seems like we can solve that just be splitting up `html/dom/interfaces.html` into separate files; e.g., we could make an `html/dom/mediaelements-interfaces.html` file, etc.

> Yeah, splitting would help.

@Ms2ger @jgraham would you be OK with the `html/dom/interfaces.html` file being split?

Ms2ger:
In theory, sure. The one thing I'm worried about is making it harder to update to the spec.

foolip:
You mean because it would be harder to find where the IDL is pasted to update it?

youennf:
Splitting html/dom/interfaces.html would be great for WebKit too.

Ms2ger:
I filed #4002 to continue that discussion.

PR #4008, opened 2016-10-18
Import tests for MathML lengths.

fred-wang:
These tests pass on Gecko but some of them are failing on WebKit because of some bugs in the MathML implementation.

@karlt Can you please review these tests, so that we can import them into Mozilla test suite?


mathml/relations/css-styling/lengths-1.html
```diff
@@ -0,0 +1,119 @@
...
+      <!-- pc -->
+      <span style="top: 120px"><math><mspace width="12.5pc"/></math></span>
+      <span style="top: 130px; width: 200px"></span>
+
+      <!-- pt -->
+      <span style="top: 140px"><math><mspace width="150pt"/></math></span>
+      <span style="top: 150px; width: 200px"></span>
+
+      <!-- % -->
+      <span style="top: 160px"><math><mstyle mathsize="200%"><mspace width="1em"/></mstyle></math></span>
```
karlt: Did you mean 2000% here?

mathml/relations/css-styling/lengths-1.html
```diff
@@ -0,0 +1,119 @@
...
+      <span style="top: 170px; width: 200px"></span>
+
+      <!-- unitless -->
+      <span style="top: 180px"><math><mstyle mathsize="2.0"><mspace width="1em"/></mstyle></math></span>
```
karlt: And 20.0 here?

mathml/relations/css-styling/lengths-3.html
```diff
@@ -0,0 +1,158 @@
...
+    font-family: TestFont;
+    src: url("/fonts/math/xheight500.woff");
+  }
+  math {
+    font-family: TestFont;
+    font-size: 10px;
+  }
+</style>
+<script>
+  var epsilon = 1;
```
karlt: Seems like at least most of these comparisons should be exact?

Picking a multiplier for named spaces that is a multiple of 18 could make their use exact.

I'm not clear about the intentions for the different fractions in non-negative and non-positive numbers.  These could be tested less precisely than those expected to be exact, but still more precisely than a 2 pixel accepted range.

karlt (approved):
Looks good.  Just some suggestions to make the tests more precise.

mathml/relations/css-styling/lengths-1.html
fred-wang: Yes

mathml/relations/css-styling/lengths-3.html
fred-wang: > Seems like at least most of these comparisons should be exact?
> Picking a multiplier for named spaces that is a multiple of 18 could make their use exact.
Indeed, that seems to work for Gecko and WebKit on Desktop.

> I'm not clear about the intentions for the different fractions in non-negative and non-positive numbers. These could be tested less precisely than those expected to be exact, but still more precisely than a 2 pixel accepted range.
The point was to test the various syntax for a number, as Gecko uses its own parsing function for that. epsilon = .5 seems to work.

PR #4010, opened 2016-10-18
Import tests to check MathML integration points.

fred-wang:
cc @karlt 

These tests pass in Gecko and WebKit.

### DIFF
--- a/mathml/relations/html5-tree/integration-point-1.html
+++ b/mathml/relations/html5-tree/integration-point-1.html
@@ -1,0 +1,25 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8"/>
+<title>MathML inside foreignObject</title>
+<link rel="help" href="http://www.mathml-association.org/MathMLinHTML5/S2.html#SS1.SSS1"/>
+<link rel="match" href="integration-point-1-ref.html"/>
+<meta name="assert" content="Verify that MathML can be used inside a foreignObject element.">
+</head>
+<body>
+  <p>Test passes if there is a green square and no red.</p>
+  <div>
+    <div style="position: absolute; width: 200px; height: 200px; background: red"></div>
+    <div style="position: absolute;">
+      <svg width="200px" height="200px">
+        <rect width="200px" height="100px" fill="red"/>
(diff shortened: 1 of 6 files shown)


fred-wang:
@karlt Can you please take a look at that commit one too?

fred-wang:
cc' @davidcarlisle @kohlhase @rgaiacs @physikerwelt @dginev

mathml/relations/html5-tree/integration-point-1.html
```diff
@@ -0,0 +1,25 @@
...
+<body>
+  <p>Test passes if there is a green square and no red.</p>
+  <div>
+    <div style="position: absolute; width: 200px; height: 200px; background: red"></div>
+    <div style="position: absolute;">
+      <svg width="200px" height="200px">
+        <rect width="200px" height="100px" fill="red"/>
+        <foreignObject width="200px" height="200px"
+                       requiredExtensions="http://www.w3.org/1998/Math/MathML">
+          <div style="width: 200px; height: 200px; background: green"></div>
```
karlt: This is testing "MathML inside foreignObject" and yet I don't see any MathML.
I'm not familiar with svg or requiredExtensions, and I don't know the purpose of the <rect/>.
Can you explain what this is testing please? Or if someone else can review this test, then that would be great, thanks.

karlt (commented):
2 and 3 look good to me thanks.

mathml/relations/html5-tree/integration-point-1.html
fred-wang: Thanks Karl. This test actually only verifies that the MathML namespace is recognized as a supported requiredExtensions and so that the content of the foreignObject is rendered. The actual foreign content does not matter here, so that's why we don't need to put MathML inside. The SVG rect is indeed not important here. Both the red HTML div and red SVG rect will be covered by the green div.

Maybe someone more familiar with SVG can check this... @dirkschulze Could you please take a look since you review the corresponding patches in WebKit?

fred-wang:
I'm going to merge this since the MathML part has been checked by Karl and equivalent tests have been reviewed in Gecko/WebKit for the SVG foreignObject part.

PR #4012, opened 2016-10-18
Fix #3982: Remove check for remove method failure on temporary sessions

### LINKED ISSUE #3982: [eme] Syntax tests should expect remove() to work for all session types
The `remove()` tests in `syntax-mediakeysession.js` expect that "remove() on a temporary session should fail." However, the behavior of [`remove()`](https://w3c.github.io/encrypted-media/#remove) was changed such that it works for all session types.

Chrome will fail the corrected test currently: https://bugs.chromium.org/p/chromium/issues/detail?id=616166

mwatson2:


ddorwin:
Shouldn't we instead check that remove() resolves? I guess that goes beyond syntax tests and should be covered by `remove()` tests, which we don't have.

mwatson2:
Yep, ideally we need a separate `remove()` test, which ideally should check that there is the expected `keystatuseschange` event and that, indeed, you can no longer play back.

ddorwin:
LGTM to fix the bug in syntax tests.

PR #4013, opened 2016-10-18
Fix #4011: Support event sequences with multiple license request rounds

### LINKED ISSUE #4011: [eme] playback-persistent-license-events.js does not allow for a server certificate request
https://w3c-test.org/encrypted-media/drm-mp4-playback-persistent-license-events.html uses `playback-persistent-license-events.js`, which currently expects the following actions and events:
```
[
    'generaterequest',
    'license-request',
    'license-request-response',
    'updated',
    'keystatuseschange',
    'playing',
    'remove',
    'keystatuseschange',
    'license-release',
    'license-release-response',
    'closed-promise',
    'updated'
]
```
Chrome OS 54.0.2840.51 has the following results:

>0: "generaterequest"
**1: "license-request"**
**2: "license-request-response"**
**3: "updated"**
4: "license-request"
5: "license-request-response"
6: "updated"
7: "keystatuseschange"
8: "playing"
9: "license-release"
10: "license-release-response"
11: "updated"

The Chrome OS results show a second pair of license exchange messages. This is likely caused by a request for a server certificate in 1-3, meaning 4-6 are the events the test actually expects.

<br>For reference, here's an analysis of the remaining results. We'll verify the expectations and check a newer version of Chrome OS.
* After that, the events match through "playing".
* We do not see the remove() promise resolved or the corresponding keystatuseschange on Chrome OS.
* Then the license release exchange works as expected except the closed promise is not resolved.

mwatson2:


encrypted-media/scripts/playback-persistent-license-events.js
```diff
@@ -49,21 +49,21 @@ function runTest(config,qualifier) {
...
-                                    'remove',
-                                    'keystatuseschange',
-                                    'license-release',
-                                    'license-release-response',
-                                    'closed-promise',
-                                    'updated'
-                                ],
+                    assert_equals(_events[0], 'generaterequest',"First event is generateRequest");
+                    var i = 1;
+                    for(; i < _events.length-8; i+=3 ) {
```
ddorwin: I worry that the literal `8` could rot. One option would be to move the array at line 59 up as a variable and use its length here.

While less likely to change, `3` could also be problematic, and the same solution could be applied. (This would also address the formatting nit below.)

encrypted-media/scripts/playback-persistent-license-events.js
```diff
@@ -49,21 +49,21 @@ function runTest(config,qualifier) {
...
+                        assert_array_equals(_events.slice(i,i+3),['license-request', 'license-request-response', 'updated'],
```
ddorwin: nit: It would be nice if the events were easily readable. At least start the array on a new line. Possibly expand each element to a new line like below.

encrypted-media/scripts/playback-temporary-events.js
```diff
@@ -78,19 +78,22 @@ function runTest(config,qualifier) {
...
+                            "Expected one or more license request sequences");
+                }
+                assert_greater_than(i,1,"Expected at least one license request sequence");
+                assert_array_equals(_events.slice(i), [ 'allkeysusable',
+                                                        'playing',
+                                                        'closed',
+                                                        'closed-promise',
+                                                        'emptykeyslist' ],
+                            "Expected events sequence" );
+
```
ddorwin: nit: lots of extra whitespace

ddorwin (approved):
Looks good. I have some minor suggestions, but those could be addressed separately.

ddorwin (approved):
Thanks!

PR #4014, opened 2016-10-18
[eme] Improve reporting of failures for event sequence tests

mwatson2:


encrypted-media/util/utils.js
```diff
@@ -268,4 +268,27 @@ function test_exception(testCase /*...*/) {
     }
 }
 
+// Check that the events sequence (array of strings) matches the pattern (array of either strings, or
+// arrays of strings, with the latter representing a possibly repeating sub-sequence)
+function checkEventSequence(events,pattern) {
+    function th(i) { return i + (i < 4 ? ["th", "st", "nd", "rd"][i] : "th"); }
+    var i = 0, j=0, k=0;
```
ddorwin: nit: More descriptive variable names, would be helpful, though also make the following code much longer.

Even commenting what the variables are would help. Using something like e, p, and pe _might_ be better. (Those are short for eventsIndex, patternIndex, and patternElementIndex.)

encrypted-media/util/utils.js
```diff
@@ -268,4 +268,27 @@ function test_exception(testCase /*...*/) {
...
+    while(i < events.length && j < pattern.length) {
+        if (!Array.isArray(pattern[j])) {
+            assert_equals(events[i], pattern[j], "Expected " + th(i+1) + " event to be '" + pattern[j] + "'");
```
ddorwin: Suggestion for future improvement:
Does evaluation terminate after the assert? Is there a way to log the entire `events` array on failure? That makes debugging issues much easier. Currently, I have to set a breakpoint and look at the value of the array.

I was going to file a request against `assert_array_equals` to log the arrays to the console log, but now we're not using that, so we'd need to implement something ourselves.

What happens if we add the input arrays to the message? Do they get printed? Otherwise, I guess we'd have to use if statements to compare then `console.log()` and `assert(false, <message>)`.

ddorwin (approved):
Thank you!

I left some comment on possible improvements

PR #4019, opened 2016-10-19
Fix bugs in some CEReactions tests.

rniwa:
DOMTokenList.html: One of the test cases was erroneously asserting that DOMTokenList shouldn't
enqueue attributeChangedCallback when the removed token doesn't exist. DOMTokenList's remove doesn't check such a condition: https://dom.spec.whatwg.org/#dom-domtokenlist-remove
WebKit will pass the newly asserted behavior once https://webkit.org/b/163653 and Chrome fails.

Document.html: The test about importNode was erroneously asserting that a custom element
imported into a window-less document would be upgraded. Fixed this test case and added a test case for adopting a node from a template content. Both WebKit (after the aforementioned bug) and Chrome pass the newly added test case.

reactions.js: Fixed the assertion for the prototype of custom elements. We need to assert that customElement.**proto** is equal to CustomElement.prototype, not the constructor itself.

### DIFF
--- a/custom-elements/reactions/DOMTokenList.html
+++ b/custom-elements/reactions/DOMTokenList.html
@@ -89,8 +89,10 @@ test(function () {
     instance.setAttribute('class', 'hello world');
     assert_array_equals(element.takeLog().types(), ['constructed', 'attributeChanged']);
     instance.classList.remove('foo');
-    assert_array_equals(element.takeLog().types(), []);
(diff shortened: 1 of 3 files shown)


sideshowbarker:
Should be reviewed by @dominiccooney, @domenic, @hayatoito or somebody else who has good familiarity with the currently specced custom-elements requirements.

foolip:
Blink folks, I imported this with https://codereview.chromium.org/2448433002/ which causes a new failure in DOMTokenList.html as expected. Can one of you triage the failure, creating a new bug if necessary?

PR #4021, opened 2016-10-19
Import MathML tests for the mathvariant attribute.

fred-wang:
These tests pass in Gecko and WebKit and equivalent tests exist in their test suites.

@karlt Can you please review these tests? They are essentially automatically generated from the unicode.xml file maintained by @davidcarlisle


davidcarlisle:
@fred-wang , happy to see movement on getting the mathml tests into the web-platform test suite, I've been a bit distracted by the day job recently, but if you want me to look or comment on any of the tests you are adding, happy to be pinged here.

fred-wang:
@davidcarlisle Thank you. I think moving the mathml tests to web-platform-tests had been quite fast because things had already been prepared from the beginning and most of the tests had already been reviewed by WebKit people. Help is appreciated for the two remaining pull requests: the present one and https://github.com/w3c/web-platform-tests/pull/4010

karlt:
@davidcarlisle it would be great if you are able to review these.  If not, i'll see if i can do so next week.

davidcarlisle:
Hi I've _eyeballed_ the tests and they look reasonable but I'm not sure
that's a review, having just being pinged to this repository I don't have
it checked out and am not set up with the test runner etc, I'd be
interested to do that but can't promise to do it immediately, certainly not
today (can't really do it during office hours).

Or if there is a specific more lightweight test I could do that you would
consider a review that doesn't require setting up the whole testing
framework let me know.

fred-wang:
@davidcarlisle You can check tests.mathml-association.org for the mathvariant tests. For reftests, the idea is just to compare the rendering of the page with its reference. FYI, the test runner of web-platform-tests will actually just ask you to check it manually (in mozilla or webkit test suites the visual comparison is performed automatically).

The mathvariant tests have some web font tricks to make the test readable by users, but I guess the idea of the review is just to check that (for the same font) one html page and its ref should render the same according to the MathML spec.

fred-wang:
@davidcarlisle Oh, and you can actually just clone the git repo and open the HTML pages in your browser if you don't want to use the test runner.

karlt (approved):
It is clear from https://www.w3.org/TR/MathML3/chapter7.html#chars.BMP-SMP
that processors "must treat a mathematical alphanumeric character (when it
appears) as identical to the corresponding combination of the unstyled
character and mathvariant attribute value." when the mathematical alphanumeric
character existed at the time of Unicode 3.1.

There is also mention of "dotless i and j" to imply this extends to characters
introduced in Unicode 4.1.

I haven't seen anything in MathML3 clearly requiring renderers to extend this
to the Arabic Mathematical Alphabetic Symbols introduced in 6.1 and/or
indefinitely to new characters added in Mathematical Alphanumeric Symbols or
similarly named blocks.

The "Support for Plane 1 characters in currently deployed software is not
always reliable" and other sentences hint that the mathvariant mechanism
exists to bridge a gap while support for SMP characters developed.

I assume that support for SMP characters is now sufficiently advanced that
this mechanism need not be extended further.

http://www.mathml-association.org/MathMLinHTML5/S2.html#SS3.SSS1.tab2 is nice
and precise.  I don't know what authority should be given to this document,
but if this document describes and these tests align with the current behaviour
of more than one implementation, then the tests are serving a useful purpose
of testing interoperability, thanks.

fred-wang:
Thanks Karl, some comments:

> I haven't seen anything in MathML3 clearly requiring renderers to extend this to the Arabic Mathematical Alphabetic Symbols introduced in 6.1 and/or indefinitely to new characters added in Mathematical Alphanumeric Symbols or similarly named blocks.

Mmh, that's probably an issue in the MathML recommendation. I had already reported this in the past and Arabic characters are now mentioned in chapter 3. It seems that chapter 7 have not been updated, though.

> I assume that support for SMP characters is now sufficiently advanced that this mechanism need not be extended further.

I guess so. For Arabic characters there were already new mathvariant values for them, though.

> http://www.mathml-association.org/MathMLinHTML5/S2.html#SS3.SSS1.tab2 is nice and precise. I don't know what authority should be given to this document, but if this document describes and these tests align with the current behaviour of more than one implementation, then the tests are serving a useful purpose of testing interoperability, thanks.

As a general rule of thumb, the tests in mathml/ are going to verify this MathML implementation note, not the official MathML3 recommendation. I already discussed with @jgraham and @youennf and it seems that it is fine to do so, even if it is not an official W3C standard. The MathML4 threads on the Math WG mailing list have not done much progress so far (@kohlhase and @davidcarlisle may know more about the internal discussions and status) so we don't need to wait any further to import the interoperability tests into that W3C repository.

In this particular case of mathvariant, the implementation note just describes the core content of https://www.w3.org/TR/MathML3/chapter3.html#presm.commatt and avoids ambiguous statements from the official recommendation such as:

"Renderers have complete freedom in mapping mathematics style attributes to specific rendering properties. However, in practice, the mathematics style attribute names and values suggest obvious typographical properties, and renderers should attempt to respect these natural interpretations as far as possible. For example, it is reasonable to render a token with the mathvariant attribute set to "sans-serif" in Helvetica or Arial. However, rendering the token in a Times Roman font could be seriously misleading and should be avoided. "

"Renderers should support those combinations of character data and mathvariant values that correspond to Unicode characters, and that they can visually distinguish using available font characters. Renderers may ignore or support those combinations of character data and mathvariant values that do not correspond to an assigned Unicode code point, and authors should recognize that support for mathematical symbols that do not correspond to assigned Unicode code points may vary widely from one renderer to another."

Such statements have led to serious inconsistent implementations in the past because some mathvariant values were implemented with CSS properties. I hope it will be fixed in a future Math WG document...

fred-wang:
> I assume that support for SMP characters is now sufficiently advanced that this mechanism need not be extended further.

@karlt Oh, and a little bird told me that some Unicode characters are likely to be added in the future for "calligraphic" style...

PR #4023, opened 2016-10-19
Fix #3938: Report failures as FAIL instead of TIMEOUT

### LINKED ISSUE #3938: [eme] Tests should FAIL rather than TIMEOUT when a feature is not supported
We should only see `TIMEOUT` in the results if the test did not complete, such as an event never fired. Otherwise, the tests should be `FAIL`. This makes it clearer that the test ran to some completion and differentiates from, for example, test problems. Timeouts are also harder to investigate since there is not clear failing assert in the test output.

Currently, the tests for the optional persistent session types show `TIMEOUT` for all browsers. At least for browsers that do not support them, these should be `FAIL`. We might want to look at other remaining `TIMEOUT` results after the next results update.

mwatson2:


ddorwin (approved):
LGTM, assuming this works and there are no issues related to the discussion in the issue.

mwatson2:
Well, it _appears_ to work ...

PR #4024, opened 2016-10-19
Add Beacon tests http://www.w3.org/TR/beacon/

toddreifsteck:
Note that only Microsoft Edge implements sendBeacon on worker. Also... we need to have someone from Firefox analyze their failures.

### DIFF
--- a/beacon/beacon-basic-blob.html
+++ b/beacon/beacon-basic-blob.html
@@ -3,11 +3,11 @@
 <head>
     <title>W3C Beacon Basic Blob Test</title>
     <meta name="timeout" content="long">
-    <script src="../resources/testharness.js"></script>
-    <script src="../resources/testharnessreport.js"></script>
+    <script src="/resources/testharness.js"></script>
+    <script src="/resources/testharnessreport.js"></script>
 </head>
 <body>
-    <script src="../common/utils.js"></script>
+    <script src="/common/utils.js"></script>
     <script src="beacon-common.js?pipe=sub"></script>
     <script>
         "use strict";
--- a/beacon/beacon-basic-blob.html
+++ b/beacon/beacon-basic-blob.html
@@ -1,0 +1,17 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>W3C Beacon Basic Blob Test</title>
+    <meta name="timeout" content="long">
+    <script src="../resources/testharness.js"></script>
+    <script src="../resources/testharnessreport.js"></script>
+</head>
+<body>
+    <script src="../common/utils.js"></script>
+    <script src="beacon-common.js?pipe=sub"></script>
+    <script>
+        "use strict";
+        runTests(blobTests);
+    </script>
+</body>
+</html>
--- a/beacon/beacon-common.js
+++ b/beacon/beacon-common.js
@@ -1,0 +1,306 @@
+"use strict";
+
+// Different sizes of payloads to test.
+var smallPayloadSize = 10;
+var mediumPayloadSize = 10000;
+var largePayloadSize = 50000;
+var maxPayloadSize = 65536; // The maximum payload size allowed for a beacon request.
+
+// String payloads of various sizes sent by sendbeacon. The format of the payloads is a string:
+//     <numberOfCharacters>:<numberOfCharacters *'s>
+//     ex. "10:**********"
+var smallPayload = smallPayloadSize + ":" + Array(smallPayloadSize).fill('*').join("");
+var mediumPayload = mediumPayloadSize + ":" + Array(mediumPayloadSize).fill('*').join("");
+var largePayload = largePayloadSize + ":" + Array(largePayloadSize).fill('*').join("");
+// Subtract 6 from maxPayloadSize because 65536 is 5 digits, plus 1 more for the ':'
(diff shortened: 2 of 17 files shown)


beacon/beacon-basic-blob.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>W3C Beacon Basic Blob Test</title>
+    <meta name="timeout" content="long">
+    <script src="../resources/testharness.js"></script>
+    <script src="../resources/testharnessreport.js"></script>
```
sideshowbarker: In all these files, the `resources` URLs in those script elements need to be changed to an absolute path:

```html
```

sideshowbarker (changes requested):

sideshowbarker (approved):

beacon/beacon-common.js
```diff
@@ -0,0 +1,306 @@
+"use strict";
+
+// Different sizes of payloads to test.
+var smallPayloadSize = 10;
+var mediumPayloadSize = 10000;
+var largePayloadSize = 50000;
+var maxPayloadSize = 65536; // The maximum payload size allowed for a beacon request.
```
igrigorik: Technically, we did not spec a max limit in the spec.. should we be testing it here?

One solution is that we spec 64kb. Another is that we either remove this test, or (better), pump up the value to some "obviously bad" value and test that it fails? E.g. 1MB or some such.

beacon/beacon-error.js
```diff
@@ -0,0 +1,30 @@
...
+    importScripts("beacon-common.js?pipe=sub");
+}
+
+test(function() {
+    // Payload that should cause sendBeacon to return false because it exceeds the maximum payload size.
+    var exceedPayload = Array(maxPayloadSize + 1).fill('z').join("");
+
+    var success = navigator.sendBeacon("http://doesnotmatter", exceedPayload);
+    assert_false(success, "calling 'navigator.sendBeacon()' with payload size exceeding the maximum size must fail");
+}, "Verify calling 'navigator.sendBeacon()' with a large payload returns 'false'.");
```
igrigorik: Same comment as above on maxsize.

beacon/beacon-error.js
```diff
@@ -0,0 +1,30 @@
...
+
+test(function() {
+    var invalidUrl = "http://invalid:url";
+    // http://osgvsowi/7877590 - Fetch consistently throws TypeMismatchError instead of TypeError
+    assert_throws("TypeMismatchError", function() { navigator.sendBeacon(invalidUrl, smallPayload); },
```
igrigorik: Per spec...

> Set parsedUrl to the result of the URL parser steps with url and base. If the algorithm returns an error, or if parsedUrl's scheme is not "http" or "https", throw a "SyntaxError" exception and terminate these steps.

- Chrome throws `SyntaxError`
- Firefox throws `URLMismatchError`
- Edge throws a `TypeMismatchError`

Not sure what to do about this one. Wondering if we should simply relax the spec and indicate that an error should be throw, without declaring a type. Similarly, update the test to check that an error is thrown, without checking its type.

igrigorik: @travisleithead @annevk any thoughts or recommendations on how to proceed with this one?

annevk: The specification has to be precise and the testcase has to be precise, matching the specification. DOMException "SyntaxError" seems okayish. Note that Fetch throws JavaScript's TypeError, not DOMException "TypeMismatchError".

igrigorik: Fetch will throw a `TypeError` if it fails to parse the URL, correct? Given that all three implementations are different.. does it make sense to align with Fetch and spec that?

tds1307:
low audio response on y/t

beacon/beacon-error.js
annevk: Yeah, a JS TypeError. I think that would make sense, but I'm biased.

igrigorik: I'm leaning towards `TypeError` because we'll also be telling folks to use Fetch for customized cases (e.g. non-POST method) of beacon..

@toddreifsteck wdyt?

igrigorik: Done: https://github.com/w3c/beacon/commit/ff26009247cc3c76eb9c7b33cd9185082cff0a01

Filed: crbug.com/660241 and https://bugzilla.mozilla.org/show_bug.cgi?id=1313533.

toddreifsteck:
Just realized this is still sitting here open.. Brandon will submit his updates to our branch then I'll rebase this on Master so we can bring it back to life.

igrigorik:
*friendly bump* :-)

annevk:
As some high-level feedback, you should use the .any.js format described here: http://web-platform-tests.org/writing-tests/testharness.html. That way you only need a single JavaScript resource to test windows and workers and no longer need the boilerplate HTML.

See https://github.com/w3c/web-platform-tests/pull/5129 for an equivalent PR for Fetch. Simplifies things dramatically.

igrigorik:
@toddreifsteck any updates on this one? :)

toddreifsteck:
This PR is being abandoned. We have another fork that removes Workers, updates quotas, etc. I just reviewed our internal code and it is being updated. Closing this PR to avoid noise.

PR #4025, opened 2016-10-19
Test that various MessageEvents are trusted

domenic:
See https://github.com/whatwg/html/pull/1935.

### DIFF
--- a/webmessaging/MessageEvent-trusted.html
+++ b/webmessaging/MessageEvent-trusted.html
@@ -26,13 +26,12 @@ async_test(t => {
 async_test(t => {
   assert_true("BroadcastChannel" in self, "The browser must support BroadcastChannel");
 
-  const channel = new BroadcastChannel("name");
+  new Worker("MessageEvent-trusted-worker.js");
+  const channel = new BroadcastChannel("channel name");
 
   channel.onmessage = t.step_func_done(e => {
     assert_equals(e.isTrusted, true);
   });
-
-  channel.postMessage("ping");
 }, "With a BroadcastChannel");
 
 async_test(t => {
--- a/webmessaging/MessageEvent-trusted.html
+++ b/webmessaging/MessageEvent-trusted.html
@@ -1,0 +1,45 @@
+<!DOCTYPE html>
+<title>MessagePort message events are trusted</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<link rel="help" href="https://html.spec.whatwg.org/multipage/comms.html#dom-messageport-postmessage">
+<!-- See also:
+- https://github.com/whatwg/html/issues/1602
+- https://github.com/whatwg/html/pull/1935
+-->
+
+<script>
+"use strict";
+
+async_test(t => {
+  assert_true("MessageChannel" in self, "The browser must support MessageChannel");
+
+  const channel = new MessageChannel();
+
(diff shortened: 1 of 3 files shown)


domenic:
Note that I cannot get the BroadcastChannel tests to work in Chrome or Firefox---they timeout instead---so if someone can figure out my mistake that would be great.

annevk:
You need to have at least two `BroadcastChannel` objects I think. Ideally in separate globals. (Maybe use an `<iframe>`?)

sideshowbarker:
> > I cannot get the BroadcastChannel tests to work in Chrome or Firefox---they timeout instead

> You need to have at least two `BroadcastChannel` objects I think. Ideally in separate globals.

The following works (though not sure it’s the best way to do it…)

```diff
diff --git a/webmessaging/MessageEvent-trusted.html b/webmessaging/MessageEvent-trusted.html
index e3d26f8..502cd51 100644
--- a/webmessaging/MessageEvent-trusted.html
+++ b/webmessaging/MessageEvent-trusted.html
@@ -1,3 +1,8 @@
+<!--
+const channel = new BroadcastChannel("name");
+channel.postMessage("ping");
+/*
+-->
 <!DOCTYPE html>
 <title>MessagePort message events are trusted</title>
 <script src="/resources/testharness.js"></script>
@@ -26,13 +31,13 @@ async_test(t => {
 async_test(t => {
   assert_true("BroadcastChannel" in self, "The browser must support BroadcastChannel");

+  const worker = new Worker("#");
   const channel = new BroadcastChannel("name");

   channel.onmessage = t.step_func_done(e => {
     assert_equals(e.isTrusted, true);
   });

-  channel.postMessage("ping");
 }, "With a BroadcastChannel");

 async_test(t => {
@@ -43,3 +48,6 @@ async_test(t => {
   window.postMessage("ping", "*");
 }, "With window");
 </script>
+<!--
+*/
+-->
```

(btw would be nice if we could just use template literals in WPT tests but I’m told we shouldn’t yet)

domenic:
Thanks all, fixed the test, and now it indeed passes in Chrome and Firefox.

cdumez:
Wow, it now runs tests in browsers!? so cool!
I wish it would cover Safari (TP) too :D

webmessaging/MessageEvent-trusted.html
```diff
@@ -0,0 +1,44 @@
...
+    assert_equals(e.isTrusted, true);
+  });
+
+  channel.port1.postMessage("ping");
+}, "With a MessageChannel and its MessagePorts");
+
+async_test(t => {
+  assert_true("BroadcastChannel" in self, "The browser must support BroadcastChannel");
+
+  new Worker("MessageEvent-trusted-worker.js");
```
mkruisselbrink: It is probably safer to not create this worker until after you've created the BroadcastChannel instance and added the onmessage event handler. I don't think there is anything that guarantees the worker won't be created, ran, and posts its message to the channel before this method has a chance to connect to the channel itself (unless I'm missing something in how cross-thread-synchronization is specified in the html spec)

domenic:
@mkruisselbrink fixed, thanks!

domenic:
Ping @mkruisselbrink for a final sign-off? This seems to have gotten left behind at some point.

mkruisselbrink:
> Ping @mkruisselbrink for a final sign-off? This seems to have gotten left behind at some point.

Looks good to me.

PR #4028, opened 2016-10-19
[eme] Fix #4026: Allow server some leeway in expired license response

### LINKED ISSUE #4026: [eme] Expired license test fails with Widevine due to incorrect license
The test https://www.w3c-test.org/encrypted-media/drm-mp4-playback-temporary-expired.html is failing on Chrome 56.0.2895.0 (latest canary build) with:
assert_approx_equals: expiration attribute should equal provided expiration time expected 1476901853505 +/- 2000 but got NaN

In investigating the reason for the failure, the license provided to update() contains a license that never expires, so returning NaN is correct.

I think that the license issued for this test needs to have a duration specified in order for the session expiration to be set.

mwatson2:


encrypted-media/scripts/playback-temporary-expired.js
```diff
@@ -44,16 +44,18 @@ function runTest(config,qualifier) {
...
-            config.messagehandler(event.messageType, event.message, expiration).then(function(response) {
-                return event.target.update(response);
-            }).then(test.step_func(function(){
-                assert_approx_equals(event.target.expiration, expiration, 2000, "expiration attribute should equal provided expiration time");
-                test.step_timeout(function() {
-                    _video.play();
-                    test.step_timeout(function() { test.done(); }, 2000);
-                }, 5000);
-            })).catch(onFailure);
+            var expiration = Date.now().valueOf() - 60000;
```
SingingTree: I've been testing this and am having my licence requests rejected if I issue them with an already expired time. If I alter this time to be in the future, but such that the licence expires by the time that wait below is done, then I get expected behaviour. However, then we have a race on the network: e.g. if the network is slow on the request reaching the licence server it may reject the request.

mwatson2: Hmm, I was finding that the server rounds up the expiration time to the (server's) current time.

Anyway, if we do not have robust server support for obtaining an expired license, we cannot implement this test.

mwatson2:
Fixed by #4070.

PR #4033, opened 2016-10-20
We decided to make abort() no-op more

annevk:
See https://github.com/whatwg/xhr/issues/88 for the discussion that led
to these changes.

### DIFF
--- a/XMLHttpRequest/abort-during-open.js
+++ b/XMLHttpRequest/abort-during-open.js
@@ -7,8 +7,8 @@ test.step(function() {
       assert_unreached()
     })
   }
+  assert_equals(client.readyState, 1, "before abort()")
   client.abort()
-  assert_equals(client.readyState, 0)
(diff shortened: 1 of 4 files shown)


berniegp:
To cover all issues raised in #88, I believe [open-during-abort](https://github.com/w3c/web-platform-tests/blob/master/XMLHttpRequest/open-during-abort.htm) should also be modified by adding this final check:

    assert_equals(client.readyState, 1)

Since abort() does not fire a readystatechange event in step 3, the test cannot currently catch an erroneous state change.

Apologies if this is already covered in this PR.

annevk:
@berniegp since assert_array_equals in that test happens after the `abort()` call, aren't we good?

annevk:
Never mind, I see what you mean.

wisniewskit (approved):

wisniewskit (approved):

PR #4034, opened 2016-10-20
Manual test for click in progress flag

annevk:
See https://github.com/whatwg/html/pull/1925 for context.


zcorpan (approved):
Should we hold on merging this until the spec change is merged?

annevk:
Yeah maybe, would be good if @foolip wrote down the guidelines for that somewhere. And maybe we should add a WPT label for that kind of thing too.

foolip:
The spec change has been merged, is this ready now?

PR #4035, opened 2016-10-20
Moved to new Cast app ID, stored in common config file

tidoust:
This PR introduces a common.js file that exposes a `presentationUrls` variable which contains the list of valid presentation URLs to use to create `PresentationRequest` objects in tests. An array of two URLs is returned with:
1. a relative link to the main support/presentation.html receiver app
2. a link to the publicly available custom receiver Cast application. Note this Cast app ID is maintained by W3C team.

I haven't tried to solve the "how to run tests in a local environment?" issue (relative links will get resolved to the `web-platform.test` domain name which can only be resolved on the local computer, unless one also updates its LAN DNS server). That needs more thoughts.

The commit also ensures that the action button gets disabled as soon as the user clicks it to avoid running tests more than once (which the test harness does not like).

When possible, calls to the `PresentationRequest` constructor have been moved to within the `async_test` or `promise_test` callback function. This guarantees that the test will fail properly when run in user agents that do not implement the Presentation API at all.

I dropped the invalid presentation URL test from the getAvailability tests, because the test seems invalid: "really invalid" URLs will be caught when the PresentationRequest object is constructed. The behavior of "getAvailability" when an incorrect Cast URL is used is implementation-specific otherwise (and implementations may not support Cast URLs at all in particular).

I think there are a couple of bugs in `PresentationConnection_onxxx-manual.html` tests which I haven't tried to fix in this PR. Also, the `defaultRequest_success-manual.html` would benefit from an update to disable the button, and not crash when `PresentationRequest` is not supported at all. Same thing for `reconnectToPresentation_success-manual.html`. I'll raise separate issues.

### DIFF
--- a/presentation-api/controlling-ua/PresentationConnection_onclosed-manual.html
+++ b/presentation-api/controlling-ua/PresentationConnection_onclosed-manual.html
@@ -7,6 +7,7 @@
 <link rel="help" href="http://w3c.github.io/presentation-api/#interface-presentationconnection">
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
+<script src="common.js"></script>
 <h2>Description</h2>
 <p>
   This test validates that after connection close,<br/>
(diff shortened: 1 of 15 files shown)


sideshowbarker:
@tidoust In order to test the stability-checker update that was merged to master today, I rebased your branch for this PR with master and then force-pushed it back to your repo. So to pull back to the branch on your local clone, instead of `git pull` you’ll need to run `git fetch` and then `git rebase` or `git reset`. If you have problems please just ping .me

louaybassbouss:
Thx @tidoust LGTM

PR #4036, opened 2016-10-20
Test URL setters on <a> and <area> too

annevk:


zcorpan (approved):

annevk:
Ah shit, I think I might have overlooked something. `pathname` tests won't fail because `<a>` and `<area>` have a base URL, whereas `URL` might not. So we'd need to add `<base href=about:blank>` or some such, I think. Thoughts?

zcorpan:
I don't follow. It sets `href` to the test's `href`. Which tests are affected by the base URL here?

annevk:
When you start out with `mailto:...` and then set pathname to `/foo`. There's at least one such test.

zcorpan:
OK. It seems it doesn't affect the test results in Chromium or Gecko though AFAICT.

annevk:
Look for the results of `Setting <mailto:me@example.net>.pathname = '/foo'` and see how they differ between `URL` and `<a>`/`<area>`.

annevk:
Ooh, only affects Firefox Nightly maybe?

annevk:
Arguably that's a bug, even, so maybe this is all good.

PR #4038, opened 2016-10-20
Make sure setting pathname to "?" encodes it

annevk:


zcorpan (approved):
New test passes in WebKit/Chromium/Gecko. LGTM but I'll let you merge 🙂

PR #4043, opened 2016-10-20
Two more tests for Event-dispatch-click.html

annevk:


annevk:
Something strange happened when running this test locally in Firefox, it redirects to `javascript:globalCounter++`. Not entirely sure what's going on or how to prevent that.

zcorpan:
I guess that has to do with navigating to a `javascript:` URL when the return type is a number... You need to make it return undefined, e.g. with `javascript:(function() { globalCounter++; })()`

annevk:
@domenic? Given that javascript URLs are still changing, what is going to be reliable here?

annevk:
@zcorpan I'm surprised that (()=>globalCounters++)() does not work.

domenic:
> I'm surprised that (()=>globalCounters++)() does not work.

That function now returns globalCounters++ :). `(() => { globalCounters++; })()` would do the trick. Also probably `javascript:globalCounters++,undefined`. But yeah, any version with a function invocation to return undefined should be safe, I think.

dom/events/Event-dispatch-click.html
```diff
@@ -116,4 +116,57 @@
   })
   input.dispatchEvent(clickEvent)
 }, "redispatch during post-click handling")
+
+async_test(function(t) {
+  var input = document.createElement("input")
+  input.type = "checkbox"
+  dump.appendChild(input)
+  var child = input.appendChild(document.createElement("input"))
```
foolip: Just a note from IRC for posterity:
```
09:40 < foolip> annevk: but why is the disabled test using nested input elements, is there some significance to that?
09:40 < annevk> foolip: yeah, to make sure the disabled input is not skipped when finding an element with activation
09:41 < foolip> annevk: oh, so it could have been any parent with activation behavior
09:41 < annevk> Yeah
```

foolip (approved):

PR #4044, opened 2016-10-20
Test that radio button group and usemap name matching is case-sensitive

domenic:
This follows the spec change proposed at https://github.com/whatwg/html/pull/1941.

### DIFF
--- a/html/semantics/embedded-content/the-object-element/usemap-casing.html
+++ b/html/semantics/embedded-content/the-object-element/usemap-casing.html
@@ -1,0 +1,79 @@
+<!DOCTYPE html>
+<meta charset=utf-8>
+<title>object usemap case-sensitive</title>
+<link rel="author" title="Domenic Denicola" href="mailto:d@domenic.me">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/infrastructure.html#rules-for-parsing-a-hash-name-reference">
+<!-- See also: https://github.com/whatwg/html/issues/1666 -->
+
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<div id="log"></div>
+
+<object data="/images/threecolors.png" usemap="#sImPlE" width="300" height="300"></object>
+<map name="simple"><area shape="rect" coords="0,0,300,300"></map>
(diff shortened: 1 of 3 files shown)


annevk:
I don't understand the results from @wpt-stability-bot. @jgraham?

annevk:
Tests look good. Marking "do not merge yet" to land them together with the specification change once approved.

domenic:
I force-pushed a few times, so it got confused. I guess we should not do that in this new world.

html/semantics/embedded-content/the-object-element/usemap-casing.html
```diff
@@ -0,0 +1,79 @@
...
+<link rel="author" title="Domenic Denicola" href="mailto:d@domenic.me">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/infrastructure.html#rules-for-parsing-a-hash-name-reference">
+<!-- See also: https://github.com/whatwg/html/issues/1666 -->
+
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<div id="log"></div>
+
+<object data="/images/threecolors.png" usemap="#sImPlE" width="300" height="300"></object>
```
zcorpan: Can you add a test with case-sensitive match, as a sanity test (with the opposite pass condition)? And similarly for img and radio button tests.

(It appears that Safari passes all the `object` tests but if I add another with a case-sensitive match, it still returns the `object` element rather than the `area`.)

zcorpan (changes requested):

html/semantics/embedded-content/the-object-element/usemap-casing.html
domenic: Interesting. Some, but not all, browsers seem to fail such a sanity check for object elements. I also am getting different results between the wpt-stability-bot (presumably Linux) and my Windows 10 computer.

I guess it'll be a good test to add regardless.

zcorpan (approved):

zcorpan:
Thanks. The sanityCheck test fails in Firefox because of https://bugzilla.mozilla.org/show_bug.cgi?id=1227469

PR #4046, opened 2016-10-21
Fix handling when PresentationRequest not found

### LINKED ISSUE #4039: Test on presentation.defaultRequest may crash without reporting results
A couple of bugs on the [defaultRequest_success-manual.html](https://github.com/w3c/web-platform-tests/blob/master/presentation-api/controlling-ua/defaultRequest_success-manual.html) test:
1. The test assumes that the `PresentationRequest` constructor exists and will crash with no way to report results if it is run on a user agent that does not support the Presentation API at all. Wrap the call to `new PresentationRequest()` in a `try... catch`?
2. The "Not Supported" button should be disabled (the whole "If your browser does not support" text could perhaps be hidden) as soon as the "connectionavailable" event is detected, or as soon as the button is clicked.

tomoyukilabs:
This PR fixes #4039.
- `new PresentationRequest()` is moved to `async_test(function(t){ ... })` for proper error report.
- The "Not Supported" button is disabled when the button is clicked.
- The whole "If your browser does not support" text including the button is hidden when the `connectionavailable` event is fired.

### DIFF
--- a/presentation-api/controlling-ua/defaultRequest_success-manual.html
+++ b/presentation-api/controlling-ua/defaultRequest_success-manual.html
@@ -8,12 +8,18 @@
 <script src="/resources/testharnessreport.js"></script>
 
 <p>Click the button or the menu item for presentation on your browser (for example, "Cast"), to start the manual test.</p>
-<p>If your browser does not support <code>defaultRequest</code>, please click this button: <button id="notsupported">Not Supported</button>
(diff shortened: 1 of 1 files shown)


tomoyukilabs:
#4039 seems to be already addressed by #4035. I should cancel this PR.

Closed.

tomoyukilabs:
Sorry, I misunderstood.

Anyway, I should wait until #4035 is merged. After it is merged, I will submit a PR again.

PR #4050, opened 2016-10-21
Make stability checker report link to w3c-test.org URLs

sideshowbarker:


check_stability.py
```diff
@@ -427,7 +427,9 @@ def write_inconsistent(inconsistent, iterations):
 def write_results(results, iterations):
     logger.info("## All results ##\n")
     for test, test_results in results.iteritems():
-        logger.info("### %s ###" % test)
+        baseurl = "https://w3c-test.org/submissions"
```
jgraham: You need to test for `.https.` in the filename to decide if it's a http or https url.

lint.whitelist
```diff
@@ -80,6 +80,7 @@ W3C-TEST.ORG:tools/*
 PRINT STATEMENT:tools/*
 W3C-TEST.ORG:*/tools/*
 PRINT STATEMENT:*/tools/*
+W3C-TEST.ORG:check_stability.py
```
jgraham: I think this already had a section further down.

check_stability.py
```diff
@@ -427,7 +427,9 @@ def write_inconsistent(inconsistent, iterations):
...
+        pr = os.environ['TRAVIS_PULL_REQUEST']
```
jgraham: This case be `"false"`, which is probably not what you want. Pass through args.comment_pr and check that it parses as an int.

jgraham (changes requested):

check_stability.py
```diff
@@ -424,10 +424,22 @@ def write_inconsistent(inconsistent, iterations):
     table(["Test", "Subtest", "Results"], strings, logger.error)
 
 
-def write_results(results, iterations):
+def write_results(results, iterations, comment_pr):
     logger.info("## All results ##\n")
     for test, test_results in results.iteritems():
-        logger.info("### %s ###" % test)
+        baseurl = "http://w3c-test.org/submissions"
+        if test.split('.')[-2] == "https":
```
jgraham: Technically I think you want `"https" in os.path.splitext(test)[0].split(".")[1:]` or something.

check_stability.py
```diff
@@ -424,10 +424,22 @@ def write_inconsistent(inconsistent, iterations):
...
+            baseurl = "https://w3c-test.org/submissions"
+        pr_number = None
+        if comment_pr:
+            try:
+                pr_number = int(comment_pr)
+            except ValueError:
+                pass
+        if (pr_number):
```
jgraham: No need for parens.

PR #4062, opened 2016-10-22
Add RTCPeerConnection constructor tests

foolip:


webrtc/rtcpeerconnection/rtcpeerconnection-constructor.html
```diff
@@ -0,0 +1,136 @@
...
+  'iceConnectionState': 'new',
+  'connectionState': 'new',
+  'canTrickleIceCandidates': null,
+  // TODO: defaultIceServers
+};
+for (const attr in initialState) {
+  test(function() {
+    const pc = new RTCPeerConnection;
+    assert_equals(pc[attr], initialState[attr]);
+  }, attr + ' initial value');
```
alvestrand: This test seems strange. You're creating a set of peerconnections, but then you're only checking one attribute each round. Wouldn't it be easier / simpler / better to read something like

test(function()) {
    const pc = new RTCPeerConnection;
    for (const attr in initialState) {
        assert_equals(pc[attr], initialState[attr])
   }
 }, 'all initial values have their correct defaults');

The advantage I can see of the current state is that you get the attribute tested into the name of the test; the advantage of the other looping is that it actually tests that the attributes have the correct values at the same time - it's theoretically possible that a bad implementation could pass the first test but not the second, if signalingState was stable on run 5 and iceGatheringState was new on run 6, but not at the same time - would be completely perverse and unlikely, though.

webrtc/rtcpeerconnection/rtcpeerconnection-constructor.html
```diff
@@ -0,0 +1,136 @@
...
+<script src=/resources/testharnessreport.js></script>
+<script>
+test(function() {
+  assert_equals(RTCPeerConnection.length, 0);
+}, 'RTCPeerConnection.length');
+
+const toStringThrows = { toString: function() { throw new Error; } };
+const toNumberThrows = Symbol();
+
+const argTests = {
```
alvestrand: Shouldn't this be called "testArgs", since it's not the tests, it's the args?

It's nice to have a comment here saying that the name is the argument to pass, and the value is either false (if it's expected to succeed) or the error that is supposed to be thrown. Without it, understanding that false means success requires reading the code in detail.

webrtc/rtcpeerconnection/rtcpeerconnection-constructor.html
```diff
@@ -0,0 +1,136 @@
...
+  '{ certificates: null }': new TypeError,
+  '{ certificates: undefined }': false,
+  '{ certificates: [] }': false,
+  '{ certificates: [null] }': new TypeError,
+  '{ certificates: [undefined] }': new TypeError,
+
+  // iceCandidatePoolSize
+  '{ iceCandidatePoolSize: toNumberThrows }': new TypeError,
+}
+for (const arg in argTests) {
```
alvestrand: I like blank lines between top level constructs.

alvestrand (approved):
Very nice-looking and thorough test. Learning test styling! Do you know what caused the timeout of the Firefox bot?

foolip:
The reason for the Firefox timeout is that `new RTCPeerConnection({ iceTransportPolicy: "none" })` crashes. When writing the tests I had to comment that out. I'll file a bug for this test as a whole when it's landed. @jan-ivar FYI.

webrtc/rtcpeerconnection/rtcpeerconnection-constructor.html
foolip: Renamed and documented.

webrtc/rtcpeerconnection/rtcpeerconnection-constructor.html
foolip: Done.

webrtc/rtcpeerconnection/rtcpeerconnection-constructor.html
```diff
@@ -0,0 +1,154 @@
...
+<title>RTCPeerConnection constructor</title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<script>
+test(function() {
+  assert_equals(RTCPeerConnection.length, 0);
+}, 'RTCPeerConnection.length');
+
+// These are used for string and number dictionary members to see if they are
+// being accesssed at all.
```
phoglund: Nit: accessed

webrtc/rtcpeerconnection/rtcpeerconnection-constructor.html
```diff
@@ -0,0 +1,154 @@
...
+const toStringThrows = { toString: function() { throw new Error; } };
+const toNumberThrows = Symbol();
+
+// Test the first argument of the constructor. The key is the argument itself,
+// and the value is the first argument for assert_throws, or false if no
+// exception should be thrown.
+const testArgs = {
+  // no argument or equivalent
```
phoglund: Nit: Uppercase N, end with .

Apply throughout

webrtc/rtcpeerconnection/rtcpeerconnection-constructor.html
```diff
@@ -0,0 +1,154 @@
...
+      });
+    } else {
+      eval(expr);
+    }
+  }, expr);
+}
+
+promise_test(function() {
+  return RTCPeerConnection.generateCertificate({ name: "ECDSA", namedCurve: "P-256" })
+    .then(certificate => new RTCPeerConnection({ certificates: [certificate] }));
```
phoglund: Nit: should be indent 4 whenever you break a long line.

webrtc/rtcpeerconnection/rtcpeerconnection-constructor.html
```diff
@@ -0,0 +1,154 @@
...
+  '{ certificates: undefined }': false,
+  '{ certificates: [] }': false,
+  '{ certificates: [null] }': new TypeError,
+  '{ certificates: [undefined] }': new TypeError,
+
+  // iceCandidatePoolSize
+  '{ iceCandidatePoolSize: toNumberThrows }': new TypeError,
+}
+
+for (const arg in testArgs) {
```
phoglund: Normally I'm skeptical to this kind of complex construct inside a test, but you do have a lot of test cases so I can live with it. One way of making it simpler is to group throwing and non-throwing test cases in two different lists, but maybe it's actually nicer the way it is.

phoglund (commented):
I think it looks alright, just some minor comments.

webrtc/rtcpeerconnection/rtcpeerconnection-constructor.html
foolip: Fixed, but I find no other cases. "username" below is the name of the dictionary member.

webrtc/rtcpeerconnection/rtcpeerconnection-constructor.html
foolip: I started out with just individual tests, then had two buckets, but ended up here because it was much easier to group the tests and see what was already tested.

phoglund: Ok

phoglund:
lgtm

foolip:
https://travis-ci.org/w3c/web-platform-tests/jobs/170074906 is upset because "No output has been received in the last 10m0s" but I'm not sure why. The tests should all fail, and they do. Will try rebasing to retrigger, maybe flaky infrastructure.

jan-ivar:
> The reason for the Firefox timeout is that new RTCPeerConnection({ iceTransportPolicy: "none" }) crashes.

@foolip It's [been fixed in 52](https://bugzil.la/1312304), but still affects 51. I've requested uplift. Thanks!

FYI `"none"` is [no longer in the spec](http://w3c.github.io/webrtc-pc/#idl-def-rtcicetransportpolicy), so should it still be in web-platform-tests?

foolip:
The test is explicitly testing that it's not supported. Some parts of web-platform-tests have separate historical.html tests, but it was just very convenient to put them in the same test here.

https://crbug.com/659134 is the Chromium bug, I added use counters to begin with. Do you happen to know what it is? Was it ever in the spec, and does it do anything useful?

jan-ivar:
It [used to be](https://github.com/w3c/webrtc-pc/issues/384). Some sort of gate on the ice agent.

foolip:
I'm going to wait for the use counter numbers, but do you think it's a good idea to remove this? I presume it does *something* useful?

jan-ivar:
I very much doubt it does anything useful. Are you on mac by chance? If so, then the firefox crash stats may have been you. ;) I don't think anyone uses the "none" setting.

foolip:
Yep, I submitted a crash report from Mac when running the tests :)

PR #4073, opened 2016-10-25
Fix #4058: Correct expected order of closed promise resolutions

### LINKED ISSUE #4058: [eme] playback-temporary-events.js: Expected order of resolving close-related promises does not match the spec
`playback-temporary-events.js`, which is used in `*-mp4-playback-temporary-events.html`, checks the order of resolution of the `closed` promise attribute and the promise returned by the `close()` call.

It currently expects the latter (logged as `'closed'`) to be resolved before the former (logged as `'closed-promise'`).

I believe the spec says the order should be the opposite. The [close()](https://w3c.github.io/encrypted-media/#close) algorithm queues a task to:
 1. Run the [Session Closed](https://w3c.github.io/encrypted-media/#session-closed) algorithm on the session.
 1. Resolve the returned promise.

That first step synchronously resolves the `closed` attribute promise. Thus, the promises are resolved in series. It would seem that they should be resolved in order.

I don't recall whether such ordering is guaranteed. If not, we should probably allow both orders. It's also possible that the test and harness (e.g. `test.step_func()`) are interfering.

Currently, Chrome and Firefox pass both `*-mp4-playback-temporary-events.html` test, meaning they resolve the returned promise before executing all steps of the [Session Closed](https://w3c.github.io/encrypted-media/#session-closed) algorithm. Implementing the current spec would require user agents to wait for the "cdm to close the session associated with session" then execute the Session Closed algorithm before resolving the returned promise.
@cpearce @jdsmith3000

P.S.
The naming of these events is confusing. We should rename them to `'closed-attributed-resolved'` and `'close-returned-promise-resolved'` or similar.

mwatson2:


encrypted-media/scripts/playback-temporary-events.js
```diff
@@ -76,15 +76,15 @@ function runTest(config,qualifier) {
         }
 
         function onClosed(event) {
-            _events.push('closed-promise');
+            _events.push('closed-method-promise-resolved');
```
ddorwin: These are backwards. This is the `closed-attribute-resolved` handler.

mwatson2: Doh, yes, you're right. But it passes on Chrome as is. Is that expected ?

ddorwin: Yes, Chrome does not currently implement the correct order. I filed https://crbug.com/659374. I have https://codereview.chromium.org/2446993003/ to fix Clear Key, but fixing Widevine will take more time.

ddorwin (approved):
Thanks.

ddorwin:
@cpearce, this is going to create two new failures for Firefox.

@jdsmith3000, Chrome will fail clearkey- until https://codereview.chromium.org/2446993003/ lands. drm- will fail until https://crbug.com/659374 is fixed. We can consider this a browser bug.

PR #4076, opened 2016-10-25
[eme] Correct expected readyState value when waitingforkey

ddorwin:
https://github.com/w3c/encrypted-media/issues/338 changed the expected
readyState value when the current frame has been decrypted and decoded, which
is the case in this test.


ddorwin:
@cpearce, does this let Firefox get further in this test?

cpearce (approved):
Looks good.

PR #4077, opened 2016-10-26
serve: Add empty "aliases" config so that it can be overridden.

tkent-google:
This is a counterpart of https://github.com/w3c/wpt-tools/pull/131


tkent-google:
@ayg, why did you close this?  This should be reviewed and merged after w3c/wpt-tools#131.

PR #4078, opened 2016-10-26
[html] Reduce the number of tests in reflection-*.html test files.

tkent-google:
Remove "... should not throw" tests by testing it in "... followed by ..."
tests. This change removes 22,044 tests without reducing test coverage.

Background:
In Chromium project, we check in text dump of testharnessreport.js output.  The
text dumps for these test files are too large to handle with our code review
tool, which has 1MB limit for a single file.

### DIFF
--- a/html/dom/reflection.js
+++ b/html/dom/reflection.js
@@ -722,31 +722,45 @@ ReflectionTests.doReflects = function(data, idlName, idlObj, domName, domObj) {
     }
 
     for (var i = 0; i < idlTests.length; i++) {
+        var valueDesc = ReflectionHarness.stringRep(idlTests[i]);
         if ((data.type == "limited long" && idlTests[i] < 0) ||
             (data.type == "limited unsigned long" && idlTests[i] == 0)) {
             ReflectionHarness.testException("INDEX_SIZE_ERR", function() {
                 idlObj[idlName] = idlTests[i];
-            }, "IDL set to " + ReflectionHarness.stringRep(idlTests[i]) + " must throw INDEX_SIZE_ERR");
+            }, "IDL set to " + valueDesc + " must throw INDEX_SIZE_ERR");
         } else {
-            ReflectionHarness.run(function() {
-                idlObj[idlName] = idlTests[i];
-                if (data.type == "boolean") {
-                    // Special case yay
-                    ReflectionHarness.test(domObj.hasAttribute(domName), Boolean(idlTests[i]), "IDL set to " + ReflectionHarness.stringRep(idlTests[i]) + " followed by hasAttribute()");
-                } else if (idlDomExpected[i] !== null || data.isNullable) {
-                    var expected = idlDomExpected[i] + "";
-                    if (data.isNullable && idlDomExpected[i] === null) {
-                        expected = null;
-                    }
-                    ReflectionHarness.test(domObj.getAttribute(domName), expected, "IDL set to " + ReflectionHarness.stringRep(idlTests[i]) + " followed by getAttribute()");
+            var descPrefix = ReflectionHarness.getTypeDescription() + ": IDL set to " + valueDesc;
+            var testOnlyNoThrow = true;
+            if (data.type == "boolean") {
+                // Special case yay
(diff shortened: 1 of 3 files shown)


tkent-google:
Honestly, I'm not sure if it's ok to modify ReflectionHarness code.  ReflectionHarness doesn't look so helpful to me, and we can simplify reflection.js, new-harness.js, and original-harness.js much more.

ayg:
What do you mean, not sure it's okay to modify ReflectionHarness code?  Which code are you referring to?

html/dom/reflection.js
```diff
@@ -722,31 +722,45 @@ ReflectionTests.doReflects = function(data, idlName, idlObj, domName, domObj) {
...
     for (var i = 0; i < idlTests.length; i++) {
+        var valueDesc = ReflectionHarness.stringRep(idlTests[i]);
         if ((data.type == "limited long" && idlTests[i] < 0) ||
             (data.type == "limited unsigned long" && idlTests[i] == 0)) {
             ReflectionHarness.testException("INDEX_SIZE_ERR", function() {
                 idlObj[idlName] = idlTests[i];
-            }, "IDL set to " + ReflectionHarness.stringRep(idlTests[i]) + " must throw INDEX_SIZE_ERR");
+            }, "IDL set to " + valueDesc + " must throw INDEX_SIZE_ERR");
         } else {
-            ReflectionHarness.run(function() {
```
ayg: I don't think the approach of this patch is correct.  I think the problem is that when we call ReflectionHarness.test() inside ReflectionHarness.run(), it should call only assert_equals(), not test() (because ReflectionHarness.run() already calls test()).  Either modify ReflectionHarness.test() in new-harness.js to check if we're already inside a run() and if so don't call test(), or introduce a new ReflectionHarness.assertEquals() that's like ReflectionHarness.test() but without calling test() in new-harness.js, or something like that.  The way you do it here breaks reflection-original.html AFAICT -- there should be no testharness.js-specific code in reflection.js.

In the bigger picture, the right way to fix your problem is to condense test failures in a manner similar to how reflection-original.html does.  So perhaps instead of having a separate test() for every boolean attribute that if the content attribute is set to "foo" then the IDL property returns true, have a single test() that tests that for every attribute that's supposed to be boolean.  Or something like that.  I have some ideas on how to do this, but haven't figured them out in detail.  If this is blocking Blink/WebKit adopting this, maybe I could spend a couple of hours on it.

ayg (changes requested):

ayg:
(Closing/reopening was a mistake, not related to my review comments.)

tkent-google:
I wasn't aware of reflection-original.html, and now I understand this change is not compatible with it.  I understand the current structure of reflection.js, original-harness.js, and new-harness.js is necessary.

Then, my new question is: Do we need reflection-original.html?  As a browser vendor, reflection-original.html is useless at all because automated reflection-*.html exist.

ayg:
@tkent-google Have you tried identifying browser bugs using the testharness.js ones, and the original one?  In the testharness.js versions, you get huge numbers of extremely repetitive failures, which are condensed in the original one.  I have used the original harness in the last year and would prefer to keep it around if it's not causing problems, unless the testharness.js one is revamped to also condense repetitive failures somehow (which will be hard if you want to run a fixed set of tests irrespective of whether they fail).

ayg:
Also, as mentioned in #4087, I think Google needs to make your logs less verbose to solve this problem.  I think Mozilla's current testharness.js integration doesn't emit log lines for expected passes/failures.  Ask @jgraham about it.  That said, the current test is wrong (it's just a bug) and I don't mind fixing it as well.  If #4087 is accepted, this is moot and you should have no problem running the tests.

tkent-google:
We decided to reduce the size of text dumps in our testharnessreport.js.  So, this and #4087 are unnecessary for us.  Closing.

https://codereview.chromium.org/2466233002/

ayg:
#4139 does this more correctly and more throughly (reduces number of tests even more), in addition to some other cleanup.

Issue #4079, opened 2016-10-26
Expired license test's timeout too restrictive, races


SingingTree:
encrypted-media/drm-mp4-playback-temporary-expired.html uses the normal test time limit. However, the test has a number of waits in it, and when you get network latency on top of these waits it's possible to race on if the tests passes or times out.

I see fairly consistent failures when running against the live tests from NZ, and I'm able to reproduce this by serving locally, but introducing lag into the requests. With an extra latency of 1100 milliseconds on the video request locally I can reliably race with both ok and timeout results.

This is remedied if the test instead you the long timeout. I will submit a PR shortly with the change made.

ddorwin:
Fixed by #4080.

PR #4080, opened 2016-10-26
Extend the timeout of expired license test to long.

SingingTree:
In response to issue #4079

The test currently races against the 10 second normal test timeout. If latency
is introduced by network conditions this test will often fail with a timeout.
Extending this to be a long test mitigates this.


ddorwin (approved):
LGTM. I think @jrummell-chromium had some thoughts on not waiting as long, but this will improve the reliability for now.

PR #4082, opened 2016-10-26
Fix tests for customized built-in elements in custom-elements/htmlconstructor/newtarget.html

EdgarChen:
If I understand correctly, the fallback prototype of customized built-in element is the interface prototype object from NewTarget's realm, instead of the interface object itself.


domenic:
LGTM, thanks for catching this. Not sure why Travis won't let us merge; you might need to force-push again to re-trigger things.

PR #4084, opened 2016-10-26
Test that removed legacycallers are not supported

zcorpan:
Test for https://github.com/whatwg/html/pull/1979

Also test the previously-removed legacycallers on
HTMLFormControlsCollection and HTMLOptionsCollection.

---

I found that the added `object` and `embed` tests fail in Chrome...


foolip:
So Edge and Safari don't support legacycaller here?

zcorpan:
Correct. (Unless Edge has implemented them sometime after Edge 13.)

foolip:
"Some checks were not successful"

Can you investigate and merge when fixed @zcorpan?

PR #4086, opened 2016-10-26
Fix some nits in annotation protocol tests

halindrome:
CR testers reported that the order of arguments was wrong when checking for use of https.

They also reported that an exception was thrown when an OPTIONS request returned no Content-Type header

@azaroth42 and @BigBlueHat please review and verify these changes so we can merge them up.

### DIFF
--- a/annotation-protocol/tools/protocol-server.py
+++ b/annotation-protocol/tools/protocol-server.py
@@ -312,7 +312,7 @@ def annotation_head(request, response):
         response.status = 404
 
     add_cors_headers(response)
-    response.content = "Annotation Options\n"
+    # response.content = "Annotation Options\n"
 
 @wptserve.handlers.handler
 def annotation_options(request, response):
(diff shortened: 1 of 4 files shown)


BigBlueHat:
Looks good @halindrome!
👍

azaroth42:
LGTM too! :+1:

azaroth42:
Who can merge? There are testers running into this and walking away...

annotation-protocol/tools/protocol-server.py
```diff
@@ -312,7 +312,7 @@ def annotation_head(request, response):
         response.status = 404
 
     add_cors_headers(response)
-    response.content = "Annotation Options\n"
+    # response.content = "Annotation Options\n"
```
gsnedders: I'd rather this was just gone rather than commented out, unless there's any expectation it'll ever come back (in which case there should be a comment saying when it will!).

(That said, if you don't change that before I go to bed I'll merge anyway. :))

PR #4087, opened 2016-10-26
Run one reflection test per property

ayg:
Blink/WebKit are having trouble with these tests because they're too
big, even after splitting up.  #4078 makes a slight dent, but it looks
like my original intent was that each property should be one test, but I
didn't wind up implementing it.  This cuts the number of tests run by a
factor of something like 80 (e.g., 15,065 to 191 for embedded).

Naturally, it has the disadvantage that if an implementation fails
multiple assertions in a property, there will be no more test coverage
for that property, neither for regressions nor standards conformance.
In some cases a minor bug that's hard for a certain implementation to
fix could indefinitely disable further test coverage for a wide array of
properties.  As such, I'm not sure this is a good idea.  If the problem
that Blink/WebKit have is that the log messages are too verbose, perhaps
they shouldn't print log lines for expected passes/fails, which I
believe is what Mozilla does (speak to jgraham).

If we do want this change, we could also probably recombine the test
files.

@tkent-google This would certainly solve your problem, but might be too drastic, and I'm not sure we want it.

### DIFF
--- a/html/dom/new-harness.js
+++ b/html/dom/new-harness.js
@@ -2,21 +2,23 @@
 // original-harness.js.  Polymorphism, kind of.
 ReflectionHarness.catchUnexpectedExceptions = false;
 
+ReflectionHarness.testWrapper = function(fun) {
+  test(fun, this.getTypeDescription());
+}
+
 ReflectionHarness.test = function(expected, actual, description) {
-  test(function() {
-    assert_equals(expected, actual);
-  }, this.getTypeDescription() + ": " + description);
+  assert_equals(expected, actual,
+                this.getTypeDescription() + ": " + description);
   // This is the test suite that will rate conformance, so we don't want to
   // bail out early if a test fails -- we want all tests to always run.
   return true;
 }
 
 ReflectionHarness.run = function(fun, description) {
-  test(fun, this.getTypeDescription() + ": " + description);
+  fun();
 }
 
 ReflectionHarness.testException = function(exceptionName, fn, description) {
-  test(function() {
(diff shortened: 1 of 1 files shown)


tkent-google:
WebKit should not have the 1MB limit issue.  The issue comes from Google infrastructure.

I'm happy with this change.  The coverage difference doesn't matter because the major part of reflected attributes already pass their tests, and we can detect regressions even after this change.

ayg:
Sorry, I got this confused with html/dom/interfaces.html, which I understand is a problem for WebKit also.

I don't think this change would be particularly advantageous for Mozilla.  AFAIK the extra test() calls shouldn't add much overhead for us, but we do fail some of the tests right now and this change would reduce our test coverage slightly.  If Google has a problem with log length, they'll have that problem with other tests too, and should stop logging expected passes/fails or otherwise deal with it.  @jgraham, what do you think?

ayg:
Also interested in feedback from @Ms2ger or @gsnedders or anyone else who might have an opinion.

Ms2ger:
I would generally prefer not to (slightly) reduce Gecko's coverage, and do agree that it would be better for Chromium to fix the way it runs the tests.

OTOH, getting Chromium to run those tests might be more valuable at this point in time.

ayg:
Perhaps Chromium could take this patch locally until they sort out their logging issues.

tkent-google:
We decided to reduce the size of text dumps in our testharnessreport.js. So, Chromium project doesn't need this any longer.
Thank you for the proposed change and comments.

https://codereview.chromium.org/2466233002/

PR #4088, opened 2016-10-26
Test that events are created in the relevant realm of their target

domenic:
See discussions at https://github.com/whatwg/dom/pull/349. We should not merge until we figure out what to do there, but this helps inform discussion.

### DIFF
--- a/html/semantics/forms/resetting-a-form/reset-form-event-realm.html
+++ b/html/semantics/forms/resetting-a-form/reset-form-event-realm.html
@@ -1,0 +1,38 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>reset() event firing realm</title>
+<link rel="help" href="https://html.spec.whatwg.org/#resetting-a-form">
+<link rel="help" href="https://dom.spec.whatwg.org/#concept-event-fire">
+<link rel="author" title="Domenic Denicola" href="mailto:d@domenic.me">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<iframe src="reset-form-event-realm-support.html"></iframe>
+<iframe></iframe>
+
+<script>
+"use strict";
+
+async_test(t => {
+  window.onload = t.step_func_done(() => {
+    const frame0Form  = frames[0].document.forms[0];
+    const frame1Body = frames[1].document.body;
+
+    frame1Body.appendChild(frame0Form);
+
+    let resetCalled = false;
+    frame0Form.onreset = t.step_func(ev => {
+      resetCalled = true;
+
+      const functionConstructorInEvRealm = ev.constructor.constructor;
+      const functionConstructorInFormRealm = frame0Form.constructor.constructor;
+
+      assert_equals(functionConstructorInEvRealm, functionConstructorInFormRealm,
+        "the event must be created in the realm of the target");
+    });
+
+    frame0Form.reset();
+    assert_true(resetCalled, "The reset event handler must have been called");
+  });
+}, "reset()'s event must be fired in the Realm of the target")
+</script>
--- a/html/semantics/forms/resetting-a-form/reset-form-event-realm-support.html
+++ b/html/semantics/forms/resetting-a-form/reset-form-event-realm-support.html
@@ -1,0 +1,3 @@
+<!DOCTYPE html>
+
+<form></form>


html/semantics/forms/resetting-a-form/reset-form-event-realm.html
```diff
@@ -0,0 +1,37 @@
...
+
+  let resetCalled = false;
+  frame0Form.onreset = ev => {
+    resetCalled = true;
+
+    assert_equals(ev.constructor, frames[0].Event, "the event must be created in the realm of the target");
+    assert_not_equals(ev.constructor, frames[1].Event,
+      "the event must not be created in the realm into which the target was adopted");
+    assert_not_equals(ev.constructor, window.Event,
+      "the event must not be created in the realm that triggered the reset");
```
annevk: These last two asserts are redundant, no? And wouldn't help debugging since a user agent that does it wrong would stop at the first assert.

html/semantics/forms/resetting-a-form/reset-form-event-realm.html
```diff
@@ -0,0 +1,37 @@
...
+  const frame0Form  = frames[0].document.forms[0];
+  const frame1Body = frames[1].document.body;
+
+  frame1Body.appendChild(frame0Form);
+
+  let resetCalled = false;
+  frame0Form.onreset = ev => {
+    resetCalled = true;
+
+    assert_equals(ev.constructor, frames[0].Event, "the event must be created in the realm of the target");
```
bzbarsky: This is making assumptions about "the Realm of target" which are based on not-yet-decided spec text, no?

I would prefer it if you actually compared to the Realm of target, not to some Realm that may or may not actually be the Realm of target....

The question is how to extract the actual Realm from the target, right?  I don't think we have anything that allows us to do that, except observations like this...  In particular, the Realm of the target may or may not match the Realm of its prototype, and you can't tell short of observing the Realm of the target indirectly via things created off the target or something.

Perhaps it would make sense to require that the Realm of target in fact matches its proto chain in cases when no manual proto sets are being done in the script.  In that case we should be able to compare the Realms of any two objects by comparing Object.getPrototypeOf(obj.constructor), right?

domenic: Well, from my point of view it's based on decided-upon spec text that Gecko does not plan to follow. But I'm happy to change the test if we can figure out how.

I think an assumption like that makes sense. So we'd be comparing `Object.getPrototypeOf(ev.constructor)` with `Object.getPrototypeOf(frame0Form.constructor)`, right?

bzbarsky: That would make sense to me, yes.

domenic: Apparently Firefox implements class-side inheritance and Chrome doesn't, so that test isn't great -_-. I guess I'll crawl the prototype until I find a constructor named Function...

domenic: Nevermind, they both do, but in Chrome the error is swallowed for other reasons... anyway, check back when I push a new commit.

domenic:
@Ms2ger @jgraham I am now in a very strange situation, maybe a testharness.js bug. Firefox passes. Chrome passes, however, the developer console contains an AssertionError, and if I check the thing being asserted manually, it is false. So Chrome is getting away with a free pass somehow. This seems pretty bad...

domenic:
OK, @jgraham helped find the test bug; thanks.

Now, unfortunately, it looks like Chrome is failing the test. @dtapuska, can you help? Test is at http://w3c-test.org/submissions/4088/html/semantics/forms/resetting-a-form/reset-form-event-realm.html. I tried tracing through Chrome's code but couldn't find at what point it decides which realm to create the Event wrapper object in.

bzbarsky:
> OK, @jgraham helped find the test bug; thanks.

Out of curiosity, what was it?

domenic:
So, we have merged the relevant spec change in https://github.com/whatwg/dom/commit/5f39374c6f1d7893ece8101b083b8f6ccf8d1d72, but we did so before realizing that Chrome does not follow the proposed spec text :(. I have opened https://bugs.chromium.org/p/chromium/issues/detail?id=664676 to discuss on the Chrome side; my hope is we can get sign-off over there for a change.

annevk:
By the way, these tests seem very similar to (perhaps misnamed in retrospect) WebIDL/current-realm.html. Should we group them or cross-link them?

domenic:
Chrome seems happy with the change, yay :).

It seems like they're testing similar things to current-realm.html, but this is more specifically about DOM's event firing. I don't think it's necessary to cross-link.

html/semantics/forms/resetting-a-form/reset-form-event-realm.html
```diff
@@ -0,0 +1,38 @@
...
+      resetCalled = true;
+
+      const functionConstructorInEvRealm = ev.constructor.constructor;
+      const functionConstructorInFormRealm = frame0Form.constructor.constructor;
+
+      assert_equals(functionConstructorInEvRealm, functionConstructorInFormRealm,
+        "the event must be created in the realm of the target");
+    });
+
+    frame0Form.reset();
```
rniwa: This is an oddly specific test. We should probably add more test cases for other kinds of events.

rniwa:
Filed https://bugs.webkit.org/show_bug.cgi?id=167624 to track this in WebKit.

rniwa:
We can add more tests but this is a good start anyway.

PR #4089, opened 2016-10-26
Add tests for the event handler IDL attribute compilation realm

domenic:
See https://github.com/whatwg/html/issues/1956. Not ready to merge until we figure that one out.

### DIFF
--- a/workers/support/error-after-some-time.js
+++ b/workers/support/error-after-some-time.js
@@ -1,0 +1,22 @@
+"use strict";
+
+importScripts("/resources/testharness.js");
+
+self.onmessage = ev => {
+  switch (ev.data) {
+    case "error expected": {
+      self.addEventListener("error", ev => {
+        assert_equals(ev.constructor, ErrorEvent, "The WorkerGlobalScope error event must fire");
+        assert_equals(ev.message, "boo", "The WorkerGlobalScope error event must have the correct message");
+      });
+    }
+
+    case "error not expected": {
+      self.addEventListener("error", ev => {
+        throw new Error("Error event must not be fired");
+      });
+    }
+  }
+
+  step_timeout(() => { throw new Error("boo"); }, 100);
+};
--- a/html/webappapis/animation-frames/callback-exception-support-1.html
+++ b/html/webappapis/animation-frames/callback-exception-support-1.html
@@ -1,0 +1,4 @@
+<!DOCTYPE html>
+<title>This frame will get navigated from</title>
+
+<h1>Navigate from here</h1>
--- a/html/webappapis/animation-frames/callback-exception-support-2.html
+++ b/html/webappapis/animation-frames/callback-exception-support-2.html
@@ -1,0 +1,4 @@
+<!DOCTYPE html>
+<title>This frame will get navigated to</title>
+
+<h1>Navigate to here</h1>
--- a/html/webappapis/scripting/events/event-handler-idl-attribute-realm-support-1.html
+++ b/html/webappapis/scripting/events/event-handler-idl-attribute-realm-support-1.html
@@ -1,0 +1,4 @@
+<!DOCTYPE html>
+<title>This frame will get navigated from</title>
+
+<h1>Navigate from here</h1>
--- a/html/webappapis/scripting/events/event-handler-idl-attribute-realm-support-2.html
+++ b/html/webappapis/scripting/events/event-handler-idl-attribute-realm-support-2.html
@@ -1,0 +1,4 @@
+<!DOCTYPE html>
+<title>This frame will get navigated to</title>
+
+<h1>Navigate to here</h1>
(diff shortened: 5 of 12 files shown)


workers/support/error-after-some-time.js
```diff
@@ -0,0 +1,23 @@
...
+  switch (ev.data) {
+    case "error expected": {
+      self.addEventListener("error", ev => {
+        assert_equals(ev.constructor, ErrorEvent, "The WorkerGlobalScope error event must fire");
+        assert_equals(ev.message, "boo", "The WorkerGlobalScope error event must have the correct message");
+      });
+    }
+
+    case "error not expected": {
+      console.log("got here");
```
zcorpan: This fails lint check

html/webappapis/animation-frames/callback-exception.html
```diff
@@ -1,27 +1,53 @@
...
+      assert_equals(windowProxy.onerror, null, "Sanity check: after navigation reset onerror");
+
+      windowProxy.onerror = t.unreached_func("The post-navigation Window's onerror must not be called");
+
+      oldRAF(() => {
+        throw new Error(exceptionMessage);
+      });
+
+      // The test passes if no error events have been fired within this time.
+      setTimeout(t.step_func_done(), 10);
```
zcorpan: I think you're supposed to use `t.step_timeout(() => { t.done(); }, 10);` so certain environments that need longer timeouts can do that. But also 10ms is too short for stable results; I believe @jgraham has recommended no less than 2000ms.

Ms2ger:
Is there a particular reason to change the tools submodule?

domenic:
Sigh, no, just some branch-switching confusion. Will fix.

jdm:
https://github.com/whatwg/html/issues/1956 is fixed. @domenic Can we merge this now?

domenic:
I've rebased and force-pushed so that we can get a look at how current browsers do on these 9-year-old tests.

Note that this PR mixes up tests for https://github.com/whatwg/html/pull/1994, which was merged, and https://github.com/whatwg/html/pull/1995, which was never merged. The majority of tests seem to be for the latter, in fact.

My default course of action here will be to extract out the portions that are about the event handler realm into a separate PR, and get that merged. Then, close this PR, and close https://github.com/whatwg/html/pull/1995, and open a new whatwg/html issue requesting that someone investigate how error events should behave. I don't think we really have enough of a conclusion on the current behavior to merge these tests as-is.

But, if CI reveals that all browsers pass these tests, then I might push harder to get them landed, along with the corresponding spec fix.

domenic:
I've extracted out the portions of this that seem solid to me into https://github.com/web-platform-tests/wpt/pull/54656. Notes below on what remains:

- `requestAnimationFrame callback exception reported to error handler`
  - `requestAnimationFrame errors in an active document should fire an error event`: this is a fine test, but a bit boring. Probably it is tested elsewhere, and more could be tested. In terms of realms, [Web IDL invoke](https://webidl.spec.whatwg.org/#invoke-a-callback-function) reports the exception to the callack function's realm; the test could e.g. make that different from `requestAnimationFrame()`'s realm or the `thisArg`'s realm. @jeremyroman worked on that during https://github.com/whatwg/html/pull/10404 and I think submitted some web platform tests, although I can't find them right now.
  - `requestAnimationFrame errors in navigated-away-from documents should not fire any error events`: This is testing nothing, as `oldRAF(() => { throw new Error(); })` will not run the passed callback at all. I can't see how this test really ever made any sense, except for by me not understanding "fully active".

- `Event handler IDL attribute realm` / `Event handler IDL attribute compilation errors in navigated-away-from documents should not fire any error events`: this test is getting at something interesting.
  - In the current spec, the `error` event is fired if the navigated-away-from document is in bfcache, or dropped if it is not in bfcache, because of the "scripting is disabled" check in step 3.2 of [get the current value of the event handler](https://html.spec.whatwg.org/#getting-the-current-value-of-the-event-handler). This seems bad.
  - I think we could fix this either by making "get the current value of the event handler" have a fully active check, or by having "report an exception" have such a check. https://github.com/whatwg/html/pull/1995 goes for the latter.

- `setTimeout compilation errors`
  - `setTimeout compilation errors in an active document should fire an error event`: similar to the RAF case above, this is a fine test but could be testing more.
  - `setTimeout compilation errors in navigated-away-from documents should not fire any error events`: similar to the RAF case above, this test doesn't make much sense. The compilation for `setTimeout()` is specced to happen on a task, not synchronously. Since tasks can only run in fully active documents, there's no way we'd expect to see a compilation error.

- `Worker error reporting in navigated-away-from documents`
  - `Worker error reporting must work as normal, when no navigation is involved`: Almost certainly redundant with existing tests. Even more boring than the RAF and setTimeout cases since inside workers you can't get access to other global scopes, so there's no way to make this interesting.
  - `Worker error reporting must not work at all if the creator document has been navigated away from`: in theory, I think worker lifetimes are subtle enough that you could get one to last past its owner document's destruction, e.g. by using a busy-loop, and then see what happens to the error, and whether it bubbles to the `Worker` object's `error` handler. I don't think this test is written well enough to do that, and I don't think the spec guarantees that you could ever keep such a worker alive, as the browser is [allowed to terminate](https://html.spec.whatwg.org/#terminate-rampant-workers) after all the owner documents are destroyed.

PR #4097, opened 2016-10-28
Fix stability-checker URL-generation bug

sideshowbarker:


check_stability.py
```diff
@@ -429,7 +429,7 @@ def write_results(results, iterations, comment_pr):
     logger.info("## All results ##\n")
     for test, test_results in results.iteritems():
         baseurl = "http://w3c-test.org/submissions"
-        if os.path.splitext(test)[0].split(".")[1:][0] == "https":
+        if "".join(os.path.splitext(test)[0].split(".")[1:]) == "https":
```
jgraham: Oh, sorry dunno how I missed this. Should be `"https" in os.path.splitext(test)[0].split(".")[1:]`

jgraham (changes requested):

PR #4100, opened 2016-10-28
Find affected test files when support files change, and run stability checker on the tests

sideshowbarker:
Currently if the only changes in a PR are to `.js[on]` file, the stability checker doesn’t run any tests—because it doesn’t know which tests to run.

The patch in this PR causes the stability checker to:
1. Walk the parent of the subdirectory the changed `.js[on]` file is from.
2. Find any test files that contain the name of the changed `.js[on]` file.
3. Append the found test files to the list of test files given to `wptrunner` to run.

### DIFF
--- a/check_stability.py
+++ b/check_stability.py
@@ -14,6 +14,7 @@ import zipfile
 from cStringIO import StringIO
 from collections import defaultdict
 from urlparse import urljoin
+from tools.manifest import manifest
 
 import requests
 
@@ -26,7 +27,6 @@ TbplFormatter = None
 reader = None
 wptcommandline = None
 wptrunner = None
-
 
 logger = logging.getLogger(os.path.splitext(__file__)[0])
 
@@ -217,7 +217,6 @@ class Firefox(Browser):
         assert latest_release != 0
         return "v%s.%s.%s" % tuple(str(item) for item in latest_release)
 
-
     def install_webdriver(self):
         version = self._latest_geckodriver_version()
         logger.debug("Latest geckodriver release %s" % version)
@@ -357,6 +356,7 @@ def get_sha1():
     git = get_git_cmd(os.path.join(os.path.abspath(os.curdir), "w3c", "web-platform-tests"))
     return git("rev-parse", "HEAD").strip()
 
+
 def build_manifest():
     with pwd(os.path.join(os.path.abspath(os.curdir), "w3c", "web-platform-tests")):
         # TODO: Call the manifest code directly
@@ -382,6 +382,52 @@ def get_files_changed():
     assert files[-1] == "\0"
     return ["%s/w3c/web-platform-tests/%s" % (root, item)
             for item in files[:-1].split("\0")]
+
+
+def get_affected_testfiles(files_changed):
+    affected_testfiles = []
+    all_tests = set()
+    nontests_changed = set(files_changed)
+    repo_root = os.path.abspath(os.path.join(os.path.abspath(os.curdir), "w3c", "web-platform-tests"))
+    manifest_file = os.path.join(repo_root, "MANIFEST.json")
+    for test, _ in manifest.load(repo_root, manifest_file):
+        test_full_path = os.path.join(repo_root, test)
+        all_tests.add(test_full_path)
+        if test_full_path in nontests_changed:
+            # Reduce the set of changed files to only non-tests.
+            nontests_changed.remove(test_full_path)
+    for changedfile_pathname in nontests_changed:
+        changed_file_repo_path = os.path.join(os.path.sep, os.path.relpath(changedfile_pathname, repo_root))
+        os.path.normpath(changed_file_repo_path)
+        path_components = changed_file_repo_path.split(os.sep)[1:]
+        if len(path_components) < 2:
+            # This changed file is in the repo root, so skip it
+            # (because it's not part of any test).
+            continue
+        top_level_subdir = path_components[0]
+        if top_level_subdir in ["conformance-checkers", "docs"]:
+            continue
+        # OK, this changed file is the kind we care about: It's something
+        # other than a test (e.g., it's a .js or .json file), and it's
+        # somewhere down beneath one of the top-level "spec" directories.
+        # So now we try to find any tests that reference it.
+        for root, dirs, fnames in os.walk(os.path.join(repo_root, top_level_subdir)):
+            # Walk top_level_subdir looking for test files containing either the
+            # relative filepath or absolute filepatch to the changed file.
+            for fname in fnames:
+                testfile_full_path = os.path.join(root, fname)
+                # Skip any test file that's already in files_changed.
+                if testfile_full_path in files_changed:
+                    continue
+                # Skip any file that's not a test file.
+                if testfile_full_path not in all_tests:
+                    continue
+                with open(testfile_full_path, "r") as fh:
+                    file_contents = fh.read()
+                    changed_file_relpath = os.path.relpath(changedfile_pathname, root).replace(os.path.sep, "/")
+                    if changed_file_relpath in file_contents or changed_file_repo_path.replace(os.path.sep, "/") in file_contents:
+                        affected_testfiles.append(testfile_full_path)
+    return affected_testfiles
 
 
 def wptrunner_args(root, files_changed, iterations, browser):
@@ -473,7 +519,7 @@ def table(headings, data, log):
 def write_inconsistent(inconsistent, iterations):
     logger.error("## Unstable results ##\n")
     strings = [("`%s`" % markdown_adjust(test), ("`%s`" % markdown_adjust(subtest)) if subtest else "", err_string(results, iterations))
-                for test, subtest, results in inconsistent]
+               for test, subtest, results in inconsistent]
     table(["Test", "Subtest", "Results"], strings, logger.error)
 
 
@@ -569,6 +615,12 @@ def main():
         do_delayed_imports()
 
         logger.debug("Files changed:\n%s" % "".join(" * %s\n" % item for item in files_changed))
+
+        affected_testfiles = get_affected_testfiles(files_changed)
+
+        logger.debug("Affected tests:\n%s" % "".join(" * %s\n" % item for item in affected_testfiles))
+
+        files_changed.extend(affected_testfiles)
 
         browser = browser_cls(args.gh_token)
 


sideshowbarker:
@jgraham @Ms2ger @gsnedders This is ready for review.

sideshowbarker:
@jgraham @Ms2ger unless you have any objections I’m hoping we can get this merged

If after it’s merged we later find out it has ended up producing too many this-test-probably-is-affected false positives, I’d be very happy to spend more time then refining it further then.

check_stability.py
```diff
@@ -353,6 +353,57 @@ def get_files_changed():
             for item in files[:-1].split("\0")]
 
 
+def get_affected_testfiles(files_changed):
+    affected_testfiles = []
+    # FIXME: treat ".xml" files as test files? or support files? or both? or neither?
+    testfile_extensions = [".html", ".htm", ".xhtml", ".xht", ".xml", ".svg"]
```
jgraham: Couldn't we use the manifest to work out what's a test file and what's a support file?

check_stability.py
```diff
@@ -353,6 +353,57 @@ def get_files_changed():
...
+    for changedfile_pathname in files_changed:
+        # Skip any changed file that is actually a test file.
+        if os.path.splitext(changedfile_pathname)[1] in testfile_extensions:
+            continue
+        # Skip any changed file that is not a support file.
+        if os.path.splitext(changedfile_pathname)[1] not in supportfile_extensions:
+            continue
+        # We have a changed file that is a support file.
+        supportfile_name = os.path.basename(changedfile_pathname)
+        travis_root = os.path.join(os.path.abspath(os.curdir), "w3c", "web-platform-tests")
```
jgraham: Can we call this `repo_root` and perhaps make it an abspath.

check_stability.py
```diff
@@ -353,6 +353,57 @@ def get_files_changed():
...
+        basedir = None
+        if len(path_components) == 2:
+            # Parent of this changed support file is a top-level subdir.
+            basedir = path_components[0]
+        elif len(path_components) > 2:
+            # This changed support file some number of subdirs down. So as
+            # the directory we walk to look for test files in, we use the
+            # parent directory of the directory this changed file is in.
+            basedir = os.path.dirname(os.path.dirname(repo_path))[1:]
+        else:
```
jgraham: We also want to exclude tools + etc.

check_stability.py
```diff
@@ -353,6 +353,57 @@ def get_files_changed():
...
+        path_components = repo_path.split(os.sep)[1:]
+        basedir = None
+        if len(path_components) == 2:
+            # Parent of this changed support file is a top-level subdir.
+            basedir = path_components[0]
+        elif len(path_components) > 2:
+            # This changed support file some number of subdirs down. So as
+            # the directory we walk to look for test files in, we use the
+            # parent directory of the directory this changed file is in.
+            basedir = os.path.dirname(os.path.dirname(repo_path))[1:]
```
jgraham: This doesn't seem quite right? Why wouldn't we start in the directory containing the support file? There are certainly support files not in a support/ subdirectory, and where they are in a subdirectory that directory typically won't add any tests. A more pressing concern is what we do about directory structures like `foo/support` `foo/subfoo/test_referencing_support`. I think a simpler algorithm would just be to look for the filename in any files in the same top-level directory.

check_stability.py
```diff
@@ -353,6 +353,57 @@ def get_files_changed():
...
+            # This changed support file is in the repo root, so skip it
+            # (because it's not part of any test).
+            continue
+        for root, dirs, fnames in os.walk(os.path.join(travis_root, basedir)):
+            # Walk basedir looking for test files containing the name of the
+            # changed support file, and add those to affected_files.
+            for fname in fnames:
+                testfile_name = os.path.join(travis_root, root, fname)
+                # Skip any file that's already in files_changed.
+                if testfile_name in files_changed:
```
jgraham: I fell like maybe this should be a set if we are going to do lots of membership tests.

jgraham (changes requested):

check_stability.py
sideshowbarker: > > `travis_root = os.path.join(os.path.abspath(os.curdir), "w3c", "web-platform-tests")`

> Can we call this `repo_root` and perhaps make it an abspath

Yes, changed in https://github.com/w3c/web-platform-tests/pull/4100/commits/d76f05591bb438c8f897b10b05641f576e74594f

check_stability.py
sideshowbarker: > I think a simpler algorithm would just be to look for the filename in any files in the same top-level directory.

Yes, changed so that we now start from the top-level directory and walk down looking for any test files that contain either the absolute path to the support file, or the relative path from that test file to that support file)

check_stability.py
sideshowbarker: >> `if testfile_name` in files_changed`

> I feel like maybe this should be a set if we are going to do lots of membership tests.

So instead the following?

``` python
files_changed_set = set(files_changed)
for root, dirs, fnames in os.walk(os.path.join(repo_root, basedir)):
…
    if testfile_full_path in files_changed_set:
```

check_stability.py
sideshowbarker: > Couldn't we use the manifest to work out what's a test file and what's a support file?

How? Just by checking if manifest has a `path` value (or `url` value?) that matches the pathname of the changed file? (and then just stopping if it does, because that means it’s a test file, not a support file)

check_stability.py
sideshowbarker: > We also want to exclude tools + etc.

But what other directory names does “etc.” cover?

sideshowbarker: > We also want to exclude tools + etc.

OK, see https://github.com/w3c/web-platform-tests/pull/4100/commits/3fd790a7cd615f88afc8d15d51dec9c43e38f3f4 which causes  changed files to be ignored if in any of the following subdirs:

* common
* conformance-checkers
* docs
* lint
* resources
* tools
* \_\_pycache__
* _certs

check_stability.py
sideshowbarker: > Couldn't we use the manifest to work out what's a test file and what's a support file?

See https://github.com/w3c/web-platform-tests/pull/4100/commits/987f506fb1b31c26490361034ef1d2f749720968, which is an attempt at that.

As I mentioned on IRC I still wonder why the manifest contains `.worker.js` files and `.any.js` files, and what tests if any the stability checker runs of those change.

sideshowbarker:
Realizing now I guess we’re also going to need to deal with the "*.sub.js" case too

jgraham:
Pretty sure that's a bug in the test.

check_stability.py
```diff
@@ -353,6 +353,53 @@ def get_files_changed():
...
+def get_affected_testfiles(files_changed):
+    affected_testfiles = []
+    ignore_dirs = ["conformance-checkers", "docs"]
+    manifest_file = os.path.join(os.path.abspath(os.curdir), "w3c", "web-platform-tests", "MANIFEST.json")
+    with open(manifest_file, "r") as fh:
+        manifest_contents = fh.read()
+    for changedfile_pathname in files_changed:
+        repo_root = os.path.abspath(os.path.join(os.path.abspath(os.curdir), "w3c", "web-platform-tests"))
+        changed_file_repo_path = changedfile_pathname[len(repo_root):]
+        if changed_file_repo_path in manifest_contents:
```
jgraham: That won't work when we change the manifest format, and it's kind of a hack now.

I think you want to read the manifest using the `manifest` module, and , outside this loop, do something like:

    files_changed = set(files_changed)
    for item in manifest:
        if item.path in files_changed:
            files_changes.remove(item.path)

check_stability.py
```diff
@@ -353,6 +353,53 @@ def get_files_changed():
...
 
+def get_affected_testfiles(files_changed):
+    affected_testfiles = []
+    ignore_dirs = ["conformance-checkers", "docs"]
+    manifest_file = os.path.join(os.path.abspath(os.curdir), "w3c", "web-platform-tests", "MANIFEST.json")
+    with open(manifest_file, "r") as fh:
+        manifest_contents = fh.read()
+    for changedfile_pathname in files_changed:
+        repo_root = os.path.abspath(os.path.join(os.path.abspath(os.curdir), "w3c", "web-platform-tests"))
+        changed_file_repo_path = changedfile_pathname[len(repo_root):]
```
jgraham: This could probably use `os.path.relpath` but it's not critical.

check_stability.py
```diff
@@ -353,6 +353,53 @@ def get_files_changed():
...
+        if changed_file_repo_path in manifest_contents:
+            # This changed file is actually a test, so skip it.
+            continue
+        os.path.normpath(changed_file_repo_path)
+        path_components = changed_file_repo_path.split(os.sep)[1:]
+        if len(path_components) < 2:
+            # This changed file is in the repo root, so skip it
+            # (because it's not part of any test).
+            continue
+        top_level_subdir = changed_file_repo_path.split(os.sep)[1]
```
jgraham: This is just `path_components[0]`, right?

check_stability.py
jgraham: (or, reading some later code, just do

    manifest_files = {os.path.join(repo_root, item.path) for item in manifest.itertypes("reftest", "testharness", [etc.])}

where the point of enumerating the test types is that the new manifest will also include support files. Of course once that lands we can just check the file is in the list of support files and so things will get simpler.)

check_stability.py
```diff
@@ -353,6 +353,53 @@ def get_files_changed():
...
+            # relative filepath or absolute filepatch to the changed file.
+            for fname in fnames:
+                testfile_full_path = os.path.join(repo_root, root, fname)
+                # Skip any file that's already in files_changed.
+                if testfile_full_path in files_changed:
+                    continue
+                # Skip any file that's not a test file.
+                if os.path.splitext(testfile_full_path)[1] not in manifest_contents:
+                    continue
+                if not os.path.isfile(testfile_full_path):
```
jgraham: I don't think this can ever be true as directories end up in `dirs`. Or did you have something else in mind?

check_stability.py
```diff
@@ -353,6 +353,53 @@ def get_files_changed():
...
+                    continue
+                with open(testfile_full_path, "r") as fh:
+                    file_contents = fh.read()
+                    curdir = os.path.dirname(testfile_full_path)
+                    changed_file_relpath = os.path.relpath(changedfile_pathname, curdir)
+                    if changed_file_relpath in file_contents or changedfile_pathname in file_contents:
```
jgraham: This should be changed_file_repo_path, right? (possibly modulo a leading `/`).

Also you should replace the path separator with `/` for the benefit of Windows in case we ever run this there.

jgraham (changes requested):

check_stability.py
sideshowbarker: > This should be `changed_file_repo_path`, right?

Yes—thanks for catching that. Fixed. 

> (possibly modulo a leading `/`)

 `changed_file_repo_path` values have a leading `/` so I think we’re good there?

> Also you should replace the path separator with `/` for the benefit of Windows in case we ever run this there.

OK, did:

```diff
+ changed_file_relpath = os.path.relpath(changedfile_pathname, curdir).replace(os.path.sep, "/")
+ if changed_file_relpath in file_contents or changed_file_repo_path.replace(os.path.sep, "/") in file_contents: 
```

check_stability.py
sideshowbarker: > > `top_level_subdir = changed_file_repo_path.split(os.sep)[1]`

> This is just `path_components[0]`, right?

Yep—now changed it to that

check_stability.py
sideshowbarker: > > `if not os.path.isfile(testfile_full_path):`

> I don't think this can ever be true as directories end up in `dirs`. Or did you have something else in mind?

Didn’t have anything else in mind, so have now just dropped that check

check_stability.py
sideshowbarker: > This could probably use `os.path.relpath`

Yeah I thought the same when I wrote it

> but it's not critical

Maybe not, but worth doing right anyway :) So, changed it now to use `os.path.relpath`

check_stability.py
```diff
@@ -386,24 +387,25 @@ def get_files_changed():
...
-        repo_root = os.path.abspath(os.path.join(os.path.abspath(os.curdir), "w3c", "web-platform-tests"))
-        changed_file_repo_path = changedfile_pathname[len(repo_root):]
-        if changed_file_repo_path in manifest_contents:
-            # This changed file is actually a test, so skip it.
-            continue
+    all_tests = []
+    nontests_changed = set(files_changed)
+    repo_root = os.path.abspath(os.path.join(os.path.abspath(os.curdir), "w3c", "web-platform-tests"))
+    manifest_file = os.path.join(repo_root, "MANIFEST.json")
+    for test,_ in manifest.load(repo_root, manifest_file):
```
jgraham: Add a space after the `,`

check_stability.py
```diff
@@ -386,24 +387,25 @@ def get_files_changed():
...
-    manifest_file = os.path.join(os.path.abspath(os.curdir), "w3c", "web-platform-tests", "MANIFEST.json")
-    with open(manifest_file, "r") as fh:
-        manifest_contents = fh.read()
-    for changedfile_pathname in files_changed:
-        repo_root = os.path.abspath(os.path.join(os.path.abspath(os.curdir), "w3c", "web-platform-tests"))
-        changed_file_repo_path = changedfile_pathname[len(repo_root):]
-        if changed_file_repo_path in manifest_contents:
-            # This changed file is actually a test, so skip it.
-            continue
+    all_tests = []
```
jgraham: Make this a set and use `.add` below.

check_stability.py
```diff
@@ -386,24 +387,25 @@ def get_files_changed():
...
+    nontests_changed = set(files_changed)
+    repo_root = os.path.abspath(os.path.join(os.path.abspath(os.curdir), "w3c", "web-platform-tests"))
+    manifest_file = os.path.join(repo_root, "MANIFEST.json")
+    for test,_ in manifest.load(repo_root, manifest_file):
+        all_tests.append(os.path.join(repo_root, test))
+        if test in nontests_changed:
```
jgraham: I think `test` here is a relative path and `nontests_changed` is absolute paths. Am I missing something or should you convert test to an absolute path as in `all_tests.append` above?

check_stability.py
```diff
@@ -413,20 +415,18 @@ def get_affected_testfiles(files_changed):
...
                     continue
                 # Skip any file that's not a test file.
-                if os.path.splitext(testfile_full_path)[1] not in manifest_contents:
-                    continue
-                if not os.path.isfile(testfile_full_path):
+                if testfile_full_path not in all_tests:
                     continue
                 with open(testfile_full_path, "r") as fh:
                     file_contents = fh.read()
                     curdir = os.path.dirname(testfile_full_path)
```
jgraham: Isn't this just `root` above?

jgraham (approved):
OK to merge with comments addressed. I think only the relative path vs absolute path one is actually important.

PR #4101, opened 2016-10-28
URL: setting port to the empty string should null port

annevk:


zcorpan (approved):
So this is for https://github.com/whatwg/url/pull/113. LGTM

PR #4109, opened 2016-10-29
Drop include_package_data from wptrunner/setup.py

sideshowbarker:


sideshowbarker:
The stability-checker parts of the travis build seem to be working as expected again without this “fix”, so closing this for now (can re-open if/when the problem ever happens again).

PR #4111, opened 2016-10-29
Put test/subtest name in backticks in s-c report

### LINKED ISSUE #4110: markup chars in test/subtest names break rendering of stability-checker report table
Some test/subtest names (e.g., many in http://w3c-test.org/html/infrastructure/urls/resolving-urls/query-encoding/windows-1251.html) have stuff like literal "`<table>`" in their names—that is, markup characters, and in particular "`<`".

When the stability checker output gets run through a markdown processor to render the report table it generates, those markup chars are treated as actual HTML markup—which breaks the rendering.

So we need to make `check_stability.py` output test/subtest names with backticks around them.

Also, some test/subtest names have line breaks in them (intentionally) as well as tabs and literal backticks. The newlines break the markdown formatting of the stability-checker report table (which expects all content for a table row to be on a single line). The backticks of course cause the markdown processor to think a span of code has ended or started, and the tab characters just show up as spaces.

So we need to translate all of those cases into something that works as expected with markdown.

sideshowbarker:
Fixes #4110

### DIFF
--- a/check_stability.py
+++ b/check_stability.py
@@ -402,6 +402,14 @@ def process_results(log, iterations):
     return results, inconsistent
 
 
+def markdown_adjust(s):
+    s = s.replace('\t', u'\\t')
+    s = s.replace('\n', u'\\n')
+    s = s.replace('\r', u'\\r')
+    s = s.replace('`',  u'\\`')
+    return s
+
+
 def table(headings, data, log):
     cols = range(len(headings))
     assert all(len(item) == len(cols) for item in data)
@@ -420,7 +428,7 @@ def table(headings, data, log):
 
 def write_inconsistent(inconsistent, iterations):
     logger.error("## Unstable results ##\n")
-    strings = [(test, subtest if subtest else "", err_string(results, iterations))
+    strings = [("`%s`" % markdown_adjust(test), ("`%s`" % markdown_adjust(subtest)) if subtest else "", err_string(results, iterations))
                 for test, subtest, results in inconsistent]
     table(["Test", "Subtest", "Results"], strings, logger.error)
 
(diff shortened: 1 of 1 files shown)


check_stability.py
```diff
@@ -26,6 +27,10 @@
 wptrunner = None
 
 
+reload(sys)
+sys.setdefaultencoding('utf8')
```
jgraham: This is a hack that seems to be considered bad form (and isn't python 3 compatible). Let's explicitly encode things instead, as required.

check_stability.py
```diff
@@ -402,6 +407,14 @@ def process_results(log, iterations):
     return results, inconsistent
 
 
+def markdown_adjust(s):
+    s = s.replace('\t', u'\u2409')
```
jgraham: I'm not sure these first three aren't confusing. I guess to do this correctly we would need real table markup. In the interim replacing them with `\\t` etc. could be better.

sideshowbarker:
Added https://github.com/w3c/web-platform-tests/pull/4111/commits/6bb279ee9c88a12309bc9573da1fdcd5c7f33678 in order to test the effects on rendering of the report table. For the resulting comment, see https://github.com/w3c/web-platform-tests/pull/4111#issuecomment-257553852.

https://github.com/w3c/web-platform-tests/pull/4111/commits/6bb279ee9c88a12309bc9573da1fdcd5c7f33678 needs to be dropped from the branch before merging to master.

check_stability.py
```diff
@@ -4,6 +4,7 @@
 import os
 import re
 import stat
+import string
```
jgraham: This seems to be unused.

jgraham: This seems to be unused.

jgraham (approved):
lgtm with the unused import and test commit removed. Feel free to squash + merge with those changes.

PR #4112, opened 2016-10-30
webrtc: update fake devices doc in simplecall.html

fippo:
@alvestrand @dontcallmedom can you take a look?


alvestrand:
lgtm, but I'd have swapped the two sentences: "You must have .... " "To iterate fast....", and perhaps shorten the second to "To iterate fast, use fake devices as described above" - that saves us from updating 2 places in the future.

alvestrand:
lgtm!

PR #4113, opened 2016-10-30
webrtc: use navigator.mediaDevices.getUserMedia

fippo:
make simplecall use navigator.mediaDevices.getUserMedia

@alvestrand @dontcallmedom PTAL -- not sure what the indent style is here.
(I am also surprised that @jan-ivar hasn't killed this yet :-))

### DIFF
--- a/webrtc/simplecall.html
+++ b/webrtc/simplecall.html
@@ -108,9 +108,8 @@ instance --use-fake-device-for-media-stream for Chrome.
 
   // This function starts the test.
   test.step(function() {
-    navigator.getUserMedia({ video: true, audio: true },
-        getUserMediaOkCallback,
-        failed('getUserMedia'));
+    navigator.mediaDevices.getUserMedia({ video: true, audio: true })
+      .then(getUserMediaOkCallback, failed('getUserMedia'));
   });
 </script>
 


webrtc/simplecall.html
```diff
@@ -108,9 +108,8 @@
 
   // This function starts the test.
   test.step(function() {
-    navigator.getUserMedia({ video: true, audio: true },
-        getUserMediaOkCallback,
-        failed('getUserMedia'));
+    navigator.mediaDevices.getUserMedia({ video: true, audio: true })
+      .then(getUserMediaOkCallback, failed('getUserMedia'));
```
jan-ivar: Conversely, the bad thing about callback style is it totally ignores errors from callbacks, which is no way to solve the distinction. Besides, the problem you mention applies equally to exceptions. 

All the browsers implement `error.stack` these days, so this is a bit redundant IMHO, but we could do both:

```js
navigator.mediaDevices.getUserMedia({ video: true, audio: true })
  .then(stream => {
    document.getElementById('local-view').srcObject = stream;
    gFirstConnection = new RTCPeerConnection(null);
    gFirstConnection.onicecandidate = e =>
        gSecondConnection.addIceCandidate(e.candidate).catch(e => failed(e, 'addIceCandidate'));
    gFirstConnection.addStream(stream);
    return gFirstConnection.createOffer();
  }, e => failed(e, `getUserMedia`))
  .then(offer => gFirstConnection.setLocalDescription(offer), e => failed(e, 'createOffer'))
  .then(() => {
    // This would normally go across the application's signaling solution.
    // In our case, the "signaling" is to call this function.
    return receiveCall(gFirstConnection.localDescription.sdp);
  }, e => failed(e, 'setLocalDescription'))
  // etc.
  .catch(e => failed(e))
```
Btw, just doing this I fixed several bugs:
* We were calling the promise-version of `setLocalDescription` without any error handling.
* We were calling the promise-version of `addIceCandidate` without any error handling.
* We were sending offer before knowing if setLocalDescription would fail.
* We were checking for `null` before calling `addIceCandidate`.

But realize this amounts to doing a try/catch around every call. E.g. compare async functions with and without:

```js
let stream = await navigator.mediaDevices.getUserMedia({ video: true, audio: true });
document.getElementById('local-view').srcObject = stream;
gFirstConnection = new RTCPeerConnection(null);
gFirstConnection.onicecandidate = async e => {
  try {
    await gSecondConnection.addIceCandidate(e.candidate);
  } catch(e) {
    failed(e);
  }
};
gFirstConnection.addStream(stream);
await gFirstConnection.setLocalDescription(await gFirstConnection.createOffer());
// This would normally go across the application's signaling solution.
// In our case, the "signaling" is to call this function.
await receiveCall(gFirstConnection.localDescription.sdp);
// etc.
```
Sweet. Now compare to if we insist on function-tagging every error, like we do today:

```js
let stream;
try {
  await navigator.mediaDevices.getUserMedia({ video: true, audio: true });
} catch(e) {
  failed (e, 'getUsermedia');
  throw e;
}
document.getElementById('local-view').srcObject = stream;
gFirstConnection = new RTCPeerConnection(null);
gFirstConnection.onicecandidate = async e => {
  try {
    await gSecondConnection.addIceCandidate(e.candidate);
  } catch(e) {
    failed (e, 'addIceCandidate');
  }
};
gFirstConnection.addStream(stream);
let offer;
try {
  offer = await gFirstConnection.createOffer();
} catch(e) {
  failed (e, 'createOffer');
  throw e;
}
try {
  await gFirstConnection.setLocalDescription(offer);
} catch(e) {
  failed (e, 'setLocalDescription');
  throw e;
}
// This would normally go across the application's signaling solution.
// In our case, the "signaling" is to call this function.
await receiveCall(gFirstConnection.localDescription.sdp);
// etc.
```

This becomes insanely hard to read and maintain (I had several labels wrong before I re-checked), and we have to re-throw the failure since we don't want the code to resume as if things succeeded, which leads to double reporting. Clearly this is not where JS is going.

What we're trying to do here anyway (without the double-reporting) amounts to adding a function name to the thrown error, which is effectively what `error.stack` does.

jan-ivar: The latter example of try/catching everything also ignores that synchronous functions can throw as well. IMHO try/catching every call is swimming upstream of the exception pattern.

fippo: different PR please @jan-ivar. For now lets no longer show something the spec says exists only for backward compability reasons. If we keep showing the old stuff in examples it will take two decades to get rid of it.

PR #4114, opened 2016-10-30
webrtc: require remote video to be loaded

fippo:
This changes the test definition slightly and requires that
the remote videos loadedmetadata has been called. This will
only happen when the ICE and DTLS connections are up and the
videos readyState is HAVE_ENOUGH_DATA.

Also adds addIceCandidate(null) to signal end-of-candidates
for Microsoft Edge.

### DIFF
--- a/webrtc/simplecall.html
+++ b/webrtc/simplecall.html
@@ -34,6 +34,14 @@ instance --use-fake-device-for-media-stream for Chrome.
   var gFirstConnection = null;
   var gSecondConnection = null;
 
+  // if the remote video gets video data that implies the negotiation
+  // as well as the ICE and DTLS connection are up.
+  document.getElementById('remote-view')
+      .addEventListener('loadedmetadata', function() {
+    // Call negotiated: done.
+    test.done();
+  });
+
   function getUserMediaOkCallback(localStream) {
     gFirstConnection = new RTCPeerConnection(null);
     gFirstConnection.onicecandidate = onIceCandidateToFirst;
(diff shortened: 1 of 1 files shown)


fippo:
Stopping things after setting the remote description turned out to be quite odd when I looked at the test in webrtc-internals since addIceCandidate was never called on the remote peerconnection.

I'd rather have a readyStateChange event listener to HAVE_ENOUGH_DATA but I think loadedmetadata implies this. @samdutton: holler if i'm wrong or you know a better way.

@shijuns FYI, this (the addIceCandidate(null) specifically) causes exceptions and a test failure in Edges native 1.0 :-/

fippo:
addIceCandidate(null) is probably no longer what the WG wants (see [here](https://github.com/w3c/webrtc-pc/issues/760)) but I can fix that once its decided.

alvestrand:
lgtm
are there cases where the candidate can be falsy and not null, so that addIceCandidate(event.candidate) (no if test) isn't the Right Thing?

fippo:
addIceCandidate(event.candidate) will do addIceCandidate(undefined) if event.candidate is not set which is not recognized by adapter.js. I could do addIceCandidate(event.candidate || null)?

alvestrand:
Hm. The events are generated straight from the other PC's onicecandidate handler, which means that it's the implementation under test that's generating whatever is there (null or undefined).
If it's not generating null when it should, the test should fail.
And if it's not clear from the spec what happens when you do addIceCandidate(undefined), then that's a spec bug :-)
Let's make the code simpler and just call addIceCandidate(event.candidate), with a comment that the candidate is supposed to be null for the end-of-candidates.

fippo:
@alvestrand that requires (minor) adapter.js changes. Can do ;-)

fippo:
changed (and squashed) to pass everything directly into the remote addIceCandidate which is what http://w3c.github.io/webrtc-pc/#simple-peer-to-peer-example shows as well

alvestrand:
Nice ... seems that I can just press merge again :-)

PR #4117, opened 2016-10-30
Don't require legacy Event initializers

ayg:
UI Events says that these should only be implemented by user agents that
require comparibility with legacy software.  In practice, Firefox does
not support initKeyboardEvent because it caused compatibility problems
(https://bugzilla.mozilla.org/show_bug.cgi?id=999645).  I'm not sure I
agree with this way of speccing things, but I think that's what the spec
says.

### DIFF
--- a/dom/events/Event-init-while-dispatching.html
+++ b/dom/events/Event-init-while-dispatching.html
@@ -9,7 +9,13 @@
 var events = {
   'KeyboardEvent': {
     'constructor': function() { return new KeyboardEvent("type", {key: "A"}); },
-    'init': function(ev) { ev.initKeyboardEvent("type2", true, true, null, "a", 1, "", true, "") },
+    'init': function(ev) {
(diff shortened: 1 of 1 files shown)


ayg:
Never mind, I'm going to do more work on this test.

ayg:
On second thought, I don't know if it's worth the effort to test this area more comprehensively.  In Gecko's case we'd theoretically want coverage for all the different event types, since we have to manually code the short-circuit behavior for each.  But adding all the different types is a pain, and it's unlikely to make any difference to real-world compat.  So I'll just ask that this fix be taken, and if someone else wants to add all the other event types, they can feel free.

zcorpan:
This seems like it goes the wrong way for reaching interop. Filed spec issue https://github.com/w3c/uievents/issues/112

annevk:
Agreed, closing this as this is not the way to go.

PR #4122, opened 2016-10-31
fix existing test expectations and add new tests for HTMLOptionsCollection

Ms2ger:


Ms2ger:
Code reviewed upstream.

PR #4123, opened 2016-10-31
Add a test for changing the media attribute on a link element.

Ms2ger:


Ms2ger:
Code reviewed upstream.

PR #4124, opened 2016-10-31
adding macro to return string or url,

Ms2ger:
using that macro with the form action,
making the form submit process use base url,
adding tests.


Ms2ger:
Code reviewed upstream.

PR #4125, opened 2016-10-31
Add a test for stylesheet links with an empty href.

Ms2ger:


Ms2ger:
Code reviewed upstream.

PR #4142, opened 2016-11-01
Extend timeout on all DRM tests that were not already long timeouts.

SingingTree:
Extend the timeouts on DRM tests to long if they were not already. This reduces
issues with racy timeouts in the tests due to network conditions. This is
relevant to the DRM tests specifically because they always hit remote machines
to fetch licenses.

### DIFF
--- a/encrypted-media/drm-mp4-setmediakeys.html
+++ b/encrypted-media/drm-mp4-setmediakeys.html
@@ -2,6 +2,7 @@
 <html>
   <head>
     <meta charset=utf-8>
+    <meta name="timeout" content="long">
     <title>Encrypted Media Extensions: setMediaKeys with DRM</title>
     <link rel="help" href="https://w3c.github.io/encrypted-media/">
 
--- a/encrypted-media/drm-mp4-setmediakeys.html
+++ b/encrypted-media/drm-mp4-setmediakeys.html
@@ -2,7 +2,6 @@
 <html>
   <head>
     <meta charset=utf-8>
-    <meta name="timeout" content="long">
     <title>Encrypted Media Extensions: setMediaKeys with DRM</title>
     <link rel="help" href="https://w3c.github.io/encrypted-media/">
 
@@ -17,9 +16,6 @@
 
     <!-- Content metadata -->
     <script src=/encrypted-media/content/content-metadata.js></script>
-
-    <!-- Message handler for DRM servers -->
-    <script src=/encrypted-media/util/drm-messagehandler.js></script>
 
     <!-- The script for this specific test -->
     <script src=/encrypted-media/scripts/setmediakeys.js></script>
--- a/encrypted-media/drm-not-callable-after-createsession.html
+++ b/encrypted-media/drm-not-callable-after-createsession.html
@@ -2,7 +2,6 @@
 <html>
   <head>
     <meta charset=utf-8>
-    <meta name="timeout" content="long">
     <title>Encrypted Media Extensions: Test MediaKeySession not callable immediately after CreateSession(), DRM.</title>
     <link rel="help" href="https://w3c.github.io/encrypted-media/">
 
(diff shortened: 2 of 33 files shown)


ddorwin:
Thanks. This looks good, but on second thought, it might be too broad. There are some tests that don't actually talk to a server. At a minimum, the syntax tests but probably others. We should probably only change tests that include `drm-messagehandler.js`, but though that may not be perfect.

SingingTree:
Cheers. PR has been updated to remove the longer timeouts on tests that were not including the message handler. I've also update the commit message to reflect the changes are specific to tests that handle messages.

encrypted-media/drm-not-callable-after-createsession.html
```diff
@@ -2,6 +2,7 @@
 <html>
   <head>
     <meta charset=utf-8>
+    <meta name="timeout" content="long">
```
ddorwin: I don't think this test need needs `drm-messagehandler.js`. `not-callable-after-createsession.js` does not handle messages.

We should remove it from `clearkey-` too.

encrypted-media/drm-mp4-setmediakeys.html
```diff
@@ -2,6 +2,7 @@
 <html>
   <head>
     <meta charset=utf-8>
+    <meta name="timeout" content="long">
```
ddorwin: I don't think this test need needs `drm-messagehandler.js`. `setmediakeys.js` does not handle messages.

We should remove it from `clearkey-` too.

ddorwin (approved):
Thanks. LGTM, but I think there are two tests where we should instead remove `drm-messagehandler.js`.

ddorwin (approved):

ddorwin:
This is ready to merge, but the checks have taken over half an hour and do not show any progress. If this doesn't resolve, you could try pushing again, maybe with a merge, or contacting an admin.

ddorwin:
@SingingTree, do you want to try merging master to see if that kicks off a successful run?

ddorwin:
@SingingTree, this seems to have included all the commits rather than being just a merge. Also, the commit message seems odd compared to, for example, "Merge remote-tracking branch 'upstream/gh-pages' into foo."

I don't know if this is the most optimal process, but I always `git fetch upstream` then `git merge upstream/master`.

@jgraham, any idea why this PR is failing the checks? Normally, runs for EME tests complete in two minutes without running any tests.

SingingTree:
Hrmm yeah, lemme try and remedy that. I may just rebase this, ontop of master to make life less painful, as the original master it came off has marched on.

ddorwin (approved):
LGTM. This PR is still failing the presubmit check. Maybe try merging in master one more time.

sideshowbarker:
> This PR is still failing the presubmit check. Maybe try merging in master one more time.

Yeah that would be good, because the failure is due to the stability checker not being able to run at all. Getting the latest `check_stability.py` in from master should fix it, and then it can run and actually check to see if find any stability problem with the change.

(nit: I think usually rather merging in master to a PR branch it’s better to rebase the branch against master and then force-push it back to the remote.)

SingingTree:
Having look at this I think the problem is this: the check stability run checks 10 iterations of the changed test files. It does not check for passes or fails, rather it checks to see if the output is consistent. In the case of the EME tests this appears to be a timeout of each test. So for small changes it's no problem, but if you change a lot of tests the iterations take longer. There's a global timeout, where if the 10 iterations take longer than, the suite fails. As a result of a large number of files being changed, combined with their increased timeouts we're running into the global limit. 

I'm mindful that if a large number of encrypted-media tests are changed in future this issue will repeat. Thoughts?

cpearce:
Should the global timeout be made longer so that we are less likely to get failures when making large changes in future?

Or we should we just merge the big change here, ignore the failure, and then keep an eye on the subsequent change's results?

jgraham:
This is just exceeding the travis timeout, which we can't control. If you haven't made any functional changes other than increasing timeouts, I think it's fine to merge it. An admin (e.g. @sideshowbarker or I) can bypass the presubmit requirements.

SingingTree:
The majority of this is extending timeouts. There are also some files that have been modified to remove unused message handler includes based on the recommendations above. If there are no concerns about either of these changes I'm happy to have this merged bypassing the presubmit reqs.

PR #4147, opened 2016-11-02
Back out bug 1268047, because the spec it tried to implement backs the web.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1306472


jgraham:
Code reviewed upstream.

PR #4148, opened 2016-11-02
followup.  Fix the timeout on the html/browsers/browsing-the-web/navigating-across-documents/014.html web platform test.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1306472


jgraham:
Code reviewed upstream.

PR #4150, opened 2016-11-02
Add reference test for <ol reversed> with display:none children;

jgraham:
MozReview-Commit-ID: KWgQeeV9IB6

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1300944


jgraham:
Code reviewed upstream.

PR #4156, opened 2016-11-02
Port test_bug411103.html to wpt;

jgraham:
This allows other UAs to use it, removes duplicated checks, and
increases the chance of us noticing if the spec changes.  Some of the
expected values in our mochitest were contrary to the spec.

I checked the new expected failures against the spec and the other UAs.
I filed a spec bug for one group because it was contrary to all UAs
(although IMO the spec makes more sense and the UAs are buggy), and the
others are fixed in the next patch.

MozReview-Commit-ID: 1j11XgfuErB

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1298818


jgraham:
Code reviewed upstream.

PR #4162, opened 2016-11-02
Fix a streams test that forgot to return a promise

domenic:
Caught by the latest testharness.js.

@isonmad maybe you can rubber-stamp this?


isonmad (approved):
LGTM.

PR #4164, opened 2016-11-02
Import a WebKit test for the element queue.

rniwa:
Trunk WebKit and Chrome Canary pass all test cases.

### DIFF
--- a/custom-elements/custom-element-reaction-queue.html
+++ b/custom-elements/custom-element-reaction-queue.html
@@ -1,0 +1,154 @@
+<!DOCTYPE html>
+<html>
+<head>
+<title>Custom Elements: Each element must have its own custom element reaction queue</title>
+<meta name="author" title="Ryosuke Niwa" href="mailto:rniwa@webkit.org">
+<meta name="assert" content="Each element must have its own custom element reaction queue">
+<meta name="help" content="https://html.spec.whatwg.org/multipage/scripting.html#custom-element-reaction-queue">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="./resources/custom-elements-helpers.js"></script>
+</head>
+<body>
+<div id="log"></div>
+<script>
+
+test_with_window(function (contentWindow) {
+    const contentDocument = contentWindow.document;
+    contentDocument.write('<test-element id="first-element">');
+    contentDocument.write('<test-element id="second-element">');
+
+    const element1 = contentDocument.getElementById('first-element');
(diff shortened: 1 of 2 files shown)


custom-elements/custom-element-reaction-queue.html
```diff
@@ -0,0 +1,154 @@
...
+
+    assert_equals(log.length, 7);
+    assert_constructor_log_entry(log[0], element1);
+    assert_attribute_log_entry(log[1], {name: 'id', oldValue: null, newValue: 'first-element', namespace: null});
+    assert_connected_log_entry(log[2], element1);
+    assert_constructor_log_entry(log[3], element2);
+    assert_attribute_log_entry(log[4], {name: 'id', oldValue: null, newValue: 'second-element', namespace: null});
+    assert_attribute_log_entry(log[5], {name: 'class', oldValue: null, newValue: 'foo', namespace: null});
+    assert_connected_log_entry(log[6], element2);
+}, 'Mutating a undefined custom element while upgrading a custom element must not enqueue or invoke reactions on the mutated element');
```
domenic: Well, it enqueues them, right? They just execute after the first one finishes.

custom-elements/custom-element-reaction-queue.html
```diff
@@ -0,0 +1,154 @@
...
+    assert_equals(log.length, 7);
+    assert_adopted_log_entry(log[0], element1);
+    assert_adopted_log_entry(log[1], element3);
+    assert_connected_log_entry(log[2], element3);
+    assert_attribute_log_entry(log[3], {name: 'id', oldValue: null, newValue: 'foo', namespace: null});
+    assert_connected_log_entry(log[4], element1);
+    assert_adopted_log_entry(log[5], element2);
+    assert_connected_log_entry(log[6], element2);
+
+}, 'Mutating another custom element inside adopted callback must invoke all pending callbacks on the mutated element');
```
domenic: Wow, good test; took a while to work through but yeah, this is as expected.

domenic (approved):
Minor question about one of the test names but LGTM otherwise.

custom-elements/custom-element-reaction-queue.html
rniwa: No, what's happening here is that the reaction isn't enqueued by the mutation because the element hadn't been upgraded. The reactions get enqueued later when the element is upgraded.

That is, if you mutated an attribute by removing & adding & adding an attribute, you'd still get a single reaction afterwards when the element is upgraded.

domenic: Ah OK, all good then, that makes sense; we can merge as-is.

PR #4167, opened 2016-11-03
Add new ImageBitmapRenderingContext tests

xidachen:
The specs for ImageBitmapRenderingContext is here:
https://html.spec.whatwg.org/multipage/scripting.html#the-imagebitmap-rendering-context

This pull request adds some new tests to ensure spec compliance.


xidachen:
@junov : Please review.

junov:
The test code looks great but the file naming needs to be improved. The directory for imageBitmapRenderingContext should be a sibling, not a child, of 2dcontext. Also, the name ibcontext is not very readable nor is it descriptive of what is being tested.

junov:
I also think the file should be split by tested sub-feature.  For example, a file just for the alpha creation attribute.

xidachen:
@junov Thank you for your review. In the second commit, I have put all the tests in a folder imagebitmap-renderingcontext, which is parallel to 2dcontext, and I have break the tests into 5 test files. Please take a look. Thank you.

imagebitmap-renderingcontext/context-creation-with-alpha.html
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    var srcCanvas = document.createElement('canvas');
+    srcCanvas.width = width;
+    srcCanvas.height = height;
+    var ctx = srcCanvas.getContext('2d');
+    ctx.fillStyle = 'rgba(0, 255, 0, 0.5)';
+    ctx.fillRect(0, 0, width, height);
+    createImageBitmap(srcCanvas).then(consumeImageBitmapAlphaDefault);
+}, "Test that canvas.getContext('bitmaprenderer') preserves alpha");
```
junov: I think the test should go last in this file, and it should entitled "Test that the 'alpha' context creation attribute is true by default."

imagebitmap-renderingcontext/context-creation-with-alpha.html
```diff
@@ -0,0 +1,92 @@
...
+    srcCanvas.width = width;
+    srcCanvas.height = height;
+    var ctx = srcCanvas.getContext('2d');
+    ctx.fillStyle = 'rgba(0, 255, 0, 0.5)';
+    ctx.fillRect(0, 0, width, height);
+    createImageBitmap(srcCanvas).then(consumeImageBitmapAlphaFalse);
+}, "Test that canvas.getContext('bitmaprenderer', alpha:false) sets the canvas opaque");
+
+promise_test(function() {
+    function consumeImageBitmapAlphaTrue(image) {
```
junov: Could this be refactored to avoid so much repeated code?

imagebitmap-renderingcontext/null-source.html
```diff
@@ -0,0 +1,75 @@
...
+        bitmapCtx.transferFromImageBitmap(image2);
+        var myCanvas3 = document.createElement('canvas');
+        myCanvas3.width = width;
+        myCanvas3.height = height;
+        var myCtx3 = myCanvas3.getContext('2d');
+        myCtx3.drawImage(bitmapCanvas, 0, 0);
+        testCanvas(myCtx3, 255, 0, 0, 255);
+    }
+
+    var canvas = document.createElement('canvas');
```
junov: call this greenCanvas

imagebitmap-renderingcontext/null-source.html
```diff
@@ -0,0 +1,75 @@
...
+    canvas.width = width;
+    canvas.height = height;
+    var ctx = canvas.getContext('2d');
+    ctx.fillStyle = '#0f0';
+    ctx.fillRect(0, 0, width, height);
+
+    var canvas2 = document.createElement('canvas');
```
junov: redCanvas

imagebitmap-renderingcontext/null-source.html
```diff
@@ -0,0 +1,75 @@
...
+    canvas2.width = width;
+    canvas2.height = height;
+    var ctx2 = canvas2.getContext('2d');
+    ctx2.fillStyle = '#f00';
+    ctx2.fillRect(0, 0, width, height);
+
+    return Promise.all([
+        createImageBitmap(canvas),
+        createImageBitmap(canvas2),
+    ]).then(([image, image2]) => {
```
junov: [greenImage, redImage]

imagebitmap-renderingcontext/null-source.html
```diff
@@ -0,0 +1,75 @@
...
+        testTransferFromImageBitmapNullability(image, image2);
+    });
+},'Test that transferFromImageBitmap(null) should not return any error');
```
junov: "Test that transferFromImageBitmap(null) discards the previously transferred image"

imagebitmap-renderingcontext/context-creation-with-alpha.html
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    var srcCanvas = document.createElement('canvas');
+    srcCanvas.width = width;
+    srcCanvas.height = height;
+    var ctx = srcCanvas.getContext('2d');
+    ctx.fillStyle = 'rgba(0, 255, 0, 0.5)';
+    ctx.fillRect(0, 0, width, height);
+    createImageBitmap(srcCanvas).then(consumeImageBitmapAlphaTrue);
+}, "Test that canvas.getContext('bitmaprenderer', alpha:true) preserves the alpha");
```
junov: I think what you mean is "Test that an ImageBitmapRenderingContext with alpha enabled preserves alpha".

junov (changes requested):
Requesting a few tweaks.  Also, we're missing test coverage for the behaviors related to consolidating differences between the canvas element's intrinsic dimensions and the ImageBitmap's dimensions.

xidachen:
@junov I have addressed all comments, and added this test: bitmaprenderer-as-imagesource.html

PTAL.

imagebitmap-renderingcontext/detached-source.html
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE html>
```
junov: could you put transferFromImageBitmap in the file name?
Something like transferFromImageBitmap-detached.html

imagebitmap-renderingcontext/null-source.html
```diff
@@ -0,0 +1,75 @@
+<!DOCTYPE html>
```
junov: tranferFromImageBitmap-null.html ?

junov (changes requested):
Looks good. Last nits about file naming.

xidachen:
@junov: File names are changed. Thanks.

junov:
One more thing we need to test. When a canvas with a bitmap render is used as an image source, and the transferFromImageBitmap() has never been called, it is supposed to produce a transparent black bitmap of the dimensions given by canvas.width/height.  Same thing should happen after calling transferFromImageBitmap(null)

xidachen:
@junov I have added two new test cases in bitmaprenderer-as-imagesource.html. PTAL. Thanks

junov (approved):

imagebitmap-renderingcontext/bitmaprenderer-as-imagesource.html
```diff
@@ -0,0 +1,86 @@
...
+    }));
+}, 'Test that createImageBitmap from a bitmaprenderer canvas produces correct result');
+
+async_test(function(t) {
+    var canvas = document.createElement("canvas");
+    canvas.width = width;
+    canvas.height = height;
+    var ctx = canvas.getContext('bitmaprenderer');
+    createImageBitmap(canvas).then(t.step_func_done(function(image) {
+        assert_equals(image.width, 10);
```
junov: should be assert_equals(image.width, width);

imagebitmap-renderingcontext/bitmaprenderer-as-imagesource.html
```diff
@@ -0,0 +1,86 @@
...
+
+
+async_test(function(t) {
+    var canvas = document.createElement("canvas");
+    canvas.width = width;
+    canvas.height = height;
+    var ctx = canvas.getContext('bitmaprenderer');
+    ctx.transferFromImageBitmap(null);
+    createImageBitmap(canvas).then(t.step_func_done(function(image) {
+        assert_equals(image.width, 10);
```
junov: same here

imagebitmap-renderingcontext/bitmaprenderer-as-imagesource.html
```diff
@@ -0,0 +1,86 @@
...
+}
+
+function consumeImageBitmap(image, t)
+{
+    var myCanvas = document.createElement('canvas');
+    myCanvas.width = myCanvas.height = 20;
+    var myCtx = myCanvas.getContext('bitmaprenderer');
+    myCtx.transferFromImageBitmap(image);
+
+    return createImageBitmap(myCanvas).then(t.step_func_done(function(imageBitmap) {
```
junov: why "return"?

xidachen:
@junov Done, all comments are addressed.

imagebitmap-renderingcontext/bitmaprenderer-as-imagesource.html
```diff
@@ -0,0 +1,86 @@
...
+        assert_equals(imageBitmap.width, 20);
```
junov: According to the spec imageBitmap should be 10x10, the size of the image that was transferred into myCtx.  What does Firefox do here?

junov (changes requested):

xidachen:
OK, I have update the tests. Firefox gives 20*20. I have updated the bug here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1316420

Chrome crashes when doing createImageBitmap(myCanvas), bug filed here:
https://bugs.chromium.org/p/chromium/issues/detail?id=663793

junov:
> Firefox gives 20*20. I have updated the bug here:
> https://bugzilla.mozilla.org/show_bug.cgi?id=1316420

Actually, what I was going to say in that case is that we should update the spec to reflect the only currently shipping implementation.

xidachen:
I see, that makes sense. I have reverted the last commit.

junov:
I retract that last comment, back in April I had an e-mail exchange with the implementer from Mozilla, Morris Tseng, around the time I was finalizing the spec. He was in agreement with what I was proposing, which was different from the Firefox implementation:

>After reading those information, I realized that the purpose of ImageBitmapRenderingContext provided a faster way to display ImageBitmap. So I think option 2 is better approach. If we choose option 2, we don't need keep both original bitmap and displayed bitmap which causes an extra copy. If user want to display ImageBitmap with different intrinsic size (stretch, crop..etc) , he should use drawImage instead.

The option 2 he refers to is the option of using the size of the transferred bitmap as the intrinsic size of the canvas (hence ignoring canvas.width/height)

So let's stick with the spec as it is. It is actually important that it remain this way for perf+memory reasons. 

I guess they just forgot to address this before shipping the feature in FF.

xidachen:
@junov In the new commits, I have changed the test to follow the spec. A comment is added too. Please review.

imagebitmap-renderingcontext/bitmaprenderer-as-imagesource.html
```diff
@@ -0,0 +1,90 @@
...
+function consumeImageBitmap(image, t)
+{
+    var myCanvas = document.createElement('canvas');
+    myCanvas.width = myCanvas.height = 20;
+    var myCtx = myCanvas.getContext('bitmaprenderer');
+    myCtx.transferFromImageBitmap(image);
+
+    createImageBitmap(myCanvas).then(t.step_func_done(function(imageBitmap) {
+        // Per spec, when transferFromImageBitmap happens, the transferred
+        // ImageBitmap (|image| here) should be the intrinsic size of
```
junov: "should be" -> "must determine"

imagebitmap-renderingcontext/bitmaprenderer-as-imagesource.html
```diff
@@ -0,0 +1,90 @@
...
+        // myCanvas, and hence myCanvas.width/height is ignored. Therefore,
+        // this created |imageBitmap| should have the same size as the |image|.
```
junov: No need for "this created"

junov (approved):

PR #4171, opened 2016-11-04
Add tests for the document promises

domenic:
Introduced in https://github.com/whatwg/html/pull/1936.


zcorpan (approved):

gsnedders:
Why has this not get merged?

domenic:
The feature proposal is kind of in limbo right now; it has vague support from two implementers, but no concrete movement toward implementation. As such it seems better to keep the spec and tests as PRs until we get some assurance that merging them would not just create more spec-fiction.

Issue #4041, opened 2016-10-20
reconnectToPresentation_success-manual test is incorrect


tidoust:
The [reconnect to a presentation](https://github.com/w3c/web-platform-tests/blob/master/presentation-api/controlling-ua/reconnectToPresentation_success-manual.html) test needs to be updated:
1. It should not use `localStorage`: it may not be supported and there does not seem to be any need for it (unless the goal is to test reconnection after a refresh of the page, which could be the purpose of another test)
2. Most of the code gets run outside any `promise_test` or `async_test` container, which means the test will just silently crash e.g. if the presentation connection cannot be established
3. The two buttons should not appear at the same time.

Also, since reconnection does not require a user gesture, I wonder if the test should not be re-written to attempt the reconnection automatically: it could call `close()`, wait for the `close` event and call `reconnect` at that point. That would slightly improve the automation of the test, as it would avoid requiring two user inputs.

louaybassbouss:
Thx @tidoust, @obstschale  can you please submit a PR that fixes this issue

louaybassbouss:
@tidoust PR #4172 should fix this issue. Feel free to close after the PR is merged.
@obstschale please focus on other tests

PR #4173, opened 2016-11-05
fix issue #4041

louaybassbouss:
This PR fixes issue #4041

### DIFF
--- a/presentation-api/controlling-ua/reconnectToPresentation_success-manual.html
+++ b/presentation-api/controlling-ua/reconnectToPresentation_success-manual.html
@@ -19,7 +19,7 @@
     startBtn.onclick = function () {
         startBtn.disabled = true;
         promise_test(function (t) {
-            var request = new PresentationRequest(presentationUrls[1]);
+            var request = new PresentationRequest(presentationUrls);
             var presentationId = null;
             return request.start().then(function (connection) {
                 presentationId = connection.id;
--- a/presentation-api/controlling-ua/reconnectToPresentation_success-manual.html
+++ b/presentation-api/controlling-ua/reconnectToPresentation_success-manual.html
@@ -1,72 +1,49 @@
 <!DOCTYPE html>
 <meta charset="utf-8">
-<title>Presentation API, reconnect to presentation tests for Controlling User Agent (success - manual test)</title>
+<title>Reconnect to presentation success manual test</title>
 <link rel="author" title="Marius Wessel" href="http://www.fokus.fraunhofer.de">
-<link rel="help" href="http://w3c.github.io/presentation-api/#dfn-controlling-user-agent">
+<link rel="author" title="Louay Bassbouss" href="http://www.fokus.fraunhofer.de">
+<link rel="help" href="http://w3c.github.io/presentation-api/#dom-presentationrequest-reconnect">
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <script src="common.js"></script>
 
-<ol>
-    <li>Start the presentation with the blue button.</li>
-    <li>After the presentation ist running successfully, disconnect the connected device.</li>
-    <li>Reconnect the device and press the green button, to start the reconnection process.</li>
-</ol>
-
-<button id="startBtn" onclick="startPresentation()" style="background-color: #2aabd2;">Start Presentation</button>
-<button id="reconnectBtn" onclick="reconnectToPresentation()" style="background-color: #5cb85c;">Reconnect Presentation
-    Test
-</button>
+<p>Click the button below to start the manual test. Select a presentation device after the selection dialog is prompted. The test assumes that at least one presentation device is available. The test passes if a "PASS" result appears.</p>
+<button id="startBtn">Start Test</button>
 
 <script>
-    // disable the timeout function for the tests
-    // and call 'done()' when the tests cases are finished.
+    // disable timeout for manual tests
     setup({explicit_timeout: true});
-
-    // ----------
-    // DOM Object
-    // ----------
-    var startBtn = document.getElementById("startBtn"),
-            reconnectBtn = document.getElementById("reconnectBtn");
-
-    // ------------
-    // Request init
-    // ------------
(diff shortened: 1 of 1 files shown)


presentation-api/controlling-ua/reconnectToPresentation_success-manual.html
```diff
@@ -1,72 +1,49 @@
...
-        }, 'The reconnection was successful.');
-    }
-    // ----------------------------------------------
-    // Reconnect to Presentation Test (success) - end
-    // ----------------------------------------------
+                // close connection. test continue after onclose event is fired.
+                oldConnection.close();
+                // the test fails if after 5s the close event is not fired
+                t.step_timeout(function () {
+                    !oldConnectionClosed && assert_unreached("Close event on old connection not fired");
```
tidoust: The assert will make the test `FAIL` but I would rather expect the test to fail with a `TIMEOUT` status.

The "Harness timeout" section in the [Testharness.js API Documentation](http://testthewebforward.org/docs/testharness-library.html) mentions the `force_timeout` method (not exactly for the same purpose and as "last resort"): "it is possible to use `test.force_timeout()` in place of `assert_unreached()`, to immediately fail the test but with a status of `TIMEOUT`".

Beware though, you need to call `t.done()` as well after the call to `t.force_timeout()`.

presentation-api/controlling-ua/reconnectToPresentation_success-manual.html
```diff
@@ -1,72 +1,49 @@
...
-    // ----------------------------------------
-    var startPresentation = function () {
-        return request.start()
-                .then(function (connection) {
-                    // save the presentation id within the
-                    // local storage of the browser
-                    localStorage["presId"] = connection.id;
+    async_test(function (t) {
+        var startBtn = document.getElementById("startBtn");
+        startBtn.onclick = t.step_func(function () {
```
tidoust: I'm the one who commented that the code was not wrapped in an `async_test` call but I think this goes too far. I would follow a "user clicks the button to start the test" approach and re-write as:

```js
startBtn.onclick = function () {
  startBtn.disabled = true;
  async_test(function (t) { ... });
};
```

(or use an `onclick` attribute that call a `startPresentation` function as in other tests)

presentation-api/controlling-ua/reconnectToPresentation_success-manual.html
```diff
@@ -1,72 +1,49 @@
...
-    // ----------------------------------------------
+                // close connection. test continue after onclose event is fired.
+                oldConnection.close();
+                // the test fails if after 5s the close event is not fired
+                t.step_timeout(function () {
+                    !oldConnectionClosed && assert_unreached("Close event on old connection not fired");
+                },5000);
+            })).catch(t.step_func(function () {
+                assert_unreached("Error on start presentation");
+            }));
```
tidoust: So, I believe the whole code is correct, but I wonder whether we could improve readability a bit. With the Presentation API, we are in `Promise` land and yet the code needs to catch and assert rejected promises as in the above lines, which would better be caught by the test harness itself. The code also needs to call `step_func` here and there. This makes it hard to read and maintain.

Concretely, I wonder if the following approach could be easier to read and maintain:

```js
function startPresentation() {
        document.getElementById("startBtn").disabled = true;

        promise_test(function (t) {
            var presentationId = null;
            return request.start()
                .then(function (connection) {
                    // No more user input needed, re-enable test timeout
                    t.step_timeout(function () {
                        t.force_timeout();
                        t.done();
                    }, 10000);

                    // Close connection and wait for "close" event
                    return new Promise(function (resolve) {
                        connection.onclose = function () {
                            presentationId = connection.id;
                            resolve();
                        };
                        connection.close();
                    });
                })
                .then(function () {
                    // Connection now closed, let's reconnect to it
                    return request.reconnect(presentationId);
                })
                .then(function (connection) {
                    assert_equals(connection.state, "connecting", "connection should be in 'connecting' state");
                    assert_equals(connection.id, presentationId, "Ids of old and new connections must be equal");
                });
        });
    }
```

I don't think that's perfect either. Using `promise_test` has the advantage of no longer requiring "catch and assert" code. One drawback is the need to create new Promise to wrap the `close` event handler.

(Note I switched to `force_timeout` in that code and use it as a "global" timeout for the test. This is on purpose to avoid looping forever in the "automated" part of the test if e.g. the `reconnect` Promise never gets resolved or rejected. It would be better to have a way to tell the test harness to simply restart the timeout, but I don't think that's possible)

Anyway, this is editorial, feel free to discard. I find these tests hard to write, so I'm trying to find recipes to simplify authoring... but what works for me may not work for others :)

tidoust (commented):
Thanks, @louaybassbouss! See comments inline

presentation-api/controlling-ua/reconnectToPresentation_success-manual.html
tidoust: Thinking about this again, the test harness `EventWatcher` class could further help streamline the above code (avoiding the call to `new Promise` in particular):

```js
function startPresentation() {
    document.getElementById("startBtn").disabled = true;

    promise_test(function (t) {
        var presentationId = null;
        return request.start()
            .then(function (connection) {
                presentationId = connection.id;

                // No more user input needed, re-enable test timeout
                t.step_timeout(function () {
                    t.force_timeout();
                    t.done();
                }, 5000);

                // Close connection and wait for "close" event
                connection.close();
                var eventWatcher = new EventWatcher(t, connection, 'close');
                return eventWatcher.wait_for('close');
            })
            .then(function () {
                // Connection now closed, let's reconnect to it
                return request.reconnect(presentationId);
            })
            .then(function (connection) {
                assert_equals(connection.state, "connecting", "connection should be in 'connecting' state");
                assert_equals(connection.id, presentationId, "Ids of old and new connections must be equal");
            });
    });
}
```

louaybassbouss:
Thanks @tidoust for review I addressed you comments in last commit

tidoust (approved):
Thks, @louaybassbouss! That looks good to me.

PR #4174, opened 2016-11-07
[presentation-api] update a test to check start() event order

### LINKED ISSUE #4168: [presentation-api] Test coverage for start() event order
The Presentation API controller tests have coverage for `onconnectionavailable`, `onconnect`, `PresentationConnection.state`, and resolution of the Promise from `start()` in isolation, but I don't see a test case that makes sure these event loops happen in a specific order.  

This issue tracks the work to add a test case to check the order of the following:
1. Promise from `start()` resolves
2. `connectionavailable` fires
3. state is `connecting`
4. `connect` fires
5. state is `connected`

tomoyukilabs:
This PR fixes #4168. startNewPresentation_success-manual.html is modified to check the following event order:

1. Promise from start() resolves
2. connectionavailable fires and then state is `connecting`
3. connect fires and then state is `connected`

### DIFF
--- a/presentation-api/controlling-ua/startNewPresentation_success-manual.html
+++ b/presentation-api/controlling-ua/startNewPresentation_success-manual.html
@@ -34,7 +34,14 @@
     presentBtn.onclick = function () {
         presentBtn.disabled = true;
         promise_test(function (t) {
-            var phase = -1, actual = -1;
+            var request = new PresentationRequest(presentationUrls);
+            return new Promise(function(resolve, reject) {
+                request.onconnectionavailable = t.step_func(function(evt) {
+                    // Check the order #2: "connectionavailable" event fires
+                    assert_equals(order[step], 'FIRE_CONNECTIONAVAILABLE', 'A "connectionavailable" event fires after Promise from PresentationRequest.start() resolves.');
+                    step++;
+                    assert_equals(connection, evt.connection, 'Both Promise from PresentationRequest() and a "connectionavailable" event handler receive the same presentation connection.');
+                });
 
             // increment the count in the actual event order
             var count = function(evt) { actual++; return evt; };
--- a/presentation-api/controlling-ua/startNewPresentation_success-manual.html
+++ b/presentation-api/controlling-ua/startNewPresentation_success-manual.html
@@ -48,13 +48,15 @@
                         assert_equals(connection.state, 'connecting', 'The initial state of the presentation connection is "connecting".');
                         // Check the order #3: "connected" event fires
                         connection.onconnect = function() {
-                            assert_equals(order[step], 'FIRE_CONNECTED', 'A connect event fires after Promise from PresentationRequest.start() resolves and a "connectionavailable" event fires.');
-                            assert_equals(connection.state, 'connected', 'The state of the presentation connection is "connected" when a "connect" event fires.');
+                            assert_equals(order[step], 'FIRE_CONNECTED', 'A connected event fires after Promise from PresentationRequest.start() resolves and a "connectionavailable" event fires.');
+                            assert_equals(connection.state, 'connected', 'The state of the presentation connection is "connected" when a "connected" event fires.');
                             resolve();
                         };
+
                         t.step_timeout(function() {
                             reject('A "connect" event did not fire when the presentation connection was established.')
                         }, 5000);
+
                         // Check, if the connection ID is set
                         assert_true(!!connection.id, 'The connection ID is set.');
                         // Check the type of the connection.id
--- a/presentation-api/controlling-ua/startNewPresentation_success-manual.html
+++ b/presentation-api/controlling-ua/startNewPresentation_success-manual.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <meta charset="utf-8">
-<title>Checking the chain of events when starting a new presentation</title>
+<title>Presentation API, start new presentation tests for Controlling User Agent (success - manual test)</title>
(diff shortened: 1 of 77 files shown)


presentation-api/controlling-ua/startNewPresentation_success-manual.html
```diff
@@ -21,16 +31,39 @@
...
+                            assert_equals(order[step], 'FIRE_CONNECTED', 'A connect event fires after Promise from PresentationRequest.start() resolves and a "connectionavailable" event fires.');
+                            assert_equals(connection.state, 'connected', 'The state of the presentation connection is "connected" when a "connect" event fires.');
+                            resolve();
+                        };
+                        t.step_timeout(function() {
+                            reject('A "connect" event did not fire when the presentation connection was established.')
+                        }, 5000);
+
                         // Check, if the connection ID is set
                         assert_true(!!connection.id, 'The connection ID is set.');
```
tidoust: Could you move these `assert_true` checks together with the check on the initial connection state? Here, they seem to be done later for a specific reason, and I don't think there is any reason to do so in practice.

Also, you can probably drop the comments, these asserts are self-explanatory.

presentation-api/controlling-ua/startNewPresentation_success-manual.html
```diff
@@ -21,16 +31,39 @@
...
+                        connection = c;
+
+                        // Check the order #1: Promise from start() resolves
+                        assert_equals(order[step], 'RESOLVE_START', 'Promise from PresentationRequest.start() resolves before connectionavailable event is fired.');
+                        step++;
 
                         // Check the initial state of a presentation connection
                         assert_equals(connection.state, 'connecting', 'The initial state of the presentation connection is "connecting".');
 
+                        // Check the order #3: "connected" event fires
```
tidoust: s/connected/connect/

presentation-api/controlling-ua/startNewPresentation_success-manual.html
```diff
@@ -39,11 +72,11 @@
 
                         // Check the instance of the connection
                         assert_true(connection instanceof PresentationConnection, 'The connection is an instance of PresentationConnection.');
-                    });
+                    }), function(e) { reject(e); });
+            });
         }, "The presentation was started successfully.");
```
tidoust: Not your fault, but that test label is not very explicit. I would drop it from there and update the `<title>` to something like: "Checking the chain of events when starting a new presentation"

presentation-api/controlling-ua/startNewPresentation_success-manual.html
```diff
@@ -39,11 +72,11 @@
 
                         // Check the instance of the connection
                         assert_true(connection instanceof PresentationConnection, 'The connection is an instance of PresentationConnection.');
-                    });
+                    }), function(e) { reject(e); });
+            });
```
tidoust: As often with these tests that mix promises and events, I'm struggling to find the right "recipe" for the test to be clear and simple. Having a `catch` function that calls `reject` works but does not "look nice" for instance.

In this case, I wonder if we could not benefit from the `EventWatcher` class that the test harness exposes and which allows to handle events through promises. Applied to this test, this could give:

```js
        promise_test(function (t) {
            return request.start()
                .then(function (c) {
                    // Phase #1: Promise is resolved
                    connection = c;

                    // No more user input needed, re-enable timeout
                    t.step_timeout(function() {
                        t.force_timeout();
                        t.done();
                    }, 5000);

                    // Check the initial state of the presentation connection
                    assert_equals(connection.state, 'connecting', 'Check the initial state of the presentation connection.');
                    assert_true(!!connection.id, 'The connection ID is set.');
                    assert_true(typeof connection.id === 'string', 'The connection ID is a string.');
                    assert_true(connection instanceof PresentationConnection, 'The connection is an instance of PresentationConnection.');

                    var eventWatcher = new EventWatcher(t, request, 'connectionavailable');
                    return eventWatcher.wait_for('connectionavailable');
                })
                .then(function (evt) {
                    // Phase #2: "connectionavailable" event fired
                    assert_equals(connection, evt.connection, 'Both Promise from PresentationRequest() and a "connectionavailable" event handler receive the same presentation connection.');
                    
                    var eventWatcher = new EventWatcher(t, connection, 'connect');
                    return eventWatcher.wait_for('connect');
                })
                .then(function () {
                    // Phase #3: "connect" event fired
                    assert_equals(connection.state, 'connected', 'The state of the presentation connection is "connected" when a "connect" event fires.');
                });
        });
```

This code would be slightly less complete than yours: it ensures things happen in the right order, but e.g. will not fail if a browser fires a `connectionavailable` event, then resolves the start promise and then fires _another_ `connectionavailable` event for instance. That said, one advantage is that the different phases appear more clearly and in logical order in the code. Also, `EventWatcher` will make the test fail if other events get fired at the request or at the connection object.

What do you think?

tidoust (commented):
See comments inline, including a proposal to refactor the test using `EventWatcher`.

presentation-api/controlling-ua/startNewPresentation_success-manual.html
tomoyukilabs: Thanks! It looks almost good to me. My concern is simply that we must wait for timeout even if the event is fired in incorrect order. I will PR which adds checking the event order.

tomoyukilabs:
Sorry, I made a mistake and failed to make a commit correctly. I close this, and submit a PR again.

tomoyukilabs:
I submitted a PR again: #4187

PR #4175, opened 2016-11-07
[presentation-api] Fix handling when PresentationRequest not found

### LINKED ISSUE #4039: Test on presentation.defaultRequest may crash without reporting results

tomoyukilabs:
This PR fixes #4039.

* `new PresentationRequest()` is moved within `async_test()`.
* The "Not Supported" button is disabled or the whole "If your browser does not support" text is hidden when unnecessary.

### DIFF
--- a/presentation-api/controlling-ua/defaultRequest_success-manual.html
+++ b/presentation-api/controlling-ua/defaultRequest_success-manual.html
@@ -37,9 +37,6 @@
             assert_true(typeof connection.id === 'string', 'The connection ID is a string.');
             assert_true(connection instanceof PresentationConnection, 'The connection is an instance of PresentationConnection.');
         });
-        document.getElementById('notsupported').onclick = t.step_func_done(function() {
-            assert_unreached('This browser does not support defaultRequest.');
-        });
         button.onclick = t.step_func_done(function() {
             button.disabled = true;
             assert_unreached('This browser does not support defaultRequest.');
--- a/presentation-api/controlling-ua/defaultRequest_success-manual.html
+++ b/presentation-api/controlling-ua/defaultRequest_success-manual.html
@@ -1,15 +1,21 @@
 <!DOCTYPE html>
 <meta charset="utf-8">
-<title>Presentation API, testing to start a new presentation with an default request setup. (success - manual)</title>
+<title>[Optional] Starting a presentation from the browser using a default presentation request. (success - manual)</title>
 <link rel="author" title="Marius Wessel" href="http://www.fokus.fraunhofer.de">
 <link rel="author" title="Tomoyuki Shimizu" href="https://github.com/tomoyukilabs">
-<link rel="help" href="http://w3c.github.io/presentation-api/#dfn-controlling-user-agent">
+<link rel="help" href="http://w3c.github.io/presentation-api/#dom-presentation-defaultrequest">
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <script src="common.js"></script>
 
-<p>Click the button or the menu item for presentation on your browser (for example, "Cast"), to start the manual test.</p>
-<p id="notice">If your browser does not support <code>defaultRequest</code>, please click this button: <button id="notsupported">Not Supported</button>
+<p>
+    Click the button or the menu item for presentation on your browser (for example, "Cast"), <br>
+    to start the manual test, and select a presentation display when prompted to do so.<br>
+    The test passes if a "PASS" result appears.
+</p>
+<p id="notice">
+    If your browser does not support <code>defaultRequest</code>, please click this button: <button id="notsupported">Not Supported</button>
+</p>
 
 <script>
     // disable the timeout function for the tests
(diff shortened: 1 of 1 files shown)


presentation-api/controlling-ua/defaultRequest_success-manual.html
```diff
@@ -9,34 +9,52 @@
 <script src="common.js"></script>
 
 <p>Click the button or the menu item for presentation on your browser (for example, "Cast"), to start the manual test.</p>
```
tidoust: I suggest to complete with "and select a presentation display when prompted to do so. The test passes if a "PASS" result appears" to clarify what the user needs to do.

presentation-api/controlling-ua/defaultRequest_success-manual.html
```diff
@@ -9,34 +9,52 @@
...
 
     // ------------------------------
     // Start New Presentation with
     // 'default request' Test - BEGIN
     // ------------------------------
     async_test(function(t) {
-        navigator.presentation.defaultRequest.onconnectionavailable = t.step_func_done(function (evt) {
-            var connection = evt.connection;
+        var request = new PresentationRequest(presentationUrls);
+        if('defaultRequest' in navigator.presentation) {
```
tidoust: If I understand things correctly, the reason you checks for attribute presence is to remove the notice before the end of the call. These checks would not really be needed otherwise: the code would simply throw an exception that the test harness would pick up and report.

I wonder if that could better be expressed using `add_cleanup`, to reduce the number of nested levels, as in:

```js
    async_test(function(t) {
        t.add_cleanup(function () {
            notice.parentNode.removeChild(notice);
        });
        button.onclick = t.step_func_done(function () {
            assert_unreached('This browser does not support defaultRequest.');
        });
        navigator.presentation.defaultRequest = request;
        request.onconnectionavailable = t.step_func_done(function (evt) {
            var connection = evt.connection;
            assert_equals(connection.state, 'connecting', 'The initial state of the presentation connection is "connecting".');
            assert_true(!connection.id, 'The connection ID is set.');
            assert_true(typeof connection.id === 'string', 'The connection ID is a string.');
            assert_true(connection instanceof PresentationConnection, 'The connection is an instance of PresentationConnection.');
        });
    });
```

This code would remove the notice paragraph even when the user clicks the "Not supported" button, but that does not seem to be a big deal. Anyway, that's all purely "cosmetic", so feel free to discard this comment.

presentation-api/controlling-ua/defaultRequest_success-manual.html
```diff
@@ -9,34 +9,52 @@
...
+                notice.parentNode.removeChild(notice);
+                assert_unreached('This browser does not support connectionavailable event.');
+            }
+        }
+        else {
+            notice.parentNode.removeChild(notice);
             assert_unreached('This browser does not support defaultRequest.');
-        });
+        }
     }, '[Optional] The presentation was started successfully.');
```
tidoust: The test label is not very descriptive. I would drop it from here and update the `<title>` tag to something like "[Optional] Starting a presentation from the browser using a default presentation request"

tidoust (commented):
See inline for a couple of additional comments on parts you did not touch ;)

Also, one suggestion to reduce the number of nested levels in the code, but the code should work as-is, feel free to discard.

presentation-api/controlling-ua/defaultRequest_success-manual.html
tomoyukilabs: LGTM!

tomoyukilabs:
Thanks so much, @tidoust! I updated the test. Could you please review it?

presentation-api/controlling-ua/defaultRequest_success-manual.html
```diff
@@ -27,35 +33,30 @@
...
+        // set an event handler to make the test fail when the button is clicked
+        button.onclick = t.step_func_done(function() {
             assert_unreached('This browser does not support defaultRequest.');
-        }
-    }, '[Optional] The presentation was started successfully.');
+        });
+        // check if defaultRequest is an instance of PresentationRequest or null
+        var defaultRequest = navigator.presentation.defaultRequest;
+        assert_true(defaultRequest instanceof PresentationRequest || defaultRequest === null,
+            'navigator.presentation.defaultRequest must be an instance of PresentationRequest or null.');
```
tomoyukilabs: [The current spec](http://w3c.github.io/presentation-api/#dom-presentation-defaultrequest) says:

> "The `defaultRequest` attribute MUST return the default presentation request if any, `null` otherwise."

However, its default value is not clearly mentioned yet. Therefore, this assertion checks if the default value is either an instance of PresentationRequest or null, just in case.

tidoust: Hmm, not sure I get this. The Common Idioms section says: "the default presentation request, **which is initially set to null**, represents..."
http://w3c.github.io/presentation-api/#dfn-default-presentation-request

presentation-api/controlling-ua/defaultRequest_success-manual.html
```diff
@@ -1,43 +1,62 @@
 <!DOCTYPE html>
 <meta charset="utf-8">
-<title>Presentation API, testing to start a new presentation with an default request setup. (success - manual)</title>
+<title>[Optional] Starting a presentation from the browser using a default presentation request. (success - manual)</title>
```
tidoust: Not a big deal, but any particular reason to keep "(success - manual)" in this title? They don't mean much to me at least.

presentation-api/controlling-ua/defaultRequest_success-manual.html
tomoyukilabs: Sorry, I missed it.

tidoust (approved):

tomoyukilabs:
Thanks a lot, @tidoust!

PR #4176, opened 2016-11-07
Improve navigate-child-function.html and navigate-child-function-src.html.

Ms2ger:


Ms2ger:
Code reviewed upstream.

PR #4179, opened 2016-11-07
Properly expect a syntax error with :scope selectors in Element.closest

Ms2ger:


Ms2ger:
Code reviewed upstream.

ayg:
@Ms2ger @nox I don't understand how this follows the spec.  The spec says to throw a syntax error only if the selector doesn't parse, and it seems like this selector does parse?  I don't know how :scope works, so perhaps it should return null, but I don't see how it can throw.  Firefox and Chrome fail the first three of these four tests, although Edge passes.

Ms2ger:
https://github.com/servo/servo/pull/14080

CC @SimonSapin

ayg:
Oh, it was Servo, that's why I couldn't find it.  I assumed Gecko.

That PR refers to w3c/csswg-drafts@9b7faa3, which removes :scope from some version of the spec.  It seems to be in the current version: https://drafts.csswg.org/selectors-4/#the-scope-pseudo  So is it still a parse error?  Firefox and Chrome both support it in practice, and I don't see why it should be considered non-conforming just because it was dropped from some particular version of a standard for procedural reasons (I assume to get to REC).

ayg:
@Ms2ger @SimonSapin Ping?  Should I submit a PR to revert this?

Ms2ger:
I don't have an opinion, I just upstreamed the patch.

nox:
> @Ms2ger @nox I don't understand how this follows the spec. The spec says to throw a syntax error only if the selector doesn't parse, and it seems like this selector does parse? I don't know how :scope works, so perhaps it should return null, but I don't see how it can throw. Firefox and Chrome fail the first three of these four tests, although Edge passes.

I remember absolutely nothing about this, sorry.

PR #4182, opened 2016-11-08
Move URL origin tests to new file

achristensen07:
Origin differences cover up parser differences in the results.  Origin conformance is implemented differently than other attributes in the constructor test

### DIFF
--- a/url/url-origin.html
+++ b/url/url-origin.html
@@ -1,0 +1,37 @@
+<!doctype html>
+<meta charset=utf-8>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<div id=log></div>
+<script>
+function runURLOriginTests() {
+  var setup = async_test("Loading data…")
+  setup.step(function() {
+    var request = new XMLHttpRequest()
+    request.open("GET", "urltestdata.json")
+    request.send()
+    request.responseType = "json"
+    request.onload = setup.step_func(function() {
+      runURLTests(request.response)
+      setup.done()
+    })
+  })
+}
+
+function bURL(url, base) {
+  return new URL(url, base || "about:blank")
+}
+
+function runURLTests(urltests) {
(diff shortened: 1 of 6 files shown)


url/url-origin.html
```diff
@@ -0,0 +1,46 @@
...
+}
+
+function bURL(url, base) {
+  return new URL(url, base || "about:blank")
+}
+
+function runURLTests(urltests) {
+  for(var i = 0, l = urltests.length; i < l; i++) {
+    var expected = urltests[i]
+    if (typeof expected === "string") continue // skip comments
```
domenic: Maybe just skip the test `if (!("origin" in expected))` since it's already tested elsewhere? Then you can omit the expected.failure stuff too.

url/url-origin.html
```diff
@@ -0,0 +1,46 @@
...
+          bURL(expected.input, expected.base)
+        })
+        return
+      }
+
+      var url = bURL(expected.input, expected.base)
+      if ("origin" in expected) {
+        assert_equals(url.origin, expected.origin, "origin")
+      }
+    }, "Parsing: <" + expected.input + "> against <" + expected.base + ">")
```
domenic: Probably change "Parsing: " to "Origin parsing: " or something

domenic (commented):
Shouldn't this kind of change also be done for a-element.js too?

achristensen07:
I believe I have addressed domenic's feedback.  Could this be merged?

url/url-origin.html
```diff
@@ -0,0 +1,37 @@
+<!doctype html>
```
youennf: Shouldn't this file be named url-constructor-origin.html

youennf:
Sounds good to me. If a reviewer can validate this, I'll be happy to merge it.

domenic:
@achristensen07 sorry, I didn't see that you pushed a new commit. Thanks for the ping!

PR #4187, opened 2016-11-09
[presentation-api] update a test to check start() event order (again)

### LINKED ISSUE #4168: [presentation-api] Test coverage for start() event order

tomoyukilabs:
This PR replaces #4174, and fixes #4168.

### DIFF
--- a/presentation-api/controlling-ua/startNewPresentation_success-manual.html
+++ b/presentation-api/controlling-ua/startNewPresentation_success-manual.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <meta charset="utf-8">
-<title>Presentation API, start new presentation tests for Controlling User Agent (success - manual test)</title>
+<title>Checking the chain of events when starting a new presentation</title>
 <link rel="author" title="Marius Wessel" href="http://www.fokus.fraunhofer.de">
 <link rel="author" title="Tomoyuki Shimizu" href="https://github.com/tomoyukilabs">
 <link rel="help" href="http://w3c.github.io/presentation-api/#dfn-controlling-user-agent">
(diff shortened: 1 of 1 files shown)


tomoyukilabs:
@tidoust This code is based on https://github.com/w3c/web-platform-tests/pull/4174#discussion_r87003676, and I added code to check event order. If this code might look not so clear, I could revert to your code. What do you think?

tidoust:
Hey, that looks nice, @tomoyukilabs! :)

For what it's worth, on top of making the test fail nicely (and without timeout) when the order of events is incorrect, your code avoids having to answer the following question that came to mind yesterday evening: my suggested code started to listen to the `connect` event as part of the resolution (a) of a `Promise` created under the hoods while handling the `connectionavailable` event. User agents may queue a task to fire the `connect` event (b) at any time after the `connectionavailable` event.

In theory, I think (b) cannot happen before (a) because user agents queue a _microtask_ for promise resolution, whereas they queue a _macrotask_ for event firing. Microtasks should run to completion first, or rather they should run right after the current macrotask and before the next one.

In practice, some web browsers do not seem to run Promise resolutions at the end of a callback as they should, see related [Firefox ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1193394) for instance. I'm not really sure whether that would have affected this code, but it's probably a good idea to take a conservative approach in tests and not rely on promise resolution code being run before other kinds of tasks.

tomoyukilabs:
@tidoust Thanks so much for review and invaluable comments!

> In practice, some web browsers do not seem to run Promise resolutions at the end of a callback as they should, see related [Firefox ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1193394) for instance.

Oh, I was surprised to see the results of [the demo](http://jsbin.com/kaqume/1/edit?js,console,output) mentioned in that ticket...

PR #4188, opened 2016-11-09
Test that SVGViewElement's viewTarget attribute is removed

foolip:
https://svgwg.org/svg2-draft/single-page.html#changes-linking
https://codereview.chromium.org/2488703003/


foolip:
Results for the new test are as expected. Chrome will start passing soonish.

nikosandronikos:
Looks good.

Issue #4181, opened 2016-11-08
[presentation-api] complement PresentationAvailability tests


tomoyukilabs:
I am now considering adding several test cases to complement PresentationAvailability tests. So I would like to confirm if there is any person currently preparing PresentationAvailability tests or not.

If not, I would like to first add a test case for `change` event, and then I will try to find any other missing test cases in existing tests, e.g. getAvailability.html.

tomoyukilabs:
Many thanks, @tidoust !

PR #4192, opened 2016-11-10
[presentation-api] Add a test case to check "change" event on PresentationAvailability

tomoyukilabs:
This PR partially addresses #4181. This test checks if a "change" event is fired when a presentation display become available and unavailable.

Currently, the timeout is set to 90 seconds to observe change of the presentation availability twice; when a presentation display becomes available, and when it becomes unavailable. Especially, it would often be possible that detecting transition to "unavailable" might take a long time. I wonder if the timeout duration could be much smaller, though.


presentation-api/controlling-ua/PresentationAvailability_onchange.html
```diff
@@ -0,0 +1,66 @@
...
+        previousState = availability.value;
+
+        // add an instruction notice
+        notice.innerHTML += 'Please make your presentation display '
+                          + (previousState ? 'unavailable' : 'available')
+                          + ' and wait for a minute...';
+
+        // wait until a "change" event is fired
+        if(!eventWatcher)
+          eventWatcher = new EventWatcher(t, availability, ['change', 'change']);
```
tidoust: Hmmm, `change` should not appear twice here. That list represents the list of event types that you want the test harness to listen to, not the sequence of events that you want to check. The sequence of events should rather be given to `wait_for`.

With this code, the test harness will attach two event listeners to the `change` event, which will fire one after the other when a `change` event is received. Now, it will work in our case but that's just "luck": the first one will process the event and resolve the Promise returned by `wait_for`; the second one will try to reject the same Promise (but fail since the Promise has already been resolved) because the event has already been processed. This would break if we were waiting for two change events, as in `wait_for(['change', 'change'])`, before continuing the test.

presentation-api/controlling-ua/PresentationAvailability_onchange.html
```diff
@@ -0,0 +1,66 @@
...
+      }, 90000);
+
+      var wait = t.step_func(function(a) {
+        availability = a;
+
+        // save the current value of the presentation availability
+        previousState = availability.value;
+
+        // add an instruction notice
+        notice.innerHTML += 'Please make your presentation display '
```
tidoust: The user may have more than one presentation display to unplug, I would use a plural form in "Please make your presentation displays..."

tidoust (commented):
Thanks @tomoyukilabs !

A couple of comments inline and a couple of comments below:
1. the test should be flagged as `-manual`. While the page itself does not require user input, it does require user action, so it should not run with automated tests.
2. I think there should be no timeout while the user performs an action. The user may need to go to the other side of the room to find the right cable to unplug for instance. This may take time. I would:
   i.  reset the timeout at the beginning of the `wait` function (calling `clearTimeout` on the timeout returned by the call to `step_timeout`).
   ii. ask the user to click a button when she's done making her presentation displays (un-)available (but note you should start waiting for the `change` event right away not to miss it). When the button gets clicked, disable (or remove) it and call `step_timeout` again to re-enable the timeout and return the `wait_for` promise.

I do not know whether we can reduce the timeout each time. Transition to unavailable may indeed take time.

tomoyukilabs:
@tidoust Thanks for your review! I've changed the file name and fixed the code. Could you please review it?

tidoust (dismissed):

tidoust (commented):
The test does not check that the initial availability value is `true` and probably should. I would add `assert_true(availability.value)` at the beginning of the `setup` method.

tomoyukilabs:
Should we assume that any presentation display must be made available before starting this test?

tidoust:
Ah, right, I thought the test always started by asking the user to *unplug* presentation displays, but it actually adapts to the starting state of the user. Sorry, I missed that! It's all fine then.

The other tests in the Presentation API test suite somewhat assume that there is a presentation display available when the test starts. That's probably something that should be made explicit at some point, in a README and/or in tests. That's one of the reasons why I put "Manual test descriptions crystal clear on test conditions" near the end of the [Proposed action plan](https://tidoust.github.io/presentation-api-testcoverage/#by-mid-december-2016). This can wait until the test suite is more advanced.

tomoyukilabs:
@tidoust Thanks for reviewing and merging! By the way, I would like to ask one question:

> The test does not check that the initial availability value is true

I wonder whether the initial availability value is always `true` or not, when any presentation display has been available. I am afraid that there might be possibility a browser might start monitoring when `getAvailability()` would be invoked, fire a `change` event with the value `false` due to timeout for example, then soon find a presentation display and fire a `change` event with the value `true`. WDYT?

tidoust:
@tomoyukilabs Right, the spec does not mandate any discovery mechanism so tests cannot assume much about the parameters and outcomes of the discovery process. In practice, it seems reasonable to expect that `getAvailability` gets resolved with an availability value set to `true` if there is a presentation display available, and to expect that availability value to be stable over time if conditions do not change.

I would keep that approach in tests (as was done in this pull request)... at least until we run in a situation where availability fluctuates over time.

PR #4194, opened 2016-11-10
Upstream 'javascript:' URL navigation tests.

mikewest:
This patch upstreams tests from [1] for 'javscript:' URL navigation's
interaction with CSP's 'script-src' directive. The only changes are
the substitution of a PHP script for a Python script due to differences
in backends.

[1]: https://codereview.chromium.org/2490943002

### DIFF
--- a/content-security-policy/navigation/to-javascript-url.html
+++ b/content-security-policy/navigation/to-javascript-url.html
@@ -1,0 +1,57 @@
+<!DOCTYPE html>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<meta http-equiv="Content-Security-Policy" content="script-src 'nonce-abc'">
+<body>
+<script nonce="abc">
+  function assert_csp_event_for_element(test, element) {
+    document.addEventListener("securitypolicyviolation", test.step_func(e => {
+      if (e.target != element)
+        return;
+      assert_equals(e.blockedURI, "inline");
+      assert_equals(e.effectiveDirective, "script-src");
+      assert_equals(element.contentDocument.body.innerText, "");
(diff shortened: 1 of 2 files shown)


mikewest:
@foolip: Mind looking at this, since you're reviewing the Blink side of things?

@annevk: FYI. This is in support of https://github.com/whatwg/html/pull/2025.

mikewest:
The failures are expected, as Firefox doesn't support the `securitypolicyviolation` event yet, and the relevant Blink patch hasn't yet landed on tip-of-tree.

foolip:
Is there any way to make the tests fail faster? Some `assert_unreached` somewhere?

mikewest:
For Firefox, I can exit early if `typeof SecurityPolicyViolationEvent === 'undefined'`, I suppose.

For Chrome, I'll try to come up with something faster. We're waiting for an event, though, and "Event didn't fire!" is a hard thing to test a priori.

mikewest:
Nope. Neither Chrome nor Firefox trigger a load or error event when preventing `javascript:` navigation.

mikewest:
(For clarity, according to https://html.spec.whatwg.org/multipage/browsers.html#navigating-across-documents:javascript-protocol, we exit early, and treat the execution as resolving to `undefined`. Since `undefined` is not a string, we act like we got a 204 response)

foolip:
OK, I'm out of ideas :)

mikewest:
I had more ideas. Also, I haven't responded to your blink-side comments yet. I thought we'd finish the review there, I'd update this patch, and we'd land it. :)

I guess I'll send you another PR in a bit. ;)

foolip:
Oops, I just assumed this wouldn't change more. Sorry a about that

PR #4196, opened 2016-11-11
Added various features to permit running of tests by others

halindrome:
A tester commented that they need to pass additional parameters to their protocol engine.  Also that their engine limits the scope of targets.  These changes allow for optional specification of those, where the default behavior is as it was earlier.

@BigBlueHat can you verify that this looks okay?  Rob and Hugo have already tested using it.


halindrome:
Note that this PR also addresses the annotation content of a 'page'.  We were checking for @context, but that is not required because it cascades.  Changed that to checking for 'type'.

azaroth42:
:+1: Independently verified that the changes fix the known issues with the tests.  Please merge, so we can get back to the implementers and they can resume checking their implementations :)

halindrome:
Sorry for the nonsense in this branch.  In order to proceed with testing I needed to rebase so that all the other changes were available in this branch (since this is not yet merged) so that CR testing could proceed.

halindrome:
@BigBlueHat and @azaroth42 can you please review again now?  I split out the canonical test as per the request from Europeana.  I have re-run against the RI and CC implementations and updated the test-results at https://w3c.github.io/test-results/annotation-protocol/all.html

PR #4202, opened 2016-11-12
Catch send failures in datachannel-emptystring

nils-ohlmeier:
Without this exceptions when sending the test messages go unnoticed by the test framework.


phoglund:
lgtm

PR #4206, opened 2016-11-14
Clean up ImageBitmapRenderingContext tests

xidachen:
@junov Please take a look. Some of the tests were not written correctly.


junov:
Why is the "return" needed?

junov:
Oh, it is because these use promise_test(). Got it.

PR #4211, opened 2016-11-15
Test default style of <multicol>

zcorpan:
Test for https://github.com/whatwg/html/pull/238

Gecko bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1317665


html/rendering/non-replaced-elements/margin-collapsing-quirks/multicol-quirks-mode.html
```diff
@@ -0,0 +1,6 @@
+<title>multicol default styles (quirks mode)</title>
```
rniwa: Can we add a link to https://html.spec.whatwg.org/#multicol?

zcorpan: Done.

PR #4212, opened 2016-11-15
Use textContent rather than innerHTML for the fetch IDL tests.

Ms2ger:
There is no good reason for the extra complexity of invoking the HTML
serializer here. It might catch bugs in the serializer implementation
(as it does in Servo: <https://github.com/servo/servo/issues/14226>),
but that's out of scope for this test.


jdm:
Thanks!

PR #4215, opened 2016-11-15
Remove gratuitous metadata and self-loop in reftest graph

gsnedders:
Ideally, we'd have a lint for this. Possibly actually walking the graph, as a first step towards doing more complex graph-based lints.


gsnedders:
@Ms2ger @sideshowbarker plz review.

PR #4216, opened 2016-11-16
Fixed expectation relying on :last-child bug in Blink.

ghost:
Running querySelector inside an unclosed element like the snippet below
would never return anything for :last-child in Blink due to
https://crbug.com/662036 :

<body>
  <div></div>
  <script>
    document.querySelector("div:last-child");
  </script>
</body>

The test expectation here was relying on that bug. A fix is about to
land in Blink: https://codereview.chromium.org/2505543004


zcorpan:
Firefox passes the tests with this change.

PR #4217, opened 2016-11-16
elementFromPoint: fix scrolled out of view cases

RByers:
@AutomatedTester @zcorpan This test was failing in a number of browsers because it was calling `elementFromPoint` for co-ordinates that were scrolled off screen.  I added a number of `scrollIntoView` calls to fix it and it's now passing in Chrome.

However the final test case still fails in Firefox and Safari because `getBoundingClientRect` on an `<area>` returns the rect of the image it's used in, and `elementFromPoint` then just gets the `` not the `<area>`.  I don't see this behavior specified anywhere.  OK to just leave a TODO in the test for now?  I see @AutomatedTester added it explicitly about a year ago - did it work in Firefox at some point?

### DIFF
--- a/cssom-view/elementFromPoint.html
+++ b/cssom-view/elementFromPoint.html
@@ -149,14 +149,23 @@
       }, "No viewport available");
 
       test(function () {
-          // TODO: Is this behavior specified anywhere?  This passes in Chrome but not Firefox
-          // or Safari.
+          // HTML says:
+          // Pointing device interaction with an image associated with a set of layered shapes per
+          // the above algorithm must result in the relevant user interaction events being first
+          // fired to the top-most shape covering the point that the pointing device indicated, if
+          // any, or to the image element itself, if there is no shape covering that point.
+          // https://html.spec.whatwg.org/multipage/embedded-content.html#image-map-processing-model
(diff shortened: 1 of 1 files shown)


RByers:
@gsnedders WDYT?

zcorpan:
HTML says

> Pointing device interaction with an image associated with a set of layered shapes per the above algorithm must result in the relevant user interaction events being first fired to the top-most shape covering the point that the pointing device indicated, if any, or to the image element itself, if there is no shape covering that point. User agents may also allow individual area elements representing hyperlinks to be selected and activated (e.g. using a keyboard).

https://html.spec.whatwg.org/multipage/embedded-content.html#image-map-processing-model

It's a bit handwavy but it tries to hook into hit testing, and `elementFromPoint` is defined in terms of hit testing (which is undefined, but, yeah). I would like it to work because it makes testing possible. :-)

zcorpan:
It seems `getBoundingClientRect` on the `rect` has dimensions 0x0 in Chromium, but it doesn't affect the result of the test...

zcorpan (approved):

RByers:
@zcorpan thank you!

PR #4218, opened 2016-11-16
scrollingElement: fix test timeout

RByers:
@Ms2ger when you refactored this test it looks like you omitted a `done` call.  It now passes in Firefox.  I filed [a blink bug](https://bugs.chromium.org/p/chromium/issues/detail?id=665927) to track the remaining failure on Chrome.


gsnedders (approved):

Ms2ger:
So I did. Thanks!

PR #4219, opened 2016-11-16
Test SVGUnitTypes changes (no longer a mixin)

foolip:
Tests for https://github.com/w3c/svgwg/issues/291 and
https://github.com/w3c/fxtf-drafts/issues/90

### DIFF
--- a/svg/historical.html
+++ b/svg/historical.html
@@ -32,7 +32,6 @@ var mixinInterfaces = [
   "SVGFitToViewBox",
   "SVGTests",
   "SVGURIReference",
-  "SVGUnitTypes",
   "SVGZoomAndPan"
 ]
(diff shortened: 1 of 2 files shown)


foolip:
@nikosandronikos, can you review?

nikosandronikos:
@foolip Test looks good. Looks like the Filter Effects drafts still have `SVGFilterElement implements SVGUnitTypes`. Would you be able to submit a PR on Filter Effects Level 1?

foolip:
How do you edit that spec? I filed https://github.com/w3c/fxtf-drafts/issues/90 because it looked complicated.

nikosandronikos:
@foolip That's fair. As long as there's an issue raised.

foolip:
Thanks @nikosandronikos!

PR #4220, opened 2016-11-16
Add test for IndexedDB name scoping.

pwnall:
Tests pass in Firefox Nightly, Chrome Canary, and Safari Tech Preview. The tests do not pass in Safari, and that's a bug on its behalf.

/cc @inexorabletash

### DIFF
--- a/IndexedDB/name-scopes.html
+++ b/IndexedDB/name-scopes.html
@@ -1,0 +1,134 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>
+  IndexedDB: scoping for database / object store / index names, and index keys
+</title>
+<link rel="help" href="https://w3c.github.io/IndexedDB/#constructs">
+<link rel="author" href="pwnall@chromium.org" title="Victor Costan">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="support-promises.js"></script>
+<script>
+'use strict';
+
+// Creates the structure inside a test database.
+//
+// The structure includes two stores with identical indexes and nearly-similar
+// records. The records differ in the "path" attribute values, which are used to
+// verify that IndexedDB returns the correct records when queried.
+//
+// databaseName appears redundant, but we don't want to rely on database.name.
+const buildStores = (database, databaseName, useUniqueKeys) => {
+  for (let storeName of ['x', 'y']) {
+    const store = database.createObjectStore(
+        storeName, { keyPath: 'pKey', autoIncrement: true });
+    for (let indexName of ['x', 'y']) {
+      store.createIndex(
+          indexName, `${indexName}Key`, { unique: useUniqueKeys });
+    }
+
+    for (let xKeyRoot of ['x', 'y']) {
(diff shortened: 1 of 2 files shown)


IndexedDB/name-scopes.html
```diff
@@ -0,0 +1,151 @@
...
+    return database.close();
+  });
+};
+
+// Reads all the store's values using an index.
+//
+// Returns a Promise that resolves with an array of values.
+const readIndex = (testCase, index) => {
+  let resolve = null;
+  const promise = new Promise((_resolve, _reject) => resolve = _resolve);
```
inexorabletash: Why this pattern (capturing `resolve`) rather than the more idomatic style of putting everything in the body of the Promise callback?

IndexedDB/name-scopes.html
```diff
@@ -0,0 +1,151 @@
...
+      promises.push(promise);
+    }
+  }
+
+  return Promise.all(promises);
+}
+
+promise_test(testCase => {
+  return buildDatabases(testCase, false).then(() => {
+    return openNamedDatabase(testCase, 'x', 1);
```
inexorabletash: I think this would more readable with some style changes:

* Don't use `x => { return y; }` when you can use just `x => y`
* Try line breaking before the `.then` 
* Avoid nested `then()`s if you can, i.e. just add more steps to the outer chain, if possible

IndexedDB/name-scopes.html
pwnall: I was afraid I might lose the transaction if I wait until the promise's function is called to make database requests.

pwnall:
@inexorabletash Can you please take another look? Is this closer to what you had in mind?

IndexedDB/name-scopes.html
```diff
@@ -0,0 +1,133 @@
...
+const buildStores = (database, databaseName, useUniqueKeys) => {
+  for (let storeName of ['x', 'y']) {
+    const store = database.createObjectStore(
+        storeName, { keyPath: 'pKey', autoIncrement: true });
+    for (let indexName of ['x', 'y']) {
+      store.createIndex(
+          indexName, `${indexName}Key`, { unique: useUniqueKeys });
+    }
+
+    for (xKeyRoot of ['x', 'y']) {
```
inexorabletash: Add `let` here. And probably `'use strict';` up top?

IndexedDB/name-scopes.html
```diff
@@ -0,0 +1,133 @@
...
+      for (yKeyRoot of ['x', 'y']) {
```
inexorabletash: `let` here too

inexorabletash (approved):
lgtm with two nits

> I was afraid I might lose the transaction if I wait until the promise's function is called to make database requests.

The Promise constructor callback runs synchronously, i.e.

``` js
console.log(1);
new Promise(r => {console.log(2)});
console.log(3)
```

... must log `1, 2, 3`.

pwnall:
Thank you for the education on Promises! I'm mildly embarrassed that I didn't look up that detail, and very grateful!

PR #4221, opened 2016-11-17
Remove a URL fragment from a test expectation of ErrorEvent.filename

nhiroki:
ErrorEvent.filename is an absolute URL of the script URL[1], and an absolute URL string does not contain a fragment string[2], so ErrorEvent.filename shouldn't contain a URL fragment.

[1] https://html.spec.whatwg.org/multipage/webappapis.html#dom-errorevent-filename
[2] https://url.spec.whatwg.org/#syntax-url-absolute

### DIFF
--- a/workers/interfaces/WorkerGlobalScope/onerror/exception-in-onerror.html
+++ b/workers/interfaces/WorkerGlobalScope/onerror/exception-in-onerror.html
@@ -19,7 +19,7 @@ async_test(function() {
   worker.onerror = this.step_func(function(e) {
     assert_true(e instanceof ErrorEvent, 'e instanceof ErrorEvent');
     assert_equals(typeof e.message, 'string', 'typeof e.message');
-    assert_equals(e.filename, document.URL+'#', 'e.filename');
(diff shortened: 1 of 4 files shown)


Ms2ger:
This is changing from Gecko's to Chrome's behaviour?

CC @domenic @annevk @bzbarsky

annevk:
Yeah, I'm not sure the wording in HTML was meant as such. Whenever HTML says "absolute URL" it's something that likely needs updating.

bzbarsky:
What do other UAs do here?

zcorpan:
For http://w3c-test.org/workers/interfaces/WorkerGlobalScope/onerror/exception-in-onerror.html

Edge 13, IE 10-11, Gecko, WebKit pass. Chromium fails (does not include the hash).

zcorpan:
I clarified the spec in https://github.com/whatwg/html/pull/2075

I also made a new PR to test this aspect for `<script src>` in https://github.com/w3c/web-platform-tests/issues/4236

PR #4223, opened 2016-11-17
Indexed DB: Update IDL test logic and WebIDL files

inexorabletash:
Pick up the additions and refinemens made in the "v2"
spec: https://w3c.github.io/IndexedDB/

### DIFF
--- a/IndexedDB/interfaces.html
+++ b/IndexedDB/interfaces.html
@@ -15,7 +15,7 @@ async_test(function(t) {
   var request = new XMLHttpRequest();
   request.open("GET", "interfaces.idl");
   request.send();
-  request.onload = function() {
+  request.onload = t.step_func(function() {
     var idlArray = new IdlArray();
     var idls = request.responseText;
 
@@ -53,6 +53,6 @@ async_test(function(t) {
 
     idlArray.test();
     t.done();
-  };
+  });
 });
 </script>
--- a/IndexedDB/interfaces.html
+++ b/IndexedDB/interfaces.html
@@ -11,28 +11,35 @@
 
 <script>
 "use strict";
-setup(function() {
-  var idlArray = new IdlArray();
-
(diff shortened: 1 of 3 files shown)


inexorabletash:
@foolip - can you review?

Failures in Chrome and Firefox are mostly as expected: new "v2" things that haven't shipped yet, Chrome's "class string" issues w/ prototypes, and other stuff from https://crbug.com/435727

One thing this did point out: the definition of `DOMStringList` I introduced into https://w3c.github.io/IndexedDB/#domstringlist-api is not marked `[Exposed=(Window,Worker)]`. Firefox does expose it, which shows up as a failing assert in the worker test. Chrome does not expose it.

IndexedDB/interfaces.html
```diff
@@ -11,28 +11,35 @@
...
 <script>
 "use strict";
-setup(function() {
-  var idlArray = new IdlArray();
-
+async_test(function(t) {
   var request = new XMLHttpRequest();
   request.open("GET", "interfaces.idl");
   request.send();
   request.onload = function() {
```
foolip: Needs to be wrapped in t.step_func in case of exceptions within, or t.step_func_done to save one line.

IndexedDB/interfaces.worker.js
```diff
@@ -3,35 +3,48 @@
...
-    IDBVersionChangeEvent: ["new IDBVersionChangeEvent('foo')"],
-  });
-  idlArray.test();
-  done();
-};
+async_test(function(t) {
+  var request = new XMLHttpRequest();
+  request.open("GET", "interfaces.idl");
+  request.send();
+  request.onload = function() {
```
foolip: Here too

foolip (approved):
LGTM, one nit, and obviously I just skimmed the IDL itself.

PR #4224, opened 2016-11-17
Update Blob-slice.html

LoonyBean:
Blob.slice() type argument should not be nullable. Added 3 tests when: null, undefined, and no argument is passed.

### DIFF
--- a/FileAPI/blob/Blob-slice.html
+++ b/FileAPI/blob/Blob-slice.html
@@ -45,7 +45,7 @@ test(function() {
   });
 
   test_blob(function() {
-    return blob4 = new Blob().slice(0,0,undefined);
+    return new Blob().slice(0,0,undefined);
   }, {
     expected: "",
     type: "",
@@ -53,7 +53,7 @@ test(function() {
   });
 
   test_blob(function() {
-    return blob5 = new Blob().slice(0,0);
+    return new Blob().slice(0,0);
   }, {
     expected: "",
     type: "",
--- a/FileAPI/blob/Blob-slice.html
+++ b/FileAPI/blob/Blob-slice.html
@@ -39,20 +39,27 @@ test(function() {
   test_blob(function() {
     return new Blob().slice(0,0,null);
   }, {
-    type: "null"
+    expected: "",
+    type: "null",
+    desc: "null type Blob slice"
   });
 
   test_blob(function() {
     return blob4 = new Blob().slice(0,0,undefined);
   }, {
-    type: ""
+    expected: "",
+    type: "",
+    desc: "undefined type Blob slice"
   });
 
   test_blob(function() {
     return blob5 = new Blob().slice(0,0);
   }, {
-    type: ""
+    expected: "",
+    type: "",
+    desc: "no type Blob slice"
   });
+
   var arrayBuffer = new ArrayBuffer(16);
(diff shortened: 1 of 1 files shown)


LoonyBean:
Blob.slice() type argument should not be nullable

FileAPI/blob/Blob-slice.html
```diff
@@ -35,7 +35,24 @@
     type: "content/type",
     desc: "blob2."
   });
+  
+  test_blob(function() {
+    return blob3 = new Blob().slice(0,0,null);
```
foolip: Apparently the reason that blob1 and blob2 are assigned above is that they're used in later tests. Since these aren't, you can just `return new Blob()...`

FileAPI/blob/Blob-slice.html
```diff
@@ -35,7 +35,24 @@
...
+  }, {
+    type: "null"
```
foolip: Is `desc` needed here, or where does the test description come from? If you haven't done so, make sure to set up the test environment and run ./serve to test with the changes in https://codereview.chromium.org/2509833002/ so that they pass for certain.

FileAPI/blob/Blob-slice.html
```diff
@@ -35,7 +35,24 @@
     type: "content/type",
     desc: "blob2."
   });
+
```
sideshowbarker: trailing whitespace on this line needs to be removed https://travis-ci.org/w3c/web-platform-tests/jobs/176824939

FileAPI/blob/Blob-slice.html
```diff
@@ -35,7 +35,24 @@
     type: "content/type",
     desc: "blob2."
   });
+  
+  test_blob(function() {
+    return blob3 = new Blob().slice(0,0,null);
+  }, {
+    type: "null"
+  });
+
```
sideshowbarker: trailing whitespace on this line needs to be removed https://travis-ci.org/w3c/web-platform-tests/jobs/176824939

foolip:
Ping @LoonyBean :)

foolip:
w3c-test:mirror

FileAPI/blob/Blob-slice.html
```diff
@@ -36,6 +36,23 @@
...
   });
 
+  test_blob(function() {
+    return new Blob().slice(0,0,null);
+  }, {
+    type: "null"
+  });
+
+  test_blob(function() {
+    return blob4 = new Blob().slice(0,0,undefined);
```
foolip: blob4/5 not used, can remove.

LoonyBean:
Made changes to code review. Thanks

FileAPI/blob/Blob-slice.html
```diff
@@ -36,6 +36,30 @@
...
+  test_blob(function() {
+    return new Blob().slice(0,0,null);
+  }, {
+    expected: "",
+    type: "null",
+    desc: "null type Blob slice"
+  });
+
+  test_blob(function() {
+    return blob4 = new Blob().slice(0,0,undefined);
```
foolip: Can drop blob4 and blob5, since they're not references later.

foolip (approved):
I can confirm that "null type Blob slice" fails on old Chrome and now passes, as expected. All LGTM except the blob4/5 thing.

PR #4226, opened 2016-11-18
Add referrer policy test case for srcdoc iframes with own policy

estark37:
This commit modifies the img-tag referrer policy tests to test a
srcdoc iframe with its own referrer policy set via meta tag, adding
on to the existing tests for srcdoc iframes that inherit the referrer
policy of their parent document.

The corresponding HTML change is
https://github.com/whatwg/html/pull/1871.

This also corrects a bug in
https://github.com/w3c/web-platform-tests/pull/3700, which was
comparing undefined to undefined (using the 'referrer' property of
image data, when it should be headers.referer).

### DIFF
--- a/referrer-policy/generic/common.js
+++ b/referrer-policy/generic/common.js
@@ -118,36 +118,45 @@ function queryImage(url, callback, attributes, referrerPolicy) {
   // - images in a top-level document,
   // - and images in a `srcdoc` frame with its own referrer policy to
   //   override its parent.
-  var iframeWithoutOwnPolicy = document.createElement('iframe');
-  iframeWithoutOwnPolicy.srcdoc = "Hello, world.";
 
-  iframeWithoutOwnPolicy.onload = function () {
+  var noSrcDocPolicy = new Promise((resolve, reject) => {
+    var iframeWithoutOwnPolicy = document.createElement('iframe');
+    iframeWithoutOwnPolicy.srcdoc = "Hello, world.";
+    iframeWithoutOwnPolicy.onload = function () {
+      var nextUrl = url + "&cache_destroyer2=" + (new Date()).getTime();
+      loadImageInWindow(nextUrl, function (img) {
+        resolve(decodeImageData(extractImageData(img)));
+      }, attributes, iframeWithoutOwnPolicy.contentWindow);
+    };
+    document.body.appendChild(iframeWithoutOwnPolicy);
+  });
+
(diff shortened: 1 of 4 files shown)


estark37:
@mikewest can you please review?

referrer-policy/generic/common.js
```diff
@@ -111,23 +111,43 @@ function queryIframe(url, callback, referrer_policy) {
...
+      var iframePolicy = (referrerPolicy === "no-referrer") ? "unsafe-url" : "no-referrer";
+      iframeWithOwnPolicy.srcdoc = "<meta name='referrer' content='" + iframePolicy + "'>Hello world.";
+
+      iframeWithOwnPolicy.onload = function () {
+        var nextUrl = url + "&cache_destroyer2=" + (new Date()).getTime();
+        loadImageInWindow(nextUrl, function (img) {
+          assert_equals((iframePolicy === "no-referrer" ? undefined : document.location.href), decodeImageData(extractImageData(img)).headers.referer, "Referrer inside 'srcdoc' should use the iframe's policy if it has one");
+
+          nextUrl = url + "&cache_destroyer3=" + (new Date()).getTime();
+          loadImageInWindow(nextUrl, function (img) {
```
estark37: I don't love the extra indentation here, but such is life in javascript, I guess...

mikewest: Indeed, this is pretty tough to follow. Especially since we've pushed the frame in which we're working all the way to the end of the function.

Since none of the frames depends on the other, could you perhaps rearrange things like:

```
var iframeWithoutOwnPolicy = ...;
iframeWithoutOwnPolicy.onload = function () {
  loadImageInWindow({ .... }, iframeWithoutOwnPolicy.contentWindow);
};
var iframeWithOwnPolicy = ...;
iframeWithOwnPolicy.onload = function () {
  loadImageInWindow({ .... }, iframeWithOwnPolicy.contentWindow);
};
```

And gate the call to `callback()` on both completing? You could set a bool for both, and end both with `maybeCallCallback()` which would be something like `if (firstFrameIsDone && secondFrameIsDone) { callback(wrapResult(...), url)`?

mikewest: Or turn everything into promises and do `Promise.all([ ... ]).then(_ => { callbackStuff() });`?

estark37: Oh, yeah, promises. That's much nicer, thanks! Done.

mikewest (approved):
Mich nicer! Thanks for the rewrite. :)

PR #4229, opened 2016-11-21
[html] HTMLFieldSetElement.prototype.elements should return HTMLColle…

tkent-google:
…ction.

Follow a specification change [1]. The test passes with Firefox, and fails with
Google Chrome 56.

Note that |children_outer instanceof HTMLCollection| doesn't work because
HTMLFormControlsCollections inherits from HTMLCollection.

[1] https://github.com/whatwg/html/commit/934887d313305448b7ba6dd3af6f7434e1b600c8


TakayoshiKochi:
LGTM

PR #4232, opened 2016-11-21
Annotation model fixreqs

halindrome:
These changes were already reviewed by the Web Annotation working group.

@tcole3 and @bigbluehat please review so we can merge.


BigBlueHat:
@halindrome the fixreq fixes look fine. 👍 

Should @tcole3's changes be added as a separate PR?

Given that we're considering [bringing back Composite, Independents, and List](https://github.com/w3c/web-annotation/issues/387), I want to be sure we can simply revert that commit (or commits) to get the tests back.

Thoughts?

halindrome:
I am not too worried about losing those tests - they are in the web-annotation-test repo.

tcole3:
I've finished double-checking that the changes in this PR, including the ones made to align with CR3 version of the model, work as expected. 

Regarding these last changes, I favor including in this PR. The deprecated keys (Composite, List, and Independents) are no longer in our context document and no longer in CR3 version of model.  Tests can be retrieved if needed by developers. Presence of keys not an error, just won't count towards meeting CR exit criteria.

PR #4233, opened 2016-11-21
Cache Storage: Make test setup resilient against buggy implementations

inexorabletash:
One test asserts that the cache name "unpaired\uD800" does not get mapped to "unpaired\uFFFD", but relies on a cache by that name not existing. If an implementation is buggy it will implicitly create the cache, causing the test to fail the setup checks on subsequent runs, as opposed to hitting the actual test assertions.

Change the test to delete the caches so that it starts with a clean slate regardless of whether or it passes the test.

### DIFF
--- a/service-workers/cache-storage/script-tests/cache-storage.js
+++ b/service-workers/cache-storage/script-tests/cache-storage.js
@@ -169,20 +169,28 @@ promise_test(function(t) {
   }, 'CacheStorage.delete with nonexistent cache');
 
 promise_test(function(t) {
-    var bad_name = 'unpaired\uD800';
-    var converted_name = 'unpaired\uFFFD'; // Don't create cache with this name.
-    return self.caches.has(converted_name)
+    var unpaired_name = 'unpaired\uD800';
+    var converted_name = 'unpaired\uFFFD';
+
+    // The test assumes that a cache with converted_name does not
+    // exist, but if the implementation fails the test then such
+    // a cache will be created. Start off in a fresh state by
+    // deleting all caches.
+    return delete_all_caches()
+      .then(function() {
+          return self.caches.has(converted_name);
+      })
       .then(function(cache_exists) {
           assert_false(cache_exists,
                        'Test setup failure: cache should not exist');
       })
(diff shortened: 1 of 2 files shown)


inexorabletash:
@wanderview - can you take a peek?

service-workers/cache-storage/script-tests/cache-storage.js
```diff
@@ -170,8 +170,15 @@ promise_test(function(t) {
 
 promise_test(function(t) {
     var bad_name = 'unpaired\uD800';
-    var converted_name = 'unpaired\uFFFD'; // Don't create cache with this name.
-    return self.caches.has(converted_name)
+    var converted_name = 'unpaired\uFFFD';
+
+    return Promise.all([
+          self.caches.delete(bad_name),
+          self.caches.delete(converted_name)
```
wanderview: I'm confused.  Why is this necessary?

inexorabletash: Quoting from above:

> One test asserts that the cache name "unpaired\uD800" does not get mapped to "unpaired\uFFFD", but relies on a cache by that name not existing. If an implementation is buggy it will implicitly create the cache, causing the test to fail the setup checks on subsequent runs, as opposed to hitting the actual test assertions.
>
> Change the test to delete the caches so that it starts with a clean slate regardless of whether or it passes the test.

wanderview: Can you just add a clarifying comment to the test?  Looking at it on its own suggests that it should be deleting something, but its really there as a failsafe.  It might be a bit clearer if you just iterated cache.keys() and deleted all cache objects to get a blank slate.

wanderview (approved):

inexorabletash:
Dang it, sorry, rebased. :P Let's see if I can make this not suck....

inexorabletash:
@wanderview - can you take another quick look?

(I also renamed the string variable per review of Blink's copy of the test)

wanderview (approved):

PR #4235, opened 2016-11-21
SW Cache Storage: use promise_rejects helper

inexorabletash:
Convert tests to use the standard testharness.js promise_rejects() helper and remove the legacy assert_promise_rejects() helper and the now unused resources/testharness-helpers.js file.

Also renames credentials test with .https.html suffix since this is a [SecureContext] feature.


inexorabletash:
@wanderview - can you review?

wanderview (approved):
Thanks!

PR #4236, opened 2016-11-22
Test window.onerror when <script src> URL contains a hash

zcorpan:
See https://github.com/whatwg/html/pull/2075

### DIFF
--- a/html/webappapis/scripting/processing-model-2/compile-error-same-origin-with-hash.html
+++ b/html/webappapis/scripting/processing-model-2/compile-error-same-origin-with-hash.html
@@ -1,0 +1,36 @@
+<!doctype html>
+<html>
+ <head>
+  <title>window.onerror - compile error in &lt;script src=...> with hash</title>
+  <script src="/resources/testharness.js"></script>
+  <script src="/resources/testharnessreport.js"></script>
+ </head>
+ <body>
+  <div id=log></div>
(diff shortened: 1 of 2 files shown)


zcorpan:
@nhiroki r? 🙂

html/webappapis/scripting/processing-model-2/compile-error-same-origin-with-hash.html
```diff
@@ -0,0 +1,36 @@
+<!doctype html>
+<html>
+ <head>
+  <title>window.onerror - compile error in &lt;script src=...> with hash</title>
```
nhiroki: Can you replace ```src=...>``` with ```src=...&gt;```?

nhiroki:
LGTM. Thank you :)

html/webappapis/scripting/processing-model-2/compile-error-same-origin-with-hash.html
zcorpan: Why? > does not need escaping in HTML.

nhiroki: Sorry for my delayed reply. I thought it's more consistent with "&lt;", but as you commented, it's not necessary. I'm ok with leaving it as is.

PR #4246, opened 2016-11-24
Test for boundary events while capturing changes

### LINKED ISSUE #145: Need to clarify what triggers boundary events when releasing a capture
When a pointer capture is released from an element A while the pointer is over a different element B, we expect <code>pointerout</code>/<code>pointerleave</code> events for <code>A</code> and <code>pointerover</code>/<code>pointerenter</code> events for B before sending out future events (say a <code>pointermove</code>) to the hit-test node B. The current spec wording covers this through the Note in [Section 5.1.3](https://w3c.github.io/pointerevents/#firing-events-using-the-pointerevent-interface).

The wording needs to be a bit more precise, see [the discussion in a previous PR](https://github.com/w3c/pointerevents/pull/138#r77719958).

NavidZ:
@mustaqahmed
closes w3c/pointerevents#145


NavidZ:
Also closes w3c/pointerevents#156

pointerevents/pointerevent_support.js
```diff
@@ -240,5 +241,9 @@ MultiPointerTypeTest.prototype.createNextTest = function() {
 
 
 function setup_pointerevent_test(testName, supportedPointerTypes) {
-   return globalPointerEventTest = new MultiPointerTypeTest(testName, supportedPointerTypes);
+    return globalPointerEventTest = new MultiPointerTypeTest(testName, supportedPointerTypes);
+}
+
+function checkPointerEventType(event) {
+    assert_equals(event.pointerType, expectedPointerType, "pointerType should be the same as the requested device.");
```
mustaqahmed: Nit: "pointerType should be the one specified in the test page"

pointerevents/pointerevent_boundary_events_in_capturing-manual.html
```diff
@@ -0,0 +1,97 @@
...
+                                        'lostpointercapture and pointerout/leave should be dispatched to the capturing target and pointerover/enter should be dispatched to the hit-test element before the first pointermove event after releasing pointer capture');
+                                }, expectedPointerType + " pointer events boundary events when releasing capture"); 
+                                phase = PhaseEnum.WaitingForUp;
+                            }
+                        }
+                    });
+                    on_event(listener, eventName, function (event) {
+                        if (phase == PhaseEnum.WaitingForDown)
+                          return;
+                        eventsRecieved.push(event.type + '@listener');
```
mustaqahmed: Please add an assert that this line is reached only in phase==WaitingForFirstMove. The listener shouldn't receive any event after that phase.

pointerevents/pointerevent_boundary_events_in_capturing-manual.html
```diff
@@ -0,0 +1,97 @@
...
+                        if (eventName == 'pointermove') {
+                          test(function () {
+                              checkPointerEventType(event);
+                              assert_array_equals(eventsRecieved, ['pointerout@target', 'pointerleave@target', 'pointerover@listener', 'pointerenter@listener', 'gotpointercapture@listener', 'pointermove@listener'],
```
mustaqahmed: Isn't it clearer to check the eventsReceived array only once, right before done()?

pointerevents/pointerevent_boundary_events_in_capturing-manual.html
NavidZ: But it does. When the captured is released this line adds a few more events to the array which will be at the beginning of the array above (i.e. 'lostpointercapture@listener', 'pointerout@listener', 'pointerleave@listener').

pointerevents/pointerevent_boundary_events_in_capturing-manual.html
NavidZ: The only reason was that I didn't want to add the logic to suppress the next move events which is one solution. The other solution would be not to add any pointermove at all in the array so that it doesn't make the array equal logic any more complicated.

pointerevents/pointerevent_support.js
NavidZ: done

pointerevents/pointerevent_boundary_events_in_capturing-manual.html
mustaqahmed: You are right about boundary & got/lost events but it looks like a crack that WaitingForFirstMove phase is never checked anywhere. The transition pattern
<code>if (oldPhase) {...; phase=newPhase;}</code>
should be there for every phase.

Change the following <code>if</code> to include phase==WaitingForFirstMove then?

NavidZ: It does make sense to have that condition in the if. I added it.

mustaqahmed (approved):

mustaqahmed:
LGTM

PR #4247, opened 2016-11-25
Fetch: align multipart/form-data with HTML

annevk:
See https://github.com/whatwg/fetch/issues/422 for rationale.


annevk:
The reason Travis is failing is because Firefox cannot be downloaded. I don't think that should block landing a change necessarily. @jgraham @Ms2ger?

annevk:
(Not that I'd want to land this without review, but I don't want reviewers to be deterred either by a failing Travis.)

annevk:
I wish you would not have merged this until one of the owners reviewed it. It turns out @youennf has concerns.

PR #4248, opened 2016-11-25
Add multipart/form-data part header field and parameter tests

eehakkin:
The TODOs should probably be resolved before merging this pull request.

For details, please see:
 * https://github.com/whatwg/fetch/issues/392
 * https://github.com/whatwg/fetch/pull/424

### DIFF
--- a/fetch/api/response/response-consume.html
+++ b/fetch/api/response/response-consume.html
@@ -12,6 +12,25 @@
   </head>
   <body>
     <script>
+    function blobToFormDataResponse(name, blob) {
+      var formData = new FormData();
+      formData.append(name, blob);
+      return new Response(formData);
+    }
+
+    function readBlobAsArrayBuffer(blob) {
+      return new Promise(function(resolve, reject) {
+        var reader = new FileReader();
+        reader.onload = function(evt) {
+          resolve(reader.result);
+        };
+        reader.onerror = function(evt) {
+          reject("Blob's reader failed");
+        };
+        reader.readAsArrayBuffer(blob);
+      });
+    }
+
     function responsePromise(body, responseInit) {
       return new Promise(function(resolve, reject) {
         resolve(new Response(body, responseInit));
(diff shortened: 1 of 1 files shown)


youennf:
@eehakkin, are you planning to add commits to fix the TODOs you mentionned, or do you think this PR can be reviewed and/or merged as is?

eehakkin:
@youennf, I fixed the TODOs.

fetch/api/response/response-consume.html
```diff
@@ -205,6 +224,72 @@
...
+        // The name of a File must be taken from the filename parameter in
+        // the Content-Disposition header field.
+        assert_equals(bodyAsFormData.get("file").name, "file1", "Retrieve and verify response's 5th entry name property");
+        assert_equals(bodyAsFormData.get("file-without-type").name, "file2", "Retrieve and verify response's 6th entry name property");
+        // The type of a File must be taken from the Content-Type header field
+        // which defaults to "text/plain".
+        assert_equals(bodyAsFormData.get("file").type, "application/octet-stream; x-param=x-value", "Retrieve and verify response's 5th entry type property");
+        assert_equals(bodyAsFormData.get("file-without-type").type, "text/plain", "Retrieve and verify response's 6th entry type property");
+
+        return Promise.all([
```
youennf: I wonder whether we should move away from Promise.all here and do the tests one after the other instead.
There might be a chance that in case asserts fail in both sub-tests, the test becomes flaky as one time it hits the first assertion and next the second one.

youennf:
The tests look good to me.
If you are planning to add more form data tests, I wonder whether we should not move those tests to a dedicated file.

Also, would it be good to run those tests in workers?
If so, you might just need to add a response-consume-formdata.any.js file with your current code.
See other *.any.js files in this repo as examples.

fetch/api/response/response-consume.html
eehakkin: Yes, Promise.all may lead to flaky results. Fixed.

PR #4253, opened 2016-11-28
Test document.fullscreenElement.requestFullscreen()

foolip:


foolip:
@upsuper, PTAL. Chrome Edge and Firefox pass, but Chrome follows an older version of the spec where the check is synchronous in `requestFullscreen()` and I wanted something to guard against regressions when refactoring.

foolip:
@upsuper, I noticed a mistake in the test and pushed an update, but it still passes in Firefox. Can you review?

foolip:
Ping @upsuper @aliams @jernoble for review.

foolip:
@mounirlamouri, would you mind reviewing this random Fullscreen test?

upsuper (approved):

mounirlamouri (approved):

foolip:
Thanks for the reviews!

PR #4256, opened 2016-11-29
Fix three Presentation API tests.

markafoltz:
- PresentationConnection_onconnected-manual.html: The test was closing the connection before the `connect` event fired.
- PresentationConnection_onterminated-manual.html: Typos
- presentation-api/controlling-ua/support/presentation.html: Bugs

PTAL @tomoyukilabs

### DIFF
--- a/presentation-api/controlling-ua/PresentationConnection_onconnected-manual.html
+++ b/presentation-api/controlling-ua/PresentationConnection_onconnected-manual.html
@@ -29,7 +29,6 @@
           connection.onconnect = t.step_func_done(function() {
             assert_equals(connection.state, "connected");
           });
-          connection.close();
       })
       .catch(function(ex) {
         assert_unreached(ex.name + ":" + ex.message);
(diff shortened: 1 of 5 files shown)


mfoltzgoogle:
CC @zhaobin2016

tomoyukilabs:
I have a couple of comments below, about PresentationConnection_onconnected-manual.html and PresentationConnection_onterminated-manual.html:

* If I understand this correctly, it fixes the items 2. and 3. mentioned in https://github.com/w3c/web-platform-tests/issues/4040#issue-184229288. I believe that we can also fix 1. by adding `step_timeout()` like below:
```js
      step_timeout({
        t.force_timeout();
        t.done();
      });
      request.start()
        .then(function(connection) { // ...
```
* These files should be renamed to PresentationConnection_onconnect-manual.html and PresentationConnection_onterminate-manual.html.

tidoust:
On top of comments raised by @tomoyukilabs, I think the call to `terminate` should be moved to an `add_cleanup` call, to make sure it runs even if the assertion fails. I raised issue #4258 to track this across tests, because that affects other tests as well.

@tomoyukilabs I don't think the call to `step_timeout` can be done *before* the call to `start()`: timeout should only start once the user is done with the dialog box, meaning when the `Promise` is resolved.

Note on-going PR #4222 from @obstschale to fix related issue #4040 (no conflict but it seems good to make sure people are aware of related efforts).

@mfoltzgoogle let us know if you'd rather merge this and let others work on remaining comments (timeout, cleanup step, file renaming, considering move to `promise_test`) in another PR.

mfoltzgoogle:
Yes, I think it would make sense to merge this and handle the other changes in other PRs.

PR #4260, opened 2016-11-29
OffscreenCanvas conformance tests

xidachen:
The spec for OffscreenCanvas has officially landed here:
https://html.spec.whatwg.org/multipage/scripting.html#the-offscreencanvas-interface

This pull request adds conformance tests for OffscreenCanvas and its 2D context.

Notice that in the spec for OffscreenCanvasRenderingContext2D, it looks like this:
OffscreenCanvasRenderingContext2D implements CanvasState;
which means that the tests for OffscreenCanvasRenderingContext2D can follow the tests for CanvasRenderingContext2D. Luckily we have a loooot of tests under 2dcontext/, and most of them are automatically generated by using the scripts in 2dcontext/tools/gentest.py

So this pull requests add the following scripts:
1. offscreen-canvas/tools/gentest.py: this is the main auto-generation script that is modified based on 2dcontext/tools/gentest.py
2. offscreen-canvas/tools/name2dir.yaml: a mapping for each test to its directory.
3. offscreen-canvas/tools/templates.yaml: this is also modified based on 2dcontext/tools/templates.yaml, notice that applying the templates here also generates tests in a worker, which is the main difference compared to the one under 2dcontext.
4. offscreen-canvas/tools/tests2d.yaml: modified based on 2dcontext/tools/tests2d.yaml and 2dcontext/tools/tests.yaml

When running the script gentest.py, there will be 1458 tests (729 in main and 729 in a worker) generated. For now, I have not included these automatically generated tests because it would be really distracting. Once the scripts are approved, I will include all the tests.

@junov @RByers @foolip @mephisto41 Please review.


offscreen-canvas/tools/templates.yaml
```diff
@@ -0,0 +1,58 @@
...
+    <h1>%(name)s</h1>
+    <p class="desc">%(desc)s</p>
+
+    %(notes)s
+    <script>
+    fetch_tests_from_worker(new Worker("%(name)s.js"));
+    </script>
+
+w3cworker: |
+    importScripts("/resources/testharness.js");
```
junov: The worker script should also have a comment stating that the file was generated and should not be edited directly.

offscreen-canvas/tools/gentest.py
```diff
@@ -0,0 +1,757 @@
+# This script is modified based on 2dcontext/tools/gentest.py
```
junov: Could we avoid so much code duplication? I think you should factor-out all the shared stuff into a module that is imported by both generators.  That will make these tests much easier to maintain in the future, and it will prevent them from forking apart.

junov (changes requested):

xidachen:
@junov I have updated the auto generation script. Basically I made the 2dcontext/tools/gentest.py a util function, and let offscreen-canvas/tools/gentest.py call that util function.

Please review.

junov (approved):

junov:
This looks great to me.  Push the generated files at your convenience.

xidachen:
@junov Thanks. All the generated tests are included in the new commit (> 2000 files).

jgraham:
w3c-test:mirror

junov:
We will push the 729 new tests progressively to avoid timeouts in the stability tests.  Now this PR only has changes to the generator scripts.

RByers:
My advice to Xida was not to let limitations of the stability checks to get in his way too much (as this is an unusual case of adding MANY tests).  If he can manually validate that the tests do indeed pass (or fail) consistently in both Firefox and Chrome when run at least 10 times in a row, then that should be good enough to land them.

PR #4261, opened 2016-11-29
<template>.content.ownerDocument properties

annevk:
A test to make sure we end up fixing
https://github.com/whatwg/html/issues/2109 correctly.

### DIFF
--- a/html/semantics/scripting-1/the-template-element/template-element/template-content-node-document-document-properties.sub.html
+++ b/html/semantics/scripting-1/the-template-element/template-element/template-content-node-document-document-properties.sub.html
@@ -32,7 +32,8 @@
     assert_equals(t.content.ownerDocument.compatMode, "CSS1Compat")
   }, "document.compatMode")
 
-  test(() => {
+  test((test) => {
+    test.add_cleanup(() => { document.cookie = "" })
     var t = document.createElement("template")
     assert_throws("SecurityError", () => { t.content.ownerDocument.cookie = "test" })
     document.cookie = "test2"
--- a/html/semantics/scripting-1/the-template-element/template-element/template-content-node-document-document-properties.sub.html
+++ b/html/semantics/scripting-1/the-template-element/template-element/template-content-node-document-document-properties.sub.html
@@ -1,0 +1,41 @@
+<meta charset=windows-1252>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div id="log"></div>
+<script>
+  test(() => {
+    assert_equals(document.origin, "{{location[scheme]}}://{{location[host]}}")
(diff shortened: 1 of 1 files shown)


html/semantics/scripting-1/the-template-element/template-element/template-content-node-document-document-properties.sub.html
```diff
@@ -0,0 +1,41 @@
...
+<div id="log"></div>
+<script>
+  test(() => {
+    assert_equals(document.origin, "{{location[scheme]}}://{{location[host]}}")
+    var t = document.createElement("template")
+    assert_equals(t.content.ownerDocument.origin, "{{location[scheme]}}://{{location[host]}}")
+  }, "document.origin")
+
+  test(() => {
+    assert_equals(document.domain, "www.web-platform.test")
```
zcorpan: This might need to be a bit more involved to work in existing setups I believe...

html/semantics/scripting-1/the-template-element/template-element/template-content-node-document-document-properties.sub.html
```diff
@@ -0,0 +1,41 @@
...
+    assert_equals(document.compatMode, "BackCompat")
+    assert_equals(t.content.ownerDocument.compatMode, "CSS1Compat")
+  }, "document.compatMode")
+
+  test(() => {
+    var t = document.createElement("template")
+    assert_throws("SecurityError", () => { t.content.ownerDocument.cookie = "test" })
+    document.cookie = "test2"
+    assert_equals(t.content.ownerDocument.cookie, "")
+  }, "document.cookie")
```
zcorpan: Maybe use `add_cleanup(() => { ... })` to remove the cookies again.

annevk: Where do I add that line?

html/semantics/scripting-1/the-template-element/template-element/template-content-node-document-document-properties.sub.html
annevk: Yeah, do you have advice? Having said that, this might just need to throw based on what Chrome/Safari appear to be doing.

zcorpan: Create an iframe that loads a file with some subdomain using `{{domain[www]}}`, and then chop off the subdomain with e.g. `substring`+`indexOf` when setting `document.domain`. 

I *think* this would work for w3c-test.org and for browsers that use wptserve to run the tests. It still wouldn't work where wptserve is not used, but not much to do about that here.

But if setting should just throw then the test could instead just do `document.domain = document.domain` and assert that it throws, without iframes.

PR #4263, opened 2016-11-30
Update webmessaging tests to match current HTML spec.

mkruisselbrink:
This removes two tests that relied on data: iframes to be same origin with
their opening window (HTML specifies documents with data: URL to always
have an opaque origin), and fixes the test expectations for passing null
to an optional sequence<> argument.


Ms2ger:
`postMessage` is supposed to work on cross-origin windows, right? Can we rewrite the test in a better way instead? Perhaps the child frame can `postMessage` its results back to the parent.

mkruisselbrink:
> postMessage is supposed to work on cross-origin windows, right? Can we rewrite the test in a better way instead? Perhaps the child frame can postMessage its results back to the parent.

Maybe that makes sense, yes... I originally tried to do something like that, but it got pretty annoying to write, and it wasn't clear to me what I'd still end up testing that isn't already being tested by other tests. I'll give it another go and see what I come up with (although not sure if it will still be testing the same thing this test was originally trying to test).

mkruisselbrink:
Okay, changed the data: tests to something that works when data: iframes are correctly treated as opaque origin (which makes them fail in firefox, since firefox still considers those iframes same origin).

Ms2ger:
Thanks!

PR #4265, opened 2016-11-30
OffscreenCanvas 2d context conformance tests - part 2

xidachen:
Previously in here:
https://github.com/w3c/web-platform-tests/commit/72f095bf6fc9c181162212ced32ed9423eb81578
we updated the scripts that automatically generates the tests for OffscreenCanvas. We have not yet included the tests because travis system times out.

We now found that the reason these test timed out is that we do
importScripts("/resources/testharnessreport.js"), which has window.xxx in it. Once we remove this line, tests should not time out.

This pull request will have 2 commits. The first commit fix the templates.yaml. The second commit will include all the generated tests.


xidachen:
w3c-test:mirror

sideshowbarker:
w3c-test:mirror

sideshowbarker:
@xidachen the “w3c-test:mirror” thing only works if the comment is posted by somebody who has write perms to the repo (it’s a security restriction to prevent arbitrary users from pushing untrusted/unreviewed python code to w3c-test.org)

xidachen:
@sideshowbarker Thank you so much for the help.

Question, when I use https://w3c-test.org/submissions/4265/ to run the tests that I added, I see every test failed with this message: OffscreenCanvas is not defined. But the result is "FAIL", not "ERROR", any idea why Travis would give "ERROR"?

sideshowbarker:
> any idea why Travis would give "ERROR"?

No—but I see you’ve also asked on the #testing channel and that is probably the better place to try to troubleshoot it

xidachen:
@sideshowbarker Thank you.
@jgraham is investigating now.

xidachen:
@junov In the first commit, I fix the test generator. Once I see all tests passed, I will have another commit with all generated tests.

xidachen:
@junov In the new commit, I included two folders of tests where folders' name starts with c. Travis check shows that firefox failed all those worker tests, I think that's fine. But somehow chrome gives:
u'log' (u'debug', {'message': 'Hanging up on Selenium session'})

I think it should be fine to merge, because locally I have ran those tests without experimental flag many times, and they fail consistently and very quickly.

junov:
I don't feel comfortable merging until we at least understand the reason why the check is failing.

xidachen:
@junov I did an experiment. If I include another folder start with 'f' which adds 420 tests, Travis times out. So let's keep this pull request as it is. I will follow up with more pull requests.

PR #402, opened 2013-11-09
Initial test for resource timing

plehegar:
This is just an initial test to adjust all of the tests before moving them all.

PR #4266, opened 2016-11-30
review some initial tests for resource timing

siusin:
review tests from pull request #402 

* RT Level 1 indicated that aborted requests or requests that don't return a response may be included as PerformanceResourceTiming objects in the Performance Timeline of the relevant context, but later versions would like to included these aborted requests in the Perf Timeline.


resource-timing/resource_connection_reuse.html
```diff
@@ -0,0 +1,55 @@
...
+    if (window.performance.getEntriesByType === undefined) {
+      done();
+      return;
+    }
+    var context = new PerformanceContext(iframe.contentWindow.performance);
+    var entries = context.getEntriesByType('resource');
+
+    test_equals(entries.length, 2, 'There should be 2 PerformanceEntries');
+
+    var entry = entries[1];
```
igrigorik: Might be worth adding a comment above, explaining that we're counting on second request of same resource to reuse the socket..

igrigorik:
> RT Level 1 indicated that aborted requests or requests that don't return a response may be included as PerformanceResourceTiming objects in the Performance Timeline of the relevant context, but later versions would like to included these aborted requests in the Perf Timeline.

Yikes, looks like we introduced some confusion here. 

- RT L1 ([section 4.2](https://www.w3.org/TR/resource-timing/#resources-included)) states: _"Aborted requests or requests that don't return a response may be included as PerformanceResourceTiming objects in the Performance Timeline of the relevant context."_
- RT L2 changes via https://github.com/w3c/resource-timing/issues/12:
  1. https://github.com/w3c/resource-timing/pull/19/commits/0eb0f6997fc3f8a70a556212b45fa9ce5cfe7631 landed an update stating that aborted requests **MUST** be included, but it didn't update the processing section..
  1. After that was landed we re-opened discussion once more, concluded that it is (in fact) the correct behavior, and Wes started working on updating the processing section.
  1. At TPAC we concluded that we'll push Wes's update to L3, but we never backed out (i). 

The end result, as of right now, is that L2 states that failed resources MUST be surfaced, but doesn't have the processing section updated. 

---

1. I think "MAY" for L1 is correct, and we shouldn't change that -- some browsers surface such requests, and that's good. 
1. We need to revisit how we want to treat this in L2: 
   1. We can carry over MAY from L1, back out https://github.com/w3c/resource-timing/pull/19/commits/0eb0f6997fc3f8a70a556212b45fa9ce5cfe7631, and target that + processing update to L3. 
   2. We can revisit our earlier decision to push MUST to L3...

I believe the reason we went with (i) at TPAC is that it would make existing L2 implementations (e.g. IE, which is not going to change?) incompatible... Which hints that we should stick with that.

@toddreifsteck @plehegar wdyt?

plehegar:
For L2, I'd rather use (i) since doing (ii) would probably require hooking ourselves into Fetch, which is for L3.

toddreifsteck:
I agree that we should go with (i) for L2 and push any further specification to L3.

siusin:
Shall we drop the test case for L1 (resource_ignore_failures_level1.html) in that case? I'm not sure if a test case for a "MAY" will make sense, knowing that it produces different test results among the major browsers.

resource-timing/resource_dynamic_insertion.html
```diff
@@ -0,0 +1,64 @@
...
+    expected_entries[ pathname + 'resources/inject_resource_test.html' ] = 'iframe',
+
+    test_resource_entries(entries, expected_entries);
+}
+window.setup_iframe = setup_iframe;
+</script>
+</head>
+<body>
+<h1>Description</h1>
+<p>This test validates that all of the initiator types are represented even when dynamically inserted.</p>
```
siusin: done!

resource-timing/resource_connection_reuse.html
siusin: Yeah. Please look at commit #b1bbb10

plehegar:
@siusin imho, we should drop it. Until we better define what it means in L3, it might be harmful to have a test related to resource failures in the test suite.

igrigorik:
+1 to removing those tests. Also, opened https://github.com/w3c/resource-timing/pull/85 for spec update, ptal.

siusin:
Ok, thanks!

plehegar:
LGTM

PR #4268, opened 2016-11-30
Test more Document object clone aspects

annevk:
See https://github.com/whatwg/dom/issues/378 for context.


zcorpan:
For the test "implementation.createDocument"...

Gecko fails because `doc.origin` is not implemented.

WebKit fails because `doc.origin` is "http_web-platform.test_8000" instead of "null".

Chromium fails because `doc.URL` is "" instead of "about:blank", and `doc.origin` is "http://web-platform.test:8000" instead of "null".

But those are bugs from `createDocument`, not from `cloneNode` -- they both copy `origin` from `doc` to `copy`, which is what the spec change is about.

LGTM.

PR #4276, opened 2016-12-02
Update expectations on innerText's getter for <rp>

zcorpan:
Test update for https://github.com/whatwg/html/pull/2129


zcorpan:
The `Lone <rp>` test fails in Chromium and WebKit because it is not `display:none` outside `<ruby>`. According to https://html.spec.whatwg.org/multipage/rendering.html#hidden-elements it should be.

domenic:
LGTM. Will leave it up to you whether to file an issue on HTML or browsers regarding the mismatch around lone rps.

PR #4283, opened 2016-12-05
Add tests that <a>/<area> can be downloaded without user interaction

domenic:
That is, that calling .click() does not throw at least, as it used to
per-spec. Matches the HTML change at
https://github.com/whatwg/html/pull/2136 per the HTML bug report at
https://github.com/whatwg/html/issues/2116.

/cc @cdumez


zcorpan:
LGTM.

@jgraham this pops up a download dialog in Firefox. Is that a problem?

PR #4284, opened 2016-12-06
Test that form data name/values are converted to USVs

domenic:
Tests https://github.com/whatwg/html/pull/2124, which is a fix for
https://github.com/whatwg/html/issues/1490.

---

These pass in Firefox and Chrome but Edge fails by somehow adding three `\uFFFD` replacement characters instead of one.

I am pretty sure Python and the text/plain serialization is not messing with things, but careful review would be appreciated.

### DIFF
--- a/html/semantics/forms/form-submission-0/form-echo.py
+++ b/html/semantics/forms/form-submission-0/form-echo.py
@@ -1,5 +1,7 @@
 def main(request, response):
+    bytes = bytearray(request.raw_input.read())
+    bytes_string = " ".join("%02x" % b for b in bytes)
     return (
         [("Content-Type", "text/plain")],
-        request.body
+        bytes_string
     )
--- a/html/semantics/forms/form-submission-0/form-echo.py
+++ b/html/semantics/forms/form-submission-0/form-echo.py
@@ -1,0 +1,5 @@
+def main(request, response):
+    return (
+        [("Content-Type", "text/plain")],
+        request.body
+    )
--- a/html/semantics/forms/form-submission-0/form-data-set-usv-form.html
+++ b/html/semantics/forms/form-submission-0/form-data-set-usv-form.html
@@ -1,0 +1,27 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>This is the form that will be submitted</title>
+
+<form action="form-echo.py" method="post" enctype="text/plain">
+  <input id="input1" type="text">
+  <select id="input2">
+    <option selected>option
+  </select>
+  <input id="input3" type="radio" checked>
+  <input id="input4" type="checkbox" checked>
+</form>
+
+<script>
+"use strict";
+
+const form = document.querySelector("form");
+
+for (let el of Array.from(form.querySelectorAll("input"))) { // Firefox/Edge support
+  el.name = el.id + "\uDC01";
+  el.value = el.id + "\uDC01";
+}
+
+const select = document.querySelector("select");
+select.name = select.id + "\uDC01";
+select.firstElementChild.value = select.id + "\uDC01";
+</script>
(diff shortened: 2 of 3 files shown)


html/semantics/forms/form-submission-0/form-echo.py
```diff
@@ -0,0 +1,5 @@
+def main(request, response):
+    return (
+        [("Content-Type", "text/plain")],
```
zcorpan: Need to set `; charset=utf-8` here

zcorpan (changes requested):
Possibly Edge has a default charset of text/plain that is us-ascii, and their us-ascii decoder replaces bytes above 0x7F with U+FFFD. That would explain why it has three U+FFFDs. But I haven't tested Edge to confirm.

In Opera I get this in Packet Peeper for form-echo.py request/response:
```
POST /html/semantics/forms/form-submission-0/form-echo.py HTTP/1.1
Host: web-platform.test:8000
Connection: keep-alive
Content-Length: 82
Pragma: no-cache
Cache-Control: no-cache
Origin: http://web-platform.test:8000
Upgrade-Insecure-Requests: 1
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.44 Safari/537.36 OPR/42.0.2393.65 (Edition beta)
Content-Type: text/plain
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8
Referer: http://web-platform.test:8000/html/semantics/forms/form-submission-0/form-data-set-usv-form.html
Accept-Encoding: gzip, deflate, lzma
Accept-Language: sv,en-US;q=0.8,en;q=0.6

input1...=input1...
input2...=input2...
input3...=input3...
input4...=input4...HTTP/1.1 200 OK
Content-Type: text/plain
Server: BaseHTTP/0.3 Python/2.7.10
Date: Tue, 06 Dec 2016 21:18:20 GMT
Content-Length: 82

input1...=input1...
input2...=input2...
input3...=input3...
input4...=input4...
```

("..." means 3 non-ASCII bytes.)

domenic:
Hmm. What I really want to do is to test the bytes that the server sees. Is anyone familiar enough with the WPT framework and with Python to help me with that? I'm thinking that maybe I should write out the body as a series of numeric byte sequences and then I can test against the string containing those numeric byte sequences.

That would at least avoid the bytes -> text/plain avenue for confusion, although the form data set -> text/plain encoded bytes step is still mixed up in here.

domenic:
OK, I think I got that part...

```python
def main(request, response):
    bytes = bytearray(request.raw_input.read())
    bytes_string = " ".join("%02x" % b for b in bytes)
    return (
        [("Content-Type", "text/plain")],
        bytes_string
    )
```

Will incorporate that later

zcorpan:
> the form data set -> text/plain encoded bytes step is still mixed up in here.

Maybe there could be several tests (but that could be a followup).

* https://html.spec.whatwg.org/#submit-mutate-action
* https://html.spec.whatwg.org/#submit-body - 3 tests, one for each `enctype`
* https://html.spec.whatwg.org/#submit-get-action
* https://html.spec.whatwg.org/#submit-dialog

(Skip mailto: since that would require manual tests I suppose.)

gsnedders:
@domenic You don't need the `bytearray` call, you can just have the return result of `.read()`. I'm thinking there should be a more performant way of doing it, but idk what.

domenic:
@gsnedders it's needed to do array comprehensions over it, it appears.

@zcorpan agreed those would be good follow-up tests.

I updated this to test the bytes more directly. It will now fail in Chrome because Chrome does not append \r\n. It seemed a bit silly to remove a trailing \r\n from the byte sequence.

I guess we should figure out whether that's a spec bug or a Chrome bug...

gsnedders:
@domenic oh, the difference is an iterator over a `bytearray` returns ints, whereas over a `str` object it gives `str` objects with length one.

zcorpan (approved):

domenic:
@gsnedders oh, so `raw_input.read()` returns a string? That's not great :(. Is there a way to get the bytes, without any text decoding?

gsnedders:
@domenic This is all Python 2, so `str` is a sequence of bytes.

PR #4289, opened 2016-12-06
Add tests to verify behavior of workers in an opaque origin document.

mkruisselbrink:
Kind of an edge case, but all of these tests are currently failing in chrome for various reasons, and this is all stuff that is supposed to work (and works in firefox). So possibly still valuable to have tests for this.


mkruisselbrink:
Could anybody take a look at this?

PR #4290, opened 2016-12-06
Encoding: Consolidate label tests to one per label

inexorabletash:
Restructure the encoding label tests to be one per label (~200) containing multiple assertions, rather than one per assertion (~3000).

### DIFF
--- a/encoding/textdecoder-labels.html
+++ b/encoding/textdecoder-labels.html
@@ -4,33 +4,32 @@
 <script src="/resources/testharnessreport.js"></script>
 <script src="resources/encodings.js"></script>
 <script>
-var tests = [];
-setup(function() {
-  var whitespace = [' ', '\t', '\n', '\f', '\r'];
-  encodings_table.forEach(function(section) {
-    section.encodings.filter(function(encoding) {
-      return encoding.name !== 'replacement';
-    }).forEach(function(encoding) {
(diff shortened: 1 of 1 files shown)


inexorabletash:
@annevk or @hsivonen - can you take a peek?

This is purely motivated by the test being particularly slow when run under the "MSan" bot in Blink's CI. If anyone prefers it as-is we can leave it.

encoding/textdecoder-labels.html
```diff
@@ -4,33 +4,32 @@
...
+  }).forEach(function(encoding) {
+    encoding.labels.forEach(function(label) {
+      test(function(t) {
+        assert_equals(
+          new TextDecoder(label).encoding, encoding.name,
+          'label for encoding should match');
+        assert_equals(
+          new TextDecoder(label.toUpperCase()).encoding, encoding.name,
+          'label matching should be case-insensitive');
+        var whitespace = [' ', '\t', '\n', '\f', '\r'];
```
annevk: Seems better to define this outside of the loops?

Anyway, looks okay to me overall. This is your test and @Ms2ger modified it a bit. I don't really care how it's formatted personally, as long as it still tests the same things and we get the same relative failure rates.

annevk (approved):

PR #4294, opened 2016-12-07
Align BroadcastChannel test with the HTML Standard

annevk:
Per https://github.com/whatwg/html/issues/1319 this should simply work.


mkruisselbrink:
Thanks. We should probably have some tests that make sure this actually works correctly, but this is at least better than the current test.

PR #4297, opened 2016-12-08
EE CSP: Adding tests for csp attribute on iframe

aubakirova:
@mikewest


mikewest:
Reviewed upstream in Blink. LGTM!

PR #4300, opened 2016-12-08
Add tests for fullscreenchange and fullscreenerror event timing


jeffcarp:
Although HTML does not yet define animation frame tasks as used by
Fullscreen, it does have "run the fullscreen rendering steps". That's
the old hook, making the timing of these events relative to resize and
animation frame callbacks unambiguous:
https://fullscreen.spec.whatwg.org/#dom-element-requestfullscreen
https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-processing-model

Spec issue: https://github.com/whatwg/html/issues/707

document-exit-fullscreen-timing-manual.html consistently matches
-expected.txt, with (accidentally) correct timing but no resize event.

element-request-fullscreen-timing-manual.html usually matches
-expected.txt, with incorrect timing, but can accidentally get the
timing right and instead fail due to no resize event.

Both are marked as flaky to be safe, and will be enabled in a coming CL.

Note: This CL is also a test for the WPT export process.

Firefox passes both fullscreechange event tests, but fails the
fullscreenerror test. (Implementation preceded spec changes.)

BUG=402376,672436

Review-Url: https://codereview.chromium.org/2564543002
Cr-Commit-Position: refs/heads/master@{#437283}

### DIFF
--- a/fullscreen/api/document-exit-fullscreen-timing-manual-expected.txt
+++ b/fullscreen/api/document-exit-fullscreen-timing-manual-expected.txt
@@ -1,0 +1,4 @@
+This is a testharness.js-based test.
+FAIL Timing of fullscreenchange and resize events assert_array_equals: event order lengths differ, expected 2 got 1
+Harness: the test ran to completion.
+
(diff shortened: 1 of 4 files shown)


jeffcarp:
Incorrect patch, trying again.

PR #4306, opened 2016-12-09
Test for the absence of DOMStringList.

Ms2ger:


domenic:
This seems incorrect given https://w3c.github.io/IndexedDB/#domstringlist-api

zcorpan:
Huh. I guess https://dom.spec.whatwg.org/#domstringlist should point to that, then...

zcorpan:
PR for DOM standard: https://github.com/whatwg/dom/pull/385

PR #4308, opened 2016-12-09
EE: Embedding-CSP Header

aubakirova:


Sun77789:
@mikewest 
Corresponding test in [chrome](https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/http/tests/security/contentSecurityPolicy/embeddedEnforcement/embedding_csp-header.html?sq=package:chromium&dr=C)

content-security-policy/support/echo-embedding-csp.py
```diff
@@ -0,0 +1,11 @@
+import json
+def main(request, response):
+    header = request.headers.get("Embedding-CSP");
+    message = {}
+    message['embedding_csp'] = header if header else None
+    return [("Content-Type", "text/html")], '''
```
mikewest: I'd suggest adding the `allow` header as well so you can do cross-origin tests.

content-security-policy/embedded-enforcement/embedding_csp-header.sub.html
```diff
@@ -0,0 +1,109 @@
...
+        i.src = embedding_src;
+
+        window.addEventListener('message', t.step_func(e => {
+          if (e.source != i.contentWindow)
+              return;
+          assert_equals(e.data['embedding_csp'], null);
+          t.done();
+        }));
+
+        document.body.appendChild(i);
```
mikewest: Several of these tests could be combined into something a little more streamlined. Perhaps two new helpers:

```
function wait_for_message_from_frame(i) {
    window.addEventListener("message", test.step_func(e => {
      if (e.source == i.contentWindow)
        resolve(e);
    }));
  });
}

function assert_embedding_csp(test, src, csp, expected) {
  var i = document.createElement('iframe');
  if (csp !== null)
    i.csp = csp;
  i.src = src;

  wait_for_message_from_frame(i)
    .then(t.step_func_done(e => {
      assert_equals(e.data['embedding_csp'], expected);
      i.remove();
    }));

  document.body.appendChild(i);
}
```

And then tests like:

```
var tests = [
  { "src": embedding_src, "csp": null, "expected": null },
  { "src": embedding_src, "csp": "script-src 'unsafe-inline'", "expected": "script-src 'unsafe-inline'" },
  { "src": embedding_src, "csp": "totally invalid csp", "expected": null },
  ...
];

tests.forEach(test => {
  async_test(t => assert_embedding_csp(t, test.src, test.csp, test.expected));
});
```

content-security-policy/support/echo-embedding-csp.py
Sun77789: I tried added `("Access-Control-Allow-Origin", "*")`
but that didn't seem to help with the commented out code in the html file?

content-security-policy/embedded-enforcement/embedding_csp-header.https.html
```diff
@@ -0,0 +1,63 @@
...
+    var tests = [
+      { "name": "Embedding-CSP is not sent if csp attribute is not set on <iframe>.", 
+          "csp": null, "expected": null,"new_csp": null },
+      { "name": "Send Embedding-CSP when csp attribute of <iframe> is not empty.", 
+          "csp": "script-src 'unsafe-inline'", "expected": "script-src 'unsafe-inline'", "new_csp": null },
+      { "name": "Send Embedding-CSP Header on change of `src` attribute on iframe.", 
+          "csp": "script-src 'unsafe-inline'", "expected": "script-src 'unsafe-inline'", "new_csp": "default-src 'unsafe-inline'"},
+      // TODO(amalika): Validate csp attribute to be a proper policy.
+      // { "name": "Send Embedding-CSP Header on change of `src` attribute on iframe.", 
+      //     "csp": "completely wrong csp", "expected": null, "new_csp": null},
```
Sun77789: Right now we have no validation of the correctness of the csp attribute so we send Embedding-CSP even if the value is not a proper CSP.

content-security-policy/embedded-enforcement/embedding_csp-header.https.html
```diff
@@ -0,0 +1,63 @@
+<!DOCTYPE html>
```
Sun77789: This file is of type `https.html` which I copied from the upgrade-insecure-request, but not sure it is necessary here?

content-security-policy/support/echo-embedding-csp.py
mikewest: Sorry, I meant the `Allow-CSP-From` header you added.

Sun77789: I think the tests would be slightly different. Here we expect to get Embedding-CSP (from say a browser) and if so, post a message.
In the other case, depending on what browser sent, we decide to send or not Allow-CSP-From header. But I am working on that PR now!

content-security-policy/embedded-enforcement/embedding_csp-header.https.html
```diff
@@ -0,0 +1,63 @@
...
+      }, "Test cross origin redirect: " + test.name);
+    });
+
+    // tests.forEach(test => {
+    //   async_test(t => {
+    //     var url = generateURL(Host.CROSS_ORIGIN, PolicyHeader.EMBEDDING_CSP);
+    //     var redirect_url = generateRedirect(Host.CROSS_ORIGIN, url);
+    //     assert_embedding_csp(t, redirect_url, test.csp, test.new_csp, null);
+    //   }, "Test cross origin redirect of cross origin iframe: " + test.name);
+    // });
```
Sun77789: This times out. Not sure if it is supposed to work?

content-security-policy/embedded-enforcement/embedding_csp-header.https.html
Sun77789: Great! Thanks :)

content-security-policy/support/testharness-helper.sub.js
```diff
@@ -0,0 +1,47 @@
...
+  CROSS_ORIGIN: "cross-origin",
+};
+
+const PolicyHeader = {
+  CSP: "echo-policy.py?policy=",
+  EMBEDDING_CSP: "echo-embedding-csp.py",
+} 
+
+function generateURL(host, path) {
+  var url = new URL("http://{{host}}:{{ports[http][0]}}/content-security-policy/support/");
```
mikewest: you'll need either to remove the `.https` suffix from the test, or change this to load over HTTPS (by changing the protocol to `https`, and the port to `ports[https][0]`.

content-security-policy/support/testharness-helper.sub.js
```diff
@@ -0,0 +1,47 @@
...
+  url.hostname = host == Host.SAME_ORIGIN ? "{{host}}" : "{{domains[天気の良い日]}}";
+  url.pathname += path;
+
+  return url.toString();
+}
+
+function generateRedirect(host, target) {
+  var url = "http://{{host}}:{{ports[http][0]}}/common/redirect.py?location=" + encodeURIComponent(target);
```
mikewest: Ditto.

mikewest: Also, you can write this as `new URL(...)` and change line 22 to set the URL's hostname (as above).

content-security-policy/support/testharness-helper.sub.js
```diff
@@ -0,0 +1,47 @@
...
+  if (host == Host.CROSS_ORIGIN) 
+    url = "http://{{domains[天気の良い日]}}:{{ports[http][0]}}/common/redirect.py?location=" + encodeURIComponent(target);
+
+  return url.toString();
+}
+
+function assert_embedding_csp(t, url, csp, expected, newCsp=null, newSrc=null) {
```
mikewest: Nit: Drop `=null`. They'll be undefined by default, which is falsey.

content-security-policy/support/testharness-helper.sub.js
```diff
@@ -0,0 +1,47 @@
...
+      expected = newCsp;
+    assert_equals(expected, e.data['embedding_csp']);
+    t.done();
+  }));
+
+  document.body.appendChild(i);
+  if (newCsp)
+    i.csp = newCsp;
+  if (newSrc)
+    i.src = newSrc;
```
mikewest: It's not clear to me what you're testing here. If you want to verify that the current attribute is used for the current request, then I think you also need to verify that the initial request has the expected policy. That is, something like listening for the first message, then changing the attribute and navigating again.

As written, I don't think the first request will complete (because you'll reset `src` before the frame's `onload` event fires.

mikewest (commented):
Thanks!

content-security-policy/embedded-enforcement/embedding_csp-header.https.html
mikewest: It might timeout because you're loading an insecure frame into a secure document (and is blocked by mixed content).

content-security-policy/support/testharness-helper.sub.js
Sun77789: I was thinking it would be sufficient to check that the latest CSP is always used since we append the iframe and then change it. Since old and new csp are different in the tests, this seemed sufficient.

Would you prefer if I remove these bits or rather if add additional check for the first message as you mentioned?

content-security-policy/support/testharness-helper.sub.js
Sun77789: I tried changing everything to `https`, but then I get "Aw, Snap!" page.
When I downgrade to `http` the last `for_each` test (cross origin redirect for cross origin iframe) is still failing :(

content-security-policy/support/testharness-helper.sub.js
mikewest: > I was thinking it would be sufficient to check that the latest CSP is always used since we append the iframe and then change it.

Right, but you're setting up a race condition by setting `i.src` to `url`, then appending the element (which kicks off a load), then modifying `i.src` to `newSrc` (which kicks off another load, probably before the first load has finished, but maybe not). That means that the `message` event you're listening for might come from either the first or second load, based purely upon how the browser decided to execute your code.

If you want to verify that setting the `csp` attribute doesn't affect an already-loaded page, I think a better way of doing so would be to wait for the `load` event. If you want to verify that setting the `csp` attribute before a page completes its load, then I think you need to set the expectations more strictly (e.g. drop the `e.data['expected_csp'] == null` carveout).

content-security-policy/support/testharness-helper.sub.js
mikewest: I think cross-origin things will fail as long as you aren't sending an `Allow-CSP-From` header, or a `Content-Security-Policy` header that matches the embedder's request. I mean, that's the whole feature, right? :)

content-security-policy/support/echo-embedding-csp.py
mikewest: See below:

mikewest:
Skimmed through this, I think the attribute tests are great, but the `embedding-csp` header needs some work.

content-security-policy/support/testharness-helper.sub.js
Sun77789: I added it as a separate test to make sure the iframe has loaded and then changing `src` attribute to cross origin url.

content-security-policy/support/testharness-helper.sub.js
Sun77789: Ahhh.. Thinking that Embedding-CSP should be sent no matter what, I forgot that however post message wouldn't be sent if it is not loaded! All tests fixed ( except the one on the wrong value of csp that we don't validate)

mikewest (approved):
LGTM, thanks!

PR #4309, opened 2016-12-09
Test that PaymentRequest throws when its document is not active

zcorpan:
See
https://github.com/whatwg/html/pull/2160
https://github.com/w3c/browser-payment-api/issues/361


zcorpan:
https://github.com/w3c/web-platform-tests/pull/4309/commits/9f9123869594478a832c19efd67a8b46eaec6a4d tests requestFullscreen, and passes in Gecko. It fails in Chromium but the test I based it on (element-ready-check-containing-iframe-manual.html) also fails, unclear to me what's up. @annevk @foolip ?

zcorpan:
The PaymentRequest tests need Experimental Web Platform features enabled in Chrome. For me those 3 tests pass in Chrome canary 57.0.2951.0 with experimental features enabled.

zcorpan:
I see now why `requestFullscreen` fails in Chromium. It's still prefixed, `webkitRequestFullScreen`.

If I change the test to use the prefixed name then Chromium neither enters fullscreen but also doesn't fire a `fullscreenerror` event, so the test times out.

zcorpan:
Unprefixing is https://bugs.chromium.org/p/chromium/issues/detail?id=383813

> If I change the test to use the prefixed names then Chromium neither enters fullscreen but also doesn't fire a webkitfullscreenerror event, so the test times out.

Filed https://bugs.chromium.org/p/chromium/issues/detail?id=674454

fullscreen/api/active-document-manual.html
```diff
@@ -0,0 +1,18 @@
...
+<script src="/resources/testharnessreport.js"></script>
+<script src="../trusted-click.js"></script>
+<div id="log"></div>
+<iframe allowfullscreen></iframe>
+<script>
+async_test((t) => {
+  onload = () => {
+    var iframes = document.getElementsByTagName("iframe");
+    trusted_request(iframes[0].contentDocument.body, document.body);
+    window[0].location.href = '/common/blank.html';
```
foolip: This throws an exception I guess?

fullscreen/api/active-document-manual.html
```diff
@@ -0,0 +1,18 @@
...
+<title>Element ready check for containing iframe, element's document is not the active document</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="../trusted-click.js"></script>
+<div id="log"></div>
+<iframe allowfullscreen></iframe>
+<script>
+async_test((t) => {
+  onload = () => {
+    var iframes = document.getElementsByTagName("iframe");
```
foolip: Maybe `document.querySelector("iframe")` to save some later `[0]`?

fullscreen/api/active-document-manual.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
+<title>Element ready check for containing iframe, element's document is not the active document</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="../trusted-click.js"></script>
+<div id="log"></div>
+<iframe allowfullscreen></iframe>
+<script>
+async_test((t) => {
+  onload = () => {
```
foolip: Don't need to wait for load I think, but if you `t.step_func` is needed. (Or is it? I can never remember if testharness.js gets it right if there's just a single async_test.

foolip (commented):
Suggest splitting Fullscreen into a separate PR as tests for https://github.com/whatwg/fullscreen/issues/67

fullscreen/api/active-document-manual.html
zcorpan: No?

fullscreen/api/active-document-manual.html
zcorpan: Right onload is not necessary. Will fix.

fullscreen/api/active-document-manual.html
foolip: What does `window[0]` mean? Isn't it just undefined?

zcorpan:
Fullscreen test moved to https://github.com/w3c/web-platform-tests/issues/4334

PR #4310, opened 2016-12-09
EE: Allow-CSP-From Header Tests

aubakirova:
@mikewest 
Some of the code repeats because I didnt want to rebase on the other PR.
(Otherwise, github would display all the changes).

### DIFF
--- a/content-security-policy/embedded-enforcement/allow_csp_from-header.html
+++ b/content-security-policy/embedded-enforcement/allow_csp_from-header.html
@@ -1,0 +1,88 @@
+<!DOCTYPE html>
+<html>
+<head>
+<title>Embedded Enforcement: Allow-CSP-From header.</title>
+  <script src="/resources/testharness.js"></script>
+  <script src="/resources/testharnessreport.js"></script>
+  <script src="../support/testharness-helper.sub.js"></script>
+</head>
+<body>
+  <script>
+    var tests = [
+      { "name": "Same origin iframes are always allowed.", 
+        "origin": Host.SAME_ORIGIN,
+        "csp": "style-src 'unsafe-inline'; script-src 'unsafe-inline'", 
+        "allow_csp_from": "¢¥§",
+        "expected": IframeLoad.EXPECT_LOAD,
+        "blockedURI": null},
+      { "name": "Same origin iframes are allowed even if the Allow-CSP-From is empty.", 
+        "origin": Host.SAME_ORIGIN,
+        "csp": "style-src 'unsafe-inline'; script-src 'unsafe-inline'", 
+        "allow_csp_from": "",
+        "expected": IframeLoad.EXPECT_LOAD,
+        "blockedURI": null},
+      { "name": "Same origin iframes are allowed even if the Allow-CSP-From is not present.", 
+        "origin": Host.SAME_ORIGIN,
+        "csp": "style-src 'unsafe-inline'; script-src 'unsafe-inline'", 
+        "allow_csp_from": null,
+        "expected": IframeLoad.EXPECT_LOAD,
+        "blockedURI": null},
+      { "name": "Same origin iframes are allowed even if Allow-CSP-From does not match origin.", 
+        "origin": Host.SAME_ORIGIN,
+        "csp": "style-src 'unsafe-inline'; script-src 'unsafe-inline'", 
+        "allow_csp_from": "http://example.com:888",
+        "expected": IframeLoad.EXPECT_LOAD,
+        "blockedURI": null},
+      { "name": "Cross origin iframe with an empty Allow-CSP-From header gets blocked.", 
+        "origin": Host.CROSS_ORIGIN,
+        "csp": "script-src 'unsafe-inline'", 
+        "allow_csp_from": "",
+        "expected": IframeLoad.EXPECT_BLOCK,
+        "blockedURI": null},
+      { "name": "Cross origin iframe without Allow-CSP-From header gets blocked.", 
+        "origin": Host.CROSS_ORIGIN,
+        "csp": "script-src 'unsafe-inline'", 
+        "allow_csp_from": null,
+        "expected": IframeLoad.EXPECT_BLOCK,
+        "blockedURI": null},
+      { "name": "iframe from cross origin does not load without Allow-CSP-From header.", 
+        "origin": Host.CROSS_ORIGIN,
+        "csp": "style-src 'unsafe-inline'; script-src 'unsafe-inline'", 
+        "allow_csp_from": getOrigin(),
+        "expected": IframeLoad.EXPECT_LOAD,
+        "blockedURI": null},
+      { "name": "Iframe with improper Allow-CSP-From header gets blocked.", 
+        "origin": Host.CROSS_ORIGIN,
+        "csp": "script-src 'unsafe-inline'", 
+        "allow_csp_from": "* ¢¥§",
+        "expected": IframeLoad.EXPECT_BLOCK,
+        "blockedURI": null},
+      { "name": "Allow-CSP-From header with a star value can be returned.", 
+        "origin": Host.CROSS_ORIGIN,
+        "csp": "script-src 'unsafe-inline'", 
+        "allow_csp_from": "*",
+        "expected": IframeLoad.EXPECT_LOAD,
+        "blockedURI": null},
+      { "name": "Star Allow-CSP-From header enforces EmbeddingCSP.", 
+        "origin": Host.CROSS_ORIGIN,
+        "csp": "script-src 'nonce-123'", 
+        "allow_csp_from": "*",
+        "expected": IframeLoad.EXPECT_LOAD,
+        "blockedURI": "inline"},
+      { "name": "Allow-CSP-From header enforces EmbeddingCSP.", 
+        "origin": Host.CROSS_ORIGIN,
+        "csp": "style-src 'none'; script-src 'nonce-123'", 
+        "allow_csp_from": getOrigin(),
+        "expected": IframeLoad.EXPECT_LOAD,
+        "blockedURI": "inline"},
+    ];
+
+    tests.forEach(test => {
+      async_test(t =>  {
+        var url = generateUrlWithAllowCSPFrom(test.origin, test.allow_csp_from);
+        assert_iframe_with_csp(t, url, test.csp, test.expected, test.name, test.blockedURI);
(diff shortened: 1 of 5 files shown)


content-security-policy/embedded-enforcement/allow_csp_from-header.html
```diff
@@ -0,0 +1,64 @@
...
+<title>Embedded Enforcement: Allow-CSP-From header.</title>
+  <script src="/resources/testharness.js"></script>
+  <script src="/resources/testharnessreport.js"></script>
+  <script src="../support/testharness-helper.sub.js"></script>
+</head>
+<body>
+  <script>
+    var tests = [
+      { "name": "Same origin iframes are always allowed.", 
+          "origin": Host.SAME_ORIGIN,
```
mikewest: Nit: Spacing is weird here.

content-security-policy/support/testharness-helper.sub.js
```diff
@@ -0,0 +1,81 @@
...
+  if (shouldBlock) {
+    // Assert iframe does not load.
+    window.onmessage = function (e) {
+      if (e.source != i.contentWindow)
+          return;
+      t.unreached_func('No message should be sent from the frame.');
+    }
+    i.onload = t.step_func(function () {
+      // Delay the check until the postMessage has a chance to execute.
+      setTimeout(t.step_func_done(function () {
```
mikewest: You should also be able to assert that you can't access the DOM of this document (meaning that it's either cross-origin, or a blocked same-origin document). Something like `assert_throws(SecurityError, _ => { var x = i.contentWindow.location.href; });` (I forget the syntax; skim through other tests for `assert_throws` :)

content-security-policy/support/testharness-helper.sub.js
```diff
@@ -0,0 +1,81 @@
...
+           assert_equals(loaded[urlId], undefined);
+      }), 1);
+    });
+  } else if (blockedURI) {
+    // Assert iframe loads with an expected violation.
+    window.addEventListener('message', t.step_func(e => {
+      if (e.source != i.contentWindow || e.data["securitypolicyviolation"] != true)
```
mikewest: Shouldn't the test fail if `securitypolicyviolation` is `false` and `blockedURI` isn't?

content-security-policy/support/testharness-helper.sub.js
```diff
@@ -0,0 +1,81 @@
...
+        return;
+      assert_equals(e.data["blockedURI"], blockedURI);
+      t.done();
+    }));
+  } else {
+    // Assert iframe loads.
+    document.addEventListener("securitypolicyviolation",
+      t.unreached_func("There should not be any violations."));
```
mikewest: There should never be violations in the parent document, right?

mikewest (commented):
Thanks, this looks like a good start!

content-security-policy/support/testharness-helper.sub.js
Sun77789: That indeed seems redundant. Removed!

content-security-policy/embedded-enforcement/allow_csp_from-header.html
Sun77789: Not sure why I for some reasons thought the lines after the first should be more indented  ¯\_(ツ)_/¯
Also, I tried running "./lint" but it does not display anything in the terminal, but fixed the  spacing here!

content-security-policy/support/echo-allow-csp-from.py
```diff
@@ -0,0 +1,37 @@
...
+    <style>
+        body {
+            background-color: maroon;
+        }
+    </style>
+    <script nonce="abc"> 
+        var response = {};
+        response["id"] = "%s";
+        response["loaded"] = true;
+        window.onload = window.top.postMessage(response, '*');
```
mikewest: If you just execute `postMessage` rather than waiting for `onload`, then you can drop the timeout in the test file.

content-security-policy/support/testharness-helper.sub.js
```diff
@@ -39,3 +55,53 @@ function assert_embedding_csp(t, url, csp, expected) {
...
+      }), 1);
+
+      assert_throws("SecurityError", () => {
+        var x = i.contentWindow.location.href;
+      });
+    });
+  } else if (blockedURI) {
+    // Assert iframe loads with an expected violation.
+    window.addEventListener('message', t.step_func(e => {
+      if (e.source != i.contentWindow || !e.data["blockedURI"])
```
mikewest: Is it ever acceptable for a message from a frame which expects |blockedURI| to send a message that doesn't contain the value? Shouldn't we consider that a failure?

mikewest (changes requested):
small suggestions, thanks!

mikewest:
Can you rebase this one as well?

Sun77789:
Rebased!

content-security-policy/embedded-enforcement/allow_csp_from-header.html
```diff
@@ -0,0 +1,64 @@
...
+  <script src="/resources/testharness.js"></script>
+  <script src="/resources/testharnessreport.js"></script>
+  <script src="../support/testharness-helper.sub.js"></script>
+</head>
+<body>
+  <script>
+    var tests = [
+      { "name": "Same origin iframes are always allowed.", 
+        "origin": Host.SAME_ORIGIN,
+        "csp": "style-src 'unsafe-inline'; script-src 'unsafe-inline';",
```
mikewest: Nit: Do you need the trailing `;` at the end of all of these? It's not necessary from a grammar perspective, so... :)

content-security-policy/embedded-enforcement/allow_csp_from-header.html
```diff
@@ -0,0 +1,64 @@
...
+  <script src="/resources/testharnessreport.js"></script>
+  <script src="../support/testharness-helper.sub.js"></script>
+</head>
+<body>
+  <script>
+    var tests = [
+      { "name": "Same origin iframes are always allowed.", 
+        "origin": Host.SAME_ORIGIN,
+        "csp": "style-src 'unsafe-inline'; script-src 'unsafe-inline';", 
+        "allow_csp_from": "",
```
mikewest: Can you add some more variants of this test? Invalid values, valid values that don't match the domain or port or scheme, valid values that do match, etc.

mikewest: (Basically, same origin should always load)

content-security-policy/embedded-enforcement/allow_csp_from-header.html
```diff
@@ -0,0 +1,64 @@
...
+        "expected": IframeLoad.EXPECT_LOAD,
+        "blockedURI": null},
+      { "name": "Cross origin iframe without Allow-CSP-From header gets blocked.", 
+        "origin": Host.CROSS_ORIGIN,
+        "csp": "script-src 'unsafe-inline';", 
+        "allow_csp_from": "",
```
mikewest: Can you distinguish between an _empty_ `allow-csp-from` and response where the header isn't present? Perhaps `null` as opposed to `""`?

mikewest (changes requested):
Mostly LGTM, thanks!

content-security-policy/embedded-enforcement/allow_csp_from-header.html
Sun77789: Removed!

content-security-policy/embedded-enforcement/allow_csp_from-header.html
Sun77789: Added :) thanks!

content-security-policy/embedded-enforcement/allow_csp_from-header.html
Sun77789: Added checks for empty allow_csp_from parameter and null checks

content-security-policy/support/testharness-helper.sub.js
```diff
@@ -41,6 +41,13 @@ function generateUrlWithPolicies(host, policy) {
     return url + "?policy=" + encodeURIComponent(policy);
 }
 
+function generateUrlWithAllowCSPFrom(host, allowCspFrom) {
+  var url = generateURL(host, PolicyHeader.ALLOW_CSP_FROM);
+  if (!allowCspFrom)
+    return url + "?allow_csp_from=";
```
mikewest: I'm pretty sure that this is still going to treat `""` and `null` the same (because both are `== false`).

mikewest: Also, it's going to look the same to Python. Perhaps not sending `allow_csp_from` at all would be a better way to signal to the backend that it shouldn't set the header at all (as opposed to setting it to an empty value).

mikewest (changes requested):

content-security-policy/support/testharness-helper.sub.js
Sun77789: I also added a check in the py file `request.GET["allow_csp_from"]`` but indeed then both versions do not send Allow-CSP-From header. 
I did not remove `"?allow_csp_from="` because otherwise GET parsing is weird since "&id" is added to url in `assert_iframe_with_csp`. So I guess better solution is just adding "?" to the url. Updated!

content-security-policy/support/testharness-helper.sub.js
```diff
@@ -41,6 +41,13 @@ function generateUrlWithPolicies(host, policy) {
     return url + "?policy=" + encodeURIComponent(policy);
 }
 
+function generateUrlWithAllowCSPFrom(host, allowCspFrom) {
+  var url = generateURL(host, PolicyHeader.ALLOW_CSP_FROM);
+  if (allowCspFrom == null)
+    return url + "?";
+  return url + "?allow_csp_from=" + encodeURIComponent(allowCspFrom);
```
mikewest: We can simplify this:

```
 var url = generateURL(host, PolicyHeader.ALLOW_CSP_FROM);
 if (allowCspFrom != null)
   url.searchParams.append("allow_csp_from", allowCspFrom);
 return url;
```

`URLSearchParams` will take care of the encoding for you; sorry I didn't think of that yesterday.

Likewise, you can update the other GET string operations to use that interface on the URL object. That should make the logic easier to follow.

mikewest (approved):
Thanks!

PR #4314, opened 2016-12-12
Show the line where JSON.parse throw an error

rianby64:
Currently in Chrome and Opera it's not working as expected.
I want to see the line that fired the following error:

  Uncaught (in promise) SyntaxError: Unexpected end of JSON input

Firefox shows correctly the line where this error was originated.

### DIFF
--- a/html/webappapis/scripting/processing-model-2/JSON-runtime-error.html
+++ b/html/webappapis/scripting/processing-model-2/JSON-runtime-error.html
@@ -1,0 +1,39 @@
+<!doctype html>
+<html>
+ <head>
+  <title>window.onerror - JSON runtime error in &lt;script></title>
+  <script src="/resources/testharness.js"></script>
+  <script src="/resources/testharnessreport.js"></script>
+ </head>
+ <body>
+  <div id=log></div>
+  <script>
+    setup({allow_uncaught_exception:true});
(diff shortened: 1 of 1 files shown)


rianby64:
I've some questions about this test

1. Where's a test for JSON.parse?
2. Did I place the test in the correct place?
3. Is it useful?

zcorpan:
1. https://github.com/tc39/test262/search?utf8=✓&q=json.parse

For (2) and (3), while having line numbers certainly seems like a useful thing, the spec for this feature currently [has issues](https://github.com/whatwg/html/issues/958). There are also a bazillion things that can cause an error, and it doesn't scale well to copy-paste this test for each thing. The existing tests are mostly trying to cover each aspect of the HTML spec's requirements, but not testing each possible scenario for each requirement.

That said, testing multiple scenarios and in particular cases people run into that are a problem for them is great to have tests for. So I think:

* The existing test(s) should be modified to also test exact line (and column?), but those should probably be new separate `test()`s.
* We should figure out a way to test several scenarios for each test in a DRY way, so it's reasonably easy to maintain the tests. Maybe a build script could work? (Maybe this could be a follow-up though; if so, please file a new issue.)

foolip:
Closing due to no submitter response.

PR #4317, opened 2016-12-12
Add basic tests for the Clear-Site-Data header

msramek:
Clear-Site-Data (https://www.w3.org/TR/clear-site-data/) is a header
instructing the user agent to delete certain datatypes.

This commit adds the testing infrastructure (a server serving this
header and JS utility functions), and a basic test verifying that
the header is respected for cookies and storage deletions on
a navigation.

Cache tests will be added in a followup change. Subresource request
tests are not yet added, as they are not implemented by any browser.


msramek:
Hello @mikewest ! Could you have a look at this?

clear-site-data/support/server.py
```diff
@@ -0,0 +1,11 @@
+import json
```
mikewest: Nit: We usually call files like this something like `echo-clear-site-data.py`, just so that the intent is clear.

clear-site-data/support/server.py
```diff
@@ -0,0 +1,11 @@
...
+
+KNOWN_DATATYPES = ["cookies", "storage", "cache"]
+
+# A support server that receives a list of datatypes in the GET query
+# and returns a Clear-Site-Data header with those datatypes.
+def main(request, response):
+    types = [key for key in request.GET.keys() if key in KNOWN_DATATYPES]
```
mikewest: Don't you want to test the effect of invalid types too?

clear-site-data/navigation.html
```diff
@@ -0,0 +1,76 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <script src="/resources/testharness.js"></script>
+    <script src="/resources/testharnessreport.js"></script>
+    <script src="support/test_utils.js"></script>
+
+    <script>
+      setup({ "explicit_done": true });
```
mikewest: I don't think you need this.

clear-site-data/navigation.html
```diff
@@ -0,0 +1,76 @@
...
+          });
+
+          iframe.src = TestUtils.getClearSiteDataUrl(
+              TestUtils.COMBINATIONS[index]);
+        });
+      }
+
+      // Run the test for the first combination. The rest will be executed
+      // recursively.
+      window.onload = function() { testCombinationWithIndex(0); }
```
mikewest: If you need a set of tests to run in serial, use `promise_test` rather than `async_test`. The former waits for each test to finish before running the next. Setting things up that way would remove a lot of the complexity here (as well as the `explicit_done` bits).

Something like:

```
// You need to make this async anyway in order to deal with things like IDB.
function populateDatatypes() {
    for (var i = 0; i < TestUtils.DATATYPES.length; i++) {
      TestUtils.DATATYPES[i].add();
      assert_false(
        TestUtils.DATATYPES[i].isEmpty(),
        TestUtils.DATATYPES[i].name + " has to be nonempty before the test starts.");
    }
    resolve();
  });
}

TestUtils.COMBINATIONS.forEach(combo => {
  promise_test(t => {
    return populateDatatypes()
      .then(_ => {
          var i = document.createElement('iframe');
          i.src = TestUtils.getClearSiteDataUrl(combo);
          i.addEventListener("load", _ => {
            // assert results here
            resolve();
          });
        });
      });
  }, "Clear datatypes on navigation: " + combination.map(function(e) { return e.name; }).join(", "));
});
```

clear-site-data/navigation.html
```diff
@@ -0,0 +1,76 @@
...
+                TestUtils.DATATYPES[i].name +
+                    " has to be nonempty before the test starts.");
+          }
+
+          // Navigate to a resource with a Clear-Site-Data header in an iframe.
+          // Verify that the correct types have been deleted.
+          var iframe = document.createElement("iframe");
+          document.body.appendChild(iframe);
+
+          iframe.addEventListener("load", test.step_func(function() {
```
mikewest: The timing here seems wrong. That is, appending the `iframe` to the body will cause a `load` event pretty quickly, as you haven't set `src` yet, and `about:blank` loads.

mikewest: (Maybe we don't trigger the `about:blank` load until the current task is finished, but that seems like a bad idea to rely upon)

clear-site-data/support/server.py
```diff
@@ -0,0 +1,11 @@
...
+    header = json.dumps({ "types": types })
+    response = "Clear-Site-Data: %s" % header
+    return [("Clear-Site-Data", header)], response
```
mikewest: Did you consider having this page `postMessage` the state of various storage types back up to the parent? That might be simpler (and would allow you to do cross-origin tests in the future).

mikewest (changes requested):
This is looking good! I've left some comments inline.

clear-site-data/navigation.html
msramek: Removed. I needed it with the async tests, since I didn't run them on the same time, to indicate which one was last.

clear-site-data/navigation.html
msramek: Ah, thanks. I tested it and we actually load "about:blank" immediately and synchronously, i.e. before iframe.addEventListener() has a chance to run. So it actually worked correctly (on Chromium). Anyways, we now test the "message" event, and that one comes only from the correctly loaded page.

clear-site-data/navigation.html
msramek: Thanks! I rewrote this section.

We don't currently test the Indexed DB (this test focuses on the recognition of datatypes; I'll later add another test that checks whether 'storage' contains all data storage backends it should). However, adding/checking a cache entry (in a followup pull request) will be asynchronous (because XHR). So I changed add() and isEmpty() to also work with Promise.

clear-site-data/support/server.py
msramek: Done.

clear-site-data/support/server.py
msramek: Not in these particular tests, but eventually yes, so we can remove the restriction.

clear-site-data/support/echo-clear-site-data.py
```diff
@@ -0,0 +1,14 @@
...
+# of the response is "report-status.html", a html site using postMessage
+# to report the status of the datatypes, so that if used in an iframe,
+# it can inform the embedder whether the data deletion succeeded.
+def main(request, response):
+    types = [key for key in request.GET.keys()]
+    header = json.dumps({ "types": types })
+    response = "Clear-Site-Data: %s" % header
+    return ([("Clear-Site-Data", header),
+             ("Content-Type", "text/html")],
+            open("clear-site-data/support/report-status.html", "r").read())
```
mikewest: Nit: Rather than reading a file here, I'd prefer that you put it into heredoc. WPT is sloooow, and I'd prefer to minimize file IO.

mikewest (approved):
LGTM % nit. Thanks for rewriting this!

clear-site-data/support/echo-clear-site-data.py
msramek: Done!

mikewest (approved):
\o/

PR #4319, opened 2016-12-12
Add `step_timeout` in PresentationConnection_onxxx

obstschale:
This adds `step_timeout` to all 3 PresentationConnection_on* tests and relates to number 1 in #4040

### DIFF
--- a/presentation-api/controlling-ua/PresentationConnection_onclosed-manual.html
+++ b/presentation-api/controlling-ua/PresentationConnection_onclosed-manual.html
@@ -28,7 +28,10 @@
         .then(function(connection) {
 
           // Enable timeout again, cause no user action is needed from here.
-          t.step_timeout(function() { t.done(); }, 2000);
+          t.step_timeout(function() {
+              t.force_timeout();
+              t.done();
+          }, 5000);
 
           assert_true(connection instanceof PresentationConnection, 'the connection is setup');
           connection.onclose = t.step_func_done(function(evt) {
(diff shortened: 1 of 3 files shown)


tidoust:
Unless I missed something, this code will not make the test harness report a "timeout" status. There needs to be a call to `force_timeout` before the call to `done`, as in:
https://github.com/w3c/web-platform-tests/blob/master/presentation-api/controlling-ua/startNewPresentation_success-manual.html#L58

Also, could you update the timeout to 5 seconds instead of 2, as in other tests?

obstschale:
I did so. I thought 2 sec would be enough.

Can you explain why we need to call `force_timeout` before `done`?

tidoust:
> I thought 2 sec would be enough.

Right. 2s might be enough, but 5s is the default timeout of the test harness so we should stick to it unless there's a good reason to depart. In an ideal world, we would not need to hardcode this number in tests but would rather have a method to tell the harness to just restart the timeout using whatever duration it uses by default. That method does not exist though.

> Can you explain why we need to call force_timeout before done?

Well, `step_timeout` merely sets a timeout in the JS sense of that world, it does not tell the test harness to consider that the test times out after some duration.

In other words, If you just call `done`, all you're telling the test harness is "look, I'm done with this test". Since no expectation failed, the test harness will happily report a `PASS` outcome. The call to `force_timeout` tells the test harness "look, you should consider that this test timed out". The test harness will then report a `TIMEOUT` outcome.

tidoust (approved):

PR #4320, opened 2016-12-13
Add some tests for the MessagePort.close() method.

mkruisselbrink:
Not a single webmessaging test was calling MessagePort.close(). This adds some basic tests to verify it's behavior. All browsers mostly agree that this is the correct behavior. Among these tests transferring a closed message port is the odd one out. Firefox passes that test, Edge throws an exception when trying to transfer the port, and in Chrome transferring the closed port re-opens the port...

I didn't include tests for what to do when the destination port is closed after postMessage was called on the source port, but before the event loop has run. It's a bit of an edge case, web browsers disagree (only Firefox seems to comply with a strict reading of the spec I believe), and in most cases (where both ports are in different contexts) the difference is indistinguishable anyway.

### DIFF
--- a/webmessaging/message-channels/close.html
+++ b/webmessaging/message-channels/close.html
@@ -1,0 +1,69 @@
+<!DOCTYPE html>
+<meta charset=utf-8>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+async_test(t => {
+    const c = new MessageChannel();
+    c.port1.onmessage = t.unreached_func('Should not have delivered message');
+    c.port1.close();
+    c.port2.postMessage('TEST');
+    setTimeout(t.step_func_done(), 100);
+  }, 'Message sent to closed port should not arrive.');
+
+async_test(t => {
+    const c = new MessageChannel();
+    c.port1.onmessage = t.unreached_func('Should not have delivered message');
+    c.port2.close();
+    c.port2.postMessage('TEST');
+    setTimeout(t.step_func_done(), 100);
+  }, 'Message sent from closed port should not arrive.');
+
+async_test(t => {
+    const c = new MessageChannel();
+    c.port1.onmessage = t.unreached_func('Should not have delivered message');
+    c.port1.close();
+    const c2 = new MessageChannel();
+    c2.port1.onmessage = t.step_func(e => {
+        e.ports[0].postMessage('TESTMSG');
+        setTimeout(t.step_func_done(), 100);
+      });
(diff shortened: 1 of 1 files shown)


webmessaging/message-channels/close.html
```diff
@@ -0,0 +1,69 @@
+<!DOCTYPE html>
+<meta charset=utf-8>
```
pwnall: Please consider adding a `<link rel="help" href="https://html.spec.whatwg.org/multipage/comms.html#dom-messageport-close">`

webmessaging/message-channels/close.html
```diff
@@ -0,0 +1,69 @@
...
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+async_test(t => {
+    const c = new MessageChannel();
+    c.port1.onmessage = t.unreached_func('Should not have delivered message');
+    c.port1.close();
+    c.port2.postMessage('TEST');
+    setTimeout(t.step_func_done(), 100);
```
pwnall: Does this timeout value rely on anything in the spec? (sorry if I missed that)
If not, is there any chance you can make it a constant at the top of the script, so it's easy to tweak either way?

Alternatively, is it possible to have a more robust testing scheme that avoids the timeout altogether? For example, you could create two MessageChannels, close the first one, send messages to both (in that order), and end the test when you get the message on the second channel.

I read https://html.spec.whatwg.org/multipage/comms.html#implementation-advice very briefly, but I think it entails that you'd get the messages in order, since they'd be queued on the same task source?

pwnall:
Thank you for writing the tests, this looks like an important gap that must be filled!

webmessaging/message-channels/close.html
mkruisselbrink: Done

webmessaging/message-channels/close.html
mkruisselbrink: It doesn't really rely on anything in the spec, no. I made it a constant now.

I don't think there really is a way to make this more robust (but at least the worst that can happen is that the test passes when it shouldn't). MessagePorts are specifically defined with each port having its own task source (https://html.spec.whatwg.org/multipage/comms.html#port-message-queue). And browser are free to use separate queues for separate task sources (https://html.spec.whatwg.org/multipage/webappapis.html#task-source) (in fact in the near future Chrome is likely going to start relying on that).

pwnall: Thanks for explaining! I should've read the spec with more care 😄

PR #4322, opened 2016-12-14
SVG 2 marker tests

nikosandronikos:
Some new tests for SVG 2 markers.

**marker-009** tests that the correct number of overlaid markers are rendered on paths with multiple sub-paths and corresponds to the Figure in [13.7.2. Vertex markers: the ‘marker-start’, ‘marker-mid’ and ‘marker-end’ properties](https://svgwg.org/svg2-draft/painting.html#VertexMarkerProperties).

**marker-orient-001** adds additional tests for the `auto` and `auto-start-reverse` values of the [orient](https://svgwg.org/svg2-draft/painting.html#OrientAttribute) attribute.
Note: Every browser tested currently fails this test for one reason or another.

### DIFF
--- a/svg/painting/marker-orient-001-ref.svg
+++ b/svg/painting/marker-orient-001-ref.svg
@@ -1,0 +1,31 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="500" height="500">
+    <defs>
+        <marker id="arrowNW" orient="225" markerUnits="userSpaceOnUse" style="overflow:visible">
+            <path d="M25,0 L 0,-25 L0,25 z" fill="green"/>
+        </marker>
+        <marker id="arrowNE" orient="315" markerUnits="userSpaceOnUse" style="overflow:visible">
+            <path d="M25,0 L 0,-25 L0,25 z" fill="green"/>
+        </marker>
+        <marker id="arrowSE" orient="45" markerUnits="userSpaceOnUse" style="overflow:visible">
+            <path d="M25,0 L 0,-25 L0,25 z" fill="green"/>
+        </marker>
+        <marker id="arrowSW" orient="135" markerUnits="userSpaceOnUse" style="overflow:visible">
+            <path d="M25,0 L 0,-25 L0,25 z" fill="green"/>
+        </marker>
+        <marker id="arrowW" orient="180" markerUnits="userSpaceOnUse" style="overflow:visible">
+            <path d="M25,0 L 0,-25 L0,25 z" fill="green"/>
+        </marker>
+        <marker id="arrowS" orient="90" markerUnits="userSpaceOnUse" style="overflow:visible">
+            <path d="M25,0 L 0,-25 L0,25 z" fill="green"/>
+        </marker>
+        <marker id="arrowAuto" orient="auto" markerUnits="userSpaceOnUse" style="overflow:visible">
+            <path d="M25,0 L 0,-25 L0,25 z" fill="green"/>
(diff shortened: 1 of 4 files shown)


nikosandronikos:
Suggesting @AmeliaBR for review also.

nikosandronikos:
ping @AmeliaBR in case these are still useful.

svg/painting/marker-orient-001-ref.svg
```diff
@@ -0,0 +1,31 @@
...
+        <marker id="arrowAuto" orient="auto" markerUnits="userSpaceOnUse" style="overflow:visible">
+            <path d="M25,0 L 0,-25 L0,25 z" fill="green"/>
+        </marker>
+     </defs>
+    <!-- Test path, with markers -->
+    <path d="M 40,40 l140,140 240,240" marker-start="url(#arrowNW)" marker-mid="url(#arrowSE)" marker-end="url(#arrowSE)" stroke="black" stroke-width="2"/>
+    <path d="M 40,400 l100,-100 100,-100" marker-start="url(#arrowSW)" marker-mid="url(#arrowNE)" marker-end="url(#arrowNE)" stroke="black" stroke-width="2"/>
+    <path d="M 160,40 c40,0 40,0 40,40" marker-start="url(#arrowW)" marker-end="url(#arrowS)" stroke="black" fill="none" stroke-width="2"/>
+    <path d="M 260,140 a 100,100 0 1 1 100,100" marker-start="url(#arrowS)" marker-end="url(#arrowW)" stroke="black" fill="none" stroke-width="2"/>
+    <path d="M 340,100 h50 v50 h-50 z" marker-start="url(#arrowSW)" marker-mid="url(#arrowAuto)" marker-end="url(#arrowNE)" stroke="black" fill="none" stroke-width="2"/>
```
AmeliaBR: If I'm reading [the spec](https://svgwg.org/svg2-draft/painting.html#VertexMarkerProperties) correctly, the start and end markers on this closed path should have the same orientation & therefore overlap.

(Note, this isn't what the browsers I checked do, but the browsers aren't consistent with each other, either!)

```suggestion
    <path d="M 340,100 h50 v50 h-50 z" marker-start="url(#arrowNE)" marker-mid="url(#arrowAuto)" marker-end="url(#arrowNE)" stroke="black" fill="none" stroke-width="2"/>
```

AmeliaBR (commented):
I'm not sure why this wasn't merged three years ago. Maybe just waiting on me to take a look at it.

I'm pretty sure one of the references is incorrect relative to the spec text, see the suggested change.

PS, Is there any way to trigger the browser runs again?

Ms2ger:
@AmeliaBR see the new results at https://wpt.fyi/results/svg/painting?label=pr_head&max-count=1&pr=4322

svg/painting/marker-orient-001-ref.svg
AmeliaBR: Wait, no, this is entirely my error: I missed that this is an `auto-start-reverse` marker, so of course the end marker is flipped!

AmeliaBR:
Thanks @Ms2ger. So, doesn't look like the anti-aliasing issue is a problem any more. (The Chrome & WebKit fails are for obvious stuff.)

And I was wrong about my previous review comment: the ref test (and Firefox rendering) are correct. This is good to merge.

AmeliaBR (approved):

Ms2ger:
Thanks for following up!

PR #4325, opened 2016-12-14
EE: Subsumption Algorithm  -- Part 1

aubakirova:
Basic implementation tests for Subsumption Algorithm
Depends on https://github.com/w3c/web-platform-tests/pull/4310


content-security-policy/embedded-enforcement/subsumption_algorithm-general.html
```diff
@@ -0,0 +1,54 @@
...
+        "required_csp": "style-src 'none' 8 script-src 'unsafe-inline'", 
+        "returned_csp": "style-src 'none' 8 script-src 'unsafe-inline'", 
+        "expected": IframeLoad.EXPECT_LOAD },
+      { "name": "Iframe with more restricting CSP should load.", 
+        "required_csp": "script-src 'nonce-abc' 'nonce-123'", 
+        "returned_csp": "script-src 'nonce-abc'", 
+        "expected": IframeLoad.EXPECT_LOAD },
+      { "name": "Iframe with less restricting CSP should be blocked.", 
+        "required_csp": "style-src 'none' 8 script-src 'none'", 
+        "returned_csp": "style-src 'none' 8 script-src 'self'",
```
Sun77789: `8` instead of a semicolon. Otherwise, it was split into two parameters

content-security-policy/support/echo-policy-multiple.py
```diff
@@ -0,0 +1,38 @@
+def main(request, response):
+    headers = [("Content-Type", "text/html")]
+    if "policy" in request.GET:
+        headers.append(("Content-Security-Policy", request.GET["policy"].replace("8", ";")))
```
Sun77789: @mikewest  
It seems like GET parameters are split on a semicolon so it was not getting a full policy. This seems like a bad solution but I just used `8` as a special character and replaced it with a semicolon.

content-security-policy/support/echo-policy-multiple.py
```diff
@@ -0,0 +1,28 @@
...
+    return headers, '''
+<!DOCTYPE html>
+<html>
+<head>
+    <title>This page enforces embedder's policies</title>
+    <style>
+        body {
+            background-color: maroon;
+        }
+    </style>
```
Sun77789: If I remove this, then it seems like post message is not delivered on time for the check in `assert_iframe_with_csp`. So I have to either set a timeout or leave this here.

content-security-policy/embedded-enforcement/subsumption_algorithm-general.html
mikewest: That seems very strange. I'll look at this with you when I get back to my desk. We shouldn't need to do this...

Are you running `encodeURIComponent` on the string you're passing into the URL?

content-security-policy/support/echo-policy-multiple.py
mikewest: Funky. Set the timeout, I suppose.

mikewest (approved):
Thanks!

PR #4327, opened 2016-12-14
EE: Subsumption Algorithm -- Part 3

aubakirova:
Tests for keyword `self`.
@mikewest 
(Files other than `subsumption_algorithm-self.html` should be rebased on other PRs)

### DIFF
--- a/content-security-policy/embedded-enforcement/subsumption_algorithm-self.html
+++ b/content-security-policy/embedded-enforcement/subsumption_algorithm-self.html
@@ -1,0 +1,49 @@
+<!DOCTYPE html>
+<html>
+<head>
+<title>Embedded Enforcement: Subsumption Algorithm - 'self' keyword.</title>
+  <script src="/resources/testharness.js"></script>
+  <script src="/resources/testharnessreport.js"></script>
+  <script src="support/testharness-helper.sub.js"></script>
+</head>
+<body>
+  <script>
+    var tests = [
+      { "name": "'self' keywords should match.", 
+        "required_csp": "img-src 'self' http://b.com:*", 
+        "returned_csp": "img-src 'self' http://b.com:*",
+        "expected": IframeLoad.EXPECT_LOAD },
+      { "name": "Returned CSP does not have to specify 'self'.", 
+        "required_csp": "img-src 'self' http://b.com:*", 
+        "returned_csp": "img-src http://b.com:*",
+        "expected": IframeLoad.EXPECT_LOAD },
+      { "name": "Returned CSP must not allow 'self' if required CSP does not.", 
+        "required_csp": "img-src http://b.com:*", 
+        "returned_csp": "img-src 'self' http://b.com:*",
+        "expected": IframeLoad.EXPECT_BLOCK },
+      { "name": "Returned 'self' should match to an origin's url.", 
+        "required_csp": "img-src 'self' http://b.com:*", 
+        "returned_csp": "img-src " + getCrossOrigin(),
+        "expected": IframeLoad.EXPECT_LOAD },
+      { "name": "Required 'self' should match to a origin's url.", 
+        "required_csp": "img-src " +  getCrossOrigin() + " http://b.com:*", 
+        "returned_csp": "img-src 'self'",
+        "expected": IframeLoad.EXPECT_LOAD },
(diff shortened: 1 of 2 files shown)


content-security-policy/support/testharness-helper.sub.js
```diff
@@ -5,8 +5,30 @@ const Host = {
...
+  return url.toString();
+}
+
+function getCrossOrigin() {
+  var url = new URL("http://{{domains[天気の良い日]}}:{{ports[http][0]}}/");
+  return url.toString();
+}
+
+function getSecureCrossOrigin() {
+  var url = new URL("https://{{domains[天気の良い日]}}:{{ports[http][0]}}/");
```
Sun77789: Note the `http` port for `https` scheme. 
Otherwise, server specifies a specific port that is not necessarily a default port for `http` nor the first port of `https` is a default for `https`. So it happens that the ports do not match and subsumption is `false`.

mikewest: Can you add a comment explaining this? I never would have noticed it if you didn't point it out (and I'm still not sure I understand why it's necessary).

mikewest:
Also it looks like you need to rebase this on top of the other two patches.

content-security-policy/support/testharness-helper.sub.js
Sun77789: It seems like server is that such that ` {{ports[https][0]}}` is 8443, whereas ` {{ports[http][0]}}` is 8000.
So that `'self'` is http://xn--n8j6ds53lwwkrqhv28a.web-platform.test:8000/
and secure server https give `https://xn--n8j6ds53lwwkrqhv28a.web-platform.test:8443/`

I just wanted to add checks for 'self' that are not exact matches (for example, more secure scheme). But since ports differ and 8443 is not default for `https`, subsumption is false. So I wanted to use more secure scheme but the same port. For now I added a comment, but we could also just drop this test?

content-security-policy/support/testharness-helper.sub.js
```diff
@@ -20,6 +20,18 @@ function getOrigin() {
...
 }
 
+function getCrossOrigin() {
+  var url = new URL("http://{{domains[天気の良い日]}}:{{ports[http][0]}}/");
+  return url.toString();
+}
+
+function getSecureCrossOrigin() {
+  // Note that `http` port is used. This is to check subsumption of `self`,
+  // since ports are not wildcards and are not necessarily default ports.
```
mikewest: Can you add a little more detail, like:

"Since `wptserve` spins up servers on non-default port, `'self'` matches `http://[host]:[specified-port]` and `https://[host]:[specified-port]`, but not `https://[host]:[https-port]`. So, we use the `http` port for this `https` origin in order to verify that a secure variant of a non-secure URL matches `'self'`".

mikewest (approved):
LGTM % nit.

Sun77789:
Updated!
But I think it might have merge conflicts after https://github.com/w3c/web-platform-tests/pull/4310. So I will rebase it after the other one is merged.

content-security-policy/support/testharness-helper.sub.js
```diff
@@ -20,6 +20,20 @@ function getOrigin() {
...
+  var url = new URL("http://{{domains[天気の良い日]}}:{{ports[http][0]}}/");
+  return url.toString();
+}
+
+function getSecureCrossOrigin() {
+  // Since wptserve spins up servers on non-default port, 'self' matches
+  // http://[host]:[specified-port] and https://[host]:[specified-port], but not
+  // https://[host]:[https-port]. So, we use the http port for this https origin
+  // in order to verify that a secure variant of a non-secure URL matches 'self'.
+  var url = new URL("https://{{domains[天気の良い日]}}:{{ports[http][0]}}");
```
mikewest: I just realized that these helpers are in the `content-security-policy/support` file: let's move them to `content-security-policy/embedded-enforcement/support`. This `https` scheme/`http` port mismatch isn't going to be useful generally, and could cause confusion in the future if someone sees `getSecureCrossOrigin` and thinks "Hey, that sounds useful!"

mikewest (approved):
LGTM % locational nit.

PR #4328, opened 2016-12-14
Expand on tests for textarea.selectionStart/End/Direction

foolip:
Tests for https://github.com/whatwg/html/pull/2176


foolip:
This now fails everywhere because of reasons unrelated to the nullability. The most interesting failure seems to be with Safari, where there "updated selectionStart" assert fails. @rniwa @cdumez, would either of you care to review?

foolip:
Happy new year and ping @rniwa @cdumez :)

rniwa:
r=me.

PR #4330, opened 2016-12-14
Use promise_test instead of async_test

obstschale:
The tests use now `promise_test` instead of `async_test`. I also removed the `step_func`s.

The only thing I am not sure is whether to remove the `catch` blocks or not.

ping #4040

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/4330)

### DIFF
--- a/presentation-api/controlling-ua/PresentationConnection_onclosed-manual.html
+++ b/presentation-api/controlling-ua/PresentationConnection_onclosed-manual.html
@@ -38,12 +38,15 @@
 
           assert_true(connection instanceof PresentationConnection, 'the connection is setup');
 
-          connection.onclose = function (evt) {
+          const watcher = new EventWatcher(t, connection, 'close');
+
+          watcher.wait_for('close').then(function () {
+            connection.close();
+            return watcher.wait_for('close');
+          }).then(function (evt) {
             assert_equals(evt.type, "close");
             assert_equals(connection.state, "closed");
-          };
-
-          connection.close();
+          });
 
       });
 
--- a/presentation-api/controlling-ua/PresentationConnection_onclosed-manual.html
+++ b/presentation-api/controlling-ua/PresentationConnection_onclosed-manual.html
@@ -21,28 +21,36 @@
 <script>
   setup({explicit_timeout: true});
   var startPresentation = function () {
+
     document.getElementById('presentBtn').disabled = true;
-    async_test(function(t) {
+
+    promise_test(function (t) {
       var request = new PresentationRequest(presentationUrls);
-      request.start()
-        .then(function(connection) {
+
+      return request.start()
+        .then(function (connection) {
 
           // Enable timeout again, cause no user action is needed from here.
-          t.step_timeout(function() {
+          t.step_timeout(function () {
               t.force_timeout();
               t.done();
           }, 5000);
 
           assert_true(connection instanceof PresentationConnection, 'the connection is setup');
-          connection.onclose = t.step_func_done(function(evt) {
+
+          connection.onclose = function (evt) {
             assert_equals(evt.type, "close");
             assert_equals(connection.state, "closed");
-          });
+          };
+
           connection.close();
+
       })
-      .catch(function(ex) {
+      .catch(function (ex) {
         assert_unreached(ex.name + ":" + ex.message);
       });
+
     }, "the onclose is fired and the connection state is closed.");
+
   }
 </script>
--- a/presentation-api/controlling-ua/PresentationConnection_onclosed-manual.html
+++ b/presentation-api/controlling-ua/PresentationConnection_onclosed-manual.html
@@ -45,9 +45,6 @@
 
           connection.close();
 
-      })
-      .catch(function (ex) {
-        assert_unreached(ex.name + ":" + ex.message);
       });
 
     }, "the onclose is fired and the connection state is closed.");
--- a/presentation-api/controlling-ua/PresentationConnection_onterminated-manual.html
+++ b/presentation-api/controlling-ua/PresentationConnection_onterminated-manual.html
@@ -52,9 +52,6 @@
               assert_unreached("Wrong, the onclose shouldn't be triggered!");
           };
 
-      })
-      .catch(function (ex) {
-        assert_unreached(ex.name + ":" + ex.message);
       });
 
     }, "the onterminate is fired and the connection state is terminated");
--- a/presentation-api/controlling-ua/PresentationConnection_onterminated-manual.html
+++ b/presentation-api/controlling-ua/PresentationConnection_onterminated-manual.html
@@ -21,34 +21,43 @@
 <script>
   setup({explicit_timeout: true});
   var startPresentation = function () {
+
     document.getElementById('presentBtn').disabled = true;
-    async_test(function(t) {
+
+    promise_test(function (t) {
       var request = new PresentationRequest(presentationUrls);
-      request.start()
-        .then(function(connection) {
+
+      return request.start()
+        .then(function (connection) {
 
           // Enable timeout again, cause no user action is needed from here.
-          t.step_timeout(function() {
+          t.step_timeout(function () {
               t.force_timeout();
               t.done();
           }, 5000);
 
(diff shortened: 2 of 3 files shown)


presentation-api/controlling-ua/PresentationConnection_onclosed-manual.html
```diff
@@ -21,28 +21,36 @@
...
-          t.step_timeout(function() {
+          t.step_timeout(function () {
               t.force_timeout();
               t.done();
           }, 5000);
 
           assert_true(connection instanceof PresentationConnection, 'the connection is setup');
-          connection.onclose = t.step_func_done(function(evt) {
+
+          connection.onclose = function (evt) {
```
mkruisselbrink: If you want the test to wait for onclose to be called, you'll need to return some promise that doesn't resolve until onclose is called. EventWatcher might help with that, something like (untested):
```js
const w = new EventWatcher(t, connection, 'close');
const p = w.wait_for('close');
connection.close();
return p;
}.then(function (evt) {
  // assertions
});

File: presentation-api/controlling-ua/PresentationConnection_onclosed-manual.html [hunk 2]
```diff
@@ -21,28 +21,36 @@
...
             assert_equals(evt.type, "close");
             assert_equals(connection.state, "closed");
-          });
+          };
+
           connection.close();
+
       })
-      .catch(function(ex) {
+      .catch(function (ex) {
```
mkruisselbrink: You don't need this catch in a promise_test. The promise_test will fail if the promise passed to it rejects.

mkruisselbrink (changes requested):

presentation-api/controlling-ua/PresentationConnection_onclosed-manual.html
obstschale: That's what I thought. I will remove the `catch` blocks

presentation-api/controlling-ua/PresentationConnection_onclosed-manual.html
obstschale: I added an EventWatcher in `PresentationConnection_onclosed-manual.html`

obstschale (commented):
Changed files like requested

sideshowbarker:
@mkruisselbrink is this ready to land?

presentation-api/controlling-ua/PresentationConnection_onclosed-manual.html
```diff
@@ -21,28 +21,36 @@
...
           }, 5000);
 
           assert_true(connection instanceof PresentationConnection, 'the connection is setup');
-          connection.onclose = t.step_func_done(function(evt) {
+
+          const watcher = new EventWatcher(t, connection, 'close');
+
+          watcher.wait_for('close').then(function () {
+            connection.close();
+            return watcher.wait_for('close');
```
mkruisselbrink: I don't know the spec this is testing, but it seems a bit odd to me that in the new test we're waiting for two close events, while in the old test you only waited for one close event?

presentation-api/controlling-ua/PresentationConnection_onterminated-manual.html
```diff
@@ -21,34 +21,40 @@
...
-          t.step_timeout(function() {
+          t.step_timeout(function () {
               t.force_timeout();
               t.done();
           }, 5000);
 
           assert_true(connection instanceof PresentationConnection);
-          connection.onconnect = t.step_func(function(evt) {
+
+          connection.onconnect = function (evt) {
```
mkruisselbrink: for asserts to work correctly in functions that aren't part of the promise chain (such as this event handler) you still need to wrap them in t.step_func

presentation-api/controlling-ua/PresentationConnection_onterminated-manual.html
```diff
@@ -21,34 +21,40 @@
...
             assert_equals(connection.state, "connected");
             connection.terminate();
-          });
-          connection.onterminate = t.step_func_done(function(evt) {
+          };
+
+          connection.onterminate = function (evt) {
```
mkruisselbrink: presumably you'll want to use EventWatcher or something similar to wait for onterminate to actually be be called?

presentation-api/controlling-ua/PresentationConnection_onterminated-manual.html
```diff
@@ -21,34 +21,40 @@
...
             assert_equals(evt.type, "terminate");
             assert_equals(connection.state, "terminated");
-          });
-          connection.onclose = t.step_func_done(function(evt) {
+          };
+
+          connection.onclose = function (evt) {
```
mkruisselbrink: nit: could just use `connection.onclose = t.unreached_func()`. Or if you use EventWatcher, add the close event to the list of events being watched, and it'll cause a failure if the event happens before the events you want to get.

mkruisselbrink:
> @mkruisselbrink is this ready to land?

somebody that actually knows something about this spec should probably also have a look? I've just been looking at correctness with how promise_test is used.

sideshowbarker:
> somebody that actually knows something about this spec should probably also have a look?

fair enough

@tidoust, back to you?

presentation-api/controlling-ua/PresentationConnection_onclosed-manual.html
obstschale: I did this because the API uses a similar example. See [here](https://github.com/w3c/testharness.js/blob/master/docs/api.md#promise-tests). There also the watcher is returned with all events to watch for.

presentation-api/controlling-ua/PresentationConnection_onterminated-manual.html
obstschale: Are you sure? @tidoust explained it in #4040 like this. In his example he left out the `step_func`.

presentation-api/controlling-ua/PresentationConnection_onterminated-manual.html
tidoust: Right. Which should boil down to something like:

```js
let connection = null;
return request.start()
  .then(function (c) {
    connection = c;
    [...]
    const watcher = new EventWatcher(t, connection, 'connect');
    return watcher.wait_for('connect');
  })
  .then(function () {
    connection.terminate();
    const watcher = new EventWatcher(t, connection, ['terminate', 'close']);
    return watcher.wait_for('terminate');
  })
  .then(function () {
    assert_equals(connection.state, 'terminated');
  });
```

or you could perhaps use the same `EventWatcher` as in:

```js
let connection = null;
let watcher = null;
return request.start()
  .then(function (c) {
    connection = c;
    [...]
    watcher = new EventWatcher(t, connection, ['connect', 'close', 'terminate']);
    return watcher.wait_for('connect');
  })
  .then(function () {
    connection.terminate();
    return watcher.wait_for('terminate');
  })
  .then(function () {
    assert_equals(connection.state, 'terminated');
  });
```

tidoust (changes requested):

tomoyukilabs:
BTW, can we use ES6 syntax to write codes of test cases? As far as I have seen before, all existing test cases for Presentation API seem to be written in ES5.

gsnedders:
@tomoyukilabs Yes; the requirement is that they must run in the latest stable release of each major browser (Chrome, Edge, Firefox, Safari), and with Safari 10 having shipped ES6 syntax is fine.

tidoust:
Note test files also now need to be renamed to end with `.https.html` now that the Presentation API was restricted to secure contexts.

@obstschale Are you still working on this?

tomoyukilabs:
I would suggest that I could take this work over if this PR would be merged.

tidoust:
You're more than welcome to take this work over @tomoyukilabs. Instead of merging this PR and then work on a follow-up PR to fix remaining issues, could you retrieve the code of that PR and prepare another one?

tomoyukilabs:
@tidoust okay, I'll submit another PR which includes all updates in that PR.

tidoust:
Pull request superseded by PR #5794, now merged.

PR #4332, opened 2016-12-15
Test that prefixed Fullscreen API is not supported

zcorpan:


upsuper:
I don't understand why do we want this?

zcorpan:
@upsuper it's good to test that APIs we do *not* want in the Web platform are in fact not supported. This is applying some pressure on Chromium to unprefix and remove the prefixed names, so other engines don't end up having to implement the prefixed names for web compat.

fullscreen/api/historical.html
```diff
@@ -0,0 +1,16 @@
+<!doctype html>
+<title>Historical Fullscreen features</title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<body>
```
foolip: Not minimal enough! ;)

zcorpan: Hah I think I first wanted to check against `document.body` but then didn't.

foolip:
I am OK with having failing tests for our prefixed APIs until we make the effort to standardize them. The full list of stuff is:
* `document.onwebkitfullscreenchange`
* `document.onwebkitfullscreenerror`
* `document.webkitCurrentFullScreenElement`
* `document.webkitCurrentFullScreenElement`
* `document.webkitExitFullscreen`
* `document.webkitFullscreenElement`
* `document.webkitFullscreenEnabled`
* `document.webkitIsFullScreen`
* `element.onwebkitfullscreenchange`
* `element.onwebkitfullscreenerror`
* `element.webkitRequestFullScreen`
* `element.webkitRequestFullscreen`
* `video.webkitDisplayingFullscreen`
* `video.webkitEnterFullScreen`
* `video.webkitEnterFullscreen`
* `video.webkitExitFullScreen`
* `video.webkitExitFullscreen`
* `video.webkitSupportsFullscreen`

Awesome, huh?

upsuper:
Then do you want to also add moz-prefixed ones and probably mz-prefixed ones to the list as well?

I suppose dropping all the prefixed API would actually hurt web compatibility for a while... but if that's the plan for all browsers, I guess I'm fine with that. (Actually if all browsers can coordinate to drop prefixed API together, I would be happy.)

foolip:
I'm not at all assuming that we'll be successful in removing the prefixed APIs, only after successfully shipping the unprefixed API will that be possible to get some data on I think. But a failing test to remind us that it's on the TODO list of things to sort out is OK I think. It'll also reveal that Edge in fact implements some or all of these.

zcorpan:
Now tested all of the above with webkit/moz/ms prefixes.

fullscreen/api/historical.html
```diff
@@ -0,0 +1,35 @@
...
+  "webkitExitFullscreen",
+  "webkitSupportsFullscreen",
+].forEach(function(member) {
+  ["webkit", "moz", "ms"].forEach(function(prefix) {
+    var alias = member.replace("webkit", prefix);
+    var clarifyTestName = (alias.indexOf('FullScreen') != -1 ? ' (uppercase S)' : '');
+
+    test(function() {
+      assert_false(alias in document.createElement('video'));
+    }, '<video> member must be nuked: ' + alias + clarifyTestName);
```
foolip: This is efficient and OK with me, but a comment somewhere saying that lots of the things that should be nuked have never existed would be nice.

foolip (approved):

zcorpan:
https://bugs.chromium.org/p/chromium/issues/detail?id=383813
https://bugzilla.mozilla.org/show_bug.cgi?id=1249225
https://bugs.webkit.org/show_bug.cgi?id=158125
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8438881/

PR #4334, opened 2016-12-15
Test active document check for requestFullscreen

zcorpan:
This is cherry-picked from https://github.com/w3c/web-platform-tests/pull/4309


zcorpan:
https://github.com/whatwg/html/pull/2160

upsuper (approved):

zcorpan:
https://github.com/whatwg/fullscreen/issues/67 points out that the queued event can't really be fired in an inactive document, so this *should* time out. Need to figure out what correct behavior should be and test that.

Should add:
* Test the promise.
* Test navigating back so the document becomes active again, does the event fire at that point?

fullscreen/api/active-document-manual.html
```diff
@@ -0,0 +1,18 @@
...
+<div id="log"></div>
+<iframe allowfullscreen></iframe>
+<script>
+async_test((t) => {
+  onload = () => {
+    var iframes = document.getElementsByTagName("iframe");
+    trusted_request(iframes[0].contentDocument.body, document.body);
+    window[0].location.href = '/common/blank.html';
+    iframes[0].contentDocument.onfullscreenchange = t.unreached_func("fullscreenchange event");
+    iframes[0].contentDocument.onfullscreenerror = t.step_func_done();
```
upsuper: Oh, wait. This is wrong. `iframes[0].contentDocument` is a different than the previous document you request fullscreen on, so these events are never triggered.

You need to record the previous document and set event handlers on that.

Also, shouldn't it easier to just check `document.fullscreenEnabled`?

upsuper (changes requested):

upsuper:
I think you can simply check `documentBeforeNav.fullscreenEnabled` is `false`.

zcorpan:
OK, yeah that seems reasonable. It may be worth testing events and the promise as well, so the behavior there is interoperable.

upsuper:
In that case, I suggest you split the events test out, because that test would still need manual action, which is not great.

foolip (approved):

foolip:
Options nit to make file name and test title derivative of document-fullscreen-enabled.html, otherwise OK to merge.

PR #4335, opened 2016-12-15
Create a webvr directory for layout tests

LoonyBean:


foolip:
@toji @kenrussell are you OK with this? Anyone else who should be in the OWNERS file? Ideally people from all engines are involved.

kenrussell:
I'm not the best person to be an owner of this directory -- my involvement with WebVR is peripheral. Brandon's out for the rest of the year, so not sure who would be a good reviewer right now. I'll ask one of Brandon's teammates to review this and suggest.

klausw:
I'm working with Brandon on WebVR implementation in Chrome, and I volunteer as owner to help cover during Brandon's leave. Note that I'm still pretty new to WebVR and not familiar with how specs testing works, so I'd prefer sticking to simple and obviously correct tests at this point. Is there someone willing to help with reviews to ensure that they follow best practices for such tests, independent of WebVR specifics?

foolip:
I've written and reviewed quite a bit in web-platform-tests, and can help take a look when needed. I would prefer to not be in OWNERS though, to only be pinged when actually needed :)

foolip:
The background for this PR is that @LoonyBean was going to write a simple test for some bit of WebVR, maybe an event constructor, and I suggested doing it inside Chromium and exporting using the process in https://crbug.com/657117, but then the directory has to exist first.

Gecko already has a similar process. In other words, don't be alarmed if "export" commits show up that get merged without review, they should always have been reviewed elsewhere, and if further changes are needed everyone can just get together on GitHub to discuss.

(If you could find people from other engines working on WebVR interested in sharing tests, that'd be great too.)

LoonyBean:
Updated OWNERS file

PR #4337, opened 2016-12-15
Update some Fullscreen test titles for consistency

foolip:


foolip:
Merging without review because I don't imagine anyone much cares about the titles except me.

PR #4343, opened 2016-12-15
Add IDL test for InputEvent

chong-z:
This is the test for partial `InputEvent` IDL only, additional tests should be added later for `StaticRange` and the other part of `InputEvent` interface in `UIEvent`.

Testing results:
  * Passed on Safari Technology Preview Release 18 and Chrome Canary 57 (with `InputEvent` enabled).

References:
1. InputEvent spec: https://w3c.github.io/input-events/#h-interface-inputevent
2. InputEvent in UIEvent spec: https://w3c.github.io/uievents/#interface-inputevent

### DIFF
--- a/input-events/idlharness.html
+++ b/input-events/idlharness.html
@@ -1,0 +1,38 @@
+<title>Input Event IDL tests</title>
+<link rel="help" href="https://w3c.github.io/input-events/#h-interface-inputevent"/>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/resources/WebIDLParser.js"></script>
+<script src="/resources/idlharness.js"></script>
+
+<pre id="untested_idl">
+interface InputEvent {
+};
+</pre>
+
+<pre id='idl'>
+partial interface InputEvent {
+    readonly attribute DOMString     inputType;
+    readonly attribute DataTransfer? dataTransfer;
+    sequence<StaticRange> getTargetRanges();
+};
+
+partial dictionary InputEventInit {
+    DOMString             inputType = "";
+    DataTransfer?         dataTransfer = null;
+    sequence<StaticRange> targetRanges = [];
+};
+</pre>
+
+<script>
+(function(){
+    "use strict";
+    const idl_array = new IdlArray();
+    idl_array.add_untested_idls(document.getElementById("untested_idl").textContent);
+    idl_array.add_idls(document.getElementById("idl").textContent);
+    idl_array.add_objects({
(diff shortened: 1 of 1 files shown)


input-events/idlharness.html
```diff
@@ -0,0 +1,40 @@
...
+interface UIEvent {
+};
+
+interface StaticRange {
+};
+</pre>
+
+<pre id='idl'>
+[Constructor(DOMString type, optional InputEventInit eventInitDict)]
+interface InputEvent : UIEvent {
```
foolip: This doesn't match https://w3c.github.io/uievents/#events-inputevents

In order to make it easy to keep these tests up to date, it'd be good to copy spec IDL verbatim, and link to where it was copied from.

choniong: Thanks for the review!

I've added a link to https://w3c.github.io/input-events/#h-interface-inputevent and removed the UIEvent portion. Updated test will pass on both Safari and Chrome.

Should we test https://w3c.github.io/uievents/#events-inputevents inside "uievents/interface/"?

foolip: Yeah, testing the main interface under uievents/interface sounds right.

Note that idlharness.js doesn't actually try invoking any constructors, so one still has to do things like `assert_true(new Event('type', { bubbles: true}).bubbles)` as separate tests, to ensure that the dictionaries are actually hooked up correctly.

choniong: > Yeah, testing the main interface under uievents/interface sounds right.

Can I leave it to another patch using the two-way sync process?

> Note that idlharness.js doesn't actually try invoking any constructors, so one still has to do things like assert_true(new Event('type', { bubbles: true}).bubbles) as separate tests, to ensure that the dictionaries are actually hooked up correctly.

Will add constructor tests using the two-way sync process.

Regarding this PR, do you want me to squash the commits?

choniong:
I've updated 1. PR description, 2. IDL with latest `InputEvent` interface and 3. squashed into 1 commit. 
Please take a look again, thanks!

input-events/idlharness.html
foolip: Sure, I'd love for you to write some tests in Chromium for exporting, so we can uncover ever more bugs before a full launch.

I like squashed commits, but if I'm the one merging a PR I'll just use the "Squash and merge" button and tweak the message to my liking, so it doesn't matter much.

choniong: Thanks for merging! 

> I like squashed commits, but if I'm the one merging a PR I'll just use the "Squash and merge" button and tweak the message to my liking, so it doesn't matter much.

Sweet, never know that before.

PR #4347, opened 2016-12-16
EE: Splitting Host Source tests

aubakirova:
I realized the other file for host source expressions was too big so some tests were sometimes timing out before getting a post message. 
I split the tests and added more cases for scheme checks.


Sun77789:
@mikewest

PR #4350, opened 2016-12-16
Test document.exitFullscreen() for an inactive document

foolip:
Test for https://github.com/whatwg/fullscreen/pull/68

### DIFF
--- a/fullscreen/api/document-exit-fullscreen-active-document.html
+++ b/fullscreen/api/document-exit-fullscreen-active-document.html
@@ -1,0 +1,26 @@
+<!DOCTYPE html>
+<title>Document#exitFullscreen() when the document is not the active document</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div id="log"></div>
+<iframe allowfullscreen></iframe>
+<script>
+var t = async_test();
+
+onload = t.step_func(() => {
+  var iframe = document.querySelector("iframe");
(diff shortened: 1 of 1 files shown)


fullscreen/api/document-exit-fullscreen-active-document.html
```diff
@@ -0,0 +1,26 @@
...
+<script>
+var t = async_test();
+
+onload = t.step_func(() => {
+  var iframe = document.querySelector("iframe");
+  var documentBeforeNav = iframe.contentDocument;
+
+  iframe.onload = t.step_func(() => {
+    var p = documentBeforeNav.exitFullscreen();
+    assert_true(p instanceof Promise, 'exitFullscreen() returns promise');
```
zcorpan: The message is shown when the assert fails, and this message says the opposite of what happened when it fails. Maybe 'exitFullscreen() should return a promise'

zcorpan (approved):
LGTM % nit

fullscreen/api/document-exit-fullscreen-active-document.html
foolip: The failure message is "assert_true: exitFullscreen() returns promise expected true got false", isn't that appropriate? I've come to to just describe the expected value for assert_equals and the statement that is true of assert_true, based on what the errors end up looking like.

zcorpan: Hmm I guess it is indeed clear. OK.

PR #4354, opened 2016-12-16
Test snapshotting of allowfullscreen attribute

zcorpan:
Test for https://github.com/whatwg/html/pull/2187

Edit: this ought to follow https://github.com/whatwg/html/pull/3287 but I don't know right now (2020-01-27) whether it does.


fullscreen/api/document-fullscreen-enabled-removing-allowfullscreen.sub.html
```diff
@@ -0,0 +1,36 @@
...
+<script src="/resources/testharnessreport.js"></script>
+<div id="log"></div>
+<script>
+async_test((t) => {
+  const iframe = document.createElement('iframe');
+  iframe.allowFullscreen = true;
+
+  let i = 0;
+
+  iframe.src = "http://{{domains[www1]}}:{{ports[http][0]}}/fullscreen/api/echo-fullscreenEnabled.html";
```
foolip: I don't recognize this syntax, is this using a wptserve feature, to make it a cross-origin iframe?

foolip (approved):
Seems to be testing what it intends to. Can you see who passes and fails?

fullscreen/api/document-fullscreen-enabled-removing-allowfullscreen.sub.html
zcorpan: Yes.

zcorpan:
Gecko and Edge 13 fail because of live behavior.

Chromium and WebKit have prefixed names. Testing with `webkitFullscreenEnabled` instead they still fail because of live behavior.

zcorpan:
Should also test the timing of when it's snapshotted; at document creation time vs completed response headers.

foolip:
@zcorpan can you rebase to trigger new CI runs?

zcorpan:
Rebased.

zcorpan:
I rebased again to rerun CI checks.

gsnedders:
@zcorpan 

```
ERROR:lint:fullscreen/api/document-fullscreen-enabled-setting-allowfullscreen-timing.sub.html:26: setTimeout used (SET TIMEOUT)
```

zcorpan:
Lint error was fixed, but CI still failed. The logs were no longer available. Rebasing again.

https://github.com/whatwg/html/pull/3287 replaced https://github.com/whatwg/html/pull/2187 . @clelland , do these tests match what the spec now requires?

fullscreen/api/echo-fullscreenEnabled.html
```diff
@@ -0,0 +1,11 @@
+<!doctype html>
+<script>
+window.onmessage = (e) => {
+  if (e.data === 'What is document.fullscreenEnabled?') {
```
foolip, line 4: I'll move this into resources/

fullscreen/api/document-fullscreen-enabled-setting-allowfullscreen-timing.sub.html
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
+<title>Document#fullscreenEnabled setting allowfullscreen after document creation, before response</title>
```
foolip, line 2: Nice and evil test, I like it!

foolip (approved):

foolip:
The spec for this ended up as https://html.spec.whatwg.org/multipage/iframe-embed-object.html#allowed-to-use. The warning there summarizes it well:
> Because they only influence the permissions policy of the nested browsing context's active document, the allow and allowfullscreen attributes only take effect when the nested browsing context of the iframe is navigated. Adding or removing them has no effect on an already-loaded document.

zcorpan:
Thanks @foolip!

PR #4356, opened 2016-12-16
Images are always sniffed

annevk:
See https://github.com/whatwg/fetch/issues/395 for details.


annevk:
@mikewest can maybe review these as well.

zcorpan (approved):

PR #4358, opened 2016-12-16
Pointer Events: Add IDL test

RByers:
Add an [idlharness test](testthewebforward.org/docs/testharness-idlharness.html) for Pointer Events.  This is a little redundant (but more precise) with some of our other basic interface tests, but I think that's fine for now (we've got some plans to clean them up anyway).  We'll have to keep the IDL here in sync with spec changes, but I think it's worth the effort - gives us a bunch of test coverage pretty cheaply.  Thoughts?

There are a few test failures in Chrome (tested 56.0.2924.18 Linux) today:
1. Missing ongotpointercapture/onlostpointercapture on Window and Document: [recent spec change](https://github.com/w3c/pointerevents/issues/157), @NavidZ will land a change in blink soon
2. Other Window on... attributes reporting value `undefined` instead of having a getter function: A long-standing [blink bindings bug](https://bugs.chromium.org/p/chromium/issues/detail?id=475556).  But this has just been (mostly) fixed in Chrome 57.
3. class string of PointerEvent.prototype expected "[object PointerEventPrototype]" but got "[object PointerEvent]":  A long standing [blink](https://bugs.chromium.org/p/chromium/issues/detail?id=643712) / [WebIDL spec](https://www.w3.org/Bugs/Public/show_bug.cgi?id=28244) disagreement.  I'll try to push some progress on this.
4. Missing  `twist` and `tangentialPressure` properties: tracked [here](https://bugs.chromium.org/p/chromium/issues/detail?id=649376)

@NavidZ PTAL

### DIFF
--- a/pointerevents/idlharness.html
+++ b/pointerevents/idlharness.html
@@ -1,0 +1,104 @@
+<!doctype html>
+<meta charset=utf-8>
+<title>idlharness test</title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<script src="/resources/WebIDLParser.js"></script>
+<script src="/resources/idlharness.js"></script>
+
+<pre id='untested_idl' style='display:none'>
+[PrimaryGlobal]
+interface Window {
+};
+
+[TreatNonObjectAsNull]
+callback EventHandlerNonNull = any (Event event);
+typedef EventHandlerNonNull? EventHandler;
+
+[NoInterfaceObject]
+interface GlobalEventHandlers {
+};
+Window implements GlobalEventHandlers;
+
+interface Navigator {
+};
+
+interface Element {
+};
+
+interface HTMLElement : Element {
+};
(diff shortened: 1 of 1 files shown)


NavidZ:
Would this make this [other test](https://github.com/w3c/web-platform-tests/pull/4357) redundant?

RByers:
> Would this make this other test redundant?

Not completely, no.  This doesn't test that the properties actually install listeners that get invoked, just tests the exposed API surface area shape.  Left a comment there suggesting a simplification.

NavidZ:
lgtm. I also verified that all the failures due to the first reason are resolved after testing it locally with the CL.

RByers:
Thanks.  It's kinda too bad that tests and specs are in separate repos, I'd really rather just have the test pull the IDL from the spec than having to keep copy/paste it, but hopefully it's not that much of a hassle...

PR #4359, opened 2016-12-16
The value of the srclang and language attributes must be a valid BCP 47 language tag.

johnwilander:
The srclang attribute: https://html.spec.whatwg.org/multipage/embedded-content.html#attr-track-srclang. The language attribute is a BCP 47 language tag representing the language of the text track's cues: https://html.spec.whatwg.org/multipage/embedded-content.html#text-track-language. BCP 47 does not allow null characters in the language tag: https://tools.ietf.org/html/bcp47#section-2.1.

### DIFF
--- a/html/semantics/embedded-content/media-elements/interfaces/TextTrack/language.html
+++ b/html/semantics/embedded-content/media-elements/interfaces/TextTrack/language.html
@@ -21,9 +21,9 @@ test(function(){
     assert_equals(t2.language, '');
 });
 test(function(){
-    track.srclang = '\u0000a';
-    assert_equals(t2.language, '\u0000a', 'IDL attribute');
-    track.setAttribute('srclang', '\u0000b');
-    assert_equals(t2.language, '\u0000b', 'content attribute');
-}, document.title+', \\u0000');
+    track.srclang = 'zh-Hant-CN-x-private1-private2';
+    assert_equals(t2.language, 'zh-Hant-CN-x-private1-private2', 'IDL attribute');
+    track.setAttribute('srclang', 'i-klingon');
+    assert_equals(t2.language, 'i-klingon', 'content attribute');
+}, document.title+', BCP47 language tag');
 </script>


zcorpan:
That is an authoring requirement, it does not affect user agents. The test is correct as-is.

johnwilander:
The page author sets srclang to something not compliant with BCP 47 and the test checks that the language attribute contains that non-compliant value. Then how can user agents make sure that a read of language returns a valid BCP 47 language tag?

cdumez:
@zcorpan I agree with you that setting srclang to a non BCP-47 value and then querying it back will return the same value (because this is simple reflection).

However, how about TextTrack.language [1]? It says the user agent should return a BCP 47 language tag. Why is it OK for the user-agent to return something that is not BCP 47?

Note that the web-platform-test is reading TextTrack.language, not srclang.

[1] https://html.spec.whatwg.org/multipage/embedded-content.html#text-track-language

zcorpan:
https://html.spec.whatwg.org/multipage/embedded-content.html#sourcing-out-of-band-text-tracks

> The text track language is the element's track language, if any, or the empty string otherwise.

https://html.spec.whatwg.org/multipage/embedded-content.html#track-language

> If the element has a srclang attribute whose value is not the empty string, then the element's track language is the value of the attribute. Otherwise, the element has no track language.

cdumez:
Based on what @zcorpan is quoting, I now also believe the test is right after all.

johnwilander:
> If the element has a srclang attribute whose value is not the empty string, then the element's track language is the value of the attribute. Otherwise, the element has no track language.

The paragraph above says:

> The srclang attribute gives the language of the text track data. The value must be a valid BCP 47 language tag.

That's very confusing to me.

zcorpan:
That HTML has different rules for UAs and for authors is known to cause confusion. Does it help to read https://html.spec.whatwg.org/multipage/infrastructure.html#conformance-classes ?

I think the spec for text track language can be improved to be clearer though. File an issue? ☺

PR #4362, opened 2016-12-19
Add a fetch test for upload streaming

yutakahirano:
Adding a test for uploading a request made from a ReadableStream.

### DIFF
--- a/fetch/api/basic/request-upload.js
+++ b/fetch/api/basic/request-upload.js
@@ -31,5 +31,10 @@ testUpload("Fetch with POST with Float32Array body", url, "POST", new Float32Arr
 testUpload("Fetch with POST with Float64Array body", url, "POST", new Float64Array(1), "\0\0\0\0\0\0\0\0");
 testUpload("Fetch with POST with DataView body", url, "POST", new DataView(new ArrayBuffer(8), 0, 4), "\0\0\0\0");
 testUpload("Fetch with POST with Blob body with mime type", url, "POST", new Blob(["Test"], { type: "text/maybe" }), "Test");
+testUpload("Fetch with POST with ReadableStream", url, "POST", new ReadableStream(controller => {
+    controller.enqueue("Te");
+    controller.enqueue("st");
+    controller.close();
+  }), "Test");
 
 done();
--- a/fetch/api/basic/request-upload.js
+++ b/fetch/api/basic/request-upload.js
@@ -18,6 +18,20 @@ function testUpload(desc, url, method, body, expectedBody) {
   }, desc);
 }
 
+function testUploadFailure(desc, url, method, body) {
+  const requestInit = {"method": method};
+  promise_test(test => {
+    if (typeof body === "function")
+      body = body();
+    if (body)
+      requestInit["body"] = body;
+    return fetch(url, requestInit).then(() => {
+      assert_unreached("fetch should not succeeed");
+    }, () => {
+    });
+  }, desc);
+}
+
 var url = RESOURCES_DIR + "echo-content.py"
 
 testUpload("Fetch with PUT with body", url, "PUT", "Request's body", "Request's body");
@@ -36,5 +50,17 @@ testUpload("Fetch with POST with ReadableStream", url, "POST", new ReadableStrea
     controller.enqueue(encoder.encode("Test"));
     controller.close();
   }}), "Test");
+testUploadFailure("Fetch with POST with ReadableStream containing String", url, "POST", new ReadableStream({start: controller => {
+    controller.enqueue("Test");
+    controller.close();
+  }}));
+testUploadFailure("Fetch with POST with ReadableStream containing null", url, "POST", new ReadableStream({start: controller => {
+    controller.enqueue(null);
+    controller.close();
+  }}));
+testUploadFailure("Fetch with POST with ReadableStream containing number", url, "POST", new ReadableStream({start: controller => {
+    controller.enqueue(99);
+    controller.close();
+  }}));
 
 done();
--- a/fetch/api/basic/request-upload.js
+++ b/fetch/api/basic/request-upload.js
@@ -31,10 +31,10 @@ testUpload("Fetch with POST with Float32Array body", url, "POST", new Float32Arr
 testUpload("Fetch with POST with Float64Array body", url, "POST", new Float64Array(1), "\0\0\0\0\0\0\0\0");
 testUpload("Fetch with POST with DataView body", url, "POST", new DataView(new ArrayBuffer(8), 0, 4), "\0\0\0\0");
 testUpload("Fetch with POST with Blob body with mime type", url, "POST", new Blob(["Test"], { type: "text/maybe" }), "Test");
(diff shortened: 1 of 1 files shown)


yutakahirano:
@annevk: This is for https://github.com/whatwg/fetch/pull/425

annevk:
This looks okay to me. Though ideally we also have some tests around redirect handling and different kinds of methods. For a new feature that will need tests written anyway as part of the implementation this might be good enough though.

@foolip @domenic thoughts?

foolip:
Yeah, testing non-trivials things blindly is very hard, and a good reason to not write all tests up front. Keeping track somewhere of what testing coverage is missing would be nice of course, but only if someone is then using that to guide testing priorities.

domenic:
Wait, we're allowing strings to be enqueued? I didn't get that from https://github.com/whatwg/fetch/pull/425 .

annevk:
That is a good point and needs to be fixed in the test.

Where do we currently place restrictions on what can be on a `ReadableStream` object used in Fetch? It seems like we don't...

yutakahirano:
> Wait, we're allowing strings to be enqueued? I didn't get that from whatwg/fetch#425 .

Sorry that was my mistake. Fixed.

> Where do we currently place restrictions on what can be on a ReadableStream object used in Fetch? It seems like we don't...

At least, Request.text(), etc. expect that as they are using https://fetch.spec.whatwg.org/#concept-read-all-bytes-from-readablestream.

annevk:
We should also make the network layer expect that though. And whatever ends up reading a response.

yutakahirano:
At https://fetch.spec.whatwg.org/#read-a-request?

annevk:
Yeah I suppose. And we probably need something similar for responses then and make sure all specifications use it… 😟

guest271314:
@annevk Following https://www.w3.org/2015/05/Web_Platform_Testing_Latest_News.pdf, is the current test able to be run in browser at http://w3c-test.org/fetch/api/ ?

annevk:
I'm not sure what you're asking for.

guest271314:
@annevk Tried to run the current test in browser. Checked all fetch tests, browser crashed at a worker test. Just want to run the current test. How to do this at *nix, chromium 52?

annevk:
What current test? This is a pull request for a test change. It hasn't landed...

guest271314:
@annevk Ok. Was not certain about that. Not well-versed in the terminology and mechanics of git, pull requests. Is it not currently possible to try this now? Is a "pull request" code written that is tested before being published at large?

annevk:
You can locally check out the code and run it https://github.com/w3c/web-platform-tests/blob/master/README.md. I believe for some PRs there's also a URL you can visit, but not for this PR it seems (might depend on where the branch is situated).

guest271314:
@annevk Is the above referenced test in that repository? Do not need to run all of the tests in the repo, just the newest one, or only the `fetch` tests. Is it necessary to download the entire repo? And install the python server? Was there changes made to the way `fetch` `body` functions in order to pass the test?

annevk:
The test of this PR is in https://github.com/yutakahirano/web-platform-tests/tree/upload-streaming, a branch in a fork of this repository. If you just want to know whether any browser passes the test, I can tell you that none do, since as I told you elsewhere, this is not implemented yet.

guest271314:
@annevk Actually seeking more knowledge than just to know if currently implemented; also how implemented at tests? Or, that is, how is the test run if not implemented? The author(s) composed code which is being tested? A custom browser build implementing the changes necessary to pass the test? Gracious for your patience here. Again, just attempting to learn. If this exchange is not helpful for the topic, here, can stop now and try to gather pieces of information to possibly answer questions, at large.

guest271314:
@annevk Am attempting to gain more knowledge of the process. Your feedback has been helpful.

guest271314:
@annevk  Is the appropriate venue to ask questions about this implementation https://github.com/whatwg/fetch/pull/425 ?

annevk:
@yutakahirano perhaps you can also add a test that expects strings and numbers and such to fail.

annevk:
Thanks, @domenic could you give it another review?

fetch/api/basic/request-upload.js
```diff
@@ -18,6 +18,20 @@ function testUpload(desc, url, method, body, expectedBody) {
...
 }
 
+function testUploadFailure(desc, url, method, body) {
+  const requestInit = {"method": method};
+  promise_test(test => {
+    if (typeof body === "function")
+      body = body();
+    if (body)
+      requestInit["body"] = body;
+    return fetch(url, requestInit).then(() => {
```
domenic: This should use `return promise_rejects(new TypeError(), fetch(url, requestInit))`

fetch/api/basic/request-upload.js
```diff
@@ -31,5 +45,22 @@ testUpload("Fetch with POST with Float32Array body", url, "POST", new Float32Arr
...
+    controller.close();
+  }}));
+testUploadFailure("Fetch with POST with ReadableStream containing null", url, "POST", new ReadableStream({start: controller => {
+    controller.enqueue(null);
+    controller.close();
+  }}));
+testUploadFailure("Fetch with POST with ReadableStream containing number", url, "POST", new ReadableStream({start: controller => {
+    controller.enqueue(99);
+    controller.close();
+  }}));
```
domenic: We should also test ArrayBuffer and Blob are not allowed

domenic (changes requested):
Some minor test nits. Additionally I have something I am confused about; forgive me if we have discussed it before.

These tests seem to imply that fetch() will read the whole request body stream before fulfilling or rejecting the fetch promise. But, doesn't that hurt our ability to do full-duplex communication with streams? I thought the idea was that you could pass a request body stream, and get back a response body stream, and use them both at the same time. If we wait for the request body stream to be fully read before vending the response body stream, this is not possible, right?

annevk:
I agree with your assessment of how things should be with regards to full-duplex communication. These tests are simply not testing that scenario, or am I missing something?

annevk:
I would expect that the default server setup is such that it doesn't emit a response before getting the full request body. It would be good to have tests for early transmitted responses though.

domenic:
I see. So is the intent of the spec that there's a race then? If the fetch gets terminated before the promise settles, then the promise is rejected (as in this case); if the fetch gets terminated after the response comes back, the promise is still fulfilled?

annevk:
Yeah, the fetch gets terminated so you might not get the whole response body. We should really figure out a way to write full duplex tests since I suspect it is all kinds of broken.

domenic:
OK, if that's intended then the tests are good, just the nits in my review remaining.

fetch/api/basic/request-upload.js
yutakahirano: Done.

fetch/api/basic/request-upload.js
yutakahirano: Done.

domenic (approved):

PR #4363, opened 2016-12-19
Test same-URL navigation

annevk:
See https://github.com/whatwg/html/pull/2114 for the change to the HTML
Standard.

### DIFF
--- a/html/browsers/browsing-the-web/history-traversal/resources/a.html
+++ b/html/browsers/browsing-the-web/history-traversal/resources/a.html
@@ -1,0 +1,1 @@
+Welcome to A.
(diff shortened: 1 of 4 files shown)


Ms2ger:
Missing some `t.step_func()`s

annevk:
Ah right.

annevk:
I'll fix that, though as I noted in the parent issue I'm not even sure if this test is going to help us.

domenic:
This passes in Edge at least, and looks super-reasonable. I'll file some browser bugs.

domenic:
Filed https://bugs.chromium.org/p/chromium/issues/detail?id=676050 and https://bugzilla.mozilla.org/show_bug.cgi?id=1324861

PR #4364, opened 2016-12-19
Added a test for event sequence after implicit release.

mustaqahmed:
Fixes https://github.com/w3c/pointerevents/issues/142.

### DIFF
--- a/pointerevents/pointerevent_sequence_at_implicit_release-manual.html
+++ b/pointerevents/pointerevent_sequence_at_implicit_release-manual.html
@@ -1,0 +1,112 @@
+<!doctype html>
+<html>
+  <head>
+    <title>Pointer Event: Event sequence at implicit release</title>
+    <meta content="text/html; charset=UTF-8" http-equiv="Content-Type"/>
+    <link rel="author" title="Google" href="http://www.google.com "/>
+    <meta name="assert" content="When a captured pointer is implicitly released, the boundary events should follow the lostpointercapture event."/>
+    <link rel="stylesheet" type="text/css" href="pointerevent_styles.css">
+    <script src="/resources/testharness.js"></script>
+    <script src="/resources/testharnessreport.js"></script>
+    <script type="text/javascript" src="pointerevent_support.js"></script>
+    <script type="text/javascript">
+      var test_pointerEvent_0 = async_test("Event sequence at implicit release after click");
+      var test_pointerEvent_1 = async_test("Event sequence at implicit release after drag");
+
+      add_completion_callback(end_of_test);
+
+      var detected_pointertypes = {};
+
+      var test_num = 0;
+      var event_log = [];
+      var start_logging = false;
+      var seen_pointercancel = false;
+
+      function end_of_test() {
+        showLoggedEvents();
+        showPointerTypes();
+      }
+
+      function end_of_current_test() {
+        if (test_num == 0) {
+          var expected_events = "pointerup, lostpointercapture, pointerout, pointerleave";
+          test(function () {
+            assert_equals(event_log.join(", "), expected_events);
+          }, "Event log after click");
+
+          test_pointerEvent_0.done();
+          test_num++;
+          event_log = [];
+          start_logging = false;
+          seen_pointercancel = false;
+
+        } else {
+          var expected_events = seen_pointercancel ?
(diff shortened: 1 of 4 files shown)


pointerevents/pointerevent_sequence_at_implicit_release-manual.html
```diff
@@ -0,0 +1,112 @@
...
+            if (event.type == "pointerdown") {
+              event.target.setPointerCapture(event.pointerId);
+
+            } else if (event.type == "gotpointercapture") {
+              start_logging = true;
+
+            } else if (event.type != "pointermove" && start_logging) {
+              event_log.push(event.type);
+              if (event.type == "pointercancel")
+                seen_pointercancel = true;
```
NavidZ: Why should we expect a pointercancel while the touch-action is none?

pointerevents/pointerevent_sequence_at_implicit_release-manual.html
```diff
@@ -0,0 +1,112 @@
...
+    <meta content="text/html; charset=UTF-8" http-equiv="Content-Type"/>
+    <link rel="author" title="Google" href="http://www.google.com "/>
+    <meta name="assert" content="When a captured pointer is implicitly released, the boundary events should follow the lostpointercapture event."/>
+    <link rel="stylesheet" type="text/css" href="pointerevent_styles.css">
+    <script src="/resources/testharness.js"></script>
+    <script src="/resources/testharnessreport.js"></script>
+    <script type="text/javascript" src="pointerevent_support.js"></script>
+    <script type="text/javascript">
+      var test_pointerEvent_0 = async_test("Event sequence at implicit release after click");
+      var test_pointerEvent_1 = async_test("Event sequence at implicit release after drag");
```
NavidZ: Any reason we put these two tests in the same file? I mean they don't seem to share the instructions either.
Also is it possible to use setup_pointerevent_test if we expect multiple pointers to be used for this test?

pointerevents/pointerevent_sequence_at_implicit_release-manual.html
```diff
@@ -0,0 +1,112 @@
...
+      var seen_pointercancel = false;
+
+      function end_of_test() {
+        showLoggedEvents();
+        showPointerTypes();
+      }
+
+      function end_of_current_test() {
+        if (test_num == 0) {
+          var expected_events = "pointerup, lostpointercapture, pointerout, pointerleave";
```
NavidZ: This doesn't seem right for the mouse case. If we just click in the black why do we get pointerout/leave for mouse?

mustaqahmed: We get out/leave right before clicking the green.

mustaqahmed:
ptal.

pointerevents/pointerevent_sequence_at_implicit_release_on_click-manual.html
```diff
@@ -0,0 +1,84 @@
...
+        event_log = [];
+        start_logging = false;
+      }
+
+      function run() {
+        var test_pointer_event = setup_pointerevent_test("Event sequence at implicit release on click", ALL_POINTERS);
+
+        on_event(document.getElementById("done"), "click", function() {
+          var expected_events = "pointerup, lostpointercapture, pointerout, pointerleave";
+          test(function () {
```
NavidZ: nit: We could have done the test_pointer_event.step(function() ...) as this is the only thing that we test here. We can do that to reduce one line in the output as the async test doesn't have any other assert and always passes.
But we also need to add the step function to the MultiPointerTypeTest to call step function on the current active test.

NavidZ (commented):
lgtm

pointerevents/pointerevent_sequence_at_implicit_release_on_click-manual.html
NavidZ: I would have preferred defining the step function for MultiPointerTypeTest in pointerevent_support.js so that we could have just changed the test_pointer_event with a normal async test and everything should just work. Basically I tried to hide the fact that there are multiple tests from the test itself except in a few places.

mustaqahmed: Here it is, ptal.

NavidZ (approved):

PR #4365, opened 2016-12-19
Add URL tests for IPv4 addresses

targos:
These tests are part of a PR that fixes a bug in the Node.js implementation of the URL spec.
Since we use a copy of urltestdata.json, it was suggested that I propose the changes here.

Ref: https://github.com/nodejs/node/pull/10317


domenic:
Confirmed passing against jsdom/whatwg-url and thus very likely against the spec.

PR #4367, opened 2016-12-20
Upstream some file URL tests from jsdom

domenic:
Not sure if we want to count these as "reviewed upstream" or not


zcorpan:
I don't see any review having taken place in https://github.com/jsdom/whatwg-url/pull/60 ? Was it reviewed?

domenic:
Yeah, good point, I guess it was not :).

zcorpan (approved):

PR #4369, opened 2016-12-20
Test snapshotting of allowpaymentrequest attribute

zcorpan:
Test for https://github.com/whatwg/html/pull/2195

### DIFF
--- a/payment-request/allowpaymentrequest/echo-PaymentRequest.html
+++ b/payment-request/allowpaymentrequest/echo-PaymentRequest.html
@@ -1,0 +1,18 @@
+<!doctype html>
+<script>
+window.onmessage = (e) => {
+  const paymentArgs = [[{supportedMethods: ['foo']}], {total: {label: 'label', amount: {currency: 'USD', value: '5.00'}}}];
+
+  if (e.data === 'What is the result of new PaymentRequest(...)?') {
+    try {
+      new PaymentRequest(...paymentArgs);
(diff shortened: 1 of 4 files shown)


zcorpan:
I'm not sure why @wpt-stability-bot's harness status for Chrome is `Error`. If I run the tests on w3c-test.org in Chrome, with or without Experimental Web Platform features enabled, the harness status is `OK`.

zcorpan:
@clelland @foolip @bzbarsky

payment-request/allowpaymentrequest/echo-PaymentRequest.html
```diff
@@ -0,0 +1,18 @@
+<!doctype html>
+<script>
+window.onmessage = (e) => {
+  const paymentArgs = [[{supportedMethods: ['foo']}], {total: {label: 'label', amount: {currency: 'USD', value: '5.00'}}}];
```
foolip: So this test actually works even with 'foo' as the method?

foolip (approved):
Very similar to https://github.com/w3c/web-platform-tests/pull/4354 which I already reviewed.

payment-request/allowpaymentrequest/echo-PaymentRequest.html
zcorpan: Yes, as far as I can tell. It would fail in `show()` though if 'foo' is not an accepted method.

PR #4371, opened 2016-12-20
Add auxclick idl test

NavidZ:
Add the idl test for onauxclick
@dtapuska can you have a quick look at this.


domenic:
I think this would be better as a patch to html/dom/interfaces.html

NavidZ:
It seems that the onauxclick attribute is already added over there. So I guess that makes this change unnecessary.

PR #4372, opened 2016-12-20
Test <meta> refresh behavior when moving between documents

domenic:
Tests for https://github.com/whatwg/html/pull/2198.

### DIFF
--- a/html/semantics/document-metadata/the-meta-element/pragma-directives/attr-meta-http-equiv-refresh/allow-scripts-flag-changing-1.html
+++ b/html/semantics/document-metadata/the-meta-element/pragma-directives/attr-meta-http-equiv-refresh/allow-scripts-flag-changing-1.html
@@ -5,14 +5,16 @@
 <script src="/resources/testharnessreport.js"></script>
 <link rel="help" href="https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-http-equiv-refresh">
 
-<iframe id="source" src="support/refresh.sub.html?input=1%3B%20url%3Dfoo" sandbox="allow-same-origin"></iframe>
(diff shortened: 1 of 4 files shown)


html/semantics/document-metadata/the-meta-element/pragma-directives/attr-meta-http-equiv-refresh/allow-scripts-flag-changing-1.html
```diff
@@ -0,0 +1,42 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Meta refresh is blocked by the allow-scripts sandbox flag at its creation time, not when refresh comes due</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<link rel="help" href="https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-http-equiv-refresh">
+
+<iframe id="source" src="support/refresh.sub.html?input=1%3B%20url%3Dfoo" sandbox="allow-same-origin"></iframe>
```
zcorpan: Can you create the iframe elements in script to avoid race problem?

domenic: That seems bad because then they navigate from about:blank to the new URL. Maybe instead onload="" content attributes is the way to go.

zcorpan: It doesn't load about:blank when not in a document. When it is inserted to the document it processes the iframe attributes "for the first time", and then the src attribute is already set.

PR #4373, opened 2016-12-21
parser-uses-constructed-element.html should use assert_true for bools.

dominiccooney:
This test does assert_equals(bool_expr, string_expr) when it probably
intends to assert the bool_expr is true, with the string_expr
describing the assertion.


dominiccooney:
The Chrome failures are expected; see [Issue 676232.](https://crbug.com/676232)

PR #4374, opened 2016-12-21
[html] Remove a testcase to check case-insensitivity of radio button …

tkent-google:
…group name

The testcase doesn't match to the current specification, and
radio-groupname-case.html covers this feature.

### DIFF
--- a/html/semantics/forms/the-input-element/radio.html
+++ b/html/semantics/forms/the-input-element/radio.html
@@ -9,6 +9,9 @@
 <input type=radio name=group1 id=radio1>
 <input type=radio name=group1 id=radio2>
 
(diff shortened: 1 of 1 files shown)


tkent-google:
@domenic @TakayoshiKochi , would you review this please?

zcorpan:
Can you change it so radio3 and radio4 both have the name `groüp2`?

radio-groupname-case.html doesn't test that two controls with the *same* name with non-ASCII characters are in the same group, as far as I can tell.

tkent-google:
@zcorpan thank you for the comment.  New commit addressed it.

zcorpan:
Thanks!

PR #4378, opened 2016-12-21
Update current-realm.html test

danelphick:
Update current-realm.html test to ignore errors when loading the FontFace from about:blank since it's not a valid URL and the result of the load is not important for the test.

This helps fix https://bugs.chromium.org/p/chromium/issues/detail?id=402694 as chromium has tests that check for diffs in the console log and they trigger on this error message.

### DIFF
--- a/WebIDL/current-realm.html
+++ b/WebIDL/current-realm.html
@@ -134,6 +134,9 @@
    test(function() {
      var c = new self[0].FontFace("test", "about:blank"),
          obj = c.load()
+     obj.catch(function(reason) {
+       // Ignore exception when it fails to load because the URL is invalid.
+     });
      assert_global(obj)
 
      obj = FontFace.prototype.load.call(c)


rmcilroy:
@RByers could you give this a review for Dan? He's working on making setTimeout(0) not clamp to 1ms, and this change was needed to avoid this test being flaky.

RByers:
I don't have any particular context on this logic, but since this was previously an uncaught promise rejection  that's now just silently caught there's no real change in test semantics - just in console output.  So LGTM.

annevk:
Wait this looks bogus. That URL is not invalid.

annevk:
What specification defines rejection here?

danelphick:
The URL was always invalid and causes the test to log an error when the load promise completes.

But the test isn't testing that c.load() actually loads the test, just that the promise is global.

We could could also suppress the console error message by adding a valid font URL, but then we need a valid loadable font URL and we still won't ever look at the result.

annevk:
To be clear, it's a valid URL. It returns an empty HTML document. So what does the `c.load()` specification state? Does it actually check the bytes of the returned resource before resolving?

RByers:
The load algorithm is defined [here](https://drafts.csswg.org/css-font-loading/#font-face-load).  My reading is that yes, successfully loading a resource that is not a valid font should reject the promise.  But it's not spelled out entirely explicitly (and the spec appears to have no tests of it's own 😢).  I've filed https://github.com/w3c/csswg-drafts/issues/852 to try to get that clarified.  /cc @tabatkins

But whether it should fail or not is not being tested here.  Even if it's not supposed to fail, what should this test do when it does fail?  Isn't explicitly ignoring rejection better stylistically than implicitly letting the rejection go unhandled when the intention of the test is to not care about any failures here?

annevk:
@RByers is this test failing in all browsers due to this? The impression I got was that this change is made specifically for Chrome. If other browsers are not failing due to this, the test is finding a bug (either in Chrome or the other browsers) that is now being silenced.

annevk:
Also, that algorithm purely talks about a successful load (which isn't defined (is 404 successful?), but doesn't seem to include parsing the response).

danelphick:
Sorry, my explanation of the need for this change wasn't very clear. I'm attempting to fix the behaviour in Chrome where setTimeout always waits at least 1ms even for setTimeout(..., 0). Only after doing this, did this test start to fail.

The reason it fails is that the test completed before the font load failed and logged to the console, as previously this was taking somewhere between 0 and 1ms to occur. When the test runner compares the output to the expected output (https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/imported/wpt/WebIDL/current-realm-expected.txt), it finds that the CONSOLE ERROR message is missing. We could fix this by updating the expected output, but then we would still have a chance that the message could reappear possibly making the test flaky.

I don't know how other browsers check for test success. If they are checking the console output and their tests are terminating after the log message is printed, then their golden files would need updating as well.

jgraham:
Tests are not expected to contain console.log messages (there is supposed to be a lint that checks for this, but obviously it can't catch everything).

In general I would recommend a system where you read the test results directly rather than writing a golden file with console output and whatever. This is what Mozilla use.

domenic:
+1 on not using console output comparison, but instead comparing results from the testharness directly. I've had severe flakiness problems in the past with importing WPTs into Chrome due to that.

mkruisselbrink:
In general chrome doesn't actually use console output comparison for testharness.js based tests, but instead just directly checks if the test passed/failed. Which works fine for most cases. But if a particular file has multiple tests, most of which pass, but some of which fail, we have a few choices (with the current system): either mark the entire file as failing, losing any useful test coverage from the file, or falling back to comparing the output. So what chrome's system really needs here of course is a way to mark particular sub-tests as failing, rather than only having test expectations at the file level.

domenic:
Yeah, or maybe some way of comparing the testharness output instead of the console output.

RByers:
Sorry for the delay, @annevk thanks for doing the revert - I agree that's the right thing to do for now.  I filed [this bug](https://bugs.chromium.org/p/chromium/issues/detail?id=679742) to track improving the blink infrastructure here.

Still, aren't unhandled promise rejections generally considered bugs (the way unhandled exceptions are)?  Isn't it likely that someone debugging an unrelated failure in this tests will be thrown off by the warning that a promise rejection wasn't handled?

annevk:
I don't know. Exceptions would cause an actual test failure so they are somewhat different. Also, part of the problem here is that the specification is unclear so whether rejection needs to happen is unknown.

PR #4382, opened 2016-12-22
More tests for file URL Window drive letter quirk

zcorpan:


Sebmaster:
There are a few issues with these:
```
  1) Web platform tests parsing </> against <file:///C:/a/b>:

      AssertionError: href
      + expected - actual

      -file:///C:/
      +file:///

      at Context.<anonymous> (test/web-platform.js:20:12)

  2) Web platform tests parsing <//d:> against <file:///C:/a/b>:

      AssertionError: href
      + expected - actual

      -file:///d:
      +file://d:/

      at Context.<anonymous> (test/web-platform.js:20:12)

  3) Web platform tests parsing <//d:/..> against <file:///C:/a/b>:

      AssertionError: href
      + expected - actual

      -file:///d:/
      +file://d:/

      at Context.<anonymous> (test/web-platform.js:20:12)
```

Pretty sure the second and third one are correct failures, not quite sure about the first one.

domenic:
Not sure exactly what you mean by "correct failures" @Sebmaster, but I agree with jsdom/whatwg-url that the first result should be `file:///C:/`. I don't really know what things of the form `//driveletter:` are supposed to do, but I trust jsdom/whatwg-url, so the "actual" results are probably correct. I think at least the spec takes great pains to ensure that there are always three slashes after `file:`

Sebmaster:
"correct failures" as in: I'm pretty sure what whatwg-url does is correct here, mainly because of the triple slash thing (also because Chrome agrees with that resolution too).

zcorpan:
I guess I was confused by reading the spec. But would be nice with confirmation from @annevk that the expected results are actually per spec.

url/urltestdata.json
```diff
@@ -4587,5 +4587,76 @@
     "pathname": "/",
     "search": "",
     "hash": ""
+  },
+  "# More file URL tests by zcorpan",
```
annevk: Why do we put the author here? Should I do that too?

url/urltestdata.json
```diff
@@ -4587,5 +4587,76 @@
...
+    "hostname": "",
+    "port": "",
+    "pathname": "/",
+    "search": "",
+    "hash": ""
+  },
+  {
+    "input": "//d:",
+    "base": "file:///C:/a/b",
+    "href": "file:///d:",
```
annevk: This should end in a slash, no?

annevk (changes requested):

url/urltestdata.json
zcorpan: I just wanted to separate these tests from the ones above to direct any blame at me, but of course `git blame` works too. Can remove it though.

url/urltestdata.json
zcorpan: https://url.spec.whatwg.org/#file-state "/"
https://url.spec.whatwg.org/#file-slash-state "/"
https://url.spec.whatwg.org/#file-host-state "d:" appended to buffer, then EOF. Buffer is a Windows drive letter.
https://url.spec.whatwg.org/#path-state step 1.4.2 appends buffer to path, without adding a trailing slash.

annevk: That seems correct, but then pathname wouldn't end in a "/" either. It's also not entirely clear to me that is the desired outcome.

zcorpan: https://github.com/whatwg/url/issues/193

url/urltestdata.json
```diff
@@ -4652,7 +4652,7 @@
   {
     "input": "/",
     "base": "file:///C:/a/b",
-    "href": "file:///",
+    "href": "file:///C:/",
     "protocol": "file:",
     "username": "",
```
domenic: pathname should be "/C:/" not "/"

url/urltestdata.json
```diff
@@ -4666,7 +4666,7 @@
   {
     "input": "//d:",
     "base": "file:///C:/a/b",
-    "href": "file://d:/",
+    "href": "file:///d:",
```
domenic: pathname should be "/d:" not "/d:/"

domenic (approved):
Forgot to fixup some pathnames. Let me push an extra commit...

domenic:
@annevk all looks good to me now. Can we merge this, and deal with https://github.com/whatwg/url/issues/193 as a follow-up?

annevk (approved):

annevk:
I think that's reasonable. Will let @zcorpan merge in case there was anything else.

PR #4385, opened 2016-12-22
Make WebVTT reftests fit in 600x600 viewport

### LINKED ISSUE #4383: 2_tracks.html is too big for the viewport
CC @zcorpan 

Gecko uses 600x600 windows; I thought we agreed to standardize on that, but I can't find it in the docs anywhere.

zcorpan:
Most dimensions are divided by 4.

Two tests tested specific heights (19 and 400), and so kept their
original heights.

Move 'font-size' declaration in the refs to .video because otherwise
linebreaks are spaced according to the default font size (16px).

Fix a few tests that were missing the `<script>` to enable the track.

Fixes #4383.


zcorpan:
For /webvtt/rendering/cues-with-video/processing-model/dom_override_cue_text_while_paused.html in Gecko it sometimes throws an exception "TypeError: c is undefined"

```js
    function updateCue() {
        var t = document.getElementById('track');
        var c = t.track.cues[0];
        c.text = 'f o o';
        takeScreenshotDelayed(1000);
    }
```

I think this is a bug in Gecko; it shouldn't start playing the video until the tracks are loaded. But I should make it stable somehow and test for Gecko's bug specifically in a new test that should also be stable...

zcorpan:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1334112

zcorpan:
All results were stable!!! 🎉

webvtt/rendering/cues-with-video/processing-model/2_cues_overlapping_completely_move_up-ref.html
```diff
@@ -5,31 +5,31 @@
...
+    position: relative;
+    font-size: 9px;
 }
 #cue1 {
     position: absolute;
     top: 50%;
     left: 0;
     right: 0;
-    margin-top: -18px;
+    margin-top: -4.5px;
```
gsnedders: Are we okay generally relying on rounding being the same here? Probably?

zcorpan: I guess?

zcorpan:
@gsnedders I think it would be good to merge this asap because @BenjaminSchaaf has started to work on improving webvtt tests; having this in is a better starting point.

jgraham:
My concern here is that all? the tests seem to be failing. Do we understand why that is?

gsnedders:
Having run the tests at 1600x1200, not all tests failed. So this seems like there's some regression? (I only have a hidpi screen here and reftests are generally unstable in hipdi environments.)

zcorpan:
Many tests fail in Firefox because it does not support styling of cues, and most tests style cues to use the Ahem font.

Many tests fail in Chrome because it has a default edge margin, which the spec allows, but the tests do not expect, and because the cue background box has a default padding (not allowed by the spec). Addressing this somehow would be good but I think doesn't need to be part of this PR.

silviapfeiffer:
I suggest landing this - the test failures are not related to this change, but are general. @BenjaminSchaaf has started work on improving the test suite and this should really land beforehand.

zcorpan:
I'll merge this on Friday unless there are objections or new review comments.

gsnedders:
@jgraham I got two unexpected results comparing Firefox comparing the old tests run at 1600x1200 with the new ones at 600x600:

```
/webvtt/rendering/cues-with-video/processing-model/evil/media_404_omit_subtitles.html
-------------------------------------------------------------------------------------
TIMEOUT [Parent]

/webvtt/webvtt-file-format-parsing/webvtt-cue-text-parsing-rules/tests/timestamps.html
--------------------------------------------------------------------------------------
FAIL WebVTT cue data parser test timestamps - 398e8da1aaaf392739ca18057fef58bd5333f74d
```

The first of those, if I'm not misunderstanding, is down to the test now having `reftest-wait` and it never reaching the point where it removes `reftest-wait` due to a bug in Firefox?

The second of those is a change on master since this branches, hence can be ignored.

gsnedders:
So yes, let's land this.

zcorpan:
/webvtt/rendering/cues-with-video/processing-model/evil/media_404_omit_subtitles.html

This is a bug in the test. It assumes autoplay will happen when there is nothing to play. Oops. Can fix in followup PR. Thanks!

PR #4387, opened 2016-12-27
Dpub vocab

halindrome:
This is a simple test that will evaluate any input file for use of the dpub-aria vocabulary role values.

@iherman can you please review this so we can get it merged into the main framework?


sideshowbarker:
w3c-test:mirror

PR #4402, opened 2016-12-28
Test for the removal of keygen

domenic:
Follows https://github.com/whatwg/html/pull/2221.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/4402)


domenic:
Travis is erroring with

> ERROR:check_stability:| `/html/semantics/scripting-1/the-template-element/template-element/template-as-a-descendant.html` | `Template element as a descendant of the FRAMESET element. Template element is created by innerHTML` | **FAIL: 20/10** |

But I haven't touched that file and it seems to run the same on w3c-test.org as it does locally with my changes...

gsnedders:
@jgraham @sideshowbarker WTF is up with the stability checker on that build?

domenic:
So it seems like that test was counted as "affected" by these changes (since it has a `<script>` referencing a changed resource) and so the stability checker ran on it. It then noticed a pre-existing problem where there were duplicate test names. It didn't give the clearest error message, but fixing the duplicate test names seems to have made the new build pass, so that's good...

sideshowbarker:
> @jgraham @sideshowbarker WTF is up with the stability checker on that build?

Looks like @domenic [tracked it down](https://github.com/w3c/web-platform-tests/pull/4402#issuecomment-269554805).

Seems like a plus that the affected-tests feature in the stability checker helps catch these kinds of existing problems in other tests (though when this happens I guess we should not expect the current PR submitter to troubleshoot it and fix it but instead should assign it to one of us or to whoever last touched the other test that’s causing the problem).

domenic:
Yeah it would be better if there were an error message about duplicate test names in Travis though instead of a mysterious "20/10 failures"

sideshowbarker:
> Yeah it would be better if there were an error message about duplicate test names in Travis though instead of a mysterious "20/10 failures"

Yes I just now raised https://github.com/w3c/web-platform-tests/issues/4404 for that

zcorpan (approved):

PR #4405, opened 2016-12-29
URL: change null password handling

annevk:
URL Standard change: https://github.com/whatwg/url/pull/186.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/4405)


domenic:
One additional test seems to need a change. If we update whatwg-url to follow the latest spec, we get

```
 1) Web platform tests parsing <http://::@c@d:2> against <http://example.org/foo/bar>:

      AssertionError: href
      + expected - actual

      -http://%40c:%3A@d:2/
      +http://:%3A%40c@d:2/
```

Of course this might be a bug in whatwg-url's spec translation or in the new spec.

annevk:
I don't understand the original result.

annevk:
@domenic I have a hard time explaining the new result from either the specification or your code. Once the first ":" is seen the remainder is treated as a password, but it seems that is not happening and furthermore, somehow the password ends up before the username in the serialization. There's nothing obvious that explains such a result.

annevk:
And to be clear, `http://:%3A%40c@d:2/` is correct and expected by `urltestdata.json`. That was not immediately obvious when I first read your comment.

domenic:
@annevk here is what happens in that test case of parsing `http://::@c@d:2` with the new spec:

- (A bunch of stuff...)
- authority state, c = "@", buffer = "::"
  - Set the @ flag
  - Let the passwordTokenSeenFlag be unset
  - For each code point in buffer
   - ":"
     - code point is ":" and passwordTokenSeenFlag is unset, so set setPasswordTokenSeenFlag and continue
   - ":"
     - Let encodedCodePoints = "%3A"
     - passwordTokenSeenFlag is set, so password += encodedCodePoints
     - password is now "%3A"
  - buffer = ""
- authority state, c = "c", buffer = ""
  - buffer += "c"
- authority state, c = "@", buffer = "c"
  - @ flag is set, so prepend "%40" to buffer: buffer is now "%40c"
  - Set the @ flag
  - Let the passwordTokenSeenFlag be unset
  - For each code point in buffer:
    - "%"
      - encodedCodePoints = "%"
      - username += encodedCodePoints; username is now "%"
    - "4"
      - ditto
    - Continues until username = "%40c"
  - buffer = ""
- authority state, c = "d", buffer = ""
  - buffer += "d"
- authority state, c = ":", buffer = "d"
  - buffer += ":"
- authority state, c = "2", buffer = "d:"
  - buffer += "2"
- authority state, c = EOF, buffer = "d:2"
  - pointer -= 4
  - buffer = ""
  - state = "host"

So you can see how this ends up with a username of "%40c" and a password of "%3A"

annevk:
I see, so passwordTokenSeenFlag needs to be more global. Good thing we have this test.

annevk:
I updated the URL Standard PR.

domenic (approved):

PR #4412, opened 2017-01-03
HTML: set location.protocol to broken schemes

annevk:
See https://github.com/whatwg/url/issues/61 for context.

### DIFF
--- a/html/browsers/history/the-location-interface/location-protocol-setter.html
+++ b/html/browsers/history/the-location-interface/location-protocol-setter.html
@@ -33,16 +33,56 @@ onmessage = (e) => {
 </script>"></iframe>
 <script>
   let broken = [
-    'x\x20',
-    'x\x00',
     '\x00',
+    '\x01',
+    '\x0A',
+    '\x20',
+    '\x21',
+    '\x7F',
+    '\x80',
+    '\xFF',
     ':',
     '†',
-    'a†',
+    '\x00x',
+    '\x01x',
+    '\x0Ax',
+    '\x20x',
+    '\x21x',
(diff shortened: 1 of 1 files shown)


annevk:
Note: this requires some changes to the URL Standard that the HTML Standard has forever anticipated. PR coming up for that.

annevk:
It might be worth testing these broken schemes inside a data URL `<iframe>` as well and maybe an `<iframe srcdoc>`. So please don't land straight away.

annevk:
https://github.com/whatwg/url/pull/194 is the change to the URL Standard by the way. This is ready to land now.

domenic:
These all fail in Chrome and Firefox because of the exception type, right? I think we should change it to SyntaxError DOMException in the tests and spec before merging; no need to churn browsers over this detail.

annevk:
Well Firefox needs a change anyway since it's not "SyntaxError" there but rather some internal thing. I guess we can decide to align with Chrome, but it seems a little weird since we use TypeError normally for URL parsing issues.

domenic:
Hmm. Edge also has a weird `Error` instead of any normal error type. Safari matches Chrome though. I guess I do not feel strongly, but I do note the following places that use "SyntaxError" DOMExceptions for URL parsing failures:

- location.assign/location.replace
- registerProtocolHandler/registerContentHandler
- EventSource constructor
- WebSocket constructor
- window.postMessage origin parser
- Worker/SharedWorker constructor
- importScripts

annevk:
Okay, I will give up, but not today, dinner time!

annevk:
This is now ready to land.

PR #4413, opened 2017-01-03
URL: protocol setter and special URLs

annevk:
Closes https://github.com/whatwg/url/issues/104.

### DIFF
--- a/url/setters_tests.json
+++ b/url/setters_tests.json
@@ -164,7 +164,7 @@
             "href": "nonsense:///test",
             "new_value": "https",
             "expected": {
-                "href": "nonsense:///",
+                "href": "nonsense:///test",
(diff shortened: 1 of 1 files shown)


url/setters_tests.json
```diff
@@ -129,6 +145,30 @@
...
+            "expected": {
+                "href": "ssh://me@example.net/",
+                "protocol": "ssh:"
+            }
+        },
+        {
+            "href": "nonsense:///test",
+            "new_value": "https",
+            "expected": {
+                "href": "nonsense:///",
```
domenic: This should be `nonsense:///test` according to jsdom/whatwg-url

annevk: Indeed. My bad.

domenic (approved):

PR #4417, opened 2017-01-03
support focusin/focusout based on webkit/blink

jgraham:
Blink and webkit launch focusin after focus and focusout after blur. Despite
this contradiction with the spec, it is best to mirror this new way, as there
is little guidance or existing code to clarify implementation amiguities that
can arise from the spec.

If focus/blur is fired on a window or document, or the event triggers a change
of focus, do not fire the corresponding focusin/focusout. Otherwise, always
fire the corresponding event.

Additionally, add a mochitest and a w3c-platform-test.

MozReview-Commit-ID: AgQ8JBxKIqK

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=687787


jgraham:
Code reviewed upstream.

PR #4419, opened 2017-01-03
Fix the race condition to prevent the access of metadata in both PBackground thread and the Connection thread.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1309527


jgraham:
Code reviewed upstream.

PR #4420, opened 2017-01-03
Do not expect init segment to be parsed when appending corrupted data.

jgraham:
From the Segment Parser Loop definition:
https://w3c.github.io/media-source/index.html#sourcebuffer-segment-parser-loop
"2. If the input buffer contains bytes that violate the SourceBuffer byte stream format specification, then run the append error algorithm and abort this algorithm."

The test appends data with a valid init segment followed by corrupted content.
The corrupted content satisfies the condition that the input buffer contains bytes that violate byte stream format specification.
As such, the append error algorithm is to be run prior parsing the init segment.

Add a new test verifying the defined behavior.

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1314533

### DIFF
--- a/IndexedDB/idbindex-rename.html
+++ b/IndexedDB/idbindex-rename.html
@@ -1,4 +1,5 @@
 <!DOCTYPE html>
+<meta name="timeout" content="long">
 <title>IndexedDB: index renaming support</title>
 <link rel="help"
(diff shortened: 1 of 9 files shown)


jgraham:
Code reviewed upstream.

PR #4424, opened 2017-01-03
Revert the set-from-dom.sub.html and set-from-http.sub.html test cases of web-platform.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1316532


jgraham:
Code reviewed upstream.

PR #4425, opened 2017-01-03
P2 Add wpt test to verify timers are restored after being loaded from bfcache.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1316289


jgraham:
Code reviewed upstream.

PR #4440, opened 2017-01-03
Remove entries from a form's past names map when an element is removed from the form, even if that element doesn't have a name or id anymore.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1318576


jgraham:
Code reviewed upstream.

PR #4443, opened 2017-01-03
Wait for a single finish event itself instead of waiting two requestAnimationFrame and checking the event.

jgraham:
Also we don't need to check the event time, it's out of scope of this test case.

MozReview-Commit-ID: JkbiiVPunGH

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1316050


jgraham:
Code reviewed upstream.

PR #4445, opened 2017-01-03
followup.  Fix web platform tests that either depended on the old spec's behavior or were asserting that we fail the new one.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1315135


jgraham:
Code reviewed upstream.

PR #4447, opened 2017-01-03
Only allow valid base64-values for CSP nonce and hash sources, per spec.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1309219


jgraham:
Code reviewed upstream.

PR #4449, opened 2017-01-03
Fix <select> validity status for listboxes and for non-placeholder empty valued options.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1052045


jgraham:
Code reviewed upstream.

PR #4450, opened 2017-01-03
Do not fail event processing if an exception is thrown in ServiceWorker.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1271692


jgraham:
Code reviewed upstream.

PR #4453, opened 2017-01-03
Cross-origin objects should allow gets of certain symbol-named properties but force the value to be undefined.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1321299


jgraham:
Code reviewed upstream.

PR #4457, opened 2017-01-03
Replace a use of cStringIO with StringIO to avoid a problem with old Python versions,

jgraham:
MozReview-Commit-ID: D1wkrdaejQB

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1257249


jgraham:
Code reviewed upstream.

PR #4458, opened 2017-01-03
Ensure that we don't replace load popup.

jgraham:
This makes open-url-multi-window-6.htm always fail, which is intended.

MozReview-Commit-ID: 4OENwxMLUeh

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1315260


jgraham:
Code reviewed upstream.

PR #4462, opened 2017-01-03
Fix property enumeration on cross-origin windows to include indexed props, and add some tests for the ordering of the indexed, named, and symbol-named props on cross-origin objects (window and location).

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1322415


jgraham:
Code reviewed upstream.

PR #4469, opened 2017-01-03
navigationStart should not be exposed to workers as timing attribute,

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1323941


jgraham:
Code reviewed upstream.

PR #4474, opened 2017-01-03
Use single quote instead of double for consistency.

jgraham:
Also, dropped expected failures for properties that have been already dropped
in test cases from meta file.

MozReview-Commit-ID: I7d7bdg0nMs


jgraham:
Code reviewed upstream.

PR #4477, opened 2017-01-03
Rename type-per-property.html to interpolation-per-property.html

jgraham:
MozReview-Commit-ID: 6SHutpwLMTH


jgraham:
Code reviewed upstream.

PR #4479, opened 2017-01-03
Test properties that are animatable as length.

jgraham:
Some properies (e.g. outline-width) need prerequisite styles to compute
animating styles.

MozReview-Commit-ID: LB4eWIM07dY


jgraham:
Code reviewed upstream.

PR #4486, opened 2017-01-03
Encode URL query string segments to bytes with HTML numeric character references for unmappable characters.

jgraham:
MozReview-Commit-ID: HDJxSSZZYlw

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1225255


jgraham:
Code reviewed upstream.

PR #4490, opened 2017-01-03
Implement keyframe composite(add).

jgraham:
This patch also fixes expected computed offset values in frame at 0.5 offset for
add composite.

MozReview-Commit-ID: 8PNp237NoV4


jgraham:
Code reviewed upstream.

PR #4491, opened 2017-01-03
Fix test cases that checks keyframe composite is not specified but effect composite is specified.

jgraham:
From spec <https://w3c.github.io/web-animations/#dom-keyframeeffectreadonly-getkeyframes>:

composite
    The keyframe-specific composite operation used to combine the values
    specified in this keyframe with the underlying value.
    This member will be absent if the composite operation specified on
    the keyframe effect is being used.

MozReview-Commit-ID: 8ob59Xv6DRL


jgraham:
Code reviewed upstream.

PR #4492, opened 2017-01-03
addition result tests per properties.

jgraham:
This includes some failure test cases:

1) Color interpolation with overflowed colors.
Additive color needs to be with overflowed color components that are produced
by additive calculation. For example:

 an additive color animation between rgb(100, 100, 100) and rgb(200, 200, 200)
 on rgb(100, 100, 100) element should be an animation of colors;
 from rgb(200, 200, 200) to rgb(300, 300, 300).

This will be fixed in the next patch (part 8).

2) All transform tests.
Additive transform list has to be concatinated, it will be done in a subsequent
patch (part 9).

MozReview-Commit-ID: 5OoT6icVMcC

### DIFF
--- a/web-animations/animation-model/animation-types/addition-per-property.html
+++ b/web-animations/animation-model/animation-types/addition-per-property.html
@@ -1,6 +1,6 @@
-<!DOCTYPE html>
+<!doctype html>
 <meta charset=utf-8>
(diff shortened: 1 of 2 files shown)


jgraham:
Code reviewed upstream.

PR #4497, opened 2017-01-03
The fieldset's border drawing display item shouldn't participate in hit-testing.  That's handled by its background item already.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1326163


jgraham:
Code reviewed upstream.

PR #4498, opened 2017-01-03
Inherit a fieldset's border-radius down into its content wrapper box, so it doesn't leak out and mess up hit-testing in the corner areas when it has rounded borders.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1326163


jgraham:
Code reviewed upstream.

PR #4502, opened 2017-01-04
HTML: set location.protocol to non-broken-non-functioning schemes

annevk:
See https://github.com/whatwg/url/issues/61 for context.

https://github.com/w3c/web-platform-tests/pull/4412 tests broken
schemes.

### DIFF
--- a/html/browsers/history/the-location-interface/location-protocol-setter-non-broken-weird.html
+++ b/html/browsers/history/the-location-interface/location-protocol-setter-non-broken-weird.html
@@ -1,0 +1,38 @@
+<!doctype html>
+<title>Set location.protocol from an HTTP URL</title>
+<!-- In particular, valid non-broken schemes that are nevertheless not going to work -->
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<div id=log></div>
+<iframe src=resources/dummy.html></iframe>
+<iframe src=resources/dummy.html></iframe>
+<iframe src=resources/dummy.html></iframe>
+<iframe src=resources/dummy.html></iframe>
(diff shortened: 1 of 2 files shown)


annevk:
The -weird.html is an earlier iteration of the test (excluding data URL tests) that for some reason never completes running in Firefox. I thought I'd include it since it does show differences between the two ways of testing.

html/browsers/history/the-location-interface/location-protocol-setter-non-broken.html
```diff
@@ -0,0 +1,64 @@
...
+    'ftp',
+    'gopher',
+    'http+x'
+  ].forEach((val) => {
+    async_test((t) => {
+      // HTTP URL <iframe>
+      const frame = document.createElement("iframe")
+      frame.src = "resources/dummy.html"
+      frame.onload = t.step_func(() => {
+        try {
```
zcorpan: Remove

html/browsers/history/the-location-interface/location-protocol-setter-non-broken-weird.html
```diff
@@ -0,0 +1,38 @@
...
+          assert_equals(self[index].location.protocol, location.protocol)
+          assert_equals(self[index].location.host, location.host)
+          assert_equals(self[index].location.port, location.port)
+          t.done()
+        }, 500)
+      } catch(e) {
+        assert_unreached(e)
+      }
+    })
+  })
```
zcorpan: Include `val` in the test name

zcorpan (changes requested):

annevk:
Thanks, pushed fixes.

zcorpan (approved):

PR #4503, opened 2017-01-04
HTML: set location.pathname to ?

annevk:
Closes https://github.com/whatwg/url/issues/133.

### DIFF
--- a/html/browsers/history/the-location-interface/location-pathname-setter-question-mark.html
+++ b/html/browsers/history/the-location-interface/location-pathname-setter-question-mark.html
@@ -1,0 +1,16 @@
+<!doctype html>
+<title>Set location.pathname to ?</title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<div id=log></div>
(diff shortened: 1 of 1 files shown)


html/browsers/history/the-location-interface/location-pathname-setter-question-mark.html
```diff
@@ -0,0 +1,16 @@
+<!doctype html>
+<title>Set location.pathname to ?</title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<div id=log></div>
+<iframe src=/></iframe>
```
zcorpan: Not sure it's great to invoke a directory listing in this test... Could it use `/common/blank.html` instead?

annevk: I should probably use that more often than creating dummy resources...

annevk: Fixed.

domenic (approved):

PR #4505, opened 2017-01-04
Verify that preloads with a "fetch" `as` value have an empty destination

yoavweiss:
@annevk - As discussed on https://github.com/whatwg/fetch/pull/442, here's the corresponding test. Note that it fails in all current implementations.

### DIFF
--- a/preload/fetch_destination.https.html
+++ b/preload/fetch_destination.https.html
@@ -5,7 +5,7 @@
 <script>
 async_test(function(t) {
     var worker_url = 'resources/fetch-destination-worker.js';
-    var scope = 'resources/dummy.xml';
+    var scope = 'resources/empty.html';
     var registration;
 
     function done(registration) {
@@ -21,12 +21,16 @@ async_test(function(t) {
           return with_iframe(scope);
         }))
       .then(t.step_func(function(frame) {
+              console.log(frame);
           var link = frame.contentWindow.document.createElement("link");
           link.as = "fetch";
-          link.href = scope;
+          link.href = "resources/dummy.xml";
           link.rel = "preload";
           link.addEventListener("load", t.step_func(function() { done(registration); }));
-          link.addEventListener("error", t.step_func(function() { assert_unreached("Fetch destination preload errored"); done(registration); }));
+          link.addEventListener("error", t.step_func_done(function() {
+              registration.unregister();
+              assert_unreached("Fetch destination preload errored");
+          }));
           frame.contentWindow.document.body.appendChild(link);
         }))
       .catch(unreached_rejection(t));
--- a/preload/fetch_destination.https.html
+++ b/preload/fetch_destination.https.html
@@ -5,7 +5,7 @@
 <script>
 async_test(function(t) {
     var worker_url = 'resources/fetch-destination-worker.js';
-    var scope = 'resources/empty.html';
+    var scope = 'resources/dummy.xml';
     var registration;
 
     function done(registration) {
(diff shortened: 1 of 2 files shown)


preload/fetch_destination.https.html
```diff
@@ -0,0 +1,35 @@
+<!DOCTYPE html>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="../service-workers/service-worker/resources/test-helpers.sub.js"></script>
+<script>
+async_test(function(t) {
+    var worker_url = 'resources/fetch-destination-worker.js';
+    var scope = 'resources/empty.html';
```
annevk: You can use `/common/blank.html` instead.

annevk: Also, that URL doesn't match the preload URL from later on. Is that intentional?

yoavweiss:
@annevk - Changed the scope to be the preloaded resource, and added a bunch of needed step_func wrappers around the various async test functions. (due to timeouts I seen locally)

preload/resources/fetch-destination-worker.js
```diff
@@ -0,0 +1,11 @@
+self.addEventListener('fetch', function(event) {
+    if (event.request.url.indexOf('dummy.xml') != -1) {
+        var response = new Response();
+        if (event.request.destination == "")
+            response.status = 200;
+        else
+            response.status = 500;
```
annevk: This doesn't match the Preload specification. It doesn't say that an error event is dispatched for such status codes. That would also likely be problematic due to the same-origin policy mostly trying to keep that a secret.

preload/fetch_destination.https.html
```diff
@@ -0,0 +1,34 @@
...
+      .then(t.step_func(function() {
+          return with_iframe(scope);
+        }))
+      .then(t.step_func(function(frame) {
+          var link = frame.contentWindow.document.createElement("link");
+          link.as = "fetch";
+          link.href = scope;
+          link.rel = "preload";
+          link.addEventListener("load", t.step_func(function() { done(registration); }));
+          link.addEventListener("error", t.step_func(function() { assert_unreached("Fetch destination preload errored"); done(registration); }));
```
annevk: Does this even run? I think that an assert results in an exception.

annevk (changes requested):
Found some more issues.

preload/resources/fetch-destination-worker.js
yoavweiss: Hmm, I guess we need to tighten the definition of "load was successful" there. I understood it as including the status code. Is there another way that a SW can signal the page that a load was unsuccessful?

preload/fetch_destination.https.html
yoavweiss: It runs and fails in current implementations (which trigger an error when `as=fetch`)

preload/resources/fetch-destination-worker.js
annevk: I'm not sure. @jakearchibald?

preload/fetch_destination.https.html
annevk: But does it call `done()`?

yoavweiss: It fails (immediately) with an assert_unreached().
If this is an issue and done() must be called, what's the proper way define unreachable code without asserting?

annevk: I don't think it's an issue, but I would just leave out done(). Come to think of it, you can use `t.step_func_done()` which cleans this up automatically.

preload/resources/fetch-destination-worker.js
annevk: In theory returning a network error can be done through `Response.error()` per https://fetch.spec.whatwg.org/#dom-response-error.

yoavweiss: Cool. Using that instead

preload/fetch_destination.https.html
yoavweiss: `step_func_done` it is (plus unregistration)

yoavweiss:
@annevk - I brought back the old scope (with the empty.html) as I was hitting weird errors related to frame access with the scope and link both pointing at the same URL.

annevk:
What's wrong with using http://w3c-test.org/common/blank.html as I suggested?

yoavweiss:
Oh, should've commented on that. It's from a different directory than the SW registered, and the SW registration fails.

preload/fetch_destination.https.html
```diff
@@ -0,0 +1,37 @@
...
+        }))
+      .then(t.step_func(function() {
+          return with_iframe(scope);
+        }))
+      .then(t.step_func(function(frame) {
+          var link = frame.contentWindow.document.createElement("link");
+          link.as = "fetch";
+          link.href = "resources/dummy.xml";
+          link.rel = "preload";
+          link.addEventListener("load", t.step_func(function() { done(registration); }));
```
annevk: Use step_func_done here too and inline the unregister call?

Seems okay then.

yoavweiss: done

yoavweiss:
Would you like me to squash the commits?

PR #4506, opened 2017-01-04
Use () => element.requestFullscreen() to make errors more obvious

foolip:
If the unprefixed API isn't supported, it would currently fail with
"TypeError: Unable to get property 'bind' of undefined or null
reference" or similar, obscuring the root cause.


foolip:
Prompted by mail from @aliams. Now running the tests manually will actually time out instead, because the  code that throws isn't wrapped in `t.step_func`. That's not an improvement, so closing this.

PR #4507, opened 2017-01-04
Fullscreen: give trusted_click and trusted_click the test for step_func

foolip:
The element.requestFullscreen() is wrapped in an arrow function, which
together with the other changes should make failures more obvious.


foolip:
This seems to have worked better, can you review, @aliams?

aliams:
LGTM. The error messages are much clearer now.

As an aside, I was wondering if there are any best practices when it comes to using the new JS operators considering that you may want to run these tests on older browsers.

PR #4509, opened 2017-01-05
IndexedDB upgrade transaction lifecycle tests.

pwnall:
These tests are upstreamed from Blink - https://crrev.com/2556373003

All the tests pass in Chrome Canary.

Firefox Nightly fails one test - `in a promise microtask after abort() is called, before the transaction abort event is fired`. I think this is a reasonable test, because the microtask aspect in it can be ignored. The transaction was aborted by JavaScript by calling abort(), and the `abort` event hasn't fired yet, so object store operations should throw TransactionInactiveError, no matter how microtasks are handled.

/cc @inexorabletash @bevis-tseng

### DIFF
--- a/IndexedDB/support-promises.js
+++ b/IndexedDB/support-promises.js
@@ -1,42 +1,46 @@
-// Returns an IndexedDB database name likely to be unique to the test case.
-const databaseName = (testCase) => {
+'use strict';
+
+// Returns an IndexedDB database name that is unique to the test case.
+function databaseName(testCase) {
   return 'db' + self.location.pathname + '-' + testCase.name;
-};
+}
 
 // Creates an EventWatcher covering all the events that can be issued by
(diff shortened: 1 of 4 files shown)


bevis-tseng:
This seems to be a known issue regarding the checkpoint of the promise callback in Gecko and is going to be fixed recently. We'll revisit this test case after fixed. Thanks for catching this issue.

pwnall:
@bevis-tseng Do you mean you'd rather that we wait for the Firefox fixes before adding this test to the repository, or do you mean that you're OK with the test going in, and you'll take a look at getting Firefox to pass after the promise-related changes are done?

Thank you very much!

bevis-tseng:
It's totally fine to have this test included without waiting our fix in FF.
We'll have a follow-up bug internally for tracking this.

PR #4513, opened 2017-01-06
Test IPv4 part of IPv6

annevk:
See https://github.com/whatwg/url/issues/195 for context.

### DIFF
--- a/url/setters_tests.json
+++ b/url/setters_tests.json
@@ -631,6 +631,52 @@
                 "hostname": "example.com",
                 "port": ""
             }
+        },
+        {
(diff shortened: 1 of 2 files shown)


annevk:
@achristensen07 looks like Safari TP might have some issues with setters, including these newly added tests.

achristensen07:
Yep.  It looks like the URL object behaves correctly, but the anchor tag setters don't do the right thing when set to an invalid hostname, etc.  I need to fix those issues in WebKit.  The tests correctly test the spec.

domenic (approved):

PR #4517, opened 2017-01-10
application/x-www-form-urlencoded parser tests

annevk:
To support https://github.com/whatwg/url/pull/197.


domenic:
jsdom doesn't have a URLSearchParams or formData so I can't help in my usual way here. The %C2 failing in both Chrome and Firefox is a little worrying I guess.

annevk:
The reason for that is that the Encoding library in Chrome and Firefox doesn't detect EOF correctly. I think the current sentiment is that everyone still wants to fix that at some point.

annevk:
https://bugs.chromium.org/p/chromium/issues/detail?id=633153 deals with the other %C2 issue seen here.

annevk:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1330245 on Firefox doing something with the BOM. I should probably add a test for the UTF-16 BOM. And maybe I should also test response.formData() since it's cheap to do.

annevk:
Filed https://bugs.webkit.org/show_bug.cgi?id=166920 and https://bugs.webkit.org/show_bug.cgi?id=166921 on Safari TP.

annevk:
I ended up not filing an issue against Edge since it implements neither feature and I'm pretty sure `fetch()` and `URLSearchParams` are already filed.

annevk:
Should I also copy all of http://searchfox.org/mozilla-central/source/dom/url/tests/test_urlSearchParams.html#67 here? @bakulf?

bakulf:
Yes, please.

annevk:
I found four duplicates in those tests that I did not copy. I also did not copy the last three tests (starting with "?") as those parse differently for `URLSearchParams` (strips leading "?").

url/urlencoded-parser.html
```diff
@@ -0,0 +1,71 @@
...
+  { "input": 'a=b&c=d&', "output": [['a', 'b'], ['c', 'd']] },
+  { "input": '&&&a=b&&&&c=d&', "output": [['a', 'b'], ['c', 'd']] },
+  { "input": 'a=a&a=b&a=c', "output": [['a', 'a'], ['a', 'b'], ['a', 'c']] },
+  { "input": 'a==a', "output": [['a', '=a']] },
+  { "input": 'a=a+b+c+d', "output": [['a', 'a b c d']] },
+  { "input": '%=a', "output": [['%', 'a']] },
+  { "input": '%a=a', "output": [['%a', 'a']] },
+  { "input": '%a_=a', "output": [['%a_', 'a']] },
+  { "input": '%61=a', "output": [['a', 'a']] },
+  { "input": '%61+%4d%4D=', "output": [['a MM', '']] }
```
bakulf: , at the end of this line?

bakulf (approved):
lgtm

url/urlencoded-parser.html
annevk: I haven't done that so far. I guess I'll stick with my own convention, but if you ever change this feel free to add it.

PR #4523, opened 2017-01-11
Test new URLSearchParams constructor features

annevk:
The main thing here is new tests for
https://github.com/whatwg/url/pull/175 but I also found issues
with some existing tests that I’m attempting to fix at the same time.

The current tests assumed the constructor argument was mandatory, but
it’s in fact optional. The other issue was that the current test assume
stringifying DOMException.prototype throws, but I don’t think that’s
the case.

### DIFF
--- a/url/urlsearchparams-constructor.html
+++ b/url/urlsearchparams-constructor.html
@@ -35,7 +35,7 @@ test(() => {
 
 test(() => {
     params = new URLSearchParams({});
-    assert_equals(params + '', '%5Bobject+Object%5D=');
+    assert_equals(params + '', "");
 }, 'URLSearchParams constructor, {} as argument');
 
 test(function() {
@@ -157,7 +157,7 @@ test(() => {
   }
   let params2 = new URLSearchParams(params)
   assert_equals(params2.get("a"), "b")
-}, "Evil constructor test")
+}, "Custom [Symbol.iterator]")
 </script>
 </head>
 </html>
--- a/url/urlsearchparams-constructor.html
+++ b/url/urlsearchparams-constructor.html
@@ -20,16 +20,23 @@ test(function() {
 test(function() {
     var params = new URLSearchParams()
     assert_equals(params.toString(), "")
+}, "URLSearchParams constructor, no arguments")
 
+test(() => {
     params = new URLSearchParams(DOMException.prototype);
     assert_equals(params.toString(), "Error=")
+}, "URLSearchParams constructor, DOMException.prototype as argument")
 
+test(() => {
     params = new URLSearchParams('');
     assert_true(params != null, 'constructor returned non-null value.');
     assert_equals(params.__proto__, URLSearchParams.prototype, 'expected URLSearchParams.prototype as prototype.');
+}, "URLSearchParams constructor, empty string as argument")
+
+test(() => {
     params = new URLSearchParams({});
     assert_equals(params + '', '%5Bobject+Object%5D=');
-}, 'URLSearchParams constructor, empty and unusual input.');
+}, 'URLSearchParams constructor, {} as argument');
 
 test(function() {
     var params = new URLSearchParams('a=b');
@@ -144,12 +151,12 @@ test(function() {
 })
 
 test(() => {
-  let params1 = new URLSearchParams("a=b"),
-      params2 = new URLSearchParams()
-  params2[Symbol.iterator] = params1[Symbol.iterator]
-  let params3 = new URLSearchParams(params2)
-  assert_equals(params3.get("a"), params1.get("a"))
-  assert_equals(params3.get("a"), "b")
+  params = new URLSearchParams()
+  params[Symbol.iterator] = function *() {
+    yield ["a", "b"]
+  }
+  let params2 = new URLSearchParams(params)
+  assert_equals(params2.get("a"), "b")
 }, "Evil constructor test")
 </script>
 </head>
--- a/url/urlsearchparams-constructor.html
+++ b/url/urlsearchparams-constructor.html
@@ -18,15 +18,18 @@ test(function() {
 }, 'Basic URLSearchParams construction');
 
 test(function() {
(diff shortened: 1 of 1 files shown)


url/urlsearchparams-constructor.html
```diff
@@ -18,15 +18,18 @@
...
+    params = new URLSearchParams(DOMException.prototype);
+    assert_equals(params.toString(), "Error=")
+
+    params = new URLSearchParams('');
     assert_true(params != null, 'constructor returned non-null value.');
     assert_equals(params.__proto__, URLSearchParams.prototype, 'expected URLSearchParams.prototype as prototype.');
     params = new URLSearchParams({});
     assert_equals(params + '', '%5Bobject+Object%5D=');
-}, 'URLSearchParams constructor, empty.');
+}, 'URLSearchParams constructor, empty and unusual input.');
```
domenic: Separate tests per input will make failures clearer.

url/urlsearchparams-constructor.html
```diff
@@ -124,6 +127,30 @@
...
+})
+
+test(() => {
+  let params1 = new URLSearchParams("a=b"),
+      params2 = new URLSearchParams()
+  params2[Symbol.iterator] = params1[Symbol.iterator]
+  let params3 = new URLSearchParams(params2)
+  assert_equals(params3.get("a"), params1.get("a"))
+  assert_equals(params3.get("a"), "b")
+}, "Evil constructor test")
```
domenic: I don't think this works as expected. `params1[Symbol.iterator] === params2[Symbol.iterator] === URLSearchParams.prototype[Symbol.iterator]`. The assignment is a no-op. The iterator will look up the private data on the instance it's invoked on, so iterating over params2 will always give no params, so params3 should be empty.

domenic (changes requested):

url/urlsearchparams-constructor.html
annevk: Oh, shit. I based that on #4240 which I also wrote. How would I create a custom iterator? Would assigning an array be sufficient?

domenic: The easiest way is to use a generator:

```js
params3[Symbol.iterator] = function *() {
  yield [a, b];
  yield [c, d];
};
```

annevk: Thanks, used.

url/urlsearchparams-constructor.html
annevk: Done this as well.

url/urlsearchparams-constructor.html
```diff
@@ -18,15 +18,25 @@
...
+
+test(() => {
+    params = new URLSearchParams('');
     assert_true(params != null, 'constructor returned non-null value.');
     assert_equals(params.__proto__, URLSearchParams.prototype, 'expected URLSearchParams.prototype as prototype.');
+}, "URLSearchParams constructor, empty string as argument")
+
+test(() => {
     params = new URLSearchParams({});
     assert_equals(params + '', '%5Bobject+Object%5D=');
```
domenic: I think this should now be equivalent to no arguments per record semantics. So just empty string.

url/urlsearchparams-constructor.html
```diff
@@ -124,6 +134,30 @@
...
+})
+
+test(() => {
+  params = new URLSearchParams()
+  params[Symbol.iterator] = function *() {
+    yield ["a", "b"]
+  }
+  let params2 = new URLSearchParams(params)
+  assert_equals(params2.get("a"), "b")
+}, "Evil constructor test")
```
domenic: Maybe "overridden iterator test" is now more accurate.

annevk:
Thanks!

domenic (approved):

url/urlsearchparams-constructor.html
```diff
@@ -18,15 +18,25 @@
...
-                  'Calling \'URLSearchParams\' without \'new\' should throw.');
-    assert_throws(new TypeError(), function () { new URLSearchParams(DOMException.prototype); });
-    var params = new URLSearchParams('');
+    var params = new URLSearchParams()
+    assert_equals(params.toString(), "")
+}, "URLSearchParams constructor, no arguments")
+
+test(() => {
+    params = new URLSearchParams(DOMException.prototype);
+    assert_equals(params.toString(), "Error=")
```
cdumez: Can you clarify why this is the expected result here. My interpretation of the Web IDL spec [1] is that we should go to step 11.1 and treat the input as a sequence (not a String).

[1] https://heycam.github.io/webidl/#es-union

domenic: Yes but method is undefined so it then continues on with the rest of the steps.

cdumez: How about 11.4 ? We do have a record in there.

domenic: Yeah you're right. I think this is indeed incorrect.

annevk: https://github.com/w3c/web-platform-tests/pull/4581 has the fix. I also commented on the WebKit bug.

url/urlsearchparams-constructor.html
```diff
@@ -124,6 +134,30 @@
     params = new URLSearchParams('a%f0%9f%92%a9b=c');
     assert_equals(params.get('a\uD83D\uDCA9b'), 'c');
 }, 'Parse %f0%9f%92%a9');  // Unicode Character 'PILE OF POO' (U+1F4A9)
+
+;[
+  { "input": {"+": "%C2"}, "output": [[" ", "\uFFFD"]], "name": "object with +" },
```
TimothyGu: @annevk, I'm in the process of implementing the URL spec change in Node.js. Where does it say in the URL Standard that keys and values of a record must be unescaped?

annevk: That's a good question, it doesn't and I'm not sure it should. I filed https://github.com/whatwg/url/issues/220.

PR #4525, opened 2017-01-11
Test header values

annevk:
See https://github.com/whatwg/fetch/issues/332 for details.

Unless https://github.com/w3c/wptserve/issues/111 is fixed soon we may have to apply a temporary exception for 0x0B and 0x0C.

### DIFF
--- a/XMLHttpRequest/setrequestheader-bogus-value.htm
+++ b/XMLHttpRequest/setrequestheader-bogus-value.htm
@@ -17,10 +17,9 @@
           assert_throws("SyntaxError", function() { client.setRequestHeader("x-test", value) }, ' given value ' + value+', ');
         });
       }
+      try_value("t\x00t");
       try_value("t\rt");
       try_value("t\nt");
-      try_value("t\bt");
-      try_value("\x7f");
       test(function() {
         var client = new XMLHttpRequest();
         client.open("GET", "...");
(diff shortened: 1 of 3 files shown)


annevk:
@jgraham this test runs fine locally. Is 500 fetches really too much? What should happen instead here?

jgraham:
Try adding `<meta name=timeout content=long>` to the test (500 fetches in 10s seems like a lot to me).

youennf:
IIRC, wptserve does not support pipeline so basically, 500 fetches means creating 500 connections.
Why not adding support for it?

youennf:
See https://github.com/w3c/wptserve/pull/86 and https://github.com/w3c/wptserve/issues/108

annevk:
@youennf pipelining is not web-compatible I believe. Even so, this test seems to work fine for me locally and since the longer timeout it also runs fine on Travis.

annevk:
@youennf are you okay with this test and the change to the standard?

youennf:
Sounds ok

ricea:
500 fetches is going to cause masses of timeouts on Chromium's bots. As a rule of thumb, the slowest bots can be 100 times slower than running locally.

I suggest only testing the characters that have been known to cause problems.

annevk:
@ricea I could remove 0-9, a-Z. But there's a lot of code points here that cause a problem of sorts.

ricea:
@annevk Based on my experience, you have a budget of about 50 connections to play with. It looks like you should be able to test multiple headers on a single connection; would that work?

annevk:
That's a reasonable idea. I'll work on a PR.

annevk:
Created two PRs that do just that.

PR #4531, opened 2017-01-12
URLSearchParams.prototype.sort()

annevk:
See https://github.com/whatwg/url/issues/26 for context.

### DIFF
--- a/url/urlsearchparams-sort.html
+++ b/url/urlsearchparams-sort.html
@@ -12,6 +12,14 @@
   {
     "input": "\uFFFD=x&\uFFFC&\uFFFD=a",
     "output": [["\uFFFC", ""], ["\uFFFD", "x"], ["\uFFFD", "a"]]
+  },
+  {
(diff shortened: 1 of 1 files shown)


url/urlsearchparams-sort.html
```diff
@@ -0,0 +1,27 @@
...
+[
+  {
+    "input": "z=b&a=b&z=a&a=a",
+    "output": [["a", "b"], ["a", "a"], ["z", "b"], ["z", "a"]]
+  },
+  {
+    "input": "\uFFFD=x&\uFFFC&\uFFFD=a",
+    "output": [["\uFFFC", ""], ["\uFFFD", "x"], ["\uFFFD", "a"]]
+  }
+].forEach((val) => {
```
domenic: I think we need at least a few tests on Unicode inputs, to show what type of sorting is being applied here. I assume it's just going to be by character code, so instead of sorting é near e, it'll sort after the usual alphabet? Tests to confirm.

annevk:
What do you think?

domenic (approved):

PR #4534, opened 2017-01-12
Add tests for DOMStringList

inexorabletash:
For https://github.com/w3c/web-platform-tests/issues/4522 to unblock https://github.com/whatwg/html/pull/2192

Not the most extensive test suite in the world and the dependency on IndexedDB is unfortunate, but it was handy.

Should the idlharness test remain stand-alone or be merged into `html/dom/interfaces.html` ?

### DIFF
--- a/html/infrastructure/common-dom-interfaces/collections/domstringlist-interface.html
+++ b/html/infrastructure/common-dom-interfaces/collections/domstringlist-interface.html
@@ -1,0 +1,32 @@
+<!doctype html>
+<meta charset=utf-8>
+<title>DOMStringList IDL tests</title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<script src=/resources/WebIDLParser.js></script>
+<script src=/resources/idlharness.js></script>
+
+<h1>DOMStringList IDL tests</h1>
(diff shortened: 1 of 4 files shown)


html/infrastructure/common-dom-interfaces/collections/domstringlist-interface.html
```diff
@@ -0,0 +1,32 @@
...
+  request.open("GET", "domstringlist.idl");
+  request.send();
+  request.onload = t.step_func(function() {
+    var idlArray = new IdlArray();
+    var idls = request.responseText;
+
+    idlArray.add_idls(idls);
+
+    idlArray.add_objects({
+      DOMStringList: [],
```
zcorpan: Use 'location.ancestorOrigins' here?

zcorpan:
w3c-test:mirror

zcorpan (approved):
LGTM % 1 comment

inexorabletash:
Also added the missing `[Exposed=(Window,Worker)]` so the worker test actually does something. (Chrome is missing that `[Exposed]` per https://github.com/w3c/IndexedDB/issues/118 so it fails all over the place, as expected)

PR #4541, opened 2017-01-16
Add tests for slot.assignedElements() method

cyrilletuzi:
As requested in whatwg/html#2269

### DIFF
--- a/html/dom/interfaces.html
+++ b/html/dom/interfaces.html
@@ -1856,6 +1856,7 @@ interface HTMLTemplateElement : HTMLElement {
 interface HTMLSlotElement : HTMLElement {
   /*[CEReactions]*/ attribute DOMString name;
   sequence<Node> assignedNodes(optional AssignedNodesOptions options);
+  sequence<Element> assignedElements(optional AssignedNodesOptions options);
 };
 
 dictionary AssignedNodesOptions {
--- a/shadow-dom/HTMLSlotElement-interface.html
+++ b/shadow-dom/HTMLSlotElement-interface.html
@@ -33,7 +33,7 @@ test(function () {
     assert_equals(slotElement.getAttribute('name'), 'bar', '"name" attribute must update the "name" content attribute');
 }, '"name" attribute on HTMLSlotElement must reflect "name" attribute');
 
-function testSlotOutsideShadowTree(options)
+function testSlotNodesOutsideShadowTree(options)
 {
     test(function () {
         assert_true('assignedNodes' in HTMLSlotElement.prototype, '"assignedNodes" method must be defined on HTMLSlotElement.prototype');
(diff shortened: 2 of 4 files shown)


hayatoito:
Hmm. I think we do not want to duplicate every tests where assignedNodes() is used. We wouldn't get benefits from this kind of duplicated tests.

You might want to add a minimum test for assignedElements(), instead of duplicating every tests which use assignedNodes().

cyrilletuzi:
I followed @annevk instructions in whatwg/html/#2269

Which makes sense as `assignedElements()` will be considered as ShadowDOM v2, so previous test with only `assignedNodes()` should continue to succeed while the new ones with `assignedElements()` should fail until implementation.

But if it's better to group them, just tell me, I just need to know who has the final word on this.

cyrilletuzi:
(Just a note to mention that previous tests for `assignedNodes()` should have raised a problem, as all tests needed to do `removeWhiteSpaceOnlyTextNodes()`, which is what is solved by `assignedElements()`.)

annevk:
I tend to disagree with @hayatoito. Tests are rather cheap and it's better to be exhaustive since you don't know what implementation decisions everyone will make. If at some point it turns out that one of these methods needs some kind of fast path due to wide usage, having tests for all scenarios in both will help catch subtle bugs.

hayatoito:
Okay. We should judge this kind of *exhaustive*-ness case by case, however, in this particular case, I am not a fan of this kind of duplications. I got an impression that this duplication happened only because the usage of `assignedNodes()` were *spotted* in the original tests. Is my understanding correct? Is there any existing test which uses `assignedNodes()` API, but it was not *duplicated*, with a reason?

Now we have doubled the number of tests. Then, later, if we add yet another new API, we would double the number again, having 4 x N tests? I think we want to avoid this kind of exponential-ness so that tests can be maintenance-friendly.

Could we spend more efforts to reduce this kind of explosion by making tests more orthogonal, or having more maintenance-friendly way? For example, I think @rniwa parameterized some tests so that we can test both cases, open or closed, without *copy and pasting piece of code*.

Can we do that? or is that too early optimization? Maybe I am worrying too much on this kind of duplication. :(

@rniwa, WDYT? Is this okay for you?

cyrilletuzi:
There are some tests which add text node child to check they are filtered by `assignedElements()`. But most of them are just duplicated.

If we group tests, we will need to be careful to do `assignedElements()` tests before calling `removeWhiteSpaceOnlyTextNodes()`.

hayatoito:
@cyrilletuzi Thank you. I understand your intention. I appreciate your efforts.  :)

I prefer non-duplication and more orthogonal tests, but If @annevk and @rniwa is okay to this change, please merge this without waiting me.

I might try to refactor tests so that they can be more maintenance-friendly, but that can be done later.

annevk:
Ah, I misunderstood your concern @hayatoito. Deduplication of code does seem worthwhile, even in tests. You would still get 4 * N tests, but the amount of changes needed for that will be lower (just update an array and maybe some conditionals).

cyrilletuzi:
Any news ? Merge or regroup ?

annevk:
So I think the main question is whether you think it's feasible to parameterize the existing assignedNodes tests so that they can be reused for assignedElements without as much duplication as the current patch.

rniwa:
Oh yeah, we certainly don't want to duplicate tests like this. We should at minimum extract helper functions that create trees to test things.

Note that I usually prefer not having logics for assertions since I prefer test assertions being duplicated over test assertions dependent on complicated logic prone to bugs.

annevk:
@cyrilletuzi are you still interested in working on this?

cyrilletuzi:
Yes but I need a clear direction.

annevk:
That is fair. What I'd do is something like
```
["assignedNodes", "assignedElements"].forEach(prop => {
   test(...);

   ...
}
```
as a wrapper around the existing tests. Perhaps with slightly more complex elements in the array depending on all the parts of the tests that need templating.

@hayatoito should probably confirm that's acceptable, but I suspect it is.

hayatoito:
> ```["assignedNodes", "assignedElements"].forEach(```

Yeah, that's one of the ways to remove code duplication, which is acceptable.

PR #4542, opened 2017-01-16
Presentation API: Sandboxing test on PresentationRequest constructor

tidoust:
User agents show now check sandboxing flags when PresentationRequest objects are constructed.

This update also drops sandboxing tests that were checking failures to call "start", "reconnect", and "getAvailability" in the absence of the "allow-presentation" sandbox attribute, because there should be no longer any way to reach these steps. Such scripts will simply always fail to create PresentationRequest objects.


tidoust:
@tomoyukilabs or @louaybassbouss Could you review this pull request?

tomoyukilabs (approved):
@tidoust This looks good to me.

PR #4543, opened 2017-01-16
Update "stability" scripts to use Google Chrome instead of Chromium

jugglinmike:
Note that this changeset includes a meaningless test modification for
demonstration purposes. That should probably be removed prior to merging
(although leaving it as-is should be harmless).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 dist: trusty
+sudo: required
 language: python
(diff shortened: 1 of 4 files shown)


jugglinmike:
Re-submitted as gh-4544 in order to satisfy TravisCI's requirements for "trusted builds."

PR #4544, opened 2017-01-16
Update "stability" scripts to use Google Chrome instead of Chromium

jugglinmike:
Note that this changeset includes a meaningless test modification for
demonstration purposes. That should probably be removed prior to merging
(although leaving it as-is should be harmless).

### DIFF
--- a/ci_stability.sh
+++ b/ci_stability.sh
@@ -20,6 +20,18 @@ hosts_fixup() {
     echo "travis_fold:end:hosts_fixup"
 }
 
+install_chrome() {
+    deb_archive=google-chrome-stable_current_amd64.deb
+    wget https://dl.google.com/linux/direct/$deb_archive
+
+    # Installation will fail in cases where the package has unmet dependencies.
+    # When this occurs, attempt to use the system package manager to fetch the
+    # required packages and retry.
+    if ! sudo dpkg --install $deb_archive; then
+      sudo apt-get install --fix-broken
+      sudo dpkg --install $deb_archive
+    fi
+}
 
 test_stability() {
     python check_stability.py $PRODUCT
@@ -27,6 +39,9 @@ test_stability() {
 
 main() {
     hosts_fixup
+    if [ "$PRODUCT" == "chrome" ]; then
+        install_chrome
+    fi
     test_stability
 }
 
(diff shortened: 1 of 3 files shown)


ci_stability.sh
```diff
@@ -20,13 +20,24 @@ hosts_fixup() {
     echo "travis_fold:end:hosts_fixup"
 }
 
+install_chrome() {
+    channel=$1
+    deb_archive=google-chrome-${channel}_current_amd64.deb
+    wget https://dl.google.com/linux/direct/$deb_archive
+    sudo dpkg --install $deb_archive || true
+    sudo apt-get install --fix-broken
```
jgraham: It's not quite clear why this is needed. Leave a comment, perhaps?

url/url-origin.html
```diff
@@ -1,4 +1,4 @@
-<!doctype html>
+<!DOCTYPE html>
```
jgraham: Remove this change.

ci_stability.sh
```diff
@@ -20,13 +20,24 @@ hosts_fixup() {
...
+}
 
 test_stability() {
     python check_stability.py $PRODUCT
 }
 
 main() {
     hosts_fixup
+    if [ $(echo $PRODUCT | grep '^chrome:') ]; then
+       install_chrome $(echo $PRODUCT | grep --only-matching '\w\+$')
```
jgraham: I guess I might have used two variables rather thing string manipulation, but this is OK for now.

ci_stability.sh
jugglinmike: Sure. I've also refactored this with an `if` statement to make the relationship between the commands more clear.

url/url-origin.html
jugglinmike: Done

jgraham:
Mind gunning `git rebase --autosquash origin/master` before landing?

jgraham (approved):

jugglinmike:
Sure thing. The prior version of this branch is available at [jugglinmike/chrome-presquash](https://github.com/w3c/web-platform-tests/compare/73f4ad361ca56d890b5f46d62c3bc84371a77e10...jugglinmike:chrome-presquash) for posterity.

PR #4555, opened 2017-01-18
Attempt to fix some intermittant CORS tests

annevk:


annevk:
It seems that Travis does not notice it, but I do get timeouts on w3c-test.org...

annevk:
Maybe w3c-test.org is just really slow? There are a number of Gecko issues open too thought intermittent CORS tests.

PR #4558, opened 2017-01-18
Extend stability bot to report browser version

jugglinmike:


jugglinmike:
@rbyers The reported information for Chrome isn't quite as specific as what you requested in [our "project brainstorming" document](https://docs.google.com/document/d/10oGzrrukzuVqMydDYyDXsHi5hhjuP5OOEsT1ANTOJVo/edit): it doesn't include a revision SHA.  Now that we're using Chrome instead of Chromium, I'm not sure if we can access that information programatically (I can only find it in `chrome://version`). Do you think [reporting the version number alone](https://github.com/w3c/web-platform-tests/pull/4558#issuecomment-273510929) is acceptable?

RByers:
Nice!
>  Do you think reporting the version number alone is acceptable?

Yes, for official Chrome builds the verison number carries all the details (you can map it to a commit position hash with tools like the one at the bottom [here](https://omahaproxy.appspot.com/)).  It was only for ToT builds where we needed something more than just the version number to uniquely identify the build.

RByers (approved):
Changes look good to me, but I don't have any history with this code.  Perhaps @jgraham will want to review before we push?

jgraham:
I'm totally happy to merge this, if you remove the meaningless test commit, but for Fx nightly the version number doesn't tell you much (more or less just that it is nightly). I think you get better information by reading the `platform.ini` file in the firefox directory which contains both a build timestamp and a source SHA1.

jugglinmike:
Thanks for the tip, @jgraham! I have no insight about how stable the location or format of that file is, though I know you wouldn't make that recommendation unless it was "reasonably stable." Still, it occurred to me that it might not be intended for programmatic consumption, and then I realized the same was true for the output of `chrome --version`. It would be a shame to fail a build due to a change in an implementation detail like how version information is published. So in addition to using `platform.ini` to identify the installed version of Firefox, I've added a bit of logic to report errors for any "retrieve version" operation while continuing the build.

If this looks good to you, I'm happy to squash the relevant changes and remove the irrelevant one--let me know.

jgraham (approved):
Subject to removing the test change.

jugglinmike:
All set

PR #4561, opened 2017-01-19
selection: Split collapse.html into 2, and extend.html into 3.

tkent-google:
The text dump of collapse.html and extend.html result exceed the file size limit
in Chromium project.

This is similar to:
https://github.com/w3c/web-platform-tests/pull/4049

### DIFF
--- a/selection/collapse-00.html
+++ b/selection/collapse-00.html
@@ -1,0 +1,14 @@
+<!doctype html>
+<title>Selection.collapse() tests</title>
+<meta name=timeout content=long>
+<div id=log></div>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<script src=common.js></script>
+<script src=collapse.js></script>
+<script>
+"use strict";
+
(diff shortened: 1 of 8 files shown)


TakayoshiKochi:
Test ranges for each test is 30 for collapse and 20 for extend.
This looks somewhat arbitrary, though it should be with reason.

I'm not sure if any test cases are added or removed, but I foresee someone
who touches this test may wonder any criteria for the split length.

The `testRagnes` is defined in `common.js`, and referenced from tests other than
collapse/extend. How about adding a comment on `testRanges` so that if the list
grows/shrinks a test author should take care of?  Probably reference to this PR
should be fine.

// The length of testRanges caused a problem that results got too long, and
// some tests (collapse, extend) split the testRange into pieces.
// See https://github.com/w3c/web-platform-tests/pull/4561

tkent-google:
@TakayoshiKochi , I added a comment to common.js

TakayoshiKochi:
Thanks!  The comment looks good.

LGTM

PR #4563, opened 2017-01-19
Make sure stable sort is used for URLSearchParams's sort()

annevk:
See https://bugs.webkit.org/show_bug.cgi?id=167154#c2 for context.


annevk:
@achristensen07 I tried to attribute the commit to you since you created the test, but it appears you don't have your Apple email address associated with your GitHub account.

PR #4565, opened 2017-01-19
Streams: Test pipeTo with undefined rejections

ricea:
When the source or destination returns a rejection the promise returned
from ReadableStream.pipeTo should also reject. Verify that this works
correctly when the value of the rejection is undefined.

### DIFF
--- a/streams/piping/general.js
+++ b/streams/piping/general.js
@@ -163,7 +163,10 @@ for (const preventAbort of [true, false]) {
       }
     }, { preventAbort });
 
-    return promise_rejects(t, null, rs.pipeTo(new WritableStream()), 'promise should be rejected');
+    return rs.pipeTo(new WritableStream()).then(
+        () => assert_unreached('pipeTo promise should be rejected'),
+        value => assert_equals(value, undefined, 'rejection value should be undefined'));
+
   }, `an undefined rejection from pull should cause pipeTo() to reject when preventAbort is ${preventAbort}`);
 }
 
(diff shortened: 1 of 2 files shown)


ricea:
@domenic PTAL

streams/piping/general.js
```diff
@@ -154,4 +154,37 @@ promise_test(() => {
...
+for (const preventAbort of [true, false]) {
+  promise_test(t => {
+
+    const rs = new ReadableStream({
+      pull() {
+        return Promise.reject(undefined);
+      }
+    }, { preventAbort });
+
+    return promise_rejects(t, null, rs.pipeTo(new WritableStream()), 'promise should be rejected');
```
domenic: Hmm. I think this will allow anything through with no `name` property? Until https://github.com/w3c/testharness.js/issues/165 is landed this might be a good case for actually reverting to the assert_unreached + assert_equals combo to test that the rejection reason is actually undefined.

domenic (changes requested):
One tweak for better coverage. Also, you might want to add yourself to the OWNERS file.

streams/piping/general.js
ricea: I felt it was unnatural to put `null` in there but I concluded that it was the fact that it is rejecting at all that was most important.

I've switched to assert_unreached + assert_equals.

streams/piping/general.js
```diff
@@ -154,4 +154,42 @@ promise_test(() => {
...
 }, 'Piping from a ReadableStream for which a chunk becomes asynchronously readable after the pipeTo');
 
+for (const preventAbort of [true, false]) {
+  promise_test(t => {
+
+    const rs = new ReadableStream({
+      pull() {
+        return Promise.reject(undefined);
+      }
+    }, { preventAbort });
```
tyoshino: preventAbort passed to the ReadableStream constructor. to pipeTo() below?

streams/piping/general.js
```diff
@@ -154,4 +154,42 @@ promise_test(() => {
...
+}
+
+for (const preventCancel of [true, false]) {
+  promise_test(t => {
+
+    const rs = new ReadableStream({
+      pull(controller) {
+        controller.enqueue(0);
+      }
+    }, { preventCancel });
```
tyoshino: ditto

ricea: That's embarrassing. #4613.

PR #4566, opened 2017-01-19
make travis directories dynamic

bobholt:
This dynamically sets the directories used in the `check_stability.py` script to be based on the user running the Travis job.

This allows anyone to set up Travis builds on their fork of the web platform tests without modifying `check_stability.py`.

This opts to pass the user value to each method rather than setting a global value. Happy to change if a global value is preferred.

### DIFF
--- a/check_stability.py
+++ b/check_stability.py
@@ -27,7 +27,8 @@ TbplFormatter = None
 reader = None
 wptcommandline = None
 wptrunner = None
-user = None
+wpt_root = None
+wptrunner_root = None
 
 logger = logging.getLogger(os.path.splitext(__file__)[0])
 
@@ -188,7 +189,7 @@ class Firefox(Browser):
     product = "firefox"
 
     def install(self):
-        call("pip", "install", "-r", "%s/wptrunner/requirements_firefox.txt" % (user))
+        call("pip", "install", "-r", os.path.join(wptrunner_root, "requirements_firefox.txt"))
         resp = get("https://archive.mozilla.org/pub/firefox/nightly/latest-mozilla-central/firefox-53.0a1.en-US.linux-x86_64.tar.bz2")
         untar(resp.raw)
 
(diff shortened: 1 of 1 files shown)


jgraham:
You might need to rebase onto latest master for this to work.

jgraham:
Seems OK in principle, but it's pretty noisy. Maybe we should make the base path a global which we write to once?

bobholt:
@jgraham Agree. I'll make an update.

bobholt:
Latest push makes that user variable global. Will go further and make that whole path global now.

jgraham (approved):
Can you just demonstrate this works by pushing a no-op test change, and then removing it again once the checker reports some results?

check_stability.py
```diff
@@ -606,9 +606,8 @@ def main():
     parser = get_parser()
     args = parser.parse_args()
 
-    user_path = os.path.join(os.path.abspath(os.curdir), args.user)
-    wpt_root = os.path.join(user_path, "web-platform-tests")
-    wptrunner_root = os.path.join(user_path, "wptrunner")
+    wpt_root = os.path.join(os.path.abspath(os.curdir))
+    wptrunner_root = os.path.join(wpt_root, "..", "wptrunner")
```
jgraham: I think a normalized path here would be nice.

jgraham (changes requested):

check_stability.py
bobholt: Just had that same thought and pushed another change - rebased because I was getting so many fixups.

bobholt:
Please bear with me as I work through my python rust. If it's too noisy, I can close and reopen once the light is green on my fork.

jgraham (approved):

jgraham:
I don't know what the problem reporting the travis status is, but the build passes and tests run, so I'm happy with this as soon as you remove the example test change.

bobholt:
Yeah, that status was causing me great consternation yesterday. The test change has been removed.

PR #4571, opened 2017-01-19
Add movementxy attribute test for pointerevents

### LINKED ISSUE #567: Incorrect movementX/Y in PointerEvents
It seems that non of the implementations of Chrome and Edge conforms with the definition of movementX/Y attributes and they are always 0 in those implementations for PointerEvents as oppose to compat MouseEvents.

https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/movementX
https://w3c.github.io/pointerlock/#widl-MouseEvent-movementX

I was wondering since PointerEvents are MouseEvents do we need to do anything special about these fields? What is the expectation here?

NavidZ:
@mustaqahmed @dtapuska 

closes w3c/pointerevents#567


RByers:
w3c-test:mirror  (want to try the test out on w3c-test.org)
One initial nit: rather than a generic "interaction-with-other-specs" directory can you just use the name of the spec - "pointerlock"?  That's the pattern we've already started with the "compat" sub-directory.

RByers:
Test is [now served here](https://w3c-test.org/submissions/4571/pointerevents/interaction-with-other-specs/pointerevent_movementxy-manual.html).  I can't get it to complete on current Chrome dev channel -
 the mouse one completes, but clicking 'skip' or using touch I still get "1 remain".

I was initially confused about the iframe and requirement that movement is kept to <10 pixels per event.  This is because the events that land over the frame are lost, right?  This is a same-origin iframe so it should be easy to also have listeners inside the frame which just call into some common code in the top document so that you can test both the events delivered to the top document and those delivered inside the frame.

Big picture, what I was wondering was - is it better to rely on relatively precise positioning like this (move from red to black), and/or should our test just verify that the movement sum exactly equals the change in client (or maybe screen) co-ordinates the way [the pointerlock tests do](https://github.com/w3c/web-platform-tests/blob/master/pointerlock/movementX_Y_basic-manual.html)?

/cc @scheib who may have input.  From a quick glance I didn't see any movementX/Y tests there which look explicitly at the iframe case (but maybe it was only our originally flawed PointerEvents implementation that would even suggest that such a test was worthwhile).

NavidZ:
We don't have any further sub-directory under compat though. I thought I just put all the related interaction with other spec tests under here like Shadow dom that will come later. If I were to create a folder for each some of those folders will have only one test. Having said that which one do you think is better? Create a separate subfolder under pointerevents for each other spec we would like to add test?

I do have a handler for the inner iframe as well and I also check the events from that (i.e. innerSumX vs sumX). The reason that I wanted the movements to be small is to avoid the jump when you cross the boundary of an iframe discussed in this [issue](https://github.com/w3c/pointerlock/issues/16) on PointerLock. So I can actually sum things up inside the frame and outside the frame and be able to rely on the range of the result.

RByers:
> We don't have any further sub-directory under compat though

Right, `compat` is the name of the spec (compat.spec.whatwg.org).  I only expect the single test there for the forseable future.  We should have the spec name in the path somewhere, but as long as we're consistent with the pattern I don't really care.  Eg: we could have:
```
compat/pointerevent_touch-action_two-finger_interaction-manual.html
pointerlock/pointerevent_movementxy-manual.html
```
or
```
interaction-with-other-specs/compat-pointerevent_touch-action_two-finger_interaction-manual.html
interaction-with-other-specs/pointerlock-pointerevent_movementxy-manual.html
```

I have a slight preference for the former, but as long as we're not mixing different patterns I don't really care.

RByers:
> I do have a handler for the inner iframe as well and I also check the events from that (i.e. innerSumX vs sumX). The reason that I wanted the movements to be small is to avoid the jump when you cross the boundary of an iframe discussed in this issue on PointerLock. So I can actually sum things up inside the frame and outside the frame and be able to rely on the range of the result.

I see, thanks.  See [my comment here](https://github.com/w3c/pointerlock/issues/16#issuecomment-275151076).  I don't think it makes sense to try to test the movementX/Y for the first event inside an iframe without communicating with the outer frame in order to measure the real delta.

RByers:
w3c-test:mirror

pointerevents/pointerlock/pointerevent_movementxy-manual.html
```diff
@@ -0,0 +1,103 @@
...
+                      }
+                      isFirstMove = false;
+                      lastClientX = event.clientX;
+                      lastClientY = event.clientY;
+                    }
+                  });
+                });
+                ['pointerenter', 'pointerleave'].forEach(function(eventName) {
+                  on_event(document.getElementById('innerFrame').contentDocument, eventName, function (event) {
+                    isFirstMove = true;
```
RByers: I see, so this means we're not testing the movementX/movementY values for the event that enters/leaves the iframe.  That's OK but not ideal (misses some interesting test coverage).

The simplest way to get this additional test coverage is to use `screenX/Y` instead of `clientX/Y` (and remove all the `isFirstMove` logic).  That would also better match the spec, since `movementX/Y` is [defined in terms of screen co-ordinates](https://w3c.github.io/pointerlock/#dom-mouseevent-movementx).  Eg. if the browser content area happened to move relative to the physical screen during the test (such as when top controls show or hide in mobile browsers) then the test should still pass when using `screen` co-ordinates but should fail when using `client` (assuming we've implemented according to the spec).

If there's some reason why using client co-ordinates is preferable (eg. perhaps @schieb has a reason why the existing pointerlock tests do that), you could still get this additional test coverage as follows:
 - specify the iframe border size explicitly (with either the 'border=' HTML attribute or a CSS border rule on the iframe element) rather than relying on the browser default (not sure that's standardized).
 - on test start, compute the offset of the iframe within the parent frame by using getBoundingClientRect on innerFrame plus the border size you specified
 - in your pointermove handler, if the event target is inside the iframe, add the above offset to the client co-ordinates

In that way you'd be doing all your math in the ["root frame" coordinate space](https://www.chromium.org/developers/design-documents/blink-coordinate-spaces) (instead of in the client / local-frame co-ordinate space).  And then the math should work out in all cases.

schieb: @scheib ^

scheib: screenX/Y are good, the use of clientX/Y in other tests was an oversight. I've filed an issue to fix the pointerlock tests #4630.

RByers: Perfect, thanks!  Sorry for the mis-spell ;-)

NavidZ:
Now it looks a lot better using screenX/Y. I was able to get rid of all that isFirstMove/enter/leave logic. Thanks. ptal.

RByers:
w3c-test:mirror

RByers:
Yeah it looks great - thanks!  Just wanted to try it out again myself.

On Chrome 55 desktop (the only laptop I have handy in the meeting I'm sitting in) the mouse case fails the ASSERT.  Is there a known bug for mouse in Chrome 55?  I thought the only known bug was with touch?

I also tried Chrome 57 android but it seems the test won't work on a phone - it doesn't fit and you apparently can't scroll the page.  It's probably not worth a big effort now (eg. could be a mobile-friendly-cleanup we do later), but do you think it's worth seeing if you can tweak the layout / touch-action a bit to make this test passable on a phone?

NavidZ:
w3c-test:mirror

NavidZ:
Layout is fixed now. Chrome 55 does have the bug. That is the intention of this test. Hopefully it will be fixed in the latest version.

Ms2ger:
@RByers is this good to go now?

RByers:
w3c-test:mirror

RByers:
LGTM, thanks!

smaug----:
So this was wrong. PointerLock spec explicitly states that "movementX/Y must be zero for all mouse events except mousemove." We need a spec change, as far as I see.

PR #4572, opened 2017-01-19
Streams: add some pipeThrough tests

domenic:
This includes tests that the pipeThrough promise is handled; see https://github.com/whatwg/streams/pull/664.

### DIFF
--- a/streams/piping/pipe-through.js
+++ b/streams/piping/pipe-through.js
@@ -3,7 +3,11 @@
 if (self.importScripts) {
   self.importScripts('/resources/testharness.js');
   self.importScripts('../resources/rs-utils.js');
+  self.importScripts('../resources/test-utils.js');
 }
+
+const error1 = new Error('error1!');
+error1.name = 'error1';
 
 function duckTypedPassThroughTransform() {
   let enqueueInReadable;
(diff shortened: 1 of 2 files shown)


streams/piping/pipe-through.js
```diff
@@ -36,4 +40,55 @@ promise_test(() => {
...
+
+  assert_array_equals(calledWithArgs, [fakeWritable, arg2],
+    'The this value\'s pipeTo method should be called with the appropriate arguments');
+
+}, 'pipeThrough generically calls pipeTo with the appropriate args');
+
+test(() => {
+  const dummy = {
+    pipeTo(args) {
+      return { not: 'a promise' };
```
ricea: I was wondering about having a test for a thenable. However, looking at the change to the standard text, there doesn't seem to be a lot of point. As a compromise, how about making this object a thenable? It won't weaken the test, and has a tiny possibility of finding someone's bug.

streams/piping/pipe-through.js
```diff
@@ -36,4 +40,55 @@ promise_test(() => {
     assert_array_equals(chunks, [1, 2, 3, 4, 5]), 'chunks should match');
 }, 'Piping through a duck-typed pass-through transform stream should work');
 
+promise_test(t => {
+  const transform = {
+    writable: new WritableStream({
+      start(c) {
+        c.error(error1);
```
ricea: IIUC, the main purpose of `error1` here is to help identify the source of the problem if the test fails. In which case, wouldn't using a value like `'this rejection should not be reported as unhandled'` be even better?

ricea (approved):
Nice tests.

streams/piping/pipe-through.js
tyoshino: has this been addressed?

PR #4574, opened 2017-01-19
Add WebDriver Tests for Finding Elements

thejohnjansen:
Adding tests for findElement and findElements

### DIFF
--- a/webdriver/find_element/find_element_tests.py
+++ b/webdriver/find_element/find_element_tests.py
@@ -7,27 +7,27 @@ import base_test
 
 class FindElementTests(base_test.WebDriverBaseTest):
     def test_find_element_by_name(self):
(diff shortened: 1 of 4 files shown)


andreastt (changes requested):
This test isn’t using the wdspec test harness we integrated into wptrunner, and thus isn’t picked up by wptrunner.

I have a few general comments:

- Please convert the test to pytest following the example set out by the other files in /webdriver/*.py.
- Avoid using a single .html document for all tests. We want individual tests to be self-contained to the largest possible extent.
- Please test the protocol level formalities of the Find Element command first. This just tests that the behaviour is correct and relies too heavily on the client.

thejohnjansen:
I am going to close this after conversation in email and start again after talking with Sam.

PR #4575, opened 2017-01-19
Ensure stability bot uses intended Chrome version

jugglinmike:
By default, the TravisCI environment provides a stable build of the Google Chrome web browser. The globally-installed `google-chrome` binary from this installation takes precedence over the version introduced via the stability bot's installation script, and as a result, the provided "stable" release was always used by the bot.

Use the Debian `update-alternatives` tool to ensure that previously-installed binaries are ignored and that the automated stability checks run against the intended version of the browser. Update the "version number" pattern to match the new version format.

### DIFF
--- a/check_stability.py
+++ b/check_stability.py
@@ -264,7 +264,7 @@ class Chrome(Browser):
     def version(self, root):
         """Retrieve the release version of the installed browser."""
         output = call(self.binary, "--version")
-        return re.search(r"[0-9a-z\.]+$", output.strip()).group(0)
+        return re.search(r"[0-9\.]+( [a-z]+)?$", output.strip()).group(0)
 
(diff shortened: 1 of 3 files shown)


jugglinmike:
Compare the browser version in the above report for Chrome with [the version in the report generated in gh-4558](https://github.com/w3c/web-platform-tests/pull/4558#issuecomment-273510929):

> ## Chrome (unstable channel) ##
>
> Testing web-platform-tests at revision 1bee1eb
> Using browser at version 55.0.2883.75
> Starting 10 test iterations
> All results were stable

foolip:
lgtm, but I think @jgraham understands the code for realz.

jgraham:
Good catch. Let's remove the meaningless change though.

PR #4584, opened 2017-01-20
Stop streams/w-s/write.https.html timeouts

ricea:
write.https.html times out on some Chromium bots. Reduce the number of
iterations in the 'a large queue of writes should be processed
completely' test to make the test run faster.

Fixes whatwg/streams#660.


ricea:
@tyoshino, ptal.

domenic:
Since an equivalent PR was approved by @tyoshino over in https://github.com/whatwg/streams/pull/663 I will go ahead and merge this.

tyoshino (commented):
Yes, lgtm. Thanks

PR #4585, opened 2017-01-20
Reduce number of requests in header-values

annevk:
This makes it a little harder to study the results, but I guess that’s
fine as it runs a lot quicker.


annevk:
See https://github.com/w3c/web-platform-tests/pull/4525 for context by the way.

jdm:
This seems like a reasonable compromise.

PR #4589, opened 2017-01-20
Require HTTPS for the correct streams tests

domenic:
Previously, HTTPS was being required for the window-context tests, and not for the service-worker-context tests. This reverses that requirement.

Note that when we update the whatwg/streams submodule, we'll need to change [this line in the test runner](https://github.com/whatwg/streams/blob/94140088224c1a0e219c630378d1fa7985d657a9/reference-implementation/run-web-platform-tests.js#L20) to be a bit more sophisticated.


domenic:
Ping @tyoshino @ricea, this can probably be rubber-stamped.

ricea:
lgtm

PR #4603, opened 2017-01-23
Update the stability checker to deal with new manifest format.

jgraham:
Support files being in the manifest meant that all support files
were being skipped in stability checks. Also take the oppertunity
to ensure that we read each test file exactly once, and make some
other optimistations.

### DIFF
--- a/check_stability.py
+++ b/check_stability.py
@@ -404,52 +404,62 @@ def get_files_changed():
     if not files:
         return []
     assert files[-1] == "\0"
-    return ["%s/%s" % (wpt_root, item)
+    return [os.path.join(wpt_root, item)
             for item in files[:-1].split("\0")]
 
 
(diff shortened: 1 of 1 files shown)


check_stability.py
```diff
@@ -404,52 +404,63 @@ def get_files_changed():
...
+        rel_root = os.path.relpath(root, wpt_root)
+        if root == wpt_root:
+            for dir_name in skip_dirs:
+                dirs.remove(dir_name)
+        for fname in fnames:
+            test_full_path = os.path.join(root, fname)
+            # Skip any file that's not a test file.
+            if test_full_path not in test_files:
+                continue
+            with open(test_full_path, "r") as fh:
```
Ms2ger: "rb" while you're here

sideshowbarker (approved):

PR #4604, opened 2017-01-23
Modify and Add tests to OffscreenCanvas promise-based commit()

xlai-o:
@junov: Please review. Thanks!

This pull request mainly modifies the OffscreenCanvas.commit() tests to handle the latest change in [specification about a promise-based commit() in OffscreenCanvas](https://wiki.whatwg.org/wiki/OffscreenCanvas.requestAnimationFrame). The changes include:
1. Modify the two existing tests in offscreencanvas.commit.html and offscreencanvas.commit.w.html to promise_tests. In particular, we test the assert pixels after the promise returned by commit() is resolved.
2. Add one more test that assert the pixels in OffscreenCanvas to be non-zero after commit(). This is an adaptation of [an existing layout test in Chromium](https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/fast/canvas/OffscreenCanvas-commit-retains-backing.html).

### DIFF
--- a/common/canvas-tests.js
+++ b/common/canvas-tests.js
@@ -35,10 +35,7 @@ function _getPixel(canvas, x,y)
 function _assertPixel(canvas, x,y, r,g,b,a, pos, colour)
 {
     var c = _getPixel(canvas, x,y);
-    assert_equals(c[0], r, 'Red channel of the pixel at (' + x + ', ' + y + ')');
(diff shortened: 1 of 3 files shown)


junov:
Theses tests are great, but let's wait for the promise-returning-commit to be added to the whatwg spec before landing this.  There's a conf call tomorrow with participants of the WebGL working group where we'll discuss this.  Once multi-vendor support is established, we'll move forward with the spec change, then we can land these tests.

PR #4607, opened 2017-01-24
Upstream IDBCursor.continuePrimaryKey exception test to WPT.


jeffcarp:
BUG=683463

Review-Url: https://codereview.chromium.org/2652463002
Cr-Commit-Position: refs/heads/master@{#445616}

### DIFF
--- a/IndexedDB/idbcursor-continuePrimaryKey-exceptions.htm
+++ b/IndexedDB/idbcursor-continuePrimaryKey-exceptions.htm
@@ -1,0 +1,99 @@
+<!doctype html>
+<meta charset="utf-8" />
+<title>IndexedDB: IDBCursor continuePrimaryKey() exception throwing</title>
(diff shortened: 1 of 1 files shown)


inexorabletash:
Hey @RByers @foolip - check this out! 

(Fails in Chrome because the feature is still experimental.)

foolip:
That seems OK, but I guess it might be confusing? Trying to use this process for not-yet-enabled features won't work, but I suppose we could have a separate run with experimental features enabled?

inexorabletash:
I should have made my point clearer - I was dancing with excitement about the process being (semi)functional and awesome. :)

But yeah, maybe we do want a run with experimental features enabled.

foolip:
Ah, consider yourself a beta tester! This process will be announced for general usage real soon now, thanks for helping test it, and feel free to keep at it :)

PR #4612, opened 2017-01-24
Add tests for body property handling in Request constructor

yutakahirano:
This change is for https://github.com/whatwg/fetch/pull/425 and https://github.com/whatwg/fetch/pull/458.

### DIFF
--- a/fetch/api/request/request-disturbed.html
+++ b/fetch/api/request/request-disturbed.html
@@ -16,8 +16,17 @@
       };
 
       var noBodyConsumed = new Request("");
-      noBodyConsumed.blob();
       var bodyConsumed = new Request("", initValuesDict);
+
+      test(() => {
+        assert_equals(noBodyConsumed.body, null, "body's default value is null");
+        assert_false(noBodyConsumed.bodyUsed , "bodyUsed is false when request is not disturbed");
+        assert_not_equals(bodyConsumed.body, null, "non-null body");
+        assert_true(bodyConsumed.body instanceof ReadableStream, "non-null body type");
+        assert_false(noBodyConsumed.bodyUsed, "bodyUsed is false when request is not disturbed");
+      }, "Request's body: initial state");
+
+      noBodyConsumed.blob();
       bodyConsumed.blob();
 
       test(function() {
@@ -39,7 +48,7 @@
         assert_throws(new TypeError(), function() { new Request(bodyConsumed); });
       }, "Check creating a new request from a disturbed request");
 
-      test(function() {
+      promise_test(function() {
         var bodyRequest = new Request("", initValuesDict);
         const originalBody = bodyRequest.body;
         assert_false(bodyRequest.bodyUsed , "bodyUsed is false when request is not disturbed");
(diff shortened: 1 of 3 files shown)


jdm:
These new tests look good to me. I'm not going to merge them until someone else says that they're testing the cases we care about for the fetch PRs mentioned in the original comment.

yutakahirano:
cc: @annevk

fetch/api/request/request-disturbed.html
```diff
@@ -41,11 +41,30 @@
...
 
+      promise_test(() => {
+        const bodyRequest = new Request("", initValuesDict);
+        const originalBody = bodyRequest.body;
+        assert_false(bodyRequest.bodyUsed , "bodyUsed is false when request is not disturbed");
+        const requestFromRequest = new Request(bodyRequest, { body : "init body" });
+        assert_true(bodyRequest.bodyUsed , "bodyUsed is true when request is disturbed");
+        assert_equals(bodyRequest.body, originalBody, "body should not change");
+        assert_not_equals(originalBody, undefined, "body should not be undefined");
+        assert_not_equals(originalBody, null, "body should not be null");
```
annevk: Can we add a test here about what it should be? Either instanceof/typeof or some such?

And perhaps we should also have a test where the initial request does not have a body?

fetch/api/request/request-disturbed.html
```diff
@@ -41,11 +41,30 @@
...
       test(function() {
         var bodyRequest = new Request("", initValuesDict);
+        const originalBody = bodyRequest.body;
         assert_false(bodyRequest.bodyUsed , "bodyUsed is false when request is not disturbed");
         var requestFromRequest = new Request(bodyRequest);
         assert_true(bodyRequest.bodyUsed , "bodyUsed is true when request is disturbed");
+        assert_equals(bodyRequest.body, originalBody, "body should not change");
+        assert_not_equals(originalBody, undefined, "body should not be undefined");
+        assert_not_equals(originalBody, null, "body should not be null");
       }, "Input request used for creating new request became disturbed");
```
annevk: This test is basically duplicated no as a promise_test, no? Maybe it's better to just remove it? Or am I missing something subtle?

annevk: Oh wait, the difference is that the promise_test passes a body. Disregard the above comment please.

fetch/api/request/request-disturbed.html
yutakahirano: Oops, I found that I forgot to expose body property on Request. I'll make another fetch spec PR...

yutakahirano: Anyway, I added some assertions.

annevk: Maybe just add that to the existing PR? Then we can use these tests for both.

yutakahirano: Ok, will do.

annevk: Never mind, what you did is fine.

PR #4613, opened 2017-01-24
Fix bug in streams/piping/general.js

ricea:
The test 'an undefined rejection from pull should cause pipeTo() to
reject' was failing to correctly test the case where preventAbort is
true.

The test 'an undefined rejection from write should cause pipeTo() to
reject' was failing to correctly test the case where preventCancel is
true.

Fixed.


tyoshino (approved):
lgtm!

domenic:
Thanks for catching this @tyoshino!!

PR #4614, opened 2017-01-24
XMLHttpRequest: check more in abort() during DONE

annevk:


jdm:
These changes look good according to https://fetch.spec.whatwg.org/#concept-network-error.

PR #4617, opened 2017-01-24
Tests for UTF-8 decoding in host

annevk:
See https://github.com/whatwg/url/issues/215.


domenic (commented):
Tests LGTM but the spec change that motivates them appears to be unnecessary per discussion in https://github.com/whatwg/url/pull/216. No reason not to merge them though.

PR #4618, opened 2017-01-25
Create a change in external/wpt meant to be reverted


jeffcarp:
This is to test reverted changes in the WPT Export process.

BUG=679951

Review-Url: https://codereview.chromium.org/2651533003
Cr-Commit-Position: refs/heads/master@{#445651}


Ms2ger:
@jeffcarp so can you please revert it?

PR #4623, opened 2017-01-25
Test non-IDL bits from Location and its stringifier

annevk:
See https://github.com/whatwg/html/pull/2294.

### DIFF
--- a/html/browsers/history/the-location-interface/location-stringifier.html
+++ b/html/browsers/history/the-location-interface/location-stringifier.html
@@ -13,11 +13,12 @@ test(function() {
   const prop1 = Object.getOwnPropertyDescriptor(location, "toString"),
         prop2 = Object.getOwnPropertyDescriptor(location, "href")
 
-  assert_equals(prop1.configurable, prop2.configurable)
-  assert_equals(prop1.enumerable, prop2.enumerable)
-
   assert_true(prop1.enumerable)
   assert_false(prop1.writable)
   assert_false(prop1.configurable)
+
+  assert_true(prop2.enumerable)
+  assert_false(prop2.configurable)
+  assert_equals(typeof prop2.get, "function")
 })
 </script>
--- a/html/browsers/history/the-location-interface/location-stringifier.html
+++ b/html/browsers/history/the-location-interface/location-stringifier.html
@@ -8,4 +8,16 @@
 <div id=log></div>
 <script>
 test_stringifier_attribute(location, "href", true);
+
+test(function() {
+  const prop1 = Object.getOwnPropertyDescriptor(location, "toString"),
+        prop2 = Object.getOwnPropertyDescriptor(location, "href")
+
+  assert_equals(prop1.configurable, prop2.configurable)
+  assert_equals(prop1.enumerable, prop2.enumerable)
+
+  assert_true(prop1.enumerable)
+  assert_false(prop1.writable)
+  assert_false(prop1.configurable)
+})
 </script>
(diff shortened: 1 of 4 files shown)


annevk:
@cdumez you might want to look at these. Safari does some odd things with Object.getOwnPropertyDescriptor() and Location objects.

html/browsers/history/the-location-interface/location-stringifier.html
```diff
@@ -8,4 +8,16 @@
 <div id=log></div>
 <script>
 test_stringifier_attribute(location, "href", true);
+
+test(function() {
+  const prop1 = Object.getOwnPropertyDescriptor(location, "toString"),
+        prop2 = Object.getOwnPropertyDescriptor(location, "href")
+
+  assert_equals(prop1.configurable, prop2.configurable)
```
domenic: I think this would be clearer if you just tested prop2's properties directly with assert_true/assert_false. Also it'd be good to test that `typeof prop2.get === "function"`.

annevk:
FWIW, I addressed the feedback, but given that we might want to remove toJSON again maybe we should wait with landing this?

annevk:
If we removed toJSON, would we just test it being undefined?

domenic:
You'd want to test that hasOwnProperty(location, "toJSON") is false, and probably also that it's undefined for good measure.

I'm OK with either incremental changes or both at once.

cdumez:
@annevk Thanks for letting me know. I'll take a look once this lands. I thought I had made getOwnPropertyDescriptor() work correctly for Window / Location by now :(

annevk:
@cdumez thanks, if you have any feedback on whether or not we should add toJSON that'd be great. Seems the majority of browsers don't have it and thus far we haven't found a good reason to add it.

cdumez:
@annevk I do not have a strong opinion on this. If most browsers do not have it, I would not add it to the spec. I don't think we would be very motivated in implementing it without a good use-case, especially if other browsers do not already support it.

cdumez:
FYI, I was "watching" this repo on GitHub and was getting way too much mail. I fear this may have caused me to miss people mentioning me in comments. I have just unwatched the repo so hopefully I will be more responsible to @ mentions in the future.

sideshowbarker:
> I fear this may have caused me to miss people mentioning me in comments.

Note that when you get mentioned in a comment, the e-mail notification you receive will have a Cc to `mention@noreply.github.com`. Other notifications for comments you’re not specifically mentioned in don’t have that `mention@noreply.github.com` Cc. So can search/filter your e-mail on that.

Details are at https://help.github.com/articles/about-notification-emails/#the-cc-address (which also has info about other special Cc addresses you can filter/search your e-mail notifications on).

html/browsers/history/the-location-interface/location-valueof.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<title>Location valueOf</title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<div id=log></div>
+<script>
+test(() => {
+  assert_equals(location.valueOf, Object.prototype.valueOf)
+  assert_equals(typeof location.valueOf.call(5), "object")
+  const prop = Object.getOwnPropertyDescriptor(location, "valueOf")
```
cdumez: Why do we expect valueOf to be defined on the instance here?

annevk: Answered in https://github.com/whatwg/html/pull/2294.

domenic (approved):

PR #4624, opened 2017-01-25
Add test for WebRTC data channels conformance to IDL definition

dontcallmedom:
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/4624)

### DIFF
--- a/webrtc/datachannel-idlharness.html
+++ b/webrtc/datachannel-idlharness.html
@@ -1,0 +1,75 @@
+<!doctype html>
+<!--
+This test creates a data channel object and compares it with the WebIDL defined interface
+-->
+
+<html>
+<head>
+  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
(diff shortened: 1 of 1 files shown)


dontcallmedom:
FYI quite a few failures in Firefox and Chrome:
* Chrome still needs prefix for RTCPeerConnection
* even then, doesn't seem to assign the IDL attributes as per WebIDL
* Firefox uses `DataChannel` rather than `RTCDataChannel` for the interface name, causing lots of failures; a number of them disappear (but not all) when the test is adjusted to take that into account

alvestrand:
Which Chrome version did you test on?
Unprefixed RTCPeerConnection is supposed to have landed in M56 (or someone forgot to adjust the bug). https://bugs.chromium.org/p/chromium/issues/detail?id=658423

dontcallmedom:
@alvestrand my bad, I tested in a horribly outdated version of Chrome :)

PR #4627, opened 2017-01-25
streams: test 2nd read request with autoAllocateChunkSize

isonmad:
This just checks against the erroneous assert removed from the spec in https://github.com/whatwg/streams/pull/651.


ricea (approved):
lgtm

tyoshino (approved):
lgtm

PR #4633, opened 2017-01-26
streams/r-b-s: eliminate asserts from try blocks

ricea:
Test harness asserts inside underlying source methods do not work
properly. All source methods effectively run inside try-catch blocks. If
an assert fails within one of these methods, it will at best lose the
assert message, and sometimes fail to notice the failure at all.
 
Fix this by moving asserts out of pull() methods, usually by copying the
value being checked and verifying it after pull() has returned. Some
asserts were redundant and have been removed completely: in particular
checks that pull() wasn't called more than once were often performed in
two different ways.

Some asserts were never reached and have been removed completely.

There are also some miscellaneous cleanups, particularly around awkward
use of promises. `if` statements that are redundant with assert statements
have been removed. Some unnecessary uses of the start() method to
capture the controller have been removed.

For practical reasons, this change doesn't fix everything. In
particular:
 - Only streams/readable-byte-streams/general.js is covered. Other files
   may have similar issues.
 - Many asserts still lack descriptions.
 - Lots of hard-to-follow out-of-order logic still remains.
 - general.js is far too long and needs splitting.

### DIFF
--- a/streams/readable-byte-streams/general.js
+++ b/streams/readable-byte-streams/general.js
@@ -20,6 +20,8 @@ test(() => {
 
 promise_test(() => {
   let startCalled = false;
+  let startCalledBeforePull = false;
+  let desiredSize;
   let controller;
 
   let resolveTestPromise;
@@ -33,8 +35,8 @@ promise_test(() => {
       startCalled = true;
     },
     pull() {
-      assert_true(startCalled, 'start has been called');
-      assert_equals(controller.desiredSize, 256, 'desiredSize');
+      startCalledBeforePull = startCalled;
+      desiredSize = controller.desiredSize;
       resolveTestPromise();
     },
     type: 'bytes'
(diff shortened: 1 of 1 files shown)


domenic (approved):
Changes LGTM, but it's a long review so ideally we get another pair of eyes. Does the reference implementation still pass them?

mkruisselbrink:
> Fix this by moving asserts out of pull() methods, usually by copying the
> value being checked and verifying it after pull() has returned.

Another fix would have been to just wrap all these async methods in Test.step_func. And for things where you're trying to assert a method doesn't get called, passing Test.unreached_func() as that method seems simpler than having a variable, setting that variable, and then later checking that the variable didn't get changed.

ricea:
@domenic Yes, they pass. My guess is that the tests were ported from working tape tests.

@mkruisselbrink Thanks for telling me about unreached_func(). It has mitigated some of the size increase. If I had to do it again, I would use step_func(). But if I'd done that I wouldn't have caught so many bugs.

PR #4634, opened 2017-01-26
README: Mention proxies when describing the /etc/hosts entries.

rakuco:
People behind proxies need to go through some additional steps for `./serve`
and network lookups to work. The specifics vary depending on how the proxies
are set up, so for now just mention that this needs to be done.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,9 @@ following entries are required:
 0.0.0.0     nonexistent-origin.web-platform.test
 ```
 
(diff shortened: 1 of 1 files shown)


rakuco:
@annevk. Not sure who else to include for README patches.

annevk:
This is probably more something for @jgraham / @gsnedders / @sideshowbarker to review. Thanks though!

PR #4638, opened 2017-01-26
Test existence and typeof MediaError.message

wolenetz:
These may currently pass only on Mozilla nightly/aurora builds. Chrome support for MediaError.message is in-progress. @domenic @foolip @jyavenard  Please review.

@jdsmith3000, @travisleithead  FYI

See related WHATWG spec change https://github.com/whatwg/html/issues/2085 and communication to W3C at https://lists.w3.org/Archives/Public/www-archive/2016Nov/0004.html.

### DIFF
--- a/html/semantics/embedded-content/media-elements/error-codes/error.html
+++ b/html/semantics/embedded-content/media-elements/error-codes/error.html
@@ -28,6 +28,7 @@ function error_test(tagName, src) {
       assert_true(e.error instanceof MediaError);
       assert_equals(e.error.code, 4);
       assert_equals(e.error.code, e.error.MEDIA_ERR_SRC_NOT_SUPPORTED);
+      assert_equals(typeof e.error.message, 'string');
       t.done();
     });
   }, tagName + '.error after setting src to the empty string');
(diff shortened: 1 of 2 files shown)


html/semantics/embedded-content/media-elements/error-codes/error.html
```diff
@@ -28,6 +28,7 @@
       assert_true(e.error instanceof MediaError);
       assert_equals(e.error.code, 4);
       assert_equals(e.error.code, e.error.MEDIA_ERR_SRC_NOT_SUPPORTED);
+      assert_equals(typeof e.error.message, 'string');
```
foolip: Since this will fail initially, can you add 'error.message type' as the description (third argument)? That will make the failures more intelligible.

foolip (approved):
LGTM % nit, but should not merge this until spec PR is up and ready to merge.

foolip:
Actually, do you think the spec could require a non-empty error message, or will there be errors where we just don't know and would end up having to use "unknown decoder error" or something similarly useless? If it is always possible to have a useful message, then you could also test that the string is non-empty.

jyavenard:
@foolip We had discussion already on whether message should be an optional string or not. The consensus was that it was better not to.
Forcing message to be set fall in the same discussion I think, and you don't always know what caused the error. Like with MSE and endOfStream, there's no "message" parameter that could be passed there.

PR #4657, opened 2017-01-30
Add tests to verify the origin of Blob URLs.

mkruisselbrink:
Tests for the change from unicode serialization to ascii serialization in https://github.com/w3c/FileAPI/pull/61

### DIFF
--- a/FileAPI/url/origin.sub.html
+++ b/FileAPI/url/origin.sub.html
@@ -1,0 +1,32 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>FileAPI Test: Verify origin of Blob URL</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<body>
+<iframe src="{{location[scheme]}}://{{domains[天気の良い日]}}:{{location[port]}}/FileAPI/support/url-origin.html"></iframe>
+<script>
+test(t => {
+  const blob = new Blob(["Test Blob"]);
+  const url = URL.createObjectURL(blob);
+  assert_equals(new URL(url).origin, location.origin);
+  assert_true(url.includes(location.origin));
+}, 'Verify origin of Blob URI matches our origin');
+
+async_test(t => {
+  const frame = document.querySelector('iframe');
+  self.addEventListener('message', t.step_func(e => {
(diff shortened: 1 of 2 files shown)


FileAPI/url/origin.sub.html
```diff
@@ -0,0 +1,32 @@
...
+      if (e.source != frame.contentWindow) return;
+      const url = e.data.url;
+      assert_false(url.includes('天気の良い日'),
+          'Origin should be ascii rather than unicode');
+      assert_equals(new URL(url).origin, e.origin,
+          'Origin of URL should match origin of frame');
+      assert_true(url.startsWith('blob:{{location[scheme]}}://xn--'));
+      t.done();
+    }));
+  frame.addEventListener('load', t.step_func(() => {
```
Ms2ger: I think adding a load event handler here is vulnerable to a race condition where the event would be dispatched before the script is executed.

FileAPI/url/origin.sub.html
```diff
@@ -0,0 +1,32 @@
...
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<body>
+<iframe src="{{location[scheme]}}://{{domains[天気の良い日]}}:{{location[port]}}/FileAPI/support/url-origin.html"></iframe>
+<script>
+test(t => {
+  const blob = new Blob(["Test Blob"]);
+  const url = URL.createObjectURL(blob);
+  assert_equals(new URL(url).origin, location.origin);
+  assert_true(url.includes(location.origin));
```
Ms2ger: Also add a `startsWith` check here?

FileAPI/url/origin.sub.html
zcorpan: Fix would be to create the iframe with `createElement` and append it to body.

FileAPI/url/origin.sub.html
mkruisselbrink: Done

FileAPI/url/origin.sub.html
mkruisselbrink: Changed it to be that way. I'm always confused if it is possible or not for the load event to be fired before the first synchronous javascript executes. But dynamically creating the iframe is definitely cleaner.

PR #4658, opened 2017-01-31
Update dom/lists/DOMTokenList-iteration.html so it no longer expects duplicates

cdumez:
As per the latest DOM specification [1], when a DOMTokenList gets created, we
end up running the "attribute change steps" for element, localName, value, value,
and null. The attribute changed steps for DOMTokenList say "set tokens to value,
parsed." "parsed" links to [2] which says to run the ordered set parser, which
should get rid of duplicates. WebKit, for example, does get rid of duplicates.

[1] https://dom.spec.whatwg.org/#interface-domtokenlist
[2] https://dom.spec.whatwg.org/#concept-ordered-set-parser

### DIFF
--- a/dom/lists/DOMTokenList-iteration.html
+++ b/dom/lists/DOMTokenList-iteration.html
@@ -7,17 +7,17 @@
 <script>
   test(function() {
     var list = document.querySelector("span").classList;
-    assert_array_equals([...list], ["a", "a", "b"]);
+    assert_array_equals([...list], ["a", "b"]);
 
     var keys = list.keys();
     assert_false(keys instanceof Array);
     keys = [...keys];
-    assert_array_equals(keys, [0, 1, 2]);
+    assert_array_equals(keys, [0, 1]);
 
     var values = list.values();
     assert_false(values instanceof Array);
     values = [...values];
-    assert_array_equals(values, ["a", "a", "b"]);
+    assert_array_equals(values, ["a", "b"]);
 
     var entries = list.entries();
     assert_false(entries instanceof Array);


cdumez:
@domenic @annevk @foolip @bzbarsky

foolip:
w3c-test:mirror

cdumez:
Yes, I think the DOM spec has been changed. Let me try and find a reference.

bzbarsky:
See https://github.com/whatwg/dom/issues/105 for the relevant DOM spec discussion, such as it is.  @foolip, you were in on that.  The spec said something, then changed, then none of the UAs followed for years, until Safari sort of did in their pre-releses, then the spec got switched back to match shipping browsers, then Safari shipped the changed behavior (which now no longer matched the changed-back spec), then the spec got changed again to that behavior.  None of the other engines have updated to that last spec change, and I'm not clear on whether they plan to, unfortunately.

Speaking for Geckom I'm still not entirely convinced on the new spec behavior, but probably willing to take a patch in Gecko if there's a patch with no performance impact, as I said in .  Such a thing has not materialized yet.  I myself have no plans to work on this in the near future; I have a lot of other much higher-priority things on my plate.

cdumez:
whatwg/dom#105 is indeed related but was about the serializer AFAICT. WebKit already reverted the change to the serializer to match the latest spec. This particular test is not covering the serializer.

bzbarsky:
Hmm.  I thought that was tracking general changes to whether the uniquifying happens, but maybe I was wrong...

The general point stands: we will need a no-performance-impact change to DOMTokenList to make any changes to its behavior in Gecko, unfortunately.

foolip:
OK, so this is a pretty similar to the first mishap mentioned in https://blog.whatwg.org/improving-interoperability, so let's try to not do this kind of thing as much going forward.

@cdumez, perhaps there are other tests for this, but can you summarize where uniquifying happens in the various engines? It looks like Blink stores the tokens separately from the underlying values, so presumably it would be fairly simple for Blink to follow WebKit here. But, if the behavior of the 3 other engines kind of makes sense, then *maybe* the simplest path is to change the spec again, and be very sorry about it. (Sorry if we've already discussed this elsewhere and I've forgotten.)

ArkadiuszMichalski:
Some time ago I rise this duplicate problem here https://github.com/whatwg/dom/issues/201. Looks like Webkit finally gose like spec say (including serialising case), unfortunately still only Webkit.
But if duplicate will stay then all tests for `DOMTokenList` also must be correct. Once again I had to spend some time to find them in `web-platform-tests/dom/nodes/Element-classlist.html`, so if you take any decission then maybe move them to `dom/lists/DOMTokenList` (separate files for all methods or just use one file covers all methods) because `DOMTokenList` is used not only by the `classList`.

cdumez:
I thought these tests were supposed to match the spec and serve as incentive for browsers to adopt spec changes?

cdumez:
I don't think it is a good working model to block web-platform-tests matching the spec because some implementors disagree with the specification. If those implementators want to get the specification changed, then they can try and do so. However, I don't think this should prevent web-platform-tests matching the current specification from landing until the specification gets changed. Those tests can be updated later if the spec changes (again).

annevk:
@cdumez I tend to agree. Do you want to address @ArkadiuszMichalski feedback here or should we do that separately?

cdumez:
@ArkadiuszMichalski Do you mind if I do the refactoring in a follow-up?

ArkadiuszMichalski:
I open https://github.com/w3c/web-platform-tests/issues/4703 and explaining what is a problem, it does not lock in any way this bug.

annevk:
If anyone takes issue with my decision here please file an issue at https://github.com/whatwg/dom/issues/new and I will make sure it gets the attention it deserves, but I didn't feel blocking @cdumez longer was warranted.

foolip:
FWIW, I tried running all the tests in http://w3c-test.org/dom/lists/ in Chrome, Edge, Firefox and Safari. Safari Technology Preview passed them all, but elsewhere there are widespread failures. I had to do some manual checking on Edge, but in the end it seems like Chrome, Edge and Firefox do no uniquifying.

Since Safari is now passing everything, I'd feel like a bit of a douche trying to change it all back, so I guess let's see if anyone else picks up on the change.

harrisonwithrow-cloud (approved):
Loved it

harrisonwithrow-cloud (approved):

PR #4662, opened 2017-01-31
Fix cross-origin-objects-exceptions.html to expect indices in GetOwnP…

cdumez:
…ropertyNames

Fix cross-origin-objects-exceptions.html to expect indices in GetOwnPropertyNames
similarly to what was done to cross-origin-objects.html in 7fc836d5.

### DIFF
--- a/html/browsers/origin/cross-origin-objects/cross-origin-objects-exceptions.html
+++ b/html/browsers/origin/cross-origin-objects/cross-origin-objects-exceptions.html
@@ -71,8 +71,12 @@ addTest(function() {
  * Also tests for [[GetOwnProperty]] and [[HasOwnProperty]] behavior.
  */
 
+var whitelistedWindowIndices = ['0', '1'];
 var whitelistedWindowProps = ['location', 'postMessage', 'window', 'frames', 'self', 'top', 'parent',
                               'opener', 'closed', 'close', 'blur', 'focus', 'length'];
+whitelistedWindowProps = whitelistedWindowProps.concat(whitelistedWindowIndices);
+whitelistedWindowProps.sort();
+
 addTest(function() {
   for (var prop in window) {
     if (whitelistedWindowProps.indexOf(prop) != -1) {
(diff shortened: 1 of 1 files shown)


cdumez:
@domenic @bzbarsky

cdumez:
I think it would be good to update cross-origin-objects.html and cross-origin-objects-exceptions.html to cover named properties (not just indices). I am planning to do this in a follow-up.

domenic:
Ideally we would not have these two separate files... I think it was because Firefox wanted to pass one of the files even if they threw the wrong exception types? That doesn't seem like a good enough idea to cause this kind of error-prone duplication.

Anyway, let's merge for now pending further discussion there.

bzbarsky:
> I think it was because Firefox wanted to pass one of the files even if they threw the wrong exception types?

It was because we wanted a test that _all_ implementations could pass while implementing the observables that are actually important for security and interop, so they would notice if they ever regressed them.  A problem with wpt is that once you fail an assertion it stops testing everything else in that test, so you lose test coverage unless your tests are written very very carefully.

Looking at the proposed changes now.

bzbarsky:
Oh, and you can tell how it would work, unfortunately, by the fact that I didn't update cross-origin-objects-exceptions.html, because it never came up for me as "hey this test is failing" while I changed all sorts of stuff.  :(

html/browsers/origin/cross-origin-objects/cross-origin-objects-exceptions.html
```diff
@@ -71,8 +71,12 @@
  * Also tests for [[GetOwnProperty]] and [[HasOwnProperty]] behavior.
  */
 
+var whitelistedWindowIndices = ['0', '1'];
 var whitelistedWindowProps = ['location', 'postMessage', 'window', 'frames', 'self', 'top', 'parent',
                               'opener', 'closed', 'close', 'blur', 'focus', 'length'];
+whitelistedWindowProps = whitelistedWindowProps.concat(whitelistedWindowIndices);
```
bzbarsky: OK, so this includes the indices, but not the symbols, right?  We should really sync these tests against each other; the only differences between them should the the assert_throws lines, where cross_origin_objects has `null` as the first arg, and cross-origin-objects-exceptions has `"SecurityError"`...   That's what we should do in the followup.

Thank you for catching this!

bzbarsky:
> and cross-origin-objects-exceptions has "SecurityError"

I guess in some cases it has "TypeError"...

PR #4663, opened 2017-01-31
Add some more tests of NodeList iterability

domenic:
Adapted from https://github.com/tmpvar/jsdom/blob/8b82e0c39bf856b072130e99cc60ca06ed75f7a7/test/living-dom/node-list.js.

### DIFF
--- a/dom/nodes/NodeList-Iterable.html
+++ b/dom/nodes/NodeList-Iterable.html
@@ -8,6 +8,8 @@
     <p id="3"></p>
     <p id="4"></p>
     <p id="5"></p>
+
+    <div id="live"><b id="b1">1</b><b id="b2">2</b><b id="b3">3</b></div>
 <script>
     var paragraphs;
     setup(function() {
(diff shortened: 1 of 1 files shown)


dom/nodes/NodeList-Iterable.html
```diff
@@ -34,4 +36,24 @@
...
+        let total = 0;
+        for (const el of nodeList) {
+            ++total;
+            assert_equals(el.localName, "b");
+            if (total < 3) {
+                container.appendChild(document.createElement("b"));
+            }
+        }
+
+        assert_equals(total, 5);
```
rniwa: Can we also confirm the order in which elements appear?
e.g. we can add class/id/title on each `b` to identify each b's.

domenic: Done

domenic:
No idea what the stability bot's problem is this time...

domenic:
OK, bot problems were my bad messing up submodules. Good for a new review.

rniwa:
r=me.

PR #4668, opened 2017-01-31
Import WebDriver test harness files from wptrunner

jugglinmike:
Previously, WebDriver tests were executed against a WebDriver session
that was managed by the wptrunner tool. Beyond increasing coupling
between this project's tests and that particular test runner, this
factoring made it impossible to write tests against aspects of the
WebDriver protocol that explicitly concern session management.

Update the test logic to explicitly manage connections to the WebDriver
"remote end," (made possible by the runtime specification of host name
and port number), and import the relevant logic from the wptrunner
project.

This depends on https://github.com/w3c/wptrunner/pull/233.


webdriver/util/cleanup.py
```diff
@@ -0,0 +1,47 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
```
jugglinmike: @jgraham Is this license compatible with this project? If not, could we resolve this with a commit authored by you to remove it?

andreastt: I wrote this code. I hereby relinquish any rights I have on the code to the WPT project default license.

webdriver/conftest.py
```diff
@@ -0,0 +1,39 @@
+import os
+
+import pytest
+import webdriver
+
+import util.cleanup as cleanup
```
andreastt: This seems redundant.

webdriver/conftest.py
```diff
@@ -0,0 +1,39 @@
...
+
+@pytest.fixture(scope="session")
+def _session(request):
+    host = os.environ.get("WD_HOST", default_host)
+    port = int(os.environ.get("WD_PORT", default_port))
+
+    session = webdriver.Session(host, port)
+
+    def destroy():
+        if session.session_id:
```
andreastt: `is not None`

andreastt (changes requested):

webdriver/util/cleanup.py
jugglinmike: Ah, thanks!

webdriver/conftest.py
jugglinmike: Sorry, still getting a hang of that detail in Python.

webdriver/conftest.py
jugglinmike: Do you mean the general pattern (e.g. importing the module and defining the fixture here) or the specific aliasing (e.g. `as cleanup`)?

andreastt: Sorry, I meant you could just do `from util import cleanup`.

andreastt:
I also wonder if we should find a better name than `util`. I don’t want to create another kitchen sink.

jugglinmike:
> I also wonder if we should find a better name than util. I don’t want to
> create another kitchen sink.

I'd suggest `fixtures/`, but I'm anticipating a few more utilities for future
tests. For instance, despite its small size, the `inline` function will be good
to share between many files. Another example is domain-specific assertions.

We could just dump them in the `webdriver/` directory, but I'm pretty averse to
mixing files containing tests with non-test files (even if Pytest is clever
enough to sort out the difference). Would you feel any better about something
like this:

    - webdriver/
      |`- tests/
      |   |`navigation.py
      |    `contexts.py
      |`- fixtures/
      |   `cleanup.py
      |`- inline.py
       `- assert.py

jugglinmike:
@andreastt I know that my proposal above kind of makes the `webdriver/` directory itself a sort of "kitchen sink," but I'm open to suggestion for alternatives!

andreastt (approved):

andreastt:
I think your proposal for document structure makes sense, although I would like the tests to be directly under /webdriver. They are the focus of attention and it should be clear to contributors where to find them.

I don’t however think we need to fix that immediately in this PR. It can come later.

jugglinmike:
Thanks, Andreas!

PR #4676, opened 2017-01-31
Use 'left' and 'right' for 'clear' property animation.

jgraham:
'inline-start' and 'inline-end' are still behind the pref on beta channel
(bug 1253919).

MozReview-Commit-ID: ISI5QYbjf7X

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1333295


jgraham:
Code reviewed upstream.

PR #4681, opened 2017-01-31
Only treat actual boolean return values from OnErrorEventHandlerNonNull as being able to cancel the event.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1333838


jgraham:
Code reviewed upstream.

PR #4683, opened 2017-01-31
Rewrite preload test.

jgraham:
This test was broken as the number of resources loaded was 3 independent of
support for preload. In addition the use of setTimeout managed to make it
untable in gecko. Rewrite the test to depend on the load events rather than
on timers, and to fail early in browsers with no preload support.

### DIFF
--- a/preload/single_download_late_used_preload.html
+++ b/preload/single_download_late_used_preload.html
@@ -2,14 +2,14 @@
 <title>Ensure preloaded resources are not downloaded again when used</title>
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
-<link rel=preload href="resources/square.png" as=image>
+<link rel=preload href="resources/square.png?pipe=trickle(d1)" as=image>
 <script>
     var link = document.getElementsByTagName("link")[0]
     assert_equals(link.as, "image");
     link.addEventListener("load", () => {
         assert_equals(performance.getEntriesByType("resource").length, 3)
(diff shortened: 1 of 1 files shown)


preload/single_download_late_used_preload.html
```diff
@@ -1,19 +1,20 @@
...
-            document.getElementById("result").innerHTML = "<image src='resources/square.png'>";
-            t.step_timeout(function() {
-                assert_equals(performance.getEntriesByType("resource").length, 3);
-                t.done();
-            }, 1000);
-        }, 1000);
-    }));
+    var link = document.getElementsByTagName("link")[0]
+    assert_equals(link.as, "image");
+    link.addEventListener("load", () => {
```
yoavweiss: That seems racy if the link finishes downloading before this runs.
Maybe move that to a function higher up, and use the `load` attribute?

yoavweiss (changes requested):

yoavweiss:
Thanks for fixing this! Overall looks good, once the raciness is fixed.

preload/single_download_late_used_preload.html
jgraham: I think the easiest fix is just to make the image slow loading. Otherwise we need to insert the `<link>` with js to be totally sure it can't load before the event handler is attached.

yoavweiss:
I guess that works as well. LGTM

jgraham:
Thanks!

PR #4549, opened 2017-01-18
selection: Split addRange.html into 15.

tkent-google:
addRange.html exceeds the time limit for a single test in Chromium project, and
also the text dump of its result exceeds the file size limit of code review tool
of Chromium project.

This is similar to https://github.com/w3c/web-platform-tests/pull/4049

PR #4684, opened 2017-02-01
Truncate data written to logs in C.I. environment

### LINKED ISSUE #4550: travis-ci: Huge change can't pass stability check

The PR affects 40,000+ sub-tests, and it exceeds the log size limitation.
https://s3.amazonaws.com/archive.travis-ci.org/jobs/192883235/log.txt

> The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
>
> The job has been terminated

This termination is handled as an error in the PR.

@jgraham

jugglinmike:
My choice of maximum row count is completely arbitrary. I'm happy to change it to any other value based on the experience of my reviewers. For context:

- the log generated for the patch that identified this issue (gh-4549) triggered tests for 15 "noisy" files and produced a log that exceeded Travis CI's 4 megabyte limit.
- the log generated for this pull request (which modifies the same 15 files) is 1.1 megabytes in size.

Please note that this solution (limiting overall log size by truncating per-file tables) does *not* scale for patches that effect a large number of tests that contain many subtests. In those cases, the overall log size may grow beyond TravisCI's 4MB limit despite the truncation of each individual test's table. Supporting this case gracefully would significantly increase the script's complexity (e.g. log buffering techniques, row allocation heuristics, etc.). This is likely not warranted given the fact that such patches are equally onerous for reviewers to validate and are likely discouraged as a matter of policy.

If that assessment seems reasonable, then this should resolve gh-4550.

### DIFF
--- a/check_stability.py
+++ b/check_stability.py
@@ -33,6 +33,7 @@ wpt_root = None
 wptrunner_root = None
 
 logger = logging.getLogger(os.path.splitext(__file__)[0])
+MAX_TABLE_ROWS = 256 if "CONTINUOUS_INTEGRATION" in os.environ else float("inf")
 
 
 def do_delayed_imports():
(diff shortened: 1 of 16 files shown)


jugglinmike:
...but of course, GitHub.com doesn't take too kindly to comments containing even "just" 1MB of text. Following up with a more complete solution at gh-4697.

PR #4687, opened 2017-02-01
Add test for invoke listener of legacy event type

upsuper:
This is the test for invoke events with legacy event type.

It includes tests for my proposal in whatwg/dom#404.

cc @smaug---- @annevk @miketaylr @foolip

### DIFF
--- a/dom/events/EventListener-invoke-legacy.html
+++ b/dom/events/EventListener-invoke-legacy.html
@@ -1,0 +1,72 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Invoke legacy event listener</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<style>
+  @keyframes test {
+    0% { color: red; }
+    100% { color: green; }
+  }
+</style>
+<div id="log"></div>
+<script>
+function runLegacyEventTest(type, legacyType, ctor, setup) {
+  function createTestElem(t) {
+    var elem = document.createElement('div');
+    document.body.appendChild(elem);
+    t.add_cleanup(function() {
+      document.body.removeChild(elem);
+    });
+    return elem;
+  }
+
+  async_test(function(t) {
+    var elem = createTestElem(t);
+    var gotEvent = false;
+    elem.addEventListener(legacyType,
+      t.unreached_func("listener of " + legacyType + " should not be invoked"));
+    elem.addEventListener(type, t.step_func(function() {
+      assert_false(gotEvent, "unexpected " + type + " event");
+      gotEvent = true;
+      t.step_timeout(function() { t.done(); }, 100);
+    }));
+    setup(elem);
+  }, "Listener of " + type);
+
+  async_test(function(t) {
+    var elem = createTestElem(t);
+    var count = 0;
+    elem.addEventListener(legacyType, t.step_func(function() {
+      ++count;
+      if (count > 1) {
+        assert_unreached("listener of " + legacyType + " should not be invoked again");
+        return;
+      }
+      elem.dispatchEvent(new ctor(type));
+      t.done();
+    }));
+    setup(elem);
+  }, "Legacy listener of " + type);
+}
+
+function setupTransition(elem) {
+  elem.style.transition = '';
+  requestAnimationFrame(function() {
+    elem.style.color = 'red';
+    elem.style.transition = 'color 30ms';
+    requestAnimationFrame(function() {
+      elem.style.color = 'green';
+    });
+  });
+}
+
+function setupAnimation(elem) {
+  elem.style.animation = 'test 30ms 2';
+}
+
+runLegacyEventTest('transitionend', 'webkitTransitionEnd', TransitionEvent, setupTransition);
+runLegacyEventTest('animationend', 'webkitAnimationEnd', AnimationEvent, setupAnimation);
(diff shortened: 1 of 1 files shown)


dom/events/EventListener-invoke-legacy.html
```diff
@@ -0,0 +1,73 @@
...
+      ++count;
+      if (count > 1) {
+        assert_unreached("listener of " + legacyType + " should not be invoked again");
+        return;
+      }
+      elem.dispatchEvent(new ctor(type));
+      t.done();
+    }));
+    setup(elem);
+  }, "Legacy listener of " + type);
```
annevk: Shouldn't the listener not fire here if we make your proposed change?

annevk (changes requested):

dom/events/EventListener-invoke-legacy.html
upsuper: The `setup` function triggers the event from user agent.

annevk: Oh I see. So the counter makes sure `dispatchEvent()` doesn't have an effect. Another way to test that would be to see if the event object is the same.

Also, you can use `t.step_func_done` and drop the `t.done()` call here.

upsuper: I thought about using `t.step_func_done`, but since that listener would be invoked twice when failing, and I'd rather keep `t.done()` be called only once even when that happens, I decided to call `t.done()` explicitly.

annevk (approved):

annevk:
Should we wait with landing this until we have changed the standard or tested that changing the standard is compatible? Seems a little bad to have the tests and standard be out-of-sync.

upsuper:
I have no idea. Do we agree to change the spec in that way? If so, I can probably submit a pr to the spec.

annevk:
I'm certainly fine with it, but we need to be wary of compat fallout.

zcorpan:
cc @grorg

dom/events/EventListener-invoke-legacy.html
```diff
@@ -0,0 +1,73 @@
...
+<script src="/resources/testharnessreport.js"></script>
+<style>
+  @keyframes test {
+    0% { color: red; }
+    100% { color: green; }
+  }
+</style>
+<div id="log"></div>
+<script>
+var $test = document.getElementById('test');
```
foolip: `$test` will be null and is never used

dom/events/EventListener-invoke-legacy.html
foolip: I don't get why `setup(elem)` is called for this test as well. Unless I'm missing some interaction, I think it'd be clearer to make this a sync test with one event listener for each of type and legacyType, and immediately after `element.dispatchEvent(...)` assert that just the first callback was invoked.

upsuper: If you register a listener for type, the listener for legacyType would never be called, with or without the spec change. See the first test.

In this test, I want to also ensure that the listener registered on legacyType can actually be called when such event happens. No other existing test checks this. If we make this test sync and only test that legacyType listener would not be invoked by dispatchEvent, we may want a seperate test for that. That's fine, too.

dom/events/EventListener-invoke-legacy.html
upsuper: Good catch. I forgot to remove this line when I refactored the test.

w3c-bots:
# Chrome (unstable channel) #
Testing web-platform-tests at revision 0fd80420eab1094eff4ca7d9d7422ab6aec752a3
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary><a href="http://w3c-test.org/submissions/4687/dom/events/EventListener-invoke-legacy.html">/dom/events/EventListener-invoke-legacy.html</a></summary>


|                 Subtest                 | Results |
|-----------------------------------------|---------|
|                                         | OK      |
| `Legacy listener of animationend`       | FAIL    |
| `Listener of animationstart`            | PASS    |
| `Legacy listener of animationiteration` | FAIL    |
| `Listener of transitionend`             | PASS    |
| `Listener of animationiteration`        | PASS    |
| `Listener of animationend`              | PASS    |
| `Legacy listener of transitionend`      | FAIL    |
| `Legacy listener of animationstart`     | FAIL    |

</details>

w3c-bots:
# Firefox (nightly channel) #
Testing web-platform-tests at revision 0fd80420eab1094eff4ca7d9d7422ab6aec752a3
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary><a href="http://w3c-test.org/submissions/4687/dom/events/EventListener-invoke-legacy.html">/dom/events/EventListener-invoke-legacy.html</a></summary>


|                 Subtest                 | Results |
|-----------------------------------------|---------|
|                                         | OK      |
| `Legacy listener of animationend`       | FAIL    |
| `Listener of animationstart`            | PASS    |
| `Legacy listener of animationiteration` | FAIL    |
| `Listener of transitionend`             | PASS    |
| `Listener of animationiteration`        | PASS    |
| `Listener of animationend`              | PASS    |
| `Legacy listener of transitionend`      | FAIL    |
| `Legacy listener of animationstart`     | FAIL    |

</details>

dom/events/EventListener-invoke-legacy.html
foolip: Thanks for elaborating :)

PR #4693, opened 2017-02-01
Upstream Node.js URL test data

joyeecheung:
Refs: https://github.com/nodejs/node/pull/11079

cc @domenic


joyeecheung:
The second commit added expected origins(need more thorough reviews), currently Node.js is unable to produce these expected origins either.

domenic (changes requested):
The URL parsing looks good; the origins should all be the string "null" per https://url.spec.whatwg.org/#origin "Otherwise" + https://html.spec.whatwg.org/multipage/browsers.html#unicode-serialisation-of-an-origin step 1

joyeecheung:
Ah, forgot these schemes should've resulted int opaque origins. Thanks, I've updated them to `null`, PTAL.

domenic (approved):

PR #4651, opened 2017-01-30
Add owner-agnostic stability commenter.

bobholt:
This PR removes the GitHub commenting from `check_stability.py` and offloads it to a bot running on an external server. This allows stability comments to appear on PRs from both trusted and untrusted branches.

Landing this PR requires infrastructure setup and additional changes to the main commit as follows:

1. Set up the [`prbuildbot`](https://github.com/bobholt/prbuildbot) on a server.
  - In discussions with @RByers, this service would ideally live on a W3C-owned server. Thoughts @jgraham and @sideshowbarker?
  - Setting up a Ubuntu 16.04 box  using Ansible to run the prbuildbot as a systemd service is described in that project. I'm happy to assist figuring out any other deployment target.
  - This will require a GH Personal Access Token (with user:email and public_repo permissions)for the account you'd like to comment on PRs. This user is currently wp-stability-bot, but could be wp-pr-bot or any other user you'd like.
2. Edit the WebHook line in `.travis.yml` in this PR to point to the server created in step 1.
3. Re-trigger the Travis build on this PR. If everything works, the prbuildbot should comment on this PR.

You can see how this all works successfully on [my fork](https://github.com/bobholt/web-platform-tests/pull/4) (the comments are from a previous commit where I pointed the `.travis.yml` to a DigitalOcean droplet running the prbuildbot).

I have included an extra commit that should trigger tests that should be removed once everything works.

> Notes: There has not yet been any code review on the prbuildbot project. It's fair game, so please feel free to comment/open issues you find there. I'm also happy to turn it over to W3C or other interested party for long term ownership and/or maintenance and give it whatever license is appropriate for that owner. The behavior in the bot is not tied to any particular repository or user, and could be used to provide the same functionality to any GitHub/Travis project.

PR #4697, opened 2017-02-01
Truncate data written to logs in C.I. environment

### LINKED ISSUE #4550: travis-ci: Huge change can't pass stability check

jugglinmike:
My choice of maximum row count is completely arbitrary. I'm happy to change it to any other value based on the experience of my reviewers. For context:

- the log generated for the patch that identified this issue (gh-4549) triggered tests for 15 "noisy" files and produced a log that exceeded Travis CI's 4 megabyte limit.
- the log generated for this pull request (which modifies the same 15 files) is 1.1 megabytes in size.

Please note that this solution (limiting overall log size by truncating per-file tables) does *not* scale for patches that effect a large number of tests that contain many subtests. In those cases, the overall log size may grow beyond TravisCI's 4MB limit despite the truncation of each individual test's table. Supporting this case gracefully would significantly increase the script's complexity (e.g.  log buffering techniques, row allocation heuristics, etc.). This is likely not warranted given the fact that such patches are equally onerous for reviewers to validate and are likely discouraged as a matter of policy.

Further truncate the log posted to GitHub.com. This is a hard clamp at GitHub's maximum length, minus the text necessary to notify the reader that the truncation occurred. This operation may invalidate part of the output (e.g. Markdown tables of open HTML tags), but again, I think we should prefer simplicity to completeness in guarding against this particular edge case. A link to the Travis "build job" page (containing a "less-abbreviated" version of the output) is included where available.

If all that seems reasonable, then this should resolve gh-4550.

### DIFF
--- a/check_stability.py
+++ b/check_stability.py
@@ -10,11 +10,10 @@ import sys
 import tarfile
 import zipfile
 from abc import ABCMeta, abstractmethod
-from cStringIO import StringIO
+from cStringIO import StringIO as CStringIO
 from collections import defaultdict
 from ConfigParser import RawConfigParser
-from io import BytesIO
-from tools.manifest import manifest
+from io import BytesIO, StringIO
 
 import requests
 
@@ -24,14 +23,14 @@ LogHandler = None
 LogLevelFilter = None
 StreamHandler = None
 TbplFormatter = None
+manifest = None
 reader = None
 wptcommandline = None
 wptrunner = None
 wpt_root = None
 wptrunner_root = None
 
-logger = logging.getLogger(os.path.splitext(__file__)[0])
-
+logger = None
 
 def do_delayed_imports():
     """Import and set up modules only needed if execution gets to this point."""
@@ -39,12 +38,14 @@ def do_delayed_imports():
     global LogLevelFilter
     global StreamHandler
     global TbplFormatter
+    global manifest
     global reader
     global wptcommandline
     global wptrunner
     from mozlog import reader
     from mozlog.formatters import TbplFormatter
     from mozlog.handlers import BaseHandler, LogLevelFilter, StreamHandler
+    from tools.manifest import manifest
     from wptrunner import wptcommandline, wptrunner
     setup_log_handler()
     setup_action_filter()
@@ -57,8 +58,6 @@ def setup_logging():
     handler.setFormatter(formatter)
     logger.addHandler(handler)
     logger.setLevel(logging.DEBUG)
-
-setup_logging()
 
 
 def setup_action_filter():
@@ -105,6 +104,50 @@ class TravisFold(object):
     def __exit__(self, type, value, traceback):
         """Emit fold end syntax."""
         print("travis_fold:end:%s" % self.name, file=sys.stderr)
+
+
+class FilteredIO(object):
+    """Wrap a file object, invoking the provided callback for every call to
+    `write` and only proceeding with the operation when that callback returns
+    True."""
+    def __init__(self, original, on_write):
+        self.original = original
+        self.on_write = on_write
+
+    def __getattr__(self, name):
+        return getattr(self.original, name)
+
+    def disable(self):
+        self.write = lambda msg: None
+
+    def write(self, msg):
+        encoded = msg.encode("utf8", "backslashreplace").decode("utf8")
+        if self.on_write(self.original, encoded) is True:
+            self.original.write(encoded)
+
+
+def replace_streams(capacity, warning_msg):
+    # Value must be boxed to support modification from inner function scope
+    count = [0]
+    capacity -= 2 + len(warning_msg)
+    stderr = sys.stderr
+
+    def on_write(handle, msg):
+        length = len(msg)
+        count[0] += length
+
+        if count[0] > capacity:
+            sys.stdout.disable()
+            sys.stderr.disable()
+            handle.write(msg[0:capacity - count[0]])
+            handle.flush()
+            stderr.write("\n%s\n" % warning_msg)
+            return False
+
+        return True
+
+    sys.stdout = FilteredIO(sys.stdout, on_write)
+    sys.stderr = FilteredIO(sys.stderr, on_write)
 
 
 class Browser(object):
@@ -275,7 +318,7 @@ def seekable(fileobj):
     try:
         fileobj.seek(fileobj.tell())
     except Exception:
-        return StringIO(fileobj.read())
+        return CStringIO(fileobj.read())
     else:
         return fileobj
 
@@ -522,7 +565,6 @@ def table(headings, data, log):
         log("|%s|" % "|".join(" %s" % row[i].ljust(max_widths[i] - 1) for i in cols))
     log("")
 
-
 def write_inconsistent(inconsistent, iterations):
     """Output inconsistent tests to logger.error."""
     logger.error("## Unstable results ##\n")
@@ -582,6 +624,10 @@ def get_parser():
                         # This is a workaround to get what should be the same value
                         default=os.environ.get("TRAVIS_REPO_SLUG").split('/')[0],
                         help="Travis user name")
+    parser.add_argument("--output-bytes",
+                        action="store",
+                        type=int,
+                        help="Maximum number of bytes to write to standard output/error")
     parser.add_argument("product",
                         action="store",
                         help="Product to run against (`browser-name` or 'browser-name:channel')")
@@ -592,10 +638,18 @@ def main():
     """Perform check_stability functionality and return exit code."""
     global wpt_root
     global wptrunner_root
+    global logger
 
     retcode = 0
     parser = get_parser()
     args = parser.parse_args()
+
+    if args.output_bytes is not None:
+        replace_streams(args.output_bytes,
+                        "Log reached capacity (%s bytes); output disabled." % args.output_bytes)
+
+    logger = logging.getLogger(os.path.splitext(__file__)[0])
+    setup_logging()
 
     wpt_root = os.path.abspath(os.curdir)
     wptrunner_root = os.path.normpath(os.path.join(wpt_root, "..", "wptrunner"))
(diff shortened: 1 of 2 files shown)


check_stability.py
```diff
@@ -33,6 +33,7 @@
 wptrunner_root = None
 
 logger = logging.getLogger(os.path.splitext(__file__)[0])
+MAX_TABLE_ROWS = 256 if "CONTINUOUS_INTEGRATION" in os.environ else float("inf")
```
jgraham: Should be `else None` I think.

check_stability.py
```diff
@@ -170,16 +173,31 @@ def get(self, url, headers=None):
...
+
+            if full_log_url is not None:
+                link = " [The complete log is available here.](%s)" % full_log_url
+
+            truncation_msg = ("*This report has been truncated because the " +
+                "total content is %s characters in length, which is in " +
+                "excess of GitHub.com's limit for comments (%s characters)." +
+                "%s*\n\n") % (len(body), self.max_comment_length, link)
+
+            body = truncation_msg + \
```
jgraham: Avoid line continuation characters.

check_stability.py
```diff
@@ -211,7 +229,15 @@ def emit(self, record):
 
     def send(self):
         """Post log to GitHub and flush log."""
-        self.github.post_comment(self.pull_number, "\n".join(self.log_data))
+        full_log_url = None
+        travis_job_id = os.environ.get("TRAVIS_JOB_ID")
+
+        if travis_job_id is not None:
+            full_log_url = \
```
jgraham: Again, please remove the line continuation character.

jgraham (commented):
It seems like a more sophisticated approach would be to count the number of bytes in `log` and allow 3Mb of output from the python script in travis before printing a truncation message and then making `log()` a noop.

check_stability.py
jugglinmike: Python appears to interpret `None` as being "less than" all other number values:

    >>> [(None < x) for x in [float('inf'), 1, 0, -1, -float('inf')]]
    [True, True, True, True, True]

Since we don't need to enforce a log size limit outside of continuous integration, I don't think this appropriate for use as a maximum value.

check_stability.py
jugglinmike: Got it. (I'm making it obvious that I rarely write Python, aren't I?)

tobie:
Ping @wpt-pr-bot.

jugglinmike:
@jgraham I've pushed up a commit that implements the solution at a lower level (the "Python file descriptor" level, but not so low as the "process output stream" level). I've defined it in terms of a new command-line option for the stability script--maybe a little over-engineered, but I wanted to make the behavior as discoverable as possible.

Anyway, [the latest build logs](https://api.travis-ci.org/jobs/198201110/log.txt?deansi=true) demonstrate the intended behavior:

>     INFO:check_stability:| `Range 7 [paras[1].firstChild, 2, paras[1].firstChild, 9] followed by Range 2 [paras[0].firstChild, 2, paras[0].firstChild, 8]: first addRange() must result in the selection's last range being the same object we added`                 | PASS    |
>     INFO:check_stability:| `Range 4 [paras[1].firstChild, 0, paras[1].firstChild, 0] followed by Range 21 [foreignDoc.body, 0, foreignDoc.body, 0]: first addRange() must result in the selection's last range being the same object we added`                        | PASS    |
>     INFO:check_stability:| `Range 6 [paras[1].firstChild, 2, paras[1].firs
>     Log reached capacity (3145728 bytes); output disabled.
>
>
>     The command "bash $SCRIPT" exited with 0.
>     store build cache
>
>     change detected (content changed, file is created, or file is deleted):
>     /home/travis/.cache/pip/http/0/1/e/c/5/01ec55882d7fee6cccf9bafb035721d3b0f1c4f5d70ef515794d5153
>     /home/travis/.cache/pip/http/0/4/6/f/2/046f263a04464ad173f2b55b53089d6611b281fd92901c6030b4cd93
>     /home/travis/.cache/pip/http/0/8/d/c/e/08dcef3be0bb9cce46e42c11d622ed3fea0797e37de9ffac3f5d70d7
>     /home/travis/.cache/pip/http/0/e/6/0/8/0e6088bd8e09b5becb00f7b07c4c7e9f5ce58d23a05bb02f89d475b4
>     /home/travis/.cache/pip/http/2/7/6/b/0/276b09a4fa1b1ca2bb4fd4b43f5c3d25e8d0fea3c8af6edde614d8df
>     /home/travis/.cache/pip/http/2/7/e/3/e/27e3e8f32f819b0843f464a41afb5e987e18f8a89ab01ac2ad98f8b0
>     /home/travis/.cache/pip/http/2/d/9/f/0/2d9f0780aedbe51e7cdcb37bf4c03c32f9d05e9445fc2fb4bf91af6c
>     /home/travis/.cache/pip/http/3/8/1/8/e/3818e6cf419a275375f4fd69142c6bcb13c625d3d0a0bd5eecb07057
>     /home/travis/.cache/pip/http/3/a/f/3/a/3af3addf06e983a6c02f46e7bea70c221d3ff95bf1418fa6da354e14
>     /home/travis/.cache/pip/http/4/7/9/9/6/479960059f6a2c1bdbd0c32805cce11c8c453c4dcac2ea743c8462ff
>     /home/travis/.cache/pip/http/5/c/d/e/b/5c
>
>     ...
>
>     changes detected, packing new archive
>     uploading archive
>
>
>     Done. Your build exited with 0.

What do you think?

jugglinmike:
By the way: the modifications to the "selection" tests are a red herring. I've included them to demonstrate the change to the stability bot's behavior, but they won't actually be merged when this patch is landed.

jugglinmike:
Part of @bobholt's work in gh-4651 involved the removal of GitHub-specific logic from the stability bot. Because this patch modified the removed logic, this introduced merge conflicts here. I've resolved the conflicts and force-pushed the new version of the change set. (The original version of this branch is available on my fork for posterity: https://github.com/w3c/web-platform-tests/compare/master...jugglinmike:4550-limit-table-size-prebase.) I have ported the GitHub-specific functionality to [the new "prbuildbot" project](https://github.com/w3c/prbuildbot); [that change is under review here](https://github.com/w3c/prbuildbot/pull/2).

check_stability.py
```diff
@@ -107,6 +106,48 @@ def __exit__(self, type, value, traceback):
         print("travis_fold:end:%s" % self.name, file=sys.stderr)
 
 
+class FilteredIO(StringIO):
```
jugglinmike: I've been researching how to do this more completely, and I found the following
documentation:

> `stdout` and `stderr` needn’t be built-in file objects: any object is
> acceptable as long as it has a `write()` method that takes a string argument.

(https://docs.python.org/2/library/sys.html#sys.stdout)

It seems like for this use case, the naive "write-only" implementation is
sufficient. Would you agree?

check_stability.py
```diff
@@ -582,6 +622,11 @@ def get_parser():
                         # This is a workaround to get what should be the same value
                         default=os.environ.get("TRAVIS_REPO_SLUG").split('/')[0],
                         help="Travis user name")
+    parser.add_argument("--output-cap",
```
jugglinmike: Sure (though I'll point out that this increases the noun/verb ambiguity in the word "output")

check_stability.py
```diff
@@ -582,6 +622,11 @@ def get_parser():
...
+                        action="store",
+                        default=sys.maxint,
```
jugglinmike: Using `None` would force us to opt-out of argparse's type enforcement. What do you think about using `-1` to signify "no limit?"

check_stability.py
```diff
@@ -592,11 +637,18 @@ def main():
     """Perform check_stability functionality and return exit code."""
     global wpt_root
     global wptrunner_root
+    global logger
 
     retcode = 0
     parser = get_parser()
     args = parser.parse_args()
 
+    replace_streams(args.output_bytes,
```
jgraham: `if args.output_bytes is not None:` perhaps?

jugglinmike: Looks like I can support this by simply not specifying a "default" value when defining the option.

check_stability.py
jugglinmike: Addressed this in 1501eac729102ad130a40d0f9d3120bdd49312ff. GitHub seems a little slow today, so you may have missed that

check_stability.py
jugglinmike: I think I follow you. I've pushed up a commit that makes the class inherit from "object" and implements the `__getattr__` method as a proxy. Let me know if this is what you had in mind.

jgraham (approved):
Please remove the test changes.

ci_stability.sh
```diff
@@ -44,7 +44,7 @@ install_chrome() {
 }
 
 test_stability() {
-    python check_stability.py $PRODUCT
+    python check_stability.py $PRODUCT --output-cap $((1024 * 1024 * 3))
```
jgraham: This should be `--output-bytes`

w3c-bots:
# Chrome (unstable channel) #
Testing web-platform-tests at revision d716371c7823185ddaf668138a3a7caf76eb2a5f
Starting 10 test iterations
No tests run.

w3c-bots:
# Firefox (nightly channel) #
Testing web-platform-tests at revision d716371c7823185ddaf668138a3a7caf76eb2a5f
Starting 10 test iterations
No tests run.

jugglinmike:
Updated the command-line flag usage, rebased onto master, squashed, and removed the "demo" commit. The prior version of this branch is available here: https://github.com/w3c/web-platform-tests/compare/master...jugglinmike:4550-limit-table-size-prebase-2

Thanks for the review, James!

PR #4698, opened 2017-02-01
Sort stability test results according to test name

### LINKED ISSUE #4515: wpt-stability-bot: report results in test order rather than Python (unordered) dict order
That would make it easier to scan the results. Especially for generated tests that have some kind of sequence.

jugglinmike:
This should resolve gh-4515

### DIFF
--- a/check_stability.py
+++ b/check_stability.py
@@ -651,6 +651,7 @@ def write_inconsistent(inconsistent, iterations):
     logger.error("## Unstable results ##\n")
     strings = [("`%s`" % markdown_adjust(test), ("`%s`" % markdown_adjust(subtest)) if subtest else "", err_string(results, iterations))
                for test, subtest, results in inconsistent]
+    strings.sort()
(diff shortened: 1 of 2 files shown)


gsnedders:
That's still test name rather than test-order (i.e., order the test objects are created by the test function being called).

Also it's unstable, you could really do with reverting that last commit. :P

domenic:
This isn't quite what you want. You want the original *test* order, as written, not lexicographic order. Shuffling them around lexicographically isn't much better than shuffling them around randomly, IMO.

jgraham:
Right, if you want sorted output you'll need to change wptrunner not to use an object to communicate subtest results, but a list of lists, or similar.

jugglinmike:
Thanks for the feedback, all. Looks like my understanding of the problem was a bit naive. A true solution won't have much in common with this approach, so I'll close this pull request.

gsnedders:
@jgraham I think `OrderedDict` would be better than lists.

PR #4711, opened 2017-02-03
Updated assertions in urlsearchparams-getall tests

IvanJov:
Hi,

I've updated tests, made them nicer for debugging. Let me know what do you think.

Code was also sent as PR to Node repo: https://github.com/nodejs/node/pull/11142

Thanks!

### DIFF
--- a/url/urlsearchparams-getall.html
+++ b/url/urlsearchparams-getall.html
@@ -21,15 +21,20 @@ test(function() {
 
 test(function() {
     var params = new URLSearchParams('a=1&a=2&a=3&a');
-    assert_true(params.has('a'), 'Search params object has name "a"');
+    assert_true(params.has('a'), 'Search params object doesn\'t have name "a"');
     var matches = params.getAll('a');
-    assert_true(matches && matches.length == 4, 'Search params object has values for name "a"');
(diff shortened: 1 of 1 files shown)


tobie:
@wpt-pr-bot, are you there?

annevk:
This looks reasonable to me, but I'm never quite sure whether the assertion description needs to describe the assertion or the unexpected result.

@gsnedders @jgraham? Seems that http://testthewebforward.org/docs/testharness-library.html#making-assertions should say this, no?

IvanJov:
@annevk I think that it should describe why the test failed. This is what docs say: `The description parameter is used to present more useful error messages when a test fails`. 🙂

domenic:
I've always gone with "should" or "must" in my assertion messages. So `'Search params object must have name "a"`

zcorpan:
w3c-test:mirror

zcorpan:
For `assert_equals` I generally avoid saying what the expected value is in the message, since the message already contains that and it tends to get out of sync if the test is later updated to change what is expected.

annevk:
@IvanJov it says to be useful when the test fails, it doesn't say describe when it fails. Given the feedback from others I guess we should either close this or simply drop these messages and rely on the default handling. Are you up for doing that?

annevk:
Closing due to lack of feedback. Sorry it didn't work out.

PR #4712, opened 2017-02-03
Fix hr-time IDL test - Missing EventTarget definition

JosephPecoraro:
Tests are unexpectedly failing because EventTarget interface parent of the
Performance interface is not defined. Add it.

--

This is a progression in Safari, Firefox, and Chrome.


igrigorik:
👍

PR #4714, opened 2017-02-03
Add a missing PerformanceTiming readonly attribute name

JosephPecoraro:
Any PerformanceTiming interface readonly attribute name should cause a SyntaxError when used by performance.mark in Window. Although this attribute is specially optional, it is a readonly attribute in the interface, so it should be tested.

---

[Spec](https://w3c.github.io/user-timing/#dom-performance-mark) for `performance.mark`:

> 1. If the global object is a Window object and markName uses the same name as a read only attribute in the PerformanceTiming interface [NAVIGATION-TIMING-2], throw a SyntaxError.

[Spec](https://w3c.github.io/navigation-timing/#the-performancetiming-interface) for `PerformanceTiming`:

>     interface PerformanceTiming {
>         ...
>         readonly attribute unsigned long long secureConnectionStart;
>     }

[Spec](https://w3c.github.io/navigation-timing/#dom-performancetiming-secureconnectionstart) for `secureConnectionStart`:

> **secureConnectionStart**
> This attribute is optional. User agents that don't have this attribute available must set it as undefined. [...]

---

Firefox does not appear to support `secureConnectionStart` yet. So adding this attributes causes Firefox to fail a few tests.

An alternative solution would be to only add "secureConnectionStart" to this list if it is supported by the browser, but I'm not sure that matters given the spec text above. Any read only attribute should throw a SyntaxError. I don't think supported or not matters.

### DIFF
--- a/user-timing/resources/webperftestharness.js
+++ b/user-timing/resources/webperftestharness.js
@@ -32,6 +32,7 @@ var timingAttributes = [
     'requestStart',
     'responseEnd',
     'responseStart',
+    'secureConnectionStart',
     'unloadEventEnd',
     'unloadEventStart'
 ];


JosephPecoraro:
Hmm, the existing [test_user_timing_mark_with_name_of_navigation_timing_optional_attribute.html](https://w3c-test.org/user-timing/test_user_timing_mark_with_name_of_navigation_timing_optional_attribute.html) does attempt to special case this optional attribute.

I'll leave it up to reviewers to suggest the path forward here. Should we special case this attribute in all tests or not.

igrigorik:
I think this is blocked on https://github.com/w3c/user-timing/issues/22.

plehegar:
The attribute was added in https://github.com/w3c/web-platform-tests/pull/5247, so feel free to close this PR.

igrigorik:
sgtm, thanks!

PR #4722, opened 2017-02-04
Add tests for which global is used in blob URL origins

domenic:
Helps with https://github.com/w3c/FileAPI/issues/60.

This checks that it's the current, since that's what we want to change the spec to, not the incumbent, what the spec currently says. It passes in Gecko and Blink; Edge fails as it seems to use an entirely different blob URL format. Haven't tested Safari.


tobie:
@wpt-pr-bot, are you there?

zcorpan:
Passes in Safari TP (10.2, WebKit 12604.1.4.2).

zcorpan (approved):

PR #4728, opened 2017-02-06
Correctly handle unserializable shorthand

Ms2ger:
get_shorthand_appendable_value doesn't always return a serializable
value. This change makes it handle that case correctly.


Ms2ger:
Code reviewed upstream.

PR #4729, opened 2017-02-06
Modify test to check if @import stylesheet get applied in a alternate stylesheet

Ms2ger:


Ms2ger:
Code reviewed upstream.

PR #4730, opened 2017-02-06
Fix 096.html and remove 098.html

Ms2ger:
Deferred scripts should have all ran before DOMContentLoaded is fired.
The removed script was a duplicate.


Ms2ger:
Code reviewed upstream.

PR #4731, opened 2017-02-06
Fix 128.html

Ms2ger:
The spec says that a script must be immediately prepared whenever "the script element
is connected and a node or document fragment is inserted into the script element,
after any script elements inserted at that time."


Ms2ger:
Code reviewed upstream.

PR #4732, opened 2017-02-06
Fix and reindent /old-tests/submission/Opera/script_scheduling/119.html

Ms2ger:


Ms2ger:
Code reviewed upstream.

gsnedders:
Ergh, why are these a single commit ;_;

PR #4734, opened 2017-02-06
script: Add tests for image intrinsic width when the element is not rendered.

Ms2ger:


Ms2ger:
Code reviewed upstream.

PR #4735, opened 2017-02-06
Correct the pass condition in FileReader-multiple-reads.html.

Ms2ger:


Ms2ger:
Code reviewed upstream.

PR #4738, opened 2017-02-06
Avoid data URLs for read-media tests.

Ms2ger:


Ms2ger:
Code reviewed upstream.

PR #4739, opened 2017-02-06
Check the contentType attribute for media documents.

Ms2ger:


Ms2ger:
Code reviewed upstream.

PR #4741, opened 2017-02-06
Add tests for windowless element's event handler behaviour

Ms2ger:


Ms2ger:
Code reviewed upstream.

PR #4742, opened 2017-02-06
Extend tests for windowless element's event handler behaviour to non-native handlers

Ms2ger:


Ms2ger:
Code reviewed upstream.

PR #4754, opened 2017-02-07
Remove invalid reftest: grouping-li-reftest-no-list-owner-ref.html

emilio:
The test relies in the HTML element not generating a layout box, via using
display: contents.

It does generate it though, because display: contents gets blockified in the
root element[1].

Indeed, Gecko shows zero-counters, but it does so even without display: contents
on the body and html elements, because Gecko only checks whether it has a
containing ol/ul/dir/menu elements[2].

[1]: https://drafts.csswg.org/css-display-3/#transformations
[2]: http://searchfox.org/mozilla-central/rev/848c29538ab007fb95dc6cff194f0e3d3809613d/layout/generic/nsContainerFrame.cpp#1822


emilio:
cc @domenic (who added the reftest)

domenic:
Hmm. Is it then not possible for a `<li>` to have no ancestors that generate CSS boxes? If so we need to update the spec as well (see the bottom of https://html.spec.whatwg.org/#ordinal-value )

w3c-bots:
# Chrome (unstable channel) #
Testing web-platform-tests at revision 4d3d3c8f3a2985202bd5b081e8cb3bd9eeb94e62
Using browser at version 58.0.3000.4 dev
Starting 10 test iterations
No tests run.

w3c-bots:
# Firefox (nightly channel) #
Testing web-platform-tests at revision 4d3d3c8f3a2985202bd5b081e8cb3bd9eeb94e62
Starting 10 test iterations
No tests run.

emilio:
> Hmm. Is it then not possible for a <li> to have no ancestors that generate CSS boxes? If so we need to update the spec as well (see the bottom of https://html.spec.whatwg.org/#ordinal-value )

Yes, it's not possible as far as I know, this is a Gecko bug.

domenic:
Great, thanks. In that case, test LGTM, and I'll update the spec.

PR #4756, opened 2017-02-07
[webdriver] Update tests

jugglinmike:
Recent changes to the "wptrunner" and "wd-client" projects have
invalidated some aspects of the WebDriver tests. Update the tests to
account for these changes, to correct previously-existing test bugs, and
to make the expectations more clear by using more explicit assertions.

@andreastt these tests currently hang in Firefox. Because the same WebDriver session
is re-used between tests, there are implicit navigation events between tests.
Marionette does not currently appears to be able to navigate between (for
instance) `about:blank#a` and `about:blank#b`. [I've filed a bug for that
here](https://bugzilla.mozilla.org/show_bug.cgi?id=1337464).


jugglinmike:
Whoops, meant to mention @andreastt.

w3c-bots:
# Chrome (unstable channel) #
Testing web-platform-tests at revision 5393e1aa6420564fc6a776a598eb00d9f9558aa8
Using browser at version 58.0.3000.4 dev
Starting 10 test iterations
No tests run.

w3c-bots:
# Firefox (nightly channel) #
Testing web-platform-tests at revision 5393e1aa6420564fc6a776a598eb00d9f9558aa8
Starting 10 test iterations
No tests run.

webdriver/navigation.py
```diff
@@ -1,109 +1,96 @@
...
+from util import wd_assert
 
 alert_doc = inline("<script>window.alert()</script>")
 frame_doc = inline("<p>frame")
 one_frame_doc = inline("<iframe src='%s'></iframe>" % frame_doc)
 two_frames_doc = inline("<iframe src='%s'></iframe>" % one_frame_doc)
 
 
-@pytest.fixture
-def new_window(session):
```
jugglinmike: This is a helpful fixture! This patch re-locates it to a separate module so we can re-use it from other tests. Those aren't included here, but I have some complete already that I plan to submit as a follow-up.

webdriver/util/wd_assert.py
```diff
@@ -0,0 +1,62 @@
...
+# >         message
+# >     stacktrace
+# >         stacktrace
+# >
+# > 5. Send a response with status and data as arguments.
+def error(result, name):
+    assert result.status == errors[name]
+    assert result.body["value"]["error"] == name
+    assert isinstance(result.body["value"]["message"], basestring)
+    assert isinstance(result.body["value"]["stacktrace"], basestring)
```
andreastt: I would use `types.StringTypes` since they could be unmarshalled to Unicode strings.

webdriver/util/wd_assert.py
```diff
@@ -0,0 +1,62 @@
...
+# >
+# >     error
+# >         name
+# >     message
+# >         message
+# >     stacktrace
+# >         stacktrace
+# >
+# > 5. Send a response with status and data as arguments.
+def error(result, name):
```
andreastt: `name` is misleading. Can you rename it `code`, `error_code` or similar?

andreastt: Please add a docstring.

andreastt: I think it would be useful if this function returned `result` unmodified.

webdriver/util/wd_assert.py
```diff
@@ -0,0 +1,62 @@
...
+    assert result.status == errors[name]
+    assert result.body["value"]["error"] == name
```
andreastt: I would do a separate assertion that `value` exists on body.

webdriver/util/fixtures.py
```diff
@@ -0,0 +1,77 @@
...
+    def create_frame():
+        session.start()
+        append = """
+            var frame = document.createElement('iframe');
+            document.body.appendChild(frame);
+            return frame;
+        """
+        response = session.send_command("POST",
+                                        "execute/sync",
+                                        dict(script=append, args=[]))
```
andreastt: Replace `[]` with `()`.

Also, why do you not call `session.execute` directly?

webdriver/util/fixtures.py
```diff
@@ -0,0 +1,77 @@
...
+        return response["value"]
```
andreastt: You can add `key="value"` to `send_command`.

webdriver/util/fixtures.py
```diff
@@ -0,0 +1,77 @@
...
+    """Open new window and return the window handle."""
+    def create_window():
+        windows_before = session.handles
+        name = session.execute_script("window.open()")
+        assert len(session.handles) == len(windows_before) + 1
+        new_windows = list(set(session.handles) - set(windows_before))
+        return new_windows.pop()
+    return create_window
+
+def create_frame(session):
```
andreastt: I think this deserves some documentation since it modifies the DOM. To be honest, I’m not sure about the benefit of this as opposed to inlining a frame in a data URI in the test. That might be less surprising that having to check the fixtures to see what they actually do.

webdriver/util/wd_assert.py
```diff
@@ -0,0 +1,62 @@
...
+    assert isinstance(result.body["value"]["message"], basestring)
+    assert isinstance(result.body["value"]["stacktrace"], basestring)
+
+def success(result, value):
```
andreastt: I think this and the previous function, `error` should be renamed `assert_error` and `assert_success`. This file should also be renamed `asserts` in my opinion. We might want other custom assertions than purely WebDriver related ones.

webdriver/util/inline.py
```diff
@@ -0,0 +1,4 @@
+import urllib
+
+def inline(document_source):
+    return "data:text/html;charset=utf-8,%s" % urllib.quote(document_source)
```
andreastt: I have a more advanced version of this if you’re interested:

```py
def inline(doc, doctype="html", mime="text/html;charset=utf-8"):
    if doctype == "html":
        mime = "text/html;charset=utf-8"
    elif doctype == "xhtml":
        mime = "application/xhtml+xml"
        doc = r"""<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
    "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
  <head>
    <title>XHTML might be the future</title>
  </head>

  <body>
    {}
  </body>
</html>""".format(doc)))
    return "data:{},{}".format(mime, urllib.quote(doc))
```

webdriver/conftest.py
```diff
@@ -25,15 +26,11 @@ def destroy():
...
-
-@pytest.fixture(scope="function")
 def http(session):
     return HTTPRequest(session.transport.host, session.transport.port)
+
+pytest.fixture(scope="function")(session)
+
+pytest.fixture()(create_window)
+
+pytest.fixture()(create_frame)
```
andreastt: What do these do?

webdriver/util/fixtures.py
```diff
@@ -0,0 +1,77 @@
...
+def _windows(session, exclude=None):
+    """Set of window handles, filtered by an `exclude` list if
+    provided.
+    """
+    if exclude is None:
+        exclude = []
+    wins = [w for w in session.handles if w not in exclude]
+    return set(wins)
+
+def create_window(session):
```
andreastt: Presumably this doesn’t need to be a fixture at all? You could make it a regular function and pass `session` to it?

webdriver/navigation.py
```diff
@@ -1,109 +1,96 @@
...
     # 7.2 step 1
-    session.window_handle = new_window
+    session.window_handle = create_window()
     session.close()
-    with pytest.raises(webdriver.NoSuchWindowException):
-        session.url = "about:blank"
+
+    result = session.send_raw_command("GET", "url")
+
+    wd_assert.error(result, "no such window")
```
andreastt: What does this improve?

webdriver/navigation.py
```diff
@@ -1,109 +1,96 @@
...
 
 def test_get_current_url_alert_prompt(session):
     # 7.2 step 2
     session.url = alert_doc
-    with pytest.raises(webdriver.UnexpectedAlertOpenException):
-        session.url = "about:blank"
 
+    result = session.send_raw_command("GET", "url")
+
+    wd_assert.success(result, alert_doc)
```
andreastt: _Get Current URL_ should return an error. The old test here is wrong as it navigates. It should just be `session.url`.

webdriver/navigation.py
```diff
@@ -1,109 +1,96 @@
...
-    with http.get("/session/%s/url" % session.session_id) as resp:
-        assert resp.status == 200
-        body = json.load(resp)
-    assert "value" in body
-    assert isinstance(body["value"], types.StringTypes)
 
+    result = session.send_raw_command("GET", "url")
+
+    assert result.status == 200
+    assert isinstance(result.body["value"], basestring)
```
andreastt: Doesn’t check for Unicode strings or that the body is JSON parsable?

webdriver/navigation.py
```diff
@@ -1,109 +1,96 @@
...
 
 def test_get_current_url_special_pages(session):
     session.url = "about:blank"
-    assert session.url == "about:blank"
 
+    result = session.send_raw_command("GET", "url")
+
+    wd_assert.success(result, "about:blank")
```
andreastt: You’ve already tested the raw command, so why not use `session.url` here?

webdriver/navigation.py
```diff
@@ -1,109 +1,96 @@
...
 # TODO(ato): This test requires modification to pass on Windows
 def test_get_current_url_file_protocol(session):
     # tests that the browsing context remains the same
     # when navigated privileged documents
     session.url = "file:///"
-    assert session.url == "file:///"
 
+    result = session.send_raw_command("GET", "url")
+
+    wd_assert.success(result, "file:///")
```
andreastt: Same here.

webdriver/navigation.py
```diff
@@ -1,109 +1,96 @@
...
 # TODO(ato): Test for http:// and https:// protocols.
 # We need to expose a fixture for accessing
 # documents served by wptserve in order to test this.
 
+def test_get_current_url_after_modified_location(session):
+    session.execute_script("window.location.href = 'about:blank#wd_test_modification'")
 
-def test_get_current_url_malformed_url(session):
-    session.url = "foo"
-    assert session.url
```
andreastt: This test has gone missing. The specification has changed since I wrote this test and it should instead return an ‘invalid argument’ error on passing a malformed URL.

webdriver/navigation.py
```diff
@@ -1,109 +1,96 @@
...
 
+    wd_assert.success(result, "about:blank#wd_test_modification")
 
-def test_get_current_url_after_modified_location(session):
-    session.execute_script("window.location.href = 'about:blank'")
-    assert session.url == "about:blank"
+def test_get_current_url_nested_browsing_context(session, create_frame):
+    session.url = "about:blank#wd_from_within_frame"
 
+    session.send_command("POST", "frame", dict(id=create_frame()))
```
andreastt: Why `dict`?

andreastt (changes requested):

webdriver/navigation.py
jugglinmike: This change makes the test more direct. The current factoring "reads" more like
a test for the `wdclient` library than for the specification itself--something
like, "Setting the `url` property on the `wdclient.Client` instance should
raise a Python exception." It also places more non-critical logic in the tested
code path.

My hope is that expressing the test subject in terms of `send_raw_command` will
make the test more legible to people who are familiar with the specification
but not the W3C's library--"Sending this command should return an error."

jugglinmike (commented):
Thanks for your feedback, @andreastt! I've implemented most of your suggestions, though I do have some questions about a few things...

jugglinmike:
I neglected to push up some of the changes you requested. Sorry about that;
they're all available now. I still have some questions about some of the finer
points of your earlier review, though. Would you mind taking another look?

webdriver/util/wd_assert.py
andreastt: Well actually I guess the use cases for it would be very limited. Ignore by above request.

webdriver/util/wd_assert.py
andreastt: You’re right, it does!

webdriver/util/fixtures.py
andreastt: I buy the argument that invoking helpers in one, streamlined way is beneficial. This change is good.

webdriver/util/fixtures.py
andreastt: Right, yes we do.

webdriver/util/fixtures.py
andreastt: > could you explain why `()` is preferable here?

Arguments in Python are represented as tuples, so it is in harmony with the standard library to represent script arguments as tuples too.

webdriver/util/fixtures.py
andreastt: I think this is less surprising because there are fewer moving parts:

```py
document_el = session.execute("return document.documentElement")
frame_id = append_frame(document_el)

def append_frame(session):
    def x(parent_el = None):
        if parent_el is None:
            parent_el = session.execute("return document.documentElement")
        id = uuid.uuid4()

        return session.execute("""const [parentEl, id] = arguments;
            let frame = document.createElement("iframe");
            frame.setAttribute("id", id);
            parentEl.appendChild(frame);
            return frame;""", script_args=(parent_el, id))

    return x

wrapper = session.find.css("#wrapper")
new_frame = append_frame(wrapper)
print new_frame.get_attribute("id")
# => '17d3a557-c85a-4650-bfe9-0a8f3b624903'
```

But I’m still not convinced it is better than:

```py
def test_foo(session):
    session.navigate(inline("<iframe></iframe>"))
    iframe = session.find.css("iframe")
    …
```

webdriver/conftest.py
andreastt: Makes sense.

webdriver/navigation.py
andreastt: `basestring` appears to be fine, and `Result` does indeed parse the body.

webdriver/navigation.py
andreastt: I don’t agree with you that `wd_assert.success(session.send_raw_command("GET", "url"), "about:blank")` is conceptually any more direct or simpler than making an HTTP request and asserting its characteristics, as we do in webdriver/contexts.py for Get Window Rect:

```py
with http.get("/url") as resp:
    assert resp.status == 200
    body = json.load(resp)

assert "value" in body
assert isinstance(body["value"], basestring)
assert body["value"] == "about:blank"
```

Whether you use `wd_assert.success` or the `session.url` property getter, they are both abstractions, and either one could repeat the above assertions. I subscribe to your ideal that we should avoid higher-level abstractions and promote immediacy and simplicity when testing a command, but I don’t see `wd_assert.success` as simple.

webdriver/navigation.py
andreastt: navigation.py is intended to test the _Navigation_ chapter in the spec. When this file becomes long enough that it makes sense to split it out, we can do that.

webdriver/navigation.py
andreastt: So I guess this is meant to switch to the frame? You could probably do {"element-6066-11e4-a52e-4f735466cecf": create_frame()}`.

andreastt:
Sorry for my late response to this PR, but I’ve been sick and away from work for a week. Let me know if you need me to look at something in particular to unblock your work.

webdriver/util/fixtures.py
jugglinmike: That `append_frame` fixture has additional functionality compared to the
`create_frame` fixture submitted here. It generates a UUID and allows for the
specification of a parent element. This isn't necessarily a bad thing (although
neither is currently required by any tests), but it conflicts with your
statement about having "fewer moving parts." This makes me think I am
misunderstanding your point.

As for the use of `inline("<iframe></iframe")`, that approach depends on URL
modification and therefore cannot be used from tests that concern the value of
the current URL.

andreastt:
Also sprach jugglinmike <notifications@github.com>:

> That append_frame fixture has additional functionality compared to
> the create_frame fixture submitted here. It generates a UUID and
> allows for the specification of a parent element. This isn't
> necessarily a bad thing (although neither is currently required by
> any tests), but it conflicts with your statement about having
> "fewer moving parts."   

Sure.  This is probably unneeded complexity.

> This makes me think I am misunderstanding your point.  As for the
> use of inline("<iframe></iframe"), that approach depends on URL
> modification and therefore cannot be used from tests that concern
> the value of the current URL.

That is a valid point.

w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/4756)

w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/4756)

w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/4756)

jugglinmike:
@ato I've rebased this branch and resolved the conflicts introduced via gh-5023 and gh-5024. Beyond that:

- The original version of this patch introduced a subdirectory named `util` for non-test files. Since the above patches introduces a subdirectory named `support` with a similar function, I relocated those files there.
- Incorporating feedback from your latest review
  - Miscellaneous improvements to the assertion functions as per:
    - https://github.com/w3c/web-platform-tests/pull/4756#discussion_r100568459
    - https://github.com/w3c/web-platform-tests/pull/4756#discussion_r100568595
    - https://github.com/w3c/web-platform-tests/pull/4756#discussion_r100568990
    - https://github.com/w3c/web-platform-tests/pull/4756#discussion_r100570202
  - Documented the `create_frame` Pytest fixture function as per https://github.com/w3c/web-platform-tests/pull/4756#discussion_r100569435
  - Extended the `inline` helper function as per https://github.com/w3c/web-platform-tests/pull/4756#discussion_r100571104
  - Updated the `create_frame` Pytest fixture function to use `Session#execute_script` as per https://github.com/w3c/web-platform-tests/pull/4756#discussion_r100569066
  - Fixed an old test as per https://github.com/w3c/web-platform-tests/pull/4756#discussion_r100572301
  - Re-introduced a mistakenly-deleted test as per https://github.com/w3c/web-platform-tests/pull/4756#discussion_r100573006

jugglinmike:
@andreastt Commenting here again in case you missed my previous message.

jugglinmike:
@andreastt Any word on this?

jugglinmike:
@AutomatedTester Would you mind taking a look? I have a few patches lined up based on this work (they use some of the same code and they're formatted with similar patterns), so I'd like confirmation from a maintainer about this before moving forward.

andreastt (approved):
r+

PR #4759, opened 2017-02-08
[editing] Correct duplicated test names

jugglinmike:
Duplicated test names make interpreting results difficult for humans and
machines. In order to avoid this confusion, [an upcoming extension to the test
harness](https://github.com/w3c/testharness.js/pull/240) will cause any test
file with duplicate test names to be interpreted as an error.

This situation occurs in a few of the "editing" tests. In most cases, this is
likely an unintentional repetition. The resolution is to simply remove one of
the two tests.

It is possible, however, that the duplication is actually the result of a
transcription error--that one of the two tests was intended to assert a
slightly different condition. In this case, the duplication actually highlights
an area of missing coverage and is best resolved by modifying the input values
to make the tests distinct.

I lack the domain knowledge to definitively differentiate between these two
possibilities. I have assumed that the former is always the case, but I request
that my reviewers take care to verify this assumption.

This change set also includes an extension to the logic for generating
tests--that code was previously generating duplicate strings for distinct
steps.


jgraham (approved):

editing/include/tests.js
```diff
@@ -5582,7 +5582,7 @@ function runConformanceTest(browserTest) {
 
             assert_equals(document.execCommand(browserTest[1][i][0], false, browserTest[1][i][1]),
                 expectedExecCommandReturnValues[i]);
-        }, testName + ": execCommand(" + format_value(browserTest[1][i][0]) + ", false, " + format_value(browserTest[1][i][1]) + ") return value");
+        }, testName + ": command #" + (i+1) + " execCommand(" + format_value(browserTest[1][i][0]) + ", false, " + format_value(browserTest[1][i][1]) + ") return value");
```
jgraham: Sorry, missed this bit. Sequence numbers are generally not a great idea as they have a tendency to change. It would be better to either only use this iff there would be a dupe, or to ensure that there is never a dupe.

jgraham (changes requested):

editing/include/tests.js
jugglinmike: > Sequence numbers are generally not a great idea as they have a tendency to
> change.

True, but the command sequence numbers are more than just test implementation
details--they are an integral part of each test. In other words, if the
sequence of a given test's commands were to change, then that would be a
distinct test.

I wasn't aware of any expectations about the stability of sub-test names. Is
that detail significant to consumers?

> It would be better to either only use this iff there would be a dupe, or to
> ensure that there is never a dupe.

We can't ensure there this is never a duplicate because some tests require
repeated applications of the same command. The first option is do-able, but I
don't know if the complexity is warranted (see above).

Then again, the surrounding test generation logic is already *very* involved,
so a few more branches couldn't hurt. I've said my peace--if you'd still be
more comfortable with "conditionally-specified" sequence numbers, then say the
word and I'll patch it in.

jgraham: Yes, sub test name stability is important. We compare the result of every subtest with the previous result of that subtest by name. So if the name changes then the expected result is lost. The worst situation (which has happened) is when two tests race to get one of a pair of names (usually foo-1 and foo-2) and the tests have different expected results. That leads to randomness in the test infrastructure.

If you are convinced that the numbers you're adding here are deterministic enough that they won't change when new tests are added, that's fine by me. I haven't studied the existing code enough to have a strong opinion so I trust your judgement given the above constraints.

jugglinmike: While I'm confident that this change results in stable test names (I even wrote
up a whole response to that effect), the testing infrastructure is sufficiently
complex to make me less-than-certain. Since I'm highly averse to introducing
instability, I've explored more explicit approaches.

I've pushed up a commit that implements a more declarative mechanism. With this
latest version, test objects may optionally specify an arbitrary string with
any command. This allows the two offending tests in `multitest.js` to extend
the repeated "underline" commands with a comment (in this case, "first
application" and "second application").

This could certainly be done in an automated way, but (as stated earlier) I'm
reluctant to contribute to the complexity of the logic. For the time being,
the "repeated command" condition occurs so infrequently (i.e. just twice) that
the extra requirement seems acceptable.

jugglinmike:
@jgraham [I implemented an alternative approach for those procedurally-generated names](https://github.com/w3c/web-platform-tests/pull/4759/commits/9776b95895fc9aef2b0d1c5eea36e7c02b3b2d25), but [my comment](https://github.com/w3c/web-platform-tests/pull/4759#discussion_r101606195) got folded away. I just fixed a linting error and updated the in-line documentation, so I think this should be good to go.

jugglinmike:
@jgraham Poking you again, as requested

jgraham (approved):

jugglinmike:
Thanks, @jgraham!

ayg:
@jgraham FTR, I don't think this was a sequence number of the kind we don't want.  "i" here is not an index into an array of tests, it's an index into a description of the current test.  It will not change if tests are added or removed.  I think it would have been better to keep it as it was, but it doesn't matter so much.

PR #4763, opened 2017-02-08
Add click event target tests during pointerevent capture

### LINKED ISSUE #3213: Add a test for click event while a pointer event is captured
Even though the PointerEvent spec doesn't override the [click target definition in UI Event](https://www.w3.org/TR/DOM-Level-3-Events/#event-type-click), a captured pointer makes the outcome slightly different. See the discussion in [this spec issue](https://github.com/w3c/pointerevents/issues/75).

We need a test for correct implementation.

NavidZ:
Add click event target tests during pointerevent capture plus itsordering compare to lostpointercapture events.
closes #3213

### DIFF
--- a/pointerevents/pointerevent_click_during_capture-manual.html
+++ b/pointerevents/pointerevent_click_during_capture-manual.html
@@ -1,0 +1,90 @@
+<!doctype html>
+<html>
+    <head>
+        <title>click event target during capture</title>
+        <meta name="viewport" content="width=device-width">
+        <link rel="stylesheet" type="text/css" href="pointerevent_styles.css">
+        <script src="/resources/testharness.js"></script>
+        <script src="/resources/testharnessreport.js"></script>
+        <script src="pointerevent_support.js"></script>
+        <style>
+        .box {
+          margin: 10px;
+        }
+        #grey {
+          background: grey;
+        }
+        #blue {
+          background: blue;
+        }
+        #green {
+          background: green;
+        }
+        </style>
+        <script type="text/javascript">
+            PhaseEnum = {
+              Phase1:  "phase1",
+              Phase2:  "phase2",
+            }
+            var phase;;
+            var receivedEvents;
+
+            function resetTestState() {
+              phase = PhaseEnum.Phase1;
+              receivedEvents = [];
+            }
+        
+            function run() {
+                var test_pointerEvent = setup_pointerevent_test("click target during capture", ['mouse']);
+                var grey = document.getElementById('grey');
+                var blue = document.getElementById('blue');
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Chrome (unstable channel) #
Testing web-platform-tests at revision e20d5a70ac8b17e025c5d72228f3fa4a8fdd06d0
Starting 10 test iterations
No tests run.

w3c-bots:
# Firefox (nightly channel) #
Testing web-platform-tests at revision e20d5a70ac8b17e025c5d72228f3fa4a8fdd06d0
Starting 10 test iterations
No tests run.

NavidZ:
@Patrickkettner @teddink

pointerevents/pointerevent_click_during_capture-manual.html
```diff
@@ -0,0 +1,90 @@
...
+                            phase = PhaseEnum.Phase2;
+                            receivedEvents = [];
+                          } else {
+                            test_pointerEvent.step(function() {
+                              assert_equals(receivedEvents.join(','), 'pointerdown@blue,gotpointercapture@blue,pointerup@blue,click@blue,lostpointercapture@blue', 'An element should only receive click when it is the first common ancestor of pointerdown and pointerup targets'); 
+                            });
+                            test_pointerEvent.done();
+                          }
+                        }
+                        if (event.type == 'pointerdown' && (target == blue || target == green)) {
```
mustaqahmed: Great, this test also covers "click" target check in addition to the ordering.

Looks like we don't have any test for "click" target when not capturing, may be because UI Event tests already cover this for mousedown/up. Now that checking for uncaptured pointerdown/up/click here is simple (right?), please consider adding two more phases---like existing ones but skips setpointercapture() calls.

NavidZ: Sure. I can add that test easily here. But then that part is only testing the UI event spec. If I recall correctly from what Rick said before the norm is to only add the tests that our spec covers in this folder.
@RByers, is it okay to add what Mustaq suggests?

RByers: I don't have a problem in general testing some of UIEvents here where it's convenient.  Since PE builds in UIEvents it's probably impossible to have tests which DON'T depend on things defined in UIEvents.  But if we find such tests trigger failures in some browsers we care about, we might want to discuss moving them elsewhere (so the failures are accounted for in the right place).

NavidZ: done

mustaqahmed (approved):
LGTM

PR #4787, opened 2017-02-09
[html] Specify unique names for tests

jugglinmike:
Duplicated test names make interpreting results difficult for humans and machines. In order to avoid this confusion, [an upcoming extension to the test harness](https://github.com/w3c/testharness.js/pull/240) will cause any test file with duplicate test names to be interpreted as an error.


jugglinmike:
Thanks!

PR #4798, opened 2017-02-10
URL: Correct expected result for non-empty search attribute

TRowbotham:
This fixes test data to expect the `searchParams` idl attribute to be
"qux=" since its `search` idl attribtue expects "?qux". This was causing
a test failure in url-constructor.html. The test data was first added in 65eb641.

cc @annevk

### DIFF
--- a/url/urltestdata.json
+++ b/url/urltestdata.json
@@ -4622,7 +4622,7 @@
     "port": "",
     "pathname": "/baz",
     "search": "?qux",
-    "searchParams": "",
+    "searchParams": "qux=",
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Chrome (unstable channel) #
Testing web-platform-tests at revision 5984497336f51bd6b49c0846aaa228227b7f1afa
Starting 10 test iterations

annevk:
Thanks! (@Manishearth and @zcorpan, please also run the test through a couple of browsers before merging.)

PR #4809, opened 2017-02-13
fix check_stability script to work without TRAVIS_REPO_SLUG env var

bobholt:
allows local running of check_stability.py for testing

I considered moving the `split` out of the default declaration and into the body of the script, because it's kind of a bear, but that would require passing `--user w3c/` instead of the more sensible `--user w3c` when running from the command line.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/4809)

### DIFF
--- a/check_stability.py
+++ b/check_stability.py
@@ -622,7 +622,7 @@ def get_parser():
                         action="store",
                         # Travis docs say do not depend on USER env variable.
                         # This is a workaround to get what should be the same value
-                        default=os.environ.get("TRAVIS_REPO_SLUG").split('/')[0],
+                        default=os.environ.get("TRAVIS_REPO_SLUG", "w3c").split('/')[0],
                         help="Travis user name")
     parser.add_argument("--output-bytes",
                         action="store",


w3c-bots:
# Chrome (unstable channel) #
Testing web-platform-tests at revision b73b1ca1691e84fcedb750aa8f36223e81500af0
Starting 10 test iterations
No tests run.

w3c-bots:
# Firefox (nightly channel) #
Testing web-platform-tests at revision b73b1ca1691e84fcedb750aa8f36223e81500af0
Starting 10 test iterations
No tests run.

check_stability.py
```diff
@@ -622,7 +622,7 @@ def get_parser():
                         action="store",
                         # Travis docs say do not depend on USER env variable.
                         # This is a workaround to get what should be the same value
-                        default=os.environ.get("TRAVIS_REPO_SLUG").split('/')[0],
+                        default=(os.environ.get("TRAVIS_REPO_SLUG") or "").split('/')[0],
```
jgraham: This doesn't really make sense, does it? If you don't pass anything here you will get a runtime error. So we should either check that or make the default `w3c` i.e. 

`default=os.environ.get("TRAVIS_REPO_SLUG", "w3c").split("/")[0]`

jgraham (changes requested):

sideshowbarker (dismissed):

bobholt:
Good point, @jgraham. Updated.

check_stability.py
```diff
@@ -622,7 +622,7 @@ def get_parser():
                         action="store",
                         # Travis docs say do not depend on USER env variable.
                         # This is a workaround to get what should be the same value
-                        default=os.environ.get("TRAVIS_REPO_SLUG").split('/')[0],
+                        default=(os.environ.get("TRAVIS_REPO_SLUG") or "w3c").split('/')[0],
```
jugglinmike: It's probably more idiomatic to use `get`'s second parameter here:

```diff
-  default=os.environ.get("TRAVIS_REPO_SLUG").split('/')[0],
+  default=os.environ.get("TRAVIS_REPO_SLUG", "w3c").split('/')[0],
```

bobholt: Why do I keep doing this?

w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/4809/)

PR #4821, opened 2017-02-14
Use travis-provided branch range as commit base, where possible

jgraham:


w3c-bots:
# Chrome (unstable channel) #
Testing web-platform-tests at revision 8b0b33b3c4a257da6ff9995c71e9e42610963e62
Git command exited with status 128
fatal: Invalid revision range 91a9307de957e185cc01c1e1a3458eb93a8d0d03..

w3c-bots:
# Firefox (nightly channel) #
Testing web-platform-tests at revision 8b0b33b3c4a257da6ff9995c71e9e42610963e62
Git command exited with status 128
fatal: Invalid revision range 91a9307de957e185cc01c1e1a3458eb93a8d0d03..

foolip:
How come so many labels were added to this PR? Is there a fallback to "everything"?

jgraham:
No, to test it I had to base it on some other branch. Then when I rebased on master it added labels for everything that changed between the original branch and master.

check_stability.py
```diff
@@ -384,11 +384,32 @@ def install_wptrunner():
...
+    logger.debug("Branch point from master: %s" % branch_point)
+    return branch_point
+
+
+def get_files_changed(branch_point):
     """Get and return files changed since current branch diverged from master."""
     root = os.path.abspath(os.curdir)
     git = get_git_cmd(wpt_root)
-    branch_point = git("merge-base", "HEAD", "master").strip()
     logger.debug("Branch point from master: %s" % branch_point)
```
bobholt: This is being logged in `get_branch_point`, above and can be removed.

bobholt (changes requested):
Looks good to me. @jugglinmike is more of a git-wizard than me, so it might be worth seeing if he has any other suggestions about the `else` block in `get_branch_point`

check_stability.py
```diff
@@ -384,12 +384,32 @@ def install_wptrunner():
     call("pip", "install", wptrunner_root)
 
 
-def get_files_changed():
+def get_branch_point(user):
+    git = get_git_cmd(wpt_root)
+    if os.environ.get("TRAVIS_PULL_REQUEST", "false") != "false":
```
jugglinmike: I think `"TRAVIS_PULL_REQUEST" in os.environ` would communicate your intention more clearly.

diff-manifest.py
```diff
@@ -22,7 +22,11 @@ def main():
     after = get_manifest("HEAD")
 
     call("git", "fetch", "origin", "master:master")
-    merge_base = call("git", "merge-base", "master", "HEAD").strip()
+    commit_range = os.environ.get("TRAVIS_COMMIT_RANGE")
+    if commit_range:
+        merge_base = commit_range.split(".", 1)[0]
+    else:
+        merge_base = call("git", "merge-base", "master", "HEAD").strip()
```
jugglinmike: Why is the simpler heuristic acceptable here?

ci_lint.sh
```diff
@@ -2,4 +2,3 @@ set -ex
 
 ./manifest
 ./lint
-./diff-manifest.py
```
jugglinmike: I don't see why this should be removed--is this change intentional?

jugglinmike (commented):
Hey James! I have nothing to block on, but some questions nonetheless.

check_stability.py
jgraham: Ah, but that would unfortunately not work. If something isn't a pull request it still has the environment variable, but set to the string `"false"` (assuming my reading of the docs is correct).

ci_lint.sh
jgraham: Yeah, that's intentional because it doesn't really do anything. There's no way the check can fail and it doesn't expose any useful information because we effectively do the same thing with the list of changed tests that are posted to the PR.

diff-manifest.py
jgraham: Oh it isn't, that's the main reason I actually stopped running this file. It should just be removed.

w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/4821/)

jugglinmike:
Thanks, James!

w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/4821/)

jugglinmike:
@bobholt Above, the bot appears to have built this branch in response to the merge event. My guess is that it is subscribed to one event too many. Whatever the case, I think you'll agree that this inefficiency is pretty low-priority. Just wanted to give you a heads-up!

PR #4823, opened 2017-02-14
Fix standalone tests using queryImage()

jeisinger:
They broke after changes to queryImage were upstreamed from mozilla, but
the standalone tests weren't touched


estark37:
lgtm, thanks!

PR #4828, opened 2017-02-14
[html] Account for bug that causes instability

jugglinmike:
The Chromium browser currently implements incorrect semantics regarding
the `sheet` property of the `link` element. Defer assertions until such
time as incorrect behavior is known to be observable.

### DIFF
--- a/html/semantics/document-metadata/styling/LinkStyle.html
+++ b/html/semantics/document-metadata/styling/LinkStyle.html
@@ -31,7 +31,7 @@
     <script>
       function waitForLinks(done) {
         function poll() {
-          if (getComputedStyle(target).textDecoration === "underline") {
+          if (getComputedStyle(target).zIndex === "1234") {
             setTimeout(done, 0);
             document.body.removeChild(beacon);
           } else {
--- a/html/semantics/document-metadata/styling/LinkStyle.html
+++ b/html/semantics/document-metadata/styling/LinkStyle.html
@@ -29,14 +29,46 @@
     <div id="test" style="display:none">STYLING TEST</div>
 
     <script>
-      test(function() {
-        var style = null,
-            i;
-        for (i = 1; i < 5; i++) {
-          style = document.getElementById("style" + i);
-          assert_equals(style.sheet, null, "The sheet attribute of style" + i + " should be null.");
-          assert_false(style.disabled, "The disabled attribute of style" + i + " should be false.");
+      function waitForLinks(done) {
+        function poll() {
+          if (getComputedStyle(target).textDecoration === "underline") {
+            setTimeout(done, 0);
+            document.body.removeChild(beacon);
+          } else {
+            setTimeout(poll, 50);
+          }
         }
+
+        var beacon = document.createElement("link");
+        var target = document.getElementById("test");
+        beacon.setAttribute("rel", "stylesheet");
+        beacon.setAttribute("href", "./support/beacon.css");
+
+        document.body.appendChild(beacon);
+        poll();
+      }
+
+      /**
+       * Browsers may incorrectly issue requests for these resources and defer
+       * definition of the `sheet` attribute until after loading is complete.
+       * In such cases, synchronous assertions regarding the absence of
+       * attributes will spuriously pass.
+       *
+       * In order to account for this incorrect behavior (exhibited at the time
+       * of this writing most notably by the Chromium browser), defer the
+       * assertions until a later-specified stylesheet has been loaded.
+       */
+      async_test(function(t) {
+        waitForLinks(t.step_func(function() {
+          var style = null,
+              i;
+          for (i = 1; i < 5; i++) {
+            style = document.getElementById("style" + i);
+            assert_equals(style.sheet, null, "The sheet attribute of style" + i + " should be null.");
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Chrome (unstable channel) #
Testing web-platform-tests at revision 62cb0c66beb65c57c7b19d773ecb3f7786872f13
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="http://w3c-test.org/submissions/4828/html/semantics/document-metadata/styling/LinkStyle.html">/html/semantics/document-metadata/styling/LinkStyle.html</a></summary>


|                                                                        Subtest                                                                        | Results |
|-------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
|                                                                                                                                                       | TIMEOUT |
| `The LinkStyle interface's sheet attribute must return StyleSheet object; the disabled attribute must be same as the StyleSheet's disabled attribute` | PASS    |
| `The LinkStyle interface's sheet attribute must return null if the corresponding element is not in a Document`                                        | PASS    |
| `The title must be the same as the value of the element's title content attribute`                                                                    | PASS    |
| `The media must be the same as the value of the element's media content attribute, or the empty string if it is omitted`                              | PASS    |
| `The LinkStyle interface's sheet attribute must return null; the disabled attribute must be false`                                                    | TIMEOUT |

</details>
</details>

w3c-bots:
# Firefox (nightly channel) #
Testing web-platform-tests at revision 62cb0c66beb65c57c7b19d773ecb3f7786872f13
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="http://w3c-test.org/submissions/4828/html/semantics/document-metadata/styling/LinkStyle.html">/html/semantics/document-metadata/styling/LinkStyle.html</a></summary>


|                                                                        Subtest                                                                        | Results |
|-------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
|                                                                                                                                                       | OK      |
| `The LinkStyle interface's sheet attribute must return StyleSheet object; the disabled attribute must be same as the StyleSheet's disabled attribute` | PASS    |
| `The LinkStyle interface's sheet attribute must return null if the corresponding element is not in a Document`                                        | PASS    |
| `The title must be the same as the value of the element's title content attribute`                                                                    | PASS    |
| `The media must be the same as the value of the element's media content attribute, or the empty string if it is omitted`                              | PASS    |
| `The LinkStyle interface's sheet attribute must return null; the disabled attribute must be false`                                                    | PASS    |

</details>
</details>

zcorpan:
Is there a bug filed for this on Chromium?

Is it possible to write a test that reliably fails if the bug is present?

zcorpan:
cc @rune-opera

tkent-google:
Here is a question about the test:
https://github.com/w3c/web-platform-tests/issues/2117

jugglinmike:
Somehow I missed that issue. I've added a comment justifying the test. Here's the relevant bug for the Chromium project: https://bugs.chromium.org/p/chromium/issues/detail?id=692204&can=4&q=&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Component%20Status%20Owner%20Summary%20OS%20Modified

jugglinmike:
> Is it possible to write a test that reliably fails if the bug is present?

That's the intent of this patch. The initial version caused a timeout in Chrome
because it didn't account for variations in the `textDocoration` computed
property (which differs between Chromium and Chrome). I've updated the load
detection mechanism to use the `zIndex` property instead.

As you can see in the stability test results above, Chrome now consistently
fails while Firefox continues to pass.

zcorpan:
I don't see `textDecoration` in the diff here... Isn't the issue here that the `<script>` is executed while there are pending stylesheets, or something? It would work to work around that by waiting for `window.onload` instead, correct?

Chrome fails because it's non-null when the test expects it to be null - https://bugs.chromium.org/p/chromium/issues/detail?id=692560 (we should decide what we want to happen there, but doesn't need to block this PR, and seems orthogonal to the bug this PR is working around).

zcorpan:
The test that checks the later stylesheets (`for (i = 5; i < 8; i++) {`) should also do the waiting?

zcorpan:
Oh I see you changed `textDecoration` in your fixup.

jugglinmike:
> It would work to work around that by waiting for `window.onload` instead,
> correct?

Yes, and that is far simpler/more readable. I've updated the patch to do that.

> The test that checks the later stylesheets (`for (i = 5; i < 8; i++) {`)
> should also do the waiting?

It couldn't hurt! (For reference: this is not technically necessary because the
Chromium instability is limited to "alternate" stylesheets with a non-empty
`name` attribute. In those cases, it exhibits a strange "deferred loading"
behavior: the resource is fetched, and that operation influences the timing of
the `onload` event, but it does not block page evaluation (despite the `<link>`
tag being declared in the head).

jugglinmike:
@zcorpan Is there anything else I can do to help land this?

zcorpan:
LGTM.

I'd still like a new test for the bug -- that alternative stylesheets aren't blocking scripts (or if we should change HTML to stop blocking scripts for alternate stylesheets?). But that can be separate PR. I'll file an issue.

PR #4840, opened 2017-02-15
close() should be in DedicatedWorkerGlobalScope and SharedWorkerGlobalScope only,

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1336043


jgraham:
Code reviewed upstream.

PR #4842, opened 2017-02-15
Split effect value tests into separate files;

jgraham:
The file naming here is based on the existing effect-value-context.html file,
i.e. we break up all the tests for the calculation the effect value into
separate files named effect-value-***.html.

MozReview-Commit-ID: LY46vX3mHh7


jgraham:
Code reviewed upstream.

PR #4844, opened 2017-02-15
Move tests for effect easing to timing-model;

jgraham:
This set of tests are really just testing that we apply the timing function to
the animation effect so they belong in the appropriate part of the timing model
tests (and should check getComputedTiming not getComputedStyle).

I've also started to update tests to ES6 where appropriate since it seems
arrow functions, template literals, etc. are all supported on all UAs that
are implementing or likely to implement Web Animations.

MozReview-Commit-ID: 3kXao0Xi0BA


jgraham:
Code reviewed upstream.

PR #4845, opened 2017-02-15
Rework tests for linear-equivalent cubic-bezier timing functions from effect-easing.html;

jgraham:
The purpose of these tests appears to be to check that a linear-equivalent
cubic-bezier timing function (e.g. 'cubic-bezier(0, 0, 0, 0)') does not affect
the result such as clamping values out of the [0, 1] range.

This test really is testing the calculation of the 'transformed distance' in
the "The effect value of a keyframe effect" so we move the test there and
rework it to more clearly test what it is intended to cover.

MozReview-Commit-ID: 9sEr7MlVZKL


jgraham:
Code reviewed upstream.

PR #4846, opened 2017-02-15
Rename effect easing tests to easing tests;

jgraham:
These tests are generic enough to be used for either effect easing or keyframe
easing.

MozReview-Commit-ID: 5cpnkiCv0z1


jgraham:
Code reviewed upstream.

PR #4847, opened 2017-02-15
Simplify invalid easing tests;

jgraham:
This doesn't need to be an array of objects when a simple array would do.

MozReview-Commit-ID: 1gtdhG5RPSy


jgraham:
Code reviewed upstream.

PR #4850, opened 2017-02-15
Drop step timing function for compositor animations;

jgraham:
This seems like a Gecko-specific test and even then it's not clear why we
expect the result could be different in this case.

MozReview-Commit-ID: Ix8jZLobwcA


jgraham:
Code reviewed upstream.

PR #4851, opened 2017-02-15
Drop checking width in step timing function tests;

jgraham:
We only use this once to test the timing function when applied to a keyframe.
Now that we have tests for this in

  animation-model/keyframe-effects/effect-value-transformed-distance.html

we can drop the check here and simplify these tests considerably.

Also, 'progress' is always set so we can drop the check for an undefined value.

MozReview-Commit-ID: 39ajHZBRWBf


jgraham:
Code reviewed upstream.

PR #4852, opened 2017-02-15
Add missing tests for steps-end in delay phase;

jgraham:
(Once we remove the special clamping behvaior later in this patch series, the
added test here will fail if we don't add special handling for the case
when the progress is zero and we are in the delay phase.)

MozReview-Commit-ID: Dnon2soE1Se


jgraham:
Code reviewed upstream.

PR #4853, opened 2017-02-15
Move tests for timing functions with inputs outside [0, 1];

jgraham:
These should be part of the tests for the transformed distance since that is
currently the only place where they can occur.

This patch also revises the test descriptions somewhat to make it clearer what
is being tested.

MozReview-Commit-ID: D4YfAiZUBYR


jgraham:
Code reviewed upstream.

PR #4858, opened 2017-02-15
Strip ':' and anything following it from the string passed to the location.protocol setter.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1328894


jgraham:
Code reviewed upstream.

PR #4861, opened 2017-02-15
Implement the same behavior for <object>-inside-<object> and <object>-inside-mediaelement as we do for <embed> already.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1332956


jgraham:
Code reviewed upstream.

PR #4862, opened 2017-02-15
Remove the service-worker version of get-host-info.sub.js.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1338173


jgraham:
Code reviewed upstream.

mkruisselbrink:
This should have failed as a merge conflict, as the file that this removed changed in WPT and doesn't match what the mozilla change tried to delete. This will also cause at least some service worker tests to now start failing.

PR #4865, opened 2017-02-15
Split the IDLs out of the HTML interfaces web platform test so we can run it in both windows and workers.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1306170


jgraham:
Code reviewed upstream.

PR #4875, opened 2017-02-15
Revert "Remove the service-worker version of get-host-info.sub.js."

mkruisselbrink:
Reverts w3c/web-platform-tests#4862

The two get-host-info.sub.js files should be sync-ed before the service worker version can be deleted.


mkruisselbrink:
Also new tests got added in WPT that refer to the service worker version of get-host-info.sub.js, which would be broken after the deletion.

jgraham:
Please fix the common file rather than reverting this change.

mkruisselbrink:
> Please fix the common file rather than reverting this change.

I don't currently have the bandwidth to fix this. This change conflicts with changes made to WPT, so it seems perfectly reasonable to revert until the original author has a chance to fix the merge conflicts.

PR #4879, opened 2017-02-15
[service-workers] Avoid race condition in test

jugglinmike:
The parent document in this test first sends a message to the
ServiceWorker and later triggers a `fetch` event. Prior to this commit,
the test initiated these two operations independently but required that
the Service Worker's `onmessage` handler was invoked prior to its
`onfetch` handler. However, because user agents may define independent
task queues for `fetch` and `postMessage` [1], there the sequence in
which the tasks are handled is not guaranteed. By incorrectly assuming
the determinacy of event order, the test created a race condition in
user agents that maintain two queues.

Re-factor the test to force a consistent execution order.

[1] From https://html.spec.whatwg.org/multipage/webappapis.html#task-queue

> Each task is defined as coming from a specific task source. All the
> tasks from one particular task source and destined to a particular
> event loop (e.g. the callbacks generated by timers of a Document, the
> events fired for mouse movements over that Document, the tasks queued
> for the parser of that Document) must always be added to the same task
> queue, but tasks from different task sources may be placed in
> different task queues.

### DIFF
--- a/service-workers/service-worker/request-end-to-end.https.html
+++ b/service-workers/service-worker/request-end-to-end.https.html
@@ -22,22 +22,34 @@ t.step(function() {
 
     function onRegister(registration) {
         var sw = registration.installing;
+        sw.addEventListener('statechange', t.step_func(function(event) {
+            if (event.target.state == 'activated') {
+                onActive(sw);
+            }
+        }));
+    }
+
+    function onActive(sw) {
         var port = sendMessagePort(sw);
         port.addEventListener('message', t.step_func(function(event) {
             onMessage(event);
         }), false);
         port.start();
-        sw.addEventListener('statechange', t.step_func(function(event) {
-            if (event.target.state == 'activated')
-                onActive();
-        }));
-    }
-
-    function onActive() {
-        with_iframe(scope).then(function(f) { frames.push(f); });
     }
 
     function onMessage(event) {
+        if (event.data === 'received port') {
+            onPortReady();
+        } else {
+            onResult(event);
+        }
+    }
+
+    function onPortReady() {
+        with_iframe(scope).then(function(f) { frames.push(f); });
+    }
+
+    function onResult(event) {
         assert_equals(
             event.data.url,
             location.href.substring(0, location.href.lastIndexOf('/') + 1) +
--- a/service-workers/service-worker/resources/request-end-to-end-worker.js
+++ b/service-workers/service-worker/resources/request-end-to-end-worker.js
@@ -1,4 +1,8 @@
 var port = undefined;
+// Create a then-able object that is never resolved.
+function createPending() {
+  return { then: createPending };
+}
 
 onmessage = function(e) {
   var message = e.data;
@@ -6,6 +10,11 @@ onmessage = function(e) {
     port = message.port;
 
     port.postMessage('received port');
+    // The ServiceWorker which handles the "message" event must persist long
+    // enough to handle the subsequent "fetch" event. To promote test
+    // simplicity, the worker prevents its own termination indefinitely via a
+    // then-able that is never resolved.
+    e.waitUntil(createPending());
   }
 };
 
--- a/service-workers/service-worker/resources/request-end-to-end-worker.js
+++ b/service-workers/service-worker/resources/request-end-to-end-worker.js
@@ -4,6 +4,8 @@ onmessage = function(e) {
   var message = e.data;
   if (typeof message === 'object' && 'port' in message) {
     port = message.port;
+
+    port.postMessage('received port');
   }
 };
 


service-workers/service-worker/request-end-to-end.https.html
```diff
@@ -22,22 +22,34 @@
...
+            onResult(event);
+        }
+    }
+
+    function onPortReady() {
         with_iframe(scope).then(function(f) { frames.push(f); });
     }
 
-    function onMessage(event) {
+    function onResult(event) {
```
mkruisselbrink: unrelated to this change, but at some level the call to onResult needs to be wrapped in t.step/t.step_func to make sure the asserts get attributed to the correct test (pretty much all the other async code also needs to be wrapped to properly attribute exceptions of course). Or alternatively since the entire file only has one async test anyway, we could just get rid of the async_test completely turning this into a single page test (but some of the helper methods like service_worker_unregister_and_done wouldn't work anymore then)

service-workers/service-worker/resources/request-end-to-end-worker.js
```diff
@@ -4,6 +4,8 @@ onmessage = function(e) {
   var message = e.data;
   if (typeof message === 'object' && 'port' in message) {
     port = message.port;
+
+    port.postMessage('received port');
   }
 };
```
mkruisselbrink: nit: technically this test is still flaky, as nothing is stopping a browser from killing the service worker as soon as the message event is handled, and restarting it before the fetch event is dispatch. To fix that you should call waitUntil on the message event with a promise you either don't resolve, or at least don't resolve until the fetch event handler.

w3c-bots:
# Chrome (unstable channel) #
Testing web-platform-tests at revision 92e35a213d0ad08662ba0a4eea5246f4aff170f5
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="https://w3c-test.org/submissions/4879/service-workers/service-worker/request-end-to-end.https.html">/service-workers/service-worker/request-end-to-end.https.html</a></summary>


|        Subtest        | Results |
|-----------------------|---------|
|                       | OK      |
| `Request: end-to-end` | FAIL    |

</details>
</details>

w3c-bots:
# Firefox (nightly channel) #
Testing web-platform-tests at revision 92e35a213d0ad08662ba0a4eea5246f4aff170f5
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="https://w3c-test.org/submissions/4879/service-workers/service-worker/request-end-to-end.https.html">/service-workers/service-worker/request-end-to-end.https.html</a></summary>


|        Subtest        | Results |
|-----------------------|---------|
|                       | OK      |
| `Request: end-to-end` | PASS    |

</details>
</details>

service-workers/service-worker/request-end-to-end.https.html
jugglinmike: Happy to address this in a follow up commit

service-workers/service-worker/resources/request-end-to-end-worker.js
jugglinmike: I've pushed up a commit to address this. I'm using a custom "pending thenable" for this purpose in order to limit the number of features in this test. If you'd prefer to use a true Promise instance, say the word and I'll be happy to do so.

mkruisselbrink:
Thanks! lgtm

jugglinmike:
Thanks for the review @mkruisselbrink! Would you mind merging this for me?

service-workers/service-worker/resources/request-end-to-end-worker.js
```diff
@@ -1,9 +1,20 @@
...
   var message = e.data;
   if (typeof message === 'object' && 'port' in message) {
     port = message.port;
+
+    port.postMessage('received port');
+    // The ServiceWorker which handles the "message" event must persist long
+    // enough to handle the subsequent "fetch" event. To promote test
+    // simplicity, the worker prevents its own termination indefinitely via a
+    // then-able that is never resolved.
+    e.waitUntil(createPending());
```
wanderview: So, this changed caused a number of other tests to fail in firefox.  Consider:

1. This test uses a scope of 'resources/blank.html' which many other tests also use.
2. It keeps the service worker thread alive until the browser force kills it.
3. We have agreement to delay marking a SW redundant until it goes idle in https://github.com/w3c/ServiceWorker/issues/916

This means this change keeps the SW registration from this test alive until the force kill timeout is triggered (5 minutes on FF) and blocks any other registration from being made on the same scope.

To work around this for now I'm moving this test to a more unique scope.  In the long run, though, I think this test should change to use a promise that resolves when the test explicitly sends a "done" message or something.  There is no need to keep the SW alive indefinitely.

Issue #3018, opened 2016-05-11
URL spec test coverage report


domenic:
FYI @annevk @Sebmaster @jasnell.

I managed to cobble this together:
- [URL state machine coverage report](https://dl.dropboxusercontent.com/u/20140634/url-spec-coverage/lib/url-state-machine.js.html): ~99%
- [URL object coverage report](https://dl.dropboxusercontent.com/u/20140634/url-spec-coverage/lib/URL-impl.js.html): ~94%

Areas of lacking coverage are:
- some origin serialization stuff (some purposefully untested)
- The href setter
- One early-return case in the hash setter
- One "else" branch in IPv4 parsing

annevk:
Oh sweet, can you generate new reports? Would love to add more tests, in particular now with Safari actively implementing.

domenic:
Yeah, it's just `npm run coverage` from within the whatwg-url project. I've generated a new report at the same URL and updated the OP. Coverage went from 90 -> 95 and 85 -> 94 for those two files.

jasnell:
Very nice.

domenic:
Updated the OP after #4882. Very nice.

annevk:
Can we also get coverage on IDNA, which is just a single method in this library?

GPHemsley:
@domenic The links here are dead. Is this something that we're still looking to have? (I'm guessing yes.) Does this issue serve that purpose?

domenic:
I think we got to 100% coverage at some point. I'm no longer actively tracking this, but you can checkout jsdom/whatwg-url, run `npm install`, then `npm run coverage` to get a report. I'll close this now since this issue isn't too helpful.

PR #4882, opened 2017-02-16
URL: improve coverage of conditions in the standard

annevk:
@domenic I found:

1. One bug in whatwg-url
2. One check that should be removed from the standard
3. We don’t test empty input to the IPv4 parser. I think that’s fine as it’s currently not possible, but who knows what we do going forward. Alternatively, we add an assert to the standard or some such.
4. There’s some stuff around origins that’s not tested. Blob URLs need to be sorted still and file URLs we don’t have agreement on thus far, though arguably whatever they return should serialize to null.


annevk:
I get 98.5% with this by the way.

w3c-bots:
# Chrome (unstable channel) #
Testing web-platform-tests at revision 548e68ac8e79aaf159ec25d04d210bdcec05d640
Starting 10 test iterations

annevk:
When committing this we should probably not close #3018 quite yet. I had forgotten to study `URL-impl.js.html` in more detail. I'd rather not block landing this on doing that work though, especially as some of it such as javascript URLs requires quite a bit of work.

domenic:
> One bug in whatwg-url

What's the bug?

annevk:
The stripping of U+FFFD in paths and fragments.

PR #4745, opened 2017-02-06
Sort stability test results by execution order

jugglinmike:
This should resolve gh-4515.

PR #4916, opened 2017-02-17
Raise stability errors to be visible in PR comments

### LINKED ISSUE #4515: wpt-stability-bot: report results in test order rather than Python (unordered) dict order

bobholt:
This raises the failing test messages into the results tables in the PR comment (see screenshot below, and presumably the w3c-bots comments on this PR). (cc @RByers - is this what you're looking for?)

Note: this includes @jugglinmike's commits from https://github.com/w3c/web-platform-tests/pull/4745 (including one that causes an instability for demonstration purposes, and which should be removed). His work on that PR is required by this one.

- Closes #4745 
- Fixes #4515

### DIFF
--- a/check_stability.py
+++ b/check_stability.py
@@ -489,7 +489,11 @@ def setup_log_handler():
                 if test["name"] == data["test"]:
                     return test
 
-            test = {"name": data["test"], "subtests": [], "status": {"OK": 0, "FAIL": 0}}
+            test = {
+                "name": data["test"],
+                "subtests": [],
+                "status": {"OK": 0, "FAIL": 0}
+            }
             self.results.append(test)
             return test
 
@@ -500,7 +504,11 @@ def setup_log_handler():
                 if subtest["name"] == data["subtest"]:
                     return subtest
 
-            subtest = {"name": data["subtest"], "status": { "PASS": 0, "FAIL": 0 }}
+            subtest = {
+                "name": data["subtest"],
+                "status": {"PASS": 0, "FAIL": 0},
+                "messages": set()
+            }
             test["subtests"].append(subtest)
 
             return subtest
@@ -508,6 +516,8 @@ def setup_log_handler():
         def test_status(self, data):
             subtest = self.find_or_create_subtest(data)
             subtest["status"][data["status"]] += 1
+            if (data.get("message")):
+                subtest["messages"].add(data["message"])
 
         def test_end(self, data):
             test = self.find_or_create_test(data)
@@ -548,7 +558,7 @@ def process_results(log, iterations):
     for test in results:
         for subtest in test["subtests"]:
             if is_inconsistent(subtest["status"], iterations):
-                inconsistent.append((test["name"], subtest["name"], subtest["status"]))
+                inconsistent.append((test["name"], subtest["name"], subtest["status"], subtest["messages"]))
     return results, inconsistent
 
 
@@ -595,9 +605,14 @@ def table(headings, data, log):
 def write_inconsistent(inconsistent, iterations):
     """Output inconsistent tests to logger.error."""
     logger.error("## Unstable results ##\n")
-    strings = [("`%s`" % markdown_adjust(test), ("`%s`" % markdown_adjust(subtest)) if subtest else "", err_string(results, iterations))
-               for test, subtest, results in inconsistent]
-    table(["Test", "Subtest", "Results"], strings, logger.error)
+    strings = [(
+        "`%s`" % markdown_adjust(test),
+        ("`%s`" % markdown_adjust(subtest)) if subtest else "",
+        err_string(results, iterations),
+        ("`%s`" % markdown_adjust(";".join(messages))) if len(messages) else ""
+    )
+               for test, subtest, results, messages in inconsistent]
+    table(["Test", "Subtest", "Results", "Messages"], strings, logger.error)
 
 
 def write_results(results, iterations, comment_pr):
@@ -626,11 +641,14 @@ def write_results(results, iterations, comment_pr):
                         (baseurl, pr_number, name, name))
         else:
             logger.info("### %s ###" % name)
-        strings = [("", err_string(test["status"], iterations))]
-        strings.extend(((("`%s`" % markdown_adjust(subtest["name"])) if subtest
-                         else "", err_string(subtest["status"], iterations))
-                        for subtest in test["subtests"]))
-        table(["Subtest", "Results"], strings, logger.info)
+        strings = [("", err_string(test["status"], iterations), "")]
+
+        strings.extend(((
+            ("`%s`" % markdown_adjust(subtest["name"])) if subtest else "",
+            err_string(subtest["status"], iterations),
+            ("`%s`" % markdown_adjust(';'.join(subtest["messages"]))) if len(subtest["messages"]) else ""
+        ) for subtest in test["subtests"]))
+        table(["Subtest", "Results", "Messages"], strings, logger.info)
         if pr_number:
             logger.info("</details>\n")
 
--- a/check_stability.py
+++ b/check_stability.py
@@ -11,7 +11,6 @@ import tarfile
 import zipfile
 from abc import ABCMeta, abstractmethod
 from cStringIO import StringIO as CStringIO
-from collections import defaultdict
 from ConfigParser import RawConfigParser
 from io import BytesIO, StringIO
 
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Chrome (unstable channel) #
Testing web-platform-tests at revision 42af9e70f4d0ba6c7cf5db9dad1366f1b262a197
Using browser at version 58.0.3013.3 dev
Starting 10 test iterations
## Unstable results ##


|          Test          |                 Subtest                 |          Results           |                   Messages                   |
|------------------------|-----------------------------------------|----------------------------|----------------------------------------------|
| `/url/historical.html` | `URL.domainToASCII should be undefined` | **FAIL: 3/10, PASS: 7/10** | `assert_equals: expected true but got false` |

## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="http://w3c-test.org/submissions/4916/url/historical.html">/url/historical.html</a></summary>


|                   Subtest                    |          Results           |                                        Messages                                       |
|----------------------------------------------|----------------------------|---------------------------------------------------------------------------------------|
|                                              | OK                         |                                                                                       |
| `searchParams on location object`            | PASS                       | ``                                                                                    |
| `searchParams on a element`                  | PASS                       | ``                                                                                    |
| `searchParams on area element`               | PASS                       | ``                                                                                    |
| `Setting URL's href attribute and base URLs` | FAIL                       | `assert_throws: function "function () {\n    url.href = "./bar";\n  }" did not throw` |
| `URL.domainToASCII should be undefined`      | **FAIL: 3/10, PASS: 7/10** | `assert_equals: expected true but got false`                                          |
| `URL.domainToUnicode should be undefined`    | PASS                       | ``                                                                                    |

</details>
</details>

w3c-bots:
# Firefox (nightly channel) #
Testing web-platform-tests at revision 42af9e70f4d0ba6c7cf5db9dad1366f1b262a197
Starting 10 test iterations
## Unstable results ##


|          Test          |                 Subtest                 |          Results           |                   Messages                   |
|------------------------|-----------------------------------------|----------------------------|----------------------------------------------|
| `/url/historical.html` | `URL.domainToASCII should be undefined` | **FAIL: 5/10, PASS: 5/10** | `assert_equals: expected true but got false` |

## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="http://w3c-test.org/submissions/4916/url/historical.html">/url/historical.html</a></summary>


|                   Subtest                    |          Results           |                   Messages                   |
|----------------------------------------------|----------------------------|----------------------------------------------|
|                                              | OK                         |                                              |
| `searchParams on location object`            | PASS                       | ``                                           |
| `searchParams on a element`                  | PASS                       | ``                                           |
| `searchParams on area element`               | PASS                       | ``                                           |
| `Setting URL's href attribute and base URLs` | PASS                       | ``                                           |
| `URL.domainToASCII should be undefined`      | **FAIL: 5/10, PASS: 5/10** | `assert_equals: expected true but got false` |
| `URL.domainToUnicode should be undefined`    | PASS                       | ``                                           |

</details>
</details>

RByers:
Love it!

zcorpan:
Should malke the message cell empty instead of `` (for passing tests).

bobholt:
Thanks @zcorpan. Resolved. (you can see this in the updated w3c-bot comments, above). I'll wait to remove the intentionally-flakey test until someone else confirms.

zcorpan:
Looks good!

bobholt:
Adding screenshot of flaky test result from this PR for posterity before removing that DEBUG commit:

raise_errors

jgraham:
Apart from #4985 (which is an incomplete solution to the presented problem), I think this commit has a number of issues that need to be addressed. In particular I don't think that we should use an implementation that's quardatic in the number of tests/subtests and I don't think we should hardcode test statuses. 

Sorry for not catching this earlier, but I'm tempted to revert this and fix the issues rather than patching them piecemeal.

check_stability.py
```diff
@@ -483,31 +482,69 @@ class LogHandler(reader.LogHandler):
...
-            self.results[data["test"]][None][data["status"]] += 1
+            test = self.find_or_create_test(data)
+            test["status"][data["status"]] += 1
 
 
 def is_inconsistent(results_dict, iterations):
     """Return whether or not a single test is inconsistent."""
-    return len(results_dict) > 1 or sum(results_dict.values()) != iterations
+    values = results_dict.values()
+    nonzero_results = filter(lambda count: count > 0, values)
```
jgraham: I think this will disappear when we use a `defaultdict`, but in general using comprehensions is preferred over using a filter in cases like this e.g. `any(item for item in values if item > 0)`.

check_stability.py
```diff
@@ -483,31 +482,69 @@ class LogHandler(reader.LogHandler):
...
         def test_end(self, data):
-            self.results[data["test"]][None][data["status"]] += 1
+            test = self.find_or_create_test(data)
+            test["status"][data["status"]] += 1
 
 
 def is_inconsistent(results_dict, iterations):
     """Return whether or not a single test is inconsistent."""
-    return len(results_dict) > 1 or sum(results_dict.values()) != iterations
+    values = results_dict.values()
```
jgraham: This isn't really a change, but there isn't really any need to get an actual list of values vs getting an iterator here (not sure which is actually fastest in this case, but I would use the iterator unless we know that perf. is a concern and that the explicit list is faster).

check_stability.py
```diff
@@ -483,31 +482,69 @@ class LogHandler(reader.LogHandler):
...
+            test = self.find_or_create_test(data)
+
+            for subtest in test["subtests"]:
+                if subtest["name"] == data["subtest"]:
+                    return subtest
+
+            subtest = {
+                "name": data["subtest"],
+                "status": {"PASS": 0, "FAIL": 0},
+                "messages": set()
```
jgraham: Do we not care about the count of messages when there is more than one?

check_stability.py
```diff
@@ -483,31 +482,69 @@ class LogHandler(reader.LogHandler):
...
+        def find_or_create_subtest(self, data):
+            test = self.find_or_create_test(data)
+
+            for subtest in test["subtests"]:
+                if subtest["name"] == data["subtest"]:
+                    return subtest
+
+            subtest = {
+                "name": data["subtest"],
+                "status": {"PASS": 0, "FAIL": 0},
```
jgraham: As above.

check_stability.py
```diff
@@ -483,31 +482,69 @@ class LogHandler(reader.LogHandler):
...
+                "subtests": [],
+                "status": {"OK": 0, "FAIL": 0}
+            }
+            self.results.append(test)
+            return test
+
+        def find_or_create_subtest(self, data):
+            test = self.find_or_create_test(data)
+
+            for subtest in test["subtests"]:
```
jgraham: As above.

check_stability.py
```diff
@@ -483,31 +482,69 @@ class LogHandler(reader.LogHandler):
...
+
+        def find_or_create_test(self, data):
+            for test in self.results:
+                if test["name"] == data["test"]:
+                    return test
+
+            test = {
+                "name": data["test"],
+                "subtests": [],
+                "status": {"OK": 0, "FAIL": 0}
```
jgraham: This is not the full range of possible statuses. Let's use a defaultdict(int) to avoid depending on specific statuses here.

check_stability.py
```diff
@@ -483,31 +482,69 @@ class LogHandler(reader.LogHandler):
         Subclasses reader.LogHandler.
         """
         def __init__(self):
-            self.results = defaultdict(lambda: defaultdict(lambda: defaultdict(int)))
+            self.results = []
+
+        def find_or_create_test(self, data):
+            for test in self.results:
```
jgraham: This ends up being quadratic in the number of tests, which is bad. We could use an OrderedDict to store the top-level results and avoid this problem.

check_stability.py
```diff
@@ -483,31 +482,69 @@ class LogHandler(reader.LogHandler):
...
+    nonzero_results = filter(lambda count: count > 0, values)
+    return len(nonzero_results) > 1 or sum(values) != iterations
 
 
 def err_string(results_dict, iterations):
     """Create and return string with errors from test run."""
     rv = []
     total_results = sum(results_dict.values())
     for key, value in sorted(results_dict.items()):
+        if value == 0:
```
jgraham: This will go away.

check_stability.py
jgraham: Oh, sorry that's the wrong replacement, but still.

check_stability.py
bobholt: I was trying to be mindful of the limited space in the GitHub UI, but instead of outputting this:
```
your flakey test failed because of this reason; your flaket test failed because of this other reason
```

we could output something like this without much additional space-usage:

```
your flakey test failed because of this reason (4 time(s)); your flaket test failed because of this other reason (1 time(s))
```

PR #4941, opened 2017-02-21
Add blob URLs into urltestdata.json

watilde:
Related link:
+ https://github.com/nodejs/node/pull/11426

### DIFF
--- a/url/urltestdata.json
+++ b/url/urltestdata.json
@@ -5644,5 +5644,33 @@
     "input": "non-special://[:80/",
     "base": "about:blank",
     "failure": true
+  },
+  {
+    "input": "blob:https://example.com:443/",
+    "base": "about:blank",
+    "href": "blob:https://example.com:443/",
+    "protocol": "blob:",
+    "username": "",
+    "password": "",
+    "host": "",
+    "hostname": "",
+    "port": "",
+    "pathname": "https://example.com:443/",
+    "search": "",
+    "hash": ""
+  },
+  {
+    "input": "blob:d3958f5c-0777-0845-9dcf-2cb28783acaf",
+    "base": "about:blank",
+    "href": "blob:d3958f5c-0777-0845-9dcf-2cb28783acaf",
+    "protocol": "blob:",
+    "username": "",
+    "password": "",
+    "host": "",
+    "hostname": "",
+    "port": "",
+    "pathname": "d3958f5c-0777-0845-9dcf-2cb28783acaf",
+    "search": "",
+    "hash": ""
   }
 ]


annevk:
I guess I'm not against landing this, but https://github.com/whatwg/url/issues/127 might change the outcome of these tests with regards to the origin field. @domenic?

domenic:
I think what we're seeing here is confusion from implementers as to what the origin of completely-synthetic blob URLs should be. I.e., if I do `new URL("blob:https://example.com:443/")`, or set `a.href = "blob:https://example.com:443/"`, what should the `.origin` property return? Similarly for if I do `blob:d3958f5c-0777-0845-9dcf-2cb28783acaf` (where that is just created out of thin air, not via URL.createObjectURL) or `blob:asdf`.

I'm not sure what the desired answer to that is according to https://github.com/whatwg/url/issues/127. I guess that is what the comments from https://github.com/whatwg/url/issues/127#issuecomment-228970175 onward say. If we really don't know the answer, I think a better approach here would be to remove the `origin` field from this PR and put a big red warning in the URL spec while we figure this out.

annevk:
Per the current URL Standard it's pretty clear what it should be. That should just work. Per that issue it's something we might want to change at which point you'd get different results.

annevk:
Well, you might get different results.

domenic:
Sure, I'm just saying it sounds like this section of the spec isn't really settled, so it makes more sense to me to make it like file URLs (we don't test or spec the origin right now) than to treat it as settled and testable.

annevk:
Fair enough. @watilde want to remove the origin lines for now?

watilde:
@annevk Sure, I will do that now!

watilde:
I've removed the origin lines, and CI liked it :)

url/urltestdata.json
```diff
@@ -5644,5 +5644,15 @@
     "input": "non-special://[:80/",
     "base": "about:blank",
     "failure": true
+  },
+  {
+    "input": "blob:https://example.com:443/",
```
domenic: These tests are missing all the other fields, which causes them to fail in the WPT test runner. You need base, href, username, password, host, hostname, port, search, and hash.

domenic (changes requested):

domenic:
@watilde sorry, I did not see that you pushed a new commit addressing my review. However, why did you allow failure for parsing blob URLs?

watilde:
@domenic Oh sorry never mind it please :bow: I wanted to make sure the failed ci was from my commit or not.

domenic:
Oh I see :(. It looks like the CI's attempts to run Chrome are not working great the moment... I'll try to poke some people.

domenic:
It might be worth trying to rebase just in case they've been making changes to the CI stuff

watilde:
@domenic Hmm it seems to be still happening. I will do rebase again tomorrow or later then. Thanks anyway!

jgraham:
@annevk informs me the review issues were fixed, and the CI problem looks like a bug in Chrome, so merging.

PR #4988, opened 2017-02-22
Add tests for the realms involved in compiling event handlers

domenic:
Follows https://github.com/whatwg/html/pull/2286 and https://github.com/whatwg/html/issues/2248.

This is WIP; ideally I'd like to also test incumbent settings object. Also the spec PR needs updates based on the results of my testing.

/cc @bzbarsky @Ms2ger

### DIFF
--- a/html/webappapis/scripting/events/compile-event-handler-settings-objects.html
+++ b/html/webappapis/scripting/events/compile-event-handler-settings-objects.html
@@ -52,7 +52,6 @@ async_test(t => {
   iframe.src = "resources/compiled-event-handler-settings-objects-support.html";
   iframe.onload = t.step_func(() => {
     window.onmessage = t.step_func_done(event => {
-      console.log(event);
       assert_equals(event.source, iframe.contentWindow, "The source must be the iframe");
     });
 
--- a/html/webappapis/scripting/events/compile-event-handler-settings-objects.html
+++ b/html/webappapis/scripting/events/compile-event-handler-settings-objects.html
@@ -34,7 +34,8 @@ async_test(t => {
 
     window.onWindowLoaded = t.step_func_done(url => {
       const pathname = new URL(url).pathname;
-      assert_equals(pathname, "/html/webappapis/scripting/events/resources/foo.html");
+      assert_equals(pathname,
+        "/html/webappapis/scripting/events/resources/open-window.html");
       // This tests that the entry settings object used to resolve URLs in that window.open() was the same as that
       // of the node document (i.e. the iframe document), not e.g. this window.
     });
--- a/html/webappapis/scripting/events/compile-event-handler-settings-objects.html
+++ b/html/webappapis/scripting/events/compile-event-handler-settings-objects.html
@@ -1,0 +1,48 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Entry and incumbent settings objects when compiling an inline event handler</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<div id="log"></div>
+
+<script>
+"use strict";
+
+async_test(t => {
+  const iframe = document.createElement("iframe");
+  iframe.src = "resources/compiled-event-handler-settings-objects-support.html";
+  iframe.onload = t.step_func(() => {
+    const button = iframe.contentDocument.querySelector("button");
+    const compiled = button.onclick;
+
+    assert_equals(compiled.constructor, iframe.contentWindow.Function, "The constructor must be from the iframe");
+    assert_not_equals(compiled.constructor, Function, "The constructor must not be from this page");
+
+    t.done();
+  });
+
+  document.body.appendChild(iframe);
+
+}, "The Function instance must be created in the Realm of the node document");
+
+async_test(t => {
+  const iframe = document.createElement("iframe");
+  iframe.src = "resources/compiled-event-handler-settings-objects-support.html";
+  iframe.onload = t.step_func(() => {
+    const button = iframe.contentDocument.querySelector("button");
+    const compiled = button.onclick;
+
+    const openedWindow = compiled();
+    openedWindow.onload = t.step_func(() => {
+      const pathname = new URL(openedWindow.document.URL).pathname;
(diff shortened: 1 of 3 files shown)


w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/204410034)

# Firefox (nightly channel) #
Testing web-platform-tests at revision 362642e87b47e42b1f210f0efd1508a722a18877
Using browser at version BuildID 20170221110300; SourceStamp d84beb192e57e26846c82d3df3599381f4663792
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="http://w3c-test.org/submissions/4988/html/webappapis/scripting/events/compile-event-handler-settings-objects.html">/html/webappapis/scripting/events/compile-event-handler-settings-objects.html</a></summary>


|                                                         Subtest                                                         | Results |     Messages     |
|-------------------------------------------------------------------------------------------------------------------------|---------|------------------|
|                                                                                                                         | TIMEOUT |                  |
| `The Function instance must be created in the Realm of the node document`                                               | PASS    |                  |
| `The entry settings object while executing the compiled callback must be the entry settings object at compilation time` | TIMEOUT | `Test timed out` |

</details>
</details>

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/204410035)

# Chrome (unstable channel) #
Testing web-platform-tests at revision 362642e87b47e42b1f210f0efd1508a722a18877
Using browser at version 58.0.3018.3 dev
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="http://w3c-test.org/submissions/4988/html/webappapis/scripting/events/compile-event-handler-settings-objects.html">/html/webappapis/scripting/events/compile-event-handler-settings-objects.html</a></summary>


|                                                         Subtest                                                         | Results | Messages |
|-------------------------------------------------------------------------------------------------------------------------|---------|----------|
|                                                                                                                         | OK      |          |
| `The Function instance must be created in the Realm of the node document`                                               | PASS    |          |
| `The entry settings object while executing the compiled callback must be the entry settings object at compilation time` | PASS    |          |

</details>
</details>

domenic:
OK, these are ready to go. Encouragingly, they pass in Firefox, and the incumbent one fails in Chrome (which is encouraging because Chrome is known to "give up" on determining the incumbent in sufficiently-esoteric situations, of which this is probably one).

bzbarsky:
I don't follow the entry settings test.  What's foo.html?  Why would that onload ever fire?  I'm pretty sure error loads don't necessarily trigger onload...  It might be better to use a URL that actually resolves to something there.

Actually, I'm not sure I follow the incumbent settings test either.  How does that test the incumbent global of a compiled event handler?  Or is that not what it's trying to test?

domenic:
foo.html resolves to a 404 page, which fires the onload just fine... we can create a foo.html if you want though.

I'm not sure about the incumbent global test. I just kind of tried to follow your instructions, but I admit not understanding them super-well. Why did you think return value conversion was relevant to the incumbent global?

bzbarsky:
> foo.html resolves to a 404 page, which fires the onload just fine...

I guess I'm not sure why both Chrome and Firefox time out on that test, then, per test results above.

> Why did you think return value conversion was relevant to the incumbent global?

Because when the return value conversion is happening the incumbent global is whatever was pushed for the callback.

So this test is testing the incumbent global that gets stored when window.onbeforeunload is assigned, but that has nothing to do with event handler compilation.

You may want to leave that test and add one which does:

    <body onbeforeunload="return { toString: etc }">

to test the incumbent set up with event handler compilation.

bzbarsky:
And also possibly of interest is a `bodyElement.setAttribute('onbeforeunload', stuff)` where the current/incumbent/entry global at the setAttribute call point don't match the relevant global.  And then see which global ends up being used as incumbent.

domenic:
> I guess I'm not sure why both Chrome and Firefox time out on that test, then, per test results above.

They don't if you actually run it in a browser instead of in CI; very interesting.

I'll try to see if I can get a good incumbent test working...

domenic:
OK, I worked on that, but now the test is failing in Firefox (the "source" ends up being the parent frame, not the iframe), which is unexpected given your comments in https://github.com/whatwg/html/issues/2248#issuecomment-271342955...

bzbarsky:
That's odd.  I just tried this locally

    <!DOCTYPE html>
    <iframe src="subframe.html"></iframe>
    <script>
      onmessage = function(e) {
        console.log(e.source == frames[0]);
        console.log(e.source == this);
      }
    </script>

where `subframe.html` is:

    <body onbeforeunload="return { toString: parent.postMessage.bind(parent, 'PASS', '*') };">
    <a href="http://example.com">Click me</a>

and I get `true` and `false` logged in that order.

domenic:
Ah, I see. The problem is that a slight modification of that, where instead of clicking the button the outer frame does `frames[0].src = "about:blank"`, logs false/true. See https://ns-gooldndpxq.now.sh.

Can you think of a reason that would be?

bzbarsky:
So first of all, that test is not navigating the subframe, just assigning a .src property on its window (as opposed to the iframe element).  The beforeunload only fires when the whole page is unloaded.

But that shouldn't affect the behavior.  What does affect it is that gecko does NOT in fact set up an incumbent global on event handlers compiled from strings.  Which means the incumbent global ends up being "yeah, whatever is on the stack" if there's script on the stack (and the entry global if there's nothing on the stack).  But then if that incumbent global is not same origin-domain with the current global, we reset it to the current global.

In the "unload the page via browser UI" case, there's actually privileged script that's part of the browser UI on the stack.  So we get that as the incumbent, then it fails the same origin-domain check and we reset the incumbent to current.  And our current at that point is the global of the postMessage function, which is the parent.

Anyway, this is a bug in Gecko.  Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1342513

domenic:
Ah, oops.

But in the end, the upside of this that the tests and spec patch are good to merge? :) The test in this PR does not appear to have the iframe vs. contentWindow confusion.

bzbarsky:
Yes, the test in the PR looks good to me!

PR #5046, opened 2017-03-06
[presentation-api] Add tests to check a newly created receiving browsing context

tomoyukilabs:
This PR adds tests to check if a receiving browsing context is initialized properly so that it complies with the restrictions and the features listed in [6.6.1 Creating a receiving browsing context](https://w3c.github.io/presentation-api/#creating-a-receiving-browsing-context).

- Tests to check HTTP authentication states and application cache have not been implemented yet.
- A test to check if `navigator.presentation.receiver` in a nested browsing context returns `null` or not (as in [6.2.2 Receiving user agent](https://w3c.github.io/presentation-api/#receiving-user-agent)) is also included.

### DIFF
--- a/presentation-api/receiving-ua/common.js
+++ b/presentation-api/receiving-ua/common.js
@@ -7,4 +7,40 @@
     toController: '0c382524-5738-4df0-837d-4f53ea8addc2',
     toReceiver: 'a9618cd1-ca2b-4155-b7f6-630dce953c44'
   }
+
+  // handle a test result received from a receiving page
+  const parseValue = value => {
+    let r;
+
+    // String
+    if (r = value.match(/^"(.*)"$/))
+      return r[1];
+    // Object
+    else if (r = value.match(/^object\s+"\[object\s+(.*)\]"$/))
+      return window[r[1]].prototype;
+    // Number, boolean, null, undefined
+    else
+      return JSON.parse(value);
+  };
+
+  window.parseResult = message => {
+    let r = message.match(/^(assert_.*):\s+(.*)$/);
+    const assertion = r[1];
+    const body = r[2];
+    switch (assertion) {
+      case 'assert_equals':
+        r = body.match(/^(.*)\s+expected\s+(true|false|null|\d+|"(.*)")\s+but\s+got\s+(true|false|null|\d+|(object\s+)?"(.*)")$/);
+        window[assertion](parseValue(r[4]), parseValue(r[2]), r[1]);
+        break;
+      case 'assert_true':
+      case 'assert_false':
+        r = body.match(/^(.*)\s+expected\s+(true|false)\s+got\s+(true|false)$/);
+        window[assertion](parseValue(r[3]), r[1]);
+        break;
+      case 'assert_unreached':
+        r = body.match(/^(.*)\s+Reached\s+unreachable\s+code$/);
+        window[assertion](r[1]);
+        break;
+    }
+  };
 })(window);
--- a/presentation-api/receiving-ua/common.js
+++ b/presentation-api/receiving-ua/common.js
@@ -13,34 +13,51 @@
     let r;
 
     // String
-    if (r = value.match(/^"(.*)"$/))
-      return r[1];
+    if (r = value.match(/^(\(string\)\s+)?"(.*)"$/))
+      return r[2];
     // Object
-    else if (r = value.match(/^object\s+"\[object\s+(.*)\]"$/))
-      return window[r[1]].prototype;
+    else if (r = value.match(/^(\(object\)\s+)?object\s+"\[object\s+(.*)\]"$/))
+      return window[r[2]].prototype;
     // Number, boolean, null, undefined
-    else
-      return JSON.parse(value);
+    else {
+      if (r = value.match(/^(\(\S+\)\s+)?(\S+)$/)) {
+        try {
+          return JSON.parse(r[2]);
+        } catch(e) {
+          return value;
+        }
+      }
+      else
+        return value;
+    }
   };
 
   window.parseResult = message => {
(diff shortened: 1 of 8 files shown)


w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/208132284)

# Firefox (nightly channel) #
Testing web-platform-tests at revision 4653c9815e0e94aa1c3c6dd67a273fc421d62717
Using browser at version BuildID 20170305110158; SourceStamp eb23648534779c110f3a1f2baae1849ae4a9c570
Starting 10 test iterations
No tests run.

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/208132285)

# Chrome (unstable channel) #
Testing web-platform-tests at revision 4653c9815e0e94aa1c3c6dd67a273fc421d62717
Using browser at version 58.0.3026.3 dev
Starting 10 test iterations
No tests run.

presentation-api/receiving-ua/support/PresentationReceiver_create_receiving-ua.html
```diff
@@ -0,0 +1,317 @@
...
+
+        // The Sandboxed auxiliary navigation browsing context flag
+        assert_equals(window.open('PresentationReceiver_unreached_receiving-ua.html'), null, 'Sandboxed auxiliary navigation browsing context flag is set.');
+
+        // The sandboxed top-level navigation browsing context flag (the following codes are expected to be ignored)
+        window.close();
+        document.querySelector('a').click();
+        location.href = 'PresentationReceiver_unreached_receiving-ua.html';
+
+        // The sandboxed modals flag (codes below are expected to be ignored)
```
tidoust: How is failure to ignore the following calls going to be detected?

I suspect the user agent would prompt the user if it does not ignore the calls and, provided the user does not answer the prompt, the test will fail with a timeout.

It could be good to be explicit about that:
1. by adding a comment as in "// If user agent prompts user, a timeout will occur and the test will eventually fail", and with user-friendly messages
2. by adding messages to users in the prompt as in "If you see this prompt, do not dismiss it and wait for test to time out"

Also `confirm` and `prompt` return a value (`false` and `null`), which could be checked as well.

tidoust (commented):
Thanks @tomoyukilabs. Now, that's an "heavy" test ;)

I do not think there's any good way to simplify it though. We could make it more atomic but that would mean creating a new manual test each time, so keeping all checks together looks good to me.

I may have a few other comments to make, but please see those inline already.

tomoyukilabs:
@tidoust Many thanks for reviewing this "heavy" test! ;)

I re-considered an overall mechanism for importing receiving side's test results. Now a controlling side runs  simpler codes which include a single `promise_test`. Also, this test shows explicit messages on a receiving side  (including `<iframe>`) during steps to check modal dialogs.

On the other hand, this test does not check a value returned by `confirm` and `prompt` yet, since they return `false` and `null` in both cases that a user chooses "cancel" and that the dialog is suppressed because of the sandboxed modals flag.

presentation-api/receiving-ua/common.js
```diff
@@ -7,4 +7,40 @@
     toController: '0c382524-5738-4df0-837d-4f53ea8addc2',
     toReceiver: 'a9618cd1-ca2b-4155-b7f6-630dce953c44'
   }
+
+  // handle a test result received from a receiving page
```
tidoust: What `parseValue` and `parseResult` gives us is great! One thing that I do not like is that these two functions will break if the format of the messages reported by the test harness changes slightly for some reason (e.g. because someone decides to clarify them further).

Could we make these functions more generic so that they cannot fail? Typically, what I have in mind is:
1. in `parseValue`: wrap the call to `JSON.parse` in a `try... catch` and report the value as-is if it cannot be parsed.
2. in `parseResult`: add a `default` case in the switch statement that calls `assert_false` for instance with the full message received.
3. in `parseResult`: if match fails, report the full message with an `assert_false` call as well.

In other words, my goal here would be to make sure the error gets reported no matter what. Now, calling `assert_false` might be confusing. It might make more sense to call `assert_unreached` with the message prefixed by something like "Unexpected failure message received from test harness: "

tidoust (commented):
Ha, it's a bit of a pity to have to reverse engineer the failure messages sent by the test harness, but it does improve the overall readability of the test!

See a couple of further comments inline. I propose to merge this afterwards, it's a great step forward. \o/

presentation-api/receiving-ua/support/PresentationReceiver_create_receiving-ua_child.html
```diff
@@ -6,48 +6,64 @@
...
+    if(event.data !== 'start')
+        return;
+
+    promise_test(t => {
+        const notice = document.getElementById('notice');
+        const modal = document.getElementById('modal');
+
+        // Presentation and PresentationReceiver
+        assert_true(navigator.presentation instanceof Presentation, 'navigator.presentation in a nested browsing context is an instanceof Presentation.');
+        //assert_equals(navigator.presentation.receiver, null, 'navigator.presentation.receiver in a nested browsing context is set to null.');
```
tomoyukilabs: Oops, it's simply my mistake. I'll fix it.

w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/5046/)

presentation-api/receiving-ua/common.js
tomoyukilabs: Okay. I have fixed them.

* If any `match` fails, `assert_unreached` is called with a default message.
* `JSON.parse` is now wrapped in `try... catch`.

tidoust (approved):

PR #5069, opened 2017-03-08
streams/: Fix issues found by eslint

ricea:
There is no requirement for tests to pass eslint, but I find it a useful tool
for debugging so I cleaned up a few non-controversial issues in the streams/
test directory.

Fixed:

- Unused variables
- Missing '()' invoking a constructor
- 'cancelPromise' is never reassigned. Use 'const' instead
- Missing semicolon
- '+' should be placed at the end of the line (subjective, but probably nobody
  cares enough to object)
- This line has 2 statements. Maximum allowed is 1
- Use the rest parameters instead of 'arguments'

Not changed:

- Variable 'ignoredReadPromise' is unused on purpose
- Lines > 120 (no universal agreement on this)
- Arrow function should not return assignment (no universal agreement)
- A space is required after '{' / before '}' (no universal agreement)

### DIFF
--- a/streams/piping/general.js
+++ b/streams/piping/general.js
@@ -155,7 +155,7 @@ promise_test(() => {
 }, 'Piping from a ReadableStream for which a chunk becomes asynchronously readable after the pipeTo');
 
 for (const preventAbort of [true, false]) {
-  promise_test(t => {
+  promise_test(() => {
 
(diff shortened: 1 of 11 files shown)


w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/5069/)

PR #5102, opened 2017-03-09
Test HTTP parsing

annevk:
Closes #5064.

PR #5227, opened 2017-03-27
Report: Inconsistency in HTTP 204 response body parsing

jugglinmike:
When a 204 response specifies as `Content-Length` header, different browsers exhibit different behaviors on different platforms. Although [the HTTP specification explicitly disallows this situation](https://tools.ietf.org/html/rfc7230#section-3.3.2), @annevk suggested that this might be behavior that should be accounted for in relevant W3C/WhatWG specifications.

I've opened this pull request to demonstrate the inconsistency. Please note that although the demo takes the form of a WPT test, it is *not* suitable for a number of reasons.

After checking out this branch, navigate to http://localhost:8000/http-demo/index.html . The tests perform the following actions:

1. Browser: perform a GET request for `demo.py?seq=1` by injecting a `<script>`
   tag into the document
2. Server: respond with `HTTP/1.1 204 No Content`, specifying a non-zero value
   for the `Content-Length` header, but omitting a body
3. Browser: in the `onload` event handler for the script created in step 1,
   perform a second GET request for `demo.py?seq=2` (again, by injecting a
   `<script>` tag into the document)
4. Server: write between zero and six bytes to the response stream that do
   *not* conform to HTTP
5. Server: respond with `HTTP/1.1 200 OK`, specifying a non-zero
   `Content-Length` and a complete body.

There are six sub-tests here: the first writes 0 non-HTTP (or "garbage") bytes in step 4; the following sub-tests write successively more bytes.

Results:

Browser               | Platform   | Behavior for request 1 | Behavior for request 2 (no "garbage)" | Behavior for request 2 (1-4 "garbage" bytes) | Behavior for request 2 (5 "garbage" bytes)
----------------------|------------|------------------------|---------------------------------------|----------------------------------------------|-------------------------------------------
Chromium 56           | GNU/Linux  | resolves successfully  | resolves successfully                 | resolves successfully                        | does not resolve
Chrome 59             | GNU/Linux  | resolves successfully  | resolves successfully                 | resolves successfully                        | reports network error `net::ERR_INVALID_HTTP_RESPONSE`
~~Chrome 57~~             | ~~Windows 10~~ | ~~resolves successfully~~  | ~~resolves successfully~~                 | ~~resolves with HTTP 502 error~~                 | ~~resolves with HTTP 502 error~~
Firefox 52            | GNU/Linux  | resolves successfully  | resolves successfully                 | resolves successfully                        | resolves successfully
~~Firefox 52~~            | ~~Windows 10~~ | ~~resolves successfully~~  | ~~resolves successfully~~                 | ~~resolves with HTTP 502 error~~                 | ~~resolves with HTTP 502 error~~
~~Safari (all versions)~~ | ~~all~~        | ~~does not resolve~~       | ~~n/a~~                                   | ~~n/a~~                                          | ~~n/a~~
~~Edge 13 & 14~~          | ~~Windows 10~~ | ~~does not resolve~~       | ~~n/a~~                                   | ~~n/a~~                                          | ~~n/a~~

In all cases, the WPT server instance was running on a GNU/Linux system. Results for the GNU/Linux platform were collected using locally-running browsers. Results for Non-GNU/Linux platforms were collected using browsers running remotely (as provided by the BrowserStack service). It seems possible that this discrepancy contributed to the variation reported above.

*Note on determinism* These behaviors are only observable when requests 1 and 2 share the same connection. As far as I know, the user agent is free to distribute requests across connections in any manner. While the demonstration seems fairly reliable, I don't think it is technically stable.

### DIFF
--- a/http-demo/index.html
+++ b/http-demo/index.html
@@ -1,0 +1,88 @@
+<!DOCTYPE html>
+<head>
+  <meta charset="utf-8" />
+  <title>Inconsistency in response parsing</title>
+  <script src="/resources/testharness.js"></script>
+  <script src="/resources/testharnessreport.js"></script>
+</head>
+<body>
+<script>
+function script(src) {
+  var s = document.createElement('script');
+  s.setAttribute('src', src + '&time=' + new Date().getTime());
+  document.body.appendChild(s);
+  return new Promise(function(resolve, reject) {
+      s.onload = resolve;
+	  s.onerror = reject;
+    });
+}
+promise_test(function(t) {
+  return Promise.resolve()
+    .then(function() {
+        return script('index.py?seq=1');
+      })
+    .then(function() {
+        return script('index.py?seq=2&leading=0');
+      });
+}, 'zero leading characters in second response');
+
+promise_test(function(t) {
+  return Promise.resolve()
+    .then(function() {
+        return script('index.py?seq=1');
+      })
+    .then(function() {
+        return script('index.py?seq=2&leading=1');
+      });
+}, 'one leading character in second response');
+
+promise_test(function(t) {
+  return Promise.resolve()
+    .then(function() {
+        return script('index.py?seq=1');
+      })
+    .then(function() {
+        return script('index.py?seq=2&leading=2');
+      });
+}, 'two leading characters in second response');
+
+promise_test(function(t) {
+  return Promise.resolve()
+    .then(function() {
+        return script('index.py?seq=1');
+      })
+    .then(function() {
+        return script('index.py?seq=2&leading=3');
+      });
+}, 'three leading characters in second response');
+
+promise_test(function(t) {
+  return Promise.resolve()
+    .then(function() {
+        return script('index.py?seq=1');
+      })
+    .then(function() {
+        return script('index.py?seq=2&leading=4');
+      });
+}, 'four leading characters in second response');
+
+promise_test(function(t) {
+  return Promise.resolve()
+    .then(function() {
+        return script('index.py?seq=1');
+      })
+    .then(function() {
+        return script('index.py?seq=2&leading=5');
+      });
+}, 'five leading characters in second response');
+
+promise_test(function(t) {
+  return Promise.resolve()
+    .then(function() {
+        return script('index.py?seq=1');
+      })
+    .then(function() {
+        return script('index.py?seq=2&leading=6');
+      });
+}, 'six leading characters in second response');
+</script>
--- a/http-demo/index.py
+++ b/http-demo/index.py
@@ -1,0 +1,28 @@
+import time
+
+second = '''HTTP/1.1 200 OK
+Content-Type: application/javascript
+Content-Length: 1
+
+0'''
+
+first = '''HTTP/1.1 204 No Content
+Content-Type: application/javascript
+Content-Length: %s
+
+''' % len(second)
+
+def main(request, response):
+    payload = ''
+    leading_characters = int(request.GET.first('leading', '0'))
+
+    if leading_characters > 0:
+        payload += 'x' * leading_characters
+
+    if request.GET.first('seq') == '1':
+        payload += first
+    else:
+        payload += second
+
+    response.writer.write(payload)
+    response.writer.flush()


w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/5227/)

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/215566137)

# Firefox (nightly channel) #
Testing web-platform-tests at revision f51f0ff357093bc55417afbe6a88dd221ffd10f5
Using browser at version BuildID 20170326110231; SourceStamp f5e214144799889e2408c4841351f4053f00544e
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="http://w3c-test.org/submissions/5227/http-demo/index.html">/http-demo/index.html</a></summary>


|                    Subtest                    | Results | Messages |
|-----------------------------------------------|---------|----------|
|                                               | OK      |          |
| `zero leading characters in second response`  | PASS    |          |
| `one leading character in second response`    | PASS    |          |
| `two leading characters in second response`   | PASS    |          |
| `three leading characters in second response` | PASS    |          |
| `four leading characters in second response`  | PASS    |          |
| `five leading characters in second response`  | PASS    |          |
| `six leading characters in second response`   | PASS    |          |

</details>
</details>

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/215566138)

# Chrome (unstable channel) #
Testing web-platform-tests at revision f51f0ff357093bc55417afbe6a88dd221ffd10f5
Using browser at version 59.0.3047.0 dev
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="http://w3c-test.org/submissions/5227/http-demo/index.html">/http-demo/index.html</a></summary>


|                    Subtest                    | Results |                                 Messages                                |
|-----------------------------------------------|---------|-------------------------------------------------------------------------|
|                                               | OK      |                                                                         |
| `zero leading characters in second response`  | PASS    |                                                                         |
| `one leading character in second response`    | PASS    |                                                                         |
| `two leading characters in second response`   | PASS    |                                                                         |
| `three leading characters in second response` | PASS    |                                                                         |
| `four leading characters in second response`  | PASS    |                                                                         |
| `five leading characters in second response`  | FAIL    | `promise_test: Unhandled rejection with value: object "[object Event]"` |
| `six leading characters in second response`   | FAIL    | `promise_test: Unhandled rejection with value: object "[object Event]"` |

</details>
</details>

jugglinmike:
@foolip I'd like to get some feedback from implementers about the current behavior and what can be done about standardizing it. Do you know anyone on the Chromium team that could weigh in?

jugglinmike:
@RByers Do you know anyone on the Chromium team that could weigh in? (See my previous comment)

foolip:
This is network stuff, so maybe @ricea knows or knows someone who does.

foolip:
Is the issue here that up to 6 bytes of garbage ar skipped, and that nothing in the HTTP spec accounts for this and thus implicitly disallows it?

jugglinmike:
The HTTP spec explicitly disallows the precondition (a 204 response with a
`Content-length` header), so it's probably not accurate to say that that spec
doesn't account for this. The issue that @annevk and @rbyers are interested in
is the inconsistency between browsers. I believe they would like to consider
if/how the Fetch specification could be extended to account for this state, and
the first step there is understanding implementation reality.

ricea:
The Chromium behaviour is not specific to 204 responses. It allows 4 bytes of "slop" at the start of any HTTP/1.[01] response. This is implemented in [net::HttpUtil::LocateStartOfStatusLine](https://cs.chromium.org/chromium/src/net/http/http_util.cc?q=net::HttpUtil::LocateStartOfStatusLine).

I'm surprised that this differs from Firefox, since Chrome's handling of HTTP server quirks was largely inspired by Firefox.

annevk:
@ricea #5102 might be of interest. Lots of differences between how responses are parsed.

ericlaw1979:
Long ago, I conducted a related survey on this topic: https://blogs.msdn.microsoft.com/ieinternals/2011/03/09/content-length-in-the-real-world/.

The Microsoft Edge team noted a real-world site with a compatibility issue related to ignoring junk: The page http://www.ntvmsnbc.com/id/25015435 sent back, "GET HTTP/1.1 301 Moved Permanently”, as the status line.  This is obviously completely invalid, but since Chrome is allowing the 4 bytes of junk it succeeds.

MattMenke:
Chrome's behavior here does not vary by platform, and Chrome does not internally generate bogus 502 response codes.  Given that the chart claims both FireFox and Chrome are displaying a 502 when a server is not, according to the description, returning a 502 error, I suspect there's some sort of proxy present, or a different test server is being used on different platforms.  Anyhow, I'd treat the chart as unreliable.

Note that the new "net::ERR_INVALID_HTTP_RESPONSE" error is because we removed support for HTTP/0.9 on non-standard ports recently.  If the server were using port 80, the final request would succeed on current versions of Chrome.

jugglinmike:
Thanks for the input, @MattMenke. You are correct about my use of a proxy. I
mentioned that the OSX and Windows tests were run using BrowserStack-provided
browsers connecting to a local server; I just didn't think deeply enough about
what that entailed. Communication is indeed mediated by a tunnel in that case,
and that indirection invalidates the tests. In order to avoid further
confusion, I've struck the invalid data points from my original report.

I've conducted a few more tests. This time, I ran them all all tests on a local
network, accessing the server directly (i.e. no tunnels, and no `localhost`
connections). I also added expanded the trial matrix to include standard and
non-standard TCP ports.

The topic of "incorrect Content-Length header" seems to be well-worn ground, so
I'm focusing on the "extra bytes" tests. (Thanks for the reference,
@ericlaw1979 --I've been enjoying reading that post and the associated ones over
the last few days.)

New tests available here: https://github.com/w3c/web-platform-tests/compare/master...bocoup:http-demo-2

Sigil              | Meaning
-------------------|--------------------------------------------------
:white_check_mark: | Request completed successfully
:x:                | Request failed (`net::ERR_INVALID_HTTP_RESPONSE`)
:stopwatch:        | Request did not resolve

Browser     | Initial conditions   | Port | 0                  | 1                  | 2                   | 3                   | 4                  | 5                  | 6
------------|----------------------|------|--------------------|--------------------|---------------------|---------------------|--------------------|--------------------|-------------------
Chrome 59   | Independent response | 80   | :white_check_mark: | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | :white_check_mark: | :stopwatch:        | :stopwatch:
Chrome 59   | Independent response | 8080 | :white_check_mark: | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | :white_check_mark: | :x:                | :x:
Chrome 59   | Following a 204      | 80   | :white_check_mark: | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | :white_check_mark: | :stopwatch:        | :stopwatch:
Chrome 59   | Following a 204      | 8080 | :white_check_mark: | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | :white_check_mark: | :x:                | :x:
Firefox 55  | Independent response | 80   | :white_check_mark: | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | :white_check_mark: | :stopwatch:        | :stopwatch:
Firefox 55  | Independent response | 8080 | :white_check_mark: | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | :white_check_mark: | :stopwatch:        | :stopwatch:
Firefox 55  | Following a 204      | 80   | :white_check_mark: | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | :white_check_mark: | :white_check_mark: | :white_check_mark:
Firefox 55  | Following a 204      | 8080 | :white_check_mark: | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | :white_check_mark: | :white_check_mark: | :white_check_mark:
Edge 40     | Independent response | 80   | :white_check_mark: | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | :white_check_mark: | :stopwatch:        | :stopwatch:
Edge 40     | Independent response | 8080 | :white_check_mark: | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | :white_check_mark: | :stopwatch:        | :stopwatch:
Edge 40     | Following a 204      | 80   | :white_check_mark: | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | :white_check_mark: | :stopwatch:        | :stopwatch:
Edge 40     | Following a 204      | 8080 | :white_check_mark: | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | :white_check_mark: | :stopwatch:        | :stopwatch:
Safari 10.2 | Independent response | 80   | :white_check_mark: | :stopwatch:        | :stopwatch:         | :stopwatch:         | :stopwatch:        | :stopwatch:        | :stopwatch:
Safari 10.2 | Independent response | 8080 | :white_check_mark: | :stopwatch:        | :stopwatch:         | :stopwatch:         | :stopwatch:        | :stopwatch:        | :stopwatch:
Safari 10.2 | Following a 204      | 80   | :white_check_mark: | :stopwatch:        | :stopwatch:         | :stopwatch:         | :stopwatch:        | :stopwatch:        | :stopwatch:
Safari 10.2 | Following a 204      | 8080 | :white_check_mark: | :stopwatch:        | :stopwatch:         | :stopwatch:         | :stopwatch:        | :stopwatch:        | :stopwatch:

(Integer-labeled columns indicate the number of invalid extra bytes written
prior to a valid HTTP 200 response.)

Chrome appears to be the only browser to differentiate between standard and
non-standard ports. @MattMenke you mentioned that the "response prefixed with 5
invalid bytes" would produce a successful request on port 80, but I'm
experiencing a timeout. The timeout is consistent with Edge and supported by
the source that @ricea referenced, but then again, my testing methods may still
be invalidating the experiment. Can you confirm that Chrome is expected to
tolerate more than 4 invalid bytes in this case?

The other interesting detail (and the one that started this whole
investigation) is that Firefox becomes more tolerant of invalid data following
a successful HTTP 204 response. This behavior is specific to 204's--the 4-byte
threshold stands following an HTTP 200 response (for instance). I haven't been
able to pin down the exact threshold in these conditions because results become
unstable somewhere around 132,200 invalid bytes.

MattMenke:
I'm not sure why Chrome isn't treating it as an HTTP/0.9 response, if I'm understanding what's happening.  Is the server closing the socket after the second response, or keeping it open?  If the socket isn't closed, and it's not a fetch request, it may be held up at the mime sniffer, unless it has a mime type that doesn't trigger sniffing.

MattMenke:
Correction - since we'd be seeing it as HTTP/0.9, we wouldn't be parsing the HTTP/1.x headers, so of course we'd be MIME sniffing it.  You'd need to either close the socket or send enough data to get any data from the second request (2 kb?) - and even if you sent enough data, I'm not sure if Chrome will run any scripts on a partially received JS file (It may, or it may not, not my area of knowledge).

jugglinmike:
Ah, I see much more consistent results when the server explicitly closes the socket after writing the data. Eliding results for non-standard ports, and extending with the protocol version used to interpret the response:

Browser     | Initial conditions   | Port | 0   | 1   | 2   | 3   | 4   | 5   | 6
------------|----------------------|------|-----|-----|-----|-----|-----|-----|-----
Chrome 59   | Independent response | 80   | 1.1 | 1.1 | 1.1 | 1.1 | 1.1 | 0.9 | 0.9
Chrome 59   | Following a 204      | 80   | 1.1 | 1.1 | 1.1 | 1.1 | 1.1 | 0.9 | 0.9
Firefox 55  | Independent response | 80   | 1.1 | 1.1 | 1.1 | 1.1 | 1.1 | 0.9 | 0.9
Firefox 55  | Following a 204      | 80   | 1.1 | 1.1 | 1.1 | 1.1 | 1.1 | 1.1 | 1.1
Edge 40     | Independent response | 80   | 1.1 | 1.1 | 1.1 | 1.1 | 1.1 | 0.9 | 0.9
Edge 40     | Following a 204      | 80   | 1.1 | 1.1 | 1.1 | 1.1 | 1.1 | 0.9 | 0.9
Safari 10.2 | Independent response | 80   | 1.1 | :x: | :x: | :x: | :x: | :x: | :x:
Safari 10.2 | Following a 204      | 80   | 1.1 | :x: | :x: | :x: | :x: | :x: | :x:

Note that the surprising Firefox behavior holds even in this case. I'm following up with a Firefox bug report--hopefully, the maintainers there will be able to shed some light on this:

https://bugzilla.mozilla.org/show_bug.cgi?id=1356614

Thanks again for your help, @MattMenke!

jugglinmike:
@annevk That Firefox bug report was closed as "wontfix":

> in general there are not normative requirements for the handling invalid behavior from the peer (in this case the invalid data before the HTTP).. so writing a test case around it seems rather unlikely to lead to a satisfying conclusion..

Is this something we try to somehow account for in Fetch?

annevk:
I reopened that bug since I don't think it should be closed. I think ideally HTTP defines how to parse HTTP. Fetch has some overrides for when HTTP doesn't do its job, but I believe in general the HTTP folks are more receptive now to fully specifying their thing. @mnot?

mnot:
Relevant text in [RFC7230 section 3.3.3](http://httpwg.org/specs/rfc7230.html#message.body.length):

> If the final response to the last request on a connection has been completely received and there remains additional data to read, a user agent MAY discard the remaining data or attempt to determine if that data belongs as part of the prior response body, which might be the case if the prior message's Content-Length value is incorrect. A client MUST NOT process, cache, or forward such extra data as a separate response, since such behavior would be vulnerable to cache poisoning.

Since there is no prior response body in the case of a 204, it's clearly to be discarded.

There's also [section 3.5](http://httpwg.org/specs/rfc7230.html#message.robustness), but that focuses more on the server side.

What I think's being requested is a recommendation for how much data should be discarded before the client gives up; possibly a minimum. It feels kind of analogous to when we established the minimum URL length that should be supported by implementations, so it's not completely off base.

That said, this is truly a corner case; the right answer is "don't do that." Anyone depending on interop in this case is doing it wrong to start with.

If there's interest in doing so any way, file a bug on <https://github.com/httpwg/http11bis/>.

jugglinmike:
What do you say, @annevk? Should I take this up with the IETF?

annevk:
@jugglinmike I'd at least file a bug yes. https://wiki.whatwg.org/wiki/HTTP has some other cases I should maybe get back to at some point.

foolip:
@jugglinmike, is there something to be done with this PR still?

foolip:
Closing as inactive, and clearly not intended to be merged as-is.

PR #5316, opened 2017-04-03
[selectors4] Fix reference to avoid glitches on :focus-within tests

mrego:
Some tests were using 3 lines and other 1 line for the same text.
When comparing 3 (test) vs 1 line (reference) this was causing
very small differences and considering the test as failure in Chrome.

This text is not relevant at all for `:focus-within` tests,
so it's better to use 3 lines in all the files (tests and reference).

This is a very simple change, please @gsnedders or @frivoal 
could you take a look? Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/5316)

### DIFF
--- a/css/selectors4/focus-within-001-ref.html
+++ b/css/selectors4/focus-within-001-ref.html
@@ -9,6 +9,8 @@ div {
   border: solid 15px green;
 }
 </style>
-<p>Test passes if, when the element below is focused, it is surrounded by a thick green border. There must be no red or blue once it is focused.</p>
+<p>Test passes if, when the element below is focused,
+it is surrounded by a thick green border.
+There must be no red or blue once it is focused.</p>
 <div>Focus this element</div>
 </html>
--- a/css/selectors4/focus-within-001.html
+++ b/css/selectors4/focus-within-001.html
@@ -26,7 +26,9 @@ div:focus-within {
 border-color: green;
 }
 </style>
-<p>Test passes if, when the element below is focused, it is surrounded by a thick green border. There must be no red or blue once it is focused.</p>
+<p>Test passes if, when the element below is focused,
+it is surrounded by a thick green border.
+There must be no red or blue once it is focused.</p>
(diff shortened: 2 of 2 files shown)


w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/217947355)

# Firefox (nightly channel) #
Testing web-platform-tests at revision e3161618b432bd097f34207ffaf2aa91fa8a4d3e
Using browser at version BuildID 20170402100159; SourceStamp 7a3f514cf8490d271ee373a1d2999e4ea4dee2d7
Starting 10 test iterations
No tests run.

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/217947356)

# Chrome (unstable channel) #
Testing web-platform-tests at revision e3161618b432bd097f34207ffaf2aa91fa8a4d3e
Using browser at version 59.0.3053.3 dev
Starting 10 test iterations
No tests run.

frivoal:
It is strange that this difference caused failures, but regardless, if it does, it's an annoyance, and the tests are correct after your fix. Approved & merging.

gsnedders:
I suspect this is a problem common with hardware acceleration in general, rather than anything Chrome specific. We really just need some statement about hardware acceleration generally, rather than trying to fix everything. Much like how everything is broken at HiDPI. :\

mrego:
@frivoal thanks for the review and merge.
As this test was not checking anything like that, I think it was fine to just do this simple change.

If you're more interested on the specific issue is reported in Chrome bugtracker:
https://bugs.chromium.org/p/chromium/issues/detail?id=707720
As @gsnedders says, I believe HDPI can be related.

frivoal:
@gsnedders I don't think we should try to fix everything a priori, but when we run into failures that are blocking us from doing something else and have trivial fixes, it's worth taking them in.

PR #5334, opened 2017-04-04
[presentation-api] fix the "Creating a receiving browsing context" test

tomoyukilabs:
While the current implementation of the "Creating a receiving browsing context" test uses `EventWatcher` to handle `upgradeneeded` and `success` events fired by IndexedDB, it was found that Firefox throws an exception like `InvalidStateError: A mutation operation was attempted on a database that did not allow mutations.` when these events are fired.

This PR replaces `EventWatcher` for IndexedDB with `new Promise` so that such an exception can be avoided, although I don't think it might be a better way.

### DIFF
--- a/presentation-api/receiving-ua/PresentationReceiver_create-manual.https.html
+++ b/presentation-api/receiving-ua/PresentationReceiver_create-manual.https.html
@@ -91,16 +91,19 @@
             const openIndexedDB = () => {
                 if ('indexedDB' in window) {
                     const dbReq = indexedDB.open(dbName, 1);
-                    const eventWatcher = new EventWatcher(t, dbReq, ['upgradeneeded', 'success']);
-                    return eventWatcher.wait_for('upgradeneeded').then(() => {
-                        db = dbReq.result;
-                        const store = db.createObjectStore(storeName, { keyPath: 'id' });
-                        store.add(storeData);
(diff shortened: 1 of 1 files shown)


w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/218340726)

# Firefox (nightly channel) #
Testing web-platform-tests at revision 89930d58e349e05f09910fe98a25415a93c499c2
Using browser at version BuildID 20170403100136; SourceStamp 38894655c89e68bcd8f45d31a0d3005f2c2b53db
Starting 10 test iterations
No tests run.

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/218340727)

# Chrome (unstable channel) #
Testing web-platform-tests at revision 89930d58e349e05f09910fe98a25415a93c499c2
Using browser at version 59.0.3053.3 dev
Starting 10 test iterations
No tests run.

tidoust (approved):
Good catch.

I believe that's actually correct in any case: The `EventWatcher` turns the event in a `Promise` which cannot resolve on the exact same tick, so the user agent may indeed complete the initialization of the database before the code that creates the object store runs, leading to the reported error.

PR #5345, opened 2017-04-04
Add tests to css-timing-1.

jgraham:
MozReview-Commit-ID: HszNpNTCM1N

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1248340 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/5345)

### DIFF
--- a/css-timing-1/OWNERS
+++ b/css-timing-1/OWNERS
@@ -1,0 +1,2 @@
+@birtles
+@BorisChiou
--- a/css-timing-1/frames-timing-functions-output.html
+++ b/css-timing-1/frames-timing-functions-output.html
@@ -1,0 +1,81 @@
+<!DOCTYPE html>
+<meta charset=utf-8>
+<meta name="assert"
+content="This test checks the output of frame timing functions with different frame numbers" />
+<title>Frames timing function output tests</title>
+<link rel="help"
+href="https://drafts.csswg.org/css-timing/#frames-timing-functions">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="testcommon.js"></script>
+<style>
+@keyframes anim {
+  from { left: 0px; }
+  to   { left: 100px; }
+}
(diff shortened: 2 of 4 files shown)


jgraham:
Code reviewed upstream.

gsnedders:
@jgraham @foolip this introduces a css-timing-1 directory at the top level; I'm pretty sure we don't want versioned folders at the top level at all.

foolip:
@birtles @BorisChiou, can you rename it?

BorisChiou:
Sure. I can update the folder name to "css-timing" in Gecko.

BorisChiou:
file bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1364004

birtles:
This should remain css-timing-1 but move under the css directory.
I spoke to James about doing this a while back but don't remember what direction was going to be easiest to do it in (m-c or wpt)

gsnedders:
@birtles I don't think there's any need to move anything into the css directory. The only advantage of the css directory is that it remains things needed for the CSS test harness, which we're meant to be moving away from. We should also be following the normal WPT habit and not having versioned folders going forward.

birtles:
@gsnedders Ok, I didn't know that. That makes sense then.
Do we plan to move other folders out of 'css' when possible?

gsnedders:
@birtles Basically the answer is no. Basically:

 * We want to make it possible to use both the old CSS WG test harness and the WPT tools in the short term for progress along the REC-track.
 * Moving files breaks vendors expectation files as to what tests fail/crash/etc.

birtles:
Ok. It feels a bit awkward that when I work on transitions I need to touch files in 'css/css-transitions-1' and 'css-timing', but, oh well.

gsnedders:
Yeah, it's kinda ugly, but it was the agreed compromise to avoid delaying merging the repos for another year or more… :\

PR #5351, opened 2017-04-04
Simplify the setup around the editor state's GetSelectionRange function.

jgraham:
Really, there are only two cases we need to worry about.  Either
IsSelectionCached(), and then our SelectionProperties has the data we want, or
not and then we have a non-null mSelCon which has the data we want.

Since we are now using cached selection state a lot more (instead of
initializing the editor whenever someone asks for selection state), we need to
actually update it more correctly when .value is set.

And since we now update the cached selection state for the case when .value has
been set (to point to the end of the text), we need to change
HTMLInputElement::HasCachedSelection to return false for that case.  Otherwise
we will always do eager editor init on value set.  We handle that by not doing
eager init if the cached selection is collapsed.

The web platform test changes test the "update on .value set" behavior.  They
fail without this patch, pass with it.

MozReview-Commit-ID: DDU8U4MGb23

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1343037 [ci skip]

### DIFF
--- a/html/semantics/forms/textfieldselection/selection.html
+++ b/html/semantics/forms/textfieldselection/selection.html
@@ -14,18 +14,24 @@
   var dirs = ['forward', 'backward', 'none'];
   var sampleText = "0123456789";
 
-  var createInputElement = function(value) {
+  var createInputElement = function(value, append = true) {
     var el = document.createElement("input");
     el.type = "text";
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #5353, opened 2017-04-04
Implement nsTextEditorState::SetSelectionEnd.

jgraham:
This introduces three behavior changes:

1)  Before this change, in cached mode, we did not enforce the "start <= end"
    invariant.
2)  Before this change, in cached mode, we did not fire "select" events on
    selectionEnd changes.
3)  Changes the IDL type of HTMLInputElement's selectionEnd attribute to
    "unsigned long" to match the spec and HTMLTextareaElement.

MozReview-Commit-ID: J3Gkhr8VnbS

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1343037 [ci skip]

### DIFF
--- a/html/semantics/forms/textfieldselection/selection-start-end.html
+++ b/html/semantics/forms/textfieldselection/selection-start-end.html
@@ -55,6 +55,19 @@
 
   test(function() {
     for (let el of createTestElements(testValue)) {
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #5355, opened 2017-04-04
P2 Add WPT verifying MessageEvent.source is set correctly when sent from waiting ServiceWorker.

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1345943 [ci skip]


jgraham:
Code reviewed upstream.

PR #5363, opened 2017-04-04
Ensure that finalizers for webdriver tests always work on a valid window.

jgraham:
In case of tests are closing the current window, and do not switch back to
a valid window, the finalizers will fail because the window to operate on
doesn't exist anymore.

MozReview-Commit-ID: 8tX6oK45530

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1322383 [ci skip]

### DIFF
--- a/webdriver/conftest.py
+++ b/webdriver/conftest.py
@@ -1,5 +1,4 @@
 import pytest
-
 from support.fixtures import (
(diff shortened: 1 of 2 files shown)


jgraham:
Code reviewed upstream.

PR #5364, opened 2017-04-04
Fix crash in respondWith resolved callback. Don't reset interception if the sw throws after .respondWith().

jgraham:
Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1342255 [ci skip]


jgraham:
Code reviewed upstream.

PR #5187, opened 2017-03-21
WebDriver new session tests.

jgraham:
Tests for the new session command, covering the behaviour of
alwaysMatch and firstMatch, along with the supported properties.

PR #5506, opened 2017-04-07
Clean up fixtures for session creation in wdspec.

jgraham:
This introduces two fixtures for creating a multi-test session and a
longlived session. For performance it is important that we don't
ordinarily create a new session for each test, but it's also important
that specific tests are able to create new sessions with arbitary
capabilities.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/5506)

### DIFF
--- a/webdriver/conftest.py
+++ b/webdriver/conftest.py
@@ -1,12 +1,13 @@
 import pytest
 from support.fixtures import (
-    create_frame, create_session, create_window, http, server_config, session,
+    configuration, create_frame, create_window, http, new_session, server_config, session,
     url)
 
+pytest.fixture(scope="session")(configuration)
 pytest.fixture()(create_frame)
-pytest.fixture()(create_session)
 pytest.fixture()(create_window)
 pytest.fixture()(http)
+pytest.fixture(scope="function")(new_session)
 pytest.fixture()(server_config)
 pytest.fixture(scope="function")(session)
 pytest.fixture()(url)
--- a/webdriver/support/fixtures.py
+++ b/webdriver/support/fixtures.py
@@ -31,6 +31,7 @@ def _dismiss_user_prompts(session):
 
     session.window_handle = current_window
 
+
 def _restore_windows(session):
     """Closes superfluous windows opened by the test without ending
     the session implicitly by closing the last window.
@@ -44,12 +45,14 @@ def _restore_windows(session):
 
     session.window_handle = current_window
 
+
 def _switch_to_top_level_browsing_context(session):
     """If the current browsing context selected by WebDriver is a
     `<frame>` or an `<iframe>`, switch it back to the top-level
     browsing context.
     """
     session.switch_frame(None)
+
 
 def _windows(session, exclude=None):
     """Set of window handles, filtered by an `exclude` list if
@@ -59,6 +62,7 @@ def _windows(session, exclude=None):
         exclude = []
     wins = [w for w in session.handles if w not in exclude]
     return set(wins)
+
 
 def create_frame(session):
     """Create an `iframe` element in the current browsing context and insert it
@@ -73,6 +77,7 @@ def create_frame(session):
 
     return create_frame
 
+
 def create_window(session):
     """Open new window and return the window handle."""
     def create_window():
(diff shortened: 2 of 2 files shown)


jgraham:
@jugglinmike With this change the wdspec tests work for me locally (including the in-review new session tests).

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/219718435)

# Firefox (nightly channel) #
Testing web-platform-tests at revision 354df2805f09e29a1c4b063804553c12f157fa29
Using browser at version BuildID 20170406100208; SourceStamp 3c68d659c2b715f811708f043a1e7169d77be2ba
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="http://w3c-test.org/submissions/5506/webdriver/conftest.py">/webdriver/conftest.py</a></summary>


| Subtest | Results | Messages |
|---------|---------|----------|
|         | OK      |          |

</details>
</details>

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/219718436)

# Chrome (unstable channel) #
Testing web-platform-tests at revision 354df2805f09e29a1c4b063804553c12f157fa29
Using browser at version 59.0.3063.4 dev
Starting 10 test iterations
No tests run.

webdriver/support/fixtures.py
```diff
@@ -83,54 +88,88 @@ def create_window():
...
+            _current_session = None
+
+    def create_session(body):
+        global _current_session
+        _session = webdriver.Session(configuration["host"],
+                                     configuration["port"],
+                                     capabilities=None)
+        # TODO: merge in some capabilities from the confguration capabilities
+        # since these might be needed to start the browser
+        value = _session.send_command("POST", "session", body=body)
```
jugglinmike: What would you think about simply constructing the Session object (with
capabilities) and returning it? This would make the tests that need this
fixture a little more direct, and it would save us from having to deal with the
command return value or the object's `session_id` property.

Also, why is the "merging" operation a "TODO".

webdriver/support/fixtures.py
```diff
@@ -83,54 +88,88 @@ def create_window():
...
+        # Don't set the global session until we are sure this succeeded
+        _current_session = _session
+        _session.session_id = value["sessionId"]
+
+        return value, _current_session
+
+    end()
+    request.addfinalizer(end)
```
jugglinmike: I was tempted to suggest placing this inside the `create_session` function, but I think that would make it preclude testing (doomed) attempts to create concurrent sessions. Probably that's why you wrote it as you did, but I just wanted to raise that to make sure we're in agreement.

webdriver/conftest.py
```diff
@@ -1,10 +1,11 @@
 import pytest
 from support.fixtures import (
-    create_frame, create_session, create_window, http, server_config, session,
+    configuration, create_frame, create_window, http, new_session, server_config, session,
     url)
 
+pytest.fixture(scope="session")(configuration)
 pytest.fixture()(create_frame)
-pytest.fixture()(create_session)
+pytest.fixture(scope="function")(new_session)
```
jugglinmike: Mind re-locating this to maintain alphabetic sort order?

jugglinmike (changes requested):
Thanks for taking the lead on this, James. Only one substantive piece of feedback here, curious to hear your thoughts.

w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/5506/)

jugglinmike:
Any word on this, @jgraham? I'm hoping to see it land before moving forward with https://github.com/w3c/web-platform-tests/pull/5275 to avoid giving you new conflicts to resolve.

webdriver/support/fixtures.py
jgraham: Yes.

webdriver/support/fixtures.py
jgraham: The merging is a TODO because it's not needed for geckodriver and I don't have wdspec support for Chrome or Edgedriver in order to test it. I think making it work well might require splitting the capabilites we store into "required" and "optional" or something, where "required" would probably be just the path to the binary and optional would be anything like log level that isn't really needed to start a successful session.

Just returning the session object doesn't seem like a big win to me? We probably still need the rest of the logic here but it would end up in a function in the tests. In any case the tests in https://github.com/w3c/web-platform-tests/pull/5187 depend on this API and I don't think they're too confusing.

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/223498276)

# Lint #

Passed

webdriver/support/fixtures.py
jugglinmike: > I think making it work well might require splitting the capabilites we store
> into "required" and "optional" or something, where "required" would probably
> be just the path to the binary and optional would be anything like log level
> that isn't really needed to start a successful session.

The `merge_dictionaries` utility function gives us this functionality because it throws when the second dictionary re-defines a key that is present in the first.

> Just returning the session object doesn't seem like a big win to me? We
> probably still need the rest of the logic here but it would end up in a
> function in the tests.

The change I have in mind is minimal and would only result in one extra line of code within the tests, `session.start()` (more on this in a moment):

```diff
 def create_session(body):
     global _current_session
     _session = webdriver.Session(configuration["host"],
                                  configuration["port"],
                                  capabilities=None)
     # TODO: merge in some capabilities from the confguration capabilities
     # since these might be needed to start the browser
-    value = _session.send_command("POST", "session", body=body)
     # Don't set the global session until we are sure this succeeded
     _current_session = _session
-    _session.session_id = value["sessionId"]
 
-    return value, _current_session
+    return _current_session
```

> In any case the tests in #5187 depend on this API and I don't think they're
> too confusing.

The response body is really only useful for those tests. The majority of tests that rely on this fixture will have to shunt that value.

Of course, if we *don't* start the session in the fixture, those tests will have to do that themselves (as mentioned above, via `session.start()`. This is a different (and longer) kind of repetition. But it's also in keeping with the approach we agreed on a while back: tests ought to use high-level, exception-raising methods for details that they are not directly validating. For functionality under test, though, tests really ought to include the request-response handling. This would make [your new tests](https://github.com/w3c/web-platform-tests/pull/5187) a little more coherent to the average contributor: instead of asserting that "calling this Pytest fixture function raises a wdclient exception", the tests would say "sending a New Session command with wdclient produces an HTTP response with such-and-such status code and content."

jgraham: I think I disagree with this. I think the new session tests are the ones that are by-far the most common users of this code; although it's not impossible that we will end up with tests for "X where X is the first command in a session, or whatever" it seems reasonably unlikely there will be many cases where that is significant.

So let's focus on the new session command case for now. In that case we need the actual response (always) and the session object (but not often). We also need to be able to start mutliple sessions and implicity close the active session at the end of the test. In this patch that relies on the `_current_session` global being set so that we know which session to end (and whether to start a new long-lived session for the following tests). In your proposal how would that work?

I think my feeling is that this is solving a real problem right now and even if it could improve it's probably good enough to land and change later if needed.

jugglinmike: It's been a while since I made that suggestion, so I'm not sure if I've had a
change of heart, or if I just understand this patch better than I did back
then. In any event, I'm happy to give this a shot and iterate when/if we feel
modifications are necessary.

jugglinmike (approved):

jugglinmike:
Verified locally. So much faster!

PR #5602, opened 2017-04-18
webrtc: remove setRemoteDescription / msid test

fippo:
since track ids are essentially meaningless. If the party calling
setRemoteDescription would have added a track before calling SRD
the assertion from this test would break.

See https://github.com/w3c/webrtc-pc/issues/1128 for some discussion.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/5602)

### DIFF
--- a/webrtc/RTCPeerConnection-setRemoteDescription.html
+++ b/webrtc/RTCPeerConnection-setRemoteDescription.html
@@ -10,7 +10,7 @@
   <script src="/resources/testharnessreport.js"></script>
   <script type="text/javascript">
   // tests that ontrack is called and parses the msid information from the SDP and creates
-  // the streams and tracks with matching identifiers.
+  // the streams with matching identifiers.
   async_test(function(test) {
     const sdp = 'v=0\r\n' +
         'o=- 166855176514521964 2 IN IP4 127.0.0.1\r\n' +
@@ -34,9 +34,6 @@
     var pc = new RTCPeerConnection(null);
 
     pc.ontrack = test.step_func(function(event) {
-      assert_equals(event.track.kind, 'audio', 'ontrack fires with an audio track');
-      assert_equals(event.track.id, 'track1', 'the track name is parsed from the MSID line');
-
       assert_equals(event.streams.length, 1, 'the track belongs to one MediaStream');
       assert_equals(event.streams[0].id, 'stream1', 'the stream name is parsed from the MSID line');
       test.done();
@@ -46,7 +43,7 @@
     .catch(test.step_func(function(e) {
       assert_unreached('Error ' + e.name + ': ' + e.message);
     }));
-  }, 'Triggers ontrack when called with a remote description and the MSID is parsed.');
+  }, 'Triggers ontrack when called with a remote description and the MSID of the stream is is parsed.');
 </script>
 
 </body>


w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/223298137)

# Lint #

Passed

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/223298139)

# Firefox (nightly channel) #
Testing web-platform-tests at revision 5ee726463376166c18eb25c4eb55ef4cced3fd1c
Using browser at version BuildID 20170417100236; SourceStamp c697e756f738ce37abc56f31bfbc48f55625d617
Starting 10 test iterations
No tests run.

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/223298140)

# Chrome (unstable channel) #
Testing web-platform-tests at revision 5ee726463376166c18eb25c4eb55ef4cced3fd1c
Using browser at version 59.0.3067.0 dev
Starting 10 test iterations
No tests run.

foolip:
The existing test does pass on Firefox, is there nothing of value that is being asserted? The spec presumably requires *some* behavior, even if the test is currently wrong?

agouaillard-cosmo:
I cannot see the comments from, get notifications about or send messages to @fippo. It makes reviewing more troublesome than it should. @fippo, could you make the necessary changes so we can proceed with this PR and all. your future contributions?

foolip:
@agouaillard-cosmo, what do you mean? @fippo hasn't written any comments on this PR.

This PR has conflicts though, because I moved the file in question. Sorry?

agouaillard:
I did not get the notification of the PR creation, and of the first message, and I cannot exchange with him (using @fippo). I suppose he is blocking me.

fippo:
@foolip: it passed in Chrome with adapter too. In general, the msid to track mapping is broken since adding transceivers. While the particular case tested would probably continue to work it doesn't add value anymore. Rebased.

w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/5602/)

foolip:
OK, so this is beyond my knowledge of WebRTC. @henbos, can you check the test and confirm that it's not testing anything valuable?

henbos:
> If the party calling setRemoteDescription would have added a track before calling SRD the assertion from this test would break.

But the test does not add a track before calling setRemoteDescription, we know that. It tests a very basic case of one SRD call using msid. It looks like it is testing https://tools.ietf.org/html/draft-ietf-mmusic-msid-16#section-3. Am I missing something? (Not an SDP expert.)

agouaillard:
This is not adapter.js, we are not here to address implementation problems. We are here to test against the specs. The fact that browser implementation has been recently broken does not matter. W3C tests can break and should break when the browsers are not compliant.

Does the code test an existing API or feature against its normative definition? Unless it does not, the test should NOT be removed.

foolip:
@fippo, in https://github.com/w3c/webrtc-pc/issues/1128#issuecomment-294963117 you wrote:
> However, it does not hold if the receiver called addTrack before calling setRemoteDescription which causes a RTPReceiver to be creating with a random track id which is then used in ontrack. I'm going to remove the wpt test which made that assertion.

However, this test doesn't call addTrack. If the test is asserting something about MSID that isn't guaranteed, can you remove just that part?

fippo:
@foolip done.

agouaillard:
LGTM.

Issue #5406, opened 2017-04-06
Google Chrome CI job occassionally does not run any tests


gsnedders:
See:

https://travis-ci.org/w3c/web-platform-tests/builds/218464208

cc/ @jgraham @jugglinmike @bobholt

gsnedders:
also https://travis-ci.org/w3c/web-platform-tests/builds/218535057

gsnedders:
Potentially has overlap with #5408.

jugglinmike:
The problem here is actually specific to the Chrome job, so I'm updating the
title to reflect that.

In both reported cases, ChromeDriver failed to run any tests at all. The full
log (hosted on TraviCI but not included in the automated GitHub.com comment)
includes the following output:

    PROCESS | 12034 | Starting ChromeDriver 2.29.461571 (8a88bbe0775e2a23afda0ceaf2ef7ee74e822cc5) on port 4454
    PROCESS | 12034 | Only local connections are allowed.
    Failed to connect to navigate initial page
    Init failed 1
    u'log' (u'debug', {'message': 'Hanging up on Selenium session'})
    u'runner_teardown' ()
    PROCESS | 12165 | Starting ChromeDriver 2.29.461571 (8a88bbe0775e2a23afda0ceaf2ef7ee74e822cc5) on port 4454
    PROCESS | 12165 | Only local connections are allowed.
    Failed to connect to navigate initial page
    Init failed 2

...repeated 3 more times (for a total of 5 "Init failed" reports), followed by
a JSON-encoded version of the following error message:

    Connecting to Selenium failed:
    Traceback (most recent call last):
      File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/wptrunner/executors/executorselenium.py", line 59, in setup
        desired_capabilities=self.capabilities)
      File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/selenium/webdriver/remote/webdriver.py", line 98, in __init__
        self.start_session(desired_capabilities, browser_profile)
      File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/selenium/webdriver/remote/webdriver.py", line 185, in start_session
        response = self.execute(Command.NEW_SESSION, parameters)
      File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/selenium/webdriver/remote/webdriver.py", line 247, in execute
        response = self.command_executor.execute(driver_command, params)
[...]

PR #5626, opened 2017-04-20
Try setting DBUS_SESSION_BUS_ADDRESS when running tests in Chromedriver

jgraham:
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/5626)

### DIFF
--- a/check_stability.py
+++ b/check_stability.py
@@ -1,7 +1,6 @@
 from __future__ import print_function
 
 import argparse
-from ConfigParser import SafeConfigParser
 import logging
 import os
 import re
@@ -10,10 +9,11 @@ import subprocess
 import sys
 import tarfile
 import zipfile
+from ConfigParser import RawConfigParser, SafeConfigParser
 from abc import ABCMeta, abstractmethod
 from cStringIO import StringIO as CStringIO
 from collections import defaultdict, OrderedDict
-from ConfigParser import RawConfigParser
+from distutils.spawn import find_executable
 from io import BytesIO, StringIO
 
 import requests
@@ -172,6 +172,12 @@ class Browser(object):
     def wptrunner_args(self):
         return NotImplemented
 
+    def prepare_environment(self):
+        """Do any additional setup of the environment required to start the
+           browser successfully
+        """
+        pass
+
 
 class Firefox(Browser):
     """Firefox-specific interface.
@@ -282,6 +288,17 @@ class Chrome(Browser):
             "webdriver_binary": "%s/chromedriver" % root,
             "test_types": ["testharness", "reftest"]
         }
+
+    def prepare_environment(self):
+        # https://bugs.chromium.org/p/chromium/issues/detail?id=713947
+        logger.debug("DBUS_SESSION_BUS_ADDRESS %s" % os.environ.get("DBUS_SESSION_BUS_ADDRESS"))
+        if "DBUS_SESSION_BUS_ADDRESS" not in os.environ:
+            if find_executable("dbus-launch"):
+                logger.debug("Attempting to start dbus")
+                logger.debug(subprocess.check_output(["dbus-launch"]))
+            else:
+                logger.critical("dbus not running and can't be started")
+                sys.exit(1)
 
 
 def get(url):
@@ -842,6 +859,8 @@ def main():
                                 files_changed,
                                 args.iterations,
                                 browser)
+
+        browser.prepare_environment()
 
     with TravisFold("running_tests"):
         logger.info("Starting %i test iterations" % args.iterations)
--- a/check_stability.py
+++ b/check_stability.py
@@ -902,6 +902,8 @@ if __name__ == "__main__":
     try:
         retcode = main()
     except:
-        raise
+        import traceback
+        traceback.print_exc()
+        sys.exit(1)
     else:
         sys.exit(retcode)


w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/5626/)

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/224045392)

# Lint #

Passed

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/224045394)

# Firefox (nightly channel) #
Testing web-platform-tests at revision c14ad0df81444aa601b5438dcd308962090a305d
Using browser at version BuildID 20170420100256; SourceStamp 27311156637f9b5d4504373967e01c4241902ae7
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="http://w3c-test.org/submissions/5626/test/chrome_unstable.html">/test/chrome_unstable.html</a></summary>


|    Subtest     | Results | Messages |
|----------------|---------|----------|
|                | OK      |          |
| `Example test` | PASS    |          |

</details>
</details>

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/224045400)

# Chrome (unstable channel) #
Testing web-platform-tests at revision c14ad0df81444aa601b5438dcd308962090a305d

jugglinmike:
(This workaround is currently being discussed in gh-5406.)

jugglinmike:
After (finally) reproducing the issue locally and verifying the efficacy of the workaround, and after reading the sordid history of the problem from ChromeDriver to Chromium to glib, I'm starting to come around to this.

From a purely technical perspective: it seems a little imprecise to incorporate the workaround into the "check stability" utility directly. We now know that the problem is limited to non-interactive environments. Did you consider implementing this as part of the TravisCI test script?

check_stability.py
```diff
@@ -172,6 +172,9 @@ def version(self):
     def wptrunner_args(self):
         return NotImplemented
 
+    def extra_setup(self):
+        pass
```
jugglinmike: The term "setup" is a bit ambiguous in this context (i.e. more installation
steps?) Could we rename the `extra_setup` method to `prepare_environment`? And
defer its invocation to just prior to `run_tests`?

Also, could you add a doc string?

check_stability.py
```diff
@@ -283,6 +286,17 @@ def wptrunner_args(self, root):
             "test_types": ["testharness", "reftest"]
         }
 
+    def extra_setup(self):
+        # https://bugs.chromium.org/p/chromium/issues/detail?id=713947
+        logger.debug("DBUS_SESSION_BUS_ADDRESS %s" % os.environ.get("DBUS_SESSION_BUS_ADDRESS"))
+        if "DBUS_SESSION_BUS_ADDRESS" not in os.environ:
```
jugglinmike: This is only necessary in Linux; could we bail in other environments?

check_stability.py
```diff
@@ -283,6 +286,17 @@ def wptrunner_args(self, root):
...
+            if find_executable("dbus-launch"):
+                logger.debug("Attempting to start dbus")
+                logger.debug(subprocess.check_output(["dbus-launch"]))
+            else:
+                logger.debug("Setting DBUS_SESSION_BUS_ADDRESS to a placeholder value")
+                os.environ["DBUS_SESSION_BUS_ADDRESS"] = "/dev/null"
```
jugglinmike: I don't currently understand the implications of using this value. I am going to take some time to learn what this actually means.

jugglinmike (changes requested):
On IRC, we discussed implementing this workaround in wptrunner. I'm not sure if that's still on your radar, but even if this is just an "interim" fix, I'd like to see a little more formal structure. Nothing major, though!

check_stability.py
jgraham: Is that necessary given that there are no other environements at the moment and everything will certainly break in !linux, given that we are downloading linux-specific binaries?

check_stability.py
jgraham: Since we believe that this works and `dbus-launch` also seems to work in the travis environment can we land this whilst you do that research? As I say at some point I will move this into wptrunner and we can be more careful then.

jugglinmike: My research is complete. The short version: let's fail the build instead of setting the variable to `/dev/null`.

The manual page for `dbus-launch` reads:

> If `DBUS_SESSION_BUS_ADDRESS` is not set for a process that tries to use
> D-Bus, by default the process will attempt to invoke `dbus-launch` with the
> `--autolaunch` option to start up a new session bus or find the existing bus
> address on the X display or in a file in `~/.dbus/session-bus/`

Which means that setting the variable to *any* value will have the intended result: stopping glib from attempting to start a session. However, setting it to `/dev/null` will effect the way Chrome interacts with the system. This is unlikely to be a problem today, but I'm concerned about future scenarios where tests are being written for the Web Bluetooth API (for example). I'd hate to trip anyone up just because they assumed that this low-level integration was valid.

The threat may seem kind of protracted, but I think it's fair to be conservative (call it paranoid if you like) here. That's because in reality, we have no use-case for this branch. [The latest job log](https://travis-ci.org/w3c/web-platform-tests/jobs/224408572#L416) demonstrates that TravisCI (i.e. the only environment that requires this workaround) offers the `dbus-launch` executable:

>     DEBUG:check_stability:DBUS_SESSION_BUS_ADDRESS None
>     DEBUG:check_stability:Attempting to start dbus
>     DEBUG:check_stability:DBUS_SESSION_BUS_ADDRESS=unix:abstract=/tmp/dbus-R7UYTMZ60G,guid=5b04beaf64b61ac2fe15c33858fa3141
>     DBUS_SESSION_BUS_PID=5839
>     DBUS_SESSION_BUS_WINDOWID=2097153

So the branch in question effectively states: "Do something potentially invalid in a case that we don't currently need to support." We *could* just remove that branch, but I think it would be more responsible to report a meaningful error for this case, just in case this script winds up getting ported to some new environment (FreeBSD on Docker...?). Something like, "Due to a bug in glib, a valid DBUS session address must be defined before starting Chrome."

In effect, the current branch silently implements a workaround for that future case, but it's a workaround that might have deeper ramifications. Reporting an error would give the future caller the knowledge they need to implement a version of this workaround that is suited to their environment (even including `DBUS_SESSION_BUS_ADDRESS=/dev/null`).

check_stability.py
jugglinmike: Hm, I guess I forgot that context. Adding the condition now would make future efforts towards portability easier, but that would be a big job anyway (and one I don't know if we're even considering). I'm fine with leaving it as-is

jugglinmike (approved):
This looks good to me. I don't know how you want to merge this (or if you'd like me to do it), but note that the commit message should be updated as well ("set the relevant environment variable to a bogus value.")

PR #5627, opened 2017-04-20
Introduce automated tests for testharness.js

jugglinmike:
These tests are derived from previously-authored "API sample" files.

@jgraham If GeckoDriver is in your PATH, you can run the tests with the following command:

    $ pytest resources/test/tests

I'm wondering what you think about the patch itself, and how the tests ought to be enabled in CI.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/5627)


w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/5627/)

jgraham:
This is very exciting :)

jugglinmike:
:D

jugglinmike:
@jgraham I'm starting to wonder if these tests are too brittle. The
expectations encode the "enum" values for test and sub-test state. Would it be
better to express those more generically, then sort things out dynamically when
interpreting results?

For instance:

```diff
  "status": {
-   "status": 0,
-   "OK": 0,
-   "TIMEOUT": 2,
-   "ERROR": 1,
+   "status": "(this.OK)",
    "message": null,
    "stack": null
  },
```

jugglinmike:
Any thoughts on this, @jgraham?

jgraham:
I'm not a huge fan of adding all the enums in the output, so I think your alternate suggestion is a good idea.

---

Reviewed 4 of 47 files at r1.
Review status: 4 of 43 files reviewed at latest revision, 6 unresolved discussions, all commit checks successful.

---

*[resources/test/conftest.py, line 11 at r1](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627#-KjP4FyjXTwURTDs7tOy:-KjP4FyjXTwURTDs7tOz:b-4a1u1d) ([raw file](https://github.com/w3c/web-platform-tests/blob/a50441917699579484c6c6439de8454f368d3d00/resources/test/conftest.py#L11)):*
> ```Python
> from wptserver import WPTServer
> 
> _ENC = 'utf8'
> ```

All the leading underscores here seem pretty ugly.

---

*[resources/test/conftest.py, line 63 at r1](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627#-KjP6p0z_mrSqfyIjixP:-KjP6p0z_mrSqfyIjixQ:b1fqymw) ([raw file](https://github.com/w3c/web-platform-tests/blob/a50441917699579484c6c6439de8454f368d3d00/resources/test/conftest.py#L63)):*
> ```Python
>             raise Exception('Expected value not declared')
> 
>         expected = json.loads(unicode(self.expected))
> ```

Does it make sense to parse this earlier so that we fail sooner if the expected is not valid JSON?

---

*[resources/test/conftest.py, line 69 at r1](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627#-KjbBrlE6-8lqUbl-RtS:-KjbBrlFT0mbLwsCxVSk:bboobpg) ([raw file](https://github.com/w3c/web-platform-tests/blob/a50441917699579484c6c6439de8454f368d3d00/resources/test/conftest.py#L69)):*
> ```Python
>         # Test object ordering is not guaranteed. This weak assertion verifies
>         # that the indices are unique and sequential
>         indices = map(lambda test_obj: test_obj.get('index'), actual["tests"])
> ```

A list comprehension would be more idiomatic here.

---

*[resources/test/conftest.py, line 74 at r1](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627#-KjbKsVWldnFV8DlOZ4R:-KjbKsVWldnFV8DlOZ4S:byd7rhk) ([raw file](https://github.com/w3c/web-platform-tests/blob/a50441917699579484c6c6439de8454f368d3d00/resources/test/conftest.py#L74)):*
> ```Python
>         # Stack traces are implementation-defined
>         actual['status'] = self._scrub_stack(actual['status'])
>         actual['tests'] = map(self._scrub_stack, actual['tests'])
> ```

I think forcing people to put "(non deterministic)" in the expected data is silly; we can just assert that those things exist and them delete from the actual data before comapring with expected, or something.

---

*[resources/test/conftest.py, line 82 at r1](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627#-KjbQdwzLVE6l0PeIbWI:-KjbQdwzLVE6l0PeIbWJ:b-or9y2c) ([raw file](https://github.com/w3c/web-platform-tests/blob/a50441917699579484c6c6439de8454f368d3d00/resources/test/conftest.py#L82)):*
> ```Python
>     @staticmethod
>     def _assert_sequence(nums):
>         assert nums == range(1, nums[-1] + 1)
> ```

I'm not convinced that checking this is adding much value.

---

*[resources/test/conftest.py, line 96 at r1](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627#-KjbQjG2iz8ojpmKwFSV:-KjbQjG2iz8ojpmKwFSW:b-3d10pb) ([raw file](https://github.com/w3c/web-platform-tests/blob/a50441917699579484c6c6439de8454f368d3d00/resources/test/conftest.py#L96)):*
> ```Python
>     @staticmethod
>     def _scrub_stack(test_obj):
>         copy = dict(test_obj)
> ```

Is there a specific reason for all the copying? It does seem more pure, but I don't really know what it's buying us.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627#-:-KjOy5sFAQvXabI0WDYE:b7q901w)*

jugglinmike:
@jgraham thanks for the review! I'm working to implement all of your feedback.
I'd like to make one additional alteration as it relates to this comment:

> I think forcing people to put "(non deterministic)" in the expected data is
> silly; we can just assert that those things exist and them delete from the
> actual data before comapring with expected, or something.

I was trying to keep the expected values as literal as possible. I'm concerned
that anyone who looks at the tests' source for help might be confused by
disparities between what's declared in the file and what comes out of the test.
That disconnect is unavoidable with details like this, but I thought a value
like `(non-deterministic)` would be less likely to confuse people in that
situation.

Because there is also tension with repeatedly declaring the enum fields, I
think that I need to drop my hopes for literal expected values. I've thought of
an slight alteration that probably seems inconsequential but would go a long
way for me--renaming the top-level properties in the expected data. If tests
declared: `{ "summarized_status": {}, "summarized_tests": [] }`, then I think
we could avoid confusion and keep things concise.

Does that work for you?

jgraham:
I guess I think that it's more important to be succinct so that people aren't discouraged from writing tests due to overhead than it is to be precise in the correlation between the internal format and what we test, or explicit in the code you have to write about the fact that there's a difference. I think a better solution would be to add some docs about how to write these tests.

That said, I don't feel too strongly about this.

---

Review status: 4 of 43 files reviewed at latest revision, 6 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627#-:-Kjm1wLENud5oqvO8T7-:b-jcixmq)*

jugglinmike:
Hi @jgraham,

I've pushed up a couple of commits to address your feedback. There are a few
more things to be done here, but I could use your input:

- Running the expected timeout tests for 10 seconds is kind of annoying. It
  also won't scale well. Do you know of a good way to force the harness to time
  out sooner?
- Should the tests be hooked up to TravisCI? How so? A dedicated job seems like
  overkill. What do you think about adding it to [`ci_unittest.sh`](https://github.com/w3c/web-platform-tests/blob/421fc0fb3be38426d91a4aee29c8d1f693e87d8a/ci_unittest.sh)?

> I'm not convinced that checking this is adding much value.

It may be unlikely to fail, but as long as there is redundancy in the data
structure, I think we should assert consistency. Is there any benefit to
removing it?

> Is there a specific reason for all the copying? It does seem more pure, but I
> don't really know what it's buying us.

Yup, I just wanted to avoid surprising call sites by mutating values out from
under them. With the number of objects we expect to handle here, I don't see
memory ever being an issue. Is there any other reason to avoid copying?

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/230962937)

# Firefox (nightly) #
Testing web-platform-tests at revision 77cf4f2b4a828d8ebf8d78eb9705439c5cc48317
Ignoring 39 changed files:
No files changed

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/230962943)

# Chrome (unstable) #
Testing web-platform-tests at revision 77cf4f2b4a828d8ebf8d78eb9705439c5cc48317
Ignoring 39 changed files:
No files changed

jgraham:
Reviewed 1 of 21 files at r3.
Review status: 4 of 43 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[resources/test/conftest.py, line 104 at r3](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627#-KjsVOP4CCW7PYiEwToA:-KjsVOP5QSktk94n-ifV:b-8o8slr) ([raw file](https://github.com/w3c/web-platform-tests/blob/a1e50a865e30dd95dfe1f9a557fb273986293bcb/resources/test/conftest.py#L104)):*
> ```Python
> 
>         for key, value in full.iteritems():
>             if key != key.upper() or not isinstance(value, int):
> ```

I think this deserves a comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627)*

jgraham:
> Running the expected timeout tests for 10 seconds is kind of annoying. It also won't scale well. Do you know of a good way to force the harness to timeout sooner?

You can set a timeout multiplier, but beware of making it too short.

>Should the tests be hooked up to TravisCI? How so? A dedicated job seems like overkill. What do you think about adding it to ci_unittest.sh?

Sure.

>>I'm not convinced that checking this is adding much value.
>It may be unlikely to fail, but as long as there is redundancy in the data structure, I think we should assert consistency. Is there any benefit to removing it?

It makes the code simpler. I think it could also be removed from testharness.js. Anyway I don't mind either way.

>> Is there a specific reason for all the copying? It does seem more pure, but I don't really know what it's buying us.
>Yup, I just wanted to avoid surprising call sites by mutating values out from under them. With the number of objects we expect to handle here, I don't see memory ever being an issue. Is there any other reason to avoid copying?

Just that it's unidiomatic and so it seems like there's probably a special reason for it (but isn't).

jgraham:
Reviewed 3 of 5 files at r4.
Review status: 6 of 47 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627)*

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/234976043)

# Sauce (safari) #
Testing web-platform-tests at revision 24fb9bbc70a96a6ffb130dda61291967eff96ae5
Ignoring 41 changed files:
Using browser at version 10.0
Starting 10 test iterations
No tests run.

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/234976050)

# Sauce (MicrosoftEdge) #
Testing web-platform-tests at revision 24fb9bbc70a96a6ffb130dda61291967eff96ae5
Ignoring 41 changed files:
Using browser at version 14.14393
Starting 10 test iterations
No tests run.

jugglinmike:
Alrighty @jgraham, I think this is ready for another look. You'll notice that
I'm manually installing Firefox instead of specifying TravisCI's "Firefox
addon" for the jobs where it is required. This is because these tests need the
hostfile entires specified in the top-level addons definition, and specifying
"addons" for a specific job appears to negate that.

That said, I'd be open to suggestion about where to specify the Firefox and
geckodriver downloads. Do you think they're okay in `ci_unittest.sh`, or should
we do that work elsewhere?

jgraham:
Reviewed 2 of 3 files at r5.
Review status: 7 of 46 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[ci_unittest.sh, line 5 at r5](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627#-KkopVM134YE4eA2o0Tj:-KkopVM2QYPIMeCkEPJH:buyv9ps) ([raw file](https://github.com/w3c/web-platform-tests/blob/62dbd44914ceee3f76596c576a66d1c975a91cb5/ci_unittest.sh#L5)):*
> ```Shell
> 
> ROOT=$PWD
> geckodriver_url=https://github.com/mozilla/geckodriver/releases/download/v0.16.1/geckodriver-v0.16.1-linux64.tar.gz
> ```

Can we lean on the code in `tools/browserutils` rather than hardcoding that here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627)*

jgraham:
Reviewed 3 of 4 files at r7.
Review status: 9 of 49 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[tools/browserutils/install.py, line 4 at r7](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627#-KkvOiQ1270Lr4kH8ODN:-KkvOiQ1270Lr4kH8ODO:bbu8lxe) ([raw file](https://github.com/w3c/web-platform-tests/blob/6f14ae05e11a209d40bc848de9a858190253a313/tools/browserutils/install.py#L4)):*
> ```Python
> import sys
> 
> usage = 'Usage: %s [browser name] [component] [destination]\n' % sys.argv[0]
> ```

I know this is just a tiny script for now, but let's use `argparse`.

---

*[tools/browserutils/install.py, line 19 at r7](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627#-KkvOpJdnM4yVnChSlWL:-KkvOpJdnM4yVnChSlWM:b174fnt) ([raw file](https://github.com/w3c/web-platform-tests/blob/6f14ae05e11a209d40bc848de9a858190253a313/tools/browserutils/install.py#L19)):*
> ```Python
>     fail('Browser name and component required.\n')
> 
> Subclass = getattr(browser, sys.argv[1].title(), None)
> ```

Let's put this all in a `main` function and use use `if __name__ == "__main__": main()` pattern

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627)*

jgraham:
Reviewed 2 of 5 files at r8.
Review status: 10 of 49 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627)*

jgraham:
Reviewed 23 of 47 files at r1, 17 of 21 files at r3, 3 of 5 files at r8.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/web-platform-tests/5627)*

jgraham (approved):

jugglinmike:
> I know this is just a tiny script for now, but let's use `argparse`.

Good call. That package is so nice; I always forget it's part of the standard
library.

> Let's put this all in a `main` function and use use `if __name__ ==
> "__main__": main()` pattern

Sure thing.

Also: since this recent change switches us from a stable release to Nightly, I
had to modify the test expectations a bit. It looks like Nightly now honors
the stage 3 `Function.prototype.toString` proposal; I've modified the source
text and expectations so that the tests pass regardless. It also seems that
some Promises were previously being rejected with errors that lacked stack
traces. I don't think it's fruitful to try to account for this, so I've simply
updated the expectations according to the more correct behavior exhibited in
Nightly.

PR #5648, opened 2017-04-23
Convert a DOM Parsing TODO file into a test

foolip:
Added in https://github.com/w3c/web-platform-tests/pull/493

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/5648)


w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/5648/)

foolip:
@jgraham @Ms2ger, care to review? This will fail in Gecko and Edge, but pass in Chrome and Safari. I'll file bugs if the test is good.

foolip:
Firefox and Edge both fail for the same reason, that the internal subset is part of the serialized output. Presumably it's stored in some internal slot.

foolip:
Yep, there's an extra internal field in Gecko:
https://github.com/mozilla/gecko-dev/blob/866da9f9c63008c5f2446ce7c06fcbb0da10133d/dom/base/DocumentType.h#L85

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/224927831)

# Lint #

Passed

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/224927833)

# Firefox (nightly channel) #
Testing web-platform-tests at revision 26e5229c25c1f12a6b8158adbdb2598f68da5810
Using browser at version BuildID 20170422100218; SourceStamp 7d85b081bfabe91dbf18a421f1eee2a32f335823
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="http://w3c-test.org/submissions/5648/domparsing/DOMParser-parseFromString-xml-internal-subset.html">/domparsing/DOMParser-parseFromString-xml-internal-subset.html</a></summary>


|                        Subtest                         | Results |                                             Messages                                             |
|--------------------------------------------------------|---------|--------------------------------------------------------------------------------------------------|
|                                                        | OK      |                                                                                                  |
| `Parsing and serialization of doctype internal subset` | FAIL    | `assert_equals: expected "<!DOCTYPE foo>" but got "<!DOCTYPE foo [ <!ELEMENT foo (#PCDATA)> ]>"` |

</details>
</details>

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/224927842)

# Chrome (unstable channel) #
Testing web-platform-tests at revision 26e5229c25c1f12a6b8158adbdb2598f68da5810

foolip:
Hmm, maybe @ayg for review, then?

domparsing/DOMParser-parseFromString-xml-internal-subset.html
```diff
@@ -0,0 +1,24 @@
...
+  // https://www.w3.org/TR/xml11/#sec-prolog-dtd has intSubset as part of the
+  // syntax, which is not represented in the DocumentType interface. Check that
+  // publicId and systemId are not affected.
+  var doc = new DOMParser().parseFromString('<!DOCTYPE foo [ <!ELEMENT foo (#PCDATA)> ]><foo/>', 'text/xml');
+  var doctype = doc.doctype;
+  assert_equals(doctype.name, 'foo', 'doctype name');
+  assert_equals(doctype.publicId, '', 'doctype publicId');
+  assert_equals(doctype.systemId, '', 'doctype systemId');
+  // Sanity check: the root element is still parsed correctly.
+  var documentElement = doc.documentElement;
```
ayg: Why a local variable that's only used once?  You can keep it if you prefer it stylistically (so the next line doesn't get too long, etc.).  I'm just commenting in case it's a mistake.

domparsing/DOMParser-parseFromString-xml-internal-subset.html
```diff
@@ -0,0 +1,24 @@
...
+  assert_equals(documentElement.localName, 'foo', "root element's localName");
+  // https://w3c.github.io/DOM-Parsing/#xml-serializing-a-documenttype-node also
+  // does not have any notion of the internal subset, so also check that it
+  // isn't serialized by XMLSerializer.
```
ayg: This does match the spec, but do we really want the spec to say this?  What's the rationale for not round-tripping here?  I would think the spec should be changed to match Firefox and Edge.  If for some reason we really want browsers to throw away the internal subset when reserializing and change the semantics of the XML document, possibly converting it from valid to invalid XML, then LGTM.  But I'd like to hear why.  :)

ayg (commented):
This does match the spec, but do we really want the spec to say this?  What's the rationale for not round-tripping here?  I would think the spec should be changed to match Firefox and Edge.  It's true the internal subset isn't in the DOM, but any XML processor will have to store it somehow anyway in order to parse the rest of the document, at least for the duration of parsing.  (Maybe longer?  I don't know enough about XML to say.)

The spec as written changes the semantics of doctypes with internal subsets, and in fact can convert a valid XML document into an invalid one on round-tripping, IIUC.  So this seems like a spec bug, especially because two out of four UAs do actually reserialize the internal subset.  Did the author of the spec deliberately mean to do this, or did he overlook internal subsets when writing the algorithm?

If for some reason we really want the spec to say this, then the spec needs a note making it explicit.  If the note is added to the spec, the test LGTM.  Until then, I don't know if the test matches the intent of the spec, and don't think the test should be accepted until that's clarified.

domparsing/DOMParser-parseFromString-xml-internal-subset.html
foolip: I think I had two asserts before, this should be merged into a single line.

foolip:
@annevk @travisleithead, any opinions on what the specs *should* say here?

annevk:
> The spec as written changes the semantics of doctypes with internal subsets, and in fact can convert a valid XML document into an invalid one on round-tripping, IIUC.

How?

It was intentional to omit the internal subset since the parser normalizes any entity references away (we don't preserve those either). Therefore you don't have to store the entire internal subset for the duration of the parser either, just the bits that might end up being used later on.

Where should the note be added?

foolip:
I'm not really familiar with how internal subset is supposed to work, what would be an input to the parser that when later serialized drops the bit in the doctype and changes something elsewhere in the document? That'd be a better test to add than what I have here.

annevk:
```data:text/xml,<!DOCTYPE doc [ <!ENTITY x "y"> ] ><doc>&x;</doc>```
should come out as
```<doc>y</doc>```

foolip:
Thanks for the suggestion, @annevk. I'm now happy with this test, WDYT? (It still fails on Edge and Firefox, passes on Chrome and Safari)

annevk (approved):

annevk:
Are you going to file bugs?

foolip:
That slipped my mind, thanks for the reminder.

https://bugzilla.mozilla.org/show_bug.cgi?id=1363658
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11960323/

ayg:
Aha, I didn't realize that we expand the entities.  I don't know why that's the desired behavior, but now I see this won't invalidate the document, so I don't care too much.

annevk:
It's only desired because that's how XML parsers (including non-validating) have always worked.

PR #5693, opened 2017-04-26
Update RTCPeerConnection IDL test to latest editor draft

soareschen:
This PR updates the RTCPeerConnection IDL test to the latest editor draft. The test file is also moved to root webrtc/ directory following new naming convention in accordance to #5657.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/5693)

PR #5666, opened 2017-04-24
Start centralizing IDLs to reduce duplication and errors

ayg:
Fixes #5339 (or rather, starts to).

We had two copies of the DOM IDL, one in dom/ and one in html/, which
were out of sync.  Additionally, tons of tests copy bits and pieces of
IDLs from specs like DOM, or define stubs, instead of properly including
the whole IDL.  These should be fixed to include the whole IDL, because
some tests may need to know about the entire base interface (e.g.,
testing that objects inherit properly).

I chose to make the IDLs centralized because we might want to process
them centrally in the future.  For instance, we could have a script to
update all of them, or a test that checks for non-standard properties on
the global object and imports all the IDLs to get a list of standard
properties.  These seem easier to do if the IDLs are all in one place.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/5666)

PR #5712, opened 2017-04-27
Add centralized IDL test for WebRTC

soareschen:
This is a follow up on #5693. The code follows the convention of #5666 to have all WebRTC IDL tests in one place.

- The IDL definition is extracted from https://w3c.github.io/webrtc-pc/ and put at `/interfaces/webrtc.idl`.
- All IDL tests for WebRTC is done at `/webrtc/interfaces.html`.
- Currently only IDL tests for section 4 is implemented.

There are a few issues I encountered when writing the tests, which I will report in separate issue:

- The IDL definition for `partial dictionary RTCOfferOptions` is not recognized by idlharness. I had to manually merge the definition to make the parsing successful.

- Some IDL such as `RTCPeerConnectionIceErrorEvent` are yet to be implemented by browsers. In these cases the error occur during construction.

- Unlike the DOM IDL test, some IDL instance such as `RTCCertificate` can only be constructed asynchronously.

- `promise_test()` does not return a promise. This makes it hard to express things like "Do X after the promise test is done".

Right now I do all IDL tests in the same `interfaces.html`. To deal with undefined IDLs and async initialization, I added some non trivial code which adds complexity. An alternative approach is to use the old way of writing the tests in separate HTML files, so that failure of one test does not affect others. Suggestions are welcome on better ways to write the test.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/5712)


w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/226268533)

# Lint #

Passed

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/226268535)

# Firefox (nightly channel) #

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/226268541)

# Chrome (unstable channel) #

foolip:
> Currently only IDL tests for section 4 is implemented.

Do you mean https://w3c.github.io/webrtc-pc/#peer-to-peer-connections? AFAICT all the IDL from the spec has been copied.

> The IDL definition for `partial dictionary RTCOfferOptions` is not recognized by idlharness. I had to manually merge the definition to make the parsing successful.

Can you file an issue on idlharness.js? Any manual changes that need to be made will get in the way of eventually auto-updating these IDL files with a script.

> Unlike the DOM IDL test, some IDL instance such as `RTCCertificate` can only be constructed asynchronously.

That should be fine. The fetches are async, and so doing other async things after that but before the tests run should be possible. The basic structure of https://github.com/w3c/web-platform-tests/blob/master/dom/interfaces.html makes sense, just do more things inside the `promise_test` before the `doTest` bit.

> `promise_test()` does not return a promise. This makes it hard to express things like "Do X after the promise test is done".

Do you need to do some kind of cleanup, or what is the X?

webrtc/interfaces.html
```diff
@@ -0,0 +1,118 @@
+<!doctype html>
+<html>
```
foolip: Much of the introductory markup here can be omitted if you like, everything between `<!doctype html>` and `<script src=/resources/testharness.js></script>` in fact. People have different preferences, I won't nitpick it further, but at the very least not much is *required*. (I like very little boilerplate.)

webrtc/interfaces.html
```diff
@@ -0,0 +1,118 @@
...
+  <head>
+    <meta charset=utf-8>
+    <title>IDL check of RTCPeerConnection</title>
```
foolip: It tests all of https://w3c.github.io/webrtc-pc. Maybe just "WebRTC IDL tests" following https://github.com/w3c/web-platform-tests/blob/master/dom/interfaces.html

interfaces/webrtc.idl
```diff
@@ -0,0 +1,657 @@
+dictionary RTCConfiguration {
```
foolip: There will need to be a webrtc-stats.idl for https://w3c.github.io/webrtc-stats/ eventually, so can you name this file webrtc-pc.idl?

webrtc/interfaces.html
```diff
@@ -0,0 +1,118 @@
...
+  <body>
+    <h1 class="instructions">Description</h1>
+    <p class="instructions">This test verifies the conformance of WebIDLs defined in WebRTC</p>
+    <div id="log"></div>
+    <script src=/resources/testharness.js></script>
+    <script src=/resources/testharnessreport.js></script>
+    <script src=/resources/WebIDLParser.js></script>
+    <script src=/resources/idlharness.js></script>
+    <script>
+     (() => {
```
foolip: No need to avoid pollution the global object in tests.

webrtc/interfaces.html
```diff
@@ -0,0 +1,118 @@
...
+       'use strict'
+
+       function initRTCPeerConnection(idlArray) {
```
foolip: This is quite a bit more verbose than necessary for most of these objects that are trivial to create. Like https://github.com/ayg/web-platform-tests/blob/master/XMLHttpRequest/interfaces.html, you could do something like this for most of it:
```JavaScript
  idlArray.add_objects({
    RTCPeerConnection: ['new RTCPeerConnection'],
    RTCIceCandidate: ['new RTCIceCandidate'],
    ...
  });
```

Try to minimize the amount of code. You probably noticed that `new RTCIceCandidate({})` throws an exception in both Chrome and Firefox because the `candidate` member is missing. Be sure to not work around such issues, but instead file issues on the spec or implementations, depending on which you think is wrong.

webrtc/interfaces.html
```diff
@@ -0,0 +1,118 @@
...
+         idlArray.add_objects({'RTCIceCandidate': ['iceCandidate']});
+       }
+
+       function initRTCSessionDescription(idlArray) {
+         window.sessionDesc = new RTCSessionDescription({ type: 'offer' });
+         idlArray.add_objects({'RTCSessionDescription': ['sessionDesc']});
+       }
+
+       function initRTCIceEvent(idlArray) {
+         window.iceEvent = new RTCPeerConnectionIceEvent('ice-event', { });
```
foolip: Omit the empty dictionary.

webrtc/interfaces.html
```diff
@@ -0,0 +1,118 @@
...
+         ['Init RTCCertificate', initRTCCertificate],
+       ];
+
+       function initIdlArray(idlText) {
+         const idlArray = new IdlArray();
+
+         idlArray.add_untested_idls('interface EventHandler {};');
+         idlArray.add_untested_idls('interface MediaStreamTrack {};');
+         idlArray.add_idls(idlText);
+
```
foolip: This is where I'd suggest putting the first `idlArray.add_objects` for the synchronous things. Then, return a promise that is resolved with `idlArray` once the `RTCCertificate` has been created and the second `idlArray.add_objects` call has added it. This way, the code in the `promise_test` below should remain unchanged.

webrtc/interfaces.html
```diff
@@ -0,0 +1,118 @@
...
+         return idlArray;
+       }
+
+       function doIdlTests(idlArray) {
+         for(const [taskName, initIdl] of syncInits) {
```
foolip: Following the above suggestions, this should become a one-liner `idlArray.test()`

soareschen:
> Do you mean https://w3c.github.io/webrtc-pc/#peer-to-peer-connections? AFAICT all the IDL from the spec has been copied.

All the IDL definitions have been copied from the spec. However only objects of IDLs in https://w3c.github.io/webrtc-pc/#peer-to-peer-connections are constructed for testing.

> Can you file an issue on idlharness.js? Any manual changes that need to be made will get in the way of eventually auto-updating these IDL files with a script.

I have filed #5714 for the bug.

> Do you need to do some kind of cleanup, or what is the X?

The call to `idlArray.add_objects()` are asynchronous inside nested promise_test. The final `idlArray.test()` tests all added objects in one go, and should be called only when the sub promise_tests are done. (Is nested testing allowed with testharness?)

webrtc/interfaces.html
```diff
@@ -0,0 +1,118 @@
...
+           test(t => {
+             initIdl(idlArray);
+             t.done();
+           }, taskName);
+         }
+
+         // Async initialize all IDL instances in parallel
```
foolip: I think all of this would disappear.

webrtc/interfaces.html
```diff
@@ -0,0 +1,118 @@
...
+           });
+       }
+
+       promise_test(() => {
+         return fetch('/interfaces/webrtc.idl')
+           .then(response => response.text())
+           .then(initIdlArray)
+           .then(doIdlTests);
+
+       }, 'WebRTC IDL Tests');
```
foolip: Call this "Test driver" since it's just a mechanism to collect IDLs and set up the real tests. (Like https://github.com/ayg/web-platform-tests/blob/master/XMLHttpRequest/interfaces.html)

foolip (changes requested):

foolip:
>> Do you mean https://w3c.github.io/webrtc-pc/#peer-to-peer-connections? AFAICT all the IDL from the spec has been copied.
>
> All the IDL definitions have been copied from the spec. However only objects of IDLs in https://w3c.github.io/webrtc-pc/#peer-to-peer-connections are constructed for testing.

Can you add TODOs for the rest? It will not be obvious from the test at a glance which interfaces aren't tested with instances.

>> Do you need to do some kind of cleanup, or what is the X?
>
> The call to `idlArray.add_objects()` are asynchronous inside nested promise_test. The final `idlArray.test()` tests all added objects in one go, and should be called only when the sub promise_tests are done. (Is nested testing allowed with testharness?)

I think that with the suggested changes, you should end up with just a single explicit `promise_test`, which as its last step will call `idlArray.test()`, so hopefully this need will disappear.

(It is possible to test tests, e.g. to create one `async_test` at the end of another. I always find it harder to follow though, so avoid it where possible.)

foolip:
I'll be OOO until May 2. I'd be happy to take a second look then, but there's no need to wait if some other reviewer is happy with the changes.

webrtc/interfaces.html
soareschen: According to the spec `RTCPeerConnectionIceEvent` constructor accepts non-nullable and non-optional `eventInitDict` argument. (https://w3c.github.io/webrtc-pc/#constructors-3) Does this mean issue should be raised on the spec to make the arguments optional?

foolip: I see this:
```WebIDL
[Constructor(DOMString type, optional RTCPeerConnectionIceEventInit eventInitDict)]
interface RTCPeerConnectionIceEvent : Event {
    readonly attribute RTCIceCandidate? candidate;
    readonly attribute DOMString?       url;
};
```
I other words, it's already optional, or am I looking at the wrong thing?

soareschen: I see. I was looking at the table for the constructor below instead of the text. Guess it is supposed to be consistent with the IDL text. I will raise issue to the spec.

rtcpeerconnectioniceevent

foolip: Ah, that's wrong, good catch. Be sure to always look at the normative bits (mostly IDL, algorithms, and things that say "must") and not non-normative description. If it's ever not clear in a spec if a part is supposed to be the requirement or the explanation of it, please file spec issues :)

w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/5712/)

webrtc/interfaces.html
soareschen: Got it. I was copying the boilerplate from the existing test files. Now the redundant markups are removed.

webrtc/interfaces.html
soareschen: Missed updating it when copying boilerplate. Fixed.

webrtc/interfaces.html
soareschen: Good to know. I have now inline the sync constructions according to specs inside `add_objects`.

webrtc/interfaces.html
```diff
@@ -0,0 +1,55 @@
...
+          'RTCPeerConnection': ['new RTCPeerConnection'],
+          'RTCSessionDescription': ['new RTCSessionDescription({ type: "offer" })'],
+          'RTCIceCandidate': ['new RTCIceCandidate'],
+          'RTCPeerConnectionIceEvent': ['new RTCPeerConnectionIceEvent("ice")'],
+          'RTCPeerConnectionIceErrorEvent': ['new RTCPeerConnectionIceErrorEvent("ice-error", { errorCode: 701 });'],
+        });
+
+        return initRTCCertificate()
+          .then(cert => {
+            window.rtcCert = cert;
```
soareschen: The caveat here is that none of the IDL tests will run if `RTCPeerConnection.generateCertificate` fails. I am looking for a way to catch the promise rejection, report it to testharness, and proceed the rest of IDL tests regardless. The simplest solution without additional code is to modify `promise_test` to return a promise of the test completing.

webrtc/interfaces.html
foolip: If you like you can go further, removing this and the trailing bits:
```HTML
<html>
  <head>
    <title>"WebRTC IDL Tests</title>
  </head>
  <body>
```

webrtc/interfaces.html
```diff
@@ -0,0 +1,74 @@
+<!doctype html>
+<html>
+  <head>
+    <title>"WebRTC IDL Tests</title>
```
foolip: Stray " here

webrtc/interfaces.html
```diff
@@ -0,0 +1,74 @@
...
+        .then(cert => {
+          window.rtcCert = cert;
+          idlArray.add_objects({'RTCCertificate': ['rtcCert']});
+         });
+      }
+
+      // A customized version of test_harness that returns
+      // promise of the test completing as well as allow
+      // other promise tests to run in parallel.
+      function promise_test(testFunc, testName, properties) {
```
foolip: This is odd, can you elaborate on why you need it? If it is needed, then some other name would be clearer, but I doubt that there's anything here that couldn't be parallelized without this custom test type.

webrtc/interfaces.html
```diff
@@ -0,0 +1,74 @@
...
+          'Test Driver for RTCCertificate')
+        .then(() => {
+          idlArray.test();
+        })
+      }
+
+      promise_test(() => {
+        return fetch('/interfaces/webrtc-pc.idl')
+          .then(response => response.text())
+          .then(doIdlTest);
```
foolip: I think the structure is almost right now. If you make a `getRTCCertificate` return a promise that resolves with an RTCCertificate instance and doesn't do anything else, then you can do this, with the original `promise_test`:
```JavaScript
 promise_test(() => {
  return Promise.all([fetch('/interfaces/webrtc-pc.idl').then(response => response.text()),
                      getRTCCertificate()])
    .then(doIdlTest);
}, 'Test driver');
```

`doIdlTest` would take two arguments, `idlText` and `certificate` and use that to initialize the `idlArray` before finally calling `idlArray.test()`.

foolip (changes requested):

webrtc/interfaces.html
soareschen: The problem is that the original `promise_test()` cannot be used here, because it is nested inside another promise_test. And promise_test only execute a test after the previous test promise is resolved, thus causing deadlock and time out the entire test.

Technically I can inline the code inside `initRTCCertificate()`. But I feel that it is clearer as a separate function and more async IDL tests will be added here anyway.

I tentatively named it `promise_test` to serve as discussion on whether it makes any sense and whether something like this is worth to be added into testharness.js.

webrtc/interfaces.html
soareschen: I think the caveat of this approach is that if `getRTCCertificate()` fails, the entire IDL test suite don't get to run at all. I guess the original rationale for `idl_array.add_objects()` to take in strings and evaluate them is so that it can catch and report error on individual evaluations without affecting the other object tests. But the string approach don't work on promises so we have to hack around while maintaining the behavior.

An alternative approach is that I can do something like:

```javascript
initRTCCertificate()
  .catch(err => {
    test(t => assert_unreached('failed to initialize RTCCertificate'))
  })
```

This will report the error without affecting the other IDL tests. Though the tradeoff is the test case only appear if there is a failure.

webrtc/interfaces.html
soareschen: Mm not sure if removing even the `<html>` and `<body>` tags is a good idea. Do you have an example of other WG tests omitting them? I'd prefer to keep them at least.

foolip: Many tests are more minimal:
https://github.com/w3c/web-platform-tests/blob/master/dom/historical.html
https://github.com/w3c/web-platform-tests/blob/master/fullscreen/api/document-onfullscreenerror.html
https://github.com/w3c/web-platform-tests/blob/master/webrtc/RTCPeerConnection-createDataChannel.html

But, you get to pick your style, just wanted to make sure you weren't including stuff because you though it's required.

webrtc/interfaces.html
foolip: I see, you don't want the whole test to fail if `initRTCCertificate()` does, and apparently Edge doesn't have `RTCPeerConnection.generateCertificate()`. How about just adding `if (!RTCPeerConnection.generateCertificate) return null` and then not adding the instance if it's null?

webrtc/interfaces.html
foolip: A separate test that fails shouldn't be necessary, there should be other failures as long as `RTCPeerConnection.generateCertificate` doesn't exist that makes sure once can't fully pass the test without it.

foolip:
I tried making the suggested changes myself to make sure it's actually possible and not waste your time, but now am not sure how to get it onto this PR. Did you check [Allow edits from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) when creating the PR? I can't tell if that's enabled by default or not.

Anyway, I pushed the commit to a fork instead, review and push https://github.com/foolip/web-platform-tests/commit/cfa9ccc88f454de28387e28222d21fd46ca50129 to this PR? With those changes, this LGTM.

agouaillard:
@foolip, shares will merge in your proposed changes tomorrow (2am here), and move forward. We did a manual count today of the coverage of the spec (normative lines) and just reached 35%. will keep you informed of the coverage weekly.

soareschen:
LGTM. I also added a catch handler to ignore any promise rejection.

I remembered checking the "Allow edits from maintainers" when creating PR, but it's not enabled. Anyway I have checked it again on the sidebar.

Let's get this merged first. :)

webrtc/interfaces.html
soareschen: I see. Good to know some examples. Will make use of the simpler styles in new PRs then.

webrtc/interfaces.html
```diff
@@ -19,7 +19,8 @@
           modulusLength: 2048,
           publicExponent: new Uint8Array([1, 0, 1]),
           hash: 'SHA-256'
-        });
+        })
+        .catch(() => {}); // ignore error
```
foolip: Note that this will cause doIdlTests to later be called with undefined, while the early return will result in null. That's OK, both are falsy.

foolip:
I have squashed into a single commit and merged, please check if the commit message is what you had in mind. (Normally the commit message of the very first commit still applies, but saying something about RTCCertificate seemed useful.)

foolip:
> We did a manual count today of the coverage of the spec (normative lines) and just reached 35%. will keep you informed of the coverage weekly.

Can you describe the methodology here?

foolip:
Catching up on my inbox, I found it described in https://lists.w3.org/Archives/Public/public-webrtc/2017Apr/0080.html

PR #5738, opened 2017-04-28
[html] Add test for shared event loops

jugglinmike:
Add a test to verify that browsing contexts of similar origin share the
same event loop. Today, this test passes in the Chromium browser but
fails in the Firefox browser.

### DIFF
--- a/html/webappapis/scripting/event-loops/promise-from-dicarded-ctx.html
+++ b/html/webappapis/scripting/event-loops/promise-from-dicarded-ctx.html
@@ -21,9 +21,8 @@ async_test(function(t) {
 
       child.close();
 
-      // Create a new Promise (rather than invoke `fulfilled.then` directly) in
-      // order to append to the list stored in the "orphaned" Promise's
-      // [[PromiseFulfilledReactions]] slot.
+      // The use of Promise.resolve is necessary to trigger incorrect behavior
+      // in some browsers at the time of this writing"
       Promise.resolve(fulfilled).then(function() {
           t.done();
         });
@@ -37,9 +36,8 @@ async_test(function(t) {
 
       iframe.remove();
 
-      // Create a new Promise (rather than invoke `fulfilled.then` directly) in
-      // order to append to the list stored in the "orphaned" Promise's
-      // [[PromiseFulfilledReactions]] slot.
+      // The use of Promise.resolve is necessary to trigger incorrect behavior
+      // in some browsers at the time of this writing"
       Promise.resolve(fulfilled).then(function() {
           t.done();
         });
--- a/html/webappapis/scripting/event-loops/promise-from-dicarded-ctx.html
+++ b/html/webappapis/scripting/event-loops/promise-from-dicarded-ctx.html
@@ -1,0 +1,53 @@
+<!DOCTYPE html>
+<head>
+<meta charset=utf-8>
+<link rel=help href="https://html.spec.whatwg.org/multipage/webappapis.html#event-loops">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+</head>
+<body>
+  <script>
+// 8.1.4.1 Definitions
+//
+// > [...]
+// >
+// > There must be at least one browsing context event loop per user agent, and
+// > at most one per unit of related similar-origin browsing contexts.
+
+async_test(function(t) {
+  var child = window.open('/common/blank.html');
+  child.onload = function() {
+      var fulfilled = child.Promise.resolve();
+
+      child.close();
+
+      // Create a new Promise (rather than invoke `fulfilled.then` directly) in
+      // order to append to the list stored in the "orphaned" Promise's
+      // [[PromiseFulfilledReactions]] slot.
+      Promise.resolve(fulfilled).then(function() {
+          t.done();
+        });
+    };
+}, 'closing a window does not preclude execution of further tasks');
+
+async_test(function(t) {
+  var iframe = document.createElement('iframe');
+  iframe.onload = function() {
+      var fulfilled = iframe.contentWindow.Promise.resolve();
+
+      iframe.remove();
+
+      // Create a new Promise (rather than invoke `fulfilled.then` directly) in
+      // order to append to the list stored in the "orphaned" Promise's
+      // [[PromiseFulfilledReactions]] slot.
+      Promise.resolve(fulfilled).then(function() {
+          t.done();
+        });
+    };
+
+  iframe.src = '/common/blank.html';
+  document.body.appendChild(iframe);
+}, 'removing an iframe does not preclude execution of further tasks');
+</script>
+</body>
+</html>


w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/5738/)

html/webappapis/scripting/event-loops/shared-between-contexts.html
```diff
@@ -0,0 +1,56 @@
...
+  <script>
+// 8.1.4.1 Definitions
+//
+// > [...]
+// >
+// > There must be at least one browsing context event loop per user agent, and
+// > at most one per unit of related similar-origin browsing contexts.
+
+async_test(function(t) {
+    new Promise(function(resolve) {
```
domenic: Missing `return`

html/webappapis/scripting/event-loops/shared-between-contexts.html
```diff
@@ -0,0 +1,56 @@
...
+        child.onload = function() {
+            resolve(child);
+          };
+      })
+      .then(function(child) {
+          var fulfilled = child.Promise.resolve();
+
+          child.close();
+
+          Promise.resolve(fulfilled).then(function() {
```
domenic: What is the purpose of `Promise.resolve(fulfilled).then` instead of just `fulfilled.then`?

domenic (changes requested):
I wouldn't state this test as testing event loop sharing. You could fail this in various ways. One way is creating an event loop per window, which is bad. But another is preventing queuing tasks by promises created in dead windows, which is also bad.

I am pretty sure Firefox does not create an event loop per window, so instead this is likely triggering a bug where they prevent queuing tasks by promises created in dead windows, or similar.

So basically the test descriptions are good but the commit message/title are not.

html/webappapis/scripting/event-loops/shared-between-contexts.html
jugglinmike: I'm using `async_test` here to avoid implicitly relying on Promise operations that are defined externally.

domenic: Why wrap onload in a promise at all then? You can just put your code in there.

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/227009768)

# Lint #

Passed

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/227009770)

# Firefox (nightly channel) #
Testing web-platform-tests at revision aac553ccc4c4a4f6cf06d8b1577b826d43202fb8
Using browser at version BuildID 20170427100655; SourceStamp 0b77ed3f26c5335503bc16e85b8c067382e7bb1e
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="http://w3c-test.org/submissions/5738/html/webappapis/scripting/event-loops/shared-between-contexts.html">/html/webappapis/scripting/event-loops/shared-between-contexts.html</a></summary>


|                              Subtest                              | Results |     Messages     |
|-------------------------------------------------------------------|---------|------------------|
|                                                                   | TIMEOUT |                  |
| `closing a window does not preclude execution of further tasks`   | PASS    |                  |
| `removing an iframe does not preclude execution of further tasks` | TIMEOUT | `Test timed out` |

</details>
</details>

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/227009776)

# Chrome (unstable channel) #
Testing web-platform-tests at revision aac553ccc4c4a4f6cf06d8b1577b826d43202fb8

html/webappapis/scripting/event-loops/shared-between-contexts.html
jugglinmike: I was hoping to limit nesting, but I'm happy to simplify

html/webappapis/scripting/event-loops/shared-between-contexts.html
jugglinmike: Short answer: because Firefox passes when using `fulfilled.then`.

My guess here is that it comes down to [PerformPromiseThen](https://tc39.github.io/ecma262/#sec-performpromisethen). Firefox seems like it is correctly using the event loop when it comes to calling EnqueueJob from `fulfulled`, but that appending to the [[PromiseFulfillReactions]] of a newly-created Promise is where things go wrong.

Maybe it would be best to include tests for both cases (so four tests total). Does that sound good to you?

domenic: Oh jeez.

What happens if you do `Promise.resolve().then(() => fulfilled.then(() => test code))`, or `setTimeout(() => fulfilled.then(() => test code))`? I'm trying to get a sense if the extra Promise.resolve() just happens to make things wait a bit longer, and that's what causes the failure, or if it's actually a matter of something specific to promises.

jugglinmike:
How about

    [html] Add test for task management from discarded contexts

    Add a test to verify for the proper management of tasks created from
    Promises of discarded browsing contexts. Today, this test passes in the
    Chromium browser but fails via a timeout in the Firefox browser.

html/webappapis/scripting/event-loops/shared-between-contexts.html
jugglinmike: Hah! I ran the same experiment prior to submitting. Delaying the call to `then` via a timeout does not matter--Firefox still passes in this case

domenic: Wow, OK. At this point I'm not going to try to hypothesize why this is broken in Firefox. I'm OK with just these two tests. Maybe with a comment saying that the Promise.resolve() is important.

jugglinmike:
@domenic What about the file name? It sounds like you'd want to move it to a different directory altogether, but I'm not sure which one

domenic:
Oh, hmm. I think the directory is OK, but maybe just something like promises-in-closed-windows.html or something.

jugglinmike:
@domenic New version is up. I had to amend and force-push to implement the commit message changes, so I created another branch with the original commit in case that helps: https://github.com/w3c/web-platform-tests/compare/master...bocoup:event-loops-shared-orig

html/webappapis/scripting/event-loops/promise-from-dicarded-ctx.html
```diff
@@ -0,0 +1,53 @@
...
+async_test(function(t) {
+  var child = window.open('/common/blank.html');
+  child.onload = function() {
+      var fulfilled = child.Promise.resolve();
+
+      child.close();
+
+      // Create a new Promise (rather than invoke `fulfilled.then` directly) in
+      // order to append to the list stored in the "orphaned" Promise's
+      // [[PromiseFulfilledReactions]] slot.
```
domenic: This doesn't make sense to me. `Promise.resolve(fulfilled)` is basically `new Promise((resolve, reject) => fulfilled.then(resolve, reject))`, which doesn't append to `fulfilled`'s [[PromiseFulfillReactions]].

Indeed, given that `fulfilled`'s [[PromiseState]] is "fulfilled", I believe it's impossible to ever append to its [[PromiseFulfillReactions]].

jugglinmike: Oh, you're right. I mis-read the spec (thinking that the [[PromiseState]] was not set to "fulfilled" synchronously, even for `then`). I'll stop trying to be helpful and just write, "The use of `Promise.resolve` is necessary to trigger incorrect behavior in some browsers at the time of this writing"

domenic (approved):

jugglinmike:
Thanks, Domenic!

annevk:
Is this actually correct though?

User agents must ignore tasks queued by documents that are not fully active. It's right there in step 1 of the event loop's processing model.

annevk:
Maybe it's different because these are promises and thus end up adding a microtask, which don't have such an association (although they do end up being run as a task, even though I'm not entirely clear on what that means).

jugglinmike:
@bzbarsky This test is valid to the best of my understanding, but Anne has suggested that there may be some ambiguity in the specification that needs to be addressed. He also recommended we confer with you. Does this test look good to you?

bzbarsky:
There are outstanding issues on promises in navigated-away-from windows.  The specs literal text currently requires that such promises continue running forever, even when the window has long since been navigated away from.  This is a clear violation of user expectations and doesn't match what UAs do in practice in various situations, including Chrome.

There are also outstanding issues on promises in windows whose browsing context has gone away, including whether those should behave similarly to promises in navigated-away-from windows or not.  There's a bit less clarity around this in terms of user expectations and UA behavior, but once again UAs are not very consistent about it.  As a simple example, in Chrome last I checked if you close a tab whether promises in it keep running depends on random things like whether other tabs are observing it, which process things got scheduled on, etc.  But it's been a while since I checked; maybe that has been changed in the meantime.  Per current (broken, imo) spec, of course, promise execution never stops, ever.  There is simply no such concept in the spec.

domenic:
> User agents must ignore tasks queued by documents that are not fully active. It's right there in step 1 of the event loop's processing model.

In this case the task is queued by a document that is fully active:

> if the task was queued in the context of a browsing context, then [its Document] is the browsing context's active document at the time the task was queued

This particular test is still valid. If it were testing something like `child.then(...); window.close()`, the task should not execute. But it's not testing that; it's effectively testing `window.close(); child.then(...);`.

None of this detracts from @bzbarsky's point about overall UA inconsistency here, but I think the first step in addressing that inconsistency is to get tests like this one merged.

bzbarsky:
I think my basic position is that the spec is currently so wrong on this that no one should be implementing it as written, and in fact no one does.  What I want to avoid is tests that imply that the spec should be implemented as currently written, or worse yet "force" UA implementors who might not be as deeply involved in the standards process and hence always take spec text and WPT at face value to implement things that are just wrong conceptually.

What that means in terms of merging these specific tests, I don't know.  Maybe the failure messages should just point out the potential problems, to make it clear that simply making the tests pass may or may not be the right course of action...

annevk:
Are the issues with promises mentioned above tracked somewhere? This is the first time I hear about them. I thought we had solved promise queuing.

bzbarsky:
> Are the issues with promises mentioned above tracked somewhere?

Hmm.  I had thought I had filed them, but I'm not seeing anything in the obvious places (whatwg HTML issues, ES issues).

Filed https://github.com/whatwg/html/issues/2621

jgraham:
I think I'm right that this is marked as "Approved" but actually blocked from merging by spec issues?

domenic:
I guess that's @bzbarsky's opinion at least; I disagree and think this test is valuable as-is and that the current spec is pretty OK, but I don't feel strongly enough to push for merging over his objections.

annevk:
What does pretty OK mean? It's very much unclear who owns microtasks and what to do when the owner is no longer fully active, right? So if Firefox wanted to fix what it's doing here, it's very much unclear what the right fix would be, other than something that passes this test and does not regress others (assuming there's no contradictions).

domenic:
To be more clear: I think the spec as-written is good. We could contemplate making changes to it since it seems not everyone is implementing it interoperably, and that might make things easier for some implementers, but we could also just stay the course.

annevk:
If that's your opinion, maybe mention it in the relevant issue? It definitely seems highly incompatible with implementations that support inactive documents (which is something that the HTML Standard also tries to support).

PR #5757, opened 2017-05-02
Add a test for property lookup in event handlers

foolip:
Context:
https://bugs.chromium.org/p/chromium/issues/detail?id=663662#c20

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/5757)

### DIFF
--- a/html/webappapis/scripting/events/event-handler-scope.html
+++ b/html/webappapis/scripting/events/event-handler-scope.html
@@ -1,0 +1,24 @@
+<!DOCTYPE html>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<!-- There are two elements with id="foo" because this affects the behavior
+     in Gecko for the event handler attribute case. -->
+<button id="foo" onclick="window.fooInClick = foo">x</button>
+<object id="foo"></object>
+<script>
+function foo() { return "unique value"; }
+const fooButton = document.querySelector('button');
+const fooObject = document.querySelector('object');
+test(function() {
+  // assert_equals(foo, foo) would not be meaninful, so instead
+  // invoke and check return value to test what foo is.
+  assert_equals(foo(), "unique value");
+}, "Lookup of `foo` inside a plain script");
+test(function() {
+  // The click event handler synchronously sets fooInClick.
+  assert_false(window.hasOwnProperty('fooInClick'));
+  fooButton.click();
+  assert_true(window.hasOwnProperty('fooInClick'));
+  assert_equals(fooInClick, fooObject);
+}, "Lookup of `foo` inside an event handler attribute");
+</script>


w3c-bots:
These tests are now available on [w3c-test.org](https://w3c-test.org/submissions/5757/)

foolip:
@domenic, can you help me figure out if the tested behavior is actually required by the sum of specs? The test passes everywhere except Gecko, so bug to be filed if this test is merged.

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/227947068)

# Firefox (nightly channel) #
Testing web-platform-tests at revision da007380ce8975b861331c5a489583898cfcb047
Using browser at version BuildID 20170501100238; SourceStamp 076a7a66096f9e8d102548397254be32eb26bc3d
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="http://w3c-test.org/submissions/5757/html/webappapis/scripting/events/event-handler-scope.html">/html/webappapis/scripting/events/event-handler-scope.html</a></summary>


|                      Subtest                      | Results |                                                                      Messages                                                                      |
|---------------------------------------------------|---------|----------------------------------------------------------------------------------------------------------------------------------------------------|
|                                                   | OK      |                                                                                                                                                    |
| `Lookup of foo inside a plain script`             | PASS    |                                                                                                                                                    |
| `Lookup of foo inside an event handler attribute` | FAIL    | `assert_equals: expected (object) Element node <object id="foo"></object> but got (function) function "function foo() { return "unique value"; }"` |

</details>
</details>

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/227947074)

# Chrome (unstable channel) #
Testing web-platform-tests at revision da007380ce8975b861331c5a489583898cfcb047

annevk:
Note that the following already yields different results
```
.<button id="foo">x</button>
<object id="foo"></object>
<script>w(window.foo)</script>
```
and indicate a bug in Firefox per the HTML Standard (it doesn't return an `HTMLCollection`, it returns just the `button` element). Not entirely sure how in your scenario though Chrome ends up returning a single object from `.foo`.

foolip:
Interesting, per https://html.spec.whatwg.org/multipage/browsers.html#dom-window-nameditem an HTMLCollection should be returned, but it doesn't look like *any* engine actually does that?

annevk:
In my example above Chrome does at least (only tested Chrome and Firefox).

zcorpan:
`HTMLCollection` is returned for @annevk's test in WebKit/Chromium (and Edge?).

@foolip your test here is different because you're doing bareword `foo` in onclick attribute, which has weird scope rules per https://html.spec.whatwg.org/multipage/webappapis.html#getting-the-current-value-of-the-event-handler

annevk:
Oh right. I guess in @foolip's case because the document is in scope you'd end up using https://html.spec.whatwg.org/multipage/dom.html#dom-document-nameditem which would return the object element.

zcorpan:
See http://software.hixie.ch/utilities/js/live-dom-viewer/saved/5106

annevk:
Which means Firefox is wrong, I think.

annevk:
Anyway, it's probably better to turn this into a test for document's getter than this current convoluted setup.

sidewayss:
In order to give this some real world context (I am the one who raised the issue to @foolip):
I am linking an object to an input via the id attribute.  The object contains an SVG icon that serves as the user interface for the input, checkbox, radio button, or color picker.

This issue came up because not only did I give an input and an object the same id, I also gave the onchange event handler function the same name as the id.  It was convenient and makes sense largely because it's a checkbox - on/off, no related inputs like radio buttons.

My workaround/solution is to rename the function from "ID" to "changeID".  As I said, the input and object must share the id (although I <i>could</i> use a custom attribute), but the function name can be anything.
I don't know if that helps anything, but I thought it might be useful to provide a real world context.

foolip:
Looks like there are a bunch of tests for https://html.spec.whatwg.org/multipage/dom.html#dom-document-nameditem already in https://github.com/w3c/web-platform-tests/tree/master/html/dom/documents/dom-tree-accessors

I'll try to figure out (tomorrow?) if any test coverage is missing, and if a Gecko bug needs to be filed for this.

foolip:
Closing this in favor of https://github.com/w3c/web-platform-tests/pull/5860

PR #5803, opened 2017-05-05
Move tests for reversing an animation to the timing-model folder

jgraham:
We are gradually line the tests up better with sections in the spec. Given that
these tests are concerned with testing the "reverse an animation" procedure in
the spec (and not the API per se) they should be moved to an appropriate
subdirectory of timing-model.

We also update the test titles to make it clear that these tests are really
covering the timing model, not the API.

MozReview-Commit-ID: J5gc3HZg9qv


jgraham:
Code reviewed upstream.

PR #5805, opened 2017-05-05
Add a test for reversing with an infinite target effect end and playback rate of zero

jgraham:
We already pass this test but it seemed like a useful test to add since I was
unsure if the early return we have in Animation::Reverse() is valid or not.

MozReview-Commit-ID: J38Euno3VP6


jgraham:
Code reviewed upstream.

PR #5806, opened 2017-05-05
Add tests that the playback rate is updated silently

jgraham:
The spec[1] says:

  Silently set the animation playback rate of animation to -animation playback
  rate.

  This must be done silently or else we may end up resolving the current
  ready promise when we do the compensatory seek despite the fact that we are
  most likely not exiting the pending play state.

This patch add tests that we don't exit the pending play state when calling
reverse() or resolve the ready promise.

[1] https://w3c.github.io/web-animations/#reverse-an-animation

MozReview-Commit-ID: 1X42O5yKpk9


jgraham:
Code reviewed upstream.

PR #5812, opened 2017-05-05
Test pointer move with element origin and click with navigation;

jgraham:
Verify that there is no hang when we use pointer actions to
perform a click that results in navigation and document unload.

MozReview-Commit-ID: EO5FClnxML5

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1345653 [ci skip]


jgraham:
Code reviewed upstream.

PR #5814, opened 2017-05-05
Make callback-multiple-calls more stable.

jgraham:
Trigger more idle periods by forcing them to be short, using rAF.
Add a timeout to force idle callbacks to run.
Halve the number of rICs we schedule during the test.

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1352670 [ci skip]


jgraham:
Code reviewed upstream.

PR #5823, opened 2017-05-05
Expand classList testing and port to wpt;

jgraham:
Because it tests some Gecko-specific things as well, I'm making two
copies, as advised by bz and jgraham.  One is to be submitted upstream,
and a second one has local changes.  This means most of the test is run
twice.

This overwrites the preexisting Element-classlist.html test upstream.  I
think I took the useful bits out of it (particularly replace() testing),
but there are some things that it had that I didn't think were
necessary, including: things that belong in idlharness; .className
testing; testing .contains() and stringification and hasAttribute() and
such after add/remove/etc. (instead of just testing getAttribute()); CSS
class selector matching.

MozReview-Commit-ID: JxPK7OyVLXa

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1354066 [ci skip]

### DIFF
--- a/dom/nodes/Element-classlist.html
+++ b/dom/nodes/Element-classlist.html
@@ -1,409 +1,462 @@
 <!doctype html>
-<html>
-  <head class="test test">
-    <title class=" ">Element.classList in case-sensitive documents</title>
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #5824, opened 2017-05-05
Split MathML entity test into multiple files,

jgraham:
This was timing out and causing OOM on some platforms, so instead use
one doctype per test.

MozReview-Commit-ID: 7M2FCJVPElK

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1357844 [ci skip]


jgraham:
Code reviewed upstream.

PR #5829, opened 2017-05-05
Update Selection.webidl to match spec

jgraham:
Practical changes:

1) Some additional method arguments are nullable or optional, which
matches Chrome/WebKit.  They make more sense non-nullable and
non-optional, but Chrome is afraid of the compat impact of changing.

2) Added [CEReactions] to deleteFromDocument().

MozReview-Commit-ID: Kg9EDubnEui

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1359371 [ci skip]


jgraham:
Code reviewed upstream.

PR #5832, opened 2017-05-05
enable border-spacing interpolation test.

jgraham:
So far, we don't have a type to test anamations of a pair of length.
Since border-spacing consists two absolute lengths, we shall add this
new type.

MozReview-Commit-ID: Bo8VMWPLDHc

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1354437 [ci skip]


jgraham:
Code reviewed upstream.

PR #5833, opened 2017-05-05
only check the mid-point while testing interpolation results in animations' web-platform-tests. r=hiro DONTBUILD (test-only)

jgraham:
In case of interpolation, we shall just check at 500ms.
No need to check 0ms and 1000ms.

MozReview-Commit-ID: FldqRXj0HLo

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1359786 [ci skip]


jgraham:
Code reviewed upstream.

PR #5837, opened 2017-05-05
Fix incorrect text expectation; r=masayuki

jgraham:
These expectations were just wrong.  I don't know how they got here.  I
think Chrome matches the tests and not the spec, fortunately.

MozReview-Commit-ID: ENiO34GiZ0Y

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1297414 [ci skip]


jgraham:
Code reviewed upstream.

Issue #5843, opened 2017-05-06
Reftests pointing at files in support directories has odd behaviour


gsnedders:
There's a bunch of reftests that point at files in support directories. While this happens to work okay in the limited number of cases we currently have, it doesn't work okay in general: they're picked up as support files by the manifest, and that means if they *did* have a `<link rel=match>` they wouldn't get picked up as a reftest node.

bobholt:
Marking as backlog until @gsnedders can get back to #5844.

jgraham:
Is the expected solution here to move the existing files? Scanning all files in support directories  in case they happen to be reftests seems like it could be a performance issue.

gsnedders:
@jgraham I think that's probably the right option. Some might need duplicated as a result.

PR #5844, opened 2017-05-06
Start enforcing references are named as such

gsnedders:
See <https://lists.w3.org/Archives/Public/public-test-infra/2016OctDec/0027.html> for some background on this. Also see #5843 when it comes to what expected behaviour is around some of the lint failures on this job.

### DIFF
--- a/tools/lint/lint.py
+++ b/tools/lint/lint.py
@@ -484,12 +484,19 @@ def check_parsed(repo_root, path, f):
 
         assert ref_parts.path != ""
 
-        reference_file = os.path.join(repo_root, ref_parts.path[1:])
+        reference_sourcefile = SourceFile(repo_root, ref_parts.path[1:], "/")
+        reference_file = reference_sourcefile.path
         reference_rel = reftest_node.attrib.get("rel", "")
 
         if not os.path.isfile(reference_file):
             errors.append(("NON-EXISTENT-REF",
                      "Reference test with a non-existent '%s' relationship reference: '%s'" % (reference_rel, href), path, None))
+
+        if not reference_sourcefile.name_is_reference:
+            errors.append(("REF-IS-NOT-REF-NAMED",
+                           "Reference test points at '%s' which isn't a reference filename" % href,
+                           path,
+                           None))
 
     if len(source_file.timeout_nodes) > 1:
         errors.append(("MULTIPLE-TIMEOUT", "More than one meta name='timeout'", path, None))
--- a/tools/manifest/manifest.py
+++ b/tools/manifest/manifest.py
@@ -346,7 +346,7 @@ class Manifest(object):
         changed_hashes = {}
 
         for item in reftest_nodes:
-            if item.url in has_inbound:
+            if item.url in has_inbound or item.source_file.name_is_reference:
                 # This is a reference
                 if isinstance(item, RefTest):
                     item = item.to_RefTestNode()
@@ -360,6 +360,9 @@ class Manifest(object):
                                                                  item.item_type)
                 reftests[item.source_file.rel_path].add(item)
             self._reftest_nodes_by_url[item.url] = item
+
+        # while currently obvious from the if/else, this is part of the API contract, so check!
+        assert len(reftest_nodes) == len(references) + len(reftests)
 
         return reftests, references, changed_hashes
 
--- a/tools/manifest/tests/test_manifest.py
+++ b/tools/manifest/tests/test_manifest.py
@@ -269,6 +269,17 @@ def test_reftest_computation_chain_update_node_change():
                        ("support", test3.path, {test3})]
 
 
+def test_reftest_computation_ref_name():
+    m = manifest.Manifest()
+
+    s = SourceFileWithTest("foobar-ref.html", "0"*40, item.RefTestNode, [("/other.html", "==")])
+    test = s.manifest_items()[1][0]
+
+    assert m.update([(s, True)]) is True
+
+    assert list(m) == [("reftest_node", test.path, {test})]
+
+
 def test_iterpath():
     m = manifest.Manifest()
 


w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/229478535)

# Firefox (nightly channel) #
Testing web-platform-tests at revision 0f59867a3152c34f63cf245e691af849baf95f47
Using browser at version BuildID 20170506100311; SourceStamp 37a5b7f6f101df2eb292b1b6baaf1540c9920e20
Starting 10 test iterations
No tests run.

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/229478541)

# Chrome (unstable channel) #
Testing web-platform-tests at revision 0f59867a3152c34f63cf245e691af849baf95f47

jgraham:
So this obviously doesn't actually pass with the current repo.

Codewise it looks fine, but I would like to encourage pointing to global refs like `green-square.html` or whatever; I'm not sure this patch doesn't discourage that (unless we make a top  level `references` directory, which might be fine I suppose).

jgraham (changes requested):
<see comment above; silly GitHub>

gsnedders:
> Codewise it looks fine, but I would like to encourage pointing to global refs like green-square.html or whatever; I'm not sure this patch doesn't discourage that (unless we make a top level references directory, which might be fine I suppose).

We have `css/reference` which this is fine with. Surely we want global refs to be in a top-level `reference` directory or similar though, as where else would we put global refs? I presume you mean encourage at a review-level, rather than listing for it!

sideshowbarker:
Are we going to keep this open indefinitely or is there something specific targeted to happen that will resolve it?

gsnedders:
@sideshowbarker I just need to spend some time on this, which I should be able to once the automation stuff lands.

foolip:
Ping @gsnedders, still relevant?

gsnedders:
@foolip yeah, still worth doing, because it gets rid of a lot of oddities around reftests (reftests pointing at things that are supposedly other tests gets weird fast).

svg/shapes/reftests/disabled-shapes-01.svg
```diff
@@ -4,7 +4,7 @@
     <h:link rel="help" href="https://svgwg.org/svg2-draft/shapes.html#RectElement"/>
     <h:link rel="help" href="https://svgwg.org/svg2-draft/shapes.html#CircleElement"/>
     <h:link rel="help" href="https://svgwg.org/svg2-draft/shapes.html#EllipseElement"/>
-    <h:link rel="match" href="support/empty.svg"/>
+    <h:link rel="match" href="about:blank"/>
```
AmeliaBR: I'd rather not rely on `about:blank` as a reference in SVG ref tests, since these need to work for non-browser SVG rendering environments.

(So, keep empty.svg, moved to an appropriately named references folder, which maybe could be `svg/reference/` to create some shared common references.)

gsnedders: How much effort would it be to special-case `about:blank` in any reftest runner for other implementations (you could just replace `about:blank` with a URL to an empty SVG)? Practically it should be equivalent to an empty SVG document, but using `about:blank` should improve caching behaviour for browsers given they're more likely to already have a cached rendering of about:blank.

gsnedders:
> I had to do a search for the `name_is_reference` test to see what the linter would actually be testing:

Yeah, not yet got round to writing docs for it. :)

> 
> [wpt/tools/manifest/sourcefile.py](https://github.com/web-platform-tests/wpt/blob/6111e21ca42dea24a07df1acfbed78769a415b5b/tools/manifest/sourcefile.py#L334-L338)
> Lines 334 to 338 in [6111e21](/web-platform-tests/wpt/commit/6111e21ca42dea24a07df1acfbed78769a415b5b)
>      @property 
>      def name_is_reference(self): 
>          """Check if the file name matches the conditions for the file to 
>          be a reference file (not a reftest)""" 
>          return "/reference/" in self.url or "/reftest/" in self.url or bool(reference_file_re.search(self.name)) 
> where
> 
> [wpt/tools/manifest/sourcefile.py](https://github.com/web-platform-tests/wpt/blob/6111e21ca42dea24a07df1acfbed78769a415b5b/tools/manifest/sourcefile.py#L22)
> Line 22 in [6111e21](/web-platform-tests/wpt/commit/6111e21ca42dea24a07df1acfbed78769a415b5b)
>  reference_file_re = re.compile(r'(^|[\-_])(not)?ref[0-9]*([\-_]|$)') 
> If I'm reading that correctly, it means that any files in a `reftest` directory are accepted as reference files. However, the reftest directories (for SVG, anyway) include both the test files (`feature.svg`) and the references (`feature-ref.svg`). If the purpose of having linting is to prevent circular references, the file path test will need to be stricter than that.

So if there's a directory called `reference` or `reftest` in the path, it's treated as a ref. Alternatively, if it's called `-ref` (or similar, `_notref`, `-ref1-foo`, etc.) it's treated as a ref.

This isn't actually a problem for SVG because your directories are called `reftests` and not `reftest`!

gsnedders:
It's only:

```
./css/css-namespaces/reftest
./css/css-transforms/reftest
./css/css-writing-modes/reftest
```

that hits the `reftest` case; maybe we should just rename them.

gsnedders:
In case anyone is wondering about my current interest in getting this landed:

If we want to make `./wpt run xxx path/to/reftest.html` faster while still doing a manifest update, we have to do a full rescan of everything in case there's now a `path/to/foobar.html` with `<meta rel=match href=reftest.html>`, in which case `reftest.html` is no longer a test but instead part of a larger reftest chain. If we can classify files as a reftest or a reference by filename, we can avoid having to do the full rescan.

svg/shapes/reftests/disabled-shapes-01.svg
dirkschulze: I am sceptical about `about:blank` as well. I don't think it is feasible to request implementations or test suites to replace it with something like an empty SVG file.

gsnedders:
Currently abandoned pending resolving https://github.com/web-platform-tests/rfcs/pull/15.

zcorpan:
@gsnedders https://github.com/web-platform-tests/rfcs/pull/15 is now merged; what should happen for this PR?

gsnedders:
Given the conclusion that led to that RFC was that we'd just let references have their own references (to implement and), this seems totally gratuitous now.

PR #5891, opened 2017-05-11
Test [PutForwards] behavior

jensl:
In particular, check that custom getters and setters are called as expected when accessing [PutForwards] attribute.


zcorpan:
To summarize internal discussion, this test comes from a Blink patch @jensl just landed. I informed him that external/wpt has two-way syncing these days, but for this one we can carry on with this PR over here.

domenic (approved):
Thanks so much for the prompt upstreaming!! <3

PR #6001, opened 2017-05-19
idlharness: Make descriptions match the assert_true/false expectations.

rakuco:
An assertion in the form of `assert_true(foo.enumerable, "foo is not
enumerable")` leads to a confusing error message like this:

    assert_true: foo is not enumerable expected true got false

which is the opposite of what we are actually expecting.

Conversely, `assert_false(foo.enumerable, "foo is enumerable")` is also
more confusing than "foo is not enumerable", which is what we are expecting
and asserting.

Instead, standardize on making the descriptions match what we expect in the
calls to `assert_true()` and `assert_false()` and use constructions such as
"should be <foo>" or "should not have <foo>" instead of factual statements.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6001)

### DIFF
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -489,7 +489,7 @@ IdlArray.prototype.assert_type_is = function(value, type)
 
     if (type.sequence)
     {
-        assert_true(Array.isArray(value), "is not array");
+        assert_true(Array.isArray(value), "should be an Array");
         if (!value.length)
         {
             // Nothing we can do.
@@ -500,7 +500,7 @@ IdlArray.prototype.assert_type_is = function(value, type)
     }
 
     if (type.generic === "Promise") {
-        assert_true("then" in value, "Attribute with a Promise type has a then property");
+        assert_true("then" in value, "Attribute with a Promise type should have a then property");
         // TODO: Ideally, we would check on project fulfillment
         // that we get the right type
         // but that would require making the type check async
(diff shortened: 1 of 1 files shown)


rakuco:
@jdm @annevk

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/234050443)

# Firefox (nightly) #
Testing web-platform-tests at revision 0f3f4cab8b4ef0de636e926ebd792ab95a2c64cf
Ignoring 1 changed files:
No files changed

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/234050444)

# Sauce (safari) #
Testing web-platform-tests at revision 0f3f4cab8b4ef0de636e926ebd792ab95a2c64cf
Ignoring 1 changed files:
No files changed

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/234050450)

# Chrome (unstable) #
Testing web-platform-tests at revision 0f3f4cab8b4ef0de636e926ebd792ab95a2c64cf
Ignoring 1 changed files:
No files changed

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/234050451)

# Sauce (MicrosoftEdge) #
Testing web-platform-tests at revision 0f3f4cab8b4ef0de636e926ebd792ab95a2c64cf
Ignoring 1 changed files:
No files changed

jdm (approved):

jdm (changes requested):
There are a bunch of other similarly inverted messages in this file. We should fix them all.

rakuco:
> There are a bunch of other similarly inverted messages in this file. We should fix them all.

I think I've done that, please take another look. I've touched only the `assert_true()` and `assert_false()` calls since these are the most problematic ones. The PR description has also been updated.

resources/idlharness.js
```diff
@@ -840,9 +840,9 @@ IdlInterface.prototype.test_self = function()
         var desc = Object.getOwnPropertyDescriptor(self, this.name);
         assert_false("get" in desc, "self's property " + format_value(this.name) + " has getter");
         assert_false("set" in desc, "self's property " + format_value(this.name) + " has setter");
-        assert_true(desc.writable, "self's property " + format_value(this.name) + " is not writable");
+        assert_true(desc.writable, "self's property " + format_value(this.name) + " is writable");
         assert_false(desc.enumerable, "self's property " + format_value(this.name) + " is enumerable");
-        assert_true(desc.configurable, "self's property " + format_value(this.name) + " is not configurable");
+        assert_true(desc.configurable, "self's property " + format_value(this.name) + " is configurable");
```
jdm: What about the assert_falses in this block? There are a lot of other ones in this file with the same problem.

resources/idlharness.js
```diff
@@ -1148,7 +1148,7 @@ IdlInterface.prototype.test_self = function()
 
         if (this.is_callback()) {
             assert_false("prototype" in self[this.name],
-                         this.name + ' should not have a "prototype" property');
+                         this.name + ' has a "prototype" property');
```
jdm: This looks like it's creating the same problem that's being fixed elsewhere.

resources/idlharness.js
```diff
@@ -1040,7 +1040,7 @@ IdlInterface.prototype.test_self = function()
 
         if (this.is_callback()) {
             assert_false("prototype" in self[this.name],
-                         this.name + ' should not have a "prototype" property');
+                         this.name + ' should have a "prototype" property');
```
jdm: This looks incorrect.

jdm (changes requested):
It looks like a bunch of the assert_false changes are incorrectly changing messages which match the pattern we want, and ignoring lots of messages which are incorrectly negated.

resources/idlharness.js
rakuco: They look OK to me. The error message is going to say something like

    self's property foo is enumerable: expected false got true

which to me reads as "desc.enumerable should be false, but it is true". Are you suggesting negating that sentece instead?

resources/idlharness.js
rakuco: It's the same as the previous case IMO: we're saying "foo should have a 'prototype' property: expected false got true" instead of "foo should not have a 'prototype' property: expected false got true" which looks like "expected false to 'should not have a 'prototype' property' aka 'expected true to 'should have a 'prototype property''.

resources/idlharness.js
rakuco: Ditto.

resources/idlharness.js
annevk: Yeah, the description needs to match the expectation. See http://web-platform-tests.org/writing-tests/testharness-api.html. We could document that better though.

annevk: (To be clear, jdm is correct.)

rakuco: OK, so I take it that we should have `assert_false(desc.enumerable, "[...] is not enumerable")`? I thought doing so would make the messages more confusing, but I can do that if it's the general preference.

annevk: Instead of a factual statement you could phrase it as a requirement: "X should not be Y" or "X should be Y". That makes it more self-evident that it's currently wrong.

rakuco: To be clear, in `assert_false("value" in desc)` is the expectation "'value' should be in desc" or the whole thing, ie. "'value' should be in desc should be false" aka "'value' should not be in desc"?

annevk: The expectation/requirement is '"value" should not be in desc'.

rakuco:
I've pushed a new version that hopefully addresses all comments. Please take another look.

jdm:
The changes look good to me! I don't know what to do about the code coverage failure on TravisCI, though, since it's preventing merging.

jdm (approved):

rakuco:
It looks like some test was flaky; I've repushed my commit and everything's green now.

jdm:
Thanks!

PR #6100, opened 2017-05-30
Make ECDH derive_bits_keys tests run on Safari/Firefox and correct HK…

alanwaketan:
…DF derive_bits_keys tests

For ECDH tests, add rejection handlers for ECDH key generations such that tests can continue.

For HKDF, infos are missing for bad hash tests, and therefore TypeErrors will return instead of OperationErrors for those tests. Added infos. Also, for null length, HKDF should return OperationError instead of TypeError.

### DIFF
--- a/WebCryptoAPI/derive_bits_keys/ecdh_bits.js
+++ b/WebCryptoAPI/derive_bits_keys/ecdh_bits.js
@@ -200,6 +200,8 @@ function run_test() {
                                             false, ["deriveBits", "deriveKey"])
                             .then(function(key) {
                                 privateKeys[namedCurve] = key;
+                            }, function(key) {
+                                privateKeys[namedCurve] = undefined;
                             });
             promises.push(operation);
         });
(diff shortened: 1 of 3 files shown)


alanwaketan:
@jimsch Please review. Thanks.

jimsch:
The order of the steps in the document are considered to be normative.  Thus the test for the bad algorithm must be performed before the infos are evaluated.  It is not clear to me that this is incorrect behavior.

alanwaketan:
I see. If this logic applied, why we have salt there.

alanwaketan:
Also, from the spec: https://www.w3.org/TR/WebCryptoAPI/#algorithm-normalization
In section 18.4.4. Normalizing an algorithm, Step 6 states the WebIDL conversion of a Dictionary type while Step 10.3 states the conversion of the HashAlgorithmIdentifier. If Step 6 fails, I don't see why Step 10.3 is able to report OperationError.

alanwaketan:
Sorry, NotSupportError.

sideshowbarker:
What is this waiting on?

jimsch:
There is a disagreement on the correct error to be returned.  There was a large amount of discussion about the fact that evaluation order is to be fixed per the document, a position that I opposed, and the order in the document being fixed is the side that won.  I am not convinced that this is observed by the changes provided so I am unwilling to be the one to approve it.

Issue #6202, opened 2017-06-09
Continuous integration build timeouts


jugglinmike:
As of 10:00 EST on Friday, June 9, all TravisCI jobs that rely on automating the Firefox web browser began to time out. Although [a new version of the GeckoDriver utility (0.17.0) was published today](https://github.com/mozilla/geckodriver/releases/tag/v0.17.0), this timeout also occurs when using the prior release (0.16.1). Using version 54 of Firefox (as opposed to the "Nightly" version 55) corrects the problem. This appears to be caused by a bug in the latest Nightly release of Firefox itself.

Because pull requests are gated on positive test results in this project, the timeout is interrupting ongoing work. I am experimenting with a patch that will instead use the "Beta" release channel. If that resolves the problem, I plan to merge it to `master` as a short-term solution until the underlying WebDriver issue is addressed.

(Note that the use of Firefox's Beta channel may also be generally preferable moving forward. It will help shield this project from similar regressions. Additionally, Firefox Beta channel's release cycle is analogous to the "dev"/"unstable" channel that is currently in use for the Chrome browser.)

jugglinmike:
The patch to switch the the "Beta" release channel is available at https://github.com/w3c/web-platform-tests/pull/6204. While this addresses the timeout as intended, unrelated differences between the releases cause new failures. See the comment thread of that pull request for more information.

jugglinmike:
[I reported the problem to the Firefox team](https://bugzilla.mozilla.org/show_bug.cgi?id=1371868) and learned that it had already been identified. It was resolved in the very next release of Firefox Nightly: https://bugzilla.mozilla.org/show_bug.cgi?id=1371319

The more general risk of instability from Firefox's Nightly release channel remains. In gh-6204, we've decided that the frequency of problems this causes is acceptable for the advantages of testing the very latest build (see that discussion for more information).

PR #6204, opened 2017-06-09
Install Firefox from Beta channel in CI

jugglinmike:
The latest build of Firefox available on the "Nightly" channel suffers
from timeout errors in automation. Switch the installation procedure to
instead source from the Beta channel in order to use a build that is
known to work today and is less susceptible to future regressions.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6204)

### DIFF
--- a/check_stability.py
+++ b/check_stability.py
@@ -189,11 +189,7 @@ class Firefox(Browser):
     def install(self):
         """Install Firefox."""
         call("pip", "install", "-r", os.path.join(wptrunner_root, "requirements_firefox.txt"))
-        index = get("https://archive.mozilla.org/pub/firefox/nightly/latest-mozilla-central/")
-        latest = re.compile("<a[^>]*>(firefox-\d+\.\d(?:\w\d)?.en-US.linux-x86_64\.tar\.bz2)</a>")
-        filename = latest.search(index.text).group(1)
-        resp = get("https://archive.mozilla.org/pub/firefox/nightly/latest-mozilla-central/%s" %
-                   filename)
+        resp = get("https://download.mozilla.org/?product=firefox-beta-latest&os=linux64&lang=en-US")
         untar(resp.raw)
 
         if not os.path.exists("profiles"):
--- a/tools/browserutils/browser.py
+++ b/tools/browserutils/browser.py
@@ -59,7 +59,7 @@ class Firefox(Browser):
     requirements = "requirements_firefox.txt"
 
 
-    def platform_string(self):
+    def platform_string_certutil(self):
         platform = {
             "Linux": "linux",
             "Windows": "win",
@@ -78,7 +78,7 @@ class Firefox(Browser):
 
         return "%s%s" % (platform, bits)
 
-    def platform_string_geckodriver(self):
+    def platform_string(self):
         platform = {
             "Linux": "linux",
             "Windows": "win",
@@ -110,7 +110,7 @@ class Firefox(Browser):
         if dest is None:
             dest = os.getcwd()
 
-        resp = self.get_from_nightly("<a[^>]*>(firefox-\d+\.\d(?:\w\d)?.en-US.%s\.tar\.bz2)" % self.platform_string())
+        resp = get("https://download.mozilla.org/?product=firefox-beta-latest&os=%s&lang=en-US" % self.platform_string())
         untar(resp.raw, dest=dest)
         return os.path.join(dest, "firefox")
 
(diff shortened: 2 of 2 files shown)


w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/241372948)

# Firefox (nightly) #
Testing web-platform-tests at revision d0a56903a2704df60e9c2681a61ff83135a5f4b6
Using browser at version BuildID 20170601203257; SourceStamp cb4a4275b365b126ae56b1df5521756b693d8ff9
Starting 10 test iterations
No tests run.

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/241372949)

# Sauce (safari) #
Testing web-platform-tests at revision d0a56903a2704df60e9c2681a61ff83135a5f4b6
Using browser at version 10.0
Starting 10 test iterations
No tests run.

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/241372955)

# Chrome (unstable) #
Testing web-platform-tests at revision d0a56903a2704df60e9c2681a61ff83135a5f4b6
Using browser at version 60.0.3112.20 dev
Starting 10 test iterations
No tests run.

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/241372956)

# Sauce (MicrosoftEdge) #
Testing web-platform-tests at revision d0a56903a2704df60e9c2681a61ff83135a5f4b6

jugglinmike:
This patch has failed the CI build because the `testharness.js` tests rely on a feature that is currently only available in the "Nightly" Firefox release channel (assignment of a `stack` parameter for rejected promises). Although the new failure demonstrates that use of the "Beta" channel resolves issue gh-6202 as intended, it cannot be merged in its current state due to this new failure.

@jgraham It may be worth temporarily disabling the `testharness.js` tests and landing this. Because it is now late on a Friday, the problem is unlikely to cause immediate problems, and it's unclear what anyone's availability will be like to support changes in the interim.

[I have reported this bug to Mozilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1371868) to help see this resolved in a future Nightly release.

jgraham (changes requested):
I think this is the wrong appraoch. We really want to run stability checks on nightly builds. The problem here is with the testharness.js tests. Running those on e.g. dev edition seems fine, but there should be a separate codepath for that. Note that today is merge day so beta will soon be like current nightly (approximately). Also note that gecko's beta is pretty unlike the Chrome dev channel since it's released every six weeks and frozen.

jugglinmike:
@jgraham Judging from the latest builds on `master`, it looks like this was a hiccup that was resolved in subsequent builds of Nightly. Even so, I want to respond to the following because it sounds like I may be misunderstanding something:

> We really want to run stability checks on nightly builds. The problem here is
> with the testharness.js tests. Running those on e.g. dev edition seems fine,
> but there should be a separate codepath for that.

Because the problem concerned the Firefox-to-WebDriver communication channel, it effected any job that used WebDriver to automate Firefox Nightly. Judging from this build for this specific patch, that might appear to be only the jobs that ran the tests for the `testharness.js` framework, but that is because no "typical" WPT tests were added/modified. If they had been, then the stability checker would have attempted to validate them, and the subsequent Firefox job would have timed out as well.

For example, see [the build](https://travis-ci.org/w3c/web-platform-tests/builds/241391067) for gh-6175 (as performed on Friday afternoon).

Does that sound right to you? If so, I think we'll need to consider another strategy that will be acceptable for both the tests for the `testharness.js` framework and the "typical" WPT tests.

jgraham:
Yeah, I hadn't realised when I wrote the above that all Fx jobs were affected. But I'm now inclined to do even less; unless you are volunteering to build an infrastructure to smoketest the builds we use and roll back to an earlier one if that fails, I suggest we treat occasional failures of this kind as part of the cost of testing the bleeding edge of the platform. It seems like this particular kind of failure should be very rare (and I don't understand this one) because all wpt tests are run on m-c and I would expect this level of failure to stop the build ever becoming a nightly. Still it obviously can happen sometimes.

jugglinmike:
> unless you are volunteering to build an infrastructure to smoketest the
> builds we use and roll back to an earlier one if that fails

This sounds like too big a job to complete on a voluntary basis, but there are some folks at Bocoup who would be well-suited to this. Then again, given what you've said about the expected frequency of the issue, we will probably be hard-pressed to fund a dedicated effort.

I'm satisfied with taking this as a cost of testing the bleeding edge. We can revisit your smoketest idea if that cost starts to seem to high.

PR #6246, opened 2017-06-14
Default to json operation

tobie:
This adds initial support for testing [Default] toJSON operation in idlharness (specified in https://github.com/heycam/webidl/pull/323).

This is work in progress, but initial comments are welcome.


w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/243010367)

# Firefox (nightly) #
Testing web-platform-tests at revision 01bf7330be1c2ef35ffef5882018b75c33a095ac
Ignoring 6 changed files:
No files changed

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/243010368)

# Sauce (safari) #
Testing web-platform-tests at revision 01bf7330be1c2ef35ffef5882018b75c33a095ac
Ignoring 6 changed files:
No files changed

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/243010378)

# Chrome (unstable) #
Testing web-platform-tests at revision 01bf7330be1c2ef35ffef5882018b75c33a095ac
Ignoring 6 changed files:
No files changed

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/243010379)

# Sauce (MicrosoftEdge) #
Testing web-platform-tests at revision 01bf7330be1c2ef35ffef5882018b75c33a095ac
Ignoring 6 changed files:
No files changed

domenic (approved):
I guess by "support for" you mean "infrastructure for adding support for in the future"?

Seems pretty great to have this level of unit tests and such, although it's a bit hard to tell what the methods are in service of since there's nothing that uses them yet besides the tests.

LGTM mostly based on reviewing the tests.

tobie:
> although it's a bit hard to tell what the methods are in service of since there's nothing that uses them yet besides the tests.

Yeah. That's about to change. :) They're there to test [Default] toJSON() and toJSON().

resources/idlharness.js
```diff
@@ -322,6 +322,155 @@ IdlArray.prototype.recursively_get_implements = function(interface_name)
...
+     * Note: as dictionary inheritance is expressed identically by the AST,
+     * this works just as well for getting a stack of inherited dictionaries. 
+     */
+
+    var stack = [];
+    stack.push(idl_interface);
+    while (idl_interface.base) {
+        var base = this.members[idl_interface.base];
+        if (!base) {
+            throw new Error(idl_interface.type + " " + idl_interface.base + " not found (inherited by " + idl_interface.name + ")");
```
jugglinmike: This behavior isn't covered by a test.

resources/test/tests/idlharness/basic.html
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE HTML>
+<html>
+<head>
+<title>idlharness base</title>
+    </head>
```
jugglinmike: Extra whitespace

resources/test/tests/idlharness/basic.html
```diff
@@ -0,0 +1,34 @@
...
+        assert_true(new IdlArray() instanceof IdlArray);
+    }, 'IdlArray constructor should be constructible');
+    test(function() {
+        assert_true("WebIDL2" in window);
+    }, 'WebIDL2 namespace should be a global object');
+    test(function() {
+        assert_equals(typeof WebIDL2.parse, "function");
+    }, 'WebIDL2 namespace should have a parse method');
+    test(function() {
+        assert_throws(new TypeError(), function() {  WebIDL2.parse("I'm a syntax error") })
```
jugglinmike: Missing semicolon

jugglinmike (commented):
It's my understanding that consumers are still in varying state of completeness when it comes to ES2015. Tests for new web platform features use new syntax because those implementations skip those tests outright. Use in `idlharness.js` would force them to ignore tests that they currently rely on, so I think we should avoid ES2015 syntax here.

resources/idlharness.js
tobie: I think it is by this: https://github.com/w3c/web-platform-tests/pull/6246/files#diff-061d4b3be7a8a8e16e9b15230b603d47R25

resources/idlharness.js
```diff
@@ -322,6 +322,155 @@ IdlArray.prototype.recursively_get_implements = function(interface_name)
...
+
+    // union and record types
+    if (type.union || type.generic == "record") {
+        return idlType.every(this.is_JSON_type, this);
+    }
+
+    // sequence types
+    if (type.generic == "sequence") { return this.is_JSON_type(idlType); }
+    
+    if (typeof idlType == "string")
```
tobie: Wow, indeed, I completely missed this. Thanks for pointing it out.

tobie:
> It's my understanding that consumers are still in varying state of completeness when it comes to ES2015. Tests for new web platform features use new syntax because those implementations skip those tests outright. Use in idlharness.js would force them to ignore tests that they currently rely on, so I think we should avoid ES2015 syntax here.

I asked that question on irc a while back and got the following answers from @annevk and @gsnedders:

```
Friday, May 19th, 2017
15:45 <tobie> Can I use Map in idlharness?
15:46 <annevk> tobie: sure
15:46 <tobie> annevk: it seems no one is using => functions in there, so…
15:46 <annevk> tobie: old code most likely
15:47 <annevk> tobie: => is fine too
15:47 <tobie> annevk: I'm not sure what the browser support for these things is overall
      to be perfectly honest :)
15:47 <annevk> tobie: Map and => is universal afaik
15:47 <tobie> cool

[…]

17:08 <gsnedders> tobie: the intention has been to keep testharness.js working
      on older browsers than we generally support unless there's a big gain
17:08 tobie: for idlharness I don't think it matters so much given older browsers
      will in general fail tons anyway
17:12 <tobie> Gotcha
17:13 <gsnedders> (IIRC, we actually fixed IE6 support in testharness.js about a
      year ago, because it was like a two-line patch)
```

I don't really care either way, tbh.

resources/idlharness.js
jugglinmike: Ah, right you are. Sorry about that.

jgraham:
So to be clear, the general standard here is "should work in the latest stable releases of major browser engines i.e. Firefox, Safari, Edge, Chrome, prefereably Servo". testharness.js has been held to a slightly higher standard which is something like "no gratuitous breakage in older browsers", although the satndard here is a little fuzzy; something like "don't break the entire harness for trivial reasons and if some feature inevitably introduces breakage try to scope it to tests that actually use that feature".

tobie:
OK, I think this is now ready for review.

resources/idlharness.js
```diff
@@ -322,6 +322,122 @@ IdlArray.prototype.recursively_get_implements = function(interface_name)
...
+                       mixins = mixins.map(id => {
+                           var mixin = this.members[id];
+                           if (!mixin) {
+                               throw new Error("Interface " + id + " not found (implemented by " + thing.name + ")");
+                           }
+                           return mixin;
+                       });
+                       if (mixins.some(m => m.has_to_json_regular_operation())) { return true; }
+                   }
+                   // TODO handle consequential interfaces
```
jugglinmike: Is this something you think you'll implement in this patch?

resources/idlharness.js
```diff
@@ -1551,6 +1795,35 @@ IdlInterface.prototype.add_iterable_members = function(member)
            { name: "thisValue", idlType: {idlType: "any"}, optional: true}]}));
 };
 
+IdlInterface.prototype.test_to_json_operation = function(memberHolderObject, member) {
+    if (member.is_to_json_regular_operation()) {
```
jugglinmike: Given the name of this method, I think it's reasonable to expect the specified member to be a `toJSON` operation. This condition make it a little too lax, since callers will receive no feedback if they provide an inappropriate object. It seems like this condition really belongs at the call site. Would you agree?

resources/idlharness.js
```diff
@@ -1551,6 +1795,35 @@ IdlInterface.prototype.add_iterable_members = function(member)
...
+                }
+            }.bind(this), "Test default toJSON operation of " + this.name);
+        } else {
+            test(function() {
+                assert_true(this.array.is_json_type(member.idlType), JSON.stringify(member.idlType) + " is not an appropriate return value for the toJSON operation of " + this.name);
+                this.array.assert_type_is(memberHolderObject.toJSON(), member.idlType);
+            }.bind(this), "Test toJSON operation of " + this.name);
+        }
+    }
+}
```
jugglinmike: Missing semicolon

jugglinmike (commented):
Hey Tobie, sorry for the delay. Just a few minor things...

resources/idlharness.js
tobie: I did. Forgot to remove the todo. :(

resources/idlharness.js
tobie: Yeah, I (vaguely) considered this at some point, but then dropped the ball. You're right. Will fix.

resources/idlharness.js
```diff
@@ -1551,6 +1795,35 @@ IdlInterface.prototype.add_iterable_members = function(member)
            { name: "thisValue", idlType: {idlType: "any"}, optional: true}]}));
 };
 
+IdlInterface.prototype.test_to_json_operation = function(memberHolderObject, member) {
+    if (member.is_to_json_regular_operation()) {
+        if (member.has_extended_attribute("Default")) {
+            var map = this.default_to_json_operation();
+            test(function() {
+                var json = memberHolderObject.toJSON();
+                [...map.keys()].forEach(function(k) {
```
tobie: Indeed! o_O

resources/idlharness.js
tobie: Only one missing semi? 💃

resources/idlharness.js
```diff
@@ -777,6 +900,124 @@ IdlInterface.prototype.is_global = function()
...
+ *   G - A - H - J
+ *
+ * Invoking traverse_inherited_and_consequential_interfaces() on A
+ * would traverse the tree in the following order:
+ * C -> B -> F -> E -> I -> D -> A -> G -> H -> J
+ */
+
+IdlInterface.prototype.traverse_inherited_and_consequential_interfaces = function(callback) {
+    var stack = this.get_inheritance_stack();
+    _traverse_inherited_and_consequential_interfaces(stack, callback || function() {});
```
tobie: Poor man's `sleep`? Should I add an int so we can specify for how long?

tobie: More seriously, what would you do?

Throw if the arg is missing? Or let if fail in mysterious ways?

tobie:
Thanks for the review, @jugglinmike.

It's relatively sparse so I gather you're either super busy or in strong agreement with the overall direction taken here.

If it's the former, feel free to take a bit more time; I wan't to make sure you're comfortable with the whole thing.

resources/idlharness.js
jugglinmike: I think either is fine since (unlike a volatile assertion) both will only
effect the test author, not consumers. Obviously it's more polite to throw a
descriptive error, but we'd be here all day if we tried to exhaustively
validate every parameter with such assertions. There's a middle ground here,
and I'm happy to let you define it.

jugglinmike:
Thanks! I'm still not really familiar with WebIDL, let alone idlharness.js. From my perspective, if it helps you write tests (and doesn't include any blatant maintainability issues), then that ought to be enough. We can re-factor later if/when the need becomes apparent.

One thing, though: some mention of these new methods probably belongs in [`docs/_writing-tests/idlharness.md`](https://github.com/w3c/web-platform-tests/blob/e8cbae95a91b761a7b0194b3c0d560fd1947a29b/docs/_writing-tests/idlharness.md)

tobie:
Thanks for the review.

> One thing, though: some mention of these new methods probably belongs in docs/_writing-tests/idlharness.md

None of the added methods should affect (or be used by) people _authoring_ tests. So we don't want to document them there.

PR #6262, opened 2017-06-16
[idlharness.js] Expand test coverage

jugglinmike:
A previous patch introduced new tests intended to verify that global
platform objects were also immutable prototype objects. In actuality,
the new tests verified an unrelated condition: that interface prototype
objects were also immutable prototype objects.

Update idlharness to assert that both types of objects are immutable
prototype objects, and modify test expectation strings to correctly
describe each type. Introduce "cleanup" logic to avoid sub-test
interaction.

---

This is a correction/extension of https://github.com/w3c/web-platform-tests/pull/5788

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6262)

### DIFF
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -1137,81 +1137,14 @@ IdlInterface.prototype.test_self = function()
         }
     }.bind(this), this.name + " interface: existence and properties of interface prototype object");
 
-    if (this.is_global() && typeof Object.setPrototypeOf === "function") {
-        // These functions test WebIDL as of 2017-06-06.
-        // https://heycam.github.io/webidl/#platform-object-setprototypeof
-        test(function() {
-            var originalValue = Object.getPrototypeOf(self[this.name].prototype);
-            var newValue = Object.create(null);
-
-            assert_throws(new TypeError(), function() {
-                Object.setPrototypeOf(self[this.name].prototype, newValue);
-            });
-
-            assert_equals(
-                    Object.getPrototypeOf(self[this.name].prototype),
-                    originalValue,
-                    "original value not modified"
-                );
-        }.bind(this), this.name + " interface: internal [[SetPrototypeOf]] method " +
-            "of global platform object - setting to a new value via Object.setPrototypeOf " +
-            "should throw a TypeError");
-
-        test(function() {
-            var originalValue = Object.getPrototypeOf(self[this.name].prototype);
-            var newValue = Object.create(null);
-
-            assert_throws(new TypeError(), function() {
-                self[this.name].prototype.__proto__ = newValue;
-            });
-
-            assert_equals(
-                    Object.getPrototypeOf(self[this.name].prototype),
-                    originalValue,
-                    "original value not modified"
-                );
-        }.bind(this), this.name + " interface: internal [[SetPrototypeOf]] method " +
-            "of global platform object - setting to a new value via __proto__ " +
-            "should throw a TypeError");
-
-        test(function() {
-            var originalValue = Object.getPrototypeOf(self[this.name].prototype);
-            var newValue = Object.create(null);
-
-            assert_false(Reflect.setPrototypeOf(self[this.name].prototype.__proto__, newValue));
-
-            assert_equals(
-                    Object.getPrototypeOf(self[this.name].prototype),
-                    originalValue,
-                    "original value not modified"
-                );
-        }.bind(this), this.name + " interface: internal [[SetPrototypeOf]] method " +
-            "of global platform object - setting to a new value via Reflect.setPrototypeOf " +
-            "should return false");
-
-        test(function() {
-            var originalValue = Object.getPrototypeOf(self[this.name].prototype);
-
-            Object.setPrototypeOf(self[this.name].prototype, originalValue);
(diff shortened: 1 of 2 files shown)


w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/243783720)

# Firefox (nightly) #
Testing web-platform-tests at revision e780cae872580cfdf12cb9de4cb0700abfe8e590
Ignoring 2 changed files:
No files changed

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/243783721)

# Sauce (safari) #
Testing web-platform-tests at revision e780cae872580cfdf12cb9de4cb0700abfe8e590
Ignoring 2 changed files:
No files changed

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/243783727)

# Chrome (unstable) #
Testing web-platform-tests at revision e780cae872580cfdf12cb9de4cb0700abfe8e590
Ignoring 2 changed files:
No files changed

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/243783728)

# Sauce (MicrosoftEdge) #
Testing web-platform-tests at revision e780cae872580cfdf12cb9de4cb0700abfe8e590
Ignoring 2 changed files:
No files changed

domenic (approved):
So this appears to now test

- `window`'s [[Prototype]] cannot change (will always be `Window.prototype`)
- `Window.prototype`'s [[Prototype]] cannot change (will always be window's named properties object)

or in the service worker case:

- `self`'s [[Prototype]] cannot change (will always be `ServiceWorkerGlobalScope.prototype`)
- `ServiceWorkerGlobalScope.prototype`'s [[Prototype]] cannot change (will always be `WorkerGlobalScope.prototype`)

But it is not testing the rest of the objects in the prototype chain, which are tested by https://codereview.chromium.org/2933913002 for the service worker case:

- Window case
   - Named properties object's [[Prototype]] cannot change (will always be `EventTarget.prototype`)
   - `EventTarget.prototype`'s [[Prototype]] cannot change (will always be `Object.prototype`)
   - `Object.prototype`'s [[Prototype]] cannot change (will always be `null`)
- Service worker case
   - `WorkerGlobalScope.prototype`'s [[Prototype]] cannot change (will always be `EventTarget.prototype`)
   - `EventTarget.prototype`'s [[Prototype]] cannot change (will always be `Object.prototype`)
   - `Object.prototype`'s [[Prototype]] cannot change (will always be `null`)


I think this would be better if it crawled the prototype chain, like https://codereview.chromium.org/2933913002 does.

But overall the refactoring LGTM, just not sure it accomplishes everything you're hoping for.

resources/idlharness.js
```diff
@@ -1251,6 +1184,110 @@ IdlInterface.prototype.test_self = function()
 };
 
 //@}
+IdlInterface.prototype.test_immutable_prototype = function(type, obj)
+//@{
+{
+    if (typeof Object.setPrototypeOf !== "function") {
```
tobie: So I know this isn't introduced by this PR, but the refactoring makes it stand out. This creates bad incentives (i.e. you're better off not implementing `Object.setPrototypeOf` than implementing and not some of what follows.

Can we have a strategy to avoid this?

jugglinmike:
@domenic Thanks! You're right that this patch is in service of that Chromium test migration. The prototype traversal something that I don't quite like about that test. It makes the test more of an integration test for multiple web platform features. With the proper support in `idlharness.js`, the test could be re-factored as a series of independent tests--one for each platform object/interface prototype object pairing. If we had coverage in that way, then verification of the complete chain would be provided on a distributed basis by [the current behavior of `idlharness.js`](https://github.com/w3c/web-platform-tests/blob/1e717b0165a6a1ef809de55dc2d16b4cf01eddbc/resources/idlharness.js#L884), and WPT's coverage would be equivalent to that Chromium test. Factored in that way, a Service Worker test wouldn't fail for a bug in EventTarget, but the bug would still be detected.

domenic:
Hmm, I'd prefer we keep at least some integration test of the prototype migration, as it tests the actual intent. If you want to add more tests, that's reasonable, but let's keep the existing one.

resources/idlharness.js
jugglinmike: This is related to [your question in gh-6246](https://github.com/w3c/web-platform-tests/pull/6246#issuecomment-308867046). As I mentioned in IRC, I'm not in a position to make a judgement call here, but I agree that we could use some clarity. I've opened gh-6266 to facilitate a discussion along these lines. Hopefully that results in a formally-documented policy.

tobie: Well, sort of. The broader question here is how do we handle tests we're unable to run because of lack of support for X.

I'm of the opinion we should throw in such cases, not silently avoid running a bunch of tests.

jugglinmike:
@domenic I'll see what I can do in the Chromium patch, but in the mean time, would you mind merging this?

domenic:
Sure. In the future if I give the green checkmark feel free to merge yourself; no need to block on me finding free time.

PR #6279, opened 2017-06-19
Ignore mime type parameters in nosniff

jdm:
This patch implements the following changes:
- Only check for the toplevel/sublevel part of the mime type when
  deciding if it's a js or css resource, ignoring the mime parameters.
- Fix the wpt tests that did not escape url parameters properly and
  also used an invalid syntax for the mime parameter.
- Update the wpt manifest.

Upstreamed from https://github.com/servo/servo/pull/16869 [ci skip]


jdm:
Code reviewed upstream.

PR #6333, opened 2017-06-23
Document expectations for usage of new JS features

### LINKED ISSUE #6266: Formalize policy for usage of new language features
The usage of new ECMAScript features has been a topic of discussion in a few recent pull requests. Currently, there does not seem to be consensus around in what contexts it is appropriate to use such features in WPT. Having clarity here would help everyone involved with this project: contributors (so they could have some confidence that their stylistic choices would be accepted), reviewers (so they wouldn't have to resolve discrepancies between the preferences of different contributors), and consumers (so they would have a realistic expectation about when they might start to fail tests due to orthogonal issues in their JS runtime).

Unfortunately, the desires of those audiences is somewhat at odds. Test contributors will tend to want to use all the tools that are available to them. On the other hand, implementers will generally prefer to avoid practices that could prevent them from executing tests. (For their part, maintainers probably just want clarity--the specific decision doesn't effect them as much.)

This tension might be partially addressed by **a "contextual" policy**. Maybe new ECMAScript features are acceptable in tests written against specifications that are similarly new. Along these lines, new features would also be barred from shared utilities such as those found in the `common/` and `resources/` top-level directories (since those files may be included by any test and would therefore be a lowest-common-denominator). Defining, documenting, and enforcing this kind of policy is a little tricky, though, because it assumes familiarity with the publication history of various specifications.

Under **a "blanket" policy**, WPT would simply declare, "any test may use any language feature that was standardized at least X years prior to the current date." For example, setting up a blanket policy of 1 year would today translate to: "you may use the exponentiation operator, but you may not use async functions." This would be easier to follow in practice, although it might also needlessly shut some consumers out from tests that would otherwise help them ensure conformance.

I would love to get input on this topic from members of all three audiences. My goal is to amend [`docs/_writing-tests/assumptions.md`](https://github.com/w3c/web-platform-tests/blob/master/docs/_writing-tests/assumptions.md) and/or [`docs/_writing-tests/general-guidelines.md`](https://github.com/w3c/web-platform-tests/blob/master/docs/_writing-tests/general-guidelines.md) with a policy by Friday June 23.

jugglinmike:
These expectations were discussed by project maintainers at:

https://github.com/w3c/web-platform-tests/issues/6266

This is intended to resolve gh-6266.

### DIFF
--- a/docs/_writing-tests/general-guidelines.md
+++ b/docs/_writing-tests/general-guidelines.md
@@ -122,6 +122,30 @@ either installed or to have specific metrics. As such, in most cases
 when a known font is needed Ahem should be used. In other cases,
 `@font-face` should be used.
 
+#### Be Conservative
+
+Sometimes, when authoring tests for one web platform feature, it may be
+convenient to use some unrelated feature (e.g. web browser APIs or ECMAScript
+language features). Reliance on such extraneous features tends to make tests
+more difficult to consume, so care should be taken when considering their use.
+Tests may only rely on on language features that are available in the latest
+version of all of the following browsers:
+
+- [Mozilla Firefox][mozilla-firefox]
+- [Google Chrome][google-chrome]
+- [Apple Safari][apple-safari]
+- [Microsoft Edge][microsoft-edge]
+
+For JavaScript code that is re-used across many tests (e.g. `testharness.js`
+and the files located in the directory named `common`), language features that
+were introduced in [ECMAScript 2015][es2015] or later should be avoided. This
+practice avoids introducing spanning test failures for consumers maintaining
+older JavaScript runtimes.
+
+Existing tests can be used as a guide to identify acceptable features.  For
+Community-maintained projects such as [the ECMAScript compatibility
+tables][es-compat] and [caniuse.com][caniuse] provide an overview of basic
+feature support across the browsers listed above.
 
 ### Be Self-Contained
 
@@ -180,7 +204,6 @@ for CSS have some additional requirements for:
 * [Metadata][css-metadata], and
 * [User style sheets][css-user-styles].
 
-
 [server features]: {{ site.baseurl }}{% link _writing-tests/server-features.md %}
 [assumptions]: {{ site.baseurl }}{% link _writing-tests/assumptions.md %}
 [ahem]: {{ site.baseurl }}{% link _writing-tests/ahem.md %}
@@ -188,3 +211,10 @@ for CSS have some additional requirements for:
 [css-metadata]: {{ site.baseurl }}{% link _writing-tests/css-metadata.md %}
 [css-user-styles]: {{ site.baseurl }}{% link _writing-tests/css-user-styles.md %}
 [file-name-flags]: {{ site.baseurl }}{% link _writing-tests/file-names.md %}
+[mozilla-firefox]: https://mozilla.org/firefox
+[google-chrome]: https://google.com/chrome/browser/desktop/
+[apple-safari]: https://apple.com/safari
+[microsoft-edge]: http://microsoftedge.com/
+[es2015]: http://www.ecma-international.org/ecma-262/6.0/
+[es-compat]: https://kangax.github.io/compat-table/
+[caniuse]: https://caniuse.com/
--- a/docs/_writing-tests/general-guidelines.md
+++ b/docs/_writing-tests/general-guidelines.md
@@ -97,7 +97,7 @@ the test (for a typical testharness test, the only content on the page
 will be rendered by the harness itself).
 
 
-### Be Minimal
+### Be Conservative
 
 Tests should generally avoid depending on edge case behavior of
 features that they don't explicitly intend on testing. For example,
@@ -108,6 +108,30 @@ This is not, however, to discourage testing of edge cases or
 interactions between multiple features; such tests are an essential
 part of ensuring interoperability of the web platform.
 
+Tests should pass when the feature under test exposes the expected behavior,
+and they should fail when the feature under test is not implemented or is
+implemented incorrectly. Tests should not rely on unrelated features if doing
+so causes failures in the latest stable release of [Apple
+Safari][apple-safari], [Google Chrome][google-chrome], [Microsoft
+Edge][micosoft-edge], or [Mozilla Firefox][mozilla-firefox]. They should,
+therefore, not rely on any features aside from the one under test unless they
+are supported in all four browsers.
+
+Existing tests can be used as a guide to identify acceptable features. For
+features that are not used in existing tests, community-maintained projects
+such as [the ECMAScript compatibility tables][es-compat] and
+[caniuse.com][caniuse] provide an overview of basic feature support across the
+browsers listed above.
+
+For JavaScript code that is re-used across many tests (e.g. `testharness.js`
+and the files located in the directory named `common`), language features that
+were introduced in [ECMAScript 2015][es2015] or later should be avoided. This
+practice avoids introducing test failures for consumers maintaining older
+JavaScript runtimes.
+
+Patches to make tests run on older versions or other browsers will be accepted
+provided they are relatively simple and do not add undue complexity to the
+test.
 
 ### Be Cross-Platform
 
@@ -121,33 +145,6 @@ Aside from the [Ahem font][ahem], fonts cannot be relied on to be
 either installed or to have specific metrics. As such, in most cases
 when a known font is needed Ahem should be used. In other cases,
 `@font-face` should be used.
-
-#### Be Conservative
-
-Sometimes, when authoring tests for one web platform feature, it may be
-convenient to use some other feature (e.g. a web browser API or ECMAScript
-language feature) that is not directly related to the test itself. Reliance on
-such extraneous features tends to make tests more difficult to consume, so care
-should be taken when considering their use.  Tests may only rely on on language
-features that are available in the latest version of all of the following
-browsers:
-
-- [Mozilla Firefox][mozilla-firefox]
-- [Google Chrome][google-chrome]
-- [Apple Safari][apple-safari]
-- [Microsoft Edge][microsoft-edge]
-
-For JavaScript code that is re-used across many tests (e.g. `testharness.js`
-and the files located in the directory named `common`), language features that
-were introduced in [ECMAScript 2015][es2015] or later should be avoided. This
-practice avoids introducing test failures for consumers maintaining older
-JavaScript runtimes.
-
-Existing tests can be used as a guide to identify acceptable features. For
-features that are not used in existing tests, community-maintained projects
-such as [the ECMAScript compatibility tables][es-compat] and
-[caniuse.com][caniuse] provide an overview of basic feature support across the
-browsers listed above.
 
 ### Be Self-Contained
 
--- a/docs/_writing-tests/general-guidelines.md
+++ b/docs/_writing-tests/general-guidelines.md
@@ -125,11 +125,12 @@ when a known font is needed Ahem should be used. In other cases,
 #### Be Conservative
 
 Sometimes, when authoring tests for one web platform feature, it may be
-convenient to use some unrelated feature (e.g. web browser APIs or ECMAScript
-language features). Reliance on such extraneous features tends to make tests
-more difficult to consume, so care should be taken when considering their use.
-Tests may only rely on on language features that are available in the latest
-version of all of the following browsers:
+convenient to use some other feature (e.g. a web browser API or ECMAScript
+language feature) that is not directly related to the test itself. Reliance on
+such extraneous features tends to make tests more difficult to consume, so care
+should be taken when considering their use.  Tests may only rely on on language
+features that are available in the latest version of all of the following
+browsers:
 
 - [Mozilla Firefox][mozilla-firefox]
 - [Google Chrome][google-chrome]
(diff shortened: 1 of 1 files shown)


tobie:
LGTM

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/246387523)

# Firefox (nightly) #
Testing web-platform-tests at revision 8b77d6bad8fb2e9d906b527a488473cfc6c9c361
Starting 10 test iterations
No tests run.

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/246387524)

# Sauce (safari) #
Testing web-platform-tests at revision 8b77d6bad8fb2e9d906b527a488473cfc6c9c361
Using browser at version 10.0
Starting 10 test iterations

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/246387530)

# Chrome (unstable) #
Testing web-platform-tests at revision 8b77d6bad8fb2e9d906b527a488473cfc6c9c361

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/246387531)

# Sauce (MicrosoftEdge) #
Testing web-platform-tests at revision 8b77d6bad8fb2e9d906b527a488473cfc6c9c361

sideshowbarker (commented):
LGTM but I think @gsnedders should be the one to OK this

gsnedders:
Two thoughts:

 * This should be under "Be cross-platform" given that talks about what can assumed.
 * It should probably be more generic—the policy isn't specific to JS features, but rather features in general.

jugglinmike:
You got it. I moved the text to a new sub-section of "Be Cross Platform" named
"Be Conservative". I also included a reference to https://caniuse.com as this
is more helpful for determining browser support of non-ECMAScript features.

This change called for a modification of the commit message, so I've
force-pushed a new commit to this branch. The original commit is available at:
https://github.com/bocoup/web-platform-tests/tree/document-language-features-orig

tobie (commented):
So I'm not sure how much has change between the two commits, but I'm really uneasy with the tone here. It sounds paternalistic, tbh.

Using modern features is sort of described as being lazy and we should aim to "be conservative" instead. I don't think I can stand behind this depiction.

I think this points to a deeper lack in consensus as to which direction (and underlying goals and purpose) this project should hold.

Are we trying to support IE6, or bleeding edge browsers?

Are we there to help implementors implement or sites like caniuse, test?

This middle-ground leaves me uneasy. We should either embrace modern JS or backwards compat, but not a little bit of both.

tobie:
Note I don't think your effort to write this down is poor or in vain. Quite the contrary, actually. It reflects the lack of consensus around major project goals. It's great we're having this conversation. So thank you!

jgraham:
I think only `testharness.js` (and maybe related things like `idlharness.js`) is really controversial here. What I personally care about for those is the usual criteria (Chrome, Edge Firefox, Safari) + works in Servo. I am prepared to enforce that level of support.

However there have been cases where people have asked for changes to support older browsers, and where they have been small we have usually accepted those changes. So empirically some people care about that, and I don't think it's worth gratuitously breaking them. But I would certainly consider any such uses to be "Tier 2", by which i mean, worth taking patches for and bearing in mind, but not worth blocking features on.

docs/_writing-tests/general-guidelines.md
```diff
@@ -112,6 +112,30 @@ either installed or to have specific metrics. As such, in most cases
...
+
+- [Mozilla Firefox][mozilla-firefox]
+- [Google Chrome][google-chrome]
+- [Apple Safari][apple-safari]
+- [Microsoft Edge][microsoft-edge]
+
+For JavaScript code that is re-used across many tests (e.g. `testharness.js`
+and the files located in the directory named `common`), language features that
+were introduced in [ECMAScript 2015][es2015] or later should be avoided. This
+practice avoids introducing spanning test failures for consumers maintaining
```
jugglinmike: I was trying to communicate the severity of the failures--that they will span
multiple tests across all of WPT. But that detail is not particularly important
for this document, and the statement is no less accurate without the word, so
I'm happy to remove it.

docs/_writing-tests/general-guidelines.md
```diff
@@ -112,6 +112,30 @@ either installed or to have specific metrics. As such, in most cases
 when a known font is needed Ahem should be used. In other cases,
 `@font-face` should be used.
 
+#### Be Conservative
+
+Sometimes, when authoring tests for one web platform feature, it may be
+convenient to use some unrelated feature (e.g. web browser APIs or ECMAScript
```
jugglinmike: That phrase only describes features that would *not* be acceptable according to
the criteria we've identified here. This text needs to apply to *any* feature
since the explanation that follows describes how to distinguish between
"acceptable" and "unacceptable." I chose the words "unrelated" and "extraneous"
to differentiate the features under test (whose validity is unquestionable)
from those that are there to facilitate the test (whose validity is determined
by the criteria documented here).

To try make this more clear, I've rewritten that sentence:

> it may be convenient to use some other feature (e.g. a web browser API or
> ECMAScript language feature) that is not directly related to the test itself

jugglinmike:
> So I'm not sure how much has change between the two commits, but I'm really
> uneasy with the tone here. It sounds paternalistic, tbh.

It sounds like my personal bias leaked into the copy. I think paternalistic is
a somewhat negative interpretation, but I certainly didn't mean to offend.
While I'm happy to substitute in some more neutral words, this is a more
subjective change, and I'd like to avoid potential back-and-forth in satisfying
the expectations of multiple reviewers along these lines. That's why I'm
requesting a second opinion on the tone: @gsnedders, do you agree with
@tobie's suggestion?

tobie:
> I think paternalistic is a somewhat negative interpretation.

Yeah, I struggled to find a more toned down word but couldn't. It was late and I went with that. Sorry. :-/

jugglinmike:
No worries!

docs/_writing-tests/general-guidelines.md
```diff
@@ -112,6 +112,32 @@ either installed or to have specific metrics. As such, in most cases
 when a known font is needed Ahem should be used. In other cases,
 `@font-face` should be used.
 
+#### Be Conservative
+
+Sometimes, when authoring tests for one web platform feature, it may be
```
gsnedders: This seems to be duplicating what is under "Be Minimal" above. (Though "Be Conservative" is probably a better header!) We should probably add a paragraph there rather than this.

My suggestion would be something like:

> Tests should ensure they run correctly and fail when appropriate (including, e.g., when the feature is unsupported) on the latest stable release of Apple Safari, Google Chrome, Microsoft Edge, and Mozilla Firefox. They should, therefore, not rely on any features aside from the one under test unless they are supported in all four browsers.
>
> Patches to make tests run on older versions or other browsers will be accepted provided they are relatively simple and do not add undue complexity to the test.

docs/_writing-tests/general-guidelines.md
```diff
@@ -112,6 +112,32 @@ either installed or to have specific metrics. As such, in most cases
...
+should be taken when considering their use.  Tests may only rely on on language
+features that are available in the latest version of all of the following
+browsers:
+
+- [Mozilla Firefox][mozilla-firefox]
+- [Google Chrome][google-chrome]
+- [Apple Safari][apple-safari]
+- [Microsoft Edge][microsoft-edge]
+
+For JavaScript code that is re-used across many tests (e.g. `testharness.js`
```
gsnedders: Let's just keep the `testharness.js` exception (and I think it is the only exception) documented in the header of it, and not have it here. I think it's sufficiently obscure that it needn't be here.

gsnedders (changes requested):

jugglinmike:
@gsnedders Thanks! I've removed that new section, renamed the existing one, and
added your test. I modified your suggestion a bit by being explicit about "run
correctly and fail when appropriate":

> Tests should pass when the feature under test exposes the expected behavior,
> and they should fail when the feature under test is not implemented or is
> implemented incorrectly.

And as discussed on IRC, I've persisted the references to shared resources and
to the community-run projects that track implementation status.

docs/_writing-tests/general-guidelines.md
```diff
@@ -98,6 +98,30 @@ This is not, however, to discourage testing of edge cases or
 interactions between multiple features; such tests are an essential
 part of ensuring interoperability of the web platform.
 
+Tests should pass when the feature under test exposes the expected behavior,
+and they should fail when the feature under test is not implemented or is
+implemented incorrectly. Tests should not rely on unrelated features if doing
+so causes failures in the latest stable release of [Apple
+Safari][apple-safari], [Google Chrome][google-chrome], [Microsoft
```
tobie: Should we talk about rendering engines, here, instead of browsers?

snuggs:
Fixes #6266

docs/_writing-tests/general-guidelines.md
```diff
@@ -98,6 +98,30 @@ This is not, however, to discourage testing of edge cases or
...
+Edge][micosoft-edge], or [Mozilla Firefox][mozilla-firefox]. They should,
+therefore, not rely on any features aside from the one under test unless they
+are supported in all four browsers.
```
zcorpan: May be worth encouraging that when a contributor runs into an unrelated feature not working as expected in the 4 browsers, that they write a separate test for that, or at least report an issue about it so somebody else can follow up.

docs/_writing-tests/general-guidelines.md
foolip: Rendering engines don't have releases, so that wouldn't quite work.

docs/_writing-tests/general-guidelines.md
```diff
@@ -98,6 +98,30 @@ This is not, however, to discourage testing of edge cases or
...
+
+Existing tests can be used as a guide to identify acceptable features. For
+features that are not used in existing tests, community-maintained projects
+such as [the ECMAScript compatibility tables][es-compat] and
+[caniuse.com][caniuse] provide an overview of basic feature support across the
+browsers listed above.
+
+For JavaScript code that is re-used across many tests (e.g. `testharness.js`
+and the files located in the directory named `common`), language features that
+were introduced in [ECMAScript 2015][es2015] or later should be avoided. This
```
foolip: @jugglinmike, depending on Promises is fine now, but was introduced in ECMAScript 2015.

I suggest dropping this paragraph, and pointing to https://wpt.fyi/results/?label=stable in the previous. No specific guidance we give will stay up to date.

foolip (approved):
LGTM % nits

jugglinmike:
> I suggest dropping this paragraph,

The policy [was](https://github.com/web-platform-tests/wpt/issues/6266#issuecomment-309444814) and [continues to be](https://github.com/web-platform-tests/wpt/pull/12399#issuecomment-412470957) different for shared infrastructure. We can remove this paragraph, but we'll need to replace it with something that documents the distinction.

> and pointing to https://wpt.fyi/results/?label=stable in the previous.

The preceding paragraph describes ECMAScript conformance, but wpt.fyi doesn't help people understand that. To make this more clear, I've changed "For features that are not used in existing tests" to "For language features that are not used in existing tests"

> No specific guidance we give will stay up to date.

Given that WPT's policy is partially based on Servo's implementation status and that [Servo's conformance is not publicly visible](https://github.com/web-platform-tests/wpt/issues/6266#issuecomment-309547561), I think we're forced to maintain a static/"rot-prone" description in this document. I've updated the patch to reference ES2017 in recognition of Servo's lack of support for async functions.

gsnedders:
@jugglinmike I *think* Servo's support should basically match FF50, which es-compat at least used to have… :/

jugglinmike:
Rebased again. The prior version of this branch is available here:

https://github.com/bocoup/wpt/tree/document-language-features-2

gsnedders:
Looking over this again, this looks broadly good to me?

gsnedders (approved):

sideshowbarker:
Can we merge this? (After rebasing and fixing the conflicts.)

docs/_writing-tests/general-guidelines.md
```diff
@@ -108,6 +108,30 @@ This is not, however, to discourage testing of edge cases or
...
+
+Existing tests can be used as a guide to identify acceptable features. For
+language features that are not used in existing tests, community-maintained
+projects such as [the ECMAScript compatibility tables][es-compat] and
+[caniuse.com][caniuse] provide an overview of basic feature support across the
+browsers listed above.
+
+For JavaScript code that is re-used across many tests (e.g. `testharness.js`
+and the files located in the directory named `common`), language features that
+were introduced in [ECMAScript 2017][es2017] or later should be avoided. This
```
stephenmcgruer, line 128: This hasn't aged well, imo. The only thing in es2017 I'm aware of that isn't well supported across all major browsers (outside of IE11) is `SharedArrayBuffer`. More importantly, `async` functions are in es2017 and we certainly have a lot of those!

I think this would need to change to at least 2018 if not 2019, and moreso I'm concerned with us forgetting to update this every few years as features become widely available.

sideshowbarker:
```suggestion
are relatively new and not yet supported in all current browser engines should not be used. This
```

stephenmcgruer:
Again, jugglinmike no longer works in the WPT space afaik, so I've uploaded a new version of this at https://github.com/web-platform-tests/wpt/pull/26204 which incorporates the concern about the es2017 line. Hopefully we can get that merged.

PR #6374, opened 2017-06-28
Correct test.

jgraham:
the durationchange event will be queued during the initialization segment received algorithm.
Only once the init segment has been fully processed will update/updateend be queued.

As such, the updateend event must be fired before being able to call appendBuffer once again.

MozReview-Commit-ID: GYQNOwWZ7DH

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1362165 [ci skip]


jgraham:
Code reviewed upstream.

PR #6379, opened 2017-06-28
Allow running wpt directories by path on the command line,

jgraham:
We regressed running entire directories when wptrunner was switched to
look up command parameters directly in the manifest rather than
iterating over prefix matches. For files that's OK, but for
directories the old behaviour is required.

MozReview-Commit-ID: HVL7rL1YuZx

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1369471 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6379)

### DIFF
--- a/tools/manifest/manifest.py
+++ b/tools/manifest/manifest.py
@@ -51,6 +51,15 @@ class Manifest(object):
             for test in type_tests.get(path, set()):
                 yield test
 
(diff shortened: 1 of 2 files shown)


jgraham:
Code reviewed upstream.

PR #6397, opened 2017-06-28
Don't infinite loop when a test expects to crash,

jgraham:
In order to avoid the leak checker complaining about missing output,
we restart the browser before running tests that are expected to
crash. But as part of that restart we end up checking again if a
restart is required and so end up in an infinite loop. To break out of
that loop we simply check if this is the same test as during the last
iteration, and don't ask to restart in that case.

MozReview-Commit-ID: 90gsmqVCRsD

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1373709 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6397)


jgraham:
Code reviewed upstream.

PR #6403, opened 2017-06-28
Move to RectResponse aligning with the WebDriver crate;

jgraham:
This uses the merged RectResponse from the WebDriver crate. It also moves the
types in the struct to floats which aligns with the WebDriver
specification as in
https://w3c.github.io/webdriver/webdriver-spec.html#get-element-rect
and in https://w3c.github.io/webdriver/webdriver-spec.html#dfn-window-rect

MozReview-Commit-ID: DWeXJEnc1p

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1371405 [ci skip]

### DIFF
--- a/webdriver/contexts.py
+++ b/webdriver/contexts.py
@@ -19,24 +19,6 @@ def window_position_supported(session):
     except webdriver.InvalidArgumentException:
         return True
 
-def test_window_size_types(http, session):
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #6405, opened 2017-06-28
Disable loading mozlog plugin with pytest for wpt,

jgraham:
Loading mozlog from within the executor process makes us prone to
deadlocks since we can't guarantee that the log mutex isn't acquired
when we fork() the parent process, and the python multiprocessing
module doesn't respect posix guidelines about execv()ing a new process
after a fork().

To avoid this specific instance of the probelm we stop pytest loading
the mozlog plugin which we aren't actually using but is loaded by
default when mozlog is available.

MozReview-Commit-ID: IIllNZVOUJz

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1354750 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6405)


jgraham:
Code reviewed upstream.

PR #6411, opened 2017-06-28
Switch wptrunner to use a deque for test groups, r=ato,jdm

jgraham:
Initially wptrunner had a single test queue that was shared between
all processes. Then for --run-by-dir it changed to a queue of
queues. This change makes it a queue of deques, which is simpler,
since the test queues themselves are no longer shared between
processes. It also changes the implementation when we aren't using
--run-by-dir but are using multiple processes to pre-group the tests
into N queues rather than sharing a single queue between all
processes. This is necessary to use the deque of course, but
importantly anticipates a change in which we will pre-compute per
queue metdata for each queue; that doesn't work well with one shared
queue. The downside of this change is that there is no work stealing,
so it may be less efficient if we randomly assign many slow jobs to
one particular process.

MozReview-Commit-ID: 7e0Odk7yDwr

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6411)

### DIFF
--- a/tools/wptrunner/wptrunner/testloader.py
+++ b/tools/wptrunner/wptrunner/testloader.py
@@ -564,74 +564,72 @@ class TestLoader(object):
 class TestSource(object):
     __metaclass__ = ABCMeta
 
-    @abstractmethod
-    def queue_tests(self, test_queue):
-        pass
+    def __init__(self, test_queue):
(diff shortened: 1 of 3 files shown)


jgraham:
Code reviewed upstream.

PR #6412, opened 2017-06-28
Add per-test-queue metadata to wptrunner, r=ato,maja_zf

jgraham:
This adds a metadata object associated with each test queue, and uses
it to pass cache information into the marionette internal reftest
implementation so that we are able to cache only those canvases that
will be reused.

MozReview-Commit-ID: zASrlvnri3

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6412)


jgraham:
Code reviewed upstream.

PR #6414, opened 2017-06-28
Use the internal reftest implementation by default on Linux and OSX,

jgraham:
This is faster and works better in some cases (e.g. with SVG). However
it doesn't work correctly on Windows yet, so remains disabled there by default.

MozReview-Commit-ID: AXyeFUGfVgx

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6414)


jgraham:
Code reviewed upstream.

PR #6468, opened 2017-07-05
Rewrite tests for RTCConfiguration iceServers

soareschen:
The latest editor's draft have removed the `token` `credentialType` since 20170508. Instead a new `oath` `credentialType` is added. As a result, some of the existing tests for RTCPeerConnection-constructor is no longer valid.

I have extracted the existing tests for iceServers from RTCPeerConnection-constructor.html and rewrite them in RTCConfiguration-iceServers.html. The rewrite includes coverage annotation with current spec, and test helper function to test the same configuration with `new RTCPeerConnection()` and `setConfiguration()`.

The initial commit for this PR points to the tip of tree draft. The main difference is that `NotSupportedError` is thrown instead of `SyntaxError` when unknown URL scheme is provided. (w3c/webrtc-pc#1433). To make ease of the transition, I added a commit 0dc620f that revert the test back to the latest editor draft. Hopefully with this approach, this PR can be merged first without issue, and when new editor's draft is published, commit 0dc620f can be reverted so that the tests can be updated without too much effort.

### DIFF
--- a/webrtc/RTCConfiguration-iceServers.html
+++ b/webrtc/RTCConfiguration-iceServers.html
@@ -7,8 +7,8 @@
 <script>
   'use strict';
 
-  // Test is based on the following tip of tree draft (20170630):
-  // https://rawgit.com/w3c/webrtc-pc/3345dadfd5cf9ed177b70b6191b4de55af40f5e6/webrtc.html
+  // Test is based on the following editor's draft:
+  // https://w3c.github.io/webrtc-pc/archives/20170605/webrtc.html
 
   // The following helper function is called from
   // RTCConfiguration-helper.js:
@@ -275,40 +275,15 @@
 
   /*
     4.3.2.  To set a configuration
-      11.3.  For each url in server.urls parse the url using the generic URI syntax defined
-             in [RFC3986] and obtain the scheme name.
-             - If the parsing based on the syntax defined in [RFC3986] fails, throw a SyntaxError.
+      11.3. For each url in server.urls parse url and obtain scheme name.
+        - If the scheme name is not implemented by the browser, throw a SyntaxError.
+        - or if parsing based on the syntax defined in [ RFC7064] and [RFC7065] fails,
+          throw a SyntaxError.
 
-    [RFC3986] Uniform Resource Identifier (URI): Generic Syntax
-    3.  Syntax Components
-      URI         = scheme ":" hier-part [ "?" query ] [ "#" fragment ]
-   */
-  config_test(makePc => {
-    assert_throws('SyntaxError', () =>
-      makePc({ iceServers: [{
-        urls: 'relative-url'
-      }] }));
-  }, 'with relative url should throw SyntaxError');
-
-  /*
-    4.3.2.  To set a configuration
-      11.3.  For each url in server.urls parse the url using the generic URI syntax defined
-             in [RFC3986] and obtain the scheme name.
-             - If the scheme name is not implemented by the browser throw a NotSupportedError.
-   */
-  config_test(makePc => {
-    assert_throws('NotSupportedError', () =>
-      makePc({ iceServers: [{
-        urls: 'http://example.com'
-      }] }));
-  }, 'with http url should throw NotSupportedError');
-
-  /*
-    4.3.2.  To set a configuration
-      11.3.  For each url in server.urls parse the url using the generic URI syntax defined
-             in [RFC3986] and obtain the scheme name.
-              - If scheme name is turn or turns, and parsing the url using the syntax defined
-                in [RFC7064] fails, throw a SyntaxError.
+    [RFC7064] URI Scheme for the Session Traversal Utilities for NAT (STUN) Protocol
+    3.1.  URI Scheme Syntax
+      stunURI       = scheme ":" host [ ":" port ]
+      scheme        = "stun" / "stuns"
 
     [RFC7065] Traversal Using Relays around NAT (TURN) Uniform Resource Identifiers
     3.1.  URI Scheme Syntax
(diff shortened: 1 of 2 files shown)


sideshowbarker:
w3c-test:mirror

webrtc/RTCConfiguration-iceServers.html
```diff
@@ -0,0 +1,570 @@
...
+                 DOMString                          username;
+                 (DOMString or RTCOAuthCredential)  credential;
+                 RTCIceCredentialType               credentialType = "password";
+      };
+   */
+
+  test(() => {
+    const pc = new RTCPeerConnection();
+    assert_equals(pc.getConfiguration().iceServers, undefined);
+  }, 'new RTCPeerConnection() should have default configuration.iceServers of []');
```
rwaldron: This: 

> 'new RTCPeerConnection() should have default configuration.iceServers of []'

Does not reflect the test: 

> `assert_equals(pc.getConfiguration().iceServers, undefined);`

If `iceServers` is an empty array, it would not equal `undefined`

rwaldron: Also consider: https://heycam.github.io/webidl/#es-sequence

webrtc/RTCConfiguration-iceServers.html
```diff
@@ -0,0 +1,570 @@
...
+
+  config_test(makePc => {
+    assert_throws(new TypeError(), () =>
+      makePc({ iceServers: null }));
+  }, '{ iceServers: null } should throw TypeError');
+
+  config_test(makePc => {
+    const pc = makePc({ iceServers: undefined });
+    assert_equals(pc.getConfiguration().iceServers, undefined);
+  }, '{ iceServers: undefined } should succeed');
```
rwaldron: See comment above

webrtc/RTCConfiguration-iceServers.html
```diff
@@ -0,0 +1,570 @@
...
+  // token credentialType was removed from the spec since 20170508
+  config_test(makePc => {
+    assert_throws(new TypeError(), () =>
+      makePc({ iceServers: [{
+        urls: [],
+        credentialType: 'token'
+      }] }));
+  }, 'with credentialType token should throw TypeError');
+
+  // Blink and Gecko fall back to url, but it's not in the spec.
```
rwaldron: Then it's wrong and shouldn't be included as a test of spec implementation

webrtc/RTCConfiguration-iceServers.html
```diff
@@ -0,0 +1,570 @@
...
+        credential: 'cred'
+      }] }));
+  }, 'with turns server and only credential should throw InvalidAccessError');
+
+  /*
+    4.3.2.  To set a configuration
+      11.3. For each url in server.urls parse url and obtain scheme name.
+        - If the scheme name is not implemented by the browser, throw a SyntaxError.
+        - or if parsing based on the syntax defined in [ RFC7064] and [RFC7065] fails,
+          throw a SyntaxError.
```
rwaldron: Unrelated to these tests: this is another misuse of SyntaxError by this spec :(

webrtc/RTCConfiguration-iceServers.html
```diff
@@ -0,0 +1,570 @@
...
+      }] }));
+  }, 'with turns server and only username should throw InvalidAccessError');
+
+  config_test(makePc => {
+    assert_throws('InvalidAccessError', () =>
+      makePc({ iceServers: [{
+        urls: 'turns:turn.example.net',
+        credential: 'cred'
+      }] }));
+  }, 'with turns server and only credential should throw InvalidAccessError');
```
rwaldron: Thorough!

webrtc/RTCConfiguration-iceServers.html
```diff
@@ -0,0 +1,570 @@
...
+    assert_equals(iceServers.length, 1);
+
+    const server = iceServers[0];
+    assert_array_equals(server.urls, ['stun:stun1.example.net']);
+    assert_equals(server.credentialType, 'password');
+    assert_equals(server.username, 'user');
+
+    const { credential } = server;
+    assert_equals(credential.macKey, '');
+    assert_equals(credential.accessToken, '');
```
rwaldron: Not one single browser has implemented this correctly :(

webrtc/RTCConfiguration-iceServers.html
```diff
@@ -0,0 +1,570 @@
...
+    Untestable
+      4.3.2.  To set a configuration
+        11.7.  Append server to validatedServers.
+
+    Coverage Report
+      Tested         9
+      Not Tested     0
+      Untestable     1
+      Total         10
+   */
```
rwaldron: Confirmed. I've updated several sections of the coverage spec.

rwaldron (changes requested):
This appears to have conflicts, please rebase

webrtc/RTCConfiguration-iceServers.html
soareschen: Oh yeah, I think I was checking whether an optional sequence field in dictionary has default value of `[]`. It is hard to search in WebIDL but I believe the default is still `undefined`. So the description should be updated to reflect the test.

webrtc/RTCConfiguration-iceServers.html
soareschen: This has been updated in webrtc-pc master branch to throw `NotSupportedError` instead. Unfortunately the editor's draft have not been updated so I put this in a separate commit that is revertable once the next editor's draft is published.

webrtc/RTCConfiguration-iceServers.html
soareschen: This is one of the original tests written by @foolip. I guess it might be legacy features in the spec that was later removed. Since the test cases still fail in current browsers I decided to keep them during refactoring.

webrtc/RTCConfiguration-iceServers.html
rwaldron: Oh, that's great news.

rwaldron (approved):

webrtc/RTCConfiguration-iceServers.html
rwaldron: If there is no spec for the behavior, then there shouldn't be any tests. Legacy or not. I will file an issue to remove it.

foolip: I added this comment in 089c33e7267854a391cfd8ce38929c52425f1252 to explain the `'{ iceServers: [{ url: "stun:stun1.example.net" }] }': new TypeError,` line that followed but which isn't in this file.

It's an historical test, to make sure that some test is failing as long as a formerly-standard thing hasn't been removed from implementations. These are often in historical.html tests. I'm adding one for WebRTC in https://github.com/w3c/web-platform-tests/pull/6686, once that's landed the test for non-support of `url` could be moved there.

PR #6470, opened 2017-07-05
Data URL <iframe>s are cross-origin

annevk:
This modifies an XMLHttpRequest test so it runs in more browsers and
removes two DOM tests that can no longer be tested as data URLs are now
considered cross-origin, including per standard.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1377557.


jdm:
Couldn't the data URLs include a `<script>` that used postMessage to report their contentType?

annevk:
I suppose we could do that for the second example. I don't see how you'd do it for the first.

jdm:
Yes, I was specifically looking at contenttype_datauri_01.html and contenttype_datauri_02.html when I suggested that.

annevk:
I meant 01 when I said I don't see how that'd work. But I'll do it for 02.

jdm (approved):

annevk:
Since we're revamping this code in Firefox we might want to land this on our side so I'd appreciate if folks would not land this for now.

sideshowbarker:
w3c-test:mirror

annevk:
Closing as this will get upstreamed from Mozilla's side eventually.

PR #6477, opened 2017-07-06
Add test for RTCPeerConnection.getDefaultIceServers

soareschen:
This does simple test that the `getDefaultIceServers()` static method exists, and that it return result of the right type. Since idlharness do not support automatic testing of dictionary objects, the test does manual type checking of the dictionary fields.

### DIFF
--- a/webrtc/RTCPeerConnection-getDefaultIceServers.html
+++ b/webrtc/RTCPeerConnection-getDefaultIceServers.html
@@ -1,0 +1,99 @@
+<!doctype html>
+<meta charset=utf-8>
+<title>RTCPeerConnection.getDefaultIceServers</title>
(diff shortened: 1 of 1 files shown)


webrtc/RTCPeerConnection-getDefaultIceServers.html
```diff
@@ -0,0 +1,99 @@
...
+
+  }, 'RTCPeerConnection.getDefaultIceServers() should return array of RTCIceServer');
+
+  /*
+    Coverage Report
+      Since there is no steps involved and we are only checking basic call,
+      This is counted as 1 trivial test coverage.
+
+      Tested    1
+      Total     1
```
rwaldron: 👍

rwaldron (approved):

PR #6478, opened 2017-07-06
Add new test for RTCConfiguration bundlePolicy

soareschen:
This moves the constructor test for `bundlePolicy` out from RTCPeerConnection-constructor.html to a new file RTCConfiguration-bundlePolicy.html. It also tests on `getConfiguration()` and `setConfiguration()` and that `bundlePolicy` cannot be changed via `setConfiguration()`.


rwaldron (approved):
Thorough!

PR #6484, opened 2017-07-06
Aborting fetch

jakearchibald:
For https://github.com/whatwg/fetch/pull/523

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6484)


fetch/api/abort/general.js
```diff
@@ -0,0 +1,288 @@
...
+    method: 'POST',
+    body: 'foo',
+    headers: { 'Content-Type': 'text/plain' }
+  });
+
+  await fetch(request).catch(() => {});
+
+  assert_false(request.bodyUsed, "Body has not been used");
+  assert_equals(await request.text(), 'foo', "Correct body");
+}, "Request is not 'used' if signal is aborted before fetching");
```
jakearchibald: @annevk I felt like the above was the expected behaviour, but seems like it'll be a pain to spec. What do you think?

fetch/api/abort/general.js
```diff
@@ -0,0 +1,288 @@
...
+  const controller = new AbortController();
+  const signal = controller.signal;
+  const key = token();
+  controller.abort();
+
+  await fetch(`../resources/infinite-slow-response.py?key=${key}`, { signal }).catch(() => {});
+  
+  // I'm hoping this will give the browser enough time to (incorrectly) make the request
+  // above, if it intends to.
+  await fetch('../resources/data.json').then(r => r.json());
```
jakearchibald: @wanderview Any better ideas than the hack above?

tools/wptserve/wptserve/response.py
```diff
@@ -465,9 +465,11 @@ def encode(self, data):
...
-        """Flush the output."""
+        """Flush the output. Returns False if the flush failed due to
+        the socket being closed by the remote end."""
         try:
             self._wfile.flush()
+            return True
         except socket.error:
             # This can happen if the socket got closed by the remote end
-            pass
+            return False
```
jakearchibald: @annevk any idea who owns this?

fetch/api/abort/general.js
wanderview: Not really.  Doing another operation that takes the same path is what we do other places in wpt to approximate this.

tools/wptserve/wptserve/response.py
annevk: @jgraham should review that.

fetch/api/abort/general.js
annevk: So the idea is that if the signal is aborted it's a no-op when passed to `fetch()`? It seems like that should be fairly easy to specify (just don't pass it along), but I'm not sure it's the model we want.

jakearchibald: The signal is used, but the request isn't, as in the body isn't read.

My thinking is that the abort should happen straight away if the signal is already aborted. No request should be made to the server. Therefore, `request.body` shouldn't be consumed/locked/disturbed.

tools/wptserve/wptserve/response.py
jakearchibald: @jgraham I'm using this to stop an infinite response, but also to track when/if a browser closes the connection related to an aborted fetch.

fetch/api/abort/general.js
annevk: Oh, I missed the subtle distinction. I guess the `Request` constructor (which is what `fetch()` uses, can just bypass a whole bunch of things when passed an aborted signal?

jakearchibald: This is where it gets messy I think. Here are my gut feelings:

```js
const request = new Request(url, { signal: abortedSignal, body });
const anotherRequest = new Request(request);
```

It feels like `request`'s body should be disturbed/locked, since the body is being explicitly moved from one request to another.

```js
fetch(invalidURL, { signal: abortedSignal });
```

The above should reject based on the invalid URL, not aborted.

```js
fetch(request, { signal: abortedSignal, body });
```

In the above, the body shouldn't be consumed, as it doesn't need to be sent to the server, as the signal is already aborted.

So, from the developer's point of view, the order is:

1. Validate the args.
1. Abort if already aborted.
1. Make the request.

Locking/disturbing the body as part of "Validate the args" doesn't fit well with me, although I realise that's how it's specced now.

I guess fetch could undisturb/unlock the body if it's already aborted?

annevk: So you basically want `fetch()` to do something else than invoke the constructor. We could do that, but it'll be tricky to ensure it's correct and has no regressions. (I don't like undisturbing/unlocking. I'd rather not do the "damage" to begin with.)

jakearchibald: Given that mixed-content and CSP could be considered aborting early before using the body (yet they disturb/lock the body), I think I'm over complicating it by trying to make aborting different.

Happy to remove this test unless you disagree.

annevk: Yeah, it does seem like a lot of refactoring effort for unclear benefit. If we had other features that needed that kind of refactoring it might be worth looking into, but not sure.

fetch/api/abort/general.js
```diff
@@ -0,0 +1,335 @@
...
+    assert_abort_error(err);
+  });
+}, "Aborting rejects with AbortError");
+
+test(() => {
+  // TODO: we may want to discuss this design idea
+  const request = new Request('');
+  assert_true(Boolean(request.signal), "Signal member is present & truthy");
+  assert_equals(request.signal.constructor, AbortSignal);
+}, "Request objects have a signal property");
```
jakearchibald: @annevk what do you think about this? I'm undecided.

The idea is that requests will *always* have a signal. If you pass a signal to the fetch constructor, it'll listen to it & repeat the same action on its own signal.

I figured this would make things easier when it came to replacing this with `fetchSignal` later, as you don't have to worry that `request.signal` is one of a variety of types (or null).

annevk: So instead of forwarding the signal it gets copied? It seems you might still get different types or shapes over time as we add more features, but copying rather than forwarding might be nice if you want to reuse the request... It's not clear if we should take care of that or whether there should be a utility to copy signals.

jakearchibald: This is what I was thinking: https://github.com/whatwg/fetch/pull/523/files#diff-e8c7e042697e1e48e4466c612dda6a3bR5006

Then, we can later change all `request.signal`s to `FetchSignal`s, and developers can rely on behaviour like "If `self.FetchSignal` is supported, all requests will have one as `.signal`".

fetch/api/abort/cache.https.html
```diff
@@ -0,0 +1,60 @@
...
+    const requests = await cache.keys();
+
+    assert_equals(requests.length, 1, 'Ensuring cleanup worked');
+
+    const [cachedRequest] = requests;
+
+    await new Promise(resolve => {
+      signal.addEventListener('abort', resolve);
+      controller.abort();
+    });
```
annevk: What's the point of the promise here?

fetch/api/abort/general.js
```diff
@@ -0,0 +1,335 @@
...
+promise_test(async () => {
+  const controller = new AbortController();
+  const signal = controller.signal;
+  controller.abort();
+
+  const request = new Request('../resources/data.json', { signal });
+
+  // TODO: we may want to discuss this design idea
+  assert_true(Boolean(request.signal), "Signal member is present & truthy");
+  assert_equals(request.signal.constructor, AbortSignal);
```
annevk: I hope folks won't do these kind of checks when branching...

fetch/api/abort/general.js
```diff
@@ -0,0 +1,335 @@
...
+// otherwise you can end up running out of HTTP connections.
+let requestKeys = [];
+
+function abortRequests() {
+  const keys = requestKeys;
+  requestKeys = [];
+  return Promise.all(
+    keys.map(key => fetch(`../resources/stash-put.py?key=${key}&value=close`))
+  );
+}
```
annevk: Move requestKeys and abortRequests done to where they are actually used? I guess I can kinda see they are global, but to me at least I kept wondering when they'd show up and when they did I had forgotten the setup.

fetch/api/abort/general.js
```diff
@@ -0,0 +1,470 @@
...
+    while (true) {
+      // Stop spinning if 10 seconds have passed
+      if (Date.now() - start > 10000) throw Error('Timed out');
+
+      const afterAbortResult = await fetch(`../resources/stash-take.py?key=${stateKey}`).then(r => r.json());
+      if (afterAbortResult == 'closed') break;
+    }
+  }, `Fetch aborted & connection closed when aborted after calling response.${bodyMethod}()`);
+}
+
```
jakearchibald: @domenic I don't suppose you have time to check these aborting-streams tests?

fetch/api/abort/cache.https.html
jakearchibald: I thought it queued a task, but it doesn't. Will fix.

fetch/api/abort/general.js
jakearchibald: I could change it to `instanceof`?

annevk: It's fine as-is, I doubt anyone would copy from the tests.

annevk: Though maybe `instanceof` is better so we don't have to update this tests and it'll keep passing in newer UAs.

jakearchibald: Won't we update it to `instanceof FetchSignal` later?

annevk: Yeah, I guess we might do that anyway. Okay, never mind.

jakearchibald:
I've written all the tests I intended to write.

fetch/api/resources/basic.html
```diff
@@ -0,0 +1 @@
+<!DOCTYPE html>
```
domenic: You should be able to use /common/blank.html for this

fetch/api/abort/serviceworker-intercepted.https.html
```diff
@@ -0,0 +1,111 @@
...
+<head>
+  <meta charset="utf-8">
+  <title>Aborting fetch when intercepted by a service worker</title>
+  <script src="/resources/testharness.js"></script>
+  <script src="/resources/testharnessreport.js"></script>
+  <script src="../../../service-workers/service-worker/resources/test-helpers.sub.js"></script>
+</head>
+<body>
+<script>
+  function assert_abort_error(err) {
```
domenic: Use `promise_rejects(t, "AbortError", promise)` instead of this function.

fetch/api/abort/serviceworker-intercepted.https.html
```diff
@@ -0,0 +1,111 @@
...
+          }
+        ),
+        Promise.resolve().then(() => log.push('next-microtask'))
+      ]);
+
+      assert_array_equals(log, [`${bodyMethod}-reject`, 'next-microtask']);
+    }, `response.${bodyMethod}() rejects if already aborted`);
+  }
+
+  promise_test(async t => {
```
domenic: Apart from the promise_rejects issue, this looks good. You may also want to test reader.closed.

domenic (commented):
In general the same comments everywhere: use promise_rejects instead of your custom assertion function, and consider also testing .closed.

jakearchibald:
This is a high-level overview of the intent behind the controller, signal, & fetch abort design.

# What does "aborting a fetch" entail?

[Fetching](https://fetch.spec.whatwg.org/#main-fetch) can involve setting up connections, making an HTTP request, receiving headers, and reading a reaponse body. A single fetch can involve many iterations of these, through preflights and redirects. The aim is to make the process abortable during all of these phases. The spec should be explicit about how to react to an abort signal at various points during the process, including which points cannot be aborted.

# Abort controllers and abort signals

```js
const controller = new AbortController();
const signal = controller.signal;

const fetchPromise = fetch(url, {signal});
```

The [feedback we got](https://github.com/whatwg/fetch/issues/27) from developers was to avoid the `fetchPromise` being externally abortable – it should be possible to hand someone the deferred result of the fetch without giving them the ability to abort.

As a result, we designed `AbortController` and `AbortSignal` to decouple receiving the result, controlling the result, and signalling intent to the result.

```js
signal.addEventListener('abort', () => {
  console.log('Abort signalled!');
});

controller.abort();
```

The controller gives you the ability to say "I would like to signal 'abort'". The signal broadcasts that intent.

A single signal can be passed to many APIs – a bunch of fetches can be aborted by a single signal.

The signal is just a signal, it doesn't reflect whether the anything was actually aborted. For instance, you can signal abort long after the fetch has completed and been fully read. In this case the signal will be in the aborted state, but the fetch will have successfully completed.

A signal may be `postMessage`d to another thread. It achieves this by being cloned, but the clone listens to the original's signals, and mimics them. Note that this is a clone, not a transfer, the original signal continues to operate.

The design here is deliberately generic, as streams and [webauthn](https://github.com/w3c/webauthn/issues/380) want this ability. In future, `AbortController` and `AbortSignal` may be subclassed to provide additional signals. For example, fetch may gain the ability to signal priority changes.

[`AbortController` and `AbortSignal` spec](https://dom.spec.whatwg.org/#aborting-ongoing-activities).

# `request.signal`

```js
const request = new Request(url, {signal});
console.log(request.signal);
```

All requests have a signal property whether one is provided in the constructor or not. When a signal is given to the constructor, `request.signal` listens to its signals, and mimics them. This means `request.signal` never equals the signal passed into the constructor.

This means, in future, we can change `request.signal` to a subclass such as `FetchSignal`, without `request.signal` being multiple types within a single user agent.

# What does aborting look like to the developer?

If the abort occurs before response headers have been received, `fetch()`'s promise will reject with a `DOMException` named `AbortError`.

If the abort occurs when reading the body of the response, the promise (in the case of `response.text()` etc) will reject or the stream (in the case of `response.body`) will error, in either case with a `DOMException` named `AbortError`.

Aborting can be the result of the developer signalling, but also something lower-level such as the browser no longer wanting to finish the request (eg the user hit the "stop" button).

Aborting should be predictable. If aborting was signalled in the same thread as the fetch promise, the promise should reject immediately, as in rejection callbacks being called in the next microtask. Likewise, the pending/next reads of the body stream should error.

If the abort was signalled in another thread, the rejection should happen as soon as the message can be sent to the other thread.

If the browser has received the whole response, but it hasn't been read, unread data is dropped when abort is signalled, [due to how streams work](https://streams.spec.whatwg.org/#readable-stream-default-controller-error). Methods like `response.text()` should follow this behaviour and reject.

# What does abort look like to the server?

The actual aborting of network actions happens independently of rejecting promises and erroring streams. The browser should close the connection (HTTP/1.1) or reset the stream (HTTP/2).

If aborting has been signalled before calling `fetch()`, no network activity should take place as a result.

# How does this work within a service worker's fetch event?

`fetchEvent.request.signal` will signal the page's desire to abort the fetch. This may come from calling `controller.abort()` in the page, or the browser wanting to abort the request for some other reason. For example: if the page closes and the browser no longer needs the response, or the user presses the browser's "stop" button.

This means:

```js
addEventListener('fetch', event => {
  event.respondWith(fetch(event.request));
});
```

…works for free. However, if the developer wants to rewrite the request (or turn it into multiple requests), they'll need to pass the signal manually:

```js
addEventListener('fetch', event => {
  event.respondWith(
    fetch(somethingElse, { signal: event.request.signal })
  );
});
```

If a fetch is aborted within a page, but the service worker ignores that signal, the fetch should still abort from the perspective of the page.

tools/wptserve/wptserve/response.py
jakearchibald: @jgraham I want to use this in other tests too, eg to stop generating content when the client is no longer interested.

fetch/api/resources/basic.html
jakearchibald: That means I'd have to install a service worker at that scope. Is that likely to interfere with other tests?

annevk: I don't think tests run concurrently so if you clean up that could work, but it might be clearer to just add a comment as to why you duplicated and keep the scope as-is.

fetch/api/abort/serviceworker-intercepted.https.html
jakearchibald: Fixed the `promise_rejects` stuff & added `reader.closed` tests.

fetch/api/abort/general.js
```diff
@@ -0,0 +1,465 @@
...
+    headers: {
+      'Content-Type': 'text/plain',
+      'Content-Length': '2097152'
+    }
+  });
+
+  assert_true(!!cancelReason, 'Cancel called sync');
+  assert_equals(err.constructor, DOMException);
+  assert_equals(err.name, 'AbortError');
+}, "Readable stream synchronously cancels with AbortError if aborted before reading");
```
jakearchibald: @domenic does the test above make sense?

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/258540644)

# Firefox (nightly) #

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/258540645)

# Sauce (safari) #

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/258540651)

# Chrome (unstable) #

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/258540653)

# Sauce (MicrosoftEdge) #

jakearchibald:
I can't really figure out what Travis is complaining about.

annevk:
It's some intermittent issue with the Firefox bot (that also seems to be affecting the Chrome bot, but the Chrome bot doesn't count as blocking since it does that a lot). However, I can't really decipher the log either. I asked @jgraham to weigh in on IRC.

jgraham:
w3c-test:mirror

jgraham:
idk why the travis failure isn't being reported here correctly, but @bobholt is working on improved infrastructure there.

Anyway the problem in this case is a legit bug in the tests that needs to be fixed. In the log there are lots of lines at the end like

```
ERROR:/home/travis/build/w3c/web-platform-tests/tools/ci/check_stability:| `/fetch/api/abort/general.html` | `Signal can be used to abort other fetches, even if another fetch succeeded before aborting`             | **FAIL: 30/10** | `promise_test: Unhandled rejection with value: object "ReferenceError: AbortController is not defined"` |
```

The important point there is `FAIL: 30/10` meaning we did 10 iterations, but got 30 failures for tests that are indistinguishable to the infrastructure i.e. each test ran three times. That appears to be because `/fetch/api/abort/general.html` is running the same tests in Window, Worker and SharedWorker context, without distinguishing which is which.

jakearchibald:
Thanks! Fixed now.

tools/wptserve/wptserve/response.py
jgraham: I think this is fine.

fetch/api/abort/general.js
```diff
@@ -0,0 +1,469 @@
...
+
+  fetch('../resources/empty.txt', {
+    body, signal,
+    method: 'POST',
+    headers: {
+      'Content-Type': 'text/plain'
+    }
+  });
+
+  assert_true(!!cancelReason, 'Cancel called sync');
```
domenic: This test makes sense. Can we not assert anything more specific about cancelReason than that it's truthy?

fetch/api/abort/general.html
```diff
@@ -0,0 +1,19 @@
...
+  <script src="/resources/testharnessreport.js"></script>
+  <script src="/common/utils.js"></script>
+  <script src="general.js"></script>
+</head>
+<body>
+<script>
+  fetch_tests_from_worker(new Worker("general.js"));
+  if (window.SharedWorker) {
+    fetch_tests_from_worker(new SharedWorker("general.js"));
+  }
```
domenic: FWIW I found generating multiple tests with a wrapper simpler than contextualTestName: see https://github.com/w3c/web-platform-tests/tree/master/streams readme and https://github.com/w3c/web-platform-tests/blob/master/streams/generate-test-wrappers.js

domenic (commented):
The portions I have been asked to comment on look good :)

fetch/api/abort/general.js
jakearchibald: Doh, that's supposed to be the next two lines. Copy & paste error.

fetch/api/resources/sw-intercept.js
```diff
@@ -0,0 +1,10 @@
+async function broadcast(msg) {
+  for (const client of await clients.matchAll()) {
+    client.postMessage(msg);
+  }
+}
+
+addEventListener('fetch', event => {
+  event.waitUntil(broadcast(event.request.url));
+  event.respondWith(fetch(event.request));
+});
```
annevk: Add a newline here?

fetch/api/resources/basic.html
annevk: R+ with added newline and the comment.

fetch/api/abort/serviceworker-intercepted.https.html
```diff
@@ -0,0 +1,102 @@
...
+    const reader = response.body.getReader();
+
+    controller.abort();
+
+    await promise_rejects(t, "AbortError", reader.read());
+    await promise_rejects(t, "AbortError", reader.closed);
+  }, "Stream errors once aborted.");
+</script>
+</body>
+</html>
```
annevk: Add newline.

fetch/api/abort/general.html
```diff
@@ -0,0 +1,19 @@
...
+</script>
+</body>
+</html>
```
annevk: Add newline.

fetch/api/abort/cache.https.html
```diff
@@ -0,0 +1,57 @@
...
+    const [cachedRequest] = requests;
+
+    assert_false(cachedRequest.signal.aborted, "Request from cache shouldn't be aborted");
+
+    const data = await fetch(cachedRequest).then(r => r.json());
+    assert_equals(data.key, 'value', 'Fetch fully completes');
+  }, "Signals are not stored in the cache API, even if they're already aborted");
+</script>
+</body>
+</html>
```
annevk: Add newline.

fetch/api/abort/general-serviceworker.https.html
```diff
@@ -0,0 +1,23 @@
...
+    let reg = await navigator.serviceWorker.getRegistration(scope);
+    if (reg) await reg.unregister();
+
+    reg = await navigator.serviceWorker.register('general.js', {scope});
+
+    fetch_tests_from_worker(reg.installing);
+  })();
+</script>
+</body>
+</html>
```
annevk: Add newline.

annevk (approved):
Couple nits, but I think this is good to go. I'm personally happy with this landing before the change to the Fetch Standard, but I'd like to hear from @wanderview and @youennf as well. We'll start implementing this a month from now or so, so there's no particular rush.

jakearchibald:
Nits addressed. Thanks for the reviews @annevk @domenic @jgraham!

jakearchibald:
https://bugs.chromium.org/p/chromium/issues/detail?id=750599

fetch/api/resources/basic.html
```diff
@@ -1 +1,5 @@
-<!DOCTYPE html>
\ No newline at end of file
+<!DOCTYPE html>
+<!--
+  Duplicating this resource to make service worker scoping simpler in
```
annevk: "Duplicating /common/blank.html..."

annevk:
Bug for Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1378342.

jakearchibald:
Bug for WebKit: https://bugs.webkit.org/show_bug.cgi?id=174980
Bug for Edge: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/13009916/

royling:
>In future, AbortController and AbortSignal may be subclassed to provide additional signals. For example, fetch may gain the ability to signal priority changes.

if that's the intent, `SignalController` and `Signal` may be more generic to subclass than `Abort*`.

jakearchibald:
We don't have use cases for that right now, but we can add it later if need be.

PR #6486, opened 2017-07-06
wai-aria: Correct aria-haspopup assertions for macOS

joanmarie:
AXPress is independent of aria-haspopup. The Core AAM has been updated
accordingly. This brings the tests in alignment with the spec.


halindrome (approved):
Nice catch!

PR #6499, opened 2017-07-07
Initial Merge of WritableStreams Tests w/Upstream W3C

Brandr0id:
Initial merge of writablestream tests into upstream W3C version.
Contains some formatting and minor functionality fixes (timing issues in
a small subset of tests).

### DIFF
--- a/streams/writable-streams/general.js
+++ b/streams/writable-streams/general.js
@@ -112,55 +112,45 @@ promise_test(() => {
 }, 'closed and ready on a released writer');
 
 promise_test(() => {
-  const promises = {};
-  const resolvers = {};
-  for (const methodName of ['start', 'write', 'close', 'abort']) {
-    promises[methodName] = new Promise(resolve => {
-      resolvers[methodName] = resolve;
-    });
-  }
-
+  var thisObject = null;
   // Calls to Sink methods after the first are implicitly ignored. Only the first value that is passed to the resolver
   // is used.
   class Sink {
     start() {
       // Called twice
-      resolvers.start(this);
+      assert_equals(thisObject, this, 'start should be called as a method');
     }
 
     write() {
-      resolvers.write(this);
+      assert_equals(thisObject, this, 'write should be called as a method');
     }
 
     close() {
-      resolvers.close(this);
+      assert_equals(thisObject, this, 'close should be called as a method');
     }
 
     abort() {
-      resolvers.abort(this);
+      assert_equals(thisObject, this, 'abort should be called as a method');
     }
   }
 
   const theSink = new Sink();
+  thisObject = theSink;
   const ws = new WritableStream(theSink);
 
   const writer = ws.getWriter();
 
   writer.write('a');
-  writer.close();
+  var closePromise = writer.close();
 
   const ws2 = new WritableStream(theSink);
   const writer2 = ws2.getWriter();
-  writer2.abort();
+  var abortPromise = writer2.abort();
 
-  return promises.start
-      .then(thisValue => assert_equals(thisValue, theSink, 'start should be called as a method'))
-      .then(() => promises.write)
-      .then(thisValue => assert_equals(thisValue, theSink, 'write should be called as a method'))
-      .then(() => promises.close)
-      .then(thisValue => assert_equals(thisValue, theSink, 'close should be called as a method'))
-      .then(() => promises.abort)
-      .then(thisValue => assert_equals(thisValue, theSink, 'abort should be called as a method'));
+  return Promise.all([
+    closePromise,
+    abortPromise
+  ]);
 }, 'WritableStream should call underlying sink methods as methods');
 
 promise_test(t => {
--- a/streams/writable-streams/general.js
+++ b/streams/writable-streams/general.js
@@ -111,26 +111,34 @@ promise_test(() => {
   );
 }, 'closed and ready on a released writer');
 
-promise_test(() => {
-  var thisObject = null;
+promise_test(t => {
+  let thisObject = null;
   // Calls to Sink methods after the first are implicitly ignored. Only the first value that is passed to the resolver
   // is used.
   class Sink {
     start() {
       // Called twice
-      assert_equals(thisObject, this, 'start should be called as a method');
+      t.step(() => {
+        assert_equals(this, thisObject, 'start should be called as a method');
+      });
     }
 
     write() {
-      assert_equals(thisObject, this, 'write should be called as a method');
+      t.step(() => {
+        assert_equals(this, thisObject, 'write should be called as a method');
+      });
     }
 
     close() {
-      assert_equals(thisObject, this, 'close should be called as a method');
+      t.step(() => {
+        assert_equals(this, thisObject, 'close should be called as a method');
+      });
     }
 
     abort() {
-      assert_equals(thisObject, this, 'abort should be called as a method');
+      t.step(() => {
+        assert_equals(this, thisObject, 'abort should be called as a method');
+      });
     }
   }
 
@@ -141,11 +149,11 @@ promise_test(() => {
   const writer = ws.getWriter();
 
   writer.write('a');
-  var closePromise = writer.close();
+  const closePromise = writer.close();
 
   const ws2 = new WritableStream(theSink);
   const writer2 = ws2.getWriter();
-  var abortPromise = writer2.abort();
+  const abortPromise = writer2.abort();
 
   return Promise.all([
     closePromise,
--- a/streams/writable-streams/start.js
+++ b/streams/writable-streams/start.js
@@ -154,7 +154,7 @@ promise_test(t => {
     start() {
       return Promise.reject();
     }
-  }, new CountQueuingStrategy({ highWaterMark: 0 }));
+  }, { highWaterMark: 0 });
   const writer = ws.getWriter();
   catchAndRecord(writer.ready, 'ready');
   catchAndRecord(writer.closed, 'closed');


sideshowbarker:
w3c-test:mirror

Brandr0id:
@annevk , @domenic could you please take a look @ the test updates? There shouldn't be too much functionally different but I'd like to make sure the style updates work well for this repro as well so we can better keep these tests in-sync as they are updated.

domenic:
Please don't change the formatting of the HTML wrapper files. They are auto-generated, and the changes introduced are unnecessary.

Happy to re-review once those are reverted.

domenic:
Actually, please do not change any formatting at all (such as whitespace, which I see has changed dramatically in the .js file). That just makes Git history useless and doesn't help anything.

Brandr0id:
Happy to revert the wrapper files for this change, would you be receptive to a separate PR that updates generate-test-wrappers.js and re-generates the wrapper files? 

For the other formatting changes this was a direct merge of our forked version that contains updated formatting (consistent spacing/line breaks etc). Since the functional updates are fairly small I should be able to pull them out pretty easy (It'll make the PR a lot smaller 👍 ). While I don't want to pollute the commit history with formatting changes I would like to get a feel for what, if any, updates you'd be receptive to (e.g. 2sp => 4 sp, function calls either consistently on one line, or all params on a new line, etc).

domenic:
I'd rather not change the wrapper files; they are nice and minimal right now, and I don't see the motivation.

Similarly, I don't understand the motivation for any of the formatting changes. Are you updating other web platform test directories with 4 spaces as well? Without a better understanding of what value they bring, I can't really say if any of them would make sense.

annevk:
Yeah, I think two spaces in JavaScript (and one or two spaces in markup) is what we somewhat consistently use. It seems fine for new resources to deviate from that, but I'd rather not reformat just for the sake of it. If we do that we should start with some kind of style guide (though that has its own problems, you might be testing the indentation doesn't change meaning or some such).

Brandr0id:
Alright thanks for the feedback, I've pushed a new version w/o the majority of the formatting diffs (whitespace/wrappers).

tyoshino:
We had 120 col limit on the size of each line when we were developing the tests in the Streams Standard repo. That's another point @domenic may want to say something though as we've moved to wpt, it might be acceptable.

streams/writable-streams/aborting.js
```diff
@@ -60,50 +70,46 @@ promise_test(t => {
...
+  }).then(() => {
+    assert_array_equals(ws.events, ['abort', undefined]);
+  });
 }, 'Aborting a WritableStream immediately prevents future writes');
 
 promise_test(t => {
   const ws = recordingWritableStream();
   const results = [];
 
-  return delay(0)
```
domenic: Can you undo all these formatting changes?

streams/writable-streams/aborting.js
```diff
@@ -13,8 +13,16 @@ const error2 = new Error('error2');
 error2.name = 'error2';
 
 promise_test(t => {
+  let startResolve;
```
domenic: What is the motivation for this change?

streams/writable-streams/aborting.js
```diff
@@ -60,50 +70,46 @@ promise_test(t => {
 promise_test(t => {
   const ws = recordingWritableStream();
 
-  return delay(0)
```
domenic: Can you undo all these formatting changes?

streams/writable-streams/aborting.js
```diff
@@ -39,7 +50,6 @@ promise_test(t => {
   writer.write('a');
 
   const readyPromise = writer.ready;
-
```
domenic: Can you leave this blank line here?

streams/writable-streams/aborting.js
```diff
@@ -127,8 +133,8 @@ promise_test(t => {
   });
   const writer = ws.getWriter();
 
-  return promise_rejects(t, error1, writer.abort(undefined),
-    'rejection reason of abortPromise must be the error thrown by abort');
+  var errorToValidate = error1;
+  return promise_rejects(t, errorToValidate, writer.abort(undefined), 'rejection reason of abortPromise must be the error thrown by abort');
```
domenic: Can you undo this change? It makes the test less clear. Same below.

streams/writable-streams/aborting.js
```diff
@@ -157,16 +163,15 @@ promise_test(t => {
...
-  writer.write().catch(() => {});
+  writer.write(1).catch(() => {});
   return flushAsyncEvents().then(() => {
     const abortPromise = writer.abort(undefined);
 
     resolveWritePromise();
-    return promise_rejects(t, error1, abortPromise,
-      'rejection reason of abortPromise must be the error thrown by abort');
+
+    return promise_rejects(t, error1, abortPromise, 'rejection reason of abortPromise must be the error thrown by abort');
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -157,16 +163,15 @@ promise_test(t => {
...
   });
-}, 'WritableStream if sink\'s abort throws, for an abort performed during a write, the promise returned by ' +
-   'ws.abort() rejects');
+}, 'WritableStream if sink\'s abort throws, for an abort performed during a write, the promise returned by ws.abort() rejects');
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -157,16 +163,15 @@ promise_test(t => {
 
   const writer = ws.getWriter();
 
-  writer.write().catch(() => {});
+  writer.write(1).catch(() => {});
```
domenic: Can you undo this change? It's not relevant to what's being tested.

streams/writable-streams/aborting.js
```diff
@@ -204,34 +209,64 @@ promise_test(t => {
 }, 'Aborting a WritableStream puts it in an errored state, with a TypeError as the stored error');
 
 promise_test(t => {
-  const ws = new WritableStream();
+  let resolveWritePromise;
```
domenic: What is the motivation for this change? It seems like a new test; perhaps we can leave the old one alone, and add a second one?

streams/writable-streams/aborting.js
```diff
@@ -204,34 +209,64 @@ promise_test(t => {
...
+      resolveWritePromise();
 
-  return writePromise;
+      return writePromise;
+  });
 }, 'Aborting a WritableStream causes any outstanding write() promises to be rejected with a TypeError');
 
 promise_test(t => {
-  const ws = recordingWritableStream();
+  let resolveWritePromise;
```
domenic: Again, the old test was fine. Please add a new test if you feel there's additional coverage to be gained.

You mention race conditions, but per spec there are no race conditions here; the order is well-defined.

streams/writable-streams/aborting.js
```diff
@@ -204,34 +209,64 @@ promise_test(t => {
...
+          promise_rejects(t, new TypeError(), closePromise, 'close() should reject with a TypeError'),
+          abortPromise
+      ]).then(() => {
+          assert_array_equals(ws.events, ['write', 'a', 'abort', error1]);
+      });
   });
 }, 'Closing but then immediately aborting a WritableStream causes the stream to error');
 
-promise_test(() => {
+promise_test(t => {
```
domenic: Can you undo the addition of this unnecessary variable?

streams/writable-streams/aborting.js
```diff
@@ -332,15 +367,15 @@ promise_test(t => {
     const writePromise = writer.write('a');
     const abortPromise = writer.abort(error2);
     let closedRejected = false;
+
```
domenic: Can you undo this whitespace change and the below formatting changes?

streams/writable-streams/aborting.js
```diff
@@ -349,16 +384,14 @@ promise_test(t => {
     write() {
       return flushAsyncEvents();
     }
-  }, new CountQueuingStrategy(4));
+  }, { highwatermark: 4 });
```
domenic: This is incorrect. Please leave it as CountQueuingStrategy(4). highwatermark is not a valid value (must be highWaterMark). Here and elsewhere.

streams/writable-streams/aborting.js
```diff
@@ -349,16 +384,14 @@ promise_test(t => {
...
   const writer = ws.getWriter();
   return writer.ready.then(() => {
     const settlementOrder = [];
     return Promise.all([
       writer.write('1').then(() => settlementOrder.push(1)),
-      promise_rejects(t, new TypeError(), writer.write('2'), 'first queued write should be rejected')
```
domenic: Can you undo these formatting changes?

streams/writable-streams/aborting.js
```diff
@@ -369,18 +402,17 @@ promise_test(t => {
...
       return Promise.reject(error1);
     }
-  }, new CountQueuingStrategy(4));
+  }, { highwatermark: 4 });
   const writer = ws.getWriter();
+  var errorToValidate = error1;
   return writer.ready.then(() => {
     const settlementOrder = [];
     return Promise.all([
-      promise_rejects(t, error1, writer.write('1'), 'in-flight write should be rejected')
```
domenic: Can you undo these formatting changes?

streams/writable-streams/aborting.js
```diff
@@ -392,12 +424,13 @@ promise_test(t => {
...
   return writer.ready.then(() => {
     return Promise.all([
-      promise_rejects(t, error1, writer.write('a'), 'writer.write() should reject with error from underlying write()'),
-      promise_rejects(t, new TypeError(), writer.close(),
-                      'writer.close() should reject with error from underlying write()'),
-      writer.abort()
+    promise_rejects(t, error1, writer.write('a'), 'writer.write() should reject with error from underlying write()'),
+    promise_rejects(t, new TypeError(), writer.close(), 'writer.close() should reject with error from underlying write()'),
+    // Write() rejecting on our UnderlyingSink should error the stream, so abort should be rejected when we try to call it.
+    promise_rejects(t, error1, writer.abort(), 'writer.abort() should reject with error from underlying write()')
```
domenic: I don't believe this is a correct change per spec. Perhaps tracing through the reference implementation would help us understand it better.

streams/writable-streams/aborting.js
```diff
@@ -478,8 +511,7 @@ promise_test(t => {
       });
     });
   });
-}, 'if underlying close() has started and then rejects, the abort() and close() promises should reject with the ' +
-   'underlying close rejection reason');
+}, 'if underlying close() has started and then rejects, the abort() and close() promises should reject with the underlying close rejection reason');
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -684,48 +716,37 @@ promise_test(t => {
     const writePromise3 = writer.write('a');
 
     return Promise.all([
-      promise_rejects(t, new TypeError(), writePromise3,
-                      'writePromise3 must reject with an error indicating abort'),
-      promise_rejects(t, new TypeError(), writer.ready,
-                      'writer.ready must be still rejected with the error indicating abort'),
+      promise_rejects(t, new TypeError(), writePromise3, 'writePromise3 must reject with the error passed to the controller\'s error method'),
+      promise_rejects(t, new TypeError(), writer.ready, 'writer.ready must be still rejected with the error indicating abort'),
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -684,48 +716,37 @@ promise_test(t => {
...
       flushAsyncEvents()
     ]);
   }).then(() => {
-    assert_array_equals(
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -684,48 +716,37 @@ promise_test(t => {
...
-        'writePromise, abortPromise and writer.closed must not be fulfilled/rejected yet even after ' +
-            'controller.error() call');
+    assert_array_equals(events, [], 'writePromise, abortPromise and writer.closed must not be fulfilled/rejected yet even after controller.error() call');
 
     resolveWrite();
 
     return Promise.all([
       writePromise,
       abortPromise,
-      promise_rejects(t, new TypeError(), writer.closed,
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -684,48 +716,37 @@ promise_test(t => {
...
-                      'writer.closed must reject with an error indicating abort'),
+      promise_rejects(t, new TypeError(), writer.closed, 'writer.closed must reject with the error passed to the controller\'s error method'),
       flushAsyncEvents()
     ]);
   }).then(() => {
-    assert_array_equals(events, ['writePromise', 'abortPromise', 'closed'],
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -684,48 +716,37 @@ promise_test(t => {
...
-                        'writePromise, abortPromise and writer.closed must settle');
+    assert_array_equals(events, ['writePromise', 'abortPromise', 'closed'], 'writePromise, abortPromise and writer.closed must reject');
 
     const writePromise4 = writer.write('a');
 
     return Promise.all([
       writePromise,
-      promise_rejects(t, new TypeError(), writePromise4,
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -684,48 +716,37 @@ promise_test(t => {
...
-      promise_rejects(t, new TypeError(), writer.ready,
-                      'writer.ready must be still rejected with the error indicating abort')
+      promise_rejects(t, new TypeError(), writePromise4, 'writePromise4 must reject with the error passed to the controller\'s error method'),
+      promise_rejects(t, new TypeError(), writer.ready, 'writer.ready must be still rejected with the error indicating abort')
     ]);
   }).then(() => {
     writer.releaseLock();
 
     return Promise.all([
-      promise_rejects(t, new TypeError(), writer.ready,
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -768,8 +789,7 @@ promise_test(t => {
     });
 
     return Promise.all([
-      promise_rejects(t, new TypeError(), writer.close(),
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -779,18 +799,12 @@ promise_test(t => {
     controller.error(error2);
 
     return Promise.all([
-      promise_rejects(t, new TypeError(), writer.close(),
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -779,18 +799,12 @@ promise_test(t => {
...
-        'writer.close() must reject with an error indicating already closing'),
-      promise_rejects(t, new TypeError(), writer.ready,
-                      'writer.ready must be still rejected with the error indicating abort'),
+      promise_rejects(t, new TypeError(), writer.close(), 'writer.close() must reject with an error indicating already closing'),
+      promise_rejects(t, new TypeError(), writer.ready, 'writer.ready must be still rejected with the error indicating abort'),
       flushAsyncEvents()
     ]);
   }).then(() => {
-    assert_array_equals(
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -800,25 +814,19 @@ promise_test(t => {
       flushAsyncEvents()
     ]);
   }).then(() => {
-    assert_array_equals(events, ['closePromise', 'abortPromise', 'closed'],
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -800,25 +814,19 @@ promise_test(t => {
...
-                        'closedPromise, abortPromise and writer.closed must fulfill');
-
+    assert_array_equals(events, ['closePromise', 'abortPromise', 'closed'], 'closedPromise, abortPromise and writer.closed must fulfill');
     return Promise.all([
       promise_rejects(t, new TypeError(), writer.close(),
-        'writer.close() must reject with an error indicating already closing'),
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -800,25 +814,19 @@ promise_test(t => {
...
-      promise_rejects(t, new TypeError(), writer.ready,
-                      'writer.ready must be still rejected with the error indicating abort')
+      'writer.close() must reject with an error indicating already closing'),
+      promise_rejects(t, new TypeError(), writer.ready, 'writer.ready must be still rejected with the error indicating abort')
     ]);
   }).then(() => {
     writer.releaseLock();
 
     return Promise.all([
-      promise_rejects(t, new TypeError(), writer.close(),
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -963,35 +971,28 @@ promise_test(t => {
     });
 
     return Promise.all([
-      promise_rejects(t, error2, writer.ready,
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -963,35 +971,28 @@ promise_test(t => {
...
-                      'writer.ready must reject with the error passed to the controller\'s error method'),
+      promise_rejects(t, error2, writer.ready, 'writer.ready must reject with the error passed to the controller\'s error method'),
       flushAsyncEvents()
     ]);
   }).then(() => {
-    assert_array_equals(
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -963,35 +971,28 @@ promise_test(t => {
...
-        events, [],
-        'closePromise and writer.closed must not be fulfilled/rejected yet even after writer.abort()');
+    assert_array_equals( events, [], 'closePromise and writer.closed must not be fulfilled/rejected yet even after writer.abort()');
 
     resolveClose();
 
     return Promise.all([
       closePromise,
-      promise_rejects(t, error2, writer.ready,
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -963,35 +971,28 @@ promise_test(t => {
...
-                      'writer.ready must be still rejected with the error passed to the controller\'s error method'),
+      promise_rejects(t, error2, writer.ready, 'writer.ready must be still rejected with the error passed to the controller\'s error method'),
       writer.closed,
       flushAsyncEvents()
     ]);
   }).then(() => {
-    assert_array_equals(events, ['closePromise', 'abortPromise', 'closed'],
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -963,35 +971,28 @@ promise_test(t => {
...
-                        'abortPromise, closePromise and writer.closed must fulfill/reject');
+    assert_array_equals(events, ['closePromise', 'abortPromise', 'closed'], 'abortPromise, closePromise and writer.closed must fulfill/reject');
   }).then(() => {
     writer.releaseLock();
 
     return Promise.all([
-      promise_rejects(t, new TypeError(), writer.ready,
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -1091,9 +1092,7 @@ promise_test(() => {
   const abortPromise = ws.abort('done');
   controller.error(error1);
   resolveStart();
-  return abortPromise.then(() =>
```
domenic: Can you undo this formatting change?

streams/writable-streams/aborting.js
```diff
@@ -1119,7 +1118,14 @@ promise_test(t => {
 
 promise_test(t => {
   const events = [];
-  const ws = recordingWritableStream();
+  let startResolve;
```
domenic: It looks like this is a different test. The old one was correct; can you add a new one that tests the new code path you're working through, instead of modifying the existing one?

domenic (changes requested):
I made it through one file, but there are still too many formatting changes for me to usefully review this.

It may be best to start from scratch, instead of trying to upstream from your copy.

streams/writable-streams/aborting.js
Brandr0id: It looks like this test has been working fine for Chrome's impl due to the requirement of async events needing to be manually processed during the test (e.g. calls to flushAsyncEvents()). Since flushAsyncEvents() isn't called in this test, the stream is constructed but never actually finishes calling start so start is still pending due to this requirement. However it doesn't look like there's a spec'd reason why start couldn't/shouldn't start immediately when constructed.

I found in our implementation that this was happening and so the test wasn't behaving as intended. By ensuring that start doesn't resolve until the test intends (regardless of async event processing) this test now passes in both impls.

streams/writable-streams/aborting.js
Brandr0id: This potentially could be a bug in our implementation. However it looks to be unspec'd behavior? Calling write with no chunk throws so I substituted an arbitrary value for chunk here (otherwise what are you writing?). Please advise if calling write sans chunk is desired behavior.

streams/writable-streams/aborting.js
Brandr0id: Similar to the first test this appears to have been working due to async events not being processed so the stream isn't fully starting, or calling write in response to the writer's write being called. This ends up queuing a write (waiting for the start) and the test ends up testing what it was trying to. However it doesn't look like anything (other than implementation details) should actually prevent the stream from constructing, write from being called (and completing), then abort being called (but all the previous calls completed before abort happened).

I guess this could also be achieved by having start() return a promise rather than seed the test with an initial write, but it should be testing the same behavior just in a more deterministic manner and not relying on events not yet being processed by the browser.

streams/writable-streams/aborting.js
Brandr0id: I believe the race is more in the implementation details, similar to a couple of the other changes above, it only works when async events are queued and not yet processed forcing the intended ordering. However these events could be processed earlier causing a different outcome (since say the write is no longer in-flight or queued).

streams/writable-streams/aborting.js
Brandr0id: same issue as the other tests, start is just returning a promise to complete in a deterministic manner.

streams/writable-streams/aborting.js
ricea: I'm leaving the main review to @domenic. I'm just adding a couple of drive-by comments to reduce latency before he can get back.

This is a little bit subtle, but the step "Upon fulfillment of startPromise," makes the following steps be executed asynchronously, even if start() itself returned synchronously.

Aside: almost all WritableStream tests were written against the reference implementation, not Chrome's implementation. There's a few that I initially wrote against Chrome's implementation, but I always verified against the reference implementation before upstreaming.

streams/writable-streams/aborting.js
ricea: We treat `write()` as `write(undefined)`. `undefined` is a valid chunk.

I think you might be right that this is unspec'd. The "in something close to the syntax of [ECMASCRIPT]," sections are non-normative, and the steps don't say what happens if arguments are omitted. @domenic is this covered somewhere, or do we need to add an explicit normative note?

domenic: In ECMAScript, calling write() and doing write(undefined) are the same, so it's not possible for us to spec something that handles write() differently. This is already in the spec.

streams/writable-streams/aborting.js
domenic: The way in which tasks are processed by the browser is deterministically governed by web specifications. If your browser has a bug in processing tasks, then yeah, it can fail a lot of the tests. But we're not going to modify the tests for that purpose.

domenic:
(This still has too many formatting changes for me to review usefully, by the way. I won't bother commenting on each individual one again.)

domenic:
It seems so far in fact all these changes are invalid:

- They are working around bugs in Edge's event loop implementation where it does not resolve promises in the same order they were enqueued (i.e. Edge apparently processes promises nondeterministically).
- One stems from misunderstanding write() vs. write(undefined) in JavaScript.
- The rest are formatting changes.

At least, in what I've reviewed so far. As such, I'm inclined to close this without merging... What do you think, @Brandr0id?

ricea:
Supplemental note: The strictness about Promise resolution order in the tests is intentional, since it is an observable side-effect of the order of steps in the standard.

Ideally I would like all user-visible behaviour to be the same in every implementation, so that developers never get any nasty surprises. I don't want to accept changes that make the tests less strict. But I don't think failing some of the strictest expectations should block shipping.

Brandr0id:
Thanks @ricea and @domenic , I could squash the two commits to make it easier to see the lack of formatting but I think it probably isn't needed at this point. I tested some changes locally that ensure we always resolve a promise for the UnderlyingSink methods, along with updates due to a spec change (adding the erroring state) and I think the rest of the aborting test changes likely aren't needed now.

Could you please advise on two further items though:
1) It looks like all the tests are validating that the closed promise is a new TypeError() when rejected, but it looks like potentially it should be the storedError?

2) In the constructor set of tests the typeof 'WritableStreamDefaultWriter' and 'WritableStreamDefaultController' are tested to be undefined. Even though the constructors are not exposed, when defined in WebIdl, these are exposed as functions. Are there any issues with updating this to validate the function type?

domenic:
> I could squash the two commits to make it easier to see the lack of formatting but I think it probably isn't needed at this point

I'm actaully reviewing via https://github.com/w3c/web-platform-tests/pull/6499/files, so I can see them. This does seem very much cleaned up now, so thanks for that. It looks like there are still some good things to test here, e.g. removing the CountQueuingStrategy dependency in start.js, and adding the `this` tests in general.js. So I take back my desire to close this without merging.

But yeah, let's get things straightened out first. Great to hear that always-resolving fixed most issues; I look forward to a push without those changes.

> It looks like all the tests are validating that the closed promise is a new TypeError() when rejected, but it looks like potentially it should be the storedError?

This did change somewhat recently, and is still kind of confusing, so I'm not sure I can give the right answer off the top of my head. (Maybe @ricea can.) Can you give a small example we can walk through together?

> In the constructor set of tests the typeof 'WritableStreamDefaultWriter' and 'WritableStreamDefaultController' are tested to be undefined. Even though the constructors are not exposed, when defined in WebIdl, these are exposed as functions. Are there any issues with updating this to validate the function type?

Per https://streams.spec.whatwg.org/#globals, these are not exposed. This is the second bit of feedback we've gotten about this section of the spec being too hidden though. You may enjoy:

- https://github.com/whatwg/streams/issues/586
- https://github.com/whatwg/streams/issues/732
- https://github.com/whatwg/streams/issues/45

as you can see this issue of Web IDL vs. ES-style specification has been a bit problematic for us. It's good to get feedback that Edge is implementing with Web IDL; that helps shape the decision there.

For now though, we should probably not expose those classes, and keep testing that we don't expose them.

Brandr0id:
One of the existing test cases is a good example for the closed promise:
`promise_test(t => {
  const ws = new WritableStream({
    write() {
      return Promise.reject(error1);
    }
  });
  const writer = ws.getWriter();
  return writer.ready.then(() => {
    const writePromise = writer.write('a');
    const abortPromise = writer.abort(error2);
    let closedRejected = false;
    return Promise.all([
      promise_rejects(t, error1, writePromise, 'write() should reject')
          .then(() => assert_false(closedRejected, '.closed should not resolve before write()')),
      promise_rejects(t, new TypeError(), writer.closed, '.closed should reject')
          .then(() => {
            closedRejected = true;
          }),
      abortPromise
    ]);
  });
}, '.closed should not resolve before rejected write(); write() error should not overwrite abort() error');`

It seems like when closeandclosed promises are rejected with the storedError, it should be populated by the rejection of the write callback and be rejected with error1 not a new TypeError. FWIW I do sort of prefer it rejecting with a new TypeError in this case, but would like to understand why it's not using error1.

domenic:
So in that test case, the relevant happenings go something like this:

- writer.write('a')
  - WritableStreamDefaultWriterWrite
    - WritableStreamDefaultControllerWrite
      - WritableStreamDefaultControllerAdvanceQueueIfNeeded
        - WritableStreamDefaultControllerProcessWrite
          - Let sinkWritePromise = PromiseInvokeOrNoop(underlyingSink, "write", « 'a', controller »)
          - Further processing happens in response to sinkWritePromise
- writer.abort(error2)
  - WritableStreamDefaultWriterAbort
    - WritableStreamAbort
      - State is "writable"; the above writer.write() did not get to the point where it changes the state.
      - WritableStreamStartErroring(stream, new TypeError)
        - Sets [[storedError]] to the TypeError


In other words, even though the write() call "happened first", the rejection with error1 isn't seen until after the abort() call already transitioned the state to errored with the aborting TypeError.

Make sense?

streams/writable-streams/general.js
```diff
@@ -119,48 +119,48 @@ promise_test(() => {
...
     });
   }
-
+  var thisObject = null;
   // Calls to Sink methods after the first are implicitly ignored. Only the first value that is passed to the resolver
   // is used.
   class Sink {
     start() {
-      // Called twice
+      assert_equals(thisObject, this, 'start should be called as a method');
```
ricea: asserts don't work properly in sink methods because the exceptions they throw are caught by the WritableStream. This is a serious trap: we used to have lots of tests that looked fine but weren't actually testing anything.

I mostly fixed these by moving the asserts out of the methods, but another way that works is to replace `start() { assert_something(); }` with `start: t.step_func(() => { assert_something(); })`.

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/258688308)

# Firefox (nightly) #

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/258688309)

# Sauce (safari) #

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/258688317)

# Chrome (unstable) #

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/258688318)

# Sauce (MicrosoftEdge) #

domenic:
Sorry for the continued delay on this @Brandr0id! I finally got around to it and suggest the following diff:

```diff
diff --git a/streams/writable-streams/general.js b/streams/writable-streams/general.js
index 0f4dae9f1a..74929ea745 100644
--- a/streams/writable-streams/general.js
+++ b/streams/writable-streams/general.js
@@ -111,26 +111,34 @@ promise_test(() => {
   );
 }, 'closed and ready on a released writer');
 
-promise_test(() => {
-  var thisObject = null;
+promise_test(t => {
+  let thisObject = null;
   // Calls to Sink methods after the first are implicitly ignored. Only the first value that is passed to the resolver
   // is used.
   class Sink {
     start() {
       // Called twice
-      assert_equals(thisObject, this, 'start should be called as a method');
+      t.step(() => {
+        assert_equals(this, thisObject, 'start should be called as a method');
+      });
     }
 
     write() {
-      assert_equals(thisObject, this, 'write should be called as a method');
+      t.step(() => {
+        assert_equals(this, thisObject, 'write should be called as a method');
+      });
     }
 
     close() {
-      assert_equals(thisObject, this, 'close should be called as a method');
+      t.step(() => {
+        assert_equals(this, thisObject, 'close should be called as a method');
+      });
     }
 
     abort() {
-      assert_equals(thisObject, this, 'abort should be called as a method');
+      t.step(() => {
+        assert_equals(this, thisObject, 'abort should be called as a method');
+      });
     }
   }
 
@@ -141,11 +149,11 @@ promise_test(() => {
   const writer = ws.getWriter();
 
   writer.write('a');
-  var closePromise = writer.close();
+  const closePromise = writer.close();
 
   const ws2 = new WritableStream(theSink);
   const writer2 = ws2.getWriter();
-  var abortPromise = writer2.abort();
+  const abortPromise = writer2.abort();
 
   return Promise.all([
     closePromise,
```

Then we can merge. Alternately you can click "allow edits from maintainers" and I can merge this myself.

@ricea, if you have time an extra review (with my above patch) would be appreciated.

ricea:
With @domenic's change, lgtm.

ricea:
@domenic I merged your patch and tested against the reference implementation. lgtm.

domenic (approved):

PR #6505, opened 2017-07-10
Initial tests for WebDriver's Get Element Text command

shs96c:
A couple of simple tests for `Get Element Text`, one
checking the error condition, the other making sure
that a simple string is returned.

Making sure that the approach taken meets with the
expectations of the other people working on the suite.

### DIFF
--- a/webdriver/state/text/get_text.py
+++ b/webdriver/state/text/get_text.py
@@ -14,7 +14,7 @@ def test_getting_text_of_a_non_existant_element_is_an_error(new_session):
        "session/%s/element/%s/text" % (session.session_id, id))
 
    assert_error(result, "no such element")
-    
+
 
 def test_read_element_text(new_session):
     _, session = new_session({})
--- a/webdriver/state/text/get_text.py
+++ b/webdriver/state/text/get_text.py
@@ -4,8 +4,10 @@ import uuid
 from support.asserts import assert_error, assert_success
 from support.inline import inline
 
-def test_getting_text_of_a_non_existant_element_is_an_error(new_session):
-   _, session = new_session({})
+# For failing tests, the Get Element Text end-point is used
+# directly. In all other cases, the Element.text() function is used.
+
+def test_getting_text_of_a_non_existant_element_is_an_error(session):
    session.url = inline("""<body>Hello world</body>""")
    id = uuid.uuid4()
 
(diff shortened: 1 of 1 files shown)


shs96c:
Open question: since these tests are effectively stateless, should I be creating a new session per test, or re-using one across all tests?

webdriver/state/text/get_text.py
```diff
@@ -0,0 +1,28 @@
...
+from support.asserts import assert_error, assert_success
+from support.inline import inline
+
+def test_getting_text_of_a_non_existant_element_is_an_error(session):
+   session.url = inline("""<body>Hello world</body>""")
+   id = uuid.uuid4()
+
+   result = session.transport.send(
+       "GET",
+       "session/%s/element/%s/text" % (session.session_id, id))
```
jgraham: This is fine, but would like to see a PR to add this endpoint to the webdriver client in the repo as well, so that it isn't necessary for all PRs to use the endpoint directly.

webdriver/state/text/get_text.py
```diff
@@ -0,0 +1,28 @@
...
+import uuid
+
+from support.asserts import assert_error, assert_success
+from support.inline import inline
+
+def test_getting_text_of_a_non_existant_element_is_an_error(session):
+   session.url = inline("""<body>Hello world</body>""")
+   id = uuid.uuid4()
+
+   result = session.transport.send(
```
jgraham: This could be `send_session_command`, right?

shs96c: `send_session_command` assumes the session succeeds, and fails the test if it doesn't. In this one case, we want to check the actual result that comes back.

webdriver/state/text/get_text.py
shs96c: It's already present as `Element.text()`. I thought we'd agreed at the F2F sessions to use the URLs when testing a particular function. Would be more than happy to use it here.

sideshowbarker:
w3c-test:mirror

webdriver/state/text/get_text.py
jgraham: I'm not sure I agree with that, although I see the logic. And I don't think there's much precedent for actaully doing that. But I guess it's fine. A comment would be nice.

webdriver/state/text/get_text.py
jgraham: OK, I guess this is clearer than using `pytest.raises` and assuming that the error text is correct in the client.

jgraham (approved):

jgraham (approved):

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/259825691)

# Firefox (nightly) #

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/259825692)

# Sauce (safari) #

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/259825700)

# Chrome (unstable) #

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/259825702)

# Sauce (MicrosoftEdge) #

AutomatedTester:
Just an FYI, I have most of the rest of the element state tests in https://bugzilla.mozilla.org/show_bug.cgi?id=1376625

AutomatedTester:
Just an FYI, I have most of the rest of the element state tests in https://bugzilla.mozilla.org/show_bug.cgi?id=1376625

PR #6514, opened 2017-07-11
Add addTrack-based test for removeTrack

### LINKED ISSUE #6453: removeTrack test should not rely on Transceivers
@soareschen It's great to see modern tests for `addTrack`/`removeTrack`!

Even better would be them following the [evolution](https://blog.mozilla.org/webrtc/the-evolution-of-webrtc/) that justifies their existence.

Specifically, I don't think we should require `addTransceiver` (a stage 3 api) to test `removeTrack` (a stage 2 api), when we can use `addTrack` (a stage 2 api) which returns a sender. IMHO we should use the same abstraction level.

E.g. Firefox has supported `removeTrack` for [almost 3 years](https://hg.mozilla.org/mozilla-central/rev/79525105ef9a), yet won't pass any of the `removeTrack` tests [as written](https://github.com/w3c/web-platform-tests/pull/6251/files) because they require `addTransceiver`.

soareschen:
Fixes #6453.

This adds addTrack-based tests for the removeTrack test cases that does not involve transceivers. It is basically duplicating existing tests that change the method of adding tracks to use `getUserMedia()` and `addTrack()`.

### DIFF
--- a/webrtc/RTCPeerConnection-removeTrack.html
+++ b/webrtc/RTCPeerConnection-removeTrack.html
@@ -24,6 +24,13 @@
       };
    */
 
+  // Before calling removeTrack can be tested, one needs to add MediaStreamTracks to
+  // a peer connection. There are two ways for adding MediaStreamTrack: addTrack and
+  // addTransceiver. addTransceiver is a newer API while addTrack has been implemented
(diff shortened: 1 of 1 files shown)


rwaldron:
Awesome :D

rwaldron (approved):

PR #6515, opened 2017-07-11
Add a test to verify that allow-top-navigation-by-user-activation is …

fred-wang:
…ignored when allow-top-navigation is present.

This is similar to https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/http/tests/security/frameNavigation/sandbox-ALLOWED-top-navigation-with-two-flags.html but navigation is performed without user gesture (so allow-top-navigation is needed and allow-top-navigation-by-user-activation is not enough).


mikewest (approved):
LGTM! Thanks.

PR #6531, opened 2017-07-12
Check default media type support with MediaSource.isTypeSupported()

naeloued:
Use MediaSource.isTypeSupported() instead of HTMLMediaElement.canPlayType() to avoid
failure when HTMLMediaElement supports default media type but not MediaSource.


sideshowbarker:
w3c-test:mirror

wolenetz (approved):
Looks good to me. Good catch.

PR #6535, opened 2017-07-12
Fix 2 broken tests: Visible test description did not match reference …

atotic:
Found this while testing Chrome. 

These two tests are clearly incorrect, the visible test description did not match reference description.


kojiishi (approved):
lgtm, thank you.

PR #6538, opened 2017-07-13
"Cross-origin objects" are to have enumerable properties

annevk:
See https://github.com/whatwg/html/pull/2777.

### DIFF
--- a/html/browsers/origin/cross-origin-objects/cross-origin-objects.html
+++ b/html/browsers/origin/cross-origin-objects/cross-origin-objects.html
@@ -187,7 +187,7 @@ function checkPropertyDescriptor(desc, propName, expectWritable) {
   var isSymbol = (typeof(propName) == "symbol");
   propName = String(propName);
   assert_true(isObject(desc), "property descriptor for " + propName + " should exist");
-  assert_equals(desc.enumerable, false, "property descriptor for " + propName + " should be non-enumerable");
+  assert_equals(desc.enumerable, true, "property descriptor for " + propName + " should be enumerable");
   assert_equals(desc.configurable, true, "property descriptor for " + propName + " should be configurable");
   if (isSymbol) {
     assert_true("value" in desc,
@@ -274,9 +274,15 @@ addTest(function() {
   assert_array_equals(Object.getOwnPropertyNames(C).sort(),
                       whitelistedWindowPropNames,
                       "Object.getOwnPropertyNames() gives the right answer for cross-origin Window");
+  assert_array_equals(Object.keys(C).sort(),
+                      whitelistedWindowPropNames,
+                      "Object.keys() gives the right answer for cross-origin Window");
   assert_array_equals(Object.getOwnPropertyNames(C.location).sort(),
                       whitelistedLocationPropNames,
                       "Object.getOwnPropertyNames() gives the right answer for cross-origin Location");
+  assert_array_equals(Object.keys(C.location).sort(),
+                      whitelistedLocationPropNames,
+                      "Object.keys() gives the right answer for cross-origin Location");
 }, "[[OwnPropertyKeys]] should return all properties from cross-origin objects");
 
 addTest(function() {
--- a/html/browsers/origin/cross-origin-objects/frame.html
+++ b/html/browsers/origin/cross-origin-objects/frame.html
@@ -37,6 +37,6 @@
 <body>
   <!-- Two subframes to give us some indexed properties -->
   <iframe></iframe>
-  <iframe></iframe>
+  <iframe name=donotleakme></iframe><!-- "donotleakme" is excluded as cross-origin named property due to [[HideFromKeys]] -->
 </body>
 </html>


domenic (approved):
LGTM, although maybe the commit message should also mention that it's testing the change to [[OwnPropertyKeys]]\()

cdumez:
Do we still want to keep those checks in the test?

/*
 * [[Enumerate]]
 */

addTest(function() {
  for (var prop in C)
    assert_unreached("Shouldn't have been able to enumerate " + prop + " on cross-origin Window");
  for (var prop in C.location)
    assert_unreached("Shouldn't have been able to enumerate " + prop + " on cross-origin Location");
}, "[[Enumerate]] should return an empty iterator");

cdumez:
Claiming the properties are enumerable but then not being able to see them when enumerating seems broken to me?

annevk:
Sigh. I wish I had spotted that before. https://tc39.github.io/ecma262/#sec-enumerate-object-properties uses [[OwnPropertyKeys]] ([[Enumerate]] is no more it seems) so named properties will still be hidden, but generally everything else should indeed enumerate. I'll write a patch.

annevk:
See https://github.com/w3c/web-platform-tests/pull/6571 for the follow-up.

bzbarsky:
Shouldn't there be tests for the behavior on non-cross-origin objects too?  Or are there already?

html/browsers/origin/cross-origin-objects/cross-origin-objects.html
```diff
@@ -187,7 +187,7 @@
   var isSymbol = (typeof(propName) == "symbol");
   propName = String(propName);
   assert_true(isObject(desc), "property descriptor for " + propName + " should exist");
-  assert_equals(desc.enumerable, false, "property descriptor for " + propName + " should be non-enumerable");
+  assert_equals(desc.enumerable, true, "property descriptor for " + propName + " should be enumerable");
```
bzbarsky: This change is wrong.  It should be comparing desc.enumerable to !isSymbol, because per spec the symbols should still be non-enumerable.

bzbarsky: Or branch on isSymbol to have nicer messages...

annevk:
I think same-origin is tested by html/dom/interfaces.html already.

html/browsers/origin/cross-origin-objects/cross-origin-objects.html
annevk: Surprised @cdumez didn't find this. Created #6583.

annevk: (Thanks by the way!)

bzbarsky:
> I think same-origin is tested by html/dom/interfaces.html already.

Does that test the indexed props?

annevk:
Adding a test for that now in the other PR. It seems that named properties are still non-enumerable in the same-origin case due to the named properties object setup and Window having LegacyUnenumerableNamedProperties.

annevk:
(And FTR, that's tested already by `html/browsers/the-window-object/window-named-properties.html`.)

html/browsers/origin/cross-origin-objects/cross-origin-objects.html
cdumez: My initial WebKit patch was also wrong so it was passing all the checks :(

domenic:
> It seems that named properties are still non-enumerable in the same-origin case due to the named properties object setup and Window having LegacyUnenumerableNamedProperties.

Uhhh I see... Maybe the consistency argument is not so strong for switching "named properties" (i.e. those from the document-tree child browsing context name property set) to enumerable for cross-origin then? :-/.

annevk:
@domenic given that cross-origin they're on a totally different object I think it's fine. We could maybe consider changing the same-origin setup, but maybe it's best to just leave it alone.

domenic:
Well, but https://github.com/whatwg/html/pull/2777's intent was

> It makes all cross-origin properties that would normally be enumerable on same-origin objects, enumerable also on WindowProxy and Location objects (including when accessed same-origin).

Before that the spec had them as non-enumerable (i.e. same-origin WindowProxy and cross-origin WindowProxy both exposed them as non-enumerable). So maybe we should revert that part of the spec?

annevk:
Well in one case they're on WindowProxy and in the other case they're on some prototype object (the named properties object). Looking at the Gecko patch it's probably easier if all cross-origin properties are enumerable consistently, though @bzbarsky can correct me.

bzbarsky:
In Gecko it's easy either way.  The patch in Gecko just removes code that _forced_ all cross-origin properties to be non-enumerable.  The code that puts the named props on the cross-origin window is at http://searchfox.org/mozilla-central/rev/a83a4b68974aecaaacdf25145420e0fe97b7aa22/js/xpconnect/wrappers/XrayWrapper.cpp#2059-2076 and it would be quite simple to have it pass `false` for the optional fifth argument (`enumerable`) to `FillPropertyDescriptor`.

cdumez:
I am a bit confused here. I understand we expose named properties on cross-origin windows (i.e. we allow looking up iframe by name). However, those frame names are NOT exposed cross-origin, are they? I believe doing so would leak cross-origin information and I thought this is why we had [[HideFromKeys]]: true for those crossOriginWindowProperties in the spec. If this is true, I don't think they should be enumerable either.

bzbarsky:
The set of frame names is not exposed cross-origin.  That is, `[[OwnPropertyKeys]]` does not include them in its return value.

If you know a specific frame name, you can get the property.

Since the names are not returned by `[[OwnPropertyKeys]]`, they're cannot be enumerated viaa `for...in`.  So the only thing affected by whether they're enumerable is what `getOwnPropertyDescriptor` returns, afaict.

annevk:
I'm happy with the specification as it stands now, but if folks feel strongly I can make more changes.

cdumez:
@bzbarsky Ok, this was my understanding. As a result, I don't think getOwnPropertyDescriptor() should lie and claim those properties are enumerable (as they are actually not).

bzbarsky:
It's not lying.  It's just that "enumerable" in `getOwnPropertyDescriptor` is totally unrelated to whether `for..in` sees the property, if `[[OwnPropertyKeys]]` lies about whether the property exists at all...

Anyway, I don't have a very strong feeling about the named props issue, because it doesn't have any real practical impact; happy to implement whatever people decide on there.

domenic:
I'm happy to stop churning on this and call it done.

cdumez:
Your call. I feel it would be cleaner for enumerable to be false but in the end, it does not really matter since there is no behavior difference.

PR #6563, opened 2017-07-16
[intersection-observer] Make testing for expected time origins more reliable.

tbtlr:
There is a hard assertion in intersection/observer.html that checks if topWindowTimeBeforeNotification is greater than iframeWindowTimeAfterNotification with a note that following tests are only significant if so. We had several intermittent failures in Firefox were this wasn't the case. IMO this shouldn't fail the tests. I removed the assert but left the comment.

@szager-chromium any comments why this was added in first place? And if there was a legit reason, shouldn't it at least be changed to topWindowTimeBeforeNotification > iframeWindowTimeBeforeNotification?


w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/254055793)

# Firefox (nightly) #
Testing web-platform-tests at revision d58ba571b5717142d666824392b52601c28e1046
Using browser at version BuildID 20170715100214; SourceStamp bd1b6a4e5d8e1184a89ae60a74fd86d3d4a9e95c
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="http://w3c-test.org/submissions/6563/intersection-observer/timestamp.html">/intersection-observer/timestamp.html</a></summary>


|                                          Subtest                                          | Results | Messages |
|-------------------------------------------------------------------------------------------|---------|----------|
|                                                                                           | OK      |          |
| `Check that timestamps correspond to the to execution context that created the observer.` | PASS    |          |
| `First rAF after iframe is loaded.`                                                       | PASS    |          |
| `Generate notifications.`                                                                 | PASS    |          |

</details>
</details>

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/254055794)

# Sauce (safari) #
Testing web-platform-tests at revision d58ba571b5717142d666824392b52601c28e1046
Using browser at version 10.0
Starting 10 test iterations
All results were stable
## All results ##
<details>
<summary>1 test ran</summary>
<details>
<summary><a href="http://w3c-test.org/submissions/6563/intersection-observer/timestamp.html">/intersection-observer/timestamp.html</a></summary>


|                                          Subtest                                          | Results |     Messages     |
|-------------------------------------------------------------------------------------------|---------|------------------|
|                                                                                           | ERROR   |                  |
| `Check that timestamps correspond to the to execution context that created the observer.` | TIMEOUT | `Test timed out` |

</details>
</details>

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/254055800)

# Chrome (unstable) #
Testing web-platform-tests at revision d58ba571b5717142d666824392b52601c28e1046

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/254055801)

# Sauce (MicrosoftEdge) #
Testing web-platform-tests at revision d58ba571b5717142d666824392b52601c28e1046

sideshowbarker:
w3c-test:mirror

tobytailor:
@philipwalton, any thoughts on this?

philipwalton:
@tobytailor, no thoughts here. My main experience with this API was from the polyfill perspective, which doesn't cover iframes at all.

szager-chromium:
The purpose of the assert is to verify that there's a sufficient difference between document.performance.now() and iframe.contentWindow.performance.now().

The whole point of the test is to verify that timestamps are relative to the correct execution context.  That's why there's a 40ms setTimeout before creating the iframe -- to ensure that the performance clock for the iframe is sufficiently skewed from the performance clock of the parent document.

Without that assert, it's possible that the test could pass when it should fail, because timestamps generated by the iframe execution context could be close enough to timestamps generated by the parent document to make it unclear which context's clock was used.

szager-chromium:
We haven't seen any flakiness of this test in chromium.

This should be the numerical order of the timestamp variables:

iframeWindowTimeBeforeNotification
iframeWindowTimeAfterNotification
topWindowTimeBeforeNotification
topWindowTimeAfterNotification

You can imagine this as two disjoint ranges:

[iframeWindowTimeBeforeNotification, iframeWindowTimeAfterNotification]
[topWindowTimeBeforeNotification, topWindowTimeBeforeNotification]

Here are some assumptions about the timestamps:

- topWindowTimeBeforeNotification should be at least 40ms greater (i.e., later) than iframeWindowTimeBeforeNotification, because the iframe was created at least 40ms after the top document began running script.

- topWindowTimeAfterNotification should be approximately 16ms (one frame) after topWindowTimeBeforeNotification.  Likewise for iframeWindowTime(Before|After)Notification.

- By extrapolation, topWindowTimeBeforeNotification should be approximately 32ms (40ms - 16ms) greater than iframeWindowTimeAfterNotification.  That 32ms is an arbitrary fudge factor I chose to be quite sure that topWindowTimeBeforeNotification is greater than iframeWindowTimeAfterNotification, without causing the test to run overly long.

The test verifies the timestamps on two notifications:

For the observer created in the iframe, the notification timestamp should be within the iframeWindow range.
For the observer created in the top document, the notification timestamp should be within the topWindow range.

If there is any overlap between the two ranges, then there's a possibility that, e.g., a timestamp created in the iframe context could be within the time range for the parent document context.  In that case, the test cannot verify which context actually created the timestamp.

Hopefully that helps; if it's still unclear, I can draw a diagram that might help.

szager-chromium:
Wow, my arithmetic is terrible :)

Please s/32ms/24ms/g in my previous comment.

tobytailor:
@szager-chromium thanks a lot for the very comprehensive explanation. It help me a lot finding the real issue why this test is so flaky in FF. The way the value for timeSkew is picked (currently 40ms) is loosely tight to the test running at 60Hz. We can't guarantee that always being the case, especially when running it automated on different platforms and hosts. This explains why we mostly saw this testing failing on very slow machines (mostly debug build running on vm's). When running slower than 30Hz your assumptions don't apply any more. I think we need a better, more reliable way to choose a value.

jgraham:
40ms timeouts are totally unsuitable for running on many kinds of CI infrastructure; we see many tests be flaky on all sorts of browsers when they depend on such short intervals. In general I would always  use a timeout at least 10x that value. However if this is tied to the framerate then maybe the thing to do is to set the test up so that a certain number of rAF intervals have passed before creating the iframe? Maybe that will allow you to get away with a shorter total timeout.

szager-chromium:
If you can guarantee that frame timing will be consistent with low variance during a particular test run, then using a three-rAF delay rather than 40ms should work.  But if there's a possibility that frame timing will vary significantly during the test run, then increasing timeSkew may be the best we can do.

tobytailor:
I updated the PR with a commit to change the test so that it waits for a certain amount of frames (I picked 3 for now) before creating the iframe instead of a static amount of time.

szager-chromium (approved):

tobytailor:
I spent a bit more time collecting exact timer values in failing test situations over the past 10 days. Given the results I would like to add one more assumption to @szager-chromium's list in comment 19:

The time passed between a top window notification's time and iframeWindowTimeBeforeNotification should be the same (with a relatively small tolerance) as the time passed between an iframe notificaton's time and iframeWindowTimeBeforeNotification. Since these should usually processed in a tight loop, the tolerance should be very small (even on slow machines where the drop in framerate was mostly due to script execution).

This means that in case of notification times being relative to the wrong time origin (which is what the assert in questions is testing for), the differences should be way off.

So I suggest replacing the current assertion with

`assert_approx_equals(topWindowEntries[1].time - topWindowTimeBeforeNotification, iframeWindowEntries[1].time - iframeWindowTimeBeforeNotification, tolerance)`

Value for tolerance is yet to be figured out (tho I'm pretty sure 16ms should be enough, even on slow frame rates).

foolip:
I see that @szager-chromium approved this back in August, so presumably now all that needs to happen is to rebase to resolve the conflicts? If that's right, I can re-approve and merge it once Travis is happy with the test.

foolip:
Ping @tobytailor?

zcorpan:
Trying to cherry-pick cf899510013f9657217d34b491fb538835ce1c48 to a new branch based on master and resolving the conflict results in an empty commit. Has this change reached master somehow?

gsnedders:
Landed in https://github.com/web-platform-tests/wpt/commit/e8d533c159c3ebad464e39378b70cff90f9281eb via Gecko.

PR #6647, opened 2017-07-28
Refactor @font-feature-values test and make it match with spec.

canova:
According to [spec][spec-link],
> If the same identifier is defined mulitple times for a given feature type and font family, the last defined value is used.

But currently we have two tests that checks otherwise. Tests should match with spec here.
cc @upsuper 

[spec-link]: https://drafts.csswg.org/css-fonts-3/#basic-syntax

### DIFF
--- a/css/css-fonts-3/test_font_feature_values_parsing.html
+++ b/css/css-fonts-3/test_font_feature_values_parsing.html
@@ -152,8 +152,8 @@ var testrules = [
   { rule: makeRule("--bongo", "@styleset { blah: 1; }"), invalid: true },
 
   /* ident tests */
-  { rule: _("@styleset { blah: 1; blah: 1; }"), serializationSame: true },
-  { rule: _("@styleset { blah: 1; de-blah: 1; blah: 2; }"), serializationSame: true },
+  { rule: _("@styleset { blah: 1; blah: 1; }"), serialization: _("@styleset { blah: 1; }") },
+  { rule: _("@styleset { blah: 1; de-blah: 1; blah: 2; }"), serialization: _("@styleset { blah: 2; de-blah: 1; }") },
   { rule: _("@styleset { \\tra-la: 1; }"), serialization: _("@styleset { tra-la: 1; }") },
   { rule: _("@styleset { b\\lah: 1; }"), serialization: _("@styleset { blah: 1; }") },
   { rule: _("@styleset { \\62 lah: 1; }"), serialization: _("@styleset { blah: 1; }") },


css/css-fonts-3/test_font_feature_values_parsing.html
```diff
@@ -152,8 +152,8 @@
   { rule: makeRule("--bongo", "@styleset { blah: 1; }"), invalid: true },
 
   /* ident tests */
-  { rule: _("@styleset { blah: 1; blah: 1; }"), serializationSame: true },
-  { rule: _("@styleset { blah: 1; de-blah: 1; blah: 2; }"), serializationSame: true },
+  { rule: _("@styleset { blah: 1; blah: 1; }"), serialization: _("@styleset { blah: 1; }") },
+  { rule: _("@styleset { blah: 1; de-blah: 1; blah: 2; }"), serialization: _("@styleset { blah: 2; de-blah: 1; }") },
```
canaltinova: We should definitely use second value for `blah` here but I'm not sure about if it should be in first place or second.

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/258426868)

# Firefox (nightly) #

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/258426869)

# Sauce (safari) #

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/258426875)

# Chrome (unstable) #

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/258426876)

# Sauce (MicrosoftEdge) #

sideshowbarker:
w3c-test:mirror

litherum:
I don't understand this, so I don't think I can review it.

sideshowbarker:
@gsnedders hope you might be able to figure out who else could review this so it can be merged

sideshowbarker:
Actually seems like a review from @upsuper would be good

upsuper:
The serialization of `@font-feature-values` is apparently underspecified. The spec wording doesn't really imply that the serialization should only contain the last item.

But I agree that it makes more sense to only serialize each item once, as it matches what browsers currently do for serializing declaration block. (That doesn't seem to exactly match the spec, though.)

For duplicate items in declaration block, Blink and Gecko serialize it at the position of its last appearance, while Edge serializes it at the first appearance. I guess the former seems more intuitive, but either way is probably fine. Since there are two impls for the former behavior, I'm inclined to that one.

I'm not sure what should we do here, though, since it is unspecified, although the change is probably desirable.

Ms2ger:
@upsuper can you file a spec issue?

svgeesus (changes requested):
Changes look good, but this now needs to be a Fonts 4 test rather than a fonts 3 test. Feature was already at risk, has only one implementation, and CSS WG resolved to move it to Fonts 4. I haven't done the spec edits yet, plan to get to that soon. The section heading is there as a stub in Fonts 4 though, so the test can link there.

I'm not sure why there is a conflict on the patch.

svgeesus:
@upsuper the specification is [now clear on duplicate values](https://drafts.csswg.org/css-fonts-3/#basic-syntax):

> If the same identifier is defined multiple times for a given feature type and font family, the last defined value is used.

Issue #6602, opened 2017-07-22
Investigate Safari flakiness in existing tests


sideshowbarker:
With at least two PRs recently we have run into Safari flakiness with existing tests that didn’t have anything specifically to do with the new changes introduced in the PRs.

* https://github.com/w3c/web-platform-tests/pull/6578
* https://github.com/w3c/web-platform-tests/pull/6595

sideshowbarker:
Also https://github.com/w3c/web-platform-tests/pull/6588 I think

annevk:
cc @cdumez @rniwa

sideshowbarker:
We are running into this more with other PRs too so I’m not going to try to update this issue to explicitly itemize all the failures. I expect (hope) the ones here are already illustrative enough but if not am happy to make time to try to help further isolate the cause(s).

sideshowbarker:
Note that over at https://github.com/w3c/web-platform-tests/pull/6655#discussion_r131495010 @cdumez says:

> FYI, the LargeString check is slow is causes the test to flakily timeout for WebKit

…about the particular test in that PR at least

annevk:
Also affects #6606 I think.

PR #6655, opened 2017-07-28
Initial Commit of sendBeacon Tests

Brandr0id:

### DIFF
--- a/beacon/beacon-basic-string.html
+++ b/beacon/beacon-basic-string.html
@@ -1,0 +1,18 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>W3C Beacon Basic String Test</title>
+    <meta name="timeout" content="long">
+    <meta name="author" title="Microsoft Edge" href="https://www.microsoft.com">
+    <script src="/resources/testharness.js"></script>
+    <script src="/resources/testharnessreport.js"></script>
+</head>
+<body>
+    <script src="/common/utils.js"></script>
+    <script src="beacon-common.js?pipe=sub"></script>
+    <script>
+        "use strict";
+        runTests(stringTests);
+    </script>
+</body>
+</html>
--- a/beacon/beacon-basic-string.html
+++ b/beacon/beacon-basic-string.html
@@ -9,7 +9,7 @@
 </head>
 <body>
     <script src="/common/utils.js"></script>
-    <script src="beacon-common.js?pipe=sub"></script>
+    <script src="beacon-common.sub.js"></script>
     <script>
         "use strict";
         runTests(stringTests);
--- a/beacon/beacon-common.js
+++ b/beacon/beacon-common.js
@@ -1,0 +1,353 @@
+"use strict";
+
+// Different sizes of payloads to test.
+var smallPayloadSize = 10;
+var mediumPayloadSize = 10000;
+var largePayloadSize = 50000;
+var maxPayloadSize = 65536; // The maximum payload size allowed for a beacon request.
+
+// String payloads of various sizes sent by sendbeacon. The format of the payloads is a string:
+//     <numberOfCharacters>:<numberOfCharacters *'s>
+//     ex. "10:**********"
+var smallPayload = smallPayloadSize + ":" + Array(smallPayloadSize).fill('*').join("");
+var mediumPayload = mediumPayloadSize + ":" + Array(mediumPayloadSize).fill('*').join("");
+var largePayload = largePayloadSize + ":" + Array(largePayloadSize).fill('*').join("");
+// Subtract 6 from maxPayloadSize because 65536 is 5 digits, plus 1 more for the ':'
+var maxPayload = (maxPayloadSize - 6) + ":" + Array(maxPayloadSize - 6).fill('*').join("")
+
+// Test case definitions.
+//      id: String containing the unique name of the test case.
+//      data: Payload object to send through sendbeacon.
+var noDataTest = { id: "NoData" };
+var nullDataTest = { id: "NullData", data: null };
+var undefinedDataTest = { id: "UndefinedData", data: undefined };
+var smallStringTest = { id: "SmallString", data: smallPayload };
+var mediumStringTest = { id: "MediumString", data: mediumPayload };
+var largeStringTest = { id: "LargeString", data: largePayload };
+var maxStringTest = { id: "MaxString", data: maxPayload };
+var emptyBlobTest = { id: "EmptyBlob", data: new Blob() };
+var smallBlobTest = { id: "SmallBlob", data: new Blob([smallPayload]) };
+var mediumBlobTest = { id: "MediumBlob", data: new Blob([mediumPayload]) };
+var largeBlobTest = { id: "LargeBlob", data: new Blob([largePayload]) };
+var maxBlobTest = { id: "MaxBlob", data: new Blob([maxPayload]) };
+var emptyBufferSourceTest = { id: "EmptyBufferSource", data: new Uint8Array() };
+var smallBufferSourceTest = { id: "SmallBufferSource", data: CreateArrayBufferFromPayload(smallPayload) };
+var mediumBufferSourceTest = { id: "MediumBufferSource", data: CreateArrayBufferFromPayload(mediumPayload) };
+var largeBufferSourceTest = { id: "LargeBufferSource", data: CreateArrayBufferFromPayload(largePayload) };
+var maxBufferSourceTest = { id: "MaxBufferSource", data: CreateArrayBufferFromPayload(maxPayload) };
+var emptyFormDataTest = { id: "EmptyFormData", data: CreateEmptyFormDataPayload() };
+var smallFormDataTest = { id: "SmallFormData", data: CreateFormDataFromPayload(smallPayload) };
+var mediumFormDataTest = { id: "MediumFormData", data: CreateFormDataFromPayload(mediumPayload) };
+var largeFormDataTest = { id: "LargeFormData", data: CreateFormDataFromPayload(largePayload) };
+// We don't test maxFormData because the extra multipart separators make it difficult to
+// calculate a maxPayload.
+
+// Test case suites.
+// Due to quota limits we split the max payload tests into their own bucket.
+var stringTests = [noDataTest, nullDataTest, undefinedDataTest, smallStringTest, mediumStringTest, largeStringTest];
+var stringMaxTest = [maxStringTest];
+var blobTests = [emptyBlobTest, smallBlobTest, mediumBlobTest, largeBlobTest];
+var blobMaxTest = [maxBlobTest];
+var bufferSourceTests = [emptyBufferSourceTest, smallBufferSourceTest, mediumBufferSourceTest, largeBufferSourceTest];
+var bufferSourceMaxTest = [maxBufferSourceTest];
+var formDataTests = [emptyFormDataTest, smallFormDataTest, mediumFormDataTest, largeFormDataTest];
+var formDataMaxTest = [largeFormDataTest];
+var allTests = [].concat(stringTests, stringMaxTest, blobTests, blobMaxTest, bufferSourceTests, bufferSourceMaxTest, formDataTests, formDataMaxTest);
+
+// This special cross section of test cases is meant to provide a slimmer but reasonably-
+// representative set of tests for parameterization across variables (e.g. redirect codes,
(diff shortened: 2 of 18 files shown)


Brandr0id:
@toddreifsteck @igrigorik

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/258726553)

# Firefox (nightly) #

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/258726554)

# Sauce (safari) #

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/258726560)

# Chrome (unstable) #

w3c-bots:
[View the complete job log.](https://travis-ci.org/w3c/web-platform-tests/jobs/258726561)

# Sauce (MicrosoftEdge) #

beacon/beacon-common.js
```diff
@@ -0,0 +1,353 @@
...
+function failSession(session, reason) {
+    session.testCases.forEach(function(testCase) {
+        var test = testCase.test;
+        test.unreached_func(reason)();
+    });
+}
+
+// Creates an iframe on the document's body and runs the sample tests from the iframe.
+// The iframe is navigated immediately after it sends the data, and the window verifies
+// that the data is still successfully sent.
```
igrigorik: 👍

beacon/beacon-redirect.window.js
```diff
@@ -0,0 +1,25 @@
+// META: script=/common/utils.js
+// META: script=beacon-common.js?pipe=sub
+
+"use strict";
+
+// Execute each sample test per redirect status code.
+// Note that status codes 307 and 308 are the only codes that will maintain POST data
```
igrigorik: Awesome 👍

igrigorik (approved):
Looks great, kudos for the thorough work Brandon!

igrigorik:
@siusin @jgraham any idea on why the bots are failing here and not reporting the results? One of the builds appears to be [erroring out with a Selenium error](https://travis-ci.org/w3c/web-platform-tests/jobs/258726554), but I don't see how or why that would be caused by the code here.

In the meantime, I ran the tests locally on my machine and the results make sense -- e.g. Blob tests fail on Chrome because those are currently disabled.

beacon/beacon-cors.window.js
```diff
@@ -0,0 +1,41 @@
...
+    self.buildBaseUrl = function(baseUrl) {
+        return "http://{{domains[www]}}:{{ports[http][0]}}";
+    };
+    // Implement the self.buildTargetUrl extension to append a directive
+    // to the handler, that it should return CORS headers, if 'allowCors'
+    // is true.
+    self.buildTargetUrl = function(targetUrl) {
+        // Note that 'allowCors=true' is not necessary for the sendBeacon() to reach
+        // the server. Beacons use the HTTP POST method, which is a CORS-safelisted
+        // method, and thus they do not trigger preflight. If the server does not
```
igrigorik: > Beacons use the HTTP POST method, which is a CORS-safelisted method, and thus they do not trigger preflight.

This is not true, the behavior here is also conditional on the type of data being transferred. Specifically, "no-cors" applies for CORS-safelisted content-types (`application/x-www-form-urlencoded`, `multipart/form-data`, or `text/plain`), other types change cors mode to "cors" and require a preflight.

Processing section covers this: https://w3c.github.io/beacon/#sec-processing-model

Could we extend these tests to account for the content-type differences?

igrigorik (changes requested):
On second pass, a few CORS-related questions before we land this..

beacon/beacon-basic-string.html
```diff
@@ -0,0 +1,18 @@
...
+    <meta name="author" title="Microsoft Edge" href="https://www.microsoft.com">
+    <script src="/resources/testharness.js"></script>
+    <script src="/resources/testharnessreport.js"></script>
+</head>
+<body>
+    <script src="/common/utils.js"></script>
+    <script src="beacon-common.js?pipe=sub"></script>
+    <script>
+        "use strict";
+        runTests(stringTests);
```
cdumez: FYI, the LargeString check is slow is causes the test to flakily timeout for WebKit.

igrigorik: The string is large, but not that large.. Do you have any suggestions for how to improve this?

igrigorik:
/me nudges @siusin @jgraham, any ideas for: https://github.com/w3c/web-platform-tests/pull/6655#issuecomment-319775273 -- thanks :)

siusin:
pinged @sideshowbarker 

Mike said this Safari failure is a known issue #6602 , he will look at the Chrome failure later.

beacon/beacon-basic-string.html
cdumez: I'll will investigate on WebKit side. The test takes up to ~130 seconds on some of our debug bots.
In any case, feel free to land as is. It could simply be an issue with our implementation as it is very new.

sideshowbarker:
The Chrome failure is not due to flakiness but instead just due to the build taking longer than the limit that Travis sets for builds to complete. So as soon as all review comments are resolved and a reviewer approves the PR, I can merge it

igrigorik:
@sideshowbarker interesting, presumably the long build is not due to the tests introduced here? Are you seeing similar behavior on other PRs?

@Brandr0id  do you have bandwidth to iterate on CORS issue we flagged above? https://github.com/w3c/web-platform-tests/pull/6655#discussion_r130978823. If not, we can land this and take that as a separate PR.. I don't want this PR to bitrot.

sideshowbarker:
> @sideshowbarker interesting, presumably the long build is not due to the tests introduced here? Are you seeing similar behavior on other PRs?

Actually I misspoke—I restarted the job and it now is (sorta) indicating unstable results:

https://github.com/w3c/web-platform-tests/pull/6655#issuecomment-318779277
https://travis-ci.org/w3c/web-platform-tests/jobs/258726560

But those results themselves are weird—notice the `TIMEOUT: 20/10`, which is saying that when the 10 iterations of test were run, it timed out in 20 of those iterations. So that basically doesn’t make sense and I can vaguely recall the cause is a known issue in the test infrastructure that’s happened a lot of time before, but I don’t remember the cause. Anyway, that doesn’t need to block merging.

beacon/beacon-common.js
```diff
@@ -0,0 +1,353 @@
...
+            var iframeWindow = document.getElementById("iframe").contentWindow;
+            // We run into problems passing the testCase over the document boundary,
+            // because of structured cloning constraints. Instead we'll send over the
+            // test case id, and the iFrame can load the static test case by including
+            // beacon-common.js.
+            tests[testCase.origId] = testCase;
+            iframeWindow.postMessage([testCase.origId, testCase.url, funcName], "*");
+        };
+
+        runTests(sampleTests);
```
cdumez: Running this in WebKit & Blink, I see runTests() getting called twice (so we run all 7 sample tests twice, a total of 14 tests). This seems to indicate that the onload handler is called twice. Isn't it the expected behavior that unload gets called twice given that the iframe navigates itself to Microsoft.com at some point?

beacon/beacon-common.js
```diff
@@ -0,0 +1,353 @@
...
+// The iframe is navigated immediately after it sends the data, and the window verifies
+// that the data is still successfully sent.
+//    funcName: "beacon" to send the data via navigator.sendBeacon(),
+//              "fetch" to send the data via fetch() with the keepalive flag.
+function runSendInIframeAndNavigateTests(funcName) {
+    var iframe = document.createElement("iframe");
+    iframe.id = "iframe";
+    iframe.onload = function() {
+        var tests = Array();
+
```
cdumez: Adding "iframe.onload = null;" here makes WebKit / Blink run only 7 tests as expected (runTests is only called once).
That said, all 7 tests time out in WebKit & Blink and I have not figured out why yet. Firefox seems to run all 7 tests fine though..

beacon/fetch-keepalive-navigate.iFrame.html
```diff
@@ -0,0 +1,49 @@
...
+            }
+            else {
+                throw new Error(func + " is an invalid function");
+            }
+
+            // Let the main page continue the test if we don't immediately throw an exception.
+            parent.postMessage(testCaseId, "*");
+
+            // Now navigate ourselves.
+            if (tests == sampleTests.Length) {
```
cdumez: Shouldn't this be .length instead of .Length ?

toddreifsteck:
@igrigorik @Brandr0id Lets avoid taking new tests into this PR. After it is landed, new tests are "easy". Brandon will have enough on his hands to land this for us while juggling the other tasks he has on his plate.

igrigorik:
@toddreifsteck sgtm.

beacon/beacon-common.js
```diff
@@ -0,0 +1,353 @@
...
+            // beacon-common.js.
+            tests[testCase.origId] = testCase;
+            iframeWindow.postMessage([testCase.origId, testCase.url, funcName], "*");
+        };
+
+        runTests(sampleTests);
+    };
+
+    document.body.appendChild(iframe);
+    iframe.src = "fetch-keepalive-navigate.iFrame.html";
```
cdumez: So one issue with Blink / WebKit is that you need to set the src attribute *before* adding the iframe to the document. Otherwise, onload gets called twice, once for about:blank load, and a second time for the fetch-keepalive-navigate.iFrame.html load.

beacon/beacon-common.js
cdumez: In WebKit / Blink, I need to do:
iframe.onload = null;

here, otherwise, the onload handler gets called again for the navigation to Microsoft.com at the end of the test and we run 14 tests instead of 7.

beacon/fetch-keepalive-navigate.iFrame.html
```diff
@@ -0,0 +1,49 @@
...
+                window.location = "http://microsoft.com";
```
cdumez: Can we navigate to "about:blank" instead. Using this URL, I see this in the console:
Refused to display 'https://www.microsoft.com/en-us/' in a frame because it set 'X-Frame-Options' to 'sameorigin'.

youennf:
Renaming beacon-common.js as beacon-common.sub.js might be nice.
This would remove the need for ?pipe=sub.

Looking at the polling mechanism, raising an error after some long enough time like 5 seconds would help limit the time needed to run the tests in case they time out.

beacon/beacon-common.js
Brandr0id: Swapped this around :)

beacon/beacon-common.js
Brandr0id: This may also be why the Chrome test pass is seeing 20/10 failures. Updated this.

beacon/fetch-keepalive-navigate.iFrame.html
Brandr0id: I updated it to use the root of the test server, does that work?

beacon/fetch-keepalive-navigate.iFrame.html
Brandr0id: indeed it should be, updated this.

beacon/beacon-cors.window.js
Brandr0id: I agree with the other comments, let's land this PR then expand/add the extra test cases :)

Brandr0id:
@jgraham \ @toddreifsteck This isn't really specific to this PR but I noticed the Edge run failing some tests due to behavior that was changed with the Creators Update. It looks like sauce labs is hard coded to use the anniversary update. Is there any particular reason this isn't either using the creators update '15.15063', or 'latest' which I think Sauce supports?

Brandr0id:
Ping @igrigorik

beacon/beacon-basic-string.html
igrigorik: @cdumez any findings on your end for this one?

igrigorik (approved):
As noted earlier, the CORS logic in https://github.com/w3c/web-platform-tests/pull/6655/files#diff-fcdc2e3ae8101103a5474d1a0f415b56 needs work as it doesn't cover the non-safelisted content type cases, but we'll handle that in a separate commit. 

Otherwise, LGTM. Thanks a lot for the great work on this!

PR #6689, opened 2017-08-02
Return single cookie for GetNamedCookie;

jgraham:
The GetNamedCookie command currently returns a JSON Array of one,
retained cookie, after it has removed the cookies that don't match by
name.

This is in violation of the WebDriver specification, which says it must
return the cookie serialisation directly.

MozReview-Commit-ID: 9yEiarEGBez

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1372595 [ci skip]


jgraham:
Code reviewed upstream.

PR #6693, opened 2017-08-02
Move all wdspec tests into a package;

jgraham:
We need a top-level package for tests and support files so that
support modules are importable from subpackages. For example,
previously webdriver/support/inline.py could not be imported from
webdriver/actions/mouse.py.

I'm creating the package in a subdirectory under webdriver in
order to avoid a name-clash with webdriver, the WebDriver client
library. Part of the motivation for adding the subdirectory rather
than renaming the client library is that it's a relatively
isolated change.

MozReview-Commit-ID: BihAKcNLbXK

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1353895 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6693)


jgraham:
Code reviewed upstream.

PR #6707, opened 2017-08-02
Ensure we have a valid window handle before trying to wait,

jgraham:
If we switched to a window that is now closed (possible in a reftest
run, for example), we might end up calling execute_async_script
without a valid current window. The runner window is always supposed
to be open (and if it isn't then waiting won't make much sense
anyway), so switch to that before executing script in wait().

MozReview-Commit-ID: ERoN4sz5SuH

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1381858 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6707)


jgraham:
Code reviewed upstream.

PR #6709, opened 2017-08-02
add wpt interpolation test for mismatched 3D transforms.

jgraham:
This test aims for forcing the two mismatched transforms to be decomposed into
matrix3d before interpolation. To do so, we need to:

1. make 'From' and 'To' to be two mismatched transforms,
2. at least one of the two transforms has to be a 3D transform.

In this way, we not only test the interpolation, but also test the correctness
of 3D matrix decomposition.

MozReview-Commit-ID: FAKu6GFVZRm

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1382517 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6709)


jgraham:
Code reviewed upstream.

PR #6755, opened 2017-08-04
Rewrite pointerevents test to not use generate_tests

gsnedders:
cc/ @NavidZ who wrote at least some of this

### DIFF
--- a/pointerevents/extension/pointerevent_constructor.html
+++ b/pointerevents/extension/pointerevent_constructor.html
@@ -17,36 +17,79 @@
             var target0 = document.getElementById("target0");
             // set values for non-default constructor
             on_event(target0, "pointermove", this.step_func(function(event) {
-                generate_tests(assert_equals, [
-                    ["getCoalescedEvents().length", event.getCoalescedEvents().length, 2],
-                    ["event.target", event.target, target0],
-                    ["event.currentTarget", event.currentTarget, target0],
-                    ["event.eventPhase", event.eventPhase, Event.AT_TARGET],
-                    ["event.clientX", event.clientX, 310],
-                    ["getCoalescedEvents()[0].clientX", event.getCoalescedEvents()[0].clientX, 300],
-                    ["getCoalescedEvents()[1].clientX", event.getCoalescedEvents()[1].clientX, 310],
-                ]);
-                for (var i=0; i<event.getCoalescedEvents().length; i++) {
-                    var coalescedEvent = event.getCoalescedEvents()[i];
-                    generate_tests(assert_equals, [
-                        ["getCoalescedEvents()[" + i + "].pointerId", coalescedEvent.pointerId, event.pointerId],
-                        ["getCoalescedEvents()[" + i + "].pointerType", coalescedEvent.pointerType, event.pointerType],
-                        ["getCoalescedEvents()[" + i + "].isPrimary", coalescedEvent.isPrimary, event.isPrimary],
-                        ["getCoalescedEvents()[" + i + "].getCoalescedEvents().length", coalescedEvent.getCoalescedEvents().length, 0],
-                        ["getCoalescedEvents()[" + i + "].target", coalescedEvent.target, target0],
-                        ["getCoalescedEvents()[" + i + "].currentTarget", coalescedEvent.currentTarget, null],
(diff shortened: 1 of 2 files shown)


pointerevents/extension/pointerevent_constructor.html
```diff
@@ -17,36 +17,79 @@
             var target0 = document.getElementById("target0");
             // set values for non-default constructor
             on_event(target0, "pointermove", this.step_func(function(event) {
-                generate_tests(assert_equals, [
-                    ["getCoalescedEvents().length", event.getCoalescedEvents().length, 2],
-                    ["event.target", event.target, target0],
```
gsnedders: Why is this here rather than in `pointerevents/pointerevent_constructor.html`, @NavidZ?

pointerevents/extension/pointerevent_constructor.html
```diff
@@ -17,36 +17,79 @@
...
+                        var coalescedEvent = event.getCoalescedEvents()[i];
+                        assert_equals(coalescedEvent.offsetX, event.offsetX + (i==0?-10:0));
+                    }, "getCoalescedEvents()[" + i + "].offsetX");
+                    test(function() {
+                        var coalescedEvent = event.getCoalescedEvents()[i];
+                        assert_equals(coalescedEvent.offsetY, event.offsetY);
+                    }, "getCoalescedEvents()[" + i + "].offsetY");
+
+                    // FIXME: gsnedders doesn't see what guarantees this but
+                    // NavidZ had this above before
```
gsnedders: @NavidZ ??

pointerevents/extension/pointerevent_constructor.html
NavidZ: The getCoalescedEvents is only in the extension API and is not going to be part of PointerEvent spec v2. So getCoalescedEvents is not event defined in the main folder. For the target the goal was to just have the check for the target of the coalesced events. But I guess it was more of a copy paste that we have the event.target tested here too.

pointerevents/extension/pointerevent_constructor.html
NavidZ: I don't understand what you meant by "what guarantees this". This is what the test and the spec expects. If browser doesn't do that that means it fails the test. Note that I'm creating these events with the specified clientX/Y values myself here in this file.

pointerevents/extension/pointerevent_constructor.html
gsnedders: Right, I got that; just wondering why we have `event.target` tested here and seemingly otherwise have them tested in manual tests.

pointerevents/extension/pointerevent_constructor.html
gsnedders: Maybe I'm just not clear on the definition of `getCoalescedEvents`. :)

The definition talks about a whole load of properties, including offsetX/offsetY, but not clientX/clientY, and it's not clear to me what those rules mean when it comes to the value of offsetX. (This is really leaning towards a spec bug given I can't interpret this. :))

NavidZ: That is correct that the spec hasn't talked about clientX/Y at all. This was just one example of the event properties that we wanted to make sure it is set correctly (the same as what it is passed in the dictionary). It can be argued that we should have done this with all the properties and we just tested the clientX/Y as an example here.

pointerevents/extension/pointerevent_constructor.html
NavidZ: Sure. This part can certainly be moved to the main folder.

sideshowbarker:
w3c-test:mirror

pointerevents/extension/pointerevent_constructor.html
gsnedders: I've filed https://github.com/w3c/pointerevents/issues/215 based on my problem understanding this.

Ms2ger:
@gsnedders, what's the plan here?

gsnedders:
@Ms2ger fix up a bunch of stuff, land it, sometime

staktrace (changes requested):
Trying to clear out my review queue; I can't just remove myself as reviewer so I'm going to "request changes" since it sounds like this PR needs changes anyway (per last comment)

PR #6761, opened 2017-08-04
verify that ArrayBuffer Content-Type is empty

igrigorik:
Discussion https://github.com/w3c/beacon/issues/40.

### DIFF
--- a/beacon/headers/header-content-type.html
+++ b/beacon/headers/header-content-type.html
@@ -35,7 +35,7 @@ function testContentTypeHeader(what, contentType, title) {
     assert_true(navigator.sendBeacon(testUrl, what), "SendBeacon Succeeded");
     return pollResult(test, id) .then(result => {
       if (contentType == "multipart/form-data") {
-        assert_true(result.startsWith(contentType), "Correct referrer header result");
+        assert_true(result.startsWith(contentType), "Correct Content-Type header result");
       } else {
         assert_equals(result, contentType, "Correct Content-Type header result");
       }
--- a/beacon/headers/header-content-type.html
+++ b/beacon/headers/header-content-type.html
@@ -34,7 +34,11 @@ function testContentTypeHeader(what, contentType, title) {
   promise_test(function(test) {
     assert_true(navigator.sendBeacon(testUrl, what), "SendBeacon Succeeded");
     return pollResult(test, id) .then(result => {
-      assert_true(result.startsWith(contentType), "Correct referrer header result");
+      if (contentType.length == 0) {
+        assert_true(result.length == 0, "Content-Type header for ArrayBuffer should be empty");
+      } else {
+        assert_true(result.startsWith(contentType), "Correct Content-Type header result");
+      }
     });
   }, "Test content-type header for a body " + title);
 }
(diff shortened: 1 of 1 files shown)


beacon/headers/header-content-type.html
```diff
@@ -34,7 +34,11 @@
   promise_test(function(test) {
     assert_true(navigator.sendBeacon(testUrl, what), "SendBeacon Succeeded");
     return pollResult(test, id) .then(result => {
-      assert_true(result.startsWith(contentType), "Correct referrer header result");
+      if (contentType.length == 0) {
```
cdumez: As Anne commented on the bug:
""
Generally Content-Type's value is always a value that is completely known, other than the boundary parameter of multipart/form-data. It seems you could be much more strict overall here.
"""

Can we use assert_equals() in all cases and startsWith() only when the parameter is a FormData?

igrigorik:
@cdumez yep, ptal. Confused myself earlier.. thanks for the nudge.

beacon/headers/header-content-type.html
```diff
@@ -34,7 +34,11 @@
   promise_test(function(test) {
     assert_true(navigator.sendBeacon(testUrl, what), "SendBeacon Succeeded");
     return pollResult(test, id) .then(result => {
-      assert_true(result.startsWith(contentType), "Correct referrer header result");
+      if (contentType == "multipart/form-data") {
+        assert_true(result.startsWith(contentType), "Correct referrer header result");
```
cdumez: Typo: Content-Type not referrer.

cdumez:
LGTM otherwise.

igrigorik:
Yikes, friday.. fixed.

igrigorik:
@siusin can you please take a pass? Need your LGTM to land this.

cdumez:
Oh, I thought I could approve :) I'll let @siusin approve then.

igrigorik:
@cdumez if you have write access, you can.. use the GitHub review tool; I can't merge until I have the approve bit through GitHub :)

cdumez (approved):
LGTM

igrigorik:
@cdumez awesome, thank you!

cdumez:
Note that I am still not convinced the test is behaving as expected. The new version of the test still passes in WebKit even though I am pretty sure we are not sending an empty Content-Type for ArrayBuffer payloads.

I investigated a little in WebKit and I believe the issue is that the second test (the ArrayBuffer one) requires a CORS preflight. This is because the destination is cross-origin and the payload had a non-safelisted MIME type. I believe WebKit acts as per spec but please correct me if this is wrong.

Since the python script we're calling does not deal with OPTIONS requests, the preflight fails and we end up not sending the beacon. Apparently, when no beacon is received, the test considers the received Content-Type is the empty string so WebKit passes the check.

cdumez:
An easy fix would be to use same origin in this test since we are mostly interested in Content-Type header, not CORS. We can test CORS separately (I actually have a couple of Beacon CORS tests downstream that I could contribute upstream).

I am happy to write a PR for switching the test to same origin if you agree.

igrigorik:
@cdumez yep, that sounds reasonable, please do.. :)

cdumez:
Ok, https://github.com/w3c/web-platform-tests/pull/6793

PR #6762, opened 2017-08-04
Streams: consolidate pipeThrough tests

domenic:
This merges readable-streams/pipe-through.js into piping/pipe-through.js. "ReadableStream.prototype.pipeThrough should work generically on its this and its arguments" was merged into "pipeThrough generically calls pipeTo with the appropriate args".


sideshowbarker:
w3c-test:mirror

ricea (approved):
Sorry I lost track of this.

PR #6765, opened 2017-08-06
Remove outdated comment

ayg:
#442 was resolved a long time ago.

Protip: set `git config --global core.commentChar "auto"` to avoid having commit summaries like this eaten when you try to edit them.


annevk (approved):

annevk:
Be careful with using #... when you actually want to reference another repository. Can end up being confusing. I fixed it when landing this change.

ayg:
Oops, thanks!

PR #6792, opened 2017-08-09
Redo autogenerated metadata

### LINKED ISSUE #4210: Make .any.js tests work for service/shared workers too
Right now streams has its own hacky setup for running tests in three environments (window, dedicated worker, shared worker, service worker). Previously discussed in e.g. https://github.com/w3c/testharness.js/issues/168 (cc @youennf). It would be ideal if we could get rid of that and just use the .any.js setup. Right now we use https://github.com/w3c/web-platform-tests/blob/master/streams/resources/test-initializer.js which causes duplicate test names.

I am not sure how the service worker part of this interacts with the HTTPS requirement (which seems to involve adding .https to test filenames for some reason).

I also can't find the docs for the .any.js thing. I've had this problem before. If they could be made more prominent somehow that would be great.

annevk:
Fixes #4210.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6792)

### DIFF
--- a/tools/manifest/sourcefile.py
+++ b/tools/manifest/sourcefile.py
@@ -49,6 +49,87 @@ def read_script_metadata(f, regexp):
         yield (m.groups()[0], m.groups()[1])
 
 
+_any_variants = {
+    b"default": {"longhand": {b"window", b"dedicatedworker"}},
+    b"window": {"suffix": ".any.html"},
+    b"serviceworker": {"force_https": True},
+    b"sharedworker": {},
+    b"dedicatedworker": {"suffix": ".any.worker.html"},
+    b"worker": {"longhand": {b"dedicatedworker", b"sharedworker", b"serviceworker"}}
+}
+
+
+def get_any_variants(item):
+    """
+    Returns a set of variants (bytestrings) defined by the given keyword.
+    """
+    assert isinstance(item, binary_type), item
+    assert not item.startswith(b"!"), item
+
+    variant = _any_variants.get(item, None)
+    if variant is None:
+        return set()
+
+    return variant.get("longhand", {item})
+
+
+def get_default_any_variants():
+    """
+    Returns a set of variants (bytestrings) that will be used by default.
+    """
+    return set(_any_variants[b"default"]["longhand"])
+
+
+def parse_variants(value):
+    """
+    Returns a set of variants (bytestrings) defined by a comma-separated value.
+    """
+    assert isinstance(value, binary_type), value
+
+    globals = get_default_any_variants()
+
+    for item in value.split(b","):
+        item = item.strip()
+        if item.startswith(b"!"):
+            globals -= get_any_variants(item[1:])
+        else:
+            globals |= get_any_variants(item)
+
+    return globals
+
+
+def global_suffixes(value):
+    """
+    Yields the relevant filename suffixes (strings) for the variants defined by
+    the given comma-separated value.
+    """
+    assert isinstance(value, binary_type), value
+
+    rv = set()
+
+    global_types = parse_variants(value)
+    for global_type in global_types:
+        variant = _any_variants[global_type]
+        suffix = variant.get("suffix", ".any.%s.html" % global_type.decode("utf-8"))
+        if variant.get("force_https", False):
+            suffix = ".https" + suffix
+        rv.add(suffix)
+
+    return rv
+
+
+def global_variant_url(url, suffix):
+    """
+    Returns a url created from the given url and suffix (all strings).
+    """
+    url = url.replace(".any.", ".")
+    # If the url must be loaded over https, ensure that it will have
+    # the form .https.any.js
+    if ".https." in url and suffix.startswith(".https."):
+        url = url.replace(".https.", ".")
+    return replace_end(url, ".js", suffix)
+
+
 class SourceFile(object):
     parsers = {"html":lambda x:html5lib.parse(x, treebuilder="etree"),
                "xhtml":lambda x:ElementTree.parse(x, XMLParser.XMLParser()),
@@ -510,12 +591,17 @@ class SourceFile(object):
             rv = VisualTest.item_type, [VisualTest(self, self.url)]
 
         elif self.name_is_multi_global:
-            rv = TestharnessTest.item_type, [
-                TestharnessTest(self, replace_end(self.url, ".any.js", ".any.html"),
-                                timeout=self.timeout),
-                TestharnessTest(self, replace_end(self.url, ".any.js", ".any.worker.html"),
-                                timeout=self.timeout),
+            globals = b""
+            for (key, value) in self.script_metadata:
+                if key == b"global":
+                    globals = value
+                    break
+
+            tests = [
+                TestharnessTest(self, global_variant_url(self.url, suffix), timeout=self.timeout)
+                for suffix in sorted(global_suffixes(globals))
             ]
+            rv = TestharnessTest.item_type, tests
 
         elif self.name_is_worker:
             rv = (TestharnessTest.item_type,
--- a/docs/_writing-tests/file-names.md
+++ b/docs/_writing-tests/file-names.md
@@ -52,7 +52,7 @@ themselves precede any test type flag, but are otherwise unordered.
 
 `.any`
  : (js files only) Indicates that the file generates tests in which it
-    is run in Window and dedicated worker environments.
+    is [run in multiple scopes][multi-global-tests].
 
 `.tentative`
  : Indicates that a test makes assertions not yet required by any specification,
@@ -63,3 +63,4 @@ themselves precede any test type flag, but are otherwise unordered.
 
 
 [server-side substitution]: https://wptserve.readthedocs.io/en/latest/pipes.html#sub
+[multi-global-tests]: {{ site.baseurl }}{% link _writing-tests/testharness.md %}#multi-global-tests
--- a/docs/_writing-tests/testharness.md
+++ b/docs/_writing-tests/testharness.md
@@ -59,13 +59,11 @@ This test could then be run from `FileAPI/FileReaderSync.worker.html`.
 
 ### Multi-global tests
 
-Tests for features that exist in multiple global scopes can be written
-in a way that they are automatically run in a window scope and a
-worker scope.
+Tests for features that exist in multiple global scopes can be written in a way
+that they are automatically run in several scopes. In this case, the test is a
+JavaScript file with extension `.any.js`, which can use all the usual APIs.
 
-In this case, the test is a JavaScript file with extension `.any.js`.
-The test can then use all the usual APIs, and can be run from the path to the
-JavaScript file with the `.js` replaced by `.worker.html` or `.html`.
+By default, the test runs in a window scope and a dedicated worker scope.
 
 For example, one could write a test for the `Blob` constructor by
 creating a `FileAPI/Blob-constructor.any.js` as follows:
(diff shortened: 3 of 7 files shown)


w3c-bots:
# Build BROKEN

Started: 2017-08-09 14:04:21
Finished: 2017-08-09 14:24:14



- [WPT Results Dashboard](http://pulls.web-platform-tests.org/build/17169)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/262679047)

jgraham:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[tools/manifest/sourcefile.py, line 204 at r1](https://reviewable.io:443/reviews/w3c/web-platform-tests/6792#-Kr6PV_Dzto7YYVvTzLi:-Kr6PV_ErJxo-4nfvsRW:bvqclop) ([raw file](https://github.com/w3c/web-platform-tests/blob/9830c5a474a361e2ee71cd26d8d5f234d081545d/tools/manifest/sourcefile.py#L204)):*
> ```Python
> 
>     @property
>     def name_is_multi_global(self):
> ```

FWIW I would leave `.worker.`, `.window` and `.any.` tests working for now so you don't have to replace all of them at once.

---

*[tools/manifest/sourcefile.py, line 207 at r1](https://reviewable.io:443/reviews/w3c/web-platform-tests/6792#-Kr6PgbAsAfC9T-1i4F2:-Kr6PgbB2tNTD1rrsrmT:b-sq5gox) ([raw file](https://github.com/w3c/web-platform-tests/blob/557156e603e1141ffa61028f26b0e78b21884f9c/tools/manifest/sourcefile.py#L207)):*
> ```Python
>         """Check if the file name matches the conditions for the file to
>         be a multi-global js test file"""
>         return "global" in self.meta_flags and self.ext == ".js"
> ```

`.global.` seems like an odd flag because the technical meaning contradicts the standard meaning here (it makes it sound like the test is somehow global rather than like it runs in multiple globals.

---

*[tools/manifest/sourcefile.py, line 487 at r1](https://reviewable.io:443/reviews/w3c/web-platform-tests/6792#-Kr6Q8xhZdRm78Fwo2w7:-Kr6Q8xi0ZxtEth7xffV:bbjmssp) ([raw file](https://github.com/w3c/web-platform-tests/blob/557156e603e1141ffa61028f26b0e78b21884f9c/tools/manifest/sourcefile.py#L487)):*
> ```Python
>         elif self.name_is_global:
>             rv = TestharnessTest.item_type, []
>             valueTypeSuffixes = []
> ```

Variables should be formatted with underscores i.e. `value_type_suffix`.

---

*[tools/manifest/sourcefile.py, line 491 at r1](https://reviewable.io:443/reviews/w3c/web-platform-tests/6792#-Kr6QlFp19pzrP42dKW2:-Kr6QlFp19pzrP42dKW3:b-xafvsj) ([raw file](https://github.com/w3c/web-platform-tests/blob/557156e603e1141ffa61028f26b0e78b21884f9c/tools/manifest/sourcefile.py#L491)):*
> ```Python
>                 if key == b"global":
>                     valueTypes = value.split(b",")
>                     for valueType in valueTypes:
> ```

So it seems like this would be clearer by building up a set e.g.

    variants = {}
    value_types = sorted(item.strip() for item in value.split(","), key=lambda x:(x[0] == "!", x))
    expansions = {"workers": {"worker", "sharedworker", "serviceworker"}}
    for item in value_types:
        values = expansions.get(item, {item})
        if item.startswith("!"):
            variants -= values
        else:
            variants |= values

    for variant in variants:
        rv[1].append(TestharnessTest(self, replace_end(self.url, ".global.js", ".global.%s.html" % variant), timeout=self.timeout))

I would also put the first bit into its own method. This does mean that window tests aren't just `.html`, but you can special case that in the substitution if you like.

---

*[tools/manifest/sourcefile.py, line 508 at r1](https://reviewable.io:443/reviews/w3c/web-platform-tests/6792#-Kr6QS1kZ8TQ5kb_bbI8:-Kr6QS1l0MLhr7ggmvVE:b-agexu0) ([raw file](https://github.com/w3c/web-platform-tests/blob/557156e603e1141ffa61028f26b0e78b21884f9c/tools/manifest/sourcefile.py#L508)):*
> ```Python
> 
>             if valueTypeSuffixes == []:
>                 raise ValueError("You need to use `// META: global=...` in your resource")
> ```

This type of check should be part of the lint.

---

*[tools/serve/serve.py, line 142 at r1](https://reviewable.io:443/reviews/w3c/web-platform-tests/6792#-Kr6SYkrp19mxEJgsYUd:-Kr6SYkrp19mxEJgsYUe:b-igh5xi) ([raw file](https://github.com/w3c/web-platform-tests/blob/9830c5a474a361e2ee71cd26d8d5f234d081545d/tools/serve/serve.py#L142)):*
> ```Python
> 
> class WindowHandler(HtmlWrapperHandler):
>     path_replace = [(".window.html", ".window.js")]
> ```

Again, I would leave these ones working for now.

---

*[tools/serve/serve.py, line 142 at r1](https://reviewable.io:443/reviews/w3c/web-platform-tests/6792#-Kr6U8PRyK7SqlqnsqIZ:-Kr6U8PRyK7SqlqnsqI_:bn8hzt1) ([raw file](https://github.com/w3c/web-platform-tests/blob/557156e603e1141ffa61028f26b0e78b21884f9c/tools/serve/serve.py#L142)):*
> ```Python
> 
>     def _meta_replacement(self, key, value):
>         if key == b"global":
> ```

I'm not really sure that this isn't just adding unnecessary complexity. In particular you end up duplicating a lot of logic here, so if you really want this I would prefer some way to share the logic with `sourcefile.py.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/web-platform-tests/6792)*

jgraham:
You also need to update the lint to allow this metadata. At the moment it will fail because this isn't a meta type it knows about.

annevk:
> FWIW I would leave .worker., .window and .any. tests working for now so you don't have to replace all of them at once.

I think it would be better to do that as part of this. Having said that, it would be trivial to keep `window.js` working so I'll add that back in. Would save from adding some metadata. `worker.js` would clash with WorkerJavaScriptHandler.

> .global. seems like an odd flag because the technical meaning contradicts the standard meaning here (it makes it sound like the test is somehow global rather than like it runs in multiple globals.

Okay, in that case I suggest we stick with .any. Shall I name the META variable any too?

> So it seems like this would be clearer by building up a set

With the updates I pushed, do you still think so? (I don't think we should support standalone sharedworker for now btw. There's no demonstrated need. And !window and !worker also don't make sense.)

> I'm not really sure that this isn't just adding unnecessary complexity. In particular you end up duplicating a lot of logic here, so if you really want this I would prefer some way to share the logic with `sourcefile.py.

How is this duplicating logic? This seems like the only place that controls whether something that gets loaded actually ends up making sense.

jgraham:
Reviewed 1 of 2 files at r2, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/web-platform-tests/6792)*

jgraham:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[tools/manifest/sourcefile.py, line 496 at r3](https://reviewable.io:443/reviews/w3c/web-platform-tests/6792#-KrGoPqNm-o9sVinpXN1:-KrGoPqNm-o9sVinpXN2:bait7dm) ([raw file](https://github.com/w3c/web-platform-tests/blob/7c12b328aa8460be557c606b5512bd60d72d31e6/tools/manifest/sourcefile.py#L496)):*
> ```Python
>             for (key, value) in self.script_metadata:
>                 if key == b"global":
>                     global_values = value.split(b",")
> ```

Forgot to mention that you should `.strip()` each item here to allow whitespace in the values. You'll need to update the lint too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/web-platform-tests/6792)*

annevk:
> Forgot to mention that you should .strip() each item here to allow whitespace in the values.

Why would we want to allow whitespace?

jgraham:
I pushed some proposed changes to https://github.com/w3c/web-platform-tests/tree/meta_global

annevk:
I think the main problem is that it requires all tests to be over HTTPS if you want to include a service worker test. As long as HTTPS requires more work to get done I don't think that's a good tradeoff.

jgraham:
It doesn't require that.

w3c-bots:
# Build BROKEN

Started: 2017-08-10 13:07:29
Finished: 2017-08-10 13:22:07





- [WPT PR Status](http://pulls.web-platform-tests.org/build/17203)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/263076271)

Ms2ger:
What's the status here?

annevk:
I want to pick this up again at some point, but I haven't made time recently. If someone else would like to take over to make this happen, be my guest. Whoever picks this up should probably take @jgraham's branch (or the commits there) as a starting point.

foolip:
> I want to pick this up again at some point, but I haven't made time recently. If someone else would like to take over to make this happen, be my guest. Whoever picks this up should probably take @jgraham's branch (or the commits there) as a starting point.

Looks like the commit from is actually included in this PR, so I guess just continuing here would work?

jgraham:
Yes

foolip:
I resolved the conflicts and made sure that `wpt lint` and `wpt run` don't just crash, although `wpt lint` did complain about things that probably need fixing to land this.

@annevk @jgraham is this in a state where careful review is "all" that's needed to get it landed, or was something tricky discovered above that needs fixing first?

foolip:
OK, failing lints:
```
css/geometry/DOMMatrix-css-string.worker.js: Support file not in support directory (SUPPORT-WRONG-DIR)
css/geometry/WebKitCSSMatrix.worker.js: Support file not in support directory (SUPPORT-WRONG-DIR)
css/geometry/interfaces.worker.js: Support file not in support directory (SUPPORT-WRONG-DIR)
```

Seems like a regression in what is considered a support file.

annevk:
@foolip OP has a list of things that need to be done before this can be merged.

Ms2ger:
@jgraham @gsnedders I think this is ready for another round of reviews

tools/manifest/sourcefile.py
```diff
@@ -49,6 +49,86 @@ def read_script_metadata(f, regexp):
         yield (m.groups()[0], m.groups()[1])
 
 
+_any_variants = {
+    b"window": {"default": True, "suffix": ".any.html"},
```
Ms2ger: Thought: maybe we should instead have `b"default": {"longhand": {b"window", b"dedicatedworker"}}`, then you could use
```js
// META: global=!default,blah
```

jgraham:
This looks great, thanks for picking it up. I think this is at least somewhat documented, we should update the documentation before landing. Otherwise my only real feedback is that I prefer the design where default can be used explicitly.

---

Reviewed 5 of 5 files at r7.
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[tools/lint/lint.py, line 653 at r7](https://beta.reviewable.io/reviews/w3c/web-platform-tests/6792#-LAYlESL1Q0-dIYPGGo2:-LAYlESNCwqfdqKdSSgU:bxiem9o) ([raw file](https://github.com/w3c/web-platform-tests/blob/5a68a36f3ab3c428799a170c2ad8813905185eac/tools/lint/lint.py#L653)):*
> ```Python
> 
>                         else:
>                             included_variants = set.union(get_default_any_variants(),
> ```

I might move this out of the loop.

---

*[tools/manifest/sourcefile.py, line 53 at r7](https://beta.reviewable.io/reviews/w3c/web-platform-tests/6792#-Kr6IjJTJa_qiEbFOFB7-r7-53:-LAYoYzlC3nHpGxFmeGW:b-vswaky) ([raw file](https://github.com/w3c/web-platform-tests/blob/5a68a36f3ab3c428799a170c2ad8813905185eac/tools/manifest/sourcefile.py#L53)):*
<details><summary><i>Previously, Ms2ger wrote…</i></summary><blockquote>

Thought: maybe we should instead have `b"default": {"longhand": {b"window", b"dedicatedworker"}}`, then you could use
```js
// META: global=!default,blah
```
</blockquote></details>

I like that suggestion better.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/w3c/web-platform-tests/6792#-:-LAYq4ai4vyFjJEKNGE3:b84k5ek)*

tools/manifest/sourcefile.py
```diff
@@ -49,6 +49,86 @@ def read_script_metadata(f, regexp):
...
+def parse_variants(value):
+    """
+    Returns a set of variants (bytestrings) defined by a comma-separated value.
+    """
+    assert isinstance(value, binary_type), value
+
+    globals = get_default_any_variants()
+    global_values = set(item.strip() for item in value.split(b","))
+
+    for item in global_values:
```
Ms2ger: Note to self: check if the iteration order here is significant

tools/manifest/sourcefile.py
```diff
@@ -49,6 +49,86 @@ def read_script_metadata(f, regexp):
...
+
+
+def parse_variants(value):
+    """
+    Returns a set of variants (bytestrings) defined by a comma-separated value.
+    """
+    assert isinstance(value, binary_type), value
+
+    globals = get_default_any_variants()
+    global_values = set(item.strip() for item in value.split(b","))
```
Ms2ger: There doesn't seem to be a reason to create the set. (Could also do the `strip()` call inside the loop).

jgraham (approved):

PR #6822, opened 2017-08-11
Export "Fix flakiness in timeOrigin test" and "Allow Date.now() rounding errors in timeOrigin test"

foolip:
Putting some things here to convince Chromium's exporter that these changes are exported:

Change-Id: I9b3610a44ca5199b8a906dbe9715d280bd3439c5
Change-Id: Iba8d3c8c6889a13fddb7ae779c6029d7664cb039

### DIFF
--- a/hr-time/timeOrigin.html
+++ b/hr-time/timeOrigin.html
@@ -10,9 +10,9 @@ const windowOrigin = performance.timeOrigin;
 
 test(() => {
   const startTime = Date.now();
-  assert_greater_than_equal(startTime, windowOrigin, 'Date.now() should be at least as large as the window timeOrigin.');
+  assert_greater_than_equal(startTime + 1, windowOrigin, 'Date.now() should be at least as large as the window timeOrigin.');
   const startNow = performance.now();
-  assert_less_than_equal(startTime, windowOrigin + startNow, 'Date.now() should be close to window timeOrigin.');
+  assert_less_than_equal(startTime, windowOrigin + startNow + 1, 'Date.now() should be close to window timeOrigin.');
 }, 'Window timeOrigin is close to Date.now() when there is no system clock adjustment.');
 
 const workerScript = 'postMessage({timeOrigin: performance.timeOrigin})';
--- a/hr-time/timeOrigin.html
+++ b/hr-time/timeOrigin.html
@@ -11,18 +11,21 @@ const windowOrigin = performance.timeOrigin;
 test(() => {
   const startTime = Date.now();
   assert_greater_than_equal(startTime, windowOrigin, 'Date.now() should be at least as large as the window timeOrigin.');
-  assert_less_than(startTime - windowOrigin, 500, 'Date.now() should be close to window timeOrigin.');
+  const startNow = performance.now();
+  assert_less_than_equal(startTime, windowOrigin + startNow, 'Date.now() should be close to window timeOrigin.');
 }, 'Window timeOrigin is close to Date.now() when there is no system clock adjustment.');
 
 const workerScript = 'postMessage({timeOrigin: performance.timeOrigin})';
 const blob = new Blob([workerScript]);
 
 async_test(function(t) {
+  const beforeWorkerCreation = performance.now();
   const worker = new Worker(URL.createObjectURL(blob));
   worker.addEventListener('message', t.step_func_done(function(event) {
     const workerOrigin = event.data.timeOrigin;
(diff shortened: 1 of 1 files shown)


foolip:
This is a manual export of https://chromium-review.googlesource.com/602381 and https://chromium-review.googlesource.com/608629, the first of which is already in https://github.com/w3c/web-platform-tests/pull/6759 but cannot be merged due to flakiness. And the follow-up PR cannot be created due to conflicts, see https://bugs.chromium.org/p/chromium/issues/detail?id=754616

@jeffcarp @Hexcles FYI.

w3c-bots:
# Build PENDING

Started: None
Finished: None



- [WPT Results Dashboard](http://pulls.web-platform-tests.org/build/17263)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/263428938)

w3c-bots:
# Build PENDING

Started: None
Finished: None



- [WPT Results Dashboard](http://pulls.web-platform-tests.org/build/17263)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/263428938)

w3c-bots:
# Build PASSED

Started: 2017-08-11 10:00:47
Finished: 2017-08-11 10:12:23



- [WPT Results Dashboard](http://pulls.web-platform-tests.org/build/17263)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/263428938)

foolip:
Already reviewed, so merging manually without further review.

w3c-bots:
# Build PASSED

Started: 2017-08-11 10:00:47
Finished: 2017-08-11 10:12:23





- [WPT PR Status](http://pulls.web-platform-tests.org/build/17263)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/263428938)

hr-time/timeOrigin.html
```diff
@@ -10,19 +10,22 @@
 
 test(() => {
   const startTime = Date.now();
-  assert_greater_than_equal(startTime, windowOrigin, 'Date.now() should be at least as large as the window timeOrigin.');
-  assert_less_than(startTime - windowOrigin, 500, 'Date.now() should be close to window timeOrigin.');
+  assert_greater_than_equal(startTime + 1, windowOrigin, 'Date.now() should be at least as large as the window timeOrigin.');
+  const startNow = performance.now();
+  assert_less_than_equal(startTime, windowOrigin + startNow + 1, 'Date.now() should be close to window timeOrigin.');
```
toddreifsteck: Although true most of the time, Date.now and performance.now have no specified relationship. We should not use Date.now in these tests as a method of testing. If we do use them, we need to allow more than an offset of 1 millisecond of error.

toddreifsteck: This may be new feedback as I did not review the previous tests carefully. Sorry if this is late news.. We can address as a separate issue if you'd prefer.

toddreifsteck (commented):
Updates look good, but using Date.now to test timeOrigin is suspect as the specifications are not directly related. @igrigorik for a 2nd pair of eyes. We don't need to block this commit given the test already depends on Date.now() but could address in an issue. I'll defer to what works best for the contributors.

hr-time/timeOrigin.html
Hexcles: This was manually exported from Chromium a while ago (I just added a missing label which was probably why you noticed this PR again). Pinging the original author @npm1 for comments.

npm1: I disagree that there should be no relationship between Date.now and performance.now if there are no system clock adjustments. The monotonic clock should behave similarly to the system clock in this case. Otherwise, developers would have no idea how to quantify the numbers obtained from the monotonic clock. The spec says that for monotonic clocks the 'reference point is the Unix time'. Maybe we should make this relationship more explicit in the spec. Nonetheless, it is easy to remove Date.now from this test if needed, by just removing the first test().

toddreifsteck: @npm1 I don't disagree as a "human" :-), but the specifications don't support that tight relationship today. 
See https://github.com/w3c/hr-time/issues/27 for background

npm1: I can remove the first test then, which means, for example, a browser could always set the window timeOrigin as 0 and still pass the tests... What about this, is this also incorrect?
https://github.com/w3c/web-platform-tests/blob/8bb0b769af8d7a6d245db31c0b6029eb3b7da861/hr-time/basic.any.js#L19

PR #6826, opened 2017-08-11
Update capabilities with value returned from the remote end

jgraham:
When a new session is created the server returns a few details of the
remote end. It replaces the contents of the capabilities variable with
what is returned so that it can be used when accessing it from the
session object.

MozReview-Commit-ID: DB4iMpYMKXU

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1364389 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6826)


jgraham:
Code reviewed upstream.

PR #6830, opened 2017-08-11
Improve crash detection in wpt with e10s enabled,

jgraham:
Set MOZ_CRASHREPORTER_SHUTDOWN so that the browser terminates
immediately in the case of a content crash. This causes an early
return from executeAsyncScript, which also happens in cases with an
unhandled alert dialog, so check  if the connection is still alive to
distinguish between a crash and an error.

MozReview-Commit-ID: 6FYDwt15tI4

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1384063 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6830)

### DIFF
--- a/tools/wptrunner/wptrunner/browsers/base.py
+++ b/tools/wptrunner/wptrunner/browsers/base.py
@@ -119,6 +119,10 @@ class Browser(object):
         with which it should be instantiated"""
         return ExecutorBrowser, {}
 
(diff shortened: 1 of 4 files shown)


jgraham:
Code reviewed upstream.

PR #6834, opened 2017-08-11
Allow assert_success to be used without an expected value.

jgraham:
It happens that we receive responses that do not have a body value.
We want to be able to use assert_success with these as well.

MozReview-Commit-ID: B1f0Hn406Nj

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1381876 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6834)


jgraham:
Code reviewed upstream.

PR #6841, opened 2017-08-11
Add client.Window#state and #rect.

jgraham:
Introduces two new APIs on client.Window in the WPT WebDriver client:
client.Window#state and client.Window#rect.  The latter is used to reduce
raw calls amongst client.Window's shorthands to GET window/rect.

MozReview-Commit-ID: Kf4P2q93QaL

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1388365 [ci skip]

### DIFF
--- a/tools/webdriver/webdriver/client.py
+++ b/tools/webdriver/webdriver/client.py
@@ -239,9 +239,14 @@ class Window(object):
 
     @property
     @command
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #6881, opened 2017-08-14
Fix too few args check for static members.

davidlehn:
- Remove test that a static name is not in an instance. The same name
  could be both a static and non-static.
- Add test that static name is in the instance constructor.
- Use the correct static or non-static function with throwOrReject.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6881)

### DIFF
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -2240,16 +2240,19 @@ IdlInterface.prototype.test_interface_of = function(desc, obj, exception, expect
             {
                 assert_equals(exception, null, "Unexpected exception when evaluating object");
                 assert_equals(typeof obj, expected_typeof, "wrong typeof object");
+                var fn;
                 if (!member["static"]) {
                     if (!this.is_global() && !member.isUnforgeable) {
                         assert_inherits(obj, member.name);
                     } else {
                         assert_own_property(obj, member.name);
                     }
+                    fn = obj[member.name];
                 }
                 else
                 {
-                    assert_false(member.name in obj);
+                    assert_own_property(obj.constructor, member.name, "interface object must have static operation as own property");
+                    fn = obj.constructor[member.name];
                 }
 
                 var minLength = minOverloadLength(this.members.filter(function(m) {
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PASSED

Started: 2017-08-14 21:49:51
Finished: 2017-08-14 21:53:51



- [WPT PR Status](http://pulls.web-platform-tests.org/build/17380)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/264531322)

w3c-bots:
# Build PASSED

Started: 2017-08-14 21:49:51
Finished: 2017-08-14 21:53:51



- [WPT PR Status](http://pulls.web-platform-tests.org/build/17380)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/264531322)

resources/idlharness.js
```diff
@@ -2240,16 +2240,19 @@ IdlInterface.prototype.test_interface_of = function(desc, obj, exception, expect
...
                         assert_inherits(obj, member.name);
                     } else {
                         assert_own_property(obj, member.name);
                     }
+                    fn = obj[member.name];
                 }
                 else
                 {
-                    assert_false(member.name in obj);
+                    assert_own_property(obj.constructor, member.name, "constructor does not have property:");
```
ayg: Nit: I don't think this matches the style of other assert messages in the file.  Maybe "Interface object must have static operation as own property" or something.

ayg (changes requested):
LGTM other than style nit.

resources/idlharness.js
davidlehn: Ok.  Your suggestion is fine, I'll update the commit.  Also a bit unclear how to separate messages.  I added the trailing ":" because otherwise it just runs into the next text in the error chain.  Result I get with combo testharness and mocha I'm using would look like:
```
AssertionError: assert_own_property: Interface object must have static operation as own property: expected property "foo" missing: expected 1 to equal 0
    at ...
```

ayg: The prevailing style is to not have a trailing ":", so please omit it.  I agree this does look a bit funny, but if that should be fixed it should be fixed in testharness.js, not by adding a colon to individual asserts in test files.  (I'm thoroughly used to it and don't notice, but maybe you're right it should be fixed, because objectively it does look odd.)

davidlehn: Updated.

ayg:
LGTM.  In the future, it's easier if you push changes as new fixup commits instead of amending the existing commits, because that way we can see what was changed from the original commit.  The one who's merging the changes will normally squash them into a single commit before merging.

ayg (approved):

PR #6890, opened 2017-08-15
data: URL tests

annevk:
Tests for https://github.com/whatwg/fetch/pull/579.


annevk:
This initial commit contains tests for base64 (and makes sure they're shared with `atob()` so any new tests will cover both).

data: URL-specific tests to follow.

annevk:
Bugs:
* https://bugs.chromium.org/p/chromium/issues/detail?id=755493
* https://bugs.webkit.org/show_bug.cgi?id=175568

w3c-bots:
# Build PENDING

Started: None
Finished: None



- [WPT PR Status](http://pulls.web-platform-tests.org/build/17411)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/264695122)

w3c-bots:
# Build PENDING

Started: None
Finished: None



- [WPT PR Status](http://pulls.web-platform-tests.org/build/17411)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/264695122)

w3c-bots:
# Build PASSED

Started: 2017-08-15 10:44:32
Finished: 2017-08-15 10:54:31



- [WPT PR Status](http://pulls.web-platform-tests.org/build/17411)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/264695122)

w3c-bots:
# Build PASSED

Started: 2017-08-17 16:53:51
Finished: 2017-08-17 17:03:39


<h2>Failing Jobs</h2>
<ul>

<li>chrome:unstable</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/17517)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/265656965)

fetch/data-urls/resources/data-urls.json
```diff
@@ -0,0 +1,175 @@
...
+  ["data:;charset=,X",
+   "text/plain;charset",
+   [88]],
+  ["data:;charset,X",
+   "text/plain;charset",
+   [88]],
+  ["data:;charset=\"x\",X",
+   "text/plain;charset=\"x\"",
+   [88]],
+  ["data:;CHARSET=\"X\",X",
```
foolip: I guess this would be the place to test the questions in https://github.com/whatwg/fetch/pull/579#issuecomment-330542244? Or are such tests already here?

annevk: Yeah, a little above this one. Line 163, 166, and various tests around those too.

fetch/data-urls/resources/data-urls.json
```diff
@@ -0,0 +1,175 @@
...
+   "text/plain;charset=x",
+   [88]],
+  ["data:; charset=x,X",
+   "text/plain;charset=x",
+   [88]],
+  ["data:;charset =x,X",
+   "text/plain;charset=x",
+   [88]],
+  ["data:;charset= x,X",
+   "text/plain;charset=x",
```
annevk: This pass condition should change to preserve the space before the lowercase x.

fetch/data-urls/resources/data-urls.json
```diff
@@ -0,0 +1,175 @@
...
+   [88]],
+  ["data:;charset=,X",
```
yutakahirano: An empty string is not a valid token and hence this will result in text/plain;charset=US-ASCII, I guess.

annevk: It'll result in `text/plain`. We only use the fallback if MIME type parsing returns missing or failure which it won't for `text/plain;charset=`.

annevk:
@bzbarsky @foolip @yutakahirano any final remarks on these data: URL tests or suggestions for new tests?

fetch/data-urls/processing.any.js
```diff
@@ -0,0 +1,23 @@
+promise_test(() => fetch("resources/data-urls.json").then(res => res.json()).then(runDataURLTests), "Setup.");
+function runDataURLTests(tests) {
+  for(let i = 0; i < tests.length; i++) {
+    const input = tests[i][0];
+    let expectedMimeType = tests[i][1];
+    const expectedBody = expectedMimeType !== null ? tests[i][2] : null;
+    promise_test(t => {
+      if(expectedMimeType === "") {
```
yutakahirano: I would like to avoid such special expectations as much as possible: Can we define defaultType = "text/plain;charset=US-ASCII" and use it from the test cases?

fetch/data-urls/resources/data-urls.json
```diff
@@ -0,0 +1,175 @@
...
+   "image/png",
+   [88, 32, 88]],
+  ["data:application/xml,X X",
+   "application/xml",
+   [88, 32, 88]],
+  ["data:unknown/unknown,X X",
+   "unknown/unknown",
+   [88, 32, 88]],
+  ["data:text/plain;a=\",\",X",
+   "data:text/plain;a=\"\\\"\"",
```
yutakahirano: Is this correct?

fetch/data-urls/resources/data-urls.json
```diff
@@ -0,0 +1,175 @@
...
+   [34, 44, 88]],
+  ["data:text/plain;a=%2C,X",
+   "data:text/plain;a=%2C",
```
yutakahirano: Is this correct?

fetch/data-urls/resources/data-urls.json
annevk: So you split on the first comma. That means `"` is the value of the `a` parameter. If you serialize that the value becomes `"\""`. Add some more slashes because it's JSON and you got the above.

fetch/data-urls/resources/data-urls.json
annevk: Yeah, you split on the first comma. `%2C` is an acceptable value and does not need to be quoted.

fetch/data-urls/resources/data-urls.json
yutakahirano: I mean, is "data:" needed in the expected value?

annevk: Aah, no it's not! Thanks. Same below indeed.

fetch/data-urls/processing.any.js
```diff
@@ -0,0 +1,23 @@
...
+        expectedMimeType = "text/plain;charset=US-ASCII";
```
domenic: Normalizing this in the data files would be better and clearer. Performing computation in the tests is best avoided if possible.

domenic:
I did a first draft of a per-spec parser and got these failures. I have not yet investigated whether they are my bugs or test data bugs.

```
● data-urls.json › data:text/plain;a=",",X

  Expected value to equal:
    "text/plain;a=\"\\\"\""
  Received:
    "text/plain"

● data-urls.json › data:x;base64x,WA

  Expected value to equal:
    "text/plain"
  Received:
    "text/plain;charset=US-ASCII"

● data-urls.json › data:x;base64;x,WA

  Expected value to equal:
    "text/plain"
  Received:
    "text/plain;charset=US-ASCII"

● data-urls.json › data:x;base64=x,WA

  Expected value to equal:
    "text/plain;base64=x"
  Received:
    "text/plain;charset=US-ASCII"

● data-urls.json › data:; base64,WA

  Expected value to equal:
    "text/plain;charset=US-ASCII"
  Received:
    "text/plain"

● data-urls.json › data:;charset="x",X

  Expected value to equal:
    "text/plain;charset=\"x\""
  Received:
    "text/plain;charset=x"

● data-urls.json › data:;CHARSET="X",X

  Expected value to equal:
    "text/plain;charset=\"X\""
  Received:
    "text/plain;charset=X"
```

domenic:
Computer-determined code coverage is almost complete except for the first item below. Just looking through though, some by-inspection coverage gaps:

- Step 6 premature termination; I believe this is when there is no comma in the data URL. (Possibly worth testing encoded commas.)
- Step 9 stripping leading and trailing ASCII whitespace from mimeType never seems to be invoked. In particular how this impacts `;base64` detection and starts-with-`;` detection.
- `data:;some text` should be interpreted as text/plain. There are similar tests but they all have MIME type parameters or are base64 so they don't test the default MIME type logic.
- `data:;text/plain,foo`, e.g. a MIME type that is valid except it starts with a `;` and so goes down that different code path.

domenic:
After some inspection the above all appear to be spec or test bugs. Somehow I wrote my implementation right the first time.

- `data:text/plain;a=",",X`: the parser does not distinguish between commas inside quotes and outside. The MIME type is thus `text/plain;a="`, which gets parsed as `text/plain`. The body bytes are then `",X`, so probably the rest of the test needs adjusting. **This may be a spec bug?**
- `data:x;base64x,WA`: the string `x;base64x` does not end with `base64`, and parsing it as a MIME type fails, so we go for the default case, which is `text/plain;charset=US-ASCII`, not just `text/plain`.
- `data:x;base64;x,WA`: same as the previous one with regard to the string `x;base64;x`.
- `data:x;base64=x,WA`: same as the previous one with regard to `x;base64=x`.
- `data:; base64,WA`: the string `; base64` does not end in `;base64` because of the space. However it does start with `;`, so we prepend `text/plain` to it and attempt to parse `text/plain; base64`. This just produces the MIME type `text/plain`. **This seems like a spec bug; spaces should be allowed before base64**.
- `data:;charset="x",X`: This seems like a test bug where you've over-quoted the serialized MIME type
- `data:;CHARSET="X",X`: ditto

annevk:
> This seems like a spec bug; spaces should be allowed before base64.

Why? They're not allowed in the original RFC.

Strictly splitting on the first comma (disregarding all other data) is done to align with implementations. It's not super great, but I don't think it's worth introducing complexity for either.

foolip:
w3c-test:mirror

foolip:
I've run these in some different browsers to check for things that are obvious broken:

http://w3c-test.org/submissions/6890/fetch/data-urls/base64.any.html

All pass in Firefox Nightly, some failures in other browsers. No overlap between the failures in Chrome and Safari, which is promising, so I didn't look very hard at Edge's results.

http://w3c-test.org/submissions/6890/fetch/data-urls/processing.any.html

As expected a bit all over the place. I'll comment on the JSON data file for any interesting things I find.

fetch/data-urls/resources/data-urls.json
```diff
@@ -0,0 +1,175 @@
...
+  ["data:;base64;base64,WA",
+   "text/plain",
+   [88]],
+  ["data:x/x;base64;base64,WA",
+   "x/x",
+   [88]],
+  ["data:x/x;base64;charset=x;base64,WA",
+   "x/x;charset=x",
+   [88]],
+  ["data:x/x;base64;base64x,WA",
```
foolip: In Chrome, Edge and Firefox this fails with "assert_array_equals: lengths differ, expected 2 got 1" and in Safari it fails on the following assert, which means the body did match. I take this to mean that Chrome, Edge and Firefox treat this as base64 while Safari does not.

I don't know what to expect web compat wise, but would there be anything wrong with aligning with the majority on this point?

fetch/data-urls/resources/data-urls.json
```diff
@@ -0,0 +1,175 @@
...
+  ["data:x;base64;x,WA",
+   "text/plain",
+   [88]],
+  ["data:x;base64=x,WA",
+   "text/plain;base64=x",
+   [87, 65]],
+  ["data:; base64,WA",
+   "",
+   [87, 65]],
+  ["data:;base64;,WA",
```
foolip: This one is also "assert_array_equals: lengths differ, expected 2 got 1" in all except Safari, which then fails because its ";base64;" instead of "text/plain". Same bucket as the above comment.

fetch/data-urls/processing.any.js
annevk: Maybe, but copypasta is a risk too.

fetch/data-urls/resources/data-urls.json
```diff
@@ -0,0 +1,175 @@
...
+  ["data:x;base64x,WA",
+   "text/plain",
+   [87, 65]],
+  ["data:x;base64;x,WA",
+   "text/plain",
+   [88]],
+  ["data:x;base64=x,WA",
+   "text/plain;base64=x",
+   [87, 65]],
+  ["data:; base64,WA",
```
foolip: Trailing "; base64" is the most interesting remaining question for me. Here it would be useful to also add a "data:text/plain; base64,WA" case to isolate entirely from differences in MIME parsing. As it is this fails everywhere, and I think splitting out that difference would be useful. (When triaging failures, it'll usually make sense to look at tests that fail everywhere last.)

foolip (approved):
Did some pretty extensive manual testing and lining up of failures of the sort that https://github.com/w3c/web-platform-tests/issues/7475 should automate away.

And I can't see anything here that doesn't match what the spec says. So LGTM unless the spec changes, which I'll comment on separately.

fetch/data-urls/processing.any.js
foolip: Now I'm confused, I'm arguing in favor of keeping this as-is, isn't copypasta the least likely with the empty string as the value?

fetch/data-urls/resources/data-urls.json
```diff
@@ -106,6 +106,9 @@
   ["data:x/x;base64;base64,WA",
    "x/x",
    [88]],
+  ["data:x/x;base64;charset=x,WA",
```
foolip: @annevk, I'm pretty sure this is correct per https://mimesniff.spec.whatwg.org/#parse-a-mime-type and https://github.com/whatwg/fetch/pull/579, but please review. Pushed so I can see what all browsers do via w3c-test.org.

foolip:
w3c-test:mirror

fetch/data-urls/processing.any.js
annevk: My reply was for @domenic. I also made my comment six minutes before yours according to the GitHub timestamp on it, so not sure why it appears to be later.

fetch/data-urls/processing.any.js
```diff
@@ -0,0 +1,23 @@
...
+      }
+      if(expectedMimeType === null) {
+        return promise_rejects(t, new TypeError(), fetch(input));
+      } else {
+        return fetch(input).then(res => {
+          return res.arrayBuffer().then(body => {
+            assert_array_equals(new Uint8Array(body), expectedBody);
+            assert_equals(res.headers.get("content-type"), expectedMimeType); // We could assert this earlier, but this fails often
```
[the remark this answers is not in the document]
annevk: My thinking was that all the additional fetch() and arrayBuffer() calls are better avoided.

fetch/data-urls/processing.any.js
```diff
@@ -0,0 +1,23 @@
...
+    const expectedBody = expectedMimeType !== null ? tests[i][2] : null;
+    promise_test(t => {
+      if(expectedMimeType === "") {
+        expectedMimeType = "text/plain;charset=US-ASCII";
+      }
+      if(expectedMimeType === null) {
+        return promise_rejects(t, new TypeError(), fetch(input));
+      } else {
+        return fetch(input).then(res => {
+          return res.arrayBuffer().then(body => {
```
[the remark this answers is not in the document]
annevk: Then _res_ would no longer be in scope as is required here.

annevk:
@domenic if you have ideas for more comma tests btw let me know.

sideshowbarker:
w3c-test:mirror

annevk:
@domenic I think I addressed all your feedback now.

fetch/data-urls/resources/data-urls.json
```diff
@@ -0,0 +1,178 @@
...
+   [88]],
+  ["data:;base64,W%0CA",
+   "",
+   [88]],
+  ["data:x;base64x,WA",
+   "",
+   [87, 65]],
+  ["data:x;base64;x,WA",
+   "",
+   [88]],
```
domenic: Per my tests this should be [87, 65] like the above (doesn't end in `;base64`).

fetch/data-urls/processing.any.js
domenic: It's still really, really confusing looking at the test files and seeing empty string as the expected MIME type. It'd be quite nice if this was fixed.

If nothing else, maybe replace it with `"__default__"` or something, with a comment at the top explaining the substitution.

domenic (dismissed):

fetch/data-urls/processing.any.js
annevk: Would using null work?

domenic: null means failure, per line 6.

domenic: The larger principle here is that any substitution requires every client of these tests to read the documentation and duplicate the substitution logic themselves, instead of just being able to write an equality test.

annevk: Well, you have to read the documentation either way. I don't understand why the focus is on consumption so much as writing tests should be easy too. Maybe once this is implemented in more places and we have a good grasp on coverage we can simplify the consumption though.

domenic: The focus is on being able to read the data file and tell what the test is. Code (and tests) are read more often than they are written.

If you're just going to generate tests from logic, instead of using a data file with the correct data, you might as well just use jsdom/data-url and check that the results match, instead of using a data file at all. That's obviously more extreme, but it's the same slippery slope.

fetch/data-urls/processing.any.js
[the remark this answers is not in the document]
foolip: The "wrap in a test() trick" would be to keep everything unchanged, but here do:
`test(() => assert_equals(res.headers.get("content-type"), expectedMimeType), format_value(input) + ', content-type)`.

As I said, not my favorite thing, but it does the trick in a situation like this.

fetch/data-urls/resources/data-urls.json
```diff
@@ -0,0 +1,178 @@
...
+  ["data:%1F,%FF",
+   "",
+   [255]],
+  ["data:\u0000,%FF",
+   "",
+   [255]],
+  ["data:%00,%FF",
+   "",
+   [255]],
+  ["data:text/html  ,X",
```
foolip: Given https://github.com/whatwg/fetch/pull/579#issuecomment-358446766 I think this would be a good place to have a negative test for the odd thing that Safari now has, for example with "data:text / html,X", and asserting that it's treated as an invalid MIME type. Should I just push that?

annevk: Sure!

domenic:
After updating to https://github.com/whatwg/fetch/pull/579/commits/cedd3f39389ee258945a989e339b92c3e70fdc59 I get the following failure with the tests, which don't seem updated yet:

- `data:; base64,WA` expected "text/plain", but implementation produced "text/plain;charset=US-ASCII"

Probably more tests should be added for those spec changes, as just that one seems pretty small.

fetch/data-urls/resources/data-urls.json
foolip: Done!

foolip (approved):
I added the promised test, have reviewed the recent additions, and can't think of anything more. Great stuff!

foolip:
Note that the "data:; base64,WA" @domenic found was fixed in https://github.com/w3c/web-platform-tests/pull/6890/commits/8d4335f5dfafc039fe5b75fe412488eb40c29440

domenic:
These tests appear to all match the spec, as tested by a JS implementation of it.

I still am pretty frustrated that my feedback on using the actual MIME type instead of a magic empty string is not being taken into account in the data file; I find this kind of mix of useful, readable, expected results with magic sentinels that cause a translation layer to be invoked frustrating. I think it would be nice to listen to your test consumers, and optimize for reading and consumption (which happen many times) more than writing (which happens once).

But I won't block merging based on that. I just want to issue a final plea.

foolip:
If "__default__" is acceptable to @domenic, then that seems like an improvement to me.

domenic:
It's not as good as the actual result, as it still requires a translation layer, but at least it's less likely to be interpreted as a literal empty string.

domenic:
Some additional fun tests might be for things that Unicode-compare equal to base64, but don't ASCII-compare equal to it. I'm not sure what those would be exactly, but if I'd written `/(.*); *base6464$/i` instead of `/(.*); *[Bb][Aa][Ss][Ee]64$/`, I would have gotten this wrong.

foolip:
Is having null as a sentinel also a problem, or just changing one string to another string? Are you doing things with the JSON file other than run these tests?

Just expanding to the real value would be fine by me I guess, when there are failures the values will already have been expanded so maybe it'll be easier to find in the JSON file then.

domenic:
A sentinel for failure (viz. null) makes sense; you actually need to have your test harness behave differently.

A sentinel that is just replacing one string with another means you can't do string comparison, or read the expected results in the JSON file, but instead have to transform the test JSON file into an actually-expected-results.json before doing the comparisons.

> Are you doing things with the JSON file other than run these tests?

Well, I'm trying to read it to figure out what the results should be, while debugging test failures. Having to mentally translate empty string to mean... what was the exact string? I guess I'll go look at README.md... is a definite burden.

foolip:
I see, that sounds like a real burden worth avoiding. I originally somewhat uncharitably took the argument to be a principled one about minimizing computation in tests, even when it does make things clearer, as I think it did for me as a reviewer.

annevk:
@domenic Unicode gets percent-encoded, no? I'm happy to expand the empty string btw, once everything else is in order.

domenic:
Confirmed the new EOF cases close that coverage gap.

I guess there is no way to trigger Unicode-but-not-ASCII case insensitive equality, you're right.

domenic (approved):
Thanks for writing out the MIME type :)

fetch/data-urls/resources/data-urls.json
```diff
@@ -0,0 +1,199 @@
+[
+  ["data://test/,X",
+   "text/plain;charset=US-ASCII",
+   [88]],
+  ["data://test:test/,X",
+   null],
```
SimonSapin: Why does this test case fail? Is it the URL parser that returns an error?

SimonSapin: This URL loads an `X` document in Firefox and Chromium. (It doesn’t in Edge, but Edge appears to reject any `data:` URL with a MIME type parsing error.)

annevk: Yeah.

PR #6418, opened 2017-06-28
PaymentRequest: fixes allowpaymentrequest test errors

AmazingJaze:
Fixes test errors causing test files to fail in Edge and Chrome. 
1) The error in removing-allowpaymentrequest.https.sub.html is that the test expects the PaymentRequest Constructor to fail before allowpaymentrequest has been turned off.
2) The error in setting-allowpaymentrequest.https.sub.html is that the test expects the PaymentRequest Constructor to succeed before allowpaymentrequest has ever been turned on.

**Edit:** Also fixes an error in setting-allowpaymentrequest-timing.https.sub.html 
3) From the comments in the test it appears that the test expects to turn on the allowpaymentrequest asynchronously before the iframe can receive the postmessage. However, the error in the test is that it expects a security error to be thrown after the allowpaymentrequest attribute has been turned on. 

//CC @marcoscaceres

PR #6905, opened 2017-08-16
Revert "PaymentRequest: fixes allowpaymentrequest test errors (#6418)"

zcorpan:
Also document what removing-allowpaymentrequest.https.sub.html is doing.

See https://github.com/w3c/web-platform-tests/pull/6418#issuecomment-322732608

cc @AmazingJaze @marcoscaceres @rsolomakhin

### DIFF
--- a/payment-request/allowpaymentrequest/removing-allowpaymentrequest.https.sub.html
+++ b/payment-request/allowpaymentrequest/removing-allowpaymentrequest.https.sub.html
@@ -14,25 +14,48 @@ async_test((t) => {
   iframe.src = "https://{{domains[www1]}}:{{ports[https][0]}}" + path + "echo-PaymentRequest.html";
   iframe.onload = t.step_func(() => {
     if (i === 0) {
+      // 2. The iframe has now loaded (the first load).
+      // The allowpaymentrequest flag is set.
       iframe.allowPaymentRequest = false;
+      // 3. The allowpaymentrequest attribute has now been removed.
+      // The allowpaymentrequest flag is *still set* for the document.
     }
+    // 4. (first load) Ask the subdocument to invoke PaymentRequest and post back the result.
+    // 7. (second load) Ask the subdocument to invoke PaymentRequest and post back the result.
     iframe.contentWindow.postMessage('What is the result of new PaymentRequest(...)?', '*');
   });
 
   window.onmessage = t.step_func((e) => {
     i++;
     if (i === 1) {
+      // 5. This is the first message we receive, from the first load.
+      // Since the allowpaymentrequest flag was set, we expect success.
       assert_equals(e.data.message, 'Success', 'before navigation');
 
-      // Navigate the iframe. This will fire a second 'load' event on the iframe.
+      // 6. Navigate the iframe. This will fire a second 'load' event on the iframe.
+      // At this time, the iframe does not have an allowpaymentrequest attribute.
+      // https://html.spec.whatwg.org/#the-location-interface:dom-location-href-3
+      // https://html.spec.whatwg.org/#location-object-setter-navigate
+      // https://html.spec.whatwg.org/#location-object-navigate
+      // https://html.spec.whatwg.org/#navigate step 12
+      // https://html.spec.whatwg.org/#process-a-navigate-response step 5
+      // https://html.spec.whatwg.org/#read-html
+      // https://html.spec.whatwg.org/#initialise-the-document-object step 8
+      // https://html.spec.whatwg.org/#set-the-allow*-flags step 3 does *not* set the allowpaymentrequest flag.
       iframe.contentWindow.location.href = iframe.src + '?2';
     } else {
+      // 8. This is the second message we receive, from the second load.
+      // Since the allowpaymentrequest flag was not set, we expect an exception.
       assert_equals(e.data.message, 'Exception', 'after navigation');
       assert_array_equals(e.data.details, [true /* ex instanceof DOMException*/, 18 /* ex.code */, 'SecurityError' /* ex.name */], 'after navigation');
       t.done();
     }
   });
 
+  // 1. The browsing context for the iframe is created when the iframe is inserted to the document.
+  // https://html.spec.whatwg.org/#the-iframe-element:creating-a-new-browsing-context
+  // https://html.spec.whatwg.org/#creating-a-new-browsing-context step 11
+  // https://html.spec.whatwg.org/#set-the-allow*-flags step 3 sets the allowpaymentrequest flag.
   document.body.appendChild(iframe);
 });
 </script>
--- a/payment-request/allowpaymentrequest/removing-allowpaymentrequest.https.sub.html
+++ b/payment-request/allowpaymentrequest/removing-allowpaymentrequest.https.sub.html
@@ -13,7 +13,7 @@ async_test((t) => {
   const path = location.pathname.substring(0, location.pathname.lastIndexOf('/') + 1);
   iframe.src = "https://{{domains[www1]}}:{{ports[https][0]}}" + path + "echo-PaymentRequest.html";
   iframe.onload = t.step_func(() => {
-    if (i === 1) {
+    if (i === 0) {
       iframe.allowPaymentRequest = false;
     }
     iframe.contentWindow.postMessage('What is the result of new PaymentRequest(...)?', '*');
--- a/payment-request/allowpaymentrequest/removing-allowpaymentrequest.https.sub.html
+++ b/payment-request/allowpaymentrequest/removing-allowpaymentrequest.https.sub.html
@@ -21,6 +21,7 @@ async_test((t) => {
       // The allowpaymentrequest flag is *still set* for the document.
     }
     // 4. (first load) Ask the subdocument to invoke PaymentRequest and post back the result.
+    // (See below for steps 5 and 6.)
     // 7. (second load) Ask the subdocument to invoke PaymentRequest and post back the result.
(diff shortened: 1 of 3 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None



- [WPT PR Status](http://pulls.web-platform-tests.org/build/17452)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/265114445)

zcorpan:
In Chrome Canary with experimental web platform features enabled, in https://w3c-test.org/submissions/6905/payment-request/allowpaymentrequest/setting-allowpaymentrequest-timing.https.sub.html I get this result

Result | Test Name | Message
-- | -- | --
Fail | PaymentRequest setting allowpaymentrequest after document creation, before response | assert_array_equals: property 0, expected true but got false

Which means that it throws an exception as expected, but the exception is not an instance of `DOMException`. @rsolomakhin known?

In https://w3c-test.org/submissions/6905/payment-request/allowpaymentrequest/removing-allowpaymentrequest.https.sub.html I get this result

Result | Test Name | Message
-- | -- | --
Fail | PaymentRequest removing allowpaymentrequest after load and then navigating | assert_equals: before navigation expected "Success" but got "Exception"

@rsolomakhin do you know why this fails?

rsolomakhin:
That's not something that I was aware of. In general, as these tests get written, my assumption has been that the tests are correct and Chrome implementation needs fixing, which is what we've been doing.

That being said, the iframe implementation in PaymentRequest  has been modified by feature policy team, so I wonder whether they have more insight here.

cc @lunalu

loonybear:
What is the plan, are we removing "allowpaymentrequest" attribute?

It wouldn't affect payment request in feature policy, in fact it will make it better. Right now we have 2 special cases for "allowpaymentrequest" and "alllowfullscreen", I can just remove that part of the code.

loonybear:
I think that could be a bug that feature policy introduced:
in HTMLIFrameElement::ParseAttribute, allowfullscreen and allowpaymentrequest attribute get parsed and update feature policy corresponding to its updated value. But I don't see any code for what will happen if the two attributes are removed.

@clelland

zcorpan:
@loonybear thanks. That should not happen when the attribute is set, per spec. The state should only update on browsing context creation and again on navigation.

clelland:
In Chrome, the policy in the frame is only updated, as you say, on context creation and on navigation (except in-document navigation, of course). As far as I know, the updates that @loonybear is talking about only affect the pending state, which we do calculate, and propagate between renderers as needed, but that shouldn't take effect until the frame is navigated again. (Any other behaviour is certainly a bug)

I'm trying to understand what the actual issue with the failures is -- it looks like two different bugs, perhaps?

Test Name | Message
-- | --
PaymentRequest setting allowpaymentrequest after document creation, before response | assert_array_equals: property 0, expected true but got false

This looks like a bug, if we're just returning the wrong kind of object (an exception which isn't a DOMException)

Test Name | Message
-- | --
PaymentRequest removing allowpaymentrequest after load and then navigating | assert_equals: before navigation expected "Success" but got "Exception"

Does this mean that the test is loading a frame with allowpaymentrequest, then removing it *and navigating the frame*? I would have expected that to recalculate the policy for the new content in the frame, and I *would* expect that to throw an exception.

zcorpan:
> This looks like a bug, if we're just returning the wrong kind of object (an exception which isn't a DOMException)

Indeed. It looks like it's a `TypeError`.

> Does this mean that the test is loading a frame with allowpaymentrequest, then removing it *and navigating the frame*? I would have expected that to recalculate the policy for the new content in the frame, and I *would* expect that to throw an exception.

Yes, but we don't get that far. The assertion that fails is before navigation.

```
// 5. This is the first message we receive, from the first load.
// Since the allowpaymentrequest flag was set, we expect success.
assert_equals(e.data.message, 'Success', 'before navigation');
```

w3c-bots:
# Build ERRORED

Started: 2017-08-23 14:37:50
Finished: 2017-08-23 14:57:46


<h2>Failing Jobs</h2>
<ul>

<li>chrome:unstable</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/17452)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/265114445)

zcorpan:
The chrome run on Travis CI errored with lots of messages like
```
PID 10379 | Starting ChromeDriver 2.31.488763 (092de99f48a300323ecf8c2a4e2e7cab51de5ba8) on port 4453
PID 10379 | Only local connections are allowed.
```
Anyone know what's up? @bobholt ?

payment-request/allowpaymentrequest/setting-allowpaymentrequest-timing.https.sub.html
```diff
@@ -18,7 +18,8 @@
   });
 
   window.onmessage = t.step_func_done((e) => {
-    assert_equals(e.data.message, 'Success');
+    assert_equals(e.data.message, 'Exception');
```
edenchuang: The test is inconsistent with its description at line 7 and line 8.

edenchuang (changes requested):
Make the testing and its description be consistent.

bobholt:
zcorpan: sorry for the late reply. `PID 10379 | Only local connections are allowed.` is a standard message when we start Chromedriver, both when it works and when it doesn't, so unfortunately is a red herring. This appears to be a Chrome timeout issue we're still wrestling with. We moved Chrome back to an allowed failure. 

Re-running this job will keep the original configuration where Chrome is not an allowed failure, but adding a commit to this PR will create a new job on Travis where Chrome should be an allowed fail.

zcorpan:
@bobholt thank you, I've rebased now.

payment-request/allowpaymentrequest/setting-allowpaymentrequest-timing.https.sub.html
zcorpan: It's not, actually, but I have expanded on the comment to explain in more detail what's going on.

payment-request/allowpaymentrequest/setting-allowpaymentrequest-timing.https.sub.html
```diff
@@ -18,7 +22,8 @@
...
+    assert_array_equals(e.data.details, [true /* ex instanceof DOMException*/, 18 /* ex.code */, 'SecurityError' /* ex.name */]);
```
marcoscaceres: nit: 18/DOMException.SECURITY_ERR/
Then we probably don't need the  /* ex.name */ as what it is is fairly clear.

payment-request/allowpaymentrequest/removing-allowpaymentrequest.https.sub.html
```diff
@@ -13,26 +13,49 @@
...
-    if (i === 1) {
+    if (i === 0) {
+      // 2. The iframe has now loaded (the first load).
+      // The allowpaymentrequest flag is set.
       iframe.allowPaymentRequest = false;
+      // 3. The allowpaymentrequest attribute has now been removed.
+      // The allowpaymentrequest flag is *still set* for the document.
     }
+    // 4. (first load) Ask the subdocument to invoke PaymentRequest and post back the result.
+    // 7. (second load) Ask the subdocument to invoke PaymentRequest and post back the result.
```
marcoscaceres: Maybe add a comment saying to see below for 5, and 6?

payment-request/allowpaymentrequest/removing-allowpaymentrequest.https.sub.html
```diff
@@ -13,26 +13,49 @@
...
     iframe.contentWindow.postMessage('What is the result of new PaymentRequest(...)?', '*');
   });
 
   window.onmessage = t.step_func((e) => {
     i++;
     if (i === 1) {
+      // 5. This is the first message we receive, from the first load.
```
marcoscaceres: In the future, we should write these as a promised-message queue. Would avoid having to keep a step index.

marcoscaceres (approved):
Couple of nits.

zcorpan:
@marcoscaceres thanks, fixed.

@edenchuang can you please take another look?

payment-request/allowpaymentrequest/removing-allowpaymentrequest.https.sub.html
```diff
@@ -13,26 +13,50 @@
   const path = location.pathname.substring(0, location.pathname.lastIndexOf('/') + 1);
   iframe.src = "https://{{domains[www1]}}:{{ports[https][0]}}" + path + "echo-PaymentRequest.html";
   iframe.onload = t.step_func(() => {
-    if (i === 1) {
+    if (i === 0) {
+      // 2. The iframe has now loaded (the first load).
+      // The allowpaymentrequest flag is set.
       iframe.allowPaymentRequest = false;
+      // 3. The allowpaymentrequest attribute has now been removed.
+      // The allowpaymentrequest flag is *still set* for the document.
```
alphan102: Would it be more readable if we move "set allowPaymentRequest as false" before navigating the iframe again(step 6)?

zcorpan: I don't want to change the logic of what this test is doing. I'd like to revert an incorrect change, and this is already taking too much time. 😐

zcorpan:
If we don't hear from @edenchuang in the next few days I suggest we dismiss their review and merge this.

edenchuang (approved):

zcorpan:
Thanks!

PR #6918, opened 2017-08-17
Add Chrome support for wdspec tests

jgraham:
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/6918)

### DIFF
--- a/tools/webdriver/webdriver/client.py
+++ b/tools/webdriver/webdriver/client.py
@@ -1,8 +1,12 @@
+import json
 import urlparse
 
 import error
 import transport
 
+from mozlog import get_default_logger
+
+logger = get_default_logger()
 
 element_key = "element-6066-11e4-a52e-4f735466cecf"
 
@@ -425,7 +429,12 @@ class Session(object):
             an error.
         """
         response = self.transport.send(method, url, body)
-        value = response.body["value"]
+
+        if "value" in response.body:
+            value = response.body["value"]
+        else:
+            raise error.UnknownErrorException("No 'value' key in response body:\n%s" %
+                                              json.dumps(response.body))
 
         if response.status != 200:
             cls = error.get(value.get("error"))
--- a/tools/webdriver/webdriver/transport.py
+++ b/tools/webdriver/webdriver/transport.py
@@ -2,6 +2,8 @@ import httplib
 import json
 import urlparse
 
+import error
+
 class Response(object):
     """Describes an HTTP response received from a remote en"Describes an HTTP
     response received from a remote end whose body has been read and parsed as
@@ -27,17 +29,12 @@ class Response(object):
         # >       "application/json; charset=utf-8"
         # >    "cache-control"
         # >       "no-cache"
-        assert http_response.getheader("Content-Type") == "application/json; charset=utf-8"
-        assert http_response.getheader("Cache-Control") == "no-cache"
 
         if body:
-            body = json.loads(body)
-
-            # SpecID: dfn-send-a-response
-            #
-            # > 4. If data is not null, let response's body be a JSON Object
-            #      with a key `value` set to the JSON Serialization of data.
-            assert "value" in body
+            try:
+                body = json.loads(body)
+            except:
+                raise error.UnknownErrorException("Failed to decode body as json:\n%s" % body)
 
         return cls(status, body)
 
--- a/tools/wptrunner/wptrunner/executors/base.py
+++ b/tools/wptrunner/wptrunner/executors/base.py
@@ -1,13 +1,19 @@
 import hashlib
-import json
+import httplib
 import os
+import threading
 import traceback
+import socket
 import urlparse
 from abc import ABCMeta, abstractmethod
 
 from ..testrunner import Stop
 
 here = os.path.split(__file__)[0]
+
+# Extra timeout to use after internal test timeout at which the harness
+# should force a timeout
+extra_timeout = 5 # seconds
 
 
 def executor_kwargs(test_type, server_config, cache_manager, **kwargs):
@@ -21,6 +27,11 @@ def executor_kwargs(test_type, server_config, cache_manager, **kwargs):
 
     if test_type == "reftest":
         executor_kwargs["screenshot_cache"] = cache_manager.dict()
+
+    if test_type == "wdspec":
+        executor_kwargs["binary"] = kwargs.get("binary")
+        executor_kwargs["webdriver_binary"] = kwargs.get("webdriver_binary")
+        executor_kwargs["webdriver_args"] = kwargs.get("webdriver_args")
 
     return executor_kwargs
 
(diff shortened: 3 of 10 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None



- [WPT PR Status](http://pulls.web-platform-tests.org/build/17502)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/265551877)

tools/webdriver/webdriver/transport.py
```diff
@@ -27,17 +29,12 @@ def from_http_response(cls, http_response):
         # >       "application/json; charset=utf-8"
         # >    "cache-control"
         # >       "no-cache"
-        assert http_response.getheader("Content-Type") == "application/json; charset=utf-8"
-        assert http_response.getheader("Cache-Control") == "no-cache"
```
andreastt: The response should have these headers, so why are the asserts going away?

tools/wptrunner/wptrunner/executors/base.py
```diff
@@ -329,3 +385,95 @@ def teardown(self):
...
+    def __init__(self, executor, browser):
+        Protocol.__init__(self, executor, browser)
+        self.webdriver_binary = executor.webdriver_binary
+        self.webdriver_args = executor.webdriver_args
+        self.capabilities = self.executor.capabilities
+        self.session_config = None
+        self.server = None
+
+    def setup(self, runner):
+        """Connect to browser via the Marionette HTTP server."""
```
andreastt: s/Marionette//

tools/wptrunner/wptrunner/executors/base.py
```diff
@@ -329,3 +385,95 @@ def teardown(self):
...
+        else:
+            self.executor.runner.send_message("init_succeeded")
+
+    def teardown(self):
+        if self.server is not None and self.server.is_alive:
+            self.server.stop()
+
+    @property
+    def is_alive(self):
+        """Test that the Marionette connection is still alive.
```
andreastt: s/Marionette/remote end/

tools/wptrunner/wptrunner/executors/executorselenium.py
```diff
@@ -13,6 +13,9 @@
                    RefTestImplementation,
                    TestExecutor,
                    TestharnessExecutor,
+                   WdspecExecutor,
+                   WdspecRun,
+                   extra_timeout,
```
andreastt: Can’t see that these are used in this file?

webdriver/tests/support/fixtures.py
```diff
@@ -140,12 +158,13 @@ def session(configuration, request):
     does not demand that we start a new session per test."""
     global _current_session
     if _current_session is None:
+        print configuration
```
andreastt: Debug?

tools/webdriver/webdriver/client.py
```diff
@@ -425,7 +429,26 @@ def send_command(self, method, url, body=None):
...
         response = self.transport.send(method, url, body)
-        value = response.body["value"]
+
+        if "value" in response.body:
+            value = response.body["value"]
+        elif "error" in response.body:
+            logger.warning("'value' key not found in response but 'error' key "
+                           "found in root. This is a compliance issue, but the tests are "
+                           "working around it.")
+            value = response.body
```
andreastt: Not sure I agree with what you say in the commit message:

> ```
> Chrome currently doesn't use anything much like the spec
> protocol. Nevertheless it's useful to run higher level tests without
> failing just because we can't create a session or similar.
> ```

Why do we want to work around this instead of just failing the tests? Logging that the implementation is incorrect but not actually failing the test gives a false impression that the implementation is conformant.

webdriver/tests/support/fixtures.py
```diff
@@ -12,6 +13,20 @@
...
+    def inner(*args, **kwargs):
+        try:
+            return f(*args, **kwargs)
+        except webdriver.error.WebDriverException as e:
+            logger.warning("Ignored exception %s" % e)
+    inner.__name__ = f.__name__
+    return inner
+
+
+@ignore_exceptions
```
andreastt: I’m OK with ignoring errors in here as the tests may also have caused the driver to end up in a bad state, and the tests should do the actual assertions that the the driver is compliant.

andreastt (changes requested):

tools/webdriver/webdriver/transport.py
jgraham: Because asserting here for something that should simply be a test in the testsuite is wrong. It doesn't help anyone to ERROR on every test just because the headers are wrong, even if the implementation otherwise works.

tools/webdriver/webdriver/client.py
jgraham: Because if we fail to start a session due to non-conformance of the new-session response we can't run any tests. I don't see why we would make getting that specific detail right a prerequisite to determining interop in other areas. To be clear we should catch this error with a specific test.

tools/wptrunner/wptrunner/executors/base.py
```diff
@@ -329,3 +385,95 @@ def teardown(self):
...
+        Because the remote communication happens over HTTP we need to
+        make an explicit request to the remote.  It is allowed for
+        WebDriver spec tests to not have a WebDriver session, since this
+        may be what is tested.
+
+        An HTTP request to an invalid path that results in a 404 is
+        proof enough to us that the server is alive and kicking.
+        """
+        conn = httplib.HTTPConnection(self.server.host, self.server.port)
+        conn.request("HEAD", self.server.base_path + "invalid")
```
shs96c: Why not use the `/status` endpoint? It's designed for things like this....

tools/wptrunner/wptrunner/webdriver_server.py
```diff
@@ -125,7 +125,7 @@ def make_command(self):
 
 
 class ChromeDriverServer(WebDriverServer):
-    default_base_path = "/wd/hub"
+    default_base_path = "/"
```
shs96c: Not for this PR, but if we could specify the base path, that'd be nifty

shs96c (changes requested):

tools/webdriver/webdriver/transport.py
jgraham: Sometimes it sends plain text as an error message.

But in any case that's not what this tells us. This tells us that they set the headers as specified. It's quite possible to be sending json with the wrong headers (especially Cache-Control). There's just no way every test should fail if this is wrong vs the test for this precise thing failing.

tools/webdriver/webdriver/client.py
jgraham: geckodriver implemented a lot of the spec before it got the new session handling right. I generally dislike this idea that people shouldn't be able to make incremental fixes  in whatever order they choose to prioritise but should be forced to implement certain things to even get tests to run.

shs96c: After discussing on the IRC channel with the developer of chromedriver, we agreed that the payload should be correctly formatted.

tools/webdriver/webdriver/transport.py
shs96c: After discussing on IRC with the developer of the chromedriver, we agreed to ignore the headers here, provided there's a test for them elsewhere.

shs96c (changes requested):

tools/wptrunner/wptrunner/executors/base.py
jgraham: It possibly didn't exist at the time, and requiring that to be correctly implemented for any tests to run seems like a bad idea. Anyway this code didn't change here, it just moved, so I don't plan to change it in this PR.

jgraham:
FTR Edge also has issues with the `new session` command, so unless they plan to fix that soon, I think we want the same kind of hacks in to allow them to run other tests.

tools/wptrunner/wptrunner/executors/base.py
andreastt: That’s correct, at the time when I wrote this `/status` didn’t exist. I agree we shouldn’t change this proven-to-work code in this patch.

tools/wptrunner/wptrunner/webdriver_server.py
andreastt: You can specify it, this is just the default.

andreastt (approved):

jgraham:
I'm reasonably sure the instability here is pre-existing.

PR #6969, opened 2017-08-22
[css-writing-modes-3] Test orthogonal table cells

frivoal:
A while back I reported to the CSS WG that we had issues with orthogonal table cells, and I was tasked with writing a TC about it. Here it is.

As of today:
* It passes in Edge
* It fails in Chrome and Safari, because they do not apply the writing-mode property to table cells
* It randomly passes or fail in Firefox, and reloading will randomly give different results. When it fails, Firefox does apply the writing-mode property to the table cell, but fails to size it properly. 

Note: I wrote this test based on what I understand table layout to be. I am not entirely sure it is fully defined for orthogonal flows, but this seems like the right answer. If it isn't let me know. Regardless, we still have an interop issue.


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/17663)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/267213114)

css/css-writing-modes-3/table-cell-001.html
```diff
@@ -0,0 +1,46 @@
...
+<meta charset="utf-8">
+<title>CSS Writing Mode test: orthogonal table cell</title>
+<link rel="help" href="https://drafts.csswg.org/css-writing-modes-3/#block-flow">
+<meta assert="The writing mode property applies to table cells: two adjacent cells can have orthogonal writing modes, and the enclosing table must be laid out correctly.">
+<meta flags="ahem">
+<link rel="match" href="reference/table-cell-001-ref.html">
+<link rel="author" title="Florian Rivoal" href="http://florian.rivoal.net/">
+
+<style>
+@font-face {
```
gsnedders: You don't need to include Ahem, it exists by definition in the testsuite.

frivoal: I usually don't include it, and then people complain that the test looks weird when they look at it on a computer they didn't set up properly, such as edge in browser stack, so I decided to play it safe and include the font.

Also, there are a bunch of version of ahem around, and it seems sane to require one that is known to work with the test (even though this particular test should work with any ahem).

Should I remove it?

gsnedders: Yes, please remove it.

Where are the other versions of Ahem around? We should only have `css/fonts/ahem/ahem.ttf` nowadays (`css/fonts/ahem-extra` is… different).

FremyCompany (commented):
I think browsers do not follow the spec when the orthogonal content is allowed+forced to wrap; could you investigate that case?

frivoal:
@FremyCompany Just pushed an other test case, where Firefox fails to wrap wrappable lines in an orthogonal table cell (and Blink/Webkit fail to make it orthogonal). Is that what you were talking about?

FremyCompany:
@frivoal Yes, that's what I meant. There is also [a case](https://wptest.center/#/json:{%22title%22:%22UntitledTest%22,%22html%22:%22%3Ctable%3E%3Ctd%3Exx%20xx%3C/td%3E%3C/table%3E%22,%22css%22:%22table%20{\r\n\theight:%20100px;\r\n\tborder-spacing:%200;\r\n}\r\n\r\ntd%20{\r\n\tpadding:%200;\r\n\tfont:%2050px%20/%2050px%20ahem;\r\n\tcolor:%20green;\r\n\twriting-mode:%20vertical-rl;\r\n\tbackground:%20red;\r\n}%22,%22jsBody%22:%22%22,%22jsHead%22:%22%22,%22watches%22:[],%22watchValues%22:[]}) where I think maybe Edge is also failing (but in a different way as Firefox), though I am not entirely sure it is fully defined in any spec and (if it is) whether the way it is currently defined makes sense at all. 

---------------------------------------

Here is what makes me think Edge could be right based on the current tables-spec but I'm not entirely sure. What is your opinion on that test case?

image

frivoal:
@FremyCompany As written, I agree that the text you quoted seems to support Edge's behavior, but whether that's intentional, or whether the text should say "block size" where it says "height", I cannot tell.

Also, for the test you asked me to add (table-cell-002.html), it seems good to me, but I am also insufficiently versed in css21 and css-tables to figure out exactly which spec (if any) supports it. Hopefully there's something, and if not, we need something, but I haven't figured it out yet.

Should we raise this as an issue/question against css-writing-modes / css-tables / css21, and see if the WG thinks the wording in the various specs is both clear enough and intentional, or if we need to clarify something?

@fantasai, your input on this would be appreciated.

FremyCompany:
I wouldn't mind this being discussed/reviewed by more people indeed

css/css-writing-modes-3/reference/table-cell-001-ref.html
```diff
@@ -0,0 +1,23 @@
...
+  z-index: -1;
+  background: green;
+  height: 4em;
+  width: 5em;
+}
+
+
+</style>
+
+<p>There should be a <strong>green recangle</strong> and <strong>no red</strong>.
```
dbaron: recangle -> rectangle

css/css-writing-modes-3/table-cell-001.html
```diff
@@ -0,0 +1,42 @@
...
+  z-index: -1;
+  background: red;
+  height: 4em;
+  width: 5em;
+}
+
+
+</style>
+
+<p>There should be a <strong>green recangle</strong> and <strong>no red</strong>.
```
dbaron: recangle -> rectangle

css/css-writing-modes-3/table-cell-001.html
```diff
@@ -0,0 +1,42 @@
...
+
+<div id=red></div>
+
+<table><td id=t1>&nbsp;&nbsp;&nbsp;&nbsp;</td><td id=t2>&nbsp;&nbsp;&nbsp;&nbsp;</td></table>
```
dbaron: better to use a &lt;tr> here rather than testing parser error handling

css/css-writing-modes-3/table-cell-002.html
```diff
@@ -0,0 +1,32 @@
...
+  font: 50px / 50px ahem;
+  color: green;
+  writing-mode: vertical-rl;
+  background: red;
+}
+</style>
+
+<p>There should be a <strong>green square</strong> and <strong>no red</strong>.
+
+<table><td>xx xx</td></table>
```
dbaron: same here, re &lt;tr>

dbaron (commented):
So I don't see any nondeterminism when I load the tests in Firefox.  But that might be because the tests don't actually link to the Ahem font with an @font-face rule.  Isn't that the normal practice for tests that use Ahem?  I think the tests also fail in unexpected ways because they're not actually using Ahem.

In any case, once I manually add an @font-face rule that links to the Ahem font, I do see the nondeterminism that you mention.  But note that changing the zoom clears up the nondeterminism in favor of making the first test pass.

On the other hand, the second test fails for me by having the square being split into two parts, and *sometimes* (always, after changing zoom) showing red between the two parts.

I'll try to investigate a bit further.

Also, a few comments on the tests themselves...

jgraham:
>  But that might be because the tests don't actually link to the Ahem font with an @font-face rule. Isn't that the normal practice for tests that use Ahem? I think the tests also fail in unexpected ways because they're not actually using Ahem.

You don't need to use `@font-face` for ahem in web-platform tests, it's assumed to be avaiable (that was historically the case for CSS tests, but not Mozilla reftests). Moreover we actually install it into the test environment (on supported platforms) if you pass `--install-fonts` to wptrunner, so this will start being true on travis, on the results dashboard, and in Mozilla CI.

dbaron:
I think the nondeterminism may well be related to intrinsic widths being dependent on layout.  I'm not planning to investigate further.

(Also, requiring test compilation or setup steps makes tests harder for developers to use.  It may be fine for automation systems, but it detracts from actual bugs being fixed.)

css/css-writing-modes-3/table-cell-001.html
FremyCompany: Technically the parser also adds <tbody> though if you want to go there

frivoal:
> (Also, requiring test compilation or setup steps makes tests harder for developers to use. It may be fine for automation systems, but it detracts from actual bugs being fixed.)

I've been instructed during code reviews to remove `@font-face` links to Ahem. I'd be happy to start putting them back in, but we need to have a consistent policy.

> So I don't see any nondeterminism when I load the tests in Firefox. [...] In any case, once I manually add an `@font-face` rule that links to the Ahem font, I do see the nondeterminism that you mention.

I seem to remember seeing the nondeterminism without the `@font-face`, but I no longer can either. Not sure if I am misremembering, or if something has changed in a recent update of FF. Either way, I agree that adding the following to the test brings back the non determinism.

```css
@font-face {
	font-family: ahem;
	src: local("ahem");
}
```

> I think the nondeterminism may well be related to intrinsic widths being dependent on layout. I'm not planning to investigate further.

I am somewhat confused. You of course do what you want with FF bug investigations, but I don't understand why the `@font-face` above should make any difference to the results, nor why having intrinsic with being dependent on layout cause non determinism, or be acceptable.

dbaron:
The `@font-face` makes a difference because the test sizes totally differently without Ahem.

The intrinsic width being dependent on layout causes non-determinism because whoever implemented it didn't redesign enough of the layout engine (a lot) in order to implement that dependency correctly, because many existing engines were designed around the assumption of not having that dependency.  Fixing that is something that I think I'm leaving to the next generation of layout engine developers.

dbaron:
And the reason I prefer the `@font-face` in is that part of the design of the reftest format is that the tests can just be loaded in any existing browser UI (e.g., a tab) without using any special testing infrastructure.  It's especially painful if I'm trying to look at tests that are in pull requests or review requests -- review tools generally give me a good way to download the test files and look at them, but building them as part of some test infrastructure is much more work.

frivoal:
> The @font-face makes a difference because the test sizes totally differently without Ahem.

Sure, but I have Ahem installed on my computer, so why is
```css
font-family: ahem;
```
 different from
```css
@font-face {
  font-family: ahem;
  src: local("ahem");
}
font-family: ahem;
```
Both are only working with local fonts...

> The intrinsic width being dependent on layout causes non-determinism because [...]

Understood. But this seems to be a bad news from the point of view of the writing-mode property applying (usefully) to table cells. Does this mean that Gecko just won't be able to do that nicely, or am I missing something and the consequence isn't that bad?

> And the reason I prefer the @font-face in is that part of the design of the reftest format is that the tests can just be loaded in any existing browser UI

I agree. This somehow seems to be against the current policy. Should we change it?

svgeesus (approved):
Reviewed by several others. lgtm. 
Make a new issue for whether all Ahem-using tests should be converted to @font-face.

Issue #6611, opened 2017-07-24
RTCPeerConnection-addIceCandidate.html causes race conditions


nils-ohlmeier:
RTCPeerConnection-addIceCandidate.html has several tests in it where the sequence is something like this:

`let ret = pc.addIceCandiddate();
 pc.close();
expect ret == reject`

Especially this test here: https://github.com/w3c/web-platform-tests/blob/master/webrtc/RTCPeerConnection-addIceCandidate.html#L574
is prone to test failures depending on the speed on the executing machine as it uses a timeout on top of the above sequence.

This is causing test failures in Mozilla's CI environment: https://bugzilla.mozilla.org/show_bug.cgi?id=1382972

I think these tests should be changed so that they look something like this:

`pc.close()
waitForSignalingStatetoReach(closed);
let ret = pc.addIceCandidate()
expect ret == rejected`

So you close the PeerConnection. Wait for the singaling state to actually reach 'closed' and then do what ever operation you need to check for against a closed PeerConnection.

soareschen:
The original idea for the test was to test for the step _"1. If connection's [[isClosed]] slot is true, abort these steps."_ in _section 4.3.1.2 Enqueue an operation_. Because if that step was to reach first, the body of `addIceCandidate()` wouldn't have reach, so no iceCandidate would have been added and the promise wouldn't have resolve.

There were some discussions about race conditions of enqueue operations in w3c/webrtc-pc#1218. The consensus is that some of the steps in the specs are specifically crafted to cater for legacy behavior of current implementations.

Given that it is difficult to reason and test statements like "If connection's [[isClosed]] slot is true, then abort these steps.", I will remove the relevant tests until we have figured better ways to test such statements.

jan-ivar:
FWIW I think it should be possible to test for [[isClosed]] in the common cases where the queue is empty. The only non-determinism would be invoking `addIceCandidate` from an `onicecandidate` callback, as those may happen at random.

soareschen:
@jan-ivar When I have time I will read through the steps again and think of proper ways to test them. Will ask for help from you and @nils-ohlmeier to review the PR since we seems to have some different understanding of the semantics for enqueue operation and concurrent steps execution.

PR #7001, opened 2017-08-24
setup onnegotiationneeded before checking it

nils-ohlmeier:
It look like @soareschen had a little misconception here: browsers do not store events for ever until the JS application adds a callback function. If there is no callback function registered at time of the event the event goes to void.

The other problem this fixes is that if you just call CreateOffer() without attaching any track/stream and without any options it results in the browser refusing to anything with an empty offer.

The current tests are causing intermittent test failures in Firefox Ci system, see https://bugzilla.mozilla.org/show_bug.cgi?id=1386143

The last test "negotiationneeded event should fire only after signaling state go back to stable" in here is still causing failures on Firefox although it should pass. I gave up trying to fix it.

### DIFF
--- a/webrtc/RTCPeerConnection-onnegotiationneeded.html
+++ b/webrtc/RTCPeerConnection-onnegotiationneeded.html
@@ -82,25 +82,18 @@
    */
   promise_test(t => {
     const pc = new RTCPeerConnection();
-    const promise = awaitNegotiation(pc);
-    pc.createDataChannel('test');
-    return promise;
-  }, 'Creating first data channel should fire negotiationneeded event');
+    const negotiated = awaitNegotiation(pc);
 
-  promise_test(t => {
-    const pc = new RTCPeerConnection();
     pc.createDataChannel('test');
-    // Attaching the event handler after the negotiation-needed steps
-    // are performed should still receive the event, because the event
-    // firing is queued as a task
-    return awaitNegotiation(pc);
-  }, 'task for negotiationneeded event should be enqueued for next tick');
+    return negotiated;
+  }, 'Creating first data channel should fire negotiationneeded event');
 
   test_never_resolve(t => {
     const pc = new RTCPeerConnection();
-    pc.createDataChannel('foo');
+    const negotiated = awaitNegotiation(pc);
 
-    return awaitNegotiation(pc)
+    pc.createDataChannel('foo');
+    return negotiated
       .then(({nextPromise}) => {
       pc.createDataChannel('bar');
       return nextPromise;
@@ -121,8 +114,10 @@
    */
   promise_test(t => {
     const pc = new RTCPeerConnection();
+    const negotiated = awaitNegotiation(pc);
+
     pc.addTransceiver('audio');
-    return awaitNegotiation(pc);
+    return negotiated;
   }, 'addTransceiver() should fire negotiationneeded event');
 
   /*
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/17747)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/268176037)

soareschen:
Thanks @nils-ohlmeier for the report and patch. I understand the part about browser events fire only to handlers attached at the time. I have read through the specs again and I think the part that I got wrong is on the concept of "queue a task".

My original assumption for the tests is that when a "queue a task" step is invoked, the steps inside the queued task should be executed only during the next turn of the event loop, which is until the synchronous part of the JavaScript test code returns. But I just read the HTML spec again on the section of task queue says: (http://w3c.github.io/html/webappapis.html#queuing)

> Each task is defined as coming from a specific task source. All the tasks from one particular task source and destined to a particular event loop (e.g., the callbacks generated by timers of a Document, the events fired for mouse movements over that Document, the tasks queued for the parser of that Document) must always be added to the same task queue, but tasks from different task sources may be placed in different task queues.

In other words, the task source for WebRTC tasks MAY not be the same as the task source for the main JavaScript code.

I guess what happened is that Firefox has a different task source for WebRTC, while Chrome might be using the same task source for both WebRTC and JS. The tests pass on Chrome and so I thought it is just a bug for Firefox.

@nils-ohlmeier please correct me if I still have any misconception this time. I think there may be other places in the tests that test this wrong behavior of "queue a task". I will try and find them all and fix them soon. Meanwhile if you find any other intermittent test failures, please do report them here so that I can go and fix them.

I will look at the last test and try to fix it after this PR is merged. I will also refactor the other tests to use `async_test` instead of `promise_test`, since wrapping events as promises turn out to be not so good idea.

soareschen (approved):

jan-ivar:
> My original assumption for the tests is that when a "queue a task" step is invoked, the steps inside the queued task should be executed only during the next turn of the event loop

@soareschen correct, inasmuch as it won't execute any sooner. E.g.
```js
pc.addTransceiver('audio');
pc.onnegotiationneeded = () => console.log("audio transceiver added"):
```
should work.

> the task source for WebRTC tasks MAY not be the same as the task source for the main JavaScript code.

Tasks still won't interrupt the current run of the event loop. @nils-ohlmeier this is [a bug](https://bugzil.la/1394072) in Firefox.

soareschen:
> @soareschen is correct, inasmuch as it won't execute any sooner. E.g.

@jan-ivar hmm I am not very sure where the actual problem lies. It seems that this is a similar problem to #6611. In both cases the tests try to test the "queue a task" semantics by racing some mutation synchronously, and in both cases they are causing flaky tests in Firefox.

If there are more flaky tests found, we might want to discuss whether "queue a task" should be rigorously tested, and whether there is a bug in Firefox.

jan-ivar:
@soareschen They're not flaky, they're always wrong. See my explanation [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1394072#c2). Yes we should test.

jan-ivar:
Note that "queue a task" and RTCPeerConnection's [enqueue an operation](https://w3c.github.io/webrtc-pc/#enqueue-an-operation) are totally different things.

PR #7010, opened 2017-08-25
Test behavior of setSinkId (automatic and manual)

dontcallmedom:
@guidou want to review? (chrome passes all the tests, congrats \o/)

### DIFF
--- a/audio-output/setSinkId-manual.https.html
+++ b/audio-output/setSinkId-manual.https.html
@@ -1,0 +1,132 @@
+<!doctype html>
+<html>
+<head>
+<title>Test setSinkId behavior with permissions / device changes</title>
+<link rel="author" title="Dominique Hazael-Massieux" href="mailto:dom@w3.org"/>
+<link rel="help" href="https://www.w3.org/TR/audio-output/#dom-htmlmediaelement-setsinkid">
+<meta name="timeout" content="long">
+
+</head>
+<body>
+<h1 class="instructions">Description</h1>
+<p class="instructions">This test checks that <code>setSinkId</code> follows the algorithm, this includes manually checking the proper rendering on new output devices.</p>
+<p class="instructions">When prompted to access microphones, please accept as this is the only current way to get permissions for associated output devices.</p>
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/17779)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/268397678)

audio-output/setSinkId-manual.https.html
```diff
@@ -0,0 +1,132 @@
...
+    span.textContent = device.label;
+    label.appendChild(input);
+    label.appendChild(span);
+    li.appendChild(label);
+    authorized.appendChild(li);
+};
+
+const authorizeDeviceTester = (t) => (e) => {
+    const groupId = e.target.dataset["groupid"];
+    navigator.mediaDevices.getUserMedia({audio: {groupId}})
```
guidou: I'm not sure that getting permission with getUserMedia should be part of the WPT test for setSinkId. 
In Chrome, gUM authorizes output devices although I see that as a Chrome-specific implementation detail. It's certainly not something in any spec. 
The spec-compliant way to get permission to use the device would be to use the permissions API, but that does not work in Chrome as of now.

dontcallmedom: The spec says
>  implementations MUST support implicit consent via the getUserMedia() permission prompt; when an audio input device is authorized and opened via getUserMedia(), this also authorizes access to any associated audio output devices (i.e., those with the same groupId)
so testing it seems reasonable to me

guidou: You are correct. 
My concern was about output devices that have no associated microphone and are thus not required to be authorized by getUserMedia, but there is currently no standard programmatic way to authorize those anyway, although Chrome authorizes them as well with gUM. I thought the permissions API would allow a standard way for those, but I just found out that's not the case.
So, this test LGTM.

guidou (approved):

PR #7025, opened 2017-08-28
Make use of RTCPeerConnection.prototype.{getConfiguration, setConfiguration} defensive

rwaldron:

### DIFF
--- a/webrtc/RTCConfiguration-iceCandidatePoolSize.html
+++ b/webrtc/RTCConfiguration-iceCandidatePoolSize.html
@@ -26,6 +26,7 @@ dictionary RTCConfiguration {
 */
 test(() => {
   const pc = new RTCPeerConnection();
+  assert_idl_attribute(pc, "getConfiguration");
   assert_equals(pc.getConfiguration().iceCandidatePoolSize, 0);
 }, "Initialize a new RTCPeerConnection with no iceCandidatePoolSize");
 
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/17830)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/269233631)

rwaldron:
Before:



After:

soareschen:
Hmm why does it throw `InvalidAccessError` on your side. I thought it would throw something like "pc.setConfiguration is not a function". It is also weird that testharness doesn't catch the error and just abort remaining tests.

webrtc/RTCConfiguration-iceCandidatePoolSize.html
```diff
@@ -63,17 +66,22 @@
 /*
 Reconfiguration
 */
-const pc = new RTCPeerConnection({});
 
 test(() => {
+  const pc = new RTCPeerConnection({});
```
soareschen: `pc` can be constructed with no argument, i.e. `new RTCPeerConnection()`

webrtc/RTCConfiguration-iceCandidatePoolSize.html
```diff
@@ -88,8 +96,9 @@
 */
 
 test(() => {
+  const pc = new RTCPeerConnection({});
+  assert_idl_attribute(pc, "setConfiguration");
   assert_equals(typeof pc.setConfiguration, "function", "RTCPeerConnection.prototype.setConfiguration is not implemented");
```
soareschen: This become redundant with the `assert_idl_attribute` above. We probably don't want to assert everything too defensively.

soareschen (changes requested):

soareschen (approved):

PR #7035, opened 2017-08-29
Streams: Check constructor and method names

ricea:
Check the .name properties of constructors and methods for the Streams API.

Also add properties tests for ReadableByteStreamController,
ReadableStreamBYOBReader and ReadableStreamBYOBRequest.

### DIFF
--- a/streams/readable-byte-streams/properties.js
+++ b/streams/readable-byte-streams/properties.js
@@ -1,0 +1,147 @@
+'use strict';
+
+if (self.importScripts) {
+  self.importScripts('../resources/rs-utils.js');
+  self.importScripts('/resources/testharness.js');
+}
+
+let ReadableStreamBYOBReader;
+
+test(() => {
+
+  // It's not exposed globally, but we test a few of its properties here.
+  ReadableStreamBYOBReader = (new ReadableStream({ type: 'bytes' }))
+      .getReader({ mode: 'byob' }).constructor;
+
+}, 'Can get the ReadableStreamBYOBReader constructor indirectly');
+
+test(() => {
+
+  assert_throws(new TypeError(), () => new ReadableStreamBYOBReader('potato'));
+  assert_throws(new TypeError(), () => new ReadableStreamBYOBReader({}));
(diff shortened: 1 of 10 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/17861)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/269548124)

streams/readable-byte-streams/properties.js
```diff
@@ -0,0 +1,147 @@
...
+      controller = c;
+    },
+    type: 'bytes'
+  });
+  const proto = Object.getPrototypeOf(controller);
+
+  assert_array_equals(Object.getOwnPropertyNames(proto).sort(), properties);
+
+  for (const m of methods) {
+    const propDesc = Object.getOwnPropertyDescriptor(proto, m);
```
domenic: Lots of duplication here and elsewhere, but meh...

domenic (approved):
Setting aside pre-existing duplication issues, LGTM.

streams/readable-byte-streams/properties.js
ricea: I know. I wanted to use the structure I used in [writable-streams/properties.js](https://github.com/w3c/web-platform-tests/blob/master/streams/writable-streams/properties.js) but the need to fetch ReadableStreamBYOBRequest asynchronously meant it wouldn't work directly. So then I just decided to go ahead and land this before the scope creep got any worse.

ricea:
I'm a bit worried that the Travis failures for Safari and Edge might mean the test never completes on those browsers. Does w-p-t have a way of dealing with that if I merge this change?

domenic:
My interpretation was that the WPT infra is just broken and not running tests in any browsers. (Since Chrome is not working either.) Let me ask in #testing IRC to see what's up.

domenic:
gsnedders helpfully said

> Chrome looks like the same old probably-ChromeDriver issue, Safari/Edge look like Sauce being down again

So I think we can merge this.

PR #7037, opened 2017-08-29
Add initial payment handler tests.

rsolomakhin:
This patch adds initial WebIDL interface tests for payment handler:

https://w3c.github.io/payment-handler/#idl-index

The partial interfaces for ServiceWorkerRegistration and
ServiceWorkerGlobalScope are omitted.

### DIFF
--- a/payment-handler/interfaces.https.any.js
+++ b/payment-handler/interfaces.https.any.js
@@ -1,0 +1,20 @@
+// META: script=/resources/WebIDLParser.js
+// META: script=/resources/idlharness.js
+
+"use strict";
+
+if (self.importScripts) {
+    importScripts("/resources/testharness.js");
+    importScripts("/resources/WebIDLParser.js", "/resources/idlharness.js");
+}
+
+// https://w3c.github.io/payment-handler/
+
+promise_test(async() => {
+    const text = await fetch("/interfaces/payment-handler.idl")
+        .then(response => response.text());
+    const idlArray = new IdlArray();
+    idlArray.add_idls(text);
+    idlArray.test();
+    done();
+}, "Payment handler interfaces.");
--- a/interfaces/payment-handler.idl
+++ b/interfaces/payment-handler.idl
@@ -1,0 +1,53 @@
+[SecureContext,
+ Exposed=(Window,Worker)]
+interface PaymentManager {
+    [SameObject]
+    readonly attribute PaymentInstruments instruments;
+    [Exposed=Window] static Promise<PermissionState> requestPermission();
+};
+interface PaymentInstruments {
+    Promise<boolean>             delete(DOMString instrumentKey);
+    Promise<PaymentInstrument>   get(DOMString instrumentKey);
+    Promise<sequence<DOMString>> keys();
+    Promise<boolean>             has(DOMString instrumentKey);
+    Promise<void>                set(DOMString instrumentKey,
+                                     PaymentInstrument details);
+    Promise<void>                clear();
+};
+dictionary PaymentInstrument {
+    required DOMString             name;
+             sequence<ImageObject> icons;
+             sequence<DOMString>   enabledMethods;
+             object                capabilities;
+};
+dictionary ImageObject {
+    required USVString src;
+             DOMString sizes;
+             DOMString type;
+};
+[Constructor(DOMString type, PaymentRequestEventInit eventInitDict),
+ Exposed=ServiceWorker]
+interface PaymentRequestEvent : ExtendableEvent {
+    readonly attribute USVString                           topLevelOrigin;
+    readonly attribute USVString                           paymentRequestOrigin;
+    readonly attribute DOMString                           paymentRequestId;
+    readonly attribute FrozenArray<PaymentMethodData>      methodData;
+    readonly attribute object                              total;
+    readonly attribute FrozenArray<PaymentDetailsModifier> modifiers;
+    readonly attribute DOMString                           instrumentKey;
+    Promise<WindowClient?> openWindow(USVString url);
+    void                   respondWith(Promise<PaymentHandlerResponse> handlerResponse);
+};
+dictionary PaymentRequestEventInit : ExtendableEventInit {
+    USVString                        topLevelOrigin;
+    USVString                        paymentRequestOrigin;
+    DOMString                        paymentRequestId;
+    sequence<PaymentMethodData>      methodData;
+    PaymentCurrencyAmount            total;
+    sequence<PaymentDetailsModifier> modifiers;
+    DOMString                        instrumentKey;
+};
+dictionary PaymentHandlerResponse {
+    DOMString methodName;
+    object    details;
+};
--- a/payment-handler/OWNERS
+++ b/payment-handler/OWNERS
@@ -1,0 +1,2 @@
+@marcoscaceres
+@rsolomakhin


rsolomakhin:
@marcoscaceres @ianbjacobs FYI

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/17869)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/269669713)

payment-handler/interfaces.https.html
```diff
@@ -0,0 +1,81 @@
...
+dictionary PaymentHandlerResponse {
+    DOMString methodName;
+    object    details;
+};
+</script>
+
+<script>
+"use strict";
+var idlArray = new IdlArray();
+[].forEach.call(document.querySelectorAll("script[type=text\\/plain]"), function(node) {
```
marcoscaceres: `[].forEach.call`? Just do:

```
Array.from(document.querySelectorAll("script[type=text\\/plain]")).forEach(node => {

});
```

marcoscaceres (approved):
Small nit. I gather you probably copy/pasted the runner code at the end... but would be great to fix that :)

payment-handler/interfaces.https.html
rsolomakhin: @marcoscaceres: Done. Please take another look.

marcoscaceres (approved):

rsolomakhin:
Added @rsolomakhin and @marcoscaceres to OWNERS file.

marcoscaceres:
@zcorpan, we are starting this new suite, but I don't know enough about how the IDL harness works. Can you check if this is right and would appreciate guidance (e.g., should we do a "historical" file also, like you added for payment request?)

zcorpan:
It's better to put IDL files under `/interfaces/` and then use `fetch()` to get the dependencies and the IDL to test.

If instances can be created, those should be provided to idlharness.

Since this has things `Exposed=(Window,Worker)`, it would be good to test in both contexts.

For an example to copy from, see

/interfaces/geometry.idl
/css/geometry-1/support/interfaces.js
/css/geometry-1/interfaces.html
/css/geometry-1/interfaces.worker.js

Actually, the interfaces.html and interfaces.worker.js could be just one file, interfaces.any.js. Like /url/interfaces.any.js but then that one embeds its IDL inline for some reason.

As for historical, that is only necessary when some API has been removed or renamed in the spec after it has been implemented, to make sure it doesn't linger around in implementations.

marcoscaceres:
> It's better to put IDL files under /interfaces/ and then use fetch() to get the dependencies and the IDL to test.

Agree... this already bit us on PaymentRequest. I'll file a bug for that one too.

marcoscaceres:
@rsolomakhin, something like this should work:

```JS
async function fetchIDL(specURL){
  const doc = document.implementation.createHTMLDocument("");
  doc.documentElement.innerHTML = await fetch(specURL).then(r => r.text());
  const idl = doc.getElementById("actual-idl-index").textContent;
  const idlArray = new IdlArray();
  idlArray.add_untested_idls(idl);
  idlArray.test();
}
fetchIDL("https://www.w3.org/TR/payment-request/");
```

zcorpan:
I did not mean to `fetch()` from the spec directly. Relying on external resources is not OK. The IDL files need to be in web-platform-tests; there's a root directory called `interfaces` for this.

I don't know if `async`/`await` is OK yet; is it supported in stable Safari/Chrome/Firefox/Edge? (Servo?)

marcoscaceres:
> Safari/Chrome/Firefox/Edge

[Yes](http://caniuse.com/#search=await). 

> Servo?

🤷🏽‍♂️

marcoscaceres:
Thanks for clarifying about `interfaces`. Wasn't aware that was there.

rsolomakhin:
> Actually, the interfaces.html and interfaces.worker.js could be just one file, interfaces.any.js. Like /url/interfaces.any.js but then that one embeds its IDL inline for some reason.

@zcorpan How do I run the interfaces.any.js locally to verify that it works?

zcorpan:
See http://web-platform-tests.org/writing-tests/testharness.html#multi-global-tests

i.e. load interfaces.any.html to run the window test, and load interfaces.worker.html to run the worker test.

rsolomakhin:
@marcoscaceres @zcorpan : Please take a look.

payment-handler/interfaces.any.js
```diff
@@ -0,0 +1,22 @@
...
+
+if (self.importScripts) {
+    importScripts("/resources/testharness.js");
+    importScripts("/resources/WebIDLParser.js", "/resources/idlharness.js");
+}
+
+// https://w3c.github.io/payment-handler/
+
+promise_test(function() {
+    return fetch("/interfaces/payment-handler.idl")
```
marcoscaceres: Nit: please switch to async/await. Much more legible and maintainable.

payment-handler/interfaces.any.js
```diff
@@ -0,0 +1,22 @@
...
+        .then(response => response.text())
+        .then(text => {
+            var idlArray = new IdlArray();
```
marcoscaceres: nit: s/var/const

marcoscaceres (approved):
Just a couple of nits

marcoscaceres:
Dropped other reviewers. No need for all those people to review :)

tobie:
> /me shakes fist at wpt-pr-bot.

wpt-pr-bot won't let you attempt to sneak in some lousy commits without making sure you're exposed as the impostor you truly are.

Alternatively I could fix the bot to allow folks with the right permission set to change previews, but I'm not sure I can do that _and_ keep the bot stateless.

marcoscaceres:
> wpt-pr-bot won't let you attempt to sneak in some lousy commits without making sure you're exposed as the impostor you truly are.

Busted! 🕵🏻

I thought it was just assigning like crazy because of a lack of owners.

tobie:
> I thought it was just assigning like crazy because of a lack of owners.

Nope, it's when it hit the `/interfaces/` directory.

zcorpan:
Hmm I think it actually needs to be `interfaces.https.any.js` so the tests are run in secure context.

rsolomakhin:
> Hmm I think it actually needs to be interfaces.https.any.js so the tests are run in secure context.

Done.

rsolomakhin:
> Nit: please switch to async/await. Much more legible and maintainable.

Done, although I find async/await less clear personally.

> nit: s/var/const

Done.

payment-handler/interfaces.https.any.js
```diff
@@ -0,0 +1,20 @@
...
+"use strict";
+
+if (self.importScripts) {
+    importScripts("/resources/testharness.js");
+    importScripts("/resources/WebIDLParser.js", "/resources/idlharness.js");
+}
+
+// https://w3c.github.io/payment-handler/
+
+promise_test(async() => {
```
marcoscaceres: Sorry for being overly picky here, but this should really be: 

```JS
promise_test(async() => {
    const text = await fetch("/interfaces/payment-handler.idl")
        .then(response => response.text());
    const idlArray = new IdlArray();
    idlArray.add_idls(text);
    idlArray.test();
    done();
}, "Payment handler interfaces.");
```

marcoscaceres (changes requested):

payment-handler/interfaces.https.any.js
rsolomakhin: No worries. So that I etch this into my memory, please provide your motivation for this change.

rsolomakhin: Done.

marcoscaceres: Oh, rule is that the variable name  should reflect its content. So, if it's `text`, it should contain text and not a promise; otherwise it should be "textPromise". See also in the spec: we use the same naming pattern for `resposePromise` etc.

Also, generally `.then()` is considered  an anti-pattern. In simple cases like the above for .text() it's usually fine tho... but chaining them or nesting them is usually not  good. 

Hope that helps.

marcoscaceres (approved):

payment-handler/interfaces.https.any.js
rsolomakhin: Kewl... Do you have a good list of such patterns and anti-patterns for future reference?

marcoscaceres: Jakes article is pretty good (but doesn't exactly include these rules... sorta implied) https://developers.google.com/web/fundamentals/getting-started/primers/async-functions

PR #7084, opened 2017-08-31
Disable pytest cache provider in WPT.

jgraham:
The pytest cache provider lets you have access to --last-failed and
--failed-first based on a cache file it puts in .cache/v/lastfailed.
This can be useful for iterating on a set of failures, but not a feature
we want to support through wptrunner.

This patch disables the cache provider so that these files are
not generated.  When calling wptrunner with --manifest-update, these
temporary files have had a tedency to be included in MANIFEST.json,
and this change will effectively prevent that from happening.

MozReview-Commit-ID: EU7zXAdxsfY

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1391662 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7084)


jgraham:
Code reviewed upstream.

PR #7086, opened 2017-08-31
Flag the in-flow frames of kids of various wrapper frames during frame construction, so we know to restyle the wrapper frames.

jgraham:
MozReview-Commit-ID: 8KNug88sGp

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1388625 [ci skip]


jgraham:
Code reviewed upstream.

PR #7087, opened 2017-08-31
Don't allow Selection in nodes not in the document;

jgraham:
This matches the spec and Chrome, and seems to bring us closer to Edge
and WebKit as well.  It also matches our own behavior for addRange(),
which was changed in bug 1341137.

For collapse and selectAllChildren, we match the tests and browsers, but
the spec is incorrect at the time of this writing:
https://github.com/w3c/selection-api/pull/86

The removeAllRanges test hadn't been updated for the spec change.

MozReview-Commit-ID: DTK8283k5IP

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1359397 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7087)

### DIFF
--- a/selection/removeAllRanges.html
+++ b/selection/removeAllRanges.html
@@ -13,8 +13,13 @@ testRanges.unshift("[]");
 var range = rangeFromEndpoints([paras[0].firstChild, 0, paras[0].firstChild, 1]);
 
 function testRange(rangeDesc, method) {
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #7103, opened 2017-08-31
Fix undefined variable error with client.Session#position().

jgraham:
The x and y variables are not defined but needs to be extracted from
new_position.

MozReview-Commit-ID: Ds8cPlufjUa

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1391691 [ci skip]


jgraham:
Code reviewed upstream.

PR #7104, opened 2017-08-31
Restore window state using Set Window Rect.

jgraham:
Because the Maximize Window, Minimize Window, and Fullscreen Window
commands are now idempotent we will want to use the Set Window Rect to
bring them back to the normal window state.

MozReview-Commit-ID: BxtAJizfti6

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1391691 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7104)


jgraham:
Code reviewed upstream.

PR #7117, opened 2017-08-31
Improve edge case in-op tests for Set Window Rect.

jgraham:
Marionette has issues with repositioning/resizing the window when one
of width/height/x/y is the rect's existing value.  This is an edge case
it is worth testing for.

MozReview-Commit-ID: 7E46ly2lIiY

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1392368 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7117)

### DIFF
--- a/webdriver/tests/set_window_rect.py
+++ b/webdriver/tests/set_window_rect.py
@@ -458,19 +458,40 @@ def test_negative_x_y(session):
                                   "state": original["state"]})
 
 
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #7213, opened 2017-08-31
Add tests for user-timing, resource-timing, and timeOrigin in workers

quietfanatic:
Submitting this from Microsoft.  Thanks @toddreifsteck for the help!

This is my first time doing this, so please let me know if I did anything wrong. : )


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/18044)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/270520281)

sideshowbarker:
w3c-test:mirror

toddreifsteck:
@igrigorik please take a look. Because Lewis altered tests previously submitted, we should also have @JosephPecoraro take a look.

toddreifsteck:
@sideshowbarker How do I find the active mirror for this PR? I'd like to test in recent Edge just to confirm all tests pass.

sideshowbarker:
> @sideshowbarker How do I find the active mirror for this PR? I'd like to test in recent Edge just to confirm all tests pass.

https://w3c-test.org/submissions/7213/

workers/worker-performance.worker.js
```diff
@@ -0,0 +1,124 @@
...
+        // We want to skip over values that are 0 because of TAO securty rescritions
+        // Or else this test will fail. This can happen for "requestStart", "responseStart".
+        if (timestamp != 0) {
+            assert_greater_than_equal(timestamp, currentTimestamp, "Resource entry " + name + " is after " + currentTimestampName);
+            currentTimestamp = timestamp;
+            currentTimestampName = name;
+        }
+    });
+    assert_greater_than(entries[0].responseEnd, entries[0].startTime, "The resource request should have taken at least some time");
+    assert_approx_equals(entries[0].responseEnd - entries[0].responseStart, 250, 50, "Resource timing numbers reflect reality somewhat");
```
igrigorik: I think we need to relax this requirement? Executing this test against w3c-test.org I'm seeing significantly higher download times, in 450ms+ range. Perhaps change to 250ms+?

image

Test url: https://w3c-test.org/submissions/7213/workers/worker-performance.worker.html

workers/worker-performance.worker.js
```diff
@@ -0,0 +1,124 @@
...
+    performance.clearResourceTimings();
+    assert_equals(performance.getEntriesByType("resource").length, 0, "clearResourceTimings cleared the resource timings");
+}, "Resource timing seems to work in workers");
+
+test(function testPerformanceSetResourceTimingBufferSize () {
+    performance.setResourceTimingBufferSize(0);
+    var xhr = new XMLHttpRequest();
```
igrigorik: Hmm, I'm seeing a test failure here in Chrome. If I set a breakpoint on line 95 + 89, I break on 95 first and I see that there 2 entries returned by `getEntriesByType('resource')`, which is correct because `setResourceTimingBufferSize()` [does not trim the buffer](https://www.w3.org/TR/resource-timing-2/#dom-performance-setresourcetimingbuffersize). 

It appears that we have two tests racing here? Would it make sense to merge these two and run the clear -> setResourceTimingBufferSize as part of one test?

igrigorik (commented):
Lewis, thanks for the great work. These look great! 👍

https://w3c-test.org/submissions/7213/hr-time/window-worker-timeOrigin.window.html - passes in Chrome.

https://w3c-test.org/submissions/7213/workers/worker-performance.worker.html - gives me two failures in Chrome, left a few notes below.

workers/worker-performance.worker.js
quietfanatic: Ah, that makes sense.  I was running this test with a local server, but there's no upper bound to how long the request can take with a remote server.

workers/worker-performance.worker.js
quietfanatic: What happened was the previous test() block was failing before it reached clearResourceTimings, so the buffer was never cleared.  To make this independent, I just brought it out into its own test.

igrigorik: Ah, doh! Thanks, that makes sense.

igrigorik (approved):
Looks great, thanks for the great work!

igrigorik:
LGTM from my end. 

@JosephPecoraro can you please take a pass and sanity check as well?

workers/worker-performance.worker.js
```diff
@@ -0,0 +1,128 @@
+"use strict";
+
+// Tests that most of the functionality of the window.performance object is available in web workers.
```
JosephPecoraro: Is this needed? Aren't most `performance` tests written to run on both a page and a worker?

JosephPecoraro: For example this test seems to duplicate some of what `user-timing/mark*.any.js`, `user-timing/clear*.any.js`, and `user-timing/measure*.any.js` cover? Is it expected to have duplication of testing across different features (here having a test in `workers` duplicating `user-timing`, `resource-timing`, etc). I'm fine with it, I just don't really know WPT policies. I do see that none of the resource-timing tests test in workers, so there is certainly some merit for this test.

workers/worker-performance.worker.js
```diff
@@ -0,0 +1,128 @@
...
+    performance.clearMarks();
+    assert_equals(performance.getEntriesByType("mark").length, 0, "clearMarks cleared the marks");
+    performance.clearMeasures();
+    assert_equals(performance.getEntriesByType("measure").length, 0, "clearMeasures cleared the measures");
+}, "Can use clearMarks and clearMeasures in workers");
+
+test(function testPerformanceResourceTiming () {  // Resource timing
+    var start = performance.now();
+    var xhr = new XMLHttpRequest();
+     // Do a synchronous request and add a little artificial delay
```
JosephPecoraro: I don't recall the spec guaranteeing that when a synchronous XHR completes that a PerformanceResourceTiming entry is in the global performance buffer. That said, I think all existing resource timing tests relied on that behavior so all implementations have since met this pseudo-requirement. I think its fine to continue this, but it is something to be aware of.

workers/worker-performance.worker.js
```diff
@@ -0,0 +1,128 @@
...
+    assert_true(!!entries[0].name.match(/\/resources\/testharness.js/), "Resource entry has loaded url as its name");
+    assert_equals(entries[0].entryType, "resource", "Resource entry has correct entryType");
+    assert_equals(entries[0].initiatorType, "xmlhttprequest", "Resource entry has correct initiatorType");
+    var currentTimestamp = start;
+    var currentTimestampName = "a time before it";
+    [
+        "startTime", "fetchStart", "requestStart", "responseStart", "responseEnd"
+    ].forEach((name) => {
+        var timestamp = entries[0][name];
+        // We want to skip over values that are 0 because of TAO securty rescritions
```
JosephPecoraro: Typo: "securty rescritions" => "security restrictions"

workers/worker-performance.worker.js
```diff
@@ -0,0 +1,128 @@
...
+    assert_equals(performance.getEntriesByName("document", "navigation").length, 0, "getEntriesByName(\"document\", \"navigation\") returns nothing");
+    assert_equals(performance.getEntriesByName("document").length, 0, "getEntriesByName(\"document\") returns nothing");
+    var hasNavigation = performance.getEntries().some((e,i,a) => {
+        return e.entryType == "navigation";
+    });
+    assert_false(hasNavigation, "getEntries should return no navigation entries.");
+
+}, "There are no navigation type performance entries in workers");
+
+done();
```
JosephPecoraro: Looks good. This test does a covers a lot of basic API surface, do you also want to sanity check `PerformanceObserver` in workers?

JosephPecoraro:
The new tests look good to me. Nice to see `timeOrigin` tests!

workers/worker-performance.worker.js
igrigorik: @plehegar any guidance here? I can see arguments both ways.. duplication is not good (tm), but at the same time, some redundancy is OK in my books -- e.g. if someone is running worker test suite only, this would help signal potential issues with perf primitives.

plehegar: Some redundancy is also ok in my books. This test can indeed encourage support of RT in workers.

workers/worker-performance.worker.js
plehegar: The spec says "All resources fetched by the current browsing [HTML5] or worker [WORKERS] context's MUST be included as PerformanceResourceTiming objects in the Performance Timeline of the relevant context." so XHR requests are expected to be included. See https://w3c.github.io/resource-timing/#resources-included

plehegar (approved):

igrigorik:
Thanks all, merging. We can refactor + update in followup PRs.

workers/worker-performance.worker.js
JosephPecoraro: My concern was not _if_ XHRs show up in the performance timeline it was _when_ they show in the performance timeline. I was thinking of https://github.com/w3c/resource-timing/issues/82 which the spec has a note about at the end of the processing model.

Re-reading the processing model, I would expect a sync XHR to have a Performance Entry added to the global performance timeline before JavaScript continues executing. So this test seems fine.

PR #7235, opened 2017-09-04
Add tests for PaymentRequest.shippingOption attr

marcoscaceres:

### DIFF
--- a/payment-request/payment-request-shippingOption-attribute.https.html
+++ b/payment-request/payment-request-shippingOption-attribute.https.html
@@ -1,0 +1,80 @@
+<!DOCTYPE html>
+<!-- Copyright © 2017 Chromium authors and World Wide Web Consortium, (Massachusetts Institute of Technology, ERCIM, Keio University, Beihang). -->
+<meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/18126)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/271553950)

edenchuang (changes requested):
Because the fix on the spec https://github.com/w3c/payment-request/pull/610 is approved, I think we need to modify this test to consider the effect of options.requestShipping.

marcoscaceres:
> Because the fix on the spec w3c/payment-request#610 is approved, I think we need to modify this test to consider the effect of options.requestShipping.

Agree. On it!

edenchuang (changes requested):
We should add a test for the case that one shippingOption is set selected as true, but options.requestShipping is false.

edenchuang (approved):
Have the case in the ctor test, approve this PR.

PR #7250, opened 2017-09-04
Add test for text track cue order

fsoder:
Per https://html.spec.whatwg.org/#text-track-cue-order

### DIFF
--- a/html/semantics/embedded-content/media-elements/track/track-element/track-cue-order.html
+++ b/html/semantics/embedded-content/media-elements/track/track-element/track-cue-order.html
@@ -1,0 +1,83 @@
+<!DOCTYPE html>
+<title>Text track cue order</title>
+<link rel="help" href="https://html.spec.whatwg.org/#text-track-cue-order">
+<script src="/resources/testharness.js"></script>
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/18160)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/271702629)

zcorpan:
In Chromium/WebKit/Gecko, there's this failure

Result | Test Name | Message
-- | -- | --
Fail | Text track cue order, equal start and end times with startTime mutations. | assert_equals: "1" moved first expected "123" but got "231"

Though since the other tests pass, including the "endTime mutations" one, it seems like this is a bug in the browser engines; it wouldn't make sense to specify that cues aren't reordered for this case only.

zcorpan (approved):

fsoder:
Right, the following part of the spec:

> ...any cues with identical end times must be sorted in the order they were last added to their respective text track list of cues...

does not seem to be adhered to (in general.)

PR #7259, opened 2017-09-05
Change expectation for WebSocket case-only sub-protocol mismatch

ricea:
A sub-protocol mismatch during the WebSocket handshake causes the browser to
fail the WebSocket connection, meaning the onopen event should never fire. Test
websockets/constructor/011.html assumed that the onopen event would fire in this
case. This doesn't match the standard or the behaviour of any current major
browser. Fix it.

Also remove the unrelated check that the onclose event is not fired twice.

### DIFF
--- a/websockets/constructor/011.html
+++ b/websockets/constructor/011.html
@@ -8,20 +8,20 @@
 <div id=log></div>
 <script>
 async_test(function(t) {
+  // Sub-protocol matching is case-sensitive.
   var ws = new WebSocket(SCHEME_DOMAIN_PORT+'/handshake_protocol', 'FOOBAR');
   var gotOpen = false;
   var gotError = false;
   ws.onopen = t.step_func(function(e) {
     gotOpen = true;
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/18192)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/272067666)

ricea:
@zcorpan Please take a look.

ricea:
@tyoshino, can you review this?

tyoshino (approved):
Shall we clarify what kind of mismatch this is testing (you've already added L11 to explain that) also at the title?

tyoshino:
Oh, sorry to be unclear. I meant changing the contents in the <title> element in the test case.

ricea:
Oh. Yes, that makes sense.

PR #7261, opened 2017-09-05
Fetch: fragments in responses

annevk:
Test for https://github.com/whatwg/fetch/issues/1167.

### DIFF
--- a/fetch/api/response/response-fragment.window.js
+++ b/fetch/api/response/response-fragment.window.js
@@ -1,13 +1,25 @@
-promise_test(() => {
-  return fetch("./#test").then(res => {
-    assert_equals(res.url, new URL("./", location).href);
-    return fetch("../resources/redirect.py?location=/#test").then(res2 => {
-      // redirect.py changes the URL a bit; the important bit here is that the fragment is dropped
-      assert_equals(res2.url, new URL("/?location=%2F&count=1", location).href);
-      // redirect.py ends up appending some stuff to the fragment, but that does not affect this
-      return fetch("../resources/redirect.py?location=/%23test").then(res3 => {
-        assert_equals(res3.url, new URL("/", location).href);
-      });
+const expectedURL = new URL("/common/blank.html#test", location).href;
+function fragment_test(url, desc) {
+  promise_test(() => {
+    return fetch(url).then(res => {
+      assert_equals(res.url, expectedURL);
     });
-  });
-}, "Fragments don't end up in responses");
+  }, "Fetch: " + desc);
+  async_test(t => {
+    const client = new XMLHttpRequest();
+    client.open("GET", url);
+    client.send();
+    client.onload = t.step_func_done(() => {
+      assert_equals(client.responseURL, expectedURL);
+    })
+  }, "XMLHttpRequest: " + desc);
+}
+
+fragment_test("/common/blank.html#test",
+              "fragment in request copied over to response");
+fragment_test("../resources/redirect.py?omit_parameters&location=/common/blank.html#test",
+              "fragment in request copied over during redirect to response");
+fragment_test("../resources/redirect.py?omit_parameters&location=/%23test",
+              "fragment in redirect copied over to response");
+fragment_test("../resources/redirect.py?omit_parameters&location=/%23test#notit",
+              "fragment in request overridden by fragment in redirect for response");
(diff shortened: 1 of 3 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/18194)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/272081378)

fetch/api/response/response-fragment.window.js
```diff
@@ -0,0 +1,8 @@
+promise_test(() => {
+  return fetch("./#test").then(res => {
+    assert_equals(res.url, new URL("./", location).href);
+    return fetch("../resources/redirect.py?location=/#test").then(res2 => {
+      assert_equals(res2.url, new URL("/?location=%2F&count=1", location).href);
```
foolip: This bit took a while to understand, but I think it's correct. You really have to read redirect.py to make sense of it though, first I assumed redirect.py would simply redirect. Maybe comment "/redirect.py redirects to this funny URL, and the UA should not append a fragment"?

foolip (approved):
LGTM % trouble of understanding redirect.py

fetch/api/response/response-fragment.window.js
annevk: Will do.

annevk:
Hmm, I should maybe also test a redirect to a URL that includes a fragment, for completeness.

fetch/fragments/subresource-redirect.html
```diff
@@ -0,0 +1,7 @@
+<p><img src=/images/green.svg>
```
foolip: I guess this isn't ready for review yet since this isn't used, will hold off until poked.

fetch/api/response/response-fragment.window.js
```diff
@@ -0,0 +1,13 @@
+promise_test(() => {
+  return fetch("./#test").then(res => {
+    assert_equals(res.url, new URL("./", location).href);
+    return fetch("../resources/redirect.py?location=/#test").then(res2 => {
+      // redirect.py changes the URL a bit; the important bit here is that the fragment is dropped
+      assert_equals(res2.url, new URL("/?location=%2F&count=1", location).href);
+      // redirect.py ends up appending some stuff to the fragment, but that does not affect this
+      return fetch("../resources/redirect.py?location=/%23test").then(res3 => {
+        assert_equals(res3.url, new URL("/", location).href);
```
foolip: I was surprised here, but I guess the deal is that fragments on redirects are not preserved and if they're included they're thrown away, *except* for navigations.

annevk: That's the current deal, yes. Perhaps we should alter the deal though.

fetch/fragments/subresource-redirect.html
annevk: Yeah, I was thinking of using a reference test for this, but I first need to determine if everyone is okay with passing along fragments all the time.

fetch/fragments/subresource-redirect.html
```diff
@@ -0,0 +1,37 @@
...
+};
+document.querySelectorAll("img").forEach(img => {
+  async_test(t => {
+    t.add_cleanup(() => {
+      img.remove();
+      ctx.clearRect(0, 0, 100, 50);
+    });
+    self.addEventListener("load", t.step_func_done(() => {
+      ctx.drawImage(img, 0, 0);
+      assert_true(isColor(ctx.getImageData(0, 0, 100, 25).data, 0, 255, 0, 255));
```
foolip: I think this bit could use a comment to explain what the structure of the image is and why these asserts would fail if something were wrong with how fragments are handled on subresource redirects.

annevk: Added an explanation.

foolip (approved):

domenic (approved):

fetch/fragments/subresource-redirect.html
```diff
@@ -0,0 +1,48 @@
...
+  }
+  return true;
+};
+document.querySelectorAll("img").forEach(img => {
+  async_test(t => {
+    t.add_cleanup(() => {
+      img.remove();
+      ctx.clearRect(0, 0, 100, 50);
+    });
+    self.addEventListener("load", t.step_func_done(() => {
```
Dlee827, line 41: Road

Dlee827: Event listener￼

PR #6645, opened 2017-07-28
Rework transition event tests

birtles:
I would like to simplify the tests in CSS transitions, beginning with the events tests so that I can add tests for the new `transitionrun`, `transitionstart`, and `transitioncancel` events.

I have a few concerns with the current tests:

* They seem overly complicated:
  * One is required to read an extra README simply to make head or tails of what the test is doing.
  * They require a lot of boilerplate.
  * They are far from minimal. When we were analyzing why `properties-value-implicit-001.html` was failing in [Mozilla bug 1356627](https://bugzilla.mozilla.org/show_bug.cgi?id=1356627) I discovered it was failing because somewhere deep in the test harness it was failing in the way it was waiting on `transitionend` events, something that seems completely unrelated to what should be under test in that file.
  * As a result it is time-consuming to reduce a failing test case for debugging.
* They make little use of existing `testharness.js` features such as `EventWatcher` etc. As a result, more unfamiliar boilerplate/libraries are required.
* They are brittle. In Gecko we had to [disable the entire css-transitions folder of tests](https://bugzilla.mozilla.org/show_bug.cgi?id=1356627) because they had so many intermittent failures.

I'd like to try rewriting these tests, keeping roughly the same test coverage, but taking a simpler approach.

Note that the changes to `EventWatcher` should be merged in #6634 and are not part of this PR.

PR #7273, opened 2017-09-07
Update more transition event tests

birtles:
Continuing on from #6645, this applies the same treatment to the remainder of the events tests.

### DIFF
--- a/css/css-transitions-1/events-004.html
+++ b/css/css-transitions-1/events-004.html
@@ -1,150 +1,61 @@
-<!DOCTYPE html>
+<!doctype html>
 <html>
-    <head>
-        <meta charset="utf-8">
-        <title>CSS Transitions Test: transitionend event with non matching lists</title>
-        <meta name="assert" content="Test checks that non-matching lists are properly resolved">
-        <link rel="help" title="2. Transitions - Example 3" href="http://www.w3.org/TR/css3-transitions/#list-matching">
-        <link rel="help" title="5. Transition Events" href="http://www.w3.org/TR/css3-transitions/#transition-events">
-        <link rel="author" title="Rodney Rehm" href="http://rodneyrehm.de/en/">
-        <meta name="flags" content="dom">
+<head>
+<meta charset=utf-8>
+<title>CSS Transitions Test: transitionend event with non matching lists</title>
+<meta name="assert" content="Test checks that non-matching lists are properly resolved">
+<link rel="help" title="2. Transitions - Example 3" href="http://www.w3.org/TR/css3-transitions/#list-matching">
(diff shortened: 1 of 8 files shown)


birtles:
CC @ewilligers

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/18257)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/272769041)

css/css-transitions-1/events-004.html
```diff
@@ -1,150 +1,61 @@
...
+<div id="log"></div>
+
+<script>
+promise_test(t => {
+  const div = addDiv(t, { style: 'padding: 1px' });
+  getComputedStyle(div).paddingLeft;
+
+  div.style.transitionProperty =
+    'padding-top, padding-right, padding-bottom, padding-left';
+  div.style.transitionDuration = '0.02s, 0.01s';
```
ewilligers: Should the durations be left unchanged for this test?

birtles: They probably could in this case. In general I've adjusted the durations to be as short as possible while still being significantly more than the tolerance we allow when comparing times so that when we run them in automation that don't add up to a significant amount of machine time.

So in this case I've changed `0.1s, 0.05s` (total 150ms) to `0.02s, 0.01s` (30ms) which is not a massive saving but between ~10 test cases means >1s saving. Some of the other tests are more significant.

css/css-transitions-1/events-007.html
```diff
@@ -1,64 +1,50 @@
...
 </head>
 <body>
-    <div id="a" class="red"></div>
+
+<div id="log"></div>
+
+<script>
+async_test(t => {
+  window.addEventListener('load', function () {
+    const div = addDiv(t, { 'style': 'transition: background-color 0.5s;' +
```
ewilligers: I suspect this should be less than 0.5s, e.g. 0.2s, so the test would usually fail if "div.style.display = 'none';" was not present.

ewilligers (approved):

css/css-transitions-1/events-007.html
birtles: Hmm, I did check that the test fails without `div.style.display = 'none';` at least in Gecko (I ran all the other tests in Chrome too but didn't check this one failed without the `display:none` in Chrome). The reason this should work is we don't start the timeout until the `requestAnimationFrame` callback by which point the transition should be running so we should have <500ms until the transition ends. We can make the transition shorter but then we increase the chance of the `requestAnimationFrame` running after the transition has already completed.

How about I make the transition `0.4s` just in case there's a UA that doesn't start the transition until the `requestAnimationFrame`.

ewilligers: Sure

PR #7275, opened 2017-09-07
Test that PaymentRequest aborts if doc not fully active

### LINKED ISSUE #7240: show() must reject when doc not fully active
In  "The show() method must act as follows": 

* [x] If document is not fully active, then return a promise rejected with an "AbortError" DOMException.

Copy: XMLHttpRequest/open-url-multi-window-6.htm

And test for:

* [x] If document stops being fully active while the user interface is being shown, or no longer is by the time this step is reached, then the user interface should be hidden, and acceptPromise should be rejected with an "AbortError" DOMException.

marcoscaceres:
closes https://github.com/w3c/web-platform-tests/issues/7240


marcoscaceres:
@domenic, @zcorpan, sorry, I'm not sure if the second test here is actually correct for "fully active". If possible, I could use a little guidance.

@rsolomakhin you might be interested in these tests... Chrome doesn't respect the first test. And for the second test, it claims `.show()` was already called :)

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/18262)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/272831404)

rsolomakhin:
@marcoscaceres Thank you for bringing this to my attention. I will be making sure that Chromium passes all web-platform-tests.

domenic (changes requested):
This isn't testing fully active, indeed :).

Here are the two cases worth testing for a document not being fully active:

## Not active

You need an iframe navigated to a non-`about:blank` location. Make sure the load event has already fired.

- Grab `iframe.contentWindow.PaymentRequest`
- Use it to construct a new `pr` object. (This step can be either before or after the next one; shouldn't matter.)
- Navigate `iframe` to a new location. Wait for the load event to fire.
- Now, `pr`'s relevant global object's document is no longer active. So, call `pr.show()`, and make sure it rejects appropriately.

## Active, but not fully active

You need an iframe containing another iframe. Again, both must have navigated to non-`about:blank` locations, and have finished loading.

- Grab `iframe.contentDocument.querySelector("#otheriframe").contentWindow.PaymentRequest`
- Use it to construct a new `pr` object.
- Navigate `iframe` (the outer iframe) to a new location. Wait for the load event to fire.
- Now, `pr`'s relevant global object's document is still active (it is the active document of the inner iframe), but is not fully active (since the parent of the inner iframe is itself no longer active). So, call `pr.show()` and make sure it rejects appropriately.

---

I suppose the iframe.remove() case is also good to test, however that's an open spec bug: https://github.com/whatwg/html/issues/1073 . Probably that's a fixable spec bug though; maybe I can get to it soon.

payment-request/throw_if_not_active.https.html
```diff
@@ -0,0 +1,161 @@
...
+  await promise_rejects(
+    t,
+    "AbortError",
+    request.show(),
+    "Active, but not fully active, so must throw AbortError"
+  );
+  iframe.remove();
+}, "PaymentRequest.show() aborts if the document is active, but not fully active");
+
+promise_test(async t => {
```
marcoscaceres: This is the test for:

> If document stops being fully active while the user interface is being shown, or no longer is by the time this step is reached, then the user interface should be hidden, and acceptPromise should be rejected with an "AbortError" DOMException.

marcoscaceres:
@domenic, all reviewers, this is ready for another round of review.

payment-request/throw_if_not_active.https.html
```diff
@@ -0,0 +1,166 @@
...
+  // check that PaymentRequests can be constructed.
+  new PaymentRequest(validMethods, validDetails);
+  // We create two iframes, and nest them. Wait for both of them to load.
+  const iframe = document.createElement("iframe");
+  iframe.allowPaymentRequest = true;
+  document.body.appendChild(iframe);
+  // We get the request from the inner-most iframe.
+  const request = await new Promise(resolve => {
```
domenic: Can't you use your super-nice helper here?

```js
const request1 = await getLoadedPaymentRequest(iframe, "/payment-request/resources/page1.html");

const innerIframe = iframe.contentDocument.createElement("iframe");
innerIframe.allowPaymentRequest = true;
iframe.contentDocument.body.appendChild(innerIframe);

const request2 = await getLoadedPaymentRequest(innerIframe, "/payment-request/resources/page2.html");
```

? If I'm missing something, I'm still hopeful that you can create some variation on your helper that does work, since it's so nice.

payment-request/throw_if_not_active.https.html
```diff
@@ -0,0 +1,166 @@
...
+    t,
+    "AbortError",
+    request.show(),
+    "Active, but not fully active, so must throw AbortError"
+  );
+  // We are done, so clean up.
+  iframe.remove();
+}, "PaymentRequest.show() aborts if the document is active, but not fully active");
+
+promise_test(async t => {
```
domenic: This test is redundant with the first one, right?

domenic (commented):
Sorry for the lag. In Tokyo now so we should be in similar timezones!

payment-request/throw_if_not_active.https.html
marcoscaceres: Um, maybe I screwed it up... In the first test, I'm testing that .show() can't be called if the document is inactive... but in this test, I'm making sure that that if the payment sheet is *already showing* then it should abort.

payment-request/throw_if_not_active.https.html
marcoscaceres: Will add tomorrow :( Ran out of time today.

payment-request/throw_if_not_active.https.html
domenic: Ah, no, you're totally right!!

I guess you can also use your helper in this test.

marcoscaceres: Used it below ... was worried about adding too much indirection, but it's all good :)

payment-request/throw_if_not_active.https.html
```diff
@@ -0,0 +1,158 @@
+<!DOCTYPE html>
+<meta charset=utf-8>
+<title>PaymentRequest show() throws if doc is not fully active</title>
```
domenic: Nit: "rejects", not throws

payment-request/throw_if_not_active.https.html
```diff
@@ -0,0 +1,159 @@
+<!DOCTYPE html>
+<meta charset=utf-8>
+<link rel="help" href="https://w3c.github.io/payment-request/#show()-method">
+<title>PaymentRequest show() throws if doc is not fully active</title>
```
domenic: Rejects, not throws

payment-request/throw_if_not_active.https.html
```diff
@@ -0,0 +1,159 @@
...
+    t,
+    "AbortError",
+    showPromise,
+    "The iframe was navigated away, so showPromise must reject with AbortError"
+  );
+  // We are done, so clean up.
+  iframe.remove();
+}, "If a payment request is showing, but its document is navigated away (so no longer fully active), the payment request aborts.");
+
+done();
```
domenic: This shouldn't be necessary

domenic (approved):
LGTM with nits. I'll be referring people to this file in the future as a great example of testing fully-activeness.

marcoscaceres:
Ok, hopefully all good now 🤞Looks like everyone fails these spectacularly 😁

marcoscaceres:
> LGTM with nits. I'll be referring people to this file in the future as a great example of testing fully-activeness.

That's great to hear. The guidance of navigating the iframes to different places really helped, so big thanks!

PR #7345, opened 2017-09-14
Fix some WebSocket test handlers

ricea:
Several of the pywebsocket handlers used in tests had the problem that they
would generate a hand-crafted header and then pywebsocket would add its own
header immediately afterwards, causing unnecessary errors.

In most cases there was no need to generate a hand-crafted header at all, so the
tests have been modified to use pywebsocket APIs to do the same
thing. simple_handshake_wsh.py still generates a hand-crafted header but now
suppresses pywebsocket's own header.

Several tests closed the WebSocket on the client side but the handler didn't
wait to receive the client-side close. These have been modified to wait for the
close frame.

Several tests had expectations that onclose would never be called but then
called close() themselves, causing onclose to be called. In some cases this was
mitigated by calling t.done() before t.unreached_func() was called, but it has
been fixed by disabling the onclose handler after calling close().

This change only fixes a subset of handlers and their affected tests. It is not
intended to be comprehensive, and many issues remain.

### DIFF
--- a/websockets/handlers/simple_handshake_wsh.py
+++ b/websockets/handlers/simple_handshake_wsh.py
@@ -1,7 +1,5 @@
 #!/usr/bin/python
 
-import struct
-
 from mod_pywebsocket import common, stream
 from mod_pywebsocket.handshake import AbortedByUserException, hybi
 
--- a/websockets/handlers/simple_handshake_wsh.py
+++ b/websockets/handlers/simple_handshake_wsh.py
@@ -1,13 +1,25 @@
 #!/usr/bin/python
 
-from mod_pywebsocket import common, msgutil, util
-from mod_pywebsocket.handshake import hybi
+import struct
+
+from mod_pywebsocket import common, stream
+from mod_pywebsocket.handshake import AbortedByUserException, hybi
 
 
 def web_socket_do_extra_handshake(request):
+    # Send simple response header. This test implements the handshake
+    # manually. It's not clear why.
     request.connection.write('HTTP/1.1 101 Switching Protocols:\x0D\x0AConnection: Upgrade\x0D\x0AUpgrade: WebSocket\x0D\x0ASet-Cookie: ws_test=test\x0D\x0ASec-WebSocket-Origin: '+request.ws_origin+'\x0D\x0ASec-WebSocket-Accept: '+hybi.compute_accept(request.headers_in.get(common.SEC_WEBSOCKET_KEY_HEADER))[0]+'\x0D\x0A\x0D\x0A')
-    return
+    # Send a clean close frame.
+    body = stream.create_closing_handshake_body(1000, '')
+    request.connection.write(stream.create_close_frame(body));
+    # Wait for the responding close frame from the user agent. It's not possible
+    # to use the stream methods at this point because the stream hasn't been
+    # established from pywebsocket's point of view. Instead just read the
+    # appropriate number of bytes and assume they are correct.
+    request.connection.read(8)
+    # Close the socket without pywebsocket sending its own handshake response.
+    raise AbortedByUserException('Abort the connection')
 
 def web_socket_transfer_data(request):
-    while True:
-        return
+    pass
--- a/websockets/cookies/002.html
+++ b/websockets/cookies/002.html
@@ -17,6 +17,7 @@ async_test(function(t) {
   ws.onopen = t.step_func(function(e) {
     assert_regexp_match(document.cookie, new RegExp('ws_test_'+cookie_id+'=test'));
     ws.close();
+    ws.onclose = null;
     t.done();
   });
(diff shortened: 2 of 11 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/18529)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/275335062)

ricea:
@tyoshino PTAL

Ms2ger:
> Several tests had expectations that onclose would never be called but then called close() themselves, causing onclose to be called. In some cases this was mitigated by calling t.done() before t.unreached_func() was called, but it has been fixed by disabling the onclose handler after calling close().

I don't think these particular changes are necessary or helpful.

ricea:
> I don't think these particular changes are necessary or helpful.

I have two problems with the way this works at the moment:

1. We're relying on the test harness to ignore assertion failures that happen after we call done(). I don't know whether this is common practice, but it seems risky and fragile to me to ignore assertion failures at any time.
2. It makes the test misleading. As written, anyone reading the test could reasonably conclude that onclose is never called, but that is not the case.

Ms2ger:
This is very much part of the design of the test harness; perhaps @gsnedders has thoughts.

gsnedders:
I've never touched any of this; defer to @jgraham.

websockets/handlers/simple_handshake_wsh.py
```diff
@@ -1,13 +1,25 @@
 #!/usr/bin/python
 
-from mod_pywebsocket import common, msgutil, util
-from mod_pywebsocket.handshake import hybi
+import struct
```
tyoshino: not used?

tyoshino (commented):
Even if the design guarantees that, I'd like to vote for the change considering ricea's second point.

websockets/handlers/simple_handshake_wsh.py
```diff
@@ -1,13 +1,25 @@
...
     request.connection.write('HTTP/1.1 101 Switching Protocols:\x0D\x0AConnection: Upgrade\x0D\x0AUpgrade: WebSocket\x0D\x0ASet-Cookie: ws_test=test\x0D\x0ASec-WebSocket-Origin: '+request.ws_origin+'\x0D\x0ASec-WebSocket-Accept: '+hybi.compute_accept(request.headers_in.get(common.SEC_WEBSOCKET_KEY_HEADER))[0]+'\x0D\x0A\x0D\x0A')
-    return
+    # Send a clean close frame.
+    body = stream.create_closing_handshake_body(1000, '')
+    request.connection.write(stream.create_close_frame(body));
+    # Wait for the responding close frame from the user agent. It's not possible
+    # to use the stream methods at this point because the stream hasn't been
+    # established from pywebsocket's point of view. Instead just read the
+    # appropriate number of bytes and assume they are correct.
+    request.connection.read(8)
```
tyoshino: Don't we need to emulate _stream_base.receive_bytes() here?

ricea: I think it's safe to assume that if we got any bytes at all then the close frame has been sent.

It's possible that a browser could send the close frame fragmented into two parts, the server closes the connection after the first fragment and the browser gets a "Connection reset by peer" error and sets wasClean to false. But I know of no reason why a browser would fragment the close frame.

This handler is already complicated and probably pointless, so I don't want to complicate it any further.

websockets/handlers/simple_handshake_wsh.py
ricea: Good catch. Must have been leftover from a previous iteration.

websockets/handlers/simple_handshake_wsh.py
tyoshino: ok

tyoshino (approved):
I'm happy with the current patch. You can wait for jgraham's comment regarding usage of done().

ricea:
@jgraham, do you have any comments on whether asserting after a async_test is done is good practice?

gsnedders:
@ricea I'd hope we ignore such asserts? Probably even set the test/harness to ERROR.

ricea:
@gsnedders The asserts were ignored. I found it surprising. I see the situation with uncaught promise rejections in promise_tests as analogous. Uncaught rejections cause the whole page to fail, even though they can't be attributed to a single test.

gsnedders:
@ricea Likewise uncaught exceptions do. We should definitely fix this. Lemme file a bug.

gsnedders:
Oh, thinking about this, it may actually be unsolvable as long as assertions aren't linked directly to tests and only are processed as a result of them throwing an exception. We definitely can't ever catch `assert_true(true)` even if it's after being done.

PR #7348, opened 2017-09-14
test that updateWith() is called immediately

marcoscaceres:

### DIFF
--- a/payment-request/PaymentRequestUpdateEvent/updateWith-call-immediate-manual.https.html
+++ b/payment-request/PaymentRequestUpdateEvent/updateWith-call-immediate-manual.https.html
@@ -1,0 +1,170 @@
+<!doctype html>
+<meta charset="utf8">
+<link rel="help" href="">
+<title>
+</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+setup({ explicit_done: true, explicit_timeout: true });
+const validMethod = Object.freeze({ supportedMethods: "basic-card" });
+const validMethods = Object.freeze([validMethod]);
+const validAmount = Object.freeze({ currency: "USD", value: "5.00" });
+const validTotal = Object.freeze({
+  label: "label",
+  amount: validAmount,
+});
+const validShippingOptionA = Object.freeze({
+  id: "a-shipping-option",
+  label: "A shipping option",
+  amount: validAmount,
+  selected: true,
+});
+const validShippingOptionB = Object.freeze({
+  id: "b-shipping-option",
+  label: "B shipping option",
+  amount: validAmount,
+});
+const validDetails = Object.freeze({
+  total: validTotal,
+  shippingOptions: [validShippingOptionA, validShippingOptionB],
+});
+const validOptions = Object.freeze({
+  requestShipping: true,
+});
+
+function testImmediateUpdate({ textContent: testName }) {
+  promise_test(async t => {
+    const request = new PaymentRequest(
+      validMethods,
+      validDetails,
+      validOptions
+    );
+    const shippingOptionEventPromise = new Promise((resolve, reject) => {
+      request.addEventListener(
+        "shippingaddresschange",
+        async ev => {
+          // spin the event loop, sets [[waitForUpdate]] to true.
+          await Promise.resolve();
+          try {
+            ev.updateWith(validDetails);
+            resolve(); // This is bad.
+          } catch (err) {
+            reject(err); // this is good.
+          }
+        },
+        { once: true }
+      );
+    });
+    const response = await request.show();
+    await promise_rejects(
+      t,
+      "InvalidStateError",
+      shippingOptionEventPromise,
+      "The event loop already spun, so [[waitForUpdate]] is now true"
+    );
+    await response.complete();
+  }, testName.trim());
+}
+
+function testSubsequentUpdateWithCalls({ textContent: testName }) {
+  promise_test(async t => {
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PASSED

Started: 2017-09-14 08:58:58
Finished: 2017-09-14 09:06:23


<h2>Failing Jobs</h2>
<ul>

<li>safari:10.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/18543)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/275368590)

payment-request/PaymentRequestUpdateEvent/updateWith-call-immediate-manual.https.html
```diff
@@ -0,0 +1,171 @@
...
+    // Let's wait for the shippingOption
+    const optionChangeEvent = await optionChangedPromise;
+
+    // The addressChangeEvent was already consumed and expired.
+    assert_throws(
+      new TypeError(),
+      () => {
+        addressChangeEvent.updateWith(validDetails);
+      },
+      "addressChangeEvent [[waitingForUpdate]] is true, so it must throw"
```
edenchuang: I think addressChangeEvent.[[WaitForUpdating]] is false after optionChangeEvent.updateWith(), because the updating of the first addChangeEvent.updateWith() must be processed before selecting an option and emitting the optionChangeEvent. Maybe we need to update the error message to the correct situation.
I am not sure what kind of error it should be, TypeError, InvalidStateError, or others.

payment-request/PaymentRequestUpdateEvent/updateWith-call-immediate-manual.https.html
```diff
@@ -0,0 +1,171 @@
...
+        try {
+          ev.updateWith(validDetails);
+          resolve(); // this is bad, we should never get to here.
+        } catch (err) {
+          reject(err); // this is good!
+        }
+      });
+    });
+    const responsePromise = request.show();
+    await promise_rejects(t, "InvalidStateError", eventPromise);
```
edenchuang: Please add some error message here.

payment-request/PaymentRequestUpdateEvent/updateWith-call-immediate-manual.https.html
```diff
@@ -0,0 +1,171 @@
...
+
+    // But optionChangeEvent is still usable tho, so...
+    optionChangeEvent.updateWith(validDetails);
+
+    assert_throws(
+      "InvalidStateError",
+      () => {
+        optionChangeEvent.updateWith(validDetails);
+      },
+      "optionChangeEvent [[waitingForUpdate]] is true, so it must throw"
```
edenchuang: Here is a rase condition. We can't not sure which one is executed between the first optionChangeEvent.updateWith() resolve callback and the second optionChangeEvent.updateWith(). If the resolve callback is executed first, optionChangeEvent.[[waitingForUpdate]] will become false.
If we want to make sure [[waitingForUpdate]] is true, using a pending promise in the first optionChangeEvent.updateWith() probably is a good solution. And we can add two assertions before and after resolving the pending promise. The before one should get InvalidStateError, since the optionChangeEvent.[waitingForUpdate] is true. And the second should throw an error, maybe InvalidStateError, since the event is expired. 

By the way, in fact, the spec doesn't mention what happens if the merchant calls PaymentRequestUpdateEvent.updateWith repeatlly. In my opinion, for each PaymentRequestUpdateEvent, updateWith can only be called one time.

edenchuang (changes requested):
See comments.

payment-request/PaymentRequestUpdateEvent/updateWith-call-immediate-manual.https.html
marcoscaceres: > I think addressChangeEvent.[[WaitForUpdating]] is false, 

We changed the spec so `[[waitingForUpdate]]`  can't ever go back to false. Once .updateWith() is called, it's always true.

marcoscaceres: > I am not sure what kind of error it should be, TypeError, InvalidStateError, or others.

The error should have been `InvalidStateError` - so good catch!

payment-request/PaymentRequestUpdateEvent/updateWith-call-immediate-manual.https.html
marcoscaceres: > Here is a rase condition. We can't not sure which one is executed between the first optionChangeEvent.updateWith() resolve callback and the second optionChangeEvent.updateWith()

In the spec, .updateWith() immediately sets `[waitingForUpdate]` to true (before we go async):

> 8. Set event.[[waitForUpdate]] to true.

And step 3 guards:

> 3. If event.[[waitForUpdate]] is true, then throw an "InvalidStateError" DOMException.

marcoscaceres: > By the way, in fact, the spec doesn't mention what happens if the merchant calls PaymentRequestUpdateEvent.updateWith repeatlly. In my opinion, for each PaymentRequestUpdateEvent, updateWith can only be called one time. 

The:

> 3. If event.[[waitForUpdate]] is true, then throw an "InvalidStateError" DOMException.

Prevents calling .updateWith() multiple times.

edenchuang (approved):

PR #7394, opened 2017-09-19
Fix --exclude and --include arguments to wptrunner,

jgraham:
We need to set_default when creating an empty IncludeManifest so that skip is set.
This is pretty unfortunate.

MozReview-Commit-ID: 43Vwtu2bhvn

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1397215 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7394)


jgraham:
Code reviewed upstream.

PR #7405, opened 2017-09-19
Relax WindowRectResponse typing.

jgraham:
The WindowProxy screenX/screenY/outerWidth/outerHeight fields are
platform-independent longs, meaning the bitness of the system is not
taken into account.  This is defined in WebIDL as "a signed integer type
that has values in the range [−2147483648, 2147483647]".  This can be
represented with an i32 internally.

Additionally, the WebDriver specification says that we should accept
JSON Numbers as input.  This means the input value can be either a float
or an integer.  rustc_serialize handles this very badly and we first
need to extract the input as f64 because this is the most liberal type
(i.e. an integer can be coerced to a float without problem).

We then we cast the number to an i64 which trims off any decimals without
worrying that the number wraps.  We then run bounds checks of an i32
and error if the number will not fit within the set range.

This patch expects the output from Marionette to be conforming.

MozReview-Commit-ID: CVdUZWh2t8X

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1396866 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7405)

### DIFF
--- a/webdriver/tests/contexts/maximize_window.py
+++ b/webdriver/tests/contexts/maximize_window.py
@@ -191,6 +191,33 @@ def test_maximize(session):
 
 
 def test_payload(session):
+    """
+    7. Return success with the JSON serialization of the current top-level
+    browsing context's window rect.
+
+    [...]
+
+    A top-level browsing context's window rect is defined as a
+    dictionary of the screenX, screenY, width and height attributes of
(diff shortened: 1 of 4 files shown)


jgraham:
Code reviewed upstream.

PR #7406, opened 2017-09-19
Disable test_maximize_when_resized_to_max_size.

jgraham:
This test causes both a failure and an error and we need to investigate
how to reuse the wptrunner session configuration when implicitly starting
a session.

MozReview-Commit-ID: 5k7nfevamZY

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1396866 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7406)


jgraham:
Code reviewed upstream.

PR #7438, opened 2017-09-21
[css-text-3] Add tests for line-break:anywhere

frivoal:
Tests for https://drafts.csswg.org/css-text-3/#valdef-line-break-anywhere, as added in https://github.com/w3c/csswg-drafts/issues/1171 / https://github.com/w3c/csswg-drafts/issues/1561 / https://github.com/w3c/csswg-drafts/commit/c2beab13bcc63aa8597ad5d97ab7faee1cc5182f

### DIFF
--- a/css/css-text-3/line-break/line-break-anywhere-001.html
+++ b/css/css-text-3/line-break/line-break-anywhere-001.html
@@ -1,0 +1,32 @@
+﻿<!DOCTYPE html>
+<html lang=en>
+<meta charset="utf-8">
+<title>CSS Text Test: line-break: anywhere</title>
+<link rel="author" title="Florian Rivoal" href="https://florian.rivoal.net/">
+<link rel="help" href="https://drafts.csswg.org/css-text-3/#valdef-line-break-anywhere">
+<link rel="match" href="reference/line-break-anywhere-001-ref.html">
+<meta name="flags" content="">
+<meta name="assert" content="line-break:anywhere puts a soft wrap opportunity around every typographic character unit,
+                             including around punctuation or in the middle of words,
+                             disregarding any prohibition against line breaks introduced by characters with the GL, JW, or ZJW character class.">
+<style>
+#green {
+  position: absolute;
+  background: green;
+  font-family: monospace;
+  width: 1ch;
+  height: 20em;
+}
+#test {
+  width: 1ch;
+  line-height: 1;
+  color: red;
+  font-family: monospace;
+  line-break: anywhere;
+}
+</style>
+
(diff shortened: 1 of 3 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/18751)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/278213833)

css/css-text-3/line-break/line-break-anywhere-001.html
```diff
@@ -0,0 +1,32 @@
...
+  color: red;
+  font-family: monospace;
+  line-break: anywhere;
+}
+</style>
+
+<p>Test passes if there is a green rectangle below and no red.</p>
+<div id=green></div>
+<!-- with line breaks everywhere, none of the following characters should stick out from under the green div -->
+<div id=test>aa-a.a)a,a）a&nbsp;a&#xfeff;a&#x2060;a&#x200d;a･a</div>
```
FremyCompany: Is it possible that some of these characters could use a non-monospace fonts on a system that doesn't have a monospace glyph for them, and they end up overflowing anyway?

Also, maybe some of these characters could end up not rendered at all (be transparent) and overflow the line but you wouldn't be able to see it. 

Maybe it's ok, not sure if this has to be catched by the test

FremyCompany: Maybe wrapping each char in an `<abbr title="name-of-the-glyph">` with `abbr { background-color: red }` would help catch the second scenario? But then there might be a bug where a browser will only break if the text nodes are not the same, so I would maybe recommend the test have the two versions (one with the spans, one without).

FremyCompany (approved):
Approved (with suggestions)

css/css-text-3/line-break/line-break-anywhere-001.html
frivoal: > Is it possible that some of these characters could use a non-monospace fonts on a system that doesn't have a monospace glyph for them, and they end up overflowing anyway?

Only to the extent that fonts may fail to do what they're supposed to be designed for. There's one character in that list that is larger: the second parenthesis is a full width one, so 2ch wide rather than 1ch wide. However, it is also supposed to only have ink in the left half, as the right half is supposed to be trimmed out in various circumstances.

So unless the font isn't actually mono space despite claiming to be, or has incorrectly designed full width punctuation, it ought to work.

The red background trick would actually fail on this full width closing parenthesis, since that would give you 2ch worth or red, instead of the expected 1ch worth of ink inside a 2ch space.

PR #7441, opened 2017-09-21
Implement Get Timeouts tests for WPT wdspec tests

alchiem:
Testing webdriver get timeouts spec.
http://www.w3.org/TR/webdriver/#get-timeouts

### DIFF
--- a/webdriver/tests/sessions/get_timeouts.py
+++ b/webdriver/tests/sessions/get_timeouts.py
@@ -21,6 +21,26 @@ def test_get_timeouts(session):
     assert "implicit" in value
     assert "pageLoad" in value
 
-    assert isinstance(response.body["value"]["script"], int)
-    assert isinstance(response.body["value"]["implicit"], int)
-    assert isinstance(response.body["value"]["pageLoad"], int)
+    assert isinstance(value["script"], int)
+    assert isinstance(value["implicit"], int)
+    assert isinstance(value["pageLoad"], int)
+
+
+def test_get_default_timeouts(session):
+    response = get_timeouts(session)
+
+    assert_success(response)
+    assert response.body["value"]["script"] == 30000
+    assert response.body["value"]["implicit"] == 0
+    assert response.body["value"]["pageLoad"] == 300000
+
+
+def test_get_new_timeouts(session):
+    session.timeouts.script = 60
+    session.timeouts.implicit = 1
+    session.timeouts.page_load = 200
+    response = get_timeouts(session)
+    assert_success(response)
+    assert response.body["value"]["script"] == 60000
+    assert response.body["value"]["implicit"] == 1000
+    assert response.body["value"]["pageLoad"] == 200000
--- a/webdriver/tests/sessions/get_timeouts.py
+++ b/webdriver/tests/sessions/get_timeouts.py
@@ -24,15 +24,3 @@ def test_get_timeouts(session):
     assert isinstance(response.body["value"]["script"], int)
     assert isinstance(response.body["value"]["implicit"], int)
     assert isinstance(response.body["value"]["pageLoad"], int)
-
-    assert response.body["value"]["script"] == 30000
-    assert response.body["value"]["implicit"] == 0
-    assert response.body["value"]["pageLoad"] == 300000
-
-    session.timeouts.script = 60
-    session.timeouts.implicit = 1
-    session.timeouts.page_load = 200
-    response = get_timeouts(session)
-    assert response.body["value"]["script"] == 60000
-    assert response.body["value"]["implicit"] == 1000
-    assert response.body["value"]["pageLoad"] == 200000
--- a/webdriver/tests/sessions/get_timeouts.py
+++ b/webdriver/tests/sessions/get_timeouts.py
@@ -24,3 +24,15 @@ def test_get_timeouts(session):
     assert isinstance(response.body["value"]["script"], int)
     assert isinstance(response.body["value"]["implicit"], int)
     assert isinstance(response.body["value"]["pageLoad"], int)
+
+    assert response.body["value"]["script"] == 30000
+    assert response.body["value"]["implicit"] == 0
+    assert response.body["value"]["pageLoad"] == 300000
+
+    session.timeouts.script = 60
+    session.timeouts.implicit = 1
+    session.timeouts.page_load = 200
+    response = get_timeouts(session)
+    assert response.body["value"]["script"] == 60000
+    assert response.body["value"]["implicit"] == 1000
+    assert response.body["value"]["pageLoad"] == 200000
--- a/webdriver/tests/sessions/get_timeouts.py
+++ b/webdriver/tests/sessions/get_timeouts.py
@@ -1,5 +1,4 @@
-from tests.support.asserts import assert_error, assert_success
-from tests.support.inline import inline
+from tests.support.asserts import assert_success
 
 
 def get_timeouts(session):
@@ -12,6 +11,15 @@ def get_timeouts(session):
 def test_get_timeouts(session):
     # 8.4 step 1
     response = get_timeouts(session)
+
+    assert response.status == 200
+    # Check if value exist
+    assert 'value' in response.body
+
+    # Check if script, implicit, pageLoad exist
+    assert 'script' in response.body["value"]
+    assert 'implicit' in response.body["value"]
+    assert 'pageLoad' in response.body["value"]
 
     assert isinstance(response.body["value"]["script"], int)
     assert isinstance(response.body["value"]["implicit"], int)
--- a/webdriver/tests/sessions/get_timeouts.py
+++ b/webdriver/tests/sessions/get_timeouts.py
@@ -12,16 +12,15 @@ def test_get_timeouts(session):
     # 8.4 step 1
     response = get_timeouts(session)
 
-    assert response.status == 200
-    # Check if value exist
-    assert 'value' in response.body
+    assert_success(response)
+    assert "value" in response.body
+    assert isinstance(response.body["value"], dict)
 
-    # Check if script, implicit, pageLoad exist
-    assert 'script' in response.body["value"]
-    assert 'implicit' in response.body["value"]
-    assert 'pageLoad' in response.body["value"]
+    value = response.body["value"]
+    assert "script" in value
+    assert "implicit" in value
+    assert "pageLoad" in value
 
     assert isinstance(response.body["value"]["script"], int)
     assert isinstance(response.body["value"]["implicit"], int)
     assert isinstance(response.body["value"]["pageLoad"], int)
-    return assert_success(response)
--- a/webdriver/tests/sessions/get_timeouts.py
+++ b/webdriver/tests/sessions/get_timeouts.py
@@ -1,0 +1,19 @@
+from tests.support.asserts import assert_error, assert_success
+from tests.support.inline import inline
+
+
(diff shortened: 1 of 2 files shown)


webdriver/tests/sessions/get_timeouts.py
```diff
@@ -0,0 +1,19 @@
...
+                                  .format(session_id=session.session_id))
+
+
+# 8.4 Get Timeouts
+
+def test_get_timeouts(session):
+    # 8.4 step 1
+    response = get_timeouts(session)
+
+    assert isinstance(response.body["value"]["script"], int)
```
andreastt: You need to verify that `value` is in `response.body`.

webdriver/tests/sessions/get_timeouts.py
```diff
@@ -0,0 +1,19 @@
...
+    assert isinstance(response.body["value"]["implicit"], int)
```
andreastt: Would probably also be good for test readability if you checked that `script`, `implicit`, and `pageLoad` were on the `value` object in `response.body` for better readability in case the test should fail.

webdriver/tests/sessions/get_timeouts.py
```diff
@@ -0,0 +1,19 @@
...
+    assert isinstance(response.body["value"]["pageLoad"], int)
+    return assert_success(response)
```
andreastt: Why are you returning here?  Also, the response code ought be checked first. If it’s not 200, there’s no point in running the rest of these checks.

andreastt (changes requested):
This looks good so far, but you probably ought to check that the values match those given in the timeouts configuration object on session creation as well.

webdriver/tests/sessions/get_timeouts.py
alchiem: I am returning assert_success because Step 8.4.2 says "Return success with data body." Is that the correct way to do it? Should I even return the data body at all?

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/18758)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/278395925)

webdriver/tests/sessions/get_timeouts.py
alchiem: fixed

webdriver/tests/sessions/get_timeouts.py
alchiem: fixed

webdriver/tests/sessions/get_timeouts.py
andreastt: A test isn’t a literal implementation of the remote end steps: its purpose is, given some input, to check both the formal aspects and the semantics of a given command.

In this case the remote end is told to return a successful response and the test needs to assert that the implementation conforms to this behaviour. If however the response from `get_timeouts` is an error, the following line will error because there is no `value` field on `response.body`:

    assert isinstance(response.body["value"]["script"], int)

You need to first check whether the response was successful because this is a precursor for testing that the successful response’s JSON body is set up in the way it’s meant to be. In this case you need to change to test to do this:

    response = get_timeout(session)

    assert response.status == 200
    assert "value" in response.body
    assert isinstance(response.body["value"], dict)

    assert "script" in value
    assert "implicit" in value
    assert "pageLoad" in value
    assert isinstance(value["script"], int)
    assert isinstance(value["implicit"], int)
    assert isinstance(value["pageLoad"], int)

webdriver/tests/sessions/get_timeouts.py
```diff
@@ -0,0 +1,27 @@
...
+    response = get_timeouts(session)
+
+    assert response.status == 200
+    # Check if value exist
+    assert 'value' in response.body
+
+    # Check if script, implicit, pageLoad exist
+    assert 'script' in response.body["value"]
+    assert 'implicit' in response.body["value"]
+    assert 'pageLoad' in response.body["value"]
```
andreastt: Can you use double-quotes consistently?

webdriver/tests/sessions/get_timeouts.py
```diff
@@ -0,0 +1,27 @@
...
+
+# 8.4 Get Timeouts
+
+def test_get_timeouts(session):
+    # 8.4 step 1
+    response = get_timeouts(session)
+
+    assert response.status == 200
+    # Check if value exist
+    assert 'value' in response.body
```
andreastt: Drop this comment. It only distracts.

webdriver/tests/sessions/get_timeouts.py
andreastt: Your last update almost fixes the issues I mentioned in the last comment. However, there is no need for this:

    return assert_success(response)

You are correct that the remote end is being asked to return a successful response. The test isn’t.

andreastt (changes requested):
Your changes to assert the response’s status and that fields appear on the response’s JSON body are good, but like I mentioned in my last pass on this review, I’m still missing a test where the timeouts configuration object is changed on starting the session.

You also need to check the defaults of the session timeouts.

webdriver/tests/sessions/get_timeouts.py
alchiem: fixed

webdriver/tests/sessions/get_timeouts.py
alchiem: fixed

webdriver/tests/sessions/get_timeouts.py
alchiem: fixed

alchiem:
Hey @andreastt, so you also want me to check the initial values of the timeouts? Looking at the client.py. The initial values are obtained by using the get timeouts route which is what I did in my test. Would that be too redundant?

"I’m still missing a test where the timeouts configuration object is changed on starting the session."
And you wanted the check where the timeout configuration object is a dict?

andreastt:
> you also want me to check the initial values of the timeouts? Looking at the client.py. The initial values are obtained by using the get timeouts route which is what I did in my test. Would that be too redundant?

I want you to test the default values of [session script timeout](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-session-script-timeout), [session page load timeout](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-session-page-load-timeout), and [session implicit wait timeout](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-session-implicit-wait-timeout). These should be, respectively, 30,000 ms, 300,000 ms, and 0 ms.

>> I’m still missing a test where the timeouts configuration object is changed on starting the session.
> 
> And you wanted the check where the timeout configuration object is a dict?

You could start a new session, give it a [session timeouts configuration object](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-session-timeouts-configuration) and test that the given values are reflected when calling Get Timeouts.

alchiem:
Hey @andreastt,

For this part "You could start a new session, give it a session timeouts configuration object and test that the given values are reflected when calling Get Timeouts."
So right now, it seems like none of the browsers support setting timeouts. Even if I pass a new Timeouts object (client.py), I have to set each timeout to another value and it will throw an error saying key is not found in the Timeout._set method. I was getting an empty dictionary from line 46: timeouts = self.session.send_session_command("POST", "timeouts", body). Is this route suppose to be returning the timeout object and returning it?

webdriver/tests/sessions/get_timeouts.py
```diff
@@ -0,0 +1,38 @@
...
+    assert isinstance(response.body["value"], dict)
+
+    value = response.body["value"]
+    assert "script" in value
+    assert "implicit" in value
+    assert "pageLoad" in value
+
+    assert isinstance(response.body["value"]["script"], int)
+    assert isinstance(response.body["value"]["implicit"], int)
+    assert isinstance(response.body["value"]["pageLoad"], int)
```
andreastt: Here you can use the `value` variable.

webdriver/tests/sessions/get_timeouts.py
```diff
@@ -0,0 +1,38 @@
...
+
+    assert response.body["value"]["script"] == 30000
+    assert response.body["value"]["implicit"] == 0
+    assert response.body["value"]["pageLoad"] == 300000
```
andreastt: Same.

webdriver/tests/sessions/get_timeouts.py
```diff
@@ -0,0 +1,38 @@
...
+
+    session.timeouts.script = 60
+    session.timeouts.implicit = 1
+    session.timeouts.page_load = 200
+    response = get_timeouts(session)
+    assert response.body["value"]["script"] == 60000
+    assert response.body["value"]["implicit"] == 1000
+    assert response.body["value"]["pageLoad"] == 200000
```
andreastt: Make this a separate test.

webdriver/tests/sessions/get_timeouts.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def get_timeouts(session):
+    return session.transport.send("GET", "session/{session_id}/timeouts"
+                                  .format(session_id=session.session_id))
+
+
+# 8.4 Get Timeouts
+
+def test_get_timeouts(session):
```
andreastt: Can you split this into three tests? One for checking the payload (up to line 26), one for checking the session’s default values (lines 28 thru 30), and one for checking that _Get Timeouts_ returns the right values after having invoked `session.timeouts.script = 42`.

andreastt (changes requested):
This looks much better now.

andreastt:
> For this part "You could start a new session, give it a session timeouts configuration object and test that the given values are reflected when calling Get Timeouts."
> 
> So right now, it seems like none of the browsers support setting timeouts. Even if I pass a new Timeouts object (client.py), I have to set each timeout to another value and it will throw an error saying key is not found in the Timeout._set method. I was getting an empty dictionary from line 46: timeouts = self.session.send_session_command("POST", "timeouts", body). Is this route suppose to be returning the timeout object and returning it?

You should confer with the specification. It has the answers to all these questions.

The [New Session](https://w3c.github.io/webdriver/webdriver-spec.html#new-session) command is meant to take a `timeouts` field from the capabilities that the same session timeouts configuration object that is received by [Set Timeouts](https://w3c.github.io/webdriver/webdriver-spec.html#set-timeouts). Upon creating a new session, this should override the default timeout values.

If none of the current implementations support this, that is a bug and we need to have a test for it. In Mozilla’s CI system specifically, the failing tests will be annotated with metadata once we import the test so that we can keep track of what features geckodriver doesn’t support. This is very useful to us an implementor. So we should definitely write tests that are failing in current implementations.

webdriver/tests/sessions/get_timeouts.py
alchiem: fixed

webdriver/tests/sessions/get_timeouts.py
alchiem: fixed

webdriver/tests/sessions/get_timeouts.py
alchiem: fixed

webdriver/tests/sessions/get_timeouts.py
alchiem: fixed

andreastt (approved):

PR #7465, opened 2017-09-25
Fix manifest/lint to deal with whitespace in link@href

gsnedders:
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7465)


tools/manifest/tests/test_sourcefile.py
```diff
@@ -542,3 +542,23 @@ def test_no_parse():
...
     s = create(input, b"")
     assert s.rel_path == expected
+
+
+@pytest.mark.parametrize("url", [b"ref.html",
+                                 b"\x20ref.html",
+                                 b"ref.html\x20",
+                                 b"\x09\x0a\x0c\x0d\x20ref.html\x09\x0a\x0c\x0d\x20"])
+def test_reftest_url_whitespace(url):
+    content = b"<link rel=match href=%s>" % url
```
Ms2ger: Does this even test this code, rather than the html parser? It seems to check

```html
<link rel=match href= ref.html>
```

rather than an actual attribute with a space.

tools/manifest/tests/test_sourcefile.py
```diff
@@ -542,3 +542,23 @@ def test_no_parse():
...
+    s = create("foo/test.html", content)
+    assert s.references == [("/foo/ref.html", "==")]
+
+
+@pytest.mark.parametrize("url", [b"http://example.com/",
+                                 b"\x20http://example.com/",
+                                 b"http://example.com/\x20",
+                                 b"\x09\x0a\x0c\x0d\x20http://example.com/\x09\x0a\x0c\x0d\x20"])
+def test_spec_links_whitespace(url):
+    content = b"<link rel=help href=%s>" % url
```
Ms2ger: Ditto

Ms2ger (changes requested):
Not sure if this is worth it, especially with just one test involved.

tools/manifest/tests/test_sourcefile.py
gsnedders: As it happens the HTML parser ignores the spaces there so the test is valid (and hence it passes).

gsnedders:
I have a slight preference for just matching HTML here rather than requiring something more specific, but maybe we should be conservative here…

w3c-bots:
# Build PASSED

Started: 2017-09-25 13:20:05
Finished: 2017-09-25 14:03:31


<h2>Failing Jobs</h2>
<ul>

<li>safari:10.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/18807)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/279498086)

Ms2ger (approved):
Feel free to land if jgraham doesn't object in the next week or so.

jgraham:
Sure.

PR #7471, opened 2017-09-25
fetch: Reduce flakiness in integrity.html.

rakuco:
Commit 96adf8a1a0cd ("P4: Modify wpt test to verify only the opaque response
with the empty string will pass the SRI check") stored the promise returned
by fetch() in a separate variable, fetchPromise, which is later used in
calls to promise_test() and promise_rejects().

In the test cases where a fetch failure is expected there is a window
between fetch() returning a rejected promise and promise_rejects() being
called to connect a proper handler for the rejection.

While everything still works correctly, when this happens testharness'
"unhandledrejection" event handler will log a non-existent failure which can be
included in the test's output and cause flakiness (it happens at least with
Blink).

Fix it by dropping fetchPromise in favor of two separate calls to fetch(),
just like the code did in the past.

### DIFF
--- a/fetch/api/basic/integrity.js
+++ b/fetch/api/basic/integrity.js
@@ -8,11 +8,10 @@ function integrity(desc, url, integrity, initRequestMode, shouldPass) {
   if (!!initRequestMode && initRequestMode !== "") {
     fetchRequestInit.mode = initRequestMode;
   }
-  var fetchPromise = fetch(url, fetchRequestInit);
 
   if (shouldPass) {
     promise_test(function(test) {
-      return fetchPromise.then(function(resp) {
+      return fetch(url, fetchRequestInit).then(function(resp) {
         if (initRequestMode !== "no-cors") {
           assert_equals(resp.status, 200, "Response's status is 200");
         } else {
@@ -23,7 +22,7 @@ function integrity(desc, url, integrity, initRequestMode, shouldPass) {
     }, desc);
   } else {
     promise_test(function(test) {
-      return promise_rejects(test, new TypeError(), fetchPromise);
+      return promise_rejects(test, new TypeError(), fetch(url, fetchRequestInit));
     }, desc);
   }
 }


annevk:
I'd like @jgraham to sign off on this.

It's also not clear to me how the "error" event handler gets invoked here.

rakuco:
> It's also not clear to me how the "error" event handler gets invoked here.

AFAICS, one of those calls to `integrity()` that is expected to fail (in particular the http/https ones) invokes `fetch()`, validation fails and we end up with a promise that is rejected with a `TypeError`. Depending on how many calls are being done, Blink (and apparently Gecko as well) can reject that promise before the code gets to the `if (shouldPass) {}` block, the unhandled rejection's exception is rethrown and we reach the `error_handler` function in `testharness.js`.

When running the tests in Blink, I can see messages such as "CONSOLE ERROR: Uncaught (in promise) TypeError: Failed to fetch"; with Firefox Nightly, launching https://w3c-test.org/fetch/api/basic/integrity.html shows me one or more error messages in the inspector console saying "TypeError: NetworkError when attempting to fetch resource".

annevk:
But don't you need a specific promise rejection event handler to catch those in code? I don't see how the harness gets them without that. The error consoles displaying something doesn't really affect the state of the harness (or the correctness of the test).

rakuco:
> But don't you need a specific promise rejection event handler to catch those in code? I don't see how the harness gets them without that.

Oops, I managed to miss the "unhandledrejection" handler that's defined right below the call to `addEventListener("error", ...")` in testharness. I've updated my commit message.

> The error consoles displaying something doesn't really affect the state of the harness (or the correctness of the test).

Yeah, these were just examples showing sometimes these exceptions were thrown. In Blink, these errors are later reported as part of the test's results, so we can end up with a line saying "Harness Error. harness_status.status = 1 , harness_status.message = Failed to fetch".

With all that said, if things are still working as intended I can start a discussion about this behavior in Blink instead.

annevk (approved):
Given the new explanation I'm happy with this landing and I don't think review from @jgraham is required at this point.

Can you land this yourself once Travis passes?

w3c-bots:
# Build PASSED

Started: 2017-09-25 17:42:26
Finished: 2017-09-25 17:58:45


<h2>Failing Jobs</h2>
<ul>

<li>chrome:unstable</li>

<li>safari:10.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/18816)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/279558329)

annevk:
I think Firefox might not implement unhandledrejection yet btw, which is probably why it's not flaky there.

rakuco:
Thanks :-)

PR #7510, opened 2017-09-27
Test SpeechSynthesisUtterance reuse within and across frames

foolip:
Tentative test for expected spec change in https://github.com/w3c/speech-api/issues/7


foolip:
w3c-test:mirror

foolip:
The test now passes in Chrome bit times out in Firefox, tested manually, and as expected. (Or maybe I had expected Firefox to throw an exception per spec, but anyway.)

https://github.com/w3c/web-platform-tests/pull/7507 adds OWNERS for speech-api, so merging that first would be nice. @andrenatal, can you review this change? Once the spec is converted to Bikeshed I'd be happy to make the same fix spec side, and then hopefully nothing is blocking a fix in Gecko.

foolip:
Also tested Edge and that passes the test, but Safari actually times out, like Firefox! So this was a 2/2 split in behavior, hope we can get it fixed!

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/18907)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/280478413)

foolip:
Hmm, that's interesting, flaky on Safari. I'm pretty sure that this is actually not a problem with the test, it's not doing anything that should reasonably be racy, undefined behavior, or very slow. @youennf, can you help me find a contact for Speech API at WebKit?

fleizach:
Safari does not allow re-usable speech utterances at this time. We'll have to fix up some stuff to allow that

foolip:
> Safari does not allow re-usable speech utterances at this time. We'll have to fix up some stuff to allow that

It actually passes for me and I hear "one one" on Safari on my own machine, it seems like it usually works but there's some flakiness going on.

BTW, can I add you to speech-api/OWNERS, so that you see test reviews for Speech API?

fleizach:
If it passes it's probably out of luck rather than design. Mileage may vary on iOS as well, where the native synthesis API (AVSpeechSynthesizer) explicitly disallows re-using the same utterance.

Yes, please add me to the owners list

Thanks

foolip:
Great, I'll add you in the other PR.

foolip:
Rebased so that the OWNERS file exists, but not sure if that's enough

foolip:
Oops, reopening :)

foolip:
Hmm, I added another test for the "Using the same SpeechSynthesisUtterance for different SpeechSynthesis instances" case, but the test is now flaky on Chrome too. Maybe it'd be best to have them as separate tests. It seems like the utterances are queued up, so the tests can interact. Better to have separate tests that deliberately tests the interaction.

foolip:
I've split the tests. Ping @andrenatal @fleizach @gshires for review.

kereliuk (approved):

PR #7530, opened 2017-10-01
Adds for XMLHttpRequest's documentXML and responseType in a worker

samweinig:
Adds tests that ensure that XMLHttpRequest's documentXML is not available in a worker and that setting responseType to 'document' is no-op in a worker.

### DIFF
--- a/XMLHttpRequest/resources/responseType-document-in-worker.js
+++ b/XMLHttpRequest/resources/responseType-document-in-worker.js
@@ -1,0 +1,9 @@
+self.importScripts('/resources/testharness.js');
+
+test(function() {
+    let xhr = new XMLHttpRequest();
+    xhr.responseType = "document";
+    assert_not_equals(xhr.responseType, "document");
+}, "Setting XMLHttpRequest responseType to 'document' in a worker should have no effect.");
+
+done();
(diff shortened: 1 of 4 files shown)


wisniewskit:
I'm okay with this given that it's how browser behave, but we should also add this requirement to the spec-text. Should we keep the two in sync or land such tests prior to the spec changing?

samweinig:
> I'm okay with this given that it's how browser behave, but we should also add this requirement to the spec-text. Should we keep the two in sync or land such tests prior to the spec changing?

This is all in the spec already.

- responseType-document-in-worker.html is testing step 1 of https://xhr.spec.whatwg.org/#the-responsetype-attribute.
- responseXML-unavailable-in-worker.html is testing the [Exposed=Window] extended attribute in the IDL.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/18998)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/281945193)

annevk (approved):

annevk:
I went ahead and merged this given @samweinig's clarification as that matches my understanding. Feel free to follow-up though if there's something still to be done.

annevk:
Oh, I didn't see this was created only 19 minutes ago. Sorry for not waiting a bit until everyone was on the same page.

wisniewskit:
Ah, you're right, I was misreading the responseXML section (just a bit tired today, I guess).

PR #7534, opened 2017-10-02
Remove EventSource tests for DNS failures

yutakahirano:
The spec says:

> Network errors that prevents the connection from being established in the first place (e.g. DNS errors), should cause the user agent to reestablish the connection in parallel, unless the user agent knows that to be futile, in which case the user agent may fail the connection.

Hence the readyState expectation for http://example.not/ and https://example.not/test is not correct.

### DIFF
--- a/eventsource/eventsource-constructor-non-same-origin.htm
+++ b/eventsource/eventsource-constructor-non-same-origin.htm
@@ -7,7 +7,6 @@
     <script src="/resources/testharnessreport.js"></script>
   </head>
   <body>
-    <div id="log"></div>
     <script>
       function fetchFail(url) {
         var test = async_test(document.title + " (" + url + ")", { timeout: 20000 })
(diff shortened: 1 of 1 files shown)


yutakahirano:
@annevk

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19012)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/282116917)

annevk:
Firefox passes these tests, so they are indicative of some subtle interoperability issue. Should we not try to resolve that somehow?

yutakahirano:
To be honest, I think `readyState` should not be CLOSED here based on the spec sentence I quoted.

annevk:
The user agent cannot know the connection is futile for these domains?

If that's the case it would be good to test the opposite I suppose, though I guess we shouldn't really use external domains in WPT.

yutakahirano:
...because of the phrase `(e.g. DNS errors)`.  If you think such errors are persistent, the phrase is confusing.

annevk (approved):

annevk:
At some point we should probably clean up the specification a bit, but I guess that won't become a priority anytime soon.

PR #7538, opened 2017-10-02
Disable unneeded Firefox features that can interfere with tests

AutomatedTester:
A number of these prefs are used with Firefox services and are not part
of the web platform.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7538)

### DIFF
--- a/tools/wptrunner/wptrunner/browsers/firefox.py
+++ b/tools/wptrunner/wptrunner/browsers/firefox.py
@@ -186,11 +186,33 @@ class FirefoxBrowser(Browser):
         self.profile = FirefoxProfile(locations=locations,
                                       preferences=preferences)
         self.profile.set_preferences({"marionette.port": self.marionette_port,
+                                      "browser.aboutHomeSnippets.updateUrl": "",
+                                      "browser.newtabpage.activity-stream.enabled": False,
+                                      "browser.newtabpage.activity-stream.feeds.snippets": False,
+                                      "browser.newtabpage.activity-stream.telemetry.ping.endpoint": "",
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PASSED

Started: 2017-10-02 11:43:28
Finished: 2017-10-02 12:07:39





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19018)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/282208478)

jgraham:
Are these prefs not set in `prefs-general.js`? That is supposed to be supplied to the profile when we start Firefox in wpt. However for wdspec tests it's possible that they *don't* get set since we pass fewer prefs in via `executor_kwargs`. But also I don't think the environment variable to crash on external network connections is set in that case.

AutomatedTester:
@jgraham perhaps I am doing something wrong then. I am getting `0:22.85 PROCESS_OUTPUT: ProcessReader (pid:75732) "FATAL ERROR: Non-local network connections are disabled and a connection attempt to location.services.mozilla.com (52.213.89.190) was made."` when running things at the moment through wptdashboard.

AutomatedTester:
ok. I managed to debug the problem. There were two problems I was hitting. First was a problem with finding geckodriver. The error wasnt obvious, to me at least, and then it started running testharness tests which started hitting `MOZ_DISABLE_NONLOCAL_CONNECTIONS` errors. This patch fixes it but might not be right. 

Feel free to use or close this issue.

jgraham:
I think any relevant parts of this patch should end up in prefs-general.js on m-c.

PR #7559, opened 2017-10-03
Introduce a new test_idl function that simplifies using idlharness with external IDL files.

Ms2ger:
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7559)


foolip (commented):
Can you also convert webrtc/interfaces.html to show that this helper can deal well (enough) with the case where the objects for testing aren't known synchronously?

cssom-view/interfaces.html
```diff
@@ -17,51 +17,34 @@
...
 };
 
-function fetchData(url) {
-  return fetch(url).then((response) => response.text());
-}
-
-function waitForLoad() {
-  return new Promise(function(resolve) {
-    addEventListener("load", resolve);
+addEventListener("load", function() {
```
foolip: Why wait for the load event?

resources/idlharness.js
```diff
@@ -122,6 +122,55 @@ var fround =
...
+        for (var name of prevent_multiple_testing) {
+            idlArray.prevent_multiple_testing(name);
+        }
+        idlArray.test();
+    };
+
+    function fetchData(urls) {
+        function handle(response) {
+            if (!response.ok) {
+                throw new TypeError("Response is not ok for " + response.url);
```
foolip: What are the cases where the response is not ok but the promise isn't rejected? Seems odd :)

cssom/interfaces.html
```diff
@@ -29,61 +29,43 @@
 
 <script>
 "use strict";
-var style_element, svg_element, xmlss_pi;
+var style_element = document.getElementById('styleElement');
+var svg_element = document.getElementById('svgElement');
+var xmlss_pi = document.getElementById('xmlssPiIframe').contentDocument.firstChild;
```
jdm: There's no guarantee that the content document has finished loading yet, right? Shouldn't this be in the load handler?

resources/idlharness.js
```diff
@@ -122,6 +122,55 @@ var fround =
 })();
 //@}
 
+self.test_idl = function({
+    untested_code = "",
+    untested_urls = [],
+    tested_code = "",
```
jdm: I don't see any users of tested_code.

jdm (changes requested):

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19087)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/282767247)

Ms2ger:
#11833 did basically this, but actually landed.

PR #7579, opened 2017-10-04
Baseline of inline-block for different values of overflow, for an errata to CSS2

bert-github:
Some tests for errata http://www.w3.org/Style/css2-updates/REC-CSS2-20110607-errata.html#s.10.8.1
(An inline-block with 'overflow' other than 'visible' but that doesn't actually overflow is aligned on the baseline of its last line box. Previously on its bottom margin.)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19147)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/283228578)

dbaron:
Sorry for taking so long to get to this:  these look good, and appear to agree with the resolution in https://lists.w3.org/Archives/Public/www-style/2015Apr/0144.html .  I'll squash and merge them.  (There's a *lot* of unrelated github history in this pull request that makes it a really bad idea to do anything else.)

It would probably be good to also have tests at the non-initial scroll position, though.

dbaron (approved):

PR #7585, opened 2017-10-05
Revert "Modify errors from try server,"

foolip:
This reverts commit 05e8454f8659d4943c8076cbcaab15085395e5d6.

This appears to have been an accident, see
https://github.com/w3c/web-platform-tests/pull/7531#issuecomment-333397939

### DIFF
--- a/cookies/secure/set-from-dom.sub.html
+++ b/cookies/secure/set-from-dom.sub.html
@@ -1,0 +1,47 @@
+<!doctype html>
+<html>
+<head>
+  <meta charset=utf-8>
+  <title>Set 'secure' cookie from `document.cookie` on a non-secure page</title>
+  <meta name=help href="https://tools.ietf.org/html/draft-west-leave-secure-cookies-alone">
+  <script src="/resources/testharness.js"></script>
+  <script src="/resources/testharnessreport.js"></script>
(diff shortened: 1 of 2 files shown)


foolip:
@mikewest, actually, these aren't tests, and look like unused files. Maybe deleting these was the right thing to do?

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19178)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/283544394)

mikewest:
Why aren't they tests? They look like tests to me?

foolip:
Oh, they are, and they passed on at least Firefox. I thought .sub. was only used resource files because that's what I've seen a lot of.

So, these should be revived?

Hexcles:
@jgraham also confirmed on IRC this was likely an accident https://logs.csswg.org/irc.w3.org/testing/2017-10-03/#e885259

We should definitely try to detect such cases in the (future) audit tool.

mikewest:
Yes! Please don't delete my tests without putting them back afterwards. :)

This patch LGTM (and `.sub.` totally works on main resources). :)

Hexcles (approved):
Let me try if I can rubber stamp. :)

PR #7632, opened 2017-10-06
[css-variables] Test attribute changes that affect generated content defined via CSS variables.

emilio:
This adds a test for:

  https://bugs.chromium.org/p/chromium/issues/detail?id=772316

Testcase passes in Gecko and WebKit, not in Blink. Haven't tested Edge.

### DIFF
--- a/css/css-variables-1/variable-generated-content-dynamic-001-ref.html
+++ b/css/css-variables-1/variable-generated-content-dynamic-001-ref.html
@@ -1,0 +1,4 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Test Reference</title>
+<div>PASS</div>
--- a/css/css-variables-1/variable-generated-content-dynamic-001.html
+++ b/css/css-variables-1/variable-generated-content-dynamic-001.html
@@ -1,0 +1,19 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Test: Dynamic attribute change handling in generated content</title>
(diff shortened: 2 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19251)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/284462309)

Ms2ger:
Is there a particular reason to use variables at all in the reference?

emilio:
Nope, but it doesn't, right?

css/css-variables-1/variable-generated-content-dynamic-001-ref.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Test Reference</title>
+<style>
+div::before {
+  content: attr(data-foo);
+}
+</style>
+<div data-foo="PASS"></div>
```
gsnedders: Does `<div>PASS</div>` not suffice here?

emilio: It probably should, but blink has had problems with reftests that exercised text-shaping across different node boundaries (see https://github.com/w3c/csswg-test/pull/1143). That being said, that is being fixed, so I'll update, I guess.

emilio: Just did that.

gsnedders (approved):

PR #7633, opened 2017-10-07
css generated content: added test for counter-* calc() values

ondras:
This new test checks support for `calc()` expressions in `counter-reset` and `counter-increment` properties.

### DIFF
--- a/css/CSS2/generated-content/content-counter-017-ref.xht
+++ b/css/CSS2/generated-content/content-counter-017-ref.xht
@@ -6,20 +6,11 @@
   <link rel="help" href="http://www.w3.org/TR/CSS21/generate.html#counters"/>
   <link rel="help" href="http://www.w3.org/TR/CSS21/generate.html#propdef-content"/>
   <link rel="help" href="http://www.w3.org/TR/CSS21/syndata.html#counter"/>
-  <link rel="match" href="content-counter-017-ref.xht"/>
   <meta name="flags" content="dom"/>
   <style type="text/css">
-
   body {
    white-space: nowrap;
   }
-
-  #a { counter-reset: a calc(3 + 5); }
-  #a::before { content: counter(a); }
-
-  #b { counter-reset: b 0; counter-increment: b calc(4 + 6); }
-  #b::before { content: counter(b); }
-
   </style>
  </head>
  <body>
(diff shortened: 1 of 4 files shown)


w3c-bots:
# Build PASSED

Started: 2017-10-07 17:21:31
Finished: 2017-10-07 17:26:25


<h2>Failing Jobs</h2>
<ul>

<li>build_css</li>

<li>safari:10.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/19255)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/284738837)

Ms2ger (dismissed):
Can you please turn this into a reftest? This will allow the test to be run automatically as part of the test infrastructure browsers use.

See <http://web-platform-tests.org/writing-tests/reftests.html> and don't hesitate to ask if you have any questions.

ondras:
> Can you please turn this into a reftest? This will allow the test to be run automatically as part of the test infrastructure browsers use.

Ah, I was not aware of reftests. Will try my best.

ondras:
@Ms2ger, I added a reftest for this.

bert-github (commented):
Test is correct, but it seems to be a test for https://www.w3.org/TR/css-values-3/#calc-notation and https://www.w3.org/TR/css-lists-3/#counter-properties rather than for CSS2.There is no calc() in CSS level 2.

So I suggest to change the `<link rel=help>` elements to the two URLs above.

ondras:
@bert-github thanks, I adjusted those URLs.

dbaron:
Sorry this seems to have languished...

This seems to be good except for the filenames, which are still placing the tests within the CSS2 directory.  It's probably best to put them in `css/css-values`, and maybe call them `calc-in-counter-001.xhtml` etc. to match the files there, to go along with addressing @bert-github's comments.  I think they could still have rel=help links that point to the CSS2 definitions of `counter-reset` and `counter-increment` as well.  (I'd note that if you want you *could* change them from XHTML to HTML, but if so that should be reflected in both the filenames and the contents.)

Two other minor changes:
* you should remove the `<meta name="flags" content="dom"/>` since the tests don't actually use script.
* you should remove the `<link rel="help" ...>` from the reference; they should be only in the test

dbaron (dismissed):
see previous comment; something this week put me into the habit of writing reviews... outside of the reviews UI

ondras:
@dbaron thanks for your comment. I moved/renamed the files as suggested, cleaned up their heads and re-introduced help links to CSS2 counter spec.

svgeesus (approved):
I believe the review commends from @dbaron and @bert-github have been addressed; the test and reference seem good to me.

Sorry the review took so long!

PR #7659, opened 2017-10-10
File/Directory Entries API: Make tests Edge friendly

inexorabletash:
Edge doesn't yet implement getFile()/getDirectory()/getParent() so don't make unrelated tests assume the existence of those methods. Add a helper to get child entries via enumeration (supported everywhere) and also split up the test files by method.

### DIFF
--- a/entries-api/filesystementry-getParent-manual.html
+++ b/entries-api/filesystementry-getParent-manual.html
@@ -1,34 +1,16 @@
 <!DOCTYPE html>
 <meta charset=utf-8>
-<title>Entries API: FileSystemEntry manual test</title>
-<link rel=help href="https://wicg.github.io/entries-api/#api-entry">
+<title>Entries API: FileSystemEntry getParent() manual test</title>
+<link rel=help href="https://wicg.github.io/entries-api/#dom-filesystementry-getparentapi-entry">
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <script src="support.js"></script>
 
 <script>
 entry_test((t, entry) => {
-  assert_idl_attribute(entry, 'isFile', 'FileSystemEntry has isFile attribute');
-  assert_equals(typeof entry.isFile, 'boolean', 'isFile is boolean');
-
-  assert_idl_attribute(entry, 'isDirectory', 'FileSystemEntry has isDirectory attribute');
-  assert_equals(typeof entry.isDirectory, 'boolean', 'isFile is boolean');
-
-  assert_idl_attribute(entry, 'name', 'FileSystemEntry has name attribute');
-  assert_equals(typeof entry.name, 'string', 'name is a string');
-
-  assert_idl_attribute(entry, 'fullPath', 'FileSystemEntry has fullPath attribute');
-  assert_equals(typeof entry.fullPath, 'string', 'fullPath is a string');
-
-  assert_idl_attribute(entry, 'filesystem', 'FileSystemEntry has filesystem attribute');
-
   assert_idl_attribute(entry, 'getParent', 'FileSystemEntry has filesystem attribute');
   assert_equals(typeof entry.getParent, 'function', 'FileSystemEntry has a getParent() method');
 
-  t.done();
-}, 'FileSystemEntry - interface');
-
-entry_test((t, entry) => {
   assert_equals(entry.getParent(), void 0, 'getParent() arguments are optional');
 
   entry.getParent(t.step_func(parent => {
--- a/entries-api/filesystementry-getParent-manual.html
+++ b/entries-api/filesystementry-getParent-manual.html
@@ -8,7 +8,7 @@
 
 <script>
 entry_test((t, entry) => {
-  assert_idl_attribute(entry, 'getParent', 'FileSystemEntry has filesystem attribute');
+  assert_idl_attribute(entry, 'getParent', 'FileSystemEntry has getParent method');
   assert_equals(typeof entry.getParent, 'function', 'FileSystemEntry has a getParent() method');
 
   assert_equals(entry.getParent(), void 0, 'getParent() arguments are optional');
--- a/entries-api/support.js
+++ b/entries-api/support.js
@@ -10,7 +10,7 @@ window.addEventListener('DOMContentLoaded', e => {
   header.innerText = document.title;
   document.body.appendChild(header);
   const elem = document.createElement('div');
-  elem.style = 'height: 50px; border: 1px dotted red;';
+  elem.style.cssText = 'height: 50px; border: 1px dotted red;';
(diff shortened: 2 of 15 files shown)


inexorabletash:
@aliams @cdumez @mkruisselbrink @pwnall - could one of you take a look at these changes?

w3c-bots:
# Build PASSED

Started: 2017-10-10 21:42:01
Finished: 2017-10-10 21:49:07


<h2>Failing Jobs</h2>
<ul>

<li>chrome:unstable</li>

<li>safari:10.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/19300)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/286245265)

entries-api/support.js
```diff
@@ -10,15 +10,16 @@ window.addEventListener('DOMContentLoaded', e => {
...
   elem.innerHTML = 'Drop the <b>support/upload</b> directory here.</div>';
   document.body.appendChild(elem);
   elem.addEventListener('dragover', e => {
     e.preventDefault();
   });
   elem.addEventListener('drop', e => {
     e.preventDefault();
-    for (const item of e.dataTransfer.items) {
+    for (var i = 0; i < e.dataTransfer.items.length; ++i) {
+      var item = e.dataTransfer.items[i];
```
mkruisselbrink: nit: const?

entries-api/support.js
```diff
@@ -10,15 +10,16 @@ window.addEventListener('DOMContentLoaded', e => {
...
+  elem.style.cssText = 'height: 50px; border: 1px dotted red;';
   elem.innerHTML = 'Drop the <b>support/upload</b> directory here.</div>';
   document.body.appendChild(elem);
   elem.addEventListener('dragover', e => {
     e.preventDefault();
   });
   elem.addEventListener('drop', e => {
     e.preventDefault();
-    for (const item of e.dataTransfer.items) {
+    for (var i = 0; i < e.dataTransfer.items.length; ++i) {
```
mkruisselbrink: nit: this also doesn't have to be var, right?

entries-api/support.js
```diff
@@ -89,3 +87,43 @@ const FILE_PATHS = [
...
+// Promise.
+
+function getEntriesAsPromise(dirEntry) {
+  return new Promise((resolve, reject) => {
+    let result = [];
+    let reader = dirEntry.createReader();
+    let doBatch = () => {
+      reader.readEntries(entries => {
+        if (entries.length > 0) {
+          entries.forEach(e => result.push(e));
```
mkruisselbrink: nit: isn't this just Array.prototype.push.apply(result, entries)? Or result = result.concat(entries); Although I guess neither of those is necessarily cleaner/simpler than what you have.

entries-api/support.js
```diff
@@ -89,3 +87,43 @@ const FILE_PATHS = [
...
+}
+
+
+// Wrapper for FileSystemDirectoryReader that yields a single entry by
+// name via a callback. Can be used instead of getFile() or
+// getDirectory() since not all implementations support those.
+
+function getChildEntry(dirEntry, name, callback) {
+  getEntriesAsPromise(dirEntry)
+    .then(function(entries) {
```
mkruisselbrink: nit: entries => {} might be more consistent with the rest of these tests?

entries-api/support.js
```diff
@@ -89,3 +87,43 @@ const FILE_PATHS = [
...
+      return entries.filter(function(entry) {
```
mkruisselbrink: nit: here too, => rather than function?

entries-api/filesystementry-getParent-manual.html
```diff
@@ -0,0 +1,27 @@
...
+<meta charset=utf-8>
+<title>Entries API: FileSystemEntry getParent() manual test</title>
+<link rel=help href="https://wicg.github.io/entries-api/#dom-filesystementry-getparentapi-entry">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="support.js"></script>
+
+<script>
+entry_test((t, entry) => {
+  assert_idl_attribute(entry, 'getParent', 'FileSystemEntry has filesystem attribute');
```
mkruisselbrink: (I know this is just copied from before) getParent() rather than filesystem, right?

entries-api/support.js
```diff
@@ -89,3 +87,43 @@ const FILE_PATHS = [
...
+        return entry.name === name;
+      })[0];
+    })
+    .then(callback);
```
mkruisselbrink: Shouldn't you handle the promise rejecting to properly make whatever test called getChildEntry fail? Of course you might need to pass in the Test to getChildEntry for that to be possible.

entries-api/support.js
inexorabletash: Made `const`.

entries-api/support.js
inexorabletash: done

entries-api/support.js
inexorabletash: All correct. Leaving as-is, though. I could make all those `let`s into `const`s though.

entries-api/support.js
inexorabletash: Thanks, meant to fix this and forgot. Done!

entries-api/support.js
inexorabletash: done

entries-api/support.js
inexorabletash: ... or update all the tests to be `promise_test` based. I'll add an `errback`

entries-api/filesystementry-getParent-manual.html
inexorabletash: D'oh, nice catch. Fixed.

aliams:
Does the name consistency and object identity test need to use getParent in filesystem-manual.html?

aliams:
Also, we apparently don't support getFile as well.

inexorabletash:
> Does the name consistency and object identity test need to use getParent in filesystem-manual.html?

Good point - I was able to switch that to the `getChildEntry()` helper as well (which uses DirectoryReader)

> Also, we apparently don't support getFile as well.

Yeah, I've sequestered getParent()/getFile()/getDirectory() into specific tests (and the IDL ones). The only exception is errors-manual.html which is verifying all the DOMException details and I think I need getFile for that one.

aliams:
Thanks @inexorabletash! Looking over the tests now, the pass rate for what's actually needed to accomplish the directory upload scenario is much better now for Edge.

inexorabletash:
Thanks for confirming, @aliams !

And since github's workflow is a bit sketchy: @mkruisselbrink - I've addressed your feedback and pushed an update to the branch if you want to take another look

aliams (approved):

mkruisselbrink (approved):

PR #7663, opened 2017-10-10
Document the .tentative file name flag test feature

foolip:
Discussed in
https://lists.w3.org/Archives/Public/public-test-infra/2017AprJun/thread.html#msg59
and now already in use, with 30 tentative tests currently existing.

### DIFF
--- a/docs/_writing-tests/file-names.md
+++ b/docs/_writing-tests/file-names.md
@@ -54,5 +54,12 @@ themselves precede any test type flag, but are otherwise unordered.
  : (js files only) Indicates that the file generates tests in which it
     is run in Window and dedicated worker environments.
 
+`.tentative`
+ : Indicates that a test makes assertions not yet required by any specification,
+   or in contradiction to some specification. This is useful when implementation
+   experience is needed to inform the specification. It should be apparent in
+   context why the test is tentative and what needs to be resolved to make it
+   non-tentative.
+
 
 [server-side substitution]: https://wptserve.readthedocs.io/en/latest/pipes.html#sub


foolip:
@mikewest I think you may be the most active user of tentative tests. Is this documentation accurate?

w3c-bots:
# Build PASSED

Started: 2017-10-10 22:21:10
Finished: 2017-10-10 22:28:58


<h2>Failing Jobs</h2>
<ul>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/19304)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/286261263)

annevk:
Can we require such tests to point to a corresponding specification issue?

Ms2ger (approved):

foolip:
@annevk, I was hoping that "It should be apparent in context why the test is tentative and what needs to be resolved to make it non-tentative" would be enough here. Do you want me to also spell out that a link to an issue is the easiest way of doing that?

I think that if we really wanted to require it, we should have a lint checking for the existence of "https://" somewhere in tentative tests, but I'm pretty sure it would occasionally be annoying. For example, the problem with domxpath/interfaces.tentative.html is that there's no spec, and the comment clearly states so. However, the link isn't to an issue, it's to https://wiki.whatwg.org/wiki/DOM_XPath

annevk:
@foolip mkay, I guess we can revisit if it becomes a problem.

foolip:
Thanks! Merging now so that it can be linked to, happy to keep tweaking as needed.

PR #7665, opened 2017-10-11
Add Acid2 and Acid3

gsnedders:
cc/ @Ms2ger @zcorpan @annevk @bzbarsky


acid/README.md
```diff
@@ -0,0 +1,17 @@
+## The Acid Tests
```
gsnedders: @hixie does this README look good to you? do you want anything else here?

gsnedders:
(Please don't squash & merge, we should preserve the history of us diverging from upstream.)

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19308)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/286315237)

gsnedders:
I think Acid2 fails everywhere because `reference.png` doesn't pixel-for-pixel match, because the nose seems to be slightly (1px?) out.

gsnedders:
Also do any of you want to volunteer to be OWNERS of acid/?

Ms2ger:
> Also do any of you want to volunteer to be OWNERS of acid/?

Sure

Ms2ger:
Do we need a note about the license?

gsnedders:
> Do we need a note about the license?

I took the last email to mean no.

annevk:
Hmm, per his last message https://github.com/w3c/web-platform-tests/blob/master/LICENSE.md doesn't apply, unless we get sign off from Google and Opera.

Hixie:
I suspect the Acid3 copy here is missing all kinds of subtleties with the Apache configuration, FWIW.

gsnedders:
@Hixie Is there anything except for the `Content-Type` headers that's relevant here? (Is anything explicitly testing `Content-Encoding`?) I did base everything off of the full responses, so there's a bunch of stuff with bogus `Content-Type` headers.

Hixie:
I do not recall the details, it's been a long time. I just remember there was a complicated .htaccess file. There's been some tarballs made of this test in the past, I'm pretty sure that Mozilla, Opera, Chrome and WebKit all have had copies in the past. So it should be pretty easy to find.

gsnedders:
@Hixie Seems like we have everything, looking at the WebKit copy at least.

jgraham:
I'm happy with this landing in principle, but I haven't reviewed in detail to ensure that it's correct. I can approve on that basis, or we can get @Hixie or @annevk or someone to do a more thorough review.

annevk:
I suggest we land this provided @gsnedders has verified in at least two browsers that there's no different results between these and the originals.

PR #7671, opened 2017-10-11
Add a test for "keepalive" handling

yutakahirano:

### DIFF
--- a/fetch/api/resources/keepalive-iframe.html
+++ b/fetch/api/resources/keepalive-iframe.html
@@ -3,12 +3,7 @@
 <meta charset="utf-8">
 <script src="/common/utils.js"></script>
 <script>
-const xhr = new XMLHttpRequest();
 window.uuid = token();
-// Use a sync XHR to reset the state.
-xhr.open('GET', `../resources/stash-take.py?key=${uuid}`, false);
-xhr.send();
-
 const URL =
   `../resources/empty.txt?pipe=status(302)|` +
   `header(Location,../resources/stash-put.py?key=${uuid}%26value=on)|` +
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19326)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/286428929)

mnot:
What are you trying to test for here? HTTP's connection behaviour depends on network conditions, and can't be relied upon. I.e., you can't guarantee that two requests to the same origin will go over the same connection, even if they're one right after the other.

yutakahirano:
> What are you trying to test for here? HTTP's connection behaviour depends on network conditions, and can't be relied upon. I.e., you can't guarantee that two requests to the same origin will go over the same connection, even if they're one right after the other.

This is not a test for HTTP keepalive, but for [the keepalive feature in fetch API](https://fetch.spec.whatwg.org/#ref-for-dom-request-keepalive). Does this answer your question?

mnot:
yes, thanks :)

yutakahirano:
Can anyone review this change?

yutakahirano:
@jakearchibald can you take a look?

fetch/api/resources/keepalive-iframe.html
```diff
@@ -0,0 +1,18 @@
+<!doctype html>
+<html>
+<meta charset="utf-8">
+<script src="/common/utils.js"></script>
+<script>
+const xhr = new XMLHttpRequest();
+window.uuid = token();
+// Use a sync XHR to reset the state.
+xhr.open('GET', `../resources/stash-take.py?key=${uuid}`, false);
```
ricea: I'm not sure this is useful. If I understand correctly, `uuid` will be different every time this test is run. So there will never be an existing key with value `uuid` and this XHR will do nothing.

yutakahirano: Sounds reasonable, fixed. Thanks.

ricea (approved):

PR #7674, opened 2017-10-11
Service worker fetch abort tests

jakearchibald:
Tests for https://github.com/w3c/ServiceWorker/pull/1178.

### DIFF
--- a/service-workers/service-worker/fetch-event-request-abort-signal.https.html
+++ b/service-workers/service-worker/fetch-event-request-abort-signal.https.html
@@ -17,18 +17,12 @@ async function nextFrame() {
   await raf();
 }
 
-function broadcastResponse(channel) {
-  const bc = new BroadcastChannel(channel);
-  const p = new Promise(resolve => {
+function broadcastResponse(bc) {
+  return new Promise(resolve => {
     bc.onmessage = event => {
-      bc.close();
       resolve(event.data);
     };
   });
-
-  p.broadcastChannel = bc;
-
-  return p;
 }
 
 function reset() {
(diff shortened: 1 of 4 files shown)


jakearchibald:
@mattto I'm not getting a fetch event in the navigation preload test, which I would expect. Have I done some really silly, or have I discovered a weird Chrome bug?

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19335)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/286515375)

service-workers/service-worker/fetch-event-request-abort-signal.https.html
```diff
@@ -0,0 +1,165 @@
...
+  for (const iframe of [...document.querySelectorAll('.test-iframe')]) {
+    iframe.remove();
+  }
+  return navigator.serviceWorker.getRegistrations().then(registrations => {
+    return Promise.all(registrations.map(r => r.unregister()));
+  });
+}
+
+promise_test(async t => {
+  await reset();
```
annevk: Shouldn't this be a test_cleanup step of sorts?

annevk (commented):
Generally this looks okay, though I'm far less of an expert. It seems it might be good to cleanup the repeated boilerplate in the promise_tests. That might make them a little easier to read or at least understand the essential difference between them.

service-workers/service-worker/fetch-event-request-abort-signal.https.html
jakearchibald: Unfortunately it doesn't support async cleanup steps

annevk: Aah okay. A single comment to that effect would be good.

annevk: And maybe an issue to add promise_cleanup?

jakearchibald:
The CI is complaining about `setTimeout` here, but it seems fair to use it in promise tests, especially as it's promisified. Am I misunderstanding?

mkruisselbrink:
> The CI is complaining about setTimeout here, but it seems fair to use it in promise tests, especially as it's promisified. Am I misunderstanding?

It should still be test.step_timeout to deal with timeout multipliers (when tests are run on slow platforms for example).

jakearchibald:
https://github.com/w3c/web-platform-tests/pull/7674/commits/fbfa457b29fe0704ef69c60a85323f0ac8fd514f

I've removed the use of `setTimeout`. I'm now doing this:

1. Start a service worker controlled fetch.
1. `requestAnimationFrame` * 2.
1. Abort the fetch.
1. `requestAnimationFrame` * 2.
1. Tell the service worker "the fetch should be aborted now" via a `BroadcastChannel`.

My assumption is that the abort will always reach the service worker before the channel broadcast. Is this a safe assumption?

jakearchibald:
@mkruisselbrink could I get a review on the fixes?

wanderview:
@jakearchibald We have a test written that might cover some of this as well:

https://bugzilla.mozilla.org/page.cgi?id=splinter.html&bug=1394102&attachment=8903478

Do the tests here cover that case?  If so, I'm inclined to just defer to these tests.

ricea:
> My assumption is that the abort will always reach the service worker before the channel broadcast. Is this a safe assumption?

No. This kind of thing is probably why setTimeout is banned in the first place. This will make the test flaky. If you need ordering guarantees, please enforce them explicitly.

PR #7680, opened 2017-10-11
IndexedDB: Add tests that verify the source of IDBRequests 

inexorabletash:
Missing coverage noted in https://github.com/w3c/IndexedDB/issues/213

Verify the source property of requests returned by operations made against stores, indexes, and cursors.

### DIFF
--- a/IndexedDB/idbcursor-request-source.html
+++ b/IndexedDB/idbcursor-request-source.html
@@ -1,0 +1,31 @@
+<!doctype html>
+<meta charset=utf-8>
+<title>IndexedDB: The source of requests made against cursors</title>
+<meta name="help" href="https://w3c.github.io/IndexedDB/#dom-idbrequest-source">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="support.js"></script>
+<script>
+
+[
+  'cursor.update(0)',
+  'cursor.delete()'
+].forEach(expr => indexeddb_test(
+  (t, db) => {
+    db.createObjectStore('store', {autoIncrement: true});
+  },
+  (t, db) => {
+    const tx = db.transaction('store', 'readwrite');
+    const store = tx.objectStore('store');
+    store.put('value');
+    store.openCursor().onsuccess = t.step_func(e => {
+      const cursor = e.target.result;
+      assert_equals(eval(expr).source, cursor,
+                    `${expr}.source should be the cursor itself`);
+      t.done();
+    });
+  },
+  `The source of the request from ${expr} is the cursor itself`
+));
+
+</script>
--- a/IndexedDB/idbcursor-request-source.html
+++ b/IndexedDB/idbcursor-request-source.html
@@ -8,9 +8,9 @@
 <script>
 
 [
-  'cursor.update(0)',
-  'cursor.delete()'
-].forEach(expr => indexeddb_test(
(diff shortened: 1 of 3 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19349)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/286724372)

inexorabletash:
Weird... Edge will throw due to unsupported functions (getAll, getAllKeys, openKeyCursor on store)... but why is it flaky? And why is it hitting the 'open should succeed' error? Hrm...

inexorabletash:
I've narrowed down the flakiness to what appears to be an Edge bug - cross-talk between operations on unrelated databases:

```js
function t(dbname) {
  indexedDB.deleteDatabase(dbname).onsuccess = e => {
    const open = indexedDB.open(dbname, 1);

    open.onupgradeneeded = e => {
      const db = open.result;
      db.createObjectStore('store');
    };

    open.onerror = e => { 
      console.error('open should succeed');
    };

    open.onsuccess = e => {
      console.log('open succeeded');
      const db = open.result;
      db.close();
      indexedDB.deleteDatabase(dbname);
    };
  };
}

t('db1');
t('db2');
```
The second open flakily succeeds or fails.

siusin:
/cc @nolanlawson 
any suggestion on this Edge bug?

inexorabletash:
I sent the repro to @aliams as well. 

In case it's not clear, the repro has nothing to do with the features under test here. It's an unfortunate coincidence. It likely causes flakiness in other tests, but it's worst here because the test bodies are effectively synchronous here; the test steps don't need to wait for a request's response to come back, so the race is "lost" far more frequently than in the more common tests with a request/response cycle.

inexorabletash:
Hah, per @aliams - @nolanlawson discovered this issue 3 years ago: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/175642/

IndexedDB/idbcursor-request-source.html
```diff
@@ -0,0 +1,31 @@
...
+  (t, db) => {
+    db.createObjectStore('store', {autoIncrement: true});
+  },
+  (t, db) => {
+    const tx = db.transaction('store', 'readwrite');
+    const store = tx.objectStore('store');
+    store.put('value');
+    store.openCursor().onsuccess = t.step_func(e => {
+      const cursor = e.target.result;
+      assert_equals(eval(expr).source, cursor,
```
pwnall: nit: You could avoid eval here by having each string be a closure instead, and using toString() on the closure. To be clear, here's some incomplete code below.

```js
[
  cursor => { cursor.update(0) },
].forEach(fn ...

assert_equals(fn(cursor), cursor, `${fn}`.source ... })
```

Same for the cases below.

pwnall (approved):
LGTM with one nit.

Ms2ger:
@inexorabletash you planning to fix the not?

inexorabletash:
Yeah, it's just been the weekend in this TZ and I'm OOO from work today.

If anyone else wants to push a change to this PR and land I'd be okay with that, though.

IndexedDB/idbcursor-request-source.html
inexorabletash: Done.

PR #7689, opened 2017-10-12
Make 039.html deterministic

hajimehoshi:
We found that 039.html was flaky after changing schedulers in Chromium (crbug.com/771729). Now the test expects that loading all the iframes ends within 200[ms] after window.onload. It looks like the scheduler change affected the result.

This PR fixes the test more deterministic by confirming all the iframe loaded before the assertion.

### DIFF
--- a/html/semantics/scripting-1/the-script-element/execution-timing/039.html
+++ b/html/semantics/scripting-1/the-script-element/execution-timing/039.html
@@ -10,10 +10,14 @@
         <div id="log">FAILED (This TC requires JavaScript enabled)</div>
         <div></div>
         <script>log('inline script #1');
+                var iframeLoaded = 0;
                 for( var i=0; i<2; i++ ){
                         var iframe=document.createElement('iframe');
                         document.getElementsByTagName('div')[1].appendChild(iframe);
                         iframe.src='pages/helloworld.html?'+i+'&'+Math.random();
+                        iframe.onload = function() {
+                                iframeLoaded++;
+                        };
                 }
                 log('end script #1');
         </script>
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19362)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/286888682)

jdm:
I don't understand how this change avoids your flakiness. The window's load event should not be dispatched until the iframes are loaded; if that occurs, that sounds like a bug in the engine that needs to be investigated.

hajimehoshi:
Thank you

> The window's load event should not be dispatched until the iframes are loaded

Is that true? An iframe's onload event are fired when the source is loaded, and I think this has nothing to do with window's onload.

jdm:
Just after [this bit of the specification](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#iframe-load-event-steps), there is this paragraph:
```
When the iframe's browsing context's active document is not ready for post-load tasks, and when anything in the iframe is delaying the load event of the iframe's browsing context's active document, and when the iframe's browsing context is in the delaying load events mode, the iframe must delay the load event of its document.
```
This means that a document that contains an iframe that has not yet received a load event will itself have its load event delayed.

gsnedders (changes requested):
As @jdm says, the flakiness seen in Chrome appears to be a bug in Chrome per spec.

hajimehoshi:
OK, thanks

PR #7754, opened 2017-10-13
Merge progress-events into XMLHttpRequest

foolip:
These tests are for https://www.w3.org/TR/progress-events/, and the
Editor's Draft link there redirects to https://xhr.spec.whatwg.org/.

Status.html isn't a test and is deleted.


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19443)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/287485344)

annevk:
One of the test was a submission from Samsung. Did you verify it was correct?

annevk:
Also, I don't think we necessarily have to move these over to the XMLHttpRequest directory. If we cross-link as we do in other situations it's fine.

foolip:
The Samsung tests passes everywhere: https://wpt.fyi/progress-events/tests/submissions/Samsung

And what it's testing seems in line with https://xhr.spec.whatwg.org/#firing-events-using-the-progressevent-interface, but I didn't go looking deeply for things that may go beyond the spec.

This actually did start out as a cross-linking change, but then I looked at the tests and there were so few, and some ProgressEvent in XMLHttpRequest already, so just merging them seems better.

foolip:
Oh, and I should update the link in those tests, now that I found a better one.

annevk (approved):

foolip:
Oh, I managed to break firing-events-http-content-length.html, good thing I took a look at https://pulls.web-platform-tests.org/job/19448.10. https://github.com/w3c/web-platform-tests/issues/7475 will be a big improvement :)

foolip:
The trouble is that firing-events-http-content-length.html requests resources/img.jpg, which was previously 404, but now exists. And wptserve doesn't send a Content-Length header for img.jpg, but does for a 404. So I'll have to make sure the resource used has a Content-Length header...

foolip:
OK, fixed by using trickle.py to control whether the header is sent, and also checking the exact value. @annevk, LGTY?

XMLHttpRequest/firing-events-http-content-length.html
```diff
@@ -17,20 +17,20 @@
         xhr.onprogress = function(pe) {
           test.step(function() {
             if(pe.type == "progress") {
-              assert_true(pe.loaded >= 0, "loaded is initialize to the number of HTTP entity body bytes transferred.");
-              assert_true(pe.lengthComputable, "lengthComputable is true.");
-              assert_not_equals(pe.total, 0, "total is not zero.");
+              assert_greater_than_equal(pe.loaded, 0, "loaded");
```
foolip: I simplified the descriptions because the failure messages actually make more sense when just the actual value is described, as opposed to what it should be.

annevk (approved):

PR #7758, opened 2017-10-13
Convert 2 Microsoft SVG-in-HTML tests to testharness.js

foolip:
That's 014 and 015. Drive-by whitespace changes in 002 and 012.

### DIFF
--- a/old-tests/submission/Microsoft/foreigncontent/foreign_content_002.html
+++ b/old-tests/submission/Microsoft/foreigncontent/foreign_content_002.html
@@ -1,11 +1,11 @@
-<!DOCTYPE html >
+<!DOCTYPE html>
 <html>
     <head>
         <title>HTML 5 Foreign Content SVG in HTML </title>
         <meta description="Test to verify SVG inside HTML I element parses correctly" />
-        <link rel="author" title="Microsoft" href="http://www.microsoft.com/" />        
+        <link rel="author" title="Microsoft" href="http://www.microsoft.com/" />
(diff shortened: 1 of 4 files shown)


foolip:
The tests fail in Chrome and Firefox. When all results are on wpt.fyi, maybe it'll be easier to decide what to do with these.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19447)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/287496733)

jdm (approved):

AmeliaBR:
As far as standardizing format goes, the pull request should probably be approved.

But the tests themselves need a review.

I'm particularly confused about 015: I don't think there's anything in the HTML parser that is expected to camelCase an HTML element name just because it happens to be a case-insensitive match for an SVG element camelCase name; the [case-correction is specific to being in an SVG namespace parsing context](https://html.spec.whatwg.org/multipage/parsing.html#parsing-main-inforeign).  Not sure where's the best place to discuss that.

014 is correct per SVG 1.1 and the HTML parser, but we probably need something added to SVG 2 that lists the now-obsolete elements and requires them to be parsed as SVG.  That should probably be an issue filed in https://github.com/w3c/svgwg/issues/

foolip:
Thanks @AmeliaBR, I didn't look too hard at the validity, once I know the results I'll get back to your comment to see what changes and issues are needed.

PR #7763, opened 2017-10-13
Update Generic Sensor IDL tests; test event constructor

foolip:
Using standalone IDL files for idlharness.js is the new thing.

The contructor test fails in Chrome because it's not wired up correctly:
https://cs.chromium.org/chromium/src/third_party/WebKit/Source/modules/sensor/SensorErrorEvent.cpp

### DIFF
--- a/generic-sensor/idlharness.https.html
+++ b/generic-sensor/idlharness.https.html
@@ -7,57 +7,27 @@
 <script src="/resources/testharnessreport.js"></script>
 <script src="/resources/WebIDLParser.js"></script>
 <script src="/resources/idlharness.js"></script>
-<div id="log"></div>
-
-<script id="idl" type="text/plain">
-interface Event {
-};
-
-interface Error {
-};
-
-dictionary EventInit {
-};
-
-[SecureContext, Exposed=Window]
-interface Sensor : EventTarget {
-  readonly attribute boolean activated;
-  readonly attribute DOMHighResTimeStamp? timestamp;
-  void start();
-  void stop();
-  attribute EventHandler onreading;
-  attribute EventHandler onactivate;
-  attribute EventHandler onerror;
-};
-
-dictionary SensorOptions {
-  double? frequency;
-};
-</script>
-
-<script id="generic-idl" type="text/plain">
-[Constructor(DOMString type, SensorErrorEventInit errorEventInitDict),
- SecureContext, Exposed=Window]
-interface SensorErrorEvent : Event {
-  readonly attribute Error error;
-};
-
-dictionary SensorErrorEventInit : EventInit {
-  required Error error;
-};
-</script>
-
 <script>
-(() => {
-  "use strict";
-  let idl_array = new IdlArray();
-  idl_array.add_untested_idls(document.getElementById('idl').textContent);
-  idl_array.add_idls(document.getElementById('generic-idl').textContent);
+"use strict";
 
+function doTest([dom, generic_sensor]) {
+  var idl_array = new IdlArray();
+  idl_array.add_untested_idls(dom);
+  idl_array.add_untested_idls('interface DOMException {};');
+  idl_array.add_idls(generic_sensor);
   idl_array.add_objects({
-    SensorErrorEvent: ['new SensorErrorEvent("SECURITY_ERR", { errorCode: 18 });']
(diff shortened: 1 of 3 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19454)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/287553816)

foolip:
@anssiko, can you review? I have some IDL tests for geolocation-sensor which is blocked on this, needs the shared generic-sensor.idl file.

foolip:
Or @rakuco? Don't know how involved with the details of sensors you are :)

generic-sensor/idlharness.https.html
```diff
@@ -7,57 +7,27 @@
...
-  "use strict";
-  let idl_array = new IdlArray();
-  idl_array.add_untested_idls(document.getElementById('idl').textContent);
-  idl_array.add_idls(document.getElementById('generic-idl').textContent);
+"use strict";
 
+function doTest([dom, generic_sensor]) {
+  var idl_array = new IdlArray();
+  idl_array.add_untested_idls(dom);
+  idl_array.add_untested_idls('interface DOMException {};');
```
Honry: `interface Sensor` can be untested here as it has already been tested in concrete sensor suites.

rakuco:
@Honry, @pozdnyakov and @alexshalamov are the best people to look at this one

generic-sensor/idlharness.https.html
foolip: Oh, that is odd, `interface Sensor` appears in:
* accelerometer/idlharness.https.html
* ambient-light/idlharness.https.html
* gyroscope/idlharness.https.html
* magnetometer/idlharness.https.html
* orientation-sensor/idlharness.https.html

And it's in the tested IDL of all of these, which means the IDL is duplicated (and missing hasReading) everywhere and the Sensor interface object is tested many times.

I think the way to go is to create standalone IDL files for all the others as well, as I was doing for geolocation-sensor when I realized I needed to do this.

Then, the problem will be that some test should test an instance of Sensor, but ideally not all of them. To make this test agnostic about what kinds of sensors exist, maybe it could enumerate all things on window that end with "Sensor", and try to create instances of them, using the first one that actually returns an instance `s` where `s instanceof Sensor` is true?

pozdnyakov (approved):
lgtm

generic-sensor/idlharness.https.html
foolip: Hmm, I tried doing just that, with this addition to the test:
```JavaScript
  // Sensor itself cannot be constructor, so look for any subinterface that can
  // be and create that instead.
  for (var name of Object.getOwnPropertyNames(self)) {
    if (!name.endsWith('Sensor'))
      continue;
    if (Object.getPrototypeOf(self[name]) != Sensor)
      continue;
    try {
      var object = new self[name];
      // If that didn't throw, then this is an instance of Sensor.
      assert_true(object instanceof Sensor);
      idl_array.add_objects({
        Sensor: ['new ' + name]
      });
      break;
    } catch(e) {}
  }
```

But that doesn't work, because there will be a "Sensor must be primary interface of new AmbientLightSensor" failing test, idlharness.js wants to test a instance of Sensor alone, and that doesn't exist.

I've filed https://github.com/w3c/web-platform-tests/issues/7781 about the fact that this is hard.

alexshalamov (approved):

foolip:
OK, so I'll deal with https://github.com/w3c/web-platform-tests/issues/7781 as I try to add initial tests for Geolocation Sensor, merging this now.

rakuco:
> The contructor test fails in Chrome because it's not wired up correctly:

I've filed https://bugs.chromium.org/p/chromium/issues/detail?id=775001 to track that one.

generic-sensor/idlharness.https.html
Honry: I will create standalone IDL files for those concrete sensor APIs once we have a conclusion in #7781.

foolip: https://github.com/w3c/web-platform-tests/pull/7861 is what I'm suggesting. I won't do any of the other concrete tests, if that's merged and you like the style, please go ahead and update the others :)

PR #7764, opened 2017-10-13
MIME type parameter parsing tests

annevk:
In the hope that the charset parameter is a good proxy for the whole thing. Would be good to run these through more endpoints though.

Relates to issues 30-41 on https://github.com/whatwg/mimesniff/issues.


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19456)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/287596125)

domenic:
Things I don't see tested:

- Code points > U+00FF
- Type longer than 127
- Subtype longer than 127
- Multiple slashes before ;
- Multiple consecutive ;
- Types and subtypes containing the full range of HTTP token code points, including especially punctuation
- Maybe these should be generated somehow; not sure how to square that with the JSON-ness:
  - Exhaustive tests for types and subtypes that contain code points less than U+00FF that are not HTTP token code points
  - Exhaustive tests for parameters that contain code points less than U+00FF that are not not HTTP quoted-string token code points
    - Both quoted and unquoted would be ideal
- Server-side tests which ensure the server only sees the parsed-then-serialized output. Ideally in a way that doesn't go through other canonicalizations like Blob or Response... maybe `<form>` can help here? XHR? Would be good to test as many code paths as possible.

annevk:
I still don't buy that 127 is an actual thing. No code has it and it seems rather arbitrary.

> Server-side tests which ensure the server only sees the parsed-then-serialized output.

What code path would this test? Note that Response/Request only store the parsed output internally for generating Blob objects...

Do you want to block on having most of those tests or can they be follow-up? I left https://github.com/whatwg/mimesniff/issues/45 open quite intentionally. It requires some additional thought into how to structure the tests as not all places will react to non-ASCII input the same way.

domenic:
> I still don't buy that 127 is an actual thing. No code has it and it seems rather arbitrary.

I think in general when changing a standard it's a good idea to test that the changes you made are supported by browsers, even if the old version doesn't have tests backing it up.

> What code path would this test? Note that Response/Request only store the parsed output internally for generating Blob objects...

The idea is to test the networking code, which at least in Chrome I believe uses a completely different MIME type parser from other parts of the codebase.

> Do you want to block on having most of those tests or can they be follow-up?

No need to block, but the more tests are in place before I start writing a JS implementation, the better.

annevk:
@domenic you're talking about the MIME type the server sees though. There's already quite a few tests that ensure the browser networking stack doesn't touch the value of a custom `Content-Type` header. So I wonder what API entry point you have in mind.

domenic:
Well so for example fetch or, perhaps more directly, XHR, can send Content-Type headers. Probably they should only send ones that have been parsed-then-serialized.

annevk:
No they shouldn't (and I already test this). We don't want to embed header-specific knowledge in the networking library. That would prevent web pages from experimenting with new formats.

annevk:
There's a specific case with XMLHttpRequest where we do modify the value and in that case it's indeed parsed and serialized; that's #8422.

annevk:
Generating tests with a Python script is doable. The way to do those tests would be to write a Python script that generates a JSON resource. Ensure both are committed and ensure `ci_built_diff.sh` runs the Python script so modifications that do not keep the script and generated resource in sync are detected. Documenting this here mostly for future reference, since I won't get to it today.

mimesniff/mime-types/charset-parameter.window.js
```diff
@@ -0,0 +1,27 @@
...
+
+function runTests(tests) {
+  tests.forEach(val => {
+    if(typeof val === "string" || val.navigable !== true || typeof val.encoding !== "string") {
+      return;
+    }
+    const mime = val.input;
+    async_test(t => {
+      const frame = document.createElement("iframe"),
+            expectedEncoding = val.encoding === "" ? "UTF-8" : val.encoding;
```
domenic: I think null, instead of the empty string, should be used to signify no parsed encoding.

mimesniff/mime-types/resources/mime-types.json
```diff
@@ -0,0 +1,213 @@
...
+  {
+    "input": "text/html;0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789=x;charset=gbk",
+    "output": "text/html;0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789=x;charset=gbk",
+    "navigable": true,
+    "encoding": "GBK"
+  },
+  "type/subtype longer than 127",
+  {
+    "input": "0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789/0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789",
+    "output": "0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789/0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789"
```
domenic: This is the only entry that parses correctly but does not have an "encoding" line. The unpredictable data format made it harder to write a test runner.

mimesniff/mime-types/resources/mime-types.json
```diff
@@ -0,0 +1,213 @@
...
+    "navigable": true,
+    "encoding": "GBK"
+  },
+  {
+    "input": "text/html;charset= gbk",
+    "output": "text/html;charset=\" gbk\"",
+    "navigable": true,
+    "encoding": "GBK"
+  },
+  "Single quotes (invalid)",
```
domenic: Single quotes are valid per https://pr-preview.s3.amazonaws.com/whatwg/mimesniff/pull/36.html#http-token-code-point, so these next three tests seem wrong.

domenic:
Coverage report at https://lcov-report-sngrqbipez.now.sh. Missing one branch in the parser, plus any mixed-case tests. Pretty good!

domenic:
Oh, I guess several "else" paths were never taken, mostly where the input ends prematurely. Those are more serious coverage gaps.

mimesniff/mime-types/resources/mime-types.json
annevk: navigable and encoding are both optional and only make sense in specific contexts. I'll document it in more clearly in the README.

annevk:
I added generated tests. I think we have ample coverage now for a v0/v1 of this.

mimesniff/mime-types/resources/generated-mime-types.json
```diff
@@ -0,0 +1,3542 @@
...
+    "input": "x/\t",
+    "output": null
+  },
+  {
+    "input": "x/x;\t=x;bonus=x",
+    "output": "x/x;bonus=x"
+  },
+  {
+    "input": "x/x;x=\t;bonus=x",
+    "output": "x/x;x=\"\t\";bonus=x"
```
domenic: My parser says this should be `"x/x;bonus=x"`. Trailing whitespace gets removed from the parameter value, then it becomes the empty string, so it's omitted.

mimesniff/mime-types/resources/generated-mime-types.json
```diff
@@ -0,0 +1,3542 @@
...
+    "input": "x/ ",
+    "output": null
+  },
+  {
+    "input": "x/x; =x;bonus=x",
+    "output": "x/x;bonus=x"
+  },
+  {
+    "input": "x/x;x= ;bonus=x",
+    "output": "x/x;x=\" \";bonus=x"
```
domenic: My parser says this should be `"x/x;bonus=x"`. Trailing whitespace gets removed from the parameter value, then it becomes the empty string, so it's omitted.

mimesniff/mime-types/resources/mime-types.json
```diff
@@ -0,0 +1,253 @@
...
+    "output": "0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789/0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789"
+  },
+  "Valid",
+  {
+    "input": "!#$%&'*+-.^_`|~0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvxyz/!#$%&'*+-.^_`|~0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvxyz;!#$%&'*+-.^_`|~0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvxyz=!#$%&'*+-.^_`|~0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvxyz",
+    "output": "!#$%&'*+-.^_`|~0123456789abcdefghijklmnopqrstuvxyzabcdefghijklmnopqrstuvxyz/!#$%&'*+-.^_`|~0123456789abcdefghijklmnopqrstuvxyzabcdefghijklmnopqrstuvxyz;!#$%&'*+-.^_`|~0123456789abcdefghijklmnopqrstuvxyzabcdefghijklmnopqrstuvxyz=!#$%&'*+-.^_`|~0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvxyz"
```
domenic: "w" is missing in a few places here; the input has 1 w (should be 2) and the output has 0.

mimesniff/mime-types/resources/mime-types.json
```diff
@@ -0,0 +1,253 @@
...
+    "output": "x/x;x=\" !\\\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~\u0080\u0081\u0082\u0083\u0084\u0085\u0086\u0087\u0088\u0089\u008A\u008B\u008C\u008D\u008E\u008F\u0090\u0091\u0092\u0093\u0094\u0095\u0096\u0097\u0098\u0099\u009A\u009B\u009C\u009D\u009E\u009F\u00A0\u00A1\u00A2\u00A3\u00A4\u00A5\u00A6\u00A7\u00A8\u00A9\u00AA\u00AB\u00AC\u00AD\u00AE\u00AF\u00B0\u00B1\u00B2\u00B3\u00B4\u00B5\u00B6\u00B7\u00B8\u00B9\u00BA\u00BB\u00BC\u00BD\u00BE\u00BF\u00C0\u00C1\u00C2\u00C3\u00C4\u00C5\u00C6\u00C7\u00C8\u00C9\u00CA\u00CB\u00CC\u00CD\u00CE\u00CF\u00D0\u00D1\u00D2\u00D3\u00D4\u00D5\u00D6\u00D7\u00D8\u00D9\u00DA\u00DB\u00DC\u00DD\u00DE\u00DF\u00E0\u00E1\u00E2\u00E3\u00E4\u00E5\u00E6\u00E7\u00E8\u00E9\u00EA\u00EB\u00EC\u00ED\u00EE\u00EF\u00F0\u00F1\u00F2\u00F3\u00F4\u00F5\u00F6\u00F7\u00F8\u00F9\u00FA\u00FB\u00FC\u00FD\u00FE\u00FF\""
```
domenic: The output slashes here (between the []s) is, per my parser, removed.

domenic (changes requested):

domenic:
Coverage is 100%, but found some test bugs. Looking good though.

mimesniff/mime-types/resources/mime-charset.py
```diff
@@ -0,0 +1,3 @@
+def main(request, response):
+    response.headers.set("Content-Type", request.GET.first("type"));
+    response.content = "<meta charset=utf-8>\n<script>document.write(document.characterSet)</script>"
```
jgraham: FWIW you can also write this as

```
def  main(request, response):
    return ([("Content-Type", request.GET.first("type"))],
            "<meta charset=utf-8>\n<script>document.write(document.characterSet)</script>")

File: tools/ci/ci_built_diff.sh [hunk 9]
```diff
@@ -18,6 +18,7 @@ main() {
     )
 
     ./update-built-tests.sh
+    python ./mimesniff/mime-types/resources/generated-mime-types.py
```
jgraham: Oh, sorry this should go in `update_built_tests.sh`. I guess the distinction is to make it easy to bulk update all the tests.

mimesniff/mime-types/resources/mime-charset.py
annevk: Is that better?

tools/ci/ci_built_diff.sh
annevk: Changed!

mimesniff/mime-types/resources/mime-types.json
```diff
@@ -177,12 +177,12 @@
...
-    "output": "!#$%&'*+-.^_`|~0123456789abcdefghijklmnopqrstuvxyzabcdefghijklmnopqrstuvxyz/!#$%&'*+-.^_`|~0123456789abcdefghijklmnopqrstuvxyzabcdefghijklmnopqrstuvxyz;!#$%&'*+-.^_`|~0123456789abcdefghijklmnopqrstuvxyzabcdefghijklmnopqrstuvxyz=!#$%&'*+-.^_`|~0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvxyz"
+    "input": "!#$%&'*+-.^_`|~0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz/!#$%&'*+-.^_`|~0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz;!#$%&'*+-.^_`|~0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz=!#$%&'*+-.^_`|~0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz",
+    "output": "!#$%&'*+-.^_`|~0123456789abcdefghijklmnopqrstuvxyzabcdefghijklmnopqrstuvwxyz/!#$%&'*+-.^_`|~0123456789abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz;!#$%&'*+-.^_`|~0123456789abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz=!#$%&'*+-.^_`|~0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
```
domenic: Still missing a "w". You can tell because the strings start in different columns but end in the same column.

mimesniff/mime-types/resources/mime-types.json
```diff
@@ -193,6 +193,23 @@
...
     "output": "x/x;test=\"\\\\\""
   },
+  "Whitespace (not handled by generated-mime-types.json)",
+  {
+    "input": "x/x;x= ",
+    "output": "x/x"
+  },
+  {
+    "input": "x/x;\" ",
+    "output": "x/x;\" \""
```
domenic: In my parser the output is `x/x`, with no parameters. There is no parameter value here, so this is expected.

mimesniff/mime-types/resources/mime-types.json
```diff
@@ -193,6 +193,23 @@
...
+  },
+  {
+    "input": "x/x;x=\t",
+    "output": "x/x"
+  },
+  {
+    "input": "x/x;x=\"\t",
+    "output": "x/x;x=\"\t\""
```
domenic: In my parser the output is `x/x`, with no parameters. Remember that you remove leading and trailing whitespace from the whole MIME type in step 1 of the entire parser algorithm, so this is equivalent to parsing `x/x;x="`

domenic (changes requested):

mimesniff/mime-types/resources/mime-charset.py
jgraham: Up to you; it was just in case you didn't know.

domenic:
Generated tests still not quite there: `x/x;x=\t;bonus=x` and `x/x;x= ;bonus=x` should get their `x` parameter removed, not quoted.

Maybe we should try to do this while we're both online? Ping me sometime perhaps?

annevk:
Are you sure you're looking at the latest version? I don't see those tests in generated-mime-types.json.

domenic (approved):
My bad, my script was not properly updating generated-mime-types.json. Woohoo!!

PR #7769, opened 2017-10-13
Add a test for a Window object being serialized

gsnedders:
Should also add a test for a frame being serialised, but I'm not sure I'm using the webdriver module API right given nothing says what the webdriver module actually is here (is it actually just the Selenium one?).

### DIFF
--- a/webdriver/tests/contexts/json_serialize_windowproxy.py
+++ b/webdriver/tests/contexts/json_serialize_windowproxy.py
@@ -1,0 +1,13 @@
+# META: timeout=long
+
+import json
+
+
+def test_window(session):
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19468)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/287666275)

andreastt (approved):
This is fine, but will need to be moved to an Execute (Async) Script section later. Because that is completely missing, it’s fine to land this as it is.

PR #7775, opened 2017-10-14
Add test to ensure UA conditionally blocks rendering on link's media attr

domfarolino:
This addresses the need for a test case in https://github.com/whatwg/html/pull/2984.

At the very least it is a start, as I'm not sure if it is written the best way since it has potential to be a tad racy I think, but seems to do the trick. Basically we load two stylesheets, one via a link element whose media attribute matches the environment, and one whose media attribute doesn't*. The style sheet that is supposedly non-matching is loaded with a 2 second delay to avoid racy-ness in the following tests.

Then I run two tests, one immediately that blocks the rendering of the page, and another that runs a few seconds later. The former tests to ensure that we've only blocked rendering on a single style sheet (since at this point only one should appear in `document.styleSheets`), and the latter checks to see that the other one was loaded a few seconds after, as it would then asynchronously appear in `document.styleSheets`

----

Things to consider:

 - \* I don't actually verify in code that the media attribute which supposedly doesn't match the environment, truly doesn't match the environment. If it did, the test would incorrectly fail. Should I verify this in another test perhaps?
 - Boris mentioned a better way would be to test the computed styles as the stylesheets loaded, however I'm not sure if this is the best way since the style sheets that we're testing the loading of are supposed to fail to match the environment, so the styles of the page should never change really, we're just looking at which style sheets are loaded sync vs async.

/cc @annevk

### DIFF
--- a/html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
+++ b/html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
@@ -7,19 +7,34 @@
 <link rel=stylesheet media="screen and (max-width:10px)" href=stylesheet.py?stylesNotMatchingEnvironment&delay=2>
 <h1>Dominic Farolino</h1>
 <script>
+  function styleExists(styleText) {
+    for (let key in document.styleSheets) {
+      if (!document.styleSheets.hasOwnProperty(key) || key == "length") continue;
+
+      let currentStyleText = document.styleSheets[key].cssRules["0"].cssText;
+      if (currentStyleText == styleText) {
+        return true;
+      }
+    }
+
+    return false;
+  }
+
   test(() => {
     const h1 = document.querySelector('h1');
     const computedColor = getComputedStyle(h1).color;
     const expectedColor = "rgb(128, 0, 128)";
 
     assert_equals(computedColor, expectedColor);
-    assert_equals(document.styleSheets.length, 1);
+    assert_true(styleExists("h1 { color: purple; }")); // first style sheet
+    assert_false(styleExists("h1 { color: brown; }")); // second style sheet (should not be loaded yet)
   }, "Only the style sheet loaded via a link element whose media attribute matches the environment should block following script execution");
 
-  const secondStylesheetTest = async_test("Two style sheets should be registered as style sheets for the document after 3 seconds");
+  const secondStylesheetTest = async_test("Both style sheets loaded via the link elements should be registered as style sheets for the document after 2 seconds");
   secondStylesheetTest.step_timeout(() => {
     secondStylesheetTest.step(() => {
-      assert_equals(document.styleSheets.length, 2);
+      assert_true(styleExists("h1 { color: purple; }")); // first style sheet
+      assert_true(styleExists("h1 { color: brown; }")); // second style sheet (loaded now!)
     });
     secondStylesheetTest.done();
   }, 3000);
--- a/html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
+++ b/html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
@@ -8,6 +8,11 @@
 <h1>Dominic Farolino</h1>
 <script>
   test(() => {
+    const h1 = document.querySelector('h1');
+    const computedColor = getComputedStyle(h1).color;
+    const expectedColor = "rgb(128, 0, 128)";
+
+    assert_equals(computedColor, expectedColor);
     assert_equals(document.styleSheets.length, 1);
   }, "Only the style sheet loaded via a link element whose media attribute matches the environment should be loaded synchronously");
 
--- a/html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
+++ b/html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
@@ -14,7 +14,7 @@
 
     assert_equals(computedColor, expectedColor);
     assert_equals(document.styleSheets.length, 1);
-  }, "Only the style sheet loaded via a link element whose media attribute matches the environment should be loaded synchronously");
+  }, "Only the style sheet loaded via a link element whose media attribute matches the environment should block following script execution");
 
   const secondStylesheetTest = async_test("Two style sheets should be registered as style sheets for the document after 3 seconds");
   step_timeout(() => {
--- a/html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
+++ b/html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
@@ -12,7 +12,7 @@
   }, "Only the style sheet loaded via a link element whose media attribute matches the environment should be loaded synchronously");
 
   const secondStylesheetTest = async_test("Two style sheets should be registered as style sheets for the document after 3 seconds");
-  setTimeout(() => {
+  step_timeout(() => {
     secondStylesheetTest.step(() => {
       assert_equals(document.styleSheets.length, 2);
     });
--- a/html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
+++ b/html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
@@ -8,10 +8,8 @@
 <h1>Dominic Farolino</h1>
 <script>
   function styleExists(styleText) {
-    for (let key in document.styleSheets) {
-      if (!document.styleSheets.hasOwnProperty(key) || key == "length") continue;
-
-      let currentStyleText = document.styleSheets[key].cssRules["0"].cssText;
+    for (let styleRule of document.styleSheets) {
+      let currentStyleText = styleRule.cssRules["0"].cssText;
       if (currentStyleText == styleText) {
         return true;
       }
@@ -32,10 +30,8 @@
 
   const secondStylesheetTest = async_test("Both style sheets loaded via the link elements should be registered as style sheets for the document after 2 seconds");
   secondStylesheetTest.step_timeout(() => {
-    secondStylesheetTest.step(() => {
       assert_true(styleExists("h1 { color: purple; }")); // first style sheet
       assert_true(styleExists("h1 { color: brown; }")); // second style sheet (loaded now!)
-    });
-    secondStylesheetTest.done();
+      secondStylesheetTest.done();
   }, 3000);
 </script>
--- a/html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
+++ b/html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
@@ -17,7 +17,7 @@
   }, "Only the style sheet loaded via a link element whose media attribute matches the environment should block following script execution");
 
   const secondStylesheetTest = async_test("Two style sheets should be registered as style sheets for the document after 3 seconds");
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19483)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/288067116)

html/semantics/document-metadata/interactions-of-styling-and-scripting/stylesheet.py
```diff
@@ -0,0 +1,10 @@
+from time import sleep
+def main(request, response):
+  if "delay" in request.GET:
+    delay = int(request.GET["delay"])
+    sleep(delay)
+
+  if "stylesNotMatchingEnvironment" in request.GET:
+    return 'h1 {color: green}'
+  else:
+    return 'h1 {color: red}'
```
annevk: Usually green is pass and red is fail.

html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
```diff
@@ -0,0 +1,21 @@
...
+
+<link rel=stylesheet href=stylesheet.py>
+<link rel=stylesheet media="screen and (max-width:10px)" href=stylesheet.py?stylesNotMatchingEnvironment&delay=2>
+<h1>Dominic Farolino</h1>
+<script>
+  test(() => {
+    assert_equals(document.styleSheets.length, 1);
+  }, "Only the style sheet loaded via a link element whose media attribute matches the environment should be loaded synchronously");
+
+  const secondStylesheetTest = async_test("Two style sheets should be registered as style sheets for the document after 3 seconds");
```
annevk: What requires it to be loaded eventualy?

html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
```diff
@@ -0,0 +1,21 @@
...
+<meta charset="utf-8">
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+
+<link rel=stylesheet href=stylesheet.py>
+<link rel=stylesheet media="screen and (max-width:10px)" href=stylesheet.py?stylesNotMatchingEnvironment&delay=2>
+<h1>Dominic Farolino</h1>
+<script>
+  test(() => {
+    assert_equals(document.styleSheets.length, 1);
```
annevk: This would be stronger if you either inspected the contents of the style sheet or used `getComputedStyle` to make sure it is applied.

annevk (commented):
@bzbarsky might have some thoughts here as well about what would be good to test.

html/semantics/document-metadata/interactions-of-styling-and-scripting/stylesheet.py
domfarolino: Ok I think I'll just make these random colors or something because I'm not trying to indicate pass/fail here. Obviously if the "stylesNotMatchingEnvironment" styles were applied over the other styles (when the environment doesn't match), the implementation fails a lot more fundamental things than this test is testing.

html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
```diff
@@ -0,0 +1,26 @@
...
+<h1>Dominic Farolino</h1>
+<script>
+  test(() => {
+    const h1 = document.querySelector('h1');
+    const computedColor = getComputedStyle(h1).color;
+    const expectedColor = "rgb(128, 0, 128)";
+
+    assert_equals(computedColor, expectedColor);
+    assert_equals(document.styleSheets.length, 1);
+  }, "Only the style sheet loaded via a link element whose media attribute matches the environment should be loaded synchronously");
```
bzbarsky: s/be loaded synchronously/block following script execution/?

html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+
+<link rel=stylesheet href=stylesheet.py>
+<link rel=stylesheet media="screen and (max-width:10px)" href=stylesheet.py?stylesNotMatchingEnvironment&delay=2>
```
bzbarsky: This test will randomly fail when run in browser automation frameworks (where we've observed seconds-long pauses in the middle of HTML parsing in the past), but there really isn't much we can do about that here.  :(

html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
```diff
@@ -0,0 +1,26 @@
...
+  test(() => {
+    const h1 = document.querySelector('h1');
+    const computedColor = getComputedStyle(h1).color;
+    const expectedColor = "rgb(128, 0, 128)";
+
+    assert_equals(computedColor, expectedColor);
+    assert_equals(document.styleSheets.length, 1);
+  }, "Only the style sheet loaded via a link element whose media attribute matches the environment should be loaded synchronously");
+
+  const secondStylesheetTest = async_test("Two style sheets should be registered as style sheets for the document after 3 seconds");
```
bzbarsky: This test will also randomly fail: it assumes that the clock of the browser and the clock of stylesheet.py at least run at the same rate (which they may not) and that stylesheet.py will get woken up expeditiously enough to service the request before the browser's timer fires (totally no guarantee of that from OS schedulers).  Again, I'm not sure what can be done about that.

But the upshot is that there is a good chance this test will end up disabled in browser automation frameworks.  :(

domfarolino:
@bzbarsky I've addressed the one comment you had on the text, but regarding the other two, is there really nothing we can do here? (aka: should we just ignore the fact that the test might randomly fail)

bzbarsky:
That's the problem with timing tests, yes....

domfarolino:
@bzbarsky Ok that makes sense. Well I guess that makes this PR about wrapped up if there's nothing else? @annevk thoughts?

html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
```diff
@@ -0,0 +1,26 @@
...
+    const h1 = document.querySelector('h1');
+    const computedColor = getComputedStyle(h1).color;
+    const expectedColor = "rgb(128, 0, 128)";
+
+    assert_equals(computedColor, expectedColor);
+    assert_equals(document.styleSheets.length, 1);
+  }, "Only the style sheet loaded via a link element whose media attribute matches the environment should block following script execution");
+
+  const secondStylesheetTest = async_test("Two style sheets should be registered as style sheets for the document after 3 seconds");
+  step_timeout(() => {
```
annevk: Shouldn't this be secondStylesheetTest.step_timeout?

domfarolino: Hm that would make sense, however it also seems to work as it is, is that just a lucky accident?

annevk: Yeah, in that case they won't be associated with this test, but with an unnamed global test. It's a little confusing, but helps reducing boilerplate for files containing a single test I believe.

domfarolino:
I'm having an issue where the first test is consistently failing after the first time I load the test. The first stylesheet seems to be racing with what I assume is a default stylesheet loaded by the harness or something:

screen shot 2017-12-19 at 3 51 27 am

Again, not sure there's anything we can do on this front, but I'd like to not have this consistently fail and what not.

domfarolino:
I guess to remedy this I could not just check the length of `document.styleSheets`, but instead go into the entries further and check the number of stylesheets in `document.styleSheets` that have CSS rules for "h1" (those that are relevant to this test). That might be a little better but feels hacky. I may do this tomorrow.

annevk:
Yeah, we shouldn't merge it in that case. Not sure who can help debug flaky tests though.

domfarolino:
@annevk The flakiness of this primarily depends on whether I check more than _just_ the number of stylesheets that are in `document.styleSheets` (by checking whether they are actually the styles I'm looking for (not harness or other styles), as previously mentioned) or not.

This would very much solidify this test and make it accurate, but if you think it might be a little too hacky then we can hold off and possibly do something else?

annevk:
I'm not sure I understand. Using `getComputedStyle()` makes it flaky or not?

domfarolino:
Ah I forgot I also did `getComputedStyle()` in the first test. So the answer is no. Basically the goal is to test that right when we start evaluating the script, there should be only 1/2 of _my_ defined stylesheets loaded (because the other one, whose media attr does not match environ) will load later (assuming it doesn't take "later" seconds for us to get to the script evaluation which hopefully it doesn't). Then, a few seconds later, we test that the second of my 2 stylesheets have been loaded (the loading of this second one should not block scripts).

My original plan was to go through and first test that there would be a single entry in `document.styleSheets` when script begins evaluation. Then a bit later one, after the second one has loaded, I check to see if there are two entries in `document.styleSheets`. The issue is that the harness seems to bring with it multiple styles that make their way into `document.styleSheets`. This makes checking the number of entries bad, because I instead want the number of entries that correspond to _my_ styles (not the harness's). So the first test is failing because the harness injects another style, so the `getComputedStyle` passes, but the other one fails.

----

I've made changes to the test to make it way less flaky. It passes consistently on all browser which implement the functionality proposed in the PR that required this test.

html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
```diff
@@ -0,0 +1,41 @@
...
+    const expectedColor = "rgb(128, 0, 128)";
+
+    assert_equals(computedColor, expectedColor);
+    assert_true(styleExists("h1 { color: purple; }")); // first style sheet
+    assert_false(styleExists("h1 { color: brown; }")); // second style sheet (should not be loaded yet)
+  }, "Only the style sheet loaded via a link element whose media attribute matches the environment should block following script execution");
+
+  const secondStylesheetTest = async_test("Both style sheets loaded via the link elements should be registered as style sheets for the document after 2 seconds");
+  secondStylesheetTest.step_timeout(() => {
+    secondStylesheetTest.step(() => {
```
annevk: Why this nested .step?

html/semantics/document-metadata/interactions-of-styling-and-scripting/conditionally-block-rendering-on-link-media-attr.html
```diff
@@ -0,0 +1,41 @@
...
+    return false;
+  }
+
+  test(() => {
+    const h1 = document.querySelector('h1');
+    const computedColor = getComputedStyle(h1).color;
+    const expectedColor = "rgb(128, 0, 128)";
+
+    assert_equals(computedColor, expectedColor);
+    assert_true(styleExists("h1 { color: purple; }")); // first style sheet
```
annevk: Somewhat disturbing the CSS parser preserves color names, but I guess that's the way it is.

domfarolino: Must admit I'm not entirely sure what you mean by this. Would you prefer an rgb or hex val instead?

annevk: All I meant is that I was surprised that the parser doesn't normalize these values meaning the browser has to use some additional bits to remember to serialize back out as a name.

annevk (approved):

PR #7778, opened 2017-10-15
Improve Microsoft's history_000.htm tests

foolip:
The onpopstate test was wrong, because the global object is special and
should have own properties where normally things would be on the
prototype.

Replace the use of 404.html, since that's a bad name for something that
exists.

Note that there's some flakiness in all tested browsers, so the tests
aren't ready to be moved into html/.

### DIFF
--- a/old-tests/submission/Microsoft/history/404.html
+++ b/old-tests/submission/Microsoft/history/404.html
@@ -1,1 +1,0 @@
-Page Not Found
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19486)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/288152407)

jgraham (approved):

PR #7787, opened 2017-10-16
Don't require test type in wpt expectation manifest,

jgraham:
The test type is already in the MANIFEST.json file and requiring it in
the expectation data doesn't make much sense. It isn't used execpt
during updates and so people often forget to add it. Therefore it
makes a lot of sense to just use the data from the main manifest.

MozReview-Commit-ID: HyOoN6T28qc

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1401613 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7787)

### DIFF
--- a/tools/wptrunner/README.rst
+++ b/tools/wptrunner/README.rst
@@ -146,8 +146,6 @@ An example of an expectation file is::
   example_default_key: example_value
 
   [filename.html]
-    type: testharness
-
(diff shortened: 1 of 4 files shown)


jgraham:
Code reviewed upstream.

PR #7791, opened 2017-10-16
add wpt test for fallback discrete type of transform animation. r=boris,hiro

jgraham:
According to the spec, if one of the matrices for transform interpolation is
non-invertible, the used animation function must fall-back to a discrete
animation. Add wpt tests so we can ensure the web compability for this behavior.

Note that we don't add 'discrete' type for transform property in property-list.js.
Because the 'discrete' type also tests discrete addition and accumulation,
however, the fallback behavior is just for interpolation and accumulation, not
for addition. So, we add tests in each part of transformListType in property-types.js.

One test is failing in Gecko, so annotate it for Gecko.

MozReview-Commit-ID: 3JGvgqbBqZp

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1394284 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7791)


jgraham:
Code reviewed upstream.

PR #7794, opened 2017-10-16
Increase the marionette connect timeout in wptrunner,

jgraham:
This switches to using a timeout of 120s * timeout_multplier, which is
twice the current value. The Marionette startup_timout value is
removed because it turns out that's only used when marionette starts
the Firefox instance and therefore is ignored in this case.

MozReview-Commit-ID: 4VA8yB6M1d5

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1393366 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7794)


jgraham:
Code reviewed upstream.

PR #7795, opened 2017-10-16
Make sure offset* is computed correctly on absolutely positioned kids of relatively positioned elements with scrolllframes and borders.

jgraham:
The check for isAbsolutelyPositioned was an old incorrect attempt to fix bug
81290.  We have since added the proper fix (not adding offsets of the offset
parent frame) in bug 375003.

MozReview-Commit-ID: 7NgnfrYcs8h

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=850684 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7795)

### DIFF
--- a/cssom-view/offsetTopLeftInScrollableParent.html
+++ b/cssom-view/offsetTopLeftInScrollableParent.html
@@ -1,0 +1,111 @@
+<!doctype html>
+<meta charset=utf-8>
+<title></title>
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #7799, opened 2017-10-16
Consistently use KeyframeEffect constructor for testing in processing-a-keyframes-argument.html;

jgraham:
KeyframeEffectReadOnly may disappear (see [1]) and is only needed for CSS
Animations and CSS Transitions so in that sense KeyframeEffect is more basic
(despite being a subclass of KeyframeEffectReadOnly) so we should prefer it to
KeyframeEffectReadOnly.

Furthermore, as the comment at the start of the file suggests, we should
consistently use the same method for testing these procedures. We currently use
the KeyframeEffect constructor because it is more direct and basic.

[1] https://github.com/w3c/web-animations/issues/185

MozReview-Commit-ID: LBrlfzyn2Ch

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7799)

### DIFF
--- a/web-animations/interfaces/KeyframeEffect/processing-a-keyframes-argument.html
+++ b/web-animations/interfaces/KeyframeEffect/processing-a-keyframes-argument.html
@@ -12,8 +12,8 @@
 <script>
 'use strict';
 
-// This file only tests the KeyframeEffectReadOnly constructor since it is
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #7800, opened 2017-10-16
Use ES6 let/const in processing-a-keyframes-argument.html;

jgraham:
Gradually we plan to move all these tests to ES6 (or at least the subset
supported by all UAs that are likely to implement this spec) so while we are
touching this file we update a few uses of 'var' to let/const.

MozReview-Commit-ID: 45OJyXmUzKu

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7800)


jgraham:
Code reviewed upstream.

PR #7801, opened 2017-10-16
Use for...of instead of forEach for several tests in processing-a-keyframes-argument.html;

jgraham:
for...of is generally preferred over forEach since it is a little easier to read
and allows using 'break' and 'continue'. Furthermore it is supported in all
major browsers. (It also makes wrapping one of the long lines in this file
easier.)

MozReview-Commit-ID: 1BuoW0QSxaG

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7801)


jgraham:
Code reviewed upstream.

PR #7803, opened 2017-10-16
Drop invalid test for offsets;

jgraham:
There is a test that assumes that an offset specified on a property-indexed
keyframe is applied to all generated keyframes but that behavior is not (yet)
specified.

This behavior will be specified in [1] but until that happens it seems invalid
to test for it. Furthermore, when that is specified we will need much more
thorough tests than this one.

[1] https://github.com/w3c/web-animations/issues/148

MozReview-Commit-ID: HUUw88dg2P7

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7803)


jgraham:
Code reviewed upstream.

PR #7804, opened 2017-10-16
Update some test descriptions;

jgraham:
* We should refer to reading or accessing properties, as opposed to
  "considering" them.
* We should use "property-indexed" consistently.

MozReview-Commit-ID: ItCE4g8LmOC


jgraham:
Code reviewed upstream.

PR #7808, opened 2017-10-16
Consistently use spaces in object notation;

jgraham:
This seems to be standard JS style recently (as used in prettier etc.): Use
spaces to separate the { and } from the properties (but not for arrays).

MozReview-Commit-ID: FRkFRwwcJJh

### DIFF
--- a/web-animations/interfaces/KeyframeEffect/processing-a-keyframes-argument.html
+++ b/web-animations/interfaces/KeyframeEffect/processing-a-keyframes-argument.html
@@ -45,7 +45,7 @@ function TestKeyframe(testProp) {
 
   Object.defineProperty(this, testProp, {
     get: function() { _propAccessCount++; },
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

web-animations/interfaces/KeyframeEffect/processing-a-keyframes-argument.html
```diff
@@ -45,7 +45,7 @@
 
   Object.defineProperty(this, testProp, {
     get: function() { _propAccessCount++; },
-    enumerable: true
+    enumerable: true,
```
snuggs: @jgraham why no trailing commas in all object literals? Or the inclusion of this one?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas

PR #7818, opened 2017-10-16
Consistently use arrow syntax for functions in processing-a-keyframes-argument-*.html tests;

jgraham:
We don't however, use arrow syntax for local functions that act as class
constructors since they don't want the lexical this that arrow functions use.

MozReview-Commit-ID: FuVhHIBFZrE

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7818)


jgraham:
Code reviewed upstream.

PR #7825, opened 2017-10-16
Add test for stale frame element.

jgraham:
The <p> element inside the <iframe> should be considered stale when
interacting with it after the current browsing context is switched
back to the top level browsing context.

MozReview-Commit-ID: 1zrnBowSpxt

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1405018 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7825)


jgraham:
Code reviewed upstream.

PR #7833, opened 2017-10-16
Tidy up simple-iteration-progress.html and current-iteration.html somewhat;

jgraham:
This is in preparation for further changes to these files later in this patch
series.

In particular this patch:

* Moves some code to more modern Javascript that should be easier to read and
  maintain

* Makes the tests more strict about when active/after values are specified
  - If the timing parameters mean there is no active phase, the test should not
    specify a value to test for that phase. If there *is* an active phase, the
    test must provide a value to test (if it does not the test will fails when
    it compares against the undefined value).
  - Likewise for the after phase
  This should make it a little easier to incorporate testing the playbackRate.

MozReview-Commit-ID: 17vihK5RSbu

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1406381 [ci skip]

### DIFF
--- a/web-animations/timing-model/animation-effects/current-iteration.html
+++ b/web-animations/timing-model/animation-effects/current-iteration.html
@@ -11,26 +11,40 @@
 'use strict';
 
 function runTests(tests, description) {
-  tests.forEach(function(currentTest) {
-    var testParams = '';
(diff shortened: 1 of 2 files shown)


jgraham:
Code reviewed upstream.

PR #7838, opened 2017-10-16
Fix cookie creation via Marionette using IP address

jgraham:
The issue here was that the cookie domain was always prepended with
'.' character. To resolve this edge-case Marionette now first checks
whether the cookie domain is in fact an IP address.

MozReview-Commit-ID: 4xBd4rscXxx

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1407675 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7838)


jgraham:
Code reviewed upstream.

PR #4953, opened 2017-02-22
XMLHttpRequest/File API: fetch revoked blob URL

annevk:
Closes https://github.com/whatwg/xhr/issues/79.

PR #7848, opened 2017-10-16
Add tests using Request/fetch similar to the XHR ones added in #4953

mkruisselbrink:
As far as I can tell the Request spec has the same behavior where on creating the Request we're supposed to parse the URL and get a reference to the underlying blob, and as such this test is supposed to pass.

### DIFF
--- a/FileAPI/blob/Blob-Request-revoke-fetch.html
+++ b/FileAPI/blob/Blob-Request-revoke-fetch.html
@@ -19,6 +19,18 @@ promise_test(function(t) {
     });
 }, "Revoke blob URL after creating Request, will fetch");
 
+promise_test(function(t) {
+    const blob = new Blob(["test"]);
+    const url = URL.createObjectURL(blob);
+
+    return fetch(url).then(response => response.text()).then(text => {
+      assert_equals(text, 'test');
+    });
+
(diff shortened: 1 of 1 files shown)


mkruisselbrink:
@annevk does my understanding here match yours?

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19573)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/288701098)

annevk (approved):
Maybe also test passing the URL directly to `fetch()`? Did you file implementation bugs?

gsnedders:
This is approved, is there any reason why it isn't merged? (Had CI not finished when it was approved?)

annevk:
The reason I didn't merge it is because I asked questions. It's easier to get questions answered if something isn't marked as resolved.

mkruisselbrink:
Sorry for the delay, I somehow missed the earlier approval/comments. Added a test passing url directly to fetch(). Haven't filed any implementation bugs, no, as I wasn't sure this was actually the behavior we want to specs to describe. But I think I now agree that it probably is the right behavior.

PR #7855, opened 2017-10-17
Add tests for the behavior of `<iterator>.return === document.all` in the iterator protocol

jswalden:
test262 peoples want this sort of test in w-p-t, notwithstanding that basically all the testing is of ECMA-262 algorithm -- see https://github.com/tc39/test262/pull/1287 for the first attempt on getting these test in an upstream.

Dunno who you want reviewing this patch, because realistically you want someone JS-ful to review it, given that.  :-\  The test262 PR has links to where in ECMA-262 is being tested, but presumably you want someone with prior familiarity with this spec reviewing it.  Maybe @anba knows who can?

### DIFF
--- a/js/behaviours/iterator-close-return-emulates-undefined-throws-when-called.html
+++ b/js/behaviours/iterator-close-return-emulates-undefined-throws-when-called.html
@@ -1,0 +1,27 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="UTF-8">
+  <title>Test for document.all as the value of &lt;iterator&gt;.return</title>
+  <script src="/resources/testharness.js"></script>
+  <script src="/resources/testharnessreport.js"></script>
+</head>
+<body>
+<div id="log"></div>
+<script>
+test(function() {
+  var iter = {
+    [Symbol.iterator]() { return this; },
+    next() { return {}; },
+    return: document.all,
+  };
+
+  assert_throws(new TypeError, function() {
+    for (var x of iter)
+      break;
+  });
+}, "When <iterator>.return is document.all, it should be invoked by the " +
+   "iterator protocol (and then throw a TypeError)");
+</script>
+</body>
+</html>
--- a/js/behaviours/star-iterable-return-emulates-undefined-throws-when-called.html
+++ b/js/behaviours/star-iterable-return-emulates-undefined-throws-when-called.html
@@ -1,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="UTF-8">
+  <title>Test for document.all as the value of &lt;iterator&gt;.return</title>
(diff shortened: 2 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19584)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/288851992)

anba:
> Maybe @anba knows who can?

Hmm, I basically don't know anything about WPT (every time I break a WPT test case in Firefox, I need to relearn how WPT works. :stuck_out_tongue_closed_eyes: )

foolip:
@jugglinmike, do you think we should be adding tests here given https://github.com/w3c/web-platform-tests/issues/6462? @rwaldron @leobalter?

foolip:
If this directory isn't slated for removal, then we need OWNERS for it. Any volunteers?

foolip:
Ah, so the test is about `document.all`, maybe html/infrastructure/common-dom-interfaces/collections/htmlallcollection.html or a sibling file would make sense?

jugglinmike:
I agree with @jswalden that "basically all the testing is of ECMA-262
algorithm." It may be pragmatic to include a test in WPT for today, but this is
not a tenable solution. Every ECMA262 test that lands in WPT is a test that
non-browser EMCAScript interprets miss out on. Jeff has offered equivalent
tests to both projects, but we can't expect most contributors to be so
thorough.

At some point we should be encouraging folks to invest their energy in
improving WPT/Test262 integration. I've already spoken with @jgraham about
extending the `wpt` tool with a "test262" test type, and [over in Test262,
we're actively exploring practices to improve the experience of consuming the
tests](https://github.com/tc39/test262/pull/1255).

foolip:
@jugglinmike, what could practically be done on WPT/Test262 integration? Having the tests split across repos will mean occasional cases like this one, and a single repo isn't really on the table.

jugglinmike:
@foolip By "integration," I was thinking about extending the `wpt` utility to recognize and execute Test262 tests. In some ways, this would be similar to today's ".any.js" tests insofar as the HTML files required for browser execution would be generated dynamically. But it would also involve injecting a definition of the "host API" to provide "web" implementations of functionality that can't be expressed in JS alone (e.g. creating a new realm).

This could even involve running those kinds of tests in multiple "scenarios". A single test concerning cross-realm semantics could be interpreted by `wpt` in multiple ways--maybe by using an iframe in one case and `window.open` in another. But that's likely more than we should aim to do in an initial
implementation.

jswalden:
These tests ended up landing in test262, and given that I don't see any particular need to land these tests here.  But I'll leave it to people here to make the final call.

foolip:
If the same tests have landed in test262, then let's close this.

PR #7860, opened 2017-10-17
Split MediaStreamTrack IDL tests into a separate file

foolip:
This is a first step towards having a standalone
mediacapture-stream.idl. It also makes the existing test fail fast in
case the getUserMedia promise is rejected.

### DIFF
--- a/mediacapture-streams/MediaStreamTrack-idl.https.html
+++ b/mediacapture-streams/MediaStreamTrack-idl.https.html
@@ -1,0 +1,52 @@
+<!doctype html>
+<title>MediaStreamTrack IDL tests</title>
+<link rel="help" href="https://w3c.github.io/mediacapture-main/#media-stream-track-interface-definition">
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<script src=/resources/WebIDLParser.js></script>
+<script src=/resources/idlharness.js></script>
+<script>
+function idl_test([dom_idl, stream]) {
+  const idl_array = new IdlArray();
+
+  idl_array.add_untested_idls(dom_idl);
+  idl_array.add_untested_idls('interface EventHandler {};');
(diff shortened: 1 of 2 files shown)


foolip:
@orphis, this is to improve the situation of test mismatch in wpt.fyi and manual runs less bad.

If you'll be doing any work in wpt, then you should also feel free to add yourself to mediacapture-streams/OWNERS, and I'll give you write access so you can merge things.

foolip:
For general improvements in this test suite, there are other tests like https://wpt.fyi/mediacapture-streams/MediaStream-video-only.https.html that are also timeout for the same reason, and could probably be fixed in the same way.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19600)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/288992413)

Orphis:
@foolip Not sure if I'll be doing any work in wpt, I merely used it for now!

foolip:
@alvestrand, can you review?

mediacapture-streams/MediaStreamTrack-idl.https.html
```diff
@@ -0,0 +1,52 @@
+<!doctype html>
+<title>MediaStreamTrack IDL tests</title>
+<link rel="help" href="https://w3c.github.io/mediacapture-main/#media-stream-track-interface-definition">
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<script src=/resources/WebIDLParser.js></script>
+<script src=/resources/idlharness.js></script>
+<script>
```
alvestrand: When I did an IDL test, I found that if I put the IDL inside <textarea>, I could cut-and-paste it without having to do end-of-line escaping.
But if you're going for separate IDL files eventually, it's not worth it.

alvestrand (approved):
LGTM!

PR #7881, opened 2017-10-18
Fix wdspec new_session create tests

carlosgcampos:
Both create_alwaysMatch.py and create_firstMatch.py were failing to
import support.create where the input data is shared.

E ValueError: Attempted relative import in non-package

Add __init__.py to support directory and remove the leading '.' from the
import command in both test files.


jgraham (approved):

jgraham:
Travis doesn't like this for some reason; merging anyway.

PR #7891, opened 2017-10-19
Add Web IDL tests for Wake Lock API

Honry:

### DIFF
--- a/wake-lock/idlharness.https.html
+++ b/wake-lock/idlharness.https.html
@@ -1,0 +1,32 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Wake Lock API IDL tests</title>
+<link rel="author" title="Intel" href="http://www.intel.com">
+<link rel="help" href="https://w3c.github.io/wake-lock/">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/resources/WebIDLParser.js"></script>
+<script src="/resources/idlharness.js"></script>
+<script>
+"use strict";
+
+function doTest([dom_idl, wakelock_idl, wakelock]) {
+  const idl_array = new IdlArray();
+  idl_array.add_untested_idls(dom_idl);
+  idl_array.add_untested_idls('interface Navigator {};');
+  idl_array.add_idls(wakelock_idl);
+  self.wakelock = wakelock;
+  self.request = wakelock.createRequest();
+  idl_array.add_objects({WakeLock: ['wakelock'], WakeLockRequest: ['request']});
+  idl_array.test();
+}
+
+promise_test(() => {
+  return Promise.all([
+    fetch("/interfaces/dom.idl").then((response) => response.text()),
+    fetch("/interfaces/wake-lock.idl").then((response) => response.text()),
+    navigator.getWakeLock("screen")
+  ]).then(doTest);
+}, "Test driver");
+</script>
+
--- a/interfaces/wake-lock.idl
+++ b/interfaces/wake-lock.idl
@@ -1,0 +1,23 @@
+enum WakeLockType {
+    "screen",
+    "system"
+};
+
+partial interface Navigator {
+    [SecureContext] Promise<WakeLock> getWakeLock(WakeLockType type);
+};
+
+[SecureContext,
+ Exposed=Window]
+interface WakeLock : EventTarget {
+    readonly attribute WakeLockType type;
+    readonly attribute boolean      active;
+             attribute EventHandler onactivechange;
+    WakeLockRequest createRequest();
+};
+
+[SecureContext,
+ Exposed=Window]
+interface WakeLockRequest {
+    void cancel();
+};
--- a/wake-lock/OWNERS
+++ b/wake-lock/OWNERS
@@ -1,0 +1,3 @@
+@andrey-logvinov
+@rakuco
+@Honry
--- a/wake-lock/OWNERS
+++ b/wake-lock/OWNERS
@@ -1,3 +1,3 @@
 @andrey-logvinov
-@rakuco
+@marcoscaceres
 @Honry
--- a/wake-lock/interfaces.https.html
+++ b/wake-lock/interfaces.https.html
@@ -13,11 +13,16 @@
 function doTest([dom_idl, wakelock_idl, wakelock]) {
   const idl_array = new IdlArray();
   idl_array.add_untested_idls(dom_idl);
+  idl_array.add_untested_idls('interface EventHandler {};');
   idl_array.add_untested_idls('interface Navigator {};');
   idl_array.add_idls(wakelock_idl);
-  self.wakelock = wakelock;
-  self.request = wakelock.createRequest();
-  idl_array.add_objects({WakeLock: ['wakelock'], WakeLockRequest: ['request']});
+  window.wakelock = wakelock;
+  window.request = wakelock.createRequest();
+  idl_array.add_objects({
+    Navigator: ['navigator'],
+    WakeLock: ['wakelock'],
+    WakeLockRequest: ['request']
+  });
   idl_array.test();
 }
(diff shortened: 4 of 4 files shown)


Honry:
@anssiko, who do you think would be appropriate reviewers for Wake Lock?

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19665)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/289783918)

anssiko:
I would suggest the Wake Lock API spec editor @andrey-logvinov and @rakuco to review.

Honry:
@anssiko, thanks for the suggestion. @andrey-logvinov, @rakuco, I've created an OWNERS file with you both included, please just let me know if you would not like to be an owner.

rakuco:
I'm glad to see these new tests, but I wouldn't add myself to the OWNERS file: I'm cheering for the spec and implementation to move forward, but haven't been involved with it myself. I guess @marcoscaceres is a better OWNER/reviewer here, as he's also a spec editor along with @andrey-logvinov and @boiler23.

Honry:
@rakuco, sure, I've replaced you with @marcoscaceres.

wake-lock/idlharness.https.html
```diff
@@ -0,0 +1,32 @@
...
+  idl_array.add_untested_idls(dom_idl);
+  idl_array.add_untested_idls('interface Navigator {};');
+  idl_array.add_idls(wakelock_idl);
+  self.wakelock = wakelock;
+  self.request = wakelock.createRequest();
+  idl_array.add_objects({WakeLock: ['wakelock'], WakeLockRequest: ['request']});
+  idl_array.test();
+}
+
+promise_test(() => {
```
marcoscaceres: This isn't actually testing anything, so we should rewrite this as:

```js
promise_test(async () => {
  const idl_array = new IdlArray();
  const domIdl = await fetch("/interfaces/dom.idl").then(r => r.text());
  const wakelockIdl = await fetch("/interfaces/wake-lock.idl").then(r =>
    r.text()
  );
  const wakelock = await navigator.getWakeLock("screen");

  idl_array.add_untested_idls(domIdl);
  idl_array.add_untested_idls("interface Navigator {};");
  idl_array.add_idls(wakelock_idl);

  window.wakelock = wakelock;
  window.request = wakelock.createRequest();

  idl_array.add_objects({
    WakeLock: ["wakelock"],
    WakeLockRequest: ["request"],
  });
  idl_array.test();
}, "Test IDL implementation of WakeLock API");
```

wake-lock/idlharness.https.html
```diff
@@ -0,0 +1,32 @@
...
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/resources/WebIDLParser.js"></script>
+<script src="/resources/idlharness.js"></script>
+<script>
+"use strict";
+
+function doTest([dom_idl, wakelock_idl, wakelock]) {
+  const idl_array = new IdlArray();
+  idl_array.add_untested_idls(dom_idl);
```
marcoscaceres: We should actually test this too. They depend on each other, so why not just test them all?

wake-lock/idlharness.https.html
```diff
@@ -0,0 +1,32 @@
...
+  idl_array.add_untested_idls('interface Navigator {};');
```
marcoscaceres: Same here. We should test navigator. Don't exclude it.

wake-lock/idlharness.https.html
```diff
@@ -0,0 +1,32 @@
...
+  idl_array.add_idls(wakelock_idl);
+  self.wakelock = wakelock;
```
marcoscaceres: Using "self" here is confusing (because self is window). We should just use "window" (see below)

marcoscaceres (changes requested):
Could you please also rename "idlharness.https.html" to "interfaces.https.html".

wake-lock/idlharness.https.html
```diff
@@ -0,0 +1,32 @@
...
+  self.request = wakelock.createRequest();
+  idl_array.add_objects({WakeLock: ['wakelock'], WakeLockRequest: ['request']});
+  idl_array.test();
+}
+
+promise_test(() => {
+  return Promise.all([
+    fetch("/interfaces/dom.idl").then((response) => response.text()),
+    fetch("/interfaces/wake-lock.idl").then((response) => response.text()),
+    navigator.getWakeLock("screen")
```
marcoscaceres: Will this prompt for permission? I can't remember :)

wake-lock/idlharness.https.html
```diff
@@ -0,0 +1,32 @@
...
+  ]).then(doTest);
+}, "Test driver");
```
marcoscaceres: Nit: This should describe the test a bit better... see above.

wake-lock/idlharness.https.html
Honry: The main DOM IDL has been tested in https://github.com/w3c/web-platform-tests/blob/master/dom/interfaces.html, this would avoid duplication.

wake-lock/idlharness.https.html
Honry: `interface Navigator {};` here is the dependency of `partial interface Navigator`, we don't need to test navigator itself, but I should add `Navigator` object to `idl_array` below to test the inheritance between `navigator` and `getWakeLock`.

wake-lock/idlharness.https.html
Honry: Thanks, I will fix it.

wake-lock/idlharness.https.html
Honry: Could you clarify why we need to rewrite like this? The test style are consistent with most IDL tests in w-p-t. e.g. https://github.com/w3c/web-platform-tests/blob/master/dom/interfaces.html

wake-lock/idlharness.https.html
Honry: This is up to user agent implementation, the spec didn't constrain to request user permission.

wake-lock/idlharness.https.html
Honry: This is a standalone test for `promise_test` itself, I just want to keep consistent with most of IDL tests in w-p-t. :)

Honry:
@marcoscaceres, thanks for the review. I've addressed part of your comments, PTAL.

andrey-logvinov:
@Honry thank you! Looks good to me.

wake-lock/idlharness.https.html
marcoscaceres: The "old" style is because there was no universal support for async/await. We now have async/await, so would prefer we use it.

wake-lock/idlharness.https.html
marcoscaceres: This might cause tests to lock up tho. Maybe we need some kind of timeout here?

marcoscaceres: (actually, the test can time itself out... ignore my last comment there.)

wake-lock/idlharness.https.html
marcoscaceres: It depends: I'm ok to not use it, so long as there is no dependency on each other. 

I can't recall if there is or not - but if this spec depends on anything in DOM being there, we should definitely test it.

wake-lock/idlharness.https.html
Honry: Yeah, fixed, it looks more simplified. Thanks!

marcoscaceres (approved):
It's beautiful 💅

Honry:
Thank you for the review, @marcoscaceres, @andrey-logvinov. I am merging it.

rakuco:
Awesome, thanks guys!

PR #7898, opened 2017-10-19
Enable uploading manifest to GitHub.

jgraham:
This uploads a MANIFEST.json.gz file for every commit to master, so
that we can build tooling that attempts to download that file as an
initial manifest that can then have a rapid update rather than a slow
build.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7898)

### DIFF
--- a/tools/ci/ci_manifest.sh
+++ b/tools/ci/ci_manifest.sh
@@ -1,0 +1,10 @@
+set -ex
+
+SCRIPT_DIR=$(dirname $(readlink -f "$0"))
+WPT_ROOT=$(readlink -f $SCRIPT_DIR/../..)
+cd $WPT_ROOT
+
+mkdir -p ~/meta
+./wpt manifest -p ~/meta/MANIFEST.json
+# Force overwrite of any existing file
+gzip -f ~/meta/MANIFEST.json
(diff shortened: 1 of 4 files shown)


tools/ci/ci_manifest.sh
```diff
@@ -0,0 +1,9 @@
+set -ex
+
+SCRIPT_DIR=$(dirname $(readlink -f "$0"))
+WPT_ROOT=$(readlink -f $SCRIPT_DIR/../..)
+cd $WPT_ROOT
+
+mkdir -p ~/meta
```
foolip: I don't understand the use of ~ here, where will the file actually end up? Is it in the wpt repo itself?

jgraham: It is generated in the `$HOME/meta` directory on the travis machine, and then finally gets uploaded to the "releases" part of the GitHub repo (if it works correctly :). But since it's not tagged  I don't think it will show up as an actual release. But I hope we can still download it.

foolip: OK, as long as it's not in the wpt repo itself, I have no concerns. Let's see what happens :)

foolip (approved):

Issue #7885, opened 2017-10-18
Should web-animations/interfaces/AnimationEffectTiming/iterationStart.html use assert_approx_equals?


stephenmcgruer:
@birtles 

Chrome currently fails this test with:

FAIL Test that changing the iterationStart affects computed timing when backwards-filling assert_equals: expected 0.5 but got 0.4999999999999999
FAIL Test that changing the iterationStart affects computed timing during the active phase assert_equals: expected 0.5 but got 0.4999999999999999
FAIL Test that changing the iterationStart affects computed timing when forwards-filling assert_equals: expected 0.5 but got 0.4999999999999996

To me that says 'rounding issues', and the test should just use assert_approx_equals instead of assert_equals. If that sounds sane and there's no reason progress should be exactly 0.5, let me know and I'll send out a fix.

stephenmcgruer:
See also web-animations/timing-model/animation-effects/simple-iteration-progress.html:

FAIL Test zero iterations: iterations:0 iterationStart:2.5 duration:100 delay:1 fill:both assert_equals: Value of progress in the before phase expected 0.5 but got 0.4999999999999999
FAIL Test zero iterations: iterations:0 iterationStart:3 duration:100 delay:1 fill:both assert_equals: Value of progress in the before phase expected 0 but got 2.7755575615628914e-16

etc.

birtles:
@stephenmcgruer Yes, that would be great. Thanks!

For `assert_computed_timing_for_each_phase` could you add some local definition `assert_property_equals` that switches between `assert_equals` and `assert_approx_equals` (or, better, `assert_times_equals` since I notice we use that for progress values in at least `web-animations/timing-model/animation-effects/active-time.html`) based on the property name? Perhaps it could also fill in the `actual` argument at the same time.

birtles:
Closed by #7903 / c2d0053

stephenmcgruer:
@birtles There is still https://github.com/w3c/web-platform-tests/pull/7902 which fixes iterationStart.html :)

PR #7902, opened 2017-10-19
Use assert_times_equal to compare 'progress' in iterationStart.html

stephenmcgruer:
As progress is computed from applying the animation effect's timing
function, its output is unbounded and not necessary integer.

See issue #7885


birtles (approved):

birtles:
Sorry, I totally missed this in my inbox somehow.

PR #7909, opened 2017-10-19
Updating WebDriver actions tests to be IE-friendly

jimevans:
Removing ES6 constructs from tests and supporting test page. Also removing
validation of keyboard event `code` property, as IE (and Edge) do not
support it.

### DIFF
--- a/webdriver/tests/actions/support/test_actions_wdspec.html
+++ b/webdriver/tests/actions/support/test_actions_wdspec.html
@@ -16,7 +16,6 @@
         "use strict";
         var els = {};
         var allEvents = { events: [] };
-        var initialMoveHandler;
         function displayMessage(message) {
             document.getElementById("events").innerHTML = "<p>" + message + "</p>";
         }
@@ -104,10 +103,11 @@
           els.dragTarget.removeEventListener("mousedown", grabOnce);
         }
 
-        function dropOnce(event) {
-          initialMoveHandler(event);
-          els.dragArea.removeEventListener("mouseup", dropOnce);
-          initialMoveHandler = null;
+        function dropOnce(moveHandler) {
+          return function (event) {
+            moveHandler(event);
+            els.dragArea.removeEventListener("mouseup", dropOnce);
+          }
         }
 
         function resetEvents() {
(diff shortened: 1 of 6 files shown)


w3c-bots:
# Build PASSED

Started: 2017-10-19 20:45:33
Finished: 2017-10-19 21:27:07


<h2>Failing Jobs</h2>
<ul>

<li>chrome:unstable</li>

<li>safari:10.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/19716)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/290155688)

andreastt:
I feel @mjzffr’s input is required here.

jgraham:
@mjzffr Want to review this?

jgraham:
FWIW I think this change is probably fine. I'd also be happy, maybe happier, if we just didn't require any properties that were undefined in the event to be present so that we could keep the stricter checks for browsers implementing `code`.

webdriver/tests/actions/support/test_actions_wdspec.html
```diff
@@ -132,27 +145,31 @@
               -(areaRect.top + (event.clientY - boxRect.top)),
               20);
           els.dragArea.addEventListener("mousemove", moveHandler);
-          els.dragArea.addEventListener("mouseup", drop(moveHandler), {once: true});
+          initialMoveHandler = drop(moveHandler);
+          els.dragArea.addEventListener("mouseup", dropOnce);
```
mjzffr: I'd like to avoid the global `initialMoveHandler`. Maybe define `dropOnce` inside `grab`, then you shouldn't need `initialMoveHandler`.

mjzffr (changes requested):

webdriver/tests/actions/support/test_actions_wdspec.html
jimevans: I've gone as far as I can trying to eliminate the global event handler. If someone would show me the way to do that, I'd be happy to oblige.

andreastt (approved):
This looks good to me. It also passes the tests.

PR #7910, opened 2017-10-19
Fixes to make the manifest upload job work/unbreak master

jgraham:
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/7910)


gsnedders:
I... don't think this will make anything worse. If this doesn't fix it, we should revert all of these changes, IMO.

gsnedders (approved):

Issue #7589, opened 2017-10-05
requestidlecallback/callback-timeout-with-raf.html assumes an idle period won't start if there are pending rAF callbacks


farre:
The idea of the test is that when a rAF callback takes longer than 40 ms, which is assumed to be longer than the time until the next time that rendering will be updated, and makes a daisy chain call to requestAnimationFrame, then an idle period will never be initiated.

As far as I can tell from the spec, there is no reason for there not being an idle period in this case, given that the next update rendering step is sufficiently far away into the future. And even if the rAF callback has taken longer than the rate that the user agent wishes to run it's rendering updates, there is no requirement that the next rendering update will happen immediately.

foolip:
@rmcilroy, is this a test you added? Can you take a look?

PR #7914, opened 2017-10-20
Change callback-timeout-with-raf to use setTimeout.

rmcilroy:
Use setTimeout instead of rAF to create a busy loop for the
callback-timeout-with-raf test. Since the user-agent can decide to
throttle frame rate, the rAF approach didn't guarantee a busy event loop,
so instead use setTimeout.

Addresses #7589

### DIFF
--- a/requestidlecallback/callback-timeout-when-busy.html
+++ b/requestidlecallback/callback-timeout-when-busy.html
@@ -8,27 +8,27 @@
 async_test(function() {
   // Check whether requestIdleCallback with a timeout works when the event loop
   // is busy.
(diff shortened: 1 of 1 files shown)


rmcilroy:
@farre PTAL - this addresses the rAF issues on Firefox as far as I can tell

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19733)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/290372077)

farre (approved):
Yeah, this looks good. We want to test step 7 in https://w3c.github.io/requestidlecallback/#start-an-idle-period-algorithm and this does exactly that.

PR #7917, opened 2017-10-20
HTML: Test behavior of HTMLElement focus(options) with preventScroll

zcorpan:
Per https://github.com/whatwg/html/pull/2787

This does not test what the expected scroll alignment should be for
focus(), other than expecting that it be scrolled entirely into view.

---

cc @jihyerish @anawhj

### DIFF
--- a/html/editing/focus/processing-model/preventScroll.html
+++ b/html/editing/focus/processing-model/preventScroll.html
@@ -1,0 +1,75 @@
+<!doctype html>
+<title>focus(options) - preventScroll</title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<style>
+#iframe { width: 500px; height: 500px; border: none }
+</style>
+<iframe id=iframe src="support/preventScroll-helper.html"></iframe>
+<script>
+function isEntirelyInView(elm, win) {
+  const inViewHorizontal = (elm.offsetLeft >= win.scrollX) &&
+                           ((elm.offsetLeft + elm.clientWidth) <= (win.scrollX + win.innerWidth));
+  const inViewVertical = (elm.offsetTop >= win.scrollY) &&
+                         ((elm.offsetTop + elm.clientHeight) <= (win.scrollY + win.innerHeight));
+  return inViewHorizontal && inViewVertical;
+}
+
+setup({explicit_done: true});
+
+function resetState(win) {
+  win.scrollTo(0, 0);
+  win.document.activeElement.blur();
+}
+
+onload = () => {
+  const win = document.getElementById('iframe').contentWindow;
+  const elm = win.document.getElementById('button');
+
+  test(() => {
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19739)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/290403421)

html/editing/focus/processing-model/preventScroll.html
```diff
@@ -0,0 +1,75 @@
...
+<script>
+function isEntirelyInView(elm, win) {
+  const inViewHorizontal = (elm.offsetLeft >= win.scrollX) &&
+                           ((elm.offsetLeft + elm.clientWidth) <= (win.scrollX + win.innerWidth));
+  const inViewVertical = (elm.offsetTop >= win.scrollY) &&
+                         ((elm.offsetTop + elm.clientHeight) <= (win.scrollY + win.innerHeight));
+  return inViewHorizontal && inViewVertical;
+}
+
+setup({explicit_done: true});
```
foolip: Why this when there aren't any async tests or anything?

foolip (approved):
LGTM % possibly unnecessary done()

html/editing/focus/processing-model/preventScroll.html
```diff
@@ -0,0 +1,75 @@
...
+  test(() => {
+    resetState(win);
+    elm.focus({preventScroll: false});
+    assert_true(isEntirelyInView(elm, win));
+  }, 'elm.focus({preventScroll: false})');
+
+  test(() => {
+    resetState(win);
+    elm.focus({preventScroll: true});
+    assert_false(isEntirelyInView(elm, win));
```
NavidZ: Instead of this shouldn't we check that scroll offset hasn't changed at all?

html/editing/focus/processing-model/preventScroll.html
domenic: This is necessary since we delay the tests until onload happens. Otherwise the test framework would see that there are no tests and say "all done, 0/0 tests passed".

domenic:
I know @zcorpan is short on time these days. @NavidZ should we land this even with your above review comment, or would you perhaps be able to provide a patch for it?

NavidZ:
@eirage can you test this on your latest Chrome patch and make the modifications suggested here?

NavidZ:
domenic@ I guess we need to have access (like being a maintainer) to be able to push changes on top of others pull requests. Right? or is there another flow I'm not aware of.
Either way feel free to merge this as is. Since we have a change to land to have this feature in Chrome we can modify the test along that change and it gets push back to w3c at the end.

NavidZ (approved):

domenic:
Yeah, some kind of write access is needed I guess, sorry about that. I'll merge this.

PR #7929, opened 2017-10-20
Test for segments that don't start with keyframe.

theodab:
On Safari, segments that do not start on a keyframe cause a decoder
error. This test exposes that bug.
See https://bugs.webkit.org/show_bug.cgi?id=160460


theodab:
I work on Shaka Player, a javascript-based player for streaming content.
As we rely on the MediaSource and Encrypted Media Extensions apis, we have found a large number of browser bugs with them over the years.
We hope to port over these test cases to the web platform tests, in order to make these problems more visible, and prevent regressions for the problems that have already been fixed.
If that's against the intended purpose of the web platform tests, or if adding new media files like this is too heavyweight or anything, please tell me. This is my first time contributing to the web platform tests.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19763)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/290549127)

wolenetz:
@theodab (https://github.com/w3c/web-platform-tests/pull/7929#issuecomment-338267215) - Thanks for submitting this PR. So far, the MSE WPT have been for basic spec conformance testing. However, issues uncovered around interoperability of the implementations of the spec are also good candidates for new tests. I'll take a closer look at this PR soon (apologies for my delay so far; I've been deep in Chromium code recently). In particular, is there no analogue in webm for an "open GOP"? I believe a webm cluster doesn't have to begin with a keyframe, so it seems there could be an avenue for exposing this similar issue in implementations also via a webm video bytestream. WDYT?

theodab:
Looking at the spec, it's possible (but highly discouraged) for webm clusters to not begin with a keyframe. In fact, apparently at some point [Chrome didn't support it](https://bugs.chromium.org/p/chromium/issues/detail?id=229412) (see also [this forum topic](https://productforums.google.com/forum/#!topic/chrome/SASW6i2qDjQ)), so it might have some value as a regression test too. I can try to generate an asset for that (or use one of the ones linked in that forum topic, I suppose).

media-source/mediasource-open-gop.html
```diff
@@ -0,0 +1,72 @@
...
+                  var sourceBufferVideo = mediaSource.addSourceBuffer(typeVideo);
+                  sourceBufferVideo.timestampOffset = 0;
+                  sourceBufferVideo.appendWindowEnd = 1;
+
+                  test.expectEvent(sourceBufferVideo, "updateend");
+                  sourceBufferVideo.appendBuffer(dataVideo);
+                  test.waitForExpectedEvents(function() {
+                      test.expectEvent(mediaElement, "playing");
+                      mediaElement.play();
+                      test.waitForExpectedEvents(function() {
```
wolenetz: This doesn't need to be nested. The util manages this.

media-source/mediasource-open-gop.html
```diff
@@ -0,0 +1,72 @@
...
+                      sourceBufferVideo.timestampOffset = 0;
+                      sourceBufferVideo.appendWindowEnd = 1;
+
+                      test.expectEvent(sourceBufferVideo, "updateend");
+                      sourceBufferVideo.appendBuffer(dataVideoInit);
+                      test.waitForExpectedEvents(function()
+                      {
+                          test.expectEvent(sourceBufferVideo, "updateend");
+                          sourceBufferVideo.appendBuffer(dataVideo);
+                          test.waitForExpectedEvents(function() {
```
wolenetz: ditto (this doesn't need to be nested)

media-source/mediasource-open-gop.html
```diff
@@ -0,0 +1,72 @@
...
+                              test.expectEvent(mediaElement, "playing");
+                              mediaElement.play();
+                              test.waitForExpectedEvents(function() {
```
wolenetz: ditto (this doesn't need to be nested).

media-source/mediasource-open-gop.html
wolenetz: ditto (this doesn't need to be nested).

media-source/mediasource-open-gop.html
```diff
@@ -0,0 +1,72 @@
...
+
+            mediasource_test(function(test, mediaElement, mediaSource) {
+              mediaElement.addEventListener("error", test.unreached_func("Unexpected event 'error'"));
+              MediaSourceUtil.fetchManifestAndData(test, manifestFilenameVideoInit, function(typeVideoInit, dataVideoInit)
+              {
+                  MediaSourceUtil.fetchManifestAndData(test, manifestFilenameVideo, function(typeVideo, dataVideo)
+                  {
+                      var sourceBufferVideo = mediaSource.addSourceBuffer(typeVideo);
+                      sourceBufferVideo.timestampOffset = 0;
+                      sourceBufferVideo.appendWindowEnd = 1;
```
wolenetz: Is this needed?

media-source/mediasource-open-gop.html
```diff
@@ -0,0 +1,72 @@
...
+            var manifestFilenameVideo = "mp4/test-v-no-keyframe-segment-manifest.json";
+
+            mediasource_test(function(test, mediaElement, mediaSource) {
+              mediaElement.addEventListener("error", test.unreached_func("Unexpected event 'error'"));
+              MediaSourceUtil.fetchManifestAndData(test, manifestFilenameVideoInit, function(typeVideoInit, dataVideoInit)
+              {
+                  MediaSourceUtil.fetchManifestAndData(test, manifestFilenameVideo, function(typeVideo, dataVideo)
+                  {
+                      var sourceBufferVideo = mediaSource.addSourceBuffer(typeVideo);
+                      sourceBufferVideo.timestampOffset = 0;
```
wolenetz: This is the default offset. Setting it to 0 on a new SourceBuffer is redundant.

media-source/mediasource-open-gop.html
```diff
@@ -0,0 +1,72 @@
...
+          var subType = MediaSourceUtil.getSubType(MediaSourceUtil.VIDEO_ONLY_TYPE);
+          if (subType == 'webm') {
+            var manifestFilenameVideo = "webm/test-v-no-keyframe-cluster-manifest.json";
+
+            mediasource_test(function(test, mediaElement, mediaSource) {
+              mediaElement.addEventListener("error", test.unreached_func("Unexpected event 'error'"));
+              MediaSourceUtil.fetchManifestAndData(test, manifestFilenameVideo, function(typeVideo, dataVideo)
+              {
+                  var sourceBufferVideo = mediaSource.addSourceBuffer(typeVideo);
+                  sourceBufferVideo.timestampOffset = 0;
```
wolenetz: ditto comment below (0 is default already, this is redundant).

media-source/mediasource-open-gop.html
```diff
@@ -0,0 +1,72 @@
...
+                  sourceBufferVideo.appendWindowEnd = 1;
```
wolenetz: ditto comment below (is setting a smaller-than-default window necessary?)

media-source/mediasource-open-gop.html
```diff
@@ -0,0 +1,72 @@
...
+            var manifestFilenameVideoInit = "mp4/test-v-no-keyframe-segment-manifest-init.json";
+            var manifestFilenameVideo = "mp4/test-v-no-keyframe-segment-manifest.json";
+
+            mediasource_test(function(test, mediaElement, mediaSource) {
+              mediaElement.addEventListener("error", test.unreached_func("Unexpected event 'error'"));
+              MediaSourceUtil.fetchManifestAndData(test, manifestFilenameVideoInit, function(typeVideoInit, dataVideoInit)
+              {
+                  MediaSourceUtil.fetchManifestAndData(test, manifestFilenameVideo, function(typeVideo, dataVideo)
+                  {
+                      var sourceBufferVideo = mediaSource.addSourceBuffer(typeVideo);
```
wolenetz: Maybe add a quick assert that typeVideoInit matches typeVideo

media-source/mp4/test-v-no-keyframe-segment-manifest.json
```diff
@@ -0,0 +1,4 @@
+{
+    "url": "mp4/mp4-main-ogop-h264ogop_mid1.mp4",
```
wolenetz: Why not just have a single mp4 for the mp4 variant of this new test, that is a concatenation of the init plus mid1? That's how it's used in the test.

media-source/mediasource-open-gop.html
```diff
@@ -0,0 +1,72 @@
...
+        <title>Open GOP test cases.</title>
+        <script src="/resources/testharness.js"></script>
+        <script src="/resources/testharnessreport.js"></script>
+        <script src="mediasource-util.js"></script>
+    </head>
+    <body>
+        <div id="log"></div>
+        <script>
+
+          var subType = MediaSourceUtil.getSubType(MediaSourceUtil.VIDEO_ONLY_TYPE);
```
wolenetz: Between <script> and </script>, start with indent of 0 at outermost level (this is common to how the intended format of these tests should be in web-platform-tests IIRC.)

media-source/mediasource-open-gop.html
```diff
@@ -0,0 +1,72 @@
+<!DOCTYPE html>
+<!-- Copyright © 2016 Chromium authors and World Wide Web Consortium, (Massachusetts Institute of Technology, ERCIM, Keio University, Beihang). -->
```
wolenetz: s/2016/2018/

wolenetz:
Looking pretty good. All my comments on https://github.com/w3c/web-platform-tests/pull/7929/commits/6348520958c703ce30ed8948ff9bac7fc48eb76f are pretty much nits. I have not proofed the binary webm nor mp4 files yet though.

theodab:
Ok I've consolidated the files and flattened the expectations. And it still fails on Safari and passes on Chrome, so I don't think this broke anything.
I wanted to proof the webm files myself but I don't really know how to analyze them. I've asked around for tips or tools, but gotten no responses yet.

wolenetz:
@ https://github.com/w3c/web-platform-tests/pull/7929#issuecomment-365444108 - Though sometimes limited (such as newer metadata tags not parsed/recognized), there is a media metadata viewer called 'hachoir-urwid' that I've found useful in the past for webm metadata. Also, as a Chrome MSE implementation owner, I also used debug builds of Chromium with extra logging to understand problematic media.

wolenetz:
Also, I would rename the tests and PR to describe the intent (since "open vs closed GOP" is *very* loosely and conflictingly defined everywhere). In particular: the conformance test you're adding is testing whether or not the MSE implementation correctly handles media segments that do not begin with a **random access point** (as defined in each specific MSE bytestream spec) for every track in that segment. In practice, this comes down to whether or not the implementation correctly drops initial nonkeyframes in a track (unless the media segment is continuous with the previously appended media segment) until the first keyframe is reached for that track (following the logic in the MSE coded frame processing algorithm).

media-source/webm/test-v-no-keyframe-cluster-manifest.json
```diff
@@ -0,0 +1,4 @@
+{
+    "url": "webm/webm-no-keyframe-cluster.webm",
```
wolenetz: Inspecting the binary webm file using hachoir-urwid:
It looks like it contains 2 tracks (track 1 = video, track 2 = audio).
Cluster[0]:
SimpleBlock[0]: track 1 (video), keyframe bit is SET
SimpleBlock[1 through 3, and 5 through 6]: track 2 (audio), keyframe bit is NOT SET
SimpleBlock[4]: track 1 (video), keyframe bit is NOT SET

Cluster[1] has 13 SimpleBlocks, following same pattern as Cluster[0]: first video frame is a keyframe, remainder of video and all of audio are not keyframes.

There are 53 total clusters in the webm file -- I didn't go through all of them.
I then ran the file through a debug-build of tip-of-tree Chromium with logging to see what the Chromium MSE parsers see.

Items in the binary that I believe need fixing:
1) All audio frames should be marked keyframe. Chromium is lenient and updates those in the parser to be keyframe. Other implementations **aren't required to be so lenient**. (You can see examples of Chromium noticing this from the player's log in chrome://media-internals having entries like: "Bytestream with audio frame PTS 0us and DTS 0us indicated the frame is not a random access point (key frame). All audio frames are expected to be key frames.")

2) These clusters' first video track frame (at least for the first couple clusters) is a keyframe. This sounds like the reverse of the scenario you intend to test, unless I'm mistaken.

3) All of the media is contained in SimpleBlocks within Clusters. Notably, having a SimpleBlock at the end of a Cluster in a muxed multi-track stream leads to inability of implementations to understand the intended duration of that block at the end of the cluster (earlier block durations in the cluster for that track can be inferred by timestamp delta; not so with the last block). This can result in the MSE coded frame processing algorithm's discontinuity detection logic kicking in, forcing all tracks to drop nonkeyframes until their next keyframe is received. So, appending this media in Chrome M64 leads to 5 disjoint buffered ranges (due to video nonkeyframes being dropped after a perceived discontinuity due to under-guessed audio block durations (5ms, instead of 23ms or 24ms) from the end of one cluster to the beginning of the next). In tip-of-tree/Canary M66, there's logic that helps a little: those nonkeyframes are still dropped, but the gaps are coalesced better. Regardless, the guidance for muxers of WebM is to use BlockGroup (which encode explicit BlockDuration) for at least the last block for each track in a cluster. That avoids requiring implementations to guess those durations (which can result in unexpected gaps/stalls/dropped nonkeyframes).

wolenetz: Item 3 can be alleviated if you need by
a) using DefaultDuration to be close to the actual block durations, or have all the blocks have very similar durations. Unfortunately, the first audio block in the first cluster was 5ms, unlike the rest that I inspected (23-24ms), and
b) using just one media track per SourceBuffer (e.g., just video for your case, since audio can't have nonkeyframes...)

media-source/mp4/test-v-no-keyframe-segment-manifest.json
```diff
@@ -0,0 +1,4 @@
+{
+    "url": "mp4/mp4-main-ogop-h264ogop.mp4",
```
wolenetz: Inspecting the binary mp4 file using Chrome/Chromium, I see:
Chrome M64 (Release build): chrome://media-internals contains 10 debug warnings: "Prepared video sample is not conformant" -- this is logged when the bitstream converter detected a problem (specifically, this method returned false for those 10 frames: https://cs.chromium.org/chromium/src/media/formats/mp4/avc.cc?rcl=61cae6f129576601cb0dd964a6ce61a41e82abe0&l=180)

tip-of-tree DEBUG build with proprietary codecs:
Same 10 warnings in chrome://media-internals
debug logs indicate these occur at the end of each media segment, from: https://cs.chromium.org/chromium/src/media/formats/mp4/avc.cc?rcl=61cae6f129576601cb0dd964a6ce61a41e82abe0&l=227

Regardless, the media parses and plays.

Also, unlike the webm binary, this file has just one track (video), and exactly one keyframe, and multiple media segments -- several that begin with nonkeyframes. I believe this matches the scenario you're trying to test support for (where not all media segments begin all their tracks with keyframes).

You might consider making the first media segment appended *not* begin with a keyframe. Compliant implementations must accept and drop such nonkeyframes until the first keyframe. This would be yet a stronger test of compliance.

wolenetz: (All 10 of the extra debug logs for those nonconformat samples indicate "Unexpected NALU type 8 in order_state 2").

wolenetz:
I've shared my private Chromium build's logs with @theodab directly. Context: https://github.com/w3c/web-platform-tests/pull/7929#pullrequestreview-96712845 and https://github.com/w3c/web-platform-tests/pull/7929#pullrequestreview-96709459

theodab:
I'm not surprised that the webm file doesn't match the intended test scenario at all, because it's just a random webm file I found online that supposedly had the same problem.
I'll see if I can use ffmpeg to generate something instead.

theodab:
Ok, I spent quite a while trying to get ffmpeg to generate a better mp4 for this issue, or to make a webm file for this issue, to no avail. It is surprisingly hard to make a purposefully malformed media file, it turns out. I don't think I can justify spending more time or effort on this one test, so I am purposefully cutting back. I removed the webm test, and just put in TODOs for possible future efforts.

jyavenard:
this test is invalid for webm.
@wolenetz " I believe a webm cluster doesn't have to begin with a keyframe, so it seems there could be an avenue for exposing this similar issue in implementations also via a webm video bytestream."

actually, it does.
the webm format spec states "Key frames SHOULD be placed at the beginning of clusters.
    Having key frames at the beginning of clusters should make seeking faster and easier for the client.
"
however, it states "Muxers should treat all guidelines marked SHOULD in this section as MUST. This will foster consistency across WebM files in the real world"

As such, no muxer can legally produce webm media segments where a cluster doesn't start with a keyframe.

joeyparrish:
@jyavenard, thank you for the info!

theodab:
Well, now I don't feel so bad about failing to make a webm for it.

alvestrand:
Removing mediacapture-streams and wg-webrtc tags, since this seems to be solely an MSE PR.

PR #7935, opened 2017-10-20
navigating-across-documents: revisions and explanations for 005, 006, 007, 008 & 009

rwaldron:
Test | Chrome Canary | Edge | Firefox Nightly | Safari Technology Preview
-- | -- | -- | -- | --
005.html | ✔ | ✔ | ✔ | ✔
006.html | ✔ | ✔ | ✔ | ✔
007.html | ✔ | ✔ | ✔ | ✔
008.html | ✘ Navigating the anchor's href is expected to abort the form submit action | ✘ Navigating the anchor's href is expected to abort the form submit action  | ✘ Navigating the anchor's href is expected to abort the form submit action | ✘ Navigating the anchor's href is expected to abort the form submit action 
009.html | ✘ Navigating the anchor's href is expected to abort the form submit action | ✘ Navigating the anchor's href is expected to abort the form submit action  | ✘ Navigating the anchor's href is expected to abort the form submit action | ✘ Navigating the anchor's href is expected to abort the form submit action 

Notes: 

All platforms fail for the same reason, which is that they do not execute `javascript:...` url code asynchronously, per [step 12](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents). This is likely a candidate for changing the specification (which would mean changes to these tests as well).


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19772)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/290621568)

html/browsers/browsing-the-web/navigating-across-documents/005.html
```diff
@@ -1,16 +1,29 @@
 <!doctype html>
-<title>Link with onclick navigation and href navigation </title>
+<link rel="help" href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents">
+<title>Link with onclick navigation is aboprted by href navigation </title>
```
bzbarsky: "aborted"

html/browsers/browsing-the-web/navigating-across-documents/005.html
```diff
@@ -1,16 +1,29 @@
...
 <a target="test" onclick="document.getElementById('test').contentWindow.location='click.html'" href="href.html">Test</a>
 <script>
-var t = async_test();
-t.step(function() {document.links[0].click()});
-onmessage = t.step_func(
+var test = async_test();
+test.step(function() {
+  document.querySelector("a").click();
+});
+window.onmessage = test.step_func(
```
bzbarsky: Just use step_func_done() here and drop the explicit done() call?

html/browsers/browsing-the-web/navigating-across-documents/005.html
```diff
@@ -1,16 +1,29 @@
...
   function(e) {
     assert_equals(e.data, "href");
-    t.done();
+    test.done();
   });
 </script>
+
+<!--
+1. This test creates a iframe, and an anchor with both an `onclick` and `href`.
```
bzbarsky: "an iframe"

html/browsers/browsing-the-web/navigating-across-documents/006.html
```diff
@@ -1,17 +1,37 @@
 <!doctype html>
-<title>Link with onclick form submit and href navigation </title>
+<link rel="help" href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#javascript-protocol">
```
bzbarsky: How is the javascript: section relevant here?

html/browsers/browsing-the-web/navigating-across-documents/007.html
```diff
@@ -1,16 +1,29 @@
 <!doctype html>
-<title>Link with onclick javascript url and href navigation </title>
+<link rel="help" href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents">
+<title>Link with onclick document.write is aborted by href navigation</title>
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <div id="log"></div>
 <iframe id="test" name="test"></iframe>
-<a target="test" onclick="document.getElementById('test').contentWindow.location = 'javascript:\'abc<script>parent.postMessage(&quot;click&quot;, &quot;*&quot;)</script>\'';" href="href.html">Test</a>
+<a target="test" onclick="document.getElementById('test').contentWindow.document.write('<script>self.postMessage(&quot;click&quot;, &quot;*&quot;)</script>');" href="href.html">Test</a>
```
bzbarsky: Hmm.  So this totally changes what this test is testing; it used to test javascript: navigation, but now it tests document.open/write.  Why the change?  I'd like to understand that before reviewing the details here.

html/browsers/browsing-the-web/navigating-across-documents/008.html
```diff
@@ -1,17 +1,37 @@
...
+<link rel="help" href="https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#submit-get-action">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#plan-to-navigate">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#planned-navigation">
+<title>Link with onclick form submit to javascript: url is aborted by href navigation</title>
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <div id="log"></div>
 <iframe id="test" name="test"></iframe>
-<form target="test" action="javascript:'<script>parent.postMessage(&quot;click&quot;, &quot;*&quot;)</script>'"></form>
+<form target="test" action="javascript:parent.postMessage('click', '*')"></form>
```
bzbarsky: This is changing the test significantly.  The old test tested what happens when a javascript: URI returns a string, which is then treated as the body of a response to navigate to and ends up running scripts itself.  The new test just runs script to post a message.  Those are different operations, and we should probably have tests for both.

It's not actually 100% clear to me from the spec what the behavior should be for this rewritten testcase (nor indeed for the pre-rewrite one)...

html/browsers/browsing-the-web/navigating-across-documents/008.html
```diff
@@ -1,17 +1,37 @@
...
+});
 </script>
+
+<!--
+1. This test creates an iframe, a form with
+    `action="javascript:parent.postMessage('click', '*')", and an anchor
+    with both an `onclick` and `href`.
+2. The test "clicks" the anchor, triggering the `onclick` handler operation,
+    which will attempt to submit the form. The form's `action` is a
+    `javascript:` url, which has special handling in step 12
```
bzbarsky: Step 12 of what algorithm?

html/browsers/browsing-the-web/navigating-across-documents/008.html
```diff
@@ -1,17 +1,37 @@
...
+3. Navigating the iframe to the anchor's `href` will abort the `action`.
```
bzbarsky: This is the part that is not obvious to me in terms of how it works.  Might need spec clarifications.

What do browsers end up doing here?

html/browsers/browsing-the-web/navigating-across-documents/009.html
```diff
@@ -1,22 +1,40 @@
...
+<link rel="help" href="https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#submit-get-action">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#plan-to-navigate">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#planned-navigation">
+<title>Link with onclick form submit to javascript: url with document.write and href navigation</title>
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <div id="log"></div>
 <iframe id="test" name="test"></iframe>
-<form target="test" action="javascript:(function() {document.write('<script>parent.postMessage(&quot;write&quot;, &quot;*&quot;)</script>'); return '<script>parent.postMessage(&quot;click&quot;, &quot;*&quot;)</script>'})()"></form>
+<form target="test" action="javascript:document.write('<script>parent.postMessage(&quot;write&quot;, &quot;*&quot;)</script>');"></form>
```
bzbarsky: This is also significantly changing the test.  The previous test was testing both whether the postMessage from the javascript: evaluation itself happened and whether the postMessage from parsing the evaluation result happened.  The new version doesn't have an evaluation result at all.

html/browsers/browsing-the-web/navigating-across-documents/009.html
```diff
@@ -1,22 +1,40 @@
...
+});
 </script>
+
+<!--
+1. This test creates an iframe, a form with
+    `action="javascript:document.write('<script>parent.postMessage(&quot;write&quot;, &quot;*&quot;)</script>');",
+    and an anchor with both an `onclick` and `href`.
+2. The test "clicks" the anchor, triggering the `onclick` handler operation,
+    which will attempt to submit the form. The form's `action` is a
+    `javascript:` url, which has special handling in step 12
```
bzbarsky: Step 12 of what?

html/browsers/browsing-the-web/navigating-across-documents/009.html
```diff
@@ -1,22 +1,40 @@
...
+3. Navigating the iframe to the anchor's `href` will abort the `action`.
```
bzbarsky: This part is, again, entirely non-obvious.

bzbarsky (commented):
Thank you for documenting more of this stuff!

bzbarsky:
Oh, and as far as the Gecko implementation goes, we most certainly execute javascript: synchronously.  But we also do `<a href>` traversal asynchronously, which may matter.  To tease those effects apart, one would need to race a javascript: execution against a known-sync way of entering navigation, like a `location.href` assignment.

html/browsers/browsing-the-web/navigating-across-documents/007.html
rwaldron: This is a mistake, I need to revisit 007, 008 and 009.

html/browsers/browsing-the-web/navigating-across-documents/008.html
rwaldron: https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents, Step 6: 

> Cancel any preexisting but not yet mature attempt to navigate browsingContext, including canceling any instances of the fetch algorithm started by those attempts. If one of those attempts has already created and initialized a new Document object, abort that Document also. (Navigation attempts that have matured already have session history entries, and are therefore handled during the update the session history with the new page algorithm, later.)

The form action navigation is started and reaches step 12 of [navigating-across-documents](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents), which enters **If resource is a request whose url's scheme is "javascript"**, and does "Queue a task to run these "javascript: URL" steps, associated with the active document of browsingContext:". At this point `onclick` returns, and the `href` navigation begins. When it reaches step 6 of [navigating-across-documents](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents), it should cancel the form action navigation—however I believe there is a spec bug and I plan to write a fix proposal: there is no explicit spec language to remove that previously queued task (for the "javascript: URL").

html/browsers/browsing-the-web/navigating-across-documents/008.html
rwaldron: > This is changing the test significantly. The old test tested what happens when a javascript: URI returns a string, which is then treated as the body of a response to navigate to and ends up running scripts itself. 
The new test just runs script to post a message. Those are different operations, and we should probably have tests for both

Yes, all of this is absolutely true. [Chrome does not honor the target attribute when handling "`javascript:` URL" navigations](https://bugs.chromium.org/p/chromium/issues/detail?id=749492), so running this test results in a blank screen, which is a trash result for wpt. To properly test this, I've created new tests that nest the entire operation into an iframe—those are in a different branch.

html/browsers/browsing-the-web/navigating-across-documents/009.html
rwaldron: Some of these had to be rewritten entirely due to Chrome/Blink not honoring the `target=` when handling a "javascript: url" (which results in this test producing a blank page in Chrome). Despite that, I believe this test was wrong because the "javascript: url" code evaluation and execution is queued in a task (step 12 of [navigating-across-documents](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents)), so when step 11 ("Return to whatever algorithm invoked the navigation steps and continue running these steps in parallel.") returns and the algorithm moves on to navigating to href.html, it will then encounter step 6 of [navigating-across-documents](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents) and "Cancel any preexisting but not yet mature attempt to navigate _browsingContext_,". This is another example of under-specified behavior that I've identified, ie: there is no way to explicitly target a "preexisting ... attempt to navigate _browsingContext_," that exists within a task that's presently in a given queue (for removal at step 6)

rwaldron:
>  we most certainly execute javascript: synchronously. But we also do <a href> traversal asynchronously, which may matter. To tease those effects apart, one would need to race a javascript: execution against a known-sync way of entering navigation, like a location.href assignment.

I'm struggling with this... At first I thought you were implying that I should try to write tests that accommodated for this, but it occurs to me that doesn't make sense: if Gecko is executing "javascript: URL" synchronously, then it's not conforming to the spec as it's presently written.

bzbarsky:
Er...  That was just a typo.  Gecko executes javascript: _asynchronously_.

So if you have `<a href="javascript:foo()">` then in Gecko if you click that link (or call `click()` on it) the following will happen:

1) Queue a task to do the link traversal.
2) When that task executes, queue a task to run the JS.

Does that help?

rwaldron:
@bzbarsky _nudge_ 

note: this will make more sense if you review https://github.com/w3c/web-platform-tests/pull/8558 first

bzbarsky:
I'm not going to get to this review.  :(

PR #7989, opened 2017-10-27
fix ref links for files with base tags

jobeland:
There are a few files that incorrectly reference their ref/match file in their link tag. The path would normally be correct, but due to the base tags added in the file, the resulting URL is incorrect. This change just modifies the href to adjust for the folder(s) added from the base.

Adding White-List entries to lint file since it seems the linter doesn't take the base tag into consideration

### DIFF
--- a/css/css-style-attr/style-attr-urls-002.xht
+++ b/css/css-style-attr/style-attr-urls-002.xht
@@ -5,7 +5,7 @@
   <base href="support/"/>
   <link rel="author" title="Elika J. Etemad" href="http://fantasai.inkedblade.net/contact"/>
   <link rel="reviewer" title="Arron Eicholz" href="mailto:Arron.Eicholz@microsoft.com"/>
-  <link rel="match" href="reference/ref-green-on-green.xht"/>
+  <link rel="match" href="../reference/ref-green-on-green.xht"/>
   <link rel="help" href="http://www.w3.org/TR/css-style-attr/#interpret"/>
   <link rel="help" href="http://www.w3.org/TR/html5/the-base-element.html#the-base-element"/>
   <meta http-equiv="Content-Style-Type" content="text/css" />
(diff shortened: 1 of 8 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/19962)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/293428673)

jobeland:
looks like the linter is failing on a file I haven't touched: 
ERROR:lint:webdriver/tests/sessions/new_session/support/__init__.py:1: CR character in line separator (CR AT EOL)

Anyone know what would be causing this?

jgraham:
This will break the test runner. Not supporting base is intentional for performance reasons (building the manifest is already slow, and getting base right is complex).

jobeland:
@jgraham how does the test runner currently find the ref/match urls? Is the manifest built beforehand from the text content of the file itself? For example, in https://w3c-test.org/css/css-style-attr/style-attr-urls-002.xht its match link points to https://w3c-test.org/css/css-style-attr/support/reference/ref-green-on-green.xht, which is a 404. If that's intentional, is there any story/guidance for anyone else who wishes to test it either manually or via webdriver?

jgraham:
It currently uses the `link rel=[mismatch]` literally as a path to the reference excluding the base URL. This is put into the `MANIFEST.json` file that any test runner is supposed to use as the definitive list of tests. Not accounting for base is unfortunate, but I would at least need to see evidence it's not a significant performance regression to check in all cases when only a few tests actually use it.

To run tests the recommended method is using `wpt run <path>` (or on Windows `python wpt run <path>`. There is also a manual runner in the repository under `tools/runner` that can be used for running tests manually which many people do use, but it isn't as well maintained.

gsnedders:
https://github.com/w3c/web-platform-tests/blob/master/docs/_writing-tests/reftests.md should probably document it ignores the document base URL and uses the document fallback URL.

PR #8005, opened 2017-10-30
Tag master with the PR id that merged at that point.

jgraham:
For the manifest download stuff we have to have tags per PR merge on
master anyway, and this seems like rather useful information.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8005)

### DIFF
--- a/tools/ci/tag_master.py
+++ b/tools/ci/tag_master.py
@@ -1,0 +1,92 @@
+import base64
+import json
+import logging
+import os
+import sys
+import urllib2
+
+here = os.path.abspath(os.path.dirname(__file__))
+wpt_root = os.path.abspath(os.path.join(here, os.pardir, os.pardir))
+
+if not(wpt_root in sys.path):
+    sys.path.append(wpt_root)
+
+from tools.wpt.testfiles import get_git_cmd
+
+logging.basicConfig()
+logger = logging.getLogger(__name__)
+
+
+def get_pr(repo, owner, rev):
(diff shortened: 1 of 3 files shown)


w3c-bots:
# Build PASSED

Started: 2017-10-30 11:27:56
Finished: 2017-10-30 11:42:50





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20011)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/294805956)

gsnedders:
Surely we need to authenticate to be able to add a tag?

jgraham:
Oh, right we are authenticated for the deploy, but not at this point

jgraham:
Something like that, perhaps

tools/ci/tag_master.py
```diff
@@ -0,0 +1,85 @@
+import json
+import logging
+import os
+import sys
+import urllib2
+
+here = os.path.abspath(os.path.dirname(__file__))
+wpt_root = os.path.abspath(os.path.join(here, os.pardir, os.pardir))
+
+if not(wpt_root) in sys.path:
```
gsnedders: This is pretty weird. `not(str)` is always `False`, and `False` will never appear in `sys.path`.

gsnedders (commented):
See the below comment.

I'd also like to see some testing of this prior to landing, given prior CI changes breaking everything. Maybe change the branch that it tags for to `tag_master` and add a prefix to the tag?

gsnedders (approved):

PR #8015, opened 2017-10-31
[MIX]: Improve Safari support.

mikewest:
Safari doesn't generate events in quite the same way as Firefox and Chrome, but
those events also aren't relevant to the MIX tests. In particular, it doesn't fire an
`error` event when blocking media requests, but instead lets the media stall out.
This patch catches that event, and ensures that the test's success/failure is
based on whether or not the resource was loaded from the server, not based on
which event fired on the page.


mikewest:
@foolip WDYT?

mikewest:
@estark37 as well, since she _loves_ mixed content (and might actually be at work today?)

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20052)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/295252005)

mixed-content/generic/expect.py
```diff
@@ -95,7 +95,7 @@ def main(request, response):
 
     response.add_required_headers = False
     response.writer.write_status(200)
-    response.writer.write_header("content-type", content_type)
-    response.writer.write_header("cache-control", "no-cache; must-revalidate")
+    response.writer.write_header("Content-Type", content_type)
+    #response.writer.write_header("cache-control", "no-cache; must-revalidate")
```
estark37: Did you mean to comment this out? And in the line above, does the capitalization of Content-Type matter? O.o

estark37 (commented):
s/loves mixed content/loves to hate mixed content/

I don't think I understand the timeout thing... is the idea that Safari doesn't fire events, but once the timeout triggers, the test will continue as normal? i.e. Safari will pass the tests but it takes a second for each request?

mixed-content/generic/common.js
```diff
@@ -313,10 +321,18 @@ function createMediaElement(type, media_attrs, source_attrs) {
   var sourceElement = createElement("source", {});
 
   mediaElement.eventPromise = new Promise(function(resolve, reject) {
-    mediaElement.addEventListener("loadeddata", resolve);
-
-    // Notice that the source element will raise the error.
-    sourceElement.addEventListener("error", reject);
+    // Setting a timeout, as some browsers don't fire `error` events.
```
foolip: This bit is surprising, the error event on source elements is  practically ancient now, I would have expected it to work. Although, I don't see why the error event should fire here, there's a src attribute on the media element so the error event would be on the media element.

Maybe use the getVideoURI() helper that's used in some places?

mixed-content/generic/expect.py
mikewest: Nope, didn't mean to commit this at all. I was poking at things to see if they would make Safari happy; this wasn't it.

mixed-content/generic/common.js
mikewest: I dug in a bit more; it looks like WebKit isn't firing an `error` event when blocking mixed content. It does fire a `stalled` event at the media element, though, so I think I can hook that rather than adding the timeout for media elements. It's not clear that I can do the same for prefetch, but perhaps I can fail those early with some feature-detection magic. Hrm. Will look into it. :)

> Maybe use the getVideoURI() helper that's used in some places?

These tests are loading a strange python file that streams bits from a real media file, while also poking at wpt's stash API to record whether the file was requested. I don't think it's worth either teaching that helper about the mechanism for these tests, or teaching these tests to look like the file that helper is loading. :)

mikewest:
Updated the patch and description. Now we still have ~14 timeouts in Safari for `<object>` tags (which also don't fire `error` events), but it's better than it was before!

mikewest:
Ping? :)

mixed-content/generic/common.js
```diff
@@ -82,9 +82,12 @@ function setAttributes(el, attrs) {
  */
 function bindEvents(element, resolveEventName, rejectEventName) {
   element.eventPromise = new Promise(function(resolve, reject) {
-    element.addEventListener(resolveEventName  || "load", resolve);
-    element.addEventListener(rejectEventName || "error",
-                             function(e) { e.preventDefault(); reject(); } );
+    element.addEventListener(resolveEventName  || "load", function (e) {
```
estark37: is this change necessary?

mixed-content/generic/common.js
```diff
@@ -313,10 +318,18 @@ function createMediaElement(type, media_attrs, source_attrs) {
   var sourceElement = createElement("source", {});
 
   mediaElement.eventPromise = new Promise(function(resolve, reject) {
-    mediaElement.addEventListener("loadeddata", resolve);
+    mediaElement.addEventListener("loadeddata", function (e) {
```
estark37: ditto

estark37 (approved):

mixed-content/generic/common.js
foolip: Oh, didn't realize this wasn't just loading a static resource.

mixed-content/generic/common.js
```diff
@@ -313,10 +318,18 @@ function createMediaElement(type, media_attrs, source_attrs) {
...
+      resolve(e);
+    });
+
+    // Safari doesn't fire an `error` event when blocking mixed content.
```
foolip: What behavior is required by specs, and is there another test that'd still be failing in Safari assuming Safari's isn't the right behavior?

foolip (approved):
Left a question, but will leave the rest to @estark37

mixed-content/generic/common.js
mikewest: > What behavior is required by specs,

The spec turns mixed content blocking into a synthetic network error. So, the element should react in whatever way it reacts to any other network error. I think `error` is correct, and I filed https://bugs.webkit.org/show_bug.cgi?id=179116 to follow up on it.

> and is there another test that'd still be failing in Safari assuming Safari's isn't the right behavior?

I don't understand the question, can you clarify?

foolip: What I mean is, is there now a test that's failing because of https://bugs.webkit.org/show_bug.cgi?id=179116 and that will start passing in WebKit if the bug is fixed?

We're building stuff for wpt.fyi to highlight tests that are failing in "your" browser but passing in some/most/all others, so making sure there are failing tests will hopefully, eventually, be a way of getting interop bugs like this higher up on the radar of the team who needs to make a change.

gsnedders:
This is approved, is there any reason why it isn't merged? (Had CI not finished when it was approved?)

foolip:
I asked an annoying question in https://github.com/w3c/web-platform-tests/pull/8015/files#r148930959, I think that's why. Fine with me to ignore it though.

PR #8023, opened 2017-11-01
Remove bits from lint.whitelist that have been removed

foolip:
Found with this script:
```bash
cat lint.whitelist | grep -E '^[A-Z].*:' | cut -d : -f2 | grep -vF ' *.' | grep -vF '/*/' | \
while read path; do
    [[ "$(ls $path 2> /dev/null)" == "" ]] && echo $path
done
```

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8023)

### DIFF
--- a/lint.whitelist
+++ b/lint.whitelist
@@ -10,7 +10,6 @@
 CR AT EOL: svg/import/*
 
 INDENT TABS: .gitmodules
-INDENT TABS: app-uri/*
 INDENT TABS: conformance-checkers/*
 INDENT TABS: content-security-policy/*
 INDENT TABS: custom-elements/*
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PASSED

Started: 2017-11-01 04:50:19
Finished: 2017-11-01 04:54:58





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20070)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/295600484)

jgraham (approved):
I assume you ran `wpt lint --all`

If you don't feel like doing this again, it seems plausible to add a mode to the lint that annotates each whitelist rule that applies, and removes the ones that don't.

foolip:
@jgraham, locally, I didn't run the lint at all, I'm trusting Travis to run the whole lint and tell me if I got it wrong. Is this a safe assumption?

jgraham:
Seems like we are now correctly accounting for the case of changing the lint itself, so yes.

PR #8028, opened 2017-11-01
Fix some setTimeout lint errors in html/

foolip:
same-dispatch-time.html is the most interesting, because it was
previously broken. As a synchronous test, it would pass regardless of
what happened in the callbacks. Fix that.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8028)

### DIFF
--- a/html/webappapis/animation-frames/same-dispatch-time.html
+++ b/html/webappapis/animation-frames/same-dispatch-time.html
@@ -9,7 +9,7 @@
   <body>
     <div id="log"></div>
     <script>
-      test(function (t) {
+      async_test(function (t) {
         var a = 0, b = 0;
 
         /* REASONING:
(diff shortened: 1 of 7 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20075)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/295600765)

jdm (approved):

foolip:
w3c-test:mirror

html/webappapis/animation-frames/same-dispatch-time.html
```diff
@@ -24,7 +24,7 @@
           assert_not_equals(b, 0);
           assert_equals(a, b);
           t.done();
-        });
+        }));
```
foolip: @lukebjerring here's another thing that I noticed only because I went hunting through the results to sniff out problem. I'll stop bugging with with every instance where I want the new thing, but this comes up every few days for me, and we'd just never accept being blind like this in any proper software project.

PR #8033, opened 2017-11-01
Add support for Opera product

jensl:
The approach taken is pretty much to copy everything that exists for
Chrome, and then adjust as seemed reasonable.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8033)

### DIFF
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -2,6 +2,7 @@ import logging
 import os
 import platform
 import re
+import shutil
 import stat
 from abc import ABCMeta, abstractmethod
 from ConfigParser import RawConfigParser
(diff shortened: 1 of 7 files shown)


jensl:
Note: when I run this, I have to explicitly specify a browser binary using e.g. `--binary /usr/bin/opera`. I don't know if I've missed something, or if this is a problem with `operadriver`.

Either way, this patch makes it easier to test in Opera than before, at least.

gsnedders:
@jensl my guess would be missing some `find_executable` call, though I don't see how it manages to work for Chrome at the moment

jensl:
The reason I thought it might be a problem with `operadriver` was that its [desktop example](https://github.com/operasoftware/operachromiumdriver/blob/master/examples/desktop.py) produced exactly the same error.

jgraham:
Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[tools/wpt/browser.py, line 261 at r1](https://reviewable.io:443/reviews/w3c/web-platform-tests/8033#-Kxskvgr_tTEo9NW9mrq:-Kxskvgr_tTEo9NW9mrr:b-3p3bwm) ([raw file](https://github.com/w3c/web-platform-tests/blob/47bf27f021e2ec4861744b9c1d7f0c41b86f2ff1/tools/wpt/browser.py#L261)):*
> ```Python
>         return re.search(r"[0-9\.]+( [a-z]+)?$", output.strip()).group(0)
> 
>     def prepare_environment(self):
> ```

Did you mean to add this part?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/web-platform-tests/8033)*

jgraham:
Or maybe you're just based on an old revision and reviewable is confused? I can't see the changes in GH anyway.

Finding the browser in the case of webdriver-backed implementations is  usually the responsibility of WebDriver so I think that's not our bug.

jgraham (approved):

jensl:
@jgraham Diff confusion because of copied code. The end of the class before is identical with the end of the class I added, so the diff can claim that I added either class ending. I seem to remember at least considering to implement some kind of workaround for the phenomenon in Critic at some point. :-)

PR #8045, opened 2017-11-02
Enumerate fewer cross-origin properties

annevk:
See https://github.com/whatwg/html/pull/3186 for context.

### DIFF
--- a/html/browsers/origin/cross-origin-objects/cross-origin-objects.html
+++ b/html/browsers/origin/cross-origin-objects/cross-origin-objects.html
@@ -287,13 +287,12 @@ addTest(function() {
                       whitelistedWindowPropNames,
                       "Object.getOwnPropertyNames() gives the right answer for cross-origin Window");
   assert_array_equals(Object.keys(C).sort(),
-                      whitelistedWindowPropNames,
+                      whitelistedWindowIndices,
                       "Object.keys() gives the right answer for cross-origin Window");
   assert_array_equals(Object.getOwnPropertyNames(C.location).sort(),
                       whitelistedLocationPropNames,
                       "Object.getOwnPropertyNames() gives the right answer for cross-origin Location");
-  assert_array_equals(Object.keys(C.location).sort(),
-                      whitelistedLocationPropNames,
+  assert_equals(Object.keys(C.location).length, 0,
                       "Object.keys() gives the right answer for cross-origin Location");
 }, "[[OwnPropertyKeys]] should return all properties from cross-origin objects");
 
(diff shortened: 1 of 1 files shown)


annevk:
So the one thing I don't get is that in https://bug1412741.bmoattachments.org/attachment.cgi?id=8924054 @bzbarsky also marked the [[OwnPropertyKeys]] test as having changed results. I don't see how that follows from the standard.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20109)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/296242747)

bzbarsky:
> also marked the [[OwnPropertyKeys]] test as having changed results

The test in question is http://searchfox.org/mozilla-central/rev/423b2522c48e1d654e30ffc337164d677f934ec3/testing/web-platform/tests/html/browsers/origin/cross-origin-objects/cross-origin-objects.html#283-296 (or https://github.com/w3c/web-platform-tests/blob/26cc7b48a87800a0e28ea765e3630b1c929cf1c9/html/browsers/origin/cross-origin-objects/cross-origin-objects.html#L283-L296 if you prefer).  It tests the two obvious spec entrypoints that exercise `[[OwnPropertyKeys]]`: `Object.getOwnPropertyNames` and `Object.keys`.  The former is not affected by enumerability, but the latter is.

I didn't want to just change the test before the spec changes, so I just marked it as failing for now, but of course we'll want to update the test...

annevk:
Thanks, I'll fix that one too. I suspect it would be better for @cdumez et al if we fixed it here straight away given that there's still some (significant) delay if we land web-platform-tests changes from the Firefox side.

annevk:
Pushed a fix.

html/browsers/origin/cross-origin-objects/cross-origin-objects.html
```diff
@@ -285,13 +287,12 @@
...
-                      whitelistedWindowPropNames,
+                      whitelistedWindowIndices,
                       "Object.keys() gives the right answer for cross-origin Window");
   assert_array_equals(Object.getOwnPropertyNames(C.location).sort(),
                       whitelistedLocationPropNames,
                       "Object.getOwnPropertyNames() gives the right answer for cross-origin Location");
-  assert_array_equals(Object.keys(C.location).sort(),
-                      whitelistedLocationPropNames,
+  assert_equals(Object.keys(C.location).length, 0,
                       "Object.keys() gives the right answer for cross-origin Location");
```
bzbarsky: Fix the indent here?

bzbarsky (approved):
Looks good, thank you!

PR #8069, opened 2017-11-05
Ensure that tests-affected picks up changes to webidl/idlharness

jgraham:
When checking the tests that are affected by a change we previously
had a rule to ignore all files under reources, so that changes to
testharness.js don't cause the entire repository to be rerun. However
changes to WebIDL2 or idlharness.js probably should cause all
interfaces.html tests to be rerun. That requires two changes:

* Support for the rewrite rule that causes /resources/WebIDLParser.js
  to be interpreted as the resources/webidl2/lib/webidl2.js path

* Support for more precise exclude rules so that we can exclude
  resources/testharness.js without excluding all of the WebIDL stuff.

For the latter --ignore-dirs has been replaced by --ignore-rules,
where the rules can include a * to match anything other than a path
separator or ** to match anything. In either case the patterns are
only valid at the end of a path segment (i.e. before a seperator or
the end of the path).

Further changes may be required if we want to allow changes to webidl2
that don't affect the webidl2.js file itself to be seen as affecting
all idlharness tests.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8069)

### DIFF
--- a/tools/wpt/testfiles.py
+++ b/tools/wpt/testfiles.py
@@ -2,6 +2,7 @@ import argparse
 import itertools
 import logging
 import os
+import re
 import subprocess
 import sys
 
@@ -79,13 +80,21 @@ def branch_point():
     return branch_point
 
 
-def files_changed(revish, ignore_dirs=None, include_uncommitted=False, include_new=False):
-    """Get and return files changed since current branch diverged from master,
-    excluding those that are located within any directory specifed by
-    `ignore_changes`."""
-    if ignore_dirs is None:
-        ignore_dirs = []
+def compile_ignore_rule(rule):
+    rule = rule.replace(os.path.sep, "/")
+    parts = rule.split("/")
+    re_parts = []
+    for part in parts:
+        if part.endswith("**"):
+            re_parts.append(re.escape(part[:-2]) + ".*")
+        elif part.endswith("*"):
+            re_parts.append(re.escape(part[:-1]) + "[^/]*")
+        else:
+            re_parts.append(re.escape(part))
+    return re.compile("^%s$" % "/".join(re_parts))
 
+
+def repo_files_changed(revish, include_uncommitted=False, include_new=False):
     git = get_git_cmd(wpt_root)
     files = git("diff", "--name-only", "-z", revish).split("\0")
     assert not files[-1]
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PASSED

Started: 2017-11-05 05:02:43
Finished: 2017-11-05 05:05:18


<h2>Failing Jobs</h2>
<ul>

<li>chrome:unstable</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/20182)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/297442424)

foolip:
Typo: reources

foolip:
Typo: seperator

tools/wpt/testfiles.py
```diff
@@ -107,27 +116,50 @@ def files_changed(revish, ignore_dirs=None, include_uncommitted=False, include_n
...
         else:
             changed.append(fullpath)
 
     return changed, ignored
 
 
+def files_changed(revish, ignore_rules=None, include_uncommitted=False, include_new=False):
+    """Get and return files changed since current branch diverged from master,
+    excluding those that are located within any directory specifed by
+    `ignore_changes`."""
```
foolip: There isn't anything called `ignore_changes`.

tools/wpt/testfiles.py
```diff
@@ -79,13 +80,21 @@ def branch_point():
...
-    excluding those that are located within any directory specifed by
-    `ignore_changes`."""
-    if ignore_dirs is None:
-        ignore_dirs = []
+def compile_ignore_rule(rule):
+    rule = rule.replace(os.path.sep, "/")
+    parts = rule.split("/")
+    re_parts = []
+    for part in parts:
+        if part.endswith("**"):
```
foolip: Parsing out what the body of this loop does isn't easy, for me. Maybe a test for this that hits each of the branches? (Don't need to apply the rules, just stringify the resulting regexes and asssert that they are what they should be.)

tools/wpt/tests/test_wpt.py
```diff
@@ -91,6 +91,14 @@ def test_files_changed(capsys):
     assert err == ""
 
 
+def test_files_changed_ignore(capsys):
+    from wpt.testfiles import exclude_ignored
+    files = ["resources/testharness.js", "resources/webidl2/index.js", "test/test.js"]
```
foolip: Having webidl2.js show up in a test would be nice, since that was the thing that originally didn't work.

jgraham: That's not entirely trivial to test since it only affects `tests_affected` rather than `get_files_changed` and that's mostly about reading files to see what data they contain, so it would either need to be refactored or we'd need to find some commit with a webidl2 change to test, but there aren't many of those. Maybe land the upgrade and use that commit for the test.

foolip: Would a previous commit that changes webidl2.js do the job? 960a3d21bab0293630da8919847f87f4af3a3198?

tools/wpt/tests/test_wpt.py
```diff
@@ -91,6 +91,24 @@ def test_files_changed(capsys):
...
+    changed, ignored = exclude_ignored(files, ignore_rules=["resources/testharness*"])
+    assert changed == [os.path.join(wpt.wpt_root, item) for item in
+                       ["resources/webidl2/index.js", "test/test.js"]]
+    assert ignored == [os.path.join(wpt.wpt_root, item) for item in
+                       ["resources/testharness.js"]]
+
+
+def test_files_changed_ignore_rules():
+    from tools.wpt.testfiles import compile_ignore_rule
+    assert compile_ignore_rule("foo*bar*/baz").pattern == "^foo\*bar[^/]*/baz$"
```
foolip: Is this really right? Why is `\*` part of the pattern here or anywhere, and not `.*`?

tools/wpt/testfiles.py
foolip: Tests now in `test_files_changed_ignore_rules()`

tools/wpt/tests/test_wpt.py
jgraham: It's just doing prefix matching on path parts. It could do something different, but this is enough to satisfy the current use cases.

foolip: Oh, the comment "* before a separator or the end matching" explains this.

foolip (approved):

PR #8075, opened 2017-11-06
Add tests for CEReactions on HTML Media Capture

Honry:
And introduce a common function to create customized build in element

### DIFF
--- a/custom-elements/reactions/HTMLInputElement.html
+++ b/custom-elements/reactions/HTMLInputElement.html
@@ -1,0 +1,38 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Custom Elements: CEReactions on HTMLInputElement interface</title>
+<link rel="author" title="Intel" href="http://www.intel.com">
+<link rel="author" title="Wanming Lin" href="mailto:wanming.lin@intel.com">
+<meta name="assert" content="capture of HTMLInputElement interface must have CEReactions">
+<meta name="help" content="https://www.w3.org/TR/html-media-capture/#the-capture-attribute">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="../resources/custom-elements-helpers.js"></script>
+<script src="./resources/reactions.js"></script>
+<body>
+<script>
+test(() => {
+    const element = define_build_in_custom_element(['capture'], HTMLInputElement, 'input');
+    const instance = document.createElement('input', { is: element.name });
+
+    assert_array_equals(element.takeLog().types(), ['constructed']);
+    instance['capture'] = 'user';
+    const logEntries = element.takeLog();
+    assert_array_equals(logEntries.types(), ['attributeChanged']);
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20194)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/297822589)

domenic (approved):
LGTM, although there are two additional useful tests:

- Setting the property to "asdf" causes an attributeChanged with new value "asdf"
- Setting the property to null causes an attributeChanged with the new value null, i.e. removes the attribute.

Honry:
@domenic, thanks for your review and suggestion.

>Setting the property to "asdf" causes an attributeChanged with new value "asdf"

"asdf" is a kind of [ invalid value default ](https://www.w3.org/TR/html51/infrastructure.html#invalid-value-default), useful test. 👍 
>Setting the property to null causes an attributeChanged with the new value null, i.e. removes the attribute.

Setting the property to null doesn't mean removing the attribute for 'capture' attribute, we can setting [missing value default](https://www.w3.org/TR/html51/infrastructure.html#missing-value-default), i.e. empty string. WDYT?

domenic:
Setting it to null should remove the attribute. The spec says it reflects, and per https://html.spec.whatwg.org/#reflect,

> If a reflecting IDL attribute is a nullable DOMString attribute whose content attribute is an enumerated attribute, then, on getting, if the corresponding content attribute is in its missing value default then the IDL attribute must return null, otherwise, the IDL attribute must return the conforming value associated with the state the attribute is in (in its canonical case). **On setting, if the new value is null, the content attribute must be removed**, and otherwise, the content attribute must be set to the specified new value.

Honry:
But the "capture" is not a **nullable DOMString** attribute.
>partial interface HTMLInputElement {
    [CEReactions]
    attribute DOMString capture;
};

I think following piece of spec should be reflected: 
>If a reflecting IDL attribute is a DOMString attribute whose content attribute is an enumerated attribute, and the IDL attribute is limited to only known values, then, on getting, the IDL attribute must return the conforming value associated with the state the attribute is in (in its canonical case), if any, or the empty string if the attribute is in a state that has no associated keyword value or if the attribute is not in a defined state (e.g. the attribute is missing and there is no missing value default). On setting, the content attribute must be set to the specified new value.

domenic:
Hmm, so what's confusing is the spec doesn't say it's limited to known values. So it actually falls into

> If a reflecting IDL attribute is a DOMString or USVString attribute but doesn't fall into any of the above categories, then the getting and setting must be done in a transparent, case-preserving manner.

In that case setting it to null should cause the attribute to change to the string `"null"`.

(The invalid value default then causes this to behave as if it's in the implementation-specific state. But it should still change the attribute to the string `"null"`)

domenic:
Anyway, in that case we're basically testing the reflection, which can be covered by other tests, not the CEReactions tests. So the tests LGTM to merge as-is.

Honry:
@domenic, thanks! Merging it.

PR #8089, opened 2017-11-07
Clean up click-cancel.html a bit

foolip:
As the first testdriver.js test, people might look. So:
 * Remove the manual instructions (which were wrong anyway)
 * Don't use a variable called done, that's confusing
 * Make the test fail fast if test_driver.click() rejects
 * Make the indentation a bit more reasonable

### DIFF
--- a/uievents/order-of-events/mouse-events/click-cancel.html
+++ b/uievents/order-of-events/mouse-events/click-cancel.html
@@ -10,36 +10,28 @@
     <script src="/resources/testdriver-vendor.js"></script>
   </head>
   <body>
-    <ol>
-      <li>Check the following item.</li>
-      <li>Hover over the button.</li>
-    </ol>
     <input type="checkbox" id="checkbox">Check me!<br>
-    <button type="button" id="done">Button</button>
-</body>
-<script>
-var t = async_test("Default event is canceled if the click event is canceled");
+    <button type="button" id="button">Button</button>
+    <script>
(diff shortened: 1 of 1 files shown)


foolip:
I may well have broken the test here. Even if stability checks pass, need to check that the test is actually passing too.

w3c-bots:
# Build PASSED

Started: 2017-11-07 07:51:03
Finished: 2017-11-07 08:02:31


<h2>Failing Jobs</h2>
<ul>

<li>build_css</li>

<li>chrome:unstable</li>

<li>safari:10.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/20251)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/298411320)

jdm (approved):

gsnedders:
Someone had suggested using a `promise_test` for this but I don't know if that works here?

foolip:
> Someone had suggested using a `promise_test` for this but I don't know if that works here?

This is actually rather a big problem, I spend maybe 30 minutes trying to decide between `async_test` and `promise_test` before concluding that all options would suck. With `async_test` we need to catch promise rejections, and propagating the error message is quite some work.

And with `promise_test`, after the second click, a promise would have be created which resolved in an event handler later, in a way that'd look quite ugly.

When we've written some more ugly tests, we should probably fix this somehow.

PR #8091, opened 2017-11-07
streams/generate-test-wrappers.js: suggest --harmony

ricea:
Add --harmony to the suggested command-line for generate-test-wrappers.js so
that it works for nodejs v6 users.


domenic:
We shouldn't do this, but instead recommend Node 8. --harmony features are unstable and may have semantic differences.

ricea:
It would be hypocritical of me to tell people to use node 8 when I'm on node 6 :smile:

domenic:
Upgrade? 6 is barely supported now that 8 is the new LTS and 9 the new stable.

ricea:
I just use whatever is in the Chrome tree. It saves me having to upgrade manually.

I'm withdrawing this PR.

PR #8103, opened 2017-11-08
Inline the Hash Function Textual Names registry for RTCCertficate test

foolip:
Follows https://github.com/w3c/webrtc-pc/issues/1663.

### DIFF
--- a/webrtc/RTCCertificate.html
+++ b/webrtc/RTCCertificate.html
@@ -172,8 +172,9 @@
         assert_equals(typeof fingerprint, 'object',
           'Expect fingerprint to be an object (dictionary)');
 
-        // Can only do simple test as the allowed values may be extended
-        assert_true(/^[a-zA-Z\-]+$/.test(fingerprint.algorithm),
+        // https://www.iana.org/assignments/hash-function-text-names/hash-function-text-names.xml
+        const algorithms = ['md2', 'md5', 'sha-1', 'sha-224', 'sha-256', 'sha-384', 'sha-512'];
+        assert_in_array(fingerprint.algorithm, algorithms,
           'Expect fingerprint.algorithm to be string of algorithm identifier');
 
         assert_true(/^([0-9a-f]{2}\:)+[0-9a-f]{2}$/.test(fingerprint.value),


foolip:
@soareschen, this test was previously wrong, because the regex didn't allow for all of the values that should be used. So I just enumerated all of them, the test can be updated if any implementation ever fails because of new algorithms being supported.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20326)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/298971516)

webrtc/RTCCertificate.html
```diff
@@ -172,8 +172,9 @@
         assert_equals(typeof fingerprint, 'object',
           'Expect fingerprint to be an object (dictionary)');
 
-        // Can only do simple test as the allowed values may be extended
-        assert_true(/^[a-zA-Z\-]+$/.test(fingerprint.algorithm),
+        // https://www.iana.org/assignments/hash-function-text-names/hash-function-text-names.xml
+        const algorithms = ['md2', 'md5', 'sha-1', 'sha-224', 'sha-256', 'sha-384', 'sha-512'];
+        assert_in_array(fingerprint.algorithm, algorithms,
```
henbos: The spec says "The browser selects the algorithm used to sign the certificate; a browser SHOULD select SHA-256 [FIPS-180-4] if a hash algorithm is needed."

Unless there is a known browser that does not use sha-256 maybe consider asserting that it is "sha-256", since this is `SHOULD`. Or do you think we should accept any of the known ones?

Also the spec's example includes `hash: "SHA-256"` with capital letters. Can you look up if it is defined if it should be upper or lower case or if both are acceptable?

soareschen (approved):

soareschen:
LGTM. Does this mean that when the IANA registry is updated we will then update the test accordingly?

foolip:
> LGTM. Does this mean that when the IANA registry is updated we will then update the test accordingly?

Yes, that's what we should do.

webrtc/RTCCertificate.html
foolip: @henbos, do you think any changes to this PR are needed for this? Can you comment on the spec side?

snuggs:
@foolip LGTM and regarding IANA update.

snuggs (commented):
LGTM @foolip. And great rec for futureproof.

PR #8129, opened 2017-11-09
Fetch: adjust Content-Length value in cors-filtering.js

### LINKED ISSUE #6471: CORS test incorrectly setting Content-Length
In https://github.com/w3c/web-platform-tests/blob/d6102e6097670a68abae60c6c24cc8060d9ac379/fetch/api/cors/cors-filtering.js#L59 there's a test:
```js
corsFilter(url, "Content-Length", "0" , true);
```
which asks the HTTP server to set the the `Content-Length` header to be `0`. Since the content is `top`, this causes `top` to be interpreted as the beginning of the HTTP response to the next HTTP request, so the next test to run on that connection fails. This causes nondeterministic test failure, as seen here: https://github.com/servo/servo/pull/16508#issuecomment-312979555

annevk:
The exact value is not important for this test so it might as well be correct.

Fixes #6471.

### DIFF
--- a/fetch/api/cors/cors-filtering.js
+++ b/fetch/api/cors/cors-filtering.js
@@ -56,14 +56,14 @@ corsFilter(url, "Pragma", "no-cache", false);
 corsFilter(url, "Age", "27", true);
 corsFilter(url, "Server", "wptServe" , true);
 corsFilter(url, "Warning", "Mind the gap" , true);
-corsFilter(url, "Content-Length", "0" , true);
+corsFilter(url, "Content-Length", "3" , true); // top.txt contains "top"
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20403)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/299594858)

jdm (approved):

jugglinmike:
Looks like I was a bit too slow on the review here. Sorry about that!

annevk:
I don't think so. @jdm was just surprisingly quick. 😊

PR #8147, opened 2017-11-10
Fix all non-zero delays in transform stream tests

ricea:
Some tests for TransformStream used the delay() function with non-zero
delays. Mostly this was done to make things happen asynchronously in a
particular order, and has been replaced with equivalent code using
promise chains.

The test "TransformStream flush is called after all queued writes
finish, once the writable is closed" was passing accidentally as it was
using test() where it should have been using promise_test(). In addition
to removing non-zero delays from the test, also fix it to use
promise_test() and remove the backpressure that was causing transform()
to never be called.

See https://github.com/whatwg/streams/issues/548 for background on this
clean-up.

### DIFF
--- a/streams/transform-streams/flush.js
+++ b/streams/transform-streams/flush.js
@@ -19,7 +19,6 @@ promise_test(() => {
   });
 }, 'TransformStream flush is called immediately when the writable is closed, if no writes are queued');
 
-// TODO
 promise_test(() => {
   let flushCalled = false;
   let resolveTransform;
--- a/streams/transform-streams/flush.js
+++ b/streams/transform-streams/flush.js
@@ -20,17 +20,20 @@ promise_test(() => {
 }, 'TransformStream flush is called immediately when the writable is closed, if no writes are queued');
 
 // TODO
-test(() => {
+promise_test(() => {
   let flushCalled = false;
+  let resolveTransform;
   const ts = new TransformStream({
     transform() {
-      return delay(10);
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20478)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/300190046)

streams/transform-streams/flush.js
```diff
@@ -20,17 +20,20 @@ promise_test(() => {
 }, 'TransformStream flush is called immediately when the writable is closed, if no writes are queued');
 
 // TODO
```
domenic: Any idea what this TODO comment is doing here?

domenic (approved):
Really nice fixes. Approving with an opportunity for removing a stray TODO that I'll leave to you.

streams/transform-streams/flush.js
ricea: I was trying to ignore it.

It turns out it was added by @marvinhagemeister a year ago when he ported the tests to wpt format in https://github.com/whatwg/streams/pull/531. I can find nothing in that test or the surrounding ones that needs doing, so I have removed it.

Marvin, if you remember what needed doing here, please speak up!

marvinhagemeister: Just went through my original PRI honestly can't remember why I put it there :/

PR #8150, opened 2017-11-10
Fix #8122: Upstream Mozilla counter-style tests

### LINKED ISSUE #8122: Move CSS Counter Styles reftests over from mozilla-central
See tests not beginning with t1202 in http://searchfox.org/mozilla-central/source/layout/reftests/counters

Note these don't cover all Counter Styles but they are reftests unlike the current visual tests.

gsnedders:
These don't cover all of the counter styles, but they give us automated tests for most of them.

They are all originally authored by @dbaron and @upsuper.

Fixes #8122.


gsnedders:
I fear these might be larger than our 600x600 or 800x600 viewport, though?

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20493)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/300342665)

upsuper:
So, it shouldn't be larger than Mozilla's reftest limit which is 800*1000. I think majority of the tests wouldn't be larger than 800*600 either becaise many of them are written intentionally use inline layout rather than using block. 

What I'm concerned is that there are tests in them checking some impl-dependent behavior. I had a list of tests which mix impl-dependent and -independent behavior, and wanted to split them properly, but never really got time to do so.

upsuper:
Oh, and thanks for doing this!

gsnedders:
@upsuper My quick glance implied that the impl-specific stuff (mostly checking rounding around 2^31-1) was in self-contained files, but plenty of these tests contain tests for values well above 2^16 so they practically rely on having at least 32-bit ints. I think we rely on 32-bit ints being used in enough places in the testsuite we should just assume it generally (maybe even just add the rounding tests on the basis we're assuming that).

gsnedders:
The lint failures appear to all be MISSING-LINK failures, as we might expect.

upsuper:
I just had a brief look at the tests being imported, and it seems these tests aren't partocularly impl-dependent. There is one area, though, which is techniquely undefined at the moment from the spec's point of view: result of numbers over 10k in longhand east asian counter styles.

We had discussed about details of the algorithm for a long time and @tabatkins eventually decided to make it undefined and allow impls to explore. I'm confident to some extent that the algorithm I implemented is correct for at least Chinese and Japanese, but that is technically not part of the spec text.

upsuper:
And... if we want to include the tests with numbers over 10k for those styles, we probably want experts of CJK languages to check with the result in the testcases, and if nothing seems to be wrong, we should probably put the algorithm I wrote into the spec so that spec and testcases here match each other.

gsnedders:
#8122 was fixed by @upsuper separately

r12a:
I very much wanted to automate the (many) tests i contributed for counter-styles, but i couldn't find a way to control the spacing around the counters in lists so that the ref test matched for all browsers.  Not sure how these tests get around that.

r12a:
I just ran one  css/css-counter-styles/counter-suffix.html  in jsbin, and seemed ok for Firefox and Chrome, but the ref failed for example for Latin suffixes in Safari even though the suffixes were ok.

gsnedders:
Closing per my above comment.

@r12a See e.g. https://hg.mozilla.org/mozilla-central/file/tip/layout/reftests/counters/counter-cjk-decimal.html from mozilla-central. Doing it with list-items is far more problematic, but we don't really need to test the entire range with lists, merely that the counter keyword is accepted as a list-item-type.

r12a:
@gsnedders not sure whether we're referring to the same thing. My concern was that if you only have two items in the test, the ref file doesn't represent pixel-by-pixel what a successful test looks like on Safari. So the test apparently fails on Safari when actually all was working as expected. It's because Safari doesn't use the same white space around counters as Firefox and Chrome (which is irrelevant for this test).

In terms of the number of items,  i agree that my tests were probably more verbose than needed, although with non-decimal list styles it's probably important to test enough values to ensure that the algorithm is working at certain key points (not many). So my comment wasn't related at all to the number of subtests on the page.

Does that clarify things, or am i misunderstanding?  (Reopening for now, just so this doesn't get lost.)

gsnedders:
@r12a My point was merely that the majority of counter style tests don't need to use lists (which have relatively poor interop, for reasons unrelated to counters themselves), and thus can avoid many of those issues.

With regards to counter-suffix.html specifically: the spec does actually define what white-space should be used, though kerning does differ between browsers (ergh!). One would hope Safari might fix its differences here (or fix the test!) to get it to pass.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24817)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/345334374)

svgeesus:
So, Travis CI is failing because the repo moved (possibly other reasons too, but that one doesn't help).

svgeesus (approved):
Don't see a particular need for the mix of html5, html401 and xhtml but whatever
Actual tests look fine to me
Hope an approving review can get these moving, worried about the travis stuff

svgeesus:
> Required statuses must pass before merging

@gsnedders magic wand required

gsnedders:
https://travis-ci.org/web-platform-tests/wpt/builds/345334374?utm_source=github_status&utm_medium=notification is the old Travis run (just rewrite w3c/web-platform-tests into web-platform-tests/wpt in the URL).

r12a:
> I just ran one css/css-counter-styles/counter-suffix.html in jsbin, and seemed ok for Firefox and Chrome, but the ref failed for example for Latin suffixes in Safari even though the suffixes were ok.

This comment was saying that a particular test in this set doesn't produce consistent results across browsers, and the results cannot therefore be tested reliably.  That point seems to have been lost in discussion of generalities - the point of which generalities i have no argument with, but the practical issue remains for the test i mentioned.

nt1m:
cc @vitorroriz in case you're interested in these tests.

nt1m:
@jgraham Hi! I'm trying to upstream these tests, and seems like the wpt-firefox-nightly-stability is the only task preventing merging.

PR #8155, opened 2017-11-10
Implement Send Alert Text tests for WPT wdspec tests

alchiem:
delete trailing whitespace

Deleted unused import. Simplified send alert text test.

Test non string input. Test text with whitespace. Formatting fixes.

Add more tests for invalid input


webdriver/tests/user_prompts/send_alert_text.py
```diff
@@ -0,0 +1,80 @@
...
+    session.url = inline("<p>This is not an editable paragraph.")
+    element = session.find.css("p", all=False)
+    arr = [1, 2, 3, 4, 5]
+    elements = {"null": None,
+                "object": element,
+                "array": arr,
+                "number": 123}
+    session.execute_script("window.result = window.prompt(\"Enter Your Name: \", \"Name\");")
+    for item in elements:
+        body = {"text": elements[item]}
```
andreastt: This iterator is not very Pythonic. Can you change the test to use [`@pytest.parametrize`](https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize)?

This would look roughly something like this:

```py
@pytest.parametrize("text", [None, {}, [], 42, True])
def test_invalid_input(session, text):
    session.execute_script("""window.result = window.prompt("foo")""")
    response = send_alert_text(session, {"text": text})
    assert_error(response, "invalid argument")
```

Note that I haven’t run the above code.

andreastt (changes requested):
Thanks for testing invalid input. The test is unnecessarily complicated and could be simplified using parametrization. I’ve included an example of how to make the test simpler.

Once you make that final change, I think this PR is ready for integration.

andreastt:
Also, there are conflicts which you must resolve.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20542)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/301646538)

andreastt (approved):

Issue #7435, opened 2017-09-21
Be able to send complex pointer-events interactions


gsnedders:
This should be based on https://w3c.github.io/webdriver/webdriver-spec.html#actions and probably expose an API similar to http://searchfox.org/mozilla-central/rev/2ef8bd8a46a02c68ddbb1d5f25fa254dd7be1fbd/testing/marionette/client/marionette_driver/marionette.py#200

cc/ @NavidZ

foolip:
@kereliuk this is still on track for "soon", right? (Doing the rotation.)

foolip:
@kereliuk, was this resolved already?

gsnedders:
Doesn't appear to be anything in testdriver.js for it.

foolip:
There's some WIP in https://github.com/web-platform-tests/wpt/pull/11931

foolip:
@NavidZ @SummerLW should this be assigned to one of you?

NavidZ:
Sure. Keep it assigned to me. I guess I need to add some documentation about the synthetic pointer actions as well. I remember there was a test driver documentation some where around here.

foolip:
Documentation is at https://web-platform-tests.org/writing-tests/testdriver.html

mdittmer:
Friendly ecosystem infra rotation nudge to @NavidZ. Any updates on adding documentation?

NavidZ:
@jgraham I was planning to only explain the builder API directly to make sure people hopefully use that instead of directly calling in action_sequence. Are you okay with that?

jgraham:
Yes, for sure.

lukebjerring:
@NavidZ - any updates?

NavidZ:
I did update the documentation. I guess I did it inline and forgot to mention it in this bug.

Issue #7434, opened 2017-09-21
User initiated keyboard input


gsnedders:
We should be able to send user initiated keyboard input.

@TakayoshiKochi wants this to be able to test `tabindex` around shadow DOM.

foolip:
Assigning to @JKereliuk as he's begun work on this.

JKereliuk:
Are there any existing manual tests that need this?

foolip:
@kojiishi, do you have some manual tests that need a tab key to move focus?

TakayoshiKochi:
I guess you meant @TakayoshiKochi :)
I haven't started yet, but testing tab focus order is what I would like to
do.

2017/11/21 7:17 "Philip Jägenstedt" <notifications@github.com>:

foolip:
Yes I did, my apologies @TakayoshiKochi!

foolip:
@kereliuk this will be resolved by https://github.com/w3c/web-platform-tests/pull/10113, right?

kereliuk:
Yep :)

foolip:
https://github.com/w3c/web-platform-tests/pull/10113 has been merged \o/

PR #8159, opened 2017-11-13
Actions automation

kereliuk:
This is a PR to get the ball rolling about automating more of the selenium capability in testdriver.js.

Want to know what people think about the way we will be extending the [python selenium bindings](http://selenium-python.readthedocs.io/).

I've included

- basic functionality for sending actions in testdriver.js tests
- an example of part of a manual test that can now be automated

AFAIK you cannot add input sources with the current selenium python bindings (It's there just not exposed?) So that would be next so we can take advantage of the W3C actions with multiple inputs.

Still a WIP (no optional parameters and missing error checks) and needs some clean up and documentation.

cc/ @gsnedders @foolip

#7434 
#7435 

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8159)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20506)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/301193563)

pointerevents/pointerevent_capture_mouse.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Action</title>
+    <link rel="help" href="https://w3c.github.io/uievents/#events-clickevent-event-order">
```
NavidZ: This should be removed?

pointerevents/pointerevent_capture_mouse.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
```
NavidZ: I assume this is just a proof of concept test here. Right? Wasn't there a simple test already in pointerevents that only needed a move or something that you could reuse?

JKereliuk: Yep, this is just proof of concept. What is the test you are thinking of?

pointerevents/pointerevent_capture_mouse.html
JKereliuk: Removed

docs/_writing-tests/testdriver.md
```diff
@@ -30,5 +30,8 @@ Note that if the element to be clicked does not have a unique ID, the
 document must not have any DOM mutations made between the function
 being called and the promise settling.
 
+### `testdriver.actions`
+
+http://selenium-python.readthedocs.io/api.html#module-selenium.webdriver.common.action_chains
```
foolip: Can this be written in terms of the WebDriver API surface, to not assume that testdriver is implemented using Selenium?

pointerevents/pointerevent_capture_mouse.html
NavidZ: How about this:
https://github.com/w3c/web-platform-tests/blob/master/pointerevents/pointerevent_element_haspointercapture-manual.html

sideshowbarker (commented):
Only reviewed the docs change, but that at least LGTM

JKereliuk:
@jgraham @gsnedders Do you have any comments for this :)

gsnedders:
At a glance this looks good, will look closer tomorrow.

jgraham:
Sorry, missed this somehow. I'm worried about the way this appears to be tightly coupled to the Selenium API rather than the underlying actions model, but I need to look more closely.

JKereliuk:
@jgraham  I have the same worry. Right now the existing click automation is in terms of the selenium API. I think we should be consistent in either using selenium or using WebDriver directly.
WDYT?

gsnedders:
> Right now the existing click automation is in terms of the selenium API. I think we should be consistent in either using selenium or using WebDriver directly.

In principle it should be possible to implement all of these things not using Selenium or WebDriver without having to implement too overhead (e.g., `click` using `gpuBenchmarking`). It's not accidental that `click` does all of the complex WebDriver-like validation in the non-`_internal` function, making it trivial to implement `_internal.click`. We should definitely expect people will want to implement this without using Selenium or WebDriver (e.g., in content_shell).

JKereliuk:
> In principle it should be possible to implement all of these things not using Selenium or WebDriver without having to implement too overhead (e.g., click using gpuBenchmarking). It's not accidental that click does all of the complex WebDriver-like validation in the non-_internal function, making it trivial to implement _internal.click.

I think it still makes sense to provide WebDriver like capability for automation. Also in this PR we have all of the validation in the non-```_internal``` function. 

> We should definitely expect people will want to implement this without using Selenium or WebDriver (e.g., in content_shell).

Why would people need other implementations of clicks (i.e. for content shell)?

If we agree to use WebDriver for automation in WPT, I think we should either use the protocol directly (send requests directly), or have a wrapper on the existing selenium API.

jgraham:
Sorry I don't have more time to look at this at the moment.

Being able to reimplement (some/all) of the API in terms of browser-internal functions was one of the design requirements, both potentially for performance reasons, but also for functional ones (i.e. WebDriver itself isn't available in all environments).

I am specifically worried about using the Selenium API in this case because it outsources the specification of what various actions mean to the Selenium implementation. That is subject to change, invalidating tests, and also hard to implement if Selenium itself is not being used. For actions it's particularly problematic because the Selenium API is much higher level than the underlying WebDriver primitives.

foolip:
> Why would people need other implementations of clicks (i.e. for content shell)?

@JKereliuk, this is actually what we did, see https://bugs.chromium.org/p/chromium/issues/detail?id=781231

We had the Q3 OKR for "Enable web-platform-tests using WebDriver to run in content_shell" that we decided not to pursue then, but something like that would be required (in all engines) in order to assume WebDriver. So defining the behavior in terms of WebDriver but allowing other implementations seems like the way to go.

kereliuk:
Closing this as I will write a similar PR that uses the WebDriver API directly and not Selenium

PR #8163, opened 2017-11-13
Improve compatibility of re-entrant pipeTo test

ricea:
The 'pipeTo() inside size() should work' test didn't pass on Chromium's
implementation as it performs the read() synchronously inside the call
to pipeTo(). This encounters the race condition described in
https://github.com/whatwg/streams/issues/794.

Although different from the reference implementation, what Chromium's
implementation does here is valid and other implementations may do the
same thing.

As a workaround, queue a second chunk inside the test to trigger the
first one to be read.

### DIFF
--- a/streams/transform-streams/reentrant-strategies.js
+++ b/streams/transform-streams/reentrant-strategies.js
@@ -129,7 +129,9 @@ promise_test(() => {
     }
   }, undefined, {
     size() {
-      pipeToPromise = ts.readable.pipeTo(ws);
+      if (!pipeToPromise) {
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20518)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/301359116)

domenic (approved):
Good to catch this. I agree we should allow pipeTo tests to be more lenient, and accept fixes like this in general.

PR #8166, opened 2017-11-13
Re-entrant strategy tests for ReadableStream

ricea:
Verify the results when calling ReadableStream and related class methods inside
the size() function of the strategy. This results in user code being executed
inside ReadableStreamDefaultControllerEnqueue, which can have surprising
results. No real user code should be doing this, but it is important that
implementations be robust in these cases.

These tests are based closely on the ones in
streams/transform-streams/reentrant-strategies.js, and are somewhat redundant
with them. Since less code is under test here, these tests are easier to
understand and debug. Additional tests cover getReader() and tee().

See https://github.com/whatwg/streams/issues/794 for background of some of the
stranger expectations here.

### DIFF
--- a/streams/readable-streams/reentrant-strategies.dedicatedworker.html
+++ b/streams/readable-streams/reentrant-strategies.dedicatedworker.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>reentrant-strategies.js dedicated worker wrapper file</title>
+
+<script src="/resources/testharness.js"></script>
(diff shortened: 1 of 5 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20522)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/301442629)

domenic (approved):
Great tests!

PR #8167, opened 2017-11-13
Implement wdpec support for Edge

thejohnjansen:
In order to run the wdspec tests in Edge, we need to implement the WdspecExecutor.

Included in this change is a hack to run the tests even though we do not return a valid w3c session ID yet.

@jgraham - I removed the lines where we Printed the session Id for debugging. You mentioned making the change a little cleaner, but I'm not sure if you wanted something different from what I have here.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8167)

### DIFF
--- a/tools/webdriver/webdriver/client.py
+++ b/tools/webdriver/webdriver/client.py
@@ -438,7 +438,7 @@ class Session(object):
             bug:
             https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/14641972
             """
-            if url == "session" and method == "POST" and "sessionId" in response.body:
+            if url == "session" and method == "POST" and "sessionId" in response.body and "sessionId" not in value:
                 value["sessionId"] = response.body["sessionId"]
         else:
             raise error.UnknownErrorException("No 'value' key in response body:\n%s" %
--- a/tools/webdriver/webdriver/client.py
+++ b/tools/webdriver/webdriver/client.py
@@ -432,7 +432,7 @@ class Session(object):
         if "value" in response.body:
             value = response.body["value"]
             """
-            Edge does not yet return the w3c session ID. 
+            Edge does not yet return the w3c session ID.
             We want the tests to run in Edge anyway to help with REC.
             In order to run the tests in Edge, we need to hack around
             bug:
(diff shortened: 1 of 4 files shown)


w3c-bots:
# Build BROKEN

Started: 2017-11-13 18:30:53
Finished: 2017-11-13 18:43:08


<h2>Failing Jobs</h2>
<ul>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/20528)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/301546671)

tools/webdriver/webdriver/client.py
```diff
@@ -431,6 +431,15 @@ def send_command(self, method, url, body=None):
...
         if "value" in response.body:
             value = response.body["value"]
+            """
+            Edge does not yet return the w3c session ID. 
+            We want the tests to run in Edge anyway to help with REC.
+            In order to run the tests in Edge, we need to hack around
+            bug:
+            https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/14641972
+            """
+            if url == "session" and method == "POST" and "sessionId" in response.body:
```
jgraham: Can we add an `and "sessionId" not in value` clause here, and log a warning in this case, please?

jgraham (commented):
Basically looks great; I'd just like to make sure we don't accidentally end up down the compat branch in too many cases and not notice.

tools/webdriver/webdriver/client.py
thejohnjansen: @jgraham I added the additional clause, but I don't see other logging in this file yet. I don't mind adding it, but I'm not sure about the best way to do it.

jgraham: `logger.warning("Got legacy new session response")`

The logger is already imported even though it's not used.

thejohnjansen: I should have mentioned that I tried logger.warning("foo") and logger.debug("foo") but both hit an error:
AttributeError: 'NoneType' object has no attribute 'warning'
Which I think means that the logger was not created successfully.

jgraham (approved):
OK, let's merge this and I'll try to look at the logger thing.

PR #8177, opened 2017-11-14
Some service worker tests rely on a document body which might be null

youennf:
On WebKit side, there is a race so that document.body might be null when with_iframe is called.
This makes the tests fail.
Although additional investigation should be done to see whether body should or should not be null in that case, this change allows making the following two tests more robust.


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20572)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/301970730)

mkruisselbrink (approved):

PR #8180, opened 2017-11-14
Fix service-workers/service-worker/getregistrations.https.html

cdumez:
The test expected the registrations returned by getRegistrations() to be
in a specific order but nothing seems the guarantee a given order in the
specification.

### DIFF
--- a/service-workers/service-worker/getregistrations.https.html
+++ b/service-workers/service-worker/getregistrations.https.html
@@ -61,10 +61,9 @@ promise_test(function(t) {
           return navigator.serviceWorker.getRegistrations();
         })
       .then(function(value) {
-          assert_array_equals(
-            value,
-            registrations,
-            'getRegistrations should resolve with array of registrations.');
+          assert_equals(value.length, registrations.length, "Should return the right number of registrations");
+          for (let registration of registrations)
+              assert_in_array(registration, value, "Should return the right registrations");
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20588)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/302117446)

mkruisselbrink:
I believe the test is actually correct here. The [scope to registration map](https://w3c.github.io/ServiceWorker/#dfn-scope-to-registration-map) is an [ordered map](https://infra.spec.whatwg.org/#ordered-map). And an ordered map seems to store its items in insertion order (and thus getRegistrations should also return the registrations in order of registration).

(there might be some edge cases where it is not immediately obvious what the order would be since different scopes have different job queues, but the Register jobs (which ultimately synchronously do the insertion into the registration map) are all run as tasks. But queued on an undefined task source/event loop, not sure how that's supposed to work. I guess that does in fact leave some situations where different same-origin Register jobs could be running at the same time on different event loops, and thus undefined ordering in what getRegistrations returns. Either way, this test is not one of the cases where the ordering is undefined, since we wait for the first registration to complete before starting the second registration).

cdumez:
Oh, interesting. I completely missed that it was supposed to be an ordered map.

PR #8205, opened 2017-11-15
Organize keyframe tests into sections;

jgraham:
This patch also drops some tests that were added at a point when shorthands were
expanded as part of processing a keyframes object. That is no longer the case
and so these tests should be removed (the equivalent logic is tested by
dom/animation/test/chrome/test_animation_properties.html which includes all
these tests).

MozReview-Commit-ID: BkL3kZsu4KX

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1398038 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8205)

### DIFF
--- a/web-animations/resources/keyframe-tests.js
+++ b/web-animations/resources/keyframe-tests.js
@@ -57,6 +57,9 @@ const keyframe = (offset, props, easing='linear', composite) => {
 };
 
 const gKeyframesTests = [
+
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #8218, opened 2017-11-15
Intermittent pointerevent_touch-action-button-test_touch-manual.html. r=kats.

jgraham:
The test case uses double intervals to make sure there is enough time to wait for the scroll event before continuing the testing. However, it doesn't work while the first timeout is delayed and the second one isn't. Tweaked the test case to start the second timer when the first one timeout.

MozReview-Commit-ID: gvhtIpzauE

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1347689 [ci skip]


jgraham:
Code reviewed upstream.

PR #8224, opened 2017-11-15
Display error when response is unsuccessful.

jgraham:
Instead of the default pytest error message when the assertion
fails, which is rather pointless because we know we did not match
the 200 HTTP status code, display the string representation of
wdclient.error.WebDriverException.

This string representation includes the remote end's stacktrace,
which is infinitely more useful in this context.

MozReview-Commit-ID: 6XUgEEc9Qoj

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1411045 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8224)


jgraham:
Code reviewed upstream.

PR #8227, opened 2017-11-15
Use "is" operator to compare responses.

jgraham:
The "is" operator tests if two objects are the same object, and
does not consider the values of the two objects.

This also gives a better pytest error message should the assertion fails.

MozReview-Commit-ID: GOUek15TUPU

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8227)


jgraham:
Code reviewed upstream.

PR #8229, opened 2017-11-15
Fix prompt handle assertions.

jgraham:
The first assertion was never run because it was passed a tuple
of a response value and a boolean.  It now compares the response
value to the boolean using "is".

The second assertion mistakenly used the "==" (equals) operator
but should use "is" to get a better pytest error message.

MozReview-Commit-ID: HCCr5rExIYu


jgraham:
Code reviewed upstream.

PR #8230, opened 2017-11-15
Use a command helper function.

jgraham:
Instead of making repeated calls to session.transport.send we can
use a helper function like most of the other tests do.  This makes
the test more succinct and easier on the eye.

MozReview-Commit-ID: 689rticvu2d


jgraham:
Code reviewed upstream.

PR #8233, opened 2017-11-15
Add stability checking to wptrunner,

jgraham:
This adds a --verify flag that is compatible with other Mozilla test
harnesses i.e. it runs each test 10 times without restarting and then
runs it 5 times with restarts, and then repeats with chaos mode
enabled.

This uses the code from, and can replace, the |wpt run --stability|
flag from upstream although that has different default behaviour
(running 10 times with restarts). More work is needed to avoid
duplicating all the code, however.

MozReview-Commit-ID: 7oUEwJk7uhZ

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1405141 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8233)

### DIFF
--- a/tools/wptrunner/wptrunner/wptcommandline.py
+++ b/tools/wptrunner/wptrunner/wptcommandline.py
@@ -76,6 +76,9 @@ scheme host and port.""")
     mode_group.add_argument("--list-tests", action="store_true",
                             default=False,
                             help="List all tests that will run")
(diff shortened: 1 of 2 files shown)


jgraham:
Code reviewed upstream.

PR #8256, opened 2017-11-15
Add equality test for webdriver.Session

jgraham:
When comparing two instances of webdriver.Session we want to first
check that there is a current session, then the type of the object
to compare with to make sure the "session_id" attribute is present,
then finally we compare the session IDs.

MozReview-Commit-ID: 6Ch4Uy2MEhB

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8256)


jgraham:
Code reviewed upstream.

PR #8257, opened 2017-11-15
Add equality test for webdriver.Element

jgraham:
When comparing two instances of webdriver.Element we want to first
check the type, to make sure the "id" attribute is present, then
compare the web element reference UUIDs.  These are supposed to be
unique across all browsing contexts.

MozReview-Commit-ID: 68PUBQxPdQ5

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8257)


jgraham:
Code reviewed upstream.

PR #8259, opened 2017-11-15
Unmarshal all responses in WPT WebDriver client

jgraham:
The WPT WebDriver client currently only unmarshals responses for some
commands (notably execute_script, execute_async_script, and find.css).
For the client API we want to unmarshal all response bodies automatically.

This patch moves all JSON serialisation/deserialisation to a new
webdriver.protocol package so that it is not scattered around
the client API.  It introduces specialisations of JSONEncoder and
JSONDecoder that allows web element references to be recognised
and converted to complex webdriver.Element objects.

This change means it is no longer necessary for callers to invoke
webdriver.Session._element to convert the response to a web element
as this will be done automatically on any request- and response
body to webdriver.Sesson.send_command.

An important thing to note is that HTTPWireProtocol.send does not  follow
this behaviour by default.  That is because session.transport.send
is used throughout WebDriver tests in WPT as a way to get the raw
JSON body without having to set up session state manually.

MozReview-Commit-ID: 5UyDAe43Hgf

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8259)

### DIFF
--- a/tools/webdriver/webdriver/client.py
+++ b/tools/webdriver/webdriver/client.py
@@ -1,7 +1,7 @@
-import json
 import urlparse
 
 import error
+import protocol
 import transport
 
 from mozlog import get_default_logger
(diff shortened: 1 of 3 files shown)


jgraham:
Code reviewed upstream.

PR #8260, opened 2017-11-15
Make assert_same_element accept webdriver.Element

jgraham:
Allow assert_same_element to compare web element references (JSON
Objects) with webdriver.Element and vice versa.

Tests will typically look up some element using traditional means
and use that as the trusted comparison when retrieving the same
element using the session.transport.send primitive that returns
the plain JSON Object.

MozReview-Commit-ID: 2DScnviOevb

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8260)


jgraham:
Code reviewed upstream.

PR #8263, opened 2017-11-15
Test element retrieval from XHTML namespace.

jgraham:
Finding elements in the XHTML namespace will fail unless
element.isDOMElement is fixed so that it does not look at a
particular namespaceURI.

It is worth noting that, generally, the WPT element retrieval tests
are pretty awful and the new tests don't exactly match the style of
the preceding tests.  This intended, because the other tests are bad.
This should be addessed individually.

MozReview-Commit-ID: 8I3VwhJirSb

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1410796 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8263)

### DIFF
--- a/webdriver/tests/retrieval/find_element.py
+++ b/webdriver/tests/retrieval/find_element.py
@@ -1,7 +1,6 @@
 import pytest
 
-
-from tests.support.asserts import assert_error, assert_success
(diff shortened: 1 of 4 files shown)


jgraham:
Code reviewed upstream.

PR #8280, opened 2017-11-16
WebSocket handshake has no referrer

annevk:
See https://github.com/whatwg/fetch/pull/632.

### DIFF
--- a/websockets/handlers/referrer_wsh.py
+++ b/websockets/handlers/referrer_wsh.py
@@ -7,6 +7,6 @@ def web_socket_do_extra_handshake(request):
 
 def web_socket_transfer_data(request):
     referrer = request.headers_in.get("Referer")
-    if not referrer:
+    if referrer is None:
         referrer = "MISSING AS PER FETCH"
     msgutil.send_message(request, referrer)
--- a/websockets/handlers/referrer_wsh.py
+++ b/websockets/handlers/referrer_wsh.py
@@ -1,0 +1,12 @@
+#!/usr/bin/python
+
+from mod_pywebsocket import msgutil
+
+def web_socket_do_extra_handshake(request):
+    pass
+
+def web_socket_transfer_data(request):
+    referrer = request.headers_in.get("Referer")
+    if not referrer:
+        referrer = "MISSING AS PER FETCH"
+    msgutil.send_message(request, referrer)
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20750)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/303059234)

websockets/referrer.any.js
```diff
@@ -0,0 +1,9 @@
+// META: script=constants.js?pipe=sub
+
+async_test(t => {
+  const ws = new WebSocket(SCHEME_DOMAIN_PORT + "/referrer");
+  ws.onmessage = t.step_func_done(e => {
+    assert_equals(e.data, "MISSING AS PER FETCH");
+    ws.close();
+  });
```
zcorpan: You can do this to avoid timeouts
```
  ws.onclose = t.unreached_func('close');
  ws.onerror = t.unreached_func('error');
```
Locally I get an error event with chrome canary running referrer.any.html. I haven't debugged why that is. Can you reproduce though?

websockets/handlers/referrer_wsh.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/python
+
+from mod_pywebsocket import msgutil
+
+def web_socket_do_extra_handshake(request):
+    pass
+
+def web_socket_transfer_data(request):
+    referrer = request.headers_in.get("Referer")
+    if not referrer:
```
zcorpan: I think this should be `if referrer is None`, at least other handlers use that form and I vaguely recall changing to more explicit `None` checks for some reason.

websockets/referrer.any.js
annevk: I cannot, Chrome 64.0.3271.0 on macOS.

zcorpan: Ah I see what I did wrong. I didn't restart wptserve after checking out your branch, so the websocket handler `/referrer` didn't exist. Sorry. Now it passes for me locally.

zcorpan (approved):

websockets/referrer.any.js
annevk: Ah yeah, took a while to figure that one out too.

tyoshino (commented):
lgtm

PR #3252, opened 2016-06-29
Split WebCryptoAPI generateKey tests into one test file per algorithm name

jgraham:
MozReview-Commit-ID: FVHQoZDV5eI

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1279799

PR #8284, opened 2017-11-16
Fix WebCryptoAPI tests

Hexcles:
This PR fixes a couple of issues in WebCryptoAPI tests:

* `crypto.subtle` now requires to run in secure context. #5730 added the `.https` filename tag to all the HTML tests, but left worker tests (`*.worker.js`) unchanged. In fact, the right way to do this is to change the generation script (`tools/generate.py`) and re-generate all tests.
* #3252 splitted a few large test files into smaller ones. However, the original, unsplit versions were kept around. They are redundant and should be removed.


engelke (approved):
Looks good to me.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20766)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/303236934)

jimsch:
@engelke  Do we still have a test someplace for failure on unsecure now?

engelke:
Not to my knowledge. There was one a long time ago that might be able to
resurrect.

Hexcles:
@jimsch There's an [`idlharness.html`](https://github.com/w3c/web-platform-tests/blob/master/WebCryptoAPI/idlharness.html) that's still running in HTTP, which shows `crypto.subtle` doesn't exist and fails a bunch of subtests. I can't find any tests that directly test the failures on unsecure.

If that's something we want to test, it'd be better to add a few new tests and explicitly assert the failures so that the tests "PASS" when the API in question doesn't work on unsecure. Regardless, we should have this PR to cover service workers in HTTPS.

jimsch:
@Hexcles  I was not objecting to the change proposed, I was trying to trigger things in the back of my mind about if we did do the specific tests that I was looking for initially or if I should be looking at doing them.

Hexcles:
@jimsch Ahh, sorry. I did understand what you meant, but didn't phrase my reply clearly enough. I was suggesting new tests to explicitly test the failures on unsecure as I checked no such tests currently exists.

I'll land the PR later. Thanks, everyone!

jimsch:
@Hexcles I think that the tests which check that the interface does not exist is probably sufficient.  I am not sure how you would test that the specific function fails if you cannot get the interface.

PR #8304, opened 2017-11-20
wpt serve: normalize the config up front and use only that

foolip:
This should make a fix for https://github.com/w3c/web-platform-tests/issues/8182
more straightforward.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8304)

### DIFF
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -637,9 +637,13 @@ def normalise_config(config, ports):
     for scheme, ports_used in ports.iteritems():
         ports_[scheme] = ports_used
 
-    return {"host": host,
-            "domains": domains,
-            "ports": ports_}
+    # make a (shallow) copy of the config and update that, so that the
+    # normalized config can be used in place of the original one.
+    config_ = config.copy()
+    config_["host"] = host
+    config_["domains"] = domains
+    config_["ports"] = ports_
+    return config_
 
 
 def get_ssl_config(config, external_domains, ssl_environment):
@@ -657,17 +661,15 @@ def start(config, ssl_environment, routes, **kwargs):
     paths = {"doc_root": config["doc_root"],
              "ws_doc_root": config["ws_doc_root"]}
 
-    external_config = normalise_config(config, ports)
-
-    ssl_config = get_ssl_config(config, external_config["domains"].values(), ssl_environment)
+    ssl_config = get_ssl_config(config, config["domains"].values(), ssl_environment)
 
     if config["check_subdomains"]:
         check_subdomains(host, paths, bind_hostname, ssl_config, config["aliases"])
 
-    servers = start_servers(host, ports, paths, routes, bind_hostname, external_config,
+    servers = start_servers(host, ports, paths, routes, bind_hostname, config,
                             ssl_config, **kwargs)
 
-    return external_config, servers
+    return servers
 
 
 def iter_procs(servers):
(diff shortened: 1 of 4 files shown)


w3c-bots:
# Build BROKEN

Started: 2017-11-20 10:51:38
Finished: 2017-11-20 11:05:28


<h2>Failing Jobs</h2>
<ul>

<li>wpt_integration in py27</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/20820)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/304675312)

tools/serve/serve.py
```diff
@@ -778,20 +780,23 @@ def run(**kwargs):
 
     setup_logger(config["log_level"])
 
+    ssl_environment = get_ssl_environment(config)
```
jgraham: Shouldn't this still be used as a context manager?

foolip: I know there's *some* difference because removing the `with` didn't work for one of the other cases last week, but it works here. Does it matter, what kinds of things should one use `with` for?

jgraham: With is used in RAII situations, usually where you have some setup you want to do on entering the block and some resource that you want to be automatically freed at the end of the block.

In this case I think some ssl types generate temporary files we create on `__enter__` and delete on `__exit__`. So those will break. It's a pity it's not more obvious that something can break if you use this without running `__enter__`. Presumably various patterns could support that but we aren't using them here.

foolip: So I guess I should just put back `with` then.

foolip: Done.

foolip:
Ping @jgraham

tools/serve/serve.py
```diff
@@ -657,17 +661,15 @@ def start(config, ssl_environment, routes, **kwargs):
     paths = {"doc_root": config["doc_root"],
              "ws_doc_root": config["ws_doc_root"]}
 
-    external_config = normalise_config(config, ports)
-
-    ssl_config = get_ssl_config(config, external_config["domains"].values(), ssl_environment)
+    ssl_config = get_ssl_config(config, config["domains"].values(), ssl_environment)
```
gsnedders: Somehow CI is failing with this throwing `KeyError: 'domains'`. I'm confused as to how that's possible, but I haven't dug into it.

Hexcles: Well, there are more than one code paths to serve.start(), all of which need to be changed if we now want to make start() take a normalized config.

The main function of serve.py is only one of the paths. Another path I know is via wptrunner, in particular `wptrunner/environment.py` when entering TestEnvironment. https://github.com/w3c/web-platform-tests/blob/0e5650daa71f750cdbdee6546b0b56b01c24128b/tools/wptrunner/wptrunner/environment.py#L107

P.S. A few lines above, still in start method, there's a `domains = get_subdomains(host)` which I think is no longer used. (GitHub wouldn't let me comment on just a few lines from the patch...)

foolip:
@gsnedders can I assign this to you to finish, given that you pushed some fixes?

gsnedders:
ping @jgraham; this seems possible to land now

jgraham (approved):
I wonder if will I remember why this was the way it was when we add android support.

jgraham:
This seems to break websockets tests.

PR #8319, opened 2017-11-20
Fix insertion point computation when display: contents pseudos are involved. r=mats,bz

jgraham:
This is a significant rework of how do we compute the insertion point of a
node.

We handle pseudos in the same function instead of out of band, and also recurse
up when the parent has display: contents, which simplifies the code IMO.

MozReview-Commit-ID: 1rSfv1Tq5gO

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1413619 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8319)


jgraham:
Code reviewed upstream.

PR #8321, opened 2017-11-20
Increase timeout for tone events in RTCDTMFSender-helper.js;

jgraham:
With the current value of 150 milliseconds we get intermittent timeouts on our
test infrastructure. We are missing by only a small amount, e.g. 6 milliseconds
in Bug 1415283, but it seems safer to increase the timeout to 250 milliseconds so
that we don't end up revisiting this problem in the future.

This also fixes the error message to say milliseconds rather than seconds.

MozReview-Commit-ID: CqjTKZn33q8

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1415283 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8321)

### DIFF
--- a/webrtc/RTCDTMFSender-helper.js
+++ b/webrtc/RTCDTMFSender-helper.js
@@ -76,8 +76,8 @@ function test_tone_change_events(testFunc, toneChanges, desc) {
         const now = Date.now();
         const duration = now - lastEventTime;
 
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #8322, opened 2017-11-20
Allow skipping web-paltform-tests that timeout,

jgraham:
For local runs it can be helpful to skip tests are expected to time
out. This adds a --skip-timeout flag that will skip tests that are
expected to time out on the current platform.

It doesn't solve the problem that the metadata might be over-specific
in the case that the expectation data isn't universal across all platforms.

MozReview-Commit-ID: 7S4EsVKv33P

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1415674 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8322)


jgraham:
Code reviewed upstream.

PR #8324, opened 2017-11-20
Part5 - Fix cross origin test failure. r=bz,jgraham

jgraham:
Bug 1322517 will make the EME APIs only run on secure context.
This bug will try to make WPT test run with https.
According to the spec https://www.w3.org/TR/secure-contexts/#is-url-trustworthy
I got "TypeError: navigator.requestMediaKeySystemAccess is not a function" by the original code.
So I need this patch to make the test work as expected.

MozReview-Commit-ID: Gp23IcscXHE

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1413427 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8324)

### DIFF
--- a/encrypted-media/scripts/unique-origin.js
+++ b/encrypted-media/scripts/unique-origin.js
@@ -10,7 +10,7 @@ function runTest(config) {
                 resolve(iframe);
             };
             iframe.sandbox = sandbox;
-            iframe.src = src;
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #8326, opened 2017-11-20
Tidy up titles of Web Animations web-platform-tests;

jgraham:
It's not necessary to say "tests" or "tests for" in the title of these tests.
It's obvious from the context that these are test files.

This patch:

* Remove "Tests for" and "tests" from the titles
* Makes the titles match the spec sections they test more closely where
  practical
* Makes the titles more consistent in general (e.g. not putting () after method
  names)
* Updates a few spec links
* Drops a few author annotations since we decided to no longer use them

There are still some inconsistencies:

- The naming of the setTarget.html test file
- The location of the effect-value-visibility.html file
- The getComputedStyle.html test
- etc.

These will be fixed in subsequent patches in this series.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8326)

### DIFF
--- a/web-animations/animation-model/animation-types/accumulation-per-property.html
+++ b/web-animations/animation-model/animation-types/accumulation-per-property.html
@@ -1,6 +1,6 @@
 <!doctype html>
 <meta charset=utf-8>
-<title>Tests for animation type of accumulation</title>
+<title>Accumulation for each property</title>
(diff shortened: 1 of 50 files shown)


jgraham:
Code reviewed upstream.

PR #8331, opened 2017-11-20
Drop getComputedStyle.html test in interfaces/AnimationEffectTiming;

jgraham:
This test feels out of place -- there's no getComputedStyle method on
AnimationEffectTiming and we generally test timing properties by calling
getComputedTiming.

This patch moves most of the tests in this file to the corresponding members
whose liveness they check while updating them to use getComputedTiming instead.

The remaining tests that seek the currentTime and check the style with various
values of endDelay should be covered by
timing-model/animation-effects/simple-iteration-progress.html

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8331)


jgraham:
Code reviewed upstream.

PR #8336, opened 2017-11-20
Change tests of calc in media-queries so that the root element style has been resolved first.

jgraham:
This modify the tests so that the media query evaluation they're testing happens
after style resolution, which would've caught a bug in rem unit evaluation
inside media queries in stylo.

MozReview-Commit-ID: ByaR4ZA995l

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1396057 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8336)


jgraham:
Code reviewed upstream.

PR #8339, opened 2017-11-20
Replace var with const/let in web-platform-tests/web-animations;

jgraham:
Because 'const' is longer than 'var', in quite a number of places this patch
also updates the indentation of subsequent lines.

Also, in a number of cases this means the line now needs to be wrapped. In
general I've used 'prettier' to do the line wrapping which means that the line
wrapping style differs a little from the existing code (which is already quite
inconsistent). Hopefully in the future we can use prettier on all files to make
this more consistent.

Also, in some cases, to avoid further line wrapping, this patch uses shorthand
property names (specifically replacing 'composite: composite' with just
'composite'). This appears to be supported in all browsers like to implement Web
Animations (Firefox 33+, Chrome 47+, Edge, Safari 9+).

MozReview-Commit-ID: 2xEaeZSYjlc

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8339)

### DIFF
--- a/web-animations/animation-model/animation-types/accumulation-per-property.html
+++ b/web-animations/animation-model/animation-types/accumulation-per-property.html
@@ -17,16 +17,16 @@ html {
 <script>
 'use strict';
 
-for (var property in gCSSProperties) {
+for (const property in gCSSProperties) {
   if (!isSupported(property)) {
     continue;
   }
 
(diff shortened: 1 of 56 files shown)


jgraham:
Code reviewed upstream.

PR #8346, opened 2017-11-20
Simplify getComputedTiming.html test somewhat;

jgraham:
This test doesn't actually need a target element or keyframes since it is only
testing timing.

Furthermore, we generally prefer to test KeyframeEffect over
KeyframeEffectReadOnly so this patch also changes to using the KeyframeEffect
constructor.

Many of the other tests in this folder could be similarly simplified in the
future.

MozReview-Commit-ID: GDwsgRCEpw1

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1417808 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8346)


jgraham:
Code reviewed upstream.

PR #8377, opened 2017-11-20
Add tests for PaymentHandler CanMakePaymentEvent.

rsolomakhin:
@marcoscaceres Please take a look at a couple of tests for pull request https://github.com/w3c/payment-handler/pull/170. The tests should be firing the `CanMakePaymentEvent` because the payment method is URL-based. No manifests should be necessary for these tests as long as they run on https://w3c-test.org.


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20893)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/304880716)

tobie:
I'd suggest removing copyright notices from source files. If everyone starts annotating the projects with custom copyright notices things are going to get complicated quickly. If I modify this file in 2018, do I change the copyright date? Add my name to the copyright? Etc.

payment-handler/app-can-make-payment-false.js
```diff
@@ -0,0 +1,6 @@
+// Copyright © 2017 Chromium authors and World Wide Web Consortium, (Massachusetts Institute of Technology, ERCIM, Keio University, Beihang).
```
marcoscaceres: I wonder if we can merge this file and the next one? Then we can send test conditions via the PMI: 

`https://w3c-test.org?canMakePayment=true`
`https://w3c-test.org?canMakePayment=false`
`https://w3c-test.org?canMakePayment=promise-false`
`https://w3c-test.org?canMakePayment=promise-true`

And so on...

payment-handler/app-can-make-payment-false.js
```diff
@@ -0,0 +1,6 @@
...
+
+// This app always responds to CanMakePaymentEvent with "false".
+self.addEventListener('canmakepayment', (canMakePaymentEvent) => {
+    canMakePaymentEvent.respondWith(false);
```
marcoscaceres: Although this gets auto wrapped with a promise, it might be good to have two tests:

```JS
canMakePaymentEvent.respondWith(false);
```

And:

```JS
canMakePaymentEvent.respondWith(Promise.resolve(false));
```

marcoscaceres: It might be good to also test for:

```JS
canMakePaymentEvent.respondWith(Promise.reject(new Error("Custom error")));
```

payment-handler/app-can-make-payment-true.js
```diff
@@ -0,0 +1,6 @@
+// Copyright © 2017 Chromium authors and World Wide Web Consortium, (Massachusetts Institute of Technology, ERCIM, Keio University, Beihang).
+
+// This app always responds to CanMakePaymentEvent with "true".
+self.addEventListener('canmakepayment', (canMakePaymentEvent) => {
+    canMakePaymentEvent.respondWith(true);
```
marcoscaceres: All the things above would of course apply here too :)

interfaces/payment-handler.idl
```diff
@@ -4,6 +4,7 @@ interface PaymentManager {
     [SameObject]
     readonly attribute PaymentInstruments instruments;
     [Exposed=Window] static Promise<PermissionState> requestPermission();
+    attribute DOMString userHint;
```
marcoscaceres: This probably shouldn't be added yet, unless it's also going to be tested

interfaces/payment-handler.idl
```diff
@@ -27,6 +28,21 @@ dictionary ImageObject {
              DOMString sizes;
              DOMString type;
 };
+[Constructor(DOMString type, CanMakePaymentEventInit eventInitDict),
```
marcoscaceres: I was expecting to also see tests for the `CanMakePaymentEvent ` constructor and `CanMakePaymentEventInit`. I guess they are forthcoming?

payment-handler/can-make-payment-event.https.html
```diff
@@ -0,0 +1,53 @@
...
+<script src="/resources/testharnessreport.js"></script>
+<script>
+promise_test(async t => {
+  await navigator.serviceWorker.register('app-can-make-payment-false.js');
+  const registration = await navigator.serviceWorker.ready;
+  await registration.paymentManager.instruments.set('instrument-key', {
+    name: 'Test Payment Method',
+    enabledMethods: 'https://w3c-test.org',
+  });
+  await promise_rejects(t, 'NotSupportedError', (new PaymentRequest([{
```
marcoscaceres: nit: please split this up a bit, otherwise, it's a bit unclear as to what might be causing the `NotSupportedError`:

```JS
const request = new PaymentRequest(...);
promise_reject(t, 'NotSupportedError', request.show())
```

payment-handler/can-make-payment-event.https.html
```diff
@@ -0,0 +1,53 @@
...
+        currency: 'USD',
+        value: '0',
+      },
+    },
+  })).show());
+  await registration.unregister();
+}, 'If CanMakePaymentEvent.respondWith(false) is called, then the payment method is not supported.');
+
+promise_test(async t => {
+  await navigator.serviceWorker.register('app-can-make-payment-true.js');
```
marcoscaceres: There is quite a bit of duplication in both test tests. I wonder if we can reduce some things into a function, like:
```
function makePaymentRequest(){
   return new PaymentRequest([{
     supportedMethods: 'https://w3c-test.org',
   }], {
     total: {
       label: 'Total',
       amount: {
         currency: 'USD',
         value: '0',
       },
     },
   });
}
```

payment-handler/can-make-payment-event.https.html
```diff
@@ -0,0 +1,53 @@
...
+        value: '0',
+      },
+    },
+  });
+  const acceptPromise = request.show();
+  await request.abort();
+  await promise_rejects(t, 'AbortError', acceptPromise);
+  await registration.unregister();
+}, 'If CanMakePaymentEvent.respondWith(true) is called, then the payment method is supported.');
+</script>
```
marcoscaceres: I was expecting to see `request.canMakePayment()` tests here. That works with `CanMakePaymentEvent`, right?

marcoscaceres (changes requested):
Thanks for making a start on these. Couple of questions and suggestions.

marcoscaceres:
In the spec's PR, there is also `attribute EventHandler oncanmakepayment;`, but it's not part of this set of tests. Should we add it here?

interfaces/payment-handler.idl
rsolomakhin: Removed 'userHint'.

interfaces/payment-handler.idl
rsolomakhin: Yep. Adding tests step-by-step.

payment-handler/app-can-make-payment-false.js
rsolomakhin: Merged the app files.

payment-handler/app-can-make-payment-false.js
rsolomakhin: Added more respondWith() patterns.

payment-handler/app-can-make-payment-true.js
rsolomakhin: Roger, roger.

payment-handler/can-make-payment-event.https.html
rsolomakhin: Split up.

payment-handler/can-make-payment-event.https.html
rsolomakhin: Places into a common function.

payment-handler/can-make-payment-event.https.html
rsolomakhin: Added the request.canMakePayment() calls as well.

rsolomakhin:
@marcoscaceres: please take another look.

> In the spec's PR, there is also `attribute EventHandler oncanmakepayment;`, but it's not part of this set of tests. Should we add it here?

That would require referencing the IDL for service worker, which is not in the `/interfaces/` directory at the moment. I'm not super familiar with how the IDL tester works. Is it possible to stub out the service worker IDL to enable testing our partial interface?

rsolomakhin:
@tobie: I'm just copy-pasting headers from other files in the repository. What's the official word from this repo owners on copyrights?

It would be cool if the linter ensure absence or presence (whichever way the wind blows) of the copyright headers.

jgraham:
Please don't add the headers; the `LICENSE.md` file in the root is authoritative and putting information in individual files is not necessary or desirable.

A patch to lint for this would be accepted.

interfaces/payment-handler.idl
marcoscaceres: Ok, cool. We can probably reuse quite a bit of: 
https://github.com/w3c/web-platform-tests/tree/master/payment-request/PaymentRequestUpdateEvent

payment-handler/app-can-make-payment.js
```diff
@@ -0,0 +1,26 @@
+// Copyright © 2017 Chromium authors and World Wide Web Consortium, (Massachusetts Institute of Technology, ERCIM, Keio University, Beihang).
+
+self.addEventListener('canmakepayment', (canMakePaymentEvent) => {
+    const methodName = canMakePaymentEvent.methodData[0].supportedMethods[0];
+    const behavior = methodName.split('/')[3];
+    switch (behavior) {
```
marcoscaceres: Nice!

payment-handler/can-make-payment-event.https.html
```diff
@@ -0,0 +1,94 @@
...
+
+promise_test(async t => {
+  const request = await registerAppAndBuildPaymentRequest(window.location.origin + '/canMakePayment-promise-true');
+  let paymentRequestCanMakePaymentResult = false;
+  try {
+    paymentRequestCanMakePaymentResult = await request.canMakePayment();
+  } catch (err) {
+    assert_equals(err.name, 'NotAllowedError', 'If it throws, then it must be NotAllowedError');
+  }
+  assert_equals(paymentRequestCanMakePaymentResult, true, 'canMakePayment() should return true.');
```
marcoscaceres: nit: assert_true

payment-handler/can-make-payment-event.https.html
```diff
@@ -0,0 +1,94 @@
...
+        currency: 'USD',
+        value: '0',
+      },
+    },
+  });
+  return request;
+}
+
+promise_test(async t => {
+  const request = await registerAppAndBuildPaymentRequest(window.location.origin + '/canMakePayment-false');
```
marcoscaceres: nit: would you mind putting the url into a variable. It helps a lot with debugging. 

```JS
const url = window.location.origin + '/canMakePayment-false'; 
```

payment-handler/can-make-payment-event.https.html
```diff
@@ -0,0 +1,94 @@
...
+  let paymentRequestCanMakePaymentResult = true;
+  try {
+    paymentRequestCanMakePaymentResult = await request.canMakePayment();
+  } catch (err) {
+    assert_equals(err.name, 'NotAllowedError', 'If it throws, then it must be NotAllowedError');
+  }
+  assert_equals(paymentRequestCanMakePaymentResult, false, 'canMakePayment() should return false.');
```
marcoscaceres: nit: assert_false

marcoscaceres: (fix this and the other globally)

payment-handler/app-can-make-payment.js
```diff
@@ -0,0 +1,24 @@
...
+            canMakePaymentEvent.respondWith(Promise.resolve(true));
+            break;
+        case 'canMakePayment-promise-false':
+            canMakePaymentEvent.respondWith(Promise.resolve(false));
+            break;
+        case 'canMakePayment-custom-error':
+            canMakePaymentEvent.respondWith(Promise.reject(new Error('Custom error')));
+            break;
+        default:
+            console.log('Unrecognized payment method name ' + methodName);
```
marcoscaceres: We should probably `.respondWith(new error())` by default here - to prevent the test from locking up. That should cause the expecting test to fail.

payment-handler/can-make-payment-event.https.html
```diff
@@ -0,0 +1,98 @@
...
+promise_test(async t => {
+  const methodName = window.location.origin + '/canMakePayment-promise-false';
+  const request = await registerAppAndBuildPaymentRequest(methodName);
+  let paymentRequestCanMakePaymentResult = false;
+  try {
+    paymentRequestCanMakePaymentResult = await request.canMakePayment();
+  } catch (err) {
+    assert_equals(err.name, 'NotAllowedError', 'If it throws, then it must be NotAllowedError');
+  }
+  assert_false(paymentRequestCanMakePaymentResult, 'canMakePayment() should return false.');
```
marcoscaceres: if `catch` happens, would this be a false positive?

payment-handler/can-make-payment-event.https.html
```diff
@@ -0,0 +1,98 @@
...
+  await promise_rejects(t, 'NotSupportedError', request.show());
+}, 'If CanMakePaymentEvent.respondWith(false) is called, then the payment method is not supported.');
+
+promise_test(async t => {
+  const methodName = window.location.origin + '/canMakePayment-promise-false';
+  const request = await registerAppAndBuildPaymentRequest(methodName);
+  let paymentRequestCanMakePaymentResult = false;
```
marcoscaceres: Maybe just: 

```
let paymentRequestCanMakePaymentResult;
```

That  should avoid the false positive below, I think,

payment-handler/can-make-payment-event.https.html
```diff
@@ -0,0 +1,98 @@
...
+    assert_equals(err.name, 'NotAllowedError', 'If it throws, then it must be NotAllowedError');
+  }
+  assert_false(paymentRequestCanMakePaymentResult, 'canMakePayment() should return false.');
+  await promise_rejects(t, 'NotSupportedError', request.show());
+}, 'If CanMakePaymentEvent.respondWith(Promise.resolve(false)) is called, then the payment method is not supported.');
+
+promise_test(async t => {
+  const methodName = window.location.origin + '/canMakePayment-true';
+  const request = await registerAppAndBuildPaymentRequest(methodName);
+  let paymentRequestCanMakePaymentResult = true;
```
marcoscaceres: As above... don't give it an initial value and let `await request.canMakePayment();` assign the real value.

payment-handler/can-make-payment-event.https.html
```diff
@@ -0,0 +1,98 @@
...
+  try {
+    paymentRequestCanMakePaymentResult = await request.canMakePayment();
+  } catch (err) {
+    assert_equals(err.name, 'NotAllowedError', 'If it throws, then it must be NotAllowedError');
+  }
+  assert_true(paymentRequestCanMakePaymentResult, 'canMakePayment() should return true.');
```
marcoscaceres: s/should/must/g

payment-handler/can-make-payment-event.https.html
```diff
@@ -0,0 +1,98 @@
...
+  const acceptPromise = request.show();
+  await request.abort();
+  await promise_rejects(t, 'AbortError', acceptPromise);
+}, 'If CanMakePaymentEvent.respondWith(true) is called, then the payment method is supported.');
+
+promise_test(async t => {
+  const methodName = window.location.origin + '/canMakePayment-promise-true';
+  const request = await registerAppAndBuildPaymentRequest(methodName);
+  let paymentRequestCanMakePaymentResult = true;
```
marcoscaceres: similarly here as above.

payment-handler/can-make-payment-event.https.html
```diff
@@ -0,0 +1,98 @@
...
+  try {
+    paymentRequestCanMakePaymentResult = await request.canMakePayment();
+  } catch (err) {
+    assert_equals(err.name, 'NotAllowedError', 'If it throws, then it must be NotAllowedError');
+  }
+  assert_true(paymentRequestCanMakePaymentResult, 'canMakePayment() should return true.');
+  const acceptPromise = request.show();
+  await request.abort();
+  await promise_rejects(t, 'AbortError', acceptPromise);
```
marcoscaceres: Nit: I wonder if we should bother having this as part of the test? We tested this fairly well in the Payment Request tests.

marcoscaceres: To be clear, I mean including this line: 

```
await promise_rejects(t, 'AbortError', acceptPromise);
```

Applied similarly to other tests above.

marcoscaceres (changes requested):
Couple of small issues, but looking overall looking good.

payment-handler/app-can-make-payment.js
rsolomakhin: Done.

payment-handler/can-make-payment-event.https.html
rsolomakhin: If `catch` happens, then comparing `paymentRequestCanMakePaymentResult` would not make sense. Let's start with an undefined value and then set it to the expected value in the `catch`. This will prevent the test failing due to `NotAllowedError`.

payment-handler/can-make-payment-event.https.html
rsolomakhin: Done.

payment-handler/can-make-payment-event.https.html
rsolomakhin: If I don't have a `promise_rejects` here, then the test shows an error message:

> Uncaught (in promise) AbortError: The website has aborted the payment.

I would like to avoid this error and `promise_rejects` seems like one good way to achieve this.

rsolomakhin:
@marcoscaceres Please take another look.

payment-handler/can-make-payment-event.https.html
```diff
@@ -0,0 +1,144 @@
...
+  let paymentRequestCanMakePaymentResult;
+  try {
+    paymentRequestCanMakePaymentResult = await request.canMakePayment();
+  } catch (err) {
+    assert_equals(
+      err.name,
+      "NotAllowedError",
+      "If it throws, then it must be NotAllowedError"
+    );
+    paymentRequestCanMakePaymentResult = false;
```
marcoscaceres: I'm still not super comfortable with this, tbh. I think we should do the assert_equals for the Error, but not `paymentRequestCanMakePaymentResult`.

This should just error at `assert_false`.

marcoscaceres (changes requested):
These are looking good now, but I still have a concern that we should just let the tests fail when throwing happens (even if we check what was thrown was ok).

marcoscaceres:
@rsolomakhin, please note that I reformatted the tests, so make sure you `git pull` before making changes.

payment-handler/can-make-payment-event.https.html
rsolomakhin: Let's discuss with more bandwidth at the upcoming editors meeting.

rsolomakhin:
Not sure what github is doing here. Going to close and re-create the pull request.

PR #8395, opened 2017-11-21
Update path to fonts/Ahem.ttf: "../../fonts/Ahem.ttf" => "./fonts/Ahem.ttf"

rwaldron:
@gsnedders does this make sense to you?

Details originally posted here: https://github.com/w3c/wptdashboard/issues/218#issuecomment-346129172

--------------------------------------------

With the following command:

```
$ ./wpt run chrome infrastructure/assumptions/ahem.html
```

I'm seeing Chrome fail:

```
Summary
=======

Ran 1 tests
Expected results: 0
Unexpected results: 1 (FAIL: 1)

Unexpected Results
==================

FAIL /infrastructure/assumptions/ahem.html
```

And I captured this: 

I went to the history and the path to `fonts/Ahem.ttf` was changed here: https://github.com/w3c/web-platform-tests/blob/master/infrastructure/assumptions/ahem.html; when I changed it from `../../fonts/Ahem.ttf` to `./fonts/Ahem.ttf`, the test passes:

```
Summary
=======

Ran 1 tests
Expected results: 1
Unexpected results: 0

OK
```

And now ahem-ref.html looks like this:

### DIFF
--- a/infrastructure/assumptions/ahem-ref.html
+++ b/infrastructure/assumptions/ahem-ref.html
@@ -4,7 +4,7 @@
 <style>
 @font-face {
   font-family: Ahem;
-  src: url("../../fonts/Ahem.ttf");
+  src: url("./fonts/Ahem.ttf");
 }
 * {
   padding: 0;


gsnedders (changes requested):
This is changing it so the URL doesn't resolve, and therefore it isn't rendering using Ahem (and instead using the fallback font).

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20946)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/305434740)

gsnedders:
@rwaldron If you try running `wpt run` with `--install-fonts` you'll see this pass without the change. The problem is that rely on Ahem being installed and it isn't. (The change just makes Ahem fail to load as a webfont, as there is no `infrastructure/assumptions/fonts/Ahem.ttf`, and as such it renders with the same fallback font as a test.)

rwaldron:
> This is changing it so the URL doesn't resolve, and therefore it isn't rendering using Ahem (and instead using the fallback font).

I'd like to learn more about this from you, so I can understand why the test goes from failing to passing. I will ping you directly in irc.w3.org#testing

rwaldron:
Closing!

PR #8400, opened 2017-11-22
Update outline-017.html to reduce flakiness in Edge

FremyCompany:
Edge runs the color animation in a render thread so the computed value of the property is not guaranteed to be synchronized immediately.

### DIFF
--- a/css/css-ui/outline-017.html
+++ b/css/css-ui/outline-017.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+a<!DOCTYPE html>
 <meta charset="utf-8">
 <title>CSS Basic User Interface Test: interpolable outline-* properties animation</title>
 <link rel="author" title="Florian Rivoal" href="http://florian.rivoal.net">
@@ -9,18 +9,18 @@
 <style>
 @keyframes outline-anim {
   from {
-    outline: solid 1px rgb(1,0,0);
+    outline: solid 1px rgba(1, 0, 0, 0.5);
     outline-offset: 1px;
   }
   to {
-    outline: solid 3px rgb(3,0,0);
+    outline: solid 3px rgb(3, 0, 0, 0.5);
     outline-offset: 3px;
   }
 }
 
 #test {
   animation: outline-anim 3s -1.5s paused linear;
-  outline: solid 1px rgb(1,0,0);
+  outline: solid 1px rgb(1, 0, 0, 0.5);
   outline-offset: 1px;
 }
 </style>
@@ -34,7 +34,7 @@
         test(
           function(){
             var test = document.getElementById("test");
-            assert_equals(getComputedStyle(test).outlineColor, 'rgb(2, 0, 0)');
+            assert_equals(getComputedStyle(test).outlineColor, 'rgba(2, 0, 0, 0.5)');
           }, "outline-color is animated as a color");
         test(
           function(){
--- a/css/css-ui/outline-017.html
+++ b/css/css-ui/outline-017.html
@@ -13,14 +13,14 @@
     outline-offset: 1px;
   }
   to {
-    outline: solid 3px rgb(3, 0, 0, 0.5);
+    outline: solid 3px rgba(3, 0, 0, 0.5);
     outline-offset: 3px;
   }
 }
 
 #test {
   animation: outline-anim 3s -1.5s paused linear;
-  outline: solid 1px rgb(1, 0, 0, 0.5);
+  outline: solid 1px rgba(1, 0, 0, 0.5);
   outline-offset: 1px;
 }
 </style>
--- a/css/css-ui/outline-017.html
+++ b/css/css-ui/outline-017.html
@@ -29,20 +29,24 @@
   <div id=log></div>
 
   <script>
-    test(
-      function(){
-        var test = document.getElementById("test");
-        assert_equals(getComputedStyle(test).outlineColor, 'rgb(2, 0, 0)');
-      }, "outline-color is animated as a color");
-    test(
-      function(){
-        var test = document.getElementById("test");
-        assert_equals(getComputedStyle(test).outlineWidth, '2px');
-      }, "outline-width is animated as a length");
-    test(
-      function(){
-        var test = document.getElementById("test");
-        assert_equals(getComputedStyle(test).outlineOffset, '2px');
-      }, "outline-offset is animated as a length");
+    setup({explicit_done:true});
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/20952)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/305555135)

css/css-ui/outline-017.html
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+a<!DOCTYPE html>
```
jdm: This looks like a mistake.

FremyCompany: Indeed. That will teach me to use the online interface to submit my changes :) I probably pressed the "a" a second time after my ctrl+a+delete :)

FremyCompany:
Looking at the test it looks like Firefox is still unstable despite that change. Not sure I can do anything to solve that. Any opinion on what to do next?

css/css-ui/outline-017.html
```diff
@@ -29,20 +29,24 @@
...
-      function(){
-        var test = document.getElementById("test");
-        assert_equals(getComputedStyle(test).outlineWidth, '2px');
-      }, "outline-width is animated as a length");
-    test(
-      function(){
-        var test = document.getElementById("test");
-        assert_equals(getComputedStyle(test).outlineOffset, '2px');
-      }, "outline-offset is animated as a length");
+    setup({explicit_done:true});
```
gsnedders: It would seem better to use `async_test` for this rather than `test` with `explicit_done`.

css/css-ui/outline-018.html
```diff
@@ -16,28 +16,43 @@
...
-        test.style.animation = "outline-anim 3s 0s paused linear"
-        assert_equals(getComputedStyle(test).outlineStyle, 'solid');
-        test.style.animation = "outline-anim 3s -1s paused linear"
-        assert_equals(getComputedStyle(test).outlineStyle, 'solid');
-        test.style.animation = "outline-anim 3s -2s paused linear"
-        assert_equals(getComputedStyle(test).outlineStyle, 'dotted');
-        test.style.animation = "outline-anim 3s 0s paused reverse-linear"
-        assert_equals(getComputedStyle(test).outlineStyle, 'dotted');
-      }, "outline-style is animated as a discrete type");
+    setup({ explicit_done: true });
```
gsnedders: Likewise: it would seem better to use `async_test` for this rather than `test` with `explicit_done`.

gsnedders (dismissed):
wrt it being flaky in Firefox: @birtles, do you have any idea how to make this not flaky?

birtles:
What sort of flakiness are you seeing? If it's a minor floating-point difference or something you might find using a step timing function helps (but it shouldn't really matter for a paused animation like this).

FremyCompany:
@birtles Not sure, it looks like the color transition isn't reported in the computed style by the first frame render. Maybe I should just do some rAF(time => rAF(time => ...)) to make sure an entire frame has lapsed and it is not possible from animation style not to have made it to the main thread by then. I'll try that sometime this week when I also convert tests to async tests.

birtles:
> @birtles Not sure, it looks like the color transition isn't reported in the computed style by the first frame render. Maybe I should just do some rAF(time => rAF(time => ...)) to make sure an entire frame has lapsed and it is not possible from animation style not to have made it to the main thread by then.

You shouldn't need to wait a frame at all ([test case](http://jsbin.com/laselut/edit?html,css,js,console,output)). All compositor animations also run on the main thread on-demand so that you don't need to wait for synchronizing with another process/thread (and I suspect Edge will hit compat problems down the line if it doesn't do that).

How can I see the test failures? The [log I checked](https://travis-ci.org/w3c/web-platform-tests/jobs/305555142) said Firefox was stable and when I run the original test locally it seems fine.

gsnedders:
@birtles https://travis-ci.org/w3c/web-platform-tests/jobs/305920557 is the latest log, containing:

| `/css/css-ui/outline-017.html` | `outline-color is animated as a color` | **FAIL: 1/10, PASS: 9/10** | `assert_equals: expected "rgba(2, 0, 0, 0.5)" but got "rgba(1, 0, 0, 0.5)"` |

Doing it locally I think would be something like `./wpt run --pause-on-unexpected --repeat-until-unexpected firefox css/css-ui/outline-017.html`.

birtles:
@gsnedders Thanks. How can I tell the exact version of Firefox being used in that log?

For my own reference, I suspect this could be [Mozilla bug 1405548](https://bugzilla.mozilla.org/show_bug.cgi?id=1405548) which should already be fixed.

FremyCompany:
> All compositor animations also run on the main thread on-demand so that you don't need to wait for synchronizing with another process/thread (and I suspect Edge will hit compat problems down the line if it doesn't do that).

That is not mandated by any spec and I would rather not have tests depend on it. We can reconsider our design if/when we hit such issues, but the test currently claims we do not interpolate outline-color which is not true, and I would like to get that fixed.

gsnedders:
@birtles I'm not sure there is any way; @jgraham? It should be a Firefox Nightly build, which given the date of the run should include the fix for 1405548.

gsnedders:
That said, if we're confident the flakiness is not caused by the changes in this PR, we should merge this PR.

birtles:
Yeah, I thought last time I checked, the nightly builds were quite old but I don't know how I checked that.

css/css-ui/outline-018.html
FremyCompany: I just reconsidered it, but after starting to rewrite this way, I realized this doesn't show the intent really well. These tests are synchronous, they are just delayed to allow the animation-sync to happen. If you use async_test and use a requestAnimationFrame in each of the test, each test is going to wait and it's fuzzier why this happens.

I'd be happy to add a comment explaining this in the file though if you believe that is a good idea.

zcorpan:
Should we land this, or is it waiting for further changes?

FremyCompany:
Is there a way to redo the CI without pushing new changes? 
I'm fairly sure we should just merge this

zcorpan:
Closed and reopened to trigger a new CI run.

gsnedders (commented):
Running with --verify, even with large numbers of iterations, I can't repro any flake in Firefox now.

zcorpan (approved):

PR #8412, opened 2017-11-23
Load document with a script instead of using execute_script in wdspec…

carlosgcampos:
… user prompts tests

The tests are expecting that after execute_script() the dialog is still
present and no error is returned. This can't happen according to the
spec, because in case of user prompt in the middle of a script
execution, the user prompt handler should be invoked. The result of the
handler will depend on the unhandled prompt behavior, but in any case
the result should be that either the dialog is no longer present (accept
or dismiss) or unexpected alert error is returned.
To avoid this, we could simply use the inline fixture to load a document
containing the script, instead of using execute_script().

### DIFF
--- a/webdriver/tests/user_prompts/accept_alert.py
+++ b/webdriver/tests/user_prompts/accept_alert.py
@@ -1,4 +1,5 @@
 from tests.support.asserts import assert_error, assert_success
+from tests.support.inline import inline
 
 
 def accept_alert(session):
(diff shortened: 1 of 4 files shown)


andreastt (approved):

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21005)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/306284990)

PR #8415, opened 2017-11-24
add note about pipes applying to static files only

gterzian:
I tried transforming the response coming from a python file handler, and it didn't seem to work, digging into the code seemed to confirm that the pipes are only applied to the response of `FileHandlers`, see https://github.com/w3c/web-platform-tests/blob/master/tools/wptserve/wptserve/handlers.py#L150.

If this is correct, a note in the docs is helpful from my point of view.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8415)

### DIFF
--- a/tools/wptserve/docs/pipes.rst
+++ b/tools/wptserve/docs/pipes.rst
@@ -11,6 +11,10 @@ functions are applied to the response from left to right. For example::
 This would serve bytes 1 to 199, inclusive, of foo.txt with the HTTP status
 code 404.
 
+.. note::
+   Pipes are only applied to static files, and will not work if applied to
+   other types of handlers, such as Python File Handlers.
+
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PASSED

Started: 2017-11-24 04:49:14
Finished: 2017-11-24 05:02:38





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21012)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/306597004)

gsnedders:
@jgraham do you know why this is the case? is there any reason why we shouldn't make it work everywhere?

jgraham (approved):

jgraham:
I think it could work in most cases, it's just that it doesn't at the moment. I would be very happy to get a patch to improve this.

PR #8416, opened 2017-11-24
WakeLockRequest is independent rather than WakeLock

### LINKED ISSUE #8411: wakelock-object-is-independent.https.html does not reflect the spec intent
https://github.com/w3c/web-platform-tests/blob/master/wake-lock/wakelock-object-is-independent.https.html

https://w3c.github.io/wake-lock/#extensions-to-the-navigator-interface

The spec intent is:

1. Once there is a wake lock resolution for a `Document` which is a `WakeLock` object, `Navigator.getWakeLock()` for the same Document must always return a promise resolved with the same object.

2. `WakeLock.active` reflects the global state of wake lock of this type. It is therefore an error for two `WakeLock` objects of the same wake lock type to return different values of `active`.

The test seems to expect the exact opposite.

It is `WakeLockRequest` objects that must be independent, not `WakeLock`. `WakeLock` is intended to be a singleton for this script execution context just like `Navigator`.

Honry:
- Test wake lock state is global
  - Test two wake lock returns same object
  - Fixed #8411


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21013)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/306621414)

wake-lock/wakelock-state-is-global.https.html
```diff
@@ -0,0 +1,74 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>wake lock state should be global</title>
+<link rel="help" href="https://w3c.github.io/wake-lock/">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<body>
+<script id="iframe" type="text/plain">
+let iframeWakeLock, iframeRequest;
```
marcoscaceres: Hmm... this is super racy... we can live with it... but not my favorite.

wake-lock/wakelock-state-is-global.https.html
```diff
@@ -0,0 +1,74 @@
...
+  assert_false(wakeLock.active, "wakeLock is initially false");
+
+  //when iframe wake lock is acquired, parent wake lock should be actived
+  iframe.contentWindow.postMessage("ACQUIRED", "*");
+  const isActive1 = await wait_for_message(iframe);
+  await eventWatcher.wait_for("activechange");
+  assert_true(isActive1, "the iframe wake lock state is actived when iframe wake lock is acquired");
+  assert_true(wakeLock.active, "the wake lock state is actived when iframe wake lock is acquired");
+
+
```
marcoscaceres: nit: remove this line.

marcoscaceres (changes requested):
I'm a bit concerned about `wakelock-state-is-global.https.html`, tbh... the design of using a global  is quite racy. I could be convinced to live with it, but can you maybe come up with a better way of keeping state than relying on globals?

marcoscaceres:
If you can't come up with anything better for wakelock-state-is-global.https.html, that's ok... let me know.

Honry:
@marcoscaceres, thanks for your review, I'm afraid I have no much better idea, @andrey-logvinov, do you have any good opinion?

marcoscaceres (approved):
I guess we have enough flow control in the test that should prevent race conditions. Hopefully this won't come back to bite us.

Honry:
@marcoscaceres, thanks, I hope so. :)

marcoscaceres:
Ok, merging to move forward, but @andrey-logvinov, please do take a look and file any bugs you might find (or if you can think of another way of testing this).

wake-lock/wakelock-promise.https.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>navigator.getWakeLock() for the same Document always return same WakeLock promise</title>
```
andrey-logvinov: It is not required to always return the same promise - it is only required that if a returned promise is ever resolved or rejected, all returned promises must be resolved or rejected with the same value.

Honry: > it is only required that if a returned promise is ever resolved, all returned promises must be resolved with the same value. 

About this checkpoint, isn't the test satisfied?  For sure, the desc is a bit misunderstanding.

> Also rejections must be consistent, e.g. if one promise is rejected with SecurityException, the other must be rejected also with SecurityException, but it is not required for two exceptions to be the same object.

But how to prove that the rejection is caused by other rejected promise in same Document? Since the rejected condition is global, they should be rejected beyond doubt.

andrey-logvinov: Sorry for late response.
If I understand it correctly the test asserts that the same promise object is returned, which necessarily implies that the resolution is also the same. But the spec does not require to return exactly the same promise. This is not very relevant from the script point of view but it might complicate an implementation e.g. if it is technically easier to return different promise objects resolved with the same value than exactly the promise object. For now I think the test is adequate.

As for the rejection I agree that it is effectively untestable because it requires an anticipated scenario where the rejection might not be consistent but there is no such scenario as rejection conditions are constant in time for the same Document.

PR #8422, opened 2017-11-24
Adjust XMLHttpRequest Content-Type handling

annevk:
See https://github.com/whatwg/mimesniff/pull/36.

### DIFF
--- a/xhr/send-content-type-charset.htm
+++ b/xhr/send-content-type-charset.htm
@@ -40,7 +40,7 @@
       )
       request(
         "text/plain;charset=utf-8",
-        "text/plain;charset=utf-8",
+        "text/plain;charset=UTF-8",
         "Correct text/plain MIME with charset"
       )
       request(
@@ -54,13 +54,18 @@
         "If no charset= param is given, implementation should not add one - known MIME"
       )
       request(
+        "text/plain;  hi=bye",
+        "text/plain;  hi=bye",
+        "If no charset= param is given, implementation should not add one - known MIME, unknown param, two spaces"
+      )
+      request(
         "text/x-thepiano;charset= waddup",
         "text/x-thepiano;charset=UTF-8",
         "charset given but wrong, fix it (unknown MIME, bogus charset)"
       )
       request(
         "text/plain;charset=utf-8;charset=waddup",
-        "text/plain;charset=utf-8;charset=UTF-8",
+        "text/plain;charset=UTF-8",
         "charset given but wrong, fix it (known MIME, bogus charset)"
       )
       request(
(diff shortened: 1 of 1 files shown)


annevk:
Always replacing with uppercase UTF-8 could be considered I think. I think Firefox does not do that because it broke some sites, but those may well be gone/fixed now.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21031)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/306798513)

annevk:
TODO: tests with double quotes.

annevk:
TODO: update XMLHttpRequest to actually invoke the MIME type parser and serializer when doing this manipulation.

annevk:
It does not seem that Chrome always replaces with UTF-8 by the way. And Firefox replaces a whole lot with UTF-8 as it seems it doesn't do double quotes correctly. This makes it a little bit unclear to me what the processing model should be here. Anyone ideas?

annevk:
This is ready now. This is what is remaining:

- [ ] Merging of https://github.com/whatwg/mimesniff/pull/36 (the new MIME type specification)
- [ ] Usage of that MIME type specification in the XMLHttpRequest Standard

annevk:
This could use review now.

xhr/send-content-type-charset.htm
```diff
@@ -70,14 +75,30 @@
...
-        "text/x-pink-unicorn; charset=UTF-8; charset=UTF-8; notrelated; charset=UTF-8",
-        "If multiple charset parameters are given, all should be rewritten"
+        "text/x-pink-unicorn;charset=UTF-8",
+        "Multiple charset parameters deduplicate, bogus parameter dropped"
       )
       request(
         null,
         "text/plain;charset=UTF-8",
         "No content type set, give MIME and charset"
       )
```
domenic: Is it worth testing that the empty byte sequence gets passed through? (Note you'd have to change the check in request from `if (input)` to `if (input !== null)`.)

domenic: Might be worth testing the other cases of HTML document and XML document. Although I guess they are probably tested elsewhere.

xhr/send-content-type-charset.htm
```diff
@@ -70,14 +75,30 @@
...
+        "charset in double quotes")
+      request(
+        "text/plain;charset=\" utf-8\"",
+        "text/plain;charset=UTF-8",
+        "charset in double quotes with space")
+      request(
+        "text/plain;charset=\"u\\t\\f-8\"",
+        "text/plain;charset=UTF-8",
+        "charset in double quotes with backslashes")
     </script>
```
domenic: Can you add tests that non-charset parameters also get preserved (and parsed/re-serialized), when a charset parameter is present?

domenic (approved):
LGTM but I think at least the tests with non-charset parameters alongside charset should probably be added.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/27437)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/367065238)

benbucksch:
re changing charset=utf-8 to charset=UTF-8:

I am trying to do autodiscover with outlook.com, and use XmlHttpRequest(), and set the Content-Type to "text/xml; charset=utf-8". Due to this bug, outlook.com server (<https://outlook.com/autodiscover/autodiscover.xml> with POST of XML) tells me:
Cannot process the message because the content type 'text/xml;charset=UTF-8' was not the expected type 'text/xml; charset=utf-8'.

Now, I agree that they are wrong and the space is optional and charsets are case insensitive. The (old) spec is clear on that and outlook.com is misbehaving. Yet, this is preventing me from contacting the service, and XmlHttpRequest doesn't let me pass the header as expected anymore.

xref https://bugzilla.mozilla.org/show_bug.cgi?id=1454325

PR #8433, opened 2017-11-24
Move `display: run-in` tests into css-display/run-in/

foolip:
Because `display: run-in` is not widely implemented, having these
tests in a directory of their own is very handy, so that they can be
skipped as a group. It also has its own spec section:
https://drafts.csswg.org/css-display/#run-in-layout

Only one test used a support file, and that support file wasn't used
by any other test at the origin directory.

Part of https://github.com/w3c/web-platform-tests/issues/7652.


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21048)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/306852677)

foolip:
Upon closer inspection it turns out that all tests in css/css-block/ have "run-in" in the title, with the exception of first-line-pseudo-009.xht which also involves run-in but is also testing other stuff. I'll close this and move all of css-block in css-display/run-in instead.

PR #8435, opened 2017-11-25
Use <mtext> instead of <mi> for mathvariant tests

fred-wang:
These tests are supposed to verify that token elements with a single character
and a mathvariant attribute render the same as if we were directly using the
transformed unicode character (instead of the mathvariant attribute). They
currently use the <mi> element which is a bit complex (e.g. its default
mathvariant value may be italic or normal depending of the text content).
In particular, WebKit still has some hacks to adjust the spacing around <mi>
elements which seems to make the mathvariant tests fail because of different
spacing ( https://bugs.webkit.org/show_bug.cgi?id=129097 ). To workaround such
issues unrelated to what is verified by the tests, this commit changes them
to rely on the somewhat simpler <mtext> elements.

@youennf Can you please review this commit? I would need it for https://bugs.webkit.org/show_bug.cgi?id=179110


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21060)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/307143845)

youennf (approved):

PR #8441, opened 2017-11-26
Changing set_timeout command to return None

jimevans:
The WebDriver spec says that the setTimeout command should return null.
This updates the client to honor that part of the spec.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8441)


shs96c (approved):
So it does.

PR #8449, opened 2017-11-27
Add more XMLHttpRequest overrideMimeType() tests

annevk:
For https://github.com/whatwg/xhr/issues/157 and https://github.com/whatwg/mimesniff/issues/42.

### DIFF
--- a/xhr/overridemimetype-blob.html
+++ b/xhr/overridemimetype-blob.html
@@ -27,46 +27,28 @@ async_test(t => {
   client.send();
 }, "Use text/xml as fallback MIME type, 2");
 
-["bogus",
- "†/†",
- "¢/¢",
- "ÿ/ÿ",
- "?/?",
- "a/",
- "/",
- " \t",
- ""
-].forEach((type, index) => {
-  async_test(t => {
-    const client = new XMLHttpRequest();
-    client.onload = t.step_func_done(() => {
-      assert_equals(client.getResponseHeader("Content-Type"), "");
-      assert_equals(client.response.type, "application/octet-stream");
-    });
-    client.open("GET", "resources/status.py");
-    client.responseType = "blob";
-    client.overrideMimeType(type);
-  client.send();
-  }, "Bogus MIME type should end up as application/octet-stream, " + (index + 1));
-});
+promise_test(() => {
+  // Don't load generated-mime-types.json as sending them all over the network would be prohibitive
+  return fetch("../mimesniff/mime-types/resources/mime-types.json").then(res => res.json()).then(runTests);
+}, "Loading data…");
 
-[
- ["HI/x;Test=test;charset=utf-8", "hi/x;test=test;charset=utf-8"],
- ["text/xml;charset=†", "text/xml"],
- ["text/html;", "text/html"],
- ["x/x;?=x;charset=y", "x/x;charset=y"],
- ["text/html;x=ÿ", "text/html;x=\"ÿ\""]
-].forEach(([inputType, outputType], index) => {
-  async_test(t => {
-    const client = new XMLHttpRequest();
-    client.onload = t.step_func_done(() => {
-      assert_equals(client.getResponseHeader("Content-Type"), "");
-      assert_equals(client.response.type, outputType);
-    })
-    client.open("GET", "resources/status.py");
-    client.responseType = "blob";
-    client.overrideMimeType(inputType);
-    client.send();
-  }, "Valid MIME types need to be parsed and serialized " + (index + 1));
-});
+function runTests(tests) {
+  tests.forEach((val, index) => {
+    if(typeof val === "string") {
+      return;
+    }
+    async_test(t => {
+      const client = new XMLHttpRequest(),
+            expectedOutput = val.output !== null ? val.output : "application/octet-stream";
+      client.onload = t.step_func_done(() => {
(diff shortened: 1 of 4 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21085)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/307883487)

annevk:
This is ready for review now. This is what is remaining:

- [ ] Merging of https://github.com/whatwg/mimesniff/pull/36 (the new MIME type specification)
- [ ] Usage of that MIME type specification in the XMLHttpRequest Standard

annevk:
This could use review now.

xhr/overridemimetype-blob.html
```diff
@@ -6,60 +6,67 @@
 <div id="log"></div>
 <script>
 async_test(t => {
-  const client = new XMLHttpRequest()
+  const client = new XMLHttpRequest();
```
domenic: This is extraordinarily hard to review because almost every line has changed, but I guess only because of semicolons? Please split up the commits at least next time.

xhr/overridemimetype-blob.html
```diff
@@ -6,60 +6,67 @@
...
+    assert_equals(client.response.type, "text/xml");
   })
-  client.open("GET", "resources/status.py?content=thisshouldnotmakeadifferencebutdoes")
-  client.responseType = "blob"
-  client.send()
-}, "Use text/xml as fallback MIME type, 2")
+  client.open("GET", "resources/status.py?content=thisshouldnotmakeadifferencebutdoes");
+  client.responseType = "blob";
+  client.send();
+}, "Use text/xml as fallback MIME type, 2");
```
domenic: I don't understand why the above two tests are in a file related to overrideMimeType()?

xhr/overridemimetype-blob.html
```diff
@@ -6,60 +6,67 @@
...
+  async_test(t => {
+    const client = new XMLHttpRequest();
+    client.onload = t.step_func_done(() => {
+      assert_equals(client.getResponseHeader("Content-Type"), "");
+      assert_equals(client.response.type, "application/octet-stream");
+    });
+    client.open("GET", "resources/status.py");
+    client.responseType = "blob";
+    client.overrideMimeType(type);
+  client.send();
```
domenic: Nit: whitespace is off here

xhr/overridemimetype-blob.html
```diff
@@ -6,60 +6,67 @@
...
-  client.overrideMimeType("HI/x;test=test")
-  client.send()
-}, "Valid MIME types need to be normalized")
+[
+ ["HI/x;Test=test;charset=utf-8", "hi/x;test=test;charset=utf-8"],
+ ["text/xml;charset=†", "text/xml"],
+ ["text/html;", "text/html"],
+ ["x/x;?=x;charset=y", "x/x;charset=y"],
+ ["text/html;x=ÿ", "text/html;x=\"ÿ\""]
+].forEach(([inputType, outputType], index) => {
```
domenic: Seems like you could use the whole JSON file for the above two tests. (If the JSON files says a parse failure, application/octet-stream; otherwise use the output.) As I've argued previously, this helps ensure that the same parser is being used in all parts of the codebase.

domenic (approved):
LGTM with comments

annevk:
I've made the change of using the smaller JSON resource here (since it's networking it doesn't seem good to use all), but I need to verify whether some tests should be added to that JSON resource resource that I've removed here or whether coverage remained equivalent. I should probably also have another look at code point usage just in case.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/27439)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/367109471)

xhr/overridemimetype-blob.html
annevk: Sorry, I figured it wouldn't be much of an issue given how GitHub highlights stuff.

thw0rted:
Can I suggest a test here?  I don't see any tests to confirm that the browser follows step 12 of the [updated spec for `open()`](https://github.com/whatwg/xhr/pull/174).  I just got bit by this behavior -- Edge respects the direction to reset the override MIME type while apparently all other modern browsers do not (!).  So, it would be helpful to have a test for:

* make XHR for a resource with MIME type `type1`
* call `overrideMimeType("type2")`
* call `open`
* verify that response has MIME type `type1` not `type2`.

wisniewskit:
@thworted, yeah, it turned out that resetting the MIME type wasn't web compatible (or at least that's how I read the outcome of the investigation in #12289).

annevk:
Web developers should probably fix their servers to not send broken MIME types and migrate to `fetch()`. 😊

thw0rted:
The first thing that comes to mind is XHR against a peer resource in a `file:///` origin.  Of course, a local file won't have a stored MIME type, as such, and while the browser (or OS?) might make a best-effort based on file extension, you aren't likely to have another place to step in and assert that you know better.

Maybe that's not compelling enough?  Still, `overrideMimeType` exists and consistent behavior is in everybody's best interest.

PR #8466, opened 2017-11-28
Record wdspec harness outcomes.

jgraham:
This fixes the problem that wdspec tests fail silently when there
is a problem with collecting the file in pytest.  For example,
if the Python test file contains a SyntaxError we currently fail
to record this as ERROR, and instead report it as OK.

MozReview-Commit-ID: 5W46gaLQa0c

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1416410 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8466)


jgraham:
Code reviewed upstream.

PR #8467, opened 2017-11-28
Move and rewrite playState.html test;

jgraham:
Currently we have a test in interfaces/Animation/playState.html that somewhat
randomly tests the result of the `playState` member.

However, there's no complex logic associated with the `playState` member in the
IDL. It simply returns "The play state of this animation". The logic we need to
test is in the definition of 'play state' which is in the timing model.

As a result we move this test to timing-model/animations/play-states.html

However, this test as it stands does not test the calculation of the play state
in a particularly thorough manner. For example, it does not contain a single
test for the 'finished' state.

Given that this patch series will change the definition of the 'play state' we
first fix this test to cover each of the different cases in the definition of
the 'play state' prior to these changes. That is, we update the tests based on
the definition of 'play state' here:

  https://www.w3.org/TR/2016/WD-web-animations-1-20160913/#play-states

(Note that at this point in the patch series the pref to turn on the changed
definition behavior has not been enabled even for tests so this patch is
actually testing the behavior when that pref is false. We'll replace much of
this test in the next patch but by updating the test first, we should be able to
more clearly see the changes in the next patch.)

MozReview-Commit-ID: 1xkOmuY1SxD

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1412765 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8467)


jgraham:
Code reviewed upstream.

PR #8468, opened 2017-11-28
Update web-platform-tests to test new pending member;

jgraham:
Hopefully most of these changes are self-explanatory however a few notes follow.

* In timing-model/animations/play-states.html, as well as making the tests match
  the updated spec, one or two tests have also been moved to better reflect the
  order in the spec (to make it obvious which branch of the algorithm is being
  tested).

* In timing-model/animations/set-the-timeline-of-an-animation.html we previously
  had two tests that check:
  a) That the playState was 'pending' before and after setting the timeline.
  b) That the playState was 'pending' before setting the timeline and then,
     after setting the timeline and waiting on the ready promise, would become
     'running'.
  Likewise we had the same test for pausing.

  Since these are basically the same test--(b) just adds the wait on the ready
  promise--we combine them here into one test that covers both (a) and (b).

MozReview-Commit-ID: CLoDJvsdwmF

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1412765 [ci skip]

### DIFF
--- a/web-animations/interfaces/Animatable/animate-no-browsing-context.html
+++ b/web-animations/interfaces/Animatable/animate-no-browsing-context.html
@@ -76,16 +76,15 @@ promise_test(t => {
     const div = xhrdoc.getElementById('test');
     anim = div.animate(null);
     anim.timeline = document.timeline;
-    assert_equals(anim.playState, 'pending',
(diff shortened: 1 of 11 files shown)


jgraham:
Code reviewed upstream.

PR #8472, opened 2017-11-28
Prevent missing click event from causing deadlock in interaction.flushEventLoop.

jgraham:
If the DOM click event does not fire, which currently happens in
this edge case where Firefox does not bubble the click event up
to <body>, interaction.flushEventLoop will hang forever.

This is believed to be a bug in Gecko, but it is probably safer to
use a TimedPromise here in case anything else bad were to happen.

Thanks-to: Alexei Barantsev <barancev@gmail.com>
MozReview-Commit-ID: KZBdR8zcfJb

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1417821 [ci skip]

### DIFF
--- a/webdriver/tests/element_click/bubbling.py
+++ b/webdriver/tests/element_click/bubbling.py
@@ -1,0 +1,78 @@
+from tests.support.asserts import assert_success
+from tests.support.inline import inline
+
+
+def click(session, element):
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #8475, opened 2017-11-28
Return no such element error when on no active element.

jgraham:
document.activeElement will return null if there is no document
element.  This may happen if, for example, in an HTML document the
<body> element is removed.

The WPT test test_sucess_without_body in get_active_element.py
is wrong.  It expects Get Active Element to return null if there
is no document element, but following a recent specification change
we want it to return a no such element error.

Specification change: https://github.com/w3c/webdriver/pull/1157

MozReview-Commit-ID: LQJ3slV9aty

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1420431 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8475)


jgraham:
Code reviewed upstream.

PR #8490, opened 2017-11-29
Adjust RGB tests in css-color

csnardi:
The RGB tests in css-color were simply copies of the RGBA tests, as they still included alpha values. Remove the alpha values, and simplify the number of tests (as without the alpha values some would be testing the same concept).


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21185)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/308800719)

dbaron:
The point of these tests is that level 4 of CSS Color adds alpha to the `rgb()` function.  I think the tests are correct as they are now, based on [the current spec draft](https://drafts.csswg.org/css-color-4/#rgb-functions).

hichris1234:
Oh, huh, I had no idea that https://www.w3.org/TR/css-color-4/ wasn't the most current version of the CSS Color 4 spec. Thanks for pointing that out.

PR #8495, opened 2017-11-29
webrtc: Test RTCSctpTransport.maxMessageSize attribute

adam-be:
There's an assumption that regarding canSendSize documented in the test. I'm starting a discussion in the spec PR.

We might want to move the generic `SDPAttributeHelper` code to RTCPeerConnection-helper.js.

Related spec change: https://github.com/w3c/webrtc-pc/pull/1656

### DIFF
--- a/interfaces/webrtc-pc.idl
+++ b/interfaces/webrtc-pc.idl
@@ -507,7 +507,7 @@ partial interface RTCPeerConnection {
 
 interface RTCSctpTransport {
     readonly attribute RTCDtlsTransport transport;
-    readonly attribute unsigned long    maxMessageSize;
+    readonly attribute unrestricted double maxMessageSize;
 };
 
 interface RTCDataChannel : EventTarget {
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21195)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/308886926)

lgrahl:
Further inspiration for tests (in case you haven't seen it already): https://github.com/webrtc/adapter/pull/702

adam-be:
Thanks for the pointer. I think we need all that, but in this particular test I'm trying to limit the functionality to the `maxMessageSize` attribute only.

adam-be:
This test probes the value of `canSendSize` (from spec algorithm) and requires it to be greater than 2. I don't think it's a great solution, but it allows us to do some tests with predictable results. See note in the test.

alvestrand (approved):

PR #8497, opened 2017-11-29
Write tests for instruments.get/set methods

romandev:
This change covers basic behavior of get/set methods.

### DIFF
--- a/payment-handler/payment-instruments-manual.https.html
+++ b/payment-handler/payment-instruments-manual.https.html
@@ -1,0 +1,52 @@
+<!DOCTYPE html>
+<meta charset=utf-8>
+<link rel="help" href="https://w3c.github.io/payment-handler/">
+<title>Test for PaymentInstrument</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/service-workers/service-worker/resources/test-helpers.sub.js"></script>
+<script>
+
+function addInstruments(instruments) {
+  return Promise.all([
+    instruments.set(
+      'test_key1',
+      {
+        name: 'test_instrument1',
+        enabledMethods: ['test_pay']
+      }),
+
+    instruments.set(
+      'test_key2',
+      {
+        name: 'test_instrument2',
+        enabledMethods: ['basic-card']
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21199)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/308908255)

romandev:
@marcoscaceres I've seen your comments in my another PR. (Thank you)
It's not related WebPayment, but I accidentally pushed it there. Sorry for the confusion.

Anyway, I addressed your comments in this change. We might always need manual tests due to requestPermission().

payment-handler/payment-instruments-manual.https.html
```diff
@@ -0,0 +1,52 @@
...
+  assert_equals(testInstrument1.name, 'test_instrument1');
+  assert_array_equals(testInstrument1.enabledMethods, ['test_pay']);
+
+  const testInstrument2 = await instruments.get('test_key2');
+  assert_equals(testInstrument2.name, 'test_instrument2');
+  assert_array_equals(testInstrument2.enabledMethods, ['basic-card']);
+
+  // PaymentInstruments.get() should resolve `undefined` if no stored key
+  const undefinedInstrument = await instruments.get('undefined_key');
+  assert_equals(undefinedInstrument, undefined);
```
marcoscaceres: Probably want a test that overrides an existing entry. 

Also, we probably want a test to check what equality of pulling the same instrument multiple times out of storage.

Additionally, we probably want to check that these objects are just copies, in that if you change them, they don't change the one in storage. 

After changing one, you have to store it again.

marcoscaceres: Oh, and we probably want to be mean... and add maybe 1024 instruments. And we want to check that the order is kept.

Also, what want to check that setting an existing key retains the order. So, setting key "test_instrument1" to a new instrument should still keep it as the first.

marcoscaceres (changes requested):
Great start on these... suggested a few more tests we should add.

PR #8519, opened 2017-11-30
Changes for early fetching of methods

ricea:
Stream classes are changing to get methods from the underlying object
during construction rather than when the methods are called.

Change expectations of when method lookup happens.

See https://github.com/whatwg/streams/issues/691 for background.

### DIFF
--- a/streams/readable-streams/bad-underlying-sources.js
+++ b/streams/readable-streams/bad-underlying-sources.js
@@ -35,16 +35,14 @@ test(() => {
 }, 'Underlying source start: throwing method');
 
 
-promise_test(t => {
+test(() => {
 
   const theError = new Error('a unique string');
-  const rs = new ReadableStream({
+  assert_throws(theError, () => new ReadableStream({
     get pull() {
       throw theError;
     }
-  });
-
-  return promise_rejects(t, theError, rs.getReader().closed);
+  }), 'constructor should throw');
 
 }, 'Underlying source: throwing pull getter (initial pull)');
 
@@ -82,9 +80,11 @@ promise_test(t => {
 
   return Promise.all([
     reader.read().then(r => {
-      assert_object_equals(r, { value: 'a', done: false }, 'the chunk read should be correct');
+      assert_object_equals(r, { value: 'a', done: false }, 'the first chunk read should be correct');
     }),
-    promise_rejects(t, theError, reader.closed)
+    reader.read().then(r => {
+      assert_object_equals(r, { value: 'a', done: false }, 'the second chunk read should be correct');
+    })
   ]);
 
 }, 'Underlying source pull: throwing getter (second pull)');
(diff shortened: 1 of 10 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21290)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/309549978)

streams/readable-streams/bad-underlying-sources.js
```diff
@@ -82,9 +80,11 @@ promise_test(t => {
...
-      assert_object_equals(r, { value: 'a', done: false }, 'the chunk read should be correct');
+      assert_object_equals(r, { value: 'a', done: false }, 'the first chunk read should be correct');
     }),
-    promise_rejects(t, theError, reader.closed)
+    reader.read().then(r => {
+      assert_object_equals(r, { value: 'a', done: false }, 'the second chunk read should be correct');
+    })
   ]);
 
 }, 'Underlying source pull: throwing getter (second pull)');
```
domenic: This test might benefit from a comment saying that the test is that it's only gotten once. Indeed, asserting that counter === 1 after the last assert_object_equals might be good.

streams/writable-streams/properties.js
```diff
@@ -194,28 +194,32 @@ for (const method in sinkMethods) {
...
+    }
+    const sink = new Proxy({}, handler);
+    const ws = new WritableStream(sink);
+    assert_array_equals(touchedProperties, ['type', 'write', 'close', 'abort', 'start'],
+                        'expected properties should be got');
+    return trigger(ws.getWriter()).then(() => {
+      assert_array_equals(touchedProperties, ['type', 'write', 'close', 'abort', 'start'],
+                          'no properties should be accessed on method call');
+    });
+  }, `unexpected properties should not be accessed when calling sink method ${method}`);
```
domenic: Should we clone this test for readable too, while we're here?

(What a great test, BTW...)

domenic (approved):
Some suggestions, but overall looks good. I think we should add transform stream tests too though, probably before merging.

streams/writable-streams/properties.js
ricea: I've been thinking about throwing away readable-stream's properties.js and use one based on this instead. But I'd have to check carefully that I wasn't losing any coverage, which is why I haven't done it yet.

streams/readable-streams/bad-underlying-sources.js
ricea: Done.

ricea:
Corresponding standard change is https://github.com/whatwg/streams/pull/860 (still WIP at this time).

domenic (approved):
LGTM!

PR #8526, opened 2017-11-30
two corrections to the four font-variant-nn tests

svgeesus:
Firstly, the tests claimed to be testing `font-variant-east-asian: jis04` and `font-feature-settings: "jp04" on`
but used the letter for `font-variant-east-asian: jis90` and `font-feature-settings: "jp90" on`. Now corrected.

Secondly, they were testing the superscript feature but that is a poor choice because the spec allows a lot of wiggle room for superscripts and subscripts, for historical reasons. That made everyone fail the test (separately from the error above). Now changed to use `font-variant-numeric: oldstyle-nums;` and `font-feature-settings: "onum" on` which is a much better test. (So the test letter is Q, instead of J previously. See https://www.w3.org/People/chris/fwf/fancytester.html for explanation of the letter-to-feature mapping in this font).

With those two changes, Firefox and Chrome both pass, so this is no longer at-risk. Win!

### DIFF
--- a/css/css-fonts/font-variant-01.html
+++ b/css/css-fonts/font-variant-01.html
@@ -23,9 +23,9 @@
   }
   .outer {
     font-variant-ligatures: common-ligatures  discretionary-ligatures historical-ligatures contextual;
-    font-variant-position: sub;
+    font-variant-numeric: oldstyle-nums;
     font-variant-caps: small-caps;
-    font-variant-east-asian: jis04;
+    font-variant-east-asian: jis90;
   }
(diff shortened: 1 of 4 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21312)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/309726143)

plehegar (approved):

PR #8537, opened 2017-12-01
Fix data document tests in custom-elements/parser/parser-uses-registry-of-owner-document.html

### LINKED ISSUE #8530: Data document tests in custom-elements/parser/parser-uses-registry-of-owner-document.html doesn't match the spec
See following test in custom-elements/parser/parser-uses-registry-of-owner-document.html,
https://github.com/w3c/web-platform-tests/blob/0b5de1b8bb072bd38896cde82dff0b201904d5ed/custom-elements/parser/parser-uses-registry-of-owner-document.html#L75-L82

From the step 3 of https://html.spec.whatwg.org/#document-write-steps: 
> If document is not an active document, then return

so the `write()` call on `windowlessDocument` would do nothing. And the `instance` would be `null`.

Maybe we should use DOMParser to test data document, instead of using write()?
Something like,
```javascript
var windowlessDocument = (new DOMParser()).parseFromString('<my-custom-element></my-custom-element>', "text/html");
  
var instance = windowlessDocument.querySelector('my-custom-element'); 

assert_true(instance instanceof HTMLElement); 
assert_false(instance instanceof MyCustomElement); 
```

@domenic  @annevk

EdgarChen:
Fix #8530.

### DIFF
--- a/custom-elements/parser/parser-uses-registry-of-owner-document.html
+++ b/custom-elements/parser/parser-uses-registry-of-owner-document.html
@@ -71,17 +71,14 @@ test(function () {
 document.body.removeChild(iframe);
 
 test(function () {
-    var windowlessDocument = document.implementation.createHTMLDocument();
-    windowlessDocument.open();
-    windowlessDocument.write('<my-custom-element></my-custom-element>');
-    windowlessDocument.close();
+    var windowlessDocument = (new DOMParser()).parseFromString('<my-custom-element></my-custom-element>', "text/html");
 
     var instance = windowlessDocument.querySelector('my-custom-element');
 
     assert_true(instance instanceof HTMLElement);
     assert_false(instance instanceof MyCustomElement);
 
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21355)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/310109507)

domenic (approved):
I hope there are tests somewhere else that test that document.write() doesn't work in inactive documents, because apparently it does work in Chrome/Safari.

EdgarChen:
I don't have permission to merge the pull request, could someone help me to merge? Thank you.

TakayoshiKochi:
merged per @domenic's approval.

EdgarChen:
Thank you, @TakayoshiKochi.

PR #8541, opened 2017-12-02
Apply pipes to python handlers

gterzian:
Follow-up on https://github.com/w3c/web-platform-tests/pull/8415#issuecomment-348445942

I only added support for applying pipes to Python script and function handlers(which covers the json handler as well),  for the following reason:
1. That was the use case I encountered myself("apply standard pipes on top of the custom logic contained in the python handler, to avoid writing custom code")
2. I'm currently wondering if there is any value in applying pipes to other handlers like directory and as_is.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8541)

### DIFF
--- a/tools/wptserve/docs/pipes.rst
+++ b/tools/wptserve/docs/pipes.rst
@@ -12,8 +12,7 @@ This would serve bytes 1 to 199, inclusive, of foo.txt with the HTTP status
 code 404.
 
 .. note::
-   Pipes are only applied to static files, and will not work if applied to
-   other types of handlers, such as Python File Handlers.
+   Pipes are only applied to static files, Python file and function handlers, and json handlers.
 
 There are several built-in pipe functions, and it is possible to add
 more using the `@pipe` decorator on a function, if required.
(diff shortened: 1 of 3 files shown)


w3c-bots:
# Build BROKEN

Started: 2017-12-02 04:14:01
Finished: 2017-12-02 04:25:10


<h2>Failing Jobs</h2>
<ul>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/21366)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/310369708)

gsnedders:
@gterzian trickle for example would still be useful on asis at least; I think for simplicity's sake we should just support it everywhere unless there's any real cost associated with doing so

jgraham:
One thing that Python handlers can do that isn't necessarily compatible with pipes is interact with the socket directly. I'm not actually sure if this can cause any issues. Maybe not, but perhaps we should have a test for it. Otherwise I'm happy to take this change, or to add support for `asis`.

gterzian:
I've experimented with applying the pipes to the 'as_is' handler, see the extra commit(to be removed or squashed), and it seems the pipes won't all apply 'as expected' to the response, for example in the case of the 'slice' pipe, since it only applies to the content of a response, it has no effect on the response generated to the as_is handler(without introducing additional logic in the pipeline to handle that response differently). 

The trickle option should work fine however.

jgraham (approved):

PR #8548, opened 2017-12-04
Avoid a serialization interop issue in test_media_queries.html

foolip:
The serialization of the text-decoration shortand is the only difference
between wpt and a copy of the same test in Chromium. It was modified here:
https://chromium.googlesource.com/chromium/src/+/672db831143590b48f931358fd01eab0dc7c453e

Rather than following that change, avoid the difference. It will be
tested separately: https://github.com/w3c/web-platform-tests/pull/8602

### DIFF
--- a/css/mediaqueries/test_media_queries.html
+++ b/css/mediaqueries/test_media_queries.html
@@ -394,7 +394,7 @@ function run() {
 
     function query_applies(q) {
       style.setAttribute("media", q);
-      return body_cs.getPropertyValue("text-decoration") == "underline";
+      return body_cs.getPropertyValue("text-decoration-line") == "underline";
     }
 
     function should_apply(q) {


foolip:
w3c-test:mirror

foolip:
@lukebjerring I promised to stop spamming with with every example, but here's a case where understanding the changes would have been extra useful, since there are so many tests, and reason to suspect that the changes might cause some new failures if the "text-decoration" shorthand isn't implemented everywhere yet.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21396)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/311250372)

foolip:
w3c-test:mirror

foolip:
OK, I manually determined that the change made 97 tests go from pass to fail in Firefox, where previously all were passing, and the opposite happens in Chrome.

I changed the test to avoid depending on this difference entirely, and created a single new test for css-text-decor that'll fail in Firefox. I'm not an expert in this area, so review from a Gecko dev would be good. @upsuper perhaps?

dbaron:
Serializing `text-decoration: underline` as `text-decoration: underline solid rgb(0, 0, 0)` seems to go against two of the core principles of serialization:

* serialize to minimal form
* serialize to the most compatible form

so if there's a spec that calls for that, the spec should probably be fixed.

upsuper:
I agree that `text-decoration` should be serialized to just `underline` when the other two longhands all have their initial value, for backward compatible to CSS2 (since `text-decoration` itself was a longhand at that time).

The current CSSOM spec somehow indicates Chrome's behavior by the following step:
> Return the result of joining values as appropriate according to the grammar of shorthand and terminate these steps. 

I guess it should be changed to allow shorthand properties to define their own serialization behavior, and add some something to CSS Text Decoration for the behavior we want. cc @kojiishi

dbaron:
But isn't that interpretation of the (somewhat unclear) spec incompatible with existing behavior for many shorthands?

foolip:
w3c-test:mirror

foolip:
@dbaron, I've updated this PR to just avoid the problem, and we can discuss https://github.com/w3c/web-platform-tests/pull/8602 separately. r+?

upsuper (approved):

PR #8559, opened 2017-12-04
Fix a couple of self.skipWaiting() Service Worker tests

cdumez:
The tests were timing out in WebKit because the service worker would get terminated before
it is done running the test. The issue is that the promise_test() in the html file sets a
cleanup handler to call unregister(), which effectively terminates the registration's
workers. However, the promise_test() calls fetch_tests_from_worker(sw_registration.active)
at the end which is asynchronous but does not return a promise. As a result, the test
running in the sw_registration.active runs after the promise_test() has completed and
its cleanup handler has been called to unregister.

### DIFF
--- a/service-workers/service-worker/skip-waiting-using-registration.https.html
+++ b/service-workers/service-worker/skip-waiting-using-registration.https.html
@@ -50,9 +50,6 @@ promise_test(function(t) {
         })
       .then(function(registration) {
           sw_registration = registration;
-          t.add_cleanup(function() {
-              registration.unregister();
-            });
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21439)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/311593625)

foolip:
@cdumez if this is still relevant, can you rebase it to resolve conflicts and ping a suitable reviewer?

PR #8565, opened 2017-12-05
TransformStream: a bad readableStrategy size errors the stream

ricea:
Verify that when the size() function of the readableStrategy returns an
invalid value the promise returned from writer.write() will reject on an
identity transform.

A custom transform() method may catch the exception from
controller.enqueue(), causing writer.write() to be fulfilled, but the
stream must still be errored in this case. Add a test for this case also.


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21454)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/311786736)

domenic (approved):

PR #8567, opened 2017-12-05
Update textfieldselection-setSelectionRange.html for platform dependency.

Ms2ger:
The standard explicitly allows falling back to 'forward' if 'none' is not
supported on the platform.


annevk (changes requested):
I think it would be better to store the value once and then require it to be identical later on. We don't want to allow a platform to do either of those depending on what it feels like, right?

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21459)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/311809065)

annevk (approved):
Thanks!

PR #8570, opened 2017-12-05
Check the return value of Get Window Rect in the payload test

gsnedders:

### DIFF
--- a/webdriver/tests/get_window_rect.py
+++ b/webdriver/tests/get_window_rect.py
@@ -151,11 +151,10 @@ def test_payload(session):
     assert response.status == 200
     assert isinstance(response.body["value"], dict)
     value = response.body["value"]
-    assert "width" in value
-    assert "height" in value
-    assert "x" in value
-    assert "y" in value
-    assert isinstance(value["width"], int)
-    assert isinstance(value["height"], int)
-    assert isinstance(value["x"], int)
-    assert isinstance(value["y"], int)
+    expected = session.execute_script("""return {
+         x: window.screenX,
+         y: window.screenY,
+         width: window.outerWidth,
+         height: window.outerHeight
+    }""")
+    assert expected == value


gsnedders:
(I suppose we could get rid of the rest of the asserts without any change in what we're testing.)

webdriver/tests/get_window_rect.py
```diff
@@ -151,11 +151,10 @@ def test_payload(session):
...
     assert isinstance(response.body["value"], dict)
     value = response.body["value"]
-    assert "width" in value
-    assert "height" in value
-    assert "x" in value
-    assert "y" in value
-    assert isinstance(value["width"], int)
-    assert isinstance(value["height"], int)
-    assert isinstance(value["x"], int)
-    assert isinstance(value["y"], int)
```
andreastt: I’m not entirely comfortable removing the explicit assertions of the data integrity, but I guess this is fine.

andreastt (approved):

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21479)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/312024537)

webdriver/tests/get_window_rect.py
gsnedders: In case `window.screenX` doesn't return an integer, for example?

andreastt: Yes, this has known to happen. But I think pytest will catch this when it does a deep dict comparison?

gsnedders: Right, it does compare the entire dict. Obviously, if `window.screenX` and what Get Window Rect return are broken in the same way then it won't catch it now. So maybe we should have these?

andreastt: That would be out of scope of the WebDriver tests, but I wouldn’t object to keeping the assertions around.

andreastt: Just to give some context: Marionette used to return floats for these values.

gsnedders: But if it did, then it wouldn't match `window.screenX`, no? (Assuming a conforming `window.screenX` implementation.)

andreastt: Yes, I would assume so. As I said, happy to land this as-is.

PR #8601, opened 2017-12-06
SW reg tests should assert that registration succeeds rather than fails

aliams:

### DIFF
--- a/service-workers/service-worker/resources/registration-tests-mime-types.js
+++ b/service-workers/service-worker/resources/registration-tests-mime-types.js
@@ -24,19 +24,13 @@ function registration_tests_mime_types(register_method, check_error_types) {
   promise_test(function(t) {
       var script = 'resources/import-mime-type-worker.py';
       var scope = 'resources/scope/no-mime-type-worker/';
-      return promise_rejects(t,
-          check_error_types ? 'SecurityError' : null,
-          register_method(script, {scope: scope}),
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21569)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/312657813)

mkruisselbrink:
I'm not sure I understand the rationale for this change? The spec in https://w3c.github.io/ServiceWorker/#importscripts does seem to say that importScripts without a valid mime type should return a network error, and thus I'd expect registration to fail?

aliams:
According to https://html.spec.whatwg.org/multipage/workers.html#dom-workerglobalscope-importscripts:
1. Fetch a classic worker-imported script given url and settings object, passing along any custom perform the fetch steps provided. If this succeeds, let script be the result. Otherwise, rethrow the exception.
2. Run the classic script script, with the rethrow errors argument set to true.

...as well as here (where it says to check it and calls out the scenario where it doesn’t)
https://html.spec.whatwg.org/multipage/webappapis.html#fetch-a-classic-worker-imported-script:
For historical reasons, fetching a classic script does not include MIME type checking. In contrast, module scripts will fail to load if they are not of a correct MIME type

PR #8612, opened 2017-12-07
Refactor and improve Blob URL tests.

mkruisselbrink:
This moves around some tests, makes some manual tests automated,
de-duplicates some tests and just generally expands the test coverage
of Blob URLs.

Also updates fetch_tests_from_window to work with cross-origin (in this
case unique origin) iframes.

These test changes go together with the spec changes in w3c/FileAPI#92


w3c-bots:
# Build ERRORED

Started: 2017-12-07 14:11:53
Finished: 2017-12-07 15:04:07


<h2>Failing Jobs</h2>
<ul>

<li>lint</li>

<li>firefox:nightly</li>

<li>chrome:unstable</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/21611)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/312970355)

mkruisselbrink:
Seems the testharness.js changes cause it to try to run all the tests, which predictably times out...

FileAPI/url/cross-global-revoke.sub.html
```diff
@@ -0,0 +1,63 @@
+<!doctype html>
+<meta charset="utf-8">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<body>
+<script>
+async_test(t => {
```
bsittler: Might it be more readable to use a "straight-line" async arrow function here instead, and `await` with one-shot handlers, e.g. `const first_message = await new Promise(resolve => self.onmessage = e => e.source == frame.contentWindow && resolve(e.data));` `assert_equals(first_message, 'ready');` and then later `const second_message = await new Promise(resolve => self.onmessage = e => e.source == frame.contentWindow && resolve(e.data));` `assert_equals(second_message, 'revoked')`, rather than `step_func` and nested sync handler scopes?

FileAPI/url/url-with-fetch.any.js
```diff
@@ -0,0 +1,52 @@
...
+  const blob = new Blob([blob_contents]);
+  const url = URL.createObjectURL(blob);
+
+  return fetch_should_succeed(t, url).then(text => {
+    assert_equals(text, blob_contents);
+  });
+
+  // Revoke the object URL. fetch should have already resolved the blob URL.
+  URL.revokeObjectURL(url);
+}, "Revoke blob URL after fetch, will fetch");
```
pwnall: nit: The text name is a bit unclear. As a browser developer, if this test fails and I see the error message, I still have to read the test's code to understand what to fix. The comment seems to plug that problem, so perhaps the comment should be merged with the test name?

FileAPI/url/resources/fetch-tests.js
```diff
@@ -0,0 +1,56 @@
+function fetch_tests(fetch_method, fetch_should_succeed, fetch_should_fail) {
+  const blob_contents = 'test blob contents';
+  const blob = new Blob([blob_contents]);
+
+  promise_test(t => {
+    const url = URL.createObjectURL(blob);
+
+    return fetch_should_succeed(t, url).then(text => {
```
pwnall: `fetch_method` seems to be displayed, but not used when issuing the actual fetch requests.

FileAPI/url/cross-global-revoke.sub.html
```diff
@@ -0,0 +1,63 @@
...
+  const blob_contents = 'test blob contents';
+  const blob = new Blob([blob_contents]);
+  const url = URL.createObjectURL(blob);
+  const frame = document.createElement('iframe');
+  frame.setAttribute('style', 'display:none;');
+
+  self.addEventListener('message', t.step_func(e => {
+    if (e.source != frame.contentWindow) return;
+    if (e.data == 'ready') {
```
pwnall: nit: would prefer `===` to `==` everywhere, so the reader doesn't have to reason about coercion.

FileAPI/url/cross-global-revoke.sub.html
```diff
@@ -0,0 +1,63 @@
...
+<script>
+async_test(t => {
+  const blob_contents = 'test blob contents';
+  const blob = new Blob([blob_contents]);
+  const url = URL.createObjectURL(blob);
+  const frame = document.createElement('iframe');
+  frame.setAttribute('style', 'display:none;');
+
+  self.addEventListener('message', t.step_func(e => {
+    if (e.source != frame.contentWindow) return;
```
pwnall: nit: would prefer `!==` to `!=` everywhere, so the reader doesn't have to reason about type coercion.

pwnall (commented):
This is a significant improvement over the previous version of the tests. Thank you very much!

I noticed one serious issue -- fetch tests claim to use different HTTP methods, but it seems to me that the implementation ignores the method name, which results in having tests only cover `GET`. All the other comments are (many) nits.

FileAPI/url/cross-global-revoke.sub.html
bsittler: Does that actually work with event.source and frame.contentWindow ? I thought it was not recommended to use === to compare proxy objects.

pwnall: Are you asking whether `===` works as opposed to `==`, or whether the equality operator works at all?

bsittler: The former. I know there are some special hacks to make == work, but couldn't remember whether === is guaranteed to consider the objects equal too.

FileAPI/url/cross-global-revoke.sub.html
mkruisselbrink: done

FileAPI/url/url-basics.any.js
```diff
@@ -0,0 +1,63 @@
+const blob = new Blob(['test']);
```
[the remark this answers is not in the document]
mkruisselbrink: done

FileAPI/url/url-basics.any.js
```diff
@@ -0,0 +1,63 @@
...
+test(() => {
+  const url_count = 10000;
+  let list = [];
+
+  for (let i = 0; i <= url_count; ++i)
+    list.push(URL.createObjectURL(blob));
+
+  list.sort();
+
+  for (let i = 0; i < url_count; ++i)
```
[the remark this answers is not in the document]
mkruisselbrink: > nit 1: url_count -> list.length makes the loop slightly more self-contained

note that list.length is url_count+1, not url_count



> nit 2: You could start with i = 1 and compare list[i] against list[i - 1] to avoid accessing the element past the array's end. In this case, the out-of-bounds access doesn't cause problems because you get undefined and it is compared against a string... but I don't think a reader should have to reason through this.

which also means that this isn't a concern, we're not doing any out-of-bounds access.

But yeah, it's a good point that the difference between <= and < in the two loops is kind of subtle, and just having < in both and starting the second loop at 1 would be more intuitive. I just copied what the existing test did....

FileAPI/url/cross-global-revoke.sub.html
bsittler: 👍 Those are good reasons. Thanks for explaining and dispelling my ignorance!

FileAPI/url/cross-global-revoke.sub.html
bsittler: Thanks! Looks like this all got clarified last year or so https://blog.whatwg.org/windowproxy-window-and-location and I missed it. Sorry for the noise!

FileAPI/url/url-format.any.js
```diff
@@ -0,0 +1,63 @@
...
+  let list = [];
+
+  for (let i = 0; i < url_count; ++i)
+    list.push(URL.createObjectURL(blob));
+
+  list.sort();
+
+  for (let i = 1; i < list.length; ++i)
+    assert_not_equals(list[i], list[i-1], 'generated Blob URLs should be unique');
+}, 'Check whether generated Blob URLs are unique');
```
pwnall: nit: In the end, good tests with any naming convention are better than no tests (hence why this is a nit), but I hope you'll at least consider removing the "{Check, Verify} {if, whether} " prefixes. I think that straight out deleting those 2 words in every test description where they're present will still result in a readable description that's a bit smaller.

pwnall (approved):
LGTM w/one nit, thank you for explaining!

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22514)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/326654510)

FileAPI/url/url-format.any.js
mkruisselbrink: Done, thanks for the review!

mkruisselbrink:
Unfortunately some of these tests are flaky in chrome, but that's because of implementation bugs in chrome where I'm working on fixing those bugs. Other than that I plan to land this tomorrow together with the spec changes.

PR #8614, opened 2017-12-07
Remove duplicate copies of subdomains/hostnames

### LINKED ISSUE #6556: Domains hardcoded all over the place
We have:

```
tools/serve/serve.py:              u"www1",
tools/wptrun.py:                          "www1.web-platform.test",
tools/wptrunner/wptrunner/browsers/server-locations.txt:http://www1.web-platform.test:8000
tools/wptrunner/wptrunner/browsers/server-locations.txt:http://www1.web-platform.test:8001
tools/wptrunner/wptrunner/browsers/server-locations.txt:https://www1.web-platform.test:8443
tools/wptrunner/wptrunner/browsers/server-locations.txt:http://www1.web-platform.test:8888
tools/wptrunner/wptrunner/browsers/servodriver.py:127.0.0.1 www1.web-platform.test
tools/wptrunner/wptrunner/environment.py:             "www1.web-platform.test",
tools/wptrunner/wptrunner/executors/executorservo.py:127.0.0.1 www1.web-platform.test
```

I believe `tools/serve/serve.py` is the actual authoritative one for the subdomain, with the others really being the subdomains from `tools/serve/serve.py` combined with `['host']` from `config.json`, and should probably be replaced with that rather than hardcoding them all over the place. (Unless, of course, we want to drop support for using hosts that aren't `web-platform.test`.)

`tools/wptrunner/wptrunner/browsers/server-locations.txt` makes absolutely no sense, given many of the ports are dynamically generated, and I don't see how its single use in `tools/wptrunner/wptrunner/browsers/firefox.py` ensures that they end up on the ports listed in that file? We should at the very least add a comment to that file saying what it is; it's not at all obvious that it relates to some Firefox feature.

gsnedders:
Fixes #6556.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8614)


tools/wptrunner/wptrunner/browsers/firefox.py
```diff
@@ -23,7 +23,6 @@
 from ..executors.executormarionette import (MarionetteTestharnessExecutor,
                                             MarionetteRefTestExecutor,
                                             MarionetteWdspecExecutor)
-from ..environment import hostnames
```
jgraham: This is also used later in the file, right?

jgraham (changes requested):

w3c-bots:
# Build PASSED

Started: 2017-12-07 16:31:46
Finished: 2017-12-07 16:43:04





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21617)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/313046939)

tools/wptrunner/wptrunner/browsers/firefox.py
gsnedders: Apparently so. But it passed CI despite that because we disable the unused variable pyflake still. :(

w3c-bots:
# Build BROKEN

Started: 2018-01-23 22:20:58
Finished: 2018-01-23 22:38:07


<h2>Failing Jobs</h2>
<ul>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/23187)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/332527055)

tools/wptserve/wptserve/pipes.py
```diff
@@ -392,6 +392,11 @@ def config_replacement(match):
             value = request.headers
         elif field == "GET":
             value = FirstWrapper(request.GET)
+        elif field == "domains":
```
gsnedders: I really don't like this, but I don't think we can reasonably merge `domains` and `not_domains` (given we need to know what's meant to resolve when setting up things like hosts files), and I think we want it like this to resolve #9232.

@jgraham @Ms2ger halp.

Ms2ger: I think this is probably fine. Maybe write as:

```python
request.server.config.get('not_domains', request.server.config['domains'])
```

(if that's a real `dict`

jgraham: This seems fine to me too.

tools/wptrunner/wptrunner/tests/browsers/test_firefox.py
```diff
@@ -0,0 +1,30 @@
...
+structured.set_default_logger(logger)
+
+
+def test_server_location():
+    browser = firefox.FirefoxBrowser(logger,
+                                     "foo/firefox",
+                                      "foo/prefs",
+                                      "testharness.js",
+                                      config={
+                                          "ports": {"http": [1234]},
```
jgraham: This test doesn't seem that useful. It ensures that you write something to some file, but it doesn't really check that the contents are right. It seems more useful to set things up so you can check you get the exact content you expect compared against the file you removed. Obviously this would then have to be changed when new servers are added or removed, but it would have the advantage of actually checking that something functional is produced.

jgraham (changes requested):
Could you test this by pushing some representative test changes to this branch and ensuring that you still get meaningful results?

tools/wptrunner/wptrunner/browsers/firefox.py
```diff
@@ -173,6 +192,8 @@ def __init__(self, logger, binary, prefs_root, test_type, extra_prefs=None, debu
         self.stylo_threads = stylo_threads
         self.chaos_mode_flags = chaos_mode_flags
 
+        self.server_locations = write_server_locations_file(self.config)
```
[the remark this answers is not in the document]
gsnedders: I was just copying what we did with Servo for the hosts file, FWIW.

`tempfile.NamedTemporaryFile` doesn't work because there's no guarantee whether you can reopen the file a second time. Per the docs:

> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later).

I'd somehow missed the `cleanup` function before, bleh.

gsnedders:
Very oddly, this doesn't fail locally like it does on Travis. o_O

jgraham (approved):

gsnedders:
Fixes #9232.

jgraham (approved):

PR #8636, opened 2017-12-08
Add tests for slot.assignedElements method

TimvdLippe:
This is the follow up of #4541 which became stale after comments regarding code duplication. The tests in `shadow-dom/HTMLSlotElement-interface.html` are deduplicated. The tests in the other two files are not deduplicated per [this comment](https://github.com/w3c/web-platform-tests/pull/4541#issuecomment-274730713). As I am eagerly waiting for this function, I went ahead and published this updated PR.

Supersedes and closes #4541
Ref https://github.com/whatwg/html/pull/2269

### DIFF
--- a/interfaces/html.idl
+++ b/interfaces/html.idl
@@ -1130,6 +1130,7 @@ interface HTMLTemplateElement : HTMLElement {
 interface HTMLSlotElement : HTMLElement {
   [CEReactions] attribute DOMString name;
   sequence<Node> assignedNodes(optional AssignedNodesOptions options);
+  sequence<Element> assignedElements(optional AssignedNodesOptions options);
 };
 
 dictionary AssignedNodesOptions {
--- a/shadow-dom/HTMLSlotElement-interface.html
+++ b/shadow-dom/HTMLSlotElement-interface.html
@@ -12,6 +12,8 @@
 <div id="log"></div>
 <script>
 
+const slotMethods = ['assignedNodes', 'assignedElements'];
+
 test(function () {
     assert_true('HTMLSlotElement' in window, 'HTMLSlotElement must be defined on window');
     assert_equals(HTMLSlotElement.prototype.__proto__, HTMLElement.prototype, 'HTMLSlotElement should inherit from HTMLElement');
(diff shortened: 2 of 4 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21684)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/313678827)

hayatoito:
Thanks, however, I am afraid that this PR is still a kind of over-engineering.
We don't need to add a test for assignedElements in *every* tests where assignedNodes is being used.
I guess this PR is trying to add a test for assignedElement everywhere assingedNode is being used, right? If so, please re-consider that, and try to add a minimum number of tests which would catch the different behavior of assignedNodes and assignedElement. That's the most important. I can't find this most important test easily in this PR.

----

Suppose that someone sends a PR which adds a huge number of tests for Node.parentElement in *every* tests where Node.parentNode is using used, 

> ["parentNode", "parentElement"].forEach(prop => {
>   test(...);
>
>   ...
> }

We would object to the PR because it wouldn't make sense in almost every cases. I'm afraid that the same thing is happening in this PR.

----

> ["assignedNodes", "assignedElements"].forEach(prop => {
>   test(...);
>
>   ...
> }

This patter is useful, but we don't want to apply this pattern everywhere only because assignedNodes appears in a test. You might want to select a representative case from existing tests.

In a summary, my recommendation:

- Don't modify existing tests only because assignedNodes() is being used there. 
- Add test cases where assignedNodes() and assignedElements() would return the different results.
- Pick a reprentative case where we might want to apply `["assignedNodes", "assignedElements"].forEach(prop => ...`  We don't need to confirm 'assignedNodes' and 'assignedElements' returns the same result in every tests.

TimvdLippe:
@hayatoito Thanks a lot for your review, that explains a lot :smile: I have updated this PR to (to my best knowledge) only include tests for the necessary cases. It covers some basic examples and also 2 complex cases. PTAL

hayatoito (approved):
Thank you for updating tests. Looks good to me!

PR #8669, opened 2017-12-12
FetchEvent.prototype.isReload is no longer part of the specification

cdumez:
Drop it from interfaces-worker.sub.js.

### DIFF
--- a/service-workers/service-worker/resources/interfaces-worker.sub.js
+++ b/service-workers/service-worker/resources/interfaces-worker.sub.js
@@ -57,21 +57,20 @@ test(function() {
       new FetchEvent('FetchEvent', {request: req}).clientId,
       '', 'Default FetchEvent.clientId should be the empty string');
     assert_equals(
-      new FetchEvent('FetchEvent', {request: req}).isReload,
-      false, 'Default FetchEvent.isReload should be false');
-    assert_equals(
       new FetchEvent('FetchEvent', {request: req, cancelable: false}).cancelable,
       false, 'FetchEvent.cancelable should be false');
     assert_equals(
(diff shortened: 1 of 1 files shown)


cdumez:
cc @beidson

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21772)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/315466252)

jdm:
Could we add a test that it does not exist? We do this for other legacy parts of specifications.

jdm (approved):

wanderview:
FYI there are existing SW spec issues about this open.  It was removed from the spec but not removed from implementations because there are not good alternatives yet.  We are trying to define a replacement in w3c/ServiceWorker#1167.  We had tentatively agreed in F2F meetings to leave this in WPT until the new spec feature was defined.

cdumez:
My understanding is that WPT is supposed to match the spec, not browsers. If browsers do not match the spec, then this is an issue with the spec.

wanderview:
Sure, but if you see the WPT passes in every shipping implementation, but doesn't match the spec, maybe it would be worth filing a spec issue.

Same reason I haven't written a WPT checking that `document.open()` creates a new window.  Currently firefox and the spec agree, but no other implementation does it.  We have a spec issue open instead.

wanderview:
And I'll just note I'm not really upset.  It sucks that spec stuff was changed without implementation happening.  I've been encouraging folks to either put the spec back or actually make the change.  I've also tried to give warning about known stuff like this to edge/webkit implementation folks.

cdumez:
In this case though, it *was* in an earlier version of the spec and was dropped in a more recent version of the spec. Therefore, my understanding was that the test was outdated and thus did not file an issue with the spec.

PR #8683, opened 2017-12-14
Add checkpoints to browser history traversal tests

xiuqijix:
According to following sentence in [spec](https://www.w3.org/TR/html51/browsers.html#the-popstateevent-interface), add a PopStateEvent.status initial value test.

>The state attribute must return the value it was initialized to. When the object is created, this attribute must be initialized to null.

According to following sentence in [spec](https://www.w3.org/TR/html51/browsers.html#history-traversal), add some hashchange event checkpoints.

 >If hash changed is true, then fire a trusted event with the name hashchange at the browsing context’s Window object, using the HashChangeEvent interface, with the oldURL attribute initialized to old URL and the newURL attribute initialized to new URL. This event must bubble but not be cancelable and has no default action.

### DIFF
--- a/html/browsers/browsing-the-web/history-traversal/PopStateEvent.html
+++ b/html/browsers/browsing-the-web/history-traversal/PopStateEvent.html
@@ -11,6 +11,11 @@ test(function () {
 }, 'initPopStateEvent');
 
 test(function () {
+  var popStateEvent = new PopStateEvent("popstate");
+  assert_equals(popStateEvent.state, null, "the PopStateEvent.state");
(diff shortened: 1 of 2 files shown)


kaixinjxq:
@Ms2ger, PTAL.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21823)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/316226902)

Ms2ger (approved):

PR #8698, opened 2017-12-15
Add HTMLScriptElement defer test

xiuqijix:

### DIFF
--- a/html/semantics/scripting-1/the-script-element/defer.js
+++ b/html/semantics/scripting-1/the-script-element/defer.js
@@ -1,0 +1,7 @@
+function afterLoad(){
+  flag += NUMBER_SECOND;
+  test(function(){
+    assert_equals(flag, EXPECTED, "The value of variable is incorrect.");
+  }, "the defer script run later");
+}
+afterLoad();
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21867)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/316797160)

Ms2ger (changes requested):
I made the following comment on critic:

Ms2ger at 2015-04-09 17:49 +02:00:
  Right now, if the deferred script runs after the load event fires, the test in
  it will be ignored. I suggest to instead use one async_test() and step in both
  scripts.

  I also think it would be clearer to use two boolean variables, say
  "ran_deferred_script" and "ran_inline_script".

kaixinjxq:
Thanks @Ms2ger, update the test with async_test and I think a 'ran_deferred_script' boolean variable should be enough, PTAL.

Ms2ger (approved):
I suppose that works too, but "ran_deferred_script" is a pretty poor name for what it does now. Possibly something generic like "script_run_status"?

kaixinjxq:
@Ms2ger, update the test with a 'script_run_status' string variable, PTAL.

Ms2ger:
Thanks, that looks a lot more readable.

PR #8704, opened 2017-12-15
encoding: Stop expecting '&' and '#' to be percent-encoded in queries.

rakuco:
According to
https://github.com/whatwg/url/commit/bbeacfe00317554daab0ece983f42692d42487fc ("Revamp
the way application/x-www-form-urlencoded is defined. Change query state
along with it as it had the same bug") and later, the basic URL parser, when
in the query state and EOF is reached, the following should be done to each
byte in `buffer`:

1. If byte is less than 0x21 (!), greater than 0x7E (~), or is 0x22 ("),
   0x23 (#), 0x3C (<), or 0x3E (>), append byte, percent encoded, to url’s
   query.

2. Otherwise, append a code point whose value is byte to url’s query.

'&' is U+0026 and '#' is U+003B, so they should not be percent-encoded at all.

### DIFF
--- a/encoding/big5-encoder.html
+++ b/encoding/big5-encoder.html
@@ -15,8 +15,8 @@
 
  encode("ab", "ab", "very basic")
  // edge cases
- encode("\u9EA6", "%26%2340614%3B", "Highest-pointer BMP character excluded from encoder");
- encode("\uD858\uDE6B", "%26%23156267%3B", "Highest-pointer character excluded from encoder");
+ encode("\u9EA6", "&%2340614;", "Highest-pointer BMP character excluded from encoder");
+ encode("\uD858\uDE6B", "&%23156267;", "Highest-pointer character excluded from encoder");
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21877)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/316966432)

rakuco:
> Failing Jobs
> * chrome:unstable
> * safari:10.0
> * MicrosoftEdge:14.14393

Firefox will start passing these tests I'm changing, and I'm working on fixing Blink to do the same (the fix is the same one we need for the second test in https://github.com/w3c/web-platform-tests/commit/bdadec77e585d63997c1d061b9fd2629ef7cdecd to pass).

inexorabletash (approved):

PR #7684, opened 2017-10-11
More dynamic import() string compilation tests

domenic:
More dynamic import() string compilation tests

Follows https://github.com/whatwg/html/pull/3117. Parser state is implicitly tested because lots of tests would otherwise fail.

Referrer policy tests omitted for now since you can only set that via modulepreload at the moment.

Issue #6075, opened 2017-05-29
testharness.js: extend `Test#add_cleanup` to honor Promises


jugglinmike:
Upon completion, some tests must use asynchronous operations to properly
restore global state. This is the case in many service worker tests, for
instance, where `ServiceWorkerRegistration#unregister` must be completed before
the test is truly "done."

In its current form, `Test#add_cleanup` is unusable for this purpose because it
assumes synchronous behavior.

Extend that method to inspect the return value of the provided callback
function. If the value is a `thenable`, the test should not be considered
"complete" until that thenable settles (along with all additional thenables
specified in this way). Additional considerations:

- Waiting should occur regardless of the sub-test's result.
- All Promises should be allowed to settle, making the semantics of the
  standard `Promise.all` method (where any one rejection prompts overall
  rejection) inappropriate
- Waiting should be subject to the same timeout restrictions as the test body
  itself.
- Rejected promises should cause test failures, but if the test body itself has
  failed, the test error should take precedence.
- If practical, the implementation should not introduce a dependency on the
  global Promise constructor or its methods. Barring that, references to those
  APIs should only be evaluated after a callback function has returned a
  thenable.

foolip:
@jugglinmike you did some work in https://github.com/w3c/web-platform-tests/pull/6151 but I guess there's more to be done here?

foolip:
In https://github.com/w3c/web-platform-tests/issues/7188#issuecomment-341720370 @jakearchibald said
> This would be really useful for the service worker tests. In some cases we're meaningless promise_tests to do this.

Can you link to some such cases? That would be a reason to bump the priority of this I think.

mkruisselbrink:
> Can you link to some such cases? That would be a reason to bump the priority of this I think.

https://github.com/w3c/web-platform-tests/blob/master/service-workers/service-worker/indexeddb.https.html is one such case (see the two nested "clean up" promise_tests inside the last test in that file).

foolip:
OK, I'll call it roadmap, the goal being to change https://github.com/w3c/web-platform-tests/blob/master/service-workers/service-worker/indexeddb.https.html to use `add_cleanup` instead. That should drive whatever else needs to change.

pwnall:
More test case data, to help prioritize:

1) This would help all the tests for the Async Cookies API.
[...]

PR #8735, opened 2017-12-19
[wptserve] Add on-demand file hash computation

jugglinmike:
Certain tests require that all HTML `<script>` tags be declared with an
`integrity` attribute which specifies the cryptographic hash of the file
to be loaded. These tests were initially implemented with a static value
for the hash of the test harness files located in the shared "resources"
directory.

In order to support continued development of those shared files, the
tests should be authored in a way that is tolerant of such
modifications.

Introduce a new templating capability to the `wptserve` utility which
allows markup to be rendered with dynamically-computed hash values.
Allowing the shared resources to change independently of the tests that
rely on them.

---

Hi @gsnedders,

I'm working on an enhancement to `testharness.js` (see gh-6075), and I found
that my latest patch broke a few tests (specifically, those introduced via
gh-7684). I was surprised to find that the error had nothing to do with the
semantics of my change--just the fact that I had made any change at all.

This is an initial attempt to facilitate continued development of
`testharness.js`. It's pretty rough because it's based on an afternoon's review
of the `wptserve` internals. I'm requesting your review because I believe you
may have some vision for the overall direction of WPT's server. If you think
there is someone better to give direction, then I'm all ears!

(@annevk and @domenic: my hope is that if this lands, future tests that require
strict file integrity checking can be a little more resilient to change.)

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8735)

### DIFF
--- a/html/semantics/scripting-1/the-script-element/module/dynamic-import/string-compilation-integrity-classic.sub.html
+++ b/html/semantics/scripting-1/the-script-element/module/dynamic-import/string-compilation-integrity-classic.sub.html
@@ -1,15 +1,15 @@
 <!DOCTYPE html>
 <meta charset="utf-8">
-<title>import() doesn't have any integrity metadata when initiated by compiled strings inside a module script</title>
+<title>import() doesn't have any integrity metadata when initiated by compiled strings inside a classic script</title>
 <link rel="author" title="Domenic Denicola" href="mailto:d@domenic.me">
 <meta http-equiv="Content-Security-Policy" content="require-sri-for script">
 
-<script src="/resources/testharness.js" integrity="sha384-4Nybydhnr3tOpv1yrTkDxu3RFpnxWAxlU5kGn7c8ebKvh1iUdfVMjqP6jf0dacrV"></script>
-<script src="/resources/testharnessreport.js" integrity="sha384-GOnHxuyo+nnsFAe4enY+RAl4/+w5NPMJPCQiDroTjxtR7ndRz7Uan8vNbM2qWKmU"></script>
+<script src="/resources/testharness.js" integrity="sha384-{{file_hash(sha384, resources/testharness.js)}}"></script>
+<script src="/resources/testharnessreport.js" integrity="sha384-{{file_hash(sha384, resources/testharnessreport.js)}}"></script>
 
 <div id="dummy"></div>
 
-<script type="module">
+<script>
 function createTestPromise() {
   return new Promise((resolve, reject) => {
     window.continueTest = resolve;
--- a/html/semantics/scripting-1/the-script-element/module/dynamic-import/string-compilation-integrity-module.sub.html
+++ b/html/semantics/scripting-1/the-script-element/module/dynamic-import/string-compilation-integrity-module.sub.html
@@ -1,15 +1,15 @@
 <!DOCTYPE html>
 <meta charset="utf-8">
-<title>import() doesn't have any integrity metadata when initiated by compiled strings inside a classic script</title>
+<title>import() doesn't have any integrity metadata when initiated by compiled strings inside a module script</title>
 <link rel="author" title="Domenic Denicola" href="mailto:d@domenic.me">
 <meta http-equiv="Content-Security-Policy" content="require-sri-for script">
 
-<script src="/resources/testharness.js" integrity="sha384-4Nybydhnr3tOpv1yrTkDxu3RFpnxWAxlU5kGn7c8ebKvh1iUdfVMjqP6jf0dacrV"></script>
-<script src="/resources/testharnessreport.js" integrity="sha384-GOnHxuyo+nnsFAe4enY+RAl4/+w5NPMJPCQiDroTjxtR7ndRz7Uan8vNbM2qWKmU"></script>
+<script src="/resources/testharness.js" integrity="sha384-{{file_hash(sha384, resources/testharness.js)}}"></script>
+<script src="/resources/testharnessreport.js" integrity="sha384-{{file_hash(sha384, resources/testharnessreport.js)}}"></script>
 
 <div id="dummy"></div>
 
-<script>
+<script type="module">
 function createTestPromise() {
   return new Promise((resolve, reject) => {
     window.continueTest = resolve;
--- a/tools/wptserve/tests/functional/docroot/sub_file_hash.sub.txt
+++ b/tools/wptserve/tests/functional/docroot/sub_file_hash.sub.txt
@@ -1,0 +1,6 @@
+md5: {{file_hash(md5, sub_file_hash_subject.txt)}}
+sha1: {{file_hash(sha1, sub_file_hash_subject.txt)}}
+sha224: {{file_hash(sha224, sub_file_hash_subject.txt)}}
+sha256: {{file_hash(sha256, sub_file_hash_subject.txt)}}
+sha384: {{file_hash(sha384, sub_file_hash_subject.txt)}}
+sha512: {{file_hash(sha512, sub_file_hash_subject.txt)}}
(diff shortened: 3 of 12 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21963)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/318412550)

jgraham:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[tools/wptserve/wptserve/pipes.py, line 376 at r1](https://reviewable.io:443/reviews/w3c/web-platform-tests/8735#-L0iVoX9mdRe_QLXANRT:-L0iVoXAHXnY0vMyv73y:b-4elcr4) ([raw file](https://github.com/w3c/web-platform-tests/blob/426c172a8010326c2ae3514d63262ec8c9258304/tools/wptserve/wptserve/pipes.py#L376)):*
> ```Python
>         self.doc_root = doc_root
> 
>     def __getitem__(self, name):
> ```

I really think this should look more like a function call. I know that will make the patch bigger and mean a change to the tokenizer and the uuid function, but the current setup looks pretty odd.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/web-platform-tests/8735)*

jugglinmike:
@jgraham That sounds good to me. Do you think we could do that in a follow-up patch, or do you consider it a prerequisite for this fix?

jugglinmike:
(I'm asking because I'd like to limit the prerequisites to introducing the new `testharness.js` feature referenced in the pull request description.)

html/semantics/scripting-1/the-script-element/module/dynamic-import/string-compilation-integrity-classic.sub.html
```diff
@@ -4,8 +4,8 @@
 <link rel="author" title="Domenic Denicola" href="mailto:d@domenic.me">
 <meta http-equiv="Content-Security-Policy" content="require-sri-for script">
 
-<script src="/resources/testharness.js" integrity="sha384-4Nybydhnr3tOpv1yrTkDxu3RFpnxWAxlU5kGn7c8ebKvh1iUdfVMjqP6jf0dacrV"></script>
-<script src="/resources/testharnessreport.js" integrity="sha384-GOnHxuyo+nnsFAe4enY+RAl4/+w5NPMJPCQiDroTjxtR7ndRz7Uan8vNbM2qWKmU"></script>
+<script src="/resources/testharness.js" integrity="sha384-{{sha384[resources/testharness.js]}}"></script>
```
foolip: Rather than using testharness.js, could we not add another file that doesn't ever change?

foolip: @domenic, WDYT?

annevk: I think you misunderstand the test. If you have a CSP stating `require-sri-for script` all scripts in the document need to have integrity defined. That includes the harness scripts.

jugglinmike:
I've re-implemented the functionality on top of a slightly modified template parsing engine. I've also expanded the feature to support all the hashing algorithms that Python provides.

The surrounding code was missing some test coverage, so I added tests for that to help give us confidence that there are no regressions here. For the sake of clarity, I'd prefer it if we could merge this work using distinct commits like the ones submitted here.

@jgraham Does this look better to you?

([The original version of this branch is available here.](https://github.com/w3c/web-platform-tests/compare/master...bocoup:compute-file-hash-orig))

jgraham:
Review status: 0 of 10 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[tools/wptserve/wptserve/pipes.py, line 306 at r3](https://reviewable.io:443/reviews/w3c/web-platform-tests/8735#-L2LA9nVoy2tw6dDSobM:-L2LA9nVoy2tw6dDSobN:bk0t33o) ([raw file](https://github.com/w3c/web-platform-tests/blob/97b36268b0f600ed9b5653f78dcbdee18f5af555/tools/wptserve/wptserve/pipes.py#L306)):*
> ```Python
>                           (r"\$?\w+", ident),
>                           (r"\[[^\]]*\]", index),
>                           (r"\(.*\)", arguments)])
> ```

This is a greedy regex I think, so it's going to break for cases like `foo(bar)baz(foobar)`. Changing it to `r"\([^\)]*\)" seems like it ought to fix the problem.

---

*[tools/wptserve/wptserve/pipes.py, line 393 at r3](https://reviewable.io:443/reviews/w3c/web-platform-tests/8735#-L2LAqsiA2M5k7blLKqP:-L2LAqsiA2M5k7blLKqQ:bpz04sy) ([raw file](https://github.com/w3c/web-platform-tests/blob/97b36268b0f600ed9b5653f78dcbdee18f5af555/tools/wptserve/wptserve/pipes.py#L393)):*
> ```Python
>     @staticmethod
>     def file_hash(request, algorithm, path):
>         assert algorithm in SubFunctions.supported_algorithms
> ```

A better error message here would be nice.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/web-platform-tests/8735)*

jgraham:
This looks very good, just a couple of minor things to fix before merge.

jugglinmike:
> Changing it to `r"([^)]*)" seems like it ought to fix the problem.

Sounds good. I've made the change, but I can't think of a way to test that, though. Any ideas?

> A better error message here would be nice.

You got it.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22811)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/329184919)

jgraham:
Reviewed 2 of 5 files at r1, 4 of 10 files at r2, 6 of 8 files at r3, 1 of 2 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[tools/wptserve/wptserve/pipes.py, line 394 at r5](https://reviewable.io:443/reviews/w3c/web-platform-tests/8735#-L6X2vyWmQdtxZH4rqfC:-L6X2vyYj2u1D1N_YAil:b-iq0jii) ([raw file](https://github.com/w3c/web-platform-tests/blob/e5d062b00b2de5375085fa9c6958ad8564a88d35/tools/wptserve/wptserve/pipes.py#L394)):*
> ```Python
>     @staticmethod
>     def file_hash(request, algorithm, path):
>         assert algorithm in SubFunctions.supported_algorithms, (
> ```

Not a blocker, but in general using `assert` in this case isn't ideal because there is a python interpreter option to strip  asserts. So `if algorithm not in supported_algorithms` seems preferable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/web-platform-tests/8735)*

jgraham (approved):

jgraham:
@jugglinmike Want to rebase this to reorganise the commits and then land it?

jugglinmike:
Sure thing, James. While implementing your last bit of feedback about assertions, I realized that that functionality wasn't tested. So in addition to explicitly raising a `ValueError`, this patch now also contains another functional test which verifies that the server responds with an HTTP error in this case.

The prior version of this change set is available on the following branch:

https://github.com/w3c/web-platform-tests/compare/master...bocoup:compute-file-hash-prebase?expand=1

jugglinmike:
I also had the pleasant surprise of failing a new test that didn't exist when I wrote this patch. Thanks, @gsnedders :) Updated that test accordingly.

jgraham:
Reviewed 4 of 4 files at r6.
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/w3c/web-platform-tests/8735)*

PR #8742, opened 2017-12-19
Do not return value from "cleanup" functions

jugglinmike:
Today, the return value of functions provided to the global
`add_cleanup` function has no effect on the behavior of the test runner.
An upcoming feature addition to `testharness.js` will cause the return
value to influence test results [1].

Despite this, some existing tests have already been authored to return a
value. Introducing the new test harness functionality may trigger
unexpected results in those contexts. To avoid this, refactor existing
tests to avoid returning a value.

[1] https://github.com/w3c/web-platform-tests/issues/6075

---

The Service Worker test I've modified returns a "thenable" which is actually
amenable to the proposed `testharness.js` extension. We could leave it as-is,
and when/if an implementation for gh-6075 lands, that test would become less
fragile automatically. Still, it seems safer to introduce the new functionality
with the confidence that it doesn't trigger *any* implicit opt-in, and then
incrementally begin to use it with intentional follow-up commits. (And for the
Service Worker test suite, this will entail updating many files besides the one
modified here.)

For those curious: I located these tests programatically by [modifying the test
runner](https://github.com/w3c/web-platform-tests/compare/w3c:de65155...bocoup:8fe4fe1) and using that to run all tests that reference `add_cleanup`:

    grep 'add_cleanup' . -rl --include '*.js' --include '*.html' | xargs ./wpt run chrome

And

    grep 'add_cleanup' . -rl --include '*.js' --include '*.html' | xargs ./wpt run firefox

Both browsers identified the same usages.

### DIFF
--- a/service-workers/service-worker/postmessage.https.html
+++ b/service-workers/service-worker/postmessage.https.html
@@ -13,7 +13,7 @@ promise_test(t => {
 
     return service_worker_unregister_and_register(t, script, scope)
       .then(r => {
-          t.add_cleanup(() => r.unregister());
+          t.add_cleanup(() => { r.unregister(); });
           registration = r;
           worker = registration.installing;
 
@@ -62,7 +62,7 @@ promise_test(t => {
 
     return service_worker_unregister_and_register(t, script, scope)
       .then(r => {
-          t.add_cleanup(() => r.unregister());
+          t.add_cleanup(() => { r.unregister(); });
 
           var ab = text_encoder.encode(message);
           assert_equals(ab.byteLength, message.length);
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21997)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/318836119)

service-workers/service-worker/postmessage.https.html
```diff
@@ -13,7 +13,7 @@
 
     return service_worker_unregister_and_register(t, script, scope)
       .then(r => {
-          t.add_cleanup(() => r.unregister());
+          t.add_cleanup(() => { r.unregister(); });
```
foolip: Can you add a TODO here to revert it back when promise cleanup has landed?

foolip (approved):
LGTM % tracking the remaining work.

service-workers/service-worker/postmessage.https.html
jugglinmike: Sure thing

foolip (approved):

PR #8743, opened 2017-12-19
Fix windowclient-navigate.https.html to expect success for uncontroll…

wanderview:
…ed clients.

The test uses `Clients.matchAll({ includeUncontrolled: true })` and
`Client.navigate()` does not care if the window is controlled.  The
navigate algorithm only checks if the final window is cross-origin.

Its unclear to me why chrome expects these navigations to fail to return a Client, but I don't see anything in the spec matching that behavior.

### DIFF
--- a/service-workers/service-worker/windowclient-navigate.https.html
+++ b/service-workers/service-worker/windowclient-navigate.https.html
@@ -44,7 +44,6 @@ navigate_test({
     src_url: '/common/blank.html',
     dest_url: 'blank.html?navigate',
     expected: normalizeURL(SCOPE) + '?navigate',
-    expected: 'TypeError',
   });
 
 navigate_test({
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/21999)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/318851388)

jungkees:
@wanderview, [`WindowClient.navigate()` step 4](https://w3c.github.io/ServiceWorker/#client-navigate) checks if the client's controlled by the service worker the API's called in. I don't remember if we posed any issue about this behavior.

wanderview:
> 

@wanderview, WindowClient.navigate() step 4  checks if the client's controlled by the service worker the API's  called in. I don't remember if we posed any issue about this behavior.

Oh, good catch.  I was not aware of that.  I was only looking at the type resolved by the navigate promise.

It is a bit odd, though, since you can in theory just postMessage() the client and have it change window.location.

wanderview:
I guess I will open a spec issue.

PR #8597, opened 2017-12-06
[testharness.js] Formalize test types

jugglinmike:
Previously, tests for the testharness.js framework were validated
according to an implicit heuristic based on the presence of a `<script>`
tag.

Introduce a declarative configuration to make this distinction explicit.
This pattern makes the validation behavior easier to understand. It also
enables future additions which further differentiate the way the two
test types are interpreted.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8597)

PR #8748, opened 2017-12-20
 [testharness.js] Honor promises from cleanup fns 

### LINKED ISSUE #6075: testharness.js: extend `Test#add_cleanup` to honor Promises

jugglinmike:
Extend the implementation of the `Test#add_cleanup` method to detect
cases where the provided function returns a "thenable" object. In such
cases, defer test completion until all such eventual values settle. If
any eventual value is rejected, report a harness error (as this
represents an unpredictable state that may invalidate subsequent tests).

This behavior prevents tests from being run synchronously. In order to
promote predictability in test harness behavior (and to avoid
invalidating existing tests which rely on synchronicity), only expose
the new behavior for tests defined using the `promise_test` API (as
these tests were originally implemented to run in separate turns of the
event loop).

Introduce `Tests#set_status` in order to de-couple status setting from
remote context closing. This is necessary because uncaught errors may
require asynchronous cleanup, making it inappropriate to synchronously
close communication channels with remote contexts in such cases.

Ensure that this implementation does not introduce a new dependency on
the global `Promise` constructor in order to avoid interfering with
tests that do not require it (as doing so would invalidate those tests
in contexts where `Promise` is not available).

---

The return value of the function provided to `Test#add_cleanup` is only
meaningful for tests created with the `promise_test` API, and in those
cases, only when it is a "thenable" object. If a cleanup function
returns an invalid value, this may reflect a programming error that will
silently degrade test stability.

Detect cases where an invalid value is returned and trigger immediate
harness failure, helping contributors identify mistakes.

---

This patch is intended to resolve gh-6075. Since filing that issue, I've
identified a few new details that warrant mentioning:

- **This feature is limited to tests created with the `promise_test` API.**
  Asynchronous cleanup logic could invalidate tests that were designed to run
  in a fully-synchronous manner. While a careful implementation could allow
  synchronous tests to "opt in" to the new behavior, APIs that are not
  consistently synchronous or asynchronous are [notoriously difficult to reason
  about](http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony).
- **Cleanup functions are invoked in parallel.** We could investigate running
  them in series, but I think the proper order in that case would be "first-in,
  last-out." Given that the fully-synchronous behavior in `master` today is
  "first-in, first-out," this could invalidate tests. So for now, I've settled
  on addressing this with documentation (suggesting that consumers introduce
  their own scheduling mechanism in cases where ordering is important).
- **Inappropriate return values trigger harness error.** Because the API is
  more implicit than a function call, it's possible that test authors may
  unintentionally provide inappropriate values. This kind of error may not be
  immediately obvious since it will introduce race conditions that may only
  occasionally influence test behavior. In order to highlight the mistakes (and
  avoid a source of instability), this patch triggers a harness error whenever
  a cleanup function returns an inappropriate value.

After a brief investigation into the current usage of `Test#add_cleanup` in WPT
today, I've identified two files that return a value. I've opened gh-8742 to
correct that. As noted there, if the maintainers would like to use this new
functionality, I think it's safer to allow them to do so intentionally (rather
than implicitly changing it out from underneath them).

This patch avoids introducing a dependency on the global `Promise` constructor
for any tests that have not "opted in" to this behavior. I've verified this
locally using an extended version of the test harness for `testharness.js`, and
I intend to submit that for review separately (since it relies on a more
generic improvement--see gh-8597).

Also note that as of yesterday, any change to the contents of `testharness.js`
will invalidate a few web platform tests. I've submitted gh-8735 to address
this. If we find that we'd rather land this patch before that one, then I will
update the file hashes which are hard-coded into those test files (as a final
step of the review process).

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8748)


w3c-bots:
# Build PASSED

Started: 2017-12-20 02:08:38
Finished: 2017-12-20 02:19:10


<h2>Failing Jobs</h2>
<ul>

<li>chrome:unstable</li>

<li>safari:10.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/22011)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/318949132)

foolip:
I'm trying the changes in https://chromium-review.googlesource.com/c/chromium/src/+/836619 since our CI here isn't able to handle changes this big.

resources/testharness.js
```diff
@@ -1621,29 +1632,99 @@ policies and contribution forms [3].
...
+                            bad_value_count += 1;
+                        }
+                    });
+            cleanup_done(this_obj, error_count, bad_value_count);
+        } else {
+            all_async(results,
+                      function(result, done) {
+                          if (result && typeof result.then === "function") {
+                              result
+                                  .then(null, on_error)
```
foolip: Newbie question, what's the difference between this and `.then(done, on_error)`?

resources/testharness.js
```diff
@@ -1621,29 +1632,99 @@ policies and contribution forms [3].
...
+    };
+
+    function cleanup_done(test, error_count, bad_value_count) {
+        if (error_count || bad_value_count) {
+            var total = test._user_defined_cleanup_count;
+
+            tests.phase = tests.phases.ABORTED;
+            forEach(tests.tests,
+                    function(test) {
+                        test.phase = test.phases.COMPLETE
```
foolip: This wasn't done before, why do the individual tests need to be set to `COMPLETE` here?

resources/testharness.js
```diff
@@ -1621,29 +1632,99 @@ policies and contribution forms [3].
...
+            tests.status.message = "Test named '" + test.name +
+                "' specified " + total +
+                " 'cleanup' function" + (total > 1 ? "s" : "");
+
+            if (error_count) {
+                tests.status.message += ", and " + error_count + " failed";
+            }
+
+            if (bad_value_count) {
+                var type = test._is_promise_test ? "non-thenable " : "";
```
foolip: Not sure if I have this exactly right, but I think undefined is a value, so maybe "non-undefined" instead of the empty string here?

resources/testharness.js
```diff
@@ -1621,29 +1632,99 @@ policies and contribution forms [3].
...
+            tests.status.message += ".";
+
             tests.status.stack = null;
         }
-    };
+
+        test.phase = test.phases.COMPLETE;
+        tests.result(test);
+        forEach(test._done_callbacks,
+                function(cb) {
```
foolip: Expand to `callback` as in most other places in this file.

resources/testharness.js
```diff
@@ -1669,6 +1751,7 @@ policies and contribution forms [3].
         Object.keys(this).forEach(
                 (function(key) {
                     var value = this[key];
+                    if (key === '_done_callbacks' ) { return; }
```
foolip: This looks a bit fishy, why was this needed here, when so many other properties are not?

resources/testharness.js
```diff
@@ -1680,7 +1763,9 @@ policies and contribution forms [3].
         return clone;
     };
 
-    RemoteTest.prototype.cleanup = function() {};
+    RemoteTest.prototype.cleanup = function() {
+        this.done();
```
foolip: Looks a bit weird, can you add a comment about why it's this way?

resources/testharness.js
```diff
@@ -1723,7 +1816,7 @@ policies and contribution forms [3].
         this.message_target = message_target;
         this.message_handler = function(message) {
             var passesFilter = !message_filter || message_filter(message);
-            if (this_obj.running && message.data && passesFilter &&
+            if (message.data && passesFilter &&
```
foolip: What's this change for? I notice I'm not getting most of the changes in RemoteTest, can you describe in the commit message how it's impacted?

resources/testharness.js
```diff
@@ -1934,9 +2032,33 @@ policies and contribution forms [3].
...
+            forEach(this.tests,
+                    function(test) {
+                        if (test.phase === test.phases.CLEANING) {
+                            cleaningTest = test;
+                        }
+
+                        test.phase = test.phases.COMPLETE;
+                    });
+
+            // Timeouts that occur while a sub-test is in the "cleanup" phase
```
foolip: s/sub-test/test/ here and below, other comments in this file just call them tests.

resources/testharness.js
```diff
@@ -1934,9 +2032,33 @@ policies and contribution forms [3].
...
 
     Tests.prototype.timeout = function() {
+        var cleaningTest = null;
+
         if (this.status.status === null) {
-            this.status.status = this.status.TIMEOUT;
+            forEach(this.tests,
+                    function(test) {
+                        if (test.phase === test.phases.CLEANING) {
+                            cleaningTest = test;
```
foolip: It looks like `cleaningTest` will be set to the last test in cleanup, but there can actually just be one. Can you add a comment to ensure the reader that this is not an oversight?

resources/testharness.js
```diff
@@ -1934,9 +2032,33 @@ policies and contribution forms [3].
...
+                    });
+
+            // Timeouts that occur while a sub-test is in the "cleanup" phase
+            // indicate that some global state was not properly reverted. This
+            // invalidates the overall test execution, so the timeout should be
+            // reported as an error and cancel the execution of any remaining
+            // sub-tests.
+            if (cleaningTest) {
+                this.status.status = this.status.ERROR;
+                this.status.message = "Cleanup function for test named \"" +
```
foolip: The cleanup function itself doesn't have a timeout, so maybe "Timeout while running cleanup for test named ..."?

resources/testharness.js
```diff
@@ -2788,6 +2920,50 @@ policies and contribution forms [3].
...
     }
 
+    /**
+     * Immediately invoke a "predicate" function with a series of values in
+     * parallel and invoke a final "done" function when all of the "predicates"
+     * invocations have signaled completion.
+     *
+     * @param {array} value Zero or more values to use in the invocation of
+     *                      `predicate`
+     * @param {function} predicate A function that will be invoked once for
```
foolip: Calling this a predicate seems a bit weird to me, at least it's not a https://en.wikipedia.org/wiki/Predicate_(mathematical_logic) which is what I usually expect in code. Maybe just "callback", or something else?

resources/testharness.js
```diff
@@ -2016,19 +2138,29 @@ policies and contribution forms [3].
...
+                  {
+                      if (test.phase === test.phases.COMPLETE) {
+                          testDone();
+                          return;
+                      }
+
+                      test._add_done_callback(testDone);
+
+                      if (test.phase < test.phases.CLEANING) {
+                          test.cleanup();
```
foolip: I see this was done before as well, but can you comment about why cleanup is started here, when in some cases apparently it can already have happened?

resources/testharness.js
```diff
@@ -2920,12 +3096,10 @@ policies and contribution forms [3].
             }
             test.set_status(test.FAIL, e.message, stack);
             test.phase = test.phases.HAS_RESULT;
-            test.done();
```
foolip: Why no `test.done()` here? Was it always unnecessary, or only now?

foolip (commented):
Lots of comments, but this is looking really good. I'll be around today and tomorrow to keep reviewing.

foolip:
There are rather a lot of new failures with new harness errors in https://chromium-review.googlesource.com/c/chromium/src/+/836619 that may be related:
```
css/css-conditional/test_group_insertRule.html
css/css-values/calc-unit-analysis.html
fetch/api/response/response-stream-disturbed-6.html
geolocation-API/getCurrentPosition_IDL.https.html
html/dom/interfaces.html
html/semantics/embedded-content/media-elements/playing-the-media-resource/play-in-detached-document.html
html/semantics/forms/the-select-element/selected-index.html
mimesniff/mime-types/parsing.any.html
mimesniff/mime-types/parsing.any.worker.html
offscreen-canvas/compositing/2d.composite.canvas.source-in.worker.html
offscreen-canvas/compositing/2d.composite.canvas.source-out.worker.html
offscreen-canvas/fill-and-stroke-styles/2d.pattern.paint.norepeat.coord2.worker.html
offscreen-canvas/fill-and-stroke-styles/2d.pattern.paint.norepeat.outside.worker.html
offscreen-canvas/fill-and-stroke-styles/2d.pattern.paint.repeatx.coord1.worker.html
service-workers/service-worker/client-navigate.https.html
user-timing/measure_exceptions_navigation_timing.html
```
There are some outside of wpt too, but probably the same root cause.

@jugglinmike, can you check the above and see if there are real regressions here?

resources/test/tests/api-tests-3.html
```diff
@@ -5,7 +5,7 @@
 </head>
 <script src="../../testharness.js"></script>
 
-<body onload="load_test_attr.done()">
```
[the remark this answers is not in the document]
jugglinmike: This test invokes the global `timeout` function in the initial turn of the
event loop.

- In `master` today, this causes the internal `notify_complete` function to be
  invoked in that same turn. The harness synchronously renders the "timeout"
  status (and reports as much to the test runner)
- On this branch,  the `notify_complete` function is scheduled to be run with a
  new task (via `setTimeout`). The document `load` event fires in the
  intervening time, so by the moment the test status is reported, the "timeout"
  has been updated to an "error".

The root cause of this is the new `all_async` function. I've implemented that
to always invoke the final "complete" callback asynchronously, even in cases
where the individual operations have completed in a fully-synchronous manner. I
mentioned why I believe consistency in synchronicity is important in the pull
request description.

I suppose we *could* preserve the existing behavior by disallowing state
changes from "TIMEOUT" to "ERROR". However, both states essentially mean the
same thing: the test needs to be updated. I think "ERROR" is a more severe
case, so allowing that transition seems valid.

Also, it's worth mentioning that the usage we're discussing is pretty
artificial: when would a test synchronously trigger a timeout?

resources/test/tests/promise-with-sync.html
```diff
@@ -0,0 +1,84 @@
+<!DOCTYPE HTML>
+<html>
+<head>
+<title>Promise Tests and Synchronous Tests</title>
+</head>
+<body>
+<h1>Promise Tests</h1>
+<p>This test demonstrates the use of <tt>promise_test</tt> alongside synchronous tests.</p>
```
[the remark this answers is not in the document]
jugglinmike: [MDN agrees](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt), but
this was an attempt to respect local convention. Would you mind if I filed a
separate issue to switch to `<code>`?

resources/test/tests/promise-with-sync.html
```diff
@@ -0,0 +1,84 @@
...
+"use strict";
+var sequence = [];
+
+test(function(t) {
+  assert_array_equals(sequence, []);
+  sequence.push(1);
+}, "first synchronous test");
+
+promise_test(function() {
+  assert_array_equals(sequence, [1, 2]);
```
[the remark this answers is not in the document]
jugglinmike: If `assert_array_equals(sequence, [1, 2]);` is wrong, then I don't want to be
right.

resources/test/tests/queue.html
```diff
@@ -0,0 +1,139 @@
...
+}, "Second async_test (run in parallel)");
+
+test(function(t) {
+  assert_true(
+    inInitialTurn, "should execute in the initial turn of the event loop"
+  );
+}, "Third synchronous test");
+
+promise_test(function(t) {
+  assert_false(
```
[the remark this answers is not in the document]
jugglinmike: Right on!

resources/test/tests/worker-dedicated.html
```diff
@@ -30,8 +30,8 @@ <h1>Dedicated Web Worker Tests</h1>
 {
   "summarized_status": {
     "status_string": "ERROR",
-    "message": "Error: This failure is expected.",
-    "stack": "(implementation-defined)"
+    "message": "Error in remote https://web-platform.test:9003/resources/test/tests/worker-error.js: Error: This failure is expected.",
```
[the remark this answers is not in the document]
jugglinmike: The `remote_done` function severs the connection that a "remote" environment
has with its parent. It is intended to be invoked when a remote test completes.

- In `master` today, `remote_done` is invoked before `remote_error`. When
  `remote_error` is later invoked, it delegates to `remote_done`. However,
  because that method has already been invoked, this causes an internal
  exception which interrupts further execution of the function body.
  `remote_error` is designed to invoke `preventDefault` on the uncaught
  exception event, but this never happens due to the runtime exception. That
  means the error "bubbles up" to the parent context. `testharness.js` (the
  instance running in the parent context) reacts to the uncaught error by
  updating the status accordingly.

  This is why the current failure message inaccurately describes an error that
  originated in the parent context.
- With this patch applied, delegation to `remote_done` in `remote_error` has
  been refactored, so in the internal exception does not occur in this test,
  and the subsequent call to `preventDefault` proceeds as intended. This means
  the parent context does not receive an "uncaught error" event. Instead, it
  accurately reports the error with the text, "Error in remote [...]".

resources/testharness.js
```diff
@@ -532,30 +531,34 @@ policies and contribution forms [3].
...
+
             test.step(function() {
                 assert_not_equals(promise, undefined);
             });
-            Promise.resolve(promise).then(
-                    function() {
-                        test.done();
-                    })
+
+            Promise.resolve(promise)
```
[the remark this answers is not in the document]
jugglinmike: Exactly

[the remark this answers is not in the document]
jugglinmike: Practically speaking, yes. Technically, the semantics are slightly different,
though.

- In `master` today, if the invocation of `test.done` itself produces an
  exception, then this will be reported as a test failure.
- With this patch applied, if the invocation of `test.done` itself produces an
  exception, then this will cause a global "unhandled rejection" event, which
  will subsequently be reported as a harness failure.

Although this is not a condition we expect to ever occur, it seems better to
report the error as a harness failure rather than a test failure.

resources/testharness.js
```diff
@@ -532,30 +531,34 @@ policies and contribution forms [3].
 
     function promise_test(func, name, properties) {
         var test = async_test(name, properties);
+        test._is_promise_test = true;
+
         // If there is no promise tests queue make one.
         if (!tests.promise_tests) {
             tests.promise_tests = Promise.resolve();
```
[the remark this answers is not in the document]
jugglinmike: Exactly

resources/testharness.js
```diff
@@ -688,10 +691,13 @@ policies and contribution forms [3].
         if (tests.tests.length === 0) {
             tests.set_file_is_test();
         }
+        var end_wait = function() { tests.end_wait(); };
         if (tests.file_is_test) {
+            tests.tests[0]._add_done_callback(end_wait);
```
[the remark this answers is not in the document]
jugglinmike: Good idea. I've included a note here and also documented `Test#done` itself.

resources/testharness.js
```diff
@@ -1421,9 +1427,12 @@ policies and contribution forms [3].
         this.stack = null;
 
         this.steps = [];
+        this._is_promise_test = false;
 
         this.cleanup_callbacks = [];
         this._user_defined_cleanup_count = 0;
+        this._cleanup_done_callbacks = [];
```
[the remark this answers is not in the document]
jugglinmike: Nice catch! Removed.

resources/testharness.js
```diff
@@ -1597,20 +1607,21 @@ policies and contribution forms [3].
...
     Test.prototype.force_timeout = Test.prototype.timeout;
 
+    Test.prototype._add_done_callback = function(callback) {
+        this._done_callbacks.push(callback);
+    };
+
     Test.prototype.done = function()
     {
-        if (this.phase == this.phases.COMPLETE) {
+        if (this.phase >= this.phases.CLEANING) {
```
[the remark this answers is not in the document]
jugglinmike: Your interpretation is sound, and the change looks good to me. Updated.

resources/testharness.js
jugglinmike: For the chain `x().then(y, z)`, a roughly equivalent synchronous version would
look like this:

```js
var xThrew = false;
try {
  x();
} catch (error) {
  xThrew = true;
  z(error);
}

if (!xThrew) {
  y();
}
```

Compare that to `x().then(null, z).then(y)`:

```js
try {
  x();
} catch (error) {
  z(error);
}

y();
```

resources/testharness.js
jugglinmike: The short answer is: so the tests pass.

The long answer is: to enable a implicit and brittle solution to function. It
turns out that the original patch set the properties and relied on the
following code to react to the new state:

```js
test_environment.add_on_loaded_callback(function() {
    if (this_obj.all_done()) {
        this_obj.complete();
    }
});
```

So as written, aborting was not conducted imperatively. The actual
implementation could be described in plain English as, "set the properties and
wait for a DOM event to notice that change." This works in the highly
controlled "`testharness.js` test" environment, but it may not apply generally.

I've updated the implementation and added a new test to verify. (This has
obviated the code you identified here.)

resources/testharness.js
jugglinmike: Works for me.

resources/testharness.js
jugglinmike: Definitely. I don't know what came over me!

resources/testharness.js
jugglinmike: As an array containing function values, the value of the `_done_callbacks`
property is the only one that cannot be serialized for transfer between realms.
I agree that this has a little bit of code smell, though. Here are a few ideas
for how we could address it:

- Add an in-line comment explaining the condition
- Define the property as non-enumerable so that it isn't visited by
  `Object.keys`
- Define a property like `RemoteTest.UNCLONABLE = ["_done_callbacks"]` and
  reference that when creating a cloned instance

Do you have any preference?

resources/testharness.js
jugglinmike: Sure. Here's what I wrote:

>     * `RemoteTest` instances are objects which represent tests running in
>     * another realm. They do not define "cleanup" functions (if necessary,
>     * such functions are defined on the associated `Test` instance within the
>     * external realm). However, `RemoteTests` may have "done" callbacks (e.g.
>     * as attached by the `Tests` instance responsible for tracking the overall
>     * test status in the parent realm). The `cleanup` method delegates to
>     * `done` in order to ensure that such callbacks are invoked following the
>     * completion of the `RemoteTest`.

resources/testharness.js
jugglinmike: This is superfluous because the property is initialized to `true` and only set
to `false` immediately after the callback function is removed.

That said, I really shouldn't be making unrelated simplifications in this
patch. I'll replace the removal with a "TODO" that suggests the change.

resources/testharness.js
jugglinmike: Sure. I've also tried to think of a way to re-write the code to better express
the expectation, but besides some sort of "assert" statement, I can't think of
anything better.

resources/testharness.js
jugglinmike: You got it. (This was an attempt to be a little more precise with the terminology, but it's not helping anyone to be inconsistent.)

resources/testharness.js
jugglinmike: Oooh, that is a really nuanced distinction. I like your text much better!

resources/testharness.js
jugglinmike: As near as I can tell, this method was originally written to invoke the
`cleanup` method as a way to simulate completion for pending tests. But that's
pretty indirect, as your question indicates. It would be better to do something
like the following:

```js
if (test.phase === test.phases.CLEANING) {
    test._add_done_callback(testDone);
    return;
}

if (test.phase < test.phases.CLEANING) {
    test.phase = test.phases.COMPLETE;
    this_obj.result(test);
}

testDone();
```

...but this version causes failures, so I will need to investigate further.
I'll have a better answer for you tomorrow.

resources/testharness.js
jugglinmike: This is tough because the API expects two callback functions. What do you think
of "iteratee", is in [`_.forEach`](https://lodash.com/docs/#forEach)?

resources/test/tests/api-tests-3.html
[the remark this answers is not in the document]
foolip: Unless you're saying that `load_test_attr.done()` is an elaborate way of throwing an exception to test what happens with exceptions thrown by load attribute event handlers, I still don't get it. I've sent https://github.com/w3c/web-platform-tests/pull/8774 to remove these bits in two tests, and the tests pass all the same.

If it is needed, can you rename it to `not_a_variable.not_a_function()` or something?

resources/test/tests/promise-with-sync.html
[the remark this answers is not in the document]
foolip: Sure, I didn't notice any of the surrounding usage. Just leaving it alone without an issue is fine.

resources/test/tests/worker-dedicated.html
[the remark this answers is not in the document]
foolip: It was the "web-platform.test:9003" bit that worried me, not the change in behavior itself, although I did not understand the cause of that either. The new behavior seems fine, and if the port will always be 9003 that's also fine. Can you confirm that it passes if run locally, not just on Travis?

resources/testharness.js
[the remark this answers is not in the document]
foolip: That makes sense, thanks!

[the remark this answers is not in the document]
foolip: Although perhaps it has something to do with the new harness errors that are seen with this change. Let's see about that...

resources/testharness.js
foolip: Newbie indeed, some playing around in DevTools confirms the difference, the basic point I was missing is that handling the rejection causes the promise to be resolve with the return value of `on_error`, which is undefined.

resources/testharness.js
foolip: This attention to detail is much appreciated, it's easy to write brittle code, but hard to understand the resulting failures and even harder to make the code less brittle over time. Yay!

resources/testharness.js
foolip: :)

resources/testharness.js
foolip: Is this the only bit of state on RemoteTest that doesn't need to be transferred? If so just a comment about how it isn't needed and why it can't be serialized would do. If there are other bits that aren't needed, excluding them all would make sense. Couldn't something more like `Test.prototype.structured_clone` work, where the needed bits are explicitly listed?

resources/testharness.js
foolip: Excellent!

resources/testharness.js
foolip: Thanks!

resources/testharness.js
foolip: Yeah, I think a comment is the best tradeoff, if we start asserting things and they ever fail, we'd have to think about where those exception end up and if it'll look like a harness error or a test error. No thanks :)

resources/testharness.js
foolip: Leaving a comment or a TODO would be fine, given that this was already odd.

resources/testharness.js
foolip: That would work for me, I have a soft spot for made up somethingee words. "iterCallback" and "doneCallback" would also WFM.

resources/test/tests/add_cleanup_async_rejection_after_load.html
```diff
@@ -0,0 +1,56 @@
...
+    document.addEventListener("DOMContentLoaded", function() {
+      setTimeout(resolve, 0)
+    });
+  });
+  return Promise.resolve();
+}, "Test with failing cleanup that completes after DOMContentLoaded event");
+
+promise_test(function(t) {
+  return Promise.resolve();
+}, "Test that should not be run due to invalid harness state ('NOT RUN')");
```
foolip: Wonderful tests here :)

resources/testharness.js
```diff
@@ -20,8 +20,8 @@ policies and contribution forms [3].
     var settings = {
         output:true,
         harness_timeout:{
-            "normal":10000,
-            "long":60000
+            "normal":1000,
```
foolip: I presume this was for testing, need to revert changes?

resources/testharness.js
```diff
@@ -693,6 +701,9 @@ policies and contribution forms [3].
         }
         var end_wait = function() { tests.end_wait(); };
         if (tests.file_is_test) {
+            // Register a "done" callback to ensure the invocation of
+            // `end_wait` is deffered until after test "cleanup" logic has
```
foolip: s/deffered/deferred/

resources/testharness.js
```diff
@@ -2112,10 +2155,11 @@ policies and contribution forms [3].
 
     Tests.prototype.result = function(test)
     {
-        if (this.phase > this.phases.HAVE_RESULTS) {
-            return;
+        // If the harness has already transitioned beyond the `HAVE_RESULTS`
+        // phase, subsequent tests should not cause it to revert.
+        if (this.phase <= this.phases.HAVE_RESULTS) {
```
foolip: Micronit: `<=` can be `<` to avoid the appearance of doing nothing.

foolip (commented):
I've reviewed all the comments and changes since yesterday. I think the only really important outstanding bits is understanding the new failures in Chromium CQ. I'll send this new version through and look into some of the new failures.

foolip:
OK, so looking at the new failures in Chromium's CQ, at least inside of wpt, it looks like it's mostly harness errors being reported that were previously silent. What do you think the relevant change causing that was?

foolip:
I fixed a thing that came up in https://chromium-review.googlesource.com/c/chromium/src/+/839769 that was actually unrelated, and now I'll do no more digging today. Over to you, @jugglinmike.

resources/testharness.js
jugglinmike: This is due to a bug in my changes to `test.done`. In the version I originally
submitted, if invoked after a test has already transitioned to the "COMPLETE"
phase, it would exit without calling any "done" callbacks. Effectively, calling
`test.done` here caused the test harness to time out.

I've been thinking about the internal API of this change set. [You previously
identified how the `done_callbacks` abstraction was a bit
leaky](https://github.com/w3c/web-platform-tests/pull/8748#discussion_r158028605), and this is another sign of that.

So I hope you don't mind, but I've refactored the internals for clarity. The
only outward change is that it allows us to re-introduce this `test.done`
invocation (which is superfluous in `master`, by the way--I've included a
comment to remove it later).

resources/test/tests/worker-dedicated.html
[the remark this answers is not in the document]
jugglinmike: > Can you confirm that it passes if run locally, not just on Travis?

Every time :)

resources/testharness.js
jugglinmike: Optimistically, I would hope that everything is actually necessary. It's
definitely possible that's not true, but I'm reluctant to make a change like
that within this patch. I've added this comment instead:

>     // `RemoteTest` instances are responsible for managing
>     // their own "done" callback functions, so those functions
>     // are not relevant in other execution contexts. Because of
>     // this (and because Function values cannot be serialized
>     // for cross-realm transmittance), the property should not
>     // be considered when cloning instances.

resources/testharness.js
jugglinmike: I was able to address this as part of the refactoring mentioned above.

resources/testharness.js
jugglinmike: I like those, too. I've used the "snake case" versions in observance of local code style convention.

resources/test/tests/add_cleanup_async_rejection_after_load.html
jugglinmike: Thanks!

resources/testharness.js
jugglinmike: Certainly.

resources/testharness.js
jugglinmike: I removed this comment as part of the refactoring mentioned above.

resources/testharness.js
jugglinmike: This code was also factored away.

jugglinmike:
> OK, so looking at the new failures in Chromium's CQ, at least inside of wpt,
> it looks like it's mostly harness errors being reported that were previously
> silent. What do you think the relevant change causing that was?

It's tough to say generally. I'm going to look into the errors you reported
now. Hoping that I can reproduce them locally, outside of Chromium's automation
environment (and I'm prepared to also work from the earlier state of the branch
in case any of the changes introduced as part of the review process have
corrected those problems).

jugglinmike:
The errors @foolip uncovered are not due to tests being invalidated, but rather the harness detecting existing errors that were previously ignored. When I first learned this, I felt that the tests were in error and needed to be corrected. But after discussing this in IRC with Philip, we decided that test authors may have been intentionally relying on this behavior.

Put differently: one of the responsibilities of `testharness.js` is to ignore errors that occur after synchronous tests are complete.

I've pushed up a commit to address this. The changes have also prompted a simplification: the removal of the "ABORTED" testharness "phase". Here's the explanation from the relevant commit:

> As originally implemented, the "ABORTED" phase did not describe a
> distinct phase of the harness lifecycle. It has been made superfluous by
> the formalization of "cleanup" error management. The expected
> functionality can now be expressed via the recently-introduced
> `Test#abort` method and the interpretation of the "ERROR" state to
> immediately halt further testing.

We should include that in the description of the commit that eventually gets merged to `master`. (I've wanted to remove that pseudo-phase since shortly after [I implemented it back in June](https://github.com/w3c/web-platform-tests/pull/6151). I'm glad to finally have a good motivation!)

The new test proves the intended behavior for uncaught errors following synchronous test completion. Unfortunately, I haven't been able to think of a non-intrusive way to automatically verify the timeout behavior (my best idea involves a test that locks the browser for 11 seconds), but I've manually confirmed that aspect is in-line with `master`, as well. (I've also since realized that the timeout errors reported in the Chromium trial run are actually unrelated because those tests use the `promise_test` API. As @mkruisselbrink [noted in IRC](https://w3.logbot.info/testing/20171221#c368646-c368650), those particular tests are known to time out occasionally.)

Philip: I think this patch is ready for another trial in Chromium's test infrastructure. Would you mind?

foolip:
I haven't looked at any of the changes yet, but I did copy them over to https://chromium-review.googlesource.com/c/chromium/src/+/836619 and start another dry run.

foolip:
https://ci.chromium.org/buildbot/tryserver.chromium.linux/linux_chromium_rel_ng/614428 has some results. Still plenty of new harness errors. And parsing.any.html is still a timeout.

It's also interesting that some FAIL have gone to NOTRUN, and in Accelerometer_onerror-manual.https.html also a PASS to NOTRUN. This is presumably because of the new logic for treating errors in cleanup, but looking at the tests to confirm would be good.

For most tests the harness errors weren't new, and for those I think just fixing the harness errors makes sense. But there are still some with new harness errors, like html/dom/interfaces.html.

Anyway, I feel like the number of affected tests is now small enough that this can't be too far from what we need :)

jugglinmike:
Thanks for the trial run and the encouragement, Philip. Hopefully we can get these kinks worked out in short order.

jugglinmike:
Hi @foolip! I'm ready to take another shot at this feature. I'm not able to reproduce any of the failures in your previous report. That was last December, though, so I can't be sure if the issue is in my process or if the problems have since been resolved. Plenty of new tests have been written in that time, too. Would you mind scheduling another trial run in Chromium's infrastructure?

jugglinmike:
Two commits have been merged to `master` since I resolved conflicts last Wednesday:

- fcb5082 Make fetch_tests_from_worker return a promise that resolves when the remote tests complete.
- 9d896a9 Feature-test global scopes without events and timeouts.

The conflicts were trivial, but since neither changeset includes tests, I can't formally validate my resolution. I plan on requesting confirmation from the two authors once we work out the final kinks on this branch.

jugglinmike:
@foolip I'm not familiar with the Gerrit interface, but it appears as though the tests were not run against your second patch. Is that true?

foolip:
Oops, I forgot to start the bots after the second upload. From the first patch, which just included the non-wpt copy of testharness.js, there were some regressions though:
https://test-results.appspot.com/data/layout_results/linux_chromium_rel_ng/93662/layout-test-results/results.html
https://test-results.appspot.com/data/layout_results/mac_chromium_rel_ng/46501/layout-test-results/results.html

It mostly looks like stuff that should be fixable within the tests though. Let's see if there are also regressions inside wpt, bots running now.

foolip:
OK, results including wpt are in:
https://test-results.appspot.com/data/layout_results/linux_chromium_rel_ng/94299/layout-test-results/results.html
https://test-results.appspot.com/data/layout_results/mac_chromium_rel_ng/47148/layout-test-results/results.html

Looks like a few tests are affected by the test ending later than it previously did. I don't quite recall what our conclusion on that was. If it's possible to update the tests that seems fine given that there aren't too many of them.

foolip:
css/css-fonts/variations/at-font-face-font-matching.html is pretty suspect though, why are more subtests passing than before? I wouldn't expect this refactor to fix any tests :)

jugglinmike:
Hi @foolip! I've written too much. To summarize:

- The CSS test may be invalid, so [I've proposed a modification to that test in a separate pull request](https://github.com/w3c/web-platform-tests/pull/11082)
- I've pushed up a commit which corrects 5 of the WPT failures (including the CSS test). I am conflicted about whether that commit is the best solution
- For the remaining 3 WPT tests, I am unable to reproduce the behavior from the Chromium infrastructure (neither the passing status in `master` nor the failing status on this branch)

The next steps for me:

1. Decide what to do about handling synchronicity
2. Determine how to properly execute those tests
3. Update the results collection system
4. Continue debugging the regressions introduced by this patch

For more detail, you can either slough through the novel below or give me a call to hear it spoken with breathless excitement.

---

To start, I have limited my research to the WPT tests.

Regarding the CSS test:

- `css/css-fonts/variations/at-font-face-font-matching.html`

Today, this fails in Chrome due to a timing issue; the tests execute before the required font has been applied to the element under test. With the "async cleanup" patch applied, more tests fail because [this use of `setTimeout`](https://github.com/w3c/web-platform-tests/pull/8748/files#diff-d9d64f7c061650556b18794cc9719152R1681) introduces a delay and gives the browser an opportunity to apply the fonts. I can't find any specification text that dictates the synchronous operation which this test expects, so this may be a test bug. [I've submitted a fix in a separate patch](https://github.com/w3c/web-platform-tests/pull/11082). I'm sure my reviewers will be able to shed some light on this.

The following tests produce an uncaught exception in a task that follows their final sub-test:

- `fetch/api/abort/general.any.html`
- `fetch/api/abort/general.any.sharedworker.html`
- `fetch/api/abort/general.any.worker.html`
- `fetch/api/abort/general.https.any.serviceworker.html`

They failed because they expect the test runner to detect test completion within the same task of the final test and therefore ignore that uncaught exception.

If that sounds familiar, it's because [we discussed the "uncaught exception after tests" problem last December](https://github.com/w3c/web-platform-tests/pull/8748#issuecomment-353687896).  Back then, we only considered fully-synchronous tests and the task queue. These four tests show that this also applies in asynchronous contexts, at the microtask level.

At the cost of additional complexity, I've modified the internal scheduling logic to satisfy this expectation. This is a little complicated because we cannot use `Promise.resolve` in all cases. [There are other ways to schedule a microtask](https://github.com/calvinmetcalf/immediate), but because they rely on web platform features that WPT is validating, they aren't appropriate for use in a generic utility like `testharness.js`.

Even with the assurance of our trial runs in the Chromium infrastructure, I'm still very reluctant to make this patch any more complicated than it needs to be. I'm also concerned about the cognitive overhead even this small change adds to the internals--the task/microtask distinction has a certain code smell to it. There are two alternatives that I'm aware of:

1. Revert this latest commit and update the failing tests. Allowing implementation concerns to inform the API design is admittedly a bit of a cop-out. It may be acceptable in this case because the behavior we're trying to maintain is so nuanced.
2. Dumb down the internals. `testharness.js` has organically developed a contract that is inconsistent in its synchronicity. I've been trying hard to make that distinction explicit within the internal code paths so that the expected behavior is obvious to future contributors (rather than an emergent property). However, if we remove all the explicitness, the internals become simpler, and the observable behavior is the same. [Here's what the simplification would look like.](https://github.com/bocoup/web-platform-tests/commit/testharness-cleanup-async-december-squashed-dumb) This could work if we have enough confidence in the test suite's ability to prevent future regressions. I maintain my aforementioned fears about [Zalgo's Dark͝ Tend̴r̡i҉ls sneaking out](http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony).

**I'm curious to hear your thoughts on this!**

As for the other tests:

- `web-nfc/nfc_push.https.html`
- `web-animations/timing-model/animations/canceling-an-animation.html`
- `webauthn/createcredential-timeout.https.html`

I have not been able to reproduce the expected results when running these on `master`. Although I am using Chromium locally, the test errors I am experiencing are also afflicting https://wpt.fyi (which runs Chrome):

- https://wpt.fyi/web-nfc/nfc_push.https.html
- https://wpt.fyi/web-animations/timing-model/animations/canceling-an-animation.html
- https://wpt.fyi/webauthn/createcredential-timeout.https.html

This suggests that there is some additional environment configuration that we have yet to identify. My go-to reference for this information is [the VirtualTestSuites file](https://github.com/chromium/chromium/blob/84da086ac8a21de0a177b800c8d6b1e21d867af1/third_party/WebKit/LayoutTests/VirtualTestSuites), but it doesn't have entries for any of these tests. **Do you know where else I should look?**

foolip:
> For more detail, you can either slough through the novel below or give me a call to hear it spoken with breathless excitement.

Can I get it as an audiobook? I'd love that for my morning commute :)

foolip:
Agree that "Decide what to do about handling synchronicity" is the first and most important step. Given that the number of affected tests isn't too large, we're not very constrained by what existing tests assume.

I think the following is true now and should remain true:
* for `test(fn)`, nothing that happens after `fn` returns can affect the subtest result
* for `async_test(t => ...)`, nothing that happens after `t.done()` matters
* for `promise_test(fn)`, where `fn` returns a promise (or thenable?), nothing that happens after `the_callback` in `fn().then(the_callback)` is invoked matters

All of the above points are within testharness.js even though I expressed them in terms of what the test does. For the `promise_test` case, it might be slightly more convenient for authors to say that nothing after the promise is resolved matters, but there is no synchronous "promise resolved callback" (by design, I hope) so there's no point in testharness.js to hang that onto.

Make sense so far?

Then the question is how long after each finished subtest we wait to discover new tests before declaring the whole test finished, which is also the window in which uncaught exceptions and unhandled promise rejection can happen.

Once we are past the load event, would it work to let the mentioned "nothing matters" points also be when we look for new tests, and that if there are none, the test is over?

A problem with this is that if the following test does not start to run synchronously when the previous one finished, then the last test will be special in that it doesn't have the same window following it where harness errors can happen. That's annoying when adding a new test to the end. In the worst case, with promise_test, it might be the new added test that gets the failure instead.

So: what is the necessary gap between a subtest finishing and the next one starting?

jugglinmike:
> A problem with this is that if the following test does not start to run
> synchronously when the previous one finished, then the last test will be
> special in that it doesn't have the same window following it where harness
> errors can happen.

Actually, that's already the cast in `master` today.

> So: what is the necessary gap between a subtest finishing and the next one
> starting?

It depends on how the test was defined. It's "no gap" for `test` and
`async_test` and a microtask for `promise_test`. Here it is laid out visually:

    Task | Microtask | Synchronous steps
    1    | 1         | test() ---------> { /*body*/ } --> testharness complete
    2    | 1         |                                                          (uncaught exceptions)

    Task | Microtask | Synchronous steps
    1    | 1         | async_test() ---> { t.done(); } -> testharness complete
    2    | 1         |                                                          (uncaught exceptions)

    Task | Microtask | Synchronous steps
    1    | 1         | promise_test()
    1    | 2         |        `--------> { return p; }
    1    | 3         |                        `---------> testharness complete
    2    | 1         |                                                          (uncaught exceptions)

<details>
  <summary>Test used to verify the above</summary>

```js
function microTask(callback) {
  var observer = new MutationObserver(callback);
  var element = document.createTextNode('');
  observer.observe(element, {
    characterData: true
  });

  element.data = true;
}

function mark(message) {
  console.log('sync      - ' +  message);
  microTask(() => console.log('microtask - ' + message));
  setTimeout(() => console.log('task      - ' + message));
}
setup({explicit_done: true});

add_start_callback(() => console.log('testharness.js - start'));
add_result_callback(() => console.log('testharness.js - result'));
add_completion_callback(() => console.log('testharness.js - complete'));

window.onload = () => {
  mark('before sync test');
  test(() => {
    mark('inside sync test');
  }, 'sync');

  mark('after sync test');

  async_test((t) => {
    mark('async_test: before done');
    t.done();
    mark('async_test: after done');
  }, 'async');

  mark('after async_test');

  promise_test(() => {
    mark('promise_test');
    return Promise.resolve();
  }, 'promise test');

  mark('after promise_test');

  done();
};

```
</details>

<details>
  <summary>Console output</summary>

    sync      - before sync test
    general.any.js:22 testharness.js - start
    general.any.js:16 sync      - inside sync test
    general.any.js:23 testharness.js - result
    general.any.js:16 sync      - after sync test
    general.any.js:16 sync      - async_test: before done
    general.any.js:23 testharness.js - result
    general.any.js:16 sync      - async_test: after done
    general.any.js:16 sync      - after async_test
    general.any.js:16 sync      - after promise_test
    general.any.js:17 microtask - before sync test
    general.any.js:17 microtask - inside sync test
    general.any.js:17 microtask - after sync test
    general.any.js:17 microtask - async_test: before done
    general.any.js:17 microtask - async_test: after done
    general.any.js:17 microtask - after async_test
    general.any.js:17 microtask - after promise_test
    general.any.js:16 sync      - promise_test
    general.any.js:17 microtask - promise_test
    general.any.js:23 testharness.js - result
    general.any.js:24 testharness.js - complete
    general.any.js:18 task      - before sync test
    general.any.js:18 task      - inside sync test
    general.any.js:18 task      - after sync test
    general.any.js:18 task      - async_test: before done
    general.any.js:18 task      - async_test: after done
    general.any.js:18 task      - after async_test
    general.any.js:18 task      - after promise_test
    general.any.js:18 task      - promise_test
</details>

That means that there is already a gap between sub-tests defined via
`promise_test`. It's at the *microtask* level, though. The four harness errors
that the latest trial in Chromium demonstrated occurred because this patch was
creating a new *task*:

    Task | Microtask | Synchronous steps
    1    | 1         | promise_test()
    1    | 2         |        `--------> { return p; }
    2    | 1         |                        |         (uncaught exceptions)
    3    | 1         |                        `-------------------------------> testharness complete

Commit 2c8c0e9 re-implements this behavior in as explicit a manner as possible.
Maintaining that explicitness is starting to be burdensome, though.  I'm
warming up to the idea that simplicity is more important than explicitness.
I've merged two simplifications into this branch. Bring it, Zalgo.

@foolip: can you advise on how to re-create the conditions used by Chromium for
the following WPT tests?

- `web-animations/timing-model/animations/canceling-an-animation.html`
- `web-nfc/nfc_push.https.html`
- `webauthn/createcredential-timeout.https.html`

Also, do you know if there's any way to run the LayoutTests without building
Chromium?

jugglinmike:
> Also, do you know if there's any way to run the LayoutTests without building Chromium?

I learned from #chromium that I'll need to build the project to run those tests.

jugglinmike:
We recently landed gh-11150 which is a culmination of a parallel effort I started last year (see also gh-gh-8597 and gh-11010). I've just resolved the conflicts that introduced for this branch. The end result is that we are now running all the appropriate tests in two variant environment: one variant with no modification (simulating a modern browser) and one variant with the global `Promise` constructor removed (simulating the Servo browser).

This does not address the regressions listed above (those are still on me to fix), but it does increase our confidence that we are honoring an undocumented technical requirement.

jugglinmike:
I've built Chromium and worked through most of [the regressions reported above](https://github.com/web-platform-tests/wpt/pull/8748#issuecomment-389494423).  There are two classes of regressions that call for changes to the patch:

**Tasks versus microtasks** The following tests regressed due to the aforementioned change from microtasks to tasks. They should be fixed now that we've switched back to microtasks:

- `bluetooth/descriptor/readValue/gen-io-op-fails.html`
- `bluetooth/descriptor/writeValue/gen-io-op-fails.html`
- `external/wpt/css/css-fonts/variations/at-font-face-font-matching.html`
- `external/wpt/fetch/api/abort/general.any.html`
- `external/wpt/fetch/api/abort/general.any.sharedworker.html`
- `external/wpt/fetch/api/abort/general.any.worker.html`
- `external/wpt/fetch/api/abort/general.https.any.serviceworker.html`
- `external/wpt/web-nfc/nfc_push.https.html`
- `external/wpt/webauthn/createcredential-timeout.https.html`
- `fast/canvas-api/canvas-lose-restore-max-int-size.html`
- `http/tests/streams/chromium/deep-recursion-getwriter.html`
- `usb/usbDevice-iframe.html`
- `virtual/outofblink-cors/external/wpt/fetch/api/abort/general.any.html`
- `virtual/outofblink-cors/external/wpt/fetch/api/abort/general.any.sharedworker.html`
- `virtual/outofblink-cors/external/wpt/fetch/api/abort/general.any.worker.html`
- `virtual/outofblink-cors/external/wpt/fetch/api/abort/general.https.any.serviceworker.html`

**Aborting for unhandled rejections** In testharness.js, an uncaught exception or unhandled Promise rejection will force the harness status to "ERROR". In `master` today, if there are any pending `promise_test`s, they will run as normal. With this patch applied, such tests are never executed and instead reported as "NOT RUN". This new behavior conflicts with Chromium's expectations files for `external/wpt/web-animations/timing-model/animations/canceling-an-animation.html` I've pushed up another commit to preserve the existing behavior. (For the record, I would prefer that we abort in this case, but I'll propose changing that behavior in a smaller dedicated patch.)

There are two other regressions that call for changes in the Chromium tree:

**Invalid return value** `http/tests/fetch/chromium/no-inherit-type.html` invokes `add_cleanup` with a value-returning function. Basing harness behavior off of a return value is more convenient than a more explicit API like (for example) `this.make_async()`.  That convenience can also cause test authors to make errors like unknowingly failing to enable the feature. For instance, returning the Animation instance `animation` rather than its Promise property, `animation.ready`. To help avoid these mistakes, we decided to react to any non-thenable value by causing a harness failure. You may remember this because [in preparation for this patch, we previously refactored two WPT tests to remove the return value](https://github.com/web-platform-tests/wpt/pull/8742).

**Brittle expectations file** `fast/dom/custom/svg-use-shadow-tree.html` fails simply because we inserted code into `testharness.js`. That test makes assertions outside of an asynchronous test wrapper (tisk tisk). The assertions fail in Chromium, and in order to allow this in the format of project's expectation files, the location of the `assert_false` function is hard-coded.  This is probably old news to you, though, since you and @Hexcles have both needed to [update the line numbers in the expectation file this year](https://chromium.googlesource.com/chromium/src/+log/refs/heads/master/third_party/WebKit/LayoutTests/fast/dom/custom/svg-use-shadow-tree-expected.txt) (despite [the test not having been modified since 2015](https://chromium.googlesource.com/chromium/src/+log/refs/heads/master/third_party/WebKit/LayoutTests/fast/dom/custom/svg-use-shadow-tree.html)).

What's surprising about this is that another test which exhibits the same problem was *not* flagged by the trial run.  `http/tests/webaudio/autoplay-crossorigin.html` includes testharness.js via `../resources/testharness.js`, but that doesn't resolve to a file in the repository. Probably the test infrastructure is providing it somehow. The LayoutTests suite already includes its own copy of `testharness.js`. In your commit for the most recent trial, you took care to update both the WPT version and the LayoutTest version, but this makes me wonder if maybe there's a third file that needs updating.

<details>
  <summary>command used to find that test</summary>

    $ find . -type f -name '*-expected.txt' -exec grep --color 'line.*Uncaught Error' '{}' +
    ./http/tests/webaudio/autoplay-crossorigin-expected.txt:CONSOLE ERROR: line 2684: Uncaught Error: assert_equals: stateAfterClick expected "running" but got "suspended"
    ./fast/dom/custom/svg-use-shadow-tree-expected.txt:CONSOLE ERROR: line 2752: Uncaught Error: assert_false: Should not call createdCallback in UA ShadowRoot. expected false got true
    ./fast/dom/custom/svg-use-shadow-tree-expected.txt:CONSOLE ERROR: line 2752: Uncaught Error: assert_false: Should not call attachedCallback in UA ShadowRoot. expected false got true
    ./fast/dom/custom/svg-use-shadow-tree-expected.txt:CONSOLE ERROR: line 2752: Uncaught Error: assert_false: Should not call createdCallback in UA ShadowRoot. expected false got true
    ./fast/dom/custom/svg-use-shadow-tree-expected.txt:CONSOLE ERROR: line 2752: Uncaught Error: assert_false: Should not call attachedCallback in UA ShadowRoot. expected false got true
    ./fast/xpath/nsresolver-exception-expected.txt:CONSOLE ERROR: line 14: Uncaught Error: this is an exception
    ./fast/harness/expected-error-in-async-test-expected.txt:CONSOLE ERROR: line 12: Uncaught Error: An expected error
    ./fast/events/window-onerror-07-expected.txt:CONSOLE ERROR: line 6: Uncaught Error: An exception
    ./fast/events/window-onerror-06-expected.txt:CONSOLE ERROR: line 6: Uncaught Error: Original exception.
    ./fast/events/window-onerror-06-expected.txt:CONSOLE ERROR: line 6: Uncaught Error: Nested exception.
    ./geolocation-api/callback-exception-expected.txt:CONSOLE ERROR: line 37: Uncaught Error: Exception in success callback
    ./storage/indexeddb/transaction-error-expected.txt:CONSOLE ERROR: line 82: Uncaught Error: This should *NOT* be caught!
    ./storage/indexeddb/version-change-abort-expected.txt:CONSOLE ERROR: line 52: Uncaught Error: This should *NOT* be caught!

</details>

**Unknown** I have been unable to recreate the regression in `xr/events_session_end.html`. It seems likely that the recent changes have corrected this, too. I'd like to learn if this is the case by running another trial before spending more time trying to re-create this locally.

Does that sound good to you, @foolip? If so, would you mind triggering another trial?

foolip:
Thanks @jugglinmike, sounds like we're getting to the bottom of the bucket of issues and should soon be able to land this.

I've sent https://chromium-review.googlesource.com/c/chromium/src/+/1109513 to fix http/tests/fetch/chromium/no-inherit-type.html.

There are just two copies of testharness.js in Chromium, and where it seems to not exist, "the test infrastructure is providing it somehow" is indeed the right explanation. Here's one such trick, although not the one in http/tests/ which I actually don't know how it works:
https://cs.chromium.org/chromium/src/content/shell/renderer/layout_test/blink_test_helpers.cc?l=37&rcl=9fdacf3b3440c71e50c850928b63550508ece11a

Since we have two copies of a bunch of files which aren't always in sync, I've sent https://chromium-review.googlesource.com/c/chromium/src/+/1109514 to sync them so that there are fewer differences when applying the patch.

I'll go ahead and apply this patch to our two testharness.js as they currently are and see how that fares in a CQ dry run.

foolip:
The changes applied to both of our testharness.js files, so trying it now at https://chromium-review.googlesource.com/c/chromium/src/+/1109681.

foolip:
I've also sent https://chromium-review.googlesource.com/c/chromium/src/+/1109689 to fix fast/dom/custom/svg-use-shadow-tree.html and http/tests/webaudio/autoplay-crossorigin.html.

foolip:
https://chromium-review.googlesource.com/c/chromium/src/+/1109681 has finished running and there are a bunch of failures. A bunch of fullscreen tests do `t.add_cleanup(() => document.exitFullscreen())` and because they are `async_test`s and not `promise_tests`s, that causes a harness error "specified 1 'cleanup' function, and 1 returned a non-undefined value".

Since the `exitFullscreen()` promise is rejected if not in fullscreen, it won't do to just add `{}` to return undefined. And `exitFullscreen()` doesn't return a promise in all implementations yet, and fullscreen tests have been authored to not fail because of promises not being supported if that wasn't the point of the test. So the cleanup needs to be:
```js
function fullscreen_cleanup() {
  const promise = document.exitFullscreen();
  if (promise) {
    promise.catch(() => {});
  }
}
```

Putting that in a helper and having each test call `t.add_cleanup(fullscreen_cleanup)` is probably best.

I didn't look closely at the other failures.

jugglinmike:
[Way back in December, I commented about removing the "ABORTED" phase.](https://github.com/web-platform-tests/wpt/pull/8748#issuecomment-353687896) While it's true that it is not entirely distinct from "ERROR", it is also not synonymous with "ERROR." In `master` today, the test harness may enter the "ERROR" phase but continue to execute tests.

This is a little unintuitive to me, and that may be why I didn't question the
implications of equating "ERROR" and "ABORTED". I should have, though, because
doing so changes the existing behavior of the test harness.

We're trying to limit the scope of this patch to the addition of a feature, so
I've re-introduced the distinction. I also added regression tests to prove
parity with `master`. (Note that this version is implemented with a separate
property of the `Test` object. This more clearly reflects its role as a
modifier of "ERROR" rather than a distinct phase in its own right.)

The new tests use a different approach than the "functional" tests written so
far because writing and reviewing many small files is cumbersome. They all pass
on `master` and on this branch. They also include two "TODO" comments--these
are odd but unrelated behaviors. I would like to research further once this
feature is merged.

- `external/wpt/fullscreen/model/move-to-iframe-manual.html`
- `external/wpt/fullscreen/model/move-to-inactive-document-manual.html`
- `external/wpt/fullscreen/model/remove-child-manual.html`
- `external/wpt/fullscreen/model/remove-first-manual.html`
- `external/wpt/fullscreen/model/remove-last-manual.html`
- `external/wpt/fullscreen/model/remove-parent-manual.html`
- `external/wpt/fullscreen/model/remove-single-manual.html `
- `external/wpt/html/infrastructure/urls/resolving-urls/query-encoding/attributes.sub.html?encoding=utf8`
- `external/wpt/html/infrastructure/urls/resolving-urls/query-encoding/attributes.sub.html?encoding=x-cp1251`
- `external/wpt/fetch/sec-metadata/script.tentative.https.sub.html`
- `virtual/outofblink-cors/external/wpt/fetch/sec-metadata/script.tentative.https.sub.html`

These tests return a non-thenable, so the fix is straightforward:

- https://github.com/web-platform-tests/wpt/pull/11733
- https://github.com/web-platform-tests/wpt/pull/11740
- https://github.com/web-platform-tests/wpt/pull/11741

The final four were being incorrectly reported as a timeout. This was due to
the change to exceptional condition handling described above. I've corrected
that with a new commit to this branch.

- `external/wpt/web-animations/timing-model/animations/canceling-an-animation.html`

This failure is a layover from the previous trial. I explained the issue above,
but I previously neglected to push the correction to this branch. The latest
commit includes the necessary change (i.e. removing the call to `tests.abort`
in response to uncaught exceptions and unhandled promise rejections).

- `external/wpt/webrtc/RTCPeerConnection-setRemoteDescription-tracks.https.html`

The WebRTC tests are a little creative with their usage of testharness.js. The
`RTCPeerConnection-helper.js` script implements a utility function that
combines Promises with the `step_func` API generally used within `async_test`:

    function addEventListenerPromise(t, target, type, listener) {
        target.addEventListener(type, t.step_func(e => {
          if (listener != undefined)
            e = listener(e);
          resolve(e);
        }));
      });
    }

Supporting this was a trivial change; I've documented the motivation in the
source and in the accompanying regression test.

- `fast/dom/custom/svg-use-shadow-tree.html`
- `http/tests/fetch/chromium/no-inherit-type.html`
- `virtual/outofblink-cors/http/tests/fetch/chromium/no-inherit-type.html`

As mentioned above, these needed a fix in Chromium. Since we ran this trial,
you and @Hexcles fixed this:

https://github.com/chromium/chromium/commit/f68571e38febb3a1222da1439050fd990929b274
https://github.com/chromium/chromium/commit/3d359ef54681ef365b1f72cab1c3565c728e0b29

Thanks for that!

- `virtual/outofblink-cors/http/tests/fetch/chromium/call-extra-crash-is-locked.html`

This test includes the following note:

    // This runs thousands of times, causing the console to spew "Uncaught
    // (in promise) TypeError: Response body is already used" messages, but
    // it's harmless.

However, it also invokes `add_cleanup` with a Promise returning function:

    t.add_cleanup(() => caches.delete('foo'));

Applying this patch causes the harness to wait for the item to be removed from
the cache. The `unhandledrejection` event which previously occurred after the
test was complete now occurs before it is complete and influences the test. So,
no longer "harmless." That means the test must be modified.

One seemingly-valid workaround would be to add an empty rejection handler to
the function call that is expected to fail asynchronously. However, because the
test is specifically concerned with behavior when the execution call stack is
full, such a handler cannot be invoked by the runtime. Recognizing this took me
longer than I'd care to admin.

There are two changes that *will* fix the problem: either omitting the return
value in the cleanup function:

```diff
- t.add_cleanup(() => caches.delete('foo'));
+ t.add_cleanup(() => { caches.delete('foo'); });
```

Or allowing unhandled exceptions:

```diff
+setup({
+  allow_uncaught_exception: true
+});
```

Neither is perfect, but I think the latter would be preferable.

(Changes related to this test landed since I last built Chromium. This meant I
needed to re-build Chromium, and that's unfortunately still a bit of a
production for me. Sorry for the delay!)

- `virtual/outofblink-cors/http/tests/fetch/workers/thorough/cors-preflight2-base-https-other-https.html`

I haven't been able to reproduce this error locally  It's possible that a bug
has been fixed in the time between the trial and my re-building Chromium
locally (though none of the related files have been modified in the past 6
months, so the bug would either be in the test infrastructure or in the
Chromium source). @foolip would you like me to investigate more, or do you
think we should verify the problem with another trial first?

jugglinmike:
[This patch has successfully passed through Chromium's build system!](https://chromium-review.googlesource.com/c/chromium/src/+/1142600) @jgraham is currently helping me vet it in Firefox's build system; I'll report back here when we have results.

jugglinmike:
The patch did not fare so well in Firefox's build system:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=007f47b1ca56d4eee40737c6f7dfcf0b715e5f50&selectedJob=191398326

Within the "Linux opt" build, almost all errors were cases where this patch
caused sub-tests that were expected to time out to instead pass. The test-level
status was not effected. (The only error that did not follow this pattern was
in a WebDriver spec test; I think it's safe to assume that is due to unrelated
flakiness.)

The problem stemmed from the way this patch treated all active tests whenever
the harness timed out:

    In `master` today, testharness.js responds to harness-level timeouts by
    forcing any running tests to enter their "cleanup" phase. It does this
    using the `Test#cleanup` method.

    Prior to this commit, the feature being introduced implemented this
    behavior by instead invoking `Test#done`. In addition to invoking
    `Test#cleanup`, this method also sets the test's status to "done." This
    is inappropriate for tests that have already timed out. The correction
    cannot take place within `Test#done`, however. All test's take on the
    "timeout" value when they are started. Within the `done` method, this
    makes it impossible to differentiate between a successfully completed
    test and a test that has timed out.

    Update the call site to instead invoke `Test#cleanup` (which more
    closely resembles the current implementation in `master`).

That fix is available on this branch now.

It seemed strange to me that we needed Firefox's build system to identify this
mistake, so I took a closer look at Chromium's infrastructure. While that build
system supports enforcing test expectations at a sub-test level (via files with
a `-expected.txt` suffix), it does not use the same convention for tests that
are expected to time out. Those tests are simply listed by name in a file named
`TestExpectations` and annotated with the word `TIMEOUT`. That's why we
couldn't observe this particular regression by running the patch through
Chromium's infrastructure.

The condition where the regression is expressed is interesting. It only occurs
in the face of an uncaught error during subtest execution. Unlike (for
instance) timeouts resulting from an un-implemented event handler, these
timeouts are a strong indication of deficiencies in the tests themselves (i.e.
asynchronous test code running outside of `step_func`). I've submitted a few
patches to make those tests explicitly detect the errors so that non-conforming
browsers can fail faster and with more useful messages:

- https://github.com/web-platform-tests/wpt/pull/12266
- https://github.com/web-platform-tests/wpt/pull/12267
- https://github.com/web-platform-tests/wpt/pull/12268

@jgraham There are some other failures which I don't understand, e.g.

>     IOError: [Errno 2] No such file or directory: '/builds/worker/workspace/build/tests/web-platform/tests/push-api/idlharness.any.js'

>     PROCESS-CRASH | /streams/readable-streams/tee.dedicatedworker.html | application crashed [@ ReadableStreamControllerError]

>     TEST-UNEXPECTED-ERROR | /builds/worker/checkouts/gecko/testing/web-platform/meta/MANIFEST.json:0 | resources/test/tests/functional/task-scheduling-test.html in source but not in manifest. (wpt-manifest)

I'm hopeful that some of these can be explained by external factors (maybe the
use of "inbound" as the target, or still more pre-existing flakiness). I plan
to look into this tomorrow, but if you think they could be unrelated to this
patch, then maybe we could verify by scheduling another trial in advance of
further research.

Hexcles:
@jugglinmike you're right we don't support subtest timeout in Chromium https://crbug.com/777526

jugglinmike:
@jgraham scheduled [another job on Mozilla's infrastructure](https://treeherder.mozilla.org/#/jobs?repo=try&revision=c2bbae15cee4c7299a531ba6f36ea9d803b78f6c), and we found that this patch no longer regresses there, either!

As a third (and final?) validation, I'd like to preview the effect the patch will have in Servo's infrastructure. That's because Servo does not support Promises, so it's possible that this change regresses behavior in a totally different way for that browser. [I've done my best to guard against that via extensions to the testharness.js test suite](https://github.com/web-platform-tests/wpt/pull/8748#issuecomment-397686851), but there's nothing surer than actually running the tests in the browser.

@jdm: would you mind running this patch in Servo and sharing the results here?

jdm:
No complaints from Servo!

jugglinmike:
Awesome, thanks @jdm!

With that feedback, we now have confidence that this feature does not regress in Chromium, Firefox, or Servo. @foolip are there any other checks you'd like to perform? If not, I can submit a new pull request with a single squashed commit (I'd like to preserve the history of this review process).

foolip (approved):

jugglinmike:
This was a much more challenging task than I anticipated. Thanks for your help, @jdm, @jgraham, and especially @foolip!

pwnall:
@jugglinmike Thank you very much for seeing this through!

PR #8781, opened 2017-12-22
Improve assert_array_equals to recurse into the nested arrays

### LINKED ISSUE #8769: assert_array_equals should recurse into nested arrays
When comparing arrays with nested arrays, `assert_array_equals` should recurse into the nested arrays.

Honry:
also improve the desc to showing the actual/expected array contents
  and fix issue in assertion of array object

  Fix #8769

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8781)

### DIFF
--- a/resources/testharness.js
+++ b/resources/testharness.js
@@ -996,26 +996,40 @@ policies and contribution forms [3].
 
     function assert_array_equals(actual, expected, description)
     {
-        assert(typeof actual === "object" && actual !== null && "length" in actual,
-               "assert_array_equals", description,
-               "value is ${actual}, expected array",
-               {actual:actual});
-        assert(actual.length === expected.length,
-               "assert_array_equals", description,
-               "lengths differ, expected ${expected} got ${actual}",
-               {expected:expected.length, actual:actual.length});
+        function check_equal(actual_array, expected_array)
+        {
+            assert(actual_array instanceof Array,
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build BROKEN

Started: 2017-12-22 03:02:34
Finished: 2017-12-22 03:08:25


<h2>Failing Jobs</h2>
<ul>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/22118)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/320039400)

domenic:
This seems very bad. There are many existing tests that depend on the behavior of comparing object identity inside the arrays; it should not be changed.

If you have a specific test that needs to recurse, you can create your own function. But I think we should close this.

Honry:
@domenic, thanks for your comment. 

>There are many existing tests that depend on the behavior of comparing object identity inside the arrays; it should not be changed.

I only recurse array object in my commit, other object identity inside the arrays should be the same as the one before. But I agree with you that it's a big risk for existing tests.

So let's go back to the requirement, shall we recurse into the nested arrays in "assert_array_equals"?

domenic:
No. As I said, if you need this behavior in some test (please give an example), you can write your own separate helper.

Honry:
OK, let me close it.

Issue #8479, opened 2017-11-28
Acid3 fails everywhere


gsnedders:
We seem to have Acid 3 failing everywhere; I'm not entirely sure why the reference isn't matching, maybe the screenshot is getting taken too soon.

gsnedders:
One option here is to have two tests: one which is a hacked up `testharness.js` thing that gives per-test results (which means an implementation would notice going from 99 to 98 tests passing, rather the reftest only solution we have now which means the only thing that will ever be noticed is going from all passing and rendered correctly to not) and one which is what we have now.

gsnedders:
So in Firefox:

 * The "YOU SHOULD NOT SEE THIS AT ALL" of `#linktest` is visible while it doesn't have the `pending` class, implying it isn't being rendered with the `#linktest:visited` style. I can't see any related pref being set, but might there be some pref stopping that from applying, @jgraham?
 * With the internal reftest runner, we get different sizes of scrollbars because Acid3 is larger than 600x600.

In Chrome:
 * We get different sizes of scrollbars because Acid3 is larger than 600x600.
 * The "Acid3" text-shadow is slightly different (subpixel rounding?).

gsnedders:
The viewport issue makes this part of #5403, and another case where it would be resolved if we decided in #7598 to use 800x600 or larger, or alternatively could easily be fixed by #7135.

gsnedders:
wpt.fyi still has Acid3 fail everywhere, but it passes in Chrome locally now. (Still fails in Firefox for the reason given above.)

PR #8815, opened 2017-12-28
Add a testharness.js wrapper for Acid3

gsnedders:
This allows us to track individual subtests and their results, rather than simply a PASS/FAIL result for the entirety of Acid3.

See also #8479.

### DIFF
--- a/acid/acid3/numbered-tests.html
+++ b/acid/acid3/numbered-tests.html
@@ -9,7 +9,7 @@ function gotMessage(e) {
   var m = e.data;
   if (tests === undefined && "num_tests" in m) {
     tests = [];
-    for (var i = 0; i < m.tests; i++) {
+    for (var i = 0; i < m.num_tests; i++) {
       tests.push(async_test("Test " + i));
     }
   } else if ("result" in m) {
--- a/acid/acid3/numbered-tests.html
+++ b/acid/acid3/numbered-tests.html
@@ -1,0 +1,26 @@
+<!doctype html>
+<title>Acid3 numbered tests</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+var tests = undefined;
+
+function gotMessage(e) {
+  var m = e.data;
+  if (tests === undefined && "tests" in m) {
+    tests = [];
+    for (var i = 0; i < m.tests; i++) {
+      tests.push(async_test("Test " + i));
+    }
+  } else if ("result" in m) {
+    var test = m.test;
+    var passed = m.result === "pass";
+    var message = m.message;
+    tests[test].step_func_done(function() {
+      assert_true(passed, message);
+    })();
+  }
+}
+window.addEventListener("message", gotMessage, false);
+</script>
+<iframe src="test.html"></iframe>
--- a/acid/acid3/numbered-tests.html
+++ b/acid/acid3/numbered-tests.html
@@ -7,7 +7,7 @@ var tests = undefined;
 
 function gotMessage(e) {
   var m = e.data;
-  if (tests === undefined && "tests" in m) {
+  if (tests === undefined && "num_tests" in m) {
     tests = [];
     for (var i = 0; i < m.tests; i++) {
       tests.push(async_test("Test " + i));
@@ -16,9 +16,10 @@ function gotMessage(e) {
     var test = m.test;
     var passed = m.result === "pass";
     var message = m.message;
-    tests[test].step_func_done(function() {
+    tests[test].step(function() {
       assert_true(passed, message);
-    })();
+    });
+    tests[test].done();
   }
 }
 window.addEventListener("message", gotMessage, false);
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22208)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/322615747)

acid/acid3/test.html
```diff
@@ -3419,6 +3419,7 @@
     }
 
   ];
+  window.parent.postMessage({tests: tests.length}, "*");
```
Ms2ger: "tests" doesn't sound like a number to me. "num_tests", perhaps?

acid/acid3/numbered-tests.html
```diff
@@ -0,0 +1,26 @@
+<!doctype html>
+<title>Acid3 numbered tests</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+var tests = undefined;
+
+function gotMessage(e) {
+  var m = e.data;
+  if (tests === undefined && "tests" in m) {
```
Ms2ger: Any reason for the check on `tests`?

acid/acid3/numbered-tests.html
```diff
@@ -0,0 +1,26 @@
...
+    tests = [];
+    for (var i = 0; i < m.tests; i++) {
+      tests.push(async_test("Test " + i));
+    }
+  } else if ("result" in m) {
+    var test = m.test;
+    var passed = m.result === "pass";
+    var message = m.message;
+    tests[test].step_func_done(function() {
```
Ms2ger: I think the pattern of immediately calling the result of a `_func_` method is not very clear. Just do `step` + `done`?

Ms2ger (approved):

acid/acid3/numbered-tests.html
gsnedders: Not really. Just me being paranoid not really understanding how all of the Acid 3 JS works. :)

acid/acid3/numbered-tests.html
foolip: I agree.

acid/acid3/numbered-tests.html
foolip: Check or not check seems fine with me. To really be paranoid, making something blow up if assumptions aren't true would be better.

foolip (approved):
Looks like this would work, LGTM @Ms2ger's comments.

Running all the tests twice seems kind of pointless, so maybe we can disable test.html as a standalone test at least until some browser would pass it?

gsnedders:
@foolip I *think* the failure of the reference matching might be an infra issue?

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22830)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/329282323)

foolip:
> @foolip I *think* the failure of the reference matching might be an infra issue?

Isn't it just that nobody scores 100? If not can you file a bug for whatever the problem is? If nobody does score 100, I think we should disable the reftest part of this.

Ms2ger:
At least we're getting consistent results:

Subtest | Results | Messages
-- | -- | --
Acid3 numbered tests | FAIL: 10 | TypeError: tests[test] is undefined

Subtest | Results | Messages
-- | -- | --
Acid3 numbered tests | FAIL: 10 | Uncaught TypeError: Cannot read property 'step' of undefined

Subtest | Results | Messages
-- | -- | --
Acid3 numbered tests | FAIL: 10 | TypeError: undefined is not an object (evaluating 'tests[test].step')

Subtest | Results | Messages
-- | -- | --
Acid3 numbered tests | FAIL: 10 | Unable to get property 'step' of undefined or null reference

gsnedders:
@foolip We patched Acid3, similar to how it has been before, so that everyone gets 100 now, so it doesn't contradict known specs.

gsnedders:
@Ms2ger Better now? :)

gsnedders:
@Ms2ger ping

Ms2ger (approved):

foolip:
This is flaky in Chromium, @Hexcles had to mark it as such.

gsnedders:
Hmm, do you know what exactly is flaky? Running it with `wpt run --repeat` has got it to several thousand runs with no flake locally.

Hexcles:
@gsnedders I also struggle to understand what exactly is flaky. And I also cannot reproduce the issue with `wpt run`.

The symptom is that sometimes, especially with a slower debug build, we get *empty* output, not even the testharness.js header (the first line of the result). The weird thing is that somehow Blink test runner considers the test finishes (despite empty output), instead of a timeout which would be more reasonable.

@foolip all of this makes me suspect this is more likely related to Blink's test runner and/or testharnessreport.js.

PR #8834, opened 2018-01-02
Correct CSS2/css1/c414-flt-fit-000.xht and reference

csnardi:
The reference for CSS2/css1/c414-flt-fit-000.xht did not match the test, as it added the border size into the expected width, and did not have navy everywhere the test had it. This change fixes these issues by removing the border size from the expected width, and removing navy from both the test and reference.

### DIFF
--- a/css/CSS2/css1/c414-flt-fit-000-ref.xht
+++ b/css/CSS2/css1/c414-flt-fit-000-ref.xht
@@ -17,13 +17,12 @@
   margin-left: 10px;
   padding: 1em;
   table-layout: fixed;
-  width: 16.375em;
+  width: 16em;
   /*
    14em (5em column + 5em column + 4em column)
   + 2em (left and right horizontal padding)
-  + 6px (2 vertical 3px borders)
   =====
-  16.375em
+  16em
   */
   }
 
@@ -33,7 +32,6 @@
 
   td
   {
-  color: navy;
   padding: 0;
   }
   ]]></style>
--- a/css/CSS2/css1/c414-flt-fit-000.xht
+++ b/css/CSS2/css1/c414-flt-fit-000.xht
@@ -11,7 +11,7 @@
 
   <style type="text/css"><![CDATA[
    div { border: solid blue; padding: 1em; width: 14em; margin: 10px; }
-   div p { margin: 0; width: 5em; float: left; color: navy; }
+   div p { margin: 0; width: 5em; float: left; }
   ]]></style>
 
  </head>


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22264)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/324313740)

csnardi:
@gsnedders Mind reviewing? Should be fairly quick.

TalbotG:
@csnardi 

I reviewed the test and made that reference file. I remember having difficulties on it. 
But, sorry, I do not understand the changes you do.

1-
line 15:
-  border: blue solid medium;
+  border: blue solid;

[The initial border-width value is 'medium'.](https://www.w3.org/TR/CSS22/box.html#border-width-properties) So, what is the purpose of that change?

2-
> It did not take into account that the 'special' column had a default color value

But [color property does not apply to col elements](https://www.w3.org/TR/CSS22/tables.html#columns):
https://www.w3.org/TR/CSS22/tables.html#columns
and td elements do not inherit from col elements.

3-
```
    <col></col> <col></col> <col id="special"></col>
 
-   <tr><td>1</td><td>2</td><td>3</td></tr>
+   <tr><td>1</td><td>2</td><td id="special">3</td></tr>
 
-   <tr><td>4</td><td>5</td><td></td></tr>
+   <tr><td>4</td><td id="special">5</td><td></td></tr>
```

[id is a document-wide unique attribute.](https://www.w3.org/TR/html4/struct/global.html#adef-id) So, "special" can not be assigned to 3 id attributes at the same time.

Anyway... I still do not quite understand what you are saying. 

> This change fixes these issues so the test will pass.

Which browser fails the test? Over here, the test is passed with Firefox 52.7.3 and Chromium 66.

TalbotG:
@csnardi.

4-
`+  td#special {
+    color: initial;
+  }`

In CSS2.x, 'initial' is not a valid color name.

TalbotG:
@csnardi
Oh I see now. "3" and "5" should not be navy but rather using the initial color... which is normally/usually black. So, you're right: there is problem with and in the reference file.

I think the test should have been using only black for everything (border, numbers) since the test is really about the respective positions of numbers.

csnardi:
@TalbotG I changed the commit around to address your comments (removing navy entirely from the test). I kept the border as blue, as it doesn't seem to be causing any issues.

TalbotG:
Chris, 

I still do not see why you reach 16em instead of 16.375em given that: 

> The width of the table is the distance from the left inner padding edge to the right inner padding edge (including the border spacing but excluding padding and border).
**However, in HTML and XHTML1, the width of the `<table>` element is the distance from the left border edge to the right border edge.** 

CSS2.2, section 17.6.1 The separated borders model 
https://www.w3.org/TR/CSS21/tables.html#separated-borders

> The width of the table is then the greater of the value of the 'width' property for the table element and the sum of the column widths (plus cell spacing or borders).

CSS2.2, section 17.5.2.1 Fixed table layout
https://www.w3.org/TR/CSS22/tables.html#width-layout

In
[[CSS21] WD 07 Dec. 2010: errors in section 17.6.1 Separated borders model and width of table](http://lists.w3.org/Archives/Public/www-style/2011Feb/0652.html) 
, I reported an error in a diagram..
http://www.gtalbot.org/BrowserBugsSection/css21testsuite/width-table-BC-separated.html
versus
http://www.w3.org/TR/CSS21/images/tbl-spacing.png

TalbotG:
What most likely happens is this quote from CSS2.x, section 17.5.2.1 Fixed table layout
https://www.w3.org/TR/CSS22/tables.html#width-layout
:

> The width of the table is then the greater of the value of the 'width' property for the table element and the sum of the column widths (plus cell spacing or borders).

If your 16em is not wide enough for the 3 columns and table borders, then the used width algorithm takes place: 

> The width of the table **is then the greater of the value** of the 'width' property for the table element and the sum of the column widths (plus cell spacing or borders).

so, the used width of the table should be 16.375em . Both 16em (or anything less) and 16.375em should work and would work given that statement.

TalbotG:
If I set table width to a value smaller than 16.375em, then the algorithm given in previous comment indeed is applied in both Firefox 52 and Chromium 66. Eg with **table {width: 1em}** instead of 16em or 16.375em:

http://www.gtalbot.org/BrowserBugsSection/css22testsuite/ChrisNardiGithub8834.xhtml

gsnedders:
Locally, the whitespace following the 5 is too long in Firefox for me (yay font selection, I guess?), but aside from that and the colour issue it looks like it passes locally. Let me throw together a PR doing that much smaller change.

PR #8839, opened 2018-01-03
Make more python code parse as python3.

Ms2ger:

### DIFF
--- a/cors/resources/status.py
+++ b/cors/resources/status.py
@@ -24,7 +24,7 @@ def main(request, response):
     if request.method == "OPTIONS" and "preflight" in request.GET:
         try:
             code = int(request.GET.first('preflight'))
-        except KeyError, ValueError:
+        except KeyError:
             pass
 
     status = code, text
--- a/mathml/tools/utils/misc.py
+++ b/mathml/tools/utils/misc.py
@@ -1,7 +1,10 @@
 from __future__ import print_function
 import os
 import progressbar
-import urllib2
+try:
+    from urllib.request import urlopen
+except ImportError:
+    from urllib2 import urlopen
 
 MathMLAssociationCopyright = "Copyright (c) 2016 MathML Association"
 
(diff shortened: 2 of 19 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22273)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/324560367)

Hexcles:
Did you use any automated tool?

Ms2ger:
No, just ran the lint tool under python3.

Hexcles:
Would it be a good idea to use some tools like 2to3 (and rely on six)?

Ms2ger:
Everything here except for the `import` change will most likely be fixed the exact same way regardless of the approach we take for further changes, so I don't think we need to block on that decision.

Hexcles:
Oh no. By no means that'd be a blocker. I thought you were still working in progress so suggested something to convert all the code.

Thanks.

Ms2ger:
I will probably pick up more python3-related work in the future, but only when I have nothing better to do :)

mathml/tools/utils/misc.py
```diff
@@ -13,14 +16,14 @@ def downloadWithProgressBar(url, outputDirectory="./", forceDownload=False):
...
-    request = urllib2.urlopen(url)
+    request = urlopen(url)
     totalSize = int(request.info().getheader('Content-Length').strip())
     bar = progressbar.ProgressBar(maxval=totalSize).start()
 
     chunkSize = 16 * 1024
     downloaded = 0
     print("Downloading %s" % url)
-    os.umask(0002)
+    os.umask(0o002)
```
foolip: Looked a lot like cat on the keyboard, but apparently this is an octal literal: https://www.python.org/dev/peps/pep-3127/

cors/resources/status.py
```diff
@@ -24,7 +24,7 @@ def main(request, response):
     if request.method == "OPTIONS" and "preflight" in request.GET:
         try:
             code = int(request.GET.first('preflight'))
-        except KeyError, ValueError:
+        except KeyError:
```
foolip: This looks like a mistake should ValueError really be removed? `int('bla')` throws ValueError. Should be `except (KeyError, ValueError)`?

foolip (approved):
LGTM % the `except (KeyError, ValueError)` case

cors/resources/status.py
Ms2ger: That's not what the code did, though: in python 3, it'd be `except KeyError as ValueError`.

@jgraham, you wrote this code, thoughts? :)

foolip: Oh, wow, I guess it's a good thing that was changed in Python 3 :)

PR #8846, opened 2018-01-03
Detect cyclic objects when marshaling objects.

jgraham:
Marionette does currently not test for cyclic object references as
it marshals return values for transport across the wire.

Example of cyclic object:

	let obj = {};
	obj.cyclic = obj;

Passing this through evalaute.toJSON currently causes an infinite
recursion due to obj being referenced inside itself.  We can use
JSON.stringify to test if obj contains such cyclic values.  It is
assumed that the input to assert.acyclic is already JSON safe, so it can
be parsed by JSON.stringify, because of the previous checks it has made.

MozReview-Commit-ID: 4CnY2dcW5IF

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1106913 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8846)

### DIFF
--- a/webdriver/tests/execute_script/cyclic.py
+++ b/webdriver/tests/execute_script/cyclic.py
@@ -1,0 +1,48 @@
+from tests.support.asserts import assert_error
+
+
+def execute_script(session, script, args=None):
+    if args is None:
+        args = []
+    body = {"script": script, "args": args}
+    return session.transport.send(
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #8853, opened 2018-01-03
Use WebDriver conformant interactability checks for sendKeysToElement.

jgraham:
Enables webdriver spec keyboard interactability tests for 'Element Send Keys'
by default by re-using the same capability 'moz:webdriverClick' from
'Element Click'. It can be disabled by turning off this preference.

Also various webplatform tests for webdriver spec have been added which
cover both the scroll into view action, and keyboard interactability check.

Existing Marionette unit tests will be run in both modes, until we can
get rid of the legacy mode.

MozReview-Commit-ID: dFB8sQ6CN5

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1414322 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8853)

### DIFF
--- a/webdriver/OWNERS
+++ b/webdriver/OWNERS
@@ -4,3 +4,4 @@
 @lukeis
 @mjzffr
 @shs96c
+@whimboo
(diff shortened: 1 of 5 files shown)


jgraham:
Code reviewed upstream.

PR #8865, opened 2018-01-03
test] Use a common logger across suites from |mach test| and other test commands,

jgraham:
In order for |mach test| and |mach mochitest| to log an overall summary,
every test harness invocation they make needs to use the same structured
logger (otherwise the affected suite will be missing from the summary).

MozReview-Commit-ID: 8LJw7r8SItk

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1209463 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8865)


jgraham:
Code reviewed upstream.

PR #8866, opened 2018-01-03
Consider DOCUMENT_NODE a valid web element.

jgraham:
When looking up the parent of <html> using an XPath parent expression
such as "..", the nodeType of the returned HTMLDocument will be 9
(DOCUMENT_NODE).  <html> is a valid web element and we should be
able to serialise and return it to the user.

It is worth noting that other WebDriver implementations fail this
test because they fail on the nodeType check.

MozReview-Commit-ID: 4FMJEd8B4PZ

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1424635 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8866)


jgraham:
Code reviewed upstream.

PR #8868, opened 2018-01-03
Add tests that canceling an idle animation does not reject promises or fire events;

jgraham:
Note that we don't need tests for the cancel events of CSS Animations or CSS
Transitions since we already have them in
dom/animation/test/css-{animations|transitions}/file_event-dispatch.html

(See the tests named "Call Animation.cancel after cancelling
transition/animation".)

MozReview-Commit-ID: GRtSxEvMkjz

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1422248 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8868)


jgraham:
Code reviewed upstream.

PR #8872, opened 2018-01-03
Don't return the response from python handler,

jgraham:
This isn't how the wptserve API works; the return value is supposed to
be the content itself or headers, content or status, headers, content.

MozReview-Commit-ID: 28wZC65crW4

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1396534 [ci skip]

### DIFF
--- a/http/resources/securedimage.py
+++ b/http/resources/securedimage.py
@@ -6,12 +6,10 @@ def main(request, response):
     if "authorization" not in request.headers:
         response.status = 401
         response.headers.set("WWW-Authenticate", "Basic")
-        return response
     else:
         auth = request.headers.get("Authorization")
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

http/resources/securedimage.py
```diff
@@ -6,12 +6,10 @@ def main(request, response):
     if "authorization" not in request.headers:
         response.status = 401
         response.headers.set("WWW-Authenticate", "Basic")
-        return response
     else:
         auth = request.headers.get("Authorization")
         if auth != "Basic dGVzdHVzZXI6dGVzdHBhc3M=":
             response.set_error(403, "Invalid username or password - " + auth)
-            return response
```
Ms2ger: @jgraham seems like this is changing the code flow?

PR #8875, opened 2018-01-03
Fix inheritance of animation and transition properties of mismatched length.

jgraham:
At least when the animation-name length is bigger than the animation properties,
we mess up inheritance and only set properly the specified counts, then don't
cycle it.

The nicer fix for this is making these vectors properly, and move the cycling
logic at used-value time (bug 1420928). Same for transitions.

MozReview-Commit-ID: 3cguzIvfMFU

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1426246 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8875)


jgraham:
Code reviewed upstream.

PR #8879, opened 2018-01-03
Remove animationiteration from legacy event listener test, make the transition tests more robust, and re-enable that test.

jgraham:
According to :birtles, it is not guaranteed that animationiteration
event will be fired. This event is sample-based rather than event-based,
and such behavior has been clarified in CSS Animations Level 2:
https://drafts.csswg.org/css-animations-2/#event-dispatch

Also, Chromium has the same issue with this test:
https://bugs.chromium.org/p/chromium/issues/detail?id=701445

MozReview-Commit-ID: KBCzkGHxbfc

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1351409 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/8879)

### DIFF
--- a/dom/events/EventListener-invoke-legacy.html
+++ b/dom/events/EventListener-invoke-legacy.html
@@ -51,22 +51,16 @@ function runLegacyEventTest(type, legacyType, ctor, setup) {
 }
 
 function setupTransition(elem) {
-  elem.style.transition = '';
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #8881, opened 2018-01-03
Correct CSS2/tables/table-anonymous-objects-* tests

csnardi:
The base reference for many css/CSS2/tables/table-anonymous-objects-* tests was incorrect. Additionally, some tests indicated two references when it was clear only one could match. This change adds in two additional references so that all tests can match, and corrects the original base reference.

### DIFF
--- a/css/CSS2/tables/reference/no_red_3x3_monospace_table_colors-ref.xht
+++ b/css/CSS2/tables/reference/no_red_3x3_monospace_table_colors-ref.xht
@@ -3,7 +3,6 @@
 <html xmlns="http://www.w3.org/1999/xhtml">
    <head>
       <title>Reference rendering - no red, 3x3 monospace table</title>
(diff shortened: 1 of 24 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22322)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/324699579)

gsnedders:
FWIW: I'd really want to just reimport these from Gecko, given they're somewhat different tests in the CSS test suite, and were never originally automated. The Gecko tests are much simpler and cover everything as well.

gsnedders:
Enough of these still fail in Chrome/Safari in somewhat suspicious ways. Let's just reimport from Gecko.

PR #8886, opened 2018-01-04
[css-values-4] Tests for cyclic dependency resolution for the lh/rlh units

frivoal:
Goes along with the spec changes in https://github.com/w3c/csswg-drafts/issues/2115

### DIFF
--- a/css/css-values/lh-rlh-on-root-001.html
+++ b/css/css-values/lh-rlh-on-root-001.html
@@ -3,43 +3,84 @@
 <title>CSS Values and Units Test: using lh and rlh units on the root element</title>
 <link rel="author" title="Florian Rivoal" href="https://florian.rivoal.net/">
 <link rel="help" href="https://drafts.csswg.org/css-values-4/#font-relative-lengths">
+<style>
+#measure_me { position: absolute; }
+</style>
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 
+<div id=measure_me>&nbsp;</div>
+
 <script>
   function get_root_font_size() {
-    return window.getComputedStyle(window.document.documentElement).font_size;
+    return parseFloat(window.getComputedStyle(window.document.documentElement).fontSize);
   }
   function get_root_line_height() {
-    return window.getComputedStyle(window.document.documentElement).lineHeight;
+    /* getComputedStyle returns the computed value (not the used value) for the line-height property,
+       and the computed value of line-height:normal is normal,
+       so we cannot just query the value fo the proerty directly on the root element.
+       However the height of an abspos that only contains a single character from the first available font
+       and doesn't have any ancestor that changes the font-size or line-height property
+       gives us an indirect way to measure the root line-height in px.
+    */
+    return parseFloat(window.getComputedStyle(document.getElementById("measure_me")).height);
   }
 
   window.document.documentElement.style="font-size: initial; line-height:initial;";
   initial_f_s = get_root_font_size();
-  initial_l_h = get_root_line_height()
+  initial_l_h = get_root_line_height();
 
   test(function() {
-    window.document.documentElement.style="font-size: 42px; line-height: 1lh;";
+    window.document.documentElement.style="font-size: 142px; line-height: 1lh;";
     l_h = get_root_line_height();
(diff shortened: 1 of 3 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22333)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/324882584)

css/css-values/lh-unit-001.html
```diff
@@ -0,0 +1,26 @@
...
+div {
+  font-size: 50px;
+  line-height: 1;
+  width: 100px;
+  height: 100px;
+  background: red;
+}
+aside {
+  background: green;
+  font-size: 42px; /* number doesn't matter, as long as it's not 100 */
```
FremyCompany: "not either 100 nor 50"?

css/css-values/lh-rlh-on-root-001.html
```diff
@@ -0,0 +1,45 @@
...
+    assert_equals( f_s, initial_f_s, "the lh unit on the root element's font-size property uses font metrics corresponding to the initial values of the font or line-height properties");
+  }, "lh in font-size on root");
+
+  test(function() {
+    window.document.documentElement.style="font-size: 1rlh; line-height: 42px;";
+    f_s = get_root_font_size();
+    assert_equals( f_s, initial_f_s, "the rlh unit on the root element's font-size property uses font metrics corresponding to the initial values of the font or line-height properties");
+
+  }, "rlh in font-size on root");
+</script>
```
FremyCompany: Shouldn't there be a test for what happens if you specify 2lh or 2rlh?

css/css-values/lh-unit-002.html
```diff
@@ -0,0 +1,26 @@
...
+div {
+  width: 100px;
+  height: 100px;
+  background: red;
+  line-height: 50px;
+}
+aside {
+  background: green;
+  height: 1em;
+  line-height: 42px; /* number doesn't matter, as long as it's not 50 */
```
FremyCompany: "not either 50 nor 100"?

css/css-values/lh-rlh-on-root-001.html
frivoal: No sure. In general, having a test checking that the unit can be use with numbers other than 1 sounds sane, but I am not sure it is needed as part of the unit cyclic dependency resolution problem, which is what I am testing here. On top of that, I am a little worried here that I'd easily run into rounding errors on this test if I started doing multiplications, especially for `rlh` where I have no way to for the initial value to be a round number.

As for testing that the unit in general works and makes sense with numbers other than one, I'm planning to write another series of tests about these units checking that they are sized correctly, now that we have resolved on https://github.com/w3c/csswg-drafts/issues/1253. Would that be enough, or do you think there is a need for this within the cyclic dependency tests?

css/css-values/lh-unit-001.html
frivoal: Agreed, and fixed.

css/css-values/lh-unit-002.html
frivoal: Agreed, and fixed.

frivoal: Agreed, and fixed.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22626)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/327495423)

css/css-values/lh-rlh-on-root-001.html
FremyCompany: I think you might want a test that checks that, at least, specifying 2lh does something (as opposed to just being ignored by the browser instead of correctly cycle-resolved to 2x the default value). Checking that the height becomes taller than the initial one might be enough (or maybe larger than 1.75x more and smaller than 2.25x less?)

frivoal: Actually, that was a good idea, sorry for pushing back initially. Also, while fixing that, I found other problems with the test, and fixed them as well. Can you have a look again?

FremyCompany (approved):

frivoal:
Thanks!

PR #8894, opened 2018-01-04
WebKit export of https://bugs.webkit.org/show_bug.cgi?id=181239

youennf:
Implement  https://fetch.spec.whatwg.org/#main-fetch default referrer policy setting

### DIFF
--- a/service-workers/service-worker/fetch-event-referrer-policy.https.html
+++ b/service-workers/service-worker/fetch-event-referrer-policy.https.html
@@ -7,12 +7,19 @@
 <script>
 var worker = 'resources/fetch-event-test-worker.js';
 
+function do_test(referrer, value, expected, name)
+{
+    test(() => {
+          assert_equals(value, expected);
+    }, name + (referrer ? " - Custom Referrer" : " - Default Referrer"));
+}
+
 function run_referrer_policy_tests(frame, referrer, href, origin) {
     return frame.contentWindow.fetch('resources/simple.html?referrerFull',
                                      {method: "GET", referrer: referrer})
       .then(function(response) { return response.text(); })
       .then(function(response_text) {
-          assert_equals(
+          do_test(referrer,
             response_text,
             'Referrer: ' + href + '\n' +
             'ReferrerPolicy: no-referrer-when-downgrade',
@@ -24,7 +31,7 @@ function run_referrer_policy_tests(frame, referrer, href, origin) {
         })
       .then(function(response) { return response.text(); })
       .then(function(response_text) {
-          assert_equals(
+          do_test(referrer,
             response_text,
             'Referrer: \n' +
             'ReferrerPolicy: no-referrer-when-downgrade',
@@ -34,7 +41,7 @@ function run_referrer_policy_tests(frame, referrer, href, origin) {
         })
       .then(function(response) { return response.text(); })
       .then(function(response_text) {
-          assert_equals(
+          do_test(referrer,
             response_text,
             'Referrer: ' + href + '\n' +
             'ReferrerPolicy: no-referrer-when-downgrade',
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22345)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/324932184)

youennf:
The lint issue is about the use of window.internals.
In this case, if window.internals is there, it will lower non mandatory security resource loading checks.
This test is indeed doing some HTTP loads from an HTTPS web page.

youennf:
@foolip, @gsnedders, is it fine to go on with this PR as is.

foolip:
@youennf, do you mean that you think it should be merged as-is, with a lint exception added, or not merged?

I suppose that to follow the model of Chromium, Gecko and Servo exports we shouldn't be doing any reviewing here and instead do that in follow-ups. But a few things to note:
* Is it possible to integrate `wpt lint` into WebKit bot steps so that failing lints are noticed during WebKit review?
* If this is imported into Chromium, it'll break there, because `setAllowDisplayOfInsecureContent` doesn't exist, but `window.internals.settings` does.

Not sure how best to proceed. Will the test simply fail again without the `setAllowDisplayOfInsecureContent` in WebKit? Does the test go beyond what any spec requires, or why does optional behavior need to be toggled on here?

On `internals` and `eventSender` generally, I think it would be wise to avoid that and try to put the required APIs into testdriver.js, although it is more work, I realize that. @kereliuk FYI that this is a test automation challenge, maybe a new one.

@leonhsl who created this test in https://chromium-review.googlesource.com/c/chromium/src/+/567963 for opinions.

service-workers/service-worker/fetch-event-referrer-policy.https.html
```diff
@@ -7,12 +7,22 @@
 <script>
 var worker = 'resources/fetch-event-test-worker.js';
 
+if (window.internals && window.internals.settings)
+    internals.settings.setAllowDisplayOfInsecureContent(true);
```
wanderview: @youennf What is this?  Generally this sort of thing is managed in the harness integration with the browser vendor's test suite.  We don't usually put non-standard stuff like this in the WPT to my knowledge.

mattto: JFYI for Blink we're somehow able to use testharnessreport.js to add internal settings. Can WebKit have a similar setup? For reference here's the change to the file in the patch foolip linked to above:
https://chromium-review.googlesource.com/c/chromium/src/+/567963/3/third_party/WebKit/LayoutTests/resources/testharnessreport.js

youennf: Thanks @matto.
We are also using testharnessreport.js for these kind of setup but not on a per test basis right now.
I guess this might be fine as a temporary solution.
Adding this kind of configuration within the test itself using something like window.test_driver might be a better approach as other browser teams will more easily understand what needs to be done.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/23738)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/336186167)

youennf:
> Is it possible to integrate wpt lint into WebKit bot steps so that failing lints are noticed during WebKit review?

This is indeed a good idea and is on the table.
Hopefully, there will be some progress on our side soon.

foolip:
What should we do to get this merged? Just cut out the `window.internals` bits? Or wait for a follow-up change in WebKit and that export PR, so that we can merged both PRs at once?

youennf:
For now, I removed the window.internals bits.
I will look at how using testdriver.js to relax the security checks for this particular test.

foolip (approved):
I haven't reviewed the tests as that's been done in WebKit, but approving in case that's all that's blocking merging this.

gsnedders:
Is there any reason this hasn't landed?

PR #8899, opened 2018-01-04
Add a minimal test for the :user-invalid psuedo-class

foolip:

### DIFF
--- a/css/selectors/user-invalid.html
+++ b/css/selectors/user-invalid.html
@@ -1,0 +1,20 @@
+<!doctype html>
+<title>Support for the :user-invalid pseudo-class</title>
+<link rel="help" href="https://drafts.csswg.org/selectors/#user-pseudos">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<input>
+<script>
+test(() => {
+  const input = document.querySelector('input');
+  // The selector can't match because no interaction has happened.
+  assert_false(input.matches(':user-invalid'));
+}, ':user-invalid selector should be supported');
+
+// historical: https://github.com/w3c/csswg-drafts/issues/1329
+test(() => {
+  const input = document.querySelector('input');
+  // matches() will throw if the selector isn't suppported
+  assert_throws("SyntaxError", () => input.matches(':user-error'));
+}, ':user-error selector should not be supported');
+</script>


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22362)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/325081553)

tabatkins (approved):

w3c-bots:
# Build ERRORED

Started: 2018-01-05 18:34:05
Finished: 2018-01-05 18:37:57


<h2>Failing Jobs</h2>
<ul>

<li>manifest_upload</li>

<li>lint</li>

<li>update_built</li>

<li>firefox:nightly</li>

<li>wpt_integration in py27</li>

<li>build_css</li>

<li>chrome:dev</li>

<li>safari:10.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/22362)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/325081553)

fantasai:
@foolip Please don't add tests for any CSS features that detect only parsing-level or otherwise super-basic-not-actually-functional support, unless there's a corresponding test for functional support: we don't want UAs to implement parsing support for a feature without actually supporting the feature, because this breaks CSS's forwards-compatible parsing invariants.

foolip:
@fantasai, I think you'll want to comment on https://github.com/w3c/web-platform-tests/issues/9113, which is about generating such tests, similar to idlharness.js tests. I think that the fact that they'd be in a large group of such generated tests make it less likely think that passing just those tests could suffice to implement a feature, and it'd reveal quickly which things aren't implemented at all, anywhere, which seems valuable.

In the absence of any tests, is there someone who goes to write some ad-hoc such tests at each spec state transition, to determine which things aren't implemented and should be removed?

frivoal:
> In the absence of any tests, is there someone who goes to write some ad-hoc such tests at each spec state transition, to determine which things aren't implemented and should be removed?

But such tests aren't enough to make that decision. Even if tons of parsing tests pass, the features implemented could be widely incompatible, and warrant being pushed to the next level while the actually stable part of a spec makes its way to REC.

And in the case where it has not been implemented at all, given that removing things from one level when publishing a spec is a manual decision taken in presence of browser vendor representatives, "Does any body implement this yet?" takes way less effort to ask than adding, reviewing, merging, running, and checking test results.

foolip:
OK, I think what you're saying is that determining if a feature has been implemented *at all* by asking in a F2F meeting is less work than adding a minimal test and finding the results of those test later.

How is the list of things for which to ask "has this been implemented at all" generated? Is the entire changelog since the last time around studied carefully, or do small changes that don't match implementations get to REC once in a while?

frivoal:
Recent CSS-WG policy is that every normative change since the last CR needs tests before we can even go to another CR, let alone REC.
RECs need a test report, and evaluating the coverage of the test suite is a necessary step for producing such a report.
Accidents happen, mistakes can be made, etc, but I would not expect the CSSWG to bring something to REC with failing tests, or (large) gaps in the test suite without a discussion of these specific gaps and failures, and granting exceptions for valid reasons.
For instance, in https://drafts.csswg.org/css-ui-3/implementation-report 

So, the list of "has this been implemented at all" is generated at the latest by the editor(s) when trying to exit CR (at which points tests will provide the answers), and earlier as well, by the editors or more members of the working group, in order to adequately mark things are risk (and push them to the next level) so that we don't get stuck later on why trying to exit CR.

frivoal:
For instance, in https://drafts.csswg.org/css-ui-3/implementation-report and the issue linked from there https://github.com/w3c/csswg-drafts/issues/1770, we went through the list of all optional features/normative statements that had only partial support, and made decisions about each individually.

PR #8934, opened 2018-01-05
Fetch destination - worker and document tests

### LINKED ISSUE #8810: Test values of `Request.destination`
There are multiple tests that check if `Request.destination` is exposed, but none I could find that check if the returned value is the right one (other than https://github.com/w3c/web-platform-tests/blob/master/preload/fetch-destination.https.html which tests the destination value for `<link rel=preload as=fetch>`.

We need more tests to make sure the values returned by the different engines are compatible with the spec and with each other.

yoavweiss:
Resolves #8810

This adds worker and document tests. It doesn't include all destinations, but hopefully enough to cover the "major" ones for the sake of compatibility.

### DIFF
--- a/fetch/api/request/destination/fetch-destination-no-load-event.https.html
+++ b/fetch/api/request/destination/fetch-destination-no-load-event.https.html
@@ -121,11 +121,4 @@ promise_test(async t => {
     await waitOnMessageFromSW(t);
 }, '@import fetches with a "style" Request.destination');
 
-// worker destination
-/////////////////////
-promise_test(async t => {
-    let worker = new frame.contentWindow.Worker("dummy?t=worker&dest=worker");
-    await waitOnMessageFromSW(t);
-}, 'Worker fetches with a "worker" Request.destination');
-
 </script>
--- a/fetch/api/request/destination/fetch-destination-no-load-event.https.html
+++ b/fetch/api/request/destination/fetch-destination-no-load-event.https.html
@@ -121,4 +121,11 @@ promise_test(async t => {
     await waitOnMessageFromSW(t);
 }, '@import fetches with a "style" Request.destination');
 
+// worker destination
+/////////////////////
+promise_test(async t => {
+    let worker = new frame.contentWindow.Worker("dummy?t=worker&dest=worker");
+    await waitOnMessageFromSW(t);
+}, 'Worker fetches with a "worker" Request.destination');
+
 </script>
--- a/fetch/api/request/destination/fetch-destination-worker.https.html
+++ b/fetch/api/request/destination/fetch-destination-worker.https.html
@@ -1,0 +1,53 @@
+<!DOCTYPE html>
+<title>Fetch destination tests for resources with no load event</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/common/get-host-info.sub.js"></script>
+<script src="/service-workers/service-worker/resources/test-helpers.sub.js"></script>
+<script>
+let frame;
+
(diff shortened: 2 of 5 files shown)


fetch/api/request/destination/fetch-destination-no-load-event.https.html
```diff
@@ -121,4 +121,11 @@
     await waitOnMessageFromSW(t);
 }, '@import fetches with a "style" Request.destination');
 
+// worker destination
+/////////////////////
+promise_test(async t => {
+    let worker = new frame.contentWindow.Worker("dummy?t=worker&dest=worker");
```
wanderview: Current implementations incorrectly inherit the service worker controller from the parent document for worker scripts.  The spec says they should do their own matching against the scope.  Since this is kind of besides the point of this test it would be nice if you could make this work for either case.

If you change the scope to `resources/dummy` without the trailing .html and search params then I think this test will be valid for both current implementations and what the spec says.  It will do a substring match for both the iframe and then worker script.

You would of course need to change the `iframe.src = kScope` to set the html suffix and search params as well.

wanderview (approved):

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22431)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/325543513)

yoavweiss:
@wanderview Thanks for reviewing! :) Moved the worker test to its own file and changed its scope (kept the `.html` extension though, because without it the navigation resulted in a download. PTAL

wanderview:
@yoavweiss , there are other tests using `dummy.html`.  You can't rename it without breaking those other tests.  Can you put that back?  I don't understand why it needed to be renamed.

fetch/api/request/destination/fetch-destination-worker.https.html
```diff
@@ -0,0 +1,53 @@
...
+    }).catch(() => {;
+        assert_unreached("Wrong destination.");
+    });
+    t.add_cleanup(() => { frame.contentWindow.navigator.serviceWorker.onmessage = null; });
+}
+
+// worker destination
+/////////////////////
+promise_test(async t => {
+    let worker = new frame.contentWindow.Worker("dummy.html?t=worker&dest=worker");
```
wanderview: nit:  Maybe add a comment here that we can use a .html because we don't care if it actually loads or not.  We just want to see the FetchEvent for the script load.

fetch/api/request/destination/resources/fetch-destination-worker-iframe.js
```diff
@@ -0,0 +1,22 @@
+self.addEventListener('fetch', function(event) {
+    if (event.request.url.includes('dummy')) {
+        event.waitUntil(async function() {
+            let destination = new URL(event.request.url).searchParams.get("dest");
+            await self.clients.matchAll({"includeUncontrolled": true})
+                .then(function(clients) {
```
wanderview: nit: You can avoid the `then()` and its nesting since you are in an async function.  Something like:

```javascript
let clients = await self.clients.matchAll({includeUncontrolled: true});
clients.forEach(...)
```

wanderview (approved):

fetch/api/request/destination/fetch-destination-worker.https.html
yoavweiss: Added

fetch/api/request/destination/resources/fetch-destination-worker-iframe.js
yoavweiss: good call :)

PR #8945, opened 2018-01-08
[css-writing-modes] Tests for Available Space in Orthogonal Flows

frivoal:
Relates to w3c/csswg-drafts#1391


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22464)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/326235810)

fantasai:
available-size-001
- Using the wrap-point to show off the test block's height is... interesting.

available-size-002
- OK, wow, so I think this is a spec issue: whether a max-height on the direct parent of the orthogonal flow limits the space available to the orthogonal flow. If you'd used an ancestor rather than the direct parent, then the test would undoubtedly be correct, however.

available-size-003
available-size-004
available-size-005
- Based on the assert, I would not expect a double orthogonal case here. It should be sufficient to have an orthogonal flow inside an auto-height container inside a fixed-height scroller.
- “like up” -> “line up”

available-size-006
available-size-007
- There's a fairly large series of tests called sizing-orthog-* which I believe cover these cases already. Your tests are not wrong, and they're pretty nice, but do you still want to add them?

available-size-008
- “sroller” -> “scroller”
- The scroller doesn't have to be taller, necessarily. It's just that if we would use the max-height because it's fixed, but it's taller than the ICB, we use the ICB instead. So the min-height here can go (and the assert updated).


available-size-009
- “sroller” -> “scroller”

available-size-010-ref
- I can't tell the difference between available-size-010-ref and available-size-005-ref. Any reason to have both?

available-size-010
available-size-011
- If this is the same as -001 but we're adding padding just to be sure you're using the correct box... just update 001 to have padding/border/margin and call it a day. Ditto 011.

frivoal:
> available-size-010
> available-size-011
>
>     If this is the same as -001 but we're adding padding just to be sure you're using the correct box... just update 001 to have padding/border/margin and call it a day. Ditto 011.

It's true that passing 010 is sufficient, as you probably cannot do that without also doing what's needed to pass 001, however I found it interesting to test separately because safari passes 001 but not 010. My logic was that having both was useful because:
- failing 010 would push safari to fix that part of the problem
- passing 001 would help make sure safari don't regress on that part of the solution

I've you're OK with that logic, I'd like to leave it the variants in.

frivoal:
> available-size-006
> available-size-007
> There's a fairly large series of tests called sizing-orthog-* which I believe cover these cases already. Your tests are not wrong, and they're pretty nice, but do you still want to add them?

Right, I was aware of these tests, but found them difficult enough to understand that it ended up taking less time to write my new tests than to convince me the existing tests were testing the exact same thing I wanted to. Given who's written them, I believe they are very likely to be correct, but I thought that by writing tests for that same topic independently, at worse they ended up being effectively duplicate, and at best they would be different in some subtle way that might catch bugs in some implementation. If you're convinced mine are fully redundant with the pre-existing tests, I'm OK with removing them. Otherwise, I'd live them in.

frivoal:
> Using the wrap-point to show off the test block's height is... interesting.

Is that re request to do things in a less "smart" (i.e. hard to understand) way (and if so, what do you suggest?), or just a passing comment as you read along?

I wasn't trying to be smart, that was just the approach I came with when trying to do this in a way that made sense to me.

frivoal:
> Based on the assert, I would not expect a double orthogonal case here. It should be sufficient to have an orthogonal flow inside an auto-height container inside a fixed-height scroller.

In retrospect, yes, it is overkill. On the other hand, it doesn't seem wrong. It seems that it does match the condition described in the assert, albeit in a more heavy handed way than necessary. Since that has the potential to trip up browsers **and** still seems to match the description, I'd argue for leaving them in. Adding the simpler variant you described as well does seem like a good idea though.

frivoal:
> >  Based on the assert, I would not expect a double orthogonal case here. It should be sufficient to have an orthogonal flow inside an auto-height container inside a fixed-height scroller.
>
> In retrospect, yes, it is overkill. On the other hand, it doesn't seem wrong. It seems that it does match the condition described in the assert, albeit in a more heavy handed way than necessary. Since that has the potential to trip up browsers and still seems to match the description, I'd argue for leaving them in. Adding the simpler variant you described as well does seem like a good idea though.

I've written the simplified tests as you suggested, and they do not bring out different bugs from browsers than the double orthogonal tests I had earlier, so I now agree with just simplifying, and not keeping both. Latest commit does that.

frivoal:
@fantasai Thank you for the review. I've now fixed or responded to all your points. Can you give it a second look?

fantasai (approved):
“however I found it interesting to test separately because safari passes 001 but not 010”

That's because 010 is a more rigorous test. I don't think 001 is adding anything interesting: if you can handle 10px padding, I'm pretty sure you can handle 0px padding. So I'd prefer to drop the simpler test, as I don't feel it's useful for it to take up space in the test suite.

Approving so you can merge; would definitely prefer you drop the less-rigorous version of the test unless you have some strong objection to doing so, though.

frivoal:
Thanks. I dropped 001 and 003 in favor of 010 and 011, and renumbered them accordingly.
Merging.

fantasai:
Filed spec issue as https://github.com/w3c/csswg-drafts/issues/2239

PR #8950, opened 2018-01-08
Reject the promise if the video fails to load in imagebitmap/common.js.

Ms2ger:

### DIFF
--- a/2dcontext/imagebitmap/common.js
+++ b/2dcontext/imagebitmap/common.js
@@ -38,6 +38,7 @@ function makeVideo() {
         video.oncanplaythrough = function() {
             resolve(video);
         };
+        video.onerror = reject;
         video.src = "/images/pattern.ogv";
     });
 }


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22479)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/326348512)

Ms2ger:
@youennf fyi; this change should help this test fail instead of timing out on macOS

2dcontext/imagebitmap/common.js
```diff
@@ -38,6 +38,7 @@ function makeVideo() {
         video.oncanplaythrough = function() {
             resolve(video);
         };
+        video.onerror = reject;
```
youennf: Looks good to add onerror in any case.
Test might fail in WebKit as the test tries to load ogv.
Can we use common/media.js to abstract this test of specific video formats?

youennf (changes requested):

2dcontext/imagebitmap/common.js
youennf: Look for getVideoURI for examples

Ms2ger: I think the needed pattern.mp4 doesn't exist yet; someone would need to create it first.

youennf (approved):
OK, let's go with this rejection for now.
Having pattern.mp4 would be nice.

PR #8951, opened 2018-01-08
Add testcase for unclosed attribute value selector

### LINKED ISSUE #549: Inconsistency in execution of unclosed attribute selector for querySelector
(Note: I hope this is the correct repository to file a bug. If not, please move this issue to the correct place)

In https://github.com/PolymerElements/iron-form/pull/222#discussion_r157619096 we discovered an inconsistency in parsing and execution of invalid `querySelector` with an unclosed attribute selector:

```js
element.querySelector('[name=input1')
```

In our [Travis build](https://travis-ci.org/PolymerElements/iron-form/builds/317379019#L2478-L2540), all OS X Safari tests failed with the error `SyntaxError: DOM Exception 12: An invalid or illegal string was specified.`.

However, all other browsers happily accept the unclosed attribute selector and do not throw an error. Based on this finding, I browsed [the DOM standard](https://dom.spec.whatwg.org/#dom-element-matches) which pointed to [this section in Selectors Level 4](https://www.w3.org/TR/selectors4/#attribute-representation). In this case, the DOM specification states that a failure in parsing of the selector must throw a `SyntaxError`. However, based on [the specification of invalid CSS Selectors](https://www.w3.org/TR/selectors4/#invalid), the notion of invalid denotes an unexpected token or namespace prefix. It is unclear what an unclosed selector denotes, which also means that (imo) the behavior of unclosed selectors is undefined.

Based on the undefinedness, it appears that browsers interpret the "failure of parsing" in multiple ways and depending on this condition either throw or don't throw a `SyntaxError` per the DOM standard. 

Therefore I would like to know what the specced behavior of "failure of parsing" is or should be. In the case that it is already specced, maybe a `web-platform-test` can be added to verify this behavior to reduce the inconsistency?

TimvdLippe:
Per https://github.com/whatwg/dom/issues/549, all browsers except Safari accept this selector. Therefore, this test case verifies that unclosed brackets are valid and do not throw a `SyntaxError`. I verified this test-case passing in Firefox and Chrome, but I do not have a Safari browser to test this on. Would be greatly appreciated if someone could run this in Safari and verify this test case fails (as expected).

Closes https://github.com/whatwg/dom/issues/549

### DIFF
--- a/dom/nodes/selectors.js
+++ b/dom/nodes/selectors.js
@@ -90,6 +90,7 @@ var validSelectors = [
 
   // - value                     [att=val]
   {name: "Attribute value selector, matching align attribute with value",                                    selector: "#attr-value [align=\"center\"]",                                     expect: ["attr-value-div1"], level: 2, testType: TEST_QSA | TEST_MATCH},
+  {name: "Attribute value selector, matching align attribute with value, unclosed bracket",                  selector: "#attr-value [align=\"center\"",                                      expect: ["attr-value-div1"], level: 2, testType: TEST_QSA | TEST_MATCH},
   {name: "Attribute value selector, matching align attribute with empty value",                              selector: "#attr-value [align=\"\"]",                                           expect: ["attr-value-div2"], level: 2, testType: TEST_QSA | TEST_MATCH},
   {name: "Attribute value selector, not matching align attribute with partial value",                        selector: "#attr-value [align=\"c\"]",                                          expect: [] /*no matches*/,   level: 2, testType: TEST_QSA},
   {name: "Attribute value selector, not matching align attribute with incorrect value",                      selector: "#attr-value [align=\"centera\"]",                                    expect: [] /*no matches*/,   level: 2, testType: TEST_QSA},


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22489)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/326395098)

annevk:
Verified that Safari ends up failing this. I'm also positive it's a bug in Safari, just don't know exactly how CSS goes about defining this anymore.

@TimvdLippe did you file a bug against WebKit already?

annevk (approved):

TimvdLippe:
@annevk I did not, let me try to figure out how that works.

annevk:
@TimvdLippe start at https://bugs.webkit.org/enter_bug.cgi?product=WebKit. Component doesn't matter so much.

TimvdLippe:
@annevk Thanks, I figured it out. When I wanted to create the bug, it pointed to potential duplicates and the issue appears to be filed at https://bugs.webkit.org/show_bug.cgi?id=142167 already, so Apple is aware of it.

annevk:
Found https://bugzilla.mozilla.org/show_bug.cgi?id=1137984#c13 via that which confirms that Selectors might still need to be updated to account for this.

TimvdLippe:
@annevk Ah good find! Yes maybe more tests need to be added, but let's see what @tabatkins decided after 3 years :smile:

tabatkins:
Yes, Selectors is updated now, and is written in terms of a CSS grammar, which means you do a CSS parse on the string to turn it into tokens, which means blocks get closed automatically and you can't tell the difference between `[foo` and `[foo]`.

PR #8976, opened 2018-01-10
wpt-export-for-webkit-181330

youennf:
Add testing for redirect URLs containing credentials in CORS mode for images and no cors mode for fetch, image and frame

### DIFF
--- a/fetch/security/redirect-to-url-with-credentials.https.html
+++ b/fetch/security/redirect-to-url-with-credentials.https.html
@@ -1,0 +1,69 @@
+<html>
+<header>
+    <script src="/resources/testharness.js"></script>
+    <script src="/resources/testharnessreport.js"></script>
+    <script src="/common/get-host-info.sub.js"></script>
+</header>
+<body>
+<script>
+var host = get_host_info();
+
+var sameOriginImageURL = "/common/redirect.py?location=" + host.HTTPS_ORIGIN_WITH_CREDS + "/service-workers/service-worker/resources/fetch-access-control.py?ACAOrigin= " + host.HTTPS_ORIGIN + "%26PNGIMAGE%26ACACredentials=true";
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22579)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/327262441)

annevk:
It seems the lint check is faiilng, but also, aren't we already testing this somewhere in the /cors directory? It's somewhat surprising to me this would not be covered. Although I guess for "no-cors" we might indeed not cover this. (I haven't yet read all the context though, just a drive-by comment before I have to go for a bit.)

youennf:
> It seems the lint check is faiilng, but also, aren't we already testing this somewhere in the /cors directory?

I saw tests for fetch cors mode and XHR.
I did not see any test for fetch no-cors mode, hence why I added one for cors and no-cors.
I did not see any test related to loading images but I haven't looked at the cors folder.

annevk (approved):
These match the current specification so I'm happy to land them and then we can change them if https://github.com/whatwg/fetch/pull/465 lands.

It might be good to wait with landing this until @mikewest has had a chance to take a look though given that he wants to change this.

(The only somewhat weird thing I noticed is that the fetch() calls are without credentials but for images you use "use-credentials". It shouldn't matter what that setting is though, but maybe if it did in WebKit you want to test more variants?)

annevk (changes requested):
Actually, the lint issues do need to be fixed.

foolip:
@youennf, I see the commit message for this one looks a bit different from https://github.com/w3c/web-platform-tests/pull/8986 and the other WebKit exports. Is there a format I can grep for, for tracking purposes?

mikewest:
I think these tests all match what the spec says, so LGTM from that perspective.

That said, I expect that Chrome will fail them, and I'm not enthusiastic about not failing them. I apparently dropped the ball on the Fetch patch; I'll try to pick that up again in the near future to try to get other vendors on board.

annevk:
The reason lint is complaining btw is that instead of setTimeout you should use test.step_timeout.

annevk (approved):

PR #8990, opened 2018-01-11
Parsing tests for clip, clip-path, clip-rule

ewilligers:
Spec: https://drafts.fxtf.org/css-masking-1/

Blink, Edge, Firefox, WebKit don’t support clip rect containing auto.

Blink fails to reject clip-path ellipse containing single radius.

Blink fails to support clip-path containing geometry-box

Edge, Firefox, WebKit fail to support clip-path containing basic-shape
and/or geometry-box

### DIFF
--- a/css/css-masking/parsing/clip-path-invalid.html
+++ b/css/css-masking/parsing/clip-path-invalid.html
@@ -1,0 +1,48 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>CSS Masking Module Level 1: parsing clip-path with invalid values</title>
+<link rel="author" title="Eric Willigers" href="mailto:ericwilligers@chromium.org">
+<link rel="help" href="https://drafts.fxtf.org/css-masking-1/#the-clip-path">
+<meta name="assert" content="clip-path supports only the grammar '<clip-source> | [ <basic-shape> || <geometry-box> ] | none'.">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="resources/parsing-testcommon.js"></script>
+</head>
+<body>
+<script>
+test_invalid_value("clip-path", "auto");
+test_invalid_value("clip-path", "path('m 12 34 h 56')");
+test_invalid_value("clip-path", "ray(0deg)");
+
+test_invalid_value("clip-path", "inset()");
+test_invalid_value("clip-path", "inset(123)");
+test_invalid_value("clip-path", "inset(1% 2% 3% 4% 5%)");
+test_invalid_value("clip-path", "inset(round 0)");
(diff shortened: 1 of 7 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22628)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/327508963)

css/css-masking/parsing/clip-path-invalid.html
```diff
@@ -0,0 +1,48 @@
...
+<link rel="help" href="https://drafts.fxtf.org/css-masking-1/#the-clip-path">
+<meta name="assert" content="clip-path supports only the grammar '<clip-source> | [ <basic-shape> || <geometry-box> ] | none'.">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="resources/parsing-testcommon.js"></script>
+</head>
+<body>
+<script>
+test_invalid_value("clip-path", "auto");
+test_invalid_value("clip-path", "path('m 12 34 h 56')");
```
dirkschulze: Why is this invalid? path() was added to CSS Shapes Level 2, right? Are you testing the first level of the specs only?

It probably will end up clipping the entire element because of the shape being 1D though.

css/css-masking/parsing/resources/parsing-testcommon.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+function test_valid_value(property, value, serializedValue) {
```
dirkschulze: There is nothing like that in testharness already? Wonder if that should e considered adding. After all there are many specs which define CSS properties.

css/css-masking/parsing/clip-path-invalid.html
ewilligers: I followed the link from
https://drafts.fxtf.org/css-masking-1/#the-clip-path
to
https://drafts.csswg.org/css-shapes-2/#typedef-basic-shape
"The shape is computed based on the values of one of inset(), circle(), ellipse() or polygon()."

Now removed from this pull request.

css/css-masking/parsing/resources/parsing-testcommon.js
ewilligers: I think testharness shouldn't be CSS specific. It might be good to have a css/resources directory that all the CSS specs can access.

This parsing-testcommon.js can easily move to such a directory if/when it is created.

dirkschulze (approved):

birtles (approved):
I'm just approving this as proxy for Dirk since I believe he is a suitable reviewer but it appears he doesn't have write access to wpt.

PR #8995, opened 2018-01-11
WebCryptoAPI: secure contexts cleanup

annevk:
This makes the following changes:
* Remove the duplicate idlharness.html as we now require secure contexts (idlharness.https.html serves this purpose).
* Include ".https." in the worker IDL test.
* Move WebCryptoAPI.idl to a central directory for easier management of IDL resources.


annevk:
Reading through https://bugzilla.mozilla.org/show_bug.cgi?id=1333140 it seems several more resources might need to be renamed. I'll iterate. It's a little bit unclear why Chrome didn't address this when moving these APIs to HTTPS. Did you not get failures as a result? Maybe there's not a good enforcement of ".https."?

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22643)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/327654040)

annevk:
@andypaicu did you simply forget to rename most of the worker resources in ea527511d7f742f5ad939339df3bad2619382a00 or was that intentional? Thanks!

andypaicu:
Hi @annevk I don't think it was intentional, though I can't say I remember much about that commit anymore.

foolip (approved):

Hexcles:
@annevk , `idlharness.html` was, in a way, intentionally kept to test the non-existence of `crypto.subtle` in insecure context, and the tests were expected to fail. It was also the only coverage. See the discussion in #8284.

I know it's not ideal; the ideal solution is to assert the non-existence of these APIs, i.e. pass when they do not exist. cc @jimsch & @engelke

annevk:
@Hexcles pretty sure that's already tested by the files in https://github.com/w3c/web-platform-tests/tree/master/WebCryptoAPI/secure_context. (I'm also somewhat surprised that PR didn't catch all the other worker resources...)

annevk:
(If the IDL harness were better about [SecureContext] that would be nice though. Then we could automatically generate the relevant tests for all contexts.)

Hexcles:
Oh, thanks for the pointer. I think we missed those tests back then. This is fine then. BTW, I'm totally unfamiliar with WebCrypto, that PR I wrote was just a quick drive-by fix.

foolip:
> If the IDL harness were better about [SecureContext] that would be nice though. Then we could automatically generate the relevant tests for all contexts

Oh, looks like webidl2.js and idlharness.js doesn't know anything at all about [SecureContext]. Filed https://github.com/w3c/web-platform-tests/issues/9042.

PR #8997, opened 2018-01-11
Correct CSS2/values/color-000.xht

csnardi:
According to Color 4, decimal values are acceptable in the rgb functions. Additionally, percent channels should be rounded appropriately (49.99% -> 127). Fix these two issues in color-000.xht.

### DIFF
--- a/css/CSS2/values/color-000-ref.xht
+++ b/css/CSS2/values/color-000-ref.xht
@@ -18,7 +18,5 @@ p {
 <p>This should be green</p>
 <p>This should be green</p>
 <p>This should be green</p>
-<p>This should be green</p>
-<p>This should be green</p>
 </body>
 </html>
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22648)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/327732552)

Ms2ger (changes requested):
Can you point out where we have coverage for the number-with-decimal case now?

csnardi:
See e.g. https://github.com/w3c/web-platform-tests/blob/master/css/css-color/rgb-002.html and the other even-numbered RGB test cases.

Ms2ger (approved):

PR #9008, opened 2018-01-12
Add format-specifier-applied-to-Symbol console tests

domfarolino:
This addresses some things found during the creation of https://github.com/whatwg/console/pull/123 and the filing of https://github.com/whatwg/console/issues/113.

The intention of this PR was to add tests for everything in Console that calls https://console.spec.whatwg.org/#formatter, and test the well-specified format specifiers as they are applied on Symbols, however I wasn't sure if adding `console.trace`, `console.group`, and `console.groupCollapsed` would be overkill. Thoughts, @domenic ?


w3c-bots:
# Build BROKEN

Started: 2018-01-12 03:50:23
Finished: 2018-01-12 03:55:42


<h2>Failing Jobs</h2>
<ul>

<li>lint</li>

<li>chrome:dev</li>

<li>safari:11.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/22704)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/327968258)

domfarolino:
FWIW here are some implementation bugs:

 - Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=801442
 - Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1429971
 - Safari: https://bugs.webkit.org/show_bug.cgi?id=181576
 - Edge: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/15432105/

domenic:
For errors at least we can have a non-manual test file right?

domfarolino:
Good call, I can separate the %s ones out into a manual test that should never throw, and the %i/%d/%f ones into a JS test that expects a TypeError.

console/console-format-specifier-symbol-throws.any.js
```diff
@@ -0,0 +1,53 @@
...
+const clauses = ["must throw a TypeError when applying the", "format specifier on a Symbol"];
+
+function errorTag(subs, clause1, clause2) {
+  return subs[0] + clause1 + subs[1] + clause2;
+}
+
+test(() => {
+  assert_throws({name: "TypeError"}, () => {
+    console.log("%i", Symbol.for("description"));
+  }, errorTag`log() ${clauses[0]} %i ${clauses[1]}`);
```
domenic: Can't you just use untagged template strings for these?

domfarolino: Haha was just going to do this. Sitting in class thinking "Hm I made that more complicated than it needs to be"

domenic: If the goal is really to DRY this up then you could reduce it all down to

```js
assertFailureCase("%i", "log")
```

domenic:
FWIW I think adding the other three is not overkill, but might be annoying to write unless you end up using a loop.

domenic (approved):

console/console-format-specifier-symbol-manual.html
```diff
@@ -0,0 +1,21 @@
...
+</head>
+<body>
+<p>Open the console inside the developer tools. It should contain four lines whose contents are:</p>
+<p><code>Symbol(description)</code></p>
+<p><code>Symbol(description)</code></p>
+<p><code>Symbol(description)</code></p>
+
+<script>
+console.log("%s", Symbol.for("description"));
+console.dir("%s", Symbol.for("description"));
```
JosephPecoraro: Having `console.dir` use a format string defeats the purpose of `console.dir`.

domenic: Good point, https://console.spec.whatwg.org/#dir doesn't go through Formatter.

domenic (changes requested):
Changing to "request changes" per @JosephPecoraro noticing dir()

console/console-format-specifier-symbol-manual.html
domfarolino: Good catch

JosephPecoraro: To elaborate, developers use `console.dir(value)` to force a Complete Object Tree representation of whatever `value` was. For example in some tools (Safari) a small Object/Array will show entirely on one line and not be expandable:

screen shot 2018-01-12 at 11 02 04 am

But if a developer wants to force this to be interactable, like a large object, they can:

screen shot 2018-01-12 at 11 02 11 am

Another example is logging a Node value. If you `console.log(node)` tools normally give a DOM Tree display of the node, which can be expanded to see children.

screen shot 2018-01-12 at 11 00 38 am

But using `console.dir(node)` tools will give an Object Tree display of the value, and you can expand to see properties and the prototype chain.

screen shot 2018-01-12 at 11 00 07 am

---

In summary, I don't think this is the direction we want to go for `console.dir`. Its primary purpose is defeated by format strings.

Using a format string as the single value that `console.dir` inspects should produce an Object Tree display of the string value. Likewise, if tools decided to evaluate the format string in `console.dir` it would still just always produce a String.

domfarolino: Indeed, I agree this makes sense. That was an error in the tests as I thought (incorrectly from memory) that `console.dir` actually did call the `Formatter` in the spec. Thanks for the good explanation!

console/console-format-specifier-symbol-manual.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+<head>
+<title>Console Format Specifiers on Symbols</title>
+<meta name="author" title="Dominic Farolino" href="mailto:domfarolino@gmail.com">
+<meta name="assert" content="Console format specifiers on Symbols">
+<link rel="help" href="https://console.spec.whatwg.org/#formatter">
+</head>
+<body>
+<p>Open the console inside the developer tools. It should contain five lines whose contents are:</p>
```
domenic: Now there's a mismatch between the following lines and the "five" here. Maybe "each of which are" and just one line would be easier.

domenic (changes requested):

domenic (approved):

domfarolino:
Bumping this thread to get some thoughts. What do you think about these changes and https://bugs.webkit.org/show_bug.cgi?id=181576#c3 @JosephPecoraro ? Edge has marked the bug as Confirmed and has assigned it, which seems promising. Firefox is discussing why their implementation is doing what it does which is nice.

If we can get Chrome on board would we be ok to merge this and the corresponding [spec change](github.com/whatwg/console/pull/123) @domenic ?

In summary, we're enforcing error-throwing behavior when `%i`, `%f`, and `%d` format specifiers are applied to Symbols, but yielding the descriptor of a Symbol when applying `%s`.

domenic:
I think we should merge https://github.com/whatwg/console/pull/123 ASAP as it aligns more closely with all browsers.

https://github.com/whatwg/console/issues/113 and its impact on the tests in this PR probably should get a bit more cross-browser agreement, indeed. Mainly it would be good to hear if @JosephPecoraro / WebKit still objects, despite Edge already throwing.

We could split this PR into the uncontroversial no-exception manual tests and merge those now, as tests corresponding to https://github.com/whatwg/console/pull/123. Then we'd have a separate tests PR for the throws-an-exception tests which still need discussion. And if discussion says we should censor exceptions for those cases then we'd change these tests and introduce a new spec change.

domfarolino:
Thanks. This PR has been amended to house the non-controversial spec change described by https://github.com/whatwg/console/pull/123. Chrome, Safari, and Node currently implement the changes outlined in that PR.

The PR https://github.com/w3c/web-platform-tests/pull/9488 has been added to house the tests originally from this PR that warrant further discussion between implementations.

domenic (approved):

PR #9027, opened 2018-01-13
Correct CSS2/syntax tests

csnardi:
Some css/CSS2/syntax tests were failing, either due to an incorrect reference or due to relying on `.htaccess` to provide encodings. This change fixes the incorrect references, and creates headers files so the tests can pass.

This also removes `ident-020.xht`, as it tests for identifiers with two dashes which is now defined as the syntax for CSS variables.


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22762)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/328533555)

dbaron (commented):
So far I've looked over everything *except* for:
* the removal of ident-020
* the .htaccess/.headers changes

and it looks fine.  I or someone else still needs to do those two.

csnardi:
For clarity in reviewing: I used https://github.com/w3c/web-platform-tests/blob/7ea7d829111a24d65cb447d4770532a286f0f618/approved/css2.1/src/support/.htaccess for what I added to the .headers files.

gsnedders:
r=me for the ident-020 change.

I agree ident-020 is invalid, given they're valid identifiers per CSS Syntax Level 3 and Selectors Level 4 just references that. As such, can you move it to `css/selectors` and invert the pass condition (and give it a more descriptive name, like class-name-double-hyphen or something)?

gsnedders:
(Unless @tabatkins or someone feels like that should go into `css/syntax`?)

gsnedders (changes requested):
Per above, please move ident-020 with pass condition inverted into css/syntax.

csnardi:
I realized when I moved the file that https://github.com/w3c/web-platform-tests/blob/master/css/css-syntax/ident-three-code-points.html is actually very similar; while it doesn't test with class names it tests the first condition with id names. Given this, what do you think about moving it?

gsnedders:
No,  probably not.

If anyone strongly disagrees, they can add tests for it.

gsnedders (approved):

PR #9039, opened 2018-01-15
WebKit export of https://bugs.webkit.org/show_bug.cgi?id=180347

youennf:
fetch()'s redirect feature is insecure combined with "no-cors" or "cors"

### DIFF
--- a/fetch/api/redirect/redirect-mode.js
+++ b/fetch/api/redirect/redirect-mode.js
@@ -3,15 +3,18 @@ if (this.document === undefined) {
   importScripts("/common/get-host-info.sub.js")
 }
 
-function redirectMode(desc, redirectUrl, redirectLocation, redirectStatus, redirectMode) {
+var redirectUrl = get_host_info().HTTPS_ORIGIN + "/fetch/api/resources/redirect.py";
+var redirectLocation = "top.txt";
+
+function testRedirect(redirectStatus, redirectMode, corsMode) {
   var url = redirectUrl;
   var urlParameters = "?redirect_status=" + redirectStatus;
   urlParameters += "&location=" + encodeURIComponent(redirectLocation);
 
-  var requestInit = {"redirect": redirectMode};
+  var requestInit = {redirect: redirectMode, mode: corsMode};
 
   promise_test(function(test) {
-    if (redirectMode === "error")
+    if (redirectMode === "error" || (corsMode === "no-cors" && redirectMode !== "follow"))
       return promise_rejects(test, new TypeError(), fetch(url + urlParameters, requestInit));
     if (redirectMode === "manual")
       return fetch(url + urlParameters, requestInit).then(function(resp) {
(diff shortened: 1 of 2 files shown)


youennf:
This is related to https://bugs.webkit.org/show_bug.cgi?id=180347, https://bugzilla.mozilla.org/show_bug.cgi?id=1422710 and https://bugs.chromium.org/p/chromium/issues/detail?id=791324.

These tests require changing a PR to fetch spec along what was discussed in https://bugzilla.mozilla.org/show_bug.cgi?id=1422710.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/22800)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/329147277)

annevk (commented):
@jakearchibald could you maybe look at the service worker bits? This looks good to me.

youennf:
Would be nice if we could move forward with this PR.
@wanderview, @jakearchibald could you take a look?

fetch/api/redirect/redirect-mode.js
```diff
@@ -33,9 +33,11 @@ var redirUrl = get_host_info().HTTP_ORIGIN + "/fetch/api/resources/redirect.py";
...
 for (var statusCode of [301, 302, 303, 307, 308]) {
-  redirectMode("Redirect " + statusCode + " in \"error\" mode ", redirUrl, locationUrl, statusCode, "error");
-  redirectMode("Redirect " + statusCode + " in \"follow\" mode ", redirUrl, locationUrl, statusCode, "follow");
-  redirectMode("Redirect " + statusCode + " in \"manual\" mode ", redirUrl, locationUrl, statusCode, "manual");
+  redirectMode("Redirect " + statusCode + " in \"error\" redirect, cors mode", redirUrl, locationUrl, statusCode, "error", "cors");
+  redirectMode("Redirect " + statusCode + " in \"follow\" redirect, cors mode", redirUrl, locationUrl, statusCode, "follow", "cors");
+  redirectMode("Redirect " + statusCode + " in \"manual\" redirect, cors mode", redirUrl, locationUrl, statusCode, "manual", "cors");
 }
+redirectMode("Redirect in \"error\" redirect, no cors mode", redirUrl, locationUrl, 301, "error", "no-cors");
+redirectMode("Redirect in \"manual\" redirect, no cors mode", redirUrl, locationUrl, 301, "manual", "no-cors");
```
wanderview: Shouldn't there be a "no-cors" case with redirect mode "follow" here?  Don't we want to exercise that to verify it rejects?  Also, why not just wrap the status code for-loop above in a loop that changes the mode?  That seems the best way to make sure every combination is exercised.

service-workers/service-worker/fetch-event-redirect.https.html
```diff
@@ -354,57 +318,6 @@
...
-    request_init: {
-      redirect: 'manual',
-      mode: 'no-cors'
-    },
-    should_reject: false
-  });
-}, 'Non-navigation, manual redirect, no-cors mode Request redirected to ' +
-   'cors with credentials should succeed interception ' +
-   'and response should not be redirected');
-
```
wanderview: Why delete these test cases?  They are there to cover all possible inputs.  It seems we should just set the result to the correct expectation; which is to now reject.

service-workers/service-worker/fetch-event-redirect.https.html
```diff
@@ -966,56 +842,5 @@
...
-      redirect: 'error',
-      mode: 'no-cors'
-    },
-    // should reject because requests with 'error' RequestRedirect cannot be
-    // redirected.
-    should_reject: true
-  });
-}, 'Non-navigation, error redirect, no-cors mode Request redirected to ' +
-   'cors with credentials should fail interception and response should not ' +
-   'be redirected');
```
wanderview: Same with these cases.  I'd prefer to keep coverage of all the possible inputs even if they are all hitting the same internal condition.  In the future the internal conditions may change, but the possible set of inputs is fixed.

wanderview (changes requested):
Overall looks good, but I'd like to keep some of the coverage that is being removed here.  Also, the fetch test doesn't seem to be testing no-cors follow.

fetch/api/redirect/redirect-mode.js
```diff
@@ -3,15 +3,15 @@ if (this.document === undefined) {
...
 }
 
-function redirectMode(desc, redirectUrl, redirectLocation, redirectStatus, redirectMode) {
+function redirectMode(desc, redirectUrl, redirectLocation, redirectStatus, redirectMode, corsMode) {
   var url = redirectUrl;
   var urlParameters = "?redirect_status=" + redirectStatus;
   urlParameters += "&location=" + encodeURIComponent(redirectLocation);
 
-  var requestInit = {"redirect": redirectMode};
+  var requestInit = {"redirect": redirectMode, mode: corsMode};
```
mattto: nit: remove the quotes from "redirect" or add them to "mode" also for consistency

service-workers/service-worker/fetch-event-redirect.https.html
mattto: Yes I'm curious the reason to delete these too.

youennf:
Thanks for the feedback, I'll fix the PR accordingly.
I'll put back the deleted tests even though they seem a bit redundant.

wanderview (approved):

wanderview:
@youennf Is it intentional that this test switched from http:// to https:// for the redirect URL?  That makes all these fetches cross-origin and possibly mixed content?

wanderview:
I'm going to do some fixes on the gecko side and sync back.

wanderview:
Actually, I'll do an upstream PR since the sync'ing is a bit lagged.

youennf:
> @youennf Is it intentional that this test switched from http:// to https:// for the redirect URL? That makes all these fetches cross-origin and possibly mixed content?

Probably not intentional indeed.
I guess the test filename could be updated so that it is served through HTTP.

PR #9046, opened 2018-01-16
[presentation-api] Stop using EventWatcher for IndexedDB

tomoyukilabs:
This PR fixes the "Creating a receiving browsing context" test again.

- Use of EventWatcher to check IndexedDB on top-level and nested browsing contexts in a receiving UA are deprecated, in the similar way to https://github.com/w3c/web-platform-tests/pull/5334.

### DIFF
--- a/presentation-api/receiving-ua/support/PresentationReceiver_create_receiving-ua.html
+++ b/presentation-api/receiving-ua/support/PresentationReceiver_create_receiving-ua.html
@@ -37,6 +37,11 @@ child.addEventListener('load', () => {
     const notice = document.getElementById('notice');
     const modal = document.getElementById('modal');
 
+    const dbName = {
+        controller: 'db-presentation-api-controlling-ua',
+        receiver: 'db-presentation-api-receiving-ua'
+    };
+
     promise_test(t => {
         t.add_cleanup(() => {
             document.cookie = cookieName + '=False;Expires=' + new Date().toUTCString();
@@ -46,9 +51,10 @@ child.addEventListener('load', () => {
             sessionStorage.removeItem(storageNameChild);
             localStorage.removeItem(storageNameChild);
 
-            if (db)
-                db.close();
-            indexedDB.deleteDatabase(dbName);
+            Object.values(dbName).forEach(name => {
+                indexedDB.deleteDatabase(name);
+            });
+
             if ('serviceWorker' in navigator) {
                 navigator.serviceWorker.getRegistrations().then(registrations => {
                     return Promise.all(registrations.map(reg => reg.unregister()));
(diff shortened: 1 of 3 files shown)


w3c-bots:
# Build PASSED

Started: 2018-01-16 07:30:19
Finished: 2018-01-16 07:37:00


<h2>Failing Jobs</h2>
<ul>

<li>chrome:dev</li>

<li>safari:11.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/22839)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/329352086)

presentation-api/receiving-ua/support/PresentationReceiver_create_receiving-ua_child.html
```diff
@@ -61,31 +61,43 @@
 
         // Indexed Database
         let db;
+        const storeName = 'store-controlling-ua';
         const checkIndexedDB = () => {
             if ('indexedDB' in window) {
                 message = 'Indexed Database is shared by top-level and nested browsing contexts.';
```
mfoltzgoogle: Isn't this required by same-origin policy?  Do we need to test this here?

presentation-api/receiving-ua/support/PresentationReceiver_create_receiving-ua.html
```diff
@@ -106,27 +106,30 @@
...
-                    assert_equals(req.result, null, 'Indexed Database is set to an empty storage.');
-                    req = store.put(storeData.parent);
-                    eventWatcher = new EventWatcher(t, req, 'success');
-                    return eventWatcher.wait_for('success');
-                }).then(() => {
-                    db.close();
-                    db = null;
+                let req = indexedDB.open(dbName, 1), store, transaction;
+                return new Promise((resolve, reject) => {
+                    // The test would fail when the database is already created by the controlling UA
```
mfoltzgoogle: According to the IndexedDB spec, you can test for existence of a database by a non-empty version property on the result object.

https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabase

Perhaps the test would be simpler by just checking whether a database already exists (thus indicating whether the database was leaked across receiving browsing contexts).

presentation-api/receiving-ua/support/PresentationReceiver_create_receiving-ua_child.html
tomoyukilabs: The intention of this is to check if both top-level and nested browsing contexts share the same storage, according to the following spec described in [Creating a receiving browsing context](http://w3c.github.io/presentation-api/#creating-a-receiving-browsing-context):

> All nested browsing contexts created by the presented document, i.e. that have the receiving browsing context as their top-level browsing context, [...] MUST also share the same browsing state (storage) for features 5-10 listed above.

presentation-api/receiving-ua/support/PresentationReceiver_create_receiving-ua.html
tomoyukilabs: That's a good idea. Regarding IndexedDB, we can simply check version of a database. I'll simplify and submit these codes later.

BTW, I'm afraid that the MDN article might be incorrect. With regard to `indexedDB.open`, the [IndexedDB spec](https://www.w3.org/TR/IndexedDB/#dom-idbfactory-open) says:

>If version is not given and no database with that name exists, a new database will be created with version equal to 1.

Anyway, we can still check whether the database has been previously created or not by looking at version of the database.

tomoyukilabs:
I've update the codes by simplifying how to check IndexedDB. Now EventWatcher is used here again.

Many thanks, @mfoltzgoogle!

markafoltz (approved):
LGTM, thanks for the changes.  However, I don't have write access to this repository.

tomoyukilabs:
@mfoltzgoogle Thanks for your review and approval.

@tidoust @louaybassbouss Could you review this PR?

tidoust (approved):
Thanks for the simplification!

tomoyukilabs:
@tidoust Thanks!

PR #9135, opened 2018-01-23
Switch most preload tests to PerformanceObserver instead of a timer

yoavweiss:
Currently many preload tests are slow due to their use of a timer. PerformanceObserver enables us to avoid that in most cases, by giving us a notification when resources were downloaded, independent from their load events.
This PR switches most of the preload tests to use that method. (Failed to switch all, as we don't have a notification when resources *failed* to load, nor guarantees regarding PerfObserver firing with regard to the resource's load event)

### DIFF
--- a/preload/download-resources.html
+++ b/preload/download-resources.html
@@ -47,14 +47,14 @@ function useAllResources() {
 promise_test(async t => {
     await new Promise(async (resolve, reject) => {
         verifyPreloadAndRTSupport()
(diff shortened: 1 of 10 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/23133)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/332285252)

PR #9160, opened 2018-01-23
Fix attributeChangedCallback isn't fired with correct newValue when the attribute value is an empty string;

jgraham:
MozReview-Commit-ID: L3RvNPNDfUC

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1430034 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9160)


jgraham:
Code reviewed upstream.

PR #9161, opened 2018-01-23
Add validity state check for WebDriver:ElementClear.

jgraham:
This patch checks that the element satisfies its form control
constraints, as well as being empty, before deciding not to clear
the element.  This will make it possible to clear elements that
have invalid input.

The "clear a resettable element" algorithm is missing a check of
the <input> element's ValidityState.  WebDriver:ElementClear has
a subtle bug that only manifests in Gecko because Blink rejects
invalid key input to validation fields such as <input type=number>,
but Gecko does not.

The value property of <input type=number> will not be updated unless
the input is actually valid, which means the first step of the algorithm
will pass irregardless of whether the user has actually modified it.

MozReview-Commit-ID: C2M3Fl1iKx6

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1430575 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9161)

### DIFF
--- a/webdriver/tests/interaction/element_clear.py
+++ b/webdriver/tests/interaction/element_clear.py
@@ -312,6 +312,48 @@ def test_resettable_element_focus_when_empty(session):
     assert session.execute_script("return window.events") == []
 
 
+@pytest.mark.parametrize("type,invalid_value",
+                         [("number", "foo"),
+                          ("range", "foo"),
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #9162, opened 2018-01-23
Fix slotted invalidation.

jgraham:
This is a partial revert of
https://github.com/servo/servo/commit/ce1d8cd232dfbc9e0a52f9467ba2bc209087ea63

If you're in a shadow tree, you may not be slotted but you still need to look at
the slotted rules, since a <slot> could be a descendant of yours.

Just use the same invalidation map everywhere, and remove complexity.

This means that we can do some extra work while trying to gather invalidation
if there are slotted rules, but I don't think it's a problem.

The test is ported from https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/fast/css/invalidation/slotted.html?l=1&rcl=58d68fdf783d7edde1c82a642e037464861f2787

Curiously, Blink fails the test as written, presumably because they don't flush
styles from getComputedStyle correctly (in their test they do via
updateStyleAndReturnAffectedElementCount), due to <slot>s not being in the flat
tree in their implementation.

MozReview-Commit-ID: 6b7BQ6bGMgd

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1429846 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9162)

### DIFF
--- a/css/css-scoping/slotted-invalidation.html
+++ b/css/css-scoping/slotted-invalidation.html
@@ -1,0 +1,35 @@
+<!doctype html>
+<title>CSS Test: Style invalidation for ::slotted()</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<link rel="author" title="Rune Lillesveen" href="mailto:futhark@chromium.org">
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #9165, opened 2018-01-23
Introduce assert_time_equals_literal and use it.

jgraham:
This assertion is supposed to be used where the first argument has a tolerance
but the second argument doesn't have such tolerance.  Whereas
assert_times_equal() is supposed to be used for the case both arguments have
the same tolerance, actually it hasn't, it will be fixed in a subsequent patch
in this patch series.

MozReview-Commit-ID: FEDHilbX2rm

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1430654 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9165)

### DIFF
--- a/web-animations/README.md
+++ b/web-animations/README.md
@@ -98,10 +98,19 @@ Guidelines for writing tests
     Remember, even if we do need to make all tests take, say 200s each, text
     editors are very good at search and replace.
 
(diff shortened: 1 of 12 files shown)


jgraham:
Code reviewed upstream.

Issue #6998, opened 2017-08-24
./wpt run chrome crashes Python on macOS


gsnedders:
macOS 10.12.6, Python 2.7.13 from Homebrew:

```
pnin:web-platform-tests gsnedders$ ./wpt run chrome dom/events/Event-type.html --no-manifest-update
Using webdriver binary /usr/local/bin/chromedriver
 0:04.01 LOG: MainThread INFO Using 1 client processes
 0:04.05 SUITE_START: MainThread 1
 0:04.05 LOG: MainThread INFO Running reftest tests
 0:04.06 LOG: MainThread INFO Starting http server on web-platform.test:8000
 0:04.06 LOG: MainThread INFO Starting http server on web-platform.test:8001
 0:04.07 LOG: MainThread INFO Starting http server on web-platform.test:8443
 0:04.08 LOG: MainThread INFO No reftest tests to run
 0:04.08 LOG: MainThread INFO Running wdspec tests
 0:04.08 LOG: MainThread INFO No wdspec tests to run
 0:04.08 LOG: MainThread INFO Running testharness tests
 0:04.11 PROCESS_OUTPUT: ProcessReader (pid:28698) Full command: /usr/local/bin/chromedriver --port=4448 --url-base=/
(pid:28698) "Starting ChromeDriver 2.31.488774 (7e15618d1bf16df8bf0ecf2914ed1964a387ba0b) on port 4448"
 0:04.11 PROCESS_OUTPUT: ProcessReader (pid:28698) "Only local connections are allowed."
 0:04.59 LOG: Thread-TestrunnerManager-1 INFO Starting runner
```

At around 4s, about the same time as most of these messages, I get a macOS "Python quit unexpectedly" message, with the following stack:

```
Process:               Python [32981]
Path:                  /usr/local/Cellar/python/2.7.13_1/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python
Identifier:            Python
Version:               2.7.13 (2.7.13)
Code Type:             X86-64 (Native)
Parent Process:        Python [32957]
Responsible:           Python [32981]
User ID:               501

Date/Time:             2017-08-24 17:04:59.487 +0100
OS Version:            Mac OS X 10.12.6 (16G29)
Report Version:        12
Anonymous UUID:        50D68F61-7230-1B71-16A9-0EDDFA61EC8F

Sleep/Wake UUID:       D27B603A-2403-47B5-B2B0-8F47FE630868

[...]

PR #9188, opened 2018-01-24
Simplify Chrome Android setup

Hexcles:
Automatically start adb, connect to device, and set up port forwarding.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9188)

### DIFF
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -281,6 +281,7 @@ class Chrome(Browser):
                 logger.critical("dbus not running and can't be started")
                 sys.exit(1)
 
+
 class ChromeAndroid(Browser):
     """Chrome-specific interface for android.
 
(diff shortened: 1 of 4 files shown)


w3c-bots:
# Build PASSED

Started: 2018-01-25 01:31:08
Finished: 2018-01-25 01:48:43





- [WPT PR Status](http://pulls.web-platform-tests.org/build/23260)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/333047600)

tobie:
Summoning @wpt-pr-bot.

Hexcles:
@jgraham the PR is ready for review

Most of the change is incremental improvements in our draft Chrome Android setup, except this line, which fixes the setup/cleanup imbalance I told you about:
https://github.com/w3c/web-platform-tests/pull/9188/commits/f71d30a6f878257e05432aceac7c0456857d10d8#diff-35e3a18456dc70eea5ca7fff107a1a46L232

tools/wpt/browser.py
```diff
@@ -35,6 +35,10 @@ def install(self, dest=None):
     def install_webdriver(self):
         return NotImplemented
 
+    @abstractmethod
+    def find_webdriver(self):
```
foolip: With no other changes to `find_webdriver` in this PR, an explanation of why this change is needed in the commit message would be good. Does this mean that a subclass not implementing this will blow up? (You can write a block in a comment describing what the commit message should be when merging.)

tools/wptrunner/wptrunner/browsers/chrome_android.py
```diff
@@ -76,6 +79,19 @@ def __init__(self, logger, binary, webdriver_binary="chromedriver",
...
 
+    def _adb_run(self, args):
+        self.logger.info('adb ' + ' '.join(args))
+        subprocess.check_call(['adb'] + args)
+
+    def setup(self):
+        self._adb_run(['wait-for-device'])
+        self._adb_run(['forward', '--remove-all'])
+        self._adb_run(['reverse', '--remove-all'])
+        # TODO: Use ports from config.
```
foolip: Some logging confirms that there are in `server_config` in `executor_kwargs`, and I'm guessing the `executor_kwargs` return value ends up somehow propagated to the `ChromeAndroidBrowser` instance? In other words, is this easy to fix already? If so let's fix it, if it's hard because of some mis-layering, maybe file an issue about that layering problem.

foolip (commented):
If @jgraham wants to take a look that would be most kind, but I'm happy to do my best and have tried.

I also tried running the `wpt run chrome_android speech-api` on my Mac and did see Chrome on my phone doing some things, but after a few tries I see a lot of this:
```
objc[66566]: +[__NSPlaceholderDate initialize] may have been in progress in another thread when fork() was called.
objc[66566]: +[__NSPlaceholderDate initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
```

And then nothing seems to really work. But I haven't rooted this phone to update the hosts file, and it seems to work about the same without the changes. If confirmed to actually work somewhere, that's good! :)

tools/wpt/browser.py
Hexcles: Actually, let me hack the history of this branch. I'll make a feature change commit + a code health commit. And then we can rebase-merge this PR.

tools/wptrunner/wptrunner/browsers/chrome_android.py
Hexcles: No, it doesn't. Although there is a hook for Browser modules to modify `executor_kwargs`, but the dictionary is not passed back to Browser classes, as Browsers are not Executors... I'm going to store the `server_config` in a module-local variable. It's not ideal, but probably better than hardcoding.

Hexcles:
I ended up splitting the changes into two PRs. See #9349 for the follow-up code health fixes.

foolip (approved):

tools/wptrunner/wptrunner/testrunner.py
```diff
@@ -229,7 +229,6 @@ def stop(self, force=False):
     def cleanup(self):
         if self.init_timer is not None:
             self.init_timer.cancel()
-        self.browser.cleanup()
```
foolip: Wait was this removal intentional?

Hexcles: Yes, indeed, and it's quite important. See the commit message:

> Also fix testrunner so that setup & cleanup balance out (called at enter
> & exit of a Browser context, respectively). Previously, cleaup() didn't
> do anything in all browsers, so the issue wasn't discovered. In this
> commit, cleanup() is used for uninstalling port forwarding.

which is all about the removal of this one line.

Hexcles:
Also, @foolip regarding the errors you saw, I'm quite certain they are not related to my change. And I'm quite confused: where exactly do we ever call a compiler (objc)?

gsnedders:
@foolip's errors appear to just be #6998 manifesting in a slightly less crashy way.

PR #9226, opened 2018-01-29
Run master commits on Taskcluster

jgraham:
This PR actually has two parts:

1. A docker image for running tests. I assume this will be relatively uncontroversial, although there are certainly ways that we could improve the details.
2. The required configuration to run each push on master on TaskCluster. This won't do anything unless the relevant GitHub integration is enabled. If it is, however, it will typically provide the full results of the testsuite within about an hour of a merge to master. You can see an [example](https://tools.taskcluster.net/groups/bLpNdT49RGCR7nEhELQE1A) run from my fork. Each job uploads an artifact containing the results from that chunk in `wptreport.json` format; these could be combined to give the full set of results for each push.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9226)

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -1,0 +1,687 @@
+allowPullRequests: collaborators
+tasks:
+- extra:
+    github:
+      branches: [master]
+      events: [push]
+  metadata: {description: '', name: wpt-chrome-dev-reftest-1, owner: '{{ event.head.user.email
+      }}', source: '{{ event.head.repo.url }}'}
+  payload:
+    artifacts:
+      public/results: {path: /home/test/artifacts, type: directory}
+    command: [/bin/bash, --login, -c, ">-\n            ~/start.sh &&\n           \
+        \ cd /home/test/web-platform-tests &&\n            git fetch {{event.head.repo.url}}\
+        \ &&\n            git config advice.detachedHead false &&\n            git\
+        \ checkout {{event.head.sha}} &&\n            ./tools/ci/ci_taskcluster.sh\
+        \ chrome reftest 1 6"]
+    image: harjgam/web-platform-tests:0.5
+    maxRunTime: 5400
+  provisionerId: '{{ taskcluster.docker.provisionerId }}'
+  workerType: '{{ taskcluster.docker.workerType }}'
+- extra:
+    github:
+      branches: [master]
+      events: [push]
+  metadata: {description: '', name: wpt-chrome-dev-reftest-2, owner: '{{ event.head.user.email
+      }}', source: '{{ event.head.repo.url }}'}
+  payload:
+    artifacts:
+      public/results: {path: /home/test/artifacts, type: directory}
+    command: [/bin/bash, --login, -c, ">-\n            ~/start.sh &&\n           \
+        \ cd /home/test/web-platform-tests &&\n            git fetch {{event.head.repo.url}}\
+        \ &&\n            git config advice.detachedHead false &&\n            git\
+        \ checkout {{event.head.sha}} &&\n            ./tools/ci/ci_taskcluster.sh\
+        \ chrome reftest 2 6"]
+    image: harjgam/web-platform-tests:0.5
+    maxRunTime: 5400
+  provisionerId: '{{ taskcluster.docker.provisionerId }}'
+  workerType: '{{ taskcluster.docker.workerType }}'
+- extra:
+    github:
+      branches: [master]
+      events: [push]
+  metadata: {description: '', name: wpt-chrome-dev-reftest-3, owner: '{{ event.head.user.email
+      }}', source: '{{ event.head.repo.url }}'}
+  payload:
+    artifacts:
+      public/results: {path: /home/test/artifacts, type: directory}
+    command: [/bin/bash, --login, -c, ">-\n            ~/start.sh &&\n           \
+        \ cd /home/test/web-platform-tests &&\n            git fetch {{event.head.repo.url}}\
+        \ &&\n            git config advice.detachedHead false &&\n            git\
+        \ checkout {{event.head.sha}} &&\n            ./tools/ci/ci_taskcluster.sh\
+        \ chrome reftest 3 6"]
+    image: harjgam/web-platform-tests:0.5
+    maxRunTime: 5400
+  provisionerId: '{{ taskcluster.docker.provisionerId }}'
+  workerType: '{{ taskcluster.docker.workerType }}'
+- extra:
+    github:
+      branches: [master]
+      events: [push]
+  metadata: {description: '', name: wpt-chrome-dev-reftest-4, owner: '{{ event.head.user.email
+      }}', source: '{{ event.head.repo.url }}'}
+  payload:
+    artifacts:
+      public/results: {path: /home/test/artifacts, type: directory}
+    command: [/bin/bash, --login, -c, ">-\n            ~/start.sh &&\n           \
+        \ cd /home/test/web-platform-tests &&\n            git fetch {{event.head.repo.url}}\
+        \ &&\n            git config advice.detachedHead false &&\n            git\
+        \ checkout {{event.head.sha}} &&\n            ./tools/ci/ci_taskcluster.sh\
+        \ chrome reftest 4 6"]
+    image: harjgam/web-platform-tests:0.5
+    maxRunTime: 5400
+  provisionerId: '{{ taskcluster.docker.provisionerId }}'
+  workerType: '{{ taskcluster.docker.workerType }}'
+- extra:
+    github:
+      branches: [master]
+      events: [push]
+  metadata: {description: '', name: wpt-chrome-dev-reftest-5, owner: '{{ event.head.user.email
+      }}', source: '{{ event.head.repo.url }}'}
+  payload:
+    artifacts:
+      public/results: {path: /home/test/artifacts, type: directory}
+    command: [/bin/bash, --login, -c, ">-\n            ~/start.sh &&\n           \
+        \ cd /home/test/web-platform-tests &&\n            git fetch {{event.head.repo.url}}\
+        \ &&\n            git config advice.detachedHead false &&\n            git\
+        \ checkout {{event.head.sha}} &&\n            ./tools/ci/ci_taskcluster.sh\
+        \ chrome reftest 5 6"]
+    image: harjgam/web-platform-tests:0.5
+    maxRunTime: 5400
+  provisionerId: '{{ taskcluster.docker.provisionerId }}'
+  workerType: '{{ taskcluster.docker.workerType }}'
+- extra:
+    github:
+      branches: [master]
+      events: [push]
+  metadata: {description: '', name: wpt-chrome-dev-reftest-6, owner: '{{ event.head.user.email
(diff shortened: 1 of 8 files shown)


w3c-bots:
# Build BROKEN

Started: 2018-01-29 12:03:04
Finished: 2018-01-29 12:19:25


<h2>Failing Jobs</h2>
<ul>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/23407)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/334670050)

foolip:
> A docker image for running tests

Interesting! @jugglinmike, very relevant to https://github.com/web-platform-tests/results-collection/issues/535#issuecomment-380643242

>  will typically provide the full results of the testsuite within about an hour of a merge to master

Wow. That'd achieve our [wpt.fyi results are available within 1 hour](https://docs.google.com/document/d/1hCOWgU95pN0LaZLdkoJyjlS9A5sjyQQWW6gh8YaDF8o/edit#heading=h.3fk9m1cm62pz) for at least Firefox, and way sooner than I thought was possible.

@lukebjerring @boaz @mariestaver @jugglinmike, this could change our roadmap drastically, so let's discuss at the earliest opportunity.

foolip:
> Each job uploads an artifact containing the results from that chunk in wptreport.json format; these could be combined to give the full set of results for each push.

@Hexcles, can you take a look at the bits around this, to see if it'll be a trivial integration with a wpt.fyi results POST endpoint? Seems like the token management and such would have to be very similar to what we've speculated about for PRs on Travis.

.taskcluster.yml
```diff
@@ -0,0 +1,687 @@
...
+      }}', source: '{{ event.head.repo.url }}'}
+  payload:
+    artifacts:
+      public/results: {path: /home/test/artifacts, type: directory}
+    command: [/bin/bash, --login, -c, ">-\n            ~/start.sh &&\n           \
+        \ cd /home/test/web-platform-tests &&\n            git fetch {{event.head.repo.url}}\
+        \ &&\n            git config advice.detachedHead false &&\n            git\
+        \ checkout {{event.head.sha}} &&\n            ./tools/ci/ci_taskcluster.sh\
+        \ chrome reftest 1 6"]
+    image: harjgam/web-platform-tests:0.5
```
foolip: Should we have a "web-platform-tests" organization on Docker Hub? Is a thing?

.taskcluster.yml
```diff
@@ -0,0 +1,687 @@
+allowPullRequests: collaborators
```
foolip: This configuration file is huge and fairly repetitive. I assume it has to be checked in to the repo like .travis.yml, but could parts of it be generated?

tools/ci/ci_taskcluster.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+if [ $1 == "firefox" ]; then
+    ./wpt run firefox --log-tbpl=- --log-tbpl-level=debug --log-wptreport=../artifacts/wpt_report.json --this-chunk=$3 --total-chunks=$4 --test-type=$2 -y --install-browser --no-pause --no-restart-on-unexpected
+elif [ $1 == "chrome" ]; then
+    ./wpt run chrome --log-tbpl=- --log-tbpl-level=debug --log-wptreport=../artifacts/wpt_report.json --this-chunk=$3 --total-chunks=$4 --test-type=$2 -y  --no-pause --no-restart-on-unexpected
```
foolip: Which version of chrome will this be?

.taskcluster.yml
foolip: Ha, there is a script :) Suggestion is then to add a comment at the top pointing to the script.

.taskcluster.yml
jgraham: It is and we can; I just created an org with the namespace `webplatformtests`; neither `wpt` or `web-platform-tests` worked, but I can't tell if they were taken or if there are some character/length restrictions.

tools/ci/ci_taskcluster.sh
foolip: Ah, well as long as it's not just accidentally chrome dev, that's just fine!

tools/docker/start.sh
```diff
@@ -0,0 +1,27 @@
...
+127.0.0.1	www1.web-platform.test
+127.0.0.1	www2.web-platform.test
+127.0.0.1	xn--n8j6ds53lwwkrqhv28a.web-platform.test
+127.0.0.1	xn--lve-6lad.web-platform.test
+0.0.0.0	nonexistent-origin.web-platform.test" >> /etc/hosts'
+
+cd web-platform-tests
+git pull --depth=1
+
+# Install Chome unstable
```
foolip: In https://github.com/w3c/web-platform-tests/pull/8838 I tried to banish "unstable" everywhere except package name, can you call it "Chrome dev"?

tools/docker/start.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+sudo sh -c 'echo "
```
foolip: @gsnedders wrote something to generate this output, can you use that here?

.taskcluster.yml
foolip: Where does the code live for creating the docker images?

tools/docker/start.sh
gsnedders: `./wpt make-hosts-file` does this

tools/docker/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+    wget \
+    xvfb
+
+# Installing just the deps of firefox and chrome is moderately tricky, so
+# just install the default versions of them, and some extra deps we happen
+# to know that chrome requires
+
+RUN apt-get -qqy install \
+    firefox \
+    chromium-browser \
```
Hexcles: IIUC, the extra dependencies below are dependencies of Chrome (that you happen to know) but not of chromium-browser.

IMHO, since Chrome is already distributed in a deb package, we can perhaps skip this (i.e. only install firefox). See my next comment in `start.sh` for dependency problems.

tools/docker/start.sh
```diff
@@ -0,0 +1,27 @@
...
+deb_archive=google-chrome-unstable_current_amd64.deb
+wget https://dl.google.com/linux/direct/$deb_archive
+
+# Installation will fail in cases where the package has unmet dependencies.
+# When this occurs, attempt to use the system package manager to fetch the
+# required packages and retry.
+if ! sudo dpkg --install $deb_archive; then
+    sudo apt-get -y install --fix-broken
+    sudo dpkg --install $deb_archive
+fi
```
Hexcles: I think this is sufficient for solving Chrome's dependencies, so there's no need for installing chromium-browser in the Docker image.

Besides, you might want to try `gdebi`, which installs a local deb package along with its dependencies (from apt). I think it's available by default on Ubuntu.

gsnedders: @Hexcles This does the same, FWIW. (Also, BTW, you're not on IRC.)

foolip:
I'll step away from the detailed code review now as @Hexcles will do that, makes the most sense since it needs to be integrated together with the results upload endpoint, somehow.

foolip:
I have a question about browser version management. If each shard individually downloads https://dl.google.com/linux/direct/google-chrome-unstable_current_amd64.deb then once every few days it's possible they will not get the same version. At minimum it would be good if the browser version is recorded in the wptreport JSON so that we can detect this. Even better would be if the browser version is somehow managed together with the container.

The very simplest would be to have the browsers installed in the container itself, which would mean updating the container daily, but that can be automated pretty easily. https://github.com/whatwg/html-build/blob/master/ci-deploy/outside-container.sh does this.

How we then ensure that the same Docker image is used for each Taskcluster shard I don't know. But it'd be very good if it's also recorded in the wptreport somehow. (A bit backwards, I know, but a variable could be passed in.)

tools/docker/start.sh
Hexcles: Yeah I know it's the same, but it'd save a couple lines :P

Hexcles:
Sorry I was busy (travelling) last week and not available on IRC.

I've glanced through the PR. Apart from the nit comment I made earlier, I have one high-level question: is there no way to configure the number of chunks other than using a script to dump each chunk as a task? I'm a bit surprised if Taskcluster doesn't some sharding this natively.

jgraham:
So. Tthe good way to support that is to write a decision task that schedules all the other tasks dynamically. We should do that, but it's not totally necessary for the first iteration.

It's possible that there's some syntax for parameterising the yaml file; I can look for documentation. I'm not very pleased with the current setup either, but it seemed like the fastest path to something that worked.

jgraham:
https://bugzilla.mozilla.org/show_bug.cgi?id=1335915 seems to be the bug for enabling the fancy templating stuff in `.taskcluster.yml`; since that's still open I assume it's not possible yet, but we should look to switch to it once that bug is fixed.

Hexcles (approved):
LGTM except for a nit.

(I agree with using a script to generate .taskcluster.yml for now. We can explore simplifications later.)

tools/docker/start.sh
```diff
@@ -0,0 +1,14 @@
...
+cd web-platform-tests
+git pull --depth=1
+
+sudo sh -c './wpt make-hosts-file >> /etc/hosts'
+
+# Install Chome dev
+deb_archive=google-chrome-unstable_current_amd64.deb
+wget https://dl.google.com/linux/direct/$deb_archive
+
+sudo gdebi -n $deb_archive
```
jugglinmike: > Note: Installing Google Chrome will add the Google repository so your system
> will automatically keep Google Chrome up to date. If you don’t want Google's
> repository, do `sudo touch /etc/default/google-chrome` before installing the
> package.

Source: https://www.google.com/intl/en/chrome/browser/

Is this relevant given the expected lifespan of the container?

tools/docker/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+    unzip \
+    wget \
+    xvfb
+
+# Installing just the deps of firefox and chrome is moderately tricky, so
+# just install the default versions of them, and some extra deps we happen
+# to know that chrome requires
+
+RUN apt-get -qqy install \
+    firefox \
```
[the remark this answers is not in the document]
jgraham: Installing more stuff at build time does have the advantage that we don't do so much work at runtime, so generally it makes more sense to move stuff into the docker build rather than the other way around.

In this case we agreed on irc that it doesn't make much difference either way and there are plenty of performance optimisations and potential cleanup left on the table, so it's OK to just merging what we have now.

tools/docker/start.sh
jgraham: No; any changes only persist for the lifetime of the container, which is just for a single test run.

foolip:
Awesome! @jgraham, where can one see the results of this? Looks like the very first run on master is this one, still ongoing:
https://tools.taskcluster.net/groups/Bv2TZUgeSyGhXblM8UU0aA

foolip:
And here are actual results from shards from a shard that has finished:
https://taskcluster-artifacts.net/N59MId9GQoir2KBBxEH_FA/0/public/results/wpt_report.json

This will be great :)

Issue #6289, opened 2017-06-20
/browsing-the-web/scroll-to-fragid/003.html test wrong


pyfisch:
The test at `/html/browsers/browsing-the-web/scroll-to-fragid/003.html`.

This is a standards mode document (<!doctype html>). It is scrolled to some fragments. Then document.body.scrollTop is requested. It should be nonzero if the document is scrolled. The spec states that this is only true in quirks mode, in standards mode document.documentElement is scrolled. (The test passes if one replaces document.body with document.documentElement).

So the test should either be run in quirks mode or it should be changed to use documentElement.

See also servo/servo#17385

fred-wang:
@pyfisch I submitted #9228 , but I used document.scrollingElement so that the test works in WebKit too.

fred-wang:
Fixed by https://github.com/w3c/web-platform-tests/pull/9228

PR #9228, opened 2018-01-29
Use document.scrollingElement for browsing-the-web/scroll-to-fragid/003.html

fred-wang:
browsing-the-web/scroll-to-fragid/003.html uses document.body in standard mode
to handle scrolling of the viewport while the CSSOMView specification says it
should use document.documentElement. This makes the test fails in all web
engines but WebKit due to [1]. This commit replace such instances by
document.scrollingElement so that the test works in any mode or web engine.
This closes issue #6289.

[1] https://bugs.webkit.org/show_bug.cgi?id=5991

### DIFF
--- a/html/browsers/browsing-the-web/scroll-to-fragid/003.html
+++ b/html/browsers/browsing-the-web/scroll-to-fragid/003.html
@@ -10,14 +10,14 @@
 <div id="test1">scroll 2</div>
 <script>
 test(function() {
-  assert_equals(document.body.scrollTop, 0);
+  assert_equals(document.scrollingElement.scrollTop, 0);
   location.hash = "test";
 
-  var scroll1 = document.body.scrollTop;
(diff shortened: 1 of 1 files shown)


gsnedders (approved):

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/23414)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/334711599)

gsnedders (commented):
Actually, do we have a test anywhere that tests behaviour with `document.body` after this change?

gsnedders (changes requested):
I don't think we do, so we need a test for the `document.body` case, please! We should have some test that fails in Webkit given the difference in behaviour here.

gsnedders (approved):
Oh, actually we have tests for that in css/cssom-view/scrollingElement.html

PR #9231, opened 2018-01-29
Test document.body and document.bgColor when root is not <html>

zcorpan:
See https://github.com/whatwg/html/issues/3403

### DIFF
--- a/html/dom/documents/dom-tree-accessors/Document.body.html
+++ b/html/dom/documents/dom-tree-accessors/Document.body.html
@@ -216,4 +216,12 @@ test(function() {
   assert_equals(f2.nextSibling, f1, "New frameset should have replaced the body");
 }, "Setting document.body to a frameset will replace the first existing body/frameset.");
 
+test(function() {
+  var doc = createDocument();
+  doc.appendChild(doc.createElement("test"));
+  var new_body = doc.createElement("body");
+  doc.body = new_body;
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/23418)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/334734459)

annevk (approved):
It might also be worth testing with a document inside an `<iframe>`. Sometimes a browsing context ends up mattering somehow (and it probably shouldn't here).

bzbarsky (approved):
This is generally good, but we should also have a test for the "frameset in html" case, where .fgColor is _not_ propagated to the frameset, even though document.body is the frameset.

zcorpan:
I've address comments and rebased to resolve conflict with https://github.com/w3c/web-platform-tests/commit/08cb16810a35273e5f400acd1fbdc352559addfe (which wasn't much of a conflict, just us adding new tests in the same place).

annevk (approved):

zcorpan:
https://bugs.webkit.org/show_bug.cgi?id=182484
https://bugs.chromium.org/p/chromium/issues/detail?id=809059
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/15806109/
Firefox Nightly passes these tests.

PR #9240, opened 2018-01-29
Don't strictly enforce a particular FetchEvent.clientId for navigatio…

wanderview:
…n requests.  (w3c/ServiceWorker#1266)

Note, `FetchEvent.clientId` is also covered by fetch-event.https.html, so we don't need this coverage here.  Since this part of the spec is changing and a bit confused at the moment here it would be nice to let these tests verify their other parts without this check for now.  Also, the tests are not well written and currently timeout if this check fails.

### DIFF
--- a/service-workers/service-worker/resources/clients-get-worker.js
+++ b/service-workers/service-worker/resources/clients-get-worker.js
@@ -4,12 +4,10 @@
 // the `event` object. In the case of the `onmessage` handler, it provides the
 // Client instance attributes of the requested clients.
 self.onfetch = function(e) {
-  if (e.request.mode === 'navigate' && e.clientId !== "") {
-    e.respondWith(Response.error(
-      '`clientId` incorrectly set to non-empty string for request with mode `navigate`'
-    ));
-    return;
-  }
+  // NOTE: Until FetchEvent.resultingClientId is implemented some browsers
+  // return null for the FetchEvent.clientId on navigation requests.  Since
+  // this is not in the spec yet, though, we don't strictly enforce it in the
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/23452)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/334925678)

service-workers/service-worker/resources/clients-get-worker.js
```diff
@@ -4,12 +4,10 @@
...
-  if (e.request.mode === 'navigate' && e.clientId !== "") {
-    e.respondWith(Response.error(
-      '`clientId` incorrectly set to non-empty string for request with mode `navigate`'
-    ));
-    return;
-  }
+  // NOTE: Until FetchEvent.resultingClientId is implemented some browsers
+  // return null for the FetchEvent.clientId on navigation requests.  Since
+  // this is not in the spec yet, though, we don't strictly enforce it in the
+  // test any more.
```
mattto: Yes I think we should use fetch-event.https.html to test this instead (although the current test there is a bit awkward).

See also client-navigate-worker.js which seems to explicitly test the opposite.

I don't know if we really need this comment in this file though... seems better suited for the commit description.

mfalken (approved):
lgtm, I'd also mention in the description we already have test coverage for clientId for navigations in fetch-event.https.html.

jdm (approved):
Based on mattto's approval.

PR #9258, opened 2018-01-30
Switch external_host to host_ip.

jgraham:
When running tests in Firefox we don't require that the hostname is
actually resolvable since we set up some internal DNS overrides. This
means that we don't actually need to have web-platform.test defined in
/etc/hosts. As a consequence when we check the server is responding,
the check will typically fail for Firefox developers if we try to
access by hostname. Previously we made use of a special dance in which
external_host held the ip at the point of the check and was later
replaced by host to actually run tests. But that changed, and instead
we relied on the fact that we (accidentially) didn't actually fail in
the case the servers couldn't be reached by host name. However in some
cases things still failed because local DNS resolved web-patform.test
to some ip address, causing hangs or other badness.

In this patch we simply add a host_ip property to the config, and
connect to that to test the servers are repopnsing rather than the
hostname. We also actually fil if we didn't manage to connect to all
the servers.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9258)

### DIFF
--- a/config.default.json
+++ b/config.default.json
@@ -1,7 +1,7 @@
 {"host": "web-platform.test",
  "doc_root": null,
  "ws_doc_root": null,
- "external_host": null,
+ "host_ip": null,
  "ports":{"http":[8000, "auto"],
           "https":[8443],
           "ws":["auto"],
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -644,9 +644,10 @@ def get_ports(config, ssl_environment):
 
 
 def normalise_config(config, ports):
-    host = config["external_host"] if config["external_host"] else config["host"]
+    host = config["host"]
     domains = get_subdomains(host)
     not_domains = get_not_subdomains(host)
+
     ports_ = {}
     for scheme, ports_used in ports.iteritems():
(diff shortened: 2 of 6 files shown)


w3c-bots:
# Build PASSED

Started: 2018-01-30 15:51:54
Finished: 2018-01-30 16:05:11





- [WPT PR Status](http://pulls.web-platform-tests.org/build/23524)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/335214397)

gsnedders:
@jgraham can you rebase?

gsnedders:
Hmm, I'm still not entirely happy with "host" and "host_ip"; I think that's still confusing for the same reason as "external_host" is confusing. (Ultimately, a lot of this could be solved by actually documenting our config file!).

To me, "server_host" in place of "external_host" and "browser_host" in place of "host" would be clearest, though as per on IRC you disagree.

jgraham:
Let's deal with another raound of renaming in a followup, since this fixes the original problem @bzbarsky was experiencing.

gsnedders (approved):

PR #9262, opened 2018-01-30
Avoid surrogate math in HasRTLChars().

jgraham:
Behavior change: Certain unpaired high surrogates will result in
the text being considered RTL where that wasn't the case before.

MozReview-Commit-ID: JL7wiRjhslC

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1428774 [ci skip]


jgraham:
Code reviewed upstream.

PR #9263, opened 2018-01-30
Remove disabled attribute from Element Clear test.

jgraham:
<select> is not a mutable form control element and it should not
be possible to call the WebDriver:ElementClear command on it.
We test this in test_select as part of element_clear.py in WPT,
but the test uses a <select disabled> element which triggers the
wrong code path in Marionette.

MozReview-Commit-ID: 4ybEdLY6OEo

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1432804 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9263)

### DIFF
--- a/webdriver/tests/interaction/element_clear.py
+++ b/webdriver/tests/interaction/element_clear.py
@@ -190,7 +190,7 @@ def test_input_file_multiple(session, text_file):
 
 def test_select(session):
     session.url = inline("""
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #9265, opened 2018-01-30
Add wpt-tests to check if insertParagraph command splits or doesn't split inline elements at their edge

jgraham:
Spec bug: https://github.com/w3c/editing/issues/172

Inline elements in paragraph except anchor elements which have non-empty
href attribute should be created in new paragraph even when new ones are
empty because users must expect to keep writing some text with same style,
but for anchor elements with non-empty href attribute, users must not
expect to create same link in new paragraph in most cases.

MozReview-Commit-ID: KjpOzfv4F0j

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1422234 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9265)


jgraham:
Code reviewed upstream.

PR #9267, opened 2018-01-30
Run native focussing steps on interaction commands.

jgraham:
Instead of generating custom focus events when interacting with elements,
we can run the HTMLElement.focus() function will do the correct thing.

Before this patch we only simulated focus events, whereas this
patch will actually focus the element.

MozReview-Commit-ID: IoBV2ngqOA5

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1432864 [ci skip]


jgraham:
Code reviewed upstream.

PR #9268, opened 2018-01-30
Unfocussing steps will target <body> element.

jgraham:
The WebDriver Element Send Keys command is meant to run the HTML
unfocussing steps after clearing the element.  When HTMLElement.blur()
is called on the root document element (<html>) it is the <body>
that is in focus when the unfocussing steps have run.

The same applies in the case of typing into an <iframe> element.
The controlling element of the <iframe> is the <body>, as this is
the element ath will receive focus should you blur the frame.

MozReview-Commit-ID: 19Ox3W9CF8O

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1432864 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9268)

### DIFF
--- a/webdriver/tests/element_send_keys/interactability.py
+++ b/webdriver/tests/element_send_keys/interactability.py
@@ -46,7 +46,7 @@ def test_document_element_is_interactable(session):
 
     response = send_keys_to_element(session, element, "foo")
     assert_success(response)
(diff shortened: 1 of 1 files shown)


jgraham:
Code reviewed upstream.

PR #9272, opened 2018-01-30
Move all assert helps into support.asserts.

jgraham:
For a better management of assertion helpers those should be
added to the central assertion helper module.

MozReview-Commit-ID: DYMfOFl9Fjd

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1433390 [ci skip]


jgraham:
Code reviewed upstream.

PR #9280, opened 2018-01-30
Look at the snapshots when invalidating due to stylesheet changes.

jgraham:
The selectorText test happens to pass right now because well, we don't implement
the setter yet[1], but would fail if we implemented an specific invalidation in
the way I'd have done it yesterday.

[1]: https://bugzilla.mozilla.org/show_bug.cgi?id=37468

MozReview-Commit-ID: DrMTgLzQcnk

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1432850 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9280)


jgraham:
Code reviewed upstream.

PR #9282, opened 2018-01-30
Align the .body setter with the spec a bit better.

jgraham:
There are two changes here:

1) We allow setting .body even if the root element is not an <html:html>.  This
is what the spec says to do, and what we used to do before the changes in bug
366200.  No tests for this yet, pending
https://github.com/whatwg/html/issues/3403 getting resolved.

2) We use GetBody(), not GetBodyElement(), to look for an existing thing to
replace.  This matters if there are <frameset>s involved.

MozReview-Commit-ID: JCaQGHKgzE7

Upstreamed from https://bugzilla.mozilla.org/show_bug.cgi?id=1276438 [ci skip]

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9282)


jgraham:
Code reviewed upstream.

PR #9299, opened 2018-01-31
Update move-to-fullscreen-iframe-manual.html to match the spec

foolip:
This test is from https://chromium-review.googlesource.com/776599 and
the explanation seemed to make sense spec-wise. But when matching the
spec more closely this begins to fail, and it seems that per spec
this updated behavior is what would happen.

### DIFF
--- a/fullscreen/model/move-to-fullscreen-iframe-manual.html
+++ b/fullscreen/model/move-to-fullscreen-iframe-manual.html
@@ -12,6 +12,8 @@ async_test(t => {
   // Enter fullscreen for the iframe's body element.
   trusted_request(t, iframeDoc.body, document.body);
   document.onfullscreenchange = t.step_func(() => {
+    assert_equals(document.fullscreenElement, iframe, "document's initial fullscreen element");
+    assert_equals(iframeDoc.fullscreenElement, iframeDoc.body, "iframe's initial fullscreen element");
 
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PASSED

Started: 2018-01-31 07:00:24
Finished: 2018-01-31 07:11:39


<h2>Failing Jobs</h2>
<ul>

<li>chrome:dev</li>

<li>safari:11.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/23602)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/335495786)

fullscreen/model/move-to-fullscreen-iframe-manual.html
```diff
@@ -22,10 +22,13 @@
     document.onfullscreenchange = t.step_func_done(() => {
       assert_equals(document.fullscreenElement, null, "document's fullscreen element");
```
upsuper: Probably consider adding assertions in the first level of event handler to clarify that both documents have fullscreen element at that point.

upsuper (approved):
It's definitely a weird case... but the explanation makes sense, I guess...

fullscreen/model/move-to-fullscreen-iframe-manual.html
foolip: Done!

PR #9307, opened 2018-01-31
Test that resources fetched by cross-origin CSS are not in the timeline

yoavweiss:
Closes https://github.com/w3c/resource-timing/issues/70

This is a long standing missing test. Seems like all browsers are doing the wrong thing here, which leads me to think we may need to change the spec...

### DIFF
--- a/resource-timing/no-entries-for-cross-origin-css-fetched.sub.html
+++ b/resource-timing/no-entries-for-cross-origin-css-fetched.sub.html
@@ -1,0 +1,45 @@
+<!DOCTYPE HTML>
+<meta charset=utf-8>
+<title>Make sure that resources fetched by cross origin CSS are not in the timeline.</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<body>
+<link rel=stylesheet id=cross_origin_style href="//{{domains[www1]}}:{{ports[http][1]}}{{location[path]}}/../resources/nested.css">
+<script>
+<<<<<<< HEAD
+<<<<<<< HEAD
+  let t = async_test("Make sure that resources fetched by cross origin CSS are not in the timeline.");
+  window.onload = function() {
+    // A timeout is needed as PerformanceObserver is not guaranteed to run before onload triggered.
+    t.step_timeout(function() {
+          let url = (new URL(document.getElementById("cross_origin_style").href));
+          let prefix = url.protocol + "//" + url.host;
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/23620)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/335608651)

igrigorik:
> This is a long standing missing test. Seems like all browsers are doing the wrong thing here, which leads me to think we may need to change the spec...

Mozilla claims to have address this in https://bugzilla.mozilla.org/show_bug.cgi?id=1180145 — is the test still failing for you? On Chrome side, see: https://bugs.chromium.org/p/chromium/issues/detail?id=532399#c32 — it seems like we still don't have agreement on https://github.com/w3c/ServiceWorker/issues/719. 

/cc @annevk @toddreifsteck @rniwa

yoavweiss:
I see this test failing in the latest Firefox nightly.

I agree that if this information is exposed through Service Workers, there's no point in blocking this info from Resource Timing.

annevk:
@jwatt and @wanderview can maybe help out here. I think our position is that we'd rather block it where possible, even if it's inconsistent.

yoavweiss:
I've rebased and updated the test to include an `@import` test.

Safari Tech Preview seems to pass the test.
Firefox nightly still fails because fonts are present in the timeline. There's an [open issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1423602) on that front.

resource-timing/no-entries-for-cross-origin-css-fetched.sub.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE HTML>
+<meta charset=utf-8>
+<title>Make sure that resources fetched by cross origin CSS are not in the timeline.</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<body>
+<link rel=stylesheet id=cross_origin_style href="//{{domains[www1]}}:{{ports[http][1]}}{{location[path]}}/../resources/nested.css">
```
cvazac: can you please add a comment as to what the `href` resolves to here?

yoavweiss: added

yoavweiss:
Commented on the flakiness in the [related bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1423602#c5)

cvazac (approved):

PR #9313, opened 2018-01-31
Add more createImageBitmap tests.

Ms2ger:

PR #9314, opened 2018-01-31
Imagebitmap redirect to same origin

Ms2ger:
The first commit is #9313; the second may be controversial, but I'm not in a hurry to land it.

### DIFF
--- a/2dcontext/imagebitmap/createImageBitmap-origin.sub.html
+++ b/2dcontext/imagebitmap/createImageBitmap-origin.sub.html
@@ -71,6 +71,18 @@ const arguments = [
   },
 
   {
+    name: "redirected to same-origin HTMLVideoElement",
+    factory: () => {
+      return new Promise((resolve, reject) => {
+        const video = document.createElement("video");
+        video.oncanplaythrough = () => resolve(video);
+        video.onerror = reject;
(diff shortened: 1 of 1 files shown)


2dcontext/imagebitmap/common.sub.js
```diff
@@ -43,6 +43,27 @@ function makeVideo() {
...
+        .then(response => Promise.all([response.headers.get("Content-Type"), response.arrayBuffer()]))
+        .then(([type, data]) => {
+            return new Promise(function(resolve, reject) {
+                var video = document.createElement("video");
+                video.oncanplaythrough = function() {
+                    resolve(video);
+                };
+                video.onerror = reject;
+                video.src = toDataUrl(type, data);
+                console.log(video.src)
```
annevk: Need to remove this line.

annevk (changes requested):
Also seems this needs to be rebased on top of the PR I just approved?

annevk (approved):

annevk:
I've landed this as-is for now. I think not throwing for a cross-origin to same-origin redirect might actually be okay, given how we deal with `<iframe>` (it's same-origin there too), but it's indeed not defined well at the moment and it's best to have a test of sorts in.

PR #9320, opened 2018-01-31
Add some tests for no-cors and navigate requests for Blob URLs.

mkruisselbrink:
Verifying behavior as discussed in whatwg/fetch#666. Don't land yet, since these tests currently don't match the spec.

### DIFF
--- a/FileAPI/url/fetch-navigate.window.js
+++ b/FileAPI/url/fetch-navigate.window.js
@@ -1,0 +1,110 @@
+// META: script=resources/test-helpers.sub.js
+
+// We can't really detect when a cross origin navigation has failed, so instead
+// detect when it succeeds, and assume it failed if it didn't succeed before
+// this timeout.
+const load_timeout = 1000;
+
+promise_test(t => {
+  const blob_contents = 'hello world';
+  const blob = new Blob([blob_contents], {type: 'text/html'});
+  return create_cross_origin_url(t, blob)
+    .then(url => {
+      with_iframe(url)
+        .then(t.unreached_func('Loading should have failed'));
+      return new Promise(resolve => {
+        t.step_timeout(resolve, load_timeout);
+      });
+    });
+}, 'Can not load a cross-origin blob URL in an iframe.');
+
+function receive_message_on_cross_origin_channel(t, channel_name) {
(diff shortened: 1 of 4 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/23650)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/335809214)

mkruisselbrink:
Any suggestions on how to make these tests less flaky are welcome... not sure how to reliable detect cross-origin loading failure, other than the timeout on loading success I'm currently doing (and as it shows, apparently a 1 second timeout isn't long enough to not be flaky...)

annevk (approved):
Tests look great, though I wonder if they should be in the /fetch directory somewhere instead, given that we'll probably define it there?

I'm not sure how to deal with the flaky behavior, I guess you can expand the timeout to 2s and see if that goes better?

mkruisselbrink:
> Tests look great, though I wonder if they should be in the /fetch directory somewhere instead, given that we'll probably define it there?

I'm not sure. We could indeed try to split the various blob URL loading related tests between FileAPI, fetch, url and XHR directories, but it isn't always clear where a test belongs, so it seemed easiest to just keep all the blob URL related tests together. But yeah, you're right that the spec changes to make these tests match the spec would be in the fetch spec.

> I'm not sure how to deal with the flaky behavior, I guess you can expand the timeout to 2s and see if that goes better?

I increased the timeout and changed the tests to async_test rather than promise_test to at least let the timeouts of all the tests overlap... Lets see how that goes.

mkruisselbrink:
Interestingly that's still flaky in chrome... I wonder if maybe our implementation is just flaky, although I haven't been able to reproduce that test passing locally...

youennf:
I usually prefer promise_test over async_test as they are run sequentially one after the other.
That makes the tests much more predictable. WebKit CI is checking JS console log messages for instance and the order might make tests fail.

I don't have a good answer for the iframe loading issue.
Maybe there could be some internals/webdriver API that could help there.

As of where it should be, fetch API already has some blob tests, as well as service worker.
I would tend to put these in fetch folder

PR #9349, opened 2018-02-01
[Code Health] Fix ABC wpt.browser.Browser

Hexcles:
* Make sure methods in subclasses have the same interface as the
  abstract methods in the base class.
* Move docstrings to the base class.
* Declare two more abstract methods: find_binary & find_webdriver,
  because multiple (though not all) subclasses already defined these
  methods, and the two methods are called externally (by wpt.run). They
  fall into the same category as version().

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9349)

### DIFF
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -18,26 +18,38 @@ logger = logging.getLogger(__name__)
 
 uname = platform.uname()
 
-def path(path, exe):
-    path = path.replace("/", os.path.sep)
-    if exe and uname[0] == "Windows":
-        path += ".exe"
-    return path
-
 
 class Browser(object):
     __metaclass__ = ABCMeta
 
     @abstractmethod
     def install(self, dest=None):
+        """Install the browser."""
         return NotImplemented
 
     @abstractmethod
-    def install_webdriver(self):
+    def install_webdriver(self, dest=None):
+        """Install the WebDriver implementation for this browser."""
         return NotImplemented
 
     @abstractmethod
-    def version(self):
+    def find_binary(self):
+        """Find the binary of the browser.
+
+        If the WebDriver for the browser is able to find the binary itself, this
+        method doesn't need to be implemented, in which case NotImplementedError
+        is suggested to be raised to prevent accidental use.
+        """
+        return NotImplemented
+
+    @abstractmethod
+    def find_webdriver(self):
+        """Find the binary of the WebDriver."""
+        return NotImplemented
+
+    @abstractmethod
+    def version(self, root):
+        """Retrieve the release version of the installed browser."""
         return NotImplemented
 
     @abstractmethod
@@ -56,7 +68,6 @@ class Firefox(Browser):
     binary = "firefox/firefox"
     platform_ini = "firefox/platform.ini"
     requirements = "requirements_firefox.txt"
-
 
     def platform_string(self):
         platform = {
@@ -126,8 +137,8 @@ class Firefox(Browser):
         untar(resp.raw, dest=dest)
         return find_executable("firefox", os.path.join(dest, "firefox"))
 
-    def find_binary(self, path=None):
-        return find_executable("firefox", path)
+    def find_binary(self):
+        return find_executable("firefox")
 
     def find_certutil(self):
         path = find_executable("certutil")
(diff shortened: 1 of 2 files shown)


jgraham (approved):

tools/wpt/browser.py
```diff
@@ -17,33 +17,41 @@
...
 
     @abstractmethod
     def install(self, dest=None):
+        """Install the browser."""
         return NotImplemented
 
     @abstractmethod
-    def install_webdriver(self):
+    def install_webdriver(self, dest=None):
+        """Install the Webdriver implementation for this browser."""
```
foolip: WebDriver capitalization here and elsewhere

tools/wpt/browser.py
```diff
@@ -239,11 +252,14 @@ def platform_string(self):
 
         return "%s%s" % (platform, bits)
 
+    def find_binary(self):
+        # ChromeDriver can find browser binary itself.
```
foolip: I take it this is supposed to be a no-op change and it probably is, but changing this kind of code does me nervous about ways in which wpt.fyi running might run. Are there any tests at all for this stuff? @gsnedders?

tools/wpt/browser.py
```diff
@@ -294,6 +309,9 @@ class ChromeAndroid(Browser):
     def install(self, dest=None):
         raise NotImplementedError
 
+    def find_binary(self):
+        raise NotImplementedError
```
foolip: A comment for this case as well?

tools/wpt/browser.py
```diff
@@ -396,14 +417,16 @@ class Edge(Browser):
...
```
foolip: Comment? Or maybe hoist all the comments into one place and remove them above instead.

tools/wpt/run.py
```diff
@@ -223,6 +223,7 @@ def setup_kwargs(self, kwargs):
             else:
                 raise WptrunError("Unable to locate or install chromedriver binary")
 
+
```
foolip: Accidental change? Just a whitespace change in a file looks weird.

foolip:
What is ABC?

tools/wpt/run.py
Hexcles: It's actually intentional, but only for styles... We should have two blank lines between top-level classes/functions according to PEP8. And this is the only offence in this file which I happened to spot.

tools/wpt/browser.py
Hexcles: Thanks. I'll grep and fix.

tools/wpt/browser.py
Hexcles: Let me put some texts in the base class.

tools/wpt/browser.py
Hexcles: An actual no-op would be `return None`. Throwing an exception would prevent future accidental use of these methods (these methods are not called currently).

I don't think there are any tests specifically for the stuff. But we do have integration tests for major products, don't we?

Hexcles (commented):
ABC is a Pythonic jargon: [Abstract Base Class](https://docs.python.org/2/library/abc.html)

w3c-bots:
# Build PASSED

Started: 2018-02-08 14:36:02
Finished: 2018-02-08 14:49:02





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24065)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/339005193)

tools/wpt/browser.py
```diff
@@ -17,33 +17,46 @@
...
+    def version(self, root):
+        """Retrieve the release version of the installed browser."""
         return NotImplemented
 
     @abstractmethod
     def requirements(self):
         """Name of the browser-specific wptrunner requirements file"""
         return NotImplemented
 
+    # TODO(robertma): This method doesn't seem to be called anywhere.
```
jgraham: This was used to work around a Chrome issue, but I think it's dead code now.

tools/wpt/browser.py
```diff
@@ -131,6 +143,12 @@ def find_certutil(self):
     def find_webdriver(self):
         return find_executable("geckodriver")
 
+    def _get_executive_path(self, path, exe):
```
jgraham: I'm not sure that making this a method rather than a function is an improvement. It's not like what it does is specific to this particular class.

gsnedders:
@Hexcles this has merge errors and review comments, would be nice to land this

tools/wpt/browser.py
Hexcles: Dead code has been removed.

tools/wpt/browser.py
Hexcles: This (along with its caller) has been removed.

Hexcles:
@gsnedders done.

PR #9376, opened 2018-02-04
[cssom] Add test for synchronization between css declaration block and style attribute

upsuper:
This is test for proposed spec change in w3c/csswg-drafts#2269.

### DIFF
--- a/css/cssom/css-style-attr-decl-block.html
+++ b/css/cssom/css-style-attr-decl-block.html
@@ -4,6 +4,7 @@
 <link rel="help" href="https://drafts.csswg.org/cssom-1/#css-declaration-blocks">
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
+<body>
 <script>
 function createTestElement(style) {
   let wrapper = document.createElement("div");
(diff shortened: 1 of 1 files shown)


upsuper:
Currently, Firefox fails one test that it generates mutation record even when removing non-existing property from the CSS declaration block. We are going to fix this in [Gecko bug 1435139](https://bugzilla.mozilla.org/show_bug.cgi?id=1435139).

WebKit and Blink both fail a test that they don't generate mutation record when setting the same value to CSS declaration block. Based on my reasoning in [w3c/csswg-drafts#2269 (comment)](https://github.com/w3c/csswg-drafts/pull/2269#issuecomment-362869138), I tend to keep the proposed behavior as-is, and ask those browsers to change their behavior here.

WebKit fails another test in addition that it doesn't clear the declaration block when the style attribute is removed. This is definitely a bug for them should be fixed.

I haven't tested Edge because I'm not using my Windows laptop at the moment.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/23842)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/337084028)

upsuper:
Hmmm, after fixing that issue, Gecko will fail the same test as Blink :/

(The test itself matches the proposed spec change, but we may want to decide what should we do for `setProperty`.)

zcorpan:
I added a commit to test
https://github.com/w3c/csswg-drafts/pull/2269/commits/03cef6ebb54f0c45fb06b6fec96ef059e4989ee2
by testing what bz suggested in
https://github.com/w3c/csswg-drafts/issues/1559#issuecomment-363132747

zcorpan (approved):

PR #9393, opened 2018-02-05
Ensure testharnessreport.js is loaded as utf8

jgraham:
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9393)

### DIFF
--- a/resources/idlharness.js.headers
+++ b/resources/idlharness.js.headers
@@ -1,1 +1,2 @@
 Content-Type: text/javascript; charset=utf-8
+Cache-Control: max-age=3600
--- a/resources/testdriver-vendor.js.headers
+++ b/resources/testdriver-vendor.js.headers
@@ -1,1 +1,2 @@
 Content-Type: text/javascript; charset=utf-8
+Cache-Control: max-age=3600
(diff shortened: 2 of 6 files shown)


gsnedders:
Why do we need this when we already have https://github.com/w3c/web-platform-tests/blob/master/resources/testharnessreport.js.headers?

jgraham:
Because a `StaticHandler` doesn't look at the `.headers` file, only a a `FileHandler` does. Maybe things should change so that .headers are applied to the URL rather than the fs path, but that is a big change and this is a one-line change.

w3c-bots:
# Build PASSED

Started: 2018-02-05 14:16:36
Finished: 2018-02-05 14:30:57





- [WPT PR Status](http://pulls.web-platform-tests.org/build/23890)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/337555659)

gsnedders:
@jgraham Can you add a comment to that affect so I don't end up confused several years down the line?

gsnedders:
@jgraham do we actually want to mark them as immutable? they might change, they aren't actually immutable.

jgraham:
It would certainly be fine on implementations where we have a separate profile per run. AIUI it only means we don't revalidate on reload so maybe it's not worthwhile, and actually it turns out to not work in firefox except over https so it's probably not worthwhile.

gsnedders:
Can you reword the commit message for the second commit on the branch, given it's now totally wrong? :)

gsnedders (approved):

PR #9401, opened 2018-02-06
Correct test for serialization of a FontFaceRule

csnardi:
According to https://drafts.csswg.org/cssom/#serialize-a-url, URLs should be serialized as a string (https://drafts.csswg.org/cssom/#serialize-a-string), which includes double quotes on each side. Correct the test to match the spec. Also update the link for the spec; the old link no longer exists.


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/23935)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/337826997)

dbaron (approved):

PR #9418, opened 2018-02-07
[css-sizing] Intrinsic percentage non-replaced widths

fantasai:
Pending resolutions / edits for https://github.com/w3c/csswg-drafts/issues/1132 
Wasn't sure how to make it self-describing (other than copying the reference into the test, for a fairly long and complicated reference) without compromising understandability.

### DIFF
--- a/css/css-sizing/intrinsic-percent-non-replaced-001-ref.html
+++ b/css/css-sizing/intrinsic-percent-non-replaced-001-ref.html
@@ -1,0 +1,71 @@
+<!DOCTYPE html>
+<title>Reference</title>
+
+<style>
+  /* establish context */
+  .container {
+    clear: both;
+    padding: 10px;
+    color: blue;
+    font: 20px/1 Ahem;
+  }
+  .zero {
+    width: 0;
+  }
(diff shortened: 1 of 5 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/23984)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/338286185)

FremyCompany (approved):
It seems those tests pass in all browsers, so I don't think there is anything controversial about them.

**One question though**: why do the test have `calc(80px + 0%)` entries but no `calc(40px + 50%)` entries? I would think that latter case is actually more complex that one where percentages are zero and can be safely ignored.

FremyCompany:
https://wptest.center/#/ejals5 has a test for latter case

FremyCompany:
https://wptest.center/#/l13fr2 for a test case with padding

FremyCompany:
And, just for good measure, a test that no browser passes:
https://wptest.center/#/f5atx4 for margin

PR #9421, opened 2018-02-07
Fix #8806: Allow about:blank as reference URL

### LINKED ISSUE #8806: Allow about:blank as a ref
cc/ @jgraham @Ms2ger @foolip 

At the moment, the lint disallows any absolute URL as a ref. We should probably allow `about:blank`, given plenty of Mozilla reftests use it as a ref.

This was hit by #8804, and fixed there by just using the pre-existing `flexbox_empty-ref.html` (which is no fix at all…).

vedantc98:
Fixes #8806.

This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9421)

### DIFF
--- a/tools/lint/lint.py
+++ b/tools/lint/lint.py
@@ -430,7 +430,7 @@ def check_parsed(repo_root, path, f):
     for reftest_node in source_file.reftest_nodes:
         href = reftest_node.attrib.get("href", "").strip(space_chars)
         parts = urlsplit(href)
-        if parts.scheme or parts.netloc:
+        if (parts.scheme or parts.netloc) and parts != urlsplit("about:blank"):
             errors.append(("ABSOLUTE-URL-REF",
                      "Reference test with a reference file specified via an absolute URL: '%s'" % href, path, None))
             continue
--- a/tools/lint/tests/test_lint.py
+++ b/tools/lint/tests/test_lint.py
@@ -213,6 +213,16 @@ def test_ref_absolute_url(caplog):
     assert "ref/absolute.html" in caplog.text
 
 
+def test_about_blank_as_ref(caplog):
+    with _mock_lint("check_path") as mocked_check_path:
+        with _mock_lint("check_file_contents") as mocked_check_file_contents:
+            rv = lint(_dummy_repo, ["about_blank-ref.html"], "normal")
+            assert rv == 0
+            assert not mocked_check_path.called
+            assert not mocked_check_file_contents.called
+    assert caplog.text == ""
+
+
 def test_ref_same_file_empty(caplog):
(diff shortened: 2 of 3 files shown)


w3c-bots:
# Build BROKEN

Started: 2018-02-07 03:00:04
Finished: 2018-02-07 03:13:16


<h2>Failing Jobs</h2>
<ul>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/23992)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/338312607)

tools/lint/lint.py
```diff
@@ -430,7 +430,7 @@ def check_parsed(repo_root, path, f):
     for reftest_node in source_file.reftest_nodes:
         href = reftest_node.attrib.get("href", "").strip(space_chars)
         parts = urlsplit(href)
-        if parts.scheme or parts.netloc:
+        if (parts.scheme or parts.netloc) and not href == "about:blank":
```
gsnedders: I wonder if we should make this `parts == ("about", "", "blank", "", "")` or `parts == urlsplit("about:blank")` to deal with characters being escaped in the path (`blank`). @jgraham?

jgraham: I hope no one ever does that, but I guess the latter form is OK.

vedantc98: I'll update the condition, but I'm unsure of how to test this bug.
Since `about:blank` will be treated as a relative file when I add it to the href attribute in the test file.

vedantc98: @jgraham What changes should I make?

tools/lint/tests/test_lint.py
```diff
@@ -213,6 +213,16 @@ def test_ref_absolute_url(caplog):
     assert "ref/absolute.html" in caplog.text
 
 
+def test_about_blank_as_ref(caplog):
+    with _mock_lint("check_path") as mocked_check_path:
+        with _mock_lint("check_file_contents") as mocked_check_file_contents:
+            rv = lint(_dummy_repo, ["about:blank"], "normal")
```
gsnedders: You don't want `about:blank` here (that'll just try and lint a file called `about:blank`); you want to add a file to `tools/lint/tests/dummy/ref` with a `<link rel=match href=about:blank>` in it (see the other files in that directory), and then you'll need to add a second file for the other test mentioned above.

tools/lint/lint.py
gsnedders: Replace the right hand side of the and here with `parts != urlsplit("about:blank")` and add a second test with `<link rel=match href=about:bl%61nk>`.

gsnedders (approved):

Issue #9111, opened 2018-01-19
[WebRTC] RTCIceTransport.html : dependency on SctpTransport


aboba:
The purpose of https://github.com/w3c/web-platform-tests/blob/master/webrtc/RTCIceTransport.html is to test the RTCIceTransport interface.  However, because it depends upon support for the RTCSctpTransport interface, on a browser that does not support RTCSctpTransport, but does support RTCIceTransport, the tests will fail, regardless of how the RTCIceTransport interface is implemented. 

A better design would be to first test for support of the RTCIceTransport object, and then if it is supported, test for support of other objects (e.g. DtlsTransport, RtpSender, RtpReceiver, SctpTransport).  Based on these results, an IceTransport could be obtained so to allow other tests to be run.  For example, on a browser that supports IceTransport, DtlsTransport and RtpReceiver, receiver.transport.transport could be used to obtain access to an IceTransport.

dontcallmedom:
given that this would have only provided different results in old Edge, and unless it becomes clear it would help with Firefox or Safari, I'm inclined to close this issue

aboba:
Safari supports IceTransport but not SctpTransport. So this is a potentially misleading test. Chrome and new Edge pass.

PR #9424, opened 2018-02-07
 Rewrite RTC ICE and DTLS transport tests with alternative dependencies 

### LINKED ISSUE #9110: [WebRTC] RTCDtlsTransport-getRemoteCertificates.html : dependency on SctpTransport
The purpose of https://github.com/w3c/web-platform-tests/blob/master/webrtc/RTCDtlsTransport-getRemoteCertificates.html  is to test the RTCDtlsTransport interface. However, because it depends upon support for the RTCSctpTransport interface, on a browser that does not support RTCSctpTransport, but does support RTCDtlsTransport, the tests will fail, regardless of how the RTCDtlsTransport interface is implemented.

A better design would be to first test for support of the RTCDtlsTransport object, and then if it is supported, test for support of other objects (e.g. RtpSender, RtpReceiver, SctpTransport). Based on these results, a DtlsTransport could be obtained so to allow other tests to be run. For example, on a browser that supports DtlsTransport and RtpReceiver, receiver.transport could be used to obtain access to a DtlsTransport.

soareschen:
Fixes #9110 and #9111.

This PR rewrites the tests for RTCDtlsTransport and RTCIceTransport to have separate tests for getting the transports from both RTP sender/receiver and SCTP transport. The common parts of the assertions are factored out into separate functions so that they can be tested in multiple test cases.

I have tried to write the tests as dependency-neutral as possible, and test this specifically on Microsoft Edge 42.17083. However there are still some issues that cause the tests to still fail on Edge:

- The RTCPeerConnection() constructor on Edge is non compliant and require minimally `new RTCPeerConnection({ iceServers: [] })` to work. On my Windows machine none of the WPT WebRTC tests that use `new RTCPeerConnection()` would pass. I made local modification to all constructors so that the tests can at least run on Edge locally.

- Edge only supports the legacy `addStream()` method and not `addTrack()` and `addTransceiver()`. I have specifically added a `addTrackOrTransceiver()` helper function that calls either `addTransceiver()`, `addTrack()`, or even `addStream()` so that we can continue testing the higher level features.

- Edge does not implement `getSenders()` and `getReceivers()`. This makes it impossible to retrieve the `RTCRtpSender` and `RTCRtpReceiver` instances so that we can get the DTLS and ICE transport objects. The workaround of getting hold of an `RTCRtpSender` using `addTrack()` also does not work, because Edge only implements `addStream()` which returns only `undefined`.

So despite my best effort I am unable to tweak the tests to pass on any of the current browsers. :( (perhaps the methodology also doesn't feel right)

If there are internal builds of browsers that have implemented the new features, please do run the tests and feedback on the test result is appreciated.


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24010)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/338418206)

nils-ohlmeier:
Firefox hasn't implemented any of the transport objects. So these will continue to fail in the foreseeable future.

fippo (commented):
including adapter.js will get you an Edge RTCPeerConnection which supports addTrack and ice transports as well as dtls transports. It comes with its own quirks though (e.g. icetransport.statechange is still named icestatechange). More inline...

webrtc/RTCTransport-helper.js
```diff
@@ -0,0 +1,193 @@
...
+    'Expect pc to have at least one sender');
+
+  assert_greater_than(receivers.length, 0,
+    'Expect pc to have at least one receiver');
+
+  const dtlsTransports = new Set();
+
+  for (const sender of senders) {
+    dtlsTransports.add(sender.transport);
+    dtlsTransports.add(sender.rtcpTransport);
```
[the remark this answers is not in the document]
soareschen: Hmm in the section for Set the RTCSessionDescription, the sender.transport and sender.rtcpTransport are always set together. Though I guess we can ignore the fact here and write test elsewhere for transport and rtcpTransport must both be set.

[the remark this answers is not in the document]
fippo: " instead let rtcpTransport equal transport." -- right, its not null. Still, this isn't relevant to check here.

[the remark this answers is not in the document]
soareschen: I have changed the test to continue if it can get at least one RTCDtlsTransport from the sender/receiver.

webrtc/RTCTransport-helper.js
```diff
@@ -0,0 +1,193 @@
...
+// transports for RTP and RTCP.
+function getDtlsTransportsFromSenderReceiver(pc) {
+  const senders = pc.getSenders();
+  const receivers = pc.getReceivers();
+
+  assert_greater_than(senders.length, 0,
+    'Expect pc to have at least one sender');
+
+  assert_greater_than(receivers.length, 0,
+    'Expect pc to have at least one receiver');
```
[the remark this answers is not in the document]
soareschen: I have changed the test to continue if it can get at least one RTCDtlsTransport from the senders/receivers.

webrtc/RTCIceTransport.html
```diff
@@ -114,77 +70,90 @@
...
-        iceTransport1.getRemoteCandidates().length,
-        iceTransport2.getLocalCandidates().length,
-        `Expect iceTransport1 to have same number of remote candidate as iceTransport2's local candidates`);
+    return createIceTransportsFromSenderReceiver(pc1, pc2)
+    .then(([iceTransports1, iceTransports2]) => {
+      const iceTransports = [...iceTransports1, ...iceTransports2];
 
-      const candidatePair1 = iceTransport1.getSelectedCandidatePair();
-      const candidatePair2 = iceTransport2.getSelectedCandidatePair();
+      return Promise.all(iceTransports.map(
```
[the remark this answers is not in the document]
soareschen: That's a good idea. I have changed it to wait for pc.connectionState to become `connected`.

webrtc/RTCDtlsTransport-getRemoteCertificates.html
```diff
@@ -4,102 +4,82 @@
...
-
-      const testedTransports = new Set();
-
-      // Callback function that test the respective DTLS transports
-      // when they become connected.
-      const onConnected = t.step_func(dtlsTransport => {
-        const certs = dtlsTransport.getRemoteCertificates();
+    return createDtlsTransportsFromSenderReceiver(pc1, pc2)
+    .then(([dtlsTransports1, dtlsTransports2]) => {
+      const dtlsTransports = [...dtlsTransports1, ... dtlsTransports2];
```
alvestrand: Please use spaces around the ... operator.

webrtc/RTCPeerConnection-helper.js
```diff
@@ -418,6 +418,28 @@ function getUserMediaTracksAndStreams(count, type = 'audio') {
   });
 }
 
+// A dependency agnostic way of adding track depending
+// on the supported method on a browser. Uses addTransceiver(track)
```
alvestrand: Rewrite this to be more readable:
* If addTransceiver(track) is available, use that.
* If addtrack(track, mediastream) is available, use that.
* Otherwise, use addStream(mediastream).

webrtc/RTCPeerConnection-helper.js
```diff
@@ -418,6 +418,28 @@ function getUserMediaTracksAndStreams(count, type = 'audio') {
...
+    // Note: The mediaStream argument is optional in the spec, but we require
+    // it for now so that we can test older versions of Firefox that only
+    // implements addTrack with compulsory mediaStream argument.
+    return pc.addTrack(track, mediaStream);
+  } else if (typeof pc.addStream === 'function') {
+    // FIXME: Remove this case of using addStream once most major browsers
+    // support addTrack or addTransceiver.
+    return pc.addStream(mediaStream)
+  } else {
+    throw new Error('This test requires either addTrack or addTransceiver implemented');
```
alvestrand: Make the error correct - "This test requires one of addTrack, addTransceiver or addStream to be implemented". We can remove addStream from the error message at the same time as removing it from the code.

webrtc/RTCTransport-helper.js
```diff
@@ -0,0 +1,184 @@
...
+    exchangeIceCandidates
+    doSignalingHandshake
+*/
+
+// Wait for all DTLS and ICE transport of a peer connection
+// to turn to connected or completed state. When the RTCPeerConnectionState
+// is "connected", all RTCIceTransports and RTCDtlsTransports are in the
+// "connected", "completed" or "closed" state and at least one of them is in the
+// "connected" or "completed" state.
+function waitConnectingPc(pc) {
```
alvestrand: This name is not obvious. Can we call it "waitForConnectedState"?

webrtc/RTCTransport-helper.js
```diff
@@ -0,0 +1,184 @@
...
+
+  const iceTransport = dtlsTransport.transport;
+  assert_true(iceTransport instanceof RTCIceTransport,
+    'Expect dtlsTransport.transport to be an RTCIceTransport');
+
+  return iceTransport;
+}
+
+// Get one or more RTCDtlsTransports from RTCRtpSender/Receiver
+// in pc. There may be more than one e.g. there are more than one
```
alvestrand: I think this is not quite right. "There may be more than one DTLSTransport if there are multiple underlying transports, such as when not using max-bundle, or when not using RTCP multiplexing."

webrtc/RTCTransport-helper.js
```diff
@@ -0,0 +1,184 @@
...
+
+  for (const receiver of receivers) {
+    adddDtlsTransport(receiver.transport);
+    adddDtlsTransport(receiver.rtcpTransport);
+  }
+
+  const dtlsTransports = [...dtlsTransportsSet];
+
+  if (dtlsTransports.length === 0) {
+    assert_unreached('Expect to get at least one unique RTCDtlsTransport from sender receivers');
```
alvestrand: "sender receivers" => "senders and receivers".

webrtc/RTCTransport-helper.js
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  return dtlsTransports;
+}
+
+function getIceTransportFromDtlsTransport(dtlsTransport) {
```
alvestrand: This function is so small, I don't think it should be a function.

You can replace it with dtlsTransport.transport (dtlsTransport.iceTransport after the change agreed to in Lyon goes through).

webrtc/RTCTransport-helper.js
```diff
@@ -0,0 +1,184 @@
...
+function getIceTransportsFromSenderReceiver(pc) {
+  const dtlsTransports = getDtlsTransportsFromSenderReceiver(pc);
+
+  const iceTransports = new Set(
+    dtlsTransports.map(getIceTransportFromDtlsTransport));
+
+  return [...iceTransports];
+}
+
+// Create DTLS transports by creating data channels
```
alvestrand: English: "Create a pair of DTLS transports by creating a data channel from pc1 to pc2. Returns an array containing two RTCDtlsTransports, obtained from each PC via pc.sctp."

webrtc/RTCTransport-helper.js
```diff
@@ -0,0 +1,184 @@
...
+  .then(() => {
+    const dtlsTransport1 = getDtlsTransportFromSctpTransport(pc1.sctp);
+    const dtlsTransport2 = getDtlsTransportFromSctpTransport(pc2.sctp);
+
+    return [dtlsTransport1, dtlsTransport2];
+  });
+}
+
+// Create DTLS transports by adding tracks and connecting
+// two peer connections. Returns an array of two array of
```
alvestrand: As above, but also "Returns an array of two arrays of RTCDtlsTransports."

webrtc/RTCTransport-helper.js
```diff
@@ -0,0 +1,184 @@
...
+    dtlsTransports.map(getIceTransportFromDtlsTransport));
+
+  return [...iceTransports];
+}
+
+// Create DTLS transports by creating data channels
+// and connecting two peer connections. Returns an
+// array of two RTCDtlsTransports, which obtained
+// from RTCSctpTransport in pc.sctp.
+function createDtlsTransportsFromSctp(pc1, pc2) {
```
alvestrand: Suggest renaming this function "connectPeerConnectionsAndReturnDtlsTransports".

I'd really prefer to not have this function, but instead have "connectPeerConnnections" and the (existing) getDtlsTransportFromSctpTransport called from the test functions - merging these two disparate things makes the test harder to read.

webrtc/RTCTransport-helper.js
```diff
@@ -0,0 +1,184 @@
...
+    .then(() => {
+      const dtlsTransports1 = getDtlsTransportsFromSenderReceiver(pc1);
+      const dtlsTransports2 = getDtlsTransportsFromSenderReceiver(pc2);
+
+      return [dtlsTransports1, dtlsTransports2];
+    });
+  });
+}
+
+function createIceTransportsFromSctp(pc1, pc2) {
```
alvestrand: These two functions also, I'd prefer that did not exist.

alvestrand (commented):
A number of things that I don't think are very readable. In particular, mixing state-changing (connecting PCs) and info-fetching (extracting the DTLS and ICE transports) makes the tests hard to read and maintain.

soareschen:
Closing this with #13886 as continuation.

PR #9431, opened 2018-02-07
Constraint fixes

kontomondo:
This fixes two things:
1) a typo in a loop where instead of using the length of `fm2`, `fm1` is used, even though it's iterating through `fm2`'s elements
2) `readonly` attribute does not apply to input type `file`, `colour` and `submit`, as per https://html.spec.whatwg.org/multipage/input.html#concept-input-apply


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24043)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/338767592)

domenic:
I don't think the second change is correct. While readonly may not "apply", the concept of "apply" doesn't impact the sentence in the spec that states

>  If the readonly attribute is specified on an input element, the element is barred from constraint validation.

That is, "specified" is unrelated to "apply".

It's notable that browsers pass all the tests you are removing, so probably we should not remove them.

kontomondo:
@domenic and below that there's a clarifying note:

> Only text controls can be made read-only, since for other controls (such as checkboxes and buttons) there is no useful distinction between being read-only and being disabled, so the readonly attribute does not apply.

So this to me is ambiguous at best.

domenic:
"specified" generally has a very specific meaning; it means "the element has the attribute". I agree it could probably be clearer.

I'm not sure exactly why the tests don't cover those other types. Maybe they should?

PR #9465, opened 2018-02-09
WebDriver: don't send a request body for HTTP GET requests

burg:
Due to the fact that Python's None is serializable as JSON,
the transport was always sending the string "null" as the body
for all requests which did not specify some other body. This
leads to HTTP GET requests with a non-zero Content-Length,
which is bad.

Fix this by only trying to encode the payload if the body
parameter has a non-default value. I don't think there is
a valid use case where a request payload is a top-level 'null'.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9465)

### DIFF
--- a/tools/webdriver/webdriver/transport.py
+++ b/tools/webdriver/webdriver/transport.py
@@ -131,13 +131,16 @@ class HTTPWireProtocol(object):
         if body is None and method == "POST":
             body = {}
 
-        try:
-            payload = json.dumps(body, cls=encoder, **codec_kwargs)
-        except ValueError:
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PASSED

Started: 2018-02-09 19:38:50
Finished: 2018-02-09 19:55:32





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24144)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/339606518)

jgraham (approved):

PR #9480, opened 2018-02-12
Fix hostnames

jugglinmike:
I believe there was an error in gh-8614. Since it removed the `hosts` property, [the templated `config.json` file which referenced that property](https://github.com/w3c/web-platform-tests/blob/9b23a0cbf6476d3f1d0b25b9bda8e5249725b00b/tools/wptrunner/wptrunner/config.json#L1) can no longer be [expanded](https://github.com/w3c/web-platform-tests/blob/9b23a0cbf6476d3f1d0b25b9bda8e5249725b00b/tools/wptrunner/wptrunner/environment.py#L130-L132). Here's an example of the error I encountered when running `./wpt run chrome` locally:

        wpt.main()
      File "/home/mike/projects/bocoup/google-wpt/web-platform-tests/tools/wpt/wpt.py", line 132, in main
        rv = script(*args, **kwargs)
      File "/home/mike/projects/bocoup/google-wpt/web-platform-tests/tools/wpt/run.py", line 419, in run
        rv = run_single(venv, **kwargs) > 0
      File "/home/mike/projects/bocoup/google-wpt/web-platform-tests/tools/wpt/run.py", line 426, in run_single
        return wptrunner.start(**kwargs)
      File "/home/mike/projects/bocoup/google-wpt/web-platform-tests/tools/wptrunner/wptrunner/wptrunner.py", line 295, in start
        return not run_tests(**kwargs)
      File "/home/mike/projects/bocoup/google-wpt/web-platform-tests/tools/wptrunner/wptrunner/wptrunner.py", line 185, in run_tests
        env_extras) as test_environment:
      File "/home/mike/projects/bocoup/google-wpt/web-platform-tests/tools/wptrunner/wptrunner/environment.py", line 96, in __enter__
        self.config = self.load_config()
      File "/home/mike/projects/bocoup/google-wpt/web-platform-tests/tools/wptrunner/wptrunner/environment.py", line 132, in load_config
        local_config = json.loads(data % self.options)
    KeyError: 'host'

This hasn't interrupted things on CI because [the Firefox configuration object still includes the `hosts`
property](https://github.com/w3c/web-platform-tests/blob/9b23a0cbf6476d3f1d0b25b9bda8e5249725b00b/tools/wptrunner/wptrunner/browsers/firefox.py#L120), and the builds for the effected browsers are configured as "allowed failures". [Those builds have recently been failing for this same reason](https://travis-ci.org/w3c/web-platform-tests/jobs/339769009).

The value in the Firefox configuration (`127.0.0.1`) differs from the value shared by all other browsers (`web-platform.test`). gh-8614 is fairly clear in its goal to reduce variability, so I'm assuming this disparity is not necessary/desirable. Since I can't think of any reason why the host should be different for Firefox, I've authored the patch to normalize on `web-platform.test`). If I'm wrong about that, then we can omit the second commit on this branch and apply only the first commit (which reverts the
original change).

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9480)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24212)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/340538974)

jugglinmike:
It appears that gh-8614 introduced an additional regression for the Sauce Labs integration. I'm investigating that now. In the mean time, the passing TravisCI build for Chrome demonstrates that this patch has the desired effect.

jugglinmike:
Okay, I think I've got that bug licked. I've submitted a patch at gh-9484. It depends on this patch.

tools/wptrunner/wptrunner/config.json
```diff
@@ -1,4 +1,4 @@
-{"host": "%(host)s",
+{"host": "web-platform.test",
```
gsnedders: Given the goal was to avoid hardcoding the domain everywhere; pretty sure the right fix for this is to make sure we always have an appropriate host provided when we evaluate it (or just drop it if we don't).

gsnedders (changes requested):

jugglinmike:
> Given the goal was to avoid hardcoding the domain everywhere;

I hope I didn't give the impression that I wasn't paying attention. I
interpreted the original goal to be reducing duplication, which is how I landed
on the change I'm proposing here. Now that I know the variation in the Firefox
configuration was intentional, I can see that this is an inappropriate
solution.

> pretty sure the right fix for this is to make sure we always have an
> appropriate host provided when we evaluate it (or just drop it if we don't).

I'm not sure what you mean here. Are you suggesting:

- Omit the `host` property from the "config" JSON template, and
- Dynamically extend the parsed object value with a `host` property if it is
  specified in the "environment options"

Or:

- specify `'web-platform.test'` as the default value of the `host` property in
  the "config" object, and
- allow that value to be optionally overridden via an "environment option"
  named `host`?

Or maybe something else?

gsnedders:
> I'm not sure what you mean here. Are you suggesting:

I'm not sure which I'm suggesting either. :) I think I had omitting it from the JSON and add it in the Python code in mind?

gsnedders:
@jugglinmike we (as in one of us!) should check why https://github.com/w3c/web-platform-tests/blob/master/tools/wpt/tests/test_wpt.py#L94 didn't catch this BTW (I presume, as I was talking to @jgraham about recently, the fact that the exit code is almost always zero, regardless of whether we've had a critical error or not).

tools/wptrunner/wptrunner/browsers/firefox.py
```diff
@@ -117,6 +117,11 @@ def env_extras(**kwargs):
 
 
 def env_options():
+    # > The host is deliberately not the default on Firefox, because Firefox is
+    # > set to always resolve the specified domains to localhost through its
+    # > `network.dns.localDomains` pref.
+    #
+    # https://github.com/w3c/web-platform-tests/pull/9480
```
gsnedders: Let's shorten this comment:

> The server host is set to 127.0.0.1 as Firefox is configured (through the network.dns.localDomains preference set below) to resolve the test domains to localhost without relying on the network stack.

gsnedders:
I'll take a closer look tomorrow, but:

 * config.json needs to go from tools/wptrunner/MANIFEST.in and tools/wptrunner/setup.py too (grep tools/wptrunner to make sure I haven't missed anything else)
 * can you write some tests for TestEnvironment.load_config?

jugglinmike:
> - config.json needs to go from tools/wptrunner/MANIFEST.in and
>   tools/wptrunner/setup.py too (grep tools/wptrunner to make sure I haven't
>   missed anything else)

Sure. There's a new "fixup!" commit with those deletions.

> - can you write some tests for TestEnvironment.load_config?

I've attempted to do so, but I ran up against a confusing error almost immediately (attempting to run using the command `tox` from the `tools/wptrunner` directory):

>     _______________________________________________________________________________ ERROR collecting wptrunner/tests/test_environment.py _______________________________________________________________________________
>     ImportError while importing test module '/home/mike/projects/web-platform-tests/tools/wptrunner/wptrunner/tests/test_environment.py'.
>     Hint: make sure your test modules/packages have valid Python names.
>     Traceback:
>     wptrunner/tests/test_environment.py:3: in <module>
>         from .. import environment
>     wptrunner/environment.py:12: in <module>
>         from wptserve.handlers import StringHandler
>     E   ImportError: No module named wptserve.handlers

I appreciate your interest in improving test coverage for the WPT CLI. I'd like to help out, but as you mentioned (and as my confusion above demonstrates), that will take some time. As we approach a week since the regression was introduced (and as we receive issue reports like gh-9520), I am feeling increased urgency about restoring the original functionality.

I submitted this patch as a way to restore functionality while honoring the refactoring, but maybe that is the wrong tack for this situation. If you would like to see this fix landed with tests, can we begin by reverting the regression in `master` as a stop gap?

gsnedders:
@jugglinmike I'd been hoping it wouldn't be too hard to write tests and that we'd get it landed today. The ImportError is odd…

gsnedders:
@jugglinmike I guess my suggestion is you clean up the history of this branch, get rid of the demo test file, and we land it after it runs through Travis?

jugglinmike:
You got it. I almost forgot to update the in-line comment as [you requested](https://github.com/w3c/web-platform-tests/pull/9480#discussion_r168351803). The new text is now available on this branch.

The original branch is available at https://github.com/bocoup/web-platform-tests/tree/fix-hostnames-prebase for posterity.

gsnedders:
@jugglinmike the first two commits are just a commit and a revert, no? so they can be dropped too.

jugglinmike:
That's right. I've removed them and squashed the remaining commits with a new message so this changeset has the proper context in the git history.

jugglinmike:
@gsnedders The TravisCI build passed. Could you please merge this patch on my behalf?

gsnedders (approved):

PR #9484, opened 2018-02-12
Correct integration with Sauce Labs service

jugglinmike:
A recently merged patch centralized configuration values which defined the domain names available during test execution [1]. That patch was based on a mistaken assumption regarding the location of that value and the stage of the test execution at which it would be available. (Specifically: the value is not present in the environment configuration object when the "environment extras" are initialized.) This interfered with the initialization of the "environment" for the Sauce Labs service, making it impossible to run the tests there.

Defer the retrieval of the environment configuration object to the time that object's context manager is activated. Update the call sites to provide the environment configuration at this moment. Also update the signature of the context manager for the `FontInstaller` "environment extra" to accept this second parameter.

[1] #8614

This depends on gh-9480

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9484)

### DIFF
--- a/tools/wptrunner/wptrunner/browsers/sauce.py
+++ b/tools/wptrunner/wptrunner/browsers/sauce.py
@@ -127,15 +127,29 @@ def get_tar(url, dest):
 class SauceConnect():
 
     def __init__(self, **kwargs):
-        self.config = kwargs["config"]
         self.sauce_user = kwargs["sauce_user"]
         self.sauce_key = kwargs["sauce_key"]
         self.sauce_tunnel_id = kwargs["sauce_tunnel_id"]
         self.sauce_connect_binary = kwargs.get("sauce_connect_binary")
         self.sc_process = None
         self.temp_dir = None
+        self.env_config = None
 
-    def __enter__(self, options):
+    def __call__(self, env_options, env_config):
+        self.env_config = env_config
+
+        return self
+
+    def __enter__(self):
+        # Because this class implements the context manager protocol, it is
+        # possible for instances to be provided to the `with` statement
+        # directly. This class implements the callable protocol so that data
+        # which is not available during object initialization can be provided
+        # prior to this moment. Instances must be invoked in preparation for
+        # the context manager protocol, but this additional constraint is not
+        # itself part of the protocol.
+        assert self.env_config is not None, 'The instance has been invoked.'
+
         if not self.sauce_connect_binary:
             self.temp_dir = tempfile.mkdtemp()
             get_tar("https://saucelabs.com/downloads/sc-4.4.9-linux.tar.gz", self.temp_dir)
@@ -153,7 +167,7 @@ class SauceConnect():
             "--metrics-address=0.0.0.0:9876",
             "--readyfile=./sauce_is_ready",
             "--tunnel-domains",
-            ",".join(self.config['domains'].values())
+            ",".join(self.env_config['domains'].values())
         ])
 
         # Timeout config vars
@@ -180,6 +194,7 @@ class SauceConnect():
             raise SauceException("Unable to start Sauce Connect Proxy. Process exited with code %s", self.sc_process.returncode)
 
     def __exit__(self, exc_type, exc_val, exc_tb):
+        self.env_config = None
         self.sc_process.terminate()
         if self.temp_dir and os.path.exists(self.temp_dir):
             try:
--- a/tools/wptrunner/wptrunner/environment.py
+++ b/tools/wptrunner/wptrunner/environment.py
@@ -84,18 +84,29 @@ class TestEnvironment(object):
         self.cache_manager = multiprocessing.Manager()
         self.stash = serve.stash.StashServer()
         self.env_extras = env_extras
+        self.env_extras_cms = None
 
 
     def __enter__(self):
         self.stash.__enter__()
         self.ssl_env.__enter__()
         self.cache_manager.__enter__()
-        for cm in self.env_extras:
-            cm.__enter__(self.options)
-        self.setup_server_logging()
+
         self.config = self.load_config()
+        self.setup_server_logging()
         ports = serve.get_ports(self.config, self.ssl_env)
         self.config = serve.normalise_config(self.config, ports)
+
+        assert self.env_extras_cms is None, (
+            "A TestEnvironment object cannot be nested")
+
+        self.env_extras_cms = []
+
+        for env in self.env_extras:
+            cm = env(self.options, self.config)
+            cm.__enter__()
+            self.env_extras_cms.append(cm)
+
         self.servers = serve.start(self.config, self.ssl_env,
                                    self.get_routes())
         if self.options.get("supports_debugger") and self.debug_info and self.debug_info.interactive:
@@ -108,8 +119,11 @@ class TestEnvironment(object):
         for scheme, servers in self.servers.iteritems():
             for port, server in servers:
                 server.kill()
-        for cm in self.env_extras:
+        for cm in self.env_extras_cms:
             cm.__exit__(exc_type, exc_val, exc_tb)
+
+        self.env_extras_cms = None
+
         self.cache_manager.__exit__(exc_type, exc_val, exc_tb)
         self.ssl_env.__exit__(exc_type, exc_val, exc_tb)
         self.stash.__exit__()
--- a/tools/wptrunner/wptrunner/font.py
+++ b/tools/wptrunner/wptrunner/font.py
@@ -17,7 +17,10 @@ class FontInstaller(object):
         self.created_dir = False
         self.fonts = fonts
 
-    def __enter__(self, options=None):
+    def __call__(self, env_options=None, env_config=None):
+        return self
+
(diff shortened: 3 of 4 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24221)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/340614218)

gsnedders:
(Also, ergh, need tests for `wptrunner.run_tests` presumably to catch this in future.)

jugglinmike:
As I understand it, the "config" we're interested in is not the value of [the like-named parameter of the `run_tests` function](https://github.com/w3c/web-platform-tests/blob/1e4fe87a7f01c0b5c614c8f601ffa68b4a00662a/tools/wptrunner/wptrunner/wptrunner.py#L135) (which is indeed available at [line 145](https://github.com/w3c/web-platform-tests/blob/1e4fe87a7f01c0b5c614c8f601ffa68b4a00662a/tools/wptrunner/wptrunner/wptrunner.py#L145)).  The "config" we need here is actually created (and subsequently normalized) [within the context manager for the `TestEnvironment` instance](https://github.com/w3c/web-platform-tests/blob/1e4fe87a7f01c0b5c614c8f601ffa68b4a00662a/tools/wptrunner/wptrunner/environment.py#L96-L98).

I haven't studied the surrounding code enough to understand the responsibilities of the two "config" objects, but adding some primitive debugging statements shows that the latter object has the properties we're interested in (see below).

Maybe we could move that code out of the context manager and into the constructor, and then instantiate the `TestEnvironment` before invoking `get_env_extras` on line 145. But that seems like a riskier alternative.

```diff
diff --git a/tools/wptrunner/wptrunner/environment.py b/tools/wptrunner/wptrunner/environment.py
index 764b479..f91d56f 100644
--- a/tools/wptrunner/wptrunner/environment.py
+++ b/tools/wptrunner/wptrunner/environment.py
@@ -96,6 +96,8 @@ class TestEnvironment(object):
         ports = serve.get_ports(self.config, self.ssl_env)
         self.config = serve.normalise_config(self.config, ports)

+        print 'TestEnvironment.enter: config = %s' % json.dumps(self.config, indent=2, sort_keys=True)
+
         for cm in self.env_extras:
             cm.__enter__(self.options, self.config)

diff --git a/tools/wptrunner/wptrunner/wptrunner.py b/tools/wptrunner/wptrunner/wptrunner.py
index d3d9d07..1577d78 100644
--- a/tools/wptrunner/wptrunner/wptrunner.py
+++ b/tools/wptrunner/wptrunner/wptrunner.py
@@ -142,6 +142,7 @@ def run_tests(config, test_paths, product, **kwargs):
          env_options, get_env_extras, run_info_extras) = products.load_product(config, product)

         ssl_env = env.ssl_env(logger, **kwargs)
+        print 'run_tests: config = %s' % json.dumps(config, indent=2, sort_keys=True)
         env_extras = get_env_extras(**kwargs)

         check_args(**kwargs)
```

Produces this output:

    run_tests: config = {
      "manifest:default": {
        "metadata": "/home/mike/projects/web-platform-tests/meta",
        "tests": "/home/mike/projects/web-platform-tests/tests",
        "url_base": "/"
      },
      "products": {},
      "web-platform-tests": {
        "branch": "master",
        "remote_url": "https://github.com/w3c/web-platform-tests.git",
        "sync_path": "/home/mike/projects/web-platform-tests/sync"
      }
    }

    TestEnvironment.enter: config = {
      "aliases": [],
      "bind_hostname": false,
      "certificate": "/home/mike/projects/web-platform-tests/tools/certs/web-platform.test.pem",
      "check_subdomains": false,
      "doc_root": "/home/mike/projects/web-platform-tests",
      "domains": {
        "": "web-platform.test",
        "www": "www.web-platform.test",
        "www1": "www1.web-platform.test",
        "www2": "www2.web-platform.test",
        "\u00e9l\u00e8ve": "xn--lve-6lad.web-platform.test",
        "\u5929\u6c17\u306e\u826f\u3044\u65e5": "xn--n8j6ds53lwwkrqhv28a.web-platform.test"
      },
      "external_host": "web-platform.test",
      "host": "web-platform.test",
      "key_file": "/home/mike/projects/web-platform-tests/tools/certs/web-platform.test.key",
      "log_level": "debug",
      "not_domains": {
        "nonexistent-origin": "nonexistent-origin.web-platform.test"
      },
      "ports": {
        "http": [
          8000,
          8001
        ],
        "https": [
          8443
        ],
        "ws": [
          8888
        ],
        "wss": [
          34279
        ]
      },
      "ssl": {
        "encrypt_after_connect": false,
        "none": {},
        "openssl": {
          "base_conf_path": null,
          "base_path": "_certs",
          "force_regenerate": false,
          "openssl_binary": "openssl"
        },
        "pregenerated": {
          "host_cert_path": "tools/certs/web-platform.test.pem",
          "host_key_path": "tools/certs/web-platform.test.key"
        },
        "type": "pregenerated"
      },
      "ws_doc_root": "/home/mike/projects/web-platform-tests/websockets/handlers"
    }

gsnedders:
Oh, well there's the bigger part of my mistake there. :(

tools/wptrunner/wptrunner/environment.py
```diff
@@ -84,18 +84,29 @@ def __init__(self, test_paths, ssl_env, pause_after_test, debug_info, options, e
...
         ports = serve.get_ports(self.config, self.ssl_env)
         self.config = serve.normalise_config(self.config, ports)
+
+        assert self.env_extras_cms == None, (
+            "A TestEnvironment object cannot be nested")
+
+        self.env_extras_cms = []
+
+        for env in self.env_extras:
+            cm = env.use(self.options, self.config)
```
jgraham: I think instead of requiring a class with a use method here, we should make the protocol that `env_extras` is an iterable of callables that return context managers, so this line would be:
`cms = env(self.options, self.config)`

and in `SauceConnect` we would s/`def use`/def __call__`

tools/wptrunner/wptrunner/browsers/sauce.py
```diff
@@ -127,15 +127,33 @@ def get_tar(url, dest):
...
+        # available during object initialization can be provided prior to this
+        # moment. The `use` method must be invoked in preparation for the
+        # context manager protocol, but this additional constraint is not
+        # itself part of the protocol. A dedicated flag is maintained in order
+        # to provide a more descriptive error to callers who fail to honor this
+        # constraint.
+        self.use_invoked = False
+
+    def use(self, env_options, env_config):
+        self.use_invoked = True
```
jgraham: Presumably this flag should be unset on `__exit__`. But rather than having a dedicated flag, why not use the fact that `env_config` is initially unset?

jgraham (changes requested):

tools/wptrunner/wptrunner/environment.py
jugglinmike: Speaking personally, I very much prefer callables to be named with verbs that
describe the effect of their invocation. It's hard to say what `environment()`
does. I know that `use` is a pretty generic term, but it does give a little
more indication that some resource is being utilized.

That said, my priority is fixing the regression, so I'm happy to oblige.

tools/wptrunner/wptrunner/browsers/sauce.py
jugglinmike: The thought occurred to me, but I decided an explicit flag would be easier to understand and harder to accidentally break. This is pretty subjective, though, so I'll re-implement as you suggest. Would you like me to unset `self.env_config` in `__exit__`?

jgraham: I'm happy with either.

jugglinmike:
Thanks @jgraham! The latest "fixup!" commit has an updated commit message

tools/wptrunner/wptrunner/browsers/sauce.py
```diff
@@ -127,15 +127,29 @@ def get_tar(url, dest):
...
         self.temp_dir = None
+        self.env_config = None
+
+    def __call__(self, env_options, env_config):
+        self.env_config = env_config
+
+        return self
+
+    def __enter__(self):
+        # Because this class implements the context manager protocol, it is
```
jgraham: You just need to change the comment and error message to account for the fact that `use` doesn't exist.

jgraham (approved):
OK to land once the comment and error message are fixed

jugglinmike:
That was careless of me. Sorry about that!

jugglinmike:
@jgraham Thank you for the linting fix. Now the the TravisCI build has passed, would you mind merging this patch?

jgraham:
Ah, sorry the reason I didn't was that there's a demo file still on this branch and I couldn't remove it with the GitHub UI.

jugglinmike:
Got it. I've rebased this branch and removed that commit. (The original version of the branch is available at https://github.com/bocoup/web-platform-tests/tree/fix-sauce-labs-integration-prebase for posterity.)

PR #9495, opened 2018-02-13
Test <video autoplay> with various broken <track src> URLs

foolip:

### DIFF
--- a/html/semantics/embedded-content/media-elements/autoplay-with-broken-track.html
+++ b/html/semantics/embedded-content/media-elements/autoplay-with-broken-track.html
@@ -1,0 +1,34 @@
+<!DOCTYPE html>
+<link rel="help" href="https://html.spec.whatwg.org/multipage/media.html#text-track-model">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/common/media.js"></script>
+<div id="log"></div>
+<script>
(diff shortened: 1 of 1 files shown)


foolip:
This created for discussion for a bug report I got. I'm not sure this is per spec yet.

foolip:
Can be tested at http://w3c-test.org/submissions/9495/html/semantics/embedded-content/media-elements/autoplay-with-broken-track.html.

foolip:
`<video autoplay> with <track src="" default=""> child` is the case I really wanted to test here, and I see it passing in Edge and failing everywhere else. And yet I can't see why it should fail, if the empty string resolves to the URL of the test page, then it should be just like the 404 case. So there must be checks for the empty string somewhere in the implementation where it doesn't belong.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24261)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/340900926)

fsoder (approved):
The case you're interested in should be covered by [1], which will yield an empty string in step 7 of [2], and then fail to load in step 10 (paragraph after the note.)

[1] https://html.spec.whatwg.org/multipage/media.html#attr-track-src
[2] https://html.spec.whatwg.org/multipage/media.html#start-the-track-processing-model

foolip:
Thanks @fsoder!

PR #9476, opened 2018-02-12
Fix import error for Servo executor.

jdm:
This regressed in 040654c78f3f021512a1f22b9b3b99925bd52d37 as part of #8614.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9476)

PR #9496, opened 2018-02-13
Add test for loading the various wptrunner products

gsnedders:
These are tests to make sure we don't regress #9476. Note, however, they fail because of https://github.com/w3c/web-platform-tests/pull/9476#pullrequestreview-96148793 (now addressed in second commit).

cc/ @jdm

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9496)

### DIFF
--- a/tools/ci/ci_tools_unittest.sh
+++ b/tools/ci/ci_tools_unittest.sh
@@ -5,21 +5,33 @@ SCRIPT_DIR=$(dirname $(readlink -f "$0"))
 WPT_ROOT=$(readlink -f $SCRIPT_DIR/../..)
 cd $WPT_ROOT
 
+run_applicable_tox () {
+    # instead of just running TOXENV (e.g., py27)
+    # run all environments that start with TOXENV
+    # (e.g., py27-firefox as well as py27)
+    local OLD_TOXENV="$TOXENV"
+    unset TOXENV
+    local RUN_ENVS=$(tox -l | grep "^${OLD_TOXENV}\(\-\|\$\)" | tr "\n" ",")
+    if [[ -n "$RUN_ENVS" ]]; then
+        tox -e "$RUN_ENVS"
+    fi
+    export TOXENV="$OLD_TOXENV"
+}
+
+
 if [[ $(./wpt test-jobs --includes tools_unittest; echo $?) -eq 0 ]]; then
     pip install -U tox codecov
     cd tools
-    tox
+    run_applicable_tox
     cd $WPT_ROOT
 else
     echo "Skipping tools unittest"
 fi
 
 if [[ $(./wpt test-jobs --includes wptrunner_unittest; echo $?) -eq 0 ]]; then
-    if [ $TOXENV == "py27" ] || [ $TOXENV == "pypy" ]; then
-        TOXENV="$TOXENV,py27-flake8"
-        cd tools/wptrunner
-        tox
-    fi
+    cd tools/wptrunner
+    run_applicable_tox
+    cd $WPT_ROOT
 else
     echo "Skipping wptrunner unittest"
 fi
--- a/tools/wptrunner/wptrunner/tests/test_products.py
+++ b/tools/wptrunner/wptrunner/tests/test_products.py
@@ -1,0 +1,24 @@
+import os
+import sys
+
+from os.path import join, dirname
+
+import pytest
+
+sys.path.insert(0, join(dirname(__file__), "..", ".."))
+
+from wptrunner import products
+
+
+def test_load_active_product(product):
+    """test we can successfully load the product of the current testenv"""
+    products.load_product({}, product)
+    # test passes if it doesn't throw
+
+
+def test_load_all_products(all_product):
+    """test every product either loads or throws ImportError"""
+    try:
+        products.load_product({}, all_product)
+    except ImportError:
+        pass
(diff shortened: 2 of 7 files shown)


gsnedders:
Hmm, both jobs passing on CI, that's a bad sign.

w3c-bots:
# Build PASSED

Started: 2018-02-13 14:08:10
Finished: 2018-02-13 14:21:14





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24264)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/340961693)

gsnedders:
OK, fixed this to actually run the tox envs specified on Travis, so this now works.

jgraham (changes requested):
Seems OK I guess. I see the point given we missed some regressions here, although the generally crazy import stuff seems to make the tests a bit delicate.

The tox stuff really needs documentation because it's far from obvious what you're doing or why.

tools/ci/ci_tools_unittest.sh
```diff
@@ -5,21 +5,30 @@ SCRIPT_DIR=$(dirname $(readlink -f "$0"))
 WPT_ROOT=$(readlink -f $SCRIPT_DIR/../..)
 cd $WPT_ROOT
 
+run_applicable_tox () {
```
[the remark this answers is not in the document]
gsnedders: Does this comment address that?

gsnedders:
@jgraham ping

jgraham:
So per my comments on IRC, I don't really undrestand all the complexity here. I don't understand why there have to be three different environment variables or why we run all the tests but then ignore the results sometimes. I think this should either be simplified or more  thoroughly commented to make it clear why this is the simplest possible approach.

gsnedders:
@jgraham What three environment variables? The only environment variables here are `TOXENV` (pre-existing) and `CURRENT_TOX_ENV` (new).

I don't know how to explain the issue further…

Fundamentally the problem is that we have:

```
pnin:wpt gsnedders$ cd web-platform-tests/tools/
pnin:tools gsnedders$ tox -l
py27
py36
pypy
pnin:tools gsnedders$ cd wptrunner/
pnin:wptrunner gsnedders$ tox -l
py27-base
py27-chrome
py27-firefox
py27-sauce
py27-servo
pypy-base
pypy-chrome
pypy-firefox
pypy-sauce
pypy-servo
py27-flake8
```

We run both of these in the same Travis job, with the same TOXENV set to the installed Python version (so `py27`, `py36`, or `pypy`).

Except, as you can see, wptrunner doesn't define a tox environment with any of those names, but rather many of them with suffixes, and we want to run all of them (because e.g. `py27-firefox` can't load the Chrome product and `py27-chrome` can't load the Firefox product). Now, we could split up the Travis jobs and introduce a wptrunner_unittest or similar and then have all of those tox environment names listed in TOXENV variables in a job in .travis.yml, but that seems like even more complexity (and likely to be forgotten when we add a new environment).

As for running all the tests but ignoring the results sometimes, how do you suggest we do this? We probably shouldn't break running the tests outside of tox, but within tox we could potentially try to more aggressively skip them before running them for other jobs?

tools/wptrunner/wptrunner/tests/test_products.py
```diff
@@ -0,0 +1,42 @@
...
+if "CURRENT_TOX_ENV" in os.environ:
+    current_tox_env = os.environ["CURRENT_TOX_ENV"]
+    current_tox_env_split = current_tox_env.split("-")
+
+product_env_map = {
+    "chrome_android": "chrome",
+    "servodriver": "servo",
+}
+
+@pytest.mark.parametrize("product", [
```
jgraham: Right, so if we keep doing this like this, we should create a test helper that returns the list of active products or the full list, and parameterise over that.

tools/ci/ci_tools_unittest.sh
```diff
@@ -5,21 +5,33 @@ SCRIPT_DIR=$(dirname $(readlink -f "$0"))
...
+    # instead of just running TOXENV (e.g., py27)
+    # run all environments that start with TOXENV
+    # (e.g., py27-firefox as well as py27)
+    local OLD_TOXENV="$TOXENV"
+    unset TOXENV
+    local RUN_ENVS=$(tox -l | grep "^${OLD_TOXENV}\(\-\|\$\)" | tr "\n" ",")
+    if [[ -n "$RUN_ENVS" ]]; then
+        tox -e "$RUN_ENVS"
```
jgraham: So I'm not super happy about running every test N times even when the test doesn't do anything vaugely product-dependent.

jgraham (changes requested):

tools/wptrunner/wptrunner/tests/test_products.py
gsnedders: I think in this case it actually makes sense, because we actually want to test something more: we want to check that it either that the product loads successfully or throws ImportError, and any other result should be a failure.

tools/ci/ci_tools_unittest.sh
gsnedders: I don't think that's worth optimising for at this stage when the whole testsuite takes under a second to run.

tools/wptrunner/wptrunner/tests/test_products.py
gsnedders: I've laid the groundwork to make this easy in future.

gsnedders:
Note the CI failure here is #9725.

jgraham (approved):

PR #9501, opened 2018-02-13
Test whether javascript: navigation aborts fetches in earlier navigation

zcorpan:
See https://github.com/whatwg/html/issues/2590. Tentative tests for now.

### DIFF
--- a/html/browsers/browsing-the-web/navigating-across-documents/javascript-url-abort/javascript-url-abort-return-value-string.tentative.html
+++ b/html/browsers/browsing-the-web/navigating-across-documents/javascript-url-abort/javascript-url-abort-return-value-string.tentative.html
@@ -1,0 +1,28 @@
+<!doctype html>
+<title>Aborting fetch for javascript:string navigation</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<link rel="help" href="https://html.spec.whatwg.org/multipage/#navigate">
+<link rel="help" href="https://github.com/whatwg/html/issues/2590">
+<div id="log"></div>
+<iframe src="support/iframe-and-links.html"></iframe>
+<script>
+async_test(test => {
+  onload = () => {
+    const child = document.querySelector('iframe').contentWindow;
+    child.document.querySelector("#slowLink").click();
+    // The step below is in a timeout. The framed page can't communicate back mid-parse because that
+    // would involve running script, which makes that navigation "mature", and we need to do this
+    // before it matures.
+    test.step_timeout(() => {
+      child.document.querySelector("#javascriptStringLink").click();
+      child.document.querySelector("iframe").onload = test.step_func_done(() => {
+        assert_false(child.childLoaded, 'child.childLoaded');
+      });
+    }, 100);
+  };
+  window.javascriptStringDocLoaded = test.step_func(() => {
+    assert_unreached("javascript: URL doc replaced the document, should be targeted to child iframe.");
+  });
+});
+</script>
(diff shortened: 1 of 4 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24292)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/341157248)

zcorpan:
Test result analysis for Firefox/Chrome/Safari. (Haven't yet run these in Edge.)

The javascript:undefined case appears interoperable. The previous ongoing fetch/navigation is not aborted. So the spec needs to be fixed here.

I see different behavior for the javascript:string case: 

* Firefox aborts the slow fetch when the javascript:string navigation happens.
* Chrome ignores (?) the `target` attribute for the javascript:string link, so that navigation replaces the link's document (this is why iframe-and-links.html is in an iframe, otherwise it wipes out the testharness.js and the test wouldn't complete). This seems like a bug in Chrome.
* Safari does **not** abort the slow fetch when the javascript:string navigation happens. Or at least it ends up with set-child-loaded.html in the nested iframe.

html/browsers/browsing-the-web/navigating-across-documents/javascript-url-abort/support/iframe-and-links.html
```diff
@@ -0,0 +1,15 @@
+<!doctype html>
+
+<iframe name="iframe"></iframe>
+<a target="iframe" href="set-child-loaded.html?pipe=trickle(d1)" id="slowLink">slow link</a>
```
annevk: Maybe add a comment what trickle(d1) does?

html/browsers/browsing-the-web/navigating-across-documents/javascript-url-abort/javascript-url-abort-return-value-string.tentative.html
```diff
@@ -0,0 +1,28 @@
...
+<div id="log"></div>
+<iframe src="support/iframe-and-links.html"></iframe>
+<script>
+async_test(test => {
+  onload = () => {
+    const child = document.querySelector('iframe').contentWindow;
+    child.document.querySelector("#slowLink").click();
+    // The step below is in a timeout. The framed page can't communicate back mid-parse because that
+    // would involve running script, which makes that navigation "mature", and we need to do this
+    // before it matures.
```
annevk: Isn't it already mature though if it has a tree you have access to?

zcorpan: iframe-and-links.html is fully loaded, but the further nested iframe should be loading and not mature when the next link is followed.

annevk: Ah right, because the first load takes longer than 100ms. Okay, this setup seems reasonable then.

annevk (approved):

html/browsers/browsing-the-web/navigating-across-documents/javascript-url-abort/javascript-url-abort-return-value-string.tentative.html
bzbarsky: @zcorpan @annevk What guarantees that this timeout will fire before the navigation matures? The 1s timer on the trickle is in a different process (the server). Depending on OS scheduling, it can in fact end up replying before this timer fires.

annevk: I think we rely on that assumption in a lot of tests. If we cannot rely on that, we basically cannot write these kind of tests...

bzbarsky: Sure you could.  For example, you could kick off two separate loads, to one of which the server responds after 100ms and to one of which it responds after 1000ms, and do your stuff when the 100ms load comes in....  That would ensure that you're using a single clock to time things.

Anyway, as written this test will randomly fail in testing infrastructure.

PR #9517, opened 2018-02-14
WebRTC: Add test for null ICE candidate

alvestrand:
The spec does not seem to allow for a null ICE candidate, and
explicitly test for presence of mid or midx attributes.

This should mean that a null ICE candidate is a type error.

New test presently passes on Chrome, fails on Firefox.

### DIFF
--- a/webrtc/RTCPeerConnection-addIceCandidate.html
+++ b/webrtc/RTCPeerConnection-addIceCandidate.html
@@ -334,6 +334,15 @@ a=rtcp-rsize
     return pc.setRemoteDescription(sessionDesc)
     .then(() =>
       promise_rejects(t, new TypeError(),
+                      pc.addIceCandidate(null)));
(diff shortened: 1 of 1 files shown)


fippo:
I continue to disagree with the spec on this but the test describes what is in the spec. @jan-ivar might have an opinion as well ;-)

fippo (approved):

soareschen:
Isn't this already been tested here? https://github.com/w3c/web-platform-tests/blob/master/webrtc/RTCPeerConnection-addIceCandidate.html#L142

jan-ivar:
This is a [known bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1318167) in Firefox.

alvestrand:
Yes, this is already tested. Don't know why I didn't see this. Closing.

PR #9527, opened 2018-02-14
Add tests for the breaking of a multicolumn across columns, including backgrounds and column rules

dbaron:
The \*-nobackground-\* tests pass in Firefox and Chrome, and the others
pass in Chrome and fail in Firefox due to what I (as a Gecko
implementor) consider to be a Firefox bug.  I haven't yet tested other
engines.

Given the level of ambiguity in the multicol spec, I suspect these tests
probably test some things that aren't formally specified.  However, I
think that's probably OK for now.  One issue that I know of that they
depend on the spec for is w3c/csswg-drafts#2309.

### DIFF
--- a/css/css-multicol/multicol-breaking-000-ref.html
+++ b/css/css-multicol/multicol-breaking-000-ref.html
@@ -1,0 +1,66 @@
+<!DOCTYPE HTML>
+<title>CSS Test Reference: breaking of a multicolumn</title>
+<meta charset="utf-8">
+<link rel="author" title="L. David Baron" href="https://dbaron.org/">
+<link rel="author" title="Mozilla" href="https://mozilla.org/">
+<style>
+
+.outer {
+  height: 100px;
+  width: 800px;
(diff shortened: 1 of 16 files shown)


dbaron:
Er, sorry, the 003 variant doesn't pass in Firefox even in the nobackground set, due to another Gecko bug.

Also, I'd note that the nobackground set tests strictly less than the others, but I'd like them there because they serve as regression tests for the current parts that *do* work in Firefox.

dbaron:
And I'd note these tests originated in [my review of bug 1346983](https://bugzilla.mozilla.org/show_bug.cgi?id=1346983#c33).

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24366)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/341657361)

mstensho (approved):

PR #9533, opened 2018-02-15
CSS2  tests were only correct if the default font has two weights.

faceless2:
font-weight-rule 005 and 007 both made the incorrect assumption that there were no weights lighter than 400 or greater than 700 in the default font.

Added new reference tests for them which don't depend on this.

### DIFF
--- a/css/CSS2/fonts/font-weight-900-ref.html
+++ b/css/CSS2/fonts/font-weight-900-ref.html
@@ -1,9 +1,9 @@
 <!DOCTYPE html>
 <meta charset="utf-8">
 <title>CSS Reference</title>
-<link rel="author" title="Intel" href="http://www.intel.com">
+<link rel="author" title="Mike Bremford" href="http://bfo.com">
 <style>
-div { font-weight: 100 }
+div { font-weight: 900 }
 </style>
 <body>
   <p>Test passes if the lines of "Filler Text" below match.</p>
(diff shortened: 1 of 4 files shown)


css/CSS2/fonts/font-weight-900-ref.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Reference</title>
+<link rel="author" title="Intel" href="http://www.intel.com">
```
dbaron: Seems like this should be you rather than Intel.  (Same for the other file.)

css/CSS2/fonts/font-weight-900-ref.html
```diff
@@ -0,0 +1,12 @@
...
+<style>
+div { font-weight: 100 }
```
dbaron: This needs to be 900 rather than 100!

dbaron (changes requested):

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24382)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/341698300)

dbaron (approved):

dbaron:
(I fixed up the commit message a bit when I merged.)

PR #5636, opened 2017-04-21
ReadableStream: Add a test that all queued writes complete on error

ricea:
**Do not merge this right now.**

This illustrates a behaviour apparently required by the streams standard but not implemented.

PR #9573, opened 2018-02-19
ReadableStream: Add a test that all queued writes complete when closing after first write

MattiasBuelens:
This PR adds more tests for the new behavior introduced in whatwg/streams#884, namely that all queued writes must complete before a pipe operation becomes completed.

When the readable stream closes, the pipe operation should wait for all writes to complete, then close the writable stream (only if `preventClose != true`), and then resolve the pipe operation's promise. The new tests check the handling of both `preventClose = true` and `preventClose != true`, although the latter case was already handled correctly by the reference implementation.

The new tests are very similar to the one added in #5636, which verifies similar behavior for when the readable stream becomes errored.

### DIFF
--- a/streams/piping/close-propagation-forward.js
+++ b/streams/piping/close-propagation-forward.js
@@ -458,4 +458,53 @@ promise_test(() => {
 
 }, 'Closing must be propagated forward: shutdown must not occur until the final write completes; preventClose = true');
 
+promise_test(() => {
+
+  const rs = recordingReadableStream();
+
+  let resolveWriteCalled;
+  const writeCalledPromise = new Promise(resolve => {
+    resolveWriteCalled = resolve;
+  });
+
+  let resolveWritePromise;
+  const ws = recordingWritableStream({
+    write() {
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24542)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/343566920)

MattiasBuelens:
It seems like the cases where `preventClose != true` are already handled correctly, although it's not very obvious *why*. The pipe operation calls `WritableStreamDefaultWriterCloseWithErrorPropagation` and awaits the returned promise, which resolves when the close request completes. The close request is queued after any other writes, so it will only complete after those writes are complete. In principle, the pipe operation could skip waiting for the queued writes to complete and instead just await the promise from the close request, but I wouldn't recommend doing this since this isn't very obvious. 😛

ricea (approved):
lgtm

PR #9598, opened 2018-02-21
Update reftests in manifest when reftests become non-reftests

foolip:
In https://github.com/w3c/web-platform-tests/pull/9523 some tests were
converted from reftests to testharness, but the incremental manifest
update mechanism doesn't handle this case. This appears to be the cause
of the mysterious Travis failures on that PR.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9598)

### DIFF
--- a/tools/manifest/manifest.py
+++ b/tools/manifest/manifest.py
@@ -86,13 +86,13 @@ class Manifest(object):
             if not is_new:
                 old_hash, old_type = self._path_hash[rel_path]
                 old_files[old_type].remove(rel_path)
-                if old_type == "reftest" and new_type != old_type:
-                    reftest_changes = True
                 if old_hash != file_hash:
                     new_type, manifest_items = source_file.manifest_items()
                     hash_changed = True
                 else:
                     new_type, manifest_items = old_type, self._data[old_type][rel_path]
+                if old_type == "reftest" and new_type != old_type:
+                    reftest_changes = True
             else:
                 new_type, manifest_items = source_file.manifest_items()
 
--- a/tools/manifest/manifest.py
+++ b/tools/manifest/manifest.py
@@ -86,6 +86,8 @@ class Manifest(object):
             if not is_new:
                 old_hash, old_type = self._path_hash[rel_path]
                 old_files[old_type].remove(rel_path)
+                if old_type == "reftest" and new_type != old_type:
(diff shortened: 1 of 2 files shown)


jgraham (approved):

w3c-bots:
# Build BROKEN

Started: 2018-02-21 10:48:58
Finished: 2018-02-21 11:02:03


<h2>Failing Jobs</h2>
<ul>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/24644)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/344257668)

jgraham (approved):

tools/manifest/manifest.py
```diff
@@ -91,6 +91,8 @@ def update(self, tree):
                     hash_changed = True
                 else:
                     new_type, manifest_items = old_type, self._data[old_type][rel_path]
+                if old_type == "reftest" and new_type != old_type:
```
gsnedders: This might need to include "reftest_node" too?

foolip: I actually don't quite understand what "reftest_node" is. Does it contain all of files involved in a reftest, everything that's not used a ref, or something else? What kind of renaming would one have to do to run into a problem here?

gsnedders: If we had:

a.html: `<link rel=match href=b.html>`
b.html: `<link rel=match href=c.html>`
c.html: `hello world`

Then: 

a.html would be a reftest (it's a test we want to run)
b.html would be a reftest_node (it's in the reftest graph and can itself be run)
c.html would be a support (because we can't tell anything about it statically)

@jgraham I think that's right?

jgraham: Yeah. I guess it's like theoretically possible you have A==B==C and you update B to be a testharness test but still use it as the ref for A, but that seems pretty unlikely, so this case might not get hit in practice, so it's a bit of an edge case correctness issue. File a bug and mark it "good first bug".

gsnedders: I think a less hypothetical case is if b.html changes to `support`? Filing an issue now.

gsnedders: https://github.com/w3c/web-platform-tests/issues/9671

PR #9600, opened 2018-02-21
Removing ES6 constructs from pages used in WebDriver tests

jimevans:
Internet Explorer doesn't support ES6 constructs like fat arrow functions.
So that the WebDriver tests can be executed against IE, the pages loaded
in the browser by the tests shouldn't include such constructs.

### DIFF
--- a/webdriver/tests/element_click/bubbling.py
+++ b/webdriver/tests/element_click/bubbling.py
@@ -1,6 +1,5 @@
 from tests.support.asserts import assert_success
 from tests.support.inline import inline
-from time import sleep
 
 def click(session, element):
     return session.transport.send(
--- a/webdriver/tests/element_click/bubbling.py
+++ b/webdriver/tests/element_click/bubbling.py
@@ -1,6 +1,6 @@
 from tests.support.asserts import assert_success
 from tests.support.inline import inline
-
+from time import sleep
 
 def click(session, element):
     return session.transport.send(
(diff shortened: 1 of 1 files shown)


webdriver/tests/element_click/bubbling.py
```diff
@@ -67,17 +68,19 @@ def test_spin_event_loop(session):
...
+            setTimeout(function() { window.delayedClicks.push(target); }, 100);
           });
         }
         </script>
         """)
     three, two, one = session.find.css("div")
     one.click()
 
-    delayed_clicks = session.execute_script("return window.delayedClicks")
+    delayed_clicks = session.execute_script("return window.delayedClicks;")
```
andreastt: FWIW the semicolon shouldn’t be necessary as the injected script is wrapped in a function, and semicolons aren’t required in JS.

webdriver/tests/element_click/bubbling.py
```diff
@@ -1,6 +1,6 @@
 from tests.support.asserts import assert_success
 from tests.support.inline import inline
-
+from time import sleep
```
andreastt: Unused?

andreastt (changes requested):

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24653)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/344362311)

webdriver/tests/element_click/bubbling.py
jimevans: D'oh! Fixed in another commit.

andreastt (approved):

webdriver/tests/element_click/bubbling.py
jimevans: Reverted to previous syntax, without semicolons.

PR #9601, opened 2018-02-21
Adding additional enabled <option> element in WebDriver test page

jimevans:
Clicking on a disabled <option> element should not allow it to become
selected. However, if the disabled <option> is the only child element of
the <select> element, it may be marked as selected by the user agent. This
commit allows the test to attempt to call elementClick on a disabled
<option> without it being selected.

### DIFF
--- a/webdriver/tests/element_click/select.py
+++ b/webdriver/tests/element_click/select.py
@@ -217,6 +217,7 @@ def test_option_disabled(session):
     session.url = inline("""
         <select>
           <option disabled>foo
+          <option>bar
         </select>""")
     option = session.find.css("option", all=False)
     assert not option.selected


andreastt:
Is that what HTML says? That the selected option of `<select disabled>` can be the first if there are two or more `<option>`s? Or is this working around a defect in IE?

jimevans:
I'm not entirely sure. Bear in mind, though, that we're not doing `<select disabled>`. What we have is `<select><option disabled></select>`.

andreastt:
Hm OK. I somewhat suspect this is working around an IE specific bug, but it doesn’t really matter since we’re testing WebDriver and having an additional `<option>` isn’t going to cause problems for other WebDriver imeplemtations.

andreastt (approved):

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24654)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/344364359)

PR #9628, opened 2018-02-22
Make percentage height calculation quirk tests match implementations

foolip:
These tests are all failing in the same way in Chrome, Edge, Firefox and
Safari: https://wpt.fyi/quirks/percentage-height-calculation.html

They came up when looking at tests that have been failing for a very
long time: https://docs.google.com/document/d/1kFYTqUMEbo9p87i6wAzORoicMWgHsuYwL2CgtvdhDaE/edit?usp=sharing

### DIFF
--- a/quirks/percentage-height-calculation.html
+++ b/quirks/percentage-height-calculation.html
@@ -64,7 +64,7 @@
         {style:'#foo { height:100px } #test { height:calc(100% + 1px) }', body:'<div id=foo><div id=test></div></div>', q:101, s:101},
         {style:'#foo { height:100px } #test { height:5px; height:calc(100% + 1px) }', body:'<div id=foo><div><div id=test></div></div></div>', q:0, s:0},
         {style:'html { display:inline } #test { height:100% }', body:'<div id=test></div>', q:184, s:0}, // display:inline on root has no effect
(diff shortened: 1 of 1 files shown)


foolip:
I did this without looking at what the spec says. I think the first instance may just have been a mistake, since 200-10*2 = 180, but the ones involving namespaces I'm not sure about. https://quirks.spec.whatwg.org/#the-percentage-height-calculation-quirk does not take namespaces into account.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24736)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/344704808)

zcorpan:
The ones involving namespaces now interact with the body element fills the viewport quirk, I think. An earlier revision of the spec aligned with gecko/presto but then gecko was changing or planning to change to webkit/blink behavior. The tests were not all updated.

So the behavior in impl is that the document element is not checked for being "html" in HTML namespace. The spec does check that for body fills html quirk. That's s for being consistent with how css 2.1 background applies to body.

But no objection to aligning with impl here.

Filed https://github.com/whatwg/quirks/issues/25

zcorpan (approved):

PR #9634, opened 2018-02-22
Updating test_element to correctly test for 'checked' property

jimevans:

### DIFF
--- a/webdriver/tests/state/get_element_property.py
+++ b/webdriver/tests/state/get_element_property.py
@@ -161,8 +161,7 @@ def test_element(session):
     session.url = inline("<input type=checkbox>")
     element = session.find.css("input", all=False)
     element.click()
-    assert False == session.execute_script("return arguments[0].hasAttribute('checked')",
-                                           args=[element])
+    assert session.execute_script("return arguments[0].hasAttribute('checked')", args=(element,)) is False
     result = session.transport.send("GET", "session/{session_id}/element/{element_id}/property/checked"
                                     .format(session_id=session.session_id,
                                             element_id=element.id))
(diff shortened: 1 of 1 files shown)


jimevans:
I've changed the test to (1) correctly verify there is no "checked" attribute on the element, and to (2) correctly retrieve the "checked" property of the element.

webdriver/tests/state/get_element_property.py
```diff
@@ -161,9 +161,9 @@ def test_element(session):
     session.url = inline("<input type=checkbox>")
     element = session.find.css("input", all=False)
     element.click()
-    assert None == session.execute_script("return arguments[0].hasAttribute('checked')",
-                                          args=[element])
-    result = session.transport.send("GET", "session/{session_id}/element/{element_id}/property/id"
+    assert False == session.execute_script("return arguments[0].hasAttribute('checked')",
+                                           args=[element])
```
andreastt: This is fine, but do you mean also replacing `==` with `is`, and converting the args list to a tuple? This statement should read:

```py
assert session.execute_script("return arguments[0].hasAttribute('checked')", args=(element,)) is False
```

andreastt (changes requested):
This is fine, but I wonder if you can make one additional change whilst you’re at it.

webdriver/tests/state/get_element_property.py
jimevans: Gladly. Done in next commit.

andreastt (approved):

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24756)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/344860986)

PR #9651, opened 2018-02-23
Switch external_host to host_ip.

jgraham:
When running tests in Firefox we don't require that the hostname is
actually resolvable since we set up some internal DNS overrides. This
means that we don't actually need to have web-platform.test defined in
/etc/hosts. As a consequence when we check the server is responding,
the check will typically fail for Firefox developers if we try to
access by hostname. Previously we made use of a special dance in which
external_host held the ip at the point of the check and was later
replaced by host to actually run tests. But that changed, and instead
we relied on the fact that we (accidentially) didn't actually fail in
the case the servers couldn't be reached by host name. However in some
cases things still failed because local DNS resolved web-patform.test
to some ip address, causing hangs or other badness.

In this patch we simply add a host_ip property to the config, and
connect to that to test the servers are repopnsing rather than the
hostname. We also actually fil if we didn't manage to connect to all
the servers.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9651)

### DIFF
--- a/config.default.json
+++ b/config.default.json
@@ -1,7 +1,7 @@
 {"host": "web-platform.test",
  "doc_root": null,
  "ws_doc_root": null,
- "external_host": null,
+ "host_ip": null,
  "ports":{"http":[8000, "auto"],
           "https":[8443],
           "ws":["auto"],
--- a/tools/ci/ci_wptrunner_infrastructure.sh
+++ b/tools/ci/ci_wptrunner_infrastructure.sh
@@ -16,10 +16,14 @@ test_infrastructure() {
 }
 
 main() {
-    hosts_fixup
     PRODUCTS=( "firefox" "chrome" )
     for PRODUCT in "${PRODUCTS[@]}"; do
-        if [ $(echo $PRODUCT | grep '^chrome') ]; then
+        if [ "$PRODUCT" != "firefox" ]; then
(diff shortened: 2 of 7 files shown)


gsnedders:
Do we have any test that catches the previous breakage (i.e., https://github.com/w3c/web-platform-tests/pull/9650#issuecomment-368100013)? Enough of our tests involve starting the server, so how did it pass CI?

w3c-bots:
# Build PASSED

Started: 2018-02-28 17:14:17
Finished: 2018-02-28 17:30:22





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25025)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/347358798)

jgraham:
It passed CI because if you have the hosts defined in `/etc/hosts` then nothing was broken. This only makes a difference if you are running Fx and don't have `/etc/hosts` entries.

jgraham:
We could not create `/etc/hosts` when running the infra tests in firefox. Or I guess we could check the generated config, but that wouldn't ensure that the server uses it in the right way.

gsnedders:
Waitaminute, this is exactly what #9726 tests. Can you have a look at that?

gsnedders (approved):

PR #9652, opened 2018-02-23
Check if git is a batch file on Windows

csnardi:
In the steps for manifest generation, if git is contained in a batch file, `subprocess` will not be able to find git and will throw a `WindowsError`. This change makes it so that if a `WindowsError` is thrown, it will check if using `git.bat` works.

Chromium has run into this issue on Windows (see https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/7Q0cv67h5p8). I'm not sure if this is the best solution, open to thoughts on that.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9652)

### DIFF
--- a/tools/manifest/vcs.py
+++ b/tools/manifest/vcs.py
@@ -14,7 +14,11 @@ class Git(object):
     def get_func(repo_path):
         def git(cmd, *args):
             full_cmd = ["git", cmd] + list(args)
-            return subprocess.check_output(full_cmd, cwd=repo_path, stderr=subprocess.STDOUT)
+            try:
+                return subprocess.check_output(full_cmd, cwd=repo_path, stderr=subprocess.STDOUT)
+            except WindowsError:
+                full_cmd[0] = "git.bat"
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PASSED

Started: 2018-02-23 19:43:05
Finished: 2018-02-23 19:58:29





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24828)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/345390059)

Ms2ger (approved):

csnardi:
@jgraham The other options here are switching to using `shell=True` (but I didn't see any other use of `shell=True` in the code), or basically trying every extension in `%PATHEXT%` to see if one works. I agree that it does seem overly specific, but I wasn't sure which of these options would be preferable.

gsnedders:
I'd have a slight preference for checking everything in `%PATHEXT%` (well, really, I'd have a preference for using stdlib `shutil.which`, but that's Py3.3+), FWIW.

PR #9655, opened 2018-02-24
Clicking on element before sending special keys in WebDriver tests

jimevans:
This commit is, admittedly, to work around a browser-specific (or perhaps
an operating-system-specific) issue. When the Alt key is pressed on
Windows, the system menu bar is activated. This causes the page in at
least one browser to lose focus, and the key reporting element will no
longer receive keydown and keyup events until focus is restored. This
causes failures in succeeding tests.

### DIFF
--- a/webdriver/tests/actions/special_keys.py
+++ b/webdriver/tests/actions/special_keys.py
@@ -20,6 +20,8 @@ def test_webdriver_special_key_sends_keydown(session,
             document.body.addEventListener("keydown",
                     function(e) { e.preventDefault() });
         """)
+    key_reporter.click()
+    session.execute_script("resetEvents();")
     key_chain.key_down(getattr(Keys, name)).perform()
 
     # only interested in keydown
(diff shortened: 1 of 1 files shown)


jimevans:
If it's unwise or unacceptable to implement such a platform-specific workaround, that's perfectly fine. I'm willing to withdraw the PR and live with documenting the failures if that's the case.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24841)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/345646277)

andreastt:
Can we make this conditional on IE? I believe we store the capabilities in `session.capabilities`.

jimevans:
Of course. It might make more sense to make it conditional on Windows, though, if other implementations have the same issue on that platform.

andreastt:
It doesn’t appear to be a problem in Firefox on Windows.

andreastt (approved):

PR #9667, opened 2018-02-26
Test BigInt as keys and values in IndexedDB

littledan:
BigInt and BigInt wrappers are supported in serialization, per
https://github.com/whatwg/html/pull/3480
This support allows them to be used as IndexedDB values.

However, BigInt is not supported as an IndexedDB key; support
has been proposed in the following PR, but that change has not
landed at the time this patch was written
https://github.com/w3c/IndexedDB/pull/231

### DIFF
--- a/IndexedDB/bigint_value.htm
+++ b/IndexedDB/bigint_value.htm
@@ -1,0 +1,65 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Values</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="support.js"></script>
+
+<script>
+// BigInt and BigInt wrappers are supported in serialization, per
+// https://github.com/whatwg/html/pull/3480
+// This support allows them to be used as IndexedDB values.
+
+let i = 0;
+function value(value, test) {
+    var t = async_test(document.title + " - " + (i++));
+    t.step(function() {
+        assert_true(test(value),
+                    "condition does not apply to initial value");
+    });
+
+    createdb(t).onupgradeneeded = function(e) {
+        e.target.result
+                .createObjectStore("store")
+                .add(value, 1);
+
+        e.target.onsuccess = t.step_func(function(e) {
+            e.target.result
+                    .transaction("store")
+                    .objectStore("store")
+                    .get(1)
+                    .onsuccess = t.step_func(function(e)
+            {
+                assert_true(test(e.target.result),
+                            "condition does not apply to deserialized result")
+                t.done();
+            });
+        });
+    };
+}
+
+value(1n, x => x === 1n);
+value(Object(1n), x => typeof x === 'object' && x instanceof BigInt
+                                             && x.valueOf() === 1n);
+
+// However, BigInt is not supported as an IndexedDB key; support
+// has been proposed in the following PR, but that change has not
+// landed at the time this patch was written
+// https://github.com/w3c/IndexedDB/pull/231
+
+function invalidKey(key) {
+    var t = async_test(document.title + " - " + (i++));
+
+    createdb(t).onupgradeneeded = function(e) {
+        let store = e.target.result.createObjectStore("store")
+        assert_throws('DataError', () => store.add(1, key));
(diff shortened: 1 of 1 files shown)


IndexedDB/bigint_value.htm
```diff
@@ -0,0 +1,65 @@
...
+
+let i = 0;
+function value(value, test) {
+    var t = async_test(document.title + " - " + (i++));
+    t.step(function() {
+        assert_true(test(value),
+                    "condition does not apply to initial value");
+    });
+
+    createdb(t).onupgradeneeded = function(e) {
```
inexorabletash: wrap the function in t.step_func() so that unexpected exceptions thrown in the async callback are tied to the test, e.g.
```js
createdb(t).onupgradeneeded = t.step_func(e => { ... });
```

IndexedDB/bigint_value.htm
```diff
@@ -0,0 +1,65 @@
...
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="support.js"></script>
+
+<script>
+// BigInt and BigInt wrappers are supported in serialization, per
+// https://github.com/whatwg/html/pull/3480
+// This support allows them to be used as IndexedDB values.
+
+let i = 0;
```
inexorabletash: Maybe call this `unique_test_id` to make the usage clear?

IndexedDB/bigint_value.htm
```diff
@@ -0,0 +1,65 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Values</title>
```
inexorabletash: Maybe clarify the name: BigInts as Values and Keys in IndexedDB ?

IndexedDB/bigint_value.htm
```diff
@@ -0,0 +1,65 @@
...
+<script src="support.js"></script>
+
+<script>
+// BigInt and BigInt wrappers are supported in serialization, per
+// https://github.com/whatwg/html/pull/3480
+// This support allows them to be used as IndexedDB values.
+
+let i = 0;
+function value(value, test) {
+    var t = async_test(document.title + " - " + (i++));
```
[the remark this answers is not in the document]
inexorabletash: It would be great if the test name said something about what was being tested in this case; rather than document.title, how about "BigInts as values in IndexedDB" 

This really helps implementers/reviewers when looking a big table of failures: you can tell what the failure was without reading the test.

Can this be passed in to the `value()` function rather than just using numbers, so it's clear what each case does?

IndexedDB/bigint_value.htm
```diff
@@ -0,0 +1,65 @@
...
+            {
+                assert_true(test(e.target.result),
+                            "condition does not apply to deserialized result")
+                t.done();
+            });
+        });
+    };
+}
+
+value(1n, x => x === 1n);
```
inexorabletash: Add case where it's not the top-level record, e.g. `{val: 1n}` ?

IndexedDB/bigint_value.htm
inexorabletash: But see below...

IndexedDB/bigint_value.htm
```diff
@@ -0,0 +1,65 @@
...
+
+// However, BigInt is not supported as an IndexedDB key; support
+// has been proposed in the following PR, but that change has not
+// landed at the time this patch was written
+// https://github.com/w3c/IndexedDB/pull/231
+
+function invalidKey(key) {
+    var t = async_test(document.title + " - " + (i++));
+
+    createdb(t).onupgradeneeded = function(e) {
```
inexorabletash: t.step_func is needed here too.

inexorabletash (commented):
Thanks for the tests!

littledan:
Thanks for this detailed review. I apologize for my delay in uploading a fixed-up PR. I intend to do so within 2-3 weeks; I'm just a bit busy with TC39 happening.

littledan:
Obsoleted by https://github.com/w3c/web-platform-tests/pull/10977

PR #9672, opened 2018-02-26
Test postMessage() exception order

annevk:
Helps with https://github.com/whatwg/html/pull/3508.

### DIFF
--- a/html/infrastructure/safe-passing-of-structured-data/transfer-errors.window.js
+++ b/html/infrastructure/safe-passing-of-structured-data/transfer-errors.window.js
@@ -18,14 +18,14 @@ function transfer_tests(name, create) {
     const transferable = await create();
     self.postMessage(null, "*", [transferable]);
     assert_throws("DataCloneError", () => self.postMessage(null, "*", [transferable]));
-  }, `Serialize should make the ${name} detached, so it cannot be transferred again.`);
+  }, `Serialize should make the ${name} detached, so it cannot be transferred again`);
 
   promise_test(async () => {
     const transferable = await create(),
           customError = new Error("hi");
     self.postMessage(null, "*", [transferable]);
     assert_throws(customError, () => self.postMessage({ get whatever() { throw customError } }, "*", [transferable]));
-  }, `Serialize should throw before a detached ${name} is found.`);
+  }, `Serialize should throw before a detached ${name} is found`);
 
   promise_test(async () => {
     const transferable = await create();
@@ -38,7 +38,7 @@ function transfer_tests(name, create) {
     };
     assert_throws("DataCloneError", () => self.postMessage(message, "*", [transferable]));
     assert_true(seen);
-  }, `Cannot transfer ${name} detached while the message was serialized.`);
+  }, `Cannot transfer ${name} detached while the message was serialized`);
 }
 
 transfer_tests("ArrayBuffer", () => new ArrayBuffer(1));
--- a/html/infrastructure/safe-passing-of-structured-data/transfer-errors.window.js
+++ b/html/infrastructure/safe-passing-of-structured-data/transfer-errors.window.js
@@ -1,0 +1,27 @@
+function assert_transfer_error(transferList) {
+  assert_throws("DATA_CLONE_ERR", () => self.postMessage({ get whatever() { throw 1 } }, "*", transferList));
+}
+
+test(() => {
+  const b = new ArrayBuffer(1);
+  assert_transfer_error([b, b]);
+}, "Cannot transfer the same object twice");
+
+test(() => {
+  const b = new ArrayBuffer(1);
+  self.postMessage(null, "*", [b]);
+  assert_transfer_error([b]);
+}, "Cannot transfer a detached ArrayBuffer");
+
+test(() => {
+  [self, self.document, new Image()].forEach(val => {
+    assert_transfer_error([val]);
+  });
+}, "Cannot transfer all objects");
+
+promise_test(() => {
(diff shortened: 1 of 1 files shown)


annevk:
It seems that Edge/Safari generally follow the specification here. Chrome/Firefox seem to handle the transfer errors later. I think either way we have to loop over the transferables before structured serializing to store them in memory, so there's probably no good reason to change the specification here.

annevk:
Safari not throwing for duplicates: https://bugs.webkit.org/show_bug.cgi?id=183124.

annevk:
Also:

* https://bugzilla.mozilla.org/show_bug.cgi?id=1441141
* https://bugs.chromium.org/p/chromium/issues/detail?id=816447

annevk:
And finally for Edge throwing the wrong exception for one test: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/16104788/.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24881)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/346277974)

html/infrastructure/safe-passing-of-structured-data/transfer-errors.window.js
```diff
@@ -0,0 +1,27 @@
+function assert_transfer_error(transferList) {
+  assert_throws("DATA_CLONE_ERR", () => self.postMessage({ get whatever() { throw 1 } }, "*", transferList));
```
mkruisselbrink: nit: DataCloneError (at least http://web-platform-tests.org/writing-tests/testharness-api.html says the _ERR form is only supported "for compatibility with existing tests")

mkruisselbrink (approved):

annevk:
I updated the tests to match the proposed change over at https://github.com/whatwg/html/pull/3557.

annevk:
@jeremyroman could you perhaps review these tests? Suggestions for more tests?

html/infrastructure/safe-passing-of-structured-data/transfer-errors.window.js
```diff
@@ -0,0 +1,42 @@
...
+
+test(() => {
+  const b = new ArrayBuffer(1);
+  self.postMessage(null, "*", [b]);
+  assert_transfer_error([b], new Error("hi"));
+  assert_throws("DataCloneError", () => self.postMessage(null, '*', [b]));
+}, "Serialize throws before a transferred detached ArrayBuffer is found");
+
+promise_test(() => {
+  return self.createImageBitmap(document.createElement("canvas")).then(bitmap => {
```
jeremyroman: Do we want to be more exhaustive? We could do something like:

```js
function transfer_tests(name, create) {
  promise_test(async () => {
    const transferable = await create();
    assert_transfer_error([transferable, transferable]);
  }, `Cannot transfer the same ${name} twice`);

  promise_test(async () => {
    const transferable = await create();
    self.postMessage(null, "*", [transferable]);
    assert_throws("DataCloneError", () => self.postMessage(null, "*", [transferable]);
  }, `Serialize should make the ${name} detached, so it cannot be transferred again.`);

  promise_test(async () => {
    const transferable = await create();
    self.postMessage(null, "*", [transferable]);
    assert_transfer_error([transferable], new Error("hi"));
  }, `Serialize should throw before a detached ${name} is found.`);

  promise_test(async () => {
    const transferable = await create();
    let seen = false;
    const message = {
      get a() {
        self.postMessage(null, '*', [transferable]);
        seen = true;
      }
    };
    assert_throws("DataCloneError", () => self.postMessage(message, "*", [transferable]));
    assert_true(seen);
  }, `Cannot transfer ${name} detached while the message was serialized.`);
}

transfer_tests("ArrayBuffer", () => new ArrayBuffer(1));
transfer_tests("MessagePort", () => new MessageChannel().port1);
transfer_tests("ImageBitmap", () => self.createImageBitmap(document.createElement("canvas")));
// OffscreenCanvas? I don't believe we have any other transferables, at least implemented in Blink.
```

That might be excessively clever, but I think the logic for all of these tests generalizes to all kinds of transferable checks, so why not check the whole matrix?

html/infrastructure/safe-passing-of-structured-data/transfer-errors.window.js
```diff
@@ -0,0 +1,42 @@
+function assert_transfer_error(transferList, exception = "DataCloneError") {
+  assert_throws(exception, () => self.postMessage({ get whatever() { throw exception } }, "*", transferList));
```
jeremyroman: nit: Is this trying too hard to do what could be two separate functions? Throwing the string "DataCloneError" as the default behavior seems odd.

jeremyroman (commented):
Makes sense to me overall.

html/infrastructure/safe-passing-of-structured-data/transfer-errors.window.js
annevk: Yeah, why not. I'll make it so. Probably not today unfortunately.

html/infrastructure/safe-passing-of-structured-data/transfer-errors.window.js
annevk: Will clean this up too.

annevk:
Ended up making those fixes today anyway.

The one thing this loses is testing `ImageBitmap`'s `close()` method, but that should probably be tested as part of that object.

jeremyroman (approved):
LGTM; thanks!

PR #9687, opened 2018-02-27
This adds the ability to download latest Servo for `wpt run`

kriti21:
Closes https://github.com/w3c/web-platform-tests/issues/9544

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9687)

### DIFF
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -431,8 +431,37 @@ class Servo(Browser):
     product = "servo"
     requirements = "requirements_servo.txt"
 
+    def platform_components(self):
+        platform = {
+            "Linux": "linux",
+            "Windows": "win",
+            "Darwin": "mac"
+        }.get(uname[0])
+
+        if platform is None:
+            raise ValueError("Unable to construct a valid Servo package for current platform")
+
+        if platform == "linux":
+            extension = ".tar.gz"
+            decompress = untar
+        elif platform == "win" or platform == "mac":
+            raise ValueError("Unable to construct a valid Servo package for current platform")
+
+        return (platform, extension, decompress)
+
     def install(self, dest=None):
-        raise NotImplementedError
+        """Install latest Browser Engine."""
+        if dest is None:
+            dest = os.pwd
+
+        platform, extension, decompress = self.platform_components()
+        url = "https://download.servo.org/nightly/%s/servo-latest%s" % (platform, extension)
+
+        decompress(get(url).raw, dest=dest)
+        path = find_executable("servo", os.path.join(dest, "servo"))
+        st = os.stat(path)
+        os.chmod(path, st.st_mode | stat.S_IEXEC)
+        return path
 
     def find_binary(self, path=None):
         return find_executable("servo")
@@ -444,7 +473,9 @@ class Servo(Browser):
         raise NotImplementedError
 
     def version(self, root):
-        return None
+        """Retrieve the release version of the installed browser."""
+        output = call(self.binary, "--version")
+        return re.search(r"[0-9\.]+( [a-z]+)?$", output.strip()).group(0)
 
 
 class Sauce(Browser):
--- a/tools/wpt/run.py
+++ b/tools/wpt/run.py
@@ -332,7 +332,8 @@ class Servo(BrowserSetup):
     browser_cls = browser.Servo
 
     def install(self, venv):
-        raise NotImplementedError
+        if self.prompt_install(self.name):
+            return self.browser.install(venv.path)
 
     def setup_kwargs(self, kwargs):
         if kwargs["binary"] is None:


w3c-bots:
# Build PASSED

Started: 2018-02-27 11:50:15
Finished: 2018-02-27 12:01:48





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24938)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/346737132)

tools/wpt/browser.py
```diff
@@ -431,8 +431,52 @@ class Servo(Browser):
...
+        platform = self.platform_string()
+
+        if platform == "linux":
+            extension = ".tar.gz"
+        elif platform == "win":
+            extension = ".msi"
+        else:
+            extension = ".dmg"
+
+        return "%s" % (extension)
```
Ms2ger: This is the same as `return extension`.

tools/wpt/browser.py
```diff
@@ -431,8 +431,52 @@ class Servo(Browser):
...
+
+        if platform == "linux":
+            bits = "-%s" % uname[4]
+            extension = ".tar.gz"
+        elif platform == "win":
+            bits = "64" if uname[4] == "x86_64" else "32"
+            extension = ".msi"
+        else:
+            bits = ""
+            extension = ".dmg"
```
Ms2ger: All of this is unused; I'd suggest returning a tuple (`return (platform, extension)`), though, so you don't need two methods.

Ms2ger (dismissed):

kriti21:
Sorry for the delay. I have made the suggested changes. Please review :) @Ms2ger

tools/wpt/browser.py
Ms2ger: Did you test the new code? I don't see how it can work, because you return a single string instead of a tuple.

kriti21: I am having trouble in understanding how to test my changes locally. Could you please guide me on that? @Ms2ger

Ms2ger:
I don't know about testing. @jdm or @jgraham might.

jdm:
The original issue suggested a command to run that is expected to work, I believe.

kriti21:
Here is screenshot of that command. Please help me with these errors. @Ms2ger  @jdm @gsnedders 

testm

tools/wpt/browser.py
```diff
@@ -431,8 +431,40 @@ class Servo(Browser):
...
+
+        return "%s%s" % (platform, extension)
+
+    def install_browser_engine(self, dest=None):
```
jdm: You renamed the `install` method to `install_browser_engine`, so the error message is letting you know that it expects the `install` method.

jdm (changes requested):

tools/wpt/browser.py
kriti21: Thanks I have corrected this. This is what it shows now.

testm1

jdm: That looks like you restored the original `install` method to its original implementation, which just raises a NotImplementedError exception.

jdm: Also, rather than attaching screenshots, I recommend copying the text out of your console and pasting it between blocks of ``` characters like this:
```
jdm@godot$ ./wpt run
```

kriti21: @jdm I. Isn't this the implementation of install method or I have done something wrong here? Why is it giving NotImplementedError after renaming it to `install` ?

jdm: @kriti21 It's hard to say without looking at the actual code you're running. Please push your changes if you'd like me to comment on them :)

kriti21:
@jdm pushed the changes :)

jdm:
@kriti21 What is the output from the command?

kriti21:
```
 0:00.11 INFO Installing browser
    wpt.main()
  File "/home/kriti/mozdev/web-platform-tests/tools/wpt/wpt.py", line 132, in main
    rv = script(*args, **kwargs)
  File "/home/kriti/mozdev/web-platform-tests/tools/wpt/run.py", line 404, in run
    **kwargs)
  File "/home/kriti/mozdev/web-platform-tests/tools/wpt/run.py", line 382, in setup_wptrunner
    kwargs["binary"] = setup_cls.install(venv)
  File "/home/kriti/mozdev/web-platform-tests/tools/wpt/run.py", line 335, in install
    raise NotImplementedError
NotImplementedError
```

jdm:
Oh, I see. The code is complaining about the class in run.py, but all your changes are in browser.py. You should be able to remove the install method in run.py, because it will fall back to [this one](https://github.com/kriti21/web-platform-tests/blob/7e82c9c008c452c266c9fb1897698a4bf0e09a71/tools/wpt/run.py#L149-L151) instead.

kriti21:
@jdm i have pushed with changes again. This is what it shows now. Please help correct this as well :)

```
 0:00.00 INFO Installing browser
Download and install servo [Y/n]? Y
    wpt.main()
  File "/home/kriti/mozdev/web-platform-tests/tools/wpt/wpt.py", line 132, in main
    rv = script(*args, **kwargs)
  File "/home/kriti/mozdev/web-platform-tests/tools/wpt/run.py", line 405, in run
    **kwargs)
  File "/home/kriti/mozdev/web-platform-tests/tools/wpt/run.py", line 383, in setup_wptrunner
    kwargs["binary"] = setup_cls.install(venv)
  File "/home/kriti/mozdev/web-platform-tests/tools/wpt/run.py", line 336, in install
    return self.browser.install(venv.path)
  File "/home/kriti/mozdev/web-platform-tests/tools/wpt/browser.py", line 474, in install
    unzip(get(url).raw, dest)
  File "/home/kriti/mozdev/web-platform-tests/tools/wpt/utils.py", line 88, in unzip
    with zipfile.ZipFile(fileobj) as zip_data:
  File "/usr/lib/python2.7/zipfile.py", line 770, in __init__
    self._RealGetContents()
  File "/usr/lib/python2.7/zipfile.py", line 811, in _RealGetContents
    raise BadZipfile, "File is not a zip file"
zipfile.BadZipfile: File is not a zip file
```

jdm:
It looks like it's expecting a zip file, but on linux we provide a tar.gz archive instead. I recommend using the untar function instead, and raising an exception if we try to use this on Windows or macOS, because those require even more specialized installation procedures (since they are not zips or tars).

kriti21:
@jdm This is the error that occurs now. 

```
 0:00.00 INFO Installing browser
Download and install servo [Y/n]? Y
    wpt.main()
  File "/home/kriti/mozdev/web-platform-tests/tools/wpt/wpt.py", line 132, in main
    rv = script(*args, **kwargs)
  File "/home/kriti/mozdev/web-platform-tests/tools/wpt/run.py", line 405, in run
    **kwargs)
  File "/home/kriti/mozdev/web-platform-tests/tools/wpt/run.py", line 383, in setup_wptrunner
    kwargs["binary"] = setup_cls.install(venv)
  File "/home/kriti/mozdev/web-platform-tests/tools/wpt/run.py", line 336, in install
    return self.browser.install(venv.path)
  File "/home/kriti/mozdev/web-platform-tests/tools/wpt/browser.py", line 480, in install
    st = os.stat(path)
TypeError: coercing to Unicode: need string or buffer, NoneType found
```
I know this error occurs when path specified is `None`. But the same approach seems to work for other browsers. Please help.

jdm:
`path` is the result of calling `find_executable`, so I recommend tracing the code there to figure out what's going wrong.

kriti21:
@jdm Could you tell me what should be the output of `find_executable`. On the website it says to download the tar file and run `./servo` inside servo directory (for linux). How can I do this in this code ?

jdm:
@kriti21 I think we probably want to follow the example of the Firefox installation code, which looks quite similar to what we're trying to do for Servo: https://github.com/kriti21/web-platform-tests/blob/832b3ec2294cf45d856b31c6479183eb98de9788/tools/wpt/browser.py#L132-L133

The second argument to find_executable allows us to specify the precise directory in which we expect to find the servo binary: https://github.com/python/cpython/blob/master/Lib/distutils/spawn.py#L169-L174

kriti21:
With the code that i have pushed now, the command `./wpt run --install-browser servo dom/historical.html` halts after displaying this 
```
0:00.06 INFO Installing browser
Download and install servo [Y/n]?
```

jdm:
@kriti21 I recommend tracing the code that runs and adding print statements to figure out what it's doing. You can start at https://github.com/kriti21/web-platform-tests/blob/8cfd02b23663ac97b4db0648e3656ca7b46c75e5/tools/wpt/run.py#L381.

kriti21:
Here is the output i get locally. Please review @jdm 
```
 0:00.10 INFO Installing browser
Download and install servo [Y/n]? Y
 2:35.39 INFO Updating test manifest /home/kriti/mozdev/web-platform-tests/MANIFEST.json
 2:35.39 INFO STDOUT: INFO:manifest:Skipping manifest download because existing file is recent
 2:38.87 INFO STDOUT: INFO:manifest:Updating manifest
 4:21.64 INFO STDOUT: DEBUG:manifest:Opening manifest at /home/kriti/mozdev/web-platform-tests/MANIFEST.json
 4:27.22 INFO Using 1 client processes
 4:27.37 INFO Starting http server on web-platform.test:8000
 4:27.38 INFO Starting http server on web-platform.test:8001
 4:27.38 SUITE_START: web-platform-test - running 1 tests
 4:27.38 INFO Running reftest tests
 4:27.39 INFO No reftest tests to run
 4:27.39 INFO Running wdspec tests
 4:27.39 INFO No wdspec tests to run
 4:27.39 INFO Running testharness tests
 4:27.39 INFO Starting runner
 4:27.40 INFO Starting http server on web-platform.test:8443
 4:27.42 TEST_START: /dom/historical.html
 4:48.28 INFO Pausing until the browser exits
 4:48.28 TEST_END: Test OK. Subtests passed 72/72. Unexpected 0
 4:48.28 INFO Pausing until the browser exits
 4:48.28 INFO No more tests
 4:48.30 WARNING u'runner_teardown': ()
 4:48.30 INFO Got 0 unexpected results
 4:48.30 SUITE_END

web-platform-test
~~~~~~~~~~~~~~~~~
Ran 73 checks (1 tests, 72 subtests)
Expected results: 73
OK
 4:48.39 INFO Closing logging queue
 4:48.39 INFO queue closed
```

jdm:
That looks like it's behaving as expected?

kriti21:
Yes

tools/wpt/browser.py
jdm: There is still a bunch of unused code in this method. I agree that combining the platform and extension methods together makes sense.

tools/wpt/browser.py
jdm: This is still the case.

tools/wpt/browser.py
```diff
@@ -431,8 +431,55 @@ class Servo(Browser):
...
     def install(self, dest=None):
-        raise NotImplementedError
+        """Install latest Browser Engine."""
+        if dest is None:
+            dest = os.pwd
+
+        extension = self.find_extension(self.platform_string())
+        url = "https://download.servo.org/nightly/%s/servo-latest%s" % (self.platform_string(), extension)
+
+        if format == "zip":
```
jdm: Where does this come from?

jdm (changes requested):

tools/wpt/browser.py
kriti21: If I use the same method to return tuple of platform and extension, i cannot use them separately in the generating url which is needed here.

kriti21:
@jdm I have removed the useless code. However, if I use the same function to return a tuple of platform and extension, I don't think i can use them at separate places in generating URL which is needed here.

kriti21:
Also as much as I understand the travis build fail logs, it says line 481 in `browser.py` has extra whitespace. I am not sure if its correct.

tools/wpt/browser.py
```diff
@@ -431,8 +431,45 @@ class Servo(Browser):
...
+        return "%s" % (extension)
+
     def install(self, dest=None):
-        raise NotImplementedError
+        """Install latest Browser Engine."""
+        if dest is None:
+            dest = os.pwd
+
+        extension = self.find_extension(self.platform_string())
+        url = "https://download.servo.org/nightly/%s/servo-latest%s" % (self.platform_string(), extension)
```
jdm: Rather than two separate functions, if there is a single function that returns (platform, extension), we can do:
```
def platform_components()
    ...
    return (platform, extension)

url = "https://download.servo.org/nightly/%s/servo-latest%s" % self.platform_components()
```

tools/wpt/browser.py
```diff
@@ -431,8 +431,45 @@ class Servo(Browser):
...
+
+        if self.platform_string() == "linux":
```
jdm: Since we only support linux right now, we can remove any code that checks for the platform or tries to support other platforms.

tools/wpt/browser.py
```diff
@@ -431,8 +431,45 @@ class Servo(Browser):
...
+            untar(get(url).raw, dest=dest)
+        else:
+            unzip(get(url).raw, dest=dest)
+
```
jdm: The TravisCI build is complaining that this blank line contains trailing whitespace.

jdm (changes requested):

tools/wpt/browser.py
Ms2ger: And if we keep this, maybe `platform_components` or whatever should return the `unzip`/`untar` function as well, to localize the dependency on the kind of files that are used on each platform.

kriti21: I am not sure about returning `unzip / untar` function from `platform_components` since `platform` and `extension` values are used to form a url which downloads the required file that is extracted using `untar / unzip` function. Would not including url or filename in `platform_components` make the function more complicated?

jdm: @kriti21 The previous comment meant something like this:
```python
def platform_components(self):
    if ...:
        return ("linux", ".tar.gz", untar)

platform, extension, decompress = self.platform_components()
url = "..." % (platform, extension)
decompress(get(url).raw, dest=dest)
```
Does that make sense?

kriti21: Oh okay! Thanks :)

kriti21:
@jdm Please review my changes. Also I haven't removed the options for platforms other than linux yet. Tell me if I should do that in this code.

tools/wpt/browser.py
```diff
@@ -431,8 +431,41 @@ class Servo(Browser):
...
+        if platform == "linux":
+            extension = ".tar.gz"
+            decompress = untar
+        elif platform == "win":
+            extension = ".msi"
+            decompress = unzip
+        else:
+            extension = ".dmg"
+            decompress = unzip
```
jdm: Given that unzip will not work for MSI or DMG files, we should do this instead:
```python
elif platform == "win" or platform == "mac":
    raise ValueError("Unable to extract Servo package for current platform")
```

jdm (changes requested):
With this change it should be good to merge. Thanks for sticking with this task!

kriti21:
@jdm I have made the final change. Hope this is ready to merge now. Thanks a lot. :)

jdm (approved):

PR #9693, opened 2018-02-27
Add more tests for data: URLs with space in body

foolip:
This is to bring attention to the surprising difference between
application/xml, application/javascript and text/javascript. In Chrome
and Firefox, only the application/javascript case fails.


annevk (approved):
@domenic should probably r+ too.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24947)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/346824582)

domenic (approved):
All tests pass with jsdom/data-urls, implying they most likely match the spec.

Maybe wroth adding a comment as to why we're testing these variations which should theoretically be redundant.

foolip:
Can't add comments in JSON, or did you have a trick in mind?

domenic:
Hmm, usually in these data files we do a thing where a line that is just a string (instead of an object/array or whatever) is skipped by the processing code. But I guess I don't see that here. Oh well, might as well merge as-is.

foolip:
Do you have an example of that? I was thinking of using https://github.com/json5/json5, but dependencies that we have to keep updating are annoying.

domenic:
https://github.com/w3c/web-platform-tests/blob/3678c44dfe811f93a796285e9531ee34d7d15682/url/urltestdata.json#L2 plus https://github.com/w3c/web-platform-tests/blob/3678c44dfe811f93a796285e9531ee34d7d15682/url/a-element.js#L28

foolip:
I see, I can try to do the same here if you like.

domenic:
Seems nice, but I don't insist on blocking the review :)

foolip:
Oh, and instead I just left this sitting. Retracting my promise, I'll just merge as is.

PR #9696, opened 2018-02-27
add XHR.prototype.open parameter toString test

michaelficarra:
This codifies the WHATWG algorithm. IE/Edge are known to fail this test, while other major browsers pass.

### DIFF
--- a/xhr/open-parameters-toString.htm
+++ b/xhr/open-parameters-toString.htm
@@ -1,0 +1,54 @@
+<!doctype html>
+<title>XMLHttpRequest: open() attempts to toString its string parameters</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div id="log"></div>
+<script>
+test(() => {
+  let log = [];
+  let expected = [
+    'method',
+    'url',
+    // NOTE: 'async' intentionally missing
+    'username',
+    'password',
+  ];
+
+  let xhr = new XMLHttpRequest;
+  xhr.open(
+    {
+      toString() {
+        log.push('method');
+        return 'get';
+      },
+    },
+    {
+      toString() {
+        log.push('url');
+        return location.href;
+      },
+    },
+    // NOTE: ToBoolean should not invoke valueOf
+    {
+      valueOf() {
+        log.push('async');
+        return true;
+      },
+    },
+    {
+      toString() {
+        log.push('username');
+        return 'username';
+      },
+    },
+    {
+      toString() {
+        log.push('password');
+        return 'password';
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/24962)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/347016236)

michaelficarra:
Turns out old Safari also `toString`s `url` before `method`.

xhr/open-parameters-toString.htm
```diff
@@ -0,0 +1,54 @@
...
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div id="log"></div>
+<script>
+test(() => {
+  let log = [];
+  let expected = [
+    'method',
+    'url',
+    // NOTE: 'async' intentionally missing
```
annevk: It'd be more helpful to comment on why it's missing. It seems that per https://tc39.github.io/ecma262/#sec-toboolean which is used for boolean conversion it'd end up being true because you passed an object. The note below therefore seems wrong too...

annevk:
Overall this seems fine to test, assuming IDL indeed defines this order for the conversion.

tobie (approved):
> Overall this seems fine to test, assuming IDL indeed defines this order for the conversion.

It does in step 11.5 of the [overload resolution algorithm](https://heycam.github.io/webidl/#dfn-overload-resolution-algorithm).

xhr/open-parameters-toString.htm
michaelficarra: I think it's missing not because it is `ToBoolean`'d and `ToBoolean` doesn't invoke `valueOf`, but because of [the `open` algorithm](https://xhr.spec.whatwg.org/#the-open()-method), particularly step 5

> If the async argument is omitted, set async to true, and set username and password to null. Note: Unfortunately legacy content prevents treating the async argument being undefined identical from it being omitted.

and step 12

> Set the synchronous flag, if async is false, and unset the synchronous flag otherwise.

I could be wrong though, in which case I will change the below note.

annevk: Before you get to that algorithm, the algorithm @tobie pointed out applies due to it using IDL. And that will convert the arguments to strings and booleans.

michaelficarra: @annevk Thanks. Updated.

annevk:
Thanks @michaelficarra. Will you file bugs on Edge and Safari?

Issue #5403, opened 2017-04-06
Tests relying on larger viewports


gsnedders:
The documentation states variously that reftests are taken at 600x600 and 800x600; they're mostly run at 600x600 and this apparently matches what was decided for the CSS testsuite years ago so let's just go with that.

Note we also have a <meta name='viewport-size'> that's totally undocumented and that only works for Servo per w3c/wptrunner#166.

This follows on from  #4619 but provides a far better list of affected tests and includes the now merged CSS testsuite.

- [ ] css/CSS2/abspos/abspos-containing-block-initial-001.xht (588x10110)
- [ ] css/CSS2/abspos/abspos-containing-block-initial-007.xht (700x588)
- [ ] css/CSS2/backgrounds/background-position-002.xht (700x588)
- [ ] css/CSS2/bidi-text/bidi-005b.xht (708x588)
- [ ] css/CSS2/bidi-text/bidi-006b.xht (708x588)
- [ ] css/CSS2/bidi-text/bidi-007b.xht (732x588)
- [ ] css/CSS2/bidi-text/bidi-008b.xht (732x588)
- [ ] css/CSS2/bidi-text/bidi-009b.xht (708x588)
- [ ] css/CSS2/bidi-text/bidi-010b.xht (732x588)
- [ ] css/CSS2/bidi-text/line-breaking-bidi-001.xht (3608x588)
- [ ] css/CSS2/bidi-text/line-breaking-bidi-002.xht (3608x588)
- [ ] css/CSS2/bidi-text/line-breaking-bidi-003.xht (694x588)
- [ ] css/CSS2/box-display/insert-block-in-blocks-n-inlines-begin-001.xht (588x775)
- [ ] css/CSS2/box-display/insert-block-in-blocks-n-inlines-begin-002.xht (588x775)
- [ ] css/CSS2/box-display/insert-block-in-blocks-n-inlines-begin-003.xht (588x775)
- [ ] css/CSS2/box-display/insert-block-in-blocks-n-inlines-end-001.xht (588x775)
- [ ] css/CSS2/box-display/insert-block-in-blocks-n-inlines-end-002.xht (588x775)
- [ ] css/CSS2/box-display/insert-block-in-blocks-n-inlines-end-003.xht (588x775)
- [ ] css/CSS2/box-display/insert-block-in-blocks-n-inlines-middle-001.xht (588x775)
[...]

PR #9716, opened 2018-02-28
[css-text] Modify reference tests so they fit in the viewport

mrego:
Many reftests in this suite are bigger than the default viewport size
that browsers use when run the tests.
That means that only the first cases are checked, everything that is not
in the visible area won't be compared, so differences there won't be detected.

This patch modifies a bunch of tests to make them fit in the viewport.
Some of the tests are splitted in several files, as they have a lot of content.

Tests under folder text-transform/ still have this kind of issues,
some are actually way bigger, so that would need to be fixed
in a follow-up patch.


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25018)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/347306991)

gsnedders:
This is part of #5403.

kojiishi (approved):
lgtm, thank you for working on this.

PR #9718, opened 2018-02-28
Remove duplicated fonts and put them under /fonts/

mrego:
This patch removes the duplicated fonts we have in the repository
by copying them under /fonts/ and removing the duplicates.
The paths are updated accordingly to point to the new location.

In addition this patch moves the content of /css/fonts/ under /fonts/,
so there is a single common "fonts" folder on the repository.


gsnedders:
I think I'd rather have each font in its own directory rather than putting lots directly in /fonts (so then we can have adjacent README files with something about the font, which most of them could do with!)

gsnedders:
(I fat-fingered the wrong button there)

w3c-bots:
# Build ERRORED

Started: 2018-02-28 18:56:55
Finished: 2018-02-28 19:14:06


<h2>Failing Jobs</h2>
<ul>

<li>firefox:nightly</li>

<li>chrome:dev</li>

<li>safari:11.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/25039)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/347396252)

gsnedders:
(The build failing is unsurprising given number of tests modified.)

mrego:
@gsnedders what could be the structure of `/fonts/` folder then, would it be something like this fine?
```
├── ad
│   └── AD.woff
├── adobe-fonts
│   ├── CSSFWOrientationTest.otf
│   ├── CSSHWOrientationTest.otf
│   ├── LICENSE
│   └── README.md
├── ahem-extra
│   ├── AHEM_*.TTF
├── Ahem.ttf
├── canvas-test
│   ├── CanvasTest.ttf
│   └── CanvasTest.ttf.sub.headers
├── CSSTest
│   ├── csstest-*.ttf
│   ├── LICENSE
│   └── README
├── gentiumplus-r
│   └── GentiumPlus-R.woff
├── math
│   ├── *.woff
├── mplus-1p-regular
│   └── mplus-1p-regular.woff
├── noto
│   ├── NotoSansAdlam-hinted
│   │   ├── LICENSE_OFL.txt
│   │   ├── NotoSansAdlam-Regular.ttf
│   │   └── README
│   ├── NotoSansCypriot-hinted
│   │   ├── LICENSE_OFL.txt
│   │   ├── NotoSansCypriot-Regular.ttf
│   │   └── README
│   └── NotoSansDeseret-Regular.ttf
├── OWNERS
├── README.md
├── revalia
│   └── Revalia.woff
├── sileot-webfont
│   └── sileot-webfont.woff
└── tcu-font
    └── tcu-font.woff
```

I'm asking before doing the actual changes on the patch, just in case you'd prefer something different.
Thanks.

jgraham:
I don't really care how this is organised within `/fonts` but one subdirectory per font doesn't seem like an obvious default. I understand doing it for cases like Noto where we might group lots of similar fonts together.

mrego:
@jgraham yes, that's why I was asking. Folders that have several fonts like `adobe-fonts`, `ahem-extra`, `CSSTest`, `math`, `noto` make sense to me, folders with just one font seems strange.
I understand it might be useful to add some doc about the font, but you can use the generic README for that. The problem is that in this patch I cannot add the documentation on the README files as I don't know it.

mrego:
Ok, so I've just uploaded a new version that only create subfolders under `/fonts/` when we have more than one file.
@gsnedders are you ok with something like this?

gsnedders:
@mrego yeah

gsnedders:
OK, having run this locally against all of /css given Travis can't cope, the following tests changed result:

```
/css/css-text/text-transform/text-transform-upperlower-016.html
  PASS expected FAIL [Parent]
/css/css-text/text-transform/text-transform-upperlower-028.html
  PASS expected FAIL [Parent]
/css/vendor-imports/mozilla/mozilla-central-reftests/masking/mask-image-1c.html
  FAIL [Parent]
/css/css-text/text-transform/text-transform-upperlower-035.html
  PASS expected FAIL [Parent]
/css/css-text/text-transform/text-transform-capitalize-028.html
  PASS expected FAIL [Parent]
/css/css-text/text-transform/text-transform-upperlower-029.html
  PASS expected FAIL [Parent]
/css/css-transitions/transitions-animatable-properties-01.html
  PASS expected NOTRUN font-size intermediate
```

(the expected results are the base revision of the branch)

@mrego can you quickly check there's nothing wrong with those so we can land this?

gsnedders (approved):

gsnedders:
Okay, I've moved CanvasTest back out of a directory (because it's a solo font) and reverted the changes that point to it. This should fix the update_built_tests check failing.

gsnedders (changes requested):
waiting for Travis to finish…

gsnedders (approved):

plinss:
Having all the fonts being referred to using absolute paths is breaking the css test infrastructure as the URLs are no longer within the test built test suite...

fantasai:
Please revert this change. This just broke way too much stuff.

TalbotG:
.

frivoal:
Also, the merge has invalidated all the manual testing results stored in testharness, since the files have changed :(

That doesn't mean we should never do any kind of sweeping change, but since it does cause data loss, we shouldn't do such changes lightly

mrego:
Here is a PR to revert this until for now: https://github.com/w3c/web-platform-tests/pull/9940

I'm sorry for this being breaking the CSS infrastructure, but I'm not sure I understand the reasons.
Using relative paths has also issues with the CSS infrastructure if they are like `../foo/bar.baz`,
however using absolute paths has been working fine for `/resources/testharness.js`
why that cannot be possible for `/fonts/`?

And about the manual testing results, I think that this kind of changes will only be needed once, so we'd need to assume that I guess we have no other choice.
And I guess we all agree that avoiding duplicated files in the repo is a good goal.

And my apologizes again for breaking this but it's not always easy to know what's going on in the CSS build system. :disappointed:

r12a:
Where does this leave us in terms of lodging copies of the CSS fonts under /fonts so that people can use them in that location in future?  I'm hoping that that's still ok, since i recently changed several hundred tests (only a few submitted so far) to look in /fonts.

Note that I already started that process with https://github.com/w3c/web-platform-tests/pull/9818 (and i'm relying on that location for the 500 test PR at https://github.com/w3c/web-platform-tests/pull/9871).

mrego:
I don't really know what we should do here, so I'll let other people with more experience to decide.

BTW, I did this PR after this comment:
https://github.com/w3c/web-platform-tests/pull/9689#issuecomment-369246291
Note that there are currently a bunch of duplicated fonts in WPT, so it seemed a good idea to get rid of the duplicates.

foolip:
On IRC: https://w3.logbot.info/testing/20180309#c384617

foolip:
So, I'm ready to revert https://github.com/w3c/web-platform-tests/pull/9940 but am waiting for confirmation, so that we don't flip back and forth and create more trouble.

plinss:
I don’t believe this needs to be reverted.

There’s a secondary problem that when tests are changed, the results posted to the older version no longer apply, and a lot of manual testing can be lost unnecessarily for changes like these. The CSS test harness has a mechanism where different versions of a test can be declared to equivalent (formerly used to allow certain meta data changes), this could be leveraged to reinstate the old results, but will take some work.

foolip:
@Hexcles and I were talking about what it would take to avoid the need for urgent fixing of this sort going forward, and lints that enforce the constraints seem like the obvious way. @plinss, what's a good place to learn more about how the CSS test harness works? Are there other constraints apart from required spec links and what directories can be included from?

mrego:
I'd be interested in knowing more about the CSS build system to check which are the requirements.
In my mind it seems it'd be nice to extend https://wpt.fyi to eventually replace https://test.csswg.org

PR #9719, opened 2018-02-28
manifest.py: Fixed updating reftest graph when node changes to other type

Cactusmachete:
Slightly altered line 94 so reftest_changes is also set when
reftest_node changes to support, etc.

Closes https://github.com/w3c/web-platform-tests/issues/9671

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9719)

### DIFF
--- a/tools/manifest/manifest.py
+++ b/tools/manifest/manifest.py
@@ -91,7 +91,7 @@ class Manifest(object):
                     hash_changed = True
                 else:
                     new_type, manifest_items = old_type, self._data[old_type][rel_path]
-                if (old_type == "reftest" and new_type != old_type) or (old_type=="reftest_node" and new_type!=old_type):
+                if old_type in ["reftest", "reftest_node"] and new_type != old_type:
                     reftest_changes = True
             else:
                 new_type, manifest_items = source_file.manifest_items()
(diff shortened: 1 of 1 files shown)


tools/manifest/manifest.py
```diff
@@ -91,7 +91,7 @@ def update(self, tree):
                     hash_changed = True
                 else:
                     new_type, manifest_items = old_type, self._data[old_type][rel_path]
-                if old_type == "reftest" and new_type != old_type:
+                if (old_type == "reftest" and new_type != old_type) or (old_type=="reftest_node" and new_type!=old_type):
```
jgraham: I think this can be written `old_type in ["reftest", "reftest_node"] and new_type != old_type`.

jgraham (commented):
I'm pretty sure the logic here is correct, thanks! Could you add a test in tools/manifest/tests/test_manifest.py to check that this doesn't regress?

tools/manifest/manifest.py
Cactusmachete: Sorry, didn't think about that earlier. Made that change now. 
I'm not sure about how to write tests so I'll have to look into that

gsnedders: It'd be more idiomatic to use a tuple (`("reftest", "reftest_node")`) here.

w3c-bots:
# Build PASSED

Started: 2018-02-28 18:15:16
Finished: 2018-02-28 18:31:02





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25035)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/347389260)

gsnedders:
This turned into #9732.

PR #9732, opened 2018-03-01
Update reftest graph when node changes to support

Cactusmachete:
Modified conditional in manifest.py to check for a reftest node changing
to another type.
Wrote a test in test_manifest.py for the same, but this needs some review.

Closes https://github.com/w3c/web-platform-tests/issues/9671

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9732)

### DIFF
--- a/tools/manifest/tests/test_manifest.py
+++ b/tools/manifest/tests/test_manifest.py
@@ -254,6 +254,31 @@ def test_reftest_computation_chain_update_test_type():
     assert list(m) == [("testharness", test2.path, {test2})]
 
 
+def test_reftest_computation_chain_update_node_change():
+    m = manifest.Manifest()
+
+    s1 = SourceFileWithTest("test1", "0"*40, item.RefTest, [("/test2", "==")])
+    s2 = SourceFileWithTest("test2", "0"*40, item.RefTestNode, [("/test3", "==")])
+
+    assert m.update([s1, s2]) is True
+
+    test1 = s1.manifest_items()[1][0]
+    test2 = s2.manifest_items()[1][0]
+
+    assert list(m) == [("reftest", test1.path, {test1}),
+                       ("reftest_node", test2.path, {test2})]
+
+    #test2 changes to support type
+    s3 = SourceFileWithTest("test2", "1"*40, item.SupportFile)
+
+    # This doesn't check whether the reftest graph was computed or not,
+    # the m.update([s3]) assert will just check whether m.update() returned True (I think).
+    # Regardless, the test works even when the graph isn't being recomputed.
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PASSED

Started: 2018-03-01 12:11:15
Finished: 2018-03-01 12:28:19





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25093)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/347733830)

tools/manifest/tests/test_manifest.py
```diff
@@ -254,6 +254,31 @@ def test_reftest_computation_chain_update_test_type():
...
+    assert list(m) == [("reftest", test1.path, {test1}),
+                       ("reftest_node", test2.path, {test2})]
+
+    #test2 changes to support type
+    s3 = SourceFileWithTest("test2", "1"*40, item.SupportFile)
+
+    # This doesn't check whether the reftest graph was computed or not,
+    # the m.update([s3]) assert will just check whether m.update() returned True (I think).
+    # Regardless, the test works even when the graph isn't being recomputed.
+    assert m.update([s3]) is True
```
jgraham: Then here I think you want to call `m.update([s1, s2])` again.

tools/manifest/tests/test_manifest.py
```diff
@@ -254,6 +254,31 @@ def test_reftest_computation_chain_update_test_type():
...
+    assert m.update([s1, s2]) is True
+
+    test1 = s1.manifest_items()[1][0]
+    test2 = s2.manifest_items()[1][0]
+
+    assert list(m) == [("reftest", test1.path, {test1}),
+                       ("reftest_node", test2.path, {test2})]
+
+    #test2 changes to support type
+    s3 = SourceFileWithTest("test2", "1"*40, item.SupportFile)
```
jgraham: So I think you want to redefine `s2` here rather than creating a new `s3`

jgraham (changes requested):
I think the way you've written the test effetively creates an entirely new manifest that's not related to the previous one.

Note that in general one should push additional commits to the existing PR (by pushing to the same branch) rather than creating a new PR with changes.

tools/manifest/tests/test_manifest.py
Cactusmachete: I made the changes and it works as it should now!
Still need to understand why, though. 
Thanks a lot :smile: 

And I'm sorry I made a new PR for this, I'll ensure I push to the same branch from now on.

jgraham (approved):
Thanks!

PR #9748, opened 2018-03-02
Firefox browser install now uses mozdownload and mozinstall

Cactusmachete:
mozinstall will refuse to install if firefox is already installed
(when checked on OSX). I've currently only set it to print an error message
when this happens - it'll require the user to uninstall firefox in _venv and run
the command again as uninstall in mozinstall doesn't seem to work very well.

Closes https://github.com/w3c/web-platform-tests/issues/7996

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9748)


tools/wpt/browser.py
```diff
@@ -105,36 +107,23 @@ def platform_string_geckodriver(self):
...
+
         if dest is None:
             dest = os.getcwd()
 
-        nightly_link = self.get_nightly_link(self.latest_nightly_listing(),
-                                             self.platform_string())
-        resp = get(nightly_link)
-        resp.raise_for_status()
-        untar(resp.raw, dest=dest)
+        filename = FactoryScraper('daily', branch='mozilla-central').download()
```
jgraham: I think we should set a destination here so that it ends up inside the venv directory.

tools/wpt/browser.py
```diff
@@ -105,36 +107,23 @@ def platform_string_geckodriver(self):
...
+
+        try:
+            #When tested in OSX, mozinstall will refuse to install if firefox is already installed
+            mozinstall.install(filename, dest)
+        except Exception as e:
```
jgraham: It would be nice to catch a more specific exception type here.

tools/wpt/browser.py
```diff
@@ -105,36 +107,23 @@ def platform_string_geckodriver(self):
...
+            # Instead of printing an error message we should uninstall the currently
```
jgraham: Can we just figure out what to delete based on the path, or does something more complex happen on OSX here?

jgraham (changes requested):
This is a very exciting change; I love all the deleted code :) I think a little cleanup is required to land it though.

w3c-bots:
# Build BROKEN

Started: 2018-03-02 18:14:44
Finished: 2018-03-02 18:21:50


<h2>Failing Jobs</h2>
<ul>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/25151)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/348337460)

tools/wpt/browser.py
Cactusmachete: Messed around a bit, turns out it's a simple fix. I still haven't tested this on windows though, which might require a different uninstall path. Will look into that in a few hours.

tools/wptrunner/requirements_firefox.txt
```diff
@@ -4,3 +4,5 @@ mozprocess == 0.26
 mozcrash == 1.0
 mozrunner == 6.14
 mozleak == 0.1
+mozinstall == 1.15
```
jgraham: This should't be both in here and in the commands.json file; here is better but you might need to move the `moz*` imports you added into ttthe function where they're actually used.

tools/wpt/browser.py
```diff
@@ -105,36 +107,30 @@ def platform_string_geckodriver(self):
...
+        try:
+            mozinstall.install(filename, dest)
+        except mozinstall.mozinstall.InstallError as e:
+            platform = {
+                "Linux": "linux",
+                "Windows": "win",
+                "Darwin": "mac"
+            }.get(uname[0])
+
+            if platform == "mac":
```
jgraham: `if uname[0] == "Darwin"` seems simpler if we really only need this on one platform. But it's unclear why macOS is special cased here. Does it really work differently on mac? If it does a comment explaining what's going on seems helpful, otherwise we should support all for Windows/Linux/Mac here

jgraham (changes requested):

tools/wpt/browser.py
```diff
@@ -105,36 +107,30 @@ def platform_string_geckodriver(self):
...
+                mozinstall.uninstall(dest+'/Firefox Nightly.app')
+                mozinstall.install(filename, dest)
+            else:
+                raise
+
+        os.remove(filename)
         return find_executable("firefox", os.path.join(dest, "firefox"))
```
jgraham: Should we consider `mozinstall.get_binary` here?

jgraham (changes requested):

tools/wptrunner/requirements_firefox.txt
Cactusmachete: If I remove these requirements from the commands.json file, the code crashes when I try to set up a new venv because mozdownload/install don't get installed there in it.
When I remove them from this file, it could cause the build to fail in Travis if anything calls browse.py because they don't get installed there.

tools/wpt/browser.py
Cactusmachete: Alright!
I checked it on all three OSes. mozinstall only throws an error on Mac. 
Afaict this is because mozdownload gets a  .dmg on Mac, but zipped folders in Lin/Win which it simply unzips there.

tools/wpt/browser.py
Cactusmachete: We could, but I have no way of testing this on a Mac anymore.

tools/wptrunner/requirements_firefox.txt
jgraham: So I think this is possible, but it involves ensuring that we install the product-specific requriements later on. Let's do that in a followup.

jgraham (approved):

jgraham (approved):

csnardi:
It looks like this broke CI, e.g. https://travis-ci.org/w3c/web-platform-tests/jobs/349449358.

PR #9776, opened 2018-03-02
Update the cssom IDL file

lukebjerring:

### DIFF
--- a/interfaces/cssom.idl
+++ b/interfaces/cssom.idl
@@ -128,18 +128,6 @@ interface CSSStyleDeclaration {
   [CEReactions] attribute [TreatNullAs=EmptyString] CSSOMString cssFloat;
 };
 
-partial interface CSSStyleDeclaration {
-  [CEReactions] attribute [TreatNullAs=EmptyString] CSSOMString _camel_cased_attribute;
-};
-
-partial interface CSSStyleDeclaration {
-  [CEReactions] attribute [TreatNullAs=EmptyString] CSSOMString _webkit_cased_attribute;
-};
-
-partial interface CSSStyleDeclaration {
-  [CEReactions] attribute [TreatNullAs=EmptyString] CSSOMString _dashed_attribute;
-};
-
 [Exposed=Window,
  NoInterfaceObject]
 interface ElementCSSInlineStyle {
--- a/interfaces/cssom.idl
+++ b/interfaces/cssom.idl
@@ -47,6 +47,7 @@ interface LinkStyle {
 };
 
 ProcessingInstruction implements LinkStyle;
+
 [Exposed=Window,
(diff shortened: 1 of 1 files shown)


interfaces/cssom.idl
```diff
@@ -123,6 +128,18 @@ interface CSSStyleDeclaration {
   [CEReactions] attribute [TreatNullAs=EmptyString] CSSOMString cssFloat;
 };
 
+partial interface CSSStyleDeclaration {
+  [CEReactions] attribute [TreatNullAs=EmptyString] CSSOMString _camel_cased_attribute;
```
gsnedders: This is wrong, given this is actually some magic variable thing.

interfaces/cssom.idl
```diff
@@ -123,6 +128,18 @@ interface CSSStyleDeclaration {
...
+};
+
+partial interface CSSStyleDeclaration {
+  [CEReactions] attribute [TreatNullAs=EmptyString] CSSOMString _webkit_cased_attribute;
```
gsnedders: This is wrong, given this is actually some magic variable thing.

interfaces/cssom.idl
```diff
@@ -123,6 +128,18 @@ interface CSSStyleDeclaration {
...
+};
+
+partial interface CSSStyleDeclaration {
+  [CEReactions] attribute [TreatNullAs=EmptyString] CSSOMString _dashed_attribute;
```
gsnedders: This is wrong, given this is actually some magic variable thing.

gsnedders (dismissed):

interfaces/cssom.idl
lukebjerring: @foolip - These aren't really part of the spec; do we have (or, need) a mechanism for excluding example/descriptive snippets of IDL from being scraped?

gsnedders: I think this might be a CSSOM-only oddity? Most other places the web platform does such things we rely on named getters, hence it might not be necessary to have a general solution to this problem?

foolip: IDL in examples is already excluded, but this is a macro of sorts. I guess what we should do is to generate the IDL that it expands to just like how implementations should and Gecko in fact does.

But just excluding this snippet with some string matching hack should be a good enough start.

lukebjerring: @mdittmer - I'll delegate to the scraper tool to ignore these chunks.

mdittmer: @gsnedders, if this isn't *really* IDL, would it be possible to mark it with the `note`, or even `example` CSS class?

lukebjerring:
@gsnedders can I get an LGTM

foolip (approved):

PR #9778, opened 2018-03-02
Update the orientation-event IDL file

lukebjerring:
Hello reviewer(s),

Please check that:


lukebjerring:
@foolip - stranglely, the original test `idlharness.html` had 2 failures, as seen here:
http://w3c-test.org/orientation-event/idlharness.html

However, after migrating to `idlharness.window.js` and `add_dependency_idls`, those 2 tests now pass..
I'm confused why the prototype object would suddenly have those attributes. The order of execution has changed too.  ¯\_(ツ)_/¯

orientation-event/idlharness.window.js
```diff
@@ -0,0 +1,25 @@
...
+  const idl = await fetch('/interfaces/orientation-event.idl').then(r => r.text());
+  const dom = await fetch('/interfaces/dom.idl').then(r => r.text());
+  const html = await fetch('/interfaces/html.idl').then(r => r.text());
+  const uievents = await fetch('/interfaces/uievents.idl').then(r => r.text());
+
+  var idl_array = new IdlArray();
+  idl_array.add_idls(idl);
+  idl_array.add_dependency_idls(html);
+  idl_array.add_dependency_idls(dom);
+  idl_array.add_dependency_idls(uievents);
```
reillyeon: Is this dependency necessary?

reillyeon (approved):

orientation-event/idlharness.window.js
lukebjerring: I thought it was earlier, but seems happy without it now - removed.

PR #9779, opened 2018-03-02
Update the dom IDL file

lukebjerring:
Hello reviewer(s),

Please check that:

### DIFF
--- a/interfaces/dom.idl
+++ b/interfaces/dom.idl
@@ -145,7 +145,7 @@ Text includes Slotable;
 interface NodeList {
   getter Node? item(unsigned long index);
   readonly attribute unsigned long length;
-  iterable<Node>;
+  // iterable<Node>;
 };
 
 [Exposed=Window, LegacyUnenumerableNamedProperties]
@@ -548,5 +548,5 @@ interface DOMTokenList {
   [CEReactions] boolean replace(DOMString token, DOMString newToken);
   boolean supports(DOMString token);
   [CEReactions] stringifier attribute DOMString value;
-  iterable<DOMString>;
+  // iterable<DOMString>;
 };
--- a/interfaces/dom.idl
+++ b/interfaces/dom.idl
@@ -1,9 +1,14 @@
+// GENERATED CONTENT - DO NOT EDIT
+// Content of this file was automatically extracted from the
+// "DOM Standard" spec.
+// See: https://dom.spec.whatwg.org/
+
 [Constructor(DOMString type, optional EventInit eventInitDict),
- Exposed=(Window,Worker)]
+ Exposed=(Window,Worker,AudioWorklet)]
 interface Event {
   readonly attribute DOMString type;
   readonly attribute EventTarget? target;
-  readonly attribute EventTarget? srcElement;
+  readonly attribute EventTarget? srcElement; // historical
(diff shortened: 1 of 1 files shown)


interfaces/dom.idl
```diff
@@ -108,50 +109,45 @@ Document includes ParentNode;
...
 Element includes Slotable;
 Text includes Slotable;
 
-
 [Exposed=Window]
 interface NodeList {
   getter Node? item(unsigned long index);
   readonly attribute unsigned long length;
-//  iterable<Node>;
+  iterable<Node>;
```
foolip: This and the DOMTokenList change causes new failing tests due to this:
https://github.com/w3c/web-platform-tests/blob/9192f4b6c633f7f00ad0c1b983cbed33fff6be82/resources/idlharness.js#L2047

It fails for entries(), keys() and values() for each of NodeList and DOMTokenList, so 6 failing tests.

The per-spec behavior is defined by https://heycam.github.io/webidl/#es-iterable and since this is a "value iterator" these assertions could be made:
* `Array.prototype.entries===NodeList.prototype.entries`
* `Array.prototype.keys===NodeList.prototype.keys`
* `Array.prototype.values===NodeList.prototype.values`

That does in fact hold true in Chrome and Safari (maybe more) and explains why `NodeList.prototype.entries.call({})` should't throw TypeError but actually return something.

So, to merge this we need to at least make idlharness.js not assert that this should throw.

foolip:
OK, sent https://github.com/w3c/web-platform-tests/pull/9790 to uncomment the iterable bits. (In the same PR to demonstrate that it works.)

interfaces/dom.idl
lukebjerring: Sounds reasonable - Can you write a PR that makes said change?

foolip: I wrote https://github.com/w3c/web-platform-tests/pull/9790 back then, but then wasn't sure it worked at all. Needs tests I guess.

lukebjerring: As you described, [Chromium](https://test-results.appspot.com/data/layout_results/linux_chromium_rel_ng/150187/site_per_process_webkit_layout_tests%20%28with%20patch%29/layout-test-results/results.html) shows a bunch of regressions due to the issue. Are you able to rebase/get #9790 up to date?

annevk (commented):
This looks great, modulo @foolip's comments. It's a little weird to see all the newlines disappear though. Is that a result of the scraping from the specification?

lukebjerring:
As discussed offline with @foolip, I'll just re-comment iterable<Node> for now and we'll follow up when #9790 is resolved :)

interfaces/dom.idl
```diff
@@ -1,9 +1,14 @@
+// GENERATED CONTENT - DO NOT EDIT
```
annevk: So I should no longer ask for updates to this file whenever we update DOM?

annevk (approved):

annevk:
(Please use squash & merge btw.)

interfaces/dom.idl
lukebjerring: Yep, but they're not yet regular (semi-manual) so feel free to ping me or @foolip if you'd like a more hasty update.

PR #9780, opened 2018-03-02
Update the encoding IDL file

lukebjerring:

### DIFF
--- a/encoding/idlharness.html
+++ b/encoding/idlharness.html
@@ -11,56 +11,15 @@
 <h1>idlharness test</h1>
 <p>This test validates the WebIDL included in the Encoding Living Standard.</p>
 
-<script type="text/plain" class="untested-idl">
-interface Window {};
-</script>
-
-<script type="text/plain" class="idl">
-// 8.1 Interface TextDecoder
-
-dictionary TextDecoderOptions {
-  boolean fatal = false;
-  boolean ignoreBOM = false;
-};
-
-dictionary TextDecodeOptions {
-  boolean stream = false;
-};
-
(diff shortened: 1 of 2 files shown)


foolip (approved):
Confirmed no interesting differences. Please review the idlharness.html change I've pushed.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25428)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/350164395)

inexorabletash:
Why isn't [encoding/idlharness.html](https://github.com/w3c/web-platform-tests/blob/master/encoding/idlharness.html) updated to use this instead of inlining the IDL?

Or phrased differently: can you update the test to reference this, or explain why that's not appropriate here?

lukebjerring:
You're right, it should, and Philip did, but the script blundered over the changes. @foolip can you dig them up and re-apply them. Sorry!

foolip:
I found it in my reflog and pushed it again. @inexorabletash, can you review the change as a whole?

inexorabletash (approved):
Yay, lgtm.

We should make it an 'any.js' test and should probably scrub all the idlharness tests to figure out which actually need "untested" stuff but as long as this passes (it does, right?) then it's a good improvement - thanks!

foolip:
> as long as this passes (it does, right?)

Figuring that out is unfortunately not straightforward. From https://pulls.web-platform-tests.org/build/25890 it looks like no tests were run, but that's wrong, from the Travis logs we can see that they were. Filed https://github.com/w3c/wpt-pullresults/issues/62. And yes, the tests do work still, as far as eyeballing differences goes.

PR #9785, opened 2018-03-02
Update the gamepad IDL file

lukebjerring:


foolip:
Turns out this file is in sync with the spec, and that these are only indentation differences. https://github.com/w3c/gamepad/blob/gh-pages/index.html has that indentation, but it doesn't affect https://w3c.github.io/gamepad/#gamepadbutton-interface.

I don't think we should pretty print (parse and serialize) the IDL if it can be avoided. A pretty good heuristic might be: If the first line of each IDL fragment starts with N spaces, and every subsequent line has no non-whitespace in the first N characters, then remove N spaces from the beginning of all lines.

foolip (approved):

PR #9799, opened 2018-03-03
Add first batch of tests for content change selection update

emanchado:
This is a first batch of tests for selection updates triggered by content changes. It doesn't cover all cases mentioned in https://github.com/whatwg/html/issues/2424, but covers some of the most common.

Questions:
1. Should we test setting the dirty flag in any way that isn't setting the value?
1. How to simulate users typing?
1. How to simulate browser autofill, clipboard operations, and others mentioned in https://github.com/whatwg/html/issues/2424#issuecomment-286605947?

### DIFF
--- a/html/semantics/forms/textfieldselection/selection-start-end-extra.html
+++ b/html/semantics/forms/textfieldselection/selection-start-end-extra.html
@@ -1,0 +1,123 @@
+<!doctype html>
+<meta charset=utf-8>
+<title></title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<div id=log></div>
+<form id="form"><input id="form-input" type="text" value="abc" /></form>
+<script>
+  // * Should we test setting the dirty flag in any way that isn't
+  //   setting the value?
+  // * How to simulate users typing?
+
+  test(function() {
+    var el = document.createElement("textarea");
+    assert_equals(el.selectionStart, 0);
+    assert_equals(el.selectionEnd, 0);
+    el.defaultValue = "123";
+    assert_equals(el.value.length, 3);
+    assert_equals(el.selectionStart, 3);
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25219)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/348688977)

domenic (approved):
These look amazing to me. Some specific feedback, which we can do in a followup:

- Probably should test selectionDirection too? I introduced some relevant helpers in https://github.com/emanchado/web-platform-tests/commit/9c585b4a03af8e137a451c34f796cd2139879518#diff-efe3ff105e161015b02223f18dc2399e.
- "How to simulate users typing?" This won't be possible, but a manual test would be a good idea anyway. For examples, search for files with "-manual" in their name.
- Similarly, the other cases won't really be possible without manual tests, at least until we get better web driver-style automation.

I'll go ahead and merge these since they're a great start.

domenic:
Oh, and I wasn't sure exactly what you meant by this:

> Should we test setting the dirty flag in any way that isn't setting the value?

Did you mean the dirty value flag? If so that isn't terribly related here, right?

emanchado:
Ah, I didn't think of selectionDirection. Thanks for the pointer, somehow I assumed every test was automated.

Re: the dirty flag, I meant because of your comment https://github.com/whatwg/html/issues/2424#issuecomment-286580627 ("depending on the dirty value flag"). If it doesn't always behave the same way, it would be nice to test both cases, no?

domenic:
Ah, right. No, I think since we're only testing changing the control's "relevant value", we can just use one mechanism of entangling relevant value and defaultValue/textContent. In other words, we can assume the UA implements the dirty value flag correctly, and just test its consequences that are relevant to this sphere.

PR #9804, opened 2018-03-05
Add some basic autofocus tests

domenic:
Follows https://github.com/whatwg/html/pull/3488, although certainly doesn't test the underlying issue exhaustively.

Edge, Firefox, and Chrome pass all these. Haven't tested Safari.

### DIFF
--- a/html/semantics/forms/autofocus/first-when-later.html
+++ b/html/semantics/forms/autofocus/first-when-later.html
@@ -1,0 +1,25 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>The first autofocus in the document wins, even if elements are inserted later</title>
+<link rel="help" href="https://html.spec.whatwg.org/multipage/#autofocusing-a-form-control:-the-autofocus-attribute">
+<link rel="author" title="Domenic Denicola" href="d@domenic.me">
+
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<input autofocus>
+
+<script>
+"use strict";
+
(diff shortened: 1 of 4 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25238)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/349181358)

html/semantics/forms/autofocus/first-when-later.html
```diff
@@ -0,0 +1,25 @@
...
+<script src="/resources/testharnessreport.js"></script>
+
+<input autofocus>
+
+<script>
+"use strict";
+
+const input1 = document.querySelector("input");
+const input2 = document.createElement("input");
+input2.autofocus = true;
```
annevk: You're not inserting the element. It still get focus?

annevk (commented):
first-when-later.html looks incorrect, otherwise LGTM.

html/semantics/forms/autofocus/first-when-later.html
domenic: Well, the test is that input1 keeps the focus, so the test still passes. But yeah, this is a bug that makes the pass trivial. Fixing.

annevk (approved):
It'd be a little more compelling I think if you inserted it before the other input element, but this seems fine as initial tests.

domenic:
Oh, good idea, Chrome fails that. I'll add it as a separate file and file a bug.

annevk (approved):

PR #9805, opened 2018-03-05
[css-text] Add lang=ja in css3-text-line-break-jazh tests

mrego:
Without the lang attribute, the reference was choosing
a different character for English vs Japanese in some configurations.

The patch just adds "lang='ja'" to both the test and reference elements.

### DIFF
--- a/css/css-text/i18n/css3-text-line-break-jazh-059.html
+++ b/css/css-text/i18n/css3-text-line-break-jazh-059.html
@@ -29,7 +29,7 @@
 
 
 
-<div class='ref'>中中<br/>中&#xFFE0;文</div></div>
+<div class='ref' lang='ja'>中中<br/>中&#xFFE0;文</div></div>
 
 
 <!--Notes:
--- a/css/css-text/i18n/css3-text-line-break-jazh-159.html
+++ b/css/css-text/i18n/css3-text-line-break-jazh-159.html
@@ -29,7 +29,7 @@
 
 
 
-<div class='ref'>中中中<br/>&#xFFE0;文</div></div>
+<div class='ref' lang='ja'>中中中<br/>&#xFFE0;文</div></div>
 
 
 <!--Notes:
--- a/css/css-text/i18n/css3-text-line-break-jazh-259.html
+++ b/css/css-text/i18n/css3-text-line-break-jazh-259.html
@@ -29,7 +29,7 @@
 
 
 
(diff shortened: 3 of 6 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25239)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/349250679)

gsnedders (approved):

frivoal:
It seems a bit awkward to me that the phrase "中文" which means Chinese in Chinese is tagged as being Japanese, but I guess that was that way in the original test. Since there could be differences between Chinese and Japanese line breaking, I do not suggest changing the language tag (at least not without checking with the author of the test for unintended side effects). The test is valid as merged, but maybe we could replace 中中中文 by 日日本語 or something like that. 

@r12a, any thoughts?

r12a:
Fwiw, i just spent the past four days (including weekend) porting the i18n test suite to github at https://github.com/w3c/i18n-tests/.  I improved a bunch of tests in addition to moving them.  In particular, i applied the same fix to the tests mentioned here.  It may be that there are other tests in this jazh series that should have had the lang attribute added. There are separate tests for Japanese and Chinese behaviours.

@frivoal the text  中中中￠文 isn't really chinese either ;-). I chose ideographs that appear in both chinese and japanese, but didn't really try to have end-to-end meaningful strings. I also chose those han characters, iirc, because they are quite simple (and therefore easier to detect for people unlike us who know nothing of Japanese/Chinese writing).

If you think we should change, we could perhaps use something like 文字, which is both chinese and japanese.  In that case, i or someone should resubmit the whole set of tests.

r12a:
btw, another change i made to (all) those tests was to change

> src: url('support/mplus-1p-regular.woff') format('woff');

to

> src: url('../../fonts/CSSTest/mplus-1p-regular.woff') format('woff');

Compare https://github.com/mrego/web-platform-tests/blob/98b733787a32327667a1a30b7247416186aff69b/css/css-text/i18n/css3-text-line-break-jazh-059.html and https://github.com/w3c/i18n-tests/blob/gh-pages/css-text/line-breaks/css3-text-line-break-jazh-059.html

r12a:
I just submitted the PR just above, which should make this one redundant.

PR #9814, opened 2018-03-05
Update the presentation-api IDL file

lukebjerring:
Hello reviewer(s),

Please check that:


w3c-bots:
# Build PASSED

Started: 2018-03-05 16:19:33
Finished: 2018-03-05 16:31:50


<h2>Failing Jobs</h2>
<ul>

<li>chrome:dev</li>

<li>safari:11.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/25252)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/349346957)

tomoyukilabs:
@tidoust @louaybassbouss Could you review this PR? Note that:

- The purpose of this PR seems to be mentioned in https://github.com/web-platform-tests/wpt/pull/9908#issuecomment-374449583.
- The IDL file in this PR contains all interfaces for both controlling and receiving UAs. However, the current Presentation API test suite contains the separated IDL tests.

lukebjerring:
@tomoyukilabs - thanks for pointing out those IDL tests. Updated them to use the imported IDL, and fixed up the deps too. PTAL.

foolip (approved):

lukebjerring:
@tidoust @louaybassbouss - are you happy for me to land this?

tidoust:
There's one thing specific about the Presentation API that @tomoyukilabs pointed out and that seems worth emphasizing: the Presentation API defines two conformance classes (controlling UA and receiving UA), and the IDL that needs to be supported is different for each of them.

The `presentation-api.idl` file contains the entire IDL defined in the specification. Some of it applies to controlling UAs, some of it to receiving UAs. The [conformance section](https://w3c.github.io/presentation-api/#conformance-classes) lists the interfaces that need to be supported by each class.

The `presentation-api.idl` file created by this PR will check controlling UAs. Depending on the implementation, the IDL harness may report failures for the `PresentationReceiver` and `PresentationConnectionList` interfaces, which do not need to be supported in controlling UAs (tests will pass in Chrome because it supports both classes, but that's not a requirement). These failures would be false positives.

To test the IDL in a receiving UA, one needs to run the IDL harness within a receiving context. The corresponding (manual) test is in: [receiving-ua/idlharness-manual.https.html](https://github.com/web-platform-tests/wpt/blob/master/presentation-api/receiving-ua/idlharness-manual.https.html). That file and its dependencies are not touched by this PR, and that's a good thing for now. But that also means the receiving IDL is still being maintained manually (in [receiving-ua/support/idlharness_receiving-ua.html](https://github.com/web-platform-tests/wpt/blob/master/presentation-api/receiving-ua/support/idlharness_receiving-ua.html)).

All in all, I believe we can merge as-is, although there are two things that we might want to do (either now or later on):
1. Make the receiving test mentioned above use the `presentation-api.idl` file to avoid having to maintain the IDL manually. If we do that, the test will also report false positives, because receiving UAs do not have to implement all interfaces of controlling UAs.
2. Implement custom IDL extraction rules for the Presentation API to create a `presentation-api-controlling-ua.idl` file and a `presentation-api-receiving-ua.idl` file. This way, implementers wouldn't have to filter out individual IDL failures to ignore those that do not apply to them, they would simply be able to ignore the IDL file as a whole. Now, the problem with custom rules is that the list of interfaces that compose each file would need to be hardcoded and maintained somewhere (well, these lists could perhaps be generated from the conformance section).

@lukebjerring I'm not sure what would make implementers' life easier. Would having separated IDL files be preferable, or is it fine to maintain a list of false positives?

lukebjerring:
We have a function param which supports filtering only some of the IDL (blacklist or whitelist both supported). Happy to set the test up that way.

How would I tell what's what? By looking at the inlined IDL before my changes?

PR #9809, opened 2018-03-05
Update the performance-timeline IDL file

lukebjerring:

PR #9837, opened 2018-03-05
Update the user-timing IDL file

lukebjerring:


interfaces/user-timing.idl
```diff
@@ -0,0 +1,17 @@
+
+        partial interface Performance {
```
foolip: Can you update to strip the indentation and use the file in https://github.com/w3c/web-platform-tests/blob/master/user-timing/idlharness.html?

foolip:
Hmm, and why is Travis unhappy?

interfaces/user-timing.idl
foolip: @lukebjerring, indentation problem still persists.

foolip:
@lukebjerring looks like this needs a test to exercise the IDL?

lukebjerring:
I've updated the `user-timing/idlharness.html` test to be `.any.js`, since it's exposed on both worker + window, and changed the test to use the IDL. PTAL.

lukebjerring:
cc @npm1 who was helping me work out whether using `.any.js` was the right thing to do. Note that we see a large number of `.any.worker.html` results with:

> assert_own_property: self does not have own property "PerformanceMark" expected property "PerformanceMark" missing

Goes a bit beyond my understanding of the implementation of idlharness. @foolip - any idea whether that's to be expected?

user-timing/idlharness.any.js
```diff
@@ -0,0 +1,21 @@
+// META: global=window,worker
+// META: script=/resources/WebIDLParser.js
+// META: script=/resources/idlharness.js
+
+'use strict';
+
+promise_test(async () => {
+  const idl_array = new IdlArray();
+  const idl = await fetch("/interfaces/user-timing.idl").then(r => r.text());
+  const hrtime = await fetch("/interfaces/hr-time.idl").then(r => r.text());
```
plehegar: you need to import /interfaces/performance-timeline.idl as well, since PerformanceMark depends on PerformanceEntry.

lukebjerring: Alright, I'll pull in https://github.com/web-platform-tests/wpt/pull/9809 to this PR.

foolip (approved):
Tested locally, .worker.html tests will reveal some things that don't match the spec in Chromium :) :+1:

PR #9851, opened 2018-03-05
Update the webdriver IDL file

lukebjerring:


w3c-bots:
# Build BROKEN

Started: 2018-03-05 23:53:47
Finished: 2018-03-05 23:59:00


<h2>Failing Jobs</h2>
<ul>

<li>lint</li>

<li>firefox:nightly</li>

<li>chrome:dev</li>

<li>safari:11.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/25312)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/349450545)

foolip:
https://github.com/w3c/web-platform-tests/blob/master/webdriver/tests/interface.html needs to be updated.

foolip (commented):
IDL LGTM, but needs to be used in a test.

webdriver/tests/interface.html
```diff
@@ -1,49 +1,23 @@
...
-"use strict";
-
-test(() => assert_idl_attribute(navigator, "webdriver"), "navigator.webdriver is present");
-
-// When test is run in automation navigator.webdriver is likely to
-// be true because WebDriver controls the browser instance.  To that
-// extent, this test is a bit special.  It should also be possible to
-// run the test manually, when WebDriver is not active, and so either
-// true/false outcome is OK.
-if (navigator.webdriver) {
```
lukebjerring: Note that this test is basically "if true, assert true, else assert false" which will only fail when `navigator.webdriver` is undefined - which is a test that is already covered by the `idlarray.add_objects({Navigator: ["navigator"]})` added below.
Hence, since it's confusing, I deleted it.

cc @kereliuk

webdriver/tests/interface.html
```diff
@@ -1,49 +1,23 @@
...
-
-interface mixin NavigatorAutomationInformation {
-  readonly attribute boolean webdriver;
-};
-</script>
-
 <script>
-"use strict";
-
-test(() => assert_idl_attribute(navigator, "webdriver"), "navigator.webdriver is present");
```
lukebjerring: `assert_idl_attribute` is already covered by `idlarray.test()` below - hence, redundant, and deleted.

lukebjerring:
Now used in a test, PTAL.

webdriver/tests/interface.html
```diff
@@ -1,49 +1,23 @@
...
-  if (node.className == "untested") {
-    idls.add_untested_idls(node.textContent);
-  } else {
-    idls.add_idls(node.textContent);
-  }
-};
-idls.test();
+  "use strict";
+
+  promise_test(async () => {
```
andreastt: I get this when running the test locally:

> JavaScript error: file:///home/ato/src/wpt/webdriver/tests/interface.html, line 12: ReferenceError: promise_test is not defined

andreastt (changes requested):
This seems to streamline things a bit, and appears to be how other IDLs are tested.

However, I’d like to get the test running before I proceed with a more detailed review. See more comments inline.

webdriver/tests/interface.html
foolip: Looks like you're trying to run the test from file://. Please run `wpt run` and use http://web-platform.test:8000/webdriver/tests/interface.html instead.

andreastt: That’s not what I want to test. I explicitly want the test to pass when loading it manually with WebDriver disabled.

foolip: Oops, I meant `wpt serve`. testharness.js tests can't be loaded over file:// is the important point.

lukebjerring: As foolip said, from the `wpt` root, run `./wpt serve` then head to http://localhost:8000/webdriver/tests/interface.html

andreastt: Thanks, I can confirm it passes as expected both with and without Marionette enabled in Firefox.

andreastt (approved):

andreastt:
Please note that a better commit message would be appreciated next time.

PR #9854, opened 2018-03-05
Update the webusb IDL file

lukebjerring:


w3c-bots:
# Build BROKEN

Started: 2018-03-06 00:10:31
Finished: 2018-03-06 00:15:25


<h2>Failing Jobs</h2>
<ul>

<li>firefox:nightly</li>

<li>chrome:dev</li>

<li>safari:11.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/25315)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/349450678)

interfaces/webusb.idl
```diff
@@ -174,3 +175,21 @@ interface USBEndpoint {
...
+  required octet vendorId;
+  required octet productId;
+  DOMString serialNumber;
+};
+
+dictionary USBPermissionStorage {
+  sequence<AllowedUSBDevice> allowedDevices = [];
+};
+
+interface USBPermissionResult : PermissionStatus {
```
foolip: I was expecting this to break the test, because PermissionStatus isn't known, but I think what I've seen is that if both interfaces *are* implemented, then one will reach some code path in idlharness.js where if we don't have any information about the right-hand side, the failure is pretty mysterious.

So... I added what I think will be needed for anyone who does implement these bits.

interfaces/webusb.idl
```diff
@@ -1,5 +1,6 @@
-// Copied from https://wicg.github.io/webusb/#idl-index minus the
```
foolip: @reillyeon, you wrote this bit. What do you think about our auto-updating scheme here? The resulting tests will fail, is that OK? Should it be removed from the spec until it's ready to test and implement, perhaps?

foolip (commented):
Question for Reilly.

interfaces/webusb.idl
reillyeon: I think the spec should be annotated to mark that section as unstable and if we have an automatic update mechanism it should be possible to mark IDL sections as "unstable and untested" so they aren't included.

foolip: That would work for us I think. @tabatkins, you have some output from Bikeshed that's used to scrape definitions and things, what you suggest as the "API" for a spec saying that pieces are not to be implemented/tested yet?

reillyeon (changes requested):
It looks like this change is pulling IDL definitions from https://github.com/WICG/webusb/blob/gh-pages/index.html rather than https://github.com/WICG/webusb/blob/master/index.bs. The former is out of date because a recent Bikeshed change broke the build.

foolip:
@lukebjerring is this now getting the IDL from reffy or webidl-diff?

lukebjerring:
webidl-diff. Will update the specs set to redirect.

lukebjerring:
Uh, scratch that. Its URL is https://wicg.github.io/webusb/

@reillyeon - I might need to wait until you can push the `.bs` changes?

foolip:
Does this PR match what is in https://wicg.github.io/webusb/ now? If so, is there still a problem? I guess https://github.com/WICG/webusb/pull/135 is to fix the build error, but have there been IDL changes since the last successful build that this PR is reverting?

reillyeon:
@foolip @odejesush is working on fixing WICG/webusb#135. This patch is still reverting changes that are in [index.bs](https://github.com/WICG/webusb/blob/master/index.bs) but have not been published to https://wicg.github.io/webusb/.

foolip:
Got it, thanks!

reillyeon:
The issue above has been fixed so the next update based on the published spec should contain the correct definitions.

interfaces/webusb.idl
lukebjerring: @tabatkins - Ping. Is there an annotation for ignoring an IDL chunk?

reillyeon (approved):

reillyeon:
This change has broken these tests in Chromium: https://test-results.appspot.com/data/layout_results/linux-xenial-rel/669/site_per_process_webkit_layout_tests/layout-test-results/results.html

When I looked at this change earlier it seemed that the permission descriptor bits had been filtered out but it looks like the version that was merged still contains them.

Please revert this change (or remove the untestable definitions) as soon as possible.

interfaces/webusb.idl
tabatkins: Yes, if it's in a non-normative section (`.note`, `.example`, `.non-normative` or `.informative` on an ancestor) it gets ignored for definition purposes in Bikeshed, and if it's non-normative or has `.extract` on the `pre` (I dunno where this practice came from but it's common on the WHATWG side) it gets omitted from the IDL index at the end of the spec.

reillyeon: The reason these definitions are there is partially because of how the Permissions API defines the permissions storage data model. What would the best way of defining IDL for objects that are only included in internal slots?

PR #9855, opened 2018-03-05
Update the webvtt IDL file

lukebjerring:


foolip (changes requested):
Can you update https://github.com/w3c/web-platform-tests/blob/master/webvtt/api/interfaces.html to use this?

foolip (approved):

PR #9880, opened 2018-03-06
AccName: Correct "embedded-menu" expectations; add three new test cases

joanmarie:

### DIFF
--- a/accname/description_link-with-label-manual.html
+++ b/accname/description_link-with-label-manual.html
@@ -1,7 +1,7 @@
 <!doctype html>
 <html>
   <head>
-    <title>Name file-label-embedded-menu</title>
+    <title>Description link-with-label</title>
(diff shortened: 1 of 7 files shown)


Ms2ger (approved):

halindrome (commented):
Question...  there are a bunch of changes from "Flash the screen 1 times" to "Flash the screen times".  Was that intentional?  It seems weird.

joanmarie:
They are intentional. That is the thing being corrected. Contrary to what was in the ARIA 1.0 test suite, the algorithm to calculate the name for embedded ARIA menus doesn't include using aria-selected as a means to include a value in the name. There are other tests where the 1 should be included.

The minutes don't adequately capture the discussion, but.... https://www.w3.org/2018/03/01-aria-minutes.html#item05

PR #9886, opened 2018-03-06
Updated ServiceWorker IDL file

lukebjerring:


w3c-bots:
# Build PASSED

Started: 2018-03-06 21:49:37
Finished: 2018-03-06 22:02:46


<h2>Failing Jobs</h2>
<ul>

<li>chrome:dev</li>

<li>safari:11.0</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/25393)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/350028163)

interfaces/ServiceWorker.idl
```diff
@@ -0,0 +1,3 @@
+partial interface ServiceWorkerGlobalScope {
```
foolip: There must be a bug here, https://w3c.github.io/ServiceWorker/ has loads of IDL.

foolip (changes requested):

interfaces/ServiceWorker.idl
lukebjerring: Looks to be the case.

foolip: Now fixed it seems.

foolip (changes requested):
Can you update the following in some way to make use of the file?
https://github.com/w3c/web-platform-tests/blob/master/service-workers/service-worker/interfaces-sw.https.html
https://github.com/w3c/web-platform-tests/blob/master/service-workers/service-worker/interfaces-window.https.html
https://github.com/w3c/web-platform-tests/blob/master/service-workers/service-worker/resources/interfaces-idls.js
https://github.com/w3c/web-platform-tests/blob/master/service-workers/service-worker/resources/interfaces-worker.sub.js

lukebjerring:
Updated the tests. Note that some of the IDL is untested (but, it's made more obvious with the `{only: [...]}` than trying to eyeball-diff the IDL js file and the scraped file.

interfaces/ServiceWorker.idl
mkruisselbrink: This seems incorrectly include the example-only IDL from https://w3c.github.io/ServiceWorker/#extensibility? Neither this nor the previous definition are normative (both of them are inside class=example and data-no-idl blocks, which at least tells bikeshed that it shouldn't consider them to be part of the normative spec text).

lukebjerring: Thanks for reviewing carefully enough to notice.

@mdittmer  - Is the scraper a little over-eager here?

mdittmer: That could be. Our [exclusion criteria](https://github.com/GoogleChromeLabs/webidl-diff/blob/8e6fa545cb68fada4861853d56d520d28bb357da/lib/org/chromium/webidl/IDLFragmentExtractor.js#L72) should be detecting the `example` CSS class name here. Perhaps we should file an issue.

lukebjerring: Address by https://github.com/GoogleChromeLabs/webidl-diff/pull/107

foolip (approved):

PR #9895, opened 2018-03-07
Rename generic-sensor as sensors to align with specname

Honry:
reflect to https://github.com/w3c/web-platform-tests/pull/9830/


Honry:
@foolip, PTAL.

interfaces/generic-sensor.idl
```diff
@@ -1,25 +0,0 @@
-[SecureContext, Exposed=Window]
```
foolip: Rather than removing this, can you rename it? In case https://github.com/w3c/web-platform-tests/pull/9830 has no differences, that'll leave a clearer history behind.

foolip (approved):
Thanks! Please rename and merge, then we'll see about https://github.com/w3c/web-platform-tests/pull/9830.

foolip:
If you want to rename https://github.com/w3c/web-platform-tests/tree/master/generic-sensor as well in this same PR, that'd be fine I think.

Honry:
Sure, I will do that.

interfaces/generic-sensor.idl
Honry: Sure.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25423)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/350158313)

kaixinjxq:
LGTM

Honry:
@foolip, @lukebjerring, I revisit the introduction guide at http://web-platform-tests.org/introduction.html to confirm how to define a directories.
> For W3C specs, these directories are typically named after the shortname of the spec (i.e. the name used for snapshot publications under /TR/)

So, we should use the shorname of publications under `/TR/` rather than ED spec.
For Generic Sensor, its publication is https://www.w3.org/TR/generic-sensor/

PTAL, I suppose this will affect the code in auto-sync tool.

Honry:
@foolip, @lukebjerring, friendly ping.

foolip:
Hmm, so the rename has already happened now. I didn't realize the TR name is different, that seems annoying, and certainly for the purposes of auto-updating IDL it's simpler to pick a name based on the URL we're getting the IDL from rather than where a snapshot might live.

@jgraham @gsnedders what do you think about this? What should http://web-platform-tests.org/introduction.html#test-layout say, and should we do anything about cases that don't match what it says?

My preference is the IDs in https://github.com/foolip/day-to-day/blob/gh-pages/specs.json, because I've picked my favorite ones across the board :)

gsnedders:
@foolip so I'm pretty sure it says the TR name because the TR name tends to be far more stable; I don't really have any strong preference but I'd rather not move things around too often.

foolip:
I guess I think that once something has a TR name, then the github.io name should never change, unless the TR name changes as well. I think there are a few more mismatched cases like this, but I'm not going to go looking for them and seek to unify.

Unless someone notices, I guess leave this one as sensors?

Honry:
Another instance of Battery Status API:

>Latest published version:
http://www.w3.org/TR/battery-status/
Latest editor's draft:
https://w3c.github.io/battery/

IMO, an update for test suite directory will affect many referenced link in corresponding spec documentations(spec, wiki, WG charter). I'd prefer leave it as TR name.

foolip:
I'm happy to leave that one alone too.

Honry:
A few more mismatched cases: 
> Latest published version: https://www.w3.org/TR/mediacapture-streams/
Latest editor's draft: https://w3c.github.io/mediacapture-main/
Latest published version: https://www.w3.org/TR/appmanifest/
Latest editor's draft: https://w3c.github.io/manifest/
Latest published version: http://www.w3.org/TR/screen-capture/
Latest editor's draft: https://w3c.github.io/mediacapture-screen-share/
Latest Published Version: https://www.w3.org/TR/orientation-event/
Latest Editor's Draft: https://w3c.github.io/deviceorientation/

Honry:
@foolip, see https://github.com/w3c/web-platform-tests/issues/9937

PR #9899, opened 2018-03-07
[encrypted-media] Correctly call isInitDataTypeSupported.

charlie-ht:
This test appears to have been doing no useful testing, since it was
passing invalid keysystem names to all browsers, which were all returning
false in requestMediaKeySystemAccess, and the test lets such a result
pass. The reason for that is because the standard does not impose which
of the init data types a browser must support.

### DIFF
--- a/encrypted-media/scripts/check-initdata-type.js
+++ b/encrypted-media/scripts/check-initdata-type.js
@@ -2,12 +2,12 @@
  {
     function checkInitDataType(initDataType)
     {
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25456)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/350386348)

PR #9900, opened 2018-03-07
Fixed firefox binary path for Windows and Mac in browser.py

Cactusmachete:
find_executable() earlier returned None for non-linux systems, causing
firefox browser install to fail. Added the different binary paths for each OS.
Also removed platform_string() which used to be called in install() before nightly
download depended on mozdownload, and is now dead code.
Removed xfail for firefox install and run on MacOS in test_wpt.py as it will work now.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9900)

### DIFF
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -65,28 +65,9 @@ class Firefox(Browser):
     """
 
     product = "firefox"
-    binary = "firefox/firefox"
-    platform_ini = "firefox/platform.ini"
+    binary = "browsers/firefox/firefox"
+    platform_ini = "browsers/firefox/platform.ini"
     requirements = "requirements_firefox.txt"
-
-    def platform_string(self):
-        platform = {
-            "Linux": "linux",
-            "Windows": "win",
-            "Darwin": "mac"
-        }.get(uname[0])
-
-        if platform is None:
-            raise ValueError("Unable to construct a valid Firefox package name for current platform")
-
-        if platform == "linux":
-            bits = "-%s" % uname[4]
-        elif platform == "win":
-            bits = "64" if uname[4] == "x86_64" else "32"
-        else:
-            bits = ""
-
-        return "%s%s" % (platform, bits)
 
     def platform_string_geckodriver(self):
         platform = {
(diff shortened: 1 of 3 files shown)


w3c-bots:
# Build ERRORED

Started: 2018-03-07 16:45:54
Finished: 2018-03-07 17:03:48


<h2>Failing Jobs</h2>
<ul>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/25457)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/350410249)

tools/wpt/browser.py
```diff
@@ -111,27 +92,46 @@ def install(self, dest=None):
...
+                install_dest = mozinstall.install(filename, dest)
             else:
                 raise
 
         os.remove(filename)
-        return find_executable("firefox", os.path.join(dest, "firefox"))
+
+        if platform == "linux":
+            bin_path = os.path.join(install_dest, "firefox")
+        elif platform == "win":
```
jgraham: I wonder if we should add `firefox` to the `dest` in non-linux cases, to ensure we aren't overwriting existing files.

tools/wpt/browser.py
```diff
@@ -111,27 +92,46 @@ def install(self, dest=None):
...
         import mozinstall
 
+        platform = {
+            "Linux": "linux",
+            "Windows": "win",
+            "Darwin": "mac"
+        }.get(uname[0])
+
+        if platform is None:
+            # This was earlier checked in platform_string, which isn't needed anymore since we switched
```
jgraham: I think this comment should explain why the code is as it is rather than explain the historical context (I'm not actually sure a comment is needed here at all).

Cactusmachete: Alright! Added it just to clarify why it was here instead of in platform_string rather than anywhere else. Will remove it

tools/wpt/browser.py
Cactusmachete: Afaik it shouldn't overwrite any files besides the firefox ones in non-linux cases. 
On mac, it installs the .dmg to make the firefox nightly app, and in windows, the installer sets it up so that the firefox build is within a new folder named with the nightly version and language it installed. 
Not sure if these can lead to us overwriting anything, but then again I don't have much experience with the two OSes.

tools/wpt/browser.py
```diff
@@ -111,30 +92,68 @@ def install(self, dest=None):
...
 
     def find_binary(self):
-        return find_executable("firefox")
+        platform = {
+            "Linux": "linux",
+            "Windows": "win",
+            "Darwin": "mac"
+        }.get(uname[0])
+
+        dest = os.path.join(os.getcwd(), "_venv")
```
jgraham: Instead of this I think we should figure out how to pass the virtualenv object into `find_binary`. I know that's a slightly larger refactor but it won't break if we change the venv name.

tools/wpt/browser.py
```diff
@@ -111,30 +92,68 @@ def install(self, dest=None):
...
+        elif platform == "win":
+            bin_path = os.path.join(install_dest, "core")
+        elif platform == "mac":
+            bin_path = os.path.join(install_dest, "Contents", "MacOS")
+
+        return find_executable("firefox", bin_path)
 
     def find_binary(self):
-        return find_executable("firefox")
+        platform = {
```
jgraham: I think we ended up with some duplicated code working out the binary path in the virtualenv, is that possible to factor into a function?

jgraham (changes requested):
This is going in the right direction and dI"m very pleased to be getting such a silly and non-obvious bug fixed, but I think we need a slightly larger refactor to be confident it won't break in the  future.

tools/wpt/browser.py
Cactusmachete: Alright, I'll see what I can do about this.

tools/wpt/browser.py
Cactusmachete: Something like find_binary_path is definitely doable. We can pass the virtualenv object to it (instead of to find_binary) and have find_binary use it to ensure that we return the firefox on path if its not installed in _venv

jgraham:
@Cactusmachete I think the `find_binary` will be fine if it doesn't have to search the entire virtualenv. So if we can arrange things so that the path always start with `_venv/firefox` I think that would work well.

jgraham (approved):

jgraham:
Please rebase this, and squash the commits together into something with meaningful history (probably just one commit for the whole thing). Please ask if that doesn't make sense.

jgraham:
This branch now contains a merge commit that introduces a whitespace error :) There shouldn't be an merge commits on the branch at all (you should use something like `git rebase -i origin/master` to rebase onto master).

jgraham (approved):

PR #9901, opened 2018-03-07
Use force flag on git add operations during syncing

jdm:
This ensures that files that belong in the tree but match .gitignore filters are still synced appropriately (like payment-handler/manifest.json, which matches MANIFEST.json).

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9901)

### DIFF
--- a/tools/wptrunner/wptrunner/update/tree.py
+++ b/tools/wptrunner/wptrunner/update/tree.py
@@ -178,9 +178,9 @@ class GitTree(object):
                        add all files under that path.
         """
         if prefix is None:
-            args = ("-a",)
+            args = ("-f", "-a",)
         else:
-            args = ("--no-ignore-removal", prefix)
+            args = ("-f", "--no-ignore-removal", prefix)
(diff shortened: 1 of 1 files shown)


jgraham:
I am worried about this leading to cache files (`.pyc`, `foo~` and so on) ending up in the tree. I think a better solution would be to special case the relevant files using `.gitignore` rules like
`!payment-handlers/manifest.json`
and a lint that uses `git check-ignore` to check that there aren't comitted files that match the ignore rule (this would probably have to be skipped in environments without git installed).

w3c-bots:
# Build PASSED

Started: 2018-03-07 16:56:58
Finished: 2018-03-07 17:16:56





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25458)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/350415821)

jdm:
This is superseded by #9902.

PR #9911, opened 2018-03-07
[WIP] Update testing of Request's default credentials mode

domfarolino:
This is for https://github.com/whatwg/fetch/pull/585.

I'm a little surprised if only these tests have to be updated, so I'm marking this as a WIP for now. Basically what I did though, was:

 - Ran all of the `/fetch`, `/xor`, `service-workers`, and `/cors` tests on Chromium, and noted failues (none)
 - Modified the Request constructor to use the `same-origin` credentials fallback instead of the previous `omit`
 - Ran all of the previous tests again and noted the failures (only two)
 - Updated the affected tests!

### DIFF
--- a/fetch/api/request/request-init-003.sub.html
+++ b/fetch/api/request/request-init-003.sub.html
@@ -44,7 +44,7 @@
                              "referrer" : "about:client",
                              "referrerPolicy" : "",
                              "mode" : "cors",
-                             "credentials" : "omit",
+                             "credentials" : "same-origin",
                              "cache" : "default",
                              "redirect" : "follow",
                              "integrity" : "",
--- a/fetch/api/request/request-structure.html
+++ b/fetch/api/request/request-structure.html
@@ -76,7 +76,7 @@
             break;
 
           case "credentials":
-            defaultValue = "omit";
+            defaultValue = "same-origin";
             newValue = "cors";
             break;
 


annevk:
I guess we don't really have good same-origin cookie/HTTP authentication tests then.

@wanderview does that make sense?

domfarolino:
Well this means we may have good ones, they probably just explicitly set `same-origin` or `include` whenever using them though right? I guess those tests could be updated to not explicitly provide a credentials mode since the default should suffice (let me know if you want me to hunt around for those in this PR). This makes sense, as how many tests can you really have that enforce the default credentials mode...

domfarolino:
Pinging @wanderview here. This seems to be the only thing blocking [Chrome's implementation](https://chromium-review.googlesource.com/c/chromium/src/+/981512) so the sooner we can get this in the better :)

wanderview (approved):
LGTM.  I'm surprised there aren't any service worker tests that check the default credentials via the FetchEvent.

domfarolino:
Thanks a lot! Me too, should we make an issue to add some at some point? I would expect more tests to be failing here, and we probably want more tests to be doing so.

annevk:
Thoughts on https://github.com/whatwg/fetch/pull/585#issuecomment-378843176 anyone? Sounds like @wanderview wants to go ahead with this in Firefox as well, so I'm happy to land this, just wondering about follow-up work and how to structure that.

wanderview:
BTW, the service worker tests mostly check that the correct value is passed through to FetchEvent.  The tests use explicit modes for this.  They are not really checking for defaults.  I think that its reasonable for the fetch tests to mainly be checking the defaults.

PR #9919, opened 2018-03-08
DOM: reset target/relatedTarget

annevk:
Helps with issue https://github.com/whatwg/dom/issues/562 and PR https://github.com/whatwg/dom/pull/585.


annevk:
I just realized I might also have to modify some tests added by @TakayoshiKochi in #7481. At least skimming through it seems they assume `relatedTarget` is not reset.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25597)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/350721206)

annevk:
Reviewing shadow-dom/event-post-dispatch.html it seems that expects some kind of difference in exposure depending on whether composed was set. I don't think we ever discussed that.

TakayoshiKochi:
Re https://github.com/w3c/web-platform-tests/pull/9919#issuecomment-372295695

So for when `composed` flag is set to false, do you think more changes needed for the spec, regarding computing *relatedTargets*?

annevk:
@TakayoshiKochi per the specification the "composed flag" only affects https://dom.spec.whatwg.org/#interface-shadowroot. The tests Google wrote seem to assume that it also affects other things, but I'm not sure why it would.

TakayoshiKochi:
Ah, I understand.  (maybe off-topic for this PR, discussion can be elsewhere)

At [dispatching events](https://dom.spec.whatwg.org/#dispatching-events) step 18, after dispatching the event to all targets, when composed flag is false, the event stops at the shadow root, and matches the condition of "target's root is a shadow root".  Therefore both `target` and `relatedTarget` have to be nullified.

TakayoshiKochi:
So in [your spec change](https://github.com/whatwg/dom/pull/585) *clearTargetsPostDispatch* seems intended to capture the conditions that `target` and `relatedTarget` should be nullified, but you still need another condition after dispatch, then?

dom/events/relatedTargets.window.js
```diff
@@ -0,0 +1,50 @@
+const host = document.createElement("div"),
```
TakayoshiKochi: I was not familiar with this style of writing a test. Looks nice to avoid boilerplate in the HTML.
Can we add metadata, especially corresponding spec URL as a comment?

annevk:
@TakayoshiKochi you mean in case someone moved the target inside a shadow tree? I don't think we should consider that case, but I wonder what @smaug---- thinks. We've been solely focused on pre-dispatch conditions for all kinds of parts of eventing.

smaug----:
yeah, I think it was discussed in the other issue where relatedTarget handling was fixed on spec level that moving nodes into ShadowDOM isn't really an issue, since if you do that, you already have cross ShadowDOM access and such, and are somewhat explicitly breaking the encapsulation.

TakayoshiKochi:
I see, then makes sense the latter half of the async_test cases in this test.
Yeah, I was wondering if we should take care of cases where an event listener pulled a node out of shadow tree to document, we should retain the `target/relatedTarget` or not (answer: no, still should be nullified per pre-dispatch condition).

dom/events/relatedTargets.window.js
annevk: You mean link to https://dom.spec.whatwg.org/#concept-event-dispatch? Feels a little self-explanatory, but I could do.

TakayoshiKochi: Yeah, I think it would be nice.

TakayoshiKochi:
Adding hayatoIto@ as he's looking the spec change more closely.

dom/events/relatedTarget.window.js
```diff
@@ -0,0 +1,85 @@
...
+    for (target of [new XMLHttpRequest(), self, host]) {
+      const event = new FocusEvent("demo", { relatedTarget: relatedTarget });
+      target.dispatchEvent(event);
+      assert_equals(event.target, null);
+      assert_equals(event.relatedTarget, null);
+    }
+  }
+}, "Reset if relatedTarget pointed to a shadow tree");
+
+async_test(t => {
```
TakayoshiKochi: Can this be a sync test()?
Because this seems to be testing a synchronous event dispatch and does not need to be async_test(),
and checking the side-effect of the event handler directly e.g. on line 35.

Although it is still okay to use async_test() if you prefer as is, as it covers synchronous testing. 
I'm a bit feeling redundant that at the `t.done()` is called at the end of an async_test.

annevk: I don't think you can use t.step stuff in a sync test.

TakayoshiKochi: Yes, but as you do on line 80 (`seen = true` in the event handler to negate the value) then immediately after dispatching the event, check the value on line 83, you should be able to do the same here.
In fact, you expect the side effect of the event handler at line 35, so you don't need to wait for the `step_func()` to finish on line 39.

annevk: I'm not waiting for it? Callbacks need to be wrapped in steps and you can only have steps in asynchronous tests, I believe. Is that not true?

TakayoshiKochi: You are right, so to be clear, I'm not against `async_test` - `t.step_func` - `t.done` construct used here.

What I suggested was alternatively this test could be written as a sync test, as the callback (event handler) here is expected to be called in synchronous manner, and you can expect its side effect on the next line after `dispatchEvent`, and technically I thought it would be possible.

Probably it's a matter of taste, either isn't much better than the other.

annevk: That does not work, since if the callback ends up throwing in that case it'd end up being catched by the framework and not the test.

TakayoshiKochi: Ah, okay, I understand.  Then LGTM!

TakayoshiKochi (approved):
Maybe you want to add further tests?
The current set of tests looks good to me.

annevk:
Yeah, I'm a little unsure how to proceed. I think the tests Google wrote I pointed out earlier need to be updated and there probably need to be more tests as well.

I guess I can try updating the Google tests as part of this PR and add new tests, if any, in separate PRs.

Does that seem reasonable?

TakayoshiKochi:
I think you can do either way, I prefer keeping a PR small and I don't mind small additions later, but if you prefer the spec change and the tests to have 1:1 relationship, it's also fine.

annevk:
I had a look through these and these are all still correct per the latest changes to https://github.com/whatwg/dom/pull/585.

PR #9920, opened 2018-03-08
Add tests for return value of DOMTokenList's replace() method

Zirro:
This adds tests for a boolean value being returned from the `replace()` method, as discussed in https://github.com/whatwg/dom/issues/577 and https://github.com/whatwg/dom/pull/582.

It also adds a test which attempts to replace a token that isn't present with one that is, since this case was not covered by the existing tests.

### DIFF
--- a/interfaces/dom.idl
+++ b/interfaces/dom.idl
@@ -548,7 +548,7 @@ interface DOMTokenList {
   [CEReactions] void add(DOMString... tokens);
   [CEReactions] void remove(DOMString... tokens);
   [CEReactions] boolean toggle(DOMString token, optional boolean force);
-  [CEReactions] void replace(DOMString token, DOMString newToken);
+  [CEReactions] boolean replace(DOMString token, DOMString newToken);
   boolean supports(DOMString token);
   [CEReactions] stringifier attribute DOMString value;
   //  iterable<DOMString>;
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25598)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/350721460)

foolip (dismissed):
In the diff it looks like you've changed the indentation of many lines without changing anything else. Can you revert those changes to just have a minimal diff?

Zirro:
@foolip I think you're referring to the lines where I've added an additional `null` argument. I'm not sure why they're not being highlighted like the lines further down.

foolip:
@Zirro, you're right, I didn't see that. I'll remove by request for changes, but also not reviewing this right now.

interfaces/dom.idl
```diff
@@ -548,7 +548,7 @@ interface DOMTokenList {
   [CEReactions] void add(DOMString... tokens);
   [CEReactions] void remove(DOMString... tokens);
   [CEReactions] boolean toggle(DOMString token, optional boolean force);
-  [CEReactions] void replace(DOMString token, DOMString newToken);
+  [CEReactions] boolean replace(DOMString token, DOMString newToken);
```
foolip: @lukebjerring, this is a problem we're going to run in pretty quickly with generated IDL, where people update the tests together with the spec. As a reviewer, I think we should probably just say "yes please" and if there are any differences the auto-update will take care of it. WDYT?

annevk (approved):

interfaces/dom.idl
lukebjerring: Presumably, differences would be subtle (whitespace, indenting, etc). Until the updater is mature, and we document how to run it, will just have to accept changes :)

PR #9926, opened 2018-03-08
Streams: Modify tests to expect abort reason passthrough

ricea:
This tests the changes in https://github.com/whatwg/streams/pull/903.

Replace expectations that a TypeError will be stored after abort() is
called with with expectations that the reason passed to abort() will be
stored instead.

Also add a test of the stored error on the readable side of a
TransformStream after abort() has been called on the writable size. This
was not explicitly tested.

Also fix a bug in the general.js test: it didn't wrap a call to
assert_unreached properly, so a failure would have shown up as an
unhandled rejection rather than being reported properly.

### DIFF
--- a/streams/piping/error-propagation-backward.js
+++ b/streams/piping/error-propagation-backward.js
@@ -548,7 +548,7 @@ promise_test(() => {
   return rs.pipeTo(ws).then(
     () => assert_unreached('the promise must not fulfill'),
     err => {
-      assert_equals(err.name, 'TypeError', 'the promise must reject with a TypeError (_not_ with error1)');
+      assert_equals(err, error1, 'the promise must reject with error1');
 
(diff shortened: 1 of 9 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25627)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/350883065)

domenic (approved):
Could we add a few smoke tests for passing through `undefined` implicitly (via `abort()`), `undefined` explicitly, and for a string? Definitely people should be using `Error` instances, but we should test the degenerate use cases.

ricea:
I will add tests for degeneracy in a followup change.

PR #9933, opened 2018-03-09
add test for aborting a document load

anthumchris:
Test for "abort" event firing on Window during a document load cancellation,
triggered by ESC key, window.stop(), or browser stop button.

https://github.com/whatwg/html/issues/3525

### DIFF
--- a/dom/events/Event-dispatch-abort-document-load.html
+++ b/dom/events/Event-dispatch-abort-document-load.html
@@ -1,0 +1,59 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Abort event fires when aborting a document load</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+// delay window.load event with slow imageso we can abort page load
+const img = document.createElement('img');
+img.src = '../../xhr/resources/delay.py?ms=1000'
+
+// Allow time for image to begin downloading
+const loadAbortTimeout = 50;
+
+// queue tests to run after abort
+const evalEventsTimeout = loadAbortTimeout+1;
+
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25671)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/351110662)

AnthumChris:
I'm open to any test that helps browser implementers ensure that some kind of event is fired.  This is especially crucial to the Streams and Service Worker APIs and sending proper notifications upstream to applications.

gterzian:
A maintainer might disagree with me, and as far as I can read from the spec, it seems that the [abort event](https://html.spec.whatwg.org/multipage/#event-abort) is only fired in response to explicit user interaction with the user agent(for example by pressing the Esc key), not when a script calls `window.stop`. 

Also looking at the below, it's interesting to read that the 'abort' event should be fired **before** running the [abort steps](from https://html.spec.whatwg.org/multipage/#abort-a-document). 

"User agents **may allow users to explicitly invoke the abort a document algorithm** for a Document. If the user does so, then, if that Document is an active document, the user agent should queue a task to **fire an event named abort** at that Document's Window object **before invoking the abort algorithm**."(emphasis are mine)

So if I were to write a test for this, I would probably use a [manual test](https://github.com/web-platform-tests/wpt/blob/4489ad989460d82fbf59c26c94248acd78583a0c/docs/_writing-tests/manual.md) that would simply assert that 'abort' was fired, I guess by starting a slow load like you've done with the image(maybe via a 'start' button on the page), and then have the tester push `Esc` or something similar and show the results of an assertion within `window.onabort`...

PR #9945, opened 2018-03-09
Allow passing stability options through command line (Fixes #9877)

### LINKED ISSUE #9877: Allow passing in stability options via wptrunner command line
wptrunner (and therefore `wpt run`) has a `--verify` option. This runs the tests multiple times in various configurations in order to check that they're stable. At present this has to be configured using Python code. It would be helpful to be able to set the configuration options (i.e.   the keyword parameters in `tools/wptrunner/wptrunner/stability.py::check_stability`) via the command line. At the very least the `max_time` option should be settable, so we are able to set the maximum runtime of the command to a specified number of seconds (or other time unit; a typical value might be 50 minutes).

marimeireles:
repeat_loop = --verify-repeat-loop
repeat_restart = --verify-repeat-restart
(disable) chaos_mode = --verify-no-chaos-mode
max_time = --verify-max-time
(disable) output_results = --verify-no-output-results

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9945)


w3c-bots:
# Build PASSED

Started: 2018-03-09 16:34:11
Finished: 2018-03-09 16:52:58





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25715)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/351380502)

marimeireles:
I'm not sure if the help messages I wrote are good.
Could you give me some insight on how to make them better if they're not good enough?
Thanks!

tools/wptrunner/wptrunner/wptcommandline.py
```diff
@@ -82,6 +83,24 @@ def create_parser(product_choices=None):
...
                             help="Output per-iteration test results when running verify")
+    mode_group.add_argument("--verify-repeat-loop", action="store",
+                            default=10,
+                            help="Run tests in a loop for a number of times",
+                            type=int)
+    mode_group.add_argument("--verify-repeat-restart", action="store",
+                            default=5,
+                            help="Restart the whole process",
+                            type=int)
+    mode_group.add_argument("--verify-no-chaos-mode", action="store_false",
```
jgraham: I think this should use `dest="verify_chaos_mode"` since a True value means that chaos mode is used. Possibly there should also be a `--verify-chaos-mode` option so that you can specify a value even if we change the default.

tools/wptrunner/wptrunner/wptcommandline.py
```diff
@@ -82,6 +83,24 @@ def create_parser(product_choices=None):
...
+                            default=True,
+                            help="Disable chaos mode when running on Firefox")
+    mode_group.add_argument("--verify-max-time", action="store",
+                            default=None,
+                            help="The maximum number of minutes for each test to run",
```
jgraham: This isn't a time in minutes, it's in seconds. And also the time is the maximum time for the job to run, it's not a time per test.

tools/wptrunner/wptrunner/wptcommandline.py
```diff
@@ -82,6 +83,24 @@ def create_parser(product_choices=None):
...
+                            type=lambda x: timedelta(minutes=float(x)))
+    mode_group.add_argument("--verify-no-output-results", action="store_false",
```
jgraham: This should also use `dest` so that the variable name matches the semantics.

tools/wptrunner/wptrunner/wptcommandline.py
```diff
@@ -82,6 +83,24 @@ def create_parser(product_choices=None):
     mode_group.add_argument("--verify-log-full", action="store_true",
                             default=False,
                             help="Output per-iteration test results when running verify")
+    mode_group.add_argument("--verify-repeat-loop", action="store",
+                            default=10,
+                            help="Run tests in a loop for a number of times",
+                            type=int)
+    mode_group.add_argument("--verify-repeat-restart", action="store",
+                            default=5,
+                            help="Restart the whole process",
```
jgraham: "Number of iterations, for a run that restarts the runner between each iteration"

tools/wptrunner/wptrunner/wptcommandline.py
```diff
@@ -82,6 +83,24 @@ def create_parser(product_choices=None):
     mode_group.add_argument("--verify-log-full", action="store_true",
                             default=False,
                             help="Output per-iteration test results when running verify")
+    mode_group.add_argument("--verify-repeat-loop", action="store",
+                            default=10,
+                            help="Run tests in a loop for a number of times",
```
jgraham: "Number of iterations for a run that reloads each test without restart."

jgraham (changes requested):

tools/wptrunner/wptrunner/wptcommandline.py
marimeireles: Let me see if I get this right you want me create a new --verify-chaos-mode and set both `--verify-chaos-mode` and `--verify-no-chaos-mode` with the same dest="verify_chaos_mode"? So that you can both enable and disable from the command line.

tools/wptrunner/wptrunner/wptcommandline.py
marimeireles: I've set it to a time in minutes:
https://github.com/w3c/web-platform-tests/blob/82da4d2663627032f6befd1a9618de51f31a2509/tools/wptrunner/wptrunner/wptcommandline.py#L100
Would you rather I change it to seconds?

Also, I'll fix the help string.

jgraham: Oh. Let's try this way first.

tools/wptrunner/wptrunner/wptcommandline.py
marimeireles: Hey @jgraham could you confirm this? I'm not sure if that was what you meant. Thanks!

jgraham: Ah apologies, I missed the question. Yes, your understanding is perfect.

marimeireles:
I believe I adressed all the comments @jgraham.

jgraham (approved):
Thanks!

PR #9957, opened 2018-03-10
lint.py: Adds a lint for files with the same basename in the same dir…

### LINKED ISSUE #7570: Consider adding a lint for files with the same basename in the same directory
This causes issues for Blink/WebKit at this time; see https://github.com/w3c/web-platform-tests/pull/7569. I'd argue it's not a very good idea in general anyway.

kriti21:
This adds a lint to the `all_paths_lints`
set in `tools/lint/lint.py` that checks if
there are any paths in which 'os.path.splitext(paths)[0]`
has multiple identical values.

Closes https://github.com/w3c/web-platform-tests/issues/7570

### DIFF
--- a/tools/lint/lint.py
+++ b/tools/lint/lint.py
@@ -624,6 +624,29 @@ def check_script_metadata(repo_root, path, f):
     return errors
 
 
+def check_multiple_identical_values(repo_root, paths):
+    """
+    Runs lints that check for multiple identical paths
+    in all_paths_lints.
+
+    :param repo_root: the repository root
+    :param paths: a list of all the paths within the repository
+    :returns: a list of errors found in ``path``
+    """
+
+    file_dict = dict()
+    errors = []
+    file_dict = defaultdict(list)
+    for path in paths:
+        file_name, file_extension =  os.path.splitext(path)
+        file_dict[file_name].append(file_extension)
+    for f in file_dict.keys():
+        if len(file_dict[f]) > 1:
+            msg = "Got files for path %s with extensions %s" % (f, file_dict[f])
+            errors.append(("DUPLICATE-EXTENSIONLESS-PATH", msg, f, None))
+    return []
+
+
 def check_path(repo_root, path):
     """
     Runs lints that check the file path.
@@ -837,7 +860,7 @@ def lint(repo_root, paths, output_format):
     return sum(itervalues(error_count))
 
 path_lints = [check_path_length, check_worker_collision, check_ahem_copy]
-all_paths_lints = [check_css_globally_unique]
+all_paths_lints = [check_css_globally_unique, check_multiple_identical_values]
 file_lints = [check_regexp_line, check_parsed, check_python_ast, check_script_metadata]
 
 if __name__ == "__main__":


kriti21:
I have added changes according to what i understood through the comments on the issue. However, when i run `./wpt lint` locally, it shows error -

```
    wpt.main()
  File "/home/kriti/mozdev/web-platform-tests/tools/wpt/wpt.py", line 132, in main
    rv = script(*args, **kwargs)
  File "/home/kriti/mozdev/web-platform-tests/tools/lint/lint.py", line 794, in main
    return lint(repo_root, paths, output_format)
  File "/home/kriti/mozdev/web-platform-tests/tools/lint/lint.py", line 846, in lint
    errors = check_all_paths(repo_root, paths)
  File "/home/kriti/mozdev/web-platform-tests/tools/lint/lint.py", line 670, in check_all_paths
    errors.extend(paths_fn(repo_root, paths))
  File "/home/kriti/mozdev/web-platform-tests/tools/lint/lint.py", line 639, in check_multiple_identical_values
    if len(os.path.splitext(paths_fn)) > 0:
  File "/usr/lib/python2.7/posixpath.py", line 98, in splitext
    return genericpath._splitext(p, sep, altsep, extsep)
  File "/usr/lib/python2.7/genericpath.py", line 99, in _splitext
    sepIndex = p.rfind(sep)
AttributeError: 'function' object has no attribute 'rfind'
```
Please help me in correcting this. @jgraham

w3c-bots:
# Build BROKEN

Started: 2018-03-10 17:01:20
Finished: 2018-03-10 17:17:14


<h2>Failing Jobs</h2>
<ul>

<li>lint</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/25758)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/351741842)

tools/lint/lint.py
```diff
@@ -624,6 +624,23 @@ def check_script_metadata(repo_root, path, f):
     return errors
 
 
+def check_multiple_identical_values(repo_root, path):
```
jgraham: So, this function will be passed the `repo_root` and a list of all the paths, not a single path.

tools/lint/lint.py
```diff
@@ -624,6 +624,23 @@ def check_script_metadata(repo_root, path, f):
...
+    Runs lints that check for multiple identical paths
+    in all_paths_lints.
+
+    :param repo_root: the repository root
+    :param path: the path of the file within the repository
+    :returns: a list of errors found in ``path``
+    """
+
+    errors = []
+    for paths_fn in all_paths_lints:
```
jgraham: So here you are iterating over all the lints, not over the list of paths.

tools/lint/lint.py
```diff
@@ -624,6 +624,23 @@ def check_script_metadata(repo_root, path, f):
...
+        if len(os.path.splitext(paths_fn)) > 0:
```
jgraham: If I assume that the input here is supposed to be a path, this is always going to be true because `splitext` is goingg to return a tuple of `path, ext`. So you need to extract the first part and keep some state to decide whether we've seen that path before. There are various ways to do that; if the list is already sorted we can just check if the current path is equal to the previous one. If we can't reply on that, the simplest approach may be to keep a dict of `{path: ext}`. If we encounter a path that's already in the dictionary we add it to a `defaultdict(list)` of `{path: [exts]}`, and then at the end we create an error for each entry in the latter dictionary, listing all the extensions that exist for the same path.

jgraham (changes requested):

kriti21:
@jgraham I have made the suggested changes. However, I don't understand what arguments to pass to `error.extend()` function which is showing error. Please help.

tools/lint/lint.py
```diff
@@ -624,6 +624,32 @@ def check_script_metadata(repo_root, path, f):
...
+    :param repo_root: the repository root
+    :param paths: a list of all the paths within the repository
+    :returns: a list of errors found in ``path``
+    """
+
+    file_dict = dict()
+    errors = []
+    for paths in paths:
+        file, file_extension =  os.path.splitext(paths)
+        if file not in file_dict.keys():
```
jgraham: There's a type in the standard library called `defaultdict` that encapsulates this check-and-create behaviour; let' use that.

tools/lint/lint.py
```diff
@@ -624,6 +624,32 @@ def check_script_metadata(repo_root, path, f):
...
+
+    :param repo_root: the repository root
+    :param paths: a list of all the paths within the repository
+    :returns: a list of errors found in ``path``
+    """
+
+    file_dict = dict()
+    errors = []
+    for paths in paths:
+        file, file_extension =  os.path.splitext(paths)
```
jgraham: Avoid using `file` as a variable name since that's built in.

tools/lint/lint.py
```diff
@@ -624,6 +624,32 @@ def check_script_metadata(repo_root, path, f):
...
+        if file not in file_dict.keys():
+            file_dict[file] = list()
+            file_dict[file].append(file_extension)
+        else:
+            file_dict[file].append(file_extension)
+    for f in file_dict.keys():
+        if len(file_dict[f]) > 0:
+            errors.extend(paths(repo_root, paths))
```
jgraham: So the errors list should be a list of tuples of the form `(type, error, path, line)`. In these cases error should be some code like `DUPLICATE EXTENSIONLESS PATH`, `error` is a sitring like `Got files for path %s with extensions %s, paths exclusing extensions must be unique` with the `%s` values fileld in from the data, `path` can be the path without an extension, and `line` will be `None`.

jgraham (changes requested):

kriti21:
@jgraham I have made the suggested changes. Please guide me how to fix the failing build.

tools/lint/lint.py
```diff
@@ -624,6 +624,28 @@ def check_script_metadata(repo_root, path, f):
...
+
+    :param repo_root: the repository root
+    :param paths: a list of all the paths within the repository
+    :returns: a list of errors found in ``path``
+    """
+
+    file_dict = dict()
+    errors = []
+    file_dict = defaultdict(list)
+    for paths in paths:
```
jgraham: `for path in paths`: (note no `s`).

tools/lint/lint.py
```diff
@@ -624,6 +624,28 @@ def check_script_metadata(repo_root, path, f):
...
+        file_name, file_extension =  os.path.splitext(paths)
+        file_dict[file_name].append(file_extension)
+    for f in file_dict.keys():
+        if len(file_dict[f]) > 0:
```
jgraham: I think you mean `> 1` here.

tools/lint/lint.py
```diff
@@ -624,6 +624,28 @@ def check_script_metadata(repo_root, path, f):
...
+    :param repo_root: the repository root
+    :param paths: a list of all the paths within the repository
+    :returns: a list of errors found in ``path``
+    """
+
+    file_dict = dict()
+    errors = []
+    file_dict = defaultdict(list)
+    for paths in paths:
+        file_name, file_extension =  os.path.splitext(paths)
```
jgraham: `path`

tools/lint/lint.py
```diff
@@ -624,6 +624,28 @@ def check_script_metadata(repo_root, path, f):
...
+        file_dict[file_name].append(file_extension)
+    for f in file_dict.keys():
+        if len(file_dict[f]) > 0:
+            msg = "Got files for path %s with extensions %s" % (f, file_dict[f])
+            errors.append(("DUPLICATE-EXTENSIONLESS-PATH", msg, f, None))
+
```
jgraham: If nothing fails you need to return an empty list `return []`

jgraham (changes requested):
Sorry, this fell off the radar a bit, but it's really close now.

kriti21:
@jgraham I have made the changes you suggested. Please have a look and guide me how to correct it so that it passes the tests.

PR #9960, opened 2018-03-10
Replace assert_own_property with assert_idl_attribute

### LINKED ISSUE #6203: cssom/style-sheet-interfaces-001.html fails on all browsers
`cssom/style-sheet-interfaces-001.html` fails on all browsers solely because it uses `assert_own_property(o, p)` to check if some CSSOM object `o` has property `p` rather than using `p in o`. If the calls to `assert_own_property` are changed to `assert_true(p in o)`, this test instead passes on all browsers.

Is it justified to use `assert_own_property` in some spec?

csnardi:
CSSOM objects do not implement attributes as their own properties; they are IDL attributes. Switch to use assert_idl_attribute instead of assert_own_property. Also make inline-style-001.html pass by removing assert_readonly (it resets the attribute, making future tests fail) and use assert_equals instead of assert_regexp_match.

Fixes #6203.

### DIFF
--- a/css/cssom/inline-style-001.html
+++ b/css/cssom/inline-style-001.html
@@ -16,15 +16,14 @@
  <script type="text/javascript">
     test(function() {
         var test = document.getElementById("test");
-        assert_own_property(test, "style");
-        assert_readonly(test, "style");
+        assert_idl_attribute(test, "style");
         declaration = test.style;
     }, "CSSStyleDeclaration_accessible", {
         assert: "Can access CSSStyleDeclaration through style property"
     });
 
     test(function() {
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25761)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/351777003)

dbaron (approved):
Yep, this matches [serialize a CSS declaration block](https://drafts.csswg.org/cssom/#serialize-a-css-declaration-block) and the [presence of PutForwards=cssText](https://drafts.csswg.org/cssom/#the-elementcssinlinestyle-interface).

PR #9970, opened 2018-03-12
Prevent redirect loop in fetch/api/redirect/redirect-location.html

ricea:
The value of invalidLocationUrl, "#invalidurl:" is actually treated as a valid
URL by redirect.py, leading to a redirect loop. This makes the test take over 6
seconds in Chrome, causing it to time out. Change invalidLocationUrl to
"invalidurl:" to avoid the redirect loop and make the test faster.

See Chrome issue http://crbug.com/815286 about the test timing out.

### DIFF
--- a/fetch/api/redirect/redirect-location.js
+++ b/fetch/api/redirect/redirect-location.js
@@ -32,7 +32,7 @@ function redirectLocation(desc, redirectUrl, redirectLocation, redirectStatus, r
 
 var redirUrl = RESOURCES_DIR + "redirect.py";
 var locationUrl = "top.txt";
-var invalidLocationUrl = "#invalidurl:";
+var invalidLocationUrl = "invalidurl:";
 var dataLocationUrl = "data:,data%20url";
 // FIXME: We may want to mix redirect-mode and cors-mode.
 // FIXME: Add tests for "error" redirect-mode.


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25789)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/352337672)

jdm (approved):

SimonSapin:
@ricea In what sense do you mean "invalid" for URLs? We have tests that ensure specifically that the URL parser accepts input with only a scheme followed by a colon: https://github.com/w3c/web-platform-tests/blob/3c5d8637bb/url/urltestdata.json#L4735-L4772

Runnable test case: http://software.hixie.ch/utilities/js/live-dom-viewer/saved/5824

annevk:
Yeah, `something:` would be rejected because it's not an HTTP(S) URL, not because it doesn't parse. If you want something that doesn't parse, use `https://test:test/` or some such.

SimonSapin:
Ok, sounds good.

ricea:
@SimonSapin The intent of the test is not completely clear to me. In this context, I think "invalid URL" probably means "something which you can't successfully redirect to". It doesn't appear that the redirect loop was intentional.

"invalidurl:" prevents redirect loop by causing redirect.py to parse the scheme as "invalidurl". When redirect.py parses the scheme as "", "http" or "https", then it adds its query arguments to the URL, creating the redirect loop. Something like "https://invalid.invalid./" would be another way to prevent a redirect loop, by causing redirect.py to construct a new URL that wouldn't point back to itself.

SimonSapin:
Yeah, feel free to ignore my earlier comment. I was thinking of validity in terms of URL parsing exclusively, but that’s not what’s relevant here.

PR #9973, opened 2018-03-12
prevent harness timeout after calling window.stop()

anthumchris:
window.stop() prevents window.onload from firing, which causes a harness timeout
This fixes that, and the natural flow of things returns after window.stop().

Relates to https://github.com/w3c/web-platform-tests/pull/9959

### DIFF
--- a/xhr/abort-after-stop.htm
+++ b/xhr/abort-after-stop.htm
@@ -20,8 +20,11 @@
         client.open("GET", "resources/delay.py?ms=3000", true);
         client.send(null);
         test.step_timeout(() => {
-          assert_equals(abortFired, true);
+          test.step(t => assert_equals(abortFired, true));
           test.done();
+
+          // Prevent harness timeout caused by window.stop() preventing "load" event
+          window.dispatchEvent(new Event('load'));
         }, 200);
         window.stop();
       });


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25792)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/352358103)

xhr/abort-after-stop.htm
```diff
@@ -20,8 +20,11 @@
         client.open("GET", "resources/delay.py?ms=3000", true);
         client.send(null);
         test.step_timeout(() => {
-          assert_equals(abortFired, true);
+          test.step(t => assert_equals(abortFired, true));
           test.done();
+
+          // Prevent harness timeout caused by window.stop() preventing "load" event
+          window.dispatchEvent(new Event('load'));
```
annevk: Does this emulate some code in the harness that dispatches a load event? How does this work?

cc @jgraham

AnthumChris: Exactly.  Normally, window.onload is fired, which triggers [`test_environment.add_on_loaded_callback`](https://github.com/w3c/web-platform-tests/blob/master/resources/testharness.js#L1869) to be called.  `window.stop()` prevents that being called, so we invoke it manually

annevk: Thanks. I think @jgraham needs to sign off on this. It seems we might want dedicated API instead so we can identify the tests that need this more easily.

Ms2ger:
Would it also work to move the code into an iframe or something?

zcorpan:
This was fixed in a different way in aa091291c045c0c53706d8541290e496fd8f4dd8

I think a third way to solve this would be to use `setup({explicit_done: true})` and call `done()` at the end.

PR #9976, opened 2018-03-12
Revert "Fix #9481: fix invalid output during environment initialization"

### LINKED ISSUE #9481: WPT CLI: Errors raised during environment initialization cause invalid output

gsnedders:
Reverts w3c/web-platform-tests#9633. This broke [Travis builds for new PRs](https://github.com/w3c/web-platform-tests/pull/9633#issuecomment-372347438).

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9976)

### DIFF
--- a/tools/ci/check_stability.py
+++ b/tools/ci/check_stability.py
@@ -307,81 +307,76 @@ def run(venv, wpt_args, **kwargs):
 
     pr_number = pr()
 
-    try:
-        with TravisFold("browser_setup"):
-            logger.info(markdown.format_comment_title(wpt_args.product))
+    with TravisFold("browser_setup"):
+        logger.info(markdown.format_comment_title(wpt_args.product))
 
-            if pr is not None:
-                deepen_checkout(kwargs["user"])
+        if pr is not None:
+            deepen_checkout(kwargs["user"])
 
-            # Ensure we have a branch called "master"
-            fetch_wpt(kwargs["user"], "master:master")
+        # Ensure we have a branch called "master"
+        fetch_wpt(kwargs["user"], "master:master")
 
-            head_sha1 = get_sha1()
-            logger.info("Testing web-platform-tests at revision %s" % head_sha1)
+        head_sha1 = get_sha1()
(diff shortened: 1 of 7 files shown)


jgraham (approved):

w3c-bots:
# Build PASSED

Started: 2018-03-12 16:03:50
Finished: 2018-03-12 16:17:27





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25797)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/352408692)

PR #9981, opened 2018-03-12
Cleanup gitignore

gsnedders:
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9981)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ node_modules/
 /MANIFEST.json
 
 testharness_runner.html
+!/testharness_runner.html
 !/tools/wptrunner/wptrunner/testharness_runner.html
 
 _certs
--- a/.gitignore
+++ b/.gitignore
@@ -1,20 +1,30 @@
-*#
+# Python
 *.py[co]
-*.sw[po]
-*~
+.virtualenv/
+_venv/
+.cache/
+
+# Node
+node_modules/
+
+# WPT repo stuff
 MANIFEST.json
 !content-security-policy/support/manifest.json
 !payment-handler/manifest.json
-\#*
-_certs
-.virtualenv
-config.json
-node_modules
-scratch
+
 testharness_runner.html
-webdriver/.idea
+!/tools/wptrunner/wptrunner/testharness_runner.html
+
+_certs
+config.json
+
+# Various OS/editor specific files
+*#
+*.sw[po]
+*~
+\#*
+scratch
+.idea/
 .vscode/
 .DS_Store
 *.rej
-_venv
-webdriver/.cache
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .virtualenv/
 _venv/
 .cache/
+.pytest_cache/
 
 # Node
 node_modules/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,20 +1,29 @@
...
+# WPT repo stuff
+/MANIFEST.json
+
+testharness_runner.html
+!/tools/wptrunner/wptrunner/testharness_runner.html
+
+_certs
+config.json
+
+# Various OS/editor specific files
```
gsnedders: Do we want to add a comment about some vague policy about adding stuff here? I mean, I'm assuming we don't just want to add everything from https://github.com/github/gitignore/tree/master/Global into here.

.gitignore
```diff
@@ -1,20 +1,29 @@
...
 *~
-MANIFEST.json
-!content-security-policy/support/manifest.json
-!payment-handler/manifest.json
 \#*
-_certs
-.virtualenv
-config.json
-node_modules
 scratch
```
gsnedders: What even is this? Should this be in the previous section?

.gitignore
```diff
@@ -1,20 +1,29 @@
...
-testharness_runner.html
 webdriver/.idea
 .vscode/
 .DS_Store
 *.rej
-_venv
 webdriver/.cache
```
gsnedders: What even is this? Should this be in the previous section?

w3c-bots:
# Build PASSED

Started: 2018-03-12 17:26:52
Finished: 2018-03-12 17:33:13





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25806)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/352449902)

.gitignore
jgraham: pytest junk

.gitignore
```diff
@@ -1,20 +1,30 @@
...
-!content-security-policy/support/manifest.json
-!payment-handler/manifest.json
 \#*
-_certs
-.virtualenv
-config.json
-node_modules
 scratch
-testharness_runner.html
 webdriver/.idea
```
jgraham: I imagine this should be matched anywhere?

.gitignore
jgraham: No idea

.gitignore
jgraham: "emacs is the one true editor" seems like a good policy to me :p

I expect it's not worth thinking too hard about this and we should just let the section grow in an ad-hoc way.

jgraham (approved):

.gitignore
gsnedders: oh is this where `.pytest_cache` comes from? that, but renamed?

jgraham: Yeah it seems that the name changed at some point, so maybe this rule should be updated.

.gitignore
gsnedders: Seems to be from 8267f079550cbe16e488fa85a04d7d7d309f6529, added along with node. But Google suggests it isn't node, so who knows.

gsnedders:
@jgraham plz merge if this still looks good to you once CI passes

jgraham (approved):

PR #9987, opened 2018-03-13
Ensure cloneNode() does not preserve the prototype

annevk:
Closes https://github.com/whatwg/dom/issues/565.

It would be great to hear from @bigopon since I cannot reproduce this in Chrome. Given that I'm not sure it's worth adding, but it's probably a good "defense in depth" check.

### DIFF
--- a/dom/nodes/Node-cloneNode.html
+++ b/dom/nodes/Node-cloneNode.html
@@ -286,4 +286,17 @@ test(function() {
     assert_equals(copy.childNodes.length, 0,
                   "copy.childNodes.length with non-deep copy");
 }, "node with children");
+
+test(() => {
+  const proto = Object.create(HTMLElement.prototype),
+        node = document.createElement("hi");
+  Object.setPrototypeOf(node, proto);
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25835)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/352672182)

annevk:
It can indeed not be reproduced in Chrome (anymore). I guess this is still worth adding.

foolip:
> It can indeed not be reproduced in Chrome (anymore)

I tried bisecting it using [bisect-builds.py](https://www.chromium.org/developers/bisect-builds-py) going back a year, points on master that became Chrome 60, but the test added here has been passing all along.

Whatever the original problem was, it seems like this test doesn't exercise it. It's not wrong, but also not very meaningful to test for IMHO. I'll review it and let @annevk make the call though.

foolip (approved):

PR #9990, opened 2018-03-13
show() must be triggered by user activation

marcoscaceres:
* actual test for https://github.com/w3c/payment-request/pull/655


payment-request/payment-request-show-method-manual.https.html
```diff
@@ -0,0 +1,105 @@
...
+test(() => {
+  try {
+    new PaymentRequest(defaultMethods, defaultDetails);
+  } catch (err) {
+    done();
+    throw err;
+  }
+}, "Must be possible to construct a payment request");
+
+promise_test(async t => {
```
marcoscaceres: @domenic, I'm a bit unsure if using `setTimeout` to race these two promises is ok? Basically, a conforming implementation would return `acceptPromise` as rejected with a `SecurityError`. However, in a non-conforming browser, calling `.show()` will show the payment sheet, causing the test to get stuck waiting on `acceptPromise` to resolve. 

In order to avoid the test getting stuck, I call .abort() after 100 milliseconds. Is there a better was of doing this?

domenic: Generally we let tests timeout in non-conforming browsers. That way, their CI infrastructure can control the timeout period, instead of getting it hard-coded to 100 ms as here.

marcoscaceres: Makes sense. I’ll split this into manual and automated test files. Then the automated can time out.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25898)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/353125535)

rsolomakhin (approved):

PR #9993, opened 2018-03-13
MathML: Test italic correction of sub and super scripts attached to a large operator

fred-wang:
This test italic correction for subsup script (tests for underover should be added in the future). The test passes on WebKit and will replace https://trac.webkit.org/browser/webkit/trunk/LayoutTests/mathml/opentype/large-operators-italic-correction.html (which requires Latin Modern Math to be installed on the WebKit bots).


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25848)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/352758553)

fred-wang:
screenshot-italic-correction

This is how the test is rendered in Gecko (left) and WebKit (right). Note however that the position of script is wrong in Gecko for the nonzero-italic case (second row) and the rendering of the operator is bad in WebKit in the zero-italic case (first row). WebKit still passes the test but not Gecko.

- largeop-displayoperatorminheight5000 has a large glyph for U+2AFF N-ARY WHITE VERTICAL BAR that has null italic correction (it is the rectangle displayed in the first row of the Gecko screenshot)
- The change in largeop.py adds a new file largeop-displayoperatorminheight2000-2AFF-italiccorrection3000 has a large glyph for U+2AFF N-ARY WHITE VERTICAL BAR with italic correction of 3000 (it is the inclined parallelogram displayed in the second row of the WebKit screenshot).
- minheight values are just to ensure web engines will pick the large glyph of  U+2AFF and not the base size.

I pushed a commit to better explain the tests.

cc'ing @davidcarlisle @karlt @khaledhosny for review.

mathml/presentation-markup/scripts/subsup-parameters-2.html
```diff
@@ -0,0 +1,131 @@
...
+    font-family: largeop-displayoperatorminheight5000;
+    src: url("/fonts/math/largeop-displayoperatorminheight5000.woff");
+  }
+  @font-face {
+    font-family: largeop-displayoperatorminheight2000-2AFF-italiccorrection3000;
+    src: url("/fonts/math/largeop-displayoperatorminheight2000-2AFF-italiccorrection3000.woff");
+  }
+</style>
+<script>
+  var emToPx = 10 / 1000; // font-size: 10px, font.em = 1000
```
mrego: Can we move this closer to where it's used?

mathml/presentation-markup/scripts/subsup-parameters-2.html
```diff
@@ -0,0 +1,131 @@
...
+  var epsilon = 1;
```
mrego: Ditto.

mathml/presentation-markup/scripts/subsup-parameters-2.html
```diff
@@ -0,0 +1,131 @@
...
+    - In msubsup, the scripts are shifted by the italic correction.
+    - In mmultiscripts, postscript pairs are shifted by the italic correction.
+    - In mmultiscripts, prescript pairs are vertically aligned.
+  */
+  function runTests() {
+    test(function() {
+      var v = 0;
+      assert_approx_equals(getBox("base001").right - getBox("sub001").left, v, epsilon, "msub");
+      assert_approx_equals(getBox("sup002").left, getBox("base002").right, epsilon, "msup");
+      assert_approx_equals(getBox("sup003").left - getBox("sub003").left, v, epsilon, "msubsup");
```
mrego: If both are wrongly positioned, but with the same `left` this will pass.
Wouldn't be better to always compare against "base00X"?
Dunno if that's possible when you the italic correction is applied.

mathml/presentation-markup/scripts/subsup-parameters-2.html
```diff
@@ -0,0 +1,131 @@
...
+      assert_approx_equals(getBox("sup012").left, getBox("base012").right, epsilon, "msup");
+      assert_approx_equals(getBox("sup013").left - getBox("sub013").left, v, epsilon, "msubsup");
+      assert_approx_equals(getBox("sup014").left - getBox("sub014").left, v, epsilon, "mmultiscripts postscripts");
+      assert_approx_equals(getBox("sup015").left - getBox("sub015").left, 0, epsilon, "mmultiscripts prescripts");
+    }, "NonNull Italic Correction");
+    done();
+  }
+</script>
+</head>
+<body>
```
mrego: Nit: Dunno if this is common or not on MathML test suite, so feel free to ignore these suggestions:
  * I'd add a `<div id="log"></div>` on the top of the test, so when you open it the testharness output will be on the top.
  * I'd add a title before each section (like `<h2>Null Italic Correction</h2>` and `<h2>NonNull Italic Correction</h2>`).

mathml/presentation-markup/scripts/subsup-parameters-2.html
fred-wang: Done.

mathml/presentation-markup/scripts/subsup-parameters-2.html
fred-wang: Done

mathml/presentation-markup/scripts/subsup-parameters-2.html
fred-wang: Yes, however I'm not sure that's a big deal to "wrongly pass" (Chromium and Edge have many MathML tests passing on wpt.fyi ; just because they succeeded to load the test !). I think it's clearer and more natural to test the italic correction by checking the difference of position of the two scripts. I could add extra tests to check the horizontal position but that kind of tests is already (and more completely done) done by http://w3c-test.org/mathml/presentation-markup/scripts/subsup-parameters-1.html

mathml/presentation-markup/scripts/subsup-parameters-2.html
fred-wang: Yes, I never liked that the test results don't show at the top. I actually think it's a good idea and should be done for MathML tests too.

mathml/presentation-markup/scripts/subsup-parameters-2.html
mrego: The goal of a test is to fail when something is not supported, so it'd be better if they fail.
Anyway I understand we have other tests to check the vertical position of the scripts and also the size of the operator (otherwise we need them).
So the rest can be good enough, we're at least checking against "base00X" in some cases so we'll should catch some issues with it.

mrego (approved):

PR #9995, opened 2018-03-13
Make lint run when directories are passed in

teh-f00l:
Response to issue https://github.com/w3c/web-platform-tests/issues/9879

Allows tester to pass command line arguments to lint specifying directories they want checked

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/9995)

### DIFF
--- a/tools/lint/lint.py
+++ b/tools/lint/lint.py
@@ -816,7 +816,7 @@ def lint(repo_root, paths, output_format):
     last = None
 
     with open(os.path.join(repo_root, "lint.whitelist")) as f:
-        whitelist, ignored_files = parse_whitelist(f)
+        whitelist, ignored_files=parse_whitelist(f)
 
     output_errors = {"json": output_errors_json,
                      "markdown": output_errors_markdown,
--- a/tools/lint/lint.py
+++ b/tools/lint/lint.py
@@ -59,9 +59,13 @@ you could add the following line to the lint.whitelist file.
 
 %s: %s"""
 
-def all_filesystem_paths(repo_root, extended_path):
+def all_filesystem_paths(repo_root, subdir = None):
     path_filter = PathFilter(repo_root, extras=[".git/*"])
-    for dirpath, dirnames, filenames in os.walk(extended_path):
+    if subdir:
+        expanded_path = subdir
+    else:
+        expanded_path = repo_root
+    for dirpath, dirnames, filenames in os.walk(expanded_path):
         for filename in filenames:
             path = os.path.relpath(os.path.join(dirpath, filename), repo_root)
             if path_filter(path):
@@ -754,10 +758,10 @@ def lint_paths(kwargs, wpt_root):
             # create extension for called directory and merge it with the existing path
                 new_path = os.path.realpath(os.path.join(wpt_root, path))
                 # get all files from new path
-                path_dir = list(all_filesystem_paths(wpt_root, new_Path))
+                path_dir = list(all_filesystem_paths(wpt_root, new_path))
              
                 # add all files from a given directory to our master list of files to check
-                paths = paths + path_dir
+                paths.extend(path_dir)
 
     elif kwargs["all"]:
         paths = list(all_filesystem_paths(wpt_root))
--- a/tools/lint/lint.py
+++ b/tools/lint/lint.py
@@ -754,8 +754,7 @@ def lint_paths(kwargs, wpt_root):
         paths = []
         for path in kwargs.get("paths"):
             if os.path.isdir(path):
-                new_path = os.path.realpath(os.path.join(wpt_root, path))
-                path_dir = list(all_filesystem_paths(wpt_root, new_path))
+                path_dir = list(all_filesystem_paths(wpt_root, path))
                 paths.extend(path_dir)
             elif os.path.isfile(path):
                 paths.extend(path)
--- a/tools/lint/lint.py
+++ b/tools/lint/lint.py
@@ -27,7 +27,7 @@ import logging
 
 logger = None
 
-def setup_logging(prefix = False):
+def setup_logging(prefix=False):
     global logger
     if logger is None:
         logger = logging.getLogger(os.path.basename(os.path.splitext(__file__)[0]))
@@ -59,7 +59,7 @@ you could add the following line to the lint.whitelist file.
 
 %s: %s"""
 
-def all_filesystem_paths(repo_root, subdir = None):
+def all_filesystem_paths(repo_root, subdir=None):
     path_filter = PathFilter(repo_root, extras = [".git/*"])
     if subdir:
         expanded_path = subdir
--- a/tools/lint/lint.py
+++ b/tools/lint/lint.py
@@ -60,7 +60,7 @@ you could add the following line to the lint.whitelist file.
 %s: %s"""
 
 def all_filesystem_paths(repo_root, subdir=None):
-    path_filter = PathFilter(repo_root, extras = [".git/*"])
+    path_filter = PathFilter(repo_root, extras=[".git/*"])
     if subdir:
         expanded_path = subdir
     else:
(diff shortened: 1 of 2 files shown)


tools/lint/lint.py
```diff
@@ -747,8 +747,19 @@ def changed_files(wpt_root):
 
 def lint_paths(kwargs, wpt_root):
     if kwargs.get("paths"):
-        r = os.path.realpath(wpt_root)
-        paths = [os.path.relpath(os.path.realpath(x), r) for x in kwargs["paths"]]
+        #make new list in case there are multiple added directories
+        AllPaths=[]
```
jgraham: Use `snake_case` for variable names.

tools/lint/lint.py
```diff
@@ -747,8 +747,19 @@ def changed_files(wpt_root):
...
+        for path in kwargs.get("paths"):
+            # create extension for called directory and merge it with the existing path
+            extension= path+'/'
```
jgraham: Use a space before and after operators like `=`.

tools/lint/lint.py
```diff
@@ -747,8 +747,19 @@ def changed_files(wpt_root):
...
+            newPath=(str(os.path.realpath(wpt_root))+'/'+ extension)
```
jgraham: So I feel like this is basically
`abs_path = os.path.realpath(os.path.join(wpt_root, path))`?

tools/lint/lint.py
```diff
@@ -747,8 +747,19 @@ def changed_files(wpt_root):
...
+            # get all files from new path
+            paths = list(all_filesystem_paths(newPath))
```
jgraham: Do we first want to check if it's a directory?

tools/lint/lint.py
```diff
@@ -747,8 +747,19 @@ def changed_files(wpt_root):
...
+            # add the extension to the strings for all files from the new path,
+            # so they may be called from the current directory
+            paths=list(map(lambda path : extension + path, paths))
```
jgraham: A list comprehension is more idomatic here, so this would look like `[extension + path for path in paths]`. That said, it isn't clear what you're doing here. In the clause below the results of `all_filesystem_paths` are used directly, so it seems like it should be fine to do that here too.

jgraham (changes requested):
This looks like a great start, but some changes are required before it can land.

w3c-bots:
# Build BROKEN

Started: 2018-03-13 13:27:45
Finished: 2018-03-13 13:42:38


<h2>Failing Jobs</h2>
<ul>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/25855)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/352836660)

tools/lint/lint.py
teh-f00l: I tried that initially, but it was insufficient because it assumes we are already in the directory passed to ```all_filesystem_paths``` and will not work unless we are already operating from that directory

tools/lint/lint.py
teh-f00l: kk

tools/lint/lint.py
teh-f00l: ah my bad, I got rid of my if statement. Putting that back in

teh-f00l: Although I am not sure if relpath was sufficient or not

tools/lint/lint.py
```diff
@@ -747,8 +747,21 @@ def changed_files(wpt_root):
...
 def lint_paths(kwargs, wpt_root):
     if kwargs.get("paths"):
-        r = os.path.realpath(wpt_root)
-        paths = [os.path.relpath(os.path.realpath(x), r) for x in kwargs["paths"]]
+        #make new list in case there are multiple added directories
+        paths=[]
+        for path in kwargs.get("paths"):
+            if os.path.isdir(path):
+            # create extension for called directory and merge it with the existing path
+                new_Path = os.path.realpath(os.path.join(wpt_root, path))
```
jgraham: Variable names  should be all lowercase.

tools/lint/lint.py
```diff
@@ -747,8 +747,21 @@ def changed_files(wpt_root):
...
+                extension = path+'/'
```
jgraham: So this should use `os.path.sep`, but  I'm not sure why it should be necessary.

tools/lint/lint.py
```diff
@@ -747,8 +747,21 @@ def changed_files(wpt_root):
 
 def lint_paths(kwargs, wpt_root):
     if kwargs.get("paths"):
-        r = os.path.realpath(wpt_root)
-        paths = [os.path.relpath(os.path.realpath(x), r) for x in kwargs["paths"]]
+        #make new list in case there are multiple added directories
+        paths=[]
```
jgraham: Spaces between operators and operands, please.

jgraham (changes requested):

tools/lint/lint.py
```diff
@@ -747,8 +747,18 @@ def changed_files(wpt_root):
 
 def lint_paths(kwargs, wpt_root):
     if kwargs.get("paths"):
-        r = os.path.realpath(wpt_root)
-        paths = [os.path.relpath(os.path.realpath(x), r) for x in kwargs["paths"]]
+        #make new list in case there are multiple added directories
+        paths = []
+        for path in kwargs.get("paths"):
+            if os.path.isdir(path):
```
jgraham: I think this  `if` is missing an else clause to just add the path to the list of paths.

tools/lint/lint.py
```diff
@@ -747,8 +747,18 @@ def changed_files(wpt_root):
...
+            # create extension for called directory and merge it with the existing path
```
jgraham: This  indent is wrong.

tools/lint/lint.py
```diff
@@ -59,9 +59,9 @@ def setup_logging(prefix=False):
 
 %s: %s"""
 
-def all_filesystem_paths(repo_root):
+def all_filesystem_paths(repo_root, extended_path):
```
jgraham: "`extended_path`" needs to be optional i.e. the signature should be

`all_filesystem_paths(repo_root, subdir=None):`

Also it should be relative  to `repo_root` so that in the function we do `root = os.path.join(repo_root, subdir) if subdir else repo_root` or something  similar.

tools/lint/lint.py
```diff
@@ -747,8 +747,18 @@ def changed_files(wpt_root):
...
-        r = os.path.realpath(wpt_root)
-        paths = [os.path.relpath(os.path.realpath(x), r) for x in kwargs["paths"]]
+        #make new list in case there are multiple added directories
+        paths = []
+        for path in kwargs.get("paths"):
+            if os.path.isdir(path):
+            # create extension for called directory and merge it with the existing path
+                new_path = os.path.realpath(os.path.join(wpt_root, path))
+                # get all files from new path
+                path_dir = list(all_filesystem_paths(wpt_root, new_Path))
```
jgraham: Your capitalisation is off here.

tools/lint/lint.py
```diff
@@ -747,8 +747,18 @@ def changed_files(wpt_root):
...
 def lint_paths(kwargs, wpt_root):
     if kwargs.get("paths"):
-        r = os.path.realpath(wpt_root)
-        paths = [os.path.relpath(os.path.realpath(x), r) for x in kwargs["paths"]]
+        #make new list in case there are multiple added directories
+        paths = []
+        for path in kwargs.get("paths"):
+            if os.path.isdir(path):
+            # create extension for called directory and merge it with the existing path
+                new_path = os.path.realpath(os.path.join(wpt_root, path))
```
jgraham: So this assumes that `path` is relative to the `repo_root` but above it's assumed to be relative to the current working directory. I think the latter is correct.

tools/lint/lint.py
```diff
@@ -747,8 +747,18 @@ def changed_files(wpt_root):
...
+                # get all files from new path
+                path_dir = list(all_filesystem_paths(wpt_root, new_Path))
+             
+                # add all files from a given directory to our master list of files to check
+                paths = paths + path_dir
```
jgraham: I think `paths.extend(path_dir)` is more obvious

jgraham (changes requested):

tools/lint/lint.py
teh-f00l: The later is correct. It should have been correct here as well, new_path should be the former.

jgraham (approved):

PR #10018, opened 2018-03-13
ReadableStream: Readable byte stream must call pull if needed after receiving new chunk

MattiasBuelens:
This adds a test for the new behavior to be introduced by whatwg/streams#904. More tests will follow later.

This also changes two other tests:
* _ReadableStream with byte source: autoAllocateChunkSize_  
This test previously expected `pull()` to only be called once, although `read()` is called twice. This was incorrect: `pull()` should have been called again to fulfill the second `read()`. The fixed test verifies that the second `read` is correctly fulfilled.
* _ReadableStream with byte source: Respond to pull() by enqueue() asynchronously_  
This test previously expected `pull()` to only be called once, but the amount of data enqueued by one `pull()` is not sufficient to fill the stream's queue up to its `highWaterMark` of `256`. This is now incorrect: the stream would need to repeatedly call pull until it has reached the `highWaterMark`. The fixed test sets the `highWaterMark` to `0` such that `pull()` does not get called repeatedly.

### DIFF
--- a/streams/readable-byte-streams/general.js
+++ b/streams/readable-byte-streams/general.js
@@ -693,7 +693,7 @@ promise_test(() => {
     },
     type: 'bytes'
   }, {
-    highWaterMark: 256
+    highWaterMark: 0
   });
 
   const reader = stream.getReader();
@@ -717,9 +717,9 @@ promise_test(() => {
     assert_equals(result[2].done, false, 'result[2].done');
     assert_equals(result[2].value.byteLength, 1, 'result[2].value.byteLength');
     assert_equals(byobRequest, undefined, 'byobRequest should be undefined');
-    assert_equals(desiredSizes[0], 256, 'desiredSize on pull should be 256');
-    assert_equals(desiredSizes[1], 256, 'desiredSize after 1st enqueue() should be 256');
-    assert_equals(desiredSizes[2], 256, 'desiredSize after 2nd enqueue() should be 256');
+    assert_equals(desiredSizes[0], 0, 'desiredSize on pull should be 0');
+    assert_equals(desiredSizes[1], 0, 'desiredSize after 1st enqueue() should be 0');
+    assert_equals(desiredSizes[2], 0, 'desiredSize after 2nd enqueue() should be 0');
     assert_equals(pullCount, 1, 'pull() should only be called once');
   });
 }, 'ReadableStream with byte source: Respond to pull() by enqueue() asynchronously');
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25892)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/353074717)

streams/readable-byte-streams/general.js
```diff
@@ -300,22 +300,28 @@ function extractViewInfo(view) {
 promise_test(() => {
   let pullCount = 0;
   let controller;
-  let byobRequest;
-  let viewDefined = false;
-  let viewInfo;
+  let byobRequests = [];
```
ricea: You can use "const" instead of "let" here.

I write the tests for streams in the same style as the streams repository, as I think it makes them easier to navigate. I have a more relaxed version of the eslint config from whatwg/streams that I use to check my changes. Unfortunately there doesn't seem to be any useful place where I could check that in.

streams/readable-byte-streams/general.js
```diff
@@ -326,28 +332,50 @@ promise_test(() => {
...
+    const viewInfo = byobRequest.viewInfo;
+    assert_equals(viewInfo.constructor, Uint8Array, 'first view.constructor should be Uint8Array');
+    assert_equals(viewInfo.bufferByteLength, 16, 'first view.buffer.byteLength should be 16');
+    assert_equals(viewInfo.byteOffset, 0, 'first view.byteOffset should be 0');
+    assert_equals(viewInfo.byteLength, 16, 'first view.byteLength should be 16');
+
+    return p0;
   }).then(result => {
+    assert_equals(pullCount, 2, 'pull() must have been invoked twice');
     assert_not_equals(result.value, undefined);
```
ricea: If you have the energy, it would be nice to have descriptions for these asserts. But don't worry about it, because it was already like this.

ricea (approved):
Great quality and coverage, thanks!

streams/readable-byte-streams/general.js
MattiasBuelens: Done.

---

I'm not sure if I like the assert message to restate the entire check, as in:
```js
assert_equals(viewInfo.byteOffset, 0, 'first view.byteOffset should be 0');
```
If the assert fails, you already get the expected value in the error message anyway:
```
assert_equals: first view.byteOffset should be 0 expected 0 but got 42
```
Perhaps in these cases, we should just describe the value being tested without repeating the expected value? I kind of like the terseness used in some other tests:
```js
assert_equals(view.byteOffset, 0, 'view.byteOffset');
```

ricea: I make a habit of putting the word 'should' in my descriptions as a way of forcing myself to be clear about what I'm asserting. But certainly in cases such as this it doesn't help readability. The important thing is to have a distinct message, so that someone looking at a failure doesn't have to guess which assert failed.

PR #10025, opened 2018-03-14
Adjust MathML tests to Workaround WebKit's bug with document.fonts.ready

fred-wang:
cc @mrego cc @youennf 

That seems to be an acceptable work around to me. WDYT?

See also https://bugs.webkit.org/show_bug.cgi?id=183628

### DIFF
--- a/mathml/presentation-markup/fractions/frac-parameters-1.html
+++ b/mathml/presentation-markup/fractions/frac-parameters-1.html
@@ -63,7 +63,7 @@
   setup({ explicit_done: true });
   window.addEventListener("load", function() {
     // Delay the check to workaround WebKit's bug https://webkit.org/b/174030.
-    setTimeout(() => { document.fonts.ready.then(runTests); }, 0);
+    requestAnimationFrame(() => { document.fonts.ready.then(runTests); });
   });
 
   function runTests() {
--- a/mathml/presentation-markup/fractions/frac-parameters-1.html
+++ b/mathml/presentation-markup/fractions/frac-parameters-1.html
@@ -62,7 +62,8 @@
 
   setup({ explicit_done: true });
   window.addEventListener("load", function() {
-    document.fonts.ready.then(runTests);
+    // Delay the check to workaround WebKit's bug https://webkit.org/b/174030.
+    setTimeout(() => { document.fonts.ready.then(runTests); }, 0);
   });
 
   function runTests() {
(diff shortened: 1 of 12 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/25928)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/353287128)

mathml/presentation-markup/fractions/frac-parameters-1.html
```diff
@@ -62,7 +62,8 @@
 
   setup({ explicit_done: true });
   window.addEventListener("load", function() {
-    document.fonts.ready.then(runTests);
+    // Delay the check to workaround WebKit's bug https://webkit.org/b/174030.
+    setTimeout(() => { document.fonts.ready.then(runTests); }, 0);
```
mrego: I think it's usually preferred `window.requestAnimationFrame()` instead of `setTimeout()`.
Could you try if something like that would fix the issue? Thanks!

fred-wang: The WebKit bug is that `document.fonts.ready` resolves immediately, even when I replace the CSS @font-face with a list of document.fonts.add(...) just before document.fonts.ready. Using setTimeout or  requestAnimationFrame will delay things a bit and hence solve the issue, but I'm not sure requestAnimationFrame is more appropriate here, since it's not really a layout/painting issue but a font API bug. For example the following changes do *not* allow to workaround the issue:

    document.fonts.ready.then(() => { requestAnimationFrame(runTests); });
or 
   document.fonts.ready.then(() => { setTimeout(runTests, notSmallEnoughDelta); });

mrego: I was thinking in something like:
```js
    window.requestAnimationFrame(() => { document.fonts.ready.then(runTests); });
```

However I agree with you that `setTimeout()` might be clearer, so I'm fine with keeping it.

mrego (approved):
LGTM

mathml/presentation-markup/fractions/frac-parameters-1.html
fred-wang: Yes, `requestAnimationFrame(() => { document.fonts.ready.then(runTests); });` would work too.

PR #10047, opened 2018-03-15
[web-animations] Update timing interfaces

birtles:
These are the tests for https://github.com/w3c/csswg-drafts/pull/2432


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26010)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/353693275)

web-animations/interfaces/Animatable/getAnimations.html
```diff
@@ -91,13 +93,15 @@
   const div = createDiv(t);
   const animation = div.animate(null, 100 * MS_PER_SEC);
 
-  animation.effect.timing.endDelay = -200 * MS_PER_SEC;
+  animation.effect.updateTiming({
+    endDelay: animation.effect.getTiming().endDelay - 200 * MS_PER_SEC,
```
stephenmcgruer: I don't follow this change. The previous version just set it to `-200 * MS_PER_SEC`, but this version decrements the existing value by `200 * MS_PER_SEC` - why ?

stephenmcgruer (approved):
Generally LGTM, a bit difficult to review such a large change. One comment that I don't quite follow.

I notice from this that the new API also makes doing 'double the duration' awkward which is a pity :(.

web-animations/interfaces/Animatable/getAnimations.html
birtles: That's a bug in the patch. Thanks!

birtles:
> I notice from this that the new API also makes doing 'double the duration' awkward which is a pity :(.

Yeah, that operation is definitely harder. I wonder in practice how often you do that, however, and how big a deal it is at the time anyway (React programmers anyway are used to this sort of pattern with `setState()` etc.)

PR #10055, opened 2018-03-15
Add tentative test for what ".." on documentElement returns

gsnedders:
As far as I can tell, from both DOM Level 3 XPath and XPath 1.0, this is undefined. As far as the XPath data model is concerned, documentElement can have no parent, and DOM Level 3 XPath doesn't address this.

However, at least Chrome/Edge/Firefox/Safari all pass this test, so it seems like something we should test.

### DIFF
--- a/domxpath/document.tentative.html
+++ b/domxpath/document.tentative.html
@@ -1,0 +1,20 @@
+<!doctype html>
+<title>XPath parent of documentElement</title>
+<script src='/resources/testharness.js'></script>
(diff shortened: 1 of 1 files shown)


Ms2ger (approved):
Please add a note to <https://wiki.whatwg.org/wiki/DOM_XPath>.

gsnedders:
@Ms2ger writing that up makes me realise we need tests for root adjacent comments, PIs, and doctypes (esp. doctypes, given XPath has no notion of them).

PR #10058, opened 2018-03-15
Changing WebDriver test_parent_htmldocument to expect error

jimevans:
When the context element is `document.documentElement`, and an attempt is
made to find elements from that context element using an XPath of `..`, a
snapshot is returned containing the `document` object. While this is
apparently the correct behavior for XPath, the WebDriver spec says that if
the object in the snapshot is not an element, we should return an error
with error code "invalid selector." The test_parent_htmldocument test in
both find_element_from_element.py and find_elements_from_element.py expect
a success in this case. This commit changes the tests to correctly expect
an error from the driver.

### DIFF
--- a/webdriver/tests/element_retrieval/find_element_from_element.py
+++ b/webdriver/tests/element_retrieval/find_element_from_element.py
@@ -79,4 +79,4 @@ def test_parent_htmldocument(session):
     from_element = session.execute_script("return document.documentElement")
 
     response = find_element(session, from_element.id, "xpath", "..")
-    assert_success(response)
+    assert_error(response, "invalid selector")
--- a/webdriver/tests/element_retrieval/find_element_from_element.py
+++ b/webdriver/tests/element_retrieval/find_element_from_element.py
@@ -76,6 +76,16 @@ def test_xhtml_namespace(session, using, value):
 
 def test_parent_htmldocument(session):
     session.url = inline("")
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26047)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/354017317)

andreastt (changes requested):
You are correct that the XPath selector `..` on the `<html>` element (the document element) should return an invalid selector error. This is consistent with the behaviour described in https://bugzilla.mozilla.org/show_bug.cgi?id=1424635#c0.

That said, I don’t think this is the correct patch for the issue. `test_parent_htmldocument` tests that it is possible to retrieve a parent HTMLDocument node that is `Node.DOCUMENT_NODE`. We have to test this specifically because the `nodeType` differs from other regular elements, and this matters in implementations such as Gecko.

I think we should fix this test to use `<body>` as the `from_element`, then write a new test for getting the relative ancestor of `<html>` and returning an invalid selector error.

Does that make sense?

andreastt:
See also https://hg.mozilla.org/mozilla-central/rev/3870805aea81 for the background for why we have this test.

andreastt (approved):

PR #10059, opened 2018-03-15
Editing expectations of domain attribute of returned cookie

jimevans:
For the domain attribute of a returned cookie, if the attribute value
conatains a domain with a single dot ('.'), some user agents prepend a
leading dot onto the beginning of the attribute value. This is consistent
with RFC 2965. Some user agents expressly omit the leading dot, which is
consistent with the later RFC 6265. This commit allows both values as the
returned value of the domain attribute.

### DIFF
--- a/webdriver/tests/cookies/add_cookie.py
+++ b/webdriver/tests/cookies/add_cookie.py
@@ -36,7 +36,7 @@ def test_add_domain_cookie(session, url, server_config):
 
     assert cookie["name"] == "hello"
     assert cookie["value"] == "world"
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26051)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/354054835)

andreastt (approved):
I’m somewhat surprised this passes the Python lints.

PR #10094, opened 2018-03-19
Verify that fetch API `force-cache` and `only-if-cached` respect no-s…

wanderview:
…tore/no-cache headers.

### DIFF
--- a/fetch/api/request/request-cache-force-cache.html
+++ b/fetch/api/request/request-cache-force-cache.html
@@ -71,6 +71,15 @@
         expected_validation_headers: [false],
         expected_no_cache_headers: [false],
       },
+      {
+        name: 'RequestCache "force-cache" mode checks the cache for previously cached content and goes to the network if previous response was no-store',
+        state: "stale",
+        cache_control: "no-store",
+        pragma: "no-cache",
+        request_cache: ["default", "force-cache"],
+        expected_validation_headers: [false, false],
+        expected_no_cache_headers: [false, false],
+      },
     ];
     run_tests(tests);
     </script>
--- a/fetch/api/request/request-cache-only-if-cached.html
+++ b/fetch/api/request/request-cache-only-if-cached.html
@@ -69,6 +69,16 @@
         expected_validation_headers: [false, false],
         expected_no_cache_headers: [false, false],
       },
+      {
+        name: 'RequestCache "only-if-cached" mode checks the cache for previously cached content and does not go to the network if a response was no-store',
+        state: "stale",
+        cache_control: "no-store",
+        pragma: "no-cache",
+        request_cache: ["default", "only-if-cached"],
+        response: [null, "error"],
+        expected_validation_headers: [false],
+        expected_no_cache_headers: [false]
+      },
     ];
     run_tests(tests);
     </script>
--- a/fetch/api/request/request-cache.js
+++ b/fetch/api/request/request-cache.js
@@ -64,6 +64,10 @@ function make_url(uuid, id, value, content, info) {
   if ("cache_control" in info) {
     cache_control = "&cache_control=" + info.cache_control;
   }
+  var pragma = "";
+  if ("pragma" in info) {
+    pragma = "&Pragma=" + info.pragma;
+  }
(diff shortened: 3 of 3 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26178)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/355393362)

fetch/api/request/request-cache-force-cache.html
```diff
@@ -71,6 +71,15 @@
         expected_validation_headers: [false],
         expected_no_cache_headers: [false],
       },
+      {
+        name: 'RequestCache "force-cache" mode checks the cache for previously cached content and goes to the network if previous response was no-store',
+        state: "stale",
+        cache_control: "no-store",
+        pragma: "no-cache",
+        request_cache: ["default", "force-cache"],
```
youennf: The case stale/no-store/default is probably already covered in fetch/api/request/request-cache-default.html.
We probably can remove it.

fetch/api/request/request-cache-only-if-cached.html
```diff
@@ -69,6 +69,16 @@
         expected_validation_headers: [false, false],
         expected_no_cache_headers: [false, false],
       },
+      {
+        name: 'RequestCache "only-if-cached" mode checks the cache for previously cached content and does not go to the network if a response was no-store',
+        state: "stale",
+        cache_control: "no-store",
+        pragma: "no-cache",
+        request_cache: ["default", "only-if-cached"],
```
youennf: Ditto here

fetch/api/request/request-cache.js
```diff
@@ -75,7 +79,7 @@ function make_url(uuid, id, value, content, info) {
     "&content=" + content +
     "&" + id + "=" + value +
     "&expires=" + dates[info.state] +
-    vary + cache_control + ignore_request_headers;
+    vary + cache_control + pragma + ignore_request_headers;
```
youennf: It is not clear to me what you are trying to do here with Pragma.
cache.py is grabbing the Pragma request header and storing it in its state so that it can later be validated by the JS test. Adding it in the URL should not change anything.
We are adding Cache-Control in the URL so that the response will contain Cache-Control header value.

fetch/api/request/request-cache-force-cache.html
wanderview: The purpose of this added test case is that `force-cache` does not override the browsers behavior of suppressing no-store responses that may be hanging around in the browser.  For example, the browser may still have this data resident to support devtools viewsource, etc.  Setting `force-cache` should not allow content to observe these `no-store` responses.

fetch/api/request/request-cache.js
wanderview: It was my understanding that to get a response to be consistently ignored by browser caches you needed to set both of these headers:

```
cache-control: no-store
pragma: no-cache
```

This diff hunk was about getting the pragma header set on the response toward this end.

fetch/api/request/request-cache-force-cache.html
youennf: Sure, instead of request_cache: ["default", "force-cache"], request_cache: ["force-cache"] should be sufficient.

fetch/api/request/request-cache.js
wanderview: Oh, I see.  I saw that pragma was being read from the request in `cache.py` and I guess I thought it was reflected back.

Since the test is adequate to cover the situation I'm trying to test without the pragma stuff actually working, then I guess I'll just remove it.  Looking closer, it seems its fine for a server to simply return `cache-control: no-store` in the response header.

fetch/api/request/request-cache-force-cache.html
wanderview: Don't I need the "default" there so that the http cache has a chance to see the no-store response?  Or does the test infrastructure prime cache in some other way?  Thanks.

ahmadmiz:
Thank you

PR #10030, opened 2018-03-14
Exit wptrunner with an error status if no tests ran

jgraham:
Typically no tests completing means that some error occured, or the user
supplied an invalid test path. In either case we should log the
problem and return a failure code.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10030)

Issue #9932, opened 2018-03-08
Chrome stability checks should block PR merges


Hexcles:
We have allowed failures in the Chrome job on Travis for a while due to bugs/instability.

We should investigate the current issues (e.g. it seems like tests aren't even running on Chrome at all, but the job is always green), make it reliable, and flip the switch. This will greatly reduce the flakiness that Chromium importer has to deal with.

Hexcles:
cc @foolip we talked about this earlier.

Hexcles:
A quick search gives these potentially related open issues: #5407 #6293

Hexcles:
#10064 has been merged. There should be no other infra issues AFAIK. I'd like to watch the Travis jobs for a few days before flipping the switch to make Chrome stability check blocking again.

PR #10098, opened 2018-03-19
Exit check_stability with an error when no tests run

Hexcles:
Similar to #10030 (which was for `wpt run`), this commit makes `wpt
check-stability` return a non-zero exit code when no tests run.

Also fix some code smells:

* `run` was redefined by `do_delayed_imports` to a different function.
  Rename the other `run` method in this module to `setup_and_run`.
* `retcode` was unnecessarily defined in the top scope. Remove it to
  avoid redefinition.
* `global logger` and the call to `get_parser` were extraneous in `run`
  (now `setup_and_run`) and hence are removed.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10098)

### DIFF
--- a/tools/ci/check_stability.py
+++ b/tools/ci/check_stability.py
@@ -22,7 +22,7 @@ from tools.wpt import markdown
 from tools import localpaths
 
 logger = None
-run, write_inconsistent, write_results = None, None, None
+stability_run, write_inconsistent, write_results = None, None, None
 wptrunner = None
 
 def setup_logging():
@@ -37,8 +37,9 @@ def setup_logging():
 
 
 def do_delayed_imports():
-    global run, write_inconsistent, write_results, wptrunner
-    from tools.wpt.stability import run, write_inconsistent, write_results
+    global stability_run, write_inconsistent, write_results, wptrunner
+    from tools.wpt.stability import run as stability_run
+    from tools.wpt.stability import write_inconsistent, write_results
     from wptrunner import wptrunner
 
 
@@ -250,10 +251,10 @@ def main():
     venv.install("requests")
 
     args, wpt_args = get_parser().parse_known_args()
-    return setup_and_run(venv, wpt_args, **vars(args))
+    return run(venv, wpt_args, **vars(args))
 
 
-def setup_and_run(venv, wpt_args, **kwargs):
+def run(venv, wpt_args, **kwargs):
     do_delayed_imports()
 
     retcode = 0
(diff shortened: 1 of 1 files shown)


Hexcles:
One more step towards #9932 .

Hexcles:
One example: https://travis-ci.org/w3c/web-platform-tests/jobs/355393374

This job didn't run any tests but the was still green.

tools/ci/check_stability.py
```diff
@@ -250,16 +250,13 @@ def main():
...
 
     args, wpt_args = get_parser().parse_known_args()
-    return run(venv, wpt_args, **vars(args))
+    return setup_and_run(venv, wpt_args, **vars(args))
 
 
-def run(venv, wpt_args, **kwargs):
-    global logger
-
+def setup_and_run(venv, wpt_args, **kwargs):
```
jgraham: This will break the `wpt check-stability` command since it uses `run` as the entry point.

tools/ci/check_stability.py
```diff
@@ -353,16 +350,15 @@ def run(venv, wpt_args, **kwargs):
                              status="failed" if inconsistent else "passed")
     else:
         logger.info("No tests run.")
+        retcode = 3
```
jgraham: Are we sure this never happens in practice? I'm slightly concerned that maybe we sometimes pass this a list of things that could be tests but actually aren't so nothing would actually run. I *think* it should be OK but there might be an edge case like a `.py` file under `webdriver` that doesn't have any `test` functions, or something.

jgraham (changes requested):

w3c-bots:
# Build PASSED

Started: 2018-03-19 15:54:38
Finished: 2018-03-19 16:08:15





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26186)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/355425369)

tools/ci/check_stability.py
Hexcles: Renamed the other run to `stability_run` instead.

tools/ci/check_stability.py
Hexcles: No you're right. It's totally possible to have `tests_changed` empty but `files_affected` not, though I'm not sure if the scenario you described would be an edge case (not familiar with `webdriver/*`).

We should dive deeper: raise an exception from `wptrunner.run_tests`.

Hexcles: Sorry, going down into `wptrunner.run_tests` doesn't really solve the problem at all.

Let me be conservative and only return errors if `tests_changed` is non-empty.

jgraham (approved):
Thanks!

PR #10100, opened 2018-03-19
Handle options for add_untested_idls

lukebjerring:
Currently, no way to import a subset of an IDL snippet as untested interfaces.

Note that the order of checking for `should_skip` needs to change for a case such as:
IDL:

    interface A {}
    interface B {}

    let text = await fetch('interfaces/foo.idl').then(r => r.text());
    idlArray.add_untested_idls(text, {only: ['A'] });
    idlArray.add_idls(text, {only: ['B'] });

The `add_idls` encounters 'A' for a second time, and throws `Duplicate identifier`, in spite of the fact that it will be skipped.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10100)

### DIFF
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -177,7 +177,7 @@ IdlArray.prototype.add_idls = function(raw_idls, options)
 };
 
 //@}
-IdlArray.prototype.add_untested_idls = function(raw_idls)
+IdlArray.prototype.add_untested_idls = function(raw_idls, options)
 //@{
 {
     /** Entry point.  See documentation at beginning of file. */
@@ -193,7 +193,7 @@ IdlArray.prototype.add_untested_idls = function(raw_idls)
             }
         }
     }
-    this.internal_add_idls(parsed_idls);
+    this.internal_add_idls(parsed_idls, options);
 };
 
 //@}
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build BROKEN

Started: 2018-03-19 19:13:01
Finished: 2018-03-19 19:18:02


<h2>Failing Jobs</h2>
<ul>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/26199)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/355522822)

foolip (approved):

foolip:
The new failures are surprising and are probably not due to the change itself, but can you take a look?

lukebjerring:
Those same failures seem unrelated and currently occur running tox tests under `resources/test/` in master, so I'm going to merge this.

foolip:
Hmm, this is weird, no failures from https://travis-ci.org/w3c/web-platform-tests/jobs/356918923.

@lukebjerring, I guess you're also seeing failures on master locally? Seems like we might have two different issues going on here, both a difference between local runs and Travis, and some actual change to the results as a consequence of this PR?

foolip:
Well that's interesting, the problem was presumably on master but now fixed...

@lukebjerring, is that a satisfactory "explanation" of what's going on, or do you think we should dig deeper into what's going on before merging?

lukebjerring:
Well, we can conclude that this PR is not causing the breakage, but something fishy is going on with those test.

PR #10110, opened 2018-03-20
Add HTMLOrSVGElement to HTML's IDL

annevk:
And align a couple other things while there.

Helps with https://github.com/whatwg/html/pull/3543.

### DIFF
--- a/interfaces/html.idl
+++ b/interfaces/html.idl
@@ -97,14 +97,10 @@ interface HTMLElement : Element {
   [CEReactions] attribute DOMString lang;
   [CEReactions] attribute boolean translate;
   [CEReactions] attribute DOMString dir;
-  [SameObject] readonly attribute DOMStringMap dataset;
 
   // user interaction
   [CEReactions] attribute boolean hidden;
   void click();
-  [CEReactions] attribute long tabIndex;
-  void focus(optional FocusOptions options);
-  void blur();
   [CEReactions] attribute DOMString accessKey;
   readonly attribute DOMString accessKeyLabel;
   [CEReactions] attribute boolean draggable;
@@ -123,6 +119,17 @@ HTMLElement includes ElementContentEditable;
 
 // Note: intentionally not [HTMLConstructor]
 interface HTMLUnknownElement : HTMLElement { };
+
+interface mixin HTMLOrSVGElement {
+  [SameObject] readonly attribute DOMStringMap dataset;
+  attribute DOMString nonce;
+
+  [CEReactions] attribute long tabIndex;
+  void focus(optional FocusOptions options);
+  void blur();
+};
+HTMLElement includes HTMLOrSVGElement;
+SVGElement includes HTMLOrSVGElement;
 
 [Exposed=Window,
  OverrideBuiltins]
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None

> This report has been truncated because the number of unstable tests exceeds GitHub.com's character limit for comments (65536 characters).


<h2>Failing Jobs</h2>
<ul>

<li>firefox:nightly</li>

</ul>



<h2>Unstable Browsers</h2>
  
  
  
  <h3>Browser: "Firefox Nightly"<small></small></h3>
  <p>View in: <a href="http://pulls.web-platform-tests.org/job/26230.4">WPT PR Status</a> |
      <a href="https://travis-ci.org/w3c/web-platform-tests/jobs/355892055">TravisCI</a></p>

foolip:
@Hexcles pointed out that it seems like this has broken a bunch of tests, like this:
https://chromium-review.googlesource.com/c/chromium/src/+/978021/4/third_party/WebKit/LayoutTests/external/wpt/IndexedDB/interfaces.any-expected.txt

So tests that include html.idl will now have to either define 'interface SVGElement {};' or import svg.idl.

foolip:
This kind of thing is going increasingly be a problem because the dependencies in spec IDL can be anything at all, and @lukebjerring is working on auto-updating from specs, and still the dependencies have to be managed manually in all of the tests including these files.

Other than fixing the immediate problem, ideas for how this is not going to keep happening?

Hexcles:
The full list of breakages found by Blink's CI (there might be more):

* IndexedDB/interfaces.any.html
* IndexedDB/interfaces.any.worker.html
* cookie-store/idlharness.tentative.html
* dom/interfaces.html
* hr-time/idlharness.html
* html/dom/interfaces.html
* webaudio/idlharness.https.html

https://crrev.com/c/978021

annevk:
Apologies :/

One thing that would help is having more reviewers. If we can automate all things IDL that'd help too. It seems the only thing that's really manual is defining how to allocate objects. It seems there could be a better system for that. The tests could then be generated using global knowledge of all IDL.

annevk:
(We could also move SVGElement includes ... to svg.idl... Though that would presumably also require that setup in the specification as otherwise the automation might mess it up again.)

foolip:
The most ambitious idea that @lukebjerring, @mdittmer and others have discussed would be precisely to eliminate the "defining how to allocate objects" manual bit. It'd be something like:
* For each interface for which an instance can be found, write one or more factory methods for it.
* All of the idlharness.js tests are automatically created, split either by spec with automatic dependency management, or split by interface, again with automatic dependency management.

annevk:
Yeah, what do we want to do here meanwhile? Comment out the SVGElement line?

foolip:
Yeah that'd be a good quickfix.

dstorey:
I feel it makes sense to add SVGElement includes to svg.idl and defined in the SVG spec. That was the change I was planning to make once the SVG PR to use the new mixin format is accepted. That way it is clearer in the SVG spec what members are included in the interface (specs are not really currently consistent if the implements/includes goes under the interface or the mixin. It is mostly the former except the DOM spec I think)

Hexcles:
Are we going to change html.idl (& svg.idl) today?

Otherwise, I'll send out a PR to batch add `add_untested_idls` to the affected tests.

foolip:
@Hexcles,  that would be more like a proper fix, so since the list of affected tests isn't that long, if you could try that, it'd be better I think.

inexorabletash:
Aside: short of the "most ambitious" approach, a helper script that knows how to pull in HTML + DOM as untested and has declarations for their dangling dependencies (LinkStyle from CSSOM, etc) would simplify maintenance on many of those tests (e.g. IndexedDB, cookie-store, webaudio, hr-time).

lukebjerring:
With https://github.com/w3c/web-platform-tests/pull/10100 landed, and not trying to script things, the fix for making tests robust is using the `{ only: ['MyInterface'] }` filtering for pulling in IDL interfaces (and deps) from the IDL files. That way, as interfaces are moved, the number of breakages is limited/mitigated.

I'd love to say that `only` (and `except`) are documented @ http://web-platform-tests.org/writing-tests/idlharness.html

... but that would be a lie.

annevk:
@dstorey that seems reasonable, but then we should remove it from the HTML Standard and make it a note there. For now I created #10160 to just comment this out, since it seems `svg/interfaces.html` hasn't really been converted to use mixins thus far and also duplicates some stuff (`GlobalEventHandlers`) so I'm not sure what the right strategy there would be.

PR #10127, opened 2018-03-21
Fix resources/test (when running locally)

### LINKED ISSUE #10114: ci_resources_unittest hangs when running locally
The tox setup doesn't seem to work at all. The test passes in most builds simply because the test is usually deemed unnecessary to run.

I've spotted a few issues and will try to send a fix PR.

cc @lukebjerring who informed me of the breakage. A failing example from his PR: https://travis-ci.org/w3c/web-platform-tests/jobs/353342172

Hexcles:
1. Check the HTTP port of wptserve instead of HTTPS to avoid the
   unnecessary complexities of setting up SSL context (which may fail in
   some environments, depending on the system OpenSSL version).
2. Use exponential backoff when waiting for wptserve and specify a
   maximum retry to avoid indefinite hang.
3. Use `terminate` instead of `kill` to give wptserve a chance to clean
   up, which is especially useful when running locally.

Fixes #10114.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10127)

### DIFF
--- a/resources/test/wptserver.py
+++ b/resources/test/wptserver.py
@@ -1,48 +1,49 @@
 import json
 import os
-import ssl
 import subprocess
+import time
 import urllib2
 
 _CONFIG_FILE = os.path.join(os.path.dirname(os.path.abspath(__file__)),
                             'config.test.json')
 
-with open(_CONFIG_FILE, 'r') as config_handle:
-    config = json.loads(config_handle.read())
-    host = config["host"]
-    port = config["ports"]["https"][0]
 
 class WPTServer(object):
-    base_url = 'https://%s:%s' % (host, port)
 
     def __init__(self, wpt_root):
         self.wpt_root = wpt_root
+        with open(_CONFIG_FILE, 'r') as config_handle:
+            config = json.load(config_handle)
+        self.host = config["host"]
+        self.http_port = config["ports"]["http"][0]
+        self.https_port = config["ports"]["https"][0]
+        self.base_url = 'http://%s:%s' % (self.host, self.http_port)
+        self.https_base_url = 'https://%s:%s' % (self.host, self.https_port)
 
     def start(self):
         self.devnull = open(os.devnull, 'w')
         self.proc = subprocess.Popen(
             [os.path.join(self.wpt_root, 'wpt'), 'serve', '--config=' + _CONFIG_FILE],
-            stdout=self.devnull,
             stderr=self.devnull,
             cwd=self.wpt_root)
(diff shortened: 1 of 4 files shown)


w3c-bots:
# Build BROKEN

Started: 2018-03-21 17:56:02
Finished: 2018-03-21 18:00:57


<h2>Failing Jobs</h2>
<ul>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/26289)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/356480453)

lukebjerring:
Yep, this lets me past the issue I was having.

Hexcles:
Fixed the SSL issue. The PR is now ready for review :)

jgraham (approved):

resources/test/wptserver.py
```diff
@@ -1,48 +1,51 @@
...
             cwd=self.wpt_root)
-        context = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
-        context.verify_mode = ssl.CERT_NONE
-        context.check_hostname = False
 
-        while True:
+        # Exponential backoff.
+        wait = 1
+        for _ in range(5):
+            time.sleep(wait)
```
jugglinmike: If we move this to the end of the loop body, we might avoid unnecessary startup delays. What do you think?

jugglinmike: Also, `time.sleep(_ ** 2)` might make your intention more clear (along with renaming `_` and removing `wait`)

jugglinmike (commented):
I can't reproduce the issue locally, but I can at least verify that this doesn't regress `tox` on my system.

resources/test/wptserver.py
Hexcles: I actually put it here intentionally as I noticed wptserve did take a short while to start and the first attempt often failed.

You meant `2 ** iterations` right? Yeah that'd be a bit cleaner. Let me change it.

Hexcles:
@jugglinmike thanks. A bit more details:

Creating a SSL context is actually quite tricky. Explicitly specifying TLS/SSL version isn't portable as newer OpenSSL may deprecate older, unsafe protocols (in fact, produce a fatal error). There is a new constant [`ssl.PROTOCOL_TLS`](https://docs.python.org/2/library/ssl.html#ssl.PROTOCOL_TLS) that alleviates the issue but is only available in very recent Python. A safer way is `ssl.create_default_context` and then tweaks the trust settings. At the end of the day, it's just overly and unnecessarily complicated...

PR #10131, opened 2018-03-21
Fix python gitignore for directories.

jgraham:
For top level directories like _venv/ we were creating the regexp
^\_venv/$, which we were then feeding paths like _venv/bin/foo which
it fails to match. Since we are matching against full paths, we
certainly want to allow matching _venv/* and probably also bare _venv,
so add a special case for paths with a trailing slash to allow these
options.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10131)

### DIFF
--- a/tools/gitignore/gitignore.py
+++ b/tools/gitignore/gitignore.py
@@ -17,6 +17,12 @@ def fnmatch_translate(pat, path_name=False):
     else:
         any_char = "."
         parts.append("^(?:.*/)?")
+    if pat[-1] == "/":
+        # If the last character is / match this directory or any subdirectory
+        pat = pat[:-1]
+        suffix = "(?:/|$)"
+    else:
+        suffix = "$"
     while i < len(pat):
         c = pat[i]
         if c == "\\":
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PASSED

Started: 2018-03-21 22:15:41
Finished: 2018-03-21 22:29:44





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26302)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/356597871)

tools/gitignore/gitignore.py
```diff
@@ -17,6 +17,12 @@ def fnmatch_translate(pat, path_name=False):
     else:
         any_char = "."
         parts.append("^(?:.*/)?")
+    if pat[-1] == "/":
```
Cactusmachete: I think parse_line doesn't pass the trailing `/` so we might need a workaround that condition first.
Also, afaict fnmatch_translate _does_ have a condition to deal with directories but due to parse_line, we never enter that condition. Not sure about the correctness of the regexp fnmatch came with for the directories though.

gsnedders (approved):

jgraham:
Oh, I realised later that this is harmless, but not really right. I should have closed the PR. Anyway, I think it's at least harmless.

PR #10149, opened 2018-03-23
Tests for SVG tabindex/focus/blur

dstorey:
SVG defines focus/blur/tabindex as behaving the same as HTML.

This takes some tests from HTML and adapts to the SVG template used in the SVG folder. I included them in the interact folder as that page (scripting) is the one that defines which elements are focusable in SVG.

Apologies if I got the folder structure wrong. I believe I suppose to put the tests in the in the scripted folder. 

Note: SVG say that the following are focusable by default:

- the document root element
- an ‘svg’ element which has zoom and pan controls
https://svgwg.org/svg2-draft/interact.html#Focus

I only included a test for the root-most SVG element, as SVG further defines that zoom and pan controls only work on the root most SVG element, so it will pass with or without that control.

I also assumed that video/audio with user controls meant the controls attribute. 

These tests match the behavior of the HTML equivalents. Edge fails the tabindex tests when checking the content attribute as there is a non-tabindex related bug with attribute reflection in SVG where the canonical case is lowercase. Otherwise those tests would pass. 

I couldn't find any HTML tests specific to blur() so I reused the composed test. blur() passes in Edge, but doesn't support event.composed.

I didn't add tests for dataset (which is in the same section of the spec) as there are already tests in HTML that detect dataset on SVG


w3c-bots:
# Build PENDING

Started: None
Finished: None


<h2>Failing Jobs</h2>
<ul>

<li>chrome:dev</li>

</ul>



<h2>Unstable Results</h2>
  
  
  
  <h3>Browser: "Chrome Dev"<small></small></h3>
  <p>View in: <a href="http://pulls.web-platform-tests.org/job/26366.5">WPT PR Status</a> |
      <a href="https://travis-ci.org/w3c/web-platform-tests/jobs/357188639">TravisCI</a></p>
  <table>
    <tr>
      <th>Test</th>
      <th>Subtest</th>
      <th>Results</th>
      <th>Messages</th>
    </tr>
  
  
  
  
  <tr>
    <td><code>/svg/interact/scripted/focus-events.svg</code></td>
    <td>&nbsp;</td>
    <td>TIMEOUT: 2<br />OK: 8<br /></td>
    <td></td>
  </tr>
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  <tr>
    <td>&nbsp;</td>
    <td><code>focusing on a focusable element fires a blur event at the previous focussed element</code></td>
    <td>PASS: 8<br />NOTRUN: 2<br /></td>
    <td></td>
  </tr>
  
  
  
  <tr>
    <td>&nbsp;</td>
    <td><code>focusing on a focusable element fires a focus event at the element</code></td>
    <td>PASS: 10<br /></td>
    <td></td>
  </tr>
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  
  </table>

dstorey:
Hmmm works on my machine :D

AmeliaBR:
Thanks for this, David!  (When I discovered Edge's `tabIndex` vs `tabindex` bug last week, my first thought was "this is why we need tests!!!")

I agree it makes sense to put these in the `interact` folder, since that's where the behavior is defined, even if the attribute is defined in structure. Also, non-interactive user agents can skip over these tests.

For`zoomAndPan`, the distinction from a `svg:root` was I think to cover the case of a zoomable inline SVG. But doesn't really matter, since none of the major browsers support SVG zoom and pan, other than by browser zoom and scrollbars on a document window. The spec intentionally says "*if* it has controls" for that reason, rather than talking about the attribute values.

Some thoughts:

- There should be tests that nested `svg` don't get the focusable-by-default behavior that root SVG do.

- Add tests for normally-rendered elements (e.g., `a[href]` and `path[tabindex=0]`) that are not rendered because of being inside `defs`, `symbol`, `pattern`, etc. (At least some browsers let them get focus, last I checked!) Also for elements that have `display: none` (pretty sure browsers are better with those) and shape elements that are not rendered because of invalid geometry properties (haven't tested). 

- Conversely, add a test that `visibility: hidden` doesn't cancel focusability (as defined in https://svgwg.org/svg2-draft/render.html#VisibilityControl ) (this will probably fail & so should probably be marked as risk, but it's an accessibility thing related to the fact that these elements can still be sensitive to pointer events in SVG).

- Is there a way to distinguish the "shoulds" tests (e.g., `focusable` support) from the "musts"?

- You might want to separate out tests relating to HTML embedded elements into a separate file, so it can clearly be dependent on underlying support of those elements in SVG contexts.

- Is there a purpose to [svg/interact/scripted/resources/blank.htm](https://github.com/w3c/web-platform-tests/pull/10149/files#diff-75a8a146aa5497ee357239fbe6e93034) ?


_______________________

Additional tests still required for full coverage of this section (not necessarily in this PR, just keeping note):

- tests for correct tab order calculation (e.g., when an element has a positive `tabindex` value)

- inline svg tests (check default focusable state of non-root `svg` element, make sure tabindex still works, make sure that the tab order is correctly calculated on a document-wide basis)

- Also, might want to add a test of HTML focusable elements inside `<foreignObject>`.

- make sure scroll-into-view on focus works in both SVG docs and inline SVG in HTML

- check that 'click' events are fired for keyboard activation of a focused element

- tests for focus handling of elements inside use-element shadow DOM

- tests of `:focus` styles (default styles are applied, and custom styles can be used to override)

dstorey:
Yeah, I think some of these would probably be a separate PR as they're not covered by the HTML versions either (presumably :focus is in a CSS test for example), and perhaps the existing tests in this PR would be enough to get the HTML PR merged for adding the mixin re require to move these definitions into HTML. 

I'll look into an additional PR though that adds things that don't need to be in here. the `foreignObject`/`use` maybe should be in this one. I guess I could just have a duplicate test in the flags test inside a use element for that one. I've not used `use` for a long time so will have to look it up again :) 

(Edge is fixing our reflect bugs. Now our tree work is done we're working on improving how we deal with attributes)

dstorey:
Re: should vs must: I don't think so. Edge passes them at least :D My understanding from HTML is most of the focusable by default elements are "should" anyway (or at least there is language about being up to the user agent/platform conventions, so it is probably fine to include them.

re: blank.html. The text mentioned valid (nested) browsing context, so I added a document. I guess I could add a data URI or such too but I just copied what was used in other tests and didn't want to deal with encoding quotes and angle brackets. I wasn't 100% sure if the definition meant if it is just an iframe with no html/svg document that it shouldn't be focusable. 

Re: controls: do you mean zoomAndPan or video/audio, as it isn't 100% clean that the latter means the controls attribute?

Re: non root SVG: I thought I added a basic test for this but may be mistaken. I'll look into it.

AmeliaBR:
> Re: should vs must

Ah, I was thinking of the [CSS test metadata](http://web-platform-tests.org/writing-tests/css-metadata.html). I'd like to get to the point of adding similar metadata for all SVG tests (plus SVG-specific metadata, like which [processing modes](https://svgwg.org/svg2-draft/conform.html#processing-modes) the test is relevant to), but that will wait until we have an agreed-upon metadata system.

> Re: blank.html

Ah, ok. Makes sense.

> Re: controls: do you mean zoomAndPan or video/audio, as it isn't 100% clean that the latter means the controls attribute?

I meant zoomAndPan, but the general wording is used for both, and for the same reasons. For [audio/video controls](https://w3c.github.io/html/semantics-embedded-content.html#element-attrdef-mediaelements-controls), the user agent is supposed to add controls even without the attribute if scripting is disabled, and may add controls regardless (in practice, that is usually with a context menu option), so again the focusability rule just focused on whether the controls are there.

But if the default behavior in HTML is to not make the element focusable just for context menu controls, then we can match that. Or leave out the no-attribute test & only test that the elements are focusable if the author has specifically requested user-agent controls.

> Re: non root SVG: I thought I added a basic test for this but may be mistaken. I'll look into it.

You had a test of `tabindex` working on a non-root SVG, I suggested adding a test of the default behavior being non-focusable.

dstorey:
Makes sense. I'll remove the test that tests audio/video fail if no controls as you mention as there is no way of detecting if the UI adds their own except with [controls].

svgeesus:
Hmm looks like the error is caused by WPT moving to a new repo.

svgeesus:
@AmeliaBR I would like to get these merged, because this PR is languishing and also because it may be easier to merge and then do a separate PR for any fixes, rather than ask @dstorey to fix them. I'm therefore going to give an approving review.

svgeesus (approved):
Yes, let's fix up any details after merge.

svgeesus:
@foolip I think this old PR needs admin to override the Travis CI failure caused by the repo having moved.

foolip:
@svgeesus, I've merged in origin/master into the review branch, which has re-triggered Travis. Just make sure to squash this when merging as the history will be very silly otherwise.

foolip:
@svgeesus Travis last ran 2 months ago. Those checks are no longer blocking. I restarted but it's not picking up changes to the Travis config that have happened on master it seems. I'll just merge in origin/master again.

foolip:
Alright, everything is fine now :)

foolip:
Bah! I picked the "rebase and merge" option, but that failed. When I hit the "try again" button eventually it was merged, but squashed. So there's now just a single commit in history, where it's not as clear which comments refer to which test files. Sorry :(

PR #10160, opened 2018-03-23
Comment out SVGElement dependency from html.idl

annevk:
Fixes regression caused by #10110.

### DIFF
--- a/interfaces/html.idl
+++ b/interfaces/html.idl
@@ -129,7 +129,7 @@ interface mixin HTMLOrSVGElement {
   void blur();
 };
 HTMLElement includes HTMLOrSVGElement;
-SVGElement includes HTMLOrSVGElement;
+// SVGElement includes HTMLOrSVGElement;
 
 [Exposed=Window,
  OverrideBuiltins]


Hexcles (approved):
LGTM, although I too am not sure what is the best approach here given you said "it seems svg/interfaces.html hasn't really been converted to use mixins thus far and also duplicates some stuff (GlobalEventHandlers)"

Hexcles:
BTW I'm also fixing the affected tests from the Chromium side https://crrev.com/c/978508

Both this PR and my CL can fix the current regression independently. And we may also have them both, especially since my CL narrows down the dependencies in a few tests by using `add_untested_idls(..., { only: [...] })` so it's good to have anyway.

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26406)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/357512667)

annevk:
@Hexcles @foolip I'll let you merge this (please use "rebase and merge") if you think this is needed.

foolip:
Since the problem has been fixed I think we should close this and fix https://github.com/w3c/web-platform-tests/blob/master/svg/interfaces.html by creating a new interfaces/svg.idl that matches the current spec. If the current SVG spec's IDL doesn't mix with the current HTML spec's IDL, that's a problem to fix spec-side first.

@lukebjerring, I guess that SVG isn't one of the specs that we have complete IDL to auto-update with?

PR #10164, opened 2018-03-23
Special-case idlharness.js errors as an assert_throw option

lukebjerring:
Fixes https://github.com/w3c/web-platform-tests/issues/10162

This adds support for more succinctly asserting behaviour of `idlharness.js` itself, using the same mechanism in `testharness.js`, e.g. when testing expected errors in `IdlArray.test()`.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10164)

### DIFF
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -603,8 +603,7 @@ IdlArray.prototype.assert_throws = function(error, idlArrayFunc)
 {
     try {
         idlArrayFunc.call(this, this);
-        assert(false, "assert_throws", description,
-            "${func} did not throw", { func: func });
+        throw new IdlHarnessError(`${idlArrayFunc} did not throw the expected IdlHarnessError`);
     } catch (e) {
         if (e instanceof AssertionError) {
             throw e;
@@ -616,9 +615,7 @@ IdlArray.prototype.assert_throws = function(error, idlArrayFunc)
         if (e.message !== error) {
             throw new IdlHarnessError(`${idlArrayFunc} threw ${e}, not the expected IdlHarnessError`);
         }
-        return;
     }
-    throw new IdlHarnessError(`${idlArrayFunc} did not throw the expected IdlHarnessError`);
 }
 
 //@}
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -123,6 +123,28 @@ var fround =
 })();
 //@}
 
+/// IdlHarnessError ///
+// Entry point
+self.IdlHarnessError = function(message)
+//@{
+{
+    /**
+     * Message to be printed as the error's toString invocation.
+     */
+    this.message = message;
+};
+
+IdlHarnessError.prototype = Object.create(Error.prototype);
+
+//@}
+IdlHarnessError.prototype.toString = function()
+//@{
+{
+    return this.message;
+};
+
+//@}
+
 /// IdlArray ///
 // Entry point
 self.IdlArray = function()
@@ -217,7 +239,7 @@ IdlArray.prototype.internal_add_idls = function(parsed_idls, options)
 
     if (options && options.only && options.except)
     {
-        throw "The only and except options can't be used together."
+        throw new IdlHarnessError("The only and except options can't be used together.");
     }
 
     function should_skip(name)
@@ -280,7 +302,7 @@ IdlArray.prototype.internal_add_idls = function(parsed_idls, options)
         }
         if (parsed_idl.name in this.members)
         {
-            throw "Duplicate identifier " + parsed_idl.name;
+            throw new IdlHarnessError("Duplicate identifier " + parsed_idl.name);
         }
         switch(parsed_idl.type)
         {
@@ -374,7 +396,7 @@ IdlArray.prototype.recursively_get_implements = function(interface_name)
         ret = ret.concat(this.recursively_get_implements(ret[i]));
         if (ret.indexOf(ret[i]) != ret.lastIndexOf(ret[i]))
         {
-            throw "Circular implements statements involving " + ret[i];
+            throw new IdlHarnessError("Circular implements statements involving " + ret[i]);
         }
     }
     return ret;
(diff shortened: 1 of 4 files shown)


w3c-bots:
# Build BROKEN

Started: 2018-03-23 19:21:14
Finished: 2018-03-23 19:27:48


<h2>Failing Jobs</h2>
<ul>

<li>lint</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/26414)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/357550731)

resources/testharness.js
```diff
@@ -1258,6 +1265,19 @@ policies and contribution forms [3].
                 throw e;
             }
 
+            if ((typeof IdlHarnessError !== 'undefined')
```
foolip: The way that `assert_throws` is used for the per-spec tests themselves is to just check that the right kind of exception is thrown, as well as it can be checked (see "but we can't" below) and to never assert anything about the exception message itself, since the messages aren't defined in web specs and are UA-defined.

For idlharness.js it is kinda useful to check that the right exception message has been thrown, but I wonder if it isn't better to use a custom assertion for those tests?

An alternative could be to allow messages to be asserted more generically, adding it to the `if (typeof code === "object")` branch below, that if `code.message` is a non-empty string, we assert that the throw exception matches. That would remove the need to have a special `IdlHarnessError`.

WDYT?

resources/idlharness.js
```diff
@@ -217,7 +239,7 @@ IdlArray.prototype.internal_add_idls = function(parsed_idls, options)
 
     if (options && options.only && options.except)
     {
-        throw "The only and except options can't be used together."
+        throw new IdlHarnessError("The only and except options can't be used together.");
```
foolip: See comment elsewhere about doing the message comparison more generically. That'd make it possible to just use `new Error` everywhere in idlharness.js.

resources/testharness.js
foolip: I see there's also an AssertionError in testharness.js that we could perhaps use. It's immediately rethrown by assert_throws. It really depend on what problem https://github.com/w3c/web-platform-tests/issues/10162 started out with, if it's just about fixing the inconsistency or if there's also some test that needs to be fixed.

foolip: To see if that idea is at all viable I sent https://chromium-review.googlesource.com/c/chromium/src/+/980757 to see if there are lots of existing tests that pass a message that *shouln't* be checked.

foolip: Yep, there were a handful, like this one:
https://chromium.googlesource.com/chromium/src/+/15e299fd19da9cf6f74dbcb78b6bc4865f7a5a35/third_party/WebKit/LayoutTests/external/wpt/mediasession/setactionhandler.html

Seems manageable if we want to do it that way, though.

resources/idlharness.js
```diff
@@ -123,6 +123,28 @@ var fround =
 })();
 //@}
 
+/// IdlHarnessError ///
+// Entry point
+self.IdlHarnessError = function(message)
+//@{
```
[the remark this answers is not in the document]
lukebjerring: Nope. Just copy-pasta-ing

resources/idlharness.js
```diff
@@ -567,7 +589,7 @@ function exposed_in(globals) {
         return globals.indexOf("Worker") >= 0 ||
                globals.indexOf("ServiceWorker") >= 0;
     }
-    throw "Unexpected global object";
+    // throw new IdlHarnessError("Unexpected global object");
```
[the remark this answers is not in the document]
lukebjerring: Accidental.

resources/test/tox.ini
```diff
@@ -13,4 +13,4 @@ deps =
   selenium
   requests
 
-commands = pytest {posargs} -vv tests
+commands = pytest {posargs} -vv tests/api-tests-1.html
```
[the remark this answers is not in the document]
lukebjerring: Yes, the use of {posargs} preceding means there's no simple way to run only a specific test. Reverted.

resources/testharness.js
```diff
@@ -1247,6 +1247,13 @@ policies and contribution forms [3].
     }
     expose(assert_readonly, "assert_readonly");
 
+    /**
+     * Assert a DOMException with the expected code is thrown.
```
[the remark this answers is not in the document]
lukebjerring: Done.

resources/testharness.js
```diff
@@ -1247,6 +1247,13 @@ policies and contribution forms [3].
...
+     *
+     * @param {number|string} code The expected exception code.
```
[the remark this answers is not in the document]
lukebjerring: Going to leave it alone for this PR.

resources/testharness.js
lukebjerring: I agree that it's bad to reference `IdlHarnessError` from here.

resources/idlharness.js
lukebjerring: Keeping explicit error type (I see value in it) but made sure not to ref it from `testharness.js`

lukebjerring:
PTAL

resources/idlharness.js
```diff
@@ -567,7 +589,36 @@ function exposed_in(globals) {
...
+        // Assertions for behaviour of the idlharness.js engine.
+        if (error instanceof IdlHarnessError) {
+            error = error.message;
+        }
+        if (e.message !== error) {
+            throw new IdlHarnessError(`${idlArrayFunc} threw ${e}, not the expected IdlHarnessError`);
+        }
+        return;
+    }
+    throw new IdlHarnessError(`${idlArrayFunc} did not throw the expected IdlHarnessError`);
```
foolip: I think this is unreachable because of the `assert(false, ...)` above.

foolip (approved):
LGTM % a few nits

resources/idlharness.js
```diff
@@ -567,7 +589,36 @@ function exposed_in(globals) {
...
+/**
+ * Asserts that the given error message is thrown for the given function.
+ * @param {string|IdlHarnessError} error Expected Error message.
+ * @param {Function} idlArrayFunc Function operating on an IdlArray that should throw.
+ */
+IdlArray.prototype.assert_throws = function(error, idlArrayFunc)
+//@{
+{
+    try {
+        idlArrayFunc.call(this, this);
```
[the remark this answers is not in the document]
lukebjerring: True about the first param, but `i => {` is `function(i) {` so it's definitely calling with `i = IdlArray`

resources/idlharness.js
lukebjerring: Corrected.

Issue #8581, opened 2017-12-06
Isolated domain for cookie tests?


mikewest:
We're looking into porting the IETF's cookie test suite (https://github.com/abarth/http-state/tree/master/tests) to WPT, which would be simplified if we could replicate its domain structure (https://github.com/abarth/http-state/blob/master/tests/data/hosts):

```
127.0.0.1 example.org
127.0.0.1 home.example.org
127.0.0.1 home.example.org.
127.0.0.1 sibling.example.org
127.0.0.1 subdomain.home.example.org
127.0.0.1 sibling.home.example.org
```

This basically boils down to two requests:

1.  Since y'all are looking into adding a new registrable domain in https://github.com/w3c/web-platform-tests/issues/2669, perhaps you could add a "just for cookie tests" domain? That would make the checks quite a bit simpler, as we wouldn't need to deal with noise from other cookies set by other tests on the same registrable domain.

2.  Can we add a subdomain of a subdomain (and another subdomain of the same subdomain) to the list of hosts?

foolip:
@gsnedders, can you see where this fits on your list of roadmap bugs? Seems like this could be solved together with https://github.com/w3c/web-platform-tests/issues/2669?

gsnedders:
@foolip Yes, easy to do alongside #2669.

I'd be curious as to what @jgraham thinks of adding such a domain?

foolip:
@gsnedders, this came up in issue triage today. With https://github.com/w3c/web-platform-tests/pull/8614 now landed, I guess a fix for this is pretty simple?

Also ping @jgraham for the requested opinions.

gsnedders:
@mikewest How do you want to enforce the "just for cookie tests" side? A lint that allows it only in `cookies/`? That said, given the tests can be run in any order, is there actually much advantage of limiting it to just `cookies/`? Ultimately, any arbitrary cookie test could've run before it and set any arbitrary cookie, so I'm not sure what the advantage of limiting it to just cookie tests is?

mikewest:
I wasn't thinking there would be any technical constraint for non-cookie tests. It might be nice to have a non-technical constraint (like, naming it 'for-cookies' or 'this-is-mikes-domain-go-away' or 'i-promise-this-is-a-cookie-related-test' or 'cookies-only' or 'cookies-are-the-best-really' or 'tests-that-do-not-use-cookies-should-not-use-this-domain-but-could-i-guess-if-they-really-wanted-to').

[...]

PR #10166, opened 2018-03-24
Add alternate_hosts

### LINKED ISSUE #2669: No cross-registrable-domain host is available for testing.
Would it be possible to add a cross-registrable-domain origin to WPT? That is, all of the `{{domain[*]}}` options are rooted in `web-platform.test` at the moment, which makes it super-tough to test things like https://tools.ietf.org/html/draft-west-first-party-cookies-06. `not-web-platform.test` would be ideal (along with `not-w3c-test.org`, of course).

gsnedders:
Fixes #2669.

Note this doesn't yet do #8581, but adding a new host at this point is adding a string to the config file, and adding new subdomains is just editing the list in `tools/serve/serve.py`.

@mikewest: This gives us `alt_domains[0]` and `alt_domains[1]`, which isn't quite what we were talking about in #8581 (given there's nowhere to put 'this-is-mikes-domain-go-away'). I think giving them names is going to be a fair bit of extra complexity, but let me know what you think?

Also, FWIW, I'm pretty sure this will fail CI. But opening the PR so people can look at it if they want before I touch it again on Monday.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10166)

### DIFF
--- a/config.default.json
+++ b/config.default.json
@@ -1,4 +1,7 @@
 {"browser_host": "web-platform.test",
+ "alternate_hosts": {
+     "alt": "not-web-platform.test"
+ },
  "doc_root": null,
  "ws_doc_root": null,
  "server_host": null,
--- a/docs/_writing-tests/server-features.md
+++ b/docs/_writing-tests/server-features.md
@@ -14,15 +14,18 @@ generation. This is supported through the
 
 ### Tests Involving Multiple Origins
 
-In the test environment, five subdomains are available: `www`, `www1`,
-`www2`, `天気の良い日`, and `élève`; there is also
-`nonexistent-origin` which is guaranteed not to resolve. In addition,
-the HTTP server listens on two ports, and the WebSockets server on
-one. These subdomains and ports must be used for cross-origin
-tests. Tests must not hardcode the hostname of the server that they
-expect to be running on or the port numbers, as these are not
-guaranteed by the test environment. Instead they can get this
-information in one of two ways:
+Our test servers are guaranteed to be accessible through two domains
+and five subdomains under each. The 'main' domain is unnamed; the
+other is called 'alt'. These subdomains are: `www`, `www1`, `www2`,
+`天気の良い日`, and `élève`; there is also `nonexistent-origin` which
+is guaranteed not to resolve. In addition, the HTTP server listens on
+two ports, and the WebSockets server on one. These subdomains and
+ports must be used for cross-origin tests.
+
+Tests must not hardcode the hostname of the server that they expect to
+be running on or the port numbers, as these are not guaranteed by the
+test environment. Instead they can get this information in one of two
+ways:
 
 * From script, using the `location` API.
 
@@ -33,15 +36,19 @@ form `{name}.sub.{ext}` e.g. `example-test.sub.html` or be referenced
 through a URL containing `pipe=sub` in the query string
 e.g. `example-test.html?pipe=sub`. The substitution syntax uses `{%
 raw %}{{ }}{% endraw %}` to delimit items for substitution. For
-example to substitute in the host name on which the tests are running,
-one would write: `{% raw %}{{host}}{% endraw %}`.
+example to substitute in the main host name, one would write:
+`{% raw %}{{host}}{% endraw %}`.
 
+To get full domains, including subdomains, there is the `hosts`
+dictionary, where the first dimension is the name of the domain, and
+the second the subdomain. For example, `{% raw %}{{hosts[][www]}}{%
+endraw %}` would give the `www` subdomain under the main (unnamed)
+domain, and `{% raw %}{{hosts[alt][élève]}}{% endraw %}` would give
+the `élève` subdomain under the alt domain.
 
-As well as the host, one can get full domains, including subdomains
-using the `domains` dictionary. For example, `{% raw
-%}{{domains[www]}}{% endraw %}` or `{% raw %}{{domains[élève]}}{%
-endraw %}` would be replaced by the full qualified domain name of the
-respective subdomains.
+For mostly historic reasons, the subdomains of the main domain are
+also available under the `domains` dictionary; this is identical to
+`hosts[]`.
 
 Ports are also available on a per-protocol basis. For example, `{% raw
 %}{{ports[ws][0]}}{% endraw %}` is replaced with the first (and only)
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -437,13 +437,11 @@ def check_subdomains(domains, paths, bind_address, ssl_config, aliases):
 def make_hosts_file(config, host):
     rv = []
 
-    for per_host_domains in config["domains"].values():
-        for domain in per_host_domains.values():
-            rv.append("%s\t%s\n" % (host, domain))
+    for domain in config.domains_set:
+        rv.append("%s\t%s\n" % (host, domain))
 
-    for per_host_domains in config["not_domains"].values():
-        for not_domain in per_host_domains.values():
-            rv.append("0.0.0.0\t%s\n" % not_domain)
+    for not_domain in config.not_domains_set:
+        rv.append("0.0.0.0\t%s\n" % not_domain)
 
     return "".join(rv)
 
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -423,7 +423,10 @@ def check_subdomains(domains, paths, bind_address, ssl_config, aliases):
                         "You may need to edit /etc/hosts or similar, see README.md." % (host, port))
         sys.exit(1)
 
-    for domain in domains.itervalues():
+    for domain in domains_set.itervalues():
+        if domain == host:
+            continue
+
         try:
             urllib2.urlopen("http://%s:%d/" % (domain, port))
         except Exception as e:
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -401,7 +401,7 @@ class ServerProc(object):
 
 def check_subdomains(domains, paths, bind_address, ssl_config, aliases):
     domains = domains.copy()
-    host = domains.pop("")
+    host = domains.pop("main_")
     port = get_port(host)
     logger.debug("Going to use port %d to check subdomains" % port)
 
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -437,11 +437,13 @@ def check_subdomains(domains, paths, bind_address, ssl_config, aliases):
 def make_hosts_file(config, host):
     rv = []
 
-    for domain in config["domains"].values():
-        rv.append("%s\t%s\n" % (host, domain))
+    for per_host_domains in config["domains"].values():
+        for domain in per_host_domains.values():
+            rv.append("%s\t%s\n" % (host, domain))
 
-    for not_domain in config.get("not_domains", {}).values():
-        rv.append("0.0.0.0\t%s\n" % not_domain)
+    for per_host_domains in config["not_domains"].values():
+        for not_domain in per_host_domains.values():
+            rv.append("0.0.0.0\t%s\n" % not_domain)
 
     return "".join(rv)
 
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -399,9 +399,13 @@ class ServerProc(object):
         return self.proc.is_alive()
 
 
-def check_subdomains(domains, paths, bind_address, ssl_config, aliases):
-    domains = domains.copy()
-    host = domains[""][""]
+def check_subdomains(config):
+    paths = config.paths
+    bind_address = config.bind_address
+    ssl_config = config.ssl_config
+    aliases = config.aliases
+
+    host = config.server_host
     port = get_port(host)
     logger.debug("Going to use port %d to check subdomains" % port)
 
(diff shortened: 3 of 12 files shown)


w3c-bots:
# Build BROKEN

Started: 2018-03-24 16:57:28
Finished: 2018-03-24 17:11:06


<h2>Failing Jobs</h2>
<ul>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/26422)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/357828195)

mikewest:
> @mikewest: This gives us alt_domains[0] and alt_domains[1], which isn't quite what we were talking about in #8581 (given there's nowhere to put 'this-is-mikes-domain-go-away'). I think giving them names is going to be a fair bit of extra complexity, but let me know what you think?

I guess I'd be a little surprised when writing tests if I had to shift over to an entirely different semantic in order to use a different domain. Subdomains of `web-platform.test` are referenced as `{{domains[www]}}`. I'd expect this alternate domain to work similarly. That is, `{{alt_domains[www]}}` would work as a subdomain on the alternate root, or `{{domains[alt_www]}}` would be the `www` subdomain on the alternate root.

`{{alt_domains[0]}}` feels pretty different. Can you help me understand the complexity hiding behind the simple-seeming approaches above?

(Also: I don't actually care about `this-is-mikes-domain-go-away`. I care about the technical capability to have subdomains of subdomains. The scary name seems interesting from an ergonomic perspective, but that's the least important bit of the request, and I'm totally happy to drop it entirely.)

gsnedders:
@mikewest 

So I think there's a few questions here:

1. How do we refer to the new alternate hosts from the original one (esp. given all tests are loaded from web-platform.test:8000)?
2. How do we refer to a subdomain on the current top-level host (for lack of a better term)?
3. How do we refer to the original host from the alternate ones?
4. How do we refer to alternate hosts from each other?

Half of this is complicated by the expectation that we'll have multiple alternate hosts (i.e., we can't just go with `alt_domains[www]` because which alt host is this referring to?).

The current behaviour in this PR is:

1. `alt_domains[0][]`
2. You don't reasonably (there's no way to find out what alternate host you're on)
3. `domains[]`
4. See 2.

Most of that is driven by what makes implementation simple.

I'm not entirely sure what I want the answers to be; my gut says:

1. `alt_domains[0][]`
2. `domains[www]`
3. `domains[www]` (um, that's impossible along with 2); thoughts?
4. Not sure? I don't know how much this matters though?

mikewest:
> 1.  How do we refer to the new alternate hosts from the original one (esp. given all tests are loaded from web-platform.test:8000)?
> 2.  How do we refer to a subdomain on the current top-level host (for lack of a better term)?
> 3.  How do we refer to the original host from the alternate ones?

One way to answer this question is to stop thinking of `{{domains[...]}}` as referencing a subdomain, but as referencing an entire domain (maybe we could rename it "host" or something?). Then `{{domains[www]}}` => `www.web-platform.test`, while `{{domains[alt_www]}}` => `www.not-web-platform.test`.

> 4.  How do we refer to alternate hosts from each other?

I think some of this logic needs to be hard-coded in the test. Like, we know that tests open up on `web-platform.test`, and that any reference to `{{domains[...]}}` will be "somethign else". I think it's probably reasonable to put the onus on the developer to track where they are/need to be via `window.open` or navigation, and to load the right `{{domains[...]}}` accordingly.

I guess it's not clear to me whether we really need another alternate host. I guess having two registrable domains suggests that we might need more than two? But two is different in kind from one in a way that three might not me. That is, I think I can perform all the cookie tests I care about from one registrable domain to another. Adding a third wouldn't help me in the tests I know I need to write today.

jgraham:
So changing the semantics of `{{domains[]}}` is probably risky given that it's used rather widely; a simple grep returned 230+ matches.

One could do some magic to make a two argument form work like `{{domains[alt,www]}}` or `{{domains[alt][www]}}`. Or one could make up a new thing like `{{hosts[alt][www]}}`. I think I agree that the numbered thing is bad, unless the idea is that all alternate domains are interchangable.

gsnedders:
> One way to answer this question is to stop thinking of `{{domains[...]}}` as referencing a subdomain, but as referencing an entire domain (maybe we could rename it "host" or something?). Then `{{domains[www]}}` => `www.web-platform.test`, while `{{domains[alt_www]}}` => `www.not-web-platform.test`.

We already have `{{ host }}` (and it's used all over the place), it's essentially an alias of `{{ domains[] }}` (i.e., `web-platform.test`). `alt_www` is workable if we only have one alt domain or otherwise name them.

> I guess it's not clear to me whether we really need another alternate host. I guess having two registrable domains suggests that we might need more than two? But two is different in kind from one in a way that three might not me. That is, I think I can perform all the cookie tests I care about from one registrable domain to another. Adding a third wouldn't help me in the tests I know I need to write today.

My understanding from #2669 and #8581 was that you wanted two new alternate hosts. If you think we might only need one, then that arguably makes this all simpler.

> So changing the semantics of `{{domains[]}}` is probably risky given that it's used rather widely; a simple grep returned 230+ matches.

It's low-risk because any bogus use will currently return 500 Internal Server Error (as with most bogus sub uses).

> One could do some magic to make a two argument form work like `{{domains[alt,www]}}` or `{{domains[alt][www]}}`. Or one could make up a new thing like `{{hosts[alt][www]}}`. I think I agree that the numbered thing is bad, unless the idea is that all alternate domains are interchangable.

`[alt,www]` is no easier/harder to make work than `[alt_www]`. I'm happy to name them, if that's what people want (presumably all the way to the config file, therefore), if we're going to allow more than one alternate host.

mikewest:
> My understanding from #2669 and #8581 was that you wanted two new alternate hosts. If you think we might only need one, then that arguably makes this all simpler.

Sorry for the confusion. _I_ need one. _I_ don't need two. I can imagine features that might need two at some point (if we ever get around to working through the weeds of concepts like [site affiliation](https://lists.w3.org/Archives/Public/public-webappsec/2017Apr/0038.html), for instance, or if folks decide that FIDO's [facets](https://fidoalliance.org/specs/fido-appid-and-facets-ps-20150514.pdf) were a good idea after all), but the use cases I know of today don't need more than two registrable domains total.

> `[alt,www]` is no easier/harder to make work than `[alt_www]`. I'm happy to name them, if that's what people want (presumably all the way to the config file, therefore), if we're going to allow more than one alternate host.

If we did that, how about going all the way with something like `[www.not-web-platform.test]`, where the existing names are aliased to `[www.web-platform.test]`?

gsnedders:
> Sorry for the confusion. _I_ need one. _I_ don't need two. I can imagine features that might need two at some point (if we ever get around to working through the weeds of concepts like [site affiliation](https://lists.w3.org/Archives/Public/public-webappsec/2017Apr/0038.html), for instance, or if folks decide that FIDO's [facets](https://fidoalliance.org/specs/fido-appid-and-facets-ps-20150514.pdf) were a good idea after all), but the use cases I know of today don't need more than two registrable domains total.

To me, that sounds like there's a reasonable chance of needing more than two in the future, and we probably shouldn't therefore go with something that makes ergonomics worse when we add more.

> If we did that, how about going all the way with something like `[www.not-web-platform.test]`, where the existing names are aliased to `[www.web-platform.test]`?

I did consider something like that; I'm still not opposed to it, though it does make linting for `web-platform.test` harder. 🙂

jgraham:
I wouldn't mind `[www]` == `[www.host]` and `[www.alternate]` but you would need to spell the bare alternate as `[.alternate]` or have clashing syntax.

I object to putting the full wpt domain in because that's just wrong in some scenarios, and therefore misleading.

mikewest:
Any progress on this? There's increased interested in `SameSite` from some browsers, and it would be really nice if we could share tests rather than rewriting things in our own internal testing frameworks a few times over. :)

jgraham:
The only funtional change I would like here is moving from `domains[main_foo]` to `hosts[][foo]` or `hosts[main][foo]` and `hosts[alt][foo]`. Apart from that it just needs to pass tests.

I agree this is high priority.

gsnedders:
@mikewest tl;dr: I got ill, and then I was at the CSS WG F2F and am now out in CA for BlinkOn. Still a priority for when I actually write any code.

gsnedders:
> The only funtional change I would like here is moving from `domains[main_foo]` to `hosts[][foo]` or `hosts[main][foo]` and `hosts[alt][foo]`. Apart from that it just needs to pass tests.

And have `domains[foo]` be equal to `hosts[][foo]`?

jgraham:
Yes.

gsnedders:
@jgraham that's now done

gsnedders:
Review status: 7 of 11 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[tools/wptserve/wptserve/config.py, line 185 at r4](https://reviewable.io/reviews/w3c/web-platform-tests/10166#-LBLi33EEyWpyt-AuHWt:-LBLt5sVAc_pccP03Ij4:b6ey5r9) ([raw file](https://github.com/w3c/web-platform-tests/blob/cd5bb894656e15405f1ab638cda9b24608e19f86/tools/wptserve/wptserve/config.py#L185)):*

I am a bit scared that changing the return type here is going to break stuff (again).
</blockquote></details>

There isn't that much that uses `domains` (or `not_domains`, etc.); you can quite easily grep through all of it and check.

---

*[tools/wptserve/wptserve/pipes.py, line 448 at r4](https://reviewable.io/reviews/w3c/web-platform-tests/10166#-LBLhHBI4yJ77jRzx3Gu:-LBLu1WvBXdrMoW3J483:b-wo4bkj) ([raw file](https://github.com/w3c/web-platform-tests/blob/cd5bb894656e15405f1ab638cda9b24608e19f86/tools/wptserve/wptserve/pipes.py#L448)):*

Should we just support `""` for the default domain, or should we have a named alias like `"default"`?
</blockquote></details>

I have no strong preference.

---


*Comments from [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10166)*

jgraham:
Reviewed 1 of 4 files at r5, 6 of 6 files at r6.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[tools/wptserve/wptserve/pipes.py, line 448 at r4](https://reviewable.io/reviews/w3c/web-platform-tests/10166#-LBLhHBI4yJ77jRzx3Gu:-LBM71se6pq_c_LtjHz_:b-knf65f) ([raw file](https://github.com/w3c/web-platform-tests/blob/cd5bb894656e15405f1ab638cda9b24608e19f86/tools/wptserve/wptserve/pipes.py#L448)):*

I have no strong preference.
</blockquote></details>

I think I err in the direction of having an alias, if it's not too much trouble. But don't worry if it is.

---


*Comments from [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10166)*

gsnedders:
Review status: 10 of 12 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[tools/wptserve/wptserve/pipes.py, line 445 at r4](https://reviewable.io/reviews/w3c/web-platform-tests/10166#-LBLhtAs0sOI7j0oNPs7:-LBNikwZ7-rDv7O7DtKf:b-896fix) ([raw file](https://github.com/w3c/web-platform-tests/blob/cd5bb894656e15405f1ab638cda9b24608e19f86/tools/wptserve/wptserve/pipes.py#L445)):*

This should be added to the docs.
</blockquote></details>

Done.

---

*[tools/wptserve/wptserve/pipes.py, line 448 at r4](https://reviewable.io/reviews/w3c/web-platform-tests/10166#-LBLhHBI4yJ77jRzx3Gu:-LBNimgd25d6oon_WwLo:b-cqkkfs) ([raw file](https://github.com/w3c/web-platform-tests/blob/cd5bb894656e15405f1ab638cda9b24608e19f86/tools/wptserve/wptserve/pipes.py#L448)):*

I think I err in the direction of having an alias, if it's not too much trouble. But don't worry if it is.
</blockquote></details>

I think I'm now erring against it, given it could easily lead to bugs with code expecting each thing in `domains` to be unique.

---


*Comments from [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10166)*

jgraham:
Reviewed 2 of 2 files at r7.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

*[tools/wptserve/wptserve/pipes.py, line 448 at r4](https://reviewable.io/reviews/w3c/web-platform-tests/10166#-LBLhHBI4yJ77jRzx3Gu:-LBQ34jq6ZQIRZwVOK3Y:bkhfpkg) ([raw file](https://github.com/w3c/web-platform-tests/blob/cd5bb894656e15405f1ab638cda9b24608e19f86/tools/wptserve/wptserve/pipes.py#L448)):*

I think I'm now erring against it, given it could easily lead to bugs with code expecting each thing in `domains` to be unique.
</blockquote></details>

OK, we can do it in a followup if there are issues.

---


*Comments from [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10166)*

jgraham (approved):
This looks good; happy for you to land it. Thanks!

PR #10167, opened 2018-03-25
Add tests for mathematical functions in <img sizes="">

domfarolino:
This should be enough to test all of the `math functions` here, for https://github.com/whatwg/html/pull/3084

## Edit

Also, I'm curious about the tests here. Currently we seem to be trying to test that various, sometimes media-condition adorned `sizes` attribute values resolve to the same computed value. We test this by checking that `currentSrc` for each image (in a particular group) is the same (since the computed value of `sizes` can affect the choice of image), however this is a little incomplete right?

I tested in Chrome and Firefox, and sometimes vastly different `sizes` attribute values have no effect on the chosen image. Doesn't this mean that even if the parsing of a particular `sizes` attribute failed (and we either chose some very wrong value, or fell back on the default `100vw`) we could still be choosing the correct image just based on how a particular user agent decides to choose an image? So we might fail parsing `sizes` properly, yet still pass a bunch of tests.

If this is the case, I'm wondering if it might be more beneficial to instead not have multiple sources in the `srcset`, and choose a particular non-100vw `sizes` value so we can reliably see if the actual computed width (maybe `clientWidth`?) of an image is what it should be after the [width descriptor](https://html.spec.whatwg.org/multipage/images.html#width-descriptor) is factored in with the *expected* `sizes` value?

### DIFF
--- a/html/semantics/embedded-content/the-img-element/sizes/sizes-iframed.sub.html
+++ b/html/semantics/embedded-content/the-img-element/sizes/sizes-iframed.sub.html
@@ -52,8 +52,14 @@
 <img srcset='/images/green-1x1.png?e34 50w, /images/green-16x16.png?e34 51w' sizes='\[,1px'>
 <img srcset='/images/green-1x1.png?e35 50w, /images/green-16x16.png?e35 51w' sizes='1\p\x'>
 <img srcset='/images/green-1x1.png?e36 50w, /images/green-16x16.png?e36 51w' sizes='calc(1px)'>
+<img srcset='/images/green-1x1.png?e36 50w, /images/green-16x16.png?e36 51w' sizes='min(1px, 100px)'>
+<img srcset='/images/green-1x1.png?e36 50w, /images/green-16x16.png?e36 51w' sizes='min(-100px, 1px)'>
 <img srcset='/images/green-1x1.png?e37 50w, /images/green-16x16.png?e37 51w' sizes='(min-width:0) calc(1px)'>
+<img srcset='/images/green-1x1.png?e37 50w, /images/green-16x16.png?e37 51w' sizes='(min-width:0) min(1px, 100px)'>
+<img srcset='/images/green-1x1.png?e37 50w, /images/green-16x16.png?e37 51w' sizes='(min-width:0) max(-100px, 1px)'>
 <img srcset='/images/green-1x1.png?e38 50w, /images/green-16x16.png?e38 51w' sizes='(min-width:calc(0)) 1px'>
+<img srcset='/images/green-1x1.png?e38 50w, /images/green-16x16.png?e38 51w' sizes='(min-width:min(0, 200vw)) 1px'>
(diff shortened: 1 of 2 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26423)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/357972368)

annevk:
@Wilto and @yoavweiss might have more suitable feedback. What you're suggesting sounds correct to me, but I haven't really studied this space much.

domfarolino:
Thanks. And to simplify what I was saying earlier (since it is a tad lengthy), consider the following test case:

```
<p>
<img srcset="test.png 600w, make.png 601w" sizes="40px">
<img srcset="test.png 600w, make.png 601w" sizes="min(40px, 80px)">
```

In this case, Chrome passes the test (Chrome just seems to always choose the src with the highest `w` descriptor) as it chooses the same source for each ``, even though the actual rendered sizes are completely different (because `min`, not yet implemented in Chrome, fails to parse and defaults to [`100vw`](https://cs.chromium.org/chromium/src/third_party/WebKit/Source/core/css/parser/SizesAttributeParser.cpp?sq=package:chromium&l=102)).

domfarolino:
As requested @annevk , this seems to be in limbo and I'm not familiar enough with this area to intelligently request reviewers. If you wanted to take a look at this perhaps that would be great, and if you need any clarification/information from me please let me know!

html/semantics/embedded-content/the-img-element/sizes/sizes-iframed.sub.html
```diff
@@ -52,8 +52,14 @@
 <img srcset='/images/green-1x1.png?e34 50w, /images/green-16x16.png?e34 51w' sizes='\[,1px'>
 <img srcset='/images/green-1x1.png?e35 50w, /images/green-16x16.png?e35 51w' sizes='1\p\x'>
 <img srcset='/images/green-1x1.png?e36 50w, /images/green-16x16.png?e36 51w' sizes='calc(1px)'>
+<img srcset='/images/green-1x1.png?e36 50w, /images/green-16x16.png?e36 51w' sizes='min(1px, 100px)'>
```
zcorpan: The number in the query should be unique so that we avoid having the browser prefer to load a candidate based on it already being in cache. You can use 36a or so.

zcorpan:
I think your analysis is correct and your suggested change seems interesting. I think I wanted to test selection, since that is the primary feature for sizes. I'm open to changing how we test sizes though.

zcorpan (approved):

domfarolino:
Good catch, I've addressed that. Should be all for now; should I maybe open an issue as a reminder to go back and address some of the things I previously brought up regarding testing that each group has the same image selected?

annevk:
Yeah, please.

PR #10209, opened 2018-03-28
Test document.write() on image and text documents

annevk:
See  https://bugzilla.mozilla.org/show_bug.cgi?id=1444872 for context.

### DIFF
--- a/html/dom/dynamic-markup-insertion/document-write/contentType.window.js
+++ b/html/dom/dynamic-markup-insertion/document-write/contentType.window.js
@@ -11,6 +11,7 @@ const videoURL = getVideoURI("/images/pattern"),
 ].forEach(val => {
   async_test(t => {
     const frame = document.body.appendChild(document.createElement("iframe"));
+    t.add_cleanup(() => frame.remove());
     frame.src = val[0];
     frame.onload = t.step_func_done(() => {
       assert_equals(frame.contentDocument.contentType, val[1]);
@@ -22,5 +23,5 @@ const videoURL = getVideoURI("/images/pattern"),
       // Make sure a load event is fired across browsers
       frame.contentDocument.close();
     });
-  }, "document.write() in a " + val[1] + " document");
+  }, "document.write(): " + val[1] + " document");
 });
--- a/html/dom/dynamic-markup-insertion/document-write/contentType.window.js
+++ b/html/dom/dynamic-markup-insertion/document-write/contentType.window.js
@@ -1,0 +1,16 @@
+[
+  ["/images/red.png", "image/png"],
+  ["/common/text-plain.txt", "text/plain"]
+].forEach(val => {
+  async_test(t => {
+    const frame = document.body.appendChild(document.createElement("iframe"));
+    frame.src = val[0];
+    frame.onload = t.step_func_done(() => {
+      assert_equals(frame.contentDocument.contentType, val[1]);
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26550)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/359312428)

bzbarsky:
@annevk You should call `document.close()` on those documents you open.  Otherwise the page never fires its load event, because it has still-loading iframes.

annevk:
I see, I wonder why it does pass in Chrome and Safari.

bzbarsky:
Because they don't treat opened-but-not-closed documents as "still loading".  Which arguably is a spec violation.

annevk:
Hmm, but since I do `write()` during the `load` event, why should the parent document wait for the subsequent load as well? Isn't the load event only delayed by the first load?

annevk:
HTML does have this note:

> If, during the handling of the load event, the browsing context in the iframe is again navigated, that will further delay the load event.

But the "document open steps" also say nothing of them delaying the load event, so dunno. I'd say this is undefined or a bug in Firefox...

bzbarsky:
> Isn't the load event only delayed by the first load?

The load event for the parent is delayed by any activity in the child, I would think.

> But the "document open steps" also say nothing of them delaying the load event

They do say to set the document readiness to "loading".  It's a bit weird to have that and not be considered as blocking onload.  

It's not super-clear to me whether close() fires load or not; I bet browsers differ....

More generally, this is a symptom of Firefox treating open() as "more like navigation" and Safari/Chrome treating it more as "mess with this existing thing" and the spec not having any idea what it wants to treat it as.

annevk:
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/17357282/
* https://bugs.webkit.org/show_bug.cgi?id=185240
* https://bugs.chromium.org/p/chromium/issues/detail?id=839258

html/dom/dynamic-markup-insertion/document-write/contentType.window.js
```diff
@@ -0,0 +1,27 @@
...
+    t.add_cleanup(() => frame.remove());
+    frame.src = val[0];
+    frame.onload = t.step_func_done(() => {
+      assert_equals(frame.contentDocument.contentType, val[1]);
+      frame.contentDocument.write("<b>Heya</b>");
+      assert_equals(frame.contentDocument.body.firstChild.localName, "b");
+      assert_equals(frame.contentDocument.body.firstChild.textContent, "Heya");
+      assert_equals(frame.contentDocument.contentType, val[1]);
+
+      // Make sure a load event is fired across browsers
```
foolip: Can you link to either spec or browser bugs here?

html/dom/dynamic-markup-insertion/document-write/contentType.window.js
```diff
@@ -0,0 +1,27 @@
...
+      frame.contentDocument.close();
+    });
+  }, "document.write(): " + val[1] + " document");
```
foolip: For the video, this will result in a different test name on different browsers. It'd be extra work to not have that be the case, so leaving it alone until we figure out a "rule" and how to enforce it seems right.

foolip (approved):
LGTM % link nit

PR #10226, opened 2018-03-28
WebDriver: add more WebKit workarounds to detect whether a window is fullscreen

burg:
WebKit doesn't conform to the Fullscreen API yet, so we need to use
the vendor prefixed property to check whether the window is fullscreen.

This workaround was previously applied to fullscreen_window.py, but it
needs to happen in other places as well.

### DIFF
--- a/webdriver/tests/contexts/maximize_window.py
+++ b/webdriver/tests/contexts/maximize_window.py
@@ -12,6 +12,12 @@ def maximize(session):
     return session.transport.send("POST", "session/%s/window/maximize" % session.session_id)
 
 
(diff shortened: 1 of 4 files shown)


burg:
Please advise if there is a better way to reuse the is_fullscreen helper method.

andreastt (approved):

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26597)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/359624980)

PR #10231, opened 2018-03-29
Make the normalized wptserve config a class

gsnedders:
This should make it easier once we have static typing to make sure we don't end up with non-normalized configs anywhere. This is essentially a subset of #10166, so this can land faster. (Though hopefully that'll land soon too.)

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10231)

### DIFF
--- a/tools/ci/make_hosts_file.py
+++ b/tools/ci/make_hosts_file.py
@@ -3,7 +3,7 @@ import os
 
 from ..localpaths import repo_root
 
-from ..serve.serve import load_config, normalise_config, make_hosts_file
+from ..serve.serve import load_config, make_hosts_file
 
 def create_parser():
     parser = argparse.ArgumentParser()
@@ -13,7 +13,5 @@ def create_parser():
 def run(**kwargs):
     config = load_config(os.path.join(repo_root, "config.default.json"),
                          os.path.join(repo_root, "config.json"))
-
-    config = normalise_config(config, {})
 
     print(make_hosts_file(config, kwargs["address"]))
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -23,8 +23,10 @@ import sslutils
 from manifest.sourcefile import read_script_metadata, js_meta_re
 from wptserve import server as wptserve, handlers
 from wptserve import stash
+from wptserve import config
 from wptserve.logger import set_logger
 from wptserve.handlers import filesystem_path, wrap_pipeline
+from wptserve.utils import get_port
 from mod_pywebsocket import standalone as pywebsocket
 
 def replace_end(s, old, new):
@@ -197,14 +199,6 @@ done();
 
 rewrites = [("GET", "/resources/WebIDLParser.js", "/resources/webidl2/lib/webidl2.js")]
 
-subdomains = [u"www",
-              u"www1",
-              u"www2",
-              u"天気の良い日",
-              u"élève"]
-
-not_subdomains = [u"nonexistent-origin"]
-
 class RoutesBuilder(object):
     def __init__(self):
         self.forbidden_override = [("GET", "/tools/runner/*", handlers.file_handler),
@@ -282,105 +276,6 @@ def build_routes(aliases):
     return builder.get_routes()
 
 
-def setup_logger(level):
-    import logging
-    global logger
-    logger = logging.getLogger("web-platform-tests")
-    logger.setLevel(getattr(logging, level.upper()))
-    set_logger(logger)
-
-
-def open_socket(port):
-    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-    if port != 0:
-        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-    sock.bind(('127.0.0.1', port))
-    sock.listen(5)
-    return sock
-
-def bad_port(port):
-    """
-    Bad port as per https://fetch.spec.whatwg.org/#port-blocking
-    """
-    return port in [
-        1,     # tcpmux
-        7,     # echo
-        9,     # discard
-        11,    # systat
-        13,    # daytime
-        15,    # netstat
-        17,    # qotd
-        19,    # chargen
-        20,    # ftp-data
-        21,    # ftp
-        22,    # ssh
-        23,    # telnet
-        25,    # smtp
-        37,    # time
-        42,    # name
-        43,    # nicname
-        53,    # domain
-        77,    # priv-rjs
-        79,    # finger
-        87,    # ttylink
-        95,    # supdup
-        101,   # hostriame
-        102,   # iso-tsap
-        103,   # gppitnp
-        104,   # acr-nema
-        109,   # pop2
-        110,   # pop3
-        111,   # sunrpc
-        113,   # auth
-        115,   # sftp
-        117,   # uucp-path
-        119,   # nntp
-        123,   # ntp
-        135,   # loc-srv / epmap
-        139,   # netbios
-        143,   # imap2
-        179,   # bgp
-        389,   # ldap
-        465,   # smtp+ssl
-        512,   # print / exec
-        513,   # login
-        514,   # shell
-        515,   # printer
-        526,   # tempo
-        530,   # courier
-        531,   # chat
-        532,   # netnews
-        540,   # uucp
-        556,   # remotefs
-        563,   # nntp+ssl
-        587,   # smtp
-        601,   # syslog-conn
-        636,   # ldap+ssl
-        993,   # imap+ssl
-        995,   # pop3+ssl
-        2049,  # nfs
-        3659,  # apple-sasl
-        4045,  # lockd
-        6000,  # x11
-        6665,  # irc (alternate)
-        6666,  # irc (alternate)
-        6667,  # irc (default)
-        6668,  # irc (alternate)
-        6669,  # irc (alternate)
-    ]
-
-def get_port():
-    port = 0
-    while True:
-        free_socket = open_socket(0)
-        port = free_socket.getsockname()[1]
-        free_socket.close()
-        if not bad_port(port):
-            break
-    logger.debug("Going to use port %s" % port)
-    return port
-
-
 class ServerProc(object):
     def __init__(self):
         self.proc = None
@@ -432,9 +327,11 @@ class ServerProc(object):
         return self.proc.is_alive()
 
 
-def check_subdomains(host, paths, bind_address, ssl_config, aliases):
-    port = get_port()
-    subdomains = get_subdomains(host)
+def check_subdomains(domains, paths, bind_address, ssl_config, aliases):
+    domains = domains.copy()
+    host = domains.pop("")
+    port = get_port(host)
+    logger.debug("Going to use port %d to check subdomains" % port)
 
     wrapper = ServerProc()
     wrapper.start(start_http_server, host, port, paths, build_routes(aliases), bind_address,
@@ -454,8 +351,7 @@ def check_subdomains(host, paths, bind_address, ssl_config, aliases):
                         "You may need to edit /etc/hosts or similar, see README.md." % (host, port))
         sys.exit(1)
 
-    for subdomain, (punycode, host) in subdomains.iteritems():
-        domain = "%s.%s" % (punycode, host)
+    for domain in domains.itervalues():
         try:
             urllib2.urlopen("http://%s:%d/" % (domain, port))
         except Exception as e:
@@ -464,18 +360,6 @@ def check_subdomains(host, paths, bind_address, ssl_config, aliases):
             sys.exit(1)
 
     wrapper.wait()
-
-
-def get_subdomains(host):
-    #This assumes that the tld is ascii-only or already in punycode
-    return {subdomain: (subdomain.encode("idna"), host)
-            for subdomain in subdomains}
-
-
-def get_not_subdomains(host):
-    #This assumes that the tld is ascii-only or already in punycode
-    return {subdomain: (subdomain.encode("idna"), host)
-            for subdomain in not_subdomains}
 
 
 def make_hosts_file(config, host):
@@ -630,80 +514,14 @@ def start_wss_server(host, port, paths, routes, bind_address, config, ssl_config
                            ssl_config)
 
 
-def get_ports(config, ssl_environment):
-    rv = defaultdict(list)
-    for scheme, ports in config["ports"].iteritems():
-        for i, port in enumerate(ports):
-            if scheme in ["wss", "https"] and not ssl_environment.ssl_enabled:
-                port = None
-            if port == "auto":
-                port = get_port()
-            else:
-                port = port
-            rv[scheme].append(port)
-    return rv
-
-
-
-def normalise_config(config, ports):
-    if "host" in config:
-        logger.warning("host in config is deprecated; use browser_host instead")
-        host = config["host"]
-    else:
-        host = config["browser_host"]
-
-    domains = get_subdomains(host)
-    not_domains = get_not_subdomains(host)
-
-    ports_ = {}
-    for scheme, ports_used in ports.iteritems():
-        ports_[scheme] = ports_used
-
-    for key, value in domains.iteritems():
-        domains[key] = ".".join(value)
-
-    for key, value in not_domains.iteritems():
-        not_domains[key] = ".".join(value)
-
-    domains[""] = host
-
-    if "bind_hostname" in config:
-        logger.warning("bind_hostname in config is deprecated; use bind_address instead")
-        bind_address = config["bind_hostname"]
-    else:
-        bind_address = config["bind_address"]
-
-    # make a (shallow) copy of the config and update that, so that the
-    # normalized config can be used in place of the original one.
-    config_ = config.copy()
-    config_["domains"] = domains
-    config_["not_domains"] = not_domains
-    config_["ports"] = ports_
-    config_["bind_address"] = bind_address
-    if config.get("server_host", None) is None:
-        config_["server_host"] = host
-    return config_
-
-
-def get_paths(config):
-    return {"doc_root": config["doc_root"],
-            "ws_doc_root": config["ws_doc_root"]}
-
-
-def get_ssl_config(config, ssl_environment):
-    external_domains = config["domains"].values()
-    key_path, cert_path = ssl_environment.host_cert_path(external_domains)
-    return {"key_path": key_path,
-            "cert_path": cert_path,
-            "encrypt_after_connect": config["ssl"]["encrypt_after_connect"]}
-
-
 def start(config, ssl_environment, routes, **kwargs):
     host = config["server_host"]
-    ports = get_ports(config, ssl_environment)
-    paths = get_paths(config)
+    ports = config.ports
+    paths = config.paths
     bind_address = config["bind_address"]
-    ssl_config = get_ssl_config(config, ssl_environment)
+    ssl_config = config.ssl_config
+
+    logger.debug("Using ports: %r" % ports)
 
     servers = start_servers(host, ports, paths, routes, bind_address, config,
                             ssl_config, **kwargs)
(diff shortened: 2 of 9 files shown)


w3c-bots:
# Build BROKEN

Started: 2018-03-29 04:40:35
Finished: 2018-03-29 04:54:05


<h2>Failing Jobs</h2>
<ul>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/26607)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/359683717)

jgraham:
Reviewed 7 of 7 files at r1, 1 of 1 files at r2.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

*[tools/serve/config.py, line 14 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LCIAQ4SScadj5nsRF:-L9LCIAQ4SScadj5nsRG:b-wmjrdx) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L14)):*
> ```Python
> 
> 
> with open(os.path.join(repo_root, "config.default.json"), "rb") as _fp:
> ```

I'm not thrilled by having this IO at the top level. Put it on the class and cache the result.

---

*[tools/serve/config.py, line 41 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LC0Mz-PN4CHmyyvvr:-L9LC0N-3zbQ5h5gicpY:b-cubwn3) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L41)):*
> ```Python
> 
>     def __init__(self, logger=None, **kwargs):
>         if logger is None:
> ```

I don't think a config class should eb setting up logging.

---

*[tools/serve/config.py, line 65 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LC_2i1VJ8S3C2rDfW:-L9LC_2i1VJ8S3C2rDfX:bncqun3) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L65)):*
> ```Python
>         if kwargs:
>             k = next(iter(kwargs))
>             raise TypeError("__init__() got an unexpected keyword argument '%s'" % k)
> ```

I would provide the full list of unexpected arguments (i.e. `kwargs.keys()` even though it's less like the built in message.

---

*[tools/serve/config.py, line 74 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LCpcA7P-AStd-f0xg:-L9LCpcA7P-AStd-f0xh:b-87w2km) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L74)):*
> ```Python
> 
>     def __iter__(self):
>         return iter([x for x in dir(self) if not x.startswith("_")])
> ```

Isn't this also going to return `logger`? It seems pretty fragile. One option would be to reverse the setup  so the config is still stored in a dict and we override the attribute getters and setters to use that dict if nothing is present on the class.

---

*[tools/serve/config.py, line 79 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LDefq4VO8MWF8AVKJ:-L9LDefrEyrCsGL4nZXT:b7qtl15) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L79)):*
> ```Python
>         return len([x for x in dir(self) if not x.startswith("_")])
> 
>     def load_overrides(self, override):
> ```

This could be called `update`

---

*[tools/serve/config.py, line 116 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LEJKG2sLonDTXGpV4:-L9LEJKH7Qa1VzUfilQ9:bnji4k4) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L116)):*
> ```Python
>                     port = None
>                 if port == "auto":
>                     port = serve.get_port()
> ```

Is this actually giving a consistent answer back if it's called multiple times?

---

*[tools/serve/config.py, line 156 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LGRL3BHOuhZT1ZMLk:-L9LGRL4B64KK6WSEBL2:blefo0h) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L156)):*
> ```Python
>     def domains(self):
>         domains = {k: ".".join(v)
>                    for k, v in serve.get_subdomains(self.browser_host).iteritems()}
> ```

I wonder whehter it makes sense to make these functions part of the config object.

---

*[tools/serve/config.py, line 173 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LGa3k-LIRp8nyyEna:-L9LGa3l-JLAI4cGEV-J:bbi5m93) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L173)):*
> ```Python
> 
>     @property
>     def domains_set(self):
> ```

I don't love having a separate property just to perform a type conversion.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231)*

gsnedders:
I'm increasingly convinced that this actually belongs in `wptserve` and not in `serve`? `wptserve` relies on the config, so it should probably live there rather than in `serve`.

However, we currently have a few things that rely on serve:

* `serve.get_subdomains` (and `get_not_subdomains`), but these are easily moved to config.py and made private (there's no need for them to be public, I don't think?)
* `serve.get_port`, which is used both by the config to return the generated ports and by `serve` when checking the hosts are set up and when starting the stash server. The latter of these could potentially be moved into the config as a default "auto" port, but that still leaves the former?

---

Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[tools/serve/config.py, line 41 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LC0Mz-PN4CHmyyvvr:-L9U_XZe50Yy6HHYJ1BK:b-upbfvf) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L41)):*

I don't think a config class should eb setting up logging.
</blockquote></details>

So where should we get the logger from before running `__init__`, given it depends on configuration options (`log_level`) and we want to log if you use a deprecated config option?

---

*[tools/serve/config.py, line 116 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LEJKG2sLonDTXGpV4:-L9UhSiHDdkV7gJqcVXZ:bwy8r5o) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L116)):*

Is this actually giving a consistent answer back if it's called multiple times?
</blockquote></details>

No. 🙁

---

*[tools/serve/config.py, line 156 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LGRL3BHOuhZT1ZMLk:-L9UhXRT4NRw1lEfN-Uo:b-47q5d8) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L156)):*

I wonder whehter it makes sense to make these functions part of the config object.
</blockquote></details>

IMO, yes. Where else does it make sense to put them?

---

*[tools/serve/config.py, line 173 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LGa3k-LIRp8nyyEna:-L9Uhdhg1y18GxHSYM7V:b-7co7a7) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L173)):*

I don't love having a separate property just to perform a type conversion.
</blockquote></details>

To be fair, they originally made more sense on the `alternate-hosts` branch, but I think now don't? So yeah, should probably go.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-:-L9UhqBRB_tsUr3DpQIi:b-27bt00)*

jgraham:
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[tools/serve/config.py, line 41 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LC0Mz-PN4CHmyyvvr:-L9aWicE6TV8dfbrl_P-:b-t9oegi) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L41)):*

So where should we get the logger from before running `__init__`, given it depends on configuration options (`log_level`) and we want to log if you use a deprecated config option?
</blockquote></details>

OK, I guess. It still seems weird for something that looks like a pure view of data to set up logging, but I guess that makes sense.

---

*[tools/serve/config.py, line 156 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LGRL3BHOuhZT1ZMLk:-L9aWuczBXWZgOWNsoKe:bcpj1mz) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L156)):*

IMO, yes. Where else does it make sense to put them?
</blockquote></details>

I mean `serve.get_subdomains` &c. Although I'm not sure what actually works well here; as you say this class feels like it belongs in wptserve. Possibly they should be passed in since presumably they get passed to the server as well.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231)*

gsnedders:
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[tools/serve/config.py, line 42 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9dW5qj2X6xbfVNEe3Z:-L9dW5qj2X6xbfVNEe3_:b-n7cok1) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L42)):*
> ```Python
>     def __init__(self, logger=None, **kwargs):
>         if logger is None:
>             logger = logging.getLogger("web-platform-tests")
> ```

@jgraham You mentioned on IRC you didn't like this, though I justified why. I think there's three other options: one is we make every warning into a (fatal) exception with no logger, another is we make everything into a Python warning (which IMO makes little sense), and the final is to just silently ignore the warning. I think I'm happy with the first or last? It does mean that usage of the Config class will need to create their own Logger, reading the object themself.

---

*[tools/serve/config.py, line 116 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LEJKG2sLonDTXGpV4:-L9dX1hc4BBKGBqahv6W:bx567d) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L116)):*

No. 🙁
</blockquote></details>

Also moving the class to `wptserve` means we add a dependency to `serve` for `get_port`, or perhaps more sensibly move all of the get_port stuff to somewhere in wptserve (where, though?).

---

*[tools/serve/config.py, line 156 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LGRL3BHOuhZT1ZMLk:-L9dXisxCPFvmumlig1R:b-flkl36) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L156)):*

I mean `serve.get_subdomains` &c. Although I'm not sure what actually works well here; as you say this class feels like it belongs in wptserve. Possibly they should be passed in since presumably they get passed to the server as well.
</blockquote></details>

I'm unconvinced by things that are essentially "get this computed config value", which `get_subdomains` is. My feeling is we have a `subdomains` and `not_subdomains` config option that `serve` sets for us (though nothing actually ever reads it; everything always just reads `domains`, but that's not unique—there's other places where the computed value is always the one used).

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231)*

gsnedders:
Review status: 3 of 9 files reviewed at latest revision, 8 unresolved discussions, some commit checks pending.

---

*[tools/wptserve/tests/test_config.py, line 26 at r3](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9dw8qcBF4LfggYw9Ob:-L9dw8qcBF4LfggYw9Oc:b-edlqsd) ([raw file](https://github.com/w3c/web-platform-tests/blob/ce02f621b1ba703971a29e9df61a3a9410d25331/tools/wptserve/tests/test_config.py#L26)):*
> ```Python
>     ({"a": {"b": 1}}, {"a": {"b": 2, "c": 3}}, {"a": {"b": 2}}),
>     pytest.param({"a": {"b": 1}}, {"a": 2}, {"a": 1}, marks=pytest.mark.xfail),
>     pytest.param({"a": 1}, {"a": {"b": 2}}, {"a": 1}, marks=pytest.mark.xfail),
> ```

FYI: these xfails already exist on `master`, we just don't have tests for it before.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231)*

gsnedders:
Now in `wptserve`.

---

Review status: 3 of 9 files reviewed at latest revision, 9 unresolved discussions, all commit checks successful.

---

*[tools/serve/serve.py, line 380 at r3](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9eFHX43hCRLvJGkPhi:-L9eFHX43hCRLvJGkPhj:b-tcanun) ([raw file](https://github.com/w3c/web-platform-tests/blob/95992cd324d80acc68707b203be8c6cc2aebc6d5/tools/serve/serve.py#L380)):*
> ```Python
>         if not bad_port(port):
>             break
>     logger.debug("Going to use port %s" % port)
> ```

Note moving this to wptserve has lost this logging (because wptserve.utils doesn't have a global logger). We should probably move this elsewhere and give a better message in general (as to what we're going to use the port for!).

---

*[tools/serve/config.py, line 74 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LCpcA7P-AStd-f0xg:-L9eF_NT2eABRcRfg78l:b-a087rk) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L74)):*

Isn't this also going to return `logger`? It seems pretty fragile. One option would be to reverse the setup  so the config is still stored in a dict and we override the attribute getters and setters to use that dict if nothing is present on the class.
</blockquote></details>

If we want to view `logger` as a config item (given it's used by various things), then surely it should be present here.

---

*[tools/serve/config.py, line 79 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LDefq4VO8MWF8AVKJ:-L9eFkvLD3RxapB9IIEy:b-896fix) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L79)):*

This could be called `update`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-:-L9eErUQ0eqhZeowZgoX:b-cekskq)*

jgraham:
Reviewed 4 of 10 files at r3, 2 of 3 files at r4.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks pending.

---

*[tools/serve/serve.py, line 581 at r4](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9eIESX5g-54Cg8TCA_:-L9eIESX5g-54Cg8TCAa:bf4r0p9) ([raw file](https://github.com/w3c/web-platform-tests/blob/951ade940eacfdad4de7530058690a10637a277d/tools/serve/serve.py#L581)):*
> ```Python
> _not_subdomains = {u"nonexistent-origin"}
> 
> def set_serve_config(config):
> ```

Why don't we just pass these in through the constructor?

---

*[tools/serve/config.py, line 156 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LGRL3BHOuhZT1ZMLk:-L9eI0Pe8d3njqvnwRxm:b-oic749) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L156)):*

I'm unconvinced by things that are essentially "get this computed config value", which `get_subdomains` is. My feeling is we have a `subdomains` and `not_subdomains` config option that `serve` sets for us (though nothing actually ever reads it; everything always just reads `domains`, but that's not unique—there's other places where the computed value is always the one used).
</blockquote></details>

This is fine now subdomains are stored on this object.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231)*

gsnedders:
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[tools/serve/serve.py, line 581 at r4](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9eIESX5g-54Cg8TCA_:-L9eL9hn0wmnNdIi34xT:b-ttc3n3) ([raw file](https://github.com/w3c/web-platform-tests/blob/951ade940eacfdad4de7530058690a10637a277d/tools/serve/serve.py#L581)):*

Why don't we just pass these in through the constructor?
</blockquote></details>

Because we also construct it in `wptrunner.environment.TestEnvironment.load_config`; we don't necessarily get the config from `serve.load_config`.

---

*[tools/serve/config.py, line 65 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LC_2i1VJ8S3C2rDfW:-L9eN77N9Pm8v3yPH1OC:b-896fix) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L65)):*

I would provide the full list of unexpected arguments (i.e. `kwargs.keys()` even though it's less like the built in message.
</blockquote></details>

Done.

---

*[tools/serve/config.py, line 116 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LEJKG2sLonDTXGpV4:-L9eJklh1vTln-ya2E1Z:b-896fix) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L116)):*

Also moving the class to `wptserve` means we add a dependency to `serve` for `get_port`, or perhaps more sensibly move all of the get_port stuff to somewhere in wptserve (where, though?).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231)*

gsnedders:
Review status: 7 of 9 files reviewed at latest revision, 10 unresolved discussions, some commit checks pending.

---

*[tools/serve/config.py, line 14 at r2](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9LCIAQ4SScadj5nsRF:-L9eNiNQ8aj5vZZexN4Z:b-896fix) ([raw file](https://github.com/w3c/web-platform-tests/blob/47c97a405921a671c3ca6ce73818f3a8b5bd8b25/tools/serve/config.py#L14)):*

I'm not thrilled by having this IO at the top level. Put it on the class and cache the result.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231)*

gsnedders:
Review status: 4 of 9 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[tools/serve/serve.py, line 581 at r4](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231#-L9eIESX5g-54Cg8TCA_:-L9ees8BCQvuUkCRIgRH:bn2oejb) ([raw file](https://github.com/w3c/web-platform-tests/blob/951ade940eacfdad4de7530058690a10637a277d/tools/serve/serve.py#L581)):*

Because we also construct it in `wptrunner.environment.TestEnvironment.load_config`; we don't necessarily get the config from `serve.load_config`.
</blockquote></details>

Done. Moved everything to use `serve.Config` which is a subclass of `wptserve.config.Config` which sets all the `serve` config correctly.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231)*

jgraham:
Reviewed 4 of 6 files at r5, 2 of 2 files at r6.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/w3c/web-platform-tests/10231)*

jgraham (approved):

PR #10232, opened 2018-03-29
[css-align] tests for row-gap / column-gap normal

frivoal:
Matching the https://github.com/w3c/csswg-drafts/issues/2294 spec change

Not including tests for used value in multicol, because (so far) it is
UA defined.

### DIFF
--- a/css/css-align/gaps/gap-normal-computed-001.html
+++ b/css/css-align/gaps/gap-normal-computed-001.html
@@ -14,7 +14,7 @@
      and there are implementations that support column-gap without supporting the shorthand */
   colum-gap: normal;
   row-gap: normal;
-  float: right; /* for shrinwrap*/
+  float: right; /* for shrinkwrap*/
 }
 #col {
   column-count: 2;
--- a/css/css-align/gaps/gap-normal-computed-001.html
+++ b/css/css-align/gaps/gap-normal-computed-001.html
@@ -1,0 +1,72 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Box Alignment Test: computed value of normal on *-gap properties</title>
+<link rel="author" title="Florian Rivoal" href="https://florian.rivoal.net">
+<link rel="help" href="https://www.w3.org/TR/css-align-3/#column-row-gap">
+<meta assert="The computed value of [row-|column-]?gap is normal for all elements it applies to. Checking explicitely because earlier version of the spec called for 0px in some cases.">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<style>
+#col,
+#grid,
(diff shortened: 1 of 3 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26610)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/359694037)

css/css-align/gaps/gap-normal-computed-001.html
```diff
@@ -0,0 +1,72 @@
...
+<script src="/resources/testharnessreport.js"></script>
+<style>
+#col,
+#grid,
+#flex {
+  /* Not using the shorthand because that's not what we're interested in,
+     and there are implementations that support column-gap without supporting the shorthand */
+  colum-gap: normal;
+  row-gap: normal;
+  float: right; /* for shrinwrap*/
```
emilio: typo: Shrinkwrap.

css/css-align/gaps/gap-normal-used-001.html
```diff
@@ -0,0 +1,31 @@
...
+  position: absolute;
+}
+#grid * { background: green; }
+#red {
+  width: 100px;
+  height: 100px;
+  background: red;
+}
+</style>
+<body>
```
emilio: nit: This `<body>` is unclosed, but I'd just remove it here and everywhere else... Your call.

emilio (approved):
Looks good to me, it'd be nice to add references that instead of a green square just use `column-gap: 0` and `row-gap: 0`?

Not a big deal I guess, the tests are clear anyway.

css/css-align/gaps/gap-normal-computed-001.html
```diff
@@ -0,0 +1,72 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Box Alignment Test: computed value of normal on *-gap properties</title>
+<link rel="author" title="Florian Rivoal" href="https://florian.rivoal.net">
+<link rel="help" href="https://www.w3.org/TR/css-align-3/#column-row-gap">
```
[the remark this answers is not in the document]
frivoal: I believe the policy is somewhat fuzzy, but that linking to TR when the text we want is on TR is preferred, and that linking to the ED otherwise is OK. I wouldn't lose sleep over it :)

css/css-align/gaps/gap-normal-computed-001.html
frivoal: thanks

css/css-align/gaps/gap-normal-used-001.html
frivoal: removing it altogether

PR #10239, opened 2018-03-29
document.open() and the load event

annevk:

### DIFF
--- a/html/dom/dynamic-markup-insertion/opening-the-input-stream/load.window.js
+++ b/html/dom/dynamic-markup-insertion/opening-the-input-stream/load.window.js
@@ -1,0 +1,21 @@
+async_test(t => {
+  const frame = document.createElement("iframe");
+  frame.onload = t.step_func(() => {
+    frame.contentDocument.open();
+    frame.contentDocument.write("x");
+  });
+
+  self.addEventListener("load", t.step_func_done(e => {
+    assert_true(e.whoa);
+    assert_equals(frame.contentDocument.body.textContent, "x");
+  }));
+
+  document.body.appendChild(frame);
+
+  t.step_timeout(() => {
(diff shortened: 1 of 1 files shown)


annevk:
@bzbarsky I tried to figure out what browsers do with the load event because of your comments in #10209 and I'm not much wiser. This test passes in Safari and fails everywhere else (it fails in Firefox because Firefox loses the written content; it fails in Chrome and Edge because they dispatch a normal load event on the window).

I did notice that Chrome has some logic for the load event here: https://cs.chromium.org/chromium/src/third_party/WebKit/Source/core/dom/Document.cpp?l=3031. And indeed if I don't wait for the `iframe`'s load event, but `open()` directly, the window's load event won't fire and that seems interoperable across browsers. (I will add that test later as it seems good to have as a baseline.)

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26631)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/359820240)

bzbarsky:
> it fails in Firefox because Firefox loses the written content

It fails in Firefox because `step_timeout` with a 0 delay fires before the load event for the frame and the open/write that load event triggers.  Which is totally allowed per spec, because they use different task sources ("DOM manipulation task source" for the load event, "timer task source" for the `setTimeout(func, 0)` call).

Furthermore, as the test is written right now, the `step_timeout` bit can even run before parsing of the HTML completes, hence definitely before the load event for the page would fire no matter what's happening with the subframe...

annevk:
I adjusted the test a bit, but now it fails in no browser.

I think the main difference between browsers is what happens when you invoke the document open steps during the frame's load event.

I'll add something for that.

PR #10240, opened 2018-03-29
Check inherited interfaces exist

lukebjerring:
Ensures that `B` exists if you `add_[untested_]idl('interface A : B {};')`
See https://www.w3.org/TR/WebIDL-1/#idl-interfaces for definition of interface inheritance.

Fixes https://github.com/w3c/web-platform-tests/issues/10214

Note:
Side-stepping the `Cannot read property 'has_extended_attribute' of undefined` message at its source would involve non-null checks in several places (e.g. l1318, l1524, ...).
This change is consistent with the other early existence checks, and is a fail-fast (before testing each member), which is likely to change the number of executed tests for any IDL tests which have undefined inherits.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10240)

### DIFF
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -603,7 +603,6 @@ IdlArray.prototype.assert_throws = function(error, idlArrayFunc)
 {
     try {
         idlArrayFunc.call(this, this);
-        throw new IdlHarnessError(`${idlArrayFunc} did not throw the expected IdlHarnessError`);
     } catch (e) {
         if (e instanceof AssertionError) {
             throw e;
@@ -613,9 +612,11 @@ IdlArray.prototype.assert_throws = function(error, idlArrayFunc)
             error = error.message;
         }
         if (e.message !== error) {
-            throw new IdlHarnessError(`${idlArrayFunc} threw ${e}, not the expected IdlHarnessError`);
+            throw new IdlHarnessError(`${idlArrayFunc} threw "${e}", not the expected IdlHarnessError "${error}"`);
         }
+        return;
     }
+    throw new IdlHarnessError(`${idlArrayFunc} did not throw the expected IdlHarnessError`);
 }
 
 //@}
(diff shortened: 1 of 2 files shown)


lukebjerring:
https://github.com/w3c/web-platform-tests/pull/9853 is an example of test output which, before this change, wasn't obvious which missing interface caused the problem.

w3c-bots:
# Build PASSED

Started: 2018-03-29 15:01:44
Finished: 2018-03-29 15:07:40





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26638)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/359881406)

resources/idlharness.js
```diff
@@ -682,6 +683,16 @@ IdlArray.prototype.test = function()
     }
     this["includes"] = {};
 
+    // Assert B defined for A : B
+    for (var member of Array.from(Object.values(this.members)).filter(m => m.base)) {
```
foolip: Object.values already returns a vanilla Array, so the Object.from ought not be needed.

resources/idlharness.js
```diff
@@ -613,9 +612,11 @@ IdlArray.prototype.assert_throws = function(error, idlArrayFunc)
             error = error.message;
         }
         if (e.message !== error) {
-            throw new IdlHarnessError(`${idlArrayFunc} threw ${e}, not the expected IdlHarnessError`);
+            throw new IdlHarnessError(`${idlArrayFunc} threw "${e}", not the expected IdlHarnessError "${error}"`);
```
foolip: Can you mention this drive-by fixing in the commit message when merging? Changes themselves look good!

foolip (approved):
More tests for idlharness.js, yay! LGTM % nit.

foolip:
Some breakage from this reported in https://bugs.chromium.org/p/chromium/issues/detail?id=827367

foolip:
Since reverting this won't break anything and a fix would require review, I'll go ahead and revert. This would be the  "Breakage in browser repos like Chromium which import wpt" case in https://github.com/w3c/web-platform-tests/issues/9953.

inexorabletash:
+1 to revert. 

FWIW, I did some testing with a local fix:
```diff
         if (!(lhs in this.members)) throw new IdlHarnessError(`${lhs} inherits ${rhs}, but ${lhs} is undefined.`);
-        if (!(this.members[lhs] instanceof IdlInterface)) throw new IdlHarnessError(`${lhs} inherits ${rhs}, but ${lhs} is not an interface.`);
         if (!(rhs in this.members)) throw new IdlHarnessError(`${lhs} inherits ${rhs}, but ${rhs} is undefined.`);
-        if (!(this.members[rhs] instanceof IdlInterface)) throw new IdlHarnessError(`${lhs} inherits ${rhs}, but ${rhs} is not an interface.`);
+        let lhs_is_interface = this.members[lhs] instanceof IdlInterface;
+        let rhs_is_interface = this.members[rhs] instanceof IdlInterface;
+        if (rhs_is_interface != lhs_is_interface) {
+            if (!lhs_is_interface) throw new IdlHarnessError(`${lhs} inherits ${rhs}, but ${lhs} is not an interface.`);
+            if (!rhs_is_interface) throw new IdlHarnessError(`${lhs} inherits ${rhs}, but ${rhs} is not an interface.`);
+        }
```
... and there *are* a handful of interface tests that do have LHS with an undefined RHS (e.g. _object "GamepadEventInit inherits EventInit, but EventInit is undefined."_) so when a corrected version relands we will still see some real failures. (Which is a good thing!)

foolip:
Landing this change in Chromium and exporting might be easier, tells us right away which tests will be affected.

lukebjerring:
I'll take the local fix and go via Chromium.

P.S. didn't know github supported diffs. Cool.

PR #10248, opened 2018-03-29
Build spec links with short URLs for CSS testsuites

csnardi:
Many spec links were in shortlink form (e.g. https://drafts.csswg.org/cssom/ instead of https://drafts.csswg.org/cssom-1/). However, build.py did not pick up these spec links, as it assumed that the only spec links could be in longhand form with the spec version. Update build.py to consider shortlinks when building.

### DIFF
--- a/css/tools/build.py
+++ b/css/tools/build.py
@@ -90,22 +90,24 @@ class Builder(object):
                         officialURL = officialURL[:-1]
                     self.specNames[officialURL.lower()] = [specName]
 
-                    shortOfficialURL = officialURL.lower().replace(specName, shortName)
-                    if (shortOfficialURL in self.specNames):
-                        self.specNames[shortOfficialURL].append(specName)
-                    else:
-                        self.specNames[shortOfficialURL] = [specName]
+                    if specName in officialURL:
+                        shortOfficialURL = officialURL.lower().replace(specName, shortName)
+                        if (shortOfficialURL in self.specNames):
+                            self.specNames[shortOfficialURL].append(specName)
+                        else:
+                            self.specNames[shortOfficialURL] = [specName]
                 draftURL = self._normalizeScheme(specData.get('draft_uri'))
                 if (draftURL):
                     if (draftURL.endswith('/')):
                         draftURL = draftURL[:-1]
                     self.specNames[draftURL.lower()] = [specName]
 
-                    shortDraftURL = draftURL.lower().replace(specName, shortName)
-                    if (shortDraftURL in self.specNames):
-                        self.specNames[shortDraftURL].append(specName)
-                    else:
-                        self.specNames[shortDraftURL] = [specName]
+                    if specName in draftURL:
+                        shortDraftURL = draftURL.lower().replace(specName, shortName)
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PASSED

Started: 2018-03-29 23:26:18
Finished: 2018-03-29 23:30:46


<h2>Failing Jobs</h2>
<ul>

<li>build_css</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/26675)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/360085129)

csnardi:
Hmm, this doesn't seem to fix it; not ready for review then.

plinss:
@csnardi the linking to versioned rather than un-versioned spec links was a conscious design choice. The un-versioned URL changes over time to point to the latest spec where the test may have been written against an old version with different requirements.

If there are tests using un-versioned urls, then the tests should be changed rather than the build system.

Also note, the build will also currently reject spec links that don't have an anchor. If you're seeing tests fail to get picked up, that may also be a factor. There is a plan to accept tests that aren't linked to an anchor, but this will also require other changes since the rest of the tooling currently presumes that each test is linked to a specific section of the spec (derived form the anchor).

csnardi:
@plinss Certain places (e.g. the top of https://www.w3.org/TR/cssom-1/) link directly to unversioned editor's draft links. I unintentionally introduced a bunch of these into the cssom test suite in https://github.com/w3c/web-platform-tests/commit/825f054d0d8f1f60ef13e36b46d6ea97ea87cc15, as I followed that link and assumed it would just work.

I would also argue that it makes more sense for unversioned links to be applied to all specs with that short name, as CSS generally builds off of the previous versions of a spec. If behavior changes, that would seem to suggest that either the test should be dropped (as it's no longer useful), or that the spec link should be updated to explicitly state which versions it applies to. But in most cases, tests from prior versions should still be applicable to future versions, e.g. `rgb` tests from Color 3 should apply to Color 4.

plinss:
But tests written for CSS Color 4 wouldn't work against a client that only implements CSS Color 3. If all the tests link to css-color, then there's no way to reason about them. 

Also note that sometimes anchor IDs change between versions so a link to https://drafts.csswg.org/css-color-3/#foo may not link to https://drafts.csswg.org/css-color-4/ or https://drafts.csswg.org/css-color-5/ where the #foo anchor may not exist or potentially even refer to something else. (Yes, we try to keep anchors stable, but that doesn't always happen, or it may have moved to another css module for example, which is common.)

plinss:
Also, the links to un-versioned draft links, for example the "Editors Draft" or "Latest Version" links are quite deliberate when we want to link people to the latest version. This is not to say that all links should be un-versioned, especially when referring to something in a specific version, like when testing a version of a spec.

csnardi:
I suppose then that this should be another lint, as I see more of these from other authors. You make valid points, but tooling should be helping authors, not making them do more work (especially when the easiest ways to find spec links results in intentional failure of the tools). This probably is relatively unimportant given the status of the CSS build system these days, though.

PR #10250, opened 2018-03-30
Test for used column-count when column-count/width are not auto

gsnedders:
Inspired by a bug in Prince (which I think is just what's reported at https://www.princexml.com/forum/topic/3804/column-count-anomalies-any-ideas).


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26680)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/360129089)

css/css-multicol/column-count-used-001-ref.html
```diff
@@ -0,0 +1,17 @@
+<!doctype html>
+<title>column-count used value test ref</title>
+<style>
+ul {
+    font: 20px/1 Ahem;
```
mstensho: To simplify the test, we could just use DIVs with a blank line created by BR elements. We don't need Ahem, and also not UL or LI. If you set orphans and widows to 1, the contents of the multicol container could simply be
```
<br><br><br><br>
```
(i.e. no need for any LI/DIV wrapper for each line)

It would be even better with a test that would just match one of the refs in ../reference/ , of course ("green square" something).

gsnedders: @mstensho does this look better? oddly, it fails in Safari, as the green box is 108px high, and I don't know if I'm missing something as to why.

css/css-multicol/column-count-used-001.html
```diff
@@ -1,19 +1,26 @@
...
+div {
+    position: absolute;
+}
+.bg {
+    background: red;
+    width: 100px;
+    height: 100px;
+}
+.test {
+    font: 50px/1 sans-serif;
```
mstensho: Use line-height:50px instead of specifying font, and there should be less trouble.

css/css-multicol/column-count-used-001.html
```diff
@@ -1,19 +1,26 @@
...
+    width: 100px;
+    background: green;
+    columns: 2 20px;
```
mstensho: This is pretty over-constrained, since the available width is already specified. I suggest "columns:2".

mstensho (changes requested):

css/css-multicol/column-count-used-001.html
gsnedders: This is specifically testing the case where column-width is not auto, because that's the case where Prince gets it wrong.

mstensho: Doh, right. I forgot what this was about.

mstensho (approved):

PR #10275, opened 2018-04-02
Add tests for rel, rellist, download, and text on SVGAElement

dstorey:
These are based on the equivalent tests in HTML's Anchor folder with minimal changes to put in SVG template and creating elements in the SVG namespace. 

These tests are a precursor to moving these members to a mixin defined in HTML, and as such are expected to fail in current implementations.


svg/linking/scripted/a-download-click.svg
```diff
@@ -0,0 +1,33 @@
...
+        frame.addEventListener("load", t.step_func(function () {
+            frame.contentWindow.addEventListener(
+                "beforeunload", t.unreached_func("Navigated instead of downloading"));
+            const string = "test";
+            const blob = new Blob([string], { type: "text/html" });
+
+            const link = frame.contentDocument.querySelector("#blob-url");
+            link.href.baseVal = URL.createObjectURL(blob);
+
+            link.click();
```
dstorey: Note: we haven't moved click() into the HTMLOrSVGElement mixin yet, so this will fail even if download is supported. I wasn't sure how to force click the download link without this and didn't want to change too much from the original test. We've discussed adding click() to the mixin, which I think we probably should do (will be trivial to implement in Edge at least)

w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26744)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/361408644)

dstorey:
@longsonr thanks for the spot, and the implementation. I've just updated it.

For the click() issue, are you interested in implementing that in Firefox? If so I'll open a PR to add it to Edge and we'll have the two implementations. I'll also update the HTML GitHub issue for HTMLorSVGElement to get click() added to that mixin. 

Otherwise we'll need to find another way to test download attribute.

longsonr:
While I could implement click() support provided it was added to the specification and there was agreement to implement, which it sounds as if there might be. Wouldn't it be better for now to just change the link.click() call to this: 

link.dispatchEvent(new Event('click'));

That works in my patched Firefox build

https://treeherder.mozilla.org/#/jobs?repo=try&revision=8c90e8e53643dfbb5711b1c257c7b3682454df90

longsonr:
My fix landed in Firefox. The WPT test change is in https://github.com/w3c/web-platform-tests/pull/10473

svgeesus (approved):
With changes suggested by @longsonr included, this lgtm

PR #10279, opened 2018-04-03
New MIME type parsing tests

annevk:
Tests suggested at https://github.com/w3c/web-platform-tests/issues/1851#issuecomment-378032606.

Co-Authored-By: Matt Menke <mmenke@google.com>

I didn't add these for now:

> text/html;charset=" gbk"

Already tested.

> text/html;charset= ";charset=GBK

I didn't see why this would end up as GBK. We'd have `" \""` as charset parameter value and that would simply not result in an encoding. Why would we consult the second parameter?

> text/html;charset=";charset=GBK";charset=GBK

Semi-colons are allowed, so this would not parse as GBK.

### DIFF
--- a/mimesniff/mime-types/resources/mime-types.json
+++ b/mimesniff/mime-types/resources/mime-types.json
@@ -57,6 +57,12 @@
     "navigable": true,
     "encoding": "GBK"
   },
+  {
+    "input": "text/html;charset= \"gbk\"",
+    "output": "text/html;charset=\" \\\"gbk\\\"\"",
+    "navigable": true,
+    "encoding": null
+  },
   "Single quotes are a token, not a delimiter",
   {
     "input": "text/html;charset='gbk'",
@@ -73,6 +79,12 @@
   {
     "input": "text/html;charset=gbk'",
     "output": "text/html;charset=gbk'",
+    "navigable": true,
+    "encoding": null
+  },
+  {
+    "input": "text/html;charset=';charset=GBK",
+    "output": "text/html;charset=';charset=GBK",
     "navigable": true,
     "encoding": null
   },
@@ -113,6 +125,12 @@
     "navigable": true,
     "encoding": "GBK"
   },
+  {
+    "input": "text/html;charset=();charset=GBK",
+    "output": "text/html;charset=GBK",
+    "navigable": true,
+    "encoding": "GBK"
+  },
   "Double quotes",
   {
     "input": "text/html;charset=\"gbk\"",
@@ -139,6 +157,12 @@
     "encoding": "GBK"
   },
   {
+    "input": "text/html;charset=\"gbk \"",
+    "output": "text/html;charset=\"gbk \"",
+    "navigable": true,
+    "encoding": "GBK"
+  },
+  {
     "input": "text/html;charset=\"\\ gbk\"",
     "output": "text/html;charset=\" gbk\"",
     "navigable": true,
@@ -153,6 +177,12 @@
   {
     "input": "text/html;charset=\"gbk\"x",
     "output": "text/html;charset=gbk",
+    "navigable": true,
+    "encoding": "GBK"
+  },
+  {
+    "input": "text/html;charset=\"\";charset=GBK",
+    "output": "text/html;charset=GBK",
     "navigable": true,
     "encoding": "GBK"
   },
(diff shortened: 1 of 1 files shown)


annevk:
cc @MattMenke2

MattMenke2:
Thanks for this!


For text/html;charset= ";charset=GBK - according to the algorithm, only quotes immediately after the space matters.  So this would be interpreted as:

text/html;charset=" \"";charset=GBK

Quotes aren't allowed in chraset values (See the "parameterName solely contains HTTP token code points" rule, which looks to exclude quotes), so the first charset value isn't set, so we fall back to the second.

Semicolons also don't look to be allowed:  "An HTTP token code point is U+0021 (!), U+0023 (#), U+0024 ($), U+0025 (%), U+0026 (&), U+0027 ('), U+002A (*), U+002B (+), U+002D (-), U+002E (.), U+005E (^), U+005F (_), U+0060 (`), U+007C (|), U+007E (~), or an ASCII alphanumeric."

Am I missing some reason why semicolons and quotes would be allowed?

MattMenke2:
Oops - looks like I was flipping name and value rules.  Sorry about that.

w3c-bots:
# Build PASSED

Started: 2018-04-03 13:02:41
Finished: 2018-04-03 13:12:50





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26766)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/361628500)

MattMenke2:
So instead, I'd suggest:

text/html;charset= ";\xFF;charset=GBK (comes out as GBK, since the \xFF is considered part of the first chrset.)

text/html;charset="\xFF;charset=foo";charset=GBK (Comes out as GBK)

The first is to catch incorrect handling of a quote after a space.  The second is intended to catch incorrect tokenization around semicolons.

MattMenke2:
Sorry, that first one should be:

text/html;charset= "\xFF;charset=GBK

annevk:
Did you mean 0x7F? 0xFF is in the allowed range of HTTP quoted-string token code point so the parameter would not be dropped.

MattMenke2:
Sorry again - yes, I meant \x7F (Or any disallowed character).

annevk:
Cool, added those too.

MattMenke2:
Looks good, thanks!  Looking at the others I suggested, I think if those two pass, all the others will as well (Modulo my confusion over which characters aren't allowed in values).

mimesniff/mime-types/resources/mime-types.json
```diff
@@ -113,6 +125,24 @@
     "navigable": true,
     "encoding": "GBK"
   },
+  {
+    "input": "text/html;charset=();charset=GBK",
+    "output": "text/html;charset=GBK",
```
domenic: I get `text/html;charset="()"` for this. https://mimesniff.spec.whatwg.org/#http-quoted-string-token-code-point includes ( and ) so `()` is a fine charset.

domenic (changes requested):

mimesniff/mime-types/resources/mime-types.json
MattMenke2: Yea, this we my mistake - was confused about the allowed characters.

MattMenke2: *was my mistake

ahmadmiz (approved):

domenic (approved):
LGTM, all tests pass on https://github.com/jsdom/whatwg-mimetype

MattMenke2:
I don't think this one is correct:  https://github.com/w3c/web-platform-tests/pull/10279/files#diff-4d2126605002e90f20a8cb91502ede9cR135

Per the parsing rules, the quoted string should run up to the end of GBK.  The encoding should be null, I believe.

Sorry for the late comment - was just looking over the tests where Chrome fails as I work on fixing most of the remianing difference.  One or two I'll probably leave failing, unless other browsers show the same behavior - I'm a bit paranoid about making the charset= "GBK" ones pass, in particular.

domenic:
That one still works, because of the space before the quote. The spec only interprets a parameter as quoted if it sees the exact sequence `="`.

So it sees two versions of `charset`, one with value ` "\u007F` and one with value `GBK`. But the first one is outside the range of HTTP quoted-string token code points, so it doesn't get set, and the GBK one does.

MattMenke2:
You're right, thanks...and I think it was even exactly as I suggested, the space before quotes rule is just very counter-intuitive.

PR #10197, opened 2018-03-27
Generic executor that uses WebDriver protocol directly

kereliuk:
The motivation behind this is to give the option to not use Selenium with WebDriver. There are multiple reasons one might want to do this:

- Wanting to test the WebDriver protocol directly
- Not wanting to worry about changes in Selenium affecting WPT
- Using functionality in a WebDriver implementation that is not implemented in Selenium

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10197)

PR #10312, opened 2018-04-04
Actions in executorselenium

kereliuk:
This is an implementation of actions in executorselenium. Would like some feedback before I polish it up/make more changes if necessary.

Want to open a discussion with this and show one possibility for the API.

- @jgraham Does marionette Actions use the low level protocol and is it suitable?
- how do we line up this with the selenium API for using actions and the low level protocol for #10197?

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10312)


w3c-bots:
# Build ERRORED

Started: 2018-04-04 20:57:50
Finished: 2018-04-04 21:25:10


<h2>Failing Jobs</h2>
<ul>

<li>lint</li>

<li>firefox:nightly</li>

<li>chrome:dev</li>

</ul>




- [WPT PR Status](http://pulls.web-platform-tests.org/build/26867)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/362273349)

jgraham:
So this seems to pretty directly expose the selenium client API, which isn't defined anywhere and so is hard to accurately reproduce. This is why I want to ensure we work with non-Selenium implementations for this feature. With that in mind, I think we should make the testdriver API a builder with the same primitives as the underlying webdriver API i.e. pointer up down, move; key up and down; pause, and anything else I forgot. This is obviously more implementation work, but I'm pretty sure that putting the higher-level API in the js rather than in the backend is the way to go.

From a marionette point of view it exposes the commands needed for the W3C API, but you have to opt in to avoid the legacy implementation. This will require some changes that I'm happy to make.

NavidZ:
Just wanted to get an update on this. @kereliuk are you working on this? Are you clear with what James refers to?
@jgraham regarding the shape of the API is it possible to also be able to inject multiple input at the same time (in a single frame). What we had in Chrome was something similar to this:
https://navidz.github.io/web-input-automation/#examples
that we were able to add pause and it would inject i'th entry of all those streams at the same time (i.e. in a single frame). For the starter we can definitely remove all these multiple input sources and just support one. Also since these are the raw actions do you think we need to also define some clear exceptions if say they sent a pointerup without any down before that or any other invalid input sequence?

foolip:
@gsnedders, will it make sense to do with now that we have ExecutorWebDriver? Does EdgeDriver or SafariDriver support the actions API?

foolip:
@NavidZ, perhaps this is overtaken by https://github.com/web-platform-tests/wpt/pull/12726? I see that Selenium is touched in that PR, at least.

PR #10315, opened 2018-04-04
Added testdriver automation for html/editing/focus/focus-02.html

kereliuk:
Automate all the things!

### DIFF
--- a/html/editing/focus/focus-02.html
+++ b/html/editing/focus/focus-02.html
@@ -7,12 +7,8 @@
 <meta assert="assert" content="Check if the key events received by document are targeted at the element when no element is focused">
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
-<h2>Steps:</h2>
-<ol>
-  <li>Press any key in [0-9a-zA-Z].</li>
-</ol>
-<h2>Expect results:</h2>
-<p>PASS</p>
+<script src="/resources/testdriver.js"></script>
+<script src="/resources/testdriver-vendor.js"></script>
 <div id="log"></div>
 <script>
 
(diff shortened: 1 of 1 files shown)


w3c-bots:
# Build PENDING

Started: None
Finished: None





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26866)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/362272654)

html/editing/focus/focus-02.html
```diff
@@ -7,12 +7,8 @@
 <meta assert="assert" content="Check if the key events received by document are targeted at the element when no element is focused">
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
-<h2>Steps:</h2>
```
jgraham: Do we have some convention about leaving manual steps in here so that they can be run without testdriver?

foolip: Not really. I think leaving the alone where they exist would be OK, but I wouldn't expect the instructions to stay working over time, or that they'd be added for new tests, so I'd perhaps slightly prefer removing them. WDYT?

foolip (approved):

html/editing/focus/focus-02.html
kereliuk: I think they should be removed.
When new tests are added with testdriver they generally aren't accompanied by manual steps incase they are run without testdriver, so I think we should stick to this convention. I also agree with what @foolip says with maintenance, but @jgraham you bring up a good point that makes me wonder whether we should even allow testdriver tests to be run in any other way than wpt run. I opened #10330 to further address this.

PR #10318, opened 2018-04-04
Match whitespace for custom property serialization

csnardi:
When serializing custom properties, they must be preserved exactly as authored (https://crbug.com/661854). variable-invalidation.html added additional whitespace to the expectation, going against this part of the spec. This change makes the test pass in Firefox, Chrome, and Edge.

### DIFF
--- a/css/css-variables/variable-invalidation.html
+++ b/css/css-variables/variable-invalidation.html
@@ -43,26 +43,26 @@
 
         // This is a boilerplate to build a testcase and then test the expected outcome
         function testCase(cssStyle, testProperty, testElement, testImportant) {
-            var initialCssText = testProperty + (testImportant ? ": red !important;" : ": red;");
+            var initialCssText = testProperty + (testImportant ? ":red !important;" : ":red;");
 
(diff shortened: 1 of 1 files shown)


csnardi:
@upsuper Could you review this? Fairly similar to #10353.

upsuper (changes requested):
I believe that the working group has resolved to something different that, when parsing, whitespaces are trimmed from start and end of token stream, and when serializing, a whitespace is always added after the colon.

The specs may need some update if they haven't been, so do implemenations.

upsuper:
It is possible that I was wrong... Let me find out the resolution...

csnardi:
I think you're correct @upsuper, I believe that's https://github.com/w3c/csswg-drafts/pull/1986 you're describing? If I read that correctly, the way the test currently is would be correct per that change (and so this PR should be closed).

upsuper:
So the resolution was made in w3c/csswg-drafts#774. There was a conflicting resolution in w3c/csswg-drafts#881 which describes what Firefox currently does, which we probably should close.

csnardi:
I believe this PR is then incorrect in at least some form for both the current spec and the proposed changes, so I'm closing this.

PR #10321, opened 2018-04-04
AccName: Fix several expected results

joanmarie:

### DIFF
--- a/accname/name_checkbox-label-multiple-label-alternative-manual.html
+++ b/accname/name_checkbox-label-multiple-label-alternative-manual.html
@@ -22,7 +22,7 @@
                   "property",
                   "name",
                   "is",
-                  "a test"
+                  "a test This is"
                ]
             ],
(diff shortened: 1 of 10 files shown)


joanmarie:
Regarding the results "a test This is", that is the correct result _when the AccName algorithm is applied to the test case_. (Which is what we're testing here.) So, yes, it might look weird as far as prose (or authored content) is concerned, but it is not a mistake. There is also a similar test case where the elements are reversed and the more normal-looking "This is a test" is the result.

w3c-bots:
# Build PASSED

Started: 2018-04-05 00:08:06
Finished: 2018-04-05 00:16:58





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26884)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/362338202)

Ms2ger (approved):

PR #10323, opened 2018-04-04
Handle the case when MANIFEST.json is invalid JSON

fergald:
I don't know how I got into this situation (probably hit ctrl-C at the wrong time) but I did and it was unclear how to get out. It presents as a ValueError with no useful info.

Update the code to handle ValueError to log a warning and cause a full regen of the manifest.

Also reported in https://crbug.com/822041

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10323)

### DIFF
--- a/tools/manifest/manifest.py
+++ b/tools/manifest/manifest.py
@@ -225,6 +225,9 @@ def load(tests_root, manifest):
                 rv = Manifest.from_json(tests_root, json.load(f))
         except IOError:
             return None
+        except ValueError:
+            logger.warning("%r may be corrupted", manifest)
+            return None
         return rv
 
     return Manifest.from_json(tests_root, json.load(manifest))


tools/manifest/manifest.py
```diff
@@ -225,6 +225,9 @@ def load(tests_root, manifest):
                 rv = Manifest.from_json(tests_root, json.load(f))
         except IOError:
             return None
+        except ValueError:
```
gsnedders: IMO, this should probably just capture `Exception`? @jgraham will probably disagree, so don't bother trying to be too prompt with addressing my comment, @fergald. 🙂

fergald:
I created another branch with a slightly different fix

https://github.com/fergald/web-platform-tests/tree/fergald-patch-2

Let me know.

tools/manifest/manifest.py
fergald: I created another branch base on that suggestion.

https://github.com/fergald/web-platform-tests/tree/fergald-patch-2

Let me know what you think.

w3c-bots:
# Build PASSED

Started: 2018-04-05 00:51:04
Finished: 2018-04-05 01:06:54





- [WPT PR Status](http://pulls.web-platform-tests.org/build/26893)
- [TravisCI](https://travis-ci.org/w3c/web-platform-tests/builds/362368575)

jgraham (approved):

Issue #5770, opened 2017-05-03
Add // META: spec=link or remove MISSING-LINK css lint


zcorpan:
See https://github.com/w3c/web-platform-tests/pull/5769

It should be possible to use `.worker.js` or `.any.js` under css/ without having to whitelist for MISSING-LINK lint failure. Either we add a new `// META:` thing to specify a link, or we silently allow such tests without a link.

@gsnedders

gsnedders:
We have the MISSING-LINK lint to make sure the CSS build system picks up everything, and that doesn't support .worker.js or .any.js tests. Um. idk what the right solution is here? Well, apart from replace the CSS build system with something saner.

foolip:
@gsnedders are you planning to work on this? (Came up in triage of roadmap issues older than 60 days.)

lukebjerring:
This is blocking https://github.com/w3c/web-platform-tests/pull/9884

WeizhongX:
https://github.com/web-platform-tests/wpt/pull/44631 hit this again.

PR #10332, opened 2018-04-05
Add MISSING-LINK docs

lukebjerring:
Explains what MISSING-LINK is for, and how to resolve it.

### DIFF
--- a/docs/_writing-tests/lint-tool.md
+++ b/docs/_writing-tests/lint-tool.md
@@ -58,6 +58,9 @@ below to fix all errors reported.
   element whose `content` attribute has a malformed value; **fix**: ensure
   the value of the `content` attribute starts with `?` or `#` or is empty.
 
+* **MISSING-LINK**: CSS test file is missing a link to a spec. **fix**: Ensure that there is a `<link rel="help" src="[url]">` for the spec.
+  * Note: `MISSING-LINK` is designed to ensure that the CSS build tool can find the tests. Note that the CSS build system is primarily used by [test.csswg.org/](http://test.csswg.org/), which doesn't use `wptserve`, so `*.any.js` and similar tests won't work there; stick with the `.html` equivalent.
+
 * **MISSING-TESTHARNESSREPORT**: Test file is missing an instance of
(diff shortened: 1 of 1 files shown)


lukebjerring:
Related to https://github.com/w3c/web-platform-tests/issues/5770, which says

> It should be possible to use .worker.js or .any.js under css/ without having to whitelist for MISSING-LINK lint failure

This statement is not strictly correct, as `.any.js` relies on `wptserve` which is not used by test.csswg.org (and, if we intend to support it, that's a separate issue).

docs/_writing-tests/lint-tool.md
```diff
@@ -58,6 +58,9 @@ below to fix all errors reported.
   element whose `content` attribute has a malformed value; **fix**: ensure
   the value of the `content` attribute starts with `?` or `#` or is empty.
 
+* **MISSING-LINK**: CSS test file is missing a link to a spec. **fix**: Ensure that there is a `<link rel="help" src="[url]">` for the spec.
+  * Note: `MISSING-LINK` is designed to ensure that the CSS build tool can find the tests. Note that the CSS build system is primarily used by [test.csswg.org/](http://test.csswg.org/), which doesn't use `wptserve`, so `*.any.js` and similar tests won't work there; stick with the `.html` equivalent.
```
Ms2ger: I don't think the comment about wptserve is terribly useful; we don't lint for other things that may break when you're not using it.

lukebjerring: Well, actually, what lead me to be frustrated by these missing docs:
- The only way to resolve the lint is with a `<link` tag; (see #5770)
-`.any.js` uses `// META: script`, but doesn't have a `// META: link` or `// META: spec`
- Even _if_ we [added](https://github.com/w3c/web-platform-tests/pull/10289) `// META: spec` for `any.js`, it doesn't work outside `wptserve`, so we'd have to change the build rule
- The lint is for the build rule in the first place

So, without that comment, you dive down a rabbit hole of "how do I fix this lint in my `.any.js` test?" instead of "dumb, but I'll just use the `.html` equivalent test."

foolip (approved):

PR #10336, opened 2018-04-05
Improve idlharness error message for exposure test

lukebjerring:
When `obj` is not created successfully, there's an error like

> Cannot use 'in' operator to search for [member.name] in undefined

NB: I'm not a fan of having each sub-test re-test the same thing, but this is at least _consistent_ with the other code..

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10336)


foolip (approved):
Which test did this come up in?

lukebjerring:
Can't remember - one of the tests I was adding recently.

PR #10344, opened 2018-04-06
Change a const in var in idlharness.js

nox:
See https://github.com/w3c/web-platform-tests/issues/10343#issuecomment-379211826

I know we can't stop progress in WPT just because Servo is stuck in the past,
but let's not make all IDL-related tests fail in it for something that can be
avoided so easily.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10344)

### DIFF
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -684,7 +684,7 @@ IdlArray.prototype.test = function()
     this["includes"] = {};
 
     // Assert B defined for A : B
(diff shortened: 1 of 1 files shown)


Ms2ger (approved):
Somewhat unfortunate, but as long as Servo is actively working on moving into the present, we should take this.

PR #10348, opened 2018-04-06
Allow range headers to pass through a service worker

jakearchibald:
For https://github.com/whatwg/fetch/pull/560.

### DIFF
--- a/service-workers/service-worker/resources/test-helpers.sub.js
+++ b/service-workers/service-worker/resources/test-helpers.sub.js
@@ -46,9 +46,14 @@ function unreached_rejection(test, prefix) {
     });
 }
 
-// Adds an iframe to the document and returns a promise that resolves to the
-// iframe when it finishes loading. The caller is responsible for removing the
-// iframe later if needed.
+/**
+ * Adds an iframe to the document and returns a promise that resolves to the
+ * iframe when it finishes loading. The caller is responsible for removing the
+ * iframe later if needed.
+ *
+ * @param {string} url
+ * @returns {HTMLIFrameElement}
+ */
 function with_iframe(url) {
   return new Promise(function(resolve) {
       var frame = document.createElement('iframe');
--- a/fetch/range/general.any.js
+++ b/fetch/range/general.any.js
@@ -1,0 +1,54 @@
+// Helpers that return headers objects with a particular guard
+function headersGuardNone(fill) {
+  if (fill) return new Headers(fill);
+  return new Headers();
+}
+
+function headersGuardResponse(fill) {
+  const opts = {};
+  if (fill) opts.headers = fill;
+  return new Response('', opts).headers;
+}
+
+function headersGuardRequest(fill) {
+  const opts = {};
+  if (fill) opts.headers = fill;
+  return new Request('./', opts).headers;
+}
+
+function headersGuardRequestNoCors(fill) {
+  const opts = { mode: 'no-cors' };
+  if (fill) opts.headers = fill;
+  return new Request('./', opts).headers;
+}
+
+test(() => {
+  // Setting range should work for these guards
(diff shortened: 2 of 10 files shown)


resources/testharness.js
```diff
@@ -1741,6 +1741,12 @@ policies and contribution forms [3].
             }
         };
 
+        if (self.Promise) {
+            this.done = new Promise(function(resolve) {
+                this_obj.doneResolve = resolve;
+            });
+        }
+
```
jakearchibald: @jgraham @gsnedders I've added this so developers can `await fetch_tests_from_worker(worker)`. This means:

* Promise tests can run in a predictable order.
* Service worker tests can complete before the next test unregisters it.

jakearchibald: @jgraham @gsnedders *nudge* I don't suppose one of you could take a look at this bit?

resources/testharness.js
```diff
@@ -1790,6 +1796,8 @@ policies and contribution forms [3].
         this.running = false;
         this.remote = null;
         this.message_target = null;
+        if (this.doneResolve) this.doneResolve();
```
Ms2ger: Wrap in braces and add line breaks.

resources/testharness.js
```diff
@@ -1741,6 +1741,12 @@ policies and contribution forms [3].
             }
         };
 
+        if (self.Promise) {
+            this.done = new Promise(function(resolve) {
```
Ms2ger: Is there a reason to create this property, or is it just because that's the easiest way to pass the promise to the caller?

resources/testharness.js
Ms2ger: These changes seem okay, but can you extract them into a separate PR and update the docs in `docs/_writing-tests/testharness-api.md`?

jakearchibald: @Ms2ger shall do. Thanks for taking a look.

resources/testharness.js
jakearchibald: Yeah, this is purely so it can be returned in https://github.com/w3c/web-platform-tests/pull/10348/files#diff-d9d64f7c061650556b18794cc9719152R2159.

resources/testharness.js
jakearchibald: Extracted: https://github.com/w3c/web-platform-tests/pull/10978

service-workers/service-worker/resources/test-helpers.sub.js
```diff
@@ -46,9 +46,14 @@ function unreached_rejection(test, prefix) {
...
-// iframe when it finishes loading. The caller is responsible for removing the
-// iframe later if needed.
+/**
+ * Adds an iframe to the document and returns a promise that resolves to the
+ * iframe when it finishes loading. The caller is responsible for removing the
+ * iframe later if needed.
+ *
+ * @param {string} url
+ * @returns {HTMLIFrameElement}
+ */
```
jakearchibald: This change was just to improve autocomplete in VSCode.

annevk (commented):
`range/general.any.js` and `range/partial-script.window.js` seem fine, but the service worker script is rather non-deterministic when I run it. It first finds 8 tests and then on refresh there's 4 (all saying `Error: wait_for_state must be passed a ServiceWorker`).

annevk:
It might be nice for `range/general.any.js` to split up the test a bit since if something fails it'd be rather hard to debug. Though I don't feel too strongly on that since browsers already pass.

jakearchibald:
> It first finds 8 tests and then on refresh there's 4 

Sounds like a bug. Will investigate & fix.

jakearchibald:
@annevk test is now stable in Firefox, and I've split up the general tests.

annevk (approved):

PR #10350, opened 2018-04-06
WIP tests for text/css link quirk

domenic:
These are meant to help with https://github.com/whatwg/html/pull/3544; do not merge until we have some clarity.


annevk:
I don't think these are needed because of #13144.

domenic:
It looks to me that this has some tests for moving between documents that that one doesn't, but good catch in general.

PR #10353, opened 2018-04-07
Update expected cssText for custom property serialization

csnardi:
See https://github.com/w3c/csswg-drafts/issues/2509#issuecomment-379152590 for the change to `target9`. Also remove the extra whitespace in `target8` and `target1` which causes these tests to fail in Chrome and Firefox.

Also remove `testcase.propertyName` from each test's name as this doesn't exist and just outputs undefined.

### DIFF
--- a/css/css-variables/variable-cssText.html
+++ b/css/css-variables/variable-cssText.html
@@ -28,15 +28,15 @@
     "use strict";
 
     var testcases = [
-        { element: "target1",   expectedCssText: "--var:  var1;" },
+        { element: "target1",   expectedCssText: "--var: var1;" },
         { element: "target2",   expectedCssText: "margin: var(--prop);" },
         { element: "target3",   expectedCssText: "background: var(--prop);" },
         { element: "target4",   expectedCssText: "margin: var(--prop)  !important;" },
         { element: "target5",   expectedCssText: "background: var(--prop)  !important;" },
         { element: "target6",   expectedCssText: "background: green;" },
         { element: "target7",   expectedCssText: "background: var(--prop);" },
-        { element: "target8",   expectedCssText: "color:  var(--prop);" },
-        { element: "target9",   expectedCssText: "margin-top: 10px; margin-right: ; margin-bottom: ; margin-left: ;" },
+        { element: "target8",   expectedCssText: "color: var(--prop);" },
+        { element: "target9",   expectedCssText: "margin-right: ; margin-bottom: ; margin-left: ; margin-top: 10px;" },
         { element: "target10",  expectedCssText: "" }
     ];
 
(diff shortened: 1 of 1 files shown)


csnardi:
@upsuper Could you review? I don't have access to request reviewers, unfortunately.

css/css-variables/variable-cssText.html
```diff
@@ -28,15 +28,15 @@
...
         { element: "target2",   expectedCssText: "margin: var(--prop);" },
         { element: "target3",   expectedCssText: "background: var(--prop);" },
         { element: "target4",   expectedCssText: "margin: var(--prop)  !important;" },
         { element: "target5",   expectedCssText: "background: var(--prop)  !important;" },
         { element: "target6",   expectedCssText: "background: green;" },
         { element: "target7",   expectedCssText: "background: var(--prop);" },
-        { element: "target8",   expectedCssText: "color:  var(--prop);" },
-        { element: "target9",   expectedCssText: "margin-top: 10px; margin-right: ; margin-bottom: ; margin-left: ;" },
+        { element: "target8",   expectedCssText: "color: var(--prop);" },
+        { element: "target9",   expectedCssText: "margin-right: ; margin-bottom: ; margin-left: ; margin-top: 10px;" },
```
csnardi: In Firefox this fails due to no whitespace between the property and the semicolon. I think this is a bug in Firefox per step 3 of https://drafts.csswg.org/cssom/#serialize-a-css-declaration, but just wanted to make sure I wasn't missing something.

upsuper:
To be clear, IIRC, for serialization of token stream (which is what the value is when a declaration references variable), we've agreed at some point that we should strip whitespaces when parsing, and add whitespace between colon and the value when serializing the declaration.

So in this test, `--var:␣var1;`should indeed be serialized to `--var:␣var1;`, while `--var:var1;` and `--var:␣␣var1;` should also be serialized to that. I believe we haven't updated our implementation to match that in Firefox, and maybe Chrome is more correct in those cases. And the specs may not have been updated properly either...

css/css-variables/variable-cssText.html
upsuper: I don't think this serialization is any useful. We may want to update the spec to something so that when a longhand is expanded from a shorthand which references variables, the shorthand is always serialized, and any longhand doesn't share the same shorthand expression are not considered serialized by this shorthand.

upsuper: But that's a separate issue, and doesn't affect this change, I think.

upsuper (approved):
I think this change is correct (at least for now). I don't think I'm able to merge it, though.

upsuper:
Oh it seems I can merge it. Then, sure :)

csnardi:
Thanks. :) Agreed on the serialization point above, a spec bug should probably be opened for that (if there isn't already one).

upsuper:
I don't think there is any. I'll file one later today to include my thought about that.

upsuper:
Thanks for raising that issue and updating the test!

PR #10354, opened 2018-04-07
Add test for checking property order after setting property on CSSStyleDeclaration

upsuper:
This is the test for w3c/csswg-drafts#2516.

### DIFF
--- a/css/cssom/cssstyledeclaration-setter-order.html
+++ b/css/cssom/cssstyledeclaration-setter-order.html
@@ -1,0 +1,108 @@
+<!DOCTYPE html>
+<title>CSSOM test: order of declarations after setting via CSSOM</title>
+<link rel="help" href="https://drafts.csswg.org/cssom/#set-a-css-declaration-value">
+<script src="/resources/testharness.js"></script>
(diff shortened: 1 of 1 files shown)


upsuper:
(FWIW, it fails with Firefox, Safari, and Chrome with exactly the same assertion, and it passes with a patched Firefox I'm working on.)

css/cssom/cssstyledeclaration-setter-order.html
```diff
@@ -0,0 +1,111 @@
...
+<script src="/resources/testharnessreport.js"></script>
+<div id="log"></div>
+<script>
+  function generateCSSDeclBlock(props) {
+    let elem = document.createElement("div");
+    let cssText = props.map(([prop, value]) => `${prop}: ${value};`).join(" ");
+    elem.setAttribute("style", cssText);
+    return elem.style;
+  }
+  function checkOrder(block, props, msg) {
```
emilio: I wonder if this can just do `assert_array_equals`.

emilio (approved):

PR #10357, opened 2018-04-07
Avoid async/await in /webstorage/idlharness.html

nox:
Am sorry, Servo is dumb, we are actively trying to fix that on our side.

### DIFF
--- a/webstorage/idlharness.html
+++ b/webstorage/idlharness.html
@@ -58,11 +58,7 @@ dictionary StorageEventInit : EventInit {
 </pre>
 
 <script>
-
-promise_test(async function() {
-  const dom = await fetch('/interfaces/dom.idl').then(r => r.text());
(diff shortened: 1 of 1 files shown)


webstorage/idlharness.html
```diff
@@ -73,8 +69,11 @@ <h1>Web Storage IDL tests</h1>
   idl_array.add_objects({Storage: ["window.localStorage"]});
 
   idl_array.test();
-}, 'webstorage interfaces');
+}
 
+promise_test(function () {
+  return Promise.all(['/interfaces/html.idl', '/interfaces/dom.idl'].map(fetchText)).then(doTest);
```
emilio: How does this work? Did you mean `do_test` instead of `doTest`?

nox: You saw nothing.

annevk (approved):
I assume you tried running this.

This looks like an improvement as fetching will happen concurrently with this setup.

PR #10381, opened 2018-04-09
Support adding Idlharness dependencies

lukebjerring:
This is a request for feedback on a potential solution to https://github.com/w3c/web-platform-tests/issues/10347

The PR adds a new `idlharness.js` api: `add_dependency_idls(raw_idls, options)`

EDIT: Explanation:
Currently, to get idl_array to stop complaining about missing interface definitions, the test author has a couple options: 
- Add all the idl interfaces that get referenced, one by one, using the `{only: 'Foo'}` helper
- Add all the idl interfaces and some that aren't referenced, whole file at a time (e.g. `html` / `dom`)
- "mock" the idl interfaces with inline, empty definitions, e.g. `interface Foo {};`

This PR aims to add a better, 4th option: Add all the idl that get referenced (and nothing more!), whole file at a time. It's better because real interfaces are used, without bloating the test.

Usage examples are in the couple of modified tests.

Note that the `options` object was needed for adding deps in some cases, because some interfaces are re-declared in both `html` and `dedicated-workers.idl`, e.g. `WorkerNavigator`.
See https://github.com/w3c/web-platform-tests/blob/master/budget-api/interfaces.any.js for example.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10381)


birtles (approved):
Changes to the web-animations tests look good to me.

resources/idlharness.js
```diff
@@ -218,6 +226,81 @@ IdlArray.prototype.add_untested_idls = function(raw_idls, options)
...
+{
+    const parsed_idls = WebIDL2.parse(raw_idls);
+    const new_options = { only: [] }
+    
+    const all_deps = new Set();
+    Object.values(this["inheritance"]).forEach(v => all_deps.add(v));
+    Object.values(this["implements"]).forEach(v => all_deps.add(v));
+    this.partials.map(p => p.name).forEach(v => all_deps.add(v));
+    
+    console.log('all deps', all_deps);
```
foolip: Remember to remove logging before merging :)

resources/idlharness.js
```diff
@@ -218,6 +226,81 @@ IdlArray.prototype.add_untested_idls = function(raw_idls, options)
...
+//@}
+IdlArray.prototype.add_dependency_idls = function(raw_idls, options)
+//@{
+{
+    const parsed_idls = WebIDL2.parse(raw_idls);
+    const new_options = { only: [] }
+    
+    const all_deps = new Set();
+    Object.values(this["inheritance"]).forEach(v => all_deps.add(v));
+    Object.values(this["implements"]).forEach(v => all_deps.add(v));
```
foolip: Should "includes" also be processed here? That's for mixins, added in https://github.com/w3c/web-platform-tests/pull/8727

resources/idlharness.js
```diff
@@ -218,6 +226,81 @@ IdlArray.prototype.add_untested_idls = function(raw_idls, options)
...
+    this.partials.map(p => p.name).forEach(v => all_deps.add(v));
```
foolip: Use `this["partials"]` here, or `this.inheritance` + `this.implements` above?

resources/idlharness.js
```diff
@@ -218,6 +226,81 @@ IdlArray.prototype.add_untested_idls = function(raw_idls, options)
...
+    
+    console.log('all deps', all_deps);
+
+    if (options && options.except && options.only) {
+        throw new IdlHarnessError("The only and except options can't be used together.");
+    }
+    const skip = name => {
```
foolip: This is very similar to the `function should_skip(name)` in `internal_add_idls`. Can you copy that so that any differences stand out, or, if it seems worthwhile, use a shared definition?

resources/idlharness.js
```diff
@@ -218,6 +226,81 @@ IdlArray.prototype.add_untested_idls = function(raw_idls, options)
...
+            skipped.delete(follow_up);
+            process(next);
+        }
+    }
+    for (let parsed of parsed_idls) {
+        process(parsed);
+    }
+
+    // Mark as untested.
+    for (var i = 0; i < parsed_idls.length; i++) {
```
foolip: Share with `add_untested_idls`?

resources/idlharness.js
```diff
@@ -218,6 +226,81 @@ IdlArray.prototype.add_untested_idls = function(raw_idls, options)
...
+            if (parsed[dep_type]) {
+                const dep = parsed[dep_type];
+                new_options.only.push(dep);
+                all_deps.add(dep);
+                follow_up.push(dep);
+                console.log(`Added ${dep_type} dep ${dep}`);
+            }
+        }
+
+        if (follow_up && skipped.has(follow_up)) {
```
foolip: This seems broken. `follow_up` is an array, and even an empty array is truthy in JS (unlike in Python). Should have been a loop?

resources/idlharness.js
```diff
@@ -218,6 +226,81 @@ IdlArray.prototype.add_untested_idls = function(raw_idls, options)
...
+            || parsed.type == "includes" && parsed.target;
+        if (skip(name) || !all_deps.has(name)) {
+            name && skipped.set(name, parsed);
+            return;
+        }
+
+        console.log(parsed);
+        
+        new_options.only.push(name);
+        const follow_up = [];
```
foolip: Can you add a comment about what this bookkeeping is for? I think it's in case we have a dependency on B, which in turn depends on C, but we've already processed C. However, since `add_dependency_idls` can be called multiple times, once for each IDL file, it seems like this will still be broken for cross-spec dependencies. Most likely, some spec depends on something in HTML, that depends on something in DOM, but we process DOM before HTML. Does this still work then?

resources/idlharness.js
```diff
@@ -304,6 +387,17 @@ IdlArray.prototype.internal_add_idls = function(parsed_idls, options)
         {
             throw new IdlHarnessError("Duplicate identifier " + parsed_idl.name);
         }
+
+        if (parsed_idl["inheritance"]) {
+            console.log(parsed_idl);
+            if (parsed_idl.name in this["inheritance"]
+                && parsed_idl["inheritance"] != this["inheritance"][parsed_idl.name]) {
+                throw new IdlHarnessError(
+                    `Inheritance for ${parsed_idl.name} was already defined`);
```
foolip: A new idlharness.js test for this error?

foolip (commented):
Cool! Last time I tried to review this I was confused about what it was trying to do, reading more carefully I conclude that it's to pull in untested dependencies from unparsed IDL but filtering out that which nothing depends on, in order to get rid of incidental dependency-dependencies that one currently needs to spend time sorting out. Makes sense! A bunch of questions...

resources/idlharness.js
lukebjerring: I treat it as a dep type below, so yeah, I think it should.

resources/idlharness.js

lukebjerring (commented):
PTAL!

resources/idlharness.js
foolip: I agree, something more global would have its own set of problems. ("Why does removing the *following* line cause a problem with the IDL on *this* line?")

The "Record of skipped items in case we later determine they are a dependency" comment is clear to me now. Could possibly say that it's *not* global, but that's a pretty far-fetched misunderstanding, so nah.

resources/idlharness.js
```diff
@@ -212,44 +212,56 @@ IdlArray.prototype.add_untested_idls = function(raw_idls, options)
 {
     /** Entry point.  See documentation at beginning of file. */
     var parsed_idls = WebIDL2.parse(raw_idls);
-    for (var i = 0; i < parsed_idls.length; i++)
-    {
+    this.mark_as_untested(parsed_idls);
```
foolip: This is short and sweet now, nice!

resources/idlharness.js
```diff
@@ -212,44 +212,56 @@ IdlArray.prototype.add_untested_idls = function(raw_idls, options)
...
+    Object.values(this.includes).forEach(v => all_deps.add(v));
     this.partials.map(p => p.name).forEach(v => all_deps.add(v));
-    
-    console.log('all deps', all_deps);
 
     if (options && options.except && options.only) {
         throw new IdlHarnessError("The only and except options can't be used together.");
     }
+
     const skip = name => {
```
foolip: Can you name this `should_skip` like the other one?

resources/idlharness.js
```diff
@@ -212,44 +212,56 @@ IdlArray.prototype.add_untested_idls = function(raw_idls, options)
...
     }
-    this.internal_add_idls(parsed_idls, options);
 };
+//@}
+
+//@}
+IdlArray.prototype.is_excluded_by_options = function (name, options)
+//@{
+{
+    return !name ||
```
foolip: Can you hoist the `!name` part of this to the one (right?) call site where it's possible for it to be falsy? I actually don't really get why it can be falsy, but it does seem like when there's nothing to skip the return value from this method can't matter, and the purpose of the check will be more clear closer to the origin of the value.

foolip (approved):
LGTM % small nits. Although most changes are contained to `add_dependency_idls` it's still not far fetched that there might be regressions. Before merging, can you run the updated tests and a handful of tests that haven't been changed to see if it's all good? (Or put it through Chromium's CQ.)

resources/idlharness.js

resources/idlharness.js
```diff
@@ -268,7 +267,7 @@ IdlArray.prototype.add_dependency_idls = function(raw_idls, options)
         let name = parsed.name
             || parsed.type == "implements" && parsed.target
             || parsed.type == "includes" && parsed.target;
-        if (skip(name) || !all_deps.has(name)) {
+        if (!name || should_skip(name) || !all_deps.has(name)) {
```
foolip: Still can't quite tell why it's possible for `name` to be null. Is it `parsed.target` which can be null?

lukebjerring: If the chunked is unnamed and doesn't implement/include. Can't recall the case that isn't any of these, but I definitely hit it in my testing.

lukebjerring:
@foolip - to fix the breakage of Animation idl test, there are 2 major changes:
 - Attributes of tested IDLs are included as dependencies
 - Encountering a dep which is already a member of the idlarray is ignored

The second change occurs when `A : B` and `B : C` and you want to test A and B, but not C. Adding deps for the same file means that B is re-encountered (and would clash / already exist). See the Animation test file for more context.

Note that the reason to add_dependency_idls from the same file is that only part of the web-animation IDL is (currently) tested, so to get a no-op result, needed to stay that way.

resources/idlharness.js
```diff
@@ -204,19 +212,109 @@ IdlArray.prototype.add_untested_idls = function(raw_idls, options)
...
+            all_deps.add(m.idlType.idlType);
+        });
+    });
+
+    if (options && options.except && options.only) {
+        throw new IdlHarnessError("The only and except options can't be used together.");
+    }
+
+    const should_skip = name => {
+      return this.is_excluded_by_options(name, options) || name in this.members;
```
foolip: Should this same condition also apply to the other `should_skip`? Just a comment about what the `name in this.members` bit is for would calm the nervous reader.

foolip (approved):
LGTM % possible comment.

Maybe it's just because I've read it over and over, but I think this is all nice and understandable now, which is rarely true in this file. Thanks!

resources/idlharness.js
lukebjerring: No; we don't want to be lenient with tested IDL. Large comment added.

resources/idlharness.js
```diff
@@ -269,7 +269,13 @@ IdlArray.prototype.add_dependency_idls = function(raw_idls, options)
     }
 
     const should_skip = name => {
-      return this.is_excluded_by_options(name, options) || name in this.members;
+        // NOTE: Deps are untested, so we're lenient, and skip re-encountered definitions.
+        // e.g. for 'idl' containing A:B, B:C, C:D
+        //      array.add_idls(idl, {only: ['A','B']}).
+        //      array.add_dependency_idls(idl);
+        // B would be encountered as tested, and encountered as a dep, so we ignore.
```
foolip: Ah, great comment, thank you!

PR #10329, opened 2018-04-05
Extra window.event tests

annevk:
To compliment #4790.

PR #10382, opened 2018-04-09
Test self.event in workers

annevk:
To compliment #10329. Needed for https://github.com/whatwg/dom/pull/407.

### DIFF
--- a/dom/events/event-global.worker.js
+++ b/dom/events/event-global.worker.js
@@ -1,0 +1,14 @@
+importScripts("/resources/testharness.js");
+test(t => {
+  let seen = false;
+  const event = new Event("hi");
+  assert_equals(self.event, undefined);
+  self.addEventListener("hi", t.step_func(e => {
+    seen = true;
+    assert_equals(self.event, undefined);
+    assert_equals(e, event);
+  }));
+  self.dispatchEvent(event);
+  assert_true(seen);
+}, "There's no self.event (that's why we call it window.event) in workers");
+done();


dom/events/event-global.worker.js
```diff
@@ -0,0 +1,14 @@
+importScripts("/resources/testharness.js");
+test(t => {
+  let seen = false;
+  const event = new Event("hi");
+  assert_equals(self.event, undefined);
+  self.addEventListener("hi", t.step_func(e => {
```
foolip: I think it's harmless, but the `t.step_func` isn't needed for synchronous tests like this one.

dom/events/event-global.worker.js
```diff
@@ -0,0 +1,14 @@
...
+    seen = true;
+    assert_equals(self.event, undefined);
+    assert_equals(e, event);
+  }));
+  self.dispatchEvent(event);
+  assert_true(seen);
+}, "There's no self.event (that's why we call it window.event) in workers");
```
foolip: I like this test title :)

foolip (approved):
LGTM % optional simplification

dom/events/event-global.worker.js
annevk: Yes it is, what if the callback throws?

foolip: The callback is invoked synchronously by `dispatchEvent` so the exception will reach the right place. Unless dispatchEvent suppresses exceptions and moves along to the next callback?

annevk: That's not how event listeners work. They use "report the exception" which crashes the harness.

foolip: Well there you go, I am just wrong :)

PR #10393, opened 2018-04-10
Fix #10388: Make resources/test use the upstream config

### LINKED ISSUE #10388: ./resources/test/tox fails after 10231
Issue started after https://github.com/w3c/web-platform-tests/pull/10231

Message is 

    Could not start wptserve

Running `wptserve` works fine - but not with the config used in the test suite, which produces

    KeyError: u"unknown config override 'external_host'"

gsnedders:
Fixes #10388. Essentially, the config.test.json is basically the old config.default.json but with non-standard ports (and we seem to not have backwards compatibility there). I don't see any reason for this, so let's just use the upstream config.default.json.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10393)


gsnedders:
This seems to work. Still a bit ugly given we're manually reading the JSON here instead of using the class, but let's just do this because it fixes it fast.

foolip:
Having one config file less is an improvement at least.

foolip (approved):

PR #10396, opened 2018-04-10
Automate more focus tests with testdriver

Honry:
Thanks to `testdriver.js`, we can automate more tests.

@kereliuk, could you PTAL?


html/editing/focus/sequential-focus-navigation-and-the-tabindex-attribute/focus-tabindex-order.html
```diff
@@ -2,17 +2,13 @@
 <meta charset="utf-8">
```
Honry: This test fails due to https://github.com/w3c/web-platform-tests/issues/10398

kereliuk (approved):
Thanks for automating these (and finding the bug).
Also your comments about TAB being '\ue004' has made me realize I should document where to find the escape codes in the testdriver docs. I will do that!

Honry:
@kereliuk, thank you for your review. Merging it...

PR #10397, opened 2018-04-10
Add python handler directory to Python path in wptserve

### LINKED ISSUE #10378: Local imports fail in python file handlers
```py
from utils import create_wav_header
```

The above fails with `ImportError: No module named utils`, despite `utils.py` existing in the same directory.

jgraham:
Currently making import of files relative to the handler is tricky
because, unlike in normal Python, the script directory isn't on the
path.

This patch adds it to the path when executing the handler. It also
ensures that any modifications made to the path or to sys.modules are
rolled back after the handler has run.

Fixes #10378

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10397)

### DIFF
--- a/tools/wptserve/tests/functional/docroot/subdir/example_module.py
+++ b/tools/wptserve/tests/functional/docroot/subdir/example_module.py
@@ -1,0 +1,2 @@
+def module_function():
+    return [("Content-Type", "text/plain")], "PASS"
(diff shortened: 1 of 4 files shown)


gsnedders (approved):

jakearchibald:
Ohh this is great. Thank you!

PR #10401, opened 2018-04-10
 Updating WebDriver execute[Async]Script tests for unhandled user prompts 

jimevans:
When the WebDriver specification is modified with the contents of PR 1248
(https://github.com/w3c/webdriver/pull/1248), the behavior of
implementations changes to handle user prompts on the next command after
an executeScript or executeAsyncScript command. This PR modifies the
tests for user prompts for those commands to match the spec behavior.


andreastt:
Hm, I think there are a lot more tests that also need this update?

jimevans:
Shouldn't be. The spec change should **only** affect when an unhandled user prompt is displayed by a user-executed piece of JavaScript. All of the other unhandled user prompt tests already wait for the next command for the prompt to appear.

andreastt:
Right, sorry. You’re correct.

andreastt (approved):

PR #10402, opened 2018-04-10
Added note about using special keys in send_keys to testdriver docs.

kereliuk:
Linked to the WebDriver spec that defines the codepoints.


kereliuk:
Ah my history is messed up and I didn't realize it... please ignore this one I will send another.

PR #10419, opened 2018-04-11
XMLHttpRequest: align username/password test with the standard

annevk:
Fixes https://github.com/whatwg/xhr/issues/191.

### DIFF
--- a/xhr/send-authentication-competing-names-passwords.htm
+++ b/xhr/send-authentication-competing-names-passwords.htm
@@ -15,9 +15,12 @@
                 userwin = user2 || user1,
                 passwin = pass2 || pass1;
           let urlstart = "";
-          if(user1) { // should add userinfo to URL (there is no way to create userinfo part of URL with only password in)
-            urlstart = "http://" + user1;
-            if(pass1) {
+          if (user1 || pass1) {
+            urlstart = "http://";
+            if (user1) {
+              urlstart += user1;
+            }
+            if (pass1) {
               urlstart += ":" + pass1;
             }
             urlstart += "@" + location.host + location.pathname.replace(/\/[^\/]*$/, '/');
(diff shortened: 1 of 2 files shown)


annevk:
@wisniewskit I fixed this per your suggestion. Would you be willing to review it?

annevk:
Once this lands, a bug should be filed against Firefox and Safari for aligning with the standard. (Edge also fails, but they intentionally throw for username/password in the URL.)

xhr/send-authentication-competing-names-passwords.htm
```diff
@@ -5,50 +5,41 @@
...
-            urlstart = "http://" + user1
-            if(pass1)
-              urlstart += ":" + pass1
-            urlstart += "@" + location.host + location.pathname.replace(/\/[^\/]*$/, '/')
+            urlstart = "http://" + user1;
+            if(pass1) {
+              urlstart += ":" + pass1;
+            }
+            urlstart += "@" + location.host + location.pathname.replace(/\/[^\/]*$/, '/');
           }
```
yutakahirano: else {
  assert_equals(null, pass2, 'when user1 is null then pass1 must be null');
}
?

annevk: I don't quite understand this suggestion. This is just about setting up the URL.

yutakahirano: Sorry, that should be assert_equals(null, pass1, 'when user1 is null then pass1 must be null');

The test will be broken when user1 is null (or empty) and pass1 is not, as pass1 is set to the url only when user1 is not null (or empty).

Perhaps we could construct a URL in a less hacky way, such as 

const url = new URL('resources/authentication.py', location.href);
if (user1) {
  url.username = user1;
}
if (pass1) {
  url.password = pass1;
}

yutakahirano: Then, how about this?

```
if (user1 || pass1) {
  urlstart = "http://";
  if (user1)
    urlstart += user1;
  if (pass1)
    urlstart += ":" + pass1;
  urlstart += "@" + location.host + location.pathname.replace(/\/[^\/]*$/, '/');
}
```

annevk: Yeah, I'll make that change. It does seem incorrect that it states you need user if you've been given pass.

annevk: I should probably add an additional test for that...

xhr/send-authentication-competing-names-passwords.htm
```diff
@@ -5,50 +5,46 @@
...
+      request(null, null, token(), null, "user in open()");
+      request(null, null, token(), token(), "user/pass in open()");
+      request(null, null, token(), token(), "another user/pass in open(); must override cached credentials from previous test");
+      request(null, token(), token(), null, "pass in URL, user in open()");
+      request(null, token(), token(), token(), "pass in URL, user/pass in open()");
+      request(token(), null, null, null, "user in URL");
+      request(token(), null, null, token(), "user in URL, pass in open()");
+      request(token(), token(), null, null, "user/pass in URL");
+      request(token(), null, token(), null, "user in URL and open()");
+      request(token(), null, token(), token(), "user in URL; user/pass in open()")
```
yutakahirano: ;

annevk:
Bugs:
* https://bugs.webkit.org/show_bug.cgi?id=184910
* https://bugzilla.mozilla.org/show_bug.cgi?id=1456421

xhr/send-authentication-competing-names-passwords.htm
```diff
@@ -5,50 +5,46 @@
...
-          // We want to send multiple requests to the same realm here, so we try to make the UA forget its (cached) credentials between each test..
-          // forcing a 401 response to (hopefully) "log out"
-          // NOTE: This is commented out because it causes authentication prompts while running the test
-          //client.open('GET', "resources/authentication.py?logout=1", false)
-          //client.send()
-        }, document.title+' '+name)
+          client.open("GET", urlstart + "resources/authentication.py", false, user2, pass2);
+          client.setRequestHeader("x-user", userwin);
+          client.send(null);
+          assert_true(client.responseText == ((userwin||'') + "\n" + (passwin||'')), 'responseText should contain the right user and password');
```
yutakahirano: === rather than ==

yutakahirano (approved):

PR #10420, opened 2018-04-11
[css-text-3] New test with line-break / word-break / lang combinations

faceless2:

### DIFF
--- a/css/css-text/word-line-break-001.xht
+++ b/css/css-text/word-line-break-001.xht
@@ -1,0 +1,326 @@
+<html xmlns="http://www.w3.org/1999/xhtml">
+ <head>
+  <meta charset="utf-8"/>
+  <link rel="author" title="Mike Bremford" href="http://bfo.com" />
+  <link rel="help" href="https://www.w3.org/TR/css-text-3/#line-breaking" />
+  <meta name="flags" content="" />
+  <meta name="assert" content="Testing various combinations of line-break, word-break and locale "/>
+  <style>
+   section {
+       margin-bottom: 1em;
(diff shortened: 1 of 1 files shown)


foolip:
@frivoal were you going to review the this?

frivoal:
Finally made it to the top of the todo list :) sorry for the delay.

This is an interesting test, but it needs a few tweaks before we can take it in:
* please move it to one of the subdirectories of `wpt/css/css-text`. `wpt/css/css-text/line-breaking` seems appropriate
* Please add a descriptive sentence at the top, so that someone looking at it knows what they should expect. Something like
    ```html
    <p>Test passes if the black and blue text in each pair of boxes is identical.
    ```
* This should be turned into a [ref test](https://web-platform-tests.org/writing-tests/reftest-tutorial.html)
* The automated screenshot tool cannot (reliably) work with tests bigger than 800x600, but this is larger. Suggestion:
  -  split into several files
  - turn the descriptive line at the beginning of each section into a html comment, to preserve it but hide if from the output
* Much of UAX14 is described as "tailorable", which is more RFC2119 "should" territory than "must". Tests for that sort of statements should have `<meta name="flags" content="should">`. This may be a good way to decide how to split the test into smaller files: put all the UAX14 non-tailorable requirements (such as the GL class) as well as all things directly required by css-text in some files, and all the rest (tailorable things) in other files, with the "should" meta tag.

frivoal (changes requested):
see https://github.com/web-platform-tests/wpt/pull/10420#issuecomment-561408221

PR #10424, opened 2018-04-11
XMLHttpRequest: response header value containing 0x00

annevk:
As discussed in https://github.com/whatwg/xhr/issues/165 these should turn the response into a network error.

### DIFF
--- a/xhr/headers-normalize-response.htm
+++ b/xhr/headers-normalize-response.htm
@@ -3,13 +3,15 @@
 <title>Whitespace and null in header values</title>
 <script src=/resources/testharness.js></script>
 <script src=/resources/testharnessreport.js></script>
-<script src=support.js?pipe=sub></script>
-
-<h1>Whitespace and null in response header values</h1>
-
 <div id=log></div>
-
 <script>
+function error(val) {
+  test(() => {
+    const client = new XMLHttpRequest();
+    client.open("GET", "resources/parse-headers.py?my-custom-header="+encodeURIComponent(val), false);
+    assert_throws("NetworkError", () => client.send());
+  }, "Header value: " + val.replace("\0", "\\0"));
+}
 
 function matchHeaderValue(val) {
     test(function () {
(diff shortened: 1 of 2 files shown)


annevk:
(Note that various cors/ tests use 0x00 in header values too, but those already expect a network error due to the nature of CORS.)

xhr/headers-normalize-response.htm
```diff
@@ -34,9 +36,12 @@ <h1>Whitespace and null in response header values</h1>
...
 matchHeaderValue("hello\tworld");
-matchHeaderValue("\0");
+error("\0");
 matchHeaderValue("   ");
 matchHeaderValue("\t");
 matchHeaderValue("");
 
+promise_test(t => {
+  return promise_rejects(t, new TypeError(), fetch("resources/parse-headers.py?my-custom-header="+encodeURIComponent("x\0x")));
+}, "Ensure fetch() rejects too")
```
domenic: This seems pretty out of place here; another separate file would be much better, in case people are testing fetch() and XHR in isolation.

domenic (changes requested):
Test code LGTM but fetch() should not be tested in the XHR test folder.

xhr/headers-normalize-response.htm
annevk: Why would they do that?

domenic: Because they're working on their fetch() or XHR implementation?

annevk: You need to support both and they both need to fail in the same way due to both operating on the same low-level primitive.

annevk: (And it's not like both are tested in the same test so it'd be easy to ignore some results.)

annevk: (I actually think we might want to merge these directories at some point due to the worry of missing test coverage in either API.)

domenic: I'm still not comfortable testing fetch in the XHR directory, sorry. It makes maintainers lives harder when they don't know where a given API's tests are.

annevk: That's already the case though for many cross-purpose tests. I guess we could move this into the `http/` directory...

annevk:
I moved it to a separate file for now. I think we already have this cross-dependency allowed elsewhere, but I want to get this landed.

domenic (approved):

PR #10428, opened 2018-04-11
accname: Updated tests

joanmarie:
* Add missing "hidden" style from several tests
* Fix expectations for several tests with presentation role used on img
* Fix bogus expectations on name calculation with combobox
* Add test for which coverage was missing


joanmarie:
Closing because I found more changes I need to make. Will do new pull request.

PR #10429, opened 2018-04-11
[cssom] Add a test for mutation records when CSSStyleDeclaration.setPropertyValue is invoked.

emilio:
Even if it doesn't mutate the block.

https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-setpropertyvalue always
ends up in https://drafts.csswg.org/cssom/#update-style-attribute-for (step 8),
which calls https://dom.spec.whatwg.org/#concept-element-attributes-set-value,
which does queue a mutation record even if the value doesn't change.

This makes this consistent with `setAttribute("style", getAttribute("style"))`,
for example.

### DIFF
--- a/css/cssom/cssstyledeclaration-mutationrecord-001.html
+++ b/css/cssom/cssstyledeclaration-mutationrecord-001.html
@@ -1,0 +1,20 @@
+<!doctype html>
+<meta charset="utf-8">
+<title>CSSOM: CSSStyleDeclaration.setPropertyValue queues a mutation record when not actually mutated</title>
+<link rel="help" href="https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-setpropertyvalue">
+<link rel="help" href="https://drafts.csswg.org/cssom/#update-style-attribute-for">
+<link rel="author" title="Emilio Cobos Álvarez" href="mailto:emilio@crisal.io">
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<script>
+  document.documentElement.style.top = "0px";
+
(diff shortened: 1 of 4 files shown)


css/cssom/cssstyledeclaration-mutationrecord-001.html
```diff
@@ -0,0 +1,20 @@
...
+  document.documentElement.style.top = "0px";
+
+  let test = async_test("CSSStyleDeclaration.setPropertyValue queues a mutation record, even if not mutated");
+  let m = new MutationObserver(function(r) {
+    assert_equals(r.length, 1);
+    test.done();
+  });
+
+  m.observe(document.documentElement,  { attributes: true });
+  document.documentElement.style.top = "0px";
```
upsuper: I'm less sure about whether we should optimize this case, since it's not really changing the value. But detecting equivalence of value is probably not that trivial spec-wise... so given the spec currently, I guess this is fine.

upsuper (commented):
Could you add another test for removing non-existing property (ideally from a declaration with some declarations, because Gecko optimizes the case where the declaration doesn't exist at all).

emilio:
Ok, added tests for all those.

upsuper (approved):
All the tests can really be done in a single file with multiple subtests... but it's not a big problem anyway.

emilio:
Yeah, the issue is that it complicates the tests, and if one times out or something then the rest of the tests wouldn't run.

emilio:
Yeah, the issue is that it complicates the tests, and if one times out or something then the rest of the tests wouldn't run.

PR #10430, opened 2018-04-11
Don't store logger on the Config object

jgraham:
Config objects are shared between processes and so must be
pickleable. Loggers contain locks and so are not. Instead of storing
the logger on the config directly, store the logger name and level and
get the logger from the logging module each time.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10430)


Ms2ger (approved):
Feel free to wait for @gsnedders, or land right away.

PR #10433, opened 2018-04-11
accname: Updated tests

joanmarie:
* Add missing "hidden" style from several tests
* Fix expectations for several tests with presentation role used on img
* Fix bogus expectations on name and description calculations with combobox
* Fix incorrect whitespace in expectations for name from content with spans
* Add test for which coverage was missing


halindrome (approved):
LGTM

PR #10441, opened 2018-04-12
Test vertical tab for base64

annevk:
Apparently Rust's base64 crate strips this, so it would be good to test. Safari is also an outlier here.


annevk:
Bug: https://bugs.webkit.org/show_bug.cgi?id=184529.

Ms2ger (approved):

PR #10442, opened 2018-04-12
Fetch: expand Response.redirect() assertions

annevk:
Needed for https://github.com/whatwg/fetch/issues/664 and https://github.com/whatwg/fetch/pull/600.

Safari already passes these. Chrome and Firefox have `statusText` returning `OK`.

### DIFF
--- a/fetch/api/response/response-static-redirect.html
+++ b/fetch/api/response/response-static-redirect.html
@@ -14,16 +14,24 @@
       var url = "http://test.url:1234/";
       test(function() {
         redirectResponse = Response.redirect(url);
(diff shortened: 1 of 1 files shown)


jakearchibald (changes requested):
Also test the `new Response()` case? Eg https://github.com/w3c/web-platform-tests/blob/master/fetch/api/response/response-init-001.html

annevk:
That is explicitly not changed here, see https://github.com/whatwg/fetch/issues/698.

jakearchibald (approved):
LGTM!

PR #10443, opened 2018-04-12
Use version-specific prefs files when running Firefox

jgraham:
Firefox requires a prefs file to be loaded to ensure that it doesn't
make external network connections, and to make sure that other
settings are appropriate for testing. Previously we always used the
version of the prefs file from master, which is usually fine for
nightly builds, but doesn't work with release builds if it happens
that a pref changed.

This change uses the correct release version of the prefs file for
firefox releases and beta versions. It continues to use the master
version for any nightly build; this could be fixed in some cases if we
are able to access the actual commit hash used to build Firefox, but
probably isn't too bad an approximation.

The caching algorithm was changed so that release versions of the
prefs are cached forever, and the nightly version is updated once per
day (although this doesn't quite match the nightly release cadence,
it's only going to fail in edge cases where the prefs were changed in
the file, but the nightly version was not yet updated, of vice-versa.)

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10443)

### DIFF
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -179,22 +179,70 @@ class Firefox(Browser):
     def find_webdriver(self):
         return find_executable("geckodriver")
 
-    def install_prefs(self, dest=None):
+    def get_version_number(self, binary):
+        version_re = re.compile("Mozilla Firefox (\d+\.\d+(?:\.\d+)?)(a|b)?")
+        proc = subprocess.Popen([binary, "--version"], stdout=subprocess.PIPE)
+        stdout, _ = proc.communicate()
+        stdout.strip()
+        m = version_re.match(stdout)
+        if not m:
+            return None, "nightly"
+        version, status = m.groups()
+        channel = {"a": "nightly", "b": "beta"}
+        return version, channel.get(status, "stable")
+
+    def get_prefs_url(self, version, channel):
+        if channel == "stable":
+            repo = "https://hg.mozilla.org/releases/mozilla-release"
+            tag = "FIREFOX_%s_RELEASE" % version.replace(".", "_")
+        else:
+            repo = "https://hg.mozilla.org/mozilla-central"
+            if channel == "beta":
+                tag = "FIREFOX_%s_BETA" % version.split(".", 1)[0]
+            else:
+                # Always use tip as the tag for nightly; this isn't quite right
+                # but to do better we need the actual build revision, which we
+                # can get if we have an application.ini file
+                tag = "tip"
+
+        return "%s/raw-file/%s/testing/profiles/prefs_general.js" % (repo, tag)
+
+    def install_prefs(self, binary, dest=None):
+        version, channel = self.get_version_number(binary)
+
         if dest is None:
             dest = os.pwd
 
         dest = os.path.join(dest, "profiles")
         if not os.path.exists(dest):
             os.makedirs(dest)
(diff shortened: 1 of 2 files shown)


tools/wpt/browser.py
```diff
@@ -179,22 +180,70 @@ def find_certutil(self):
...
+        version_re = re.compile("Mozilla Firefox (\d+\.\d+(?:\.\d+)?)(a|b)?")
+        proc = subprocess.Popen([binary, "--version"], stdout=subprocess.PIPE)
+        stdout, _ = proc.communicate()
+        stdout.strip()
+        m = version_re.match(stdout)
+        if not m:
+            return None, "nightly"
+        version, status = m.groups()
+        channel = {"a": "nightly", "b": "beta", None: "stable"}
+        return version, channel.get(status, "stable")
```
jugglinmike: I don't know which of the two approaches to defining the default value is more Pythonic, but I don't think we need them both.

tools/wpt/browser.py
```diff
@@ -179,22 +180,70 @@ def find_certutil(self):
...
+            repo = "https://hg.mozilla.org/releases/mozilla-release"
+            tag = "FIREFOX_%s_RELEASE" % version.replace(".", "_")
+        else:
+            repo = "https://hg.mozilla.org/mozilla-central"
+            if channel == "beta":
+                tag = "FIREFOX_%s_BETA" % version.split(".", 1)[0]
+            else:
+                # Always use tip as the tag for nightly; this isn't quite right
+                # but to do better we need the actual build revision, which we
+                # can get if we have an application.ini file
```
jugglinmike: This hint for future improvement is appreciated!

tools/wpt/browser.py
```diff
@@ -179,22 +180,70 @@ def find_certutil(self):
...
+        # If we don't have a recent download, grab the url
+        if not have_cache:
+            url = self.get_prefs_url(version, channel)
+
+            with open(cache_file, "wb") as f:
+                print("Installing test prefs from %s" % url)
+                resp = get(url)
                 f.write(resp.content)
+        else:
+            print("Using cached test prefs from %s" % cache_file)
```
jugglinmike: We have a `logger` instance in this file. Mind using that instead?

tools/wpt/browser.py
```diff
@@ -1,4 +1,5 @@
 import logging
+import glob
```
jugglinmike: This line may be removed

jugglinmike (changes requested):
Looking good, James! Just a few small requests

jgraham:
I fixed everything apart from the logger thing, which seems to be a little complicated. I need to remember what we do to make the stdlib logger work from mozlog, because at the moment it's looking pretty broken to me, but then it works later so I guess I'm missing something.

I think we should do that in a followup since it's not really a change in this PR (we were using `print` before these changes as well).

jugglinmike:
> I think we should do that in a followup since it's not really a change in this PR (we were using print before these changes as well).

Works for me. Could you add a link here when you do?

jugglinmike (approved):
Tested locally with Firefox version 59.0.2 and Firefox Nightly. Prior to application of the patch, version 59.0.2 consistently fails due to its making network requests on startup. After applying the patch, these requests are correctly disabled for both versions.

Thanks, @jgraham!

PR #10444, opened 2018-04-12
Improve handling of harness-level errors in wptrunner

jgraham:
Split the internal handling of errors during a test into two cases;
INTERNAL-ERROR that is produced if there's an exception in the harness
and ERROR that is for exceptions reported by the test. Both are still
reported as ERROR to the user, but in the INTERNAL-ERROR case the
runner is always restarted, just like EXTERNAL-TIMEOUT, since we
assume that the internal state is compromised somehow.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10444)

### DIFF
--- a/tools/wptrunner/wptrunner/testrunner.py
+++ b/tools/wptrunner/wptrunner/testrunner.py
@@ -560,13 +560,18 @@ class TestRunnerManager(threading.Thread):
                                     expected=expected,
                                     stack=result.stack)
 
-        # TODO: consider changing result if there is a crash dump file
-
-        # Write the result of the test harness
+        # We have a couple of status codes that are used internally, but not exposed to the
+        # user. These are used to indicate that some possibly-broken state was reached
+        # and we should restart the runner before the next test.
+        # INTERNAL-ERROR indicates a Python exception was caught in the harness
+        # EXTERNAL-TIMEOUT indicates we had to forcibly kill the browser from the harness
+        # because the test didn't return a result after reaching the test-internal timeout
+        status_subns = {"INTERNAL-ERROR": "ERROR",
+                        "EXTERNAL-TIMEOUT": "TIMEOUT"}
         expected = test.expected()
-        status = file_result.status if file_result.status != "EXTERNAL-TIMEOUT" else "TIMEOUT"
+        status = status_subns.get(file_result.status, file_result.status)
 
-        if file_result.status in ("TIMEOUT", "EXTERNAL-TIMEOUT"):
+        if file_result.status in ("TIMEOUT", "EXTERNAL-TIMEOUT", "INTERNAL-ERROR"):
             if self.browser.check_for_crashes():
                 status = "CRASH"
 
@@ -585,7 +590,7 @@ class TestRunnerManager(threading.Thread):
                              extra=file_result.extra)
 
         restart_before_next = (test.restart_after or
-                               file_result.status in ("CRASH", "EXTERNAL-TIMEOUT") or
+                               file_result.status in ("CRASH", "EXTERNAL-TIMEOUT", "INTERNAL-ERROR") or
                                ((subtest_unexpected or is_unexpected) and
                                 self.restart_on_unexpected))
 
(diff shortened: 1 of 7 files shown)


jgraham:
@jugglinmike I think you might be interested in this.

foolip (approved):
LGTM % docs, but I'd also like @jugglinmike to have a look.

Does any documentation of the different statuses exist? Having a description of "INTERNAL-ERROR" under docs/ (added in this PR) would make it easier to review, and benefits everyone trying to understand it in the future.

foolip:
Also, do we need to run this through Gecko's CI to have confidence that this works?

jgraham:
We don't currently have much documentation on the wptrunner internals, so there's nowhere obvious to add this. It isn't ever exposed to the user (it gets mapped down to ERROR before output because mozlog only supports a limited set of statuses, in order for consistency between users).

jgraham:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=890c9d39ed8c4239340954a1aed87fa3c02ba38e&selectedJob=173288584 is the try run started by our bot, although that doesn't include all tests by default. I can do another try run with everything if that looks OK.

tools/wptrunner/wptrunner/testrunner.py
```diff
@@ -563,10 +563,12 @@ def test_ended(self, test, results):
         # TODO: consider changing result if there is a crash dump file
 
         # Write the result of the test harness
+        result_subns = {"INTERNAL-ERROR": "ERROR",
+                        "EXTERNAL-TIMEOUT": "TIMEOUT"}
```
jugglinmike: The word "result" is being applied to the object that contains this status string. What do you think of `status_subns` or `status_map`?

jugglinmike:
@foolip's request for documentation seems reasonable to me, but I also understand @jgraham's reluctance to bloat usage instructions with implementation details. A few ideas:

James writes, "It isn't ever exposed to the user (it gets mapped down to ERROR before output because mozlog only supports a limited set of statuses, in order for consistency between users)." If this detail was included as an in-line comment, we could address Philip's concern about "everyone trying to understand it in the future" without adding to the information we present to the average user

This patch overloads the string "ERROR" a bit. Depending on the context, it could mean "test error, not harness error", but it could also mean "test error or harness error". This could be confusing for others. Since the latter definition is unrestricted, it more closely matches the string. Would it be too onerous to introduce a new value dedicated to the "test error, not harness error" status? (This would also help future contributors get their bearings since the distinction James is making would be explicitly reflected in the code.)

Thanks for taking the lead on this, James!

jgraham:
Adding a new error status to mozlog is hard, because it involves updating it, making a release, updating all formatters to ensure that they don't choke on the new value, and hoping you didn't miss any external ones.

It's about now you realise why it is that people like languages with enums and irrefutable pattern matching… molog would be great in Rust ;)

I'm happy to add a comment about what's going on in the source though.

jugglinmike:
> Adding a new error status to mozlog is hard, because it involves updating it,
> making a release, updating all formatters to ensure that they don't choke on
> the new value, and hoping you didn't miss any external ones.

I expected the path to updating mozlog would be a difficult one, but I also
don't think it's necessarily appropriate. The distinction we're discussion
doesn't necessarily make much sense in a general-purpose logging utility. What
I meant to suggest was the introduction of a new string value for internal use,
one to mirror "INTERNAL-ERROR" (maybe "TEST-ERROR") that would be likewise
translated to "ERROR" for integration with mozlog. My thinking was when reading
the source (and debugging runtime values), seeing that value would be more
helpful because you wouldn't have to wonder, "is this describing a test error?
Or has the mozlog translation already occurred, meaning it might be describing
an internal error?"

> It's about now you realise why it is that people like languages with enums
> and irrefutable pattern matching… mozlog would be great in Rust ;)

Still looking for an excuse to learn that language...

jgraham:
Oh, I see. The `"ERROR"` string comes  pretty much directly from testharness.js, so I'd prefer not to change that internally just to change it back again later.

jugglinmike:
This looks good to me. I don't know how to interpret the Treeherder UI, though. Does it make sense to move forward with that complete test run, now?

gsnedders:
> We don't currently have much documentation on the wptrunner internals, so there's nowhere obvious to add this.

tools/wptrunner/docs?

PR #10446, opened 2018-04-12
test the @font-face font feature settings descriptor (property already tested)

svgeesus:
FFS here standing for font feature settings, not the usual acronym

### DIFF
--- a/css/css-fonts/font-feature-settings-descriptor-01-ref.html
+++ b/css/css-fonts/font-feature-settings-descriptor-01-ref.html
@@ -19,5 +19,5 @@
 <section class="test">
     <p class="ref">AAAAAAAA</p>
     <p class="ref">AAAAAAAA</p>
-	<p class="ref">AAAAAAAA</p>
+    <p class="ref">AAAAAAAA</p>
 </section>
--- a/css/css-fonts/font-feature-settings-descriptor-01-ref.html
+++ b/css/css-fonts/font-feature-settings-descriptor-01-ref.html
@@ -1,0 +1,23 @@
+<!DOCTYPE html>
+<html lang="en">
+<meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
+<title>CSS Test: font-feature-settings descriptor</title>
+<style>
+  @font-face {
+    font-family: fwf;
+    src: url(support/fonts/FontWithFancyFeatures.otf);
+  }
+  .test {
+	  font-family: fwf;
+	  font-size: 2.4em;
+	  line-height: 1.1;
+  }
+</style>
+
+<body>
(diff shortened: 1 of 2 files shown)


css/css-fonts/font-feature-settings-descriptor-01-ref.html
```diff
@@ -0,0 +1,23 @@
...
+	  line-height: 1.1;
+  }
+</style>
+
+<body>
+<p>Test passes if the three lines below are identical, with eight check marks (✓). </p>
+<section class="test">
+    <p class="ref">AAAAAAAA</p>
+    <p class="ref">AAAAAAAA</p>
+	<p class="ref">AAAAAAAA</p>
```
drott: Nit, the indentation seems a bit wonky here?

css/css-fonts/font-feature-settings-descriptor-01.html
```diff
@@ -0,0 +1,39 @@
...
+  }
+  .property {
+    font-feature-settings: "liga" on, "clig" on, "calt" on, "hlig" on, "dlig" on, "onum" on, "smcp" on, "jp90" on;
+  }
+</style>
+
+<body>
+<p>Test passes if the three lines below are identical, with eight check marks (✓). </p>
+<section class="test">
+    <p class="descriptor">CDGFEJQa</p>
```
drott: I think it would be useful to add a comment of how this test works mechanically in combination with the FontWithFancyFeatures font. I assume it does because the ligatures features substitue CFGFEJQa into A's which then comes out as check marks again?

drott (changes requested):
> FFS here standing for font feature settings, not the usual acronym

While I enjoy the pun, I think the commit heading could be clearer with the spelled out word, e.g. "Test per @font-face font-feature-settings default in descriptor".

css/css-fonts/font-feature-settings-descriptor-01.html
drott: PS: Thanks for contributing this useful test case.

svgeesus: It works the same as all the other tests which use that font (there are a lot of them).

For each character in the font (apart from A and B) the resulting glyph is a checkmark if a particular opentype feature is on, and a cross if it is off. A and B are hardcoded to checkmark and cross, respectively, so are useful for reftests.

Documentation [here](https://www.w3.org/People/chris/fwf/index.html) and specific details of [which opentype feature goes with which test](https://www.w3.org/People/chris/fwf/fancytester.html)

svgeesus: The font was made by @litherum by the way, which is why I tagged him as a reviewer

drott (approved):

PR #10461, opened 2018-04-13
Streams: Constructors property lookup order

ricea:
ReadableStream, WritableStream and TransformStream look up and validate
properties on their arguments. Add tests to verify that these lookups
and validations happen in the order specified by the standard.

https://github.com/whatwg/streams/pull/922 adjusts some of the ordering
of these operations. These tests correspond to that change.

A new utility file streams/resources/constructor-ordering.js contains
functions to construct objects which track property lookups and
validations and can cause them to fail.

For each constructor the constructor.js file contains a list of the
expected operations in order, and a test case to verify it.

A test in streams/transform-streams/properties.js already detected the
ordering of property accesses by the TransformStream constructor.
writableStrategy is now accessed before readableStrategy to match the
order of the constructor arguments, and so the expectations of this test
are changed to match.


domenic (approved):
Approving since this accurately and comprehensively tests the spec. I am less sure about the spec, reading these sequences, but we can discuss that elsewhere.

PR #10462, opened 2018-04-13
Accept variation in serialization of 'background'

ewilligers:
Serialization of the CSS property 'background' varies across browsers.

There is an open spec issue:
https://github.com/w3c/csswg-drafts/issues/418

Until the issue is resolved, we accept a variety of serializations.


emilio (approved):
:sadface:

LGTM, thanks :)

PR #10478, opened 2018-04-15
Correct serialization of URL in background-332.html

csnardi:
According to https://drafts.csswg.org/cssom/#serialize-a-url, the serialization of a URL always includes double-quotes, and per https://drafts.csswg.org/css-values/#relative-urls, if a URL appears in a computed style, it is resolved to an absolute URL. Adjust `background-332.html` to match the spec.

### DIFF
--- a/css/css-backgrounds/background-332.html
+++ b/css/css-backgrounds/background-332.html
@@ -22,8 +22,8 @@
 
         test(function() {
             assert_equals(cs.getPropertyValue("background-image"),
-                'url("' + new URL("support/60x60-green.png", window.location.href).href + '")', "background specified value for background-image");
-        }, "background_specified_image");
+                'url("' + new URL("support/60x60-green.png", window.location.href).href + '")');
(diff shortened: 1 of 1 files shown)


css/css-backgrounds/background-332.html
```diff
@@ -22,7 +22,7 @@
 
         test(function() {
             assert_equals(cs.getPropertyValue("background-image"),
-                "url(support/60x60-green.png)", "background specified value for background-image");
+                'url("' + new URL("support/60x60-green.png", window.location.href).href + '")', "background specified value for background-image");
```
emilio: While at it maybe worth changing "specified" by "computed"? This is totally not testing the specified value :)

emilio (approved):

css/css-backgrounds/background-332.html
csnardi: Yep, good catch. Really, that should be done for the entire test (and in all of the `background-33*.html` files), but I just changed the individual test description for now.

PR #10484, opened 2018-04-16
null is not the correct origin for createDocument()

domenic:
See https://dom.spec.whatwg.org/#dom-domimplementation-createdocument


annevk (approved):
See also https://github.com/whatwg/dom/issues/410 about removing this feature entirely.

PR #10491, opened 2018-04-16
Install Ahem for Safari on Sauce

jgraham:
This just gets the latest master version of the font; we could change
things to get the version for the current revision, but it isn't
necessarily worth the extra complexity given how rarely this changes.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10491)

### DIFF
--- a/tools/wptrunner/wptrunner/browsers/sauce_setup/safari-prerun.sh
+++ b/tools/wptrunner/wptrunner/browsers/sauce_setup/safari-prerun.sh
@@ -1,2 +1,3 @@
 #!/bin/bash
+curl https://raw.githubusercontent.com/w3c/web-platform-tests/master/fonts/Ahem.ttf > ~/Library/Fonts/Ahem.ttf
 defaults write com.apple.Safari com.apple.Safari.ContentPageGroupIdentifier.WebKit2JavaScriptCanOpenWindowsAutomatically -bool true


tools/wptrunner/wptrunner/browsers/sauce_setup/safari-prerun.sh
```diff
@@ -1,2 +1,3 @@
 #!/bin/bash
+curl https://raw.githubusercontent.com/w3c/web-platform-tests/master/fonts/Ahem.ttf > ~/Library/Fonts/Ahem.ttf
```
gsnedders: Can we download files over the tunnel? Does the tunnel already exist at the point the prerun script runs? Because if it does, then we could just download http://web-platform.test:8000/fonts/Ahem.ttf, which would be the right version for the revision being run?

gsnedders:
@jugglinmike if you have the Edge side working, can we land that?

gsnedders (changes requested):
See https://github.com/w3c/web-platform-tests/pull/10499#pullrequestreview-112707151

tools/wptrunner/wptrunner/browsers/sauce_setup/safari-prerun.sh
jgraham: Sure, that might be a better appraoch. This one took like 5 minutes to implement and would therefore be good to land in the meantime if only it actually worked  :)

gsnedders (approved):

PR #10499, opened 2018-04-17
[wptrunner] Install Ahem in Sauce Labs Windows VM

jugglinmike:
In gh-10491, @gsnedders [suggested](https://github.com/w3c/web-platform-tests/pull/10491#discussion_r181912073):

> Can we download files over the tunnel? Does the tunnel already exist at the point the prerun script runs? Because if it does, then we could just download http://web-platform.test:8000/fonts/Ahem.ttf, which would be the right version for the revision being run?

That sounded like a great idea ([although we would have to construct the URL based on the relevant `env_config` values](https://github.com/w3c/web-platform-tests/commit/0fdf04d1a851e7717d03512533a9f9548c29a8aa)), and I almost included it here. However, the "prerun" script is currently uploaded with the same file name for every test execution. That would cause confusing problems if two test execution tasks used different configurations but the same Sauce Labs account. That might seem unlikely, but it may actually occur on a regular basis in http://builds.wpt.fyi fairly soon.

I have some thoughts on how we can address that problem, but I recommend we move forward with this simple solution for now because it can only improve result accuracy.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10499)

### DIFF
--- a/tools/wptrunner/wptrunner/browsers/sauce_setup/safari-prerun.sh
+++ b/tools/wptrunner/wptrunner/browsers/sauce_setup/safari-prerun.sh
@@ -1,2 +1,0 @@
-#!/bin/bash
-defaults write com.apple.Safari com.apple.Safari.ContentPageGroupIdentifier.WebKit2JavaScriptCanOpenWindowsAutomatically -bool true
--- a/tools/wptrunner/wptrunner/browsers/sauce_setup/safari-prerun.sh
+++ b/tools/wptrunner/wptrunner/browsers/sauce_setup/safari-prerun.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+defaults write com.apple.Safari com.apple.Safari.ContentPageGroupIdentifier.WebKit2JavaScriptCanOpenWindowsAutomatically -bool true
--- a/fonts/README.md
+++ b/fonts/README.md
@@ -1,2 +1,6 @@
 This directory only contains auxiliary font files used by other tests. See
 /css-fonts for tests covering the CSS Fonts Module specification.
+
+The font named `Ahem.ttf` is referenced from the project documentation and the
+CLI's scripts for provisioning virtual machines provided by Sauce Labs. If that
+file is re-located, the references should be updated accordingly.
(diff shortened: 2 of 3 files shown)


tools/wptrunner/wptrunner/browsers/sauce_setup/safari-prerun.sh
```diff
@@ -1,2 +0,0 @@
-#!/bin/bash
-defaults write com.apple.Safari com.apple.Safari.ContentPageGroupIdentifier.WebKit2JavaScriptCanOpenWindowsAutomatically -bool true
```
gsnedders: Don't delete this in this PR!

jugglinmike: Oh, of course!

gsnedders (changes requested):
> That sounded like a great idea (although we would have to construct the URL based on the relevant env_config values), and I almost included it here. However, the "prerun" script is currently uploaded with the same file name for every test execution. That would cause confusing problems if two test execution tasks used different configurations but the same Sauce Labs account. That might seem unlikely, but it may actually occur on a regular basis in http://builds.wpt.fyi fairly soon.

AIUI, this is a problem because we upload the script to Sauce Storage and then download it from there before running the tests.

Instead of doing this, we should just download the [pre-run script](https://wiki.saucelabs.com/display/DOCS/Setting+Up+Pre-Run+Executables) over HTTP across the tunnel (this will require whitelisting `tools/wptrunner/wptrunner/browsers/sauce_setup/` wherever we whitelist `tools/runner`), using the `sub` wptserve feature to add in the host/port, and downloading the correct version of Ahem for the revision under test.

jgraham (approved):

jugglinmike:
@gsnedders [I implemented the more advanced integration with the "prerun" feature](https://github.com/w3c/web-platform-tests/compare/master...bocoup:gh-9205-ahem-sauce-edge-tunnel) ([right alongside you, actually](https://github.com/w3c/web-platform-tests/compare/master...gsnedders:sauce-ahem)).  Unfortunately, neither of our implementations work. As far as I can tell, the Sauce Labs virtual machines are *not* configured to use the Sauce Connect Tunnel as a proxy when the pre-run script is executed. Because this problem is starting to draw more of our time (and because I consider it orthogonal to the original issue), I've created a dedicated issue to continue that work: gh-10513

In the mean time, this pull request contains a naive implementation where the script downloads from the `master` branch of the GitHub repository. It's brittle in its own way, but I can verify that it causes the tests to pass in Edge today ([demonstration available here](https://saucelabs.com/beta/tests/e26c6c5d555745c9a5265523c37db70f/commands#8)).  As my reviewer, I'll leave it to you to decide whether it should be landed as a short term improvement while we continue working on a more robust solution.

gsnedders (approved):

PR #10500, opened 2018-04-17
refactor to use test_driver

marcoscaceres:


payment-request/payment-request-abort-method.https.html
```diff
@@ -0,0 +1,77 @@
...
+  promise_test(async t => {
+    // request is in "created" state
+    const request = new PaymentRequest(defaultMethods, defaultDetails);
+    await promise_rejects(t, "InvalidStateError", request.abort());
+  }, `Throws if the promise [[state]] is not "interactive"`);
+
+  const button = document.getElementById("button");
+
+  promise_test(async t => {
+    button.onclick = async () => {
```
marcoscaceres: @gsnedders, this is my first attempt to use `test_driver`, does this look ok?

foolip:
@kereliuk, can you review?

marcoscaceres:
So, @gsnedders checked and it looks like we are going to be good on Safari soon (works in Safari TP!). However, it's unclear what the story is going to  be on Edge. @MSFTkihans, could you please investigate support of Web Driver? Or does anyone know someone we can ask over on the Microsoft side? I'd really like for the Edge team to be onboard here, as then we can migrate a bunch of other tests to use this.

foolip:
I think @thejohnjansen should be able to find the right contact for Edge.

kereliuk (approved):
r+

This looks good! Also for me this is failing on Firefox, but that might be expected.

foolip:
This is using `test_driver.click(button)`, and that already works on Edge:
https://wpt.fyi/uievents/order-of-events/mouse-events/click-cancel.html

Safari doesn't work because `document.elementsFromPoint` isn't there. Safari 11.0 is pretty old, though, maybe it's been added since. @youennf?

marcoscaceres:
> As long as they're manual tests, they're pretty much shutting off everyone as basically nobody will ever run them, hence I'm not convinced it's any regression.

Agree. We do manually run them at Mozilla, Google, and Apple to check conformance while implementing, and to help get us out of Candidate Recommendation.... but after CR... 😢.   

> Ultimately, if they decide not to (internally) support the cross-browser way to do test automation that we discussed at length, that's their problem IMO.

Easy, 🐯!  We are all doing the best we can here. 

The test automation is limited because this needs to deal with extremely complicated browser UI (the Payment Sheet). We are going to hit this a lot with a lot of the APIs coming down the standards pipe. 

I don't have a solution for that, apart from maybe having pref-enabled browser testing modes. I don't know what you'll all have been discussing as possible solutions for the web, but if there are some documents, I'd like to read them.

foolip:
@kereliuk, do we have a concise description somewhere of the WebDriver API + testdriver.js that we're pursuing for most automation needs?

kereliuk:
@foolip Nope, but I think this is a good thing to have. Will put on roadmap!

InstyleVII:
@marcoscaceres What's the question related to WebDriver here? We support the JWP dialect for older builds (RS4 and below) and are actively working on the W3C dialect for RS5.

thejohnjansen:
I'm really sorry for the delay in responding. It turns out that I am not good at GitHub notifications yet and this was buried... 

No excuses.

It's not clear to me yet what would block Edge from working with this. @marcoscaceres, do you have an example of how this fails in Edge?

marcoscaceres:
@thejohnjansen, sorry, I had misread somewhere that `test_driver` might not be supported by Edge. I'm really happy to hear that it should work.

What's confused me is that the WPT's README section on [Running Tests Automatically](https://github.com/w3c/web-platform-tests#running-tests-automatically), states that it's not possible to provide "edge" as an argument. It says: 

> where product is currently `firefox` or `chrome`

Is that expected to change?

gsnedders:
@marcoscaceres Ergh, our duplicated documentation bites us again (turns out that having to remember to update both README.md and the actual documentation every time any one of a billion things changes is hard). Can you file a bug for that? Also note that wptrunner is only relevant insofar as Mozilla's infra and wpt.fyi is concerned: every other vendor has their own running infrastructure for it.

kereliuk:
Done #10684

PR #10509, opened 2018-04-17
Fetch: MediaSource blob: URLs

annevk:
For https://github.com/whatwg/fetch/pull/708.


annevk:
It seems that Safari throwing for MediaStream is arguably correct per https://w3c.github.io/mediacapture-main/ so we shouldn't add tests that encourage different behavior from that.

annevk:
Removed that, should be all good again.

fetch/api/basic/mediasource.window.js
```diff
@@ -0,0 +1,3 @@
+promise_test(t => {
+  return promise_rejects(t, new TypeError(), fetch(URL.createObjectURL(new MediaSource())));
```
foolip: Move constructor out of `promise_rejects` here too?

foolip (approved):
LGTM with nit

PR #10515, opened 2018-04-18
URL.createObjectURL should not work with MediaStream

annevk:
Change the one existing test to no longer rely on it (and clearly mark it as manual) and add an entry to historical.html for it.

### DIFF
--- a/mediacapture-streams/historical.html
+++ b/mediacapture-streams/historical.html
@@ -17,6 +17,7 @@ test(function() {
 }, "navigator.mozGetUserMedia should not exist");
 
 test(() => {
-  assert_throws(new TypeError(), () => URL.createObjectURL(new MediaStream()));
+  const mediaStream = new MediaStream();
+  assert_throws(new TypeError(), () => URL.createObjectURL(mediaStream));
 }, "Passing MediaStream to URL.createObjectURL() should throw");
 </script>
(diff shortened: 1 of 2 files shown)


annevk:
Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1454889.

mediacapture-streams/historical.html
```diff
@@ -15,4 +15,8 @@
 test(function() {
   assert_false("mozGetUserMedia" in navigator);
 }, "navigator.mozGetUserMedia should not exist");
+
+test(() => {
+  assert_throws(new TypeError(), () => URL.createObjectURL(new MediaStream()));
```
Ms2ger: Can you move the creation of the `MediaStream` object out of the `assert_throws`?

foolip (approved):

mediacapture-streams/historical.html
foolip: Yeah, that'd be good.

foolip:
@Orphis this is why I poked you about finishing https://crbug.com/800767 :)

annevk:
Filed https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/17021183/ on Edge so all failing browsers have a bug now.

Ms2ger (approved):

PR #10522, opened 2018-04-18
[css-properties-values-api] add OWNERS

zcorpan:
(I could not select @esprehn in reviewers...)


astearns (commented):
Both @shans and @esprehn are working on other projects now, so probably shouldn't be added here.

astearns (approved):

shans (approved):
I probably don't need to approve this any more...

tabatkins (approved):

PR #10525, opened 2018-04-18
Parsing Content-Type

annevk:
See https://github.com/whatwg/mimesniff/issues/30#issuecomment-382721437 and https://github.com/whatwg/mimesniff/issues/30#issuecomment-436632541 for the processing model this is meant to test.

### DIFF
--- a/fetch/content-type/README.md
+++ b/fetch/content-type/README.md
@@ -2,7 +2,7 @@
 
 An array of tests. Each test has these fields:
 
-* `input`: an array of values for the `Content-Type` header. A harness needs to run the test twice if there are multiple values. One time with the values concatenated with `,` followed by a space and one time with multiple `Content-Type` declarations, each on their own line with one of the values, in order.
+* `contentType`: an array of values for the `Content-Type` header. A harness needs to run the test twice if there are multiple values. One time with the values concatenated with `,` followed by a space and one time with multiple `Content-Type` declarations, each on their own line with one of the values, in order.
 * `encoding`: the expected encoding, null for the default.
 * `mimeType`: the result of extracing a MIME type and serializing it.
 * `documentContentType`: the MIME type expected to be exposed in DOM documents.
@@ -13,7 +13,7 @@ An array of tests. Each test has these fields:
 
 An array of tests, surprise. Each test has these fields:
 
-* `input`: see above.
+* `contentType`: see above.
 * `executes`: whether the script is expected to execute.
 * `encoding`: how the script is expected to be decoded.
 
--- a/fetch/content-type/README.md
+++ b/fetch/content-type/README.md
@@ -1,0 +1,20 @@
+# `resources/content-types.json`
+
+An array of tests. Each test has these fields:
+
+* `input`: an array of values for the `Content-Type` header. A harness needs to run the test twice if there are multiple values. One time with the values concatenated with `,` followed by a space and one time with multiple `Content-Type` declarations, each on their own line with one of the values, in order.
+* `encoding`: the expected encoding, null for the default.
+* `mimeType`: the result of extracing a MIME type and serializing it.
+* `documentContentType`: the MIME type expected to be exposed in DOM documents.
+
+(These tests are currently somewhat geared towards browser use, but could be generalized easily enough if someone wanted to contribute tests for MIME types that would cause downloads in the browser or some such.)
+
+# `resources/script-content-types.json`
+
+An array of tests, surprise. Each test has these fields:
+
+* `input`: see above.
+* `executes`: whether the script is expected to execute.
+* `encoding`: how the script is expected to be decoded.
+
+These tests are expected to be loaded through `<script src>` and the server is expected to set `X-Content-Type-Options: nosniff`.
--- a/fetch/content-type/resources/content-types.json
+++ b/fetch/content-type/resources/content-types.json
@@ -1,0 +1,80 @@
+[
+  {
+    "input": ["text/html", "text/plain"],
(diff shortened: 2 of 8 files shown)


annevk:
@MattMenke2 could you review these too?

MattMenke2:
Happy to review the test cases, but I'm not at all familiar with the python+js environment used by the wpt tests, so at least that aspect should probably be reviewed by someone familiar with the test environment.

fetch/content-type/resources/content-types.json
```diff
@@ -0,0 +1,104 @@
...
+    "encoding": null,
+    "mimeType": "text/html",
+    "documentContentType": "text/html"
+  },
+  {
+    "input": ["text/html;\"", "\\\"", "text/plain"],
+    "encoding": null,
+    "mimeType": "text/html",
+    "documentContentType": "text/html"
+  }
```
MattMenke2: This seems like a pretty complete list to me.  Some suggestions:

// Make sure values other than charset are discarded from earlier lines (Should we test "boundary", too?  It's the only other value supported in browsers, so could imagine boundary and charset behaving the same, while every other values behaved differently):
"input": ["text/plain;charset=gbk;x=foo", "text/plain"] => "text/plain; charset=GBK"

// Make sure close quotes work across input lines.
"input": ["text/html;\"", "\\\"", "text/plain", "\";charset=GBK"] => "text/html; charset=GBK"

// This is largely the same as above, but for the MIME type itself, and without the extra entire quoted line - don't think it adds a whole lot of value, but still think it's nice to have the close quote and the final value on different lines.  Could also do a variant of this without the "\"", but think the value there is even more minimal than this one.
"input": ["text/html;\"",  "\"", "text/plain"] => "text/plain"

// Make sure that quoted MIME types don't work.  A case testing this perhaps fits better in wpt/mimesniff/mime-types/resources/mime-types.json.
["text/plain", "\"text/html\""] => "text/plain"

fetch/content-type/resources/script-content-types.json
```diff
@@ -0,0 +1,87 @@
...
+    "executes": false,
+    "encoding": null
+  },
+  {
+    "input": ["\"text/javascript\""],
+    "executes": false,
+    "encoding": null
+  },
+  {
+    "input": ["text / javascript"],
```
MattMenke2: Suggest splitting this in two, one with the space in each location.

Also, should the cases be checked (instead or additionally) in wpt/mimesniff/mime-types/resources/mime-types.json?

fetch/content-type/resources/script-content-types.json
```diff
@@ -0,0 +1,87 @@
...
+    "executes": true,
+    "encoding": null
+  },
+  {
+    "input": ["text/javascript", "error"],
+    "executes": true,
+    "encoding": null
+  },
+  {
+    "input": ["text/javascript;charset=windows-1252", "x/x", "text/javascript"],
```
MattMenke2: Should the first one also be of a type that doesn't execute, for this test and the next 3 as well?

fetch/content-type/script.window.js
```diff
@@ -0,0 +1,48 @@
...
+    const script = document.createElement("script");
+      t.add_cleanup(() => {
+      script.remove()
+      self.isThereABetterWay = undefined;
+    });
+    script.src = getURL(testData.input, singleHeader);
+    document.head.appendChild(script);
+    if (testData.executes) {
+      script.onload = t.step_func_done(() => {
+        assert_equals(self.isThereABetterWay, testData.encoding === "windows-1252" ? "â‚¬" : "€");
```
MattMenke2: Is there a reason for checking the encoding here?  Also, I have no idea what all the isThereABetterWay stuff is doing.

fetch/content-type/resources/script-content-types.json
annevk: Then the encoding wouldn't copy over.

fetch/content-type/script.window.js
annevk: It's basically making sure the script got decoded correctly. `isThereABetterWay` is a global variable that the script sets with the decoded string value. It's called that way because I was hoping someone would come up with a better mechanism than a global variable. I'll rename it to something more appropriate.

annevk:
Thanks a lot, these are great. I'm not doing `boundary` for now as that basically requires testing (and defining...) that format, which I want to do at some point, but I'm not gonna block on.

I fixed your other points (or left comments), and added documentation to avoid renaming `input` (and since documentation seems good anyway).

annevk:
@domenic note that I added some new generic MIME type tests here as well per feedback, you want to run those against your implementation.

fetch/content-type/README.md
```diff
@@ -0,0 +1,20 @@
+# `resources/content-types.json`
+
+An array of tests. Each test has these fields:
+
+* `input`: an array of values for the `Content-Type` header. A harness needs to run the test twice if there are multiple values. One time with the values concatenated with `,` followed by a space and one time with multiple `Content-Type` declarations, each on their own line with one of the values, in order.
```
MattMenke2: I think think it's better to rename input so the tests are more self-contained.  Note going to block on that, though.

MattMenke2 (approved):

domenic:
Updates to mimesniff/mime-types/resources/mime-types.json LGTM. One day maybe we'll have a jsdom implementation of fetch that can fully test the other JSON files.

annevk:
Fair enough, I renamed `input` to `contentType`.

PR #10526, opened 2018-04-18
Test padded and unpadded key values

robyoder:
* Chrome will not import a JWK object if one of the values contains a leading 0 byte. 
* Edge will not import a JWK object if one of the values is less than the correct byte length, i.e. not containing the leading 0 padding. 
* Firefox and Safari import them either way.

This means we have to convert all key values appropriately based on the current browser's behavior before importing the keys.

This PR adds a test to verify that both formats are accepted (the behavior of Safari and Firefox). It's my first PR, so I welcome any help with this description or the actual changes. I also don't know if this is the correct test, or if Edge or Chrome's behavior is correct and Safari and Firefox are wrong.

### DIFF
--- a/WebCryptoAPI/import_export/rsa_importKey_jwkPadding.https.worker.js
+++ b/WebCryptoAPI/import_export/rsa_importKey_jwkPadding.https.worker.js
@@ -1,0 +1,4 @@
+importScripts("/resources/testharness.js");
+importScripts("rsa_importKey_jwkPadding.js");
+run_test();
+done();
--- a/WebCryptoAPI/import_export/rsa_importKey_jwkPadding.js
+++ b/WebCryptoAPI/import_export/rsa_importKey_jwkPadding.js
@@ -1,0 +1,66 @@
+// Test importKey for RSA using padded and unpadded values
+
+function run_test() {
+    var subtle = crypto.subtle;
+
+    var alg = {
+        name: "RSA-OAEP",
+        modulusLength: 2048,
+        publicExponent: new Uint8Array([0x01, 0x00, 0x01]),
+        hash: {name: "SHA-1"},
+    }
+
+    var unpaddedDKey = {
+        "alg": "RSA-OAEP",
+        "d": "pU9ikDYCHpRfTmzb8bJ7PfyseNagnnr5dQ45KoK01HgiQQeL4G9CCVSd2Nv--JO9Vdv1iN_HvmvVl6kMVmdjA-PWNVevlTwBAxj8jb-LdqavnGFkVRvOmzu6SsHaL5qgN5UDBIe6u1i8lhBHaIpSFSiGpGDnwm6uvxoT8WSdhLk4uhCqn_ePPrTrIFA-rnGA7CpwTasnJhFI4C0XL--78NaF8JL6yD7lm2pJnYYqEOlZv0wRt3H7w9_FfhEv1VzQm4ODV6Bm9-Ok9mnTUppnb-5ea27jmqswDLW5KA2vmsksuKaceB0nr8Z80OCrFq7Eb7BmGWwcKXvQ4LDDsH8h",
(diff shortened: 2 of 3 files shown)


jimsch:
I have not looked in detail at the keys you have generated.  The rules specified by the JWA document are that all of these fields are encoded using the base64urlUint encoding.  The document explicitly states that adding a leading zero is incorrect behavior for n.  (it is also incorrect to remove leading zeros.)  The document is less clear if leading zeros are allowed for the other fields.  

The test should be expanded to import and then export the keys and check that n is not padded.  It seems reasonable that the unsigned values might have a leading zero for the private key values as the length is not known in advanced.  The leading zero should only be needed if the most significant bit in the value is set.

robyoder:
I was a bit confused by your comment, so I went through [the JWA spec](https://tools.ietf.org/html/rfc7518), and found some interesting info. Sorry in advance for the length of this reply. :)

```
   Base64urlUInt
      The representation of a positive or zero integer value as the
      base64url encoding of the value's unsigned big-endian
      representation as an octet sequence.  The octet sequence MUST
      utilize the minimum number of octets needed to represent the
      value.  Zero is represented as BASE64URL(single zero-valued
      octet), which is "AA".
```

This indicates that leading zeroes are incorrect for values of type `Base64urlUInt`. Leading zeroes would conflict with the "minimum number of octets needed to represent the value" rule.

```
6.3.1.1.  "n" (Modulus) Parameter

   The "n" (modulus) parameter contains the modulus value for the RSA
   public key.  It is represented as a Base64urlUInt-encoded value.

   Note that implementers have found that some cryptographic libraries
   prefix an extra zero-valued octet to the modulus representations they
   return, for instance, returning 257 octets for a 2048-bit key, rather
   than 256.  Implementations using such libraries will need to take
   care to omit the extra octet from the base64url-encoded
   representation.
```

Since the `n` parameter is to be Base64urlUInt-encoded, it should not contain any leading zeroes. So the extra note is redundant (though maybe necessary to call attention to the behavior) because any leading 0 bytes should be removed according to the definition of Base64urlUInt.

```
6.3.2.1.  "d" (Private Exponent) Parameter

   The "d" (private exponent) parameter contains the private exponent
   value for the RSA private key.  It is represented as a Base64urlUInt-
   encoded value.
```

Again, these values should not contain leading zeroes. The document goes on to define the rest of the parameters as `Base64urlUInt`, as you mentioned, meaning none should have leading zero bytes (octets).

So I believe these statements of yours are incorrect:

> it is also incorrect to remove leading zeros
> The document is less clear if leading zeros are allowed for the other fields.

As for this:

> The test should be expanded to import and then export the keys and check that n is not padded.

I assume you're talking about the extra note under the `n` spec. The current `n` value is 256 octets, but an import/export should not produce 257 octets. Is that what you are saying?

> It seems reasonable that the unsigned values might have a leading zero for the private key values as the length is not known in advanced. The leading zero should only be needed if the most significant bit in the value is set.

I'm not completely sure I'm following, but let me try. Are you saying that the 257 octets occur to prevent the most significant bit from being 1 and thus indicating a negative number? I don't follow the part about "the length is not known in advanced". But perhaps these sentences are just musings about the reason for the 257 octets and not something I need to act on.

robyoder:
In the end, I'm gathering that Chrome's behavior (rejecting padded values) is correct. My question then becomes this: should I change the test to assert that importing a key with padded values throws an error? That would make the test fail in Safari and Firefox. Do we test that way here? Or do we allow for browser leniency?

jimsch:
Making sure things are right is always worth the effort.

In terms of the length of 'n', the size is always known for encoding because a 2048 bit key has exactly 2048 bits as the most significant bit is always set.  Otherwise it would be a 2047 bit key.

In terms of what a Base64urlUInt means, I tend to agree that stripping leading zeros is generally a good idea in terms of what happens for export.  This means something for the fields other than 'n' as their length is not going to be fixed for a given size key.  

I would expect that an export would strip all leading zeros.
I would expect than an import will accept that format and could accept the leading zero padded version.

I have no problem with the import behavior that either allows for.  Be forgiving on what you accept as long as it is not going to give a wrong answer.  I care far more about the export behavior.

PR #10535, opened 2018-04-19
Flip serialization of two font-variation-settings tests

csnardi:
The expected computed valid of two font variation settings tests was opposite to the specified order (and per https://drafts.csswg.org/css-fonts-4/#propdef-font-variation-settings, the computed value is "as specified"). Reverse the two axis in order to match the specified order.

### DIFF
--- a/css/css-fonts/variations/font-variation-settings-parsing.html
+++ b/css/css-fonts/variations/font-variation-settings-parsing.html
@@ -11,10 +11,10 @@
     <script>
 
         var valueParserTests = [
-            { value: "'wght' 1000, '9 ~A' -45",           expectedComputedValue: "\"9 ~A\" -45, \"wght\" 1000",    isValid: true,  message: "Axis tag with valid non-letter ascii characters" },
+            { value: "'wght' 1000, '9 ~A' -45",           expectedComputedValue: "\"wght\" 1000, \"9 ~A\" -45",    isValid: true,  message: "Axis tag with valid non-letter ascii characters" },
             { value: "'\u001Fbdc' 123",                   expectedComputedValue: "",                               isValid: false, message: "Invalid character below allowed range (first char)"},
             { value: "'abc\u007F' 123",                   expectedComputedValue: "",                               isValid: false, message: "Invalid character above allowed range (mid char)"},
-            { value: "'wght' 1e3, 'slnt' -450.0e-1 ",     expectedComputedValue: "\"slnt\" -45, \"wght\" 1000",    isValid: true,  message: "Axis values in scientific form are valid" },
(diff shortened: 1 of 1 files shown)


csnardi:
@FremyCompany Can you review? Unless I'm missing something obvious, I can't see why the other serialization would be correct per the spec.

css/css-fonts/variations/font-variation-settings-parsing.html
```diff
@@ -35,6 +35,7 @@
         valueParserTests.forEach(function (testCase) {
             test(() => {
                 var element = document.getElementById("value-parser-test");
+                element.style.fontVariationSettings = "";
```
drott: Why is this needed?

drott (commented):
> was opposite to the specified order.

Can you add a spec reference to the commit msg explaining the serialization order?

css/css-fonts/variations/font-variation-settings-parsing.html
csnardi: The last two failures in https://crbug.com/809950 were not actual failures -- they failed because the last test failed (leading to no resetting of font-variation-settings). Since they were invalid parsing tests, they relied on font-variation-settings not being set prior to the test. While the first commit fixes the test for Chrome, this would fix it for other browsers which might run into the same problem where one test fails immediately before an invalid parsing test.

drott: Thanks for the clarification and finding that out. Perhaps it's useful adding a comment here, like: "Reset to empty in order to allow testing to continue in case of parse errors occuring for the next test value."

csnardi: Good suggestion, I added that.

drott (approved):
Thanks for the quick changes, LGTM.

csnardi:
@drott Could you also merge this? I unfortunately do not have write access to the repo.

emilio:
@csnardi feel free to reach to me if you need to get something merged. I cannot grant you access, but I sure can merge your patches.

Thanks for all the work you do!

csnardi:
@emilio Thank you, will do! :)

FremyCompany:
I think the spec should be updated, because the serialization of the property is not "as specified". I just took a look at the exact text we introduced:

> If the same axis name appears more than once, the value associated with the last appearance supersedes any previous value for that axis. This deduplication is observable by accessing the computed value of this property.

and I agree it doesn't say you should reorder the axis names, so I'm fine with the change.

FremyCompany (commented):
LGTM

csnardi:
@FremyCompany In the property definition box, it does say "Computed value: as specified", so I believe that would apply?

PR #10536, opened 2018-04-19
Update fetch and service worker tests to allow same-origin redirects …

wanderview:
…with no-cors requests.

This attempts to align the initial test changes from #9039 with what finally landed in the spec.  It also fixes some thing like https-vs-http that crept into the fetch test in the last merge.

@annevk @mattto @youennf


youennf (approved):
Thanks!
WebKit is indeed throwing in the case of no-cors/manual on a same-origin URL load while the spec allows it.

PR #10544, opened 2018-04-20
[wptrunner] Re-enable secure WebSocket server

jugglinmike:
Commit bfef1f20a419d24633e48d24c14e6a7503e1d48c organized WPT
configuration management into a dedicated class. As part of this change,
it modified the way certain properties were initialized. The original
implementation included the following code:

    config = serve.merge_json(default_config, local_config)

The `merge_json` method operated on object values recursively. Any
properties in `default_config["ports"]` that were *not* present in
`local_config["ports"]` were persisted.

The commit referenced above translated that code to:

    config = serve.Config(override_ssl_env=self.ssl_env, **default_config)
    config.ports = {
        "http": [8000, 8001],
        "https": [8443],
        "ws": [8888]
    }

In this factoring, the `config.ports` property is completely
overwritten.  Because the `default_config` object contains a "wss"
property that is not present in the override, this new approach disables
the default secure WebSocket server.

Re-implement the original behavior by selectively overwriting only the
specified configuration properties and persisting any
previously-existing properties.

---

As an alternative, it might make sense to remove the `config.ports =` statement from `environment.py` and instead declaratively set those values in `config.default.json`. That could be easier to follow (fewer sources of configuration), but I'm not sure where else `config.default.json` is used. The file currently demonstrates the use of the special value `"auto"`. Users are encouraged to copy the file and make local modifications, so the illustrative purpose may be a good reason to leave it as is.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10544)

### DIFF
--- a/infrastructure/assumptions/wpt-server-http.sub.html
+++ b/infrastructure/assumptions/wpt-server-http.sub.html
@@ -1,0 +1,172 @@
+<!doctype html>
+<html>
+  <head>
+    <title>WPT Server checker</title>
+    <meta charset="utf-8" />
+    <script src="/resources/testharness.js"></script>
+    <script src="/resources/testharnessreport.js"></script>
+  </head>
+</body>
+<body>
+<script>
+function check(protocol, domain, port, done) {
+  var url = protocol + '://' + domain + ':' + port + '/media/1x1-green.png';
+  var img = document.createElement('img');
+  img.setAttribute('src', url);
+  img.style.display = 'none';
+  img.onerror = function() {
+    done(false);
+  };
+  img.onload = function() {
+    done(true);
+  };
+
+  document.body.appendChild(img);
+}
+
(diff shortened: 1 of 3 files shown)


lukebjerring:
Related to https://github.com/w3c/web-platform-tests/pull/10557
IMHO This PR is a better fix (actually fixes the overwrite problem), and is tested. 
@jgraham @gsnedders - was #10557 preferred for some reason? if not, can this get a review?

gsnedders:
@lukebjerring Being at BlinkOn I entirely missed this PR! No strong preference as to which way we solve this.

jugglinmike:
I've rebased this changeset to just introduce the regression tests and submitted the result here: https://github.com/w3c/web-platform-tests/pull/10619

PR #10548, opened 2018-04-20
Parsing Content-Length

annevk:

### DIFF
--- a/fetch/content-length/parsing.window.js
+++ b/fetch/content-length/parsing.window.js
@@ -1,0 +1,19 @@
+promise_test(() => {
+  return fetch("resources/content-lengths.json").then(res => res.json()).then(runTests);
+}, "Loading JSON…");
+
+function runTests(testUnits) {
+  testUnits.forEach(testUnit => {
+    const input = encodeURIComponent(testUnit.input);
+    promise_test(t => {
+      const result = fetch("resources/content-length.py?length=" + input);
+      if (testUnit.output === null) {
+        return promise_rejects(t, new TypeError(), result);
+      } else {
+        return result.then(res => res.text()).then(text => {
+          assert_equals(text.length, testUnit.output);
+        });
+      }
+    }, "Content-Length: " + input);
+  });
+}
--- a/fetch/content-length/parsing.window.js
+++ b/fetch/content-length/parsing.window.js
@@ -8,7 +8,7 @@ function runTests(testUnits) {
     promise_test(t => {
       const result = fetch("resources/content-length.py?length=" + input);
       if (testUnit.output === null) {
-        return promise_rejects(t, new TypeError(), result);
+        return promise_rejects_js(t, TypeError, result);
       } else {
         return result.then(res => res.text()).then(text => {
           assert_equals(text.length, testUnit.output);
--- a/fetch/content-length/parsing.window.js
+++ b/fetch/content-length/parsing.window.js
@@ -14,6 +14,6 @@ function runTests(testUnits) {
           assert_equals(text.length, testUnit.output);
         });
       }
-    }, "Content-Length: " + input);
+    }, input);
   });
 }
--- a/fetch/content-length/resources/content-lengths.json
+++ b/fetch/content-length/resources/content-lengths.json
@@ -12,6 +12,10 @@
     "output": 42
   },
   {
+    "input": "Content-Length: 42\r\nContent-Length: 42,42",
+    "output": 42
+  },
+  {
     "input": "Content-Length: 30",
     "output": 30
   },
@@ -20,15 +24,43 @@
     "output": null
   },
   {
+    "input": "Content-Length: 30,42",
+    "output": null
+  },
+  {
     "input": "Content-Length: 42\r\nContent-Length: 30",
     "output": null
   },
   {
-    "input": "Content-Length: aaaah",
+    "input": "Content-Length: 30\r\nContent-Length: 42",
     "output": null
   },
   {
+    "input": "Content-Length: aaaah\r\nContent-Length: nah",
+    "output": null
+  },
+  {
+    "input": "Content-Length: aaaah, nah",
+    "output": null
+  },
+  {
+    "input": "Content-Length: aaaah\r\nContent-Length: aaaah",
+    "output": 42
+  },
+  {
+    "input": "Content-Length: aaaah, aaaah",
+    "output": 42
+  },
+  {
+    "input": "Content-Length: aaaah",
+    "output": 42
+  },
+  {
     "input": "Content-Length: 42s",
+    "output": 42
+  },
+  {
+    "input": "Content-Length: -1",
     "output": 42
   }
 ]
--- a/fetch/content-length/resources/content-lengths.json
+++ b/fetch/content-length/resources/content-lengths.json
@@ -1,0 +1,34 @@
+[
+  {
+    "input": "Content-Length: 42",
+    "output": 42
+  },
+  {
+    "input": "Content-Length: 42,42",
+    "output": 42
+  },
+  {
+    "input": "Content-Length: 42\r\nContent-Length: 42",
+    "output": 42
+  },
+  {
+    "input": "Content-Length: 30",
+    "output": 30
+  },
+  {
+    "input": "Content-Length: 42,30",
+    "output": null
+  },
+  {
+    "input": "Content-Length: 42\r\nContent-Length: 30",
+    "output": null
+  },
+  {
+    "input": "Content-Length: aaaah",
+    "output": null
+  },
+  {
+    "input": "Content-Length: 42s",
+    "output": 42
+  }
+]
--- a/fetch/content-length/resources/content-lengths.json
+++ b/fetch/content-length/resources/content-lengths.json
@@ -20,6 +20,18 @@
     "output": 30
   },
   {
+    "input": "Content-Length: 30,30",
+    "output": 30
+  },
+  {
+    "input": "Content-Length: 30\r\nContent-Length: 30",
+    "output": 30
+  },
+  {
+    "input": "Content-Length: 30\r\nContent-Length: 30,30",
+    "output": 30
+  },
+  {
     "input": "Content-Length: 42,30",
     "output": null
   },
@@ -36,19 +48,19 @@
     "output": null
   },
   {
-    "input": "Content-Length: 42,",
+    "input": "Content-Length: 30,",
     "output": null
   },
   {
-    "input": "Content-Length: ,42",
+    "input": "Content-Length: ,30",
     "output": null
   },
   {
-    "input": "Content-Length: 42\r\nContent-Length: \t",
+    "input": "Content-Length: 30\r\nContent-Length: \t",
     "output": null
   },
   {
-    "input": "Content-Length: \r\nContent-Length: 42",
+    "input": "Content-Length: \r\nContent-Length: 30",
     "output": null
   },
   {
--- a/fetch/content-length/resources/content-lengths.json
+++ b/fetch/content-length/resources/content-lengths.json
@@ -36,6 +36,22 @@
     "output": null
   },
   {
+    "input": "Content-Length: 42,",
+    "output": null
+  },
+  {
+    "input": "Content-Length: ,42",
+    "output": null
+  },
+  {
+    "input": "Content-Length: 42\r\nContent-Length: \t",
+    "output": null
+  },
+  {
+    "input": "Content-Length: \r\nContent-Length: 42",
+    "output": null
+  },
+  {
     "input": "Content-Length: aaaah\r\nContent-Length: nah",
     "output": null
   },
@@ -57,6 +73,10 @@
   },
   {
     "input": "Content-Length: 42s",
+    "output": 42
+  },
+  {
+    "input": "Content-Length: 30s",
     "output": 42
   },
   {
--- a/fetch/content-length/resources/content-lengths.json
+++ b/fetch/content-length/resources/content-lengths.json
@@ -94,5 +94,21 @@
   {
     "input": "Content-Length: -1",
     "output": 42
+  },
+  {
+    "input": "Content-Length: 0x20",
+    "output": 42
+  },
+  {
+    "input": "Content-Length: 030",
+    "output": 30
+  },
+  {
+    "input": "Content-Length: 030\r\nContent-Length: 30",
+    "output": null
+  },
+  {
+    "input": "Content-Length: 030, 30",
+    "output": null
   }
 ]
--- a/fetch/content-length/resources/content-lengths.json
+++ b/fetch/content-length/resources/content-lengths.json
@@ -118,5 +118,9 @@
   {
     "input": "Content-Length:30\r\nContent-Length:,\r\nContent-Length:30",
     "output": null
+  },
+  {
+    "input": "Content-Length: ",
+    "output": 42
   }
 ]
--- a/fetch/content-length/resources/content-lengths.json
+++ b/fetch/content-length/resources/content-lengths.json
@@ -110,5 +110,13 @@
   {
     "input": "Content-Length: 030, 30",
     "output": null
+  },
+  {
+    "input": "Content-Length: \"30\"",
+    "output": null
+  },
+  {
+    "input": "Content-Length:30\r\nContent-Length:,\r\nContent-Length:30",
+    "output": null
   }
 ]
--- a/fetch/content-length/resources/content-lengths.json
+++ b/fetch/content-length/resources/content-lengths.json
@@ -113,7 +113,7 @@
   },
   {
     "input": "Content-Length: \"30\"",
-    "output": null
+    "output": 42
   },
   {
     "input": "Content-Length:30\r\nContent-Length:,\r\nContent-Length:30",
--- a/fetch/content-length/resources/content-length.py
+++ b/fetch/content-length/resources/content-length.py
@@ -1,0 +1,9 @@
+def main(request, response):
+    response.add_required_headers = False
+    output =  "HTTP/1.1 200 OK\r\n"
+    output += "Content-Type: text/plain;charset=UTF-8\r\n"
+    output += request.GET.first("length") + "\r\n"
+    output += "\r\n"
+    output += "Fact: this is really forty-two bytes long."
+    response.writer.write(output)
+    response.write()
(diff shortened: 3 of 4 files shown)


annevk:
What I don't understand is why, e.g., `http://web-platform.test:8000/fetch/content-length/resources/content-length.py?length=Content-Length:aaah` reliably returns `Fact: this is really forty-two bytes long.HTTP/1.1 200 OK` in Chrome and Firefox. @jgraham could that be due to the server or this something like #5227 (doesn't seem like it though)?

annevk:
@MattMenke2 @mnot I suspect you have opinions/thoughts here (on the bits that do work properly).

MattMenke2:
My impression is that browsers have been historically reluctant to hard-fail requests when it sees weird responses, as other browsers don't, either, so any browser that does will probably get bug reports about being the only one that breaks broken sites.  Many of Chrome's old weird behaviors at the net layer are based on FireFox's corresponding weird behaviors, so it's not unexpected that they behave the same, except when one or the other has made a conscious effort to be less weird.  I'm not blaming FireFox for this, just explaining why they behave so similarly.

Chrome fails requests when the response has multiple Content-Length headers with different values, but then proceeds to ignore Content-Length headers with invalid values.

So these will work:
Content-Length: 1
Content-Length: 1
(Experimentally, a fair number of servers and/or proxies seem to do this)

Content-Length: Happy potato
Content-Length: Happy potato
(Treated as having no Content-Length field)

But these will fail (And close the connection):
Content-Length: 1
Content-Length: 2

Content-Length: Happy potato
Content-Length: Grapefruit

Since Chrome started rejecting mismatching content-lengths 4 or 5 years ago, the already small number of servers that did this has dropped precipitously, though we did get a number of bugs filed when I implemented that behavior.

I'm pretty sure both of the following would be interpreted by Chrome as having no content length:
Content-Length: 1 1

Content-Length: -1

MattMenke2:
So back to the actual question of what we should be doing:  I'd be fine with tightening up handling here - rejecting responses with invalid Content-Lengths entirely.  I don't have numbers on how common this is, but it's a small enough change that I'd be willing to push on trying and hoping for the best, rather than histograming, waiting months for numbers, and then disabling.

I'd be more concerned about the breakage from not allowing multiple identical Content-Length headers, just because I suspect there would be a lot more fallout from doing that, and I'm not sure that a simpler spec is worth the fallout, though if other browser vendors show interest in tightening up there, Chrome should certainly follow suit.

I assume we should we continue to ignore bad Content-Length values for chunked responses?  I believe the spec says to ignore the header in that case.

MattMenke2:
I can live with maintaining current behavior, though I think it has potential for weirdness.  It certainly seems closer to what browsers do, which makes migration less effort, and means fewer broken sites.

That having been said, Chrome's current behavior of allowing invalid values, but only if all other content-length headers have the same invalid value, seems a bit unexpected to me, and I'm not sure we'd want to standardize that, though the only other option that occurs to me (Without rejecting responses with bad values) is just ignoring all invalid values and only comparing valid ones, which wouldn't be hard to do, just not sure it's any better.

I also suspect it means all browsers will end up treating a Content-Length that doesn't fit in a 64-bit integer the same as an invalid value (i.e., silently ignoring it, since it fails the int64 parser).  While a file that long will lead to other problems, anyways, and this is a bit of an extreme corner case, ignoring the Content-Length in that case seems weird as well.  Admittedly, it's possible to differentiate this from other Content-Length values browsers don't like, I just don't think browsers are generally going to be that careful about it.

annevk:
We treat limits a bit differently in standards. Sometimes they are standardized, but often we just leave them open for implementations to compete on. Here it seems reasonable to not have a strict limit, though perhaps a lower bound should at some point be set somewhere (probably better for that to be in HTTP proper though, if they don't want to adopt all of this).

MattMenke2:
I'm not suggesting some limit be standardized, just saying that accepting but ignoring invalid values is likely to result in more sloppiness around values that are too large in a way that I think is sub-optimal.

annevk:
I see, sorry. Well, if we can push to be stricter that does seem better. If that doesn't work, we can always make sure to test the scenarios you mention and ensure browsers implement the desired result. (That's part of the reason why I set up this small testing framework.)

jugglinmike:
This looks good, although we could expand coverage to highlight another point of contention:

```diff
   {
     "input": "Content-Length: 42,30",
     "output": null
   },
+  {
+    "input": "Content-Length: 30,42",
+    "output": null
+  },
```

Chrome passes this while Firefox reports "30". They both reject

    Content-Length: 30
    Content-Length: 42

but it might be good to add that for symmetry.

annevk:
@jugglinmike thanks! Do you have insight into https://github.com/w3c/web-platform-tests/pull/10548#issuecomment-383099040 perhaps?

jugglinmike:
I don't think the server is at fault here. I ran these requests with cURL and each response included the headers as expected.

The only response that Firefox seems to have trouble parsing is 

- `Content-Length: 42\r\nContent-Length: 30`

In gh-5227, we were grappling with content from one response being written to the beginning of a subsequent response. I don't think that's the case here because if we run the sub-test in isolation, Firefox *still* has trouble with the response.

Along the same lines, Chrome won't parse either of:

- `Content-Length: 42\r\nContent-Length: 30`
- `Content-Length: 42,30`

But it does no better if either of those sub-tests run in isolation.

<details>
<summary>Script</summary>

```
#!/bin/bash

# See
# https://github.com/w3c/web-platform-tests/pull/10548

urls='
http://web-platform.test:8000/fetch/content-length/resources/content-length.py?length=Content-Length%3A%2042
http://web-platform.test:8000/fetch/content-length/resources/content-length.py?length=Content-Length%3A%2042%2C42
http://web-platform.test:8000/fetch/content-length/resources/content-length.py?length=Content-Length%3A%2042%0D%0AContent-Length%3A%2042
http://web-platform.test:8000/fetch/content-length/resources/content-length.py?length=Content-Length%3A%2030
http://web-platform.test:8000/fetch/content-length/resources/content-length.py?length=Content-Length%3A%2042%2C30
http://web-platform.test:8000/fetch/content-length/resources/content-length.py?length=Content-Length%3A%2042%0D%0AContent-Length%3A%2030
http://web-platform.test:8000/fetch/content-length/resources/content-length.py?length=Content-Length%3A%20aaaah
http://web-platform.test:8000/fetch/content-length/resources/content-length.py?length=Content-Length%3A%2042s
'

for url in $urls; do
  curl -sv $url > /dev/null
done
```
</details>

<details>
<summary>output</summary>

```
*   Trying 127.0.0.1...
* Connected to web-platform.test (127.0.0.1) port 8000 (#0)
> GET /fetch/content-length/resources/content-length.py?length=Content-Length%3A%2042 HTTP/1.1
> Host: web-platform.test:8000
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: text/plain;charset=UTF-8
< Content-Length: 42
< 
{ [61 bytes data]
* Excess found in a non pipelined read: excess = 19, size = 42, maxdownload = 42, bytecount = 0
* Connection #0 to host web-platform.test left intact
*   Trying 127.0.0.1...
* Connected to web-platform.test (127.0.0.1) port 8000 (#0)
> GET /fetch/content-length/resources/content-length.py?length=Content-Length%3A%2042%2C42 HTTP/1.1
> Host: web-platform.test:8000
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: text/plain;charset=UTF-8
< Content-Length: 42,42
< 
{ [42 bytes data]
* Connection #0 to host web-platform.test left intact
*   Trying 127.0.0.1...
* Connected to web-platform.test (127.0.0.1) port 8000 (#0)
> GET /fetch/content-length/resources/content-length.py?length=Content-Length%3A%2042%0D%0AContent-Length%3A%2042 HTTP/1.1
> Host: web-platform.test:8000
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: text/plain;charset=UTF-8
< Content-Length: 42
< Content-Length: 42
< 
{ [42 bytes data]
* Connection #0 to host web-platform.test left intact
*   Trying 127.0.0.1...
* Connected to web-platform.test (127.0.0.1) port 8000 (#0)
> GET /fetch/content-length/resources/content-length.py?length=Content-Length%3A%2030 HTTP/1.1
> Host: web-platform.test:8000
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: text/plain;charset=UTF-8
< Content-Length: 30
< 
{ [42 bytes data]
* Excess found in a non pipelined read: excess = 12, size = 30, maxdownload = 30, bytecount = 0
* Connection #0 to host web-platform.test left intact
*   Trying 127.0.0.1...
* Connected to web-platform.test (127.0.0.1) port 8000 (#0)
> GET /fetch/content-length/resources/content-length.py?length=Content-Length%3A%2042%2C30 HTTP/1.1
> Host: web-platform.test:8000
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: text/plain;charset=UTF-8
< Content-Length: 42,30
< 
{ [42 bytes data]
* Connection #0 to host web-platform.test left intact
*   Trying 127.0.0.1...
* Connected to web-platform.test (127.0.0.1) port 8000 (#0)
> GET /fetch/content-length/resources/content-length.py?length=Content-Length%3A%2042%0D%0AContent-Length%3A%2030 HTTP/1.1
> Host: web-platform.test:8000
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: text/plain;charset=UTF-8
< Content-Length: 42
< Content-Length: 30
< 
{ [59 bytes data]
* Excess found in a non pipelined read: excess = 29, size = 30, maxdownload = 30, bytecount = 0
* Connection #0 to host web-platform.test left intact
*   Trying 127.0.0.1...
* Connected to web-platform.test (127.0.0.1) port 8000 (#0)
> GET /fetch/content-length/resources/content-length.py?length=Content-Length%3A%20aaaah HTTP/1.1
> Host: web-platform.test:8000
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: text/plain;charset=UTF-8
< Content-Length: aaaah
< 
* Excess found in a non pipelined read: excess = 42 url = /fetch/content-length/resources/content-length.py?length=Content-Length%3A%20aaaah (zero-length body)
* Connection #0 to host web-platform.test left intact
*   Trying 127.0.0.1...
* Connected to web-platform.test (127.0.0.1) port 8000 (#0)
> GET /fetch/content-length/resources/content-length.py?length=Content-Length%3A%2042s HTTP/1.1
> Host: web-platform.test:8000
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: text/plain;charset=UTF-8
< Content-Length: 42s
< 
{ [42 bytes data]
* Connection #0 to host web-platform.test left intact
```
</details>

annevk:
@jugglinmike so where does the trailing `HTTP/1.1 200 OK` come from?

jgraham:
> @jugglinmike so where does the trailing HTTP/1.1 200 OK come from?

It comes from the fact that you call `response.writer.write()` and then `response.write()`, which doesn't know you already wrote a response and so constructs a new response with no headers or body. The second call is probably wrong. I suspect you want to set `response.close_connection = True`, otherwise the request is going to hang.

annevk:
Thanks, fixed that and added a bunch more tests.

fetch/content-length/resources/content-lengths.json
```diff
@@ -0,0 +1,66 @@
...
+    "input": "Content-Length: aaaah, aaaah",
+    "output": 42
+  },
+  {
+    "input": "Content-Length: aaaah",
+    "output": 42
+  },
+  {
+    "input": "Content-Length: 42s",
+    "output": 42
```
mnot: Should this be a distinct value? Some parser might decide to ignore the 's'...

mnot:
The test cases look reasonable, but I wonder if the number of bytes exposed as the response is always going to be what the browser parses 'content-length' as...

annevk:
What kind of thing do you have in mind?

fetch/content-length/resources/content-lengths.json
```diff
@@ -0,0 +1,66 @@
...
+  {
+    "input": "Content-Length: aaaah, aaaah",
+    "output": 42
+  },
+  {
+    "input": "Content-Length: aaaah",
+    "output": 42
+  },
+  {
+    "input": "Content-Length: 42s",
```
MattMenke2: We should have a test where this is 30s.  It's not clear if the output in this case should be 30 or 42.  Chrome will do the latter, not sure about other browsers.

MattMenke2:
Sorry for the delay.  The tests look good to me, a few other suggestions:

Content-Length: 0x20
Content-Length: 030
Content-Length: 030, 30 (If the above one should be 30, this is relevant.  If the above one is 42, this is much less interesting - I assume it should be 30)

annevk:
So based on https://github.com/whatwg/fetch/pull/813#issuecomment-429366466 I added tests for "empty values". For multiple header fields browsers indeed seem to reject the response, but `,42` or `42,` is "allowed".

annevk:
I've now also covered your suggestions @MattMenke2.

annevk:
It seems `,30` and `30,` should be rejected as well for consistency, but no browser does that at the moment (only when the empty/whitespace value and 30 are in separate headers).

MattMenke2:
I agree with you on the tests (Even if no browser currently implements them).

fetch/content-length/resources/content-lengths.json
```diff
@@ -19,6 +19,18 @@
...
+    "output": 30
+  },
+  {
+    "input": "Content-Length: 30\r\nContent-Length: 30",
+    "output": 30
+  },
+  {
+    "input": "Content-Length: 30\r\nContent-Length: 30,30",
+    "output": 30
+  },
```
MattMenke2: One other suggestion:  We could have a quoted test or two, to make sure quotes aren't handled:

Content-Length: "30"

MattMenke2 (approved):
I'm really happy with this set of tests.  Tried to come up with other weird ideas, and they were basically all covered.

only other thing I guess we could consider is:

Content-Length: 30
Content-Length: ,
Content-Length: 30

Which should result in null (i.e., if headers are merged, it should be 30,,,30, and not 30,30.  It seems unlikely to me that would happen, though, so not sure it's worth testing).

annevk:
@MattMenke2 I forgot why I dropped the ball on this. Probably in part to wait and see if HTTP would define this. It seems that meanwhile Chrome fixed all these tests! However, it does still fail your suggested tests with quotes that I had not got around to adding just yet.

If you could take another look that would be great. I hope that we can define obtaining the value of this header in Fetch, largely by deferring to HTTP now, and then use that operation in XMLHttpRequest for information needed by download progress events.

fetch/content-length/resources/content-lengths.json
```diff
@@ -0,0 +1,122 @@
...
+    "input": "Content-Length: 030\r\nContent-Length: 30",
+    "output": null
+  },
+  {
+    "input": "Content-Length: 030, 30",
+    "output": null
+  },
+  {
+    "input": "Content-Length: \"30\"",
+    "output": null
```
MattMenke2, line 116: Why is is null and not 42?  Seems like this should be analagous to "Content-Length: aaaah" (single header that does not parse as a number)

annevk: Ah right, that makes sense.

Reading https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#field.content-length it also seems that Fetch will still need to define this behavior.

annevk: I'll tackle this next week most likely as part of getting rid of the "total bytes" concept and defining progress events better. If you have suggestions for more tests let me know!

fetch/content-length/parsing.window.js
```diff
@@ -0,0 +1,19 @@
...
+    promise_test(t => {
+      const result = fetch("resources/content-length.py?length=" + input);
+      if (testUnit.output === null) {
+        return promise_rejects_js(t, TypeError, result);
+      } else {
+        return result.then(res => res.text()).then(text => {
+          assert_equals(text.length, testUnit.output);
+        });
+      }
+    }, "Content-Length: " + input);
```
MattMenke2, line 17: Note that all "input" values already start with "Content-Length", so maybe just output "input"?

fetch/content-length/resources/content-lengths.json
```diff
@@ -0,0 +1,122 @@
...
+  },
+  {
+    "input": "Content-Length:30\r\nContent-Length:,\r\nContent-Length:30",
+    "output": null
+  }
```
MattMenke2, line 121: Tests look pretty comprehensive to me, that having been said, more suggestions:

I'm not sure we're ready to standardize on it, but "Content-Length: 50" would be nice.  Ideally, I think we should get 42 bytes and then an error, though not sure browsers (or specs?) agree with that.

Just "Content-Length: " (We have an empty value combined with a non-empty one, but no single empty value test)

We could do some extra tests around different LWS characters / multiple LWS characters, but that seems not too useful.


Thinking further ahead, we should ideally have tests with a compressed body and a Content-Length that matches the uncompressed body length (both when the compressed body is larger and smaller than the Content-Length).  Chrome considers the latter case not an error in the connection-close case, I believe, but we should probably decide on a common set of behaviors there.  It never ends.

annevk: Is that primarily content codings or also other kinds of compression? Currently Fetch only "handles" content codings (I say "handles" as it has numerous issues), but maybe there's more to do?

annevk:
I also created https://github.com/web-platform-tests/wpt/pull/27837 to test this a bit with synthetic responses and to see how XMLHttpRequest uses `Content-Length` for its events.

fetch/content-length/resources/content-lengths.json
MattMenke2: It's only Content-Encoding (Well, and however H2/H3 advertise their version of it - I'm not sure if they encode that information in a different manner, and we map it to a Content-Encoding).

annevk: Okay, I'll update the content codings issue with your suggestion here so that can be tackled together.

annevk:
Fetch side of this is now up at https://github.com/whatwg/fetch/pull/1183 and https://github.com/whatwg/fetch/pull/1184. I'll prolly do the XHR side tomorrow. Review of all these things is appreciated, even if it's just high-level.

MattMenke2:
Still looks good

ghost (commented):
No spig

PR #10571, opened 2018-04-23
Make a check for the initial value of selectionDirection less strict.

Ms2ger:
The standard does not define an initial value.

### DIFF
--- a/html/semantics/forms/textfieldselection/selection-not-application.html
+++ b/html/semantics/forms/textfieldselection/selection-not-application.html
@@ -75,7 +75,7 @@
     }, `selectionEnd on an input[type=${type}] returns a value`);
 
     test(() => {
-      assert_in_array(el.selectionDirection, ["forward", "backward", "none"]);
+      assert_in_array(el.selectionDirection, ["forward", "none"]);
     }, `selectionDirection on an input[type=${type}] returns a value`);
 
     test(() => {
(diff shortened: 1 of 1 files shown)


html/semantics/forms/textfieldselection/selection-not-application.html
```diff
@@ -70,7 +75,7 @@
     }, `selectionEnd on an input[type=${type}] returns a value`);
 
     test(() => {
-      assert_equals(el.selectionDirection, "none");
+      assert_any(assert_equals, el.selectionDirection, ["forward", "backward", "none"]);
```
foolip:
```suggestion
      assert_in_array(el.selectionDirection, ["forward", "backward", "none"]);
```

foolip (approved):
LGTM with nit

tkent-google:
> The standard does not define an initial value.

I think it's defined.

https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#textFieldSelection:concept-textarea/input-selection-3
> The initial selection direction must be "none" if the platform supports that direction, or "forward" otherwise.

foolip:
The results at https://wpt.fyi/results/html/semantics/forms/textfieldselection/selection-not-application.html might shed some light on this. @Ms2ger given @tkent-google's link, are you sure these changes are right?

Ms2ger:
Thanks for the pointer, @tkent-google. I imagine this was tightened a bit since I made this PR. We could theoretically make the allowed values depend on the platform in the UA string or something like that, but let's not.

(If you merge this before I do, please squash.)

foolip:
I've rebased the branch to trigger Taskcluster. It doesn't work if the branch falls too far behind.

foolip:
Results from wpt.fyi look like an improvement across the board, merging.

PR #10575, opened 2018-04-23
Properly define the port for the wss:// protocol

nox:
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10575)


nox:
r? @jgraham or @gsnedders 

This is a patch that has already been made in Gecko, but somehow it was still not upstreamed and I want to update WPT in Servo, where the same bug is hit.

nox:
#10557

PR #10587, opened 2018-04-23
Move innerText/* to html/dom/elements/the-innertext-idl-attribute/

annevk:


annevk:
The naming scheme of html/ is somewhat complex, but I think I got this correct.

annevk:
Since @zcorpan is not asked for review, maybe the OWNERS file should be dropped here?

zcorpan (approved):

tobie:
> Seems like a bug in the bot not picking it up (because the file was moved maybe?). @tobie ?

Yeah, there's no special casing for moving files. Maybe it's something we should add if this becomes more common.

PR #10592, opened 2018-04-23
Update DOMPoint and DOMQuad tests

csnardi:
Per https://drafts.fxtf.org/geometry/#DOMPoint, `DOMPoint`'s constructor accepts zero to four values, and `DOMPointReadOnly` has the exact same constructor. Additionally, `DOMQuad` no longer has a live `bounds` attribute per https://drafts.fxtf.org/geometry/#domquad, instead replacing it with `getBounds()` which cannot be set. Fix tests of `DOMPoint` and `DOMQuad `to reflect these changes.

Also add a historical test to make sure that `DOMQuad.bounds` is removed.

### DIFF
--- a/css/geometry/DOMPoint-001.html
+++ b/css/geometry/DOMPoint-001.html
@@ -21,7 +21,7 @@
             checkDOMPoint(new DOMPoint(), {x:0, y:0, z:0, w:1});
         },'testConstructor0');
         test(function() {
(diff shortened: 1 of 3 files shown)


dirkschulze (approved):
LGTM

Ms2ger (approved):

PR #10593, opened 2018-04-23
Correct computed value of transform

csnardi:
Per https://drafts.csswg.org/css-transforms-1/#serialization-of-the-computed-value, the computed value of a `<transform-list>` is one `matrix()`. Change this test to bring it in line with the spec, matching the behavior of all major browsers.


csnardi:
@ewilligers Can you review? I believe I read the spec correctly here.

ewilligers (approved):

sideshowbarker:
@ewilligers @frivoal is this PR waiting on anything? review from @smfr?

Issue #5915, opened 2017-05-14
Unowned directories


gsnedders:
We should make sure all top-level directories have at least one OWNER:

```
pnin:wpt2 gsnedders$ find . ./css -mindepth 1 -maxdepth 1 -type d '!' -exec test -e "{}/OWNERS" ';' -print
```
* [x] ./orientation-event #10764
* [x] ./worklets #10766
* [x] ./offscreen-canvas #10767
* [x] ./mediasession #10768
* [x] ./server-timing #10770
* [x] ./client-hints #10771
* [x] ./picture-in-picture #10763
* [x] ./css
* [x] ./js https://github.com/w3c/web-platform-tests/pull/10585
* [x] ./mediacapture-fromelement #10772
* [x] ./media-capabilities #10774
* [x] ./.well-known #10761
* [x] ./wasm #10775
* [x] ./acid https://github.com/w3c/web-platform-tests/pull/10585
* [x] ./apng #10777
* [x] ./webgl #10760
* [x] ./imagebitmap-renderingcontext #10767
* [x] ./upgrade-insecure-requests #10779
* [x] ./netinfo #10780
* [x] ./old-tests #10759
* [x] ./credential-management #10781
* [x] ./keyboard-lock #10782
* [x] ./2dcontext #10767
* [x] ./audio-output #10783
* [x] ./remote-playback #10784
* [x] ./secure-contexts #10779
* [x] ./device-memory #10785
* [x] ./trusted-types #10779
* [x] ./innerText https://github.com/w3c/web-platform-tests/pull/10583
* [x] ./budget-api #10758
* [x] ./clear-site-data #10787
* [x] ./infrastructure https://github.com/w3c/web-platform-tests/pull/10612
* [x] ./x-frame-options #11063
* [x] ./BackgroundSync #10757
* [x] ./mediacapture-image #10788
* [x] ./css/css-content https://github.com/w3c/web-platform-tests/pull/10508
* [x] ./css/css-sizing https://github.com/w3c/web-platform-tests/pull/10578
* [x] ./css/css-shadow-parts https://github.com/w3c/web-platform-tests/pull/10551
* [x] ./css/css-scroll-snap https://github.com/w3c/web-platform-tests/pull/10550
* [x] ./css/vendor-imports (the only subdirectory mozilla/ has an OWNERS file)
* [x] ./css/css-font-loading https://github.com/w3c/web-platform-tests/pull/10520
[...]

PR #10594, opened 2018-04-24
Move x-frame-options/ into content-security-policy/

annevk:
The plan of record is for CSP to define X-Frame-Options (with small parts of it in HTML). Moving this directory there to reduce top-level directories (without owner).

Helps with #5915.


mikewest:
Is that the plan of record? :) It's not clear to me that anyone would ever think to look in the CSP spec to determine how XFO is defined... Isn't that what RFC7034 is there for?

Less practically, defining the behavior in terms of `frame-ancestors` means that we need to do some sort of work to make sure we're not firing events/sending violation reports, etc. That seems strange.

mikewest:
(If we actually define a `From-Origin` mechanism, perhaps we could combine it with XFO into some sort of "Limit your embedders" document?)

annevk:
I thought that RFC didn't really match implementations for certain values and also doesn't define things in enough detail (e.g., doesn't integrate with HTML). (Pretty common for RFCs unfortunately...)

And CSP has an inline issue on addressing that, so that's why I thought that was the plan, but I'd be happy with other solutions.

zcorpan:
@mikewest @annevk status on this?

annevk:
It needs to be properly defined in something other than that RFC as that RFC doesn't really go into sufficient detail. My thinking was that CSP would be a good place as it already needs an equivalent algorithm for frame-ancestors.

zcorpan:
Ok. Should I add an OWNERS file meanwhile?

annevk:
Yeah, I guess that's reasonable unless @mikewest is ready to fix all the things here soon.

annevk:
I'm happy to own this directory, FWIW.

mikewest:
@mikewest is not going to fix all the things here soon. I think it's probably worth doing in the context of `from-origin`, since that seems quite related, but it's way down at the bottom of my todo list.

annevk:
FWIW, I no longer think `From-Origin` should do ancestor checks.

PR #5148, opened 2017-03-15
Convert WebSocket tests to .any.js format

annevk:
Fixes #2557.

PR #10598, opened 2018-04-24
Convert WebSocket tests to .any.js format

### LINKED ISSUE #2557: Make websocket tests run in workers
This could follow the same pattern as the tests for the Fetch API: https://github.com/w3c/web-platform-tests/tree/master/fetch/api/basic - have two HTML files (one for workers, one for non-workers) that load the same JS file that contains the actual test.

annevk:
Take 2 (see #5148 for take 1), fixes #2557.


annevk:
(Rebasing the old PR was too annoying. I omitted the `<meta name=variant>` test here and also fixed up the Unicode error and the `window.WebSocket` bits. I think this should be good now.)

Ms2ger:
Does the unicode string need to be literal unicode, or would `\u` escapes work?

zcorpan (approved):
Optional nice-to-have would be to re-indent the files e.g. using `js-beautify`

annevk:
@zcorpan I think that would be better as a follow-up. I suspect if I did that everything would be seen as delete/add rather than a move-with-changes.

PR #10600, opened 2018-04-24
Fetch: use new any.js for general abort test

annevk:
@Ms2ger what am I doing wrong here?


annevk:
Ah, nothing, I just had to kill and startup wptserve. Pfew.

Ms2ger (approved):

PR #10610, opened 2018-04-24
[webrtc-stats] add META.yml

zcorpan:
See https://github.com/w3c/web-platform-tests/pull/10577 (cc @dontcallmedom)

### DIFF
--- a/webrtc-stats/META.yml
+++ b/webrtc-stats/META.yml
@@ -1,0 +1,3 @@
+suggested_reviewers:
+  - henbos
+  - vr000m
--- a/webrtc-stats/OWNERS
+++ b/webrtc-stats/OWNERS
@@ -1,0 +1,2 @@
+@alvestrand
+@vr000m
--- a/webrtc-stats/OWNERS
+++ b/webrtc-stats/OWNERS
@@ -1,2 +1,2 @@
-@alvestrand
+@henbos
 @vr000m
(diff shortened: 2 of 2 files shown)


vr000m (approved):

zcorpan:
@alvestrand are you OK with being listed in the OWNERS file here? It means you’ll get notified any time someone submits a PR. And the tests could only be merged to the master branch after review from one of the OWNERS.

foolip:
@henbos you'd be the best reviewer on the Chromium side for this, right?

henbos (commented):
Me being owner of testing of webrtc-stats spec SGTM, but what is this webrtc-stats/ directory?

Currently stats are tested as part of the webrtc/ directory. We have two specs, like so:

webrtc spec: Defines RTCPeerConnection.getStats(), RTCRtpSender.getStats(), RTCRtpReceiver.getStats(), RTCStatsReport, RTCStats.

webrtc-stats spec: Defines all RTCStats-derived dictionaries that an RTCStatsReport may contain.

We could move tests to a webrtc-stats/ directory there but there would be overlap of behaviors of the two specs and of helpers utils etc, so I don't think we want to do that.

foolip:
@vr000m, can you accept the invite to the web-platform-tests org? Otherwise @wpt-pr-bot will not be able to assign you as a reviewer to any PRs.

alvestrand (approved):

PR #10619, opened 2018-04-24
[wptrunner] Add regression tests for server config

jugglinmike:
A recent commit [1] organized WPT configuration management into a
dedicated class. In so doing, it unintentionally disabled the default
secure WebSocket server. A subsequent commit [2] corrected the
regression.

Add a set of tests to demonstrate the intended behavior and to help
avoid future regressions.

[1] bfef1f20a419d24633e48d24c14e6a7503e1d48c
[2] a5a8fbfafd075bf290525ad6a433ec5684ea874d


zcorpan (approved):
LGTM but leaving open for gsnedders

jgraham:
I don't think this should be under `assumptions/`. ISTM that those tests are for assumptions about the browser configuration and OS environment. I would create a new directory called `server` or something; you could also move the secure context tests there (same comment on the similar websockets tests).

Otherwise, this looks great; thanks for doing it.

gsnedders:
(my view is simply what @jgraham said, given @zcorpan asked for my views)

jugglinmike:
Okay, I've rebased with a new commit to create and document a directory named `infrastructure/server/`. The original version of this branch is available at https://github.com/w3c/web-platform-tests/compare/master...bocoup:re-enable-web-socket-server-tests-prebase

jgraham (approved):
I guess in general this would be clearer creating the tests in a loop rather than hardcoding everything, but this is fine.

PR #10620, opened 2018-04-25
[css-flex][spec-rec] Test for flex only alignment modes with writing modes

benfluleck:
#### What does this PR do? 
These tests ensure that flex only alignment modes  work as expected in different
writing modes
#### How should this be manually tested?
Pull the current branch
Follow the Readme.md to run the tests
*Testharness.js has been used to write the test cases

#### Screenshots (if appropriate)
screen shot 2018-04-25 at 12 02 04 am

### DIFF
--- a/css/css-flexbox/mix-writing-modes-and-flex-alignment.html
+++ b/css/css-flexbox/mix-writing-modes-and-flex-alignment.html
@@ -1,0 +1,45 @@
+<!DOCTYPE html>
+<title>flexbox | first-letter</title>
+<meta name="assert" content="This test ensures that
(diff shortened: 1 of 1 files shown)


gregwhitworth (changes requested):
Thanks for writing up a test. What needs to be covered here is ensuring the placement is correct. Currently you're testing that the computed values are correct, which in and of itself is not bad but is not what we're needing from the test. In order to accurately test this you're going to want to have numerous testcases that change the position of a box using the alignment properties and then also mix the different combinations of writing modes and check their corresponding positions are correct. Let me know if you have any questions.

PR #10627, opened 2018-04-25
[css-text] white-space and intrinsic sizing

frivoal:
Goes together with the spec changes for https://github.com/w3c/csswg-drafts/issues/2465

I am reasonably confident that the first 2 tests are supported by the spec.

I am less sure about the 3rd and the 4th, altought I believe that they are supported by the intent of the spec. The spec says:

> [...] must either hang the white space or visually collapse the character advance widths of any overflowing spaces such that they don’t take up space in the line.

For browsers that chose the first alternative ("hang the white space"), then the 3rd and 4th test should be correct.

For those that chose the second one ("visually collapse the character advance widths of any overflowing spaces such that they don’t take up space in the line"), I am less sure. Maybe they are correct, or maybe the intrinsic sizing happens first and makes room for the spaces, and then there's no need to collapse the advance width of overflowing spaces since no space overflows.

Browsers that chose that option (Safari and Chrome) are not in full agreement: they both fail test 4, but Safari passes test 3.


foolip:
@frivoal, I see that https://github.com/w3c/csswg-drafts/issues/2465 is closed. Does this just need review now?

frivoal:
Actually, the spec has changed since. So this needs an update first. I'll do that shortly.

frivoal:
I've now updated this to match the latest spec. Review appreciated (cc: @fantasai )

foolip:
Thanks @frivoal! (I'll review the review to suggested reviewers, but let me know if you need uninformed review to get it unstuck.)

frivoal:
By the way,
>For those that chose the second one ("visually collapse the character advance widths of any overflowing spaces such that they don’t take up space in the line"), I am less sure. Maybe they are correct, or maybe the intrinsic sizing happens first and makes room for the spaces, and then there's no need to collapse the advance width of overflowing spaces since no space overflows.

has been clarified in the spec, in favor of the interpretation I had taken: https://github.com/w3c/csswg-drafts/commit/e625f72b7dee716bef3be9ddf23dc173aba93118

foolip:
@frivoal do you need help getting this unstuck? I know how to skim and approve :)

PR #10633, opened 2018-04-25
Encoding: remove incorrect iso-2022-jp test

annevk:
There is a lot of code in this test that simply doesn't run and the the few bits that do run are incorrect as far as I can tell.


inexorabletash:
I want to verify that the %1B case is exercised elsewhere. But yeah, this test appears to be 99% useless otherwise.

inexorabletash (approved):
Oh, wow, and those encode() lines have bogus expectations too. The first one should expect "a" and the second should expect "" since `new URL('http://example.com/?\x1b').search === ''` anyway. So we're not even exercising the encoder meaningfully.

r12a:
I think this was uploaded by mistake, and was never a real test. (It doesn't appear at https://www.w3.org/International/tests/repo/results/encoding-dbl-byte.en#iso2022jp), so yes i agree we should jettison this file.  Tomorrow i'll double check whether there was anything else that shouldn't be there.

r12a (approved):

PR #10638, opened 2018-04-25
Fix test for Event.returnValue

Zirro:
The last test requires the event to be cancelable. This was discovered as part of https://github.com/jsdom/jsdom/pull/2197.

Two tests from this file don't pass in current browsers even with this change, but this appears to be due to changes made to the setter in the specification which will be implemented soon.


annevk (approved):
Thanks!

PR #10642, opened 2018-04-25
Additional fixes to DOMQuad-001.html

csnardi:
`DOMQuad` no longer has an explicit constructor from a `DOMRect`, this was replaced with `fromRect()` in https://github.com/w3c/fxtf-drafts/commit/62b9cb9d5be4982d2a9cbf314e3a59efb8a68dd6. Additionally, there was a test for the `bounds` attribute being read-only, which was replaced with `getBounds()`, so the test no longer makes sense. Update the tests to reflect these changes.

Also fix a typo for the bounds in `p1Top4Attributes1` as all values should be 0 since it was the original value for each.

### DIFF
--- a/css/geometry/DOMQuad-001.html
+++ b/css/geometry/DOMQuad-001.html
@@ -44,31 +44,31 @@
         },'testConstructor4');
 
         checkDOMQuad(
-                function() { return new DOMQuad(new DOMRect(10, 20, 100, 200)); },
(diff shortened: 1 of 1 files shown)


peterjoel (approved):
LGTM

csnardi:
@peterjoel Could you merge this in as well? I unfortunately don't have access to do that.

Issue #7958, opened 2017-10-23
[css-flex][spec-rec] Add test for fragmentation propogates to container


gregwhitworth:
We didn't see any tests covering this part of the spec so we should add one:

> In a column flex container, the break-before property on the first item and the break-after property on the last item are propagated to the flex container. Forced breaks on other items are applied to the item itself.

claytron5000:
relevant link: https://drafts.csswg.org/css-flexbox-1/#pagination

claytron5000:
I'm having trouble understanding what the expectation is on this specification. Specifically, the idea that the `break-after` and `break-before` properties propagate to the container. This sounds like a reverse cascade.

> In a column flex container, the break-before property on the first item and the break-after property on the last item are propagated to the flex container.

Consider the following example:

```
<div class="container">
  <div class="item"></div>
  <div class="item"></div>
  <div class="item"></div>
</div>
```
Does this mean, `.item:first-child { break-before: avoid; }` will "propagate" to the container element? So, the container will then have the property `break-before: avoid;`?

fantasai:
@claytron5000 Added some clarifications to the specs in https://github.com/w3c/csswg-drafts/issues/2614 Please let me know if you think further edits are necessary!

PR #10644, opened 2018-04-26
Add fragmentation propagation test

claytron5000:
#7958
I'm not convinced I totally understand the specification on this one.

> In a column flex container, the break-before property on the first item and the break-after property on the last item are propagated to the flex container. Forced breaks on other items are applied to the item itself. 

This fails on Chrome Canary, and I haven't seen it work in Firefox either.

### DIFF
--- a/css/css-flexbox/flexbox_fragmentation-propagation.html
+++ b/css/css-flexbox/flexbox_fragmentation-propagation.html
@@ -13,56 +13,55 @@
       height: 96px;
       border: solid black 2px;
     }
-    .container {
+    #container {
       display: flex;
       flex-direction: column;
       flex-wrap: wrap;
-      height: 300px;
+      height: 299px;
+      padding: 5px;
+      border: solid 7px #334488;
     }
-    #container-break-before .item:first-child {
-      break-before: avoid;
-    }
-    #container-break-after .item:last-child {
+    #container-break-before .item {
       break-before: always;
     }
-    #conatiner-forced-break .item:nth-child(2) {
-      break-after: page;
-    }
+
   </style>
 </head>
 <body>
-  <div id="container-break-before" class="container">
+  <div id="container-break-before">
     <div class="item">1</div>
     <div class="item">2</div>
     <div class="item">3</div>
   </div>
-  <div id="container-break-after" class="container">
-    <div class="item">1</div>
-    <div class="item">2</div>
-    <div class="item">3</div>
-  </div>
-  <div id="conatiner-forced-break" class="container">
-    <div class="item">1</div>
-    <div class="item">2</div>
-    <div class="item">3</div>
-  </div>
+
   <script src="/resources/testharness.js"></script>
   <script src="/resources/testharnessreport.js"></script>
   <script>
+    // These tests are all assuming the left to right writing mode.
     test(function() {
-      const containerStyles = getComputedStyle(document.querySelector("#container-break-before"));
-      assert_equals(containerStyles.breakBefore, 'avoid');
-    }, "Check if break-before propagated to container.")
+      const container = document.querySelector("#container-break-before");
+      const containerStyles = window.getComputedStyle(container);
+      const containerRect = container.getBoundingClientRect();
+      const firstItemRect = document.querySelector("#container-break-before .item:first-child").getBoundingClientRect();
 
-    test(function() {
-      const containerStyles = getComputedStyle(document.querySelector("#container-break-after"));
-      assert_equals(containerStyles.breakBefore, 'always');
-    }, "Check if break-after propagated to container.")
+      assert_equals(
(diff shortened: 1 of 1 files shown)


fantasai (changes requested):
OK, so this is totally wrong. :) The spec isn't saying that the computed values are passed up; it's talking about a different, abstract mechanism which is unrelated to computed values. What it's doing is making sure that a forced break-before or break-after doesn't create a break between a parent and its first/last child--breaks are only supposed to only happen between siblings, otherwise it looks weird with dangling empty bits of broken borders/padding. See https://www.w3.org/TR/css-break-3/#forced-breaks

To test this, you actually have to test layout using the reftest infrastructure. IIRC WPT doesn't have pagination reftest infrastructure, though (even though Mozilla has it), so the best you can do atm is write a test that'll check fragmentation in a multicol environment. It won't be able to test combined break types, but it can check for basic propagation. :/

(Might be worth trying to make the spec wording a bit clearer it seems, so I filed https://github.com/w3c/csswg-drafts/issues/2614 )

claytron5000:
@fantasai Thanks for the clarification, I had a feeling I was off base there.

If I'm reading this right, are you saying we're checking to make sure the first child element is not breaking to the second column(for example), or broken padding/border etc?

claytron5000:
In this version I'm testing for collapsing borders/padding on the top and left side of the container. I have to think about how to test the `break-after` situation.

gregwhitworth:
@fantasai thanks for jumping in and helping out here. Multicol was the only way I could think of to potentially test this. Here is an example of this, and this is why writing the testcases is so important for interop: http://jsbin.com/mewotuzugi/edit?html,css,output

image

Reading the spec text, Chrome/Edge are doing what I would expect them to do in the regular column fragmentation scenario. Edge supports avoiding column breaks, which makes testing the above a bit easier as you'll want the first item to break on the column: http://output.jsbin.com/feyibulefe/1

image

Based on the initial auto values and the way other fragmentation is occurring across pages, etc, I believe Edge is doing the intended behavior - please correct me if I'm wrong here @fantasai.

If this is indeed the case, then you could build a reftest that mimics this but you could also try using `getBoundingClientRects` to ensure its position and size.

fantasai:
@gregwhitworth As I mentioned on IRC, you cannot test this case by having the flex container be the first child of the multicol: a forced break there will have no effect because it's already at the top of the first column. You need to put content before the flex container to see if a break is taken between that content and the flex container.

The goal of this clause in the spec is to prevent any breaks between the start of the flex container and the first item, even if the first item requests a forced break before: that forced break gets propagated to the flex container, moving the flex item to the top of the next column by pushing the entire flex container over.

Edge is definitely getting it wrong (I'm assuming it's the black UA)--it's putting a break between the top content edge of the flex container and the top margin of the flex item, and that's not allowed.

gregwhitworth:
@fantasai thank you so much for the clarification.

@claytron5000 Based on the above, do you feel like you have enough info to give a test a go? If not - I can tackle it when I get a chance. Let me know.

claytron5000:
> The goal of this clause in the spec is to prevent any breaks between the start of the flex container and the first item, even if the first item [*in the flex container] requests a forced break before: that forced break gets propagated to the flex container, moving the flex item to the top of the next column by pushing the entire flex container over.

I'm starting to understand what's going on. I added one clarifying phrase above in brackets.

In this example I've added a "firsty" element that in Firefox bumps the entire container into the second column (unexpected/unwanted behavior), but in Chrome, does not bump it over. (I don't have Edge on this computer).
https://codepen.io/claytron5000/pen/BxQddb?editors=1100

FireFox:
image
Chrome:
image

Note: There seems to be a secondary behavior irregularity illustrated in @gregwhitworth 's bins that is independent(though maybe technically related) of this issue where Firefox doesn't push the `break-after` items to the second column.

gregwhitworth:
I don't think that we can trust what Firefox is doing here as they don't seem to ever actually fragment a column direction flex container (but I do agree with your premise). Here is a side-by-side of Chrome & Edge in your test (make sure to remove `break-before: column` as Edge supports this and will incorrectly "pass" your test, but in this screenshot it is removed):

image

Based on my reading of the spec and @fantasai clarification, I think that Firefox's end result is what is expected, but based on the general case I think they're doing that solely because the entirety of the flex container doesn't fit so they move it over. So I think it should look like this (according to the spec - this is using block layout [not flex] and can be the basis for the ref tests):

image

Here is the codepen of that: https://codepen.io/gregwhitworth/pen/MGbrpa

fantasai:
Also, and this is totally not required if you don't wanna, but, can we use somewhat less garish colors for the testcase? :) I tend to use orange/blue/aqua for things that are important to pay attention to in the testcase and silver/gray/black for incidentals. ^_^

PR #10645, opened 2018-04-26
 Update webidl2.js to v10.2.1

csnardi:
This was created by:

```
git revert 280005efe04503f14891d73c6ceab1077a59a735
git subtree pull --prefix=resources/webidl2 https://github.com/w3c/webidl2.js.git v10.2.1 --squash
```

and then making the merge commit have a reasonable commit message. @gsnedders does this look right?

Main motivation for this update is pulling in https://github.com/w3c/webidl2.js/pull/139, which has the potential to result in performance improvements for IDL tests.

Note that 280005efe04503f14891d73c6ceab1077a59a735 was reverted as upstream changes in https://github.com/w3c/webidl2.js/pull/139 should have made effectively the same change (and created a merge conflict); the fix was reported to work in Servo.

This **should not be merged** as of yet -- I think I did this right per https://github.com/w3c/web-platform-tests/pull/8063#issuecomment-341887244, but I need someone to sanity check me first so it doesn't mess anything up. This also cannot be squashed & rebased, it needs a pure merge.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10645)


csnardi:
I believe the other failures in `resources_unittest` might be preexisting, they all seem to have to do with promise tests.

csnardi:
No tests seem to have been run on Firefox/Chrome, I can put up a test CL on Chromium in the morning to see if this would break anything.

csnardi:
Running this on Chromium CI shows 1 regression in css-typed-om/interfaces.html (https://test-results.appspot.com/data/layout_results/linux_chromium_rel_ng/79219/layout-test-results/results.html):

```
FAIL Testing Symbol.iterator property of iterable interface CSSUnparsedValue assert_equals: @@iterator function does not have the right name expected "entries" but got "values"
FAIL Testing pair iterable interface CSSUnparsedValue assert_equals: entries method is not the same as @@iterator expected function "function entries() { [native code] }" but got function "function values() { [native code] }"
```

with that repeated twice more for other properties. Not sure if this was expected or not.

gsnedders:
@csnardi this LGTM now (and with #10661 merged in it passes resources/test); do you want to do another CL run before landing or just trust that the latest fixup did fix the sole regression?

csnardi:
@gsnedders I tested locally and I got the same results with the latest fixup as I did with https://w3c-test.org/css/css-typed-om/interfaces.html. So, I think this might be ready for merging?

I did realize that the CL didn't test with the newest version of idlharness.js (the last time Chrome pulled in a new version was ~8 weeks ago), so I could test with that or we could just merge this in and assume it should work.

csnardi:
Actually, nevermind, strike my last paragraph. I think we should be good then.

csnardi:
@nox @jdm This shouldn't break idlharness for Servo, but making you aware of this in case it actually does.

nox:
Thanks for the heads up!

Issue #10652, opened 2018-04-26
resources_unittests fails intermittently


Ms2ger:
Output:

```
+cd /home/travis/build/w3c/web-platform-tests/resources/test
+tox -- --binary=/home/travis/browsers/firefox/firefox
py27 create: /home/travis/build/w3c/web-platform-tests/resources/test/.tox/py27
py27 installdeps: html5lib, pytest>=2.9, pyvirtualdisplay, selenium, requests
py27 installed: attrs==17.4.0,certifi==2018.4.16,chardet==3.0.4,EasyProcess==0.2.3,funcsigs==1.0.2,html5lib==1.0.1,idna==2.6,more-itertools==4.1.0,pluggy==0.6.0,py==1.5.3,pytest==3.5.1,PyVirtualDisplay==0.2.1,requests==2.18.4,selenium==3.11.0,six==1.11.0,urllib3==1.22,webencodings==0.5.1
py27 runtests: PYTHONHASHSEED='2687066097'
py27 runtests: commands[0] | pytest --binary=/home/travis/browsers/firefox/firefox -vv tests
============================= test session starts ==============================
platform linux2 -- Python 2.7.14, pytest-3.5.1, py-1.5.3, pluggy-0.6.0 -- /home/travis/build/w3c/web-platform-tests/resources/test/.tox/py27/bin/python2.7
cachedir: .pytest_cache
rootdir: /home/travis/build/w3c/web-platform-tests/resources/test, inifile:
collected 39 items                                                             
::Test#add_cleanup <- . PASSED                                           [  2%]
::Test#add_cleanup reported count <- . PASSED                            [  5%]
::Test#add_cleanup: error <- . PASSED                                    [  7%]
::Test#add_cleanup: multiple functions with one in error <- . PASSED     [ 10%]
::Sample HTML5 API Tests <- . PASSED                                     [ 12%]
::Sample HTML5 API Tests <- . PASSED                                     [ 12%]
::Sample HTML5 API Tests <- . PASSED                                     [ 12%]
::Test#force_timeout <- . PASSED                                         [ 15%]
::Sample for using generate_tests to create a series of tests that share the same callback. <- . PASSED [ 17%]
::Example with iframe that notifies containing document via callbacks <- . PASSED [ 20%]
::Example with iframe that consolidates errors via fetch_tests_from_window <- . PASSED [ 23%]
::Example with iframe that consolidates tests via fetch_tests_from_window <- . FAILED [ 25%]
[...]

PR #10660, opened 2018-04-26
Update expectation for promise resource tests

csnardi:
5 tests were expecting a non-null stack but did not receive one. Update the expectations to match current behavior.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10660)


gsnedders:
This is #10652. Something kinda odd is up given it's only some of the tests are failing with null stacks. That said, given `Error.stack` is well, well into undefined territory (and implementations sometimes vary based on things like "has this function been compiled"), we should just stop testing anything related to it, which I'm very close to having a working PR for, thus let's just close this.

csnardi:
Ah yep, probably should have looked at issues before uploading this.

PR #10665, opened 2018-04-26
Increase taskcluster test chunks and timeouts.

jgraham:
Tests seem to be a little slower than in mozilla-central so increase the number of chunks
from 6->10 for reftests and 10->15 for testharness tests. Also make the task timeout 2 hours
rather than 1.5.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10665)


Hexcles (approved):
LGTM

PR #10679, opened 2018-04-27
HTML: document.open() and quirks mode

annevk:


foolip:
w3c-test:mirror

zcorpan (approved):

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/quirks.window.js
```diff
@@ -0,0 +1,69 @@
...
+  assert_equals(frame.contentDocument.compatMode, "CSS1Compat");
+  frame.contentDocument.write("<!doctype html");
+  assert_equals(frame.contentDocument.compatMode, "CSS1Compat");
+  frame.contentDocument.write(">");
+  assert_equals(frame.contentDocument.compatMode, "CSS1Compat");
+  frame.contentDocument.close();
+  assert_equals(frame.contentDocument.compatMode, "CSS1Compat");
+}, "document.open() and quirks mode (iframe, new doctype)");
+
+// Test derived from quirks/blocks-ignore-line-height.html.
```
foolip: There's enough here to figure out that the test isn't really about the layout size of those divs, but can you spell out that this is all because there's no way to directly test if one is in limited quirks mode, and that this test is only valid on the assumption that the one linked here passes?

foolip (approved):
LGTM % comment

zcorpan (approved):
LGTM % foolip's comment

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/quirks.window.js
TimothyGu: Done.

domenic (approved):

PR #10680, opened 2018-04-27
HTML: document.open() and mutation events

annevk:
See https://github.com/whatwg/html/pull/3598.


domenic (commented):
Giving the unreached_funcs names would help a lot here. Firefox says "reaches unreachable code" on http://w3c-test.org/submissions/10680/html/dom/dynamic-markup-insertion/opening-the-input-stream/mutation-events.window.html

Edge and Chrome pass at least.

annevk:
Interesting, it does not for me. Are you using Nightly?

domenic (approved):
No, stable. Confirmed nightly passes. I still think adding descriptions to the unreached_funcs is good, but let me flip this to approved.

PR #10686, opened 2018-04-27
HTML: document.open() event listener removal

annevk:
For https://github.com/whatwg/html/pull/3653.

### DIFF
--- a/html/dom/dynamic-markup-insertion/opening-the-input-stream/event-listeners.window.js
+++ b/html/dom/dynamic-markup-insertion/opening-the-input-stream/event-listeners.window.js
@@ -2,12 +2,12 @@ test(t => {
   const frame = document.body.appendChild(document.createElement("iframe")),
         body = frame.contentDocument.body;
   t.add_cleanup(() => frame.remove());
-  frame.contentDocument.addEventListener("x", t.unreached_func());
-  body.addEventListener("x", t.unreached_func());
+  frame.contentDocument.addEventListener("x", t.unreached_func("document event listener not removed"));
+  body.addEventListener("x", t.unreached_func("body event listener not removed"));
   frame.contentDocument.open();
-  frame.contentDocument.close();
   frame.contentDocument.dispatchEvent(new Event("x"));
   body.dispatchEvent(new Event("x"));
+  frame.contentDocument.close();
 }, "Event listeners are to be removed");
 
 test(t => {
@@ -16,12 +16,12 @@ test(t => {
   let once = false;
   frame.contentDocument.addEventListener("x", () => {
     frame.contentDocument.open();
-    frame.contentDocument.close();
     once = true;
   });
-  frame.contentDocument.addEventListener("x", t.unreached_func());
+  frame.contentDocument.addEventListener("x", t.unreached_func("second event listener not removed"));
   frame.contentDocument.dispatchEvent(new Event("x"));
   assert_true(once);
+  frame.contentDocument.close();
 }, "Event listeners are to be removed with immediate effect");
 
 test(t => {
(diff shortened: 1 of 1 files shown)


domenic:
w3c-test:mirror

html/dom/dynamic-markup-insertion/opening-the-input-stream/event-listeners.window.js
```diff
@@ -0,0 +1,42 @@
+test(t => {
+  const frame = document.body.appendChild(document.createElement("iframe")),
+        body = frame.contentDocument.body;
+  t.add_cleanup(() => frame.remove());
+  frame.contentDocument.addEventListener("x", t.unreached_func());
+  body.addEventListener("x", t.unreached_func());
+  frame.contentDocument.open();
+  frame.contentDocument.close();
```
domenic: The "close()" here seems not great since if someone implemented close() as removing the event listeners the test would still pass.

html/dom/dynamic-markup-insertion/opening-the-input-stream/event-listeners.window.js
```diff
@@ -0,0 +1,42 @@
+test(t => {
+  const frame = document.body.appendChild(document.createElement("iframe")),
+        body = frame.contentDocument.body;
+  t.add_cleanup(() => frame.remove());
+  frame.contentDocument.addEventListener("x", t.unreached_func());
+  body.addEventListener("x", t.unreached_func());
```
domenic: Adding descriptions to unreached_funcs makes test failures much easier to diagnose.

html/dom/dynamic-markup-insertion/opening-the-input-stream/event-listeners.window.js
```diff
@@ -0,0 +1,42 @@
...
+    frame.contentDocument.open();
+    frame.contentDocument.close();
+    once = true;
+  });
+  frame.contentDocument.addEventListener("x", t.unreached_func());
+  frame.contentDocument.dispatchEvent(new Event("x"));
+  assert_true(once);
+}, "Event listeners are to be removed with immediate effect");
+
+test(t => {
```
domenic: Might be worth gating this behind an `if (document.body.attachShadow)` so browsers without shadow DOM don't get failures.

html/dom/dynamic-markup-insertion/opening-the-input-stream/event-listeners.window.js
```diff
@@ -0,0 +1,42 @@
...
+  body.dispatchEvent(new Event("x"));
+}, "Event listeners are to be removed");
+
+test(t => {
+  const frame = document.body.appendChild(document.createElement("iframe"));
+  t.add_cleanup(() => frame.remove());
+  let once = false;
+  frame.contentDocument.addEventListener("x", () => {
+    frame.contentDocument.open();
+    frame.contentDocument.close();
```
domenic: Same concern about `close()`

domenic (commented):
Concerned about the extra close()s.

Looks like a great area of non-interop you've discovered.

- Chrome passes
- Safari and Edge fail the second test
- Firefox Nightly fails the first test
- Edge fails the first test with one of its "permission denied" lovely errors.

html/dom/dynamic-markup-insertion/opening-the-input-stream/event-listeners.window.js
annevk: I think that's bad practice. It makes it less clear what's going on. And have a varying number of tests between browsers, if that's what you're suggesting, seems even worse for tooling.

annevk:
Bugs:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1458476
* https://bugs.webkit.org/show_bug.cgi?id=185191
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/17338802/

domenic (approved):
This seems like it fixes my problems with close(), but I still don't understand why you're including close() at all? Does the test fail without it or something?

annevk:
Depending on the browser the load event might not fire and therefore the harness will timeout. I'm planning on investigating that separately in #10239. (Made progress on some other bits today though.)

PR #10691, opened 2018-04-27
Complete the migration from marionette-specific scripts.

jgraham:
Previously we removed the wrappedJSObject from
executor_marionette.js. However it turns out that this migration was
incomplete; in order to fully avoid the wrappedJSObject, we need to
ensure that we never create a sandbox on the marionette side. As a
result of this omission we were getting an error trying to call the
timeout() function, causing more EXTERNAL-TIMEOUTs and so performance
issues.

The solution is to use the same approach as geckodriver, which isn't
exposed through the js marionette client. Essentially we have to
ensure that the sandbox is None for all calls to (content) js. Since
this isn't exposed in the client we construct the message directly and
ensure that all calls to content js co through the base protocolpart
rather than calling marionette.execute_* directly.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10691)


tools/wptrunner/wptrunner/executors/executormarionette.py
```diff
@@ -59,8 +59,16 @@ def setup(self):
...
+        body = {
+            "script": script.strip(),
+            "args": [],
+            "newSandbox": False,
+            "specialPowers": None,
+            "scriptTimeout": None
+        }
+        method = "WebDriver:ExecuteAsyncScript" if async else "WebDriver:ExecuteScript"
+        rv = self.marionette._send_message(method, body, key="value")
+        return self.marionette._from_json(rv)
```
whimboo: I don't know all this code at all, so maybe I'm not the best reviewer. But I want to state that using the private message `_send_message` from Marionette isn't what I want to see. Can you please explain why this is necessary? 

Also note that a script timeout of None will result in a hang if something goes wrong in a call to an async script.

jgraham: `marionette.py` doesn't expose the possibility to not set a sandbox. This is replicating the message that's set in geckodriver and it's tested in-tree in gecko, so I'm reasonably confident it won't break.

whimboo: @andreastt I would refer to you here. I don't like the `_send_message` usage at all, but maybe you have a better picture for that in regards of wptrunner.

andreastt: Wouldn’t `method(script, sandbox=None, new_sandbox=False)` work?

The problem we’re trying to circumvent here is that the Python client uses the string `default` as the default value for the `sandbox` kwarg, but if this is explicitly set to `None` it will pass `{"sandbox": null, …}` to Marionette.

gsnedders:
This apparently fixes the `JavaScript error: executormarionette.py, line 84: TypeError: window.win.timeout is not a function` messages `wpt run` currently produces. (Adding this so I can find the issue again.)

andreastt (approved):

PR #10701, opened 2018-04-28
Add two screen-orientation tests and replace async by promise_test

xiuqijix:
- Add test that screen.orientation.unlock() returns a void value.
- Add test that screen.orientation.lock returns a promise which will be fulfilled with a void value.
- Add "" and true types in test (screen.orientation.lock() must throw given invalid input).
- Update test (Test screen.orientation default values).
- Replace async by promise_test.

### DIFF
--- a/screen-orientation/lock-basic.html
+++ b/screen-orientation/lock-basic.html
@@ -19,6 +19,7 @@ promise_test(async t => {
 promise_test(async t => {
   const orientations = [
     'any',
+    'natural',
     'portrait',
     'landscape',
     'portrait-secondary',
@@ -27,25 +28,36 @@ promise_test(async t => {
     'portrait-primary',
   ];
   for (const orientation of orientations) {
-    const returnType = screen.orientation.lock(orientation);
-    assert_true(returnType instanceof Promise);
-    await returnType;
+    const promiseToChange = screen.orientation.lock(orientation);
+    assert_true(promiseToChange instanceof Promise, "Expected an instance of Promise");
+    await promiseToChange;
+    const type = screen.orientation.type;
+    switch (orientation) {
+    case 'any':
+      break;
+    case 'natural':
+      assert_true(type == "portrait-primary" || type == "landscape-primary");
+      break;
+    case 'portrait':
+      assert_true(type == "portrait-primary" || type == "portrait-secondary");
+      break;
+    case 'landscape':
+      assert_true(type == "landscape-primary" || type == "landscape-secondary");
+      break;
+    default:
+      assert_equals(type, orientation, "Expected orientation to change");
+      break;
+    }
   }
   screen.orientation.unlock();
 }, "Test that screen.orientation.lock returns a pending promise.");
 
 promise_test(async t => {
   const preType = screen.orientation.type;
-  let newType = "";
-
-  if (preType.indexOf("portrait") != -1) {
-    newType = "landscape-primary";
-  } else {
-    newType = "portrait-primary";
-  }
-
+  const isPortrait = preType.includes("portrait");
+  const newType = `${ isPortrait ? "portrait" : "landscape" }-primary`;
   const p = screen.orientation.lock(newType);
-  assert_equals(screen.orientation.type, preType);
+  assert_equals(screen.orientation.type, preType, "Must not change orientation until next spin of event loop");
   await p;
   assert_equals(screen.orientation.type, newType);
 }, "Test that screen.orientation.lock() is actually async");
--- a/screen-orientation/lock-basic.html
+++ b/screen-orientation/lock-basic.html
@@ -1,41 +1,43 @@
 <!DOCTYPE html>
-<html>
-<body>
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <script>
-
-var previousOrientation = screen.orientation;
-
-test(function() {
-    screen.orientation.unlock();
+test(() => {
+  screen.orientation.unlock();
 }, "Test that screen.orientation.unlock() doesn't throw when there is no lock");
 
-async_test(function(t) {
-    var orientations = ['any', 'portrait', 'landscape', 'portrait-secondary',
-                        'landscape-primary', 'landscape-secondary', 'portrait-primary'];
+test(() => {
+  const value = screen.orientation.unlock();
+  assert_equals(value, undefined);
+}, "Test that screen.orientation.unlock() returns a void value");
 
-    setOrientation = function(idx) {
-        if( idx == orientations.length) {
-            screen.orientation.unlock();
-            t.done();
-            return;
-        }
-        (function(i) {
-            screen.orientation.lock(orientations[i]).then(function() {
-                setOrientation(i+1);
-            },function() {});
-        })(idx);
-    };
+promise_test(t => {
+  return screen.orientation.lock('any').then(value => {
+    assert_equals(value, undefined);
+  });
+}, "Test that screen.orientation.lock returns a promise which will be fulfilled with a void value.");
 
-    setOrientation(0);
-
+promise_test(t => {
+  const orientations = [
+    'any',
+    'portrait',
+    'landscape',
+    'portrait-secondary',
+    'landscape-primary',
+    'landscape-secondary',
+    'portrait-primary',
+  ];
+  const promises = orientations.map(type => {
+    const returnType = screen.orientation.lock(type);
+    assert_true(returnType instanceof Promise);
+  });
+  return Promise.all(promises).finally(() => {
+    screen.orientation.unlock();
+  });
 }, "Test that screen.orientation.lock returns a pending promise.");
 
-test(function() {
-    assert_equals(screen.orientation, previousOrientation);
+test(() => {
+  const previousOrientation = screen.orientation;
+  assert_equals(screen.orientation, previousOrientation);
 }, "Test that screen.orientation.lock() is actually async");
-
 </script>
-</body>
-</html>
--- a/screen-orientation/lock-basic.html
+++ b/screen-orientation/lock-basic.html
@@ -55,7 +55,7 @@ promise_test(async t => {
 promise_test(async t => {
   const preType = screen.orientation.type;
   const isPortrait = preType.includes("portrait");
-  const newType = `${ isPortrait ? "portrait" : "landscape" }-primary`;
+  const newType = `${ isPortrait ? "landscape" : "portrait" }-primary`;
   const p = screen.orientation.lock(newType);
   assert_equals(screen.orientation.type, preType, "Must not change orientation until next spin of event loop");
   await p;
--- a/screen-orientation/lock-basic.html
+++ b/screen-orientation/lock-basic.html
@@ -11,13 +11,12 @@ test(() => {
   assert_equals(value, undefined);
 }, "Test that screen.orientation.unlock() returns a void value");
 
-promise_test(t => {
-  return screen.orientation.lock('any').then(value => {
-    assert_equals(value, undefined);
-  });
+promise_test(async t => {
+  const value = await screen.orientation.lock('any');
+  assert_equals(value, undefined);
 }, "Test that screen.orientation.lock returns a promise which will be fulfilled with a void value.");
 
-promise_test(t => {
+promise_test(async t => {
   const orientations = [
     'any',
     'portrait',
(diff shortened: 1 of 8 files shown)


marcoscaceres (changes requested):
I know the original tests were not great (so, the feedback goes beyond your refactor)... but there is an opportunity here for us to improve the tests significantly.

screen-orientation/lock-basic.html
```diff
@@ -27,17 +26,27 @@
...
-  const promises = orientations.map(type => {
-    const returnType = screen.orientation.lock(type);
+  for (const orientation of orientations) {
+    const returnType = screen.orientation.lock(orientation);
     assert_true(returnType instanceof Promise);
-  });
-  return Promise.all(promises).finally(() => {
-    screen.orientation.unlock();
-  });
+    await returnType;
```
marcoscaceres: I feel like we should assert there result here:

```JS
const promiseToChange = screen.orientation.lock(orientation);
assert_true(promiseToChange instanceof Promise, "Expected an instance of Promise");
const resultingOrientation = await promiseToChange;
assert_equals(resultingOrientation, orientation, "Expected orientation to change");
```

screen-orientation/lock-basic.html
```diff
@@ -27,17 +26,27 @@
...
+  }
+  screen.orientation.unlock();
 }, "Test that screen.orientation.lock returns a pending promise.");
 
-test(() => {
-  const previousOrientation = screen.orientation;
-  assert_equals(screen.orientation, previousOrientation);
+promise_test(async t => {
+  const preType = screen.orientation.type;
+  let newType = "";
```
marcoscaceres: Nit - the `preType.indexOf("portrait") != -1)` is confusing, maybe: 
 
```JS
const isPortrait = screen.orientation.type.includes("portrait");
const newType = `${ isPortrait ? "portrait" : "landscape" }-primary`;
```

screen-orientation/lock-basic.html
```diff
@@ -27,17 +26,27 @@
...
+
+  if (preType.indexOf("portrait") != -1) {
+    newType = "landscape-primary";
+  } else {
+    newType = "portrait-primary";
+  }
+
+  const p = screen.orientation.lock(newType);
+  assert_equals(screen.orientation.type, preType);
```
marcoscaceres: Nit: good to explain what is being tested here.
```JS
assert_equals(screen.orientation.type, preType, "Must not change orientation until next spin of event loop");
```

marcoscaceres (changes requested):
Couple of tiny things! nearly there!

screen-orientation/lock-basic.html
kaixinjxq: OK, I suppose you meant checking orientation type in last assertion.

screen-orientation/lock-basic.html
kaixinjxq: Updated.

screen-orientation/lock-basic.html
kaixinjxq: Updated.

screen-orientation/onchange-event-subframe.html
```diff
@@ -1,46 +1,31 @@
...
+    'portrait-secondary',
+    'landscape-primary',
+    'landscape-secondary'
+  ];
+  const messageWatcher = new EventWatcher(t, window, "message");
+
+  let currentIndex = orientations.indexOf(screen.orientation.type);
+  let nextIndex;
+  let message;
+  for (let i = 0; i < orientations.length; i++) {
```
marcoscaceres: Nit: for ... of

screen-orientation/onchange-event-subframe.html
```diff
@@ -1,46 +1,31 @@
...
+promise_test(async t => {
+  const orientations = [
+    'portrait-primary',
+    'portrait-secondary',
+    'landscape-primary',
+    'landscape-secondary'
+  ];
+  const messageWatcher = new EventWatcher(t, window, "message");
+
+  let currentIndex = orientations.indexOf(screen.orientation.type);
```
marcoscaceres: Nit, all these can be in the for loop.

screen-orientation/onchange-event.html
```diff
@@ -1,80 +1,33 @@
...
+    'portrait-secondary',
+    'landscape-primary',
+    'landscape-secondary'
+  ];
+  const orientationWatcher = new EventWatcher(t, screen.orientation, 'change');
+
+  let currentIndex = orientations.indexOf(screen.orientation.type);
+  let nextIndex;
+  for (let i = 0; i < orientations.length; i++) {
+    nextIndex = (currentIndex + 1) % orientations.length;
```
marcoscaceres: As above. This could be cleaned up.

screen-orientation/onchange-event.html
```diff
@@ -1,80 +1,33 @@
...
-});
-
-function runNoChangeTest() {
-  screen.orientation.lock(orientations[currentIndex]).then(function() {}, function() {});
-
-  noChangeTest.step(function() {
-    assert_equals(screen.orientation.type, orientations[currentIndex]);
-    assert_equals(orientationChangeEventHandlerCalls, orientations.length);
-    assert_equals(orientationChangeEventListenerCalls, orientations.length);
+promise_test(t => {
```
marcoscaceres: Nit: async/await

marcoscaceres (changes requested):
Spotted a few new things, but this is such a massive improvement 😍

screen-orientation/onchange-event-subframe.html
kaixinjxq: Updated.

screen-orientation/onchange-event-subframe.html
kaixinjxq: Updated.

screen-orientation/onchange-event.html
kaixinjxq: Updated.

screen-orientation/onchange-event.html
kaixinjxq: Updated.

xiuqijix (commented):
@marcoscaceres, many thanks for your review and I have addressed your comments in new commit, PTAL.

marcoscaceres (approved):
Approving, but need clarification on the setTimeout. If we can get rid of it, then I'm happy for this to be merged.

screen-orientation/onchange-event.html
```diff
@@ -1,80 +1,37 @@
...
+    screen.orientation.onchange = () => { 
+      reject(new Error("change event should not be fired"));
     }
-  };
-}
-
-runChangeTest();
+    await screen.orientation.lock(type);
+    assert_equals(screen.orientation.type, type);
+    setTimeout(resolve, 1000);
```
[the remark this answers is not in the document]
kaixinjxq: Without a timeout, the Promise will resolve immediately once the above assertion is completed, even though the `onchange` is fired.
 I didn't find a good way more than using a `setTimeout` :(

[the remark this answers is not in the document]
marcoscaceres: but that's ok, I think: 

1. if `await screen.orientation.lock(type);` invokes the event handler, then we know the test has failed. 
1. resolve();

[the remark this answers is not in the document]
marcoscaceres: (i.e., we can safely drop setTimeout())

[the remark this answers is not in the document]
kaixinjxq: I've tested, if drop `setTimeout()`, the test will still pass even the event is fired. That means the `resolve();` will be invoked before event handler.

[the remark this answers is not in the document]
marcoscaceres: Interesting! Which browser is that in?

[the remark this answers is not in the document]
kaixinjxq: Chrome 68 on Android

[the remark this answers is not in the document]
marcoscaceres: @mounirlamouri is that a bug in Chrome? ☝️☝️☝️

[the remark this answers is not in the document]
mounirlamouri: We queue the event when we dispatch it so I guess it ends up being asynchronous while the promise may not be. It might be what the spec used to say. I can't remember :)

If this is passing in another browser, I would suggest to drop the `setTimeout()` and file a bug against Chrome. If it is not, I would suggest to change `setTimeout(resolve, 100);` to `setTimeout(resolve);` and add a test specifically for the race between event and promise and have Chrome fail it.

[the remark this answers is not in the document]
kaixinjxq: Here's an another example to use timeout to detect a misfire event https://github.com/w3c/web-platform-tests/blob/master/mediacapture-streams/MediaStream-idl.https.html#L52.

It's the same result in Fifefox 60 Android, will use `setTimeout(resolve)` in new commit.

>  add a test specifically for the race between event and promise and have Chrome fail it.

I'm not quite understand, is this new test specific for testing race between event and promise, and irrelevant to screen orientation?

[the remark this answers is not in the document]
mounirlamouri: I meant to test the race between both in the context of the orientation change event.

screen-orientation/onchange-event.html
```diff
@@ -1,80 +1,37 @@
...
+    screen.orientation.onchange = () => {
+      reject(new Error("change event should not be fired"));
     }
-  };
-}
-
-runChangeTest();
+    await screen.orientation.lock(type);
+    assert_equals(screen.orientation.type, type);
+    setTimeout(resolve, 100);
```
marcoscaceres: so, from a spec compliance perspective, I hold we should `resolve()` here. The even firing randomly later in time is definitely a Chrome bug.

screen-orientation/onchange-event.html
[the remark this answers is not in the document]
marcoscaceres: I think we should just follow the spec. Just test that the event doesn't immediately fire. Events randomly firing at strange times is not something we can reliably test... consider, we set timeout for 1000 and the event fires at 1001.

kaixinjxq:
@mounirlamouri, @marcoscaceres, friendly ping. :)
This one is going to be conflict with #11351

screen-orientation/onchange-event.html
```diff
@@ -15,7 +15,7 @@
       reject(e);
     }
     assert_equals(screen.orientation.type, type);
-    setTimeout(resolve);
```
marcoscaceres: 👍

marcoscaceres (approved):
I think we should merge these so we have them, we can then deal with any bugs the arise.

screen-orientation/onchange-event.html
```diff
@@ -6,10 +6,14 @@
   const type = screen.orientation.type;
   // wrap onchange listener in new promise to catch unreached assertion
   return new Promise(async (resolve, reject) => {
-    screen.orientation.onchange = () => {
-      reject(new Error("change event should not be fired"));
+    try {
```
zcorpan: This is not necessary with `promise_test`, right?

kaixinjxq: `promise_test` could not catch error in `new Promise()`, that's why test timeout if `orientation.lock` is unsupported, with `try..catch` we can reject the error to `promise_test`, you reminded me that the assertion should also be enclosed in `try...catch`

zcorpan: It...should work. This fails the test for me:
```
<!doctype html>
<script src=/resources/testharness.js></script>
<script src=/resources/testharnessreport.js></script>
<script>
  promise_test(() => {
    return new Promise(resolve => {
      something.something.something;
    })
  })
</script>
```

zcorpan: Oh I see, it's because of `await`. Hmm....

zcorpan: You can do 
```
await t.step(() => screen.orientation.lock(type));
```

Ms2ger: Would this work?

```js
promise_test(async t => {
  const type = screen.orientation.type;
  screen.orientation.onchange = t.unreached_func("change event should not be fired");
  // wrap onchange listener in new promise to catch unreached assertion
  await screen.orientation.lock(type);
  assert_equals(screen.orientation.type, type);
  await new Promise(resolve => setTimeout(resolve));
}, "Test that orientationchange event is not fired when the orientation does not change.");
```

zcorpan: @kaixinjxq can you try @Ms2ger's suggestion?

kaixinjxq: Good suggestion, it's simplified and effective, but it's back to the discussion about using `SetTimeout` at https://github.com/web-platform-tests/wpt/pull/10701#discussion_r188504124, @marcoscaceres

Ms2ger: My suggestion without knowing the API at all would be to keep the zero-timeout `setTimeout` call (though it probably should use `t.step_timeout()` instead, and remove the entry from `lint.whitelist`).

If people are concerned about that, though, just drop the `await new Promise(resolve => setTimeout(resolve));` line and keep the other changes? The current code is very hard to read.

screen-orientation/onchange-event.html
```diff
@@ -1,80 +1,41 @@
...
-      changeTest.done();
-      runNoChangeTest();
-    } else {
-      runChangeTest();
+promise_test(t => {
+  const type = screen.orientation.type;
+  // wrap onchange listener in new promise to catch unreached assertion
+  return new Promise(async (resolve, reject) => {
+    try {
+      screen.orientation.onchange = () => {
```
Ms2ger: Please update with my suggestion (or something similar), with or without the timeout.

Ms2ger (changes requested):

screen-orientation/onchange-event.html
kaixinjxq: @Ms2ger, thanks for your suggestion, I've addressed it without timeout, PTAL.

Ms2ger (approved):
No more complaints from me :)

zcorpan (approved):

PR #10705, opened 2018-04-29
CSS-Flexbox ::first-letter ref test

anneke:
My first ref-test!
This PR may be ignored as it is a first excercise in writing tests. 

**What does this test?**
This test asserts that use of the CSS pseudo-selector `::first-letter` will not trigger a new formatting context when given a `align-self: flex-end` property.

**How is it tested?**
I've created a reftest with html/css.


gregwhitworth (changes requested):

gregwhitworth:
Great work on this @anneke

PR #10710, opened 2018-04-30
Fix invalid new Document() test

domenic:
Per the spec, createElement() on new Document() documents should not create HTML elements, so the URL decoding behavior tested here is incorrect.

### DIFF
--- a/dom/nodes/Document-constructor.html
+++ b/dom/nodes/Document-constructor.html
@@ -48,8 +48,6 @@ test(function() {
 test(function() {
   var doc = new Document();
   var a = doc.createElement("a");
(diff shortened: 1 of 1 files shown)


annevk (changes requested):
Why is this test wrong? The document encoding is UTF-8, so this should work, no?

annevk:
I see. Wouldn't it be better to use `createElementNS` then?

domenic:
Maybe; that seems like a reasonable separate test. I'm mostly trying to fix the broken test.

annevk:
I don't quite understand how replacing it with another test represents a fix.

annevk:
Sure, created #10712.

PR #10712, opened 2018-04-30
Correct new Document() URL test

annevk:
Supersedes #10710.

### DIFF
--- a/dom/nodes/Document-constructor.html
+++ b/dom/nodes/Document-constructor.html
@@ -36,6 +36,7 @@ test(function() {
   assert_equals(doc.contentType, "application/xml");
   assert_equals(doc.origin, document.origin);
   assert_equals(doc.createElement("DIV").localName, "DIV");
+  assert_equals(doc.createElement("a").constructor, Element);
 }, "new Document(): metadata")
 
 test(function() {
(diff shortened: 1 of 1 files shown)


annevk:
This rather interestingly shows that browsers probably do not use the node document, but instead some global. File browser bugs for now I guess?

cc @Ms2ger @domenic

dom/nodes/Document-constructor.html
```diff
@@ -47,7 +48,7 @@
 
 test(function() {
   var doc = new Document();
-  var a = doc.createElement("a");
+  var a = doc.createElementNS("http://www.w3.org/1999/html", "a");
```
domenic: Wrong namespace, missing "x", causes http://w3c-test.org/submissions/10712/dom/nodes/Document-constructor.html to fail. (Discussed in IRC but recording for posterity.)

It may help this test to include

```js
assert_equals(a.constructor, HTMLAnchorElement, "Sanity check: the created element is a HTMLAnchorElement");
```

to make the failure mode clearer in this area.

annevk: Done!

domenic (approved):

PR #10715, opened 2018-04-30
Cleanup CSSOM View interfaces test

csnardi:
The CSSOM View interfaces test added the Geometry IDL as untested, but it is never needed (at least as far as I can tell, since removing it does not affect test results, but someone with more knowledge about idlharness should verify that). Removing it means that less IDL needs to be parsed, speeding up the test.

The test also adds Document as an object twice, and refers to the CSSOM spec instead of the CSSOM View spec. This change fixes both of these issues as well.

### DIFF
--- a/css/cssom-view/interfaces.html
+++ b/css/cssom-view/interfaces.html
@@ -3,7 +3,7 @@
 <!-- WARNING: These tests are preliminary and probably partly incorrect.  -->
 <title>CSSOM View automated IDL tests</title>
 <link rel="author" title="Ms2ger" href="mailto:Ms2ger@gmail.com">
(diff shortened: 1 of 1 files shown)


gsnedders:
Presumably it's there because the IDL references the DOMRect interface which is defined in Geometry? Though given it's only used as an arg/return type of operations, idlharness.js doesn't care.

csnardi:
Ah, okay, that was what it looked like, but I couldn't exactly tell from the the `add_untested_idls` section of http://web-platform-tests.org/writing-tests/idlharness.html.

Ms2ger (approved):

zcorpan:
> Presumably it's there because the IDL references the DOMRect interface which is defined in Geometry? 

Yes.

PR #10722, opened 2018-04-30
Add console.countReset manual test

domfarolino:
Tests for the pending [Console Standard PR](https://github.com/whatwg/console/pull/133) specifying `console.countReset`. This set of tests reflects the current state of that PR, which is subject to change, depending on implementer agreement. Do not merge yet.


console/console-countReset-logging-manual.html
```diff
@@ -0,0 +1,47 @@
+<!DOCTYPE html>
+<html>
```
pwnall: nit: You don't need `<html>` `<head>` or `<body>`.

pwnall (approved):
LGTM with one nit, thank you for the new test!

console/console-countReset-logging-manual.html
domfarolino: Ah, right! The reason all of the manual tests under `console/` have these unneeded components is because when I added the first one a while ago I just copied the format of some other manual tests in this repo. If you'd like I can remove these extra bits from this test (+ other manuals in this folder)?

domenic (approved):
LGTM. Also worth adding some non-manual tests for the string conversion.

domfarolino:
Right now that's on my backlog to look at: https://github.com/w3c/web-platform-tests/issues/7583

PR #10735, opened 2018-05-01
Test that non-Text nodes are not touched by normalization.

lddubeau:
This test makes sure that non-Text nodes are not touched even if they hold empty text strings. This test was motivated by a bug present in the JSDOM library.


lddubeau:
The test failure on Travis is what I reported in #10736

domenic:
w3c-test:mirror

annevk (approved):
Thanks!

PR #10742, opened 2018-05-01
[wptrunner] Re-enable return code

jugglinmike:
In a previous commit [1] the WPT CLI was modified to ignore the return
value of the `wptrunner.start` method. Although this behavior is
convenient in cases where there is no expectation data, it obscures
errors that are relevant in any context.

Specifically, the `start` method returns a non-zero value when no tests
are run, but this is a reliable indicator of an erroneous configuration.
Because the `wpt run` command is used to validate the infrastructure of
the web-platform-tests project, ignoring this case allows
mis-configurations to go unnoticed [2].

Re-enable observance of the value returned by `wptrunner.start` and
update the code generated for the TaskCluster service to ignore this
value.

[1] 5a1b0365267b540c7f4feac3e0a85e9146997fcd
[2] https://github.com/w3c/web-platform-tests/pull/10721

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10742)

### DIFF
--- a/tools/ci/ci_taskcluster.sh
+++ b/tools/ci/ci_taskcluster.sh
@@ -5,5 +5,5 @@
 if [ $1 == "firefox" ]; then
     ./wpt run firefox --log-tbpl=../artifacts/log_tbpl.log --log-tbpl-level=info --log-wptreport=../artifacts/wpt_report.json --log-mach=- --this-chunk=$3 --total-chunks=$4 --test-type=$2 -y --install-browser --no-pause --no-restart-on-unexpected --reftest-internal --install-fonts
 elif [ $1 == "chrome" ]; then
-    ./wpt run chrome --log-tbpl==../artifacts/log_tbpl.log --log-tbpl-level=info --log-wptreport=../artifacts/wpt_report.json --log-mach=- --this-chunk=$3 --total-chunks=$4 --test-type=$2 -y  --no-pause --no-restart-on-unexpected --install-fonts
+    ./wpt run chrome --log-tbpl=../artifacts/log_tbpl.log --log-tbpl-level=info --log-wptreport=../artifacts/wpt_report.json --log-mach=- --this-chunk=$3 --total-chunks=$4 --test-type=$2 -y  --no-pause --no-restart-on-unexpected --install-fonts
 fi
--- a/tools/ci/ci_taskcluster.sh
+++ b/tools/ci/ci_taskcluster.sh
@@ -3,7 +3,7 @@
 ./wpt manifest-download
 
 if [ $1 == "firefox" ]; then
-    ./wpt run firefox --log-tbpl=../artifacts/log_tbpl.log --log-tbpl-level=info --log-wptreport=../artifacts/wpt_report.json --log-mach=- --this-chunk=$3 --total-chunks=$4 --test-type=$2 -y --install-browser --no-pause --no-restart-on-unexpected --reftest-internal --install-fonts
+    ./wpt run firefox --log-tbpl=../artifacts/log_tbpl.log --log-tbpl-level=info --log-wptreport=../artifacts/wpt_report.json --log-mach=- --this-chunk=$3 --total-chunks=$4 --test-type=$2 -y --install-browser --no-pause --no-restart-on-unexpected --reftest-internal --install-fonts --no-fail-on-unexpected
 elif [ $1 == "chrome" ]; then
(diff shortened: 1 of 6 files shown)


tools/ci/taskgraph.py
```diff
@@ -27,12 +27,17 @@
...
             """>-
-            ~/start.sh &&
-            cd /home/test/web-platform-tests &&
-            git fetch {{event.head.repo.url}} &&
-            git config advice.detachedHead false &&
-            git checkout {{event.head.sha}} &&
-            %(command)s"""],
+            set -e
+            ~/start.sh
+            (
```
Hexcles: Why is a subshell needed here?

jugglinmike: That's a good question. It isn't!

Hexcles:
```
 1:29.49 TEST_START: /infrastructure/browsers/firefox/prefs.html
 1:29.50 INFO Setting pref browser.display.foreground_color ('#00FF00')
 1:39.60 pid:6627 JavaScript error: executormarionette.py, line 84: TypeError: window.win.timeout is not a function
 1:44.66 TEST_END: TIMEOUT, expected OK
```

Hmm, weird.

jugglinmike:
That is strange. I'll investigate tomorrow

jugglinmike:
It turns out that test has been timing out since [its introduction](https://github.com/w3c/web-platform-tests/pull/10406) (see [this build report, for example](https://travis-ci.org/w3c/web-platform-tests/jobs/373261174)), but we haven't been aware of it because we've been ignoring the status code. I've rebased this branch and added a separate commit to fix that bug ([the original version of this branch is available here](https://github.com/w3c/web-platform-tests/compare/master...bocoup:wpt-fail-on-error-orig)).

Hexcles (approved):
Yikes... Thanks for fixing this!

gsnedders:
Can you add a test in `tools/wpt/tests` for this? At least the "exits with non-zero when there are no matching tests" case should be easy enough.

jgraham (changes requested):

jgraham:
So I apparently forgot to press "submit" on my review. Sorry. The gist is that we should make it possible to return success when there are unexpected results without making it necessary to ignore errors entirely. I think both my original patch and this one address the problem in the wrong layer and we should fix wptrunner to work the way we want (e.g. with another command line flag).

jugglinmike:
Thanks for the feedback @jgraham. I've implemented the new feature as part of this pull request.

Recent changes in `master` introduced merge conflicts in this patch, so I had to rebase this branch in order for it to apply cleanly. [The original version of the branch is available here](https://github.com/w3c/web-platform-tests/compare/master...bocoup:wpt-fail-on-error-orig-2) in case that's of interest to any of you folks.

The text stating that the new tests "don't work on Windows for path reasons" is copied from the surrounding tests; the tests are failing for me in Windows, and I presume that this is the same issue.

tools/ci/ci_taskcluster.sh
```diff
@@ -3,7 +3,7 @@
 ./wpt manifest-download
 
 if [ $1 == "firefox" ]; then
-    ./wpt run firefox --log-tbpl=../artifacts/log_tbpl.log --log-tbpl-level=info --log-wptreport=../artifacts/wpt_report.json --log-mach=- --this-chunk=$3 --total-chunks=$4 --test-type=$2 -y --install-browser --no-pause --no-restart-on-unexpected --reftest-internal --install-fonts
+    ./wpt run firefox --log-tbpl=../artifacts/log_tbpl.log --log-tbpl-level=info --log-wptreport=../artifacts/wpt_report.json --log-mach=- --this-chunk=$3 --total-chunks=$4 --test-type=$2 -y --install-browser --no-pause --no-restart-on-unexpected --reftest-internal --install-fonts --no-fail-on-unexpected
 elif [ $1 == "chrome" ]; then
-    ./wpt run chrome --log-tbpl==../artifacts/log_tbpl.log --log-tbpl-level=info --log-wptreport=../artifacts/wpt_report.json --log-mach=- --this-chunk=$3 --total-chunks=$4 --test-type=$2 -y  --no-pause --no-restart-on-unexpected --install-fonts
+    ./wpt run chrome --log-tbpl==../artifacts/log_tbpl.log --log-tbpl-level=info --log-wptreport=../artifacts/wpt_report.json --log-mach=- --this-chunk=$3 --total-chunks=$4 --test-type=$2 -y  --no-pause --no-restart-on-unexpected --install-fonts --no-fail-on-unexpected
```
jgraham: You don't want to do a drive-by fixup of the `==` that somehow got landed here despite (your?) earlier review comment pointing it out, do you?

jgraham (approved):

PR #10751, opened 2018-05-02
Testdriver API to minimize window

kereliuk:
This is a PR to add an API to minimize the current window.
Window maximizing/minimizing is a feature we've seen needed to test certain behaviour with at least one existing manual test.

(It also needs some docs and an infra test)

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10751)


kereliuk:
Leaving without an infra test for now since I'm not sure how to test this. @gsnedders WYT?

Also, one use case for this is in page-visibility/test_minimize-manual.html

kereliuk:
@gsnedders PTAL :)

foolip:
This is tracked by https://github.com/web-platform-tests/wpt/issues/10666, closing.

PR #10749, opened 2018-05-02
Testdriver API to get a list of window handles

kereliuk:
This is a PR to add an API to get a list of available window handles. It depends on #10716 and shouldn't be merged until it is resolved.

(It also needs some docs and a test)

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10749)

PR #10752, opened 2018-05-02
Testdriver API to switch to a window

kereliuk:
This is a PR that adds a testdriver APi to switch to a window given a handle. There needs to be a way to get the required handle name so this feature depends on #10749 and #10747 for now.

(It also needs some docs and an infra test)

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10752)

### DIFF
--- a/resources/testdriver.js
+++ b/resources/testdriver.js
@@ -119,6 +119,20 @@
             }
 
             return window.test_driver_internal.send_keys(element, keys);
+        },
+
+        /**
+         * Switch to a window
+         *
+         * This matches the behaviour of the {@link
+         * https://w3c.github.io/webdriver/#switch-to-window|WebDriver
+         * Switch to Window command}.
+         *
+         * @returns {Promise} fulfilled window is switched, or rejected in
+         *                    the cases the WebDriver command errors
+         */
+        switch_to_window: function(window_handle) {
+            return window.test_driver_internal.switch_to_window(window_handle);
         }
     };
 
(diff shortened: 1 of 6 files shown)


kereliuk:
An example and requested use case for this is for testing lifecycle APIs with page states.

For now I am leaving the infra test for a follow-up CL since I will want other commands that get me the window handles for that :)

kereliuk:
@gsnedders PTAL :)

resources/testdriver.js
```diff
@@ -143,6 +157,20 @@
...
+         *
+         * This matches the behaviour of the {@link
+         * https://w3c.github.io/webdriver/#switch-to-window|WebDriver
+         * Switch to Window command}.
+         *
+         * @returns {Promise} fulfilled window is switched, or rejected in
+         *                    the cases the WebDriver command errors
+         */
+        switch_to_window: function(window_handle) {
+            return window.test_driver_internal.switch_to_window(window_handle);
```
gsnedders: This is just calling itself; should probably be a promise that rejects like the above

tools/wptrunner/wptrunner/executors/executorselenium.py
```diff
@@ -142,6 +143,13 @@ def setup(self):
     def send_keys(self, element, keys):
         return element.send_keys(keys)
 
+class SeleniumSwitchToWindowProtocolPart(SwitchToWindowProtocolPart):
+    def setup(self):
+        self.webdriver = self.parent.webdriver
+
+    def send_keys(self, element, keys):
```
gsnedders: This is the wrong protocol

gsnedders (changes requested):
Couple of small comments, and needs some in-tree implementation (strongly) preferably. @jgraham might have more comments?

foolip:
This is tracked by https://github.com/web-platform-tests/wpt/issues/10666, closing.

PR #10755, opened 2018-05-02
testcase fixes in css/CSS2/borders

faceless2:
Two changes:

1. border-*N*-width-applies-to-008.xht tests did not have a reftest and were slightly dependent on default font-metrics (specifically, nbsp being 0.25em wide). Redid to use an image, although they're still testing the same thing. Added a matching reftest

2. border-color-011.xht was incorrect, result of test has reversed since the rules regarding "currentColor" inheritance were clarified in css-color-4

3. additional whitespace in bidi-003.xht was causing reftest mismatch

### DIFF
--- a/css/CSS2/bidi-text/bidi-003.xht
+++ b/css/CSS2/bidi-text/bidi-003.xht
@@ -41,7 +41,7 @@
     <span class="control b end">   mmm nnn ooo </span>
    </p>
    <p>
-    <span class="test a"> aaa bbb ccc &#x202E; lll kkk jjj </span> iii hhh ggg <span class="test b"> fff eee ddd &#x202C; mmm nnn ooo </span>
+    <span class="test a"> aaa bbb ccc &#x202E; lll kkk jjj </span> iii hhh ggg <span class="test b"> fff eee ddd &#x202C;mmm nnn ooo </span>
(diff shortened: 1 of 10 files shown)


TalbotG:
Mike,

In your border-*-width-applies-to-008 tests:
Shouldn't you be saying that it's a square, not a rectangle?

line 26 or 27:  ``
is going to trigger a validation error. Same validation error in your border-n-width-applies-to-008-ref.xht reference file.

Also, 
alt="Image download support must be enabled" 
is missing. We want to know easily and quickly if there are any broken links or incorrect linkage to images.

TalbotG:
In your border-*-width-applies-to-008 tests:

Since img is used, then

line 18:                  display: inline;

can be safely removed too.

svgeesus (approved):
The change to border-color-011.xht correctly captures the currentColor changes in css-color-4 (and css-color-3, edited rec).

faceless2:
Closed; reapplied as single patches.
* see https://github.com/web-platform-tests/wpt/pull/21662
* see https://github.com/web-platform-tests/wpt/pull/21663
... and have left border-N-width-applies-to-008.xht - this is an old test and the change was was significant, probably not worth it.

PR #10756, opened 2018-05-02
HTML: document.open() and the document's encoding

annevk:


domenic (approved):
Still don't understand why you keep testing document.close() in your document.open() tests, but this will test the issue, I suppose.

domenic:
w3c-test:mirror

annevk:
Did you try removing it and running it in all browsers?

domenic:
Nope. I just think if there is some reason, a comment would help explain it, instead of relying on the reader to do such experimentation.

PR #10760, opened 2018-05-02
Add webgl/OWNERS

zcorpan:
cc @jdashg

### DIFF
--- a/webgl/OWNERS
+++ b/webgl/OWNERS
@@ -1,2 +1,4 @@
 @grorg
 @jdashg
+@kenrussell
+@RafaelCintron
--- a/webgl/OWNERS
+++ b/webgl/OWNERS
@@ -1,4 +1,1 @@
-@grorg
-@jdashg
 @kenrussell
-@RafaelCintron
--- a/webgl/OWNERS
+++ b/webgl/OWNERS
@@ -1,0 +1,2 @@
+@grorg
+@jdashg


jdashg:
What does OWNERS mean in this context?

jdashg:
@kenrussell and @RafaelCintron maybe?

Since active development here is largely in the Khronos repo still, I don't know how much activity we'll have/need here.

kenrussell:
If we're putting OWNERs in place then, yes, I would appreciate it if myself and @RafaelCintron would be added. However we do plan to continue to develop the WebGL conformance tests in https://github.com/KhronosGroup/WebGL .

zcorpan:
Being in OWNERS means that you get a review request when someone opens a PR to change something in the dir or subdir. There is no obligation though.

zcorpan:
I don't have any strong opinion on how to handle there being two places for webgl tests. @foolip @jgraham @gsnedders any thoughts?

foolip:
I filed https://github.com/KhronosGroup/WebGL/issues/2537 a while ago, so I think it'd be best if there weren't any WebGL tests in wpt.

zcorpan:
I think OWNERS file can be separate from the discussion of whether there should be any tests. If we decide webgl tests should not be here, we can leave a directory with only OWNERS and a README.md that explains the situation. I'll file an issue about removing webgl tests.

zcorpan:
@grorg @jdashg @RafaelCintron are you OK with being listed in the OWNERS file here? It means you’ll get notified any time someone submits a PR with Media Capture tests, or changes to existing tests. And the tests could only be merged to the master branch after review from one of the OWNERS.

PR #10762, opened 2018-05-02
Move some setup out of the docker file to  fix taskcluster problems.

jgraham:
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10762)

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -67,10 +67,11 @@ COPY start.sh /home/test/start.sh
 # Remove information on how to use sudo on login
 RUN sudo echo ""
 
-RUN git clone --depth=1 https://github.com/w3c/web-platform-tests.git
-
-RUN mkdir -p /home/test/.fonts && \
-    cp web-platform-tests/fonts/Ahem.ttf ~/.fonts && \
-    fc-cache -f -v
+RUN mkdir -p /home/test/web-platform-tests
+WORKDIR /home/test/web-platform-tests
+RUN git init
+RUN git remote add origin https://github.com/w3c/web-platform-tests.git
 
 RUN mkdir -p /home/test/artifacts
+
+WORKDIR /home/test/
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -60,18 +60,13 @@ USER test
 
 WORKDIR /home/test
 
-COPY .bashrc /home/test/.bashrc
-
-COPY start.sh /home/test/start.sh
-
 # Remove information on how to use sudo on login
 RUN sudo echo ""
-
-RUN mkdir -p /home/test/web-platform-tests
-WORKDIR /home/test/web-platform-tests
-RUN git init
-RUN git remote add origin https://github.com/w3c/web-platform-tests.git
 
 RUN mkdir -p /home/test/artifacts
 
 WORKDIR /home/test/
+
+COPY .bashrc /home/test/.bashrc
+
+COPY start.sh /home/test/start.sh
--- a/tools/docker/start.sh
+++ b/tools/docker/start.sh
@@ -1,16 +1,34 @@
 #!/bin/bash
 set -ex
 
+REMOTE=${1:-https://github.com/w3c/web-platform-tests}
+BRANCH=${2:-master}
+REV=${3:-FETCH_HEAD}
+BROWSER=${4:-all}
+
+cd ~
+
+# Initially we just fetch 50 commits in order to save several minutes of fetching
+git clone ${REMOTE} --single-branch --branch ${BRANCH} --no-checkout -q --depth=50 web-platform-tests
 cd web-platform-tests
-git fetch origin -n -q --depth=50
-git checkout -b master origin/master
+if [[ ! `git rev-parse --verify -q ${REV}` ]];
+then
+    # But if for some reason the commit under test isn't in that range, we give in and
+    # fetch everything
+    git fetch -q --unshallow ${REMOTE}
+    git rev-parse --verify ${REV}
+fi
+git checkout -b build ${REV}
 
 sudo sh -c './wpt make-hosts-file >> /etc/hosts'
 
-# Install Chome dev
-deb_archive=google-chrome-unstable_current_amd64.deb
-wget https://dl.google.com/linux/direct/$deb_archive
+if [[ $BROWSER == "chrome"* ]] || [[ "$BROWSER" == all ]]
+then
+    # Install Chrome dev
+    deb_archive=google-chrome-unstable_current_amd64.deb
+    wget https://dl.google.com/linux/direct/$deb_archive
 
-sudo gdebi -n $deb_archive
+    sudo gdebi -n $deb_archive
+fi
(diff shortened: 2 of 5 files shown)


Ms2ger (approved):

tools/docker/Dockerfile
```diff
@@ -67,10 +67,11 @@ COPY start.sh /home/test/start.sh
 # Remove information on how to use sudo on login
 RUN sudo echo ""
 
-RUN git clone --depth=1 https://github.com/w3c/web-platform-tests.git
-
-RUN mkdir -p /home/test/.fonts && \
-    cp web-platform-tests/fonts/Ahem.ttf ~/.fonts && \
-    fc-cache -f -v
+RUN mkdir -p /home/test/web-platform-tests
```
foolip: Can you add a comment explaining why this isn't just `git clone https://github.com/w3c/web-platform-tests` + `WORKDIR /home/test/web-platform-tests`?

And also comment that the reason for having a checkout as part of the image is to save on network traffic or speed things up?

tools/docker/start.sh
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
+set -ex
 
 cd web-platform-tests
-git pull --depth=50
+git fetch origin -n -q --depth=50
```
foolip: Isn't the `--depth=50` a bit pointless now? Presumably we'll be updating the image once in a while, and does it even work to do a shallow fetch from within a full clone?

tools/docker/start.sh
```diff
@@ -1,7 +1,9 @@
...
+git checkout -b master origin/master
```
foolip: Is there a separate step that checks out the right revision?

foolip (approved):
LGTM % documentation for everything I don't understand :)

tools/docker/Dockerfile
jgraham: Having the checkout as part of the image was the previous state, but now we don't. It probably did make things a little faster, but I imagine the gains would decline as the image gets older and on the other hand it does make it much slower to iterate on the docker image. In the future we could maybe consider a task to build a base image containing a checkout, updated e.g. weekly.

tools/docker/start.sh
jgraham: Yeah, although I have a followup to do this more obviously.

tools/docker/Dockerfile
foolip: Oh, so the whole point of this now is to not run `git fetch`. I did not spot that, comment would be good :)

tools/docker/start.sh
```diff
@@ -1,14 +1,24 @@
...
+git checkout -b build ${REV}
 
 sudo sh -c './wpt make-hosts-file >> /etc/hosts'
 
-# Install Chome dev
-deb_archive=google-chrome-unstable_current_amd64.deb
-wget https://dl.google.com/linux/direct/$deb_archive
+if [[ $BROWSER == "chrome"* ]] || [[ "$BROWSER" == all ]]
+then
+    # Install Chome dev
```
foolip: Preexisting typo: Chome

tools/docker/start.sh
```diff
@@ -1,14 +1,24 @@
...
-cd web-platform-tests
-git pull --depth=50
+REPO=${1:-origin}
+BRANCH=${2:-master}
+REV=${3:-master}
+BROWSER=${4:-all}
+
+cd ~/web-platform-tests
+git fetch ${REPO} ${BRANCH} -n -q --depth=50
+git checkout -b build ${REV}
```
foolip: Why create a build branch?

tools/docker/start.sh
```diff
@@ -1,14 +1,24 @@
...
 
 sudo sh -c './wpt make-hosts-file >> /etc/hosts'
 
-# Install Chome dev
-deb_archive=google-chrome-unstable_current_amd64.deb
-wget https://dl.google.com/linux/direct/$deb_archive
+if [[ $BROWSER == "chrome"* ]] || [[ "$BROWSER" == all ]]
```
foolip: Why the asterix after "chrome"?

tools/docker/start.sh
```diff
@@ -1,14 +1,24 @@
 #!/bin/bash
+set -ex
 
-cd web-platform-tests
-git pull --depth=50
+REPO=${1:-origin}
```
foolip: s/REPO/REMOTE/

tools/docker/start.sh
```diff
@@ -1,14 +1,24 @@
...
+BRANCH=${2:-master}
+REV=${3:-master}
```
foolip: I think FETCH_HEAD or origin/master would be better here. Otherwise, when you get rid of the "build" branch as I've suggested, the checkout step will sometimes create a new branch called master and sometimes not.

tools/docker/Dockerfile
foolip: Have the comment say "the remote is added but not fetched, the right revision is fetched inside the container".

But... then I have to wonder why bother setting this up in the Dockerfile at all, a plain git clone inside the container would be fewer lines of code and easier to understand.

tools/docker/start.sh
```diff
@@ -1,14 +1,24 @@
...
+BROWSER=${4:-all}
+
+cd ~/web-platform-tests
+git fetch ${REPO} ${BRANCH} -n -q --depth=50
```
foolip: The `--depth=50` still seems like a bit of a liability, making it impossible to restart a failed job if it failed more than 50 commits ago, and if we merge in other repos (like csswg-test) that might be just the previous commit.

How much time does it save to use `--depth=50`?

An alternative, which seems to work, is to make Taskcluster depend on the merge_pr_* tags. Then we could do `git fetch origin merge_pr_10791 --depth=1` here.

At minimum, a comment explaining why `--depth=50` is a good idea would be appreciated :)

foolip (approved):
Asking for more comments and wondering about `--depth=50` :)

tools/docker/start.sh
jgraham: It makes it a prefix match. Otherwise `chrome-dev` isn't going to match :)

tools/docker/start.sh
jgraham: Why not? It avoids a useless warning about checking out a refless revision, avoids having to suppress that warning, and potentially makes things a little easier to work with for interactive debugging.

tools/docker/start.sh
foolip: OK, but if "chrome" is given we'll still use Chrome Dev... I guess this doesn't matter unless we want to also run something other than Chrome Dev though.

tools/docker/start.sh
foolip: Yeah, let's call it a "later" problem.

tools/docker/start.sh
foolip: OK, so there's a warning to suppress :) If not for that, it'd just be more code.

foolip (approved):
Hope it works now :)

PR #10769, opened 2018-05-02
Add support for JS shell tests to the manifest code.

Ms2ger:
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10769)

### DIFF
--- a/tools/wptrunner/wptrunner/wptrunner.py
+++ b/tools/wptrunner/wptrunner/wptrunner.py
@@ -245,6 +245,12 @@ def run_tests(config, test_paths, product, **kwargs):
                             if test.testdriver and not executor_cls.supports_testdriver:
                                 logger.test_start(test.id)
                                 logger.test_end(test.id, status="SKIP")
+                            elif test.jsshell and not executor_cls.supports_jsshell:
+                                # We expect that tests for JavaScript shells
+                                # will not be run along with tests that run in
+                                # a full web browser, so we silently skip them
+                                # here.
+                                pass
                             else:
                                 run_tests["testharness"].append(test)
                     else:
--- a/tools/manifest/item.py
+++ b/tools/manifest/item.py
@@ -102,10 +102,11 @@ class URLManifestItem(ManifestItem):
 class TestharnessTest(URLManifestItem):
     item_type = "testharness"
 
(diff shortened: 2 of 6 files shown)


tools/wptrunner/wptrunner/wptrunner.py
```diff
@@ -245,6 +245,9 @@ def run_tests(config, test_paths, product, **kwargs):
                             if test.testdriver and not executor_cls.supports_testdriver:
                                 logger.test_start(test.id)
                                 logger.test_end(test.id, status="SKIP")
+                            elif test.jsshell and not executor_cls.supports_jsshell:
+                                logger.test_start(test.id)
```
jgraham: Does the jsshellness change the test id here? If not we are going to get multiple results for the "same" test.

tools/wptrunner/wptrunner/wptrunner.py
```diff
@@ -245,6 +245,9 @@ def run_tests(config, test_paths, product, **kwargs):
...
+                                logger.test_end(test.id, status="SKIP")
```
jgraham: Let's just ignore these rather than print a "skip" message; we expect a disjoint set of jsshell and non-jsshell executors.

jgraham:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[tools/wptrunner/wptrunner/wptrunner.py, line 249 at r2](https://reviewable.io/reviews/w3c/web-platform-tests/10769#-LBzDZvC4d6iml5LoYaq:-LBzDZvD1lTSw5j2XGp9:b743qlr) ([raw file](https://github.com/w3c/web-platform-tests/blob/81f0bae881758f1b1068087bec89169ce6b0afaf/tools/wptrunner/wptrunner/wptrunner.py#L249)):*
> ```Python
>                                 logger.test_end(test.id, status="SKIP")
>                             elif test.jsshell and not executor_cls.supports_jsshell:
>                                 pass
> ```

I think we shoudl add a comment here, but no need for more review when you are done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10769)*

jgraham (approved):

PR #10778, opened 2018-05-02
HTML: document.open() and the beforeunload event

annevk:
For https://github.com/whatwg/html/pull/3918.


html/webappapis/dynamic-markup-insertion/opening-the-input-stream/beforeunload.window.js
```diff
@@ -0,0 +1,12 @@
+async_test(t => {
+  const frame = document.body.appendChild(document.createElement("iframe"));
+  t.add_cleanup(() => frame.remove());
```
foolip: This doesn't help much when there's a single test, and leaving it in might make debugging easier.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/beforeunload.window.js
```diff
@@ -0,0 +1,12 @@
...
+  frame.src = "/common/blank.html";
+  frame.onload = t.step_func(() => {
+    frame.contentWindow.onbeforeunload = t.unreached_func("beforeunload should not be fired");
+    frame.contentDocument.open();
+    t.step_timeout(t.step_func_done(() => {
+      frame.contentDocument.close();
```
foolip: Can you add a comment describing why it's helpful to invoke `document.close()` here? I don't quite see why that would fire the beforeunload event, which is the only thing that would change the outcome of the test...

foolip (approved):
LGTM if my comments don't lead to any non-trivial changes.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/beforeunload.window.js
TimothyGu: Sure. The intent here is to test that `open()` does _not_ fire the `beforeunload` event. The `close()` here is merely to fire the `load` event on the document and unblock testharness from thinking that the page has not yet loaded.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/beforeunload.window.js
TimothyGu: We don't want the test to time out, and having `frame.remove()` here is the easiest way to ensure that the `load` event is fired on the top-level `Window` no matter the outcome of the test.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/beforeunload.window.js
TimothyGu: Added a comment. LMK if that works.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/beforeunload.window.js
foolip: Ah, I assumed it was copypasta from something with multiple tests. Never mind me.

foolip (approved):

PR #10782, opened 2018-05-02
Add keyboard-lock/OWNERS

zcorpan:
cc @jamiewalch


zcorpan:
@garykac are you OK with being listed in the OWNERS file here? It means you’ll get notified any time someone submits a PR with Media Capture tests, or changes to existing tests. And the tests could only be merged to the master branch after review from one of the OWNERS.

garykac:
lgtm.  @joedow-42 should also be listed here as well.  Thanks!

PR #10800, opened 2018-05-02
Fix #10741: Call Out-File with an encoding for make-hosts-file on Windows

### LINKED ISSUE #10741: Python checks for hosts file are not working anymore
After @gsnedders recent changes to run.py, tests no longer work on Windows. During the check for the hosts file, it throws an error suggesting you should create the hosts entries. However, doing so does not help.

When the entries are created, there is an additional space character injected between each char

eg:  1 2 7 . 0 . 0 . 1 	 w e b - p l a t f o r m . t e s t

So then later, the checks fail. If I remove the spaces, then all is good.

I was just going to submit a new PR for that, but I cannot find the spot where those spaces are added. I'm still checking, though.

gsnedders:
Fixes #10741.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10800)

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ For example, on most UNIX-like systems, you can setup the hosts file with:
 ./wpt make-hosts-file | sudo tee -a /etc/hosts
 ```
 
-And on Windows (note this requires in PowerShell with Administrator privileges):
+And on Windows (this must be run in a PowerShell session with Administrator privileges):
 
 ```bash
 python wpt make-hosts-file | Out-File %SystemRoot%\System32\drivers\etc\hosts -Encoding ascii -Append
--- a/README.md
+++ b/README.md
@@ -44,10 +44,10 @@ For example, on most UNIX-like systems, you can setup the hosts file with:
 ./wpt make-hosts-file | sudo tee -a /etc/hosts
 ```
 
-And on Windows (note this requires an Administrator privileged shell):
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -44,10 +44,10 @@ For example, on most UNIX-like systems, you can setup the hosts file with:
 ./wpt make-hosts-file | sudo tee -a /etc/hosts
 ```
 
-And on Windows (note this requires an Administrator privileged shell):
+And on Windows (note this requires in PowerShell with Administrator privileges):
```
jugglinmike: How about "this requires PowerShell with Administrator privileges" or "this must be run in PowerShell with Administrative privileges"?

jgraham (approved):
Would be nice to make it clear that using something other than PowerShell works like normal, otherwise fine modulo the typo

gsnedders:
@jgraham I don't know many people still using `cmd.exe` on Windows; it's either Bash on Windows or PowerShell nowadays (and we didn't deal with the former before, given we'd need to generate a different path for it).

README.md
gsnedders: I was trying to figure out how MS refer to it; I think something like the latter is probably what we want.

gsnedders: Does this now look better?

jugglinmike: Much!

sideshowbarker (approved):

PR #10803, opened 2018-05-02
[infra] Improve test scheduling in CI

jugglinmike:
Ensure that any changes to the integration tests for the project's
infrastructure are validated by the continuous integration system at the
time they are submitted.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10803)

### DIFF
--- a/tools/ci/jobs.py
+++ b/tools/ci/jobs.py
@@ -27,7 +27,6 @@ job_path_map = {
     "wptrunner_unittest": ["tools/wptrunner/*"],
     "build_css": ["css/"],
     "update_built": ["2dcontext/",
-                     "assumptions/",
                      "html/",
                      "offscreen-canvas/"],
     "wpt_integration": ["tools/"],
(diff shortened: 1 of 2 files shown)


jgraham (changes requested):
This seems like a reasonable idea, but there isn't a directory called `asumptions/` (I think you mean `infrastructure` and the commit needs a more specific summary.

jugglinmike:
> there isn't a directory called `asumptions/` (I think you mean
> `infrastructure`

You are correct. I think I was confused by the entry that precedes the modified one. I've added a commit to remove that configuration. Doing so alerted me to the presence of tests for this behavior, so I've also extended the original patch with a new test.

> and the commit needs a more specific summary.

Sure. I've amended the original commit and force-pushed the result. [The original version of this branch is available here.](https://github.com/w3c/web-platform-tests/compare/master...bocoup:improve-ci-scheduling-orig)

jgraham (approved):

Issue #10511, opened 2018-04-17
Include more metadata in wptreport JSON


foolip:
This came up in [Design Doc: WPT Results Receiver](https://docs.google.com/document/d/1F3GWJ6lA8C3S5WgyJgXxmTFzu92mqF-1ddGije67ork/edit?disco=AAAAB3EPiyA) with @jgraham and @Hexcles.

The format is currently:
```
{"results": [
  {"test": "/foo/bar/baz.html",
   "status": "OK",
   "message": null,
   "subtests": [
     {"name": "...", "status": "PASS", "message": "..."},
     ...
   ]
  },
  ...
]}
```

Add something like a `metadata` field at the top level, another object with something like:
- The wpt commit being tested
- Full `wpt run` command line (which would include sharding commands)
- OS, browser and webdriver version. Browser version also needs to include channel (stable, beta, dev) since the exact same version number might appear in two release channels.
- Datetime when `wpt run` was invoked and datetime when run finished. (duration can be computed)
- Name and revision of any runner environment being used, like https://github.com/web-platform-tests/results-collection

@jugglinmike, anything else you'd like to put in there?

gsnedders:
> OS, browser and webdriver version. Browser version also needs to include channel (stable, beta, dev) since the exact same version number might appear in two release channels.

I'd like to see the UA string, too?

gsnedders:
See also #10026.

foolip:
The UA string couldn't be trusted for the version number, as Safari's is frozen, but sure, if it's already known then including it seems fine.

jugglinmike:
> OS, browser and webdriver version. Browser version also needs to include > channel (stable, beta, dev) since the exact same version number might appear > in two release channels.

What do you think about a vendor-defined build identifier?

> Full `wpt run` command line (which would include sharding commands) 

The arguments passed to the WPT CLI may be overly specific to our current needs. They may include a lot of irrelevant information (`--yes`, `--log-tbpl`, etc.). Other folks don't use that tool, so they would have to take extra steps to interpret that information. This could be inconvenient when comparing datasets across time because the CLI is part of WPT itself--the type and meaning of arguments may change with any revision.

What do you think about distilling the relevant information down into runner-agnostic properties? That would include environment variables and command-line arguments for both the browser process and the WebDriver server process, though there might be more.

[...]

PR #10811, opened 2018-05-03
Add run_info and timestamps to wptreport

Hexcles:
The first stab at #10511 , adding the simplest fields to make sure I understand correctly @jgraham 

Meanwhile, how shall I pass the version of the browser? Via `run_info`? FWIW, there's `run_info['version']` which is the version of my OS (and it seems to overlap with `run_info['os_version']` a bit).

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10811)

### DIFF
--- a/tools/wptrunner/wptrunner/formatters.py
+++ b/tools/wptrunner/wptrunner/formatters.py
@@ -8,15 +8,20 @@ class WptreportFormatter(BaseFormatter):
 
     def __init__(self):
         self.raw_results = {}
+        self.results = {}
+
+    def suite_start(self, data):
+        self.results['run_info'] = data['run_info']
+        self.results['time_start'] = data['time']
 
     def suite_end(self, data):
-        results = {}
-        results["results"] = []
+        self.results['time_end'] = data['time']
+        self.results["results"] = []
         for test_name in self.raw_results:
             result = {"test": test_name}
             result.update(self.raw_results[test_name])
-            results["results"].append(result)
-        return json.dumps(results)
+            self.results["results"].append(result)
+        return json.dumps(self.results)
 
     def find_or_create_test(self, data):
         test_name = data["test"]


tools/wptrunner/wptrunner/formatters.py
```diff
@@ -8,15 +8,21 @@ class WptreportFormatter(BaseFormatter):
 
     def __init__(self):
         self.raw_results = {}
+        self.results = {}
+
+    def suite_start(self, data):
+        self.results['browser'] = data['run_info']['product']
```
jgraham: I would be tempted to just dump `run_info` directly into the results.

Hexcles: Yeah let's do that.

Meanwhile, could you answer the question I had regarding these fields in `run_info`?
```
"linux_distro": "Debian",
"os": "linux",
"os_version": "sid",
"version": "Debian sid",
```
They seem to present an overlapping set of information. Is this intended? Notably, what's `version`? Is it intended to be the version of the product or the OS?

jugglinmike:
Since we're interested in building on this data from external projects, what do you think about documenting this format in the `docts/` directory?

Hexcles:
@jugglinmike First of all, I'm going to document some of the mandatory fields in my [design doc](https://docs.google.com/document/d/1F3GWJ6lA8C3S5WgyJgXxmTFzu92mqF-1ddGije67ork/edit), including the browser info and revision, which will solve the most pressing issue for uploading results.

Later we can more formally document the whole wptreport here as a doc and/or JSON schema.

tools/wptrunner/wptrunner/formatters.py
jgraham: All I know is what I can figure out from `https://searchfox.org/mozilla-central/source/testing/mozbase/mozinfo/mozinfo/mozinfo.py`. So `version` is what you get from `uname` and `os_version` has a more custom  implementation.

jgraham (approved):

PR #10789, opened 2018-05-02
HTML: document.open() and aborting documents

annevk:
For https://github.com/w3c/web-platform-tests/pull/10773.

This PR (now) tests the new spec text proposed in https://github.com/whatwg/html/issues/3818.

PR #10818, opened 2018-05-03
HTML: document.open() and tasks

annevk:
For https://github.com/whatwg/html/pull/3651.


domenic:
Some other potential tests:

- An image being created and its load event being fired
- A marquee element's start event being fired
- A fetch completing

annevk:
`fetch()` is interesting, I test that in #10789 and only Firefox fires an error event there. I guess I should add an `onload` with `unreached_func` to see if some trigger that.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/tasks.window.js
```diff
@@ -0,0 +1,114 @@
...
+      assert_in_array(frame.contentDocument.readyState, ["interactive", "complete"]);
+      testBody(t, frame, doc => doc.open());
+    });
+  }, `document.open() and tasks (${description})`);
+}
+
+taskTest("timeout", (t, frame, open) => {
+  let happened = false;
+  // Work around the lint script, since we can't use step_timeout here.
+  frame.contentWindow['setTimeout'](() => happened = true, 100);
```
foolip: Rather than setting `happened` here, how about just `t.step_func_done()` as the callback to pass the test ASAP?

The following test just times out if there isn't a 2nd message event, so I'd let this test also time out. If that's no good, is there any timeout one could pick that doesn't make the test more likely to be flaky on slow runners?

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/tasks.window.js
```diff
@@ -0,0 +1,114 @@
...
+  let counter = 0;
+  frame.contentWindow.postMessage(undefined, "*");
+  open(frame.contentDocument);
+  frame.contentWindow.postMessage(undefined, "*");
+  frame.contentWindow.onmessage = t.step_func(e => {
+    assert_equals(e.data, undefined);
+    counter++;
+    assert_less_than_equal(counter, 2);
+    if (counter == 2) {
+      t.done();
```
foolip: (This is the test that will time out after 30s or so.)

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/tasks.window.js
```diff
@@ -0,0 +1,114 @@
...
+    }
+  });
+});
+
+taskTest("canvas.toBlob()", (t, frame, open) => {
```
foolip: The "queue a task" of `toBlob` happens inside "in parallel" so might not have happened when `document.open()` is called, is that a problem for this test? Does any browser currently fail this?

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/tasks.window.js
```diff
@@ -0,0 +1,114 @@
...
+  });
+  frame.contentWindow.onunhandledrejection = listener;
+  window.onunhandledrejection = listener;
+});
+
+taskTest("marquee start", (t, frame, open) => {
+  const doc = frame.contentDocument;
+  const marquee = doc.body.appendChild(doc.createElement("marquee"));
+  open(frame.contentDocument);
+  marquee.addEventListener("start", t.step_func_done());
```
foolip: I like this one. If you want one which is a bit special in practice try a new audio element and changing the volume attribute to 0.5. That should queue a task to fire volumechange. But implementations do something special to emulate that per-instance task source that media elements are supposed to have, and I wouldn't be surprised if it's different from other things tested here.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/tasks.window.js
TimothyGu: Good point. No browser actually seems to fail this, but I should note that `setTimeout` is defined the same way (wait in parallel, and then queue a task), and browsers have very different behavior with regards to that. That's why I think this is a good test.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/tasks.window.js
TimothyGu: This didn't seem to timeout in the latest Travis run…

foolip: Travis unfortunately doesn't fail unless the tests are flaky, it doesn't help catch any other problem, even totally broken tests. Did you read the logs to see that it didn't time out, or how could you tell?

TimothyGu: Oh right, I guess I meant rather that it wasn't flaky.

foolip (approved):
Only remaining comments requiring change are the "link to bug" ones, if you like.

PR #10819, opened 2018-05-03
[css-flex][spec-rec] Test for flex only alignment modes with writing modes (vertical-lr)

benfluleck:
#### What does this PR do? 
These tests ensure that flex only alignment modes work as expected in different
writing modes. 
The tests use the writing-mode: vertical-lr with the flex alignment modes 
(justify-content and align-items)

#### How should this be manually tested?
Pull the current branch
Follow the Readme.md to run the tests
*Testharness.js has been used to write the test cases*

#### Screenshots (if appropriate)
screen shot 2018-05-03 at 12 37 53 pm


gregwhitworth:
@benfluleck you and @claytron5000 should get on the same page regarding this - as his test will cover this one since he's generating it all. I recommend pinging him and doing his CR, then when you both agree that his test is sufficient - loop me in for CR.

Issue #10799, opened 2018-05-02
/workers/semantics/structured-clone/* are disabled in Mozilla and WebKit


zcorpan:
http://bocoup.github.io/wpt-disabled-tests-report/

Investigate what's up with this test:

Path | Products | Results | Bugs
-- | -- | -- | --
/workers/semantics/structured-clone/dedicated.html | mozilla webkit | disabled: true [ Skip ] | None None
/workers/semantics/structured-clone/shared.html | mozilla webkit | disabled: true [ Skip ] | None None

Possibly these tests are doing too many things and would work better if split up to several tests.

zcorpan:
Doesn't look very good for Chromium either:

https://test-results.appspot.com/dashboards/flakiness_dashboard.html#testType=webkit_layout_tests&tests=external%2Fwpt%2Fworkers%2Fsemantics%2Fstructured-clone%2F

zcorpan:
This should be fixed by @surma's PR https://github.com/w3c/web-platform-tests/pull/9218

cdumez:
@zcorpan Sure, will try.

cdumez:
Test is flaky for us:
/Volumes/Data/cdumez/WebKit/OpenSource/WebKitBuild/Debug/layout-test-results/imported/w3c/web-platform-tests/workers/semantics/structured-clone/dedicated-actual.txt
8585PASS Object Blob object, Blob paired surrogates (invalid utf-8) 
8686PASS Object Blob object, Blob empty 
8787PASS Object Blob object, Blob NUL 
88 FAIL File basic assert_equals: lastModified expected 42 but got 1526665571776
 88FAIL File basic assert_equals: lastModified expected 42 but got 1526665823670
8989PASS FileList empty 
9090PASS Array FileList object, FileList empty 
9191PASS Object FileList object, FileList empty

cdumez:
Printing out lastModified like this does not give stable results.

zcorpan:
Ok. Did you import from today i.e. including #9218?

It seems the test expects 42 and webkit gives something else. Bug?

cdumez:
Yes it is the latest version of the tests. Yes it is a bug in WebKit that the value is not 42). Still the test is flaky when it fails in WebKit because it prints out a timestamp when failing.

zcorpan:
I see. We could make it not flaky by doing an if check and calling assert_unreached instead of assert_equals. Or by fixing the bug. ☺

cdumez:
[...]

PR #10820, opened 2018-05-03
Split up and fix structured clone tests 

zcorpan:
This should help with #10799.


annevk:
Please coordinate with @surma given #9218. Ideally we can share these kind of tests across APIs.

zcorpan:
I agree @annevk. I think we can land this first as it's making existing tests actually work. Agree? @surma rather than duplicating code, can you rebase when this is merged and try to make the tests do what you want? (It looks like you have some good improvements there!)

PR #10821, opened 2018-05-03
File API: lastModifiedDate has been obsolete for a while

annevk:

### DIFF
--- a/FileAPI/historical.https.html
+++ b/FileAPI/historical.https.html
@@ -53,9 +53,13 @@
         assert_false('isClosed' in Blob.prototype, 'isClosed in Blob.prototype');
     }, 'Blob.close() should not be supported');
 
-    // Only add service worker test if service workers are actually supported.
-    if (navigator.serviceWorker)
-      service_worker_test('support/historical-serviceworker.js', 'Service worker test setup');
+    test(() => {
+      const f = new File([], "");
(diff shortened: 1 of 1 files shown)


Ms2ger (approved):

annevk:
https://bugzilla.mozilla.org/show_bug.cgi?id=1458883

FileAPI/historical.https.html
```diff
@@ -53,9 +53,13 @@
         assert_false('isClosed' in Blob.prototype, 'isClosed in Blob.prototype');
     }, 'Blob.close() should not be supported');
 
-    // Only add service worker test if service workers are actually supported.
-    if (navigator.serviceWorker)
```
Zirro: @annevk Was the removal of the condition here accidental? It causes this test to fail in implementations without Service Workers, such as jsdom.

annevk: Yeah, we shouldn't have conditional tests. That makes comparisons between implementations fragile. I'd be okay with splitting this test out, but these kind of conditionals in tests are bad.

mkruisselbrink: But now that means this test will fail on implementations that correctly don't implement what the test is trying to make sure they don't implement, just because they don't implement something else as well? That does seem weird to me...
At least that was my original reasoning for having the conditional here.

PR #10822, opened 2018-05-03
Rfc6265bis tests

mozmark:
These are a port of @mikewest 's tests on http://rfc6265.biz/tests/ to WPT with @gsnedders ' Add alternate_hosts work (merged in #10166). I have more tests to add but I wanted to make sure the straight porting of these was OK first.


annevk:
Yay! Please put these in `cookies/`; no need for a new top-level directory. Also, FWIW, the main reason this fails is due to linting errors: https://travis-ci.org/w3c/web-platform-tests/jobs/374399296#L538.

cookies/rfc6265bis/index.html
```diff
@@ -0,0 +1,364 @@
+<!DOCTYPE html>
```
mikewest: We don't need this file. It made sense for the stand-alone site, it doesn't make sense as part of WPT.

cookies/rfc6265bis/prefix/__host.document-cookie.non-secure.html
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE html>
```
mikewest: I'd suggest dropping the `rfc6265bis` directory. "prefix" is enough.

cookies/rfc6265bis/resources/testharness.css
```diff
@@ -0,0 +1,102 @@
+html {
```
mikewest: We don't need this, or `testharness.js`, or `testharnessreport.js`.

cookies/rfc6265bis/prefix/__host.document-cookie.non-secure.html
```diff
@@ -0,0 +1,33 @@
...
+<script src="/cookies/rfc6265bis/resources/testharness.js"></script>
+<script src="/cookies/rfc6265bis/resources/testharnessreport.js"></script>
```
mikewest: These should both be using WPT's `testharness.js` and `testharnessreport.js`, here and in every other HTML file.

mikewest (commented):
This looks like a good port, thanks for doing the work! A few comments inline.

mikewest (approved):
Thanks!

PR #10828, opened 2018-05-03
"Scroll an element into view" in HTML spec

nanto:
This is tests for whatwg/html#3131.

(I'm advised to write tests in https://github.com/whatwg/html/issues/3641#issuecomment-383603754)


zcorpan:
html/browsers/browsing-the-web/scroll-to-fragid/scroll-position-vertical-lr.html (and rl) fail in chromium (with "experimental web platform features" enabled). Can you report a bug if there isn't one already?

The 2dcontext tests I think are generated from tools/tests.yaml or some such. Can the new tests be there also?

nanto:
> html/browsers/browsing-the-web/scroll-to-fragid/scroll-position-vertical-lr.html (and rl) fail in chromium (with "experimental web platform features" enabled). Can you report a bug if there isn't one already?

I reported for Blink and Gecko.

* https://bugs.chromium.org/p/chromium/issues/detail?id=844868
* https://bugzilla.mozilla.org/show_bug.cgi?id=1462916

> The 2dcontext tests I think are generated from tools/tests.yaml or some such. Can the new tests be there also?

OK. I changed `tests2d.yaml`.

zcorpan (approved):

fserb (approved):
lgtm

Issue #7956, opened 2017-10-23
[css-flex][spec-rec] Mix writing modes and flex alignment


gregwhitworth:
Ensure that any of the flex only alignment modes (eg: flex-start/end) work as expected in different writing modes.

gregwhitworth:
And here is an example of how I would test them, in this case using `sidways-lr`
http://jsbin.com/geqigacaqa/1/edit?html,css,js,console,output

claytron5000:
https://drafts.csswg.org/css-flexbox-1/#alignment

PR #10831, opened 2018-05-03
writing modes html creation

claytron5000:
I'm looking for QA on the markup to make sure I'm hitting all variations that are alluded to in #7956. I've included one sample test of what I think of as the base-case.


gregwhitworth:
That looks like it will generate all of the different variations, some of them may be superflous but the more the merrier. Once you have testing logic, let me know.

css/css-flexbox/flexbox_mix-writing-flex-alignment.html
```diff
@@ -0,0 +1,114 @@
...
+    .space-around { justify-content: space-around; }
+    .space-evenly { justify-content: space-evenly; }
+    .flex-end { align-items: flex-end; }
+    .flex-start { align-items: flex-start; }
+    .baseline { align-items: baseline; }
+    .stretch { align-items: stretch; }
+    .center { align-items: center; }
+    .row { flex-direction: row; }
+    .row-reverse { flex-direction: row-reverse; }
+    .column { flex-direction: column; }
```
gregwhitworth: I don't see these ones being applied in the JS below, have you not done that logic yet or am I missing it?

claytron5000: Good catch, thanks.

css/css-flexbox/flexbox_mix-writing-flex-alignment.html
```diff
@@ -0,0 +1,114 @@
...
+    ];
+    for (let index in writingModes) {
+      const writingMode = writingModes[index];
+      const thisWriteMode = document.querySelector('#wrap').appendChild(makeContainer(writingMode));
+      for (let index2 in alignmentProps) {
+        const alignmentProp = alignmentProps[index2];
+        // warning, type massage!
+        if (index2 != (alignmentProps.length - 1)) {
+          // Clone the first node of writing mode
+          const dup = thisWriteMode.cloneNode()
```
benfluleck: Good job on the PR mate really good putting everything all together, sorry what is a dup?

benfluleck:
Pardon my ignorance but why is the cool stuff commented out? or are we not doing it like that @gregwhitworth

gregwhitworth:
@benfluleck cause he already used it to generate the HTML, he's probably wanting to point out that it's not going to be committed. At any rate, this is a work in progress.

css/css-flexbox/flexbox_mix-writing-flex-alignment.html
```diff
@@ -0,0 +1,122 @@
...
...</div></div><div class="container sideways-lr start"><div class="inner">sideways-lr<br>start</div></div><div class="container sideways-lr center"><div class="inner">sideways-lr<br>center</div></div><div class="container sideways-lr space-around"><div class="inner">sideways-lr<br>space-around</div></div><div class="container sideways-lr space-between"><div class="inner">sideways-lr<br>space-between</div></div><div class="container sideways-lr space-evenly"><div class="inner">sideways-lr<br>space-evenly</div></div><div class="container sideways-lr stretch"><div class="inner">sideways-lr<br>stretch</div></div><div class="container sideways-lr center"><div class="inner">sideways-lr<br>center</div></div><div class="container sideways-lr flex-start"><div class="inner">sideways-lr<br>flex-start</div></div><div class="container sideways-lr flex-end"><div class="inner">sideways-lr<br>flex-end</div></div><div class="container sideways-lr row"><div class="inner">sideways-lr<br>row</div></div><div class="container sideways-lr row-reverse"><div class="inner">sideways-lr<br>row-reverse</div></div><div class="container sideways-lr column"><div class="inner">sideways-lr<br>column</div></div></div>
```
gsnedders: Let's not hard-code this generated string; it's long and unwieldy and hard to verify all combinations are tested. I'd much rather just generate this dynamically with the code below.

css/css-flexbox/flexbox_mix-writing-flex-alignment.html
```diff
@@ -0,0 +1,122 @@
...
+  <script>
+    // Generating script.
+    // const writingModes = [
+    //   'horizontal-tb',
+    //   'vertical-rl',
+    //   'vertical-lr',
+    //   'sideways-rl',
+    //   'sideways-lr'
+    // ];
+    // const alignmentProps = [
```
gsnedders: Do we want to split this out even further (into `justify-content`, `align-items`, and `flex-direction`), and test even more potential combinations?

That said, if we're going to have to hard code the expected positions, that probably gets to unreasonably many tests (given it means we'd go from 5*13 to 5*5*5*6)… @gregwhitworth, thoughts?

gsnedders (commented):
A few random thoughts; listen to @gregwhitworth over me though. 🙂

css/css-flexbox/flexbox_mix-writing-flex-alignment.html
claytron5000: reiterating what @gsnedders is saying, maybe it makes more sense to break these into separate tests. Or, just simplify the tests to ones that _might_ actually break. I probably don't know enough to make that determination.

PR #10839, opened 2018-05-04
Add http/README.md

foolip:

### DIFF
--- a/http/README.md
+++ b/http/README.md
@@ -1,0 +1,2 @@
+This directory contains tests for
+[Hypertext Transfer Protocol -- HTTP/1.1](https://tools.ietf.org/html/rfc2616).


foolip:
Reviewers, I assume that this is actually wrong, so please tell which IETF specs should actually be tested in this directory :)

annevk:
The RFCs that replace that RFC: 7230, 7231, 7232, 7233, 7234, 7235. And maybe we put H/2 and such here too once we have the server-side story for that.

mnot:
There are already HTTP tests in /fetch; is the idea to move them here, or...?

foolip:
> There are already HTTP tests in /fetch; is the idea to move them here, or...?

I have no such ambition. I just noticed that this directory exists, and presumably tests something about HTTP :) @mnot, is the list of RFCs @annevk listed right, should I just put that in the README?

annevk:
We should probably stop using `http/` having seen what's in it. Fetch is probably a better place.

foolip:
OK, to just merge the directories, or is some finesse required here?

foolip:
Finesse attempted in https://github.com/w3c/web-platform-tests/pull/10870.

berniefyfe (approved):

PR #10850, opened 2018-05-04
Separate out wpt-specific parts of config from wptserve

jgraham:
I need to land something like this upstream; just creating this PR to ensure that all the tests pass in Travis.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10850)


tools/serve/serve.py
```diff
@@ -665,10 +659,47 @@ def load_config(default_path, override_path=None, **kwargs):
 
 _not_subdomains = {u"nonexistent-origin"}
 
+
 class Config(config.Config):
     """serve config
 
     this subclasses wptserve.config.Config to add serve config options"""
+
+    _default = {
```
gsnedders: This to me seems like a big loss in discoverability compared with having it in the JSON file, as it makes it much harder to know what the format is if you want to modify anything.

gsnedders (commented):
So yeah, my main thought is that this is a UX regression?

tools/serve/serve.py
jgraham: OK, but that's a problem that almost no one has because almost no one modifies this. I think you're right and I"m happy to address it in a followup e.g. by changing the documentation or adding a `config.sample.json` file. However actually using a file would require inventing some substitution syntax to get the right root directory, which seems like even more complexity, which is why I changed things.

gsnedders (approved):

PR #10862, opened 2018-05-07
DOM: fix relatedTarget reset test once more

### LINKED ISSUE #10854: Sort out what "Reset if relatedTarget pointed to a shadow tree pre-dispatch" test in dom/events/relatedTarget.window.js actually should test
I'm changing the test to actually get the listener called in https://bugzilla.mozilla.org/show_bug.cgi?id=1449560 but it isn't clear to me how to change the test to test what it apparently tries to test.

annevk:
The fix in #10859 seems incorrect. In particular the DOM Standard keeps a reference to the original relatedTarget field value so even if it's retargeted it'll still meet the "clearTargets" condition.

Fixes #10854.

### DIFF
--- a/dom/events/relatedTarget.window.js
+++ b/dom/events/relatedTarget.window.js
@@ -39,18 +39,15 @@ async_test(t => {
   t.done();
 }, "Reset if target pointed to a shadow tree pre-dispatch");
 
-async_test(t => {
+test(t => {
   const shadowChild = shadow.appendChild(document.createElement("div"));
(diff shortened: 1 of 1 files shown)


annevk:
@smaug---- can you review?

annevk:
This is wrong because of https://github.com/whatwg/dom/pull/585#issuecomment-376447286 and how clearTargets works. In particular it now looks at event's path and not any original values.

PR #10863, opened 2018-05-07
DOM: retarget relatedTarget

### LINKED ISSUE #10854: Sort out what "Reset if relatedTarget pointed to a shadow tree pre-dispatch" test in dom/events/relatedTarget.window.js actually should test

annevk:
relatedTarget gets retargeted. Also use test rather than async_test if possible.

Fixes #10854.

### DIFF
--- a/dom/events/relatedTarget.window.js
+++ b/dom/events/relatedTarget.window.js
@@ -21,13 +21,13 @@ test(() => {
     for (target of [new XMLHttpRequest(), self, host]) {
       const event = new FocusEvent("demo", { relatedTarget: relatedTarget });
       target.dispatchEvent(event);
-      assert_equals(event.target, null);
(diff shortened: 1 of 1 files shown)


hayatoito:
The change to the test looks good to me.

hayatoito (approved):

annevk:
This is tracked in https://bugzilla.mozilla.org/show_bug.cgi?id=1459584 for Firefox. I'm assuming other browsers haven't updated their dispatch algorithm just yet. Firefox will likely fix this post shadow trees "v1".

annevk:
Thanks for the review!

PR #10865, opened 2018-05-07
HTML: FormData serialization of <input type=file> without selection

annevk:
For https://github.com/whatwg/html/pull/3658.

### DIFF
--- a/html/semantics/forms/form-submission-0/form-data-set-empty-file.window.js
+++ b/html/semantics/forms/form-submission-0/form-data-set-empty-file.window.js
@@ -1,0 +1,14 @@
+promise_test(() => {
+  const form = document.body.appendChild(document.createElement("form")),
+        input = form.appendChild(document.createElement("input"));
+  input.type = "file";
+  input.name = "hi";
+  const fd = new FormData(form),
+        entry = fd.get(input.name);
+  assert_true(entry instanceof File, "entry is a File");
+  assert_equals(entry.name, "", "name");
+  assert_equals(entry.type, "application/octet-stream", "type");
+  return new Response(entry).text().then(body => {
+    assert_equals(body, "", "body");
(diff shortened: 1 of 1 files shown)


annevk:
It would probably be good to also test the server-side bit. I didn't really look into what that would require.

annevk:
Given that Chrome and Safari don't return a type here, should we consider changing the standard? Or is this inconsistent with what they do "over the wire"?

html/semantics/forms/form-submission-0/form-data-set-empty-file.window.js
```diff
@@ -0,0 +1,14 @@
+promise_test(() => {
+  const form = document.body.appendChild(document.createElement("form")),
+        input = form.appendChild(document.createElement("input"));
+  input.type = "file";
+  input.name = "hi";
+  const fd = new FormData(form),
+        entry = fd.get(input.name);
```
tkent-google: nit: I'd like to name this variable "value" rather than "entry" because "entry" is confusing with [entry](https://xhr.spec.whatwg.org/#concept-formdata-entry)

tkent-google (approved):

tkent-google:
> Given that Chrome and Safari don't return a type here, should we consider changing the standard? Or is this inconsistent with what they do "over the wire"?

I think we don't need to change the specification.
It is reasonable.  I confirmed Chrome/Safari sent application/octet-stream over the wire.

annevk:
Bugs:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1459859
* https://bugs.chromium.org/p/chromium/issues/detail?id=840714
* https://bugs.webkit.org/show_bug.cgi?id=185416

No bug for Edge as it doesn't support `get()` on `FormData`.

PR #10867, opened 2018-05-07
Add basic mesh tests.

Tavmjong:
SVG 2 Mesh Gradient reference tests.


longsonr:
shouldn't this wait till at least one browser has indicated they plan to implement this?

ewilligers:
> shouldn't this wait till at least one browser has indicated they plan to implement this?

No, mesh is in a spec so adding tests is legitimate.

If we find (e.g. by running the tests) that mesh doesn't have at least two implementations, with at least one implementation being a browser or high quality polyfill, then mesh would be removed from the SVG 2 Draft or at least marked At Risk. The feature might remain in one of the other drafts at https://svgwg.org/ (like other feature proposals cut from SVG 2), and the tests would remain relevant.

(I have no mesh expertise so I have no comment on the quality of the proposed tests.)

ewilligers:
Can someone from InkScape or Adobe please review this.

svgeesus (approved):
Unfortunate that the ref is a PNG, there may be changes due to antialiasing, but I can't see a way around that.
The tests look correct to me.
The webfont is unused and thus not needed, but is harmless.

PR #10868, opened 2018-05-07
Add basic text 'inline-size' tests.

Tavmjong:
SVG 2 Text 'inline-size' reference tests.


svgeesus (approved):
Tests both absolute and percentage values of inline-size in a variety of writing systems and writing directions. lgtm.

shape-inside interacting with inline-size is not tested, but we can always add more tests for that.

PR #10870, opened 2018-05-07
Move http/ tests into fetch/ and html/

foolip:
As discussed in https://github.com/w3c/web-platform-tests/pull/10839.


foolip:
The moved tests have been run in Chrome and/or Firefox to check that they weren't broken by the move. Some were of course, and fixed. Review with care :)

annevk (approved):
If you indeed ran these and did not observe difference I'm happy with these new locations.

foolip:
Thanks! I couldn't run them on all browsers, but Chrome and Firefox seem to not have regressed compared to the wpt.fyi results, at least.

PR #10878, opened 2018-05-07
Improve definition of non-existent origin

### LINKED ISSUE #10560: `nonexistent-origin` exists in GNU/Linux
[The documentation on WPT server features](https://github.com/w3c/web-platform-tests/blob/0628e2e73489ddd075f273d2555e504e97e28ce6/docs/_writing-tests/server-features.md) states:

> In the test environment, five subdomains are available: `www`, `www1`, `www2`, `天気の良い日`, and `élève`; there is also `nonexistent-origin` which is guaranteed not to resolve.

However, following the instructions for system configuration causes requests to the domain `nonexistent-origin.web-platform.test` to resolve to `web-platform.test`:

    $ curl nonexistent-origin.web-platform.test:8000 -o /dev/null -sv
    * Rebuilt URL to: nonexistent-origin.web-platform.test:8000/
    *   Trying 0.0.0.0...
    * Connected to nonexistent-origin.web-platform.test (127.0.0.1) port 8000 (#0)
    > GET / HTTP/1.1
    > Host: nonexistent-origin.web-platform.test:8000
    > User-Agent: curl/7.47.0
    > Accept: */*
    > 
    < HTTP/1.1 200 OK
    < Content-Type: text/html
    < Server: BaseHTTP/0.3 Python/2.7.12
    < Date: Sat, 21 Apr 2018 00:07:47 GMT
    < Content-Length: 11853
    < 
    { [8192 bytes data]
    * Connection #0 to host nonexistent-origin.web-platform.test left intact

This is because the `./wpt create-hosts-file` command includes the following directive:

    0.0.0.0 nonexistent-origin.web-platform.test

In GNU/Linux, the address 0.0.0.0 is interpreted to mean "any IP address," so the WPT server (bound to 127.0.0.1) responds to requests for this address:

    $ curl 0.0.0.0:8000 -o /dev/null -sv
    * Rebuilt URL to: 0.0.0.0:8000/
    *   Trying 0.0.0.0...
    * Connected to 0.0.0.0 (127.0.0.1) port 8000 (#0)
    > GET / HTTP/1.1
    > Host: 0.0.0.0:8000
    > User-Agent: curl/7.47.0
    > Accept: */*
    > 
    < HTTP/1.1 200 OK
    < Content-Type: text/html
    < Server: BaseHTTP/0.3 Python/2.7.12
    < Date: Sat, 21 Apr 2018 00:44:31 GMT
    < Content-Length: 11853
    < 
    { [8192 bytes data]
    * Connection #0 to host 0.0.0.0 left intact

However, in Windows desktop environments, the same address is interpreted as non-existent:

    PS C:\Users\admin> Invoke-RestMethod -Uri 0.0.0.0:8000
    Invoke-RestMethod : Unable to connect to the remote server
    At line:1 char:1
    + Invoke-RestMethod -Uri 0.0.0.0:8000
    + ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        + CategoryInfo          : InvalidOperation: (System.Net.HttpWebRequest:HttpWebRequest) [Invoke-RestMethod], WebExc
       eption
        + FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShell.Commands.InvokeRestMethodCommand

[The configuration was introduced to facilitate testing from within Windows](https://github.com/w3c/web-platform-tests/pull/2673), but I haven't been able to reproduce the problem it was intended to solve. @domenic: I can change the desktop wallpaper on a Windows machine, but that's about it. Would you mind weighing in on this?

To be clear: the domain is referenced by just 7 tests, and by chance, none of them are invalidated by this behavior in GNU/Linux. I've included some information on each test below, but I'll need someone else to check things out on Windows. I'm also curious to know the situation on macOS; @gsnedders would you mind taking a look?

Two solutions come to mind:

- Remove that entry from the hosts file, assume the domain does not exist, and add tests in `infrastructure/assumptions/`
- Update the `create-hosts-file` command to emit the entry only when run from Windows

<details>
  <summary>Background on the existing references to the domain</summary>

Across all of web-platform-tests, only 11 files reference `nonexistent-origin`:

    $ git grep -l nonexistent-origin .
    acid/acid3/reference.sub.html
    content-security-policy/base-uri/report-uri-does-not-respect-base-uri.sub.html
    docs/_running-tests/chrome_android.md
    docs/_writing-tests/server-features.md
    tools/serve/serve.py
    webrtc/RTCPeerConnection-getIdentityAssertion.html
    webrtc/identity-helper.js
    xhr/event-error-order.sub.html
    xhr/event-error.sub.html
    xhr/send-network-error-async-events.sub.htm
    xhr/send-network-error-sync-events.sub.htm

Of these references, 3 are documentation and 1 is configuration. The remaining 7 are tests:

- `/xhr/event-error-order.sub.html`
  - Intention: produce a failed XHR request
  - Expectation: an HTTP 404 error for resource
  - Actual: an HTTP 404 error for CORS pre-flight request
- `/xhr/event-error.sub.html`
  - Intention: produce a failed XHR request
  - Expectation: an HTTP 404 error for resource
  - Actual: an HTTP 404 error for CORS pre-flight request
- `/xhr/send-network-error-async-events.sub.htm`
  - Intention: produce a failed XHR request
  - Expectation: an HTTP 404 error for resource
  - Actual: an HTTP 404 error for CORS pre-flight request
- `/xhr/send-network-error-sync-events.sub.htm`
  - Intention: produce a failed XHR request
  - Expectation: an HTTP 404 error for resource (because resource is not found)
  - Actual: an HTTP 404 error for CORS pre-flight (because method is not
    available\*)
- `/acid/acid3/reference.sub.html`
  - Intention: unclear
  - Expectation: unclear
  - Actual: a favicon is not assigned to the reference document
- `/webrtc/RTCPeerConnection-getIdentityAssertion.html`
  - Intention: Promise rejected with RTCError
  - Expectation: Promise rejected with RTCError
  - Actual: Test failure--RTCError is not defined in any implementation
- `/content-security-policy/base-uri/report-uri-does-not-respect-base-uri.sub.html`
  - Intention: ensure that the base URL is not referenced
  - Expectation: tests will fail if they make requests to a non-existent domain
  - Actual: tests may pass if they make requests to the domain

In other words: only the final test is impacted by this bug. For the other six tests, the bug violates the authors' expectations, but extenuating circumstances satisfy the underlying criteria.

In the case of the final test, the domain is being used to implicitly assert that some behavior (CSP report redirection) is *not* occurring. The bug subverts that check. This does not make the test invalid; it only makes it weaker than intended.

\* - the WPT server should probably respond with 405 in this case

</details>

### LINKED ISSUE #10754: nonexistent-origin is confusing
Presumably this is a domain, so `nonexistent-domain` or `nonexistent-host` would be more accurate. Could we rename this?

Can one use the subdomains here too?

(` docs/_writing-tests/server-features.md`)

jugglinmike:
The first commit on this branch is intended to resolve gh-10560. The second is intended to resolve gh-10754.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10878)

### DIFF
--- a/docs/_running-tests/chrome_android.md
+++ b/docs/_running-tests/chrome_android.md
@@ -19,7 +19,6 @@ Until we find a better way, we need to root the Android device and update the
 127.0.0.1   www2.web-platform.test
 127.0.0.1   xn--n8j6ds53lwwkrqhv28a.web-platform.test
 127.0.0.1   xn--lve-6lad.web-platform.test
-0.0.0.0     nonexistent-origin.web-platform.test
 ```
 
 ## CA certificate
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -7,6 +7,7 @@ import argparse
 import json
 import logging
 import os
+import platform
 import socket
 import sys
 import threading
@@ -447,8 +448,17 @@ def make_hosts_file(config, host):
     for domain in config.domains_set:
         rv.append("%s\t%s\n" % (host, domain))
 
-    for not_domain in config.not_domains_set:
-        rv.append("0.0.0.0\t%s\n" % not_domain)
+    # Windows interpets the IP address 0.0.0.0 as "non-existent, making it an
+    # appropriate alias for non-existent hosts. However, UNIX-like systems
+    # interpret the same address to mean "any IP address," which is
+    # inappropraite for this context. These syystems do not reserve any value
+    # for this purpose, so the inavailability of the domains must be taken for
+    # granted.
+    #
+    # https://github.com/w3c/web-platform-tests/issues/10560
+    if platform.uname()[0] == "Windows":
+        for not_domain in config.not_domains_set:
+            rv.append("0.0.0.0\t%s\n" % not_domain)
 
     return "".join(rv)
 
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -448,12 +448,11 @@ def make_hosts_file(config, host):
     for domain in config.domains_set:
         rv.append("%s\t%s\n" % (host, domain))
 
-    # Windows interpets the IP address 0.0.0.0 as "non-existent, making it an
+    # Windows interpets the IP address 0.0.0.0 as non-existent, making it an
     # appropriate alias for non-existent hosts. However, UNIX-like systems
-    # interpret the same address to mean "any IP address," which is
-    # inappropraite for this context. These syystems do not reserve any value
-    # for this purpose, so the inavailability of the domains must be taken for
-    # granted.
+    # interpret the same address to mean any IP address, which is inappropraite
+    # for this context. These systems do not reserve any value for this
+    # purpose, so the inavailability of the domains must be taken for granted.
     #
     # https://github.com/w3c/web-platform-tests/issues/10560
     if platform.uname()[0] == "Windows":
@@ -666,7 +665,7 @@ _subdomains = {u"www",
                u"天気の良い日",
                u"élève"}
 
-_not_subdomains = {u"nonexistent-origin"}
+_not_subdomains = {u"nonexistent"}
 
 
 class Config(config.Config):
--- a/tools/serve/test_serve.py
+++ b/tools/serve/test_serve.py
@@ -1,4 +1,5 @@
 import pickle
+import platform
 import os
 
 import pytest
@@ -8,7 +9,30 @@ from . import serve
 from .serve import Config
 
 
-def test_make_hosts_file():
+@pytest.mark.skipif(platform.uname()[0] == "Windows",
+                    reason="Expected contents are platform-dependent")
+def test_make_hosts_file_nix():
+    c = Config(browser_host="foo.bar", alternate_hosts={"alt": "foo2.bar"})
+    hosts = serve.make_hosts_file(c, "192.168.42.42")
+    lines = hosts.split("\n")
+    assert set(lines) == {"",
+                          "192.168.42.42\tfoo.bar",
+                          "192.168.42.42\tfoo2.bar",
+                          "192.168.42.42\twww.foo.bar",
+                          "192.168.42.42\twww.foo2.bar",
+                          "192.168.42.42\twww1.foo.bar",
+                          "192.168.42.42\twww1.foo2.bar",
+                          "192.168.42.42\twww2.foo.bar",
+                          "192.168.42.42\twww2.foo2.bar",
+                          "192.168.42.42\txn--lve-6lad.foo.bar",
+                          "192.168.42.42\txn--lve-6lad.foo2.bar",
+                          "192.168.42.42\txn--n8j6ds53lwwkrqhv28a.foo.bar",
+                          "192.168.42.42\txn--n8j6ds53lwwkrqhv28a.foo2.bar"}
+    assert lines[-1] == ""
+
+@pytest.mark.skipif(platform.uname()[0] != "Windows",
+                    reason="Expected contents are platform-dependent")
+def test_make_hosts_file_windows():
     c = Config(browser_host="foo.bar", alternate_hosts={"alt": "foo2.bar"})
     hosts = serve.make_hosts_file(c, "192.168.42.42")
     lines = hosts.split("\n")
--- a/tools/serve/test_serve.py
+++ b/tools/serve/test_serve.py
@@ -37,8 +37,8 @@ def test_make_hosts_file_windows():
     hosts = serve.make_hosts_file(c, "192.168.42.42")
     lines = hosts.split("\n")
     assert set(lines) == {"",
-                          "0.0.0.0\tnonexistent-origin.foo.bar",
-                          "0.0.0.0\tnonexistent-origin.foo2.bar",
+                          "0.0.0.0\tnonexistent.foo.bar",
+                          "0.0.0.0\tnonexistent.foo2.bar",
                           "192.168.42.42\tfoo.bar",
                           "192.168.42.42\tfoo2.bar",
                           "192.168.42.42\twww.foo.bar",
--- a/acid/acid3/reference.sub.html
+++ b/acid/acid3/reference.sub.html
@@ -1,7 +1,7 @@
 <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0//EN">
 <html>
  <title>The Acid3 Test (Reference Rendering)</title>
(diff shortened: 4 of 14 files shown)


annevk:
Given we use it as subdomain prefix, maybe we should just call it "nonexistent"? Domain gives the impression you can use it standalone.

jugglinmike:
Yeah, that's probably better. Since that's different than what @gsnedders originally agreed to, I'd like to avoid churn and get their feedback before making the change. What do you say, Geoffrey?

gsnedders:
Ask others too, but that sounds good to me.

jugglinmike:
@gsnedders Pitched the idea to @jgraham, and he was over the moon. Normally I'd make the change with a "fixup!" commit, but recent changes to `master` introduced conflicts, so I've rebased and squashed. Mind taking another look?

jgraham (approved):

tools/serve/serve.py
```diff
@@ -447,8 +448,17 @@ def make_hosts_file(config, host):
     for domain in config.domains_set:
         rv.append("%s\t%s\n" % (host, domain))
 
-    for not_domain in config.not_domains_set:
-        rv.append("0.0.0.0\t%s\n" % not_domain)
+    # Windows interpets the IP address 0.0.0.0 as "non-existent, making it an
```
gsnedders: Needs a closing double quote. (Also, where's the quote from? I can't find this documented anywhere?)

jugglinmike: I've added the double quote in a fixup! commit. I'll respond to your question in the main thread so the conversation doesn't get lost when GitHub collapses this comment.

jugglinmike:
> (Also, where's the quote from? I can't find this documented anywhere?)

That's not cited because it's not a quotation. I haven't seen a particularly authoritative source describing this behavior; the explanation is a result of the independent research reported at the referenced GitHub issue and a smattering of articles/blog posts like https://www.lifewire.com/four-zero-ip-address-818384 I can add a link to that article if you like.

gsnedders:
@jugglinmike just the quotation marks make it look like it is a quote from somewhere; if there's nowhere authoritative I'd say just drop the quotation marks

jugglinmike:
Oh, now I understand. I've removed them and fixed an unrelated spelling error (from "syystem" to "system")

docs/_running-tests/chrome_android.md
```diff
@@ -19,7 +19,7 @@ Until we find a better way, we need to root the Android device and update the
 127.0.0.1   www2.web-platform.test
 127.0.0.1   xn--n8j6ds53lwwkrqhv28a.web-platform.test
 127.0.0.1   xn--lve-6lad.web-platform.test
-0.0.0.0     nonexistent-origin.web-platform.test
+0.0.0.0     nonexistent.web-platform.test
```
gsnedders: Presumably as with all other Linux systems, this should be dropped.

tools/serve/test_serve.py
```diff
@@ -8,13 +9,36 @@
 from .serve import Config
 
 
-def test_make_hosts_file():
+@pytest.mark.skipif(platform.uname()[0] == "Windows",
+                    reason="Expected contents are platform-dependent")
+def test_make_hosts_file_nix():
```
gsnedders: I think I'd rather just have one function here, with something like:

```
expected = {"",
                          "192.168.42.42\tfoo.bar",
                          "192.168.42.42\tfoo2.bar",
                          "192.168.42.42\twww.foo.bar",
                          "192.168.42.42\twww.foo2.bar",
                          "192.168.42.42\twww1.foo.bar",
                          "192.168.42.42\twww1.foo2.bar",
                          "192.168.42.42\twww2.foo.bar",
                          "192.168.42.42\twww2.foo2.bar",
                          "192.168.42.42\txn--lve-6lad.foo.bar",
                          "192.168.42.42\txn--lve-6lad.foo2.bar",
                          "192.168.42.42\txn--n8j6ds53lwwkrqhv28a.foo.bar",
                          "192.168.42.42\txn--n8j6ds53lwwkrqhv28a.foo2.bar"}
if platform.uname()[0] != "Windows":
    expected ^= {"0.0.0.0\tnonexistent-origin.foo.bar",
                           "0.0.0.0\tnonexistent-origin.foo2.bar"}
```

(except, like, don't have the whitespace a mess from copy/pasting stuff in GitHub :))

docs/_running-tests/chrome_android.md
jugglinmike: I haven't verified any of this from an Android system, so I don't feel comfortable changing the instructions. But if I really want to resolve this issue, then that is my responsibility. I'll work on running WPT there, and I'll let you know what I find.

tools/serve/test_serve.py
jugglinmike: I considered that, but I chose this organization to simplify the test's semantics. My preference is for tests' expectations to be static so that passing/failing always means the same thing. This has introduced some duplication, though. Is that why you're recommending the tests be consolidated?  Would you be satisfied if we kept both tests but defined the second dict in terms of the first *outside* of the test bodies?

jugglinmike:
Following a failed attempt to root an Android device, I've asked @Hexcles if
he could verify this behavior for me. Here are the steps we're using to check
the behavior:

1. Add the following entry to the `/etc/hosts` file on the Android device:

       0.0.0.0 localhost
2. Run the following command from your development machine:

       adb reverse tcp:8000 tcp:8000
3. Run a web server bound to port 8000 on your development machine, e.g.

       python -m SimpleHTTPServer 8000
4. Visit http://0.0.0.0:8000 from the Android device

We're curious to know if that resolves to the web server on your development machine or if it fails to resolve at all

jugglinmike:
Speaking on IRC, @hexcles reported:

> (01:44:49 PM) Hexcles: Tl;dr. Chrome on Android behaves the same as Linux
> regarding 0.0.0.0.

So that settles it. I've updated the documentation for Android devices by removing that entry from the recommended hosts file modification. Since this change needs to happen in the first commit but interferes with the second, I've rebased this branch to make merging easier. The original version of the branch is available here:

https://github.com/w3c/web-platform-tests/compare/master...bocoup:non-existent-orig

@gsnedders would you mind taking another look?

gsnedders (approved):

jugglinmike:
Thanks, @gsnedders!

PR #10879, opened 2018-05-07
Fix DOM interface test

dstorey:
...by removing SVGElement. This should be included in the SVG interface test (SVG needs further updates that will be handled separately)

Fixes https://w3c-test.org/dom/interfaces.html


dstorey:
CC @annevk

annevk (approved):

Issue #10716, opened 2018-04-30
Testdriver calls cannot return any values


kereliuk:
Currently there is no capability in the infrastructure for testdriver to return any information after the call.

For instance, if we want to use testdriver to get the current window size, there is no way to return that information to the testharness.

jgraham:
I agree that could be a problem, but the motivating example isn't very covincing. In particular, why is a test changing the browser window size? What's the use case that you're trying to solve?

gsnedders:
From IRC:


15:05 < gsnedders> kereliuk: what's the plan with https://github.com/w3c/web-platform-tests/issues/10716?
15:30 < kereliuk> gsnedders: I have something on a local branch that works at least for the window commands I've sent PRs for
15:31 < kereliuk> I have the protocol method return a python thing. Serialize that with json.dumps and throw it in the completion message.
15:31 < kereliuk> Then use JSON.parse to put it back into the JS object and attach it to the pending_resolve
15:31 < kereliuk> WYT?

Will respond below…

gsnedders:
@jgraham I don't know about size, but certainly other things related to window management are necessary for testing things like the Page Visibility API.

gsnedders:
@kereliuk yeah, that probably seems as reasonable as anything else (sorry, I got distracted by getting off the train a week ago!)

foolip:
@kereliuk, is this still needed for some API? Is it blocking https://github.com/web-platform-tests/wpt/pull/10749 and https://github.com/web-platform-tests/wpt/pull/10747?

foolip:
Demoting this to backlog as it's not clear which APIs we need it for. Existing PR is at https://github.com/web-platform-tests/wpt/pull/11080 and can be revived when needed for some other work.

LukeZielinski:
A use-case recently came up in the context of WebAuthN - to be able to inspect the credentials that exist on a test device and that authenticating with the wrong key fails.

Still backlog priority, will revisit once such tests are closer to reality.

PR #10882, opened 2018-05-08
Documentation/Tutorial for adding a command to testdriver

kereliuk:
This PR is a markdown file that contains a tutorial of how to add a testdriver API.

The example used here is a command that doesn't exist in our API (set window rect) so people can see it in action as a toy example.

### DIFF
--- a/docs/_writing-tests/testdriver-tutorial.md
+++ b/docs/_writing-tests/testdriver-tutorial.md
@@ -1,0 +1,290 @@
+# Adding new commands to testdriver.js
+ 
+## Assumptions
+We assume the following in this writeup:
+ - You know what web-platform-tests is and you have a working checkout and can run tests
+ - You know what WebDriver or Selenium is
+ - Familiarity with javascript and python
+ 
+## Code!
+ 
+Lets implement window resizing. We can do this via the [Set Window Rect](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-set-window-rect) command in WebDriver.
+ 
+First, we need to think of what the API will look like a little. We will be using selenium and marionette for this, so we can look and see that they take in x, y coordinates, width and height integers.
+ 
+The first part of this will be browser agnostic, but later we will need to implement a specific layer for each browser (here we will do Firefox and Chrome).
+ 
+## Code!
+
+### ```resources/testdriver.js```
+ 
+This is the main entry point the tests get. Here we need to add a function to the test_driver object that will call the test_driver_internal object.
+ 
+```javascript
+window.test_driver = {
+ 
+    {... other commands},
+ 
+    /**
+    * Triggers browser window to be resized and relocated
+    *
+    * This matches the behaviour of the {@link
+    * https://w3c.github.io/webdriver/webdriver-spec.html#dfn-set-window-rect|WebDriver
+    * Set Window Rect command}.
+    *
+    * @param {Integer} x - The x coordinate of the top left of the window
+    * @param {Integer} y - The x coordinate of the top left of the window
+    * @param {Integer} width - The width of the window
+    * @param {Integer} height - The width of the window
+    * @returns {Promise} fulfilled after window rect is set occurs, or rejected in
+    *                    the cases the WebDriver command errors
+    */
+    set_window_rect: function(x, y, width, height) {
+        return window.test_driver_internal.set_element_rect(x, y, width, height);
+    }
+```
+ 
+In the same file, lets add to the internal object. ( do we need to do this?) (make sure to do this if the internal call has different arguments than the external call, especially if it calls multiple internal calls)
+ 
+```javascript
+window.test_driver_internal = {
+ 
+    {... other commands},
+ 
+    /**
+     * Triggers browser window to be resized and relocated
+     *
+     * This matches the behaviour of the {@link
+     * https://w3c.github.io/webdriver/webdriver-spec.html#dfn-set-window-rect|WebDriver
+     * Set Window Rect command}.
+     *
+     * @param {Integer} x - The x coordinate of the top left of the window
+     * @param {Integer} y - The x coordinate of the top left of the window
+     * @param {Integer} width - The width of the window
+     * @param {Integer} height - The width of the window
+     * @returns {Promise} fulfilled after window rect is set occurs, or rejected in
+     *                    the cases the WebDriver command errors
+     */
+    set_window_rect: function(x, y, width, height) {
+        return Promise.reject(new Error("unimplemented"))
+    }
+```
+We will leave this unimplemented and override it in another file. Lets do that now!
+ 
+### ```tools/wptrunner/wptrunner/testdriver-extra.js```
+ 
+This will be the default function called when invoking the test driver commands (sometimes it is overridden by testdriver-vendor.js, but this is outside the scope of this writeup).
+ 
+```javascript
+window.test_driver_internal.set_element_rect = function(x, y, width, height) {
+    const pending_promise = new Promise(function(resolve, reject) {
+        pending_resolve = resolve;
+        pending_reject = reject;
+    });
+    window.opener.postMessage(
+        {"type": "action", "action": "set_window_rect", "x": x, "y": y, "width": width, "height": height}, "*");
+    return pending_promise;
+};
+```
+The main thing here is the postMessage argument. The first argument is a json object with properties
+ - ```type```: this always has to be the string ```"action"```
+ - ```action```: the name of the testdriver command this defines (in this case, set_window_rect)
+ - any other things you want to pass to the next point of execution (in this case, the x, y coordinates and the width and height)
+
+ The pending promise needs to be there as it is resolved when the window recieves a completion message from the executor.
+
+ Next, this is passed to the executor and protocol in wptrunner. Time to switch to python!
+ 
+```tools/wptrunner/wptrunner/executors/protocol.py```
+ 
+```python
+class SetWindowRectProtocolPart(ProtocolPart):
+    """Protocol part for resizing and changing location of window"""
+    __metaclass__ = ABCMeta
+ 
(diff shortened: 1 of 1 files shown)


kereliuk:
There is a preview [here](https://kereliuk.github.io/draft/).

kereliuk:
@gsnedders @foolip r? :)

docs/_writing-tests/testdriver-tutorial.md
```diff
@@ -0,0 +1,289 @@
...
+## Code!
+
+### ```resources/testdriver.js```
+
+This is the main entry point the tests get. Here we need to add a function to the test_driver object that will call the test_driver_internal object.
+
+```javascript
+window.test_driver = {
+
+    {... other commands},
```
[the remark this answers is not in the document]
foolip: I can edit :)

docs/_writing-tests/testdriver-tutorial.md
```diff
@@ -0,0 +1,289 @@
...
+
+Let's implement window resizing. We can do this via the [Set Window Rect](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-set-window-rect) command in WebDriver.
+
+First, we need to think of what the API will look like a little. We will be using Selenium and Marionette for this, so we can look and see that they take in x, y coordinates, width and height integers.
+
+The first part of this will be browser agnostic, but later we will need to implement a specific layer for each browser (here we will do Firefox and Chrome).
+
+## Code!
+
+### ```resources/testdriver.js```
```
foolip: Is it possible to link the files in the heading? Links would be useful, so if not linking them below would be cool.

docs/_writing-tests/testdriver-tutorial.md
```diff
@@ -0,0 +1,289 @@
...
+     */
+    set_window_rect: function(x, y, width, height) {
+        return Promise.reject(new Error("unimplemented"))
+    }
+```
+We will leave this unimplemented and override it in another file. Lets do that now!
+
+### ```tools/wptrunner/wptrunner/testdriver-extra.js```
+
+This will be the default function called when invoking the test driver commands (sometimes it is overridden by testdriver-vendor.js, but this is outside the scope of this writeup).
```
foolip: As a reader, here I wonder why the default implementation should always go here, or if there always needs to be a default implementation. Will there be any cases where `test_driver_internal` has a default implementation that builds on other commands, which can still be overridden? I'm thinking clipboard and focus manipulation.

docs/_writing-tests/testdriver-tutorial.md
```diff
@@ -0,0 +1,289 @@
...
+        {"type": "action", "action": "set_window_rect", "x": x, "y": y, "width": width, "height": height}, "*");
+    return pending_promise;
+};
+```
+The main thing here is the `postMessage` argument. The first argument is an object with properties
+ - `type`: this always has to be the string `"action"`
+ - `action`: the name of the testdriver command this defines (in this case, `set_window_rect`)
+ - any other things you want to pass to the next point of execution (in this case, the x, y coordinates and the width and height)
+
+The pending promise needs to be there as it is resolved when the window recieves a completion message from the executor.
```
foolip: This and the code above wasn't enough for me to understand. The key is that `pending_resolve` and `pending_reject` are in an outer scope, and implicitly there can only be one ongoing command at the same time or strange bugs will happen.

Can you document this restriction somewhere, and maybe make things blow up sooner if it's violated? And maybe the code to create `pending_promise` can just be a helper function if it has to be the same in every single command?

docs/_writing-tests/testdriver-tutorial.md
```diff
@@ -0,0 +1,289 @@
...
+                  SeleniumTestDriverProtocolPart]
+```
+
+
+### Firefox
+We use the [set window rect](http://marionette-client.readthedocs.io/en/master/reference.html#marionette_driver.marionette.Marionette.set_window_rect) Marionette command.
+
+We will use executormarionette and use the Marionette Python API.
+
+We have little actual work to do here! We just need to define a subclass of the protocol part we defined earlier.
```
foolip: Using the exact same phrasing again is a bit odd in this conversational style. Change some words?

docs/_writing-tests/testdriver-tutorial.md
```diff
@@ -0,0 +1,289 @@
...
+<meta charset="utf-8">
+<title>TestDriver set window rect method</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/resources/testdriver.js"></script>
+<script src="/resources/testdriver-vendor.js"></script>
+
+<script>
+async_test(t => {
+  test_driver
```
foolip: Can you make this a `promise_test`? May even use async/await. Should be a one-liner plus a comment, people will be impressed by the simplicity of it :)

foolip (commented):
Now finished reviewing, this is great stuff! I fixed my own nitpickery to save you the trouble, the remaining comments are ones that might need discussion.

Thank you!

kereliuk:
@Honry This might be helpful for adding testdriver.js APIs for sensors down the road :) Do you have any suggestions?

docs/_writing-tests/testdriver-tutorial.md
```diff
@@ -0,0 +1,289 @@
...
+
+## Code!
+
+Let's implement window resizing. We can do this via the [Set Window Rect](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-set-window-rect) command in WebDriver.
+
+First, we need to think of what the API will look like a little. We will be using Selenium and Marionette for this, so we can look and see that they take in x, y coordinates, width and height integers.
+
+The first part of this will be browser agnostic, but later we will need to implement a specific layer for each browser (here we will do Firefox and Chrome).
+
+## Code!
```
Honry: Two "Code!" sections, a bit confused.

docs/_writing-tests/testdriver-tutorial.md
```diff
@@ -0,0 +1,289 @@
...
+Here is some template code!
+
+```html
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>TestDriver set window rect method</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/resources/testdriver.js"></script>
+<script src="/resources/testdriver-vendor.js"></script>
```
Honry: Is that possible to give a brief introduction to `resources/testdriver-vendor.js`?

Honry:
@kereliuk, a beautiful tutorial! 👍 

testdriver.js APIs for sensors also require return values, similar to #10716, it'd be nice to add example code once we have an instance.

kereliuk:
@Honry 

> testdriver.js APIs for sensors also require return values, similar to #10716, it'd be nice to add example code once we have an instance.

Good suggestion. I'll add a note saying I plan to add this in the future.

docs/_writing-tests/testdriver-tutorial.md
kereliuk: Done.

docs/_writing-tests/testdriver-tutorial.md
kereliuk: Here is the bug for adding the code improvements: #10990
I've assigned it to myself.

docs/_writing-tests/testdriver-tutorial.md
```diff
@@ -16,14 +16,14 @@ The first part of this will be browser agnostic, but later we will need to imple
 
 ## Code!
 
-### ```resources/testdriver.js```
+### [resources/testdriver.js](https://github.com/w3c/web-platform-tests/blob/master/resources/testdriver.js)
```
foolip: I think just "resources/testdriver.js" or possible "/resources/testdriver.js" works here to link to the file in the same repo and revision. Will save some renaming after the repo move.

docs/_writing-tests/testdriver-tutorial.md
```diff
@@ -86,17 +86,43 @@ window.test_driver_internal.set_element_rect = function(x, y, width, height) {
...
+        if (data.status === "success") {
+            pending_resolve();
+        } else {
+            pending_reject();
+        }
+    });
+```
 
-```tools/wptrunner/wptrunner/executors/protocol.py```
+One limitation this introduces is that only one testdriver call can be made at one time since the `pending_resolve` and `pending_reject` variables are in abn outer scope.
```
foolip: Typo: abn

docs/_writing-tests/testdriver-tutorial.md
```diff
@@ -276,14 +302,18 @@ Here is some template code!
...
-  test_driver
-    .set_window_rect(100, 100, 100, 100) // watch the window resize!
-    .then(() => {
-        // do something!
-        t.done());
-    }
-    .catch(t.unreached_func("set window rect failed"));
-});
+promise_test(async t => {
+  await test_driver.set_window_rect(100, 100, 100, 100);
```
foolip: Sweet :)

docs/_writing-tests/testdriver-tutorial.md
```diff
@@ -276,14 +302,18 @@ Here is some template code!
...
+  // do something
+}
 </script>
 ```
+### What about testdriver-vendor.js?
+
+The file [testdriver-vendor.js](/Users/kereliuk/web-platform-tests/resources/testdriver-vendor.js) is the equivalent to testdriver-extra.js above, except is
```
foolip: Oops: /Users/kereliuk/web-platform-tests/resources/testdriver-vendor.js

foolip (approved):

docs/_writing-tests/testdriver-tutorial.md
kereliuk: Whooops

Issue #4934, opened 2017-02-21
Clean up html/infrastructure/urls/resolving-urls/query-encoding/


zcorpan:
* Need to file browser bugs for failing tests.
* Probably should split tests into smaller chunks for each file.
* running query-encoding/windows-1251.html I get numerous Python errors
* Check if there are problems with stability in test results

cc @annevk

annevk:
This blocks #552 basically.

annevk:
https://bugzilla.mozilla.org/show_bug.cgi?id=1427268 per location.search (not other location setters) is different per browsers.

inexorabletash:
WDYT about splitting these up using variant?
```html
<meta name="variant" content="?include=nested-browsing">
<meta name="variant" content="?include=loading">
<meta name="variant" content="?include=submit">
<meta name="variant" content="?include=base-href">
<meta name="variant" content="?include=workers">
<meta name="variant" content="?include=eventsource">
<meta name="variant" content="?include=xmldocument">
<meta name="variant" content="?include=window-open">
<meta name="variant" content="?include=hyperlink-search">
<meta name="variant" content="?include=history">
<meta name="variant" content="?include=svg">
<meta name="variant" content="?include=xhr">
<meta name="variant" content="?include=websocket">
<meta name="variant" content="?include=appcache">
<meta name="variant" content="?include=css">
<meta name="variant" content="?include=xml">
<meta name="variant" content="?include=url">
<meta name="variant" content="?include=scheme">
```
This would apply to the utf-8, utf-16be, utf-16le, windows-1251 and windows-1252 tests. Perhaps not as elegant as splitting things up into multiple files, but can get this landed pretty quickly via Chrome.

annevk:
I think that's probably fine and better than the status quo.

inexorabletash:
Per wpt.fyi on the PR, looking for consistent failures across Chrome/Firefox/Safari and encodings:

* appcache - TIMEOUTs 2/3 (CACHE and FALLBACK, NETWORK is ok)
* history - fails 2/2
* xmldocument - fails 1/1 (XMLDocument#load is gone, should just remove these)

And FWIW, only these pass everywhere (all browsers/all encodings)
* submit
* url
* websocket

annevk:
FWIW, dom/historical.html covers `load` removal already so that can indeed be removed.

annevk:
I thought this was fixed due to the splitting, but #14519 suggests this was still a problem after the fix got merged.

[...]

PR #10891, opened 2018-05-08
Extract Location object tests from query-encoding/

annevk:
Helps with #4934.

### DIFF
--- a/html/infrastructure/urls/resolving-urls/query-encoding/location.sub.html
+++ b/html/infrastructure/urls/resolving-urls/query-encoding/location.sub.html
@@ -1,0 +1,44 @@
+<!doctype html>
+<meta charset={{GET[encoding]}}> <!-- ends up as <meta charset> by default which is windows-1252 -->
+<meta name=variant content="?encoding=x-cp1251">
+<meta name=variant content="?encoding=utf8">
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<div id=log></div>
+<script>
+function expected(encoding) {
+  return "?" + {
+    "UTF-8": "%C3%BF",
+    "windows-1251": "&%23255;",
+    "windows-1252": "%FF"
+  }[encoding];
+}
+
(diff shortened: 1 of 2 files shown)


annevk:
Bugs:
* https://bugs.webkit.org/show_bug.cgi?id=185419
* https://bugs.chromium.org/p/chromium/issues/detail?id=840760

annevk:
@hsivonen this relies on defaulting to windows-1252. Is that a reasonable thing to do in a test suite?

hsivonen:
I think it's reasonable enough. Ideally, the instance of the test suite available to ad hoc running by users would be hosted on a non-com/org/net domain that's also not affiliated with an non-windows-1252 encoding. E.g. a `.io` (of to avoid `.io`, any of the new vanity TLDs) would work better than the current `.org`.

html/infrastructure/urls/resolving-urls/query-encoding/location.sub.html
```diff
@@ -0,0 +1,44 @@
...
+<meta name=variant content="?encoding=x-cp1251">
+<meta name=variant content="?encoding=utf8">
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<div id=log></div>
+<script>
+function expected(encoding) {
+  return "?" + {
+    "UTF-8": "%C3%BF",
+    "windows-1251": "&%23255;",
```
annevk: Either change this to align with https://github.com/whatwg/url/pull/386 before landing or rebase #10915 and change it there.

foolip:
Interesting, I came across this part of the test suite in https://github.com/web-platform-tests/wpt.fyi/issues/132 by a completely different way of looking for problems. @zcorpan FYI.

foolip (approved):
LGTM assuming the tests still work :)

annevk:
@zcorpan could you quickly skim this as well?

zcorpan:
Looks good to me.

PR #10892, opened 2018-05-08
add url host test: Port number is removed if new port is scheme default

AyushG3112:
This commit adds a test for a bug found with the `node.js` implementation of URL, which was not covered by the existing test cases.

Refs: https://github.com/nodejs/node/pull/20479

### DIFF
--- a/url/setters_tests.json
+++ b/url/setters_tests.json
@@ -968,6 +968,17 @@
                 "hostname": "test",
                 "port": "12"
             }
+        },
+        {
+            "comment": "Port number is removed if new port is scheme default",
+            "href": "http://example.net:8080",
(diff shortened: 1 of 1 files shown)


annevk:
Ah, we do have "Default port number is removed" tests, but they do not have a non-default port themselves (at least the host setter variants).

annevk:
Could you maybe locate this test closer to those and indicate how it's different from them? I think that'd make things easier for maintainers.

AyushG3112:
@annevk sure, but just to confirm, is the test `comment` alright does it need changing too?

annevk (approved):
Thanks, this works for me.

PR #10912, opened 2018-05-09
PaymentRequest "paymentmethodchanged" event

marcoscaceres:
Tests for https://github.com/w3c/payment-request/pull/695


payment-request/onpaymentmenthodchange-attribute.https.html
```diff
@@ -0,0 +1,69 @@
+<!DOCTYPE html>
+<!-- Copyright © 2017 Chromium authors and World Wide Web Consortium, (Massachusetts Institute of Technology, ERCIM, Keio University, Beihang). -->
```
mnoorenberghe: I've noticed you remove this comment before and it's not included on the other two tests.

payment-request/onpaymentmenthodchange-attribute.https.html
```diff
@@ -0,0 +1,69 @@
...
+test(() => {
+  const request = new PaymentRequest(defaultMethods, defaultDetails);
+  const ev = new Event("paymentmethodchange");
+  let didHandle = false;
+  request.onpaymentmethodchange = evt => {
+    assert_equals(ev, evt, "must be same event");
+    didHandle = true;
+  };
+  request.dispatchEvent(ev);
+  assert_true(didHandle, "event did not fire");
```
mnoorenberghe: Nit: I thought generally the assertion message should say what should be true (e.g. "must be same event") but for many you are indicating what doesn't happen if it fails. 

e.g. this should be "event fired"

mnoorenberghe (approved):

payment-request/onpaymentmenthodchange-attribute.https.html
marcoscaceres: Yeah, it's a bit all over the place and a matter of preference. The messages only show up when tests fail, so I tend to explain why something failed rather than why something passed. So:

Why did X fail? ..."event did not fire".

marcoscaceres:
Blocking merge, waiting on ok from implementers on the spec side.

marcoscaceres:
updated tests to match updated proposal.

lint.whitelist
```diff
@@ -326,6 +326,7 @@ OPEN-NO-MODE: css/*
 PRINT STATEMENT: css/*
 CONTENT-VISUAL: css/*
 CONTENT-MANUAL: css/*
+CONTENT-VISUAL: payment-request/PaymentMethodChangeEvent/methodName-attribute.https.html
```
gsnedders: This seems wrong. Why is it getting picked up as a visual test?

payment-request/PaymentMethodChangeEvent/methodName-attribute.https.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Test for PaymentMethodChangeEvent.methodName attribute</title>
+<link rel="help" href="https://w3c.github.io/browser-payment-api/#dom-paymentmethodchangeevent-src">
+<script="/resources/testharness.js"></script>
```
gsnedders: This needs to be `<script src=`

gsnedders (dismissed):

payment-request/PaymentMethodChangeEvent/methodName-attribute.https.html
marcoscaceres: oh, woopsy...

lint.whitelist
marcoscaceres: agree... filed #11479 because this is not documented.

marcoscaceres: (reverting this, btw)

marcoscaceres:
@gsnedders, take another look.

PR #10917, opened 2018-05-09
URL: remove invalid base URLs from urltestdata.json

### LINKED ISSUE #8707: Some a-element.html URL tests seem wrong
Scroll to the bottom of https://wpt.fyi/url/a-element.html and look at the "assert_unreached" instances that fail in all browsers. It appears the test is trying to say that parsing `test-a.html` against `a` should fail, but I believe that given how the a-element.html test runner is set up, that won't be the case; it'll just use the actual page URL as the base URL.

annevk:
As per the README.md they're not allowed there.

Fixes #8707.

### DIFF
--- a/url/failure.html
+++ b/url/failure.html
@@ -16,7 +16,11 @@ function runTests(testData) {
 
     const name = test.input + " should throw"
 
-    self.test(() => { // new URL itself is already tested by url-constructor.html
+    self.test(() => { // URL's constructor's first argument is tested by url-constructor.html
+      assert_throws(new TypeError(), () => new URL("about:blank", test.input));
+    }, "URL's constructor's base argument: " + name)
+
+    self.test(() => {
       const url = new URL("about:blank")
       assert_throws(new TypeError, () => url.href = test.input)
     }, "URL's href: " + name)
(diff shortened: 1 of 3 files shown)


domenic (changes requested):
Can we use a separate urltestdata.json, instead of directly encoding in a .window.js file? That will make it reusable by Node.js and jsdom.

annevk:
@domenic I'm fine with someone taking that on, but it's not a priority for me.

annevk:
Anyway, I think you should reconsider as the status quo is broken. The alternative is reverting https://github.com/w3c/web-platform-tests/commit/90efdbecfe3c457b36c71cf3e01a72cd5301c163.

url/failure.html
```diff
@@ -16,7 +16,11 @@
 
     const name = test.input + " should throw"
 
-    self.test(() => { // new URL itself is already tested by url-constructor.html
+    self.test(() => { // URL's constructor's first argument is tested by url-constructor.html
+      assert_throws(new TypeError(), () => new URL("about:blank", test.input));
```
domenic: It really seems like we should be only running these tests if test.base === "about:blank". But I guess it's probably true that if a URL fails to parse with a valid base, it will also fail to parse with about:blank?

annevk: I think that should be true, yes.

PR #10926, opened 2018-05-09
Add test for marker 'orient' attribute.

Tavmjong:
Basic test of marker 'orient' attribute include new 'auto-start-reverse' value.


svgeesus (approved):
okay, tests with both markers and use elements (so exact position can be checked) and reftests with just the use elements. Nice.
lgtm

PR #10468, opened 2018-04-13
More Tests for WebRTC Data Channels

lgrahl:
### Description

This PR updates all data channel related tests to reflect the current spec (at the time of writing, that was [revision 1cc5bfc3ff18741033d804c4a71f7891242fb5b3](https://rawgit.com/w3c/webrtc-pc/1cc5bfc3ff18741033d804c4a71f7891242fb5b3/webrtc.html)).

### Updated/Fixed Tests

* RTCDataChannel-id
  - *DTLS client uses odd data channel IDs*
  - *DTLS server uses even data channel IDs*
* RTCDataChannel-send-receive (renamed from RTCDataChannel-send)
  - (\<type> is of *In-band negotiated channel*, *Negotiated channel* or *Asymmetric negotiated channel*)
  - *\<type>: Calling send() when not open should throw InvalidStateError* (renamed from *Calling send() when data channel is in connecting state should throw InvalidStateError*)
  - *\<type>: Calling send() with null should throw TypeError*
  - *\<type>: Calling send() with undefined should throw TypeError*
  - *\<type>: Should be able to send (local) simple string and receive (remote) as string* (based on *Data channel should be able to send simple string and receive as string*)
  - *\<type>: Should be able to send (remote) simple string and receive (local) as string* (based on *Data channel should be able to send simple string and receive as string*)
  - *\<type>: Should be able to send unicode string and receive as unicode string* (renamed from *Data channel should be able to send unicode string and receive as unicode string*)
  - *\<type>: Should ignore binaryType and always receive string message as string* (renamed from *Data channel should ignore binaryType and always receive string message as string*)
  - *\<type>: Should be able to send Uint8Array message and receive as ArrayBuffer* (renamed from *Data channel should be able to send Uint8Array message and receive as ArrayBuffer*)
  - *\<type>: Should be able to send ArrayBuffer message and receive as ArrayBuffer* (renamed from *Data channel should be able to send ArrayBuffer message and receive as ArrayBuffer*)
  - *\<type>: Should be able to send Blob message and receive as ArrayBuffer* (renamed from *Data channel should be able to send Blob message and receive as ArrayBuffer*)
  - *\<type>: Should be able to send ArrayBuffer message and receive as Blob* (renamed from *Data channel should be able to send ArrayBuffer message and receive as Blob*)
  - *\<type>: Should be able to transmit an empty string* (moved from and based on datachannel-emptystring *Can send empty strings across a WebRTC data channel.*)
[...]

PR #10933, opened 2018-05-09
Expose done promise on Test

lgrahl:
Expose the *done* promise as `Test.done_promise` when using a `promise_test`. Allows to mix event callback handling with promise-based tests.

Real world examples: https://gist.github.com/lgrahl/de4101dc0276f28975a028a595ce05d7 and #10468

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10933)

### DIFF
--- a/docs/_writing-tests/testharness-api.md
+++ b/docs/_writing-tests/testharness-api.md
@@ -89,7 +89,7 @@ Assertions can be added to the test by calling the step method of the test
 object with a function containing the test assertions:
 
 ```js
-document.addEventListener("DOMContentLoaded", function() {
+document.addEventListener("DOMContentLoaded", function(e) {
   t.step(function() {
     assert_true(e.bubbles, "bubbles should be true");
   });
@@ -108,7 +108,7 @@ first argument. The above example can be rewritten as:
 
 ```js
 async_test(function(t) {
-  document.addEventListener("DOMContentLoaded", function() {
+  document.addEventListener("DOMContentLoaded", function(e) {
     t.step(function() {
       assert_true(e.bubbles, "bubbles should be true");
     });
@@ -127,7 +127,7 @@ callback. A convenient method of doing this is through the `step_func` method
 which returns a function that, when called runs a test step. For example:
 
 ```js
-document.addEventListener("DOMContentLoaded", t.step_func(function() {
+document.addEventListener("DOMContentLoaded", t.step_func(function(e) {
   assert_true(e.bubbles, "bubbles should be true");
   t.done();
 }));
@@ -137,7 +137,7 @@ As a further convenience, the `step_func` that calls `done()` can instead
 use `step_func_done`, as follows:
 
 ```js
-document.addEventListener("DOMContentLoaded", t.step_func_done(function() {
+document.addEventListener("DOMContentLoaded", t.step_func_done(function(e) {
   assert_true(e.bubbles, "bubbles should be true");
 }));
 ```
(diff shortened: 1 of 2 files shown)


docs/_writing-tests/testharness-api.md
```diff
@@ -250,6 +250,30 @@ created Promise waiting to be fulfilled, or if the event is of a different type
...
+
+```js
+function foo() {
+  return Promise.resolve("foo");
+}
+
+promise_test(function(t) {
+  someObject.addEventListener("someevent", t.step_func(function() {
+    t.done();
+  }));
```
annevk: When would this be called?

lgrahl: I don't understand. (Maybe you can ping me in #whatwg later)

lgrahl:
> That's a difficult requirement to document and enforce, though.

Well, I mean the `done()` method already exists, it clearly is a public method and it can be called in a `promise_test`. A suggestion for documentation:

> In a `promise_test` you have two ways to mark a test case as *done*:
>
> 1. Resolve the returned `Promise`, or
> 2. Call the `done()` method on the `Test` instance.
> 
> To use the latter it is required that you wait for the `done_promise` at the end of the test before resolving. Otherwise, the test would be marked as *done* but the returned promise could still be pending (and running).

> By awaiting the "done promise" in a test, authors are saying "this test is not done when testharness.js considers it to be done."

I would phrase it differently: "This test is not done before I explicitly mark it as done."

> From the use cases I've seen, it seems like a locally-defined promise could serve this purpose

Creating a `Promise` and waiting for it at the end would of course work but the promise and the `done()` method already exist, so it makes sense to use it for that purpose.

jugglinmike:
> Well, I mean the `done()` method already exists, it clearly is a public method and it can be called in a `promise_test`.

I agree completely. I think that using `done` in a `promise_test` is confusing, though, because it's not clear how the two features interact and because they currently don't interact very well.

> I would phrase it differently: "This test is not done before I explicitly mark it as done."

Here's what I had in mind:

```js
promise_test(async (t) => {
  setTimeout(() => t.done());

  await t.done_promise;

  // Is the test done at this moment?
});
```

It turns out that the test *is* done at that point: the harness has already gone on to execute the next test or to report the results (if this is the final test). More troubling: any error thrown here will be ignored.

> Creating a `Promise` and waiting for it at the end would of course work but the promise and the `done()` method already exist, so it makes sense to use it for that purpose.

I'm not convinced that their existence justifies the convenience. `done` has a specific meaning which is subverted by its use in this way--that's what I tried to show in my example above.

My concern is really about using `t.done` within a `promise_test`. That's not under consideration here, but since I think that usage is problematic, I'm reluctant to endorse new features that support it.

The patch I suggested earlier reflects a different way of handling asynchronous operations. Not every API is Promise-enabled, but that doesn't have to inform the structure of the entire test (and the testing framework). Instead, you can contain the awkwardness of the callback API to the specific points that require it.

lgrahl:
> It turns out that the test *is* done at that point: the harness has already gone on to execute the next test or to report the results (if this is the final test). More troubling: any error thrown here will be ignored.

That is a fair point. The docs would need to mention that it's okay to use promises for setup but if the *done* marking is done with the `done()` method (:smirk:) then it may not race with the promises.

However, I get your point that this is not easy to grasp. An alternative approach (very similar to your suggestion and I used it in an earlier prototype) is this:

```patch
 promise_test(async (t) => {
+  const future = new Future();
+ 
   const pc2 = new RTCPeerConnection();
   t.add_cleanup(() => pc2.close());

   exchangeIceCandidates(pc1, pc2);

   // Create a channel on pc1
   const dc = pc1.createDataChannel('onopen');
   dc.onopen = t.step_func(() => {
     assert_equals(dc.readyState, 'open', 'Channel should be open after onopen has fired');
-    t.done();
+    future.resolve();
   });

   await doSignalingHandshake(pc1, pc2);
-  await t.done_promise;
+  await future;
 }, 'In-band channel: Open event should be fired (local) when the data channel opens');
```

([Future](https://gist.github.com/lgrahl/1852b808273841ed806d9e0baa0aa571) definition)

docs/_writing-tests/testharness-api.md
annevk: I mean I don't understand this example. I don't see anything dispatching `someevent`.

lgrahl: That's because nothing does. :slightly_smiling_face: It could be added.

lgrahl:
So, should I use the `Future` approach instead? This is blocking #10468. :slightly_smiling_face:

lgrahl:
Will use the `Future` since I'm getting no response.

PR #10935, opened 2018-05-09
Add browser version into run_info (and hence wptreport)

Hexcles:
Another piece of #10511 

The data flow is:
* `wpt.run` gets `browser_version` from `wpt.browser` during `setup_wptrunner`
* `browser_version` is passed to `wptrunner.start` via `kwargs`
* `wptrunner` then passes `browser_version` to `RunInfo`

As a result, many pieces of code are modified along the code path above. I'm personally not 100% happy with the change in terms of elegance, too much plumbing, but I think generally speaking I'm not making things much worse... Suggestions for less bad solutions are much welcome!

Many of the `version()` methods in `wpt.browser` were actually broken, so I had to fix them as well. AFAIK, these methods weren't called anywhere before. Currently, we don't have version strings for Edge, Safari and Sauce yet.

(@kereliuk helped me to understand some of the code. Thanks! also cc @foolip )

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10935)

### DIFF
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -371,7 +371,7 @@ class ChromeAndroid(Browser):
         return chrome.install_webdriver(dest)
 
     def version(self, binary):
-        raise NotImplementedError
+        return None
 
 
 class Opera(Browser):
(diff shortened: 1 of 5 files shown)


jgraham (changes requested):
Generally looks good, but presumably we can't have `NotImplementedError`

Hexcles:
@jgraham You're right. Fixed.

jgraham (approved):

PR #10947, opened 2018-05-10
Added dictionary WebGLContextAttributes to avoid no original dictionary error

haoyunfeix:
This PR fixed "Partial dictionary WebGLContextAttributes with no
original dictionary" error in webxr idlharness test

### DIFF
--- a/webxr/interfaces.https.html
+++ b/webxr/interfaces.https.html
@@ -17,6 +17,7 @@ promise_test(async () => {
   idl_array.add_untested_idls(dom_idl);
   idl_array.add_untested_idls("interface Navigator {};");
   idl_array.add_idls(webxr_idl);
+  idl_array.add_idls("dictionary WebGLContextAttributes {};");
   idl_array.add_objects({
     Navigator:['navigator'],
   });


haoyunfeix:
@lukebjerring Could you please take a look at this PR?
I'm not sure if I did the right change, however it works now.

The reason for this PR is that I found idl files which contains "partial dictionary" are failed in idlharness test after https://github.com/w3c/web-platform-tests/pull/10159/ merged. There will be a "Partial dictionary XXX with no original dictionary" error happened and block other test cases.

Ms2ger (approved):

lukebjerring:
Preferable to this, would be to add it from the actual source ([webgl1](https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.2)?), using the `only` option. Generally this would look like:

    const webgl = await fetch('/interfaces/webgl1.idl').then(r => r.text());
    // ...
    idl_array.add_idls(webgl, {only: 'WebGLContextAttributes' })

However, I don't see `webgl` in the `interfaces/` dir of wpt yet, (and can't find it elsewhere), so this is fine for now.

lukebjerring:
Related: https://github.com/KhronosGroup/WebGL/issues/2537 - seems like @foolip was thinking WebGL tests shouldn't be in WPT?

foolip:
Yep. Splitting across repos without some principle for what kind of test goes where will be frustrating, and there are only a handful of tests in WPT. See https://github.com/w3c/web-platform-tests/issues/8044 for more background.

We could still have just the idlharness.js tests in WPT though, anywhere else would require a copy of the test framework itself too.

haoyunfeix:
@lukebjerring @foolip , thanks for the comments! And could you help merging this PR to avoid case blocking?

Ms2ger:
This kind of issue is exactly why we should have WebGL in WPT; it's part of the web platform and interacts with other parts of the web platform. `</soapbox>`

Anyway, keeping this webxr test broken serves no purpose, so I'm merging.

PR #10953, opened 2018-05-10
Partly fix #8581: add extra subdomains for cookie tests

### LINKED ISSUE #8581: Isolated domain for cookie tests?

gsnedders:
See #8581.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10953)

### DIFF
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -16,6 +16,7 @@ import traceback
 from six.moves import urllib
 import uuid
 from collections import defaultdict, OrderedDict
+from itertools import chain, product
 from multiprocessing import Process, Event
 
 from localpaths import repo_root
@@ -700,6 +701,9 @@ def build_config(override_path=None, **kwargs):
 
     return rv
 
+def _make_subdomains_product(s, depth=3):
+    return set(u".".join(x) for x in chain(*(product(s, repeat=i) for i in range(1, depth+1))))
+
 _subdomains = {u"www",
                u"www1",
                u"www2",
@@ -707,6 +711,10 @@ _subdomains = {u"www",
                u"élève"}
 
 _not_subdomains = {u"nonexistent"}
+
+_subdomains = _make_subdomains_product(_subdomains)
+
+_not_subdomains = _make_subdomains_product(_not_subdomains)
 
 
 class ConfigBuilder(config.ConfigBuilder):
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -752,9 +752,11 @@ class ConfigBuilder(config.ConfigBuilder):
     computed_properties = ["ws_doc_root"] + config.ConfigBuilder.computed_properties
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


gsnedders:
@jgraham pointed out yesterday that while @mikewest started off by saying in #8581 it "would be simplified if we could replicate its domain structure", the requests beneath it don't include using the exact same subdomains. And, of course, it'll need to be moved over to using `{{ host }}` at the very least, even if the subdomains are hardcoded, so it's not clear there's such a gain to exactly matching the hostnames?

mikewest (approved):
I assume these would be accessible via `{{hosts[][go.deeper]}}` and `{{hosts[alt][go.deeper]}}`? If so, LGTM!

gsnedders:
@mikewest yes

tools/serve/serve.py
```diff
@@ -663,7 +663,17 @@ def load_config(override_path=None, **kwargs):
                u"www1",
                u"www2",
                u"天気の良い日",
-               u"élève"}
+               u"élève",
+               u"deeper",
```
jgraham: OK, o I"m going to sound like the most boring person, but I don't really like the whimsical names.

Is there something wrong with using the existing hosts and adding e.g. 
```
sub1a.www
sub1b.www
sub2a.sub1a.www
sub2b.sub1a.www
sub3a.sub2a.sub1a.www
```
and so on for whatever combinations you need?

jgraham (changes requested):

tools/serve/serve.py
gsnedders: You're no fun!

Yours insincerely,

gsnedders.

annevk:
The commit message should probably note that this fixes/closes #11219.

gsnedders:
@jgraham Here's a less fun patch, following @annevk's suggestion from IRC to just repeat the existing subdomains.

jgraham (approved):

jugglinmike:
For default configurations, this patch expands the number of distinct origins from 12 to 62 (actually 312, but [I think that's a bug](https://github.com/web-platform-tests/wpt/pull/12459)). If I'm reading the original feature request correctly, all that is technically required is "a subdomain of a subdomain (and another subdomain of the same subdomain)". We could satisfy this by introducing 2 new entries for a total of 14.  Elsewhere, @jgraham's wrote that "[it seems messy to add a lot of additional subdomains](https://github.com/web-platform-tests/wpt/issues/8581#issuecomment-388116372)."  @gsnedders I expect this might be a result of the feedback you got from Anne in IRC; could you say a bit about why you went with a combinatoric approach?

annevk:
My suggestion was only about reusing existing names, if I remember correctly.

gsnedders:
I have literally no memory why I went for so much repetition.

PR #10955, opened 2018-05-10
Change how invalid base URLs are tested

### LINKED ISSUE #8707: Some a-element.html URL tests seem wrong

domenic:
Fixes #8707. Supercedes #10917, which it is based on.

This keeps the tests inside urltestdata.json, but adds a new requirement on consumers of that file that they should also test their base URL parsing logic by reusing the data in a different way.

It also rearranges the various cases from 90efdbecfe3c457b36c71cf3e01a72cd5301c163 under several different headings.

/cc @GPHemsley and @TimothyGu especially to review the changes to the README to see if they're clear on how to maintain and improve coverage. Existing harnesses will _lose_ coverage from this change, since urltestdata.json no longer tests the `a`, `a/`, and `a//` base URLs directly. But if they implement the tweak suggested in the README, they will gain coverage, testing not only those URLs as bases but also lots of others already in the file.

### DIFF
--- a/url/urltestdata.json
+++ b/url/urltestdata.json
@@ -6521,27 +6521,34 @@
     "search": "?a",
     "hash": "#%GH"
   },
-  "Bad bases",
+  "URLs that require a non-about:blank base. (Also serve as invalid base tests.)"
   {
-    "input": "test-a.html",
-    "base": "a",
+    "input": "a",
+    "base": "about:blank",
     "failure": true
   },
   {
-    "input": "test-a-slash.html",
-    "base": "a/",
+    "input": "a/",
+    "base": "about:blank",
     "failure": true
   },
   {
-    "input": "test-a-slash-slash.html",
-    "base": "a//",
+    "input": "a//",
+    "base": "about:blank",
     "failure": true
   },
+  "Bases that don't fail to parse but fail to be bases",
   {
     "input": "test-a-colon.html",
     "base": "a:",
     "failure": true
   },
+  {
+    "input": "test-a-colon-b.html",
+    "base": "a:b",
+    "failure": true
+  },
+  "Other base URL tests, that must succeed"
   {
     "input": "test-a-colon-slash.html",
     "base": "a:/",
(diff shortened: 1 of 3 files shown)


annevk (approved):

url/urltestdata.json
```diff
@@ -6521,27 +6521,34 @@
     "search": "?a",
     "hash": "#%GH"
   },
-  "Bad bases",
+  "URLs that require a non-about:blank base. (Also serve as invalid base tests.)"
```
GPHemsley: This is missing a trailing comma.

url/urltestdata.json
```diff
@@ -6521,27 +6521,34 @@
...
     "input": "test-a-colon.html",
     "base": "a:",
     "failure": true
   },
+  {
+    "input": "test-a-colon-b.html",
+    "base": "a:b",
+    "failure": true
+  },
+  "Other base URL tests, that must succeed"
```
GPHemsley: So is this.

GPHemsley (changes requested):
You've got invalid JSON here.

url/urltestdata.json
domenic: Thanks, I should actually test my changes against jsdom/whatwg-url as I do for everyone else's to this file -_-

domenic:
JSON fixed; confirmed working in https://github.com/jsdom/whatwg-url/pull/116. Filed https://github.com/nodejs/node/issues/20720 to alert Node.js folks. Will merge this after Travis finishes.

GPHemsley:
I'd wanted to test these changes before I commented on the clarity of the readme... I'm not entirely clear on what's changing.

Are you saying that the base URL needs to be parsed the same as the input? And that failing to parse the base URL should fail the whole thing? If so, I think I already do that...

domenic:
Sorry, I didn't realize :(.

That's basically what I'm saying, yes. I imagine most people already do this, but the test cases were previously not testing it explicitly. They had a few examples of invalid base URLs, but now we're encouraging testing all invalid URLs as invalid base URLs.

PR #10956, opened 2018-05-10
Accname: Remove two tests

joanmarie:
One test covers something governed by HTML-AAM rather than Accname.
Plus HTML-AAM doesn't currently have language for the condition in
question.

The other test covers an edge case which should be covered by Accname
but isn't. Fixing the spec for this edge case is complicated, and we
do not yet have consensus on what the correct expected results are.
Therefore, the Working Group will address this in version 1.2.


fred-wang (approved):
LGTM

PR #10957, opened 2018-05-10
Properly switching window focus after close of window

jimevans:
The test_handle_prompt_accept test for closing a window tests to see if an
alert was handled. It does so by calling the Get Alert Text end point.
However, the Get Alert Text end point requires an active, valid top-level
browsing context to have the focus. The test now switches focus back to
the previous window before checking that an alert was handled. In between
checking the alert, confirm, and prompt cases, it recreates the new window
required so it can be closed.

### DIFF
--- a/webdriver/tests/close_window/user_prompts.py
+++ b/webdriver/tests/close_window/user_prompts.py
@@ -32,6 +32,9 @@ def test_handle_prompt_accept(new_session, add_browser_capabilites):
     create_dialog(session)("alert", text="dismiss #1", result_var="dismiss1")
     response = close(session)
     assert response.status == 200
+
+    # Asserting that the dialog was handled requires valid top-level browsing
+    # context, so we must switch to the original window.
     session.window_handle = original_handle
     assert_dialog_handled(session, "dismiss #1")
 
@@ -41,6 +44,9 @@ def test_handle_prompt_accept(new_session, add_browser_capabilites):
     create_dialog(session)("confirm", text="dismiss #2", result_var="dismiss2")
     response = close(session)
     assert response.status == 200
+
+    # Asserting that the dialog was handled requires valid top-level browsing
+    # context, so we must switch to the original window.
     session.window_handle = original_handle
     assert_dialog_handled(session, "dismiss #2")
 
@@ -50,6 +56,9 @@ def test_handle_prompt_accept(new_session, add_browser_capabilites):
     create_dialog(session)("prompt", text="dismiss #3", result_var="dismiss3")
     response = close(session)
     assert response.status == 200
+
+    # Asserting that the dialog was handled requires valid top-level browsing
+    # context, so we must switch to the original window.
     session.window_handle = original_handle
     assert_dialog_handled(session, "dismiss #3")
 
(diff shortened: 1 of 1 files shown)


webdriver/tests/close_window/user_prompts.py
```diff
@@ -3,7 +3,7 @@
 from tests.support.asserts import assert_error, assert_dialog_handled
 from tests.support.fixtures import create_dialog, create_window
 from tests.support.inline import inline
-
+from time import sleep
```
andreastt: Debug leftover?

whimboo: Please remove the line which you probably used when debugging the code.

andreastt (dismissed):

andreastt:
@whmboo: Can you take a closer look at this?

andreastt:
@whimboo Can you take a closer look at this?

webdriver/tests/close_window/user_prompts.py
```diff
@@ -24,23 +24,33 @@ def test_handle_prompt_ignore():
...
+    original_handle = session.window_handle
 
+    session.window_handle = create_window(session)()
     session.url = inline("<title>WD doc title</title>")
 
     create_dialog(session)("alert", text="dismiss #1", result_var="dismiss1")
     response = close(session)
     assert response.status == 200
+    session.window_handle = original_handle
     assert_dialog_handled(session, "dismiss #1")
```
whimboo: We want to make sure that the dialog on the newly opened window has been handled and not on the original window, right? In that case resetting the `window_handle` should come after the assertion.

whimboo (changes requested):

webdriver/tests/close_window/user_prompts.py
whimboo: Hm, given that the dialog should have been accepted by the `close` command, the window is gone. Does it make sense to have the check at all for this command?

webdriver/tests/close_window/user_prompts.py
jimevans: Yep. Will do.

jimevans: Good catch. Fixing.

webdriver/tests/close_window/user_prompts.py
jimevans: I suppose it makes sense that we assert that no dialog is present, which is what assert_dialog_handled does. However, the call to Get Alert Text (which is what assert_dialog_handled uses to validate whether an alert is present) requires a valid, open top-level browsing context, so the reset of the current window handle must come first.

whimboo: So lets have a look at the spec for Switch To Window:

> If there is an active user prompt, that prevents the focussing of another top-level browsing context, return error with error code unexpected alert open. 

So similar to other commands switching to a different window will also accept the dialog in this test. So using `assert_dialog_handled` afterward would be a no-op. Best would still be to check that the newly created window is gone, and if not make the dialog check.

whimboo (approved):
I think this is fine for now. Thanks.

andreastt (approved):

Issue #10906, opened 2018-05-08
Many tests may be flaky after #10198


foolip:
https://github.com/w3c/web-platform-tests/pull/10198 was blocked for a long time because of being flaky in Chrome Dev, which was never resolved. It was finally force merged because the changes had been reverted in Chromium by automatic import, and the best way to recover was to merge the PR.

The last Travis job showing the flakiness:
https://travis-ci.org/w3c/web-platform-tests/jobs/376549386

foolip:
OK, "good", https://github.com/w3c/web-platform-tests/pull/10907 is still flaky when touching the tests, so this is possible to investigate.

Honry:
@foolip, the flakiness disappeared in my PR https://github.com/w3c/web-platform-tests/pull/10962, I didn't do much change, but it's magic.  :)

But most tests still failed in my local, so I did an investigation and have some findings.

1.  See following code snippet from [generic-sensor-tests.js](https://github.com/w3c/web-platform-tests/blob/master/generic-sensor/generic-sensor-tests.js#L47
), the code `await sensorTest.reset()` never been invoked once tests failed at `await func(t)`, which would lead to failure of mock sensor initialization in following tests.  

```
function sensor_test(func, name, properties) {
  promise_test(async (t) => {
    if (typeof GenericSensorTest === 'undefined') {
      await loadChromiumResources;
    }
    assert_true(typeof GenericSensorTest !== 'undefined');
    let sensorTest = new GenericSensorTest();
    await sensorTest.initialize();
    try {
      await func(t);
    } finally {
      await sensorTest.reset();
    };
  }, name, properties);
}
```
This can be fixed by using `add_cleanup` function support for `promise_test`, but which is still under review at https://github.com/w3c/web-platform-tests/pull/8748

2. Based on fixing in item [1], I found that the mock sensor reading is always 0, @alexshalamov, any idea?

rakuco:
> But most tests still failed in my local, so I did an investigation and have some findings.
> 
> 1. See following code snippet from generic-sensor-tests.js, the code await sensorTest.reset() never been invoked once tests failed at await func(t), which would lead to failure of mock sensor initialization in following tests.

I believe this one's been fixed by #11625 and we're no longer blocked on `add_cleanup` + promises.

> 2. Based on fixing in item [1], I found that the mock sensor reading is always 0, @alexshalamov, any idea?

[...]

PR #10962, opened 2018-05-11
Fix partial issues from #10198

Honry:
- `if (testRunner)` should be `if (window.testRunner)`
 - Use String to pass sensor object param to avoid undefined error

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10962)

### DIFF
--- a/resources/chromium/generic_sensor_mocks.js
+++ b/resources/chromium/generic_sensor_mocks.js
@@ -201,10 +201,10 @@ var GenericSensorTest = (() => {
       if (testInternal.initialized)
         throw new Error('Call reset() before initialize().');
 
-      if (testRunner) { // Grant sensor permissions for Chromium testrunner.
+      if (window.testRunner) { // Grant sensor permissions for Chromium testrunner.
         ['accelerometer', 'gyroscope',
          'magnetometer', 'ambient-light-sensor'].forEach((entry) => {
-          testRunner.setPermission(entry, 'granted',
+          window.testRunner.setPermission(entry, 'granted',
                                    location.origin, location.origin);
         });
       }
--- a/resources/chromium/generic_sensor_mocks.js
+++ b/resources/chromium/generic_sensor_mocks.js
@@ -205,7 +205,7 @@ var GenericSensorTest = (() => {
         ['accelerometer', 'gyroscope',
          'magnetometer', 'ambient-light-sensor'].forEach((entry) => {
           window.testRunner.setPermission(entry, 'granted',
-                                   location.origin, location.origin);
+                                          location.origin, location.origin);
         });
       }
 
--- a/ambient-light/AmbientLightSensor-disabled-by-feature-policy.https.html
+++ b/ambient-light/AmbientLightSensor-disabled-by-feature-policy.https.html
@@ -8,6 +8,6 @@
 <script>
 "use strict";
 
-run_fp_tests_disabled(AmbientLightSensor);
+run_fp_tests_disabled('AmbientLightSensor');
 </script>
 </body>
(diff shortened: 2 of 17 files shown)


Honry:
@pozdnyakov, @alexshalamov, PTAL.

Honry:
@foolip @rakuco, could you help review this PR? I am going to bring more methods from Chromium sensor mojo interfaces to WPT. e.g. `setUpdateSensorReadingFunction`, `setStartShouldFail`, etc., which is somehow based on this PR.

resources/chromium/generic_sensor_mocks.js
```diff
@@ -201,10 +201,10 @@ var GenericSensorTest = (() => {
       if (testInternal.initialized)
         throw new Error('Call reset() before initialize().');
 
-      if (testRunner) { // Grant sensor permissions for Chromium testrunner.
+      if (window.testRunner) { // Grant sensor permissions for Chromium testrunner.
         ['accelerometer', 'gyroscope',
          'magnetometer', 'ambient-light-sensor'].forEach((entry) => {
-          testRunner.setPermission(entry, 'granted',
+          window.testRunner.setPermission(entry, 'granted',
                                    location.origin, location.origin);
```
rakuco: You need to reindent this line as well.

resources/chromium/generic_sensor_mocks.js
```diff
@@ -201,10 +201,10 @@ var GenericSensorTest = (() => {
       if (testInternal.initialized)
         throw new Error('Call reset() before initialize().');
 
-      if (testRunner) { // Grant sensor permissions for Chromium testrunner.
+      if (window.testRunner) { // Grant sensor permissions for Chromium testrunner.
```
rakuco: Sorry if this sounds dumb, but isn't `testRunner` the same as `window.testRunner` here? Or do the Mojo JS files behave differently?

Honry: This `testRunner` is specific for Chromium testrunner, not sure if it is still useful in `LayoutTests/external/wpt`, but for sure, it is `undefined` in w-p-t runnner.

Honry: i.e. tests will fail with `Uncaught ReferenceError: testRunner is not defined`

resources/chromium/generic_sensor_mocks.js
Honry: thanks! Fixed.

foolip (changes requested):
Aren't all the tests failing now? run_fp_tests_* and runGenericSensor* need to updated, right?

Honry:
@foolip, most tests are still failing, for the reason, please see my comment at https://github.com/web-platform-tests/wpt/issues/10906#issuecomment-388394993

rakuco (approved):
The changes look fine to me, even if there might be additional work left. @foolip?

foolip:
@Honry, even if the tests are already failing, won't these changes make them further from passing? If they're already passing in Chromium, could you try making the changes in a Chromium CL and see if they stay passing?

rakuco:
@foolip AFAICS, this PR should get these tests closer, not further, to passing. `ambient-light/` and `proximity` are currently not imported into Chromium, and `geolocation-sensor/` is failing there due to `GeolocationSensor` not being implemented at all in Blink.

Other directories which call those functions (e.g. `magnetometer/`) are already using these function call forms.

foolip:
Hmm, something is very strange, run_fp_tests_disabled and runGenericSensorTests already assumes a string argument, but some tests still pass interface objects. So yes, these tests are moving things in the right direction. (I assumed that the helpers weren't updated, in which case this made no sense.)

foolip (approved):

PR #10965, opened 2018-05-11
Correcting string format for sending WebDriver command

jimevans:
The get_tag_name function was previously hard-coding the WebDriver element ID to an invalid value.

### DIFF
--- a/webdriver/tests/get_element_tag_name/user_prompts.py
+++ b/webdriver/tests/get_element_tag_name/user_prompts.py
@@ -7,9 +7,9 @@ def read_global(session, name):
     return session.execute_script("return %s;" % name)
 
 
-def get_tag_name(session, element_id):
+def get_tag_name(session, webdriver_element_id):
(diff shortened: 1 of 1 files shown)


shs96c (approved):
Oops

webdriver/tests/get_element_tag_name/user_prompts.py
```diff
@@ -7,9 +7,9 @@ def read_global(session, name):
     return session.execute_script("return %s;" % name)
 
 
-def get_tag_name(session, element_id):
+def get_tag_name(session, webdriver_element_id):
```
whimboo: All other tests use `element_id` here. Can you change that please?

jimevans: Happy to. I was simply worried that the formatter would get confused between the token name and variable name. I should’ve tried it myself first.

andreastt (approved):

PR #10973, opened 2018-05-12
Use getPropertyValue in parsing-testcommon.js

csnardi:
If a property is entirely unsupported, setting `style[property]` will create a new attribute on `style`, and getting `style[property]` will return whatever value was just set. This can create false results as the test will report that parsing succeeded even when it did not, and that parsing failed even when it did not. Switch to `getPropertyValue` to remove this problem.


csnardi:
@ewilligers Could you review?

ewilligers:
A couple more cleanups:

When the serialized value is not one of the expected values, it is helpful to report what it is:
```
<             assert_true(serializedValue.includes(readValue), "serialization should be sound");
>             assert_true(serializedValue.includes(readValue), "serialization should be sound, instead was \"" + readValue + "\"");
```

The latest parsing-testcommon.js should also be copied to
- css/motion/parsing/resources/parsing-testcommon.js
- css/filter-effects/parsing/resources/parsing-testcommon.js

ewilligers (approved):

PR #10975, opened 2018-05-12
Add descriptions to scrollIntoView-smooth asserts

stipsan:
Without it you can't tell wether it was the `scrollX` or `scrollY` that failed without modifying the test.


csnardi (approved):
Looks good, thanks!

PR #10976, opened 2018-05-13
Enable attr-referrer script-tag tests

domfarolino:
As per the spec change https://github.com/whatwg/html/pull/3678 we should probably enable the script-tag attr-referrer tests. Here I modified the spec JSON, removing `script-tag` from the disabled section and ran `generate.py` to generate the ncessary tests. I think that should be all that's necessary here?

/cc @zcorpan


zcorpan:
I restarted the failed Firefox job.

jdm (approved):
I'm familiar with these tests. The generated tests here should be fine.

domfarolino:
Just a note, I've filed #11023 to at some point figure out what might be necessary to support the referrer policy tests for `import`ed scripts.

PR #10977, opened 2018-05-13
Test BigInt as keys and values in IndexedDB

lispwitch:
Test BigInt serialization in IndexedDB values, and check that BigInt values cannot be used as IndexedDB keys. These are @littledan's tests with some changes suggested in reviews; see [PR #9667](https://github.com/w3c/web-platform-tests/pull/9667) for the original version.

### DIFF
--- a/IndexedDB/bigint_value.htm
+++ b/IndexedDB/bigint_value.htm
@@ -1,65 +1,72 @@
 <!DOCTYPE html>
 <meta charset="utf-8">
-<title>Values</title>
+<title>IndexedDB: BigInt keys and values</title>
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <script src="support.js"></script>
 
 <script>
-// BigInt and BigInt wrappers are supported in serialization, per
+// BigInt and BigInt objects are supported in serialization, per
 // https://github.com/whatwg/html/pull/3480
 // This support allows them to be used as IndexedDB values.
 
-let i = 0;
-function value(value, test) {
-    var t = async_test(document.title + " - " + (i++));
-    t.step(function() {
-        assert_true(test(value),
-                    "condition does not apply to initial value");
-    });
+function value_test(value, predicate, name) {
+    async_test(t => {
+        t.step(function() {
+            assert_true(predicate(value),
+                        "Predicate should return true for the initial value.");
+        });
 
-    createdb(t).onupgradeneeded = function(e) {
-        e.target.result
-                .createObjectStore("store")
-                .add(value, 1);
-
-        e.target.onsuccess = t.step_func(function(e) {
+        createdb(t).onupgradeneeded = t.step_func(e => {
             e.target.result
-                    .transaction("store")
-                    .objectStore("store")
-                    .get(1)
-                    .onsuccess = t.step_func(function(e)
-            {
-                assert_true(test(e.target.result),
-                            "condition does not apply to deserialized result")
-                t.done();
+                    .createObjectStore("store")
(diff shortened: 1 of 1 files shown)


IndexedDB/bigint_value.htm
```diff
@@ -0,0 +1,77 @@
...
+      "BigInt object inside object");
+
+// However, BigInt is not supported as an IndexedDB key; support
+// has been proposed in the following PR, but that change has not
+// landed at the time this patch was written
+// https://github.com/w3c/IndexedDB/pull/231
+
+function invalidKey(key, name) {
+    async_test(t => {
+        createdb(t).onupgradeneeded = t.step_func(e => {
```
inexorabletash: It's easier to just use e.g. `indexedDB.cmp(0, key)` - this will throw if the key is not valid. No need to create a database.

IndexedDB/bigint_value.htm
```diff
@@ -0,0 +1,77 @@
...
+// BigInt and BigInt wrappers are supported in serialization, per
+// https://github.com/whatwg/html/pull/3480
+// This support allows them to be used as IndexedDB values.
+
+let i = 0;
+function value(value, test, name) {
+    async_test(t => {
+        t.step(function() {
+            assert_true(test(value),
+                        "condition does not apply to initial value");
```
inexorabletash: Assertion messages should be phrased in the form of the expected behavior, e.g. "Predicate should return true for the initial value"; the error output is confusing if the inverse is used.

IndexedDB/bigint_value.htm
```diff
@@ -0,0 +1,77 @@
...
+<script src="/resources/testharnessreport.js"></script>
+<script src="support.js"></script>
+
+<script>
+// BigInt and BigInt wrappers are supported in serialization, per
+// https://github.com/whatwg/html/pull/3480
+// This support allows them to be used as IndexedDB values.
+
+let i = 0;
+function value(value, test, name) {
```
inexorabletash: Can you rename the `test` argument to `test_func` or `predicate`? Since `test` is a testharness function it's confusing to see it re-used as an argument.

inexorabletash: Using `value` as both the name of the function and the argument is confusing. How about `value_test` for the function?

IndexedDB/bigint_value.htm
```diff
@@ -0,0 +1,77 @@
...
+            assert_true(test(value),
+                        "condition does not apply to initial value");
+        });
+
+        createdb(t).onupgradeneeded = t.step_func(e => {
+            e.target.result
+                    .createObjectStore("store")
+                    .add(value, 1);
+
+            e.target.onsuccess = t.step_func(function(e) {
```
inexorabletash: Arrow function instead of `function` ?

IndexedDB/bigint_value.htm
```diff
@@ -0,0 +1,77 @@
...
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="support.js"></script>
+
+<script>
+// BigInt and BigInt wrappers are supported in serialization, per
+// https://github.com/whatwg/html/pull/3480
+// This support allows them to be used as IndexedDB values.
+
+let i = 0;
```
inexorabletash: Unused?

inexorabletash (changes requested):
Thanks for the submission!

cxielarko:
implemented the requested changes, and added a missing .valueOf() call

inexorabletash (approved):
Looks great, thanks!

PR #10988, opened 2018-05-14
stability/verify unification

### LINKED ISSUE #9874: Move stability checking to `wpt run --verify`
Since we developed the stability checker, wptrunner got a `--verify` flag for checking test stability. Ideally instead of having separate codepaths here we could keep all the travis handling logic from the check_stability script but do the actual run using the `--verify` option to wpt run, so that we are confident that the same CI checks apply to both.

### LINKED ISSUE #10921: Make --stability just use the --verify code
To roughly quote @jgraham:

> so `--stability` is almost exactly run on Travis, `--verify` is almost exactly what’s run on Mozilla infra; `--verify` is more flexible (with/without restarts, with/without chaos mode, varying number of iterations); `--stability` just does 10 iters with restarts

So basically `--stability` is a subset of the behaviour and should probably cease being a separate codepath.

Note that `--stability` is currently a feature of `wpt run` and *not* `wptrunner`.

We need to make sure the output doesn't change when we move this over.

gsnedders:
Fixes #10921, and will fix #9874.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10988)


gsnedders:
Seems like I posted https://github.com/w3c/web-platform-tests/issues/9874#issuecomment-388892276 in the wrong place:

> So what I haven't done so far is move over to running the multi-part verify (with/without restarts, with/without chaos mode in Firefox), but it does now use the verify code.
>
> At the moment, it's still calling into the code through Python, rather than letting it run separately and then reading the logs. If we want to avoid going through the logs, we'd need to either make `check_stability` (in tools/wptrunner/wptrunner/stability.py) return something useful or call get_steps ourselves.

tools/wptrunner/wptrunner/wptcommandline.py
```diff
@@ -81,9 +81,13 @@ def create_parser(product_choices=None):
...
-    mode_group.add_argument("--verify", action="store_true",
-                            default=False,
-                            help="Run a stability check on the selected tests")
+    stability_group = mode_group.add_mutually_exclusive_group()
+    stability_group.add_argument("--verify", action="store_true",
+                                 default=False,
+                                 help="Run a stability check on the selected tests")
+    stability_group.add_argument("--stability", action="store_true",
+                                 default=False,
+                                 help="(DEPRECATED) Run a basic stability check on the selected tests")
```
jgraham: You can use `help = argparse.SUPPRESS` to remove this option from the help.

tools/wptrunner/wptrunner/wptrunner.py
```diff
@@ -302,6 +302,13 @@ def run_tests(config, test_paths, product, **kwargs):
 
 def check_stability(**kwargs):
     import stability
+    if kwargs["stability"]:
+        kwargs['verify_max_time'] = None
```
jgraham: Maybe log a warning here.

jgraham (changes requested):
Just a few small changes to request.

Can you push an unstable test, and a stable one, to this PR so that we can see how the travis output looks?

gsnedders:
@jgraham now done; sorry I missed your comments before!

foolip:
> Can you push an unstable test, and a stable one, to this PR so that we can see how the travis output looks?

Ping @gsnedders

gsnedders:
Um, yeah, seems I deleted them locally. Now readded.

gsnedders:
(labelled because that commit needs to go before merging)

foolip:
The Travis output looks about the same as before, so that seems to be working.

jgraham (approved):
Please remove the [DO NOT MERGE] commit :)

PR #10989, opened 2018-05-14
Update mediasource-duration.html from Chromium

zcorpan:
See https://bugs.chromium.org/p/chromium/issues/detail?id=689781#c19

Fixes #10920.


zcorpan:
cc @jyavenard

PR #11000, opened 2018-05-14
Find Chrome binary on more platforms and fail gracefully

### LINKED ISSUE #10992: wpt run chrome does not use the correct default OSX binary location when getting version string
**What happens**

Fails out before launching the browser when I try to run wpt run on my mac.

**How to reproduce**

`./wpt run chrome infrastructure/testdriver/click.html  --log-mach-level=debug`
> Traceback (most recent call last):
    wpt.main()
  File "/Users/jonathonkereliuk/0101/web-platform-tests/tools/wpt/wpt.py", line 132, in main
    rv = script(*args, **kwargs)
  File "/Users/jonathonkereliuk/0101/web-platform-tests/tools/wpt/run.py", line 437, in run
    **kwargs)
  File "/Users/jonathonkereliuk/0101/web-platform-tests/tools/wpt/run.py", line 424, in setup_wptrunner
    kwargs['browser_version'] = setup_cls.browser.version(kwargs.get("binary"))
  File "/Users/jonathonkereliuk/0101/web-platform-tests/tools/wpt/browser.py", line 342, in version
    version_string = call(binary, "--version").strip()
  File "/Users/jonathonkereliuk/0101/web-platform-tests/tools/wpt/utils.py", line 44, in call
    return subprocess.check_output(args)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 566, in check_output
    process = Popen(stdout=PIPE, *popenargs, **kwargs)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory

It thinks that the binary should be at usr/bin/google-chrome.

I think this is low priority since the work around is explicitly specifying the binary location with --binary.

I believe this was introduced by #10935

Hexcles:
Also make similar changes to Opera.  Fixes #10992.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/11000)

### DIFF
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -4,6 +4,7 @@ import platform
 import re
 import shutil
 import stat
+import subprocess
 import sys
 import tempfile
 from abc import ABCMeta, abstractmethod
@@ -206,7 +207,6 @@ class Firefox(Browser):
 
     def install_prefs(self, binary, dest=None):
         version, channel = self.get_version_and_channel(binary)
-
         if dest is None:
             dest = os.pwd
 
@@ -293,8 +293,16 @@ class Chrome(Browser):
     """
 
     product = "chrome"
-    binary = "/usr/bin/google-chrome"
     requirements = "requirements_chrome.txt"
+
+    @property
+    def binary(self):
+        if uname[0] == "Linux":
+            return "/usr/bin/google-chrome"
+        if uname[0] == "Darwin":
+            return "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
+        # TODO Windows?
+        return None
 
     def install(self, dest=None):
         raise NotImplementedError
@@ -339,7 +347,10 @@ class Chrome(Browser):
 
     def version(self, binary=None):
         binary = binary or self.binary
-        version_string = call(binary, "--version").strip()
+        try:
+            version_string = call(binary, "--version").strip()
+        except subprocess.CalledProcessError:
+            return None
         m = re.match(r"Google Chrome (.*)", version_string)
         if not m:
             return None
@@ -379,8 +390,14 @@ class Opera(Browser):
     """
 
     product = "opera"
-    binary = "/usr/bin/opera"
     requirements = "requirements_opera.txt"
+
+    @property
+    def binary(self):
+        if uname[0] == "Linux":
+            return "/usr/bin/opera"
+        # TODO Windows, Mac?
+        return None
 
     def install(self, dest=None):
         raise NotImplementedError
@@ -430,7 +447,10 @@ class Opera(Browser):
     def version(self, binary):
         """Retrieve the release version of the installed browser."""
         binary = binary or self.binary
-        output = call(binary, "--version")
+        try:
+            output = call(binary, "--version")
+        except subprocess.CalledProcessError:
+            return None
         return re.search(r"[0-9\.]+( [a-z]+)?$", output.strip()).group(0)
 
 
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -302,6 +302,7 @@ class Chrome(Browser):
         if uname[0] == "Darwin":
             return "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
         # TODO Windows?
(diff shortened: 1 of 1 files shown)


tools/wpt/browser.py
```diff
@@ -293,9 +293,17 @@ class Chrome(Browser):
...
-    binary = "/usr/bin/google-chrome"
     requirements = "requirements_chrome.txt"
 
+    @property
+    def binary(self):
+        if uname[0] == "Linux":
+            return "/usr/bin/google-chrome"
+        if uname[0] == "Darwin":
+            return "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
+        # TODO Windows?
```
kereliuk: Maybe we should have a debug log here that states that no matching uname was found so we didn't attempt to find the version binary?

tools/wpt/browser.py
```diff
@@ -339,7 +347,10 @@ def install_webdriver(self, dest=None):
 
     def version(self, binary=None):
         binary = binary or self.binary
-        version_string = call(binary, "--version").strip()
+        try:
+            version_string = call(binary, "--version").strip()
+        except subprocess.CalledProcessError:
```
kereliuk: Same suggestion about a debug log here.

tools/wpt/browser.py
```diff
@@ -379,9 +390,15 @@ class Opera(Browser):
...
 
     product = "opera"
-    binary = "/usr/bin/opera"
     requirements = "requirements_opera.txt"
 
+    @property
+    def binary(self):
+        if uname[0] == "Linux":
+            return "/usr/bin/opera"
+        # TODO Windows, Mac?
```
kereliuk: And here :)

tools/wpt/browser.py
```diff
@@ -430,7 +447,10 @@ def install_webdriver(self, dest=None):
     def version(self, binary):
         """Retrieve the release version of the installed browser."""
         binary = binary or self.binary
-        output = call(binary, "--version")
+        try:
+            output = call(binary, "--version")
+        except subprocess.CalledProcessError:
```
kereliuk: Final spot!

kereliuk (approved):
Looks good! Just some nits about adding some debug logs for when we don't match the uname or the subprocess fails.

tools/wpt/browser.py
```diff
@@ -293,9 +293,17 @@ class Chrome(Browser):
...
     product = "chrome"
-    binary = "/usr/bin/google-chrome"
     requirements = "requirements_chrome.txt"
 
+    @property
+    def binary(self):
+        if uname[0] == "Linux":
+            return "/usr/bin/google-chrome"
+        if uname[0] == "Darwin":
+            return "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
```
gsnedders: Technically, I don't think this is guaranteed. I think technically on macOS we should read Contents/Info.plist, read the CFBundleExecutable key, and then set the binary to Contents/MacOS/[that value]. That said, probably needless pedantry.

Hexcles: You're right; this is BAD(tm). Also the bundle might be installed to a user-specific applications directory. Not to mention I made zero effort for Windows...

I'm essentially using a simplified version of Chromedriver's heuristics: https://cs.chromium.org/chromium/src/chrome/test/chromedriver/chrome/chrome_finder.cc , which is also not ideal.

Basically the only reliable way is to always explicitly specify the --binary flag.

gsnedders: I at some point had some plan to implement some `find_macos_binary_path(bundle_id)`, which took the application id (e.g., `com.google.Chrome`) and then returns the path to the binary. Probably even splitting it up into two functions (one which finds the `.app` bundle, one which finds the binary within it).

Hexcles: Let's wish it'll happen some day :-P

PR #11004, opened 2018-05-15
Rename *-expected.html to *-ref.html

foolip:
While there is no rule or lint for the naming, *-expected.html is the
convention in WebKit/Blink, and in WPT *-ref.html is the norm. (These
were the only *-expected.html, and there are 3000+ *-ref.html.)


annevk (commented):
I don't understand the setup for css/css-fonts. What's referencing them?

foolip:
@annevk is there something funny in css/css-fonts you noticed while reviewing this? I didn't touch it in this PR :)

foolip:
Aha, I didn't see the whole path in the web UI. I renamed css/css-fonts/support/fonts/font-feature-settings-rendering-2-expected.html to css/css-fonts/support/fonts/font-feature-settings-rendering-2-ref.html and yet didn't change anything else about. That is suspect. (I used scripts to rename.)

foolip:
It's from commit 5e10afdc69111c73be880f2048093aeb3fb48187, but the companion font-feature-settings-rendering-2.html (also in support/fonts/) isn't a test, and obviously not a font either as one would expect. It looks quite a lot like something that would've originated from WebKit or Blink, but I can't find anything similar in either repo.

annevk (approved):
Please file a follow-up on sorting out the css/css-fonts thing.

foolip:
Yep, will sort it out on top of this.

PR #11006, opened 2018-05-15
Attempt to fix the font-feature-settings-rendering-2.html test

foolip:
Both the test and its ref was in support/fonts/ which doesn't make
sense. Move the test and ref and make some changes in the direction of
fixing it. It still doesn't pass in Chrome or Firefox.

Discovered in https://github.com/w3c/web-platform-tests/pull/11004.

### DIFF
--- a/css/css-fonts/font-feature-settings-rendering-2-ref.html
+++ b/css/css-fonts/font-feature-settings-rendering-2-ref.html
@@ -4,7 +4,7 @@
 <style>
 @font-face {
     font-family: "FontFeaturesTest";
-    src: url("FontWithFeatures.otf") format("opentype");
+    src: url("support/fonts/FontWithFeatures.otf") format("opentype");
 }
 </style>
 </head>
--- a/css/css-fonts/font-feature-settings-rendering-2.html
+++ b/css/css-fonts/font-feature-settings-rendering-2.html
@@ -1,10 +1,12 @@
 <!DOCTYPE html>
 <html>
 <head>
+<link rel="help" href="https://drafts.csswg.org/css-fonts/#font-rend-desc">
+<link rel="match" href="font-feature-settings-rendering-2-ref.html">
 <style>
 @font-face {
     font-family: "FontFeaturesTest";
-    src: url("resources/FontWithFeatures.otf") format("opentype");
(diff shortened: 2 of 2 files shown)


foolip:
@svgeesus can you help review this? It's from commit 5e10afdc69111c73be880f2048093aeb3fb48187 but still doesn't work with these changes. In Chrome, there's a console message about the font not parsing correctly. The test also doesn't work in Firefox, but no console message.

svgeesus:
Okay, I looked into this. These two files were not intended to be WebPlatform tests. They were just used in developing the special-purpose font used for a lot of CSS Fonts 3 tests. This font was originally developed as an internal test font for Safari (which is why it uses a webkit-specific property, and also why it doesn't work well except on OSX). `FontWithFeatures.otf` was further developed into `FontWithFeatures2.otf` (which also works on Linux and Windows) and eventually `FontWithFancyFeatures.otf` which is what the WPT tests use.

I modified a local copy of this test, as follows:

```
@font-face {
    font-family: "FontFeaturesTest";
    src: url("FontWithFancyFeatures.otf") format("opentype");
}
```

and

```
function addElement(feature, c) {
    var wrapper=document.createElement("div");
    ["0", "1"].forEach(function(state) {
        var element = document.createElement("span");
        element.textContent = c;
        element.style.fontFamily = "FontFeaturesTest";
        element.style.fontFeatureSettings = '"' + feature + '" ' + state;
        wrapper.appendChild(element);
    });
    insertionpoint.appendChild(wrapper);
    insertionpoint.appendChild(document.createTextNode(" "));
}
```
The test then ran correctly in Firefox, up to the letter `P`, then failed. Looking at the old font and the current one, I see the glyph indexes changed during development.

Basically these two tests can be deleted. Their actual functionality is covered by other tests. If it is thought valuable to test all the `font-feature-settings` in one test, I could submit that test later (the newest font tests more than the original font did).

I have a [test for the new font here](https://www.w3.org/People/chris/fwf/fancytester.html) which documents the character to opentype feature mapping for `FontWithFancyFeatures.otf`.

foolip:
Thanks @svgeesus! I've sent https://github.com/w3c/web-platform-tests/pull/11086 to remove this test, the ref and the support font.

PR #11010, opened 2018-05-15
[testharness.js] Formalize test types

jugglinmike:
Previously, tests for the testharness.js framework were validated
according to an implicit heuristic based on the presence of a `<script>`
tag.

Introduce a file naming convention to make this distinction explicit.
This pattern makes the validation behavior easier to understand. It also
supports future additions which further differentiate the way the two
test types are interpreted.

---

@ms2ger @tobie @gsnedders This may look familiar because we worked on a similar patch late last year: gh-8597. That fell out of sync with `master`, but while I was rebasing it, I decided that a different approach would be better. Instead of detecting test type via a custom `<meta>` tag, this patch references the test file's location on disk. I think this will be easier for others to discover and maintain.

As noted in the commit message, this patch is intended to facilitate future work. I'd like to extend the way the functional tests are interpreted.  Strengthening the delineation between those tests and the unit test should limit confusion as they become more distinct.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/11010)


resources/test/README.md
```diff
@@ -42,10 +42,27 @@ This should be followed by one or more `<script>` tags that interface with the
...
+
+- Ensuring that some tests are not run
+- Ensuring conditions that cause test failures
+- Ensuring conditions that cause harness errors
+
+Functional tests allow for these details to be verified. Every functional test
+must include a summary of the expected results as a JSON string within a
+`<script>` tag with an `id` of `"expected"`, e.g.:
+
+     <script type="text/json" id="expected">
```
Ms2ger: Nit: indentation is odd

resources/test/conftest.py
```diff
@@ -58,6 +69,27 @@ def repr_failure(self, excinfo):
...
+
+    def _run_unit_test(self):
+        driver = self.session.config.driver
+        server = self.session.config.server
+
+        driver.get(server.url(HARNESS))
+
+        actual = driver.execute_async_script('runTest("%s", "foo", arguments[0])' % server.url(str(self.filename)))
+
+        summarized = self._summarize(actual)
```
Ms2ger: Extract everything up to and including the `self._summarize()` call into another helper method? (Unless you have specific plans to diverge the two methods.)

Ms2ger (approved):
Please do change the PR summary / commit message to clarify that this is about th.js self-tests, not all tests that use th.js.

Please do wait for jgraham to review, but this SGTM.

jugglinmike:
> Please do change the PR summary / commit message to clarify that this is about th.js self-tests, not all tests that use th.js.

I'm happy to clarify, but I'm short on ideas. As written, the commit message body begins with the phrase, "Previously, tests for the testharness.js framework [...]" and I'm not sure how/where I can better communicate the distinction. Could you say more on how this could be improved?

> I'm guessing you need to update a path in `lint.whitelist`.

Right you are

resources/test/conftest.py
jugglinmike: Yes, I do have plans for these two methods to diverge further

jugglinmike:
Strangely, GitHub.com was reporting a merge conflict with `master` that I could not reproduce locally. My best guess is that we were experiencing [the situation reported in this GitHub support ticket](https://github.community/t5/How-to-use-Git-and-GitHub/Merge-conflicts-on-GitHub-not-locally/td-p/7184)--that the conflict has been resolved since the conflict detection logic last ran.

Whatever the case, that problem will block merging from the web UI, so I've rebased the branch for convenience's sake. The original version of this branch is available at:

https://github.com/w3c/web-platform-tests/compare/master...bocoup:testharness-tests-explicit-mode-2-prebase

Do you think this is ready for `master`, @Ms2ger?

resources/test/conftest.py
```diff
@@ -17,8 +17,10 @@ def pytest_addoption(parser):
     parser.addoption("--binary", action="store", default=None, help="path to browser binary")
 
 def pytest_collect_file(path, parent):
+    test_type = os.path.relpath(str(path), HERE).split(os.path.sep)[1]
```
Ms2ger: I feel a little queasy about this line, but I guess it's fine if you add a comment along the lines of "Tests are organized in directories by type".

resources/test/conftest.py
```diff
@@ -17,8 +17,10 @@ def pytest_addoption(parser):
...
+
     if path.ext.lower() == '.html':
-        return HTMLItem(str(path), parent)
+        return HTMLItem(str(path), test_type, parent)
```
Ms2ger: Nit: only compute `test_type` if you need it.

resources/test/conftest.py
```diff
@@ -58,6 +69,27 @@ def repr_failure(self, excinfo):
         return pytest.Collector.repr_failure(self, excinfo)
 
     def runtest(self):
+        if self.type == 'unit':
+            self._run_unit_test()
+        elif self.type == 'functional':
+            self._run_functional_test()
```
Ms2ger:
```python
else:
    raise NotImplementedError
```

Ms2ger (approved):
Please fix the last few comments, and do squash when landing.

resources/test/conftest.py
jugglinmike: Sure

resources/test/conftest.py
jugglinmike: Got it

resources/test/conftest.py
jugglinmike: Will do.

I didn't know that Python instantiates classes when they are a raised value. I'm trying to think of when that would be necessary.

jugglinmike:
> I feel a little queasy about this line,

That's the critical difference between this patch and [the previous version](https://github.com/w3c/web-platform-tests/pull/8597). I'm not married to this one, but I did think that a filesystem-based approach would be easier to discover and use. If your preference is for the `<meta>` tag-based solution (and you're up for another round of review), I'd be happy to close this and rebase that patch. Let me know!

Ms2ger:
No, the filesystem-based approach is fine, but I felt there should have been a slightly more readable way to write it (which I didn't find). Anyway, this is good to go.

jugglinmike:
Thanks, @Ms2ger!

Issue #8420, opened 2017-11-24
Get rid of web-platform.test references


gsnedders:
From #6868:

- [ ] clear-site-data/support/test_utils.sub.js
- [ ] content-security-policy/base-uri/report-uri-does-not-respect-base-uri.sub.html
- [ ] content-security-policy/generic/generic-0_8.sub.html
- [ ] content-security-policy/generic/generic-0_8_1.sub.html
- [ ] content-security-policy/nonce-hiding/script-nonces-hidden-meta.tentative.html
- [ ] content-security-policy/nonce-hiding/svgscript-nonces-hidden-meta.tentative.html
- [ ] fetch/api/request/request-structure.html
- [ ] html/browsers/origin/relaxing-the-same-origin-restriction/document_domain_setter.html
- [ ] html/semantics/embedded-content/the-iframe-element/cross_origin_parentage.html
- [ ] html/semantics/forms/the-label-element/label-attributes.html
- [ ] longtask-timing/longtask-in-childiframe-crossorigin.html
- [ ] longtask-timing/longtask-in-sibling-iframe-crossorigin.html
- [ ] navigation-timing/nav2_test_attributes_values.html
- [ ] navigation-timing/nav2_test_instance_accessors.html
- [ ] service-workers/service-worker/update-bytecheck.https.html
- [ ] webdriver/tests/cookies/add_cookie.py
- [ ] webdriver/tests/cookies/get_named_cookie.py
- [ ] webrtc/RTCPeerConnection-getIdentityAssertion.html
- [ ] webrtc/identity-helper.js

jugglinmike:
I'm on the case

jugglinmike:
Resolved via gh-11011.

PR #11011, opened 2018-05-15
Parameterize domain and port references

jugglinmike:
The web-platform-tests server may be configured to run from any host name and any TCP/IP port number. It offers a text substitution feature so that tests can be written to account for this variability.

Update tests which previously referenced the default host name and port number (i.e. `web-platform.test` and 8000) to instead reference parameterized values.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/11011)

### DIFF
--- a/content-security-policy/nonce-hiding/script-nonces-hidden-meta.tentative.sub.html
+++ b/content-security-policy/nonce-hiding/script-nonces-hidden-meta.tentative.sub.html
@@ -124,7 +124,7 @@
         var script = document.querySelector('#cssTest');
         var style = getComputedStyle(script);
         assert_equals(style['display'], 'block');
-        assert_equals(style['background-image'], "url(\"http://web-platform.test:8001/security/resources/abe.png\")");
+        assert_equals(style['background-image'], "url(\"http://{{domains[]}}:8001/security/resources/abe.png\")");
       }));
     }, "Nonces leak via CSS side-channels.");
 </script>
--- a/content-security-policy/nonce-hiding/script-nonces-hidden-meta.tentative.sub.html
+++ b/content-security-policy/nonce-hiding/script-nonces-hidden-meta.tentative.sub.html
@@ -124,7 +124,7 @@
         var script = document.querySelector('#cssTest');
         var style = getComputedStyle(script);
         assert_equals(style['display'], 'block');
-        assert_equals(style['background-image'], "url(\"http://{{domains[]}}:8001/security/resources/abe.png\")");
+        assert_equals(style['background-image'], "url(\"http://{{domains[]}}:{{ports[http][0]}}/security/resources/abe.png\")");
       }));
     }, "Nonces leak via CSS side-channels.");
 </script>
--- a/content-security-policy/nonce-hiding/svgscript-nonces-hidden-meta.tentative.sub.html
+++ b/content-security-policy/nonce-hiding/svgscript-nonces-hidden-meta.tentative.sub.html
@@ -109,7 +109,7 @@
           var script = document.querySelector('#cssTest');
           var style = getComputedStyle(script);
           assert_equals(style['display'], 'block');
-          assert_equals(style['background-image'], "url(\"http://{{domains[]}}:8001/security/resources/abe.png\")");
+          assert_equals(style['background-image'], "url(\"http://{{domains[]}}:{{ports[http][0]}}/security/resources/abe.png\")");
         }));
       }, "Nonces don't leak via CSS side-channels.");
   </script>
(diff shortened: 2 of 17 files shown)


jugglinmike:
This is intended to resolve gh-8420

content-security-policy/nonce-hiding/script-nonces-hidden-meta.tentative.sub.html
```diff
@@ -124,7 +124,7 @@
         var script = document.querySelector('#cssTest');
         var style = getComputedStyle(script);
         assert_equals(style['display'], 'block');
-        assert_equals(style['background-image'], "url(\"http://web-platform.test:8001/security/resources/abe.png\")");
+        assert_equals(style['background-image'], "url(\"http://{{domains[]}}:{{ports[http][0]}}/security/resources/abe.png\")");
```
annevk: That's 8000, not 8001, right?

content-security-policy/nonce-hiding/svgscript-nonces-hidden-meta.tentative.sub.html
```diff
@@ -109,7 +109,7 @@
           var script = document.querySelector('#cssTest');
           var style = getComputedStyle(script);
           assert_equals(style['display'], 'block');
-          assert_equals(style['background-image'], "url(\"http://web-platform.test:8001/security/resources/abe.png\")");
+          assert_equals(style['background-image'], "url(\"http://{{domains[]}}:{{ports[http][0]}}/security/resources/abe.png\")");
```
annevk: Same here.

navigation-timing/nav2_test_attributes_values.html
```diff
@@ -81,7 +92,7 @@ <h1>Description</h1>
                     var entries = entryList.getEntries();
                     assert_equals(entries[0].entryType, "navigation",
                         "Expected entryType to be: navigation.");
-                    assert_equals(entries[0].name, "http://web-platform.test:8001/navigation-timing/nav2_test_attributes_values.html");
+                    assert_equals(entries[0].name, expectedUrl);
```
annevk: This is 8000, not 8001, right?

navigation-timing/nav2_test_instance_accessors.html
```diff
@@ -50,10 +54,10 @@ <h1>Description</h1>
                 t.step_func(function (entryList) {
                     var instance1 = performance.getEntries()[0];
                     var instance2 = performance.getEntriesByType("navigation")[0];
-                    var instance3 = performance.getEntriesByName("http://web-platform.test:8001/navigation-timing/nav2_test_instance_accessors.html")[0];
+                    var instance3 = performance.getEntriesByName(expectedUrl)[0];
```
annevk: Same.

navigation-timing/nav2_test_instance_accessors.html
```diff
@@ -50,10 +54,10 @@ <h1>Description</h1>
...
 
                     assert_equals(performance.getEntriesByType("navigation").length, 1, "Expected there is only one navigation timing instance.");
-                    assert_equals(performance.getEntriesByName("http://web-platform.test:8001/navigation-timing/nav2_test_instance_accessors.html").length, 1, "Expected there is only one navigation timing instance.");
+                    assert_equals(performance.getEntriesByName(expectedUrl).length, 1, "Expected there is only one navigation timing instance.");
```
annevk: Same.

jugglinmike:
@annevk You are correct that for the default configuration, this patch modifies the referenced port from 8001 to 8000. This is intentional: in that configuration, port 8001 is unavailable. Changing the hard-coded and unavailable port number to a dynamic port number that *is* available modifies the test semantics. I believe this is a bug fix for the modified tests since none of them appear to expect the reference to be unavailable. (In most cases, such tests should be using `nonexistent.{{domains[]}}` anyway).

annevk:
@jugglinmike but don't you turn a cross-origin reference into a same-origin one? Or did those tests expect to be loaded over 8001 too?

jugglinmike:
It's about time that I get less hand-wavy about semantics of the tests in question.

- `content-security-policy/nonce-hiding/script-nonces-hidden-meta.tentative.sub.html`
- `content-security-policy/nonce-hiding/svgscript-nonces-hidden-meta.tentative.sub.html`

The modified URL in these tests is used to assert the computed style value for the `backgroundImage` of an element. The value is set using a root-relative value, so the port number is expected to reflect the port of the surrounding document.

- `navigation-timing/nav2_test_attributes_values.html`
- `navigation-timing/nav2_test_instance_accessors.html`

The modified URL in these tests is used to assert the value of the `name` property of `PerformanceEntry` instances. Both tests concern [the `PerformanceNavigationTiming`](https://www.w3.org/TR/navigation-timing-2/#dom-performancenavigationtiming) specifically, and that's specified as:

> The name attribute MUST return the DOMString value of the address of the current document.

For all 5 URLs modified in these 4 files, by hard-coding the port as 8001, these tests would only pass when loaded using that particular port. Parameterizing the port allows the tests to pass for alternate configurations. Does that sound right to you?

annevk:
Yeah, thanks for verifying. These changes look good then.

annevk (approved):

jugglinmike:
Thanks, @annevk!

PR #11016, opened 2018-05-15
Refactor and split up with `variant` all encoding tests

### LINKED ISSUE #11015: /encoding/* are disabled in Chromium and WebKit
https://bocoup.github.io/wpt-disabled-tests-report/

/encoding/legacy-mb-tchinese/big5/big5-encode-form-errors-extBa.html

and

/encoding/legacy-mb-tchinese/big5/big5-encode-form-errors-extBb.html

should be split up, maybe similarly to how it was done in https://github.com/w3c/web-platform-tests/pull/10984

zcorpan:
Fixes #11015.


zcorpan:
I'll apply this to the rest of the encoding tests tomorrow if this seems OK.

common/subset-tests.js
```diff
@@ -0,0 +1,27 @@
...
+//   }
+// }
+(function() {
+  var subTestStart = 0;
+  var subTestEnd = Infinity;
+  var match;
+  if (location.search) {
+      match = /^\?(\d+)-(\d+|last)(?:&|$)/.exec(location.search);
+      if (match) {
+        subTestStart = match[1];
```
inexorabletash: nit: parseFloat(match[1]) or parseInt(match[1], 10)

common/subset-tests.js
```diff
@@ -0,0 +1,27 @@
...
+        if (match[2] !== "last") {
+            subTestEnd = match[2];
```
inexorabletash: ditto

common/subset-tests.js
inexorabletash: rather than relying on string/number comparison later on. (Of course, it's all per spec, so totally optional)

common/subset-tests.js
```diff
@@ -0,0 +1,27 @@
+// Only test a subset of tests with, e.g., ?1-10 in the URL
```
inexorabletash: Maybe add that this works well with the  `<meta name="variant" ... >` mechanism ?

encoding/legacy-mb-tchinese/big5/big5-encode-form-errors-extBa.html
```diff
@@ -4,10 +4,33 @@
...
+<meta name="variant" content="?17001-18000">
+<meta name="variant" content="?18001-19000">
+<meta name="variant" content="?19001-20000">
+<meta name="variant" content="?20001-21000">
+<meta name="variant" content="?21001-last">
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <script src="big5_index.js"></script>
 <script src="big5-encoder.js"></script>
+<script src="/common/subset-tests.js"></script>
```
inexorabletash: Just as a style thing, I'd expect /common/ scripts to be listed under /resources/ and before the module- or test-specific scripts.

common/subset-tests.js
```diff
@@ -0,0 +1,27 @@
...
+// Sample usage:
+// let currentSubTest = 0;
```
inexorabletash: Hmmm... so every test is going to end up with this sample usage. What do you think about altering (or adding) the API to take care of this, i.e. just exposing a function (e.g. "subsetTest") that takes a closure, which handles the counter/test for the author?
```

jimsch:
If this is going to become a common function rather than restricted to a single set of code, I worry about the fact that others are going to want to be able to do both multiple types of variants and combined variants together so that you say one of <variantA> and one of <variantB>  That means you might need to have a more complicated parsing function

common/subset-tests.js
zcorpan: I think this is more flexible and more straightforward to apply to existing tests, since it can vary a lot how tests are being set up. But maybe we can support both somehow...

encoding/legacy-mb-tchinese/big5/big5-encode-form-errors-extBa.html
zcorpan: Utility code is usually in common I think. I can move the `<script>` though.

inexorabletash: Yeah, sorry, I just meant the script tag

zcorpan:
OK I have addressed the comments, PTAL

annevk:
Should this land together with the Web Crypto changes? Definitely need a different commit message in that case.

It'd be good to document common/subset-tests.js in the documentation. So that when people read about variant (or more likely, forget how it works) they also learn about this.

zcorpan:
> Should this land together with the Web Crypto changes? Definitely need a different commit message in that case.

That is now https://github.com/w3c/web-platform-tests/pull/11028

> documentation

I'll look into that now.

annevk:
Thanks, that sounds pretty great. Can you put the non-test resources in a `resources/` directory? The other thing I hope we can do at some point is remove the need for `normalizeStr`, but that'll be easier with this change.

Does this not accidentally revert the removal of `normalizeStr` I did so far?

zcorpan:
@annevk PTAL

annevk (approved):
Thanks for doing this. I haven't verified before and after for all of these, but I believe you have, and you covered all the gotchas I could think of.

These tests still make me a little angsty given how much code is involved, but this makes it a lot better.

zcorpan:
master has `normalizeStr(item.expected)` instead of just `item.expected` for the test name for that test. Do you have a preference? We could leave out expected from the test name altogether.

annevk:
Ah, this was removed because of `return result.replace(/%1B%28%42$/, "");`. I guess I'm fine with your result then.

zcorpan:
The chrome and firefox builds failed, but i suppose that's because it's many slow tests.

PR #11024, opened 2018-05-16
Don't assume file_is_test when there's a top-level non-assert error

jgraham:
Previously, if we got an unexpected error before any asserts ran, we
would assume that the file represented a single-test file and imply a
subtest. However this is confusing in the case that the assumptions
are wrong because we see different subtests in browsers that have an
early error vs those that do not.

Instead of this, we only set file_is_test when we call assert()
outside a test, and in other cases set the overall status to ERROR and
leave the list of subtests empty.

### DIFF
--- a/resources/testharness.js
+++ b/resources/testharness.js
@@ -770,7 +770,7 @@ policies and contribution forms [3].
     }
 
     function done() {
-        if (tests.tests.length === 0) {
+        if (tests.tests.length === 0 && tests.status.status === tests.status.OK) {
             tests.set_file_is_test();
         }
         if (tests.file_is_test) {
@@ -3366,10 +3366,6 @@ policies and contribution forms [3].
 
     if (global_scope.addEventListener) {
         var error_handler = function(e) {
-            if (tests.tests.length === 0 && !tests.allow_uncaught_exception) {
-                tests.set_file_is_test();
-            }
-
             var stack;
             if (e.error && e.error.stack) {
                 stack = e.error.stack;
@@ -3391,6 +3387,9 @@ policies and contribution forms [3].
                 tests.status.status = tests.status.ERROR;
                 tests.status.message = e.message;
                 tests.status.stack = stack;
+                if (!tests.length) {
+                    tests.phase = tests.phases.ABORTED;
+                }
             }
             done();
         };
--- a/infrastructure/metadata/infrastructure/expected-fail/uncaught-exception.html.ini
+++ b/infrastructure/metadata/infrastructure/expected-fail/uncaught-exception.html.ini
@@ -1,4 +1,3 @@
 [uncaught-exception.html]
-  [Uncaught exception]
-    expected: FAIL
+  expected: ERROR
 


zcorpan:
I think this is an improvement, but also it would be nice if single page tests can get an OK FAIL instead of ERROR if the test fails before getting to an assert. Adding `assert_true(true, 'is single page test');` or so to achieve that seems worse than a dedicated opt-in imo.

foolip:
Looking at https://github.com/web-platform-tests/wpt/issues/16229#issuecomment-526665338 and similar queries (like `chrome:!unknown firefox:!unknown safari:unknown`) today I found that fairly often when a subtests existence is inconsistent, it's because in some browser there was an early exception which triggered the single page test logic instead of a harness error. The name of that subtest won't quite make sense, and it took more poking than ideal to understand the cause.

A harness error in these cases would be much preferred I think, even if also think https://github.com/web-platform-tests/wpt/pull/11364 would be great.

I'll rebase this branch and trigger full runs to see what tests are affected by this smaller change.

foolip:
Sent https://github.com/web-platform-tests/wpt/pull/18794 adding a test that this PR would have to update expectations for.

foolip:
I've triggered full Edge and Safari runs to begin with:
https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=29339

jugglinmike:
@foolip It looks like you've force-pushed the patch from gh-18794 to this branch. Was that intentional? If not, I can restore the branch to its prior state.

foolip:
@jugglinmike no that wasn't intentional, I merged the branch for https://github.com/web-platform-tests/wpt/pull/18794 into the branch for this locally for testing and then did `git rebase HEAD~ --onto c05529a9dfbda874a8f1f48eb20248a794c52e71` and forgot that I had that extra commit on the branch.

Rebased again to the current epochs/six_hourly and triggered a new run:
https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=29347

foolip:
Diff views for [Edge](https://wpt.fyi/results/?diff&filter=ADC&run_id=289790006&run_id=285580005) and [Safari](https://wpt.fyi/results/?diff&filter=ADC&run_id=301040005&run_id=305000008) available. Searching for `status:unknown` should filter to results where subtests didn't line up.

foolip:
Filed https://github.com/web-platform-tests/wpt.fyi/issues/1460 about the ability to spot *any* difference in these diffs, not just the ones I know to look for.

Skimming the results, a lot of the changes are exactly the sorts of improvements I hoped for, like this change from failing bogus subtest to harness error:
https://wpt.fyi/results/streams/readable-byte-streams/construct-byob-request.any.html?diff&filter=ADC&run_id=289790006&run_id=285580005
https://wpt.fyi/results/streams/readable-byte-streams/construct-byob-request.any.html?diff&filter=ADC&run_id=301040005&run_id=305000008

Even better when it also causes some of the subtests to be known and NOTRUN:
https://wpt.fyi/results/content-security-policy/nonce-hiding/svgscript-nonces-hidden.tentative.html?diff&filter=ADC&run_id=301040005&run_id=305000008

However, this needs investigating:
- Any change that doesn't fit the pattern of failing subtest &rarr; harness error. Some went from pass bogus subtest to many passing reasonably named subtests, with no error in sight.
- The runs took longer. Seems like failing in this way takes longer, but could just as well be instantaneous...?

foolip:
#18794 was just merged, and https://wpt.fyi/results/infrastructure/expected-fail/uncaught-exception.html?run_id=271660008&run_id=302890002&run_id=281490003 is a distilled example of the failure pattern that this would address.

foolip:
Rebased and triggered Chrome and Firefox this time:
https://tools.taskcluster.net/groups/XUjjdKP2RluLOegJQ5OJyw
https://wpt.fyi/results/?diff&filter=ADC&run_id=279760012&run_id=297090001
https://wpt.fyi/results/?diff&filter=ADC&run_id=291480011&run_id=279790003

Affected by https://github.com/web-platform-tests/wpt.fyi/issues/1462, so staging links are:
https://staging.wpt.fyi/results/?diff&filter=ADC&run_id=253070013&run_id=255070005
https://staging.wpt.fyi/results/?diff&filter=ADC&run_id=257120017&run_id=235160006

As with Edge and Safari, the runs now take more time, so that has to be solved somehow.

resources/testharness.js
```diff
@@ -770,7 +770,7 @@ policies and contribution forms [3].
     }
 
     function done() {
-        if (tests.tests.length === 0) {
+        if (tests.tests.length === 0 && tests.status.status === tests.status.OK) {
```
foolip: @jgraham do you recall why you made this change? `test.status.status` is initially null, so this changed the effects of very simple single-page tests like resources/tests/functional/single-page-test-no-assertions.html and content-security-policy/script-src/scripthash-default-src.sub.html.

jgraham: I don't remember, it sounds like it should also allow that the status is null.

foolip:
This has been subsumed by https://github.com/web-platform-tests/rfcs/blob/master/rfcs/single_test.md so closing.

Sandhukhizar61 (changes requested):
Good

PR #11028, opened 2018-05-16
Generalize subsetTest() in common/ and use for WebCryptoAPI tests

zcorpan:
Taken from #11016, will rebase that on top of this.


inexorabletash (approved):
Looks nice!

PR #11041, opened 2018-05-17
Handle the case where git is missing when using vcs.GitTree.

jgraham:
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/11041)


Hexcles (approved):
r+

PR #11052, opened 2018-05-17
Move entry check before sendBeacon() call

nolanlawson:
Fixes w3c/beacon#58

See discussion there for details.


toddreifsteck (approved):
Looks good!

PR #11053, opened 2018-05-17
Update sendBeacon test case array to include recently added test cases

Brandr0id:
Updating the allTests array to ensure that newly added test cases appear in the array also.


toddreifsteck (approved):
LGTM

PR #11076, opened 2018-05-18
Use add_dependency_idls in gamepad idlharness test

lukebjerring:
To test, copied the same file to Chromium and ran 

    python third_party/blink/tools/run_web_tests.py -t Default external/wpt/gamepad/idlharness.html

    
>    Found 1 test; running 1, skipping 0.               
>    Running 1 content_shell.                                                                              

>    The test ran as expected.


foolip (approved):
Note that the lack of first level of indentation inside `<script>` is very common in WPT and probably not an accident. But no style is enforced, so I'll LGTM changes in either direction :)

PR #11078, opened 2018-05-18
Service Worker: assert visibility of resources fetched via IFRAME srcdoc and src=about:blank

cvazac:
Per [spec clarification](https://github.com/whatwg/html/pull/3714).

Locally, this passes in Firefox 59, Safari 11.1. Fails in Chrome 66, 67, and 68.


wanderview:
Not sure if you are looking for review yet, but some notes:

* The test files should use the `.https.html` file extension so that they get run using TLS.  This is necessary for service workers to run reliably off localhost.
* Files that are not specifically tests (like scripts, iframe documents, etc) should go in the resources directory.
* Please use the helper functions like `service_worker_unregister_and_register()`, etc.

cvazac:
Thanks for the comments @wanderview!

> Please use the helper functions like service_worker_unregister_and_register(), etc.

I didn't use `service_worker_test` because I need to do some scripting in the html file. And I didn't use `service_worker_unregister_and_register` because I don't have a test that needs to run from inside the html, just from inside the service worker. If there's a different pattern I should adopt, please let me know.

wanderview:
I think it would be easiest to just use a promise_test in the html file and then call fetch_tests_from_worker() there.  Like this:

https://github.com/web-platform-tests/wpt/blob/master/service-workers/service-worker/client-navigate.https.html#L46

Then you can use all our helper functions.

Also, the tests need to clean up both the iframe and the registration.  Since your scopes are unique you can use `add_completion_callback()` to do that.  Something like:

```javascript
// after registering the SW...
add_completion_callback(_ => reg.unregister());

// after creating the iframe...
add_completion_callback(_ => frame.remove());
```

cvazac:
Sorry for the delay on this. Can you PTAL @wanderview? 

I couldn't get it quite right, the second test fails with:
```
wait_for_state must be passed a ServiceWorker
```

`reg.installing` exists, but `reg.installing.state` is `undefined`.

I'm now using `service_worker_unregister_and_register` and other helpers. I'm using `service_worker_unregister` instead of `add_completion_callback(_ => reg.unregister());` because I thought its promise would ensure the first test was all cleaned up before starting the second. 

One I get them both working, I'll write a helper method for both tests to share, because they share a lot of boilerplate.

cvazac:
@wanderview friendly ping :) any advice would be greatly appreciated!

wanderview:
You need to make the scope strings unique for each test case.  Reusing the same scope can cause the second test resurrect the first test's registration.  This means you end up with a null installing worker and an active worker instead.

PR #11080, opened 2018-05-18
Allow testdriver to return values

kereliuk:
This is tracked by #10716

Still need to write an infrastructure test and write some documentation.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/11080)


gsnedders (commented):
Modulo these two nits, this looks fine

gsnedders:
@kereliuk How did you plan on testing this? This seems like it shouldn't be testable in isolation? Docs would be good, though.

gsnedders:
@kereliuk ^^

tools/wptrunner/wptrunner/executors/base.py
```diff
@@ -531,14 +532,18 @@ def process_action(self, url, payload):
             except KeyError:
                 raise ValueError("Unknown action %s" % action)
             try:
-                action_handler(payload)
+                value = action_handler(payload)
+                self.logger.debug("Recieved value: %s" % value)
```
jugglinmike: Typo: "Recieved" -> "Received"

foolip:
Someone else will have to take this over, so I'll close and point here from the issue, for when we prioritize the issue..

PR #11086, opened 2018-05-21
Delete font-feature-settings-rendering-2.html with font

foolip:
Because the "test" is in support/fonts/ it has not been run as a test.

See https://github.com/w3c/web-platform-tests/pull/11006#issuecomment-390436319
for where these came from and why they can be deleted.


svgeesus (approved):
agreed

PR #11088, opened 2018-05-21
Add async-local-storage/OWNERS

zcorpan:
@hiroshige-g I'm guessing you are Hiroshige Hayashizaki in https://github.com/w3c/web-platform-tests/commits/master/async-local-storage , correct?

### DIFF
--- a/async-local-storage/OWNERS
+++ b/async-local-storage/OWNERS
@@ -1,0 +1,1 @@
+@hiroshige-g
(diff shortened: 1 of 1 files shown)


hiroshige-g:
Yes, that's me, but in this case @domenic is the proper OWNER. The original author of the test is @domenic and I committed the test on behalf.

domenic (approved):
Thanks both.

PR #11092, opened 2018-05-21
Lint that META.yml files exist

### LINKED ISSUE #10552: Idea for new lint check: missing OWNERS for new top-level directory
Related to https://github.com/w3c/web-platform-tests/issues/5915

It'd be nice if new top-level directories have OWNERS files from the start when they are created (or subdirectory of css/).

cc @gsnedders

zcorpan:
Fixes #10552.

---

This should either land after ~#11087~, ~#11088~, ~#11089~, or add them to `lint.whitelist`.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/11092)


tools/lint/lint.py
```diff
@@ -264,6 +264,55 @@ def check_css_globally_unique(repo_root, paths):
...
+        if len(parts) <= level or len(parts) > level + 1:
+            continue
+        if not (parts[level]) in dirs:
+            dirs[parts[level]] = False
+        offset = path.find("/OWNERS")
+        if offset != -1:
+            dirs[parts[level]] = True
+
+    for dir in dirs:
+        if dirs[dir] == False:
```
gsnedders: just use `if not dirs[dir]` rather than comparing to False

gsnedders (changes requested):
I think this generally looks good (slightly different approach than I'd take, but not sure it's better or worse), though lots of small things.

Also one big thing: tests! This makes a lot of the existing tests fail with MISSING-OWNERS and it doesn't have any new tests (which given I'm pretty sure I found bugs reading the code… :))

tools/lint/lint.py
```diff
@@ -264,6 +264,55 @@ def check_css_globally_unique(repo_root, paths):
...
+def check_owners(repo_root, paths):
+    """
+    Checks that all top-level directories have OWNERS files.
+
+    :param repo_root: the repository root
+    :param paths: list of all paths
+    :returns: a list of errors found in ``paths``
+
+    """
+    dirs = {}
```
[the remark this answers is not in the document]
zcorpan: Right.

tools/lint/lint.py
```diff
@@ -264,6 +264,55 @@ def check_css_globally_unique(repo_root, paths):
...
+        parts = source_file.dir_path.split(os.path.sep)
+        if (parts[0] in source_file.root_dir_non_test or
+            any(item in source_file.dir_non_test for item in parts) or
+            any(parts[:len(non_test_path)] == list(non_test_path) for non_test_path in source_file.dir_path_non_test)):
+            continue
+        level = 0
+        for dir in owner_in_subdirs:
+            if path.find(dir) == 0:
+                level = len(dir.split("/"))
+        if len(parts) <= level or len(parts) > level + 1:
```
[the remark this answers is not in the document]
zcorpan: I intended to report an error if foo/bar/OWNERS exists but foo/OWNERS does not. But you're right that this fails to report missing OWNERS if only foo/bar/test.html exists.

tools/lint/lint.py
gsnedders: This is outstanding.

tools/lint/lint.py
```diff
@@ -286,25 +287,18 @@ def check_owners(repo_root, paths):
...
+                level = dir.count("/") + 1
+        if len(parts) <= level:
             continue
-        if not (parts[level]) in dirs:
-            dirs[parts[level]] = False
-        offset = path.find("/OWNERS")
-        if offset != -1:
-            dirs[parts[level]] = True
+        key = "/".join(parts[:level + 1])
+        if not key in dirs:
```
gsnedders: We should be able to move the following if statement up here, and then have this as an `elif` branch.

zcorpan:
OK @gsnedders I'm now getting these errors in my new tests, I don't understand why. Can you help?

<details><summary>ze errors</summary>

```
====================================================================================================================================== FAILURES ======================================================================================================================================
________________________________________________________________________________________________________________________ test_check_owners_test_l1_no_owners _________________________________________________________________________________________________________________________

caplog = <_pytest.logging.LogCaptureFixture object at 0x104aeded0>

    def test_check_owners_test_l1_no_owners(caplog):
        with _mock_lint("check_path") as mocked_check_path:
            with _mock_lint("check_file_contents") as mocked_check_file_contents:
                rv = lint(_dummy_repo, ["test_owners/okay.html"], "normal")
>               assert rv == 1
E               assert 0 == 1

lint/tests/test_lint.py:414: AssertionError
________________________________________________________________________________________________________________________ test_check_owners_test_l2_owners_l1 _________________________________________________________________________________________________________________________

caplog = <_pytest.logging.LogCaptureFixture object at 0x104681190>

    def test_check_owners_test_l2_owners_l1(caplog):
        with _mock_lint("check_path") as mocked_check_path:
            with _mock_lint("check_file_contents") as mocked_check_file_contents:
                rv = lint(_dummy_repo, ["test_owners/foo/okay.html", "test_owners/OWNERS"], "normal")
>               assert rv == 0
E               assert 1 == 0

lint/tests/test_lint.py:441: AssertionError
-------------------------------------------------------------------------------------------------------------------------------- Captured stdout call --------------------------------------------------------------------------------------------------------------------------------
test_owners: Directory is missing OWNERS file. (MISSING-OWNERS)

There was 1 error (MISSING-OWNERS: 1)
You must fix all errors; for details on how to fix them, see
https://web-platform-tests.org/writing-tests/lint-tool.html

However, instead of fixing a particular error, it's sometimes
OK to add a line to the lint.whitelist file in the root of the
web-platform-tests directory to make the lint tool ignore it.

For example, to make the lint tool ignore all 'MISSING-OWNERS'
errors in the test_owners file,
you could add the following line to the lint.whitelist file.

MISSING-OWNERS: test_owners
--------------------------------------------------------------------------------------------------------------------------------- Captured log call ----------------------------------------------------------------------------------------------------------------------------------
lint.py                    785 ERROR    test_owners: Directory is missing OWNERS file. (MISSING-OWNERS)
lint.py                    816 INFO     
lint.py                    818 INFO     There was 1 error (MISSING-OWNERS: 1)
lint.py                    946 INFO     You must fix all errors; for details on how to fix them, see
lint.py                    946 INFO     https://web-platform-tests.org/writing-tests/lint-tool.html
lint.py                    946 INFO     
lint.py                    946 INFO     However, instead of fixing a particular error, it's sometimes
lint.py                    946 INFO     OK to add a line to the lint.whitelist file in the root of the
lint.py                    946 INFO     web-platform-tests directory to make the lint tool ignore it.
lint.py                    946 INFO     
lint.py                    946 INFO     For example, to make the lint tool ignore all 'MISSING-OWNERS'
lint.py                    946 INFO     errors in the test_owners file,
lint.py                    946 INFO     you could add the following line to the lint.whitelist file.
lint.py                    946 INFO     
lint.py                    946 INFO     MISSING-OWNERS: test_owners
```

</details>

gsnedders:
@zcorpan You didn't commit test_owners.

zcorpan:
OK check again. (I got the 2 failures locally so wasn't related to forgetting to commit that dir.)

gsnedders:
`test_check_owners_test_l1_no_owners` is failing because of `if len(parts) == 1: continue # Skip top-level files`, because for `foo/bar.html` that is `foo` (i.e., it needs to be `if not parts`).

gsnedders:
The above issue is actually the problem with both of the two failing tests. However, fixing it causes other existing tests to start failing MISSING-OWNERS.

zcorpan:
OK now I get one failure, after having fixed the path to about_blank-ref.html it fails.

<details><summary>failures</summary>

```
====================================================================================================================================== FAILURES ======================================================================================================================================
______________________________________________________________________________________________________________________________ test_about_blank_as_ref _______________________________________________________________________________________________________________________________

caplog = <_pytest.logging.LogCaptureFixture object at 0x10e2347b8>

    def test_about_blank_as_ref(caplog):
        with _mock_lint("check_path") as mocked_check_path:
            with _mock_lint("check_file_contents") as mocked_check_file_contents:
                rv = lint(_dummy_repo, ["ref/about_blank-ref.html", "ref/OWNERS"], "normal")
>               assert rv == 0
E               assert 1 == 0

lint/tests/test_lint.py:220: AssertionError
-------------------------------------------------------------------------------------------------------------------------------- Captured stdout call --------------------------------------------------------------------------------------------------------------------------------
ref/about_blank-ref.html: Reference test with a non-existent 'match' relationship reference: 'about:blank' (NON-EXISTENT-REF)

There was 1 error (NON-EXISTENT-REF: 1)
You must fix all errors; for details on how to fix them, see
https://web-platform-tests.org/writing-tests/lint-tool.html

However, instead of fixing a particular error, it's sometimes
OK to add a line to the lint.whitelist file in the root of the
web-platform-tests directory to make the lint tool ignore it.

For example, to make the lint tool ignore all 'NON-EXISTENT-REF'
errors in the ref/about_blank-ref.html file,
you could add the following line to the lint.whitelist file.

NON-EXISTENT-REF: ref/about_blank-ref.html
--------------------------------------------------------------------------------------------------------------------------------- Captured log call ----------------------------------------------------------------------------------------------------------------------------------
lint.py                    785 ERROR    ref/about_blank-ref.html: Reference test with a non-existent 'match' relationship reference: 'about:blank' (NON-EXISTENT-REF)
lint.py                    816 INFO     
lint.py                    818 INFO     There was 1 error (NON-EXISTENT-REF: 1)
lint.py                    946 INFO     You must fix all errors; for details on how to fix them, see
lint.py                    946 INFO     https://web-platform-tests.org/writing-tests/lint-tool.html
lint.py                    946 INFO     
lint.py                    946 INFO     However, instead of fixing a particular error, it's sometimes
lint.py                    946 INFO     OK to add a line to the lint.whitelist file in the root of the
lint.py                    946 INFO     web-platform-tests directory to make the lint tool ignore it.
lint.py                    946 INFO     
lint.py                    946 INFO     For example, to make the lint tool ignore all 'NON-EXISTENT-REF'
lint.py                    946 INFO     errors in the ref/about_blank-ref.html file,
lint.py                    946 INFO     you could add the following line to the lint.whitelist file.
lint.py                    946 INFO     
lint.py                    946 INFO     NON-EXISTENT-REF: ref/about_blank-ref.html
```

</details>

gsnedders:
@zcorpan this is now failing because staticrange in master has no OWNERS, which looks like the failure is correct!

tools/lint/lint.py
```diff
@@ -264,6 +264,54 @@ def check_css_globally_unique(repo_root, paths):
...
+        if len(parts) <= level:
+            continue
+        key = "/".join(parts[:level + 1])
+        if key not in dirs:
+            dirs[key] = False
+        if path.endswith("/OWNERS") and len(parts) == level + 1:
+            dirs[key] = True
+
+    for dir in dirs:
+        if dir == "":
```
gsnedders: So for the key to be `""`, `"/".join(parts[:level + 1])` needs to be `""`, which will happen if `parts[:level + 1]` is `[]` or `[""]`. 

The former can only happen when parts is [] which AFAICT is impossible (because it comes from str.split), so it must be `[""]`, which implies `source_file.dir_path` is `os.path.sep`, which it shouldn't ever be.

*digs into deeper*

gsnedders: So `source_file.dir_path` is of course empty for when we have files at the top-level.

`/check_stability.ini` gives a `source_file.dir_path` of `""` and `source_file.filename` of `check_stability.ini`. So where I was wrong was saying `not parts` suffices for top-level files. I'd guess we should just go for `if not source_file.dir_path`.

gsnedders (approved):
This LGTM. I note that the Travis failure is down to a new directory with no OWNERS. I suggest we land this ASAP with that whitelisted so we stop it from getting worse.

zcorpan:
I'll rename this per https://github.com/web-platform-tests/wpt/pull/11201

zcorpan:
Do we want to lint that suggested_reviewers have write access?

gsnedders:
@zcorpan No, because that violates the policy on lint not having external dependencies.

gsnedders:
We're continuing to add new directories that are failing this lint. Can you please add those directories so we can land this ASAP before it regresses further?

zcorpan:
https://github.com/web-platform-tests/wpt/pull/11559 is the only outstanding PR. I'll add it to the whitelist so we can merge this.

zcorpan:
I've added a commit to lint the contents of META.yml.

gsnedders:
Note this got reverted in #11723.

PR #11099, opened 2018-05-22
Remove PaymentCurrencyAmount's currencySystem member

foolip:
Matches https://github.com/w3c/payment-request/pull/694.

### DIFF
--- a/interfaces/payment-request.idl
+++ b/interfaces/payment-request.idl
@@ -27,8 +27,6 @@ dictionary PaymentMethodData {
 dictionary PaymentCurrencyAmount {
   required DOMString currency;
   required DOMString value;
-  // Note: currencySystem is "at risk" of being removed!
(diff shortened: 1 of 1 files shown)


foolip:
@lukebjerring, this file says "GENERATED CONTENT - DO NOT EDIT" at the top. Should we remove all of those until these files are auto-updating?

lukebjerring (approved):

lukebjerring:
No; I think we're better off having some indication of which files got some way through the automated import behaviour until we make a permanent move on how that'll be handled.

marcoscaceres:
I'm ok with keeping the "GENERATED CONTENT - DO NOT EDIT"  as long as there are some assurances about timing. `currencySystem` was dropped about a month ago. Also, new IDL is being added to the spec, so worried about things not getting sync'ed in a timely manner.

marcoscaceres (approved):

Issue #10881, opened 2018-05-07
WebDriver test:extension capability tests should be removed


InstyleVII:
The tests validating that a WebDriver imlpementation can support extension capabilities should be removed. These tests are currently located in the webdriver/tests/sessions/new_session/create_firstMatch and create_alwaysMatch files.

The argument for removing them is that they are only validating that the browsers support parsing test extension capabilities which don't match the prefix a browser would actually use (e.g. Edge uses a vendor prefix of "ms" not "test").

The other argument for removal of these tests is that they are testing a section of the WebDriver specification that is specifically supposed to be browser-specific. If a browser implementation wants to create it's own extension capability tests for it's actual extension capabilities that can be done per-browser, but validating this with fake test extension capabilities doesn't make sense.

InstyleVII:
@AutomatedTester and @shs96c thoughts on this?

PR #11108, opened 2018-05-23
Removed extension capability tests

instylevii:
Remove the extension capability tests referenced by #10881. The core issue is that to pass these tests browsers must add support for a fake capability, which doesn't even test the specific browsers capability extension but rather it's ability to parse a ":" in the capability name.

### DIFF
--- a/webdriver/tests/new_session/support/create.py
+++ b/webdriver/tests/new_session/support/create.py
@@ -11,5 +11,4 @@ valid_data = [
                   {"script": 500},
                   {}]),
     ("unhandledPromptBehavior", ["dismiss", "accept", None]),
-    ("test:extension", [True, "abc", 123, [], {"key": "value"}, None]),
 ]


andreastt:
Implementations are supposed to ignore extension capabilities they don’t recognise. There are three use cases for this:

1. For example a user may provide the equivalent of `moz:firefoxOptions` for every browser in the same capabilities object so that whichever browser is matched gets the correct configuration.

2. An intermediary may intercept any extension capability it likes.

3. Provide configuration to intermediary nodes such as Sauce, e.g. `{"sauce:token", <uuid>}`.

cc @jgraham

InstyleVII:
That seems reasonable however I'd note the specification doesn't state this which is why our first take at implanting this throws the invalid capability error for these tests. This should be added to the spec.

On the intermediary nodes I'd expect them to strip their capabilities once they encounter them as they forward the request along. I wouldn't expect a browser to accept `{"sauce:token", <uuid>}`. If they do is the expectation then that if the capability name has a ":" in it that browser should try to parse and if it doesn't work ignore it? Otherwise I don't see when invalid capabilities (aside from a bad argument to a known capability) would be thrown.

shs96c:
Step 3 of the [matching capabilities](https://w3c.github.io/webdriver/#dfn-matching-capabilities) algorithm states: 

```If name is the key of an extension capability, let value be the result of trying implementation-specific steps to match on name with value. If the match is not successful, return success with data null.```

Which gives the behaviour @andreastt points to about implementations ignoring unrecognised extension capabilities.

InstyleVII:
That makes sense to me, closing the PR.

PR #11110, opened 2018-05-23
[tools] Enable additional linting rules

jugglinmike:
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/11110)

### DIFF
--- a/tools/tox.ini
+++ b/tools/tox.ini
@@ -40,10 +40,8 @@ select = E,W,F,N
 # E901: SyntaxError or IndentationError
 # W601: .has_key() is deprecated, use ‘in’
 # F401: module imported but unused
-# F403: ‘from module import *’ used; unable to detect undefined names
-# F405: name may be undefined, or defined from star imports: module
 # N801: class names should use CapWords convention
 # N802: function name should be lowercase
-ignore = E128,E129,E221,E226,E231,E251,E265,E302,E303,E305,E402,E731,E901,W601,F401,F403,F405,N801,N802
+ignore = E128,E129,E221,E226,E231,E251,E265,E302,E303,E305,E402,E731,E901,W601,F401,N801,N802
 max-line-length = 141
 exclude = .tox,html5lib,third_party/py,third_party/pytest,third_party/funcsigs,third_party/attrs,third_party/pluggy/,pywebsocket,six,_venv,webencodings,wptserve/docs,wptserve/tests/functional/docroot/,wpt,wptrunner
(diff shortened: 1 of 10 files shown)


Hexcles (changes requested):
Custom `has_key` methods are defined in `wptrunner/wptrunner/wptmanifest/backends/{conditional,static}.py`
which is why unit tests are failing with your changes. The methods probably need to be renamed to the magic `__contains__`.

jugglinmike:
Thanks for the tip, @Hexcles! Unfortunately, enabling this linting option might take a little more work. That's because [the "conditional" `ManifestItem` already defines a `__contains__` method](https://github.com/w3c/web-platform-tests/blob/347a7974c93326910109cd0f3386cc427add4ec5/tools/wptrunner/wptrunner/wptmanifest/backends/conditional.py#L196-L197):

```python
    def __contains__(self, key):
        return key in self._data
```

And its semantics are distinct from those of [the custom `has_key` method](https://github.com/w3c/web-platform-tests/blob/347a7974c93326910109cd0f3386cc427add4ec5/tools/wptrunner/wptrunner/wptmanifest/backends/conditional.py#L216-L220)

```python
    def has_key(self, key):
        for node in [self, self.root]:
            if key in node._data:
                return True
        return False
```

There may be a case to be made against an API where `x.has_key(x)` behaves differently than `str in x` and that perhaps `has_key` should be renamed to more clearly signal the advanced lookup logic it implements. However, that discussion and that level of change are more than I really meant to start with this patch. I'm more inclined to remove the second commit and worry about `has_key` when/if we decide to take on the much larger (and more meaningful) task of supporting Python 3.

What do you think?

Hexcles:
We should probably rename `has_key`. Clashing with a deprecated built-in method name is not good.

That said, I haven't looked too closely at the related code. If you're not sure what to do, then we can always make a smaller step each time (i.e. abandon the second and third commits for now).

gsnedders:
@jgraham objected to renaming `has_key` before, fwiw.

jugglinmike:
@gsnedders In that case, we should definitely proceed with that change in a separate patch. Thanks for letting us know!

@jgraham In can't find anything about your objection on GitHub or the #testing IRC logs. Would you mind re-iterating here?

@Hexcles I've reverted the commits related to that rule. This should be good to go now.

Hexcles (approved):
r+

gsnedders (approved):

jugglinmike:
Thanks, @gsnedders!

jugglinmike:
Oh and @Hexcles of course :)

Issue #11105, opened 2018-05-22
Impossible to write .any.js test to be run by service workers with a secure origin


anitawoodruff:
My new test: 

notifications/interfaces.https.any.js

https://pastebin.com/BG6Xu6Pd

Seems the '.https.' is stripped for service workers which leaves it unable to find the file:

 ```
0:28.90 TEST_START: /notifications/interfaces.https.any.serviceworker.html
 0:29.14 WARNING Traceback (most recent call last):                   
  File "~/web-platform-tests/tools/wptserve/wptserve/handlers.py", line 262, in __call
__                                                                    
    rv = self.func(request, response)                    
  File "~/web-platform-tests/tools/serve/serve.py", line 61, in handle_request
    self.check_exposure(request)                        
  File "~/web-platform-tests/tools/serve/serve.py", line 149, in check_exposure
    for (key, value) in self._get_metadata(request):  
  File "~/web-platform-tests/tools/serve/serve.py", line 102, in _get_metadata
    with open(path, "rb") as f:
IOError: [Errno 2] No such file or directory: '~/web-platform-tests/notifications/inte
rfaces.any.js'
```

@jgraham

jgraham:
The problem is that the serviceworker test variant enforces a `https` flag on the resource, which it then removes when looking for the on-disk js file. This is obviously wrong if the js file itself happens to have the `.https.` flag already.

anitawoodruff:
And for the record, if I remove '.https.' from the .js test file name, it runs but I get the following failure:

 0:17.77 TEST_START: /notifications/interfaces.https.any.serviceworker.html
 0:18.09 TEST_STATUS: Test OK. Subtests passed 0/1. Unexpected 1
Untitled
--------
Expected PASS, got FAIL
Only secure origins are allowed (see: https://goo.gl/Y0ZkNV).

Only if I duplicate the file both with & without the '.https.' does it run as expected.

Ms2ger:
CC @annevk

annevk:
If the input file does not contain `.https.`, where does the error "only secure origins are allowed" come from? Is that a problem with Chrome's setup or is some of the infrastructure not aware of `any.js` enabling this? Or both?

Given that you could limit the global to service workers, that does seem like a viable path, but maybe we should disallow it somehow?

Now for the more general problem, where you want `https.any.js`, I guess we could fiddle with `add_mount_point` in `tools/serve/serve.py` a bit.

annevk:
[...]

PR #11122, opened 2018-05-23
Add tests for notification inline reply idl properties

anitawoodruff:
- Migrated existing html test to .any.js test + .idl interface
to allow testing the new properties from a service worker scope.

- See https://github.com/whatwg/notifications/pull/132 for the
spec change this tests.

- See https://github.com/anitawoodruff/inline-notification-replies
for an Explainer.

### DIFF
--- a/interfaces/notifications.idl
+++ b/interfaces/notifications.idl
@@ -1,27 +1,9 @@
-<!doctype html>
-<meta charset=utf-8>
-<title>Notification interface IDL tests</title>
-<div id=log></div>
-<script src=/resources/testharness.js></script>
-<script src=/resources/testharnessreport.js></script>
-<script src=/resources/WebIDLParser.js></script>
-<script src=/resources/idlharness.js></script>
-<script type=text/plain class=untested>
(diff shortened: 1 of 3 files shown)


anitawoodruff:
Note this PR adds two duplicate any.js tests as a workaround for https://github.com/w3c/web-platform-tests/issues/11105 - one can be deleted once that bug is fixed.

foolip:
@anitawoodruff, apologies for allowing this to sit unreviewed for so long. @lukebjerring and I have been working on getting the whole interfaces/ directory in sync with specs and auto-updating, and as part of that notifications.idl was updated in #11859.

https://github.com/web-platform-tests/wpt/issues/11105 which you filed (thank you!) has now been resolved, and we have results from the .serviceworker. test here:
https://wpt.fyi/results/notifications/idlharness.https.any.serviceworker.html?label=experimental

That does reveal a problem which I've sent a fix for in https://github.com/web-platform-tests/wpt/pull/12500.

Closing this PR, sorry for it being overtaken by other changes like this.

PR #11125, opened 2018-05-23
Organize url/

### LINKED ISSUE #10918: Reorganizing url/
@mikewest @rubys @xiaojunwu @smola @domenic @Sebmaster @annevk @GPHemsley @TimothyGu would any of you mind if I move `a-element-origin.js`, `a-element.js`, `setter_tests.json`, `toascii.json`, and `urltestdata.json` into a `resources/` directory?

If it helps I'll also document the other JSON resources as part of this to make it clearer how they can be consumed in non-HTML software projects.

annevk:
Move secondary resources into a resources/ directory and document the other two JSON resources.

Fixes #10918.


url/resources/a-element-origin.js
```diff
@@ -1,14 +1,4 @@
...
-  var request = new XMLHttpRequest()
-  request.open("GET", "urltestdata.json")
-  request.send()
-  request.responseType = "json"
-  request.onload = setup.step_func(function() {
-    runURLTests(request.response)
-    setup.done()
-  })
-})
+promise_test(() => fetch("resources/urltestdata.json").then(res => res.json()).then(runURLTests), "Loading data…");
```
foolip: This kind of thing suggests that maybe we could make use of async setup, something like `setup(async () => ....)` which if it fails counts as a harness error. Would you use that?

foolip (approved):
LGTM with market research question.

url/resources/a-element-origin.js
annevk: Yeah, that seems useful. I think this pattern probably warrants an abstraction too. `setup_from_json` or some such, but I haven't really been bothered enough to make it happen.

PR #11126, opened 2018-05-23
html/rendering: Test details' pseudo-elements.

emilio:
This is a test for https://github.com/whatwg/html/pull/3686.

I have a patch to fix Gecko as well, but I got distracted with higher priority work, so will submit that when I have the time.


annevk (approved):
This seems fine. I'm not entirely sure about the directory structure here, but since testing for rendering/ is largely non-existent we can always clean that up later.

I suspect we'll need a couple of additional tests for the PR, but this is a good start.

emilio:
Ok, let's get this in for now then.

PR #11129, opened 2018-05-23
Test query encoding of attributes separately

annevk:
Helps with #4934.


annevk:
Browser bugs needed: Firefox for background, ping; Safari for ping, Chrome for base, ping.

zcorpan (approved):
LGTM % nit

annevk:
@zcorpan I actually think this test is wrong for the `ping` attribute. There's no requirements on reflecting space-separated URLs, so that would just fallback to reflecting a string, as I understand it.

annevk:
I have now updated it test `ping` as a string reflected attribute. I also added tests for `<area>`.

zcorpan:
Ah yeah, ping was changed in the spec at some point.

zcorpan (approved):

annevk:
Bugs:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1464021
* https://bugs.chromium.org/p/chromium/issues/detail?id=846256

PR #11131, opened 2018-05-23
Add a test for connectedCallback

rwlbuis:
Add a test to verify connectedCallback is not called during HTML fragment parsing

Also see https://bugs.webkit.org/show_bug.cgi?id=183586

### DIFF
--- a/custom-elements/connected-callbacks-html-fragment-parsing.html
+++ b/custom-elements/connected-callbacks-html-fragment-parsing.html
@@ -1,0 +1,49 @@
+<!DOCTYPE html>
+<html>
+<head>
+<title>Custom Elements: the HTML fragment parsing algorithm must not create a custom element synchronously</title>
+<meta name="author" title="Rob Buis" href="mailto:rbuis@igalia.com">
+<meta name="assert" content="The HTML fragment parsing algorithm must enqueue a custom element upgrade reaction instead of synchronously invoking its constructor">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/parsing.html#create-an-element-for-the-token">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="resources/custom-elements-helpers.js"></script>
+</head>
+<body>
+<div id="log"></div>
+<script>
+
+let iteration = 0;
+
+document_types().forEach(function (entry) {
+    var documentName = entry.name;
+    var getDocument = entry.create;
+    let calls = 0;
+
+    promise_test(function () {
+        class Parenter extends HTMLElement {
+            connectedCallback() {
+                const child = this.firstChild;
+                this.removeChild(child);
+                this.appendChild(child);
+            }
+        }
+        class Child extends HTMLElement {
+            connectedCallback() { calls++; }
+        }
+        iteration++;
+        let parenter = 'x-parenter' + iteration;
+        let child = 'x-child' + iteration;
+        customElements.define(parenter, Parenter);
+        customElements.define(child, Child);
+        return getDocument().then(function (doc) {
+            document.documentElement.innerHTML = `<${parenter}><${child}></${child}></${parenter}>`;
+            doc.documentElement.appendChild(document.documentElement.firstChild);
+            assert_equals(calls, 1);
+        });
+    }, `Inserting a custom element into ${documentName} using HTML fragment parsing must enqueue a custom element upgrade reaction, not synchronously invoke its constructor`);
+});
+
+</script>
+</body>
+</html>


custom-elements/connected-callbacks-html-fragment-parsing.html
```diff
@@ -0,0 +1,45 @@
...
+<link rel="help" href="https://html.spec.whatwg.org/multipage/parsing.html#create-an-element-for-the-token">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="resources/custom-elements-helpers.js"></script>
+</head>
+<body>
+<div id="log"></div>
+<script>
+
+var calls = 0;
```
rniwa: Use `let`.

custom-elements/connected-callbacks-html-fragment-parsing.html
```diff
@@ -0,0 +1,45 @@
...
+class Parenter extends HTMLElement {
```
rniwa: This class should be re-defined for each document type.
e.g. you should be declaring these classes inside promise_test itself.

custom-elements/connected-callbacks-html-fragment-parsing.html
```diff
@@ -0,0 +1,45 @@
...
+}
+customElements.define('x-parenter', Parenter);
+class Child extends HTMLElement {
+    connectedCallback() { calls++; }
+}
+customElements.define('x-child', Child);
+
+document_types().forEach(function (entry) {
+    var documentName = entry.name;
+    var getDocument = entry.create;
```
rniwa: Use `let`.

custom-elements/connected-callbacks-html-fragment-parsing.html
```diff
@@ -0,0 +1,45 @@
...
+
+    promise_test(function () {
+        return getDocument().then(function (doc) {
+            calls = 0;
+            document.body.innerHTML = '<x-parenter><x-child></x-child></x-parenter>';
```
rniwa: We need to be testing this inside doc, not document, which is the top-level document we have.

rniwa (changes requested):
r-

rwlbuis:
I updated the test to try to address the review comments. 
Note that the test passes in chromium, and passes with WebKit with my patch applied (not without it).

custom-elements/connected-callbacks-html-fragment-parsing.html
```diff
@@ -0,0 +1,48 @@
...
+                const child = this.firstChild;
+                this.removeChild(child);
+                this.appendChild(child);
+            }
+        }
+        class Child extends HTMLElement {
+            connectedCallback() { calls++; }
+        }
+        let parenter = 'x-parenter' + iteration;
+        let child = 'x-child' + iteration++;
```
rniwa: Please increment iteration before at the beginning of each test instead

custom-elements/connected-callbacks-html-fragment-parsing.html
```diff
@@ -0,0 +1,48 @@
...
+        customElements.define(parenter, Parenter);
+        customElements.define(child, Child);
+        return getDocument().then(function (doc) {
+            document.documentElement.innerHTML = '<' + parenter + '><' + child + '></' + child + '></' + parenter + '>';
```
rniwa: Use template literal as in `<${parenter}><${child}></${child}></${parenter}>`.

custom-elements/connected-callbacks-html-fragment-parsing.html
```diff
@@ -0,0 +1,48 @@
...
+</head>
+<body>
+<div id="log"></div>
+<script>
+
+var iteration = 0;
+
+document_types().forEach(function (entry) {
+    let documentName = entry.name;
+    let getDocument = entry.create;
```
rniwa: These two variables are only used once. There is no need to define them as local variables.

custom-elements/connected-callbacks-html-fragment-parsing.html
```diff
@@ -0,0 +1,48 @@
+<!DOCTYPE html>
+<html>
+<head>
+<title>Custom Elements: connectedCallback not called during HTML fragment parsing</title>
```
rniwa: This title and the assertion description is a bit misleading. `connectedCallback` is indeed called on elements parsed via fragment parsing. It's just that "synchronous custom elements flag" must be unset when creating an element during fragment parsing. Put it more pragmatically, "the HTML fragment parsing algorithm must not create a custom element synchronously" or more verbosely "the HTML fragment parsing algorithm must enqueue a custom element upgrade reaction instead of synchronously invoking its constructor"

custom-elements/connected-callbacks-html-fragment-parsing.html
```diff
@@ -0,0 +1,48 @@
...
+        let parenter = 'x-parenter' + iteration;
+        let child = 'x-child' + iteration++;
+        customElements.define(parenter, Parenter);
+        customElements.define(child, Child);
+        return getDocument().then(function (doc) {
+            document.documentElement.innerHTML = '<' + parenter + '><' + child + '></' + child + '></' + parenter + '>';
+            doc.documentElement.appendChild(document.documentElement.firstChild);
+            assert_equals(calls, 1);
+        });
+    }, 'Inserting a custom element into ' + documentName + ' using HTML fragment parsing must not call connectedCallback.');
```
rniwa: Again, this description should not be as misleading as it is right now.

rniwa (changes requested):
Looks much better. Let's go through one more iteration before merging.

rwlbuis:
Thanks for the review, I think I addressed everything now.

custom-elements/connected-callbacks-html-fragment-parsing.html
```diff
@@ -0,0 +1,49 @@
...
+<link rel="help" href="https://html.spec.whatwg.org/multipage/parsing.html#create-an-element-for-the-token">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="resources/custom-elements-helpers.js"></script>
+</head>
+<body>
+<div id="log"></div>
+<script>
+
+var iteration = 0;
```
rniwa: Use let?

custom-elements/connected-callbacks-html-fragment-parsing.html
```diff
@@ -0,0 +1,49 @@
...
+        let parenter = 'x-parenter' + iteration;
+        let child = 'x-child' + iteration;
+        customElements.define(parenter, Parenter);
+        customElements.define(child, Child);
+        return getDocument().then(function (doc) {
+            document.documentElement.innerHTML = `<${parenter}><${child}></${child}></${parenter}>`;
+            doc.documentElement.appendChild(document.documentElement.firstChild);
+            assert_equals(calls, 1);
+        });
+    }, 'Inserting a custom element into ' + documentName + ' using HTML fragment parsing must enqueue a custom element upgrade reaction, not synchronously invoke its constructor');
```
rniwa: Use template literal?

rniwa (approved):

rniwa:
Please consider making those changes before merging.

PR #11143, opened 2018-05-24
Inline testharness.css into testharness.js for better performance

zcorpan:
Loading the stylesheet at the same time as inserting the results
table causes the table to be rendered twice in different layout
modes, which takes a long time for very big tables.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/11143)


jgraham (approved):

gsnedders:
This breaks testharness.js for all browsers that don't support template literals, which seems a bit sad. (Given, e.g., you now can't run any test in Presto to see what it does, for example.)

zcorpan:
How often do you check tests in Presto? Do you avoid newer features when writing tests?

jgraham:
I didn't think about this carefully but typically we have avoided breaking testharness.js itself for older browsers. I would like us to try and find a solution for this and should have noticed in the review; sorry,

zcorpan:
Would a normal string be ok? It would be slightly annoying to edit the stylesheet.

zcorpan:
https://github.com/w3c/web-platform-tests/pull/11159

PR #11151, opened 2018-05-24
Fix accessing serviceworker tests for .https.any.js files.

jgraham:
Our support for autogenerated wrappers in any.js files assumes that
there's a 1:1 mapping between the incoming URL and the underlying js
file on the fielsystem. This is nearly true, but because we force
serviceworker tests to run via https, foo.any.js and foo.https.any.js
generate the same serviceworker test url. "Fixing" this by doubling
the https in the latter case (i.e. making the test url
foo.https.https.serviceworker.any.html) is bad because it violates the
assumption that .https. is a flag. Instead, this commit adds a hacky
solution where we first look for foo.any.js and if that doesn't
succeed we then look for foo.https.any.js, if the url suffix starts
with .https. This is fine as long as no one adds foo.any.js and
foo.https.any.js expecting them to do different things.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/11151)

### DIFF
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -88,9 +88,18 @@ class WrapperHandler(object):
                 assert len(item) == 3
                 src = item[0]
                 dest = item[2 if resource_path else 1]
+            new_path = path
             if path.endswith(src):
-                path = replace_end(path, src, dest)
-        return path
+                new_path = replace_end(path, src, dest)
+                # Hack: for serviceworker tests generated from a .any.js file we remove
+                # the .https prefix when looking for the js file. That's correct for
+                # foo.any.js but incorrcet for foo.https.any.js. So if the first doesn't
+                # exist we try the second. This relies on not having both the https and
+                # non https variant exist, otherwise we use the non-https variant for
+                # serviceworker tests.
+                if not resource_path and src.startswith(".https.") and not os.path.exists(new_path):
+                    new_path = replace_end(path, src, ".https" + dest)
+        return new_path
 
     def _get_metadata(self, request):
         """Get an iterator over script metadata based on //META comments in the


tools/serve/serve.py
```diff
@@ -88,9 +88,18 @@ def _get_path(self, path, resource_path):
...
-                path = replace_end(path, src, dest)
-        return path
+                new_path = replace_end(path, src, dest)
+                # Hack: for serviceworker tests generated from a .any.js file we remove
+                # the .https prefix when looking for the js file. That's correct for
+                # foo.any.js but incorrcet for foo.https.any.js. So if the first doesn't
+                # exist we try the second. This relies on not having both the https and
+                # non https variant exist, otherwise we use the non-https variant for
+                # serviceworker tests.
+                if not resource_path and src.startswith(".https.") and not os.path.exists(new_path):
```
annevk: This seems wrong. What if they used `test.sub.https.any.js`? Also, this doesn't seem to prevent navigating to the double `.https` URL.

jgraham: `test.sub.https.any.js` would work. `src` here is coming from the first element of `self.path_replace`, which represents the suffix from the incoming url path to remove. What's in the path before that suffix doesn't matter.

The double https still working is fundamental to the problem we've set up (of course one could special case that to fail, but it would be an extra hack). The problem here is that the whole setup assumes a unique mapping between the requested url and the on-disk `.js` file. But that mapping isn't `1:1` for `.serviceworker.` urls because `/foo.https.servcieworker.any.js` could map to either `foo.any.js` or `foo.https.any.js`. So there are two possible "correct" ways to fix this:

* Don't generate serviceworker tests unless all the tests are run over https.
* Add something extra in the url for one case so they are distinguishable e.g. `foo.https.non-https-variant.serviceworker.any.js`.

annevk: I'd much prefer specialcasing non-https in the original URL in a couple places.

foolip:
Ping this PR, on behalf of https://github.com/web-platform-tests/wpt/issues/11105.

foolip:
Overtaken by https://github.com/web-platform-tests/wpt/pull/12381. @jgraham, please reopen if this was actually doing something else as well.

PR #11154, opened 2018-05-25
[html] Add tests for `window.stop`

### LINKED ISSUE #11090: No tests for window.stop()
https://html.spec.whatwg.org/multipage/window-object.html#dom-window-stop is untested inside html/

There are two tests that use it but aren't testing it directly:
eventsource/eventsource-request-cancellation.htm
xhr/abort-after-stop.htm

jugglinmike:
@annevk @foolip These tests show highly irregular behavior between Chrome, Edge, Firefox, and Safari:

sub-test                                     | Chrome 67 | Edge 17 | Firefox 61 | Safari 11.1
---------------------------------------------|-----------|---------|------------|------------
cancels navigation that is not unloading     | FAIL      | PASS    | PASS       | FAIL
does not cancel navigation that is unloading | PASS      | FAIL    | TIMEOUT    | PASS
aborts active fetches                        | FAIL      | FAIL    | PASS       | PASS
aborts child browsing contexts               | PASS      | FAIL    | PASS       | PASS
aborts parsing                               | PASS      | FAIL    | PASS       | PASS

Because of the amount of tomfoolery (it's tricky to verify that asynchronous operations *don't* occur), I expect there's something wrong with my methodology. I've included specification excerpts that I believe validate each test, but I could really use your input on whether or not they actually hold up.

If they do, I'll be happy to follow up with bug reports for Chrome and Firefox.

This is intended to resolve gh-11090

### DIFF
--- a/html/browsers/the-window-object/window-stop.html
+++ b/html/browsers/the-window-object/window-stop.html
@@ -1,0 +1,211 @@
+<!doctype html>
+<head>
+  <meta charset="utf-8">
+  <title>window.stop()</title>
+  <script src="/resources/testharness.js"></script>
+  <script src="/resources/testharnessreport.js"></script>
+</head>
+<body>
+<script>
+/**
+ * > The stop() method on Window objects should, if there is an existing
+ * > attempt to navigate the browsing context and that attempt is not currently
+ * > running the unload a document algorithm, cancel that navigation; [...]
+ *
+ * Evaluation of `javascript:` URLs precedes document unloading [1], so the
+ * navigation should be cancelled.
+ *
+ * [1] https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigate
+ */
+async_test(function(t) {
+  var iframe = document.createElement('iframe');
+  document.body.appendChild(iframe);
+
+  window.cancelNav = {
+    firstNavigation: t.step_func(function() {
+      assert_unreached('Navigation was not cancelled.');
+    }),
+    secondNavigation: t.step_func(function() {
+      t.done();
+    })
+  };
+  iframe.contentWindow.location = 'javascript:' + [
+    'window.stop();',
+    'window.location = "javascript: parent.cancelNav.secondNavigation();";',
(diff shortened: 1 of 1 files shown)


annevk:
This looks very hard to judge as it mixes a couple of aspects that we have a poor understanding and interoperability story for (`javascript:` URLs, `window.stop()`, document lifecycle). It's great to have tests, but I strongly suspect we need many more and a more concrete handle on what the overall model ought to be.

jugglinmike:
> we need many more and a more concrete handle on what the overall model ought to be.

I'm game for both of these things.

For tests: would you like to see better coverage for `javascript:` URLs and `data:` URLs?

For the model: is it that things are under-specified? Or just not properly reflective of web reality?

annevk:
@jugglinmike I think `data:` URLs are pretty solid these days (I wrote a large test suite for parsing them recently and their security model got figured out somewhere last year and tested). `javascript:` URLs could definitely use more coverage, though I recommend looking at open issues in whatwg/html and open PRs here first.

As far as the overall navigation/document lifecycle model goes, the issue is that things are under-specified, browsers are not interoperable with each other, and it's unclear what web reality is (there might be multiple realities due to browser sniffing or not caring about certain browsers, even). And on top of that there's a lot of things that influence that model one way or another.

zcorpan:
So next step for this PR is to add a few more tests, avoiding `javascript:`, and see if the spec for `stop()` needs any changes to align with implementations. I don't think this needs to wait for navigation to be fully understood, although these tests may need updating as part of figuring out navigation.

zcorpan:
FYI there are two tests with `window.stop()` currently:
```
$ grep -r "window\.stop" .
./eventsource/eventsource-request-cancellation.htm:          window.stop()
./xhr/abort-after-stop.htm:        window.stop();
```
also 2 more open PRs: #9973 #9933

jugglinmike:
The complexity of this patch is intended to avoid the race conditions present in gh-9933. I did try to avoid complexity, but I might be able to find something better if I take another look. I'm game to try to simplify as you suggested yesterday, Simon

PR #11169, opened 2018-05-25
Remove tests for SVG transform with CSS syntax - part 1

dirkschulze:
https://drafts.csswg.org/css-transforms/#svg-transform

According to CSS WG decision, the syntax of the SVG `transform` attribute differs from the CSS transform syntax.

Many tests are obsolete and it makes no sense to keep them around any longer. I tried to keep as many tests as it made sense.


dirkschulze:
The vast majority of the removed tests check the behavior on translation and rotation units that are no longer supported for SVG transform attributes now.

dirkschulze:
@gsnedders Correct, was never supported by any UA (browser or not).

smfr (approved):

smfr (approved):

grorg (approved):

PR #11171, opened 2018-05-25
Cross-Origin-Resource-Policy tests

youennf:
These are initial tests targeting https://github.com/whatwg/fetch/pull/733
These are based on https://bugs.webkit.org/show_bug.cgi?id=185840.


annevk (commented):
In general this looks great, but I found a couple issues. (I annotated them in one file, but they apply to most tests.)

annevk:
The other thing that'd be nice to test is ensuring the header is properly parsed. E.g., `same-ORIGIN` does not match, `same-origin, same-site` does not match, etc.

annevk (approved):
Thanks for addressing the nits and writing all these. I created two more small PRs that cover some remaining aspects.

Let's wait with landing this until next week so we can land all the various pieces together.

annevk:
The reason this is timing out is because it changes `common/get-host-info.sub.js` so we'll need an admin to land this next week.

foolip:
Rebase+merge button pressed as requested by @annevk on IRC.

PR #11173, opened 2018-05-25
[testdriver] Enable manual interaction

jugglinmike:
The testdriver.js utility allows tests which require user input to be
fully automated. It requires that the browser be under the control of a
WebDriver server in order to do this.

In test maintenance contexts, this requirement can slow development
because a new browsing session must be created for each trial, and
debugging tools such as the developer's console cannot be used.

This requirement also limits the tests' value to a larger audience when
published online. Visitors to sites like w3c-test.org are interested in
observing test results experimentally, but although they may be capable
of providing the expected input, the hard requirement on WebDriver will
cause tests to fail unconditionally.

Extend the framework to accept traditional human input if the test is
being run without the WebDriver technology. This allows test authors to
debug tests and other interested parties to verify results in browsers
of their choosing. It does not interfere with fully-automated test
execution.

---

@gsnedders @kereliuk I'm interested in going even further with this. We could make testdriver.js-powered tests look a lot like today's manual tests by dynamically inserting instructions for human operators. That's less compelling for test authors (since they presumably know exactly what inputs are needed by their own tests), but it could make a big difference for visitors to http://w3c-test.org.

Of course, any change to the document has the potential to invalidate the test, so it would need to be configurable.

For now, I figured this would be a simple and safe way to make test author's lives a little easier.

I'm excited to start extending tests with testdriver.js. Thanks for making it possible!

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/11173)


gsnedders:
(Basically: that doesn't actually quite work currently, but I think it should, and that would be much easier from a UX POV, given we don't need the user to directly trigger the right sequence of events.)

jugglinmike:
Call me a nerd, but that discussion is extremely interesting.

That said, my reading is that the design is heading in a direction that would *not* support a user interface like "please click this one button, and we'll take care of the rest." It sounds like we're headed towards a consumption-based model, and that would require a 1-to-1 correspondence between user input and protected API call.

kereliuk:
What use cases are there for manual interaction with testdriver APIs other than debugging (which is an important one).
For debugging at least, I'm leaning towards a 1-to-1 correspondence between user input and protected API call rather than a UI button.
Future testdriver APIs might not have an obvious way to manually interact, but with these I guess we can continue to do what we had before and have it as a not implemented error.

jugglinmike:
> What use cases are there for manual interaction with testdriver APIs other than debugging (which is an important one).

Beyond test debugging, this patch allows non-developers to run the tests (e.g. via http://w3c-test.org)

> Future testdriver APIs might not have an obvious way to manually interact, but with these I guess we can continue to do what we had before and have it as a not implemented error.

Agreed

gsnedders:
@jugglinmike then I misunderstood and we should add something more like this :)

gsnedders:
One thing that would probably good would be draw attention to the element in question, probably by adding a big `outline` around it.

jugglinmike:
Works for me, @gsnedders. As noted in the comment, I'm recommending we add that decoration even under automation in order to avoid any strange edge cases where the presence of an outline influences the test.

This also needed rebasing to accommodate [a recent addition to the testharness.js library](https://github.com/web-platform-tests/wpt/commit/3d6920b84bac82c45a8da5c8b04b6da0c64f02fd#diff-d3c29fceb5efc460656c6e691c91cdf5). The original version of the branch is available here:

https://github.com/web-platform-tests/wpt/compare/master...bocoup:testdriver-manual-prebase

jugglinmike:
Rebased again to resolve conflicts from gh-11512

gsnedders:
On the face of it, this looks good to me.

However, it does mean when we introduce a new testdriver function that the manual codepath will run where previously the Promise rejected, and hence it will run to timeout. And timeouts are going to be a problem with relying on manual interaction with the tests, too.

jugglinmike:
> However, it does mean when we introduce a new testdriver function that the
> manual codepath will run where previously the Promise rejected, and hence it
> will run to timeout.

To support the w3c-test.org/local filesystem use cases, I think that the "fully
manual" implementation has to be the default. We could attempt to detect when
the test is in automation by inspecting the state of the global
`test_driver_internal` object, and designing all of the "manual"
implementations to throw when any part of that has been modified, e.g.

```diff
diff --git a/resources/testdriver.js b/resources/testdriver.js
index 5651bf2..4c3943a 100644
--- a/resources/testdriver.js
+++ b/resources/testdriver.js
@@ -54,6 +54,12 @@
         element.style.outline = 'dashed #0366d6 5px';
     }
 
+    function inAutomation() {
+      return Object.keys(manual).some(function(name) {
+          return window.test_driver_internal[name] !== manual[name];
+      });
+    }
+
     /**
      * @namespace
      */
@@ -154,7 +160,7 @@
         }
     };
 
-    window.test_driver_internal = {
+    var manual = {
         /**
          * Waits for a user-initiated click
          *
@@ -163,7 +169,13 @@
          * @returns {Promise} fulfilled after click occurs
          */
         click: function(element, coords) {
-            return new Promise(function(resolve) {
+            return new Promise(function(resolve, reject) {
+                if (!inAutomation()) {
+                    reject(new Error('Not implemented'));
+
+                    return;
+                }
+
                 element.addEventListener("click", resolve);
             });
         },
@@ -178,6 +190,12 @@
          */
         send_keys: function(element, keys) {
             return new Promise(function(resolve, reject) {
+                if (!inAutomation()) {
+                    reject(new Error('Not implemented'));
+
+                    return;
+                }
+
                 var seen = "";
 
                 function onKeyDown(event) {
@@ -208,4 +226,7 @@
             return Promise.reject(new Error("unimplemented"));
         }
     };
+
+    window.test_driver_internal = Object.create(manual);
+
 })();
```

This is a little implicit, though. We could also require that implementers
define a property on `test_driver_internal` to make this more transparent.
That's a different kind of undesirable because it's more to document and prone
to accidental omission.

Do either of those strategies seem like a good idea to you?

> And timeouts are going to be a problem with relying on manual interaction
> with the tests, too.

That's true, but since that is a case where we know a human operator is
manually executing the test, I think the risk of a timeout is preferable to the
certainty of a failure.

gsnedders:
I think we should ask others who maintain runners for opinions here; cc @jgraham @Hexcles @youennf @thejohnjansen

Hexcles:
@gsnedders how would this affect non-wpt runners?

jugglinmike:
Good idea, @gsnedders. Here's my attempt to summarize for those folks:

Today, the default testdriver.js implementation throws a "Not Implemented" error for all available methods. This makes tests which depend on those methods fail quickly and logically until an automated version is provided.

With this patch, we want to allow people to manually provide input when the tests are run outside of automation. To do this, we plan to make the default testdriver.js implementation prompt for human interaction.

Without taking extra precautions, this will change those "Not Implemented" errors to timeouts. We plan to avoid this by designing the "prompt for user input" methods to throw if they are executed in an automated environment.

We could use your feedback on how we determine that the tests are being run in automation. Here are the strategies we're considering:

- comparing the methods defined on the `test_driver_internal` object with those on the "prompt for user input" implementation
- requiring that implementers set a flag on the `test_Driver_internal` object to explicitly indicate this state

The former is the lowest-friction approach, but we're concerned that it may be too implicit. The latter will require a change on your end, but it will make this behavior more obvious.

jugglinmike:
Thanks, @Hexcles! I'm satisfied moving forward with your feedback, but I want to give @jgraham @youennf, and @thejohnjansen one more chance to weigh in. [I summarized our question about a week ago](https://github.com/web-platform-tests/wpt/pull/11173#issuecomment-400029951); do you folks have opinions one way or the other?

lukebjerring:
My 5c:

This is a much more reasonable behaviour than the current failure.
https://github.com/web-platform-tests/wpt/pull/12010 is my use-case, where running via WPT serve and loading it up just rejects the promise_test, even though I can manually click the button and cause all the subtests to behave as expected.

jugglinmike:
@lukebjerring would you like to see specific references to `wpt serve`? I omitted that because this also assists people that don't use the WPT CLI (e.g.  http://w3c-test.org or folks that implement their own wrappers around the server). We might be able to recognize this distinction in the docs, but that might be too much information. I'm inclined to think that the hand-wavy "outside of automation contexts" is an appropriate generalization.

lukebjerring:
That makes sense to me.

gsnedders:
I'm still in favour of the above patch, but we're still waiting for others to respond two months later. :(

jgraham:
So I don't have have strong objections here, but I do have some concerns

* This seems to create a codepath that never runs in production (because no one really runs ,anual tests). A test that works when run manually doesn't necessarily work in any production case, particularly because the fallback seems to involve modifying the DOM of the test itself. Therefore the use case here seems to be "getting a test into a state where it seems to work when run manually, to avoid the overhead of using `wpt run` during initial development, but then testing with `wpt run` anyway". Which is a real but fairly narrow use case.

* I think the timeout issue is real. A test requiring complex interactions can't necessarily be completed by a human in 10s. This would have to disable the testharness timeout when it's being used.

I don't think it's a bad idea to have this kind of fallback, but I do think it's inherently problematic and fairly hard to get right.

foolip:
I've sent https://github.com/web-platform-tests/wpt/pull/14073 to fix just `test_driver.bless`, since I kept applying roughly that patch locally.

foolip:
I never commented on this PR at all, so here goes.

My main concern here is the use of `highlight` to change the style of elements to click. The extra side effects (style invalidation at least) could interfere with tests and most problematic would be when they cause tests to pass when they shouldn't. Most likely nobody would notice that, and an opt-out wouldn't mitigate the risk.

This isn't a super strong concern, but as we add to testdriver.js I think it will be increasingly difficult to allow a user to perform the same actions, certainly for https://github.com/w3c/sensors/pull/369 it won't be possible at all. I suspect many `action_sequence` uses are also very tricky.

So, I think that apart from APIs in the style of `bless`, instructions are best put in the tests where possible. They would get out of sync and often not be included at all, and I think that's acceptable.

jugglinmike:
This patch introduces the additional visual styles in both automated and non-automated scenarios. Whether contributors validate new tests via manual interaction or via testdriver.js, the state of the DOM will be consistent. There is some risk that the outline will make it impossible to express some expectation in either scenario, but that seems very unlikely to me.

> This isn't a super strong concern, but as we add to testdriver.js I think it will be increasingly difficult to allow a user to perform the same actions, certainly for w3c/sensors#369 it won't be possible at all. I suspect many action_sequence uses are also very tricky.

Taking this approach for these methods does not obligate us to do the same for any future methods. I maintain that implementing partial support here will result in many more tests being supported (and supported more consistently) than if the burden is left to test authors.

foolip:
This PR does update the general documentation of testdriver.js to say that human operators are supported, but it sounds like that's not the hope any longer.

For click and send_keys it's possible to wait for a human to take action, but I really would rather that we don't have any outline applied. Most tests will only require clicking a single thing, and it won't be a mystery which it is.

jugglinmike:
> This PR does update the general documentation of testdriver.js to say that human operators are supported, but it sounds like that's not the hope any longer.

What do you think of, "Outside of automation contexts, it allows human operators to provide expected input manually (for operations which may be described in simple terms)."

> For click and send_keys it's possible to wait for a human to take action, but I really would rather that we don't have any outline applied. Most tests will only require clicking a single thing, and it won't be a mystery which it is.

I added that at @gsnedders' [request](https://github.com/web-platform-tests/wpt/pull/11173#issuecomment-396762404), so I'd like to get their take before removing it. What do you say. Geoffrey?

gsnedders:
@foolip `outline` shouldn't cause any style/paint invalidation?

foolip:
It changes style surely it causes style to be recalculated? @lilles might be able to say more authoritatively if changing `element.style.outline` could plausibly ever interfere with the thing being tested.

lilles:
Changing the outline will typically cause style and paint invalidation. I guess that it could potentially mask browser-bugs for a given test-case.

gsnedders:
Does it actually cause a full recompute of style, given it's a non-inherited property? (And yes, obviously it triggers paint; I meant to say layout above!)

gsnedders:
I'm not that opposed, but there's plenty of other cases where we already have side-effects!

jugglinmike:
I am not convinced that the new side effect is actually a concern. However, it's really only a nicety, and @gsnedders isn't insisting on it, so I'm willing to debate it as a future enhancement. I've removed it from this changeset.

[The only other contentious issue is how we detect automated scenarios.](https://github.com/web-platform-tests/wpt/pull/11173#issuecomment-400029951) To reiterate: we wanted to require that folks implementing testdriver internal methods signal automation by setting an explicit flag. This allows us to fail quickly when automated methods are not available, and it's agnostic of the underlying technology (it will work for Selenium, Webriver, DevTools protocol, etc.). Failing to set this flag may result in timeouts, but it is not expected to cause tests to pass spuriously.

In light of @Hexcles's support and @jgraham's non-objection (their comments reflect different concerns), I have implemented that change.

docs/_writing-tests/testdriver.md
```diff
@@ -5,7 +5,9 @@ order: 8.5
 ---
 
 testdriver.js provides a means to automate tests that cannot be
-written purely using web platform APIs.
+written purely using web platform APIs. Outside of automation
+contexts, it allows human operators to provide expected input
+manually (for operations which may be described in simple terms).
```
foolip: This now LGTM.

resources/testdriver.js
```diff
@@ -137,27 +137,64 @@
         }
     };
 
-    window.test_driver_internal = {
+    var manual = {
```
foolip: I don't quite understand the naming of this, or if the use of `Object.create` with this object as the prototype is a necessary part of the change. Shadowing `in_automation` by setting it to true on the instance (leaving it false on the prototype) works, but why?

resources/testdriver.js
```diff
@@ -137,27 +137,64 @@
...
-         * @param {String} keys - keys to send to the element
-         * @returns {Promise} fulfilled after keys are sent or rejected if click fails
+         * @param {Element} element - element which should receve key presses
+         * @param {String} keys - keys to expect
+         * @returns {Promise} fulfilled after keys are received or rejected if
+         *                    an incorrect key sequence is received
          */
         send_keys: function(element, keys) {
-            return Promise.reject(new Error("unimplemented"));
+            if (!window.test_driver_internal.in_automation) {
```
foolip:
```suggestion
            if (this.in_automation) {
```

foolip: Hmm, isn't there some test that failed because of the `!` typo?

resources/testdriver.js
```diff
@@ -137,27 +137,64 @@
...
+
+                    if (keys.indexOf(seen) !== 0) {
+                        reject(new Error("Unexpected key sequence: " + seen));
+                    } else if (seen === keys) {
+                        resolve();
+                    }
+                }
+
+                element.addEventListener("keydown", onKeyDown);
+            });
```
foolip: Can you add a `.finally` that removes the event listener? Otherwise that `seen` variable will keep building up for every keydown, which is harmless, but takes effort for the reader to understand that it is.

resources/testdriver.js
```diff
@@ -137,27 +137,64 @@
...
+         * Waits for a user-initiated click
          *
          * @param {Element} element - element to be clicked
          * @param {{x: number, y: number} coords - viewport coordinates to click at
-         * @returns {Promise} fulfilled after click occurs or rejected if click fails
+         * @returns {Promise} fulfilled after click occurs
          */
         click: function(element, coords) {
-            return Promise.reject(new Error("unimplemented"));
+            if (window.test_driver_internal.in_automation) {
```
foolip:
```suggestion
            if (this.in_automation) {
```

foolip: This relates to the change to use a prototype. Do we treat `test_driver_internal` as a namespace like `console`, or an instance with a prototype like `document`? If you revert the `Object.create` change, please also reject these suggestions.

resources/testdriver.js
```diff
@@ -137,27 +137,64 @@
...
-    window.test_driver_internal = {
+    var manual = {
         /**
-         * Triggers a user-initiated click
+         * This flag should be set to `true` by any code which implements the
+         * internal methods defined below for automation purposes. Doing so
+         * allows the library to signal failure immediately when an automated
+         * implementation of one of the methods is not available.
+         */
+        in_automation: false,
```
foolip: Defaulting to manual means we'll have to react quickly after this is imported into Chromium to see what broke. @LukeZielinski can you try running this in Chromium CI and prepare any adaptation we need?

foolip:
@LukeZielinski would be good for you to review these changes too I think.

resources/testdriver.js
LukeZielinski: I wonder if we can set the default to true initially so that when this gets imported, everything works the same way. That can give us time to override it to true in the Chromium testdriver. A follow-up PR can then flip the default to false. Does that sound reasonable?

jugglinmike: I don't think there is any risk to Chrome in landing this with `in_automation` set to `false` by default.

The `in_automation` flag controls how the manual methods behave; its purpose is to account for future extensions. The value will be relevant when a new method is added to `testdriver.js` that the consumer has *not* implemented. In those cases, the "manual" implementation will detect that it is running in automation and return a rejected Promise as "not implemented".

Failing to set this flag will only result in test timeouts instead of test failures (and again, only when new methods are available in `testdriver.js`).

I don't know where/how Chrome currently implements `test_driver_internal`, but I believe it only lacks the `action_sequence` method. We have no plans to implement a manual version of that method (since communicating the expected input would be very difficult), so that will always report "not implemented", regardless of the value of `in_automation`.

Hexcles: Chrome implementation of `test_driver_internal` can be found here: https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/resources/testdriver-vendor.js?l=78&rcl=57d76f99672823e61bb01a8d132c5c663caf899b

LukeZielinski: Yep, you're right, so nevermind - Chrome implements all of the apis so we don't fall back to the timeout case.

foolip: @LukeZielinski prepared for this in https://chromium-review.googlesource.com/c/chromium/src/+/1352571 anyway, thanks Luke!

LukeZielinski (approved):
LGTM

jugglinmike:
I submitted this patch before we integrated Taskcluster into WPT. This means that Taskcluster ignores pushes to the branch, and that's a problem now that the Taskcluster check is required.

I would normally rebase the branch, but since the patch has already been reviewed and approved, I've merged `master` instead. Please take care to squash when merging this, @LukeZielinski / @gsnedders!

jugglinmike:
It looks like this is blocked on existing flakiness. @gsnedders how would you feel about admin-merging this?

resources/testdriver.js
gsnedders: @jugglinmike ?

gsnedders (dismissed):
@jugglinmike see @foolip's outstanding (non-resolved) comments

resources/testdriver.js
jugglinmike: Originally, we defined the object in this way so could automatically detect
when `test_driver_internal` had been extended (i.e. by comparing the enumerable
properties of `manual` and `test_driver_internal`).

We since decided on a more intentional override mechanism (via the
`in_automation` property), and you are right to observe that the use of the
prototype chain is no longer technically necessary. I've left it in place for
the sake of clarity: it makes the code reflect the intended usage. `manual`
describes an object that exposes the human-driven implementation, and
`test_driver_internal` describes a distinct object which defaults to that
behavior. When/if `test_driver_internal` is overridden, it does not affect the
`manual` implementation conceptually, and the state of the objects reflects
that.

Would naming this object `manuel_implementation` make this any more clear? Or
would you prefer that we simplify the internals instead?

resources/testdriver.js
gsnedders: `finally` is as old as exceptions in ES are (i.e., ES3), and I don't think anything ever supported exceptions but not a `finally` clause?

jugglinmike: The `finally` that @foolip and I are discussing is `Promise.prototype.finally` as standardized in ES2018

resources/testdriver.js
foolip: I'd prefer to have no separation of the methods that have a default implementation that waits for user input and those that always throw, and overriding all in the same manner. If using the prototype chain and overriding by setting a property shadowing it on the instance is nice, it's equally nice in both cases I think. Just a flag `test_driver_internal` as it originally was would be fine I think.

If one other reviewer does find the new structure clearer then I'll consider myself downvoted, however.

foolip:

resources/testdriver.js
gsnedders: @jugglinmike this is the one outstanding thread here; I don't think I have a strong preference either way here, FWIW.

foolip: I'll merge the PR and send a follow-up.

foolip:
I'll close and reopen to trigger new CI runs.

foolip:
Taskcluster failures.


## Chrome

|                                                   Test                                                   |                                           Subtest                                           |            Results            |                                         Messages                                        |
|----------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------|-------------------------------|-----------------------------------------------------------------------------------------|
| `/event-timing/event-timing-timingconditions.html`                                                       |                                                                                             | **OK: 5/10, TIMEOUT: 5/10**   |                                                                                         |
| `/event-timing/event-timing-timingconditions.html`                                                       | `Event Timing only times certain types of trusted event.`                                   | **FAIL: 5/10, TIMEOUT: 5/10** | `assert_equals: Should only observe one entry: []. expected 1 but got 0;Test timed out` |
| `/feature-policy/picture-in-picture-allowed-by-feature-policy-attribute-redirect-on-load.https.sub.html` |                                                                                             | **CRASH: 1/10, OK: 9/10**     |                                                                                         |
| `/feature-policy/picture-in-picture-allowed-by-feature-policy-attribute-redirect-on-load.https.sub.html` | `Feature-Policy allow="picture-in-picture" allows same-origin navigation in an iframe.`     | **PASS: 9/10, MISSING: 1/10** |                                                                                         |
| `/feature-policy/picture-in-picture-allowed-by-feature-policy-attribute-redirect-on-load.https.sub.html` | `Feature-Policy allow="picture-in-picture" disallows cross-origin navigation in an iframe.` | **PASS: 9/10, MISSING: 1/10** |                                                                                         |
| `/feature-policy/picture-in-picture-allowed-by-feature-policy.https.sub.html`                            |                                                                                             | **CRASH: 1/10, OK: 9/10**     |                                                                                         |
| `/feature-policy/picture-in-picture-allowed-by-feature-policy.https.sub.html`                            | `Feature-Policy header: picture-in-picture * allows the top-level document.`                | **PASS: 9/10, MISSING: 1/10** |                                                                                         |
| `/feature-policy/picture-in-picture-allowed-by-feature-policy.https.sub.html`                            | `Feature-Policy header: picture-in-picture * allows same-origin iframes.`                   | **PASS: 9/10, MISSING: 1/10** |                                                                                         |
| `/feature-policy/picture-in-picture-allowed-by-feature-policy.https.sub.html`                            | `Feature-Policy header: picture-in-picture * allows cross-origin iframes.`                  | **PASS: 9/10, MISSING: 1/10** |                                                                                         |
| `/speech-api/SpeechSynthesis-pause-resume.tentative.html`                                                |                                                                                             | **OK: 1/10, TIMEOUT: 9/10**   |                                                                                         |
| `/speech-api/SpeechSynthesis-pause-resume.tentative.html`                                                | `speechSynthesis.pause() and resume() state and events`                                     | **FAIL: 1/10, TIMEOUT: 9/10** | `assert_unreached: error event Reached unreachable code;Test timed out`                 |

# Firefox

|                                            Test                                           |                                             Subtest                                             |             Results              |                                           Messages                                           |
|-------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------|----------------------------------|----------------------------------------------------------------------------------------------|
| `/bluetooth/requestDevice/canonicalizeFilter/max-length-exceeded-name-unicode.https.html` |                                                                                                 | **ERROR: 1/10, OK: 9/10**        |                                                                                              |
| `/bluetooth/requestDevice/canonicalizeFilter/max-length-exceeded-name-unicode.https.html` | `Unicode string with utf8 representation longer than 248 bytes in 'name' must throw TypeError.` | **FAIL: 9/10, MISSING: 1/10**    | `promise_test: Unhandled rejection with value: object "ReferenceError: Mojo is not defined"` |
| `/feature-policy/reporting/camera-report-only.https.html`                                 |                                                                                                 | **ERROR: 1/10, TIMEOUT: 9/10**   |                                                                                              |
| `/feature-policy/reporting/camera-report-only.https.html`                                 | `Camera report only mode`                                                                       | **TIMEOUT: 9/10, MISSING: 1/10** | `Test timed out`                                                                             |
| `/pointerevents/pointerevent_setpointercapture_inactive_button_mouse.html`                |                                                                                                 | **OK: 4/10, TIMEOUT: 6/10**      |                                                                                              |
| `/pointerevents/pointerevent_setpointercapture_inactive_button_mouse.html`                | `setPointerCapture + inactive button state`                                                     | **FAIL: 4/10, MISSING: 6/10**    | `InvalidPointerId: Invalid pointer id.`                                                      |

foolip:
Changing testdriver.js affects tests that were perhaps already flaky, so that isn't very actionable.

Filed https://github.com/web-platform-tests/wpt.fyi/issues/1147 about a problem with the wpt.fyi checks.

There is some regression risk here, but I'll merge and then post a comparison between full runs before/after the change.

PR #11189, opened 2018-05-28
Enable the port message queue in /workers/opaque-origin.html

### LINKED ISSUE #11188: /workers/opaque-origin.html is disabled/flaky in chromium webkit

-- | -- | -- | --
/workers/opaque-origin.html | chromium webkit | [ Failure Crash Timeout ] [ Skip ] | https://crbug.com/626703 None

cc @mkruisselbrink

I think the issue is that the test isn't enabling the port message queue, it needs `channel.port1.start();`

zcorpan:
Fixes #11188.

### DIFF
--- a/workers/opaque-origin.html
+++ b/workers/opaque-origin.html
@@ -11,6 +11,7 @@ onmessage = e => {
   channel.port2.postMessage(e.data);
 };
 fetch_tests_from_worker(channel.port1);
+channel.port1.start();
 </script>
 <iframe sandbox="allow-scripts" src="support/sandboxed-tests.html?pipe=sub"></iframe>
 </body>


zcorpan:
`create_remote_worker` calls `start()` for service workers and shared workers, and otherwise assumes a dedicated worker. This test falls into the dedicated worker code path but needs `start()`. Do you think this is OK as-is or should we change `create_remote_worker` to call `start()` if it exists?

mkruisselbrink:
Ah, good point. So then it's only the fairly esoteric case of passing a raw MessagePort. We could add a separate if (is_message_port(worker)) case just for that, but given how uncommon this likely is I suppose we might as well do changes like this.

mkruisselbrink (approved):

PR #11192, opened 2018-05-28
Auto-update the payment-request IDL

lukebjerring:


interfaces/payment-request.idl
```diff
@@ -27,6 +27,8 @@ dictionary PaymentMethodData {
 dictionary PaymentCurrencyAmount {
   required DOMString currency;
   required DOMString value;
+  // Note: currencySystem is "at risk" of being removed!
```
foolip: Looks like data source isn't fresh. This is gone from the spec and I removed it manually in https://github.com/web-platform-tests/wpt/pull/11099.

foolip (commented):
Other changes looks like they're probably in the right direction, but I didn't check to see if there have been other manual changes also reverted. Seems safest to refresh the data source.

foolip (approved):
Looks like only whitespace/boring changes.

PR #11193, opened 2018-05-28
HTML: remove autofocused element before its task runs

annevk:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1463563 for context.

We'll need to update -2 to figure out the correct pass condition. Given Chrome and Safari, I guess we should expect input2 and then we should update the specification to note that if the element is no longer connected you just abort the algorithm, meaning no flags get set.

### DIFF
--- a/html/semantics/forms/autofocus/once-per-document-2.window.js
+++ b/html/semantics/forms/autofocus/once-per-document-2.window.js
@@ -1,0 +1,13 @@
+async_test(t => {
+  const input = document.createElement("input");
+  input.autofocus = true;
+  document.body.appendChild(input);
+  input.remove();
+  const input2 = document.createElement("input");
+  input2.autofocus = true;
+  document.body.appendChild(input2);
+  t.step_timeout(() => {
+    assert_equals(document.activeElement, document.body);
+    t.done();
+  }, 100);
+}, "autofocus should only work for the first inserted element matching the conditions???");
--- a/html/semantics/forms/autofocus/once-per-document.window.js
+++ b/html/semantics/forms/autofocus/once-per-document.window.js
@@ -1,0 +1,20 @@
+async_test(t => {
+  const input = document.createElement("input");
+  input.autofocus = true;
+  document.body.appendChild(input);
+  t.step_timeout(() => {
+    assert_equals(document.activeElement, input);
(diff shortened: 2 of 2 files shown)


domenic (approved):
-2 looks good to me per spec: it will

- queue a task to focus `input`
  - run the focusing steps which do nothing
  - set the "something has been autofocused" flag (implicit in step 8 right now)
- queue a task to focus `input2`
  - see the "something has been autofocused" flag is set, and do nothing

The first test also looks good, and is an interesting test. I'd suggest commenting the `document.body.appendChild(input)` to explain that this removes the element, which causes the focus fixup rule to run removing focus from it.

annevk:
Thanks for the review. -2 is good per spec, but per the Firefox issue it causes compat issues, so I guess we should change the spec.

gsnedders:
Should this have status:needs-spec-decision and a spec bug filed? (Trying to clear out approved and un-merged PRs!)

domenic:
By code inspection, Safari and Blink do very different things here. Safari has a per-element flag, not a per-document flag. Ick.

domenic:
I cannot understand Blink's behavior. @tkent-google, can you help? Given

```js
const input = document.createElement("input");
input.id = "i1";
input.autofocus = true;
document.body.appendChild(input);
input.remove();

const input2 = document.createElement("input");
input2.id = "i2";
input2.autofocus = true;
document.body.appendChild(input2);
```
My reading of the code at [SetAutofocusElement](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/dom/document.cc?l=7334&rcl=97b99cde75e3f066149e4bf8f7173c1e2ff7b356) and its call sites is that the second insertion should do nothing. But, somehow, Blink focuses `i2`.

Safari focuses `i2` because it appears to autofocus every element, with no global per-Document flag?? Ugh.

tkent-google:
Both of WebKit and Blink trigger autofocus processing after style resolution, not on inserting into a document.  So, autofocus of ```i1``` isn't processed.

I'll file a specification issue.

tkent-google:
> I'll file a specification issue.

Filed a specification issue: https://github.com/whatwg/html/issues/4645

PR #11196, opened 2018-05-28
Simplify the CSS Logical tests and prevent the result log from being rotated

Loirooriol:
Cc @mrego

### DIFF
--- a/css/css-logical/logicalprops-block-size-vlr.html
+++ b/css/css-logical/logicalprops-block-size-vlr.html
@@ -6,10 +6,10 @@
 <link rel="help" href="https://drafts.csswg.org/css-writing-modes-3/#logical-to-physical">
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
-<script src="resources/style-check.js"></script>
+<script src="/resources/check-layout-th.js"></script>
 
 <style>
-div {
+#tests div {
   border: 1px solid #000;
   writing-mode: vertical-lr;
 }
@@ -28,23 +28,8 @@ div {
   min-block-size: 50px;
   max-block-size: 100px;
 }
-#ref_div1 {
-  width: 40px;
-  min-width: 50px;
-  max-width: 100px;
-}
-#ref_div2 {
-  width: 100px;
-  min-width: 50px;
-  max-width: 100px;
-}
-#ref_div3 {
-  width: 120px;
-  min-width: 50px;
-  max-width: 100px;
-}
 
-p {
+#tests p {
   border: 1px solid #000;
   writing-mode: vertical-lr;
 }
@@ -55,12 +40,6 @@ p {
 #p2 {
   width: 50px;
   block-size: 100px;
-}
-#ref_p1 {
-  width: 50px;
-}
-#ref_p2 {
-  width: 100px;
 }
 
 .table {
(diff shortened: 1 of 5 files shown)


css/css-logical/logicalprops-block-size-vlr.html
```diff
@@ -93,75 +72,34 @@
...
+// Check that width override block-size and vice versa in vertical-lr
+checkLayout("#p1, #p2");
 
-test(function () {
-  assert_true(compareWidthHeight("table1_cell", "ref_table1_cell"));
-  assert_true(compareWidthHeight("table2_cell", "ref_table2_cell"));
-  assert_true(compareWidthHeight("table3_cell", "ref_table3_cell"));
-}, "Check that block-size < min-block-size or min-block-size < block-size <= max-block-size or block-size > max-block-size in table vertical-lr");
+// Check that block-size < min-block-size or min-block-size < block-size <= max-block-size or block-size > max-block-size in table vertical-lr
+checkLayout(".tablecell");
```
mrego: If you want to call `checkLayout()` more than once, you need to pass `false` as 2nd argument.
And then call manually `done()` at the end of your test.

Otherwise you can change numbers and you'll see that things don't fail as expected.

css/css-logical/logicalprops-block-size-vlr.html
```diff
@@ -93,75 +72,34 @@
...
-}
 </style>
 
-<div id="div1"></div>
-<div id="div2"></div>
-<div id="div3"></div>
-<div id="ref_div1"></div>
-<div id="ref_div2"></div>
-<div id="ref_div3"></div>
+<div id="tests">
```
mrego: I'd add `<div id="log"></div>` at the beginning of the test, so the results would appear there.
Then you could put the writing-mode directly on `#tests` and avoid repeating it on the rest of elements.

mrego (commented):
Thanks for doing these changes.
I have 2 comments but it's looking good.

Loirooriol:
I added the descriptions in `<p>`, which were used for some tests, so instead I added classes and renamed some things. Didn't set `writing-mode` on the parent to avoid affecting the descriptions.

css/css-logical/logicalprops-block-size-vlr.html
```diff
@@ -93,75 +72,52 @@
...
-<div id="ref_div2"></div>
-<div id="ref_div3"></div>
+<div id="log"></div>
 
-<p id="p1"></div>
-<p id="p2"></div>
-<p id="ref_p1"></div>
-<p id="ref_p2"></div>
+<h3>Maximum and minimim block sizes in blocks with vertical-lr</h3>
+<div class="tests">
```
mrego: Where is this "tests" class defined?

css/css-logical/logicalprops-block-size-vlr.html
```diff
@@ -93,75 +72,52 @@
...
-test(function () {
-  assert_true(compareWidthHeight("div1", "ref_div1"));
-  assert_true(compareWidthHeight("div2", "ref_div2"));
-  assert_true(compareWidthHeight("div3", "ref_div3"));
-}, "Check that block-size < min-block-size or min-block-size < block-size <= max-block-size or block-size > max-block-size in vertical-lr");
+<h3>Maximum and minimim block sizes in table cells with vertical-lr</h3>
+<div class="tests">
+  <p><code>block-size</code> &lt; <code>min-block-size</code></p>
+  <div class="table">
+    <div class="tablecell" id="table1_cell" data-expected-client-width="40" data-expected-client-height="100"></div>
```
mrego: Nit: "table-cell" sets again the vertical writing-mode, I don't think that's needed as it's already set in "table".

css/css-logical/logicalprops-block-size-vlr.html
```diff
@@ -93,75 +72,52 @@
...
+  <div class="table">
+    <div class="tablecell" id="table3_cell" data-expected-client-width="120" data-expected-client-height="100"></div>
+  </div>
+</div>
+
+<script>
+checkLayout(".block", false);
+checkLayout(".override", false);
+checkLayout(".tablecell", false);
+done();
```
mrego: Nit: I believe you can change this for `checkLayout(".block,.override,.tablecell")`.
Or instead of having a `<script>` tag do it in `<body onload="checkLayout(".block,.override,.tablecell")">`.

mrego (commented):
Thanks, a few more comments.

css/css-logical/logicalprops-block-size-vlr.html
Loirooriol: `.tests` is used in logicalprops-block-size.html in order to define the width of the containing block. Other files don't assign styles to it but I still used it for consistency and to group similar tests. But I guess I can remove the class from the element.

css/css-logical/logicalprops-block-size-vlr.html
Loirooriol: I tried that, but then the log just shows `.block,.override,.tablecell` everywhere. If you separate them, then you see there are 3 `.block`, 2 `.override` and 3 `.tablecell`; that seemed better to me.

mrego (approved):
Thanks for doing the changes. LGTM!

PR #11197, opened 2018-05-28
Call done() correctly in /workers/nested_worker.worker.js

### LINKED ISSUE #11195: /workers/nested_worker.worker.html is slow/disabled in chromium webkit

-- | -- | -- | --
/workers/nested_worker.worker.html | chromium webkit | [ Timeout ] [ Skip ] | https://crbug.com/709227 None

This test should call `done()` after `async_test()` and should call `t.done()` where it's currently calling `done()`.

zcorpan:
Fixes #11195.

### DIFF
--- a/workers/nested_worker.worker.js
+++ b/workers/nested_worker.worker.js
@@ -1,12 +1,11 @@
 importScripts("/resources/testharness.js");
 
-async_test(function(t) {
+async_test(function() {
     var worker1 = new Worker("support/WorkerBasic.js");
     worker1.postMessage("ping");
     worker1.onmessage = this.step_func_done(function(evt) {
         assert_equals(evt.data, "Pass");
         worker1.terminate();
-        t.done();
     });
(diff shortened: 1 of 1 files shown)


jugglinmike (changes requested):
> I don't understand why this makes a difference. The onmessage handler uses step_func_done which should have the same effect as explicitly calling t.done() in this case.

I agree that the `t.done` invocation is superfluous. @zcorpan would you mind removing it for clarity's sake?

The relocation of `done` is important because it de-couples the test status from the harness status. In other words: if the assertion fails, the test will fail instead of timing out.

zcorpan:
Ah yes, I missed that it was using `step_func_done`.

jugglinmike (approved):
I did, too. It was a good catch on @jdm's part.

Awaiting TravisCI report...

jugglinmike:
Thanks, Simon!

Issue #5293, opened 2017-03-31
Deprecate some requirement flags


gsnedders:
Originally posted as https://github.com/gsnedders/wpt-docs/issues/6 by @frivoal on 09 Jan 2017, 01:19 UTC:

> The [list of requirement flags](https://gsnedders.github.io/wpt-docs/writing-tests/css-metadata#requirement-flags) looks like it lists all of the historical ones.
> 
> We should keep on documenting the old ones, so that when people run into them they know what they mean, but we should move these to a "historical" subsection, and only keep the ones we actually expect people to in the main section.
> 
> Based on previous discussions, I believe we had agreed that the following ones could be deprecated:
> * ahem: treat ahem as a requirement for the entire test suite, and stop flagging individual tests
> * animated: no need to distinguish from manual tests flagged as "interact"
> * dom: can be inferred from the test using javascript
> * image / svg: modern UAs all support this, and it can be inferred form the test itself anyway
> * userstyle: no need to distinguish from manual tests, unless we also adopt a convention for linking to the user style sheet that automated tests runners can use to pick up the test file, in which case we will also not need the flag. For instance we could require the use of something like `Prior to running this test, <a rel="user-style-sheet" href="path/to/style-sheet.css">this user style sheet </a> must be <a href="doc/to/how-to-use/user/style-sheets/">installed</a>`.
> * 32bit: historical, inconsistently used anyway
> * 96dpi: required by the specs anyway.

gsnedders:
Originally posted as https://github.com/gsnedders/wpt-docs/issues/6#issuecomment-271199189 by @frivoal on 09 Jan 2017, 01:57 UTC:

> Actually, we can also deprecate this one:
> * interact: redundant with the `*-manual` extension if test file names, presuming we start applying it systematically for CSS tests as well.

gsnedders:
Originally posted as https://github.com/gsnedders/wpt-docs/issues/6#issuecomment-271295292 by @gsnedders on 09 Jan 2017, 14:19 UTC:

> We got no resolution and no real agreement as to what could be dropped, AIUI, hence I simply haven't touched the table yet.

gsnedders:
Originally posted as https://github.com/gsnedders/wpt-docs/issues/6#issuecomment-271302301 by @frivoal on 09 Jan 2017, 14:47 UTC:

> Let's try and propose one again. Does the list above seem good do you, or do you think I missed something? I know we had another list at some point, but I'm too lazy to go look for it.

[...]

PR #11200, opened 2018-05-28
Add tests to verify that animated translation and rotation freeze as …

fred-wang:
…expected when animation-play-state is set to 'paused'.

This covers iOS bugs reported by users in https://bugs.webkit.org/show_bug.cgi?id=170784#c0

### DIFF
--- a/css/css-animations/set-animation-play-state-to-paused-001.html
+++ b/css/css-animations/set-animation-play-state-to-paused-001.html
@@ -1,0 +1,89 @@
+<!DOCTYPE html>
+<html class="reftest-wait">
+  <head>
+    <meta charset="utf-8">
+    <title>Dynamically set animation-play-state to paused</title>
+    <link rel="author" title="Igalia S.L." href="https://www.igalia.com/">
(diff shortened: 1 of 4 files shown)


css/css-animations/set-animation-play-state-to-paused-001.html
```diff
@@ -0,0 +1,89 @@
+<!DOCTYPE html>
+<html class="reftest-wait">
+  <head>
+    <meta charset="utf-8">
+    <title>Dynamically set animation-play-state to paused</title>
+    <link rel="author" title="Igalia S.L." href="https://www.igalia.com/">
+    <link rel="help" href="https://drafts.csswg.org/css-animations-1/#animation-play-state">
+    <meta name="flags" content="animated">
```
fred-wang: @birtles I had copied this "animated" meta flag from other tests for CSS animations. However, I'm not exactly sure about the purpose of this flag... It seems the manual tests changed in #11215 were exactly those having the meta flag. Do you know more about this?

fred-wang: OK I removed this flag, per #5293

fred-wang:
I'll split this into two PRs.

Issue #9230, opened 2018-01-29
Rename OWNERS files


Ms2ger:
CC @foolip @tobie 

I've been getting comments from people who interpret them as "only people in this file can review tests". Unless there's a good reason to keep using the name, let's change it.

Ms2ger:
https://github.com/tobie/ganesh/search?utf8=%E2%9C%93&q=OWNERS&type=

foolip:
Renaming would be fine with me. REVIEWERS would have some of the same liabilities.

WATCHERS? Trouble there is you wouldn't really expect anyone who's just watching to do anything. But maybe that's a problem of reliable triage of outstanding reviews, it's certainly a problem now too.

gsnedders:
> WATCHERS? Trouble there is you wouldn't really expect anyone who's just watching to do anything. But maybe that's a problem of reliable triage of outstanding reviews, it's certainly a problem now too.

But per our docs we don't expect people in OWNERS to do anything today, so that doesn't seem like a change?

foolip:
I mean that the people sending PRs wouldn't expect anything, but maybe it's good that they expect nothing when that's often what they get. Although, if they don't know where the list of reviewers came from, they'll probably still be pleasantly surprised for a day or two until they notice no reactions :)

So, yes, I'm fine with renaming to something else. In Chromium we have [WATCHLISTS](https://dev.chromium.org/developers/contributing-code/watchlists) files, but that's a more complicated format.

Name ideas?

jgraham:
`NOTIFY`?

Ms2ger:
`NOTIFY` wfm.

kereliuk:
NOTIFY +1, I'll send a PR unless anyone has any objections.

foolip:
This will require at least the status:needs-owners label to be renamed, and probably other changes in tooling maintained by @tobie.

tobie:
Yeah. Let's touch-base before you push the PR or you'll get all owners pinged about it.

Hexcles:
Please also ping me when a PR is sent to rename these files. Blink's import/export might need some small adjustment. Thanks.

zcorpan:
When https://github.com/w3c/web-platform-tests/issues/10552 is in place, I think @tobie's bot's comment will no longer be relevant, so it can be removed.

As for the import/export, I think renaming these will be an improvement since Chromium has its own OWNERS files; NOTIFY files could be synced as normal and the export script would skip OWNERS as it does now.

[...]

Issue #4228, opened 2016-11-19
wpt-pr-bot text is confusing


RByers:
wpt-pr-bot adds text to PRs like: "Reviewers for this pull request are: @Rbyers".  People [tend to assume](https://github.com/w3c/web-platform-tests/pull/2163#issuecomment-260952922) this means the MUST wait for someone from that list to approve the PR, increasing the frustration at the difficulty of getting WPT PRs landed.  I've [tried to clarify this in the README](https://github.com/w3c/web-platform-tests/pull/4227) but people likely won't notice.  Can we update the text generated by wpt-pr-bot to make this more clear?

RByers:
I searched around for the code for wpt-pr-bot but can't seem to find it anywhere.

RByers:
Found [the announcement](https://lists.w3.org/Archives/Public/public-test-infra/2016JanMar/0001.html).  @tobie WDYT?  Arguably even the file name "OWNERS" is misleading (could be "WATCHLIST" or something).

tobie:
File is here: https://github.com/tobie/ganesh/blob/master/lib/reviewers.js

Yeah. Text sure needs fixing. Suggestion/PR welcome. If not I'll look at it once I get back home.

tobie:
Quick PR. LMK what you think or if you have another suggestion.

tobie:
Note I'm happy to change the OWNERS file name to something else if you prefer. That'll need a PR here first, though, and a bit of sync.

RByers:
That helps, thank you!  What about just adding a link to https://github.com/w3c/web-platform-tests/blob/master/README.md#test-review from the bot text - so it's obvious where to look for details / how to add yourself to the list of people to be notified?

IMHO it's probably not worth the hassle at this point to change the file name - maybe wait and see if there continues to be confusion?

tobie:
https://github.com/w3c/web-platform-tests/pull/4235#issuecomment-262081975

tobie:
Closing. LMK if you end-up changing the file names so I can update the bot accordingly.

PR #2163, opened 2015-09-14
Add requestIdleCallback tests.

rmcilroy:
Add requestIdleCallback tests based on the spec at https://w3c.github.io/requestidlecallback/

@plehegar, @igrigorik, could you take a look please?

PR #11201, opened 2018-05-28
Rename of OWNER files to META.yml

kereliuk:
This is just a rename PR. I found all the files doing a `find . -name "OWNERS"` so hopefully that didn't miss anything.

This is tracked by #9230


annevk:
Won't this affect @tobie's bot?

zcorpan (approved):

zcorpan:
Yes, let's wait for @tobie before merging.

jgraham (changes requested):
I think that REVIEWERS is a bad choice of name per the original rationale for renaming these files at all.

kereliuk:
I've switched it to NOTIFY and changed the PR description

zcorpan (approved):

jgraham (approved):

annevk:
Another reason I don't think NOTIFY makes sense is that the person putting themselves in that file does not expect to become a reviewer. Whereas everyone submitting a PR will get the impression they are a reviewer. REVIEWERS conveys the expectations much more clearly.

zcorpan:
I don't like doing nothing because the issue for Chromium having their own OWNERS files (and so skipping them in export) makes it annoying for Chromium to create new directories if the linter rejects directories without OWNERS.

annevk:
> On the other hand there is clear harm from someone not reviewing a change they are otherwise qualified to review because they believe that they aren't allowed to due to not being in the OWNERS or REVIEWERS file.

Has this hypothetical happened in practice? It's not all obvious when browsing PRs what that file might say so I'm rather skeptical.

tobie:
LMK once you've figured out what you want to call things?

jgraham:
https://github.com/web-platform-tests/wpt/pull/2163#issuecomment-260952922 is an example of this happening last time this was discussed (in https://github.com/web-platform-tests/wpt/issues/4228)

zcorpan:
How about having both reviewers (with write access), and watchlist (either in separate files or in one file with some syntax)?

jugglinmike:
Can the "watch" use case be satisfied using GitHub's built-in "Watch" feature and an e-mail filter?

annevk:
@zcorpan how would you mention those in the WATCHLIST? Adding comments isn't great.

@jgraham so one reported instance at the end of 2016. And it wasn't because of the file being named OWNERS, it was because of the people being asked to review, which this isn't solving. So I'm not sure how that example is relevant.

zcorpan:
A comment or editing OP.

annevk:
We moved away from comments as they're too noisy. Editing OP is somewhat reasonable if that works and if there's a delay so OP can itself do quick copy-edits and not get into conflicts with the bot.

annevk:
@jugglinmike no, at least not for folks who [consume GitHub like I do](https://annevankesteren.nl/2017/11/using-github).

zcorpan:
OK so to make a slightly more concrete proposal.

* Rename to `META.yaml`
* Contains two lists as follows
```yaml
Reviewers:
  - zcorpan
  - annevk
Watchers:
  - jgraham
```
(Including @ makes it invalid YAML)

In the migration we can put everyone as Reviewers since that is the status quo. I can envision having a bot move people from Reviewers to Watchers if they don't review anything after some time.

gsnedders:
I'm increasingly in favour of something like @zcorpan's proposal above. It allows for a distinction between the two and provides a sensible place to add further things in the future.

That said, I'd rather we kept the keys entirely lowercase, as seems to be more common.

@annevk I've certainly had to point out to others (on IRC, especially) that they don't need to wait for one of the specific auto-added reviewers to review and that they can ask others on multiple occasions. That said, given it's the reviewers on GH that's the issue here I'm not sure we can reasonably fix this without spammy comments.

annevk:
@gsnedders I'm not contesting that at all. I'm saying that the file name is not relevant to that debate and that the file name REVIEWERS does have positive attributes for folks adding themselves to it (namely that they realize they're expected to review).

I'm okay with the META approach too though.

foolip:
OK, so to move forward, I'd like to propose https://github.com/web-platform-tests/wpt/pull/11201#issuecomment-393140088 but with just the `reviewers` field to begin with, since we haven't figured out how to notify people in a non-spammy way. All people in OWNERS would be added in the reviewers list.

To ensure that people are actually able to review, at minimum PRs modifying META.yaml should be blocked on all of the reviewers being org members, and in the normal case also be in the reviewers team. The low-tech solution is to make it such that META.yaml requires review from @web-platform-tests/admins, and if they get tired of checking or make mistakes, we can automate. I'm currently trying to poke as many people as I can to join the org who are already in OWNERS.

foolip:
> AIUI @foolip is against the bot having the necessary permissions to add people to the org.

Not strongly. I think it would be *nice* if no bot has admin access. More important is that inviting people after they've been added to OWNERS doesn't have a high enough success rate. So something that gets people the right access before adding them to META.yaml would be great.

jgraham:
> More important is that inviting people after they've been added to OWNERS doesn't have a high enough success rate. So something that gets people the right access before adding them to META.yaml would be great.

So that implies that a lint to check that `OWNERS` are in the org is a good thing, independent of any other changes? That sounds reasonable, but I don't think it affects the other issues/decisions around renaming the file?

tobie:
> More important is that inviting people after they've been added to OWNERS doesn't have a high enough success rate.

I don’t really agree with this. I had at most a couple errors per month in the logs, which means pretty much every one had accepted the invite.

gsnedders:
@tobie FWIW, https://github.com/orgs/web-platform-tests/people is currently at 329 people and 150 pending invitations.

foolip:
Aha. It may in fact have been the repo transition that ruined it and made it look worse than what it was. Since everyone was invited to the new org, I removed repo-level collaborators with read access. So people who haven't accepted the invite are no longer being added to reviews. That's my bad, should have treated read access like write access, where I've verified write access through the team before removing from the repo.

jgraham:
Presumably that's just reflecting the difference between "you added yourself to an `OWNERS` file and as a result got an invitation" and "three years ago you added yourself to an `OWNERS` file and something happened and now you got another invitation"

jgraham:
Which is another way of saying that the underlying issue here is that many tests[uites] — including ones with "owners" who are still active on web standards — are not actually well owned. As usual my only suggestion there is "do stuff through vendor repos where reviews are more likely to actually happen". But that doesn't solve the problem when that process isn't suitable for whatever reason.

foolip:
FYI, everyone listed in any OWNERS file as of commit 1ee3583b5e9af0afa960990fed3ed341a194cbe8 os now either an org member or has a pending invite.

foolip:
(With one exception, a non-existing user name updated in https://github.com/web-platform-tests/wpt/pull/11254.)

gsnedders:
@jgraham In a sense, part of the problem is that the more we push "do stuff through the vendor repos" the less likely people are to look at this repo. And that has two consequences: issues raised here (because they affect all vendors) are less likely to get noticed or actioned upon, and pull requests here are less likely to get landed; I think it's worthwhile remembering we have a number of fairly regular contributors who don't really write browser code (mostly specs people).

jgraham:
@gsnedders I don't think we forget that. For example most of the people who are using GitHub as a TODO list are primarilly working on specs, and we have made a lot of changes to accomodate that workflow, and will make more in the future.

The "problem" is that we have a lot of people who feel contractually obliged to do reviews on their employer's review system, but no such pressure when it comes to web-platform-tests. And GitHub itself is relatively poor at this (e.g. there's no equivalent of `needinfo?` that is exposed in the default view and could be used to alert people that N issues require their specific attention).

I would love to solve this problem, but I honestly don't know how.

jgraham:
So one thing we could try is adding vendor metadata to the `OWNERS` files (but probably `META` in this case). For Gecko it would probably be the relevant bug components, for blink it would presumably be something derived from the in-tree `OWNERS` file. Then if a PR got ignored for too long we would open a vendor bug for getting the PR landed, enabling us to use the normal triage and review mechanisms there.

I think this would be at least somewhat unpopular, but it's pretty much the only idea I have which seems like it could make a difference.

zcorpan:
Interesting idea, I think we should try that.

jgraham:
Thinking about it some more, for gecko we actually already have a bug since we create one upfront for the sync. But a bot could adjust that existing bug so it appears on the triage radar and could maybe even figure out how to r? an actual human (I'm not sure where the data for this would come from, but we could look at e.g. other people who had done reviews in the component or something). So the trick would be working out the relevant gecko bug for a PR. The best thing would be to have that link in the PR, but without the ability to create comments, it's not clear where it would go.

zcorpan:
I think a single bot comment per PR is acceptable.

jgraham:
But we don't have any reaonable way to achieve that. At the moment the PR dashboard is broken, but eventually it will presumably come back in some form (posibly linked to wpt.fyi), and that will then take our per-PR bot comment quota. Having multiple systems update the same comment seems like a straightforward way to create race conditions.

AFAICT the only way to actually solve the problem here is to build up a second UI that is able to accept arbitary PR-associated content, and redirect all bot traffic there (this means that we have to control all bots and can't use any third-party integrations). That system could then own the one permitted comment.

Actually the PR dashboard intended to be that system, but we over-specialised it to the Travis use case. If it's going to be replaced by some specialization of wpt.fyi we should have a way to attach arbitary content to PRs.

Random design idea: use a git repo as the backend with one directory per pr and one file per metadata type. Then each system interacting with web-platform-tests would be taught to modify the associated file in the git repo, and one could write some simple tools to auto-publish the entire thing statically and avoid the need for a hosted systen with credentials (which is why the PR dashboard failed in the end).

foolip:
> At the moment the PR dashboard is broken, but eventually it will presumably come back in some form (posibly linked to wpt.fyi)

Maybe we'll do comments, but not necessarily. The ideal state would be that those checks are PR status checks like Travis, and that following the links tells you what you need to know. cc @lukebjerring.

jgraham:
> Maybe we'll do comments, but not necessarily.

I guess that makes sense if wpt.fyi is going to do additional checks beyond what's in the CI job. It isn't obvious to me that that should be the case (e.g. if the CI moved to taskcluster we could write the cross-browser comparisons as a task depending on the individual browser jobs, which seems preferable since it moves more logic in-tree).

But yes, one way to add links to externa systems is via real-or-fake status checks on the PR, since those store a URL. I'm not sure that links stored that way aren't confusing unless they actually represent a job that must pass for the PR to be mergable. In particular I would expect to get complaints about exposing a gecko bug in this way.

foolip:
> I guess that makes sense if wpt.fyi is going to do additional checks beyond what's in the CI job. It isn't obvious to me that that should be the case (e.g. if the CI moved to taskcluster we could write the cross-browser comparisons as a task depending on the individual browser jobs, which seems preferable since it moves more logic in-tree).

I agree that for checks based only on Taskcluster results, a dependent task that makes use of the results make the most sense. To check for regressions also involves a previous run, where the results are in wpt.fyi, and I think we'll want wpt.fyi to be able to detect regressions between two runs that *aren't* from PRs to begin with, so maybe that'll make sense to outsource to wpt.fyi. Let's see.

zcorpan:
Since comment with mentions seems more problematic than review requests, I now think we should have a single list `suggested reviewers`. We can add some mechanism to escalate if review doesn't happen, but we can discuss that in a new issue. Having something automated to prune suggested reviewers could be good but also doesn't need to block this.

zcorpan:
I've changed this to do the META.yml with `suggested_reviewers:`.

jgraham:
Is there a corresponding change in wpt-pr-bot? (also have we considered importing that directly into the repo?)

kereliuk:
@jgraham I'll push a PR to update the bot in a bit and link here.

jgraham:
Do we have the ability to redeploy the bot pending the resolution of #11275?

kereliuk:
Here is a [PR](https://github.com/web-platform-tests/wpt-pr-bot/pull/24) to update the bot.

jgraham:
Do we have the ability to redeploy the bot pending the resolution of #11275?

zcorpan:
I will look into redeployment.

zcorpan:
"Your app was successfully deployed."

zcorpan:
https://github.com/web-platform-tests/wpt-pr-bot/pull/24 was incomplete because it was still looking for `OWNERS` files. Fix for that is https://github.com/web-platform-tests/wpt-pr-bot/pull/26

jdm:
The CSS lint complains about these files when we run it downstream in Servo. Why did it not trigger on this PR?
```
Running the WPT lint on ./tests/wpt/web-platform-tests/...
{"path": "css/CSS1/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/CSS2/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/CSS2/i18n/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/WOFF2/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/compositing/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-align/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-animations/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-backgrounds/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-break/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-cascade/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-color/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-conditional/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-contain/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-content/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-counter-styles/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-display/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-display/run-in/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-exclusions/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-fill-stroke/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-filter/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-flexbox/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-font-loading/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-fonts/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-fonts/matching/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-fonts/variations/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-gcpm/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-grid/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-images/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-layout-api/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-lists/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-logical/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-masking/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-multicol/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-namespaces/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-overflow/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-page/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-paint-api/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-position/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-properties-values-api/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-pseudo/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-regions/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-rhythm/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-round-display/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-ruby/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-scoping/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-scroll-anchoring/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-scroll-snap/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-scrollbars/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-shadow-parts/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-shapes/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-sizing/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-speech/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-style-attr/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-syntax/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-tables/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-text-decor/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-text/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-text/i18n/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-timing/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-transforms/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-transitions/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-typed-om/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-ui/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-values/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-variables/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/css-writing-modes/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/cssom-view/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/cssom/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/filter-effects/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/geometry/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/mediaqueries/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/motion/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
{"path": "css/vendor-imports/mozilla/META.yml", "message": "Support file not in support directory", "lineno": null, "rule": "SUPPORT-WRONG-DIR"}
WPT Lint Tool:: lint error(s) in Web Platform Tests: exit status 73
```

jdm:
Ah, turns out it did: https://travis-ci.org/web-platform-tests/wpt/builds/387794570?utm_source=github_status&utm_medium=notification

zcorpan:
Apologies about that. Fix at https://github.com/web-platform-tests/wpt/pull/11486

plehegar:
This pull request has been flagged by the GitHub support team for having over 100 labels. After July 26th, this pull request will be frozen until the labels are cleared. They recommend removing unneeded ones to get the count under 100.

lovelyMircho:
Lovely Mircho come back

salah8584 (approved):

Issue #10990, opened 2018-05-14
Make testdriver blow up sooner with a nice message when trying to do two calls at the same time.


kereliuk:
As far as I know this isn't really enforced in a descriptive way at the moment.

jugglinmike:
I'm only now seeing this issue, though, and the wording makes me think that there is some fundamental reason why we can't support that. Care to fill me in, @kereliuk?

jugglinmike:
Scratch that--I found https://github.com/web-platform-tests/wpt/pull/10882 and the awesome tutorial you wrote! Nothing there indicates that this restriction is fundamental to the framework, though. If that's true, then my patch should just remove the caveat from the tutorial.

kereliuk:
@jugglinmike You're right, #11202 should fix this. Will review it now!

jugglinmike:
@kereliuk I think gh-11202 is an example of functional fixedness. I saw a technical problem and tried to solve it. I wasn't considering the more fundamental question: why would tests try to executing commands in parallel in the first place?

As far as I know, it's only because they are using the parallelized `async_test`. There's nothing about the testharness.js use case that necessitates it (if there were, then the fundamentally serial WebDriver protocol would be a dead-end).

By allowing parallel invocations, we would simplify the change required to use this library (it would be "safe" to use from `async_test`). That's valuable because larger patches are more susceptible to regression and generally take longer to review.

At the same time, we'd also be allowing tests to be written with less precision. testdriver.js could accommodate inconsistency, but it couldn't solve it. If instead, we intentionally limited testdriver.js as you described above, we'd be promoting better test-writing practices: more explicit execution order reduces potential for flakiness. I think that's worth a little extra churn.

So I'm planning on submitting a patch as you originally outlined in this issue's description. I wanted to get your input before doing so because you've invested time in reviewing gh-11202.

foolip:
It looks like https://github.com/web-platform-tests/wpt/pull/11202 ended by pointing back to this issue. @jugglinmike, can you summarize why queuing up the request isn't the right move here?

jugglinmike:
@foolip Our thinking was that rejecting attempts at parallel execution would encourage authors to write serial tests, and that this would reduce potential for flakiness.

To be honest, though, I am having second (actually third) thoughts. It's not that I am warming up to parallel testing; it's that moving away from parallel testing will take a much more concerted effort. Trying to "encourage" serial execution in an important-but-rarely-used utility may be more frustrating than helpful.

[...]

PR #11202, opened 2018-05-28
 [testdriver] Prevent protocol interference

jugglinmike:
Prevent interference between parallel invocations of the testdriver.js
API by scheduling messages in a queue.

Interference could be avoided by extending the communication protocol
with a request/response identifier, but such an extension would be of
little practical benefit because the WebDriver processing model does not
support concurrency.

### DIFF
--- a/tools/wptrunner/wptrunner/testdriver-extra.js
+++ b/tools/wptrunner/wptrunner/testdriver-extra.js
@@ -6,7 +6,7 @@
     window.addEventListener("message", function(event) {
         const data = event.data;
 
-        if (typeof data !== "object" && data !== null) {
+        if (typeof data !== "object" || data === null) {
             return;
         }
 
--- a/tools/wptrunner/wptrunner/testdriver-extra.js
+++ b/tools/wptrunner/wptrunner/testdriver-extra.js
@@ -1,25 +1,52 @@
 "use strict";
 
 (function(){
-    let pending_resolve = null;
-    let pending_reject = null;
-    window.addEventListener("message", function(event) {
-        const data = event.data;
+    let previous = Promise.resolve();
+    /**
+     * Schedule an asynchronous operation (described by a Promise-returning
+     * function) to be executed following completion of every asynchronous
+     * operation that has previously been scheduled in this way.
+     *
+     * @param {Function} async_op - Promise-returning function
+     * @returns {Promise} resolved with the result of invoking the provided
+     *                    function
+     */
+    function enqueue(async_op) {
+        const next = previous.then(async_op);
+        previous = next.catch(function() {});
+        return next;
+    }
+    /**
+     * Send a message to the opening browsing context.
+     *
+     * @param {any} value to be serialize and transmitted
+     * @returns {Promise} settled when the opening browsing context has replied
+     */
+    function send(requestData) {
+        window.opener.postMessage(requestData, "*");
 
-        if (typeof data !== "object" || data === null) {
-            return;
-        }
+        return new Promise(function(resolve, reject) {
+            window.addEventListener("message", function handle(event) {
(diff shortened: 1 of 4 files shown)


tools/wptrunner/wptrunner/testdriver-extra.js
```diff
@@ -1,25 +1,52 @@
...
+     * function) to be executed following completion of every asynchronous
+     * operation that has previously been scheduled in this way.
+     *
+     * @param {Function} async_op - Promise-returning function
+     * @returns {Promise} resolved with the result of invoking the provided
+     *                    function
+     */
+    function enqueue(async_op) {
+        const next = previous.then(async_op);
+        previous = next.catch(function() {});
```
kereliuk: Should we add a failure message here saying where we were in the chain? Might be useful to debugging, but also might add needless complexity.

jugglinmike:
I just pushed up a fixup commit to update the tutorial accordingly (that was a pleasant surprise).

kereliuk (changes requested):
I really like this solution (chaining promises)! Code here looks great.

Can you add a test to infrastructure/ that verifies parallel invocations are not interfered with?

tools/wptrunner/wptrunner/testdriver-extra.js
jugglinmike: We're returning `next` to the caller, so they still have the ability to react to errors.

That said, this pattern *does* hide unhandled rejections. That's because even though tests still have the option to react to failures in the `next` promise, the invocation of `catch` means that there will always be at least one handler.

Here's a more involved solution that avoids this:

```js
function enqueue(async_op) {
    let next;
    previous = new Promise(function(resolve) {
        next = previous.then(async_op)
          .then(resolve, function(err) {
              resolve();
              throw err;
          });
    });
    return next;
}
```

This feels a little too clever to me, but for a small and well-defined utility function like `enqueue`, it might be the right thing to do. What do you think?

jugglinmike:
> I really like this solution (chaining promises)! Code here looks great.

:) Thanks!

> Can you add a test to infrastructure/ that verifies parallel invocations are not interfered with?

I didn't realize we had tests, though I would have if I had read the docs. The good news keeps coming

gsnedders:
What's the reason to put this in the wptrunner WebDriver implementation rather than in /resources/testdriver.js itself? Given the API is defined to return promises, any implementation has this problem, no?

jugglinmike:
The problem isn't fundamental to the testdriver.js contract; it's specific to how the wptrunner implementation interacts with the postMessage API. I could imagine another implementation that uses `fetch`, where requests and responses would be associated without the need for explicit scheduling. The "human driver" extension I'm proposing in gh-11173 is a more concrete example of an implementation that doesn't need this.

I suppose we could re-locate the code as a general safety mechanism, but that seems particularly defensive. Implementations could be doing all sorts of invalid things, after all. Keeping the fix localized to the code that needs it makes the adapter easier to understand, and that seems valuable.

gsnedders:
@jugglinmike Indeed it's not inherent to the testdriver.js, but the issue is that if the majority of runners do one thing (and remember not all will use WebDriver), then I expect we'll see test writers relying upon the serial order that some impls don't provide. Therefore I think we should make testharness.js guarantee it.

gsnedders:
Having discussed this with @jugglinmike on IRC, I'm now more in favour of #10990 than this.

jugglinmike:
That's right: I talked @gsnedders and myself down from my own patch :(

gsnedders:
Given we talked ourselves away from this, I think we're going with #10990. Or did @jugglinmike have second thoughts in https://github.com/web-platform-tests/wpt/issues/10990#issuecomment-410909335?

PR #11205, opened 2018-05-28
Add staticrange/OWNERS

zcorpan:


zcorpan:
cc @foolip (may need to add @choniong as collaborator or so)

foolip:
@zcorpan you should have the power. Try inviting to the reviewers team and see if that works.

zcorpan:
Ok added

choniong:
Sorry for the delay, I just got the email from Simon.

I recalled that the only test was removed in:
https://github.com/w3c/web-platform-tests/pull/9967

Do we know why it was added back, or why we removed it in the first place?

Thanks!

foolip:
Oh... so we shouldn't have added that IDL file and test :/

lukebjerring:
Well, that's one assessment. More accurately: we're not sure why it was removed in the first place, and unsure that we should have added it back in. But, ultimately, the `StaticRange` in `dom.idl` seems to just be a placeholder and isn't actually testing anything.

zcorpan:
OK, seems like we can have the test, then, right?

@choniong are you OK with being in the OWNERS file?

foolip:
dom.idl has this:
```webidl
[Exposed=Window]
interface AbstractRange {
  readonly attribute Node startContainer;
  readonly attribute unsigned long startOffset;
  readonly attribute Node endContainer;
  readonly attribute unsigned long endOffset;
  readonly attribute boolean collapsed;
};

[Exposed=Window]
interface StaticRange : AbstractRange {
};
```

Filed https://github.com/w3c/staticrange/issues/16.

I think we should revert the changes. If the above doesn't match what implementers intend to align on, then that should be fixed in DOM.

zcorpan:
OK, no need for this given #11247

PR #11211, opened 2018-05-29
[mediacapture] Simply test and use standard API

jugglinmike:
The MediaStream's `active` event was removed from the specification in
December of 2015 [1]. The `addtrack` event offers similar semantics, and
as a standard API, it is expected to be implemented by all browsers
which follow the specification. In addition, because it does not depend
on media playback, its use allows `MediaElement#play` to be removed.
That removal simplifies the process of executing the test because
browsers are beginning to tighten their policies regarding automated
media playback [2].

[1] https://github.com/w3c/mediacapture-main/pull/291
[2] https://developers.google.com/web/updates/2017/09/autoplay-policy-changes

---

@yellowdoge @martinthomson @uysalere More context about autoplay is available [here](https://github.com/web-platform-tests/results-collection/pull/552). Other MediaCapture tests continue to rely on the `active` event and on automatic playback, but those definitely need a more advanced solution. Because a simpler solution seemed possible for this test, I thought I'd propose that first. Is there any reason to continue to us `onactive`?

### DIFF
--- a/mediacapture-fromelement/creation.html
+++ b/mediacapture-fromelement/creation.html
@@ -14,15 +14,20 @@ var makeAsyncTest = function(filename, numTracks) {
     var video = document.createElement('video');
     video.src = "/media/" + filename;
     video.onerror = this.unreached_func("<video> error");
-    video.play();
 
     assert_true('captureStream' in video);
 
     var stream = video.captureStream();
     assert_not_equals(stream, null, "error generating stream");
 
-    // onactive event is marked for deprecation (https://crbug.com/649328)
-    stream.onactive = this.step_func_done(function() {
+    stream.onaddtrack = this.step_func_done(function() {
+      var tracks = stream.getTracks();
+      var idx;
+
+      for (idx = 0; idx < tracks.length; idx += 1) {
+        assert_equals(tracks[idx].readyState, 'live')
+      }
+
       // The stream got a (number of) MediaStreamTracks added.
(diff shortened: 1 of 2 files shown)


zcorpan:
Can you add a `historical.html` test that tests that `active` is *not* supported?

jugglinmike:
@zcorpan Sure thing. Think we can land this?

zcorpan (approved):

zcorpan:
@jugglinmike can you file bugs for Chromium and Edge, for mediacapture-fromelement/historical.html ?

jugglinmike:
Good idea, @zcorpan. Here's the Chromium bug: https://bugs.chromium.org/p/chromium/issues/detail?id=849138

I'm unable to [file a bug against Edge](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/new/) because doing so requires an account on their system, and their system rejects my `@bocoup.com` e-mail address:

> You can't sign up here with a work or school email address. Use a personal email, such as Gmail or Yahoo!, or get a new Outlook email.

Did you register with your personal e-mail address?

zcorpan:
I used my gmail address.

zcorpan:
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/17758018/

jugglinmike:
Thanks! Hope that doesn't mess with your quota :)

yellowdoge:
See https://bugs.chromium.org/p/chromium/issues/detail?id=649328

PR #11223, opened 2018-05-29
Make multiple-workers tests fail if (Shared)Worker is not defined

### LINKED ISSUE #11222: /workers/semantics/multiple-workers/* are flaky/disabled in chromium webkit

-- | -- | -- | --
/workers/semantics/multiple-workers/003.html | chromium webkit | [ Failure Timeout ] [ Skip ] | https://crbug.com/655458 None

These can detect if `Worker` or `SharedWorker` are not defined, and fail the test.

zcorpan:
Fixes #11222.

### DIFF
--- a/workers/semantics/multiple-workers/006.html
+++ b/workers/semantics/multiple-workers/006.html
@@ -2,6 +2,9 @@
 if (location.hash == '#1') {
   onconnect = function(e) {
     var port = e.ports[0];
+    onerror = e => {
+      port.postMessage(String(e))
+    }
     var w2 = new Worker('#2');
     w2.onmessage = function(e) {
       port.postMessage('1'+e.data);
@@ -28,6 +31,7 @@ async_test(function() {
     assert_equals(e.data, '123');
     this.done();
   });
+  w1.onerror = this.unreached_func("error");
 });
 </script>
 <!--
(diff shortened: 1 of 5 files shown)


jugglinmike:
Verified that this transforms ugly timeouts into beautiful failures in Chromium. Good going!

I'm wondering what you think of this approach:

    w1.onerror = this.step_func(function(e) { assert_unreached(e.message); });

It's slightly less invasive and also capable of capturing errors stemming from `postMessage`. Then again, it introduces a dependency on another functionality of Workers. If the implementation is failing on part of that spec, maybe it's not such a good idea to rely on another part of the same spec.

I'm happy to merge this as is, just curious to hear your thoughts.

zcorpan:
That could work I think, didn't think of that. Probably more elegant. I can try that.

workers/semantics/multiple-workers/006.html
```diff
@@ -2,6 +2,9 @@
 if (location.hash == '#1') {
   onconnect = function(e) {
     var port = e.ports[0];
+    onerror = e => {
+      port.postMessage(String(e))
+    }
```
jugglinmike: If `#1` has an error, won't that trigger `w1.onerror`?

zcorpan: Not when it's a SharedWorker, in chromium, I found out. Also the spec has a difference between dedicated and shared workers: https://html.spec.whatwg.org/multipage/workers.html#runtime-script-errors-2

jugglinmike: Got it

jugglinmike (approved):

PR #11227, opened 2018-05-29
Unflake resource_initiator_types by forcing non existing entries to pass

yoavweiss:
This is an attempt to fix the flakiness encountered at https://github.com/web-platform-tests/wpt/pull/10840 by forcing non-existing entries to pass, rather than be ignored.

/cc @foolip


foolip:
Looks like the Travis jobs still think this test is flaky, not enough deflaking?

yoavweiss:
I now think this attempt was misguided, and the test was flaky in Firefox even before the https://github.com/web-platform-tests/wpt/pull/10840 changes. Closing

Issue #11212, opened 2018-05-29
Failing in all browsers test in dom/nodes/Element-matches.html


domenic:
`In-document Element.matches: Universal selector, matching all descendants of the specified reference element (with refNode Element): *` in https://wpt.fyi/results/dom/nodes/Element-matches.html seems suspicious.

domenic:
Same problem with the sibling file, Element-webkitMatchesSelector.html

foolip:
Source is here:
https://github.com/web-platform-tests/wpt/blob/91d7e29edf7e68ec7b0c1f17d1ea0ebbb1a29b7b/dom/nodes/selectors.js#L442

This was last touched by @Ms2ger in https://github.com/web-platform-tests/wpt/commit/6ceec439c15e6843d90ea9b4e9c3d0e8de907736, ages ago.

foolip:
Actually, it's https://github.com/web-platform-tests/wpt/blob/ccd35804dbdc6e37763a4b980e048c5f5fdf7c39/dom/nodes/selectors.js#L451

Don't know how I ended up at that older commit through the GitHub UI...

Ms2ger:
Possibly related: https://github.com/web-platform-tests/wpt/pull/1577

This is more of a Selectors test; cc @tabatkins @fantasai

foolip:
A problem I noticed trying to debug this is that selectors.js is written as if `element.matches` can still take more than one argument, with this bit of comment:
```
 * expect.matches(selector, context)    // Only where refNodes is not specified
 * expect.matches(selector, refNodes)
 * expect.matches(selector)             // Only if neither context nor refNodes is specified
```

And it's written to test `find()` and `findAll()`, which never came to be.

PR #11229, opened 2018-05-29
Move and simplify element.matches("*") test so that it works

foolip:
The reason it didn't work before was because in Element-matches.js,
it's always `someElement.matches("*")` being tested, which is the only
thing that could be tested as there's not other "context" or
"reference node" involved in the API surface.

All of TEST_FIND is actually dead code, so dropping that is harmless.

Fixes https://github.com/web-platform-tests/wpt/issues/11212.

### DIFF
--- a/dom/nodes/selectors.js
+++ b/dom/nodes/selectors.js
@@ -70,7 +70,7 @@ var validSelectors = [
   {name: "Type selector, matching body element", selector: "body", expect: [] /*no matches*/, exclude: ["document"],                        level: 1, testType: TEST_QSA},
 
   // Universal Selector
-  // Testing "*" for entire an entire context node is handled separately.
+  {name: "Universal selector, matching all elements",                                    selector: "*",              expect: ["universal", "universal-p1", "universal-code1", "universal-hr1", "universal-pre1", "universal-span1", "universal-p2", "universal-a1", "universal-address1", "universal-code2", "universal-a2"], level: 2, testType: TEST_MATCH},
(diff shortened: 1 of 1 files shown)


annevk (approved):

foolip:
From https://travis-ci.org/web-platform-tests/wpt/jobs/385179503 it looks like the "In-document Element.webkitMatchesSelector: :target pseudo-class selector, matching the element referenced by the URL fragment identifier (with no refNodes): :target" test is flaky. This was apparently already the case. It fails on http://w3c-test.org/dom/nodes/Element-webkitMatchesSelector.html reliably for me and passes on http://web-platform.test:8000/dom/nodes/Element-webkitMatchesSelector.html reliably, so it's probably something racy that depends on network speed.

foolip:
The same discrepancy can be seen in Element-matches.html, unsurprisingly.

annevk:
That test always passes in Firefox Nightly...

foolip:
The test passes in all browsers on wpt.fyi:
https://wpt.fyi/results/dom/nodes/Element-matches.html
https://wpt.fyi/results/dom/nodes/Element-webkitMatchesSelector.html

The test does wait for `frame.onload = this.step_func_done(e => init(e, "webkitMatchesSelector" ))` so this smells like a Chrome bug. The element that should match `:target` is set here:
https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/frame/local_frame_view.cc?l=1818&rcl=5dccd782be312708e659573fbba7effa07f4fb38

This hasn't been touched in ages, last time was @rwlbuis in https://trac.webkit.org/changeset/89745/webkit. But I can't get the subtest to fail in Safari 11.1.

domenic:
What's the plan for landing this?

foolip:
The flakiness is pre-existing, and I don't have the time to investigate it deeply, so I will give Travis another go, and if it still fails just open an issue for it and merge.

foolip:
Travis was happy this time. Probably just lucky, but no issue filed.

PR #11230, opened 2018-05-29
Use .any.js for WebCryptoAPI tests

zcorpan:
`encrypt_decrypt/rsa.https.any.js` is marked as [ Slow ] in WebKit.


WebCryptoAPI/idlharness.https.any.js
```diff
@@ -0,0 +1,23 @@
+// META: script=/resources/WebIDLParser.js
+// META: script=/resources/idlharness.js
+setup(function() {
+    var request = new XMLHttpRequest();
+    request.open("GET", "../interfaces/WebCryptoAPI.idl");
```
foolip: If there's a typo here, the test will just time out, because there's no such thing as async setup. Can you copy the style of encoding/idlharness.any.js instead? Thanks to @lukebjerring that style is used in lots of idlharness.js tests now, and it makes them quite a bit more compat/uniform.

foolip (approved):
LGTM % idlharness.js test. With any change like this, there's also a significant risk of accidental regression. @lukebjerring, do you have instructions for using the wpt.fyi diff feature for a case like this, in order to exhaustively compare a before/after state of a PR, if the results are collected manually?

lukebjerring:
Currently your best bet is to run the equivalent change through the Chromium CI (by copying the files to `src/third_party/WebKit/LayoutTests/external/wpt`

wpt.fyi is certainly closer to allowing arbitrary uploads, but not quite there.

foolip:
@zcorpan let me know if you'd like to run this through Chromium's CQ. Not necessary if you have confidence in the change by other means, of course.

WebCryptoAPI/idlharness.https.any.js
zcorpan: done

zcorpan:
I think the changes here are straightforward enough to not make that necessary.

WebCryptoAPI/idlharness.https.any.js
```diff
@@ -0,0 +1,14 @@
+// META: global=window,worker
+// META: script=/resources/WebIDLParser.js
+// META: script=/resources/idlharness.js
+
+promise_test(async() => {
+  const text = await (await fetch('/interfaces/WebCryptoAPI.idl')).text();
+  const idl_array = new IdlArray();
+  idl_array.add_untested_idls("[Global] interface Window { };");
+  idl_array.add_untested_idls("interface ArrayBuffer {};");
+  idl_array.add_untested_idls("interface ArrayBufferView {};");
```
lukebjerring: While you're here... you can avoid having explicit "mock" idl by fetching `/interfaces/html.idl` and using

    idl_array.add_dependency_idls(html);

zcorpan: done

foolip (approved):

lukebjerring (approved):
Looks good - would suggest a run through CI though - easy to cause regressions when automating the deps.

zcorpan:
OK. @foolip ?

foolip:
Copied changes into https://chromium-review.googlesource.com/c/chromium/src/+/1100877

zcorpan:
@foolip ping.

foolip:
There are results in the above link, for example https://test-results.appspot.com/data/layout_results/linux_chromium_rel_ng/118043/layout-test-results/results.html.

I think to make sense of the results they have to be compared against existing -expected.txt files in https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/external/wpt/WebCryptoAPI/. At least idlharness.https.any.serviceworker.html seems like real breakage.

foolip (approved):

foolip:
Looks like unstable results in https://travis-ci.org/web-platform-tests/wpt/jobs/394730448 :/

zcorpan:
@lukebjerring can you take over this? Otherwise I'll get back to this later this year.

PR #11237, opened 2018-05-29
Move input-events IDL to interfaces

csnardi:
The input-events IDL was previously directly written in the test. This change moves it to a dedicated file in interfaces.

This change also adds InputEventInit as an untested IDL at the beginning of the test so the test can actually run.


foolip (approved):
Yay!

foolip:
@lukebjerring it looks like the IDL auto-updating process never discovered this spec. Can you tell why that is? It is listed in https://foolip.github.io/day-to-day/.

lukebjerring:
Unsure - @mdittmer  ?

PR #11238, opened 2018-05-30
Escape attributes named 'required' in HTML IDL

csnardi:
Since `required` is a reserved name in WebIDL, it should be escaped with an underscore when used as an attribute name.

Change for https://github.com/whatwg/html/pull/3721, should be merged when that PR is merged.


jdm (approved):

annevk:
Closing per HTML Standard discussion.

PR #11243, opened 2018-05-30
[css-contain] Move contain-paint-013-ref.html file to reference/

mrego:
All the reference files of this test suite are in reference/ folder,
so following the same approach for contain-paint-013 test.


fred-wang (approved):
LGTM

PR #11249, opened 2018-05-30
Fetch: port blocking change (427, 548)

annevk:
See https://github.com/whatwg/fetch/pull/738.

### DIFF
--- a/fetch/api/request/request-bad-port.html
+++ b/fetch/api/request/request-bad-port.html
@@ -46,7 +46,8 @@
         143,  // imap2
         179,  // bgp
         389,  // ldap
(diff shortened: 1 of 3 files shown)


Ms2ger:
Can we deduplicate those lists? (At least the two in tests.)

annevk:
I wouldn't mind doing that as a follow-up, assuming it's okay for /websockets to take a dependency on /fetch.

I was initially planning on doing it, but it seemed bad to make `is_bad_port` depend on disk I/O, and then I wasn't quite sure whether it was worth it for the tests.

gsnedders (approved):

PR #11204, opened 2018-05-28
Split up WebCryptoAPI/derive_bits_keys/ with `variant`

zcorpan:
Also use .any.js.

Fixes #11203.

PR #11251, opened 2018-05-30
Add support for // META: title=...

zcorpan:
As suggested by @inexorabletash in https://github.com/web-platform-tests/wpt/pull/11204#discussion_r191495625

### DIFF
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -160,10 +160,10 @@ class HtmlWrapperHandler(WrapperHandler):
             if value == b"long":
                 return '<meta name="timeout" content="long">'
         if key == b"script":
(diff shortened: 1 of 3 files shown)


jgraham (approved):

tools/serve/serve.py
```diff
@@ -162,6 +162,9 @@ def _meta_replacement(self, key, value):
         if key == b"script":
             attribute = value.decode('utf-8').replace('"', "&quot;").replace(">", "&gt;")
```
zcorpan: I'm not sure why > is escaped here, can we remove that while we're here?

zcorpan: Done

Issue #10347, opened 2018-04-06
Add support for cascading IDL dependencies to idlharness.js


lukebjerring:
For new / separate specs, often dependencies on the original interface are created, e.g. when `partial`-ing out an existing interface. Often, that interface also inherits other interfaces, and can be a nightmare to isolate/filter.

Several things can happen in attempt to resolve the problem:

- We end up with a 'mock' of the interface, e.g. `idl_array.add_untested_idls('interface Document {};');`
- We import massive files like `dom.idl` in their entirety
- We filter imports, but end up filtering the whole dep chain

None of these is desirable or robust, nor documented. 

Resolving this issue would entail formalized behaviour of the framework for dependencies (and, optionally mocks) in a way that reduces boiler-plate and mitigates test breakages for changes to dependencies.

foolip:
@lukebjerring, should we consider this resolved by https://github.com/web-platform-tests/wpt/pull/10381?

lukebjerring:
(1 year later...) Yes, we should.

PR #11252, opened 2018-05-30
Add MouseEvent/MouseEventInit as untested DOM IDLs

csnardi:
The DOM idlharness test would not run as `MouseEvent`/`MouseEventInit` were not declared within any IDL files parsed. Add these in as untested IDLs.

### DIFF
--- a/dom/interfaces.html
+++ b/dom/interfaces.html
@@ -19,10 +19,7 @@ element.setAttribute("bar", "baz");
 var idlArray = new IdlArray();
 
 function doTest([html, dom]) {
-  // HTML is needed for EventHandler. Provide a dummy interface for
-  // LinkStyle which HTML depends on but we're not testing.
-  idlArray.add_untested_idls('interface LinkStyle {};');
-  idlArray.add_untested_idls(html);
+  idlArray.add_dependency_idls(html);
   idlArray.add_idls(dom);
   idlArray.add_objects({
     EventTarget: ['new EventTarget()'],


annevk:
I don't really understand this.

Presumably this is because `html.idl` is imported, but `html/dom/interfaces.https.html` does not include a mention of `MouseEvent` either. So what's going on?

csnardi:
`html/dom/interfaces.https.html` pulls in UIEvents as untested IDL, which defines `MouseEvent`/`MouseEventInit`. I could do that as well, but I figured it might just be easier to do it this way.

annevk:
If the dependencies are equivalent it seems better to copy HTML's style as that seems more forward compatible.

annevk:
Although I guess if someone updates HTML and forgets about DOM we're here again. Hmm.

annevk:
cc @kereliuk @foolip

gsnedders:
Related: #10347

foolip:
@csnardi can you try the mechanism that @lukebjerring introduced in https://github.com/web-platform-tests/wpt/pull/10381? By using `add_dependency_idls(html)` I think the issue might be avoided.

csnardi:
Using `add_dependency_idls` with UIEvents worked, so I updated the PR.

dom/interfaces.html
```diff
@@ -18,11 +18,12 @@ <h1>DOM IDL tests</h1>
 
 var idlArray = new IdlArray();
 
-function doTest([html, dom]) {
+function doTest([html, dom, uievents]) {
   // HTML is needed for EventHandler. Provide a dummy interface for
   // LinkStyle which HTML depends on but we're not testing.
   idlArray.add_untested_idls('interface LinkStyle {};');
   idlArray.add_untested_idls(html);
```
foolip: I was wondering if using `add_dependency_idls` *here* would do the trick, and avoid 'interface LinkStyle {};' and the extra bits from uievents entirely.

csnardi: Oh, that makes more sense. That does indeed work.

foolip (approved):
LGTM assuming the number of passing/failing tests is unchanged.

csnardi:
The number of passing/failing tests was unchanged from my first commit to this -- currently no tests are run due to this dependency issue.

foolip:
Awesome, merge away when Travis is happy!

Issue #10338, opened 2018-04-05
Multiple 'Exposed' attributes unhandled by idlharness.js


lukebjerring:
e.g. From https://wicg.github.io/budget-api/

    [Exposed=Window]
    partial interface Navigator {
        [SameObject] readonly attribute BudgetService budget;
    };

https://github.com/w3c/web-platform-tests/pull/10337 improved the error message somewhat, but  we should be applying the `Exposed` to the nested members and asserting that it's a subset of the original interface's exposure set.

foolip:
[Exposed] on partial interfaces is allowed, https://heycam.github.io/webidl/#Exposed says "If [Exposed] appears on a partial interface or partial namespace, then the partial’s own exposure set must be a subset of the exposure set of the partial’s original interface or namespace."

foolip:
https://github.com/w3c/web-platform-tests/pull/10337 is about multiple `Exposed` extended attributes syntactically (in the AST) but the example given has just one.

So the issue may not be with duplication as such, but with how idlharness.js merges extended attributes for partial interfaces?

https://github.com/w3c/web-platform-tests/blob/89daad32c029798b7203a36f21598fcba470e07d/resources/idlharness.js#L628

However, I'm afraid I can't spot the bug, because it seems to (correctly) just forward the extended attributes to all of the members, and the extended attributes on non-partial interfaces aren't propagated like that (they usually say something about the interface itself). So where's the duplication coming from?

lukebjerring:
The spot you link to is correct. The issue is that the `extAttrs` for partials are copied to the main interface (not just to the members inside the partial).
Instead, they should be asserted to be a subset of the existing attribute (if present).

lukebjerring:
So, taking a stab at this, using https://github.com/tidoust/reffy-reports/blob/master/whatwg/idl/webusb.idl as an example.

    [Exposed=(DedicatedWorker,SharedWorker), SecureContext]
    partial interface WorkerNavigator {
      [SameObject] readonly attribute USB usb;
    };

Produced

> "DedicatedWorker exposed on partial WorkerNavigator interface, but not on main interface."

Sure enough, `DedicatedWorker` is not exposed on `WorkerNavigator`. I'm not convinced I'm solving anything, just more yak shaving...

foolip:
OK, sounds like a bug in either WebUSB or HTML.

foolip:
[...]

PR #11253, opened 2018-05-30
Move mediasession IDL to interfaces

csnardi:
The mediasession IDL was defined directly in the idlharness test. This change moves the IDL definitions into its own file in interfaces. This change also makes the construction of `MediaMetadata` be quoted, so the test can proceed even if it is not defined.

### DIFF
--- a/mediasession/idlharness.html
+++ b/mediasession/idlharness.html
@@ -11,78 +11,28 @@
 </head>
 <body>
 <h1>Media Session IDL tests</h1>
-
-<pre id='untested_idl' style='display:none'>
-[Global=Window, Exposed=Global]
-interface Window {
-};
-
-interface Navigator {
-};
-</pre>
-
-<pre id='idl'>
-[Exposed=Window]
-partial interface Navigator {
-  [SameObject] readonly attribute MediaSession mediaSession;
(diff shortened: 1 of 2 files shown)


mediasession/idlharness.html
```diff
@@ -11,78 +11,28 @@
...
-idl_array.add_objects({
-  MediaMetadata: [new MediaMetadata()],
-  Navigator: ["navigator"]
-});
-idl_array.test();
+"use strict";
+
+function doTest([mediasession]) {
+    var idl_array = new IdlArray();
+    idl_array.add_untested_idls('interface Navigator {};');
```
lukebjerring: Same as described in https://github.com/web-platform-tests/wpt/pull/11255#pullrequestreview-124498586

Use `idl_array.add_dependency_idls`.

lukebjerring (changes requested):

mediasession/idlharness.html
lukebjerring: As discussed, that is blocked on https://github.com/web-platform-tests/wpt/issues/10338

lukebjerring (approved):

PR #11255, opened 2018-05-30
Move media-capabilities IDL to interfaces

csnardi:
The media-capabilities IDL was defined directly in the idlharness test. This change moves the IDL definitions into its own file in interfaces. This change also updates the definitions per the latest version of the spec.

### DIFF
--- a/media-capabilities/idlharness.html
+++ b/media-capabilities/idlharness.html
@@ -11,81 +11,26 @@
 </head>
 <body>
 <h1>Media Session IDL tests</h1>
-<pre id='untested_idl' style='display:none'>
-[Global=Window, Exposed=Window]
-interface Window {
-};
-interface Worker {
-};
-interface Navigator {
-};
-interface WorkerNavigator {
-};
-</pre>
-<pre id='idl'>
-dictionary MediaConfiguration {
-  VideoConfiguration video;
-  AudioConfiguration audio;
-};
-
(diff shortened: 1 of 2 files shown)


media-capabilities/idlharness.html
```diff
@@ -11,81 +11,26 @@
...
-idl_array.add_untested_idls(document.getElementById("untested_idl").textContent);
-idl_array.add_idls(document.getElementById("idl").textContent);
-idl_array.add_objects({
-  Navigator: ["navigator"]
-});
-idl_array.test();
+"use strict";
+function doTest([media_capabilities]) {
+    var idl_array = new IdlArray();
+    idl_array.add_untested_idls('interface Navigator {};');
```
lukebjerring: I recently added an idlharness.js feature:

   idl_array.add_dependency_idls(source);

Example https://github.com/web-platform-tests/wpt/pull/11076/files#diff-ad65774747290dda8f198aa62706d1c1

It's better than 'mocking' the required IDLs.

csnardi: Hm, if I do that, then we run into #10338, since the IDL marks the partial Navigator interfaces with `[Exposed=(Window)]`.

lukebjerring: ah, that ol' chestnut. Disregard my change request :)

lukebjerring (approved):

PR #11264, opened 2018-05-31
HTML: Test hidden attribute on table elements

zcorpan:
(I found this when adding the "Filter" feature in https://bocoup.github.io/wpt-disabled-tests-report/ -- typing something should not change the column widths of the tables.)

### DIFF
--- a/html/rendering/non-replaced-elements/tables/hidden-attr.html
+++ b/html/rendering/non-replaced-elements/tables/hidden-attr.html
@@ -4,6 +4,7 @@
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <table hidden></table>
+<table><caption hidden></caption></table>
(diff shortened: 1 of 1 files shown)


zcorpan:
https://bugzilla.mozilla.org/show_bug.cgi?id=1465724
https://bugs.webkit.org/show_bug.cgi?id=186131
https://bugs.chromium.org/p/chromium/issues/detail?id=848189
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/17718818/

Ms2ger (approved):
Feel free to also add a test for `caption` elements

zcorpan:
done

FremyCompany:
@zcorpan Visibility:collapse doesn't do anything on TH and TD elements. I doubt the proposed behavior is web-compatible in that case. I filed a spec issue on this.

PR #11266, opened 2018-05-31
WebKit export of https://bugs.webkit.org/show_bug.cgi?id=170784

fred-wang:
WebKit export from bug: [\[iOS\] Animations with Bézier timing function not suspended on UI process when animation-play-state is set to "paused"](https://bugs.webkit.org/show_bug.cgi?id=170784)

### DIFF
--- a/css/css-animations/set-animation-play-state-to-paused-001-ref.html
+++ b/css/css-animations/set-animation-play-state-to-paused-001-ref.html
@@ -13,7 +13,7 @@
         position: absolute;
         background: lightgreen;
         width: 150px;
-        height: 250px;
+        height: 300px;
         left: -75px;
         top: 0px;
         transform: translate(150px);
--- a/css/css-animations/set-animation-play-state-to-paused-001-ref.html
+++ b/css/css-animations/set-animation-play-state-to-paused-001-ref.html
@@ -1,0 +1,29 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Dynamically set animation-play-state to paused (reference)</title>
+    <style>
+      #container {
+        position: absolute;
+        left: 0;
+        top: 3em;
+      }
+      #coveringRect {
+        position: absolute;
+        background: lightgreen;
+        width: 150px;
+        height: 250px;
+        left: -75px;
+        top: 0px;
+        transform: translate(150px);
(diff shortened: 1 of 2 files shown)


birtles:
We can't write this as a scripted test that just compares the values returned by `getComputedStyle` across animation frames?

(I'm concerned that this test will become flaky, or simply useless because on our CI we often can have tens of seconds between animation frames when the test machine is under load whereas this test seems to be dealing with changes in hundreds of milliseconds.)

fred-wang:
> We can't write this as a scripted test that just compares the values returned by getComputedStyle across animation frames?
> (I'm concerned that this test will become flaky, or simply useless because on our CI we often can have tens of seconds between animation frames when the test machine is under load whereas this test seems to be dealing with changes in hundreds of milliseconds.)

I agree flakiness is a potential issue, that's why the test is not too strict: the animation not too fast and the covering rect is wider than the animated squares.

I initially tried doing getComputedStyle() but the issue is that the visual rendering does not necessary match the value returned from the DOM. In this particular WebKit iOS bug, the UI Process continues to animate the squares while the Web process claims they are paused.

birtles:
Looks like the test fails on Firefox.

https://travis-ci.org/web-platform-tests/wpt/jobs/386104808#L793-L798

fred-wang:
OK, it worked for me locally but probably this is because the test is flaky. I think I can verify getComputedStyle() to make the test a bit more robust and also fire the pause in a small interval around the target time instead of only after it.

fred-wang:
@birtles  What do you think of this new version?

The parsing of window.getComputedStyle(...).transform is not really nice and the minimum duration is now always 2s but at least the whole thjing should be more accurate than relying on timestamp.

fred-wang:
It's failing again under Firefox Nightly, but I'm not able to reproduce the issue locally. Is there a way to get the screenshots from travis? Otherwise, I guess I should send the test to the Mozilla try sever...

fred-wang:
@birtles  css/css-animations/set-animation-play-state-to-paused-001.html passes in Try server, so probably it's an issue with the Travis WPT bot:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=07483dcde214241025fc1ad7f0622bdcd3aeea9f

fred-wang:
.

PR #11268, opened 2018-05-31
[css-properties-values-api] Avoid uncaught exception for CSS.registerProperty()

zcorpan:
Part of #11269.


csnardi (approved):
Looks good.

Issue #11269, opened 2018-05-31
Fix tests with harness status ERROR or TIMEOUT


zcorpan:
From https://github.com/web-platform-tests/results-collection/issues/478#issuecomment-384131457

5 tests reporting ERROR in all browsers

- [x] [`websockets/interfaces/WebSocket/events/018.html`](https://wpt.fyi/websockets/interfaces/WebSocket/events/018.html?sha=6736e3f46b)
- [x] [`websockets/interfaces/WebSocket/events/018.html?wss`](https://wpt.fyi/websockets/interfaces/WebSocket/events/018.html?wss?sha=6736e3f46b)
- [x] [`css/css-properties-values-api/registered-property-cssom.html`](https://wpt.fyi/css/css-properties-values-api/registered-property-cssom.html?sha=6736e3f46b)
- [x] [`service-workers/service-worker/clients-get-client-types.https.html`](https://wpt.fyi/service-workers/service-worker/clients-get-client-types.https.html?sha=6736e3f46b)
- [x] [`service-workers/service-worker/clients-get.https.html`](https://wpt.fyi/service-workers/service-worker/clients-get.https.html?sha=6736e3f46b)

42 tests reporting TIMEOUT in all browsers

- [x] [`feature-policy/feature-policy-header-policy-allowed-for-self.https.sub.html`](https://wpt.fyi/feature-policy/feature-policy-header-policy-allowed-for-self.https.sub.html?sha=6736e3f46b)
- [x] [`feature-policy/feature-policy-header-policy-allowed-for-all.https.sub.html`](https://wpt.fyi/feature-policy/feature-policy-header-policy-allowed-for-all.https.sub.html?sha=6736e3f46b)
- [x] [`feature-policy/feature-policy-nested-header-policy-allowed-for-self.https.sub.html`](https://wpt.fyi/feature-policy/feature-policy-nested-header-policy-allowed-for-self.https.sub.html?sha=6736e3f46b)
- [x] [`feature-policy/feature-policy-header-policy-disallowed-for-all.https.sub.html`](https://wpt.fyi/feature-policy/feature-policy-header-policy-disallowed-for-all.https.sub.html?sha=6736e3f46b)
- [x] [`feature-policy/feature-policy-nested-header-policy-allowed-for-all.https.sub.html`](https://wpt.fyi/feature-policy/feature-policy-nested-header-policy-allowed-for-all.https.sub.html?sha=6736e3f46b)
[...]

PR #11270, opened 2018-05-31
Fail early if Feature Policy is not supported

zcorpan:
Part of #11269.

### DIFF
--- a/feature-policy/resources/featurepolicy.js
+++ b/feature-policy/resources/featurepolicy.js
@@ -1,3 +1,7 @@
+// Feature test to avoid timeouts
+function assert_feature_policy_supported() {
+  assert_not_equals(document.policy, undefined, 'document.policy');
+}
 // Tests whether a feature that is enabled/disabled by feature policy works
 // as expected.
 // Arguments:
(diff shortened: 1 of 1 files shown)


feature-policy/resources/featurepolicy.js
```diff
@@ -1,3 +1,7 @@
+// Feature test to avoid timeouts
+function assert_feature_policy_supported() {
+  assert_not_equals(document.policy, undefined, 'document.policy');
```
foolip: Maybe "Feature Policy is supported" as the description, since this isn't really about testing document.policy, and making it trivial to understand the failure reason?

foolip (approved):
one nit

feature-policy/resources/featurepolicy.js
zcorpan: fixed

foolip (approved):

zcorpan:
This seems to be timing out in Travis because too many tests, and too verbose output. May need to admin-merge.

foolip:
@zcorpan I'll try rebasing first since @gsnedders fixed the verbosity.

foolip:
Actually, no, there was a timeout, which is https://github.com/w3c/web-platform-tests/issues/7660, still unresolved.

PR #11273, opened 2018-05-31
[css-paint-api] Fail a reftest early if CSS.registerProperty is not s…

zcorpan:
…upported

Part of #11269.


zcorpan:
4051f1e should probably have been in #11268 instead, but maybe it's fine as-is...

shans (approved):

PR #11281, opened 2018-05-31
Allow cookie-store IDL test to run even if objects not defined

csnardi:
By quoting the object constructors/references, we prevent the test from erroring out prior to running and instead allow all of the tests to run. This means that if an implementation only defines one interface but not the other, it will make sure there is test coverage of that interface without having to implement the other.


pwnall (approved):
LGTM, thank you!

PR #11288, opened 2018-06-01
URL: correct instructions for testing URL's base argument

### LINKED ISSUE #11287: [URL] Parsing `about:blank` against `base` can't give failure
The PR #10955 changed testing requirements slightly. But I think the following pattern is incorrect::
> ... if `failure` is true, parsing `about:blank` against `base` must give failure.

The `base` is required to be valid, so this parsing should not return failure.

I guess there must be `input`, not `base`, i.e.:
> ... if `failure` is true, parsing `about:blank` against **`input`** must give failure.

The testing code uses `input` as well:
https://github.com/web-platform-tests/wpt/blob/f696c249c6d41c32bdc04d0bec688ee00228a0cf/url/failure.html#L20-L22

/cc @domenic @annevk

annevk:
Fixes #11287.

### DIFF
--- a/url/README.md
+++ b/url/README.md
@@ -23,7 +23,7 @@ The keys for each test case are:
 
 In addition to testing that parsing `input` against `base` gives the result, a test harness for the
 `URL` constructor (or similar APIs) should additionally test the following pattern: if `failure` is
-true, parsing `about:blank` against `base` must give failure. This tests that the logic for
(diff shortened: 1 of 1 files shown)


rmisev (approved):

domenic (approved):
Sorry about that

PR #11291, opened 2018-06-01
Test identity encoding for range requests

jakearchibald:
Tests for https://github.com/whatwg/fetch/pull/751.

### DIFF
--- a/fetch/range/general.window.js
+++ b/fetch/range/general.window.js
@@ -1,0 +1,30 @@
+// META: script=resources/utils.js
+// META: script=/common/utils.js
+
+const onload = new Promise(r => window.addEventListener('load', r));
+
+// It's weird that browsers do this, but it should continue to work.
+promise_test(async t => {
+  await loadScript('resources/partial-script.py?pretend-offset=90000');
+  assert_true(self.scriptExecuted);
+}, `Script executed from partial response`);
+
+promise_test(async () => {
+  const wavURL = new URL('resources/long-wav.py', location);
+  const stashTakeURL = new URL('resources/stash-take.py', location);
+  const stashToken = token();
+  wavURL.searchParams.set('accept-encoding-key', stashToken);
+  stashTakeURL.searchParams.set('key', stashToken);
+
+  // The testing framework waits for window onload. If the audio element
+  // is appended before onload, it extends it, and the test times out.
+  await onload;
+
+  const audio = appendAudio(document, wavURL);
+  await new Promise(r => audio.addEventListener('progress', r));
+  //await new Promise(r => audio.addEventListener('error', r));
+  audio.remove();
+
+  const response = await fetch(stashTakeURL);
+  assert_equals(await response.json(), 'identity', `Expect identity accept-encoding on media request`);
+}, `Fetch with range header will be sent with Accept-Encoding: identity`);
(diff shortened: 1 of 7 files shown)


fetch/range/general.window.js
```diff
@@ -0,0 +1,30 @@
+// META: script=resources/utils.js
+// META: script=/common/utils.js
+
+const onload = new Promise(r => window.addEventListener('load', r));
+
+// It's weird that browsers do this, but it should continue to work.
+promise_test(async t => {
+  await loadScript('resources/partial-script.py?pretend-offset=90000');
+  assert_true(self.scriptExecuted);
+}, `Script executed from partial response`);
```
jakearchibald: This test already existed, but I renamed the file and git didn't realise.

fetch/range/resources/range-sw.js
```diff
@@ -108,7 +111,7 @@ function rangeHeaderPassthroughTest(event) {
     promise_test(async () => {
       await fetch(event.request);
       const response = await fetch('stash-take.py?key=' + key);
-      assert_equals(await response.json(), '"range-header-received"');
+      assert_equals(await response.json(), 'range-header-received');
```
jakearchibald: Bug-fixing an existing test.

fetch/range/sw.https.window.js
```diff
@@ -4,28 +4,21 @@
...
-const SCOPE = 'resources/basic.html' + Math.random();
-
-function appendAudio(document, url) {
-  const audio = document.createElement('audio');
-  audio.muted = true;
-  audio.src = url;
-  audio.preload = true;
-  document.body.appendChild(audio);
-}
+const BASE_SCOPE = 'resources/basic.html?';
```
jakearchibald: I'm now generating a unique scope for each test, it avoid the old service worker coming back from the dead in Safari.

fetch/range/resources/long-wav.py
```diff
@@ -50,11 +50,28 @@ def main(request, response):
...
         # This is later collected using stash-take.py
-        request.stash.put(range_received_key, 'range-header-received', '/fetch/range/')
+        request.server.stash.put(range_received_key, 'range-header-received', '/fetch/range/')
+
+    if accept_encoding_key:
+        # Remove any current value
+        request.server.stash.take(
+            accept_encoding_key,
+            '/fetch/range/'
+        )
```
annevk: Why is this needed?

fetch/range/resources/long-wav.py
```diff
@@ -50,11 +50,28 @@ def main(request, response):
     response.headers.set("Cache-Control", "no-cache")
 
     range_header = request.headers.get('Range', '')
+    range_header_match = range_header and re.search(r'^bytes=(\d*)-(\d*)$', range_header)
```
annevk: So this is either None (or False?) or a regular expression result?

annevk (commented):
I don't see anything that I know to be wrong, but I have a couple minor questions.

fetch/range/resources/long-wav.py
jakearchibald: Safari appears to make two requests in quick succession. Without this, an error is thrown, since it tries to put something in the stash when that key already exists.

fetch/range/resources/long-wav.py
jakearchibald: Yep

annevk (approved):

PR #11292, opened 2018-06-01
Fix bugs in split-up encoding tests

### LINKED ISSUE #11015: /encoding/* are disabled in Chromium and WebKit

zcorpan:
Bugs were introduced in d6c29be

* Some script paths were wrong
* A big5 test referenced eucjpEncoder instead of big5Encoder
* That big5 test also referenced encode-form-common.js instead of
  encode-href-common.js

Fixes #11015.


encoding/legacy-mb-korean/euc-kr/euckr-encode-form-errors-han.html
```diff
@@ -43,7 +250,7 @@
 </head>
 <body>
 <div id="log"></div>
-<script src="../../resources/ranges.js"></script>
+<script src="../../ranges.js"></script>
```
inexorabletash: Why is `resources` being _removed_ from the path here?

encoding/legacy-mb-korean/euc-kr/euckr-encode-form-errors-han.html
```diff
@@ -52,10 +259,7 @@
 function expect(result, codepoint) {
   return "%26%23" + codepoint + "%3B";
 }
-function normalizeStr(str) {
-  return str;
-}
 </script>
-<script src="../../resources/encode-form-common.js"></script>
+<script src="../../encode-form-common.js"></script>
```
inexorabletash: Why is `resources` being _removed_ from the path here?

common/subset-tests.js
```diff
@@ -9,13 +9,30 @@
...
+      if (match[2] !== "last") {
+          subTestEnd = parseInt(match[2], 10);
       }
+    }
+    match = /(?:^\?|&)split=(\d+)(?:&|$)/.exec(location.search);
+    if (match) {
+      var testsPerVariant = parseInt(match[1], 10);
+      add_completion_callback(tests => {
+        var total = tests.length;
+        var template = '<meta name="variant" content="?%s-%s">';
```
inexorabletash: This looks like temporary code used to help generate the patch. Remove it?

(Or I'm really missing the magic here.)

inexorabletash (changes requested):

common/subset-tests.js
zcorpan: I thought I'd keep it in as utility for the future.

Hexcles:
Thanks for the fix!

Looking at the patch, I found something interesting: e.g. [encoding/legacy-mb-tchinese/big5/big5-encode-href-errors-han.html](https://github.com/web-platform-tests/wpt/pull/11292/files#diff-d29744659cdc71a6610d8fdea3b03a6b)

This is one of tests that still [time out after splitting](https://github.com/web-platform-tests/wpt/issues/11015#issuecomment-393719550). However, only some of the ranges time out (those after 15001), while the others, despite also having the size of 1000, run pretty fast. Then in this PR, the end of the range seems to be smaller than 15000. Was there a bug in generating the range indexes previously? That could explain what we are seeing.

If so, then the [timeout I described](https://github.com/web-platform-tests/wpt/issues/11015#issuecomment-393719550) are really caused by bugs (out-of-bound indexes) instead of the bucket size being too large. (I also think 1000 is small enough.) Can we try to fix the bugs first without decreasing the bucket size?

common/subset-tests.js
inexorabletash: Okay - add comments then please?

At first glance I thought it was going to new functionality to do this automagically like `<meta name="split" content="200">` or something and was very confused...

zcorpan: Sure.

zcorpan:
@Hexcles ok I'll fix on Monday.

Hexcles:
@zcorpan thanks! And I can help you to test out the patch using Chromium CI; let me know when it's ready.

common/subset-tests.js
zcorpan: Done.

zcorpan:
@Hexcles OK now done.

Hexcles:
@zcorpan trying in Chromium: https://chromium-review.googlesource.com/c/chromium/src/+/1087496

I'll let you know the results in a couple hours.

Hexcles (approved):
LGTM.

The differences caused by this patch seem quite reasonable: https://test-results.appspot.com/data/layout_results/linux_trusty_blink_rel/30577/layout-test-results/results.html

And there's no timeout!

zcorpan:
Nice, thanks @Hexcles!

@inexorabletash care to take another look?

inexorabletash (approved):

zcorpan:
Merging despite failed status, per discussion in irc
```
jgraham The issue with the PR is the normal "if tests take >5 minutes to run once, we can't run them all in 49½ minutes" issue
jgraham The other issue is that Travis is massively behind
jgraham and has had problems
Hexcles and for you restarting will never help because there's just too many of them
zcorpan_ ok so restarting will just time out again, and not make the queue shorter
Hexcles I vote +1 for admin-merging it because I've tested it in Chromium
Hexcles jgraham ?
jgraham That seems fine, if you tested that they work in Firefox & Chrome
```

PR #11294, opened 2018-06-01
HTML: Expect salvegable to be set to false for pagehide test

zcorpan:
Also rename the manual test.

Part of #11269.

---

I will report an issue for whatwg/html about this test.


html/browsers/browsing-the-web/unloading-documents/unload/006-1.html
```diff
@@ -4,9 +4,12 @@
 onpagehide = function() {
   onpagehide = null;
   setTimeout(function() {
-               parent.t.done()
+               parent.t.step_func(function() { assert_unreached('setTimeout survived navigatoin'); })
```
foolip: `parent.t.unreached_func('setTimeout survived navigatoin')` would be more compact.

html/browsers/browsing-the-web/unloading-documents/unload/pagehide-manual.html
```diff
@@ -2,4 +2,4 @@
 <title>Document salvagable state after setting pagehide handler</title>
 <script>onpagehide = function() {setTimeout(function(){document.body.innerHTML = "PASS"}, 100)}</script>
 <p>Click the link below then navigate back to this page. Shortly after returning you should see the text "PASS"</p>
-<p><A href="manual-001-1.html">Click here</a>
+<p><a href="pagehide-manual-1.html">Click here</a>
```
foolip: Clicking will now 404, is that part of the test, or accidental?

zcorpan: It's not 404, that file was renamed.

html/browsers/browsing-the-web/unloading-documents/unload/006-1.html
zcorpan: done

html/browsers/browsing-the-web/unloading-documents/unload/pagehide-manual.html
foolip: Oh, there's both pagehide-manual.html and pagehide-manual-1.html.

foolip (approved):

PR #11296, opened 2018-06-01
HTML: Run a history navigation test in a popup to report a result

zcorpan:
Part of #11269.

### DIFF
--- a/html/browsers/browsing-the-web/history-traversal/window-name-after-cross-origin-main-frame-navigation.sub.html
+++ b/html/browsers/browsing-the-web/history-traversal/window-name-after-cross-origin-main-frame-navigation.sub.html
@@ -1,13 +1,24 @@
 <!DOCTYPE HTML>
 <html>
 <head>
-    <!-- window.name should equal "" after a cross-origin main frame navigation. -->
     <script src='/resources/testharness.js'></script>
     <script src='/resources/testharnessreport.js'></script>
 </head>
 <body>
+    <button id="button" onclick="popup();">open popup</button>
     <script>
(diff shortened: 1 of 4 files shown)


html/browsers/browsing-the-web/history-traversal/window-name-after-cross-origin-main-frame-navigation.sub.html
```diff
@@ -1,13 +1,24 @@
...
     <script>
-        document.location = "support/window-name-navigation.sub.html?hostname={{domains[www1]}}&shouldhavename=false&sendmessage=false";
+    function popup() {
+      window.popupWin = window.open('support/window-name-after-cross-origin-main-frame-navigation-popup.sub.html', '_blank');
+    }
+    async_test(t => {
+      t.add_cleanup(() => {
+        popupWin.close();
+      })
+      document.getElementById('button').click();
```
foolip: Why the indirection via a button here? Popup blocking will still be in effect, so just inline all the code here?

foolip (approved):
LGTM % button indirection

html/browsers/browsing-the-web/history-traversal/window-name-after-cross-origin-main-frame-navigation.sub.html
zcorpan: Just to make it slightly more convenient when running the test manually.

foolip: Ah, OK

PR #11299, opened 2018-06-01
Ensure that Config's set of ports is generated atomically.

jdm:
My theory is that when there are several browsers pointed at the same server we can end up with two threads that enter the `ports` method simultaneously, create an empty `old_ports` dictionary, and generate unique ports for any "auto" entries. One of these threads will finish last and overwrite the other thread's results, leading to an inconsistent set of ports between threads.

### DIFF
--- a/tools/wptserve/wptserve/config.py
+++ b/tools/wptserve/wptserve/config.py
@@ -1,5 +1,6 @@
 import logging
 import os
+from threading import Lock
 
 from collections import defaultdict, Mapping
 
@@ -73,6 +74,8 @@ class Config(Mapping):
                 self.log_level = level_name
             self._logger_name = logger.name
 
+        self._ports_lock = Lock()
+
         for k, v in self._default.iteritems():
             setattr(self, k, kwargs.pop(k, v))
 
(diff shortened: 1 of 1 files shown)


gsnedders (changes requested):
This is failing CI because locks can't be pickled (yay multiprocessing!). Probably we should drop the lock in `__getstate__` and readd it in `__setstate__`?

jgraham:
Or we could just make the lifecycle of the config object simpler so that we can generate ports in the constructor and there's no race condition.

jdm:
Given that this is an attempt to alleviate a really common intermittent problem when running tests in Servo, and I really don't know this code well at all, could we limit the requests to changes that would be straightforward for me to make?

jdm:
The tests passed with my __getstate__/__setstate__ changes.

jgraham (approved):
This looks fine to fix the problem now. I still think we should fix things to prevent the config being changed arbitarily after it's created; we don't have a use case that requires that. But I didn't mean to suggest that had to be part of this PR.

gsnedders (approved):

PR #11300, opened 2018-06-01
HTML: split out navigation tests for URL query-encoding tests

zcorpan:
Part of #11269 and #4934.

The new test still times out because `ping` is not supported, but passes in Firefox with `send_pings` pref enabled.


annevk (approved):
Thanks for doing this! I have some style nits, but I'm never quite sure whether we should enforce those or not.

Will you file bugs against browsers too for stuff they fail? (Although maybe that's redundant with the bugs I filed already... Hmm.)

zcorpan:
Happy to file bugs

zcorpan:
Safari passes.
https://bugs.chromium.org/p/chromium/issues/detail?id=848756
https://bugzilla.mozilla.org/show_bug.cgi?id=401217
Edge: "You can not create more than 2 issues per hour."

PR #11302, opened 2018-06-01
Make dynamic-import tests fail early instead of time out

zcorpan:
Part of #11269.


domenic (approved):
LGTM, although a message might help explain things (e.g. "the browser must be able to parse a string containing the import() syntax into a function")

PR #11304, opened 2018-06-01
HTML: Use a URL that is more likely to fail to parse in img test

zcorpan:
Part of #1126.

---

Note that this doesn't seem to fix the timeout in browsers, but per spec it should fire an 'error' event. Will file bugs.

https://html.spec.whatwg.org/multipage/images.html#updating-the-image-data:parse-a-url-2

### DIFF
--- a/html/semantics/embedded-content/the-img-element/invalid-src.html
+++ b/html/semantics/embedded-content/the-img-element/invalid-src.html
@@ -8,7 +8,7 @@
 <script>
 async_test(function(t) {
     var img = document.getElementById("brokenurl");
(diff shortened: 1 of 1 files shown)


zcorpan:
https://bugs.chromium.org/p/chromium/issues/detail?id=627706
https://bugzilla.mozilla.org/show_bug.cgi?id=1466138
https://bugs.webkit.org/show_bug.cgi?id=186195
Edge: "You can not create more than 2 issues per hour." :(

foolip (approved):

PR #11306, opened 2018-06-01
Add filtering support to ManifestLoader.

Ms2ger:
In some use cases, the overhead of creating data structures for all tests in
the manifest, and then filtering them, is too big. This change allows users
to filter tests when initially creating the manifest.

### DIFF
--- a/tools/manifest/manifest.py
+++ b/tools/manifest/manifest.py
@@ -20,6 +20,13 @@ class ManifestVersionMismatch(ManifestError):
     pass
 
 
+def iterfilter(filters, iter):
+    for f in filters:
+        iter = f(iter)
+    for item in iter:
+        yield item
+
+
 class Manifest(object):
     def __init__(self, url_base="/"):
         assert url_base is not None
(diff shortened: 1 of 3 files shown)


Ms2ger:
@jgraham I hope you like this more than last week's eyesore :)

tools/manifest/manifest.py
```diff
@@ -20,6 +20,13 @@ class ManifestVersionMismatch(ManifestError):
     pass
 
 
+def iterfilter(filters, iter):
```
jgraham: Maybe call this `tests_iter` or something.

jgraham (approved):
Yes! Would love a followup to use this when loading tests in wptrunner generally since it probably saves a few ms sometimes.

tools/manifest/manifest.py
Ms2ger: I stole this from wptrunner; didn't seem worth it to try to find a place to share the code, but might as well keep the name

jgraham: OK

Ms2ger:
Removed in  #16363.

PR #11308, opened 2018-06-01
[encrypted-media] Add segmented clear-to-encrypted test.

TheModMaker:
This adds a test that tests transitioning from clear to encrypted content when it appears in different segments.  This can happen when streaming and the first content is a clear ad and the encrypted content
will be appended later.

This kind of content wasn't playable on Firefox until recently and I think this won't work on Edge if they supported clear-key (similar tests with PlayReady don't work).
See https://bugzilla.mozilla.org/show_bug.cgi?id=1454630


joeyparrish (approved):
LGTM

PR #11311, opened 2018-06-01
Update assert, help, reviewer and titles to match current ouput of w3c/woff2-tests

fred-wang:
The following command was applied: sed -i 's/<link  rel="reviewer"/<link rel="reviewer"/' *.xht
More documentation update has also been performed.

This matches the current test documentations at https://github.com/w3c/woff2-tests


fred-wang:
@svgeesus I suspect the only reason travis is failing is because there are two many files changed (603) so the tests take too long to run. If necessary, I can split the patch into smaller PRs.

svgeesus (approved):
Yes, looks fine.

fred-wang:
@jgraham It looks like travis fails because the CL modifies too many tests and the Firefox bot times out. How would be the proper way to land this change?

fred-wang:
cc @gsnedders

PR #11314, opened 2018-06-02
Add back the style for font install warning

fred-wang:
This is no longer used for the WPT tests but other generators from woff2-tests still require it, so let's
just keep it to make things easier.


fred-wang:
Cancelling since it seems UserAgent was actually the only test case using these notes.

fred-wang:
https://github.com/w3c/woff2-tests/pull/11

PR #11315, opened 2018-06-02
Correct importScripts() behavior test

jungkees:
- Update the after-install call to expect NetworkError instead of TypeError.
- Initialize echo_output before calling importScripts().


wanderview (approved):
Identical to what I came up with, but I didn't need the `echo_output = null` in `test_import()`.  Do you really need that?

jungkees:
Not for this particular case and status but for error prevention puppose for future changes.

PR #11324, opened 2018-06-04
Add test for logical box model longhand properties

Loirooriol:
It's a refactorization of Mozilla's [layout/style/test/test_logical_properties.html](https://searchfox.org/mozilla-central/rev/292d295d6b084b43b70de26a42e68513bb7b36a3/layout/style/test/test_logical_properties.html)

Cc @mrego

### DIFF
--- a/css/css-logical/logical-box-border-color.html
+++ b/css/css-logical/logical-box-border-color.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<meta charset="utf-8" />
+<title>CSS Logical Properties: Flow-Relative Border Colors</title>
+<link rel="author" title="Oriol Brufau" href="mailto:obrufau@igalia.com" />
+<link rel="help" href="https://drafts.csswg.org/css-logical/#border-color">
+<meta name="assert" content="This test checks the interaction of the flow-relative border-*-color longhand properties with the physical ones in different writing modes." />
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<div id="log"></div>
+
+<script src="./resources/test-box-properties.js"></script>
+<script>
+runTests(createBoxPropertyGroup("border-*-color", {type: "color"}));
+</script>
--- a/css/css-logical/logical-box-border-style.html
+++ b/css/css-logical/logical-box-border-style.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<meta charset="utf-8" />
+<title>CSS Logical Properties: Flow-Relative Border Styles</title>
+<link rel="author" title="Oriol Brufau" href="mailto:obrufau@igalia.com" />
+<link rel="help" href="https://drafts.csswg.org/css-logical/#border-style">
+<meta name="assert" content="This test checks the interaction of the flow-relative border-*-style longhand properties with the physical ones in different writing modes." />
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<div id="log"></div>
+
+<script src="./resources/test-box-properties.js"></script>
+<script>
+runTests(createBoxPropertyGroup("border-*-style", {type: "border-style"}));
+</script>
--- a/css/css-logical/logical-box-border-width.html
+++ b/css/css-logical/logical-box-border-width.html
@@ -1,0 +1,18 @@
+<!DOCTYPE html>
+<meta charset="utf-8" />
+<title>CSS Logical Properties: Flow-Relative Border Widths</title>
+<link rel="author" title="Oriol Brufau" href="mailto:obrufau@igalia.com" />
+<link rel="help" href="https://drafts.csswg.org/css-logical/#border-width">
+<meta name="assert" content="This test checks the interaction of the flow-relative border-*-width longhand properties with the physical ones in different writing modes." />
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<div id="log"></div>
+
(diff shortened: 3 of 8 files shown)


css/css-logical/logical-longhand-properties.html
```diff
@@ -0,0 +1,262 @@
+<!DOCTYPE html>
+<html>
+<meta charset="utf-8" />
+<title>Test for handling of logical and physical properties</title>
```
mrego: Nit: Other tests in this suite have "CSS Logical Properties:" in the title. You might want to add it.
This test is about "4. Flow-Relative Box Model Properties" so I'd mention that in the tittle too.
What about something like: "CSS Logical Properties: Flow-Relative Box Model Properties".

css/css-logical/logical-longhand-properties.html
```diff
@@ -0,0 +1,262 @@
...
+<link rel="author" title="Cameron McCormack" href="mailto:cam@mcc.id.au">
+<link rel="author" title="Oriol Brufau" href="mailto:obrufau@igalia.com">
+<link rel="help" href="https://drafts.csswg.org/css-logical-1/#box">
```
mrego: It's always nice to add a `<meta name="assert"` explaining what the test is specifically testing. Could you add it?

css/css-logical/logical-longhand-properties.html
```diff
@@ -0,0 +1,262 @@
+<!DOCTYPE html>
+<html>
```
mrego: Nit: you don't need the `<html>` tag.

css/css-logical/logical-longhand-properties.html
```diff
@@ -0,0 +1,262 @@
...
+//     blockStart: "margin-block-start", blockEnd: "margin-block-end",
+//   }, type: "length", prerequisites: "..." }
+//
+// where the `type` is a key from the `testValues` object and the `prerequisites`
+// is a declaration including prerequisites for all four physical properties.
+function createBoxPropertyGroup(property, descriptor) {
+  const logical = {};
+  const physical = {};
+  for (const logicalSide of ["inline-start", "inline-end", "block-start", "block-end"]) {
+    const camelCase = logicalSide.replace(/-(.)/g, (match, $1) => $1.toUpperCase());
```
mrego: Why you need this line? Wouldn't it be the same just using `logicalSide` later instead of `camelCase`.

css/css-logical/logical-longhand-properties.html
```diff
@@ -0,0 +1,262 @@
...
+<script>
+"use strict";
+const sheet = document.getElementById("sheet");
+const testElement = document.getElementById("test");
+
+// Logical properties with their descriptors.
+const logicalProperties = {
+  "border-*-color": {
+    type: "color",
+    prerequisites: {"color": "black"},
```
mrego: What are these prerequistes?

mrego (commented):
Thanks for doing this work, the test is quite complex.
I added a few comments in the review.

css/css-logical/logical-longhand-properties.html
Loirooriol: In this case I think it's because `border-*-color` defaults to `currentColor`, and the initial `color` value is UA-defined. So setting `color: black` you ensure that `border-*-color` won't be UA-defined. But here all tests that check `border-*-color` assign a value beforehand, so I can remove it.

In other cases it's needed in order to obtain a meaningful computed value, e.g. you can set `border-*-width` but if `border-*-style` is `none` then the former will compute to 0.

css/css-logical/logical-longhand-properties.html
Loirooriol: Logical properties are converted into physical using the properties from the `writingModes` items. If I use `logicalSide` here, then `writingModes` should have properties like `"block-start"` instead of `blockStart`. The lines will be longer due to the quotes and hyphen, but OK.

css/css-logical/logical-longhand-properties.html
Loirooriol: I usually add a root element for XML compatibility, but OK.

css/css-logical/logical-box-model-properties-001.html
```diff
@@ -0,0 +1,266 @@
...
+    for (const [i, logicalProp] of logicals.entries()) {
+      decl += `${logicalProp}: ${values[i]}; `;
+      expected.push([logicalProp, values[i]]);
+      expected.push([logicalToPhysical[logicalProp], values[i]]);
+    }
+    testComputedValues("logical properties on one declaration, writing " +
+                       `mode properties on another, '${writingModeDecl}'`,
+                       `.test { ${writingModeDecl} } .test { ${decl} }`,
+                       expected);
+  }
```
mrego: Why do you need these `{...}`?

mrego (approved):
The test LGTM, it's quite complex but it also checks a lot of stuff so it seems really useful.
Should we try to use more `test()` methods? Otherwise the test fails completely on Firefox, which is kind of strange... I understand they can have some bugs and some things could be failing, but most of the checks should be working on Firefox.
Maybe this grows too much I don't know and we should try to find a balance.

If the number of `test()` calls is too big, another idea could be to refactor the JavaScript part in a different file, and create 7 tests or so, one test for set of properties: `block|inline-size`, `border-*-color`, `border-*-style`, `border-*width`, `margin-*`, `padding-*`, `inset-*`.
So the number of `test()` calls in each of them is not huge.

WDYT?

mrego (approved):
This is amazing, thanks for doing it!

PR #11329, opened 2018-06-04
Feature Policy: fix tests that time out

zcorpan:
Part of #11269.

### DIFF
--- a/feature-policy/resources/picture-in-picture.js
+++ b/feature-policy/resources/picture-in-picture.js
@@ -3,14 +3,18 @@ function isPictureInPictureAllowed() {
     let video = document.createElement('video');
     video.src = '/media/movie_5.ogv';
     video.onloadedmetadata = () => {
-      video.requestPictureInPicture()
-      .then(() => resolve(document.pictureInPictureEnabled))
-      .catch(e => {
-        if (e.name == 'NotAllowedError')
-          resolve(document.pictureInPictureEnabled);
-        else
-          resolve(false);
-      });
+      try {
+        video.requestPictureInPicture()
+        .then(() => resolve(document.pictureInPictureEnabled))
+        .catch(e => {
+          if (e.name == 'NotAllowedError')
+            resolve(document.pictureInPictureEnabled);
+          else
+            resolve(false);
+        });
+      } catch(e) {
+        resolve(false)
+      }
     };
   });
 }
--- a/feature-policy/resources/picture-in-picture.js
+++ b/feature-policy/resources/picture-in-picture.js
@@ -3,18 +3,17 @@ function isPictureInPictureAllowed() {
     let video = document.createElement('video');
     video.src = '/media/movie_5.ogv';
     video.onloadedmetadata = () => {
-      try {
-        video.requestPictureInPicture()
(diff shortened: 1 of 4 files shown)


feature-policy/resources/picture-in-picture.js
```diff
@@ -3,14 +3,18 @@ function isPictureInPictureAllowed() {
...
+      try {
+        video.requestPictureInPicture()
+        .then(() => resolve(document.pictureInPictureEnabled))
+        .catch(e => {
+          if (e.name == 'NotAllowedError')
+            resolve(document.pictureInPictureEnabled);
+          else
+            resolve(false);
+        });
+      } catch(e) {
```
Hexcles: I don't know much about feature policy or picture in picture, but this broad exception catch smells a bit. Are you trying to catch exceptions from `video.requestPictureInPicture` (e.g. undefined)?

Hexcles (approved):
This generally LGTM, but I'd like confirmation from @clelland .

feature-policy/resources/picture-in-picture.js
zcorpan: Yeah. I suppose I can do `t.step_func` instead.

zcorpan: The function is sometimes called when there's no Test object. But I've changed to an `if` check instead.

beaufortfrancois:
Thanks for this patch.
Let me have a look.

beaufortfrancois:
I've uploaded a patch at https://chromium-review.googlesource.com/c/chromium/src/+/1107626 that fixes the issue. It will be upstreamed soon.

Hexcles:
We can either land that CL and rebase-merge this PR, or land this PR and abandon that CL. Either is fine to me.

beaufortfrancois:
I think https://chromium-review.googlesource.com/c/chromium/src/+/1107626 addresses all time out issues. Please let me know why you're changing some `async_test` to `promise_test`.

zcorpan:
I figured that since the tests were using promises they ought to use promise_test, but maybe it works OK either way.

beaufortfrancois:
I'm not sure to understand why but it looks like one test times out in https://travis-ci.org/web-platform-tests/wpt/jobs/396384904 while it doesn't for me locally. Do you folks have any idea why this may happening?

```js

promise_test(async t => {
  const video1 = await loadVideo();
  const video2 = await loadVideo();
  return callWithTrustedClick(() => {
    const first = video1.requestPictureInPicture();
    const second = video2.requestPictureInPicture();
    return Promise.all([
      first,
      promise_rejects(t, 'NotAllowedError', second)
    ]);
  });
}, 'request Picture-in-Picture consumes user gesture');
```

and we have now both promises which always reject... 

```js
if (!('pictureInPictureEnabled' in document)) {
  HTMLVideoElement.prototype.requestPictureInPicture = function() {
    return Promise.reject('Picture-in-Picture API is not available');
  }
}
```

Hexcles:
#11589 has landed and addressed all timeout on both Firefox and Chrome. This is no longer necessary, but feel free to reopen if I missed anything.

PR #11333, opened 2018-06-04
Update the IDL generated content headers to the content from reffy-reports

lukebjerring:
In preparation for moving to auto-updating using the tidoust/reffy-reports output, these are changed files, where the only differences were the header and lack of a trailing newline.


foolip (approved):

foolip:
The flakiness in https://travis-ci.org/web-platform-tests/wpt/jobs/388947894 is a bit surprising, and at any rate hardly because of the changes here. @lukebjerring I'll restart that job, but if it's still failing let me know and maybe I'll force merge this PR.

lukebjerring:
> Log reached capacity (3145728 bytes); output disabled.

Then dies 10m later because it hasn't received any output... lol! You'll need to force-merge.

foolip:
Ah, so this is another case of https://github.com/web-platform-tests/wpt/issues/11454.

PR #11334, opened 2018-06-04
[wptrunner] End Sauce Connect process gracefully

jugglinmike:
Previously, the Sauce Connect process was closed gracefully in response
to prolonged period of inactivity. This procedure was not used under
normal operating conditions.

Factor graceful shutdown logic into a new method, `quit`, and reuse this
method to end the process following successful test completion.

This is intended to resolve gh-10724


gsnedders (approved):

jugglinmike:
Thanks, @gsnedders!

PR #11339, opened 2018-06-05
HTML: Change <td hidden> and <th hidden> to expect display: none

zcorpan:
Follows https://github.com/whatwg/html/pull/3736.

### DIFF
--- a/html/rendering/non-replaced-elements/tables/hidden-attr.html
+++ b/html/rendering/non-replaced-elements/tables/hidden-attr.html
@@ -23,15 +23,16 @@ const expectedDisplay = {
   'tbody': 'table-row-group',
   'tfoot': 'table-footer-group',
   'tr': 'table-row',
-  'td': 'table-cell',
-  'th': 'table-cell',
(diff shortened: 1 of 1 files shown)


annevk:
Why would we only change these? Browsers fail for the other table elements too.

zcorpan:
Because `visibility: collapse` "works" on the other table elements, but acts like `visibility: hidden` on table cells.

annevk:
Do we expect implementations to align for those though?

zcorpan:
Quoting @FremyCompany in

> I understand the value of using visibility:collapse for table rows and table columns, as visibility:collapse is some sort of smart display:none, and I think a breaking change there is acceptable for these.

https://github.com/whatwg/html/issues/3728#issuecomment-394020407

https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/17718818/ is assigned; I hope the rest will follow.

annevk (approved):
Since you filed bugs, I guess we can see how it goes.

PR #11342, opened 2018-06-05
HTML: fix joint session history test to not time out

zcorpan:
Part of #11269.

---

The `t2` test in 002.html fails in Firefox and Chrome, and still times out in Safari. I'm not sure what the spec-correct behavior is for this test.

### DIFF
--- a/html/browsers/history/the-history-interface/joint_session_history/001.html
+++ b/html/browsers/history/the-history-interface/joint_session_history/001.html
@@ -4,7 +4,6 @@
 <script src="/resources/testharnessreport.js"></script>
 <div id="log"></div>
 <script>
-setup({timeout:10000});
 var t = async_test("Session history length on initial load");
 var t1 = async_test("Session history length on adding new iframe");
 var t2 = async_test("Navigating second iframe");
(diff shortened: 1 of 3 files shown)


foolip:
> I'm not sure what the spec-correct behavior is for this test.

Can you run Edge on the updated tests as well? If in the end Safari is the only timeout, then at least a spec issue for that would be good.

zcorpan:
w3c-test:mirror

foolip:
I've now tested in Edge as well, and the last test of 002.html fails in Chrome, Edge and Firefox with "assert_equals: expected 2 but got 1". It times out in Safari.

stephenmcgruer:
So this appears to take 002.html:'Navigating second iframe' from TIMEOUT to FAIL in Chrome and Firefox and change nothing for Safari.

For 001.html, nothing changes in Chrome (still 8/8), Firefox (still 5/8, with a TIMEOUT), or Safari (0/8, with a TIMEOUT).

This seems to align with the discussion in https://github.com/web-platform-tests/wpt/pull/11342#issuecomment-425744038, I'm just not sure of the utility of this PR if we don't really understand what is going on, what it is trying to test, and what the correct behavior is. Perhaps there is an html/ editor that can help there, but otherwise I might suggest just closing this. (My concern with just landing it is that the changes here may be invalidating the test and I don't know).

PR #11344, opened 2018-06-05
[wptserve] Put `// META:` scripts after testharness.js

zcorpan:
Some tests, e.g., [1], expect testharness.js to exist in the script.

Part of #11269.

[1] html/infrastructure/safe-passing-of-structured-data/shared-array-buffers/serialization-via-idb.any.html


zcorpan:
@gsnedders I have no idea what I'm doing with the tests, and I think they're failing but don't know why.

zcorpan:
OK, PTAL

Ms2ger (approved):
lgtm

jugglinmike:
I'm wary of assertions and tests declared within conditionals. A small change that would limit the risk of a spurious PASS (where the tests simply aren't executed) is to also push values to the `scripts` array from within the branches. If you agree, you might also want to re-name the variable (since it would no longer be tracking only script execution).

For test authors, "meta" means "any data declared with the `// META` syntax".  After we apply this patch, the term will have a different meaning in the context of the server's internals: "any data declared with the `// META` syntax *except* for scripts". It's unfortunate that we're bifurcating the definition in this way, but I can't think of a good alternative. My only ideas are to overhaul the data structure for the "vivified" metadata (i.e. a dict instead of a list of strings) or to update internal references to `meta` to `meta_without_scripts`. The former is way more disruptive than is probably warranted, and the latter is just hideous.

Do you think that's a problem? Do you have any other ideas for how we could avoid it?

Also, we'll need to make sure the new test only runs in secure contexts in order for the Service Worker version to pass. Typically, we could do this by renaming the test (in this case, from `infrastructure/server/order-of-metas.any.js` to `infrastructure/server/order-of-metas.https.any.js`). Unfortunately, [a previously-reported bug in the WPT CLI](https://github.com/web-platform-tests/wpt/issues/11105) interferes with this. I believe this patch is blocked on that bug.

zcorpan:
The service worker test is available at `infrastructure/server/order-of-metas.https.any.serviceworker.html` when using just `.any.js`, currently.

I don't think the naming is a problem in practice.

I can make the change you suggested about changing `scripts` to `tests`.

zcorpan:
Well, actually, making it tests instead of scripts makes it a bit awkward since then I also need some conditional on what the expected array should be. So I think I prefer leaving it as-is.

jugglinmike:
> I don't think the naming is a problem in practice.

It depends on what you mean by "in practice." Our CI service uses the WPT CLI. It loads the test over HTTP, and that's why this patch is being rejected with a TIMEOUT status:

>     1:38.00 TEST_START: /infrastructure/server/order-of-metas.https.any.serviceworker.html
>     2:38.12 TEST_END: TIMEOUT, expected OK

zcorpan:
About naming, I was referring to `META`.

zcorpan:
Do you think this should be blocked on #11105 ?

jugglinmike:
Yes. Without a fix, this new test will always fail and make it more difficult to validate other contributions. An alternative would be to merge this change without the tests and to only introduce them once gh-11105 is resolved. That might be advisable if there is urgency to the correction here.

zcorpan:
I've removed the service worker test.

zcorpan:
The failing check is because of https://github.com/web-platform-tests/wpt/pull/11405 (I can rebase this when that is merged).

jugglinmike:
I've been reviewing gh-11405 so that we can unblock this patch, but I don't
think that change is necessary. We originally theorized that the problem
stemmed from non-determinacy in the execution order of the web workers. That
theory is incorrect.

In `master` today, the test named `worker-dedicated.html` passes consistently.
This has been verified by TravisCI in the time since the test has been present
on the `master` branch, and I've also verified that locally by running the test
600 times.

It's definitely true that the sequence of worker execution is unstable. If you
run the test in a browser manually, you can see the ordering of results shift
between subsequent page refreshes. We accounted for this in the design of the
test harness, though: [the actual results are sorted according to sub-test
name before they are compared with the expected
results](https://github.com/web-platform-tests/wpt/blob/de3ae39cb59880a8245431e7f09817a2a4dad1a3/resources/test/conftest.py#L144).

The test named `worker-dedicated.html` fails when this patch is applied, and it
fails consistently. This is because this patch changes the way sub-test names
are computed. The failing sub-test which was previously named "Untitled" is now
named "worker-test.js". This must be accounted for in the expected values by
making two changes: updating the `name` property of the affected sub-test and
repositioning that sub-test in the array of sub-tests accordingly. I think it's
that second part which confused us. We have to re-order the expected data, but
it's not because the execution order has changed. It's because the title has
changed.

@zcorpan I've pushed up a commit to implement that change. If it looks good to
you, I think we can close gh-11405 and land this directly.

jugglinmike:
This looks good to me.

@jgraham and @foolip This change has an observable effect on multi-global test, so it's possible that it introduces regressions. Would you like to run a trial through the test infrastructure for Chromium/Firefox before we land it?

foolip:
Do the test changes and python changes need to go together for this to not break? Those are unfortunately separate import processes in Blink and if so it's good to know in advance.

Ms2ger:
Tests and python changes should be able to land separately; if the test changes land first, I guess they'll fail consistently until the python changes land.

foolip:
I'd count that as being coupled. @Hexcles is monitoring https://foolip.github.io/ecosystem-infra-rotation/ now and tomorrow it'll be my turn. Knowing that this change is coming through the pipeline is useful.

How about we just don't run it through Chromium's CQ in advance, but instead just monitor to see how test expectations change. If tests other than the ones changed here are affected, I'll revert or fix.

zcorpan:
SGTM, feel free to merge.

PR #11345, opened 2018-06-05
Precompute reference pass conditions

### LINKED ISSUE #11285: Move reftest pass conditions into the manifest
At the moment, we provide the reftest digraph in the manifest, but don't explicitly list the pass condition there, and as such every runner has to implement it itself. I'm not entirely convinced that wptrunner currently does this correctly, and further detaching it from test execution is probably a good idea.

I suggest, instead, that we provide an array of possible pass conditions, each pass condition being made up by an array of arrays of files that must render identically (i.e., split the set of URLs into a [quotient set](https://en.wikipedia.org/wiki/Equivalence_class)).

e.g.:

```
[
    [
        ["a.html", "b.html"]
    ]
]
```

This represents a single pass condition, where that pass condition is "a.html" and "b.html" must render identically.

```
[
    [
        ["a.html", "b.html"],
        ["c.html"]
    ]
]
```

This represents a single pass condition, where "a.html" and "b.html" must render identically, and "c.html" must render differently.

```
[
    [
        ["a.html", "b.html"],
        ["c.html"]
    ],
    [
        ["a.html", "b.html", "d.html"],
    ]
]
```

This represents two possible pass conditions: either a.html and b.html must render identically with c.html rendering differently *or* a.html, b.html, and d.html must render identically.

cc/ @jgraham @Hexcles @thejohnjansen

gsnedders:
Fixes #11285.

This is a bit rough around the edges, but does work.

- [ ] tests for pass_conditions
- [ ] documentation of this

### DIFF
--- a/tools/manifest/disjointset.py
+++ b/tools/manifest/disjointset.py
@@ -1,0 +1,173 @@
+"""
+A disjoint set implementation
+"""
+
+import array
+import ctypes
+
+try:
+    # needed for mypy
+    from typing import List, Dict, Iterable, Iterator, Optional, Set  # noqa: F401
+except ImportError:
+    pass
+
+# figure out what array size we're going to use
+_size_t_size = ctypes.sizeof(ctypes.c_size_t)
+
+for _type, _typecode in (
+    (ctypes.c_ubyte, "B"),
+    (ctypes.c_ushort, "H"),
+    (ctypes.c_uint, "I"),
+    (ctypes.c_ulong, "L"),
+    (ctypes.c_ulonglong, "Q")
+):
+    if ctypes.sizeof(_type) == _size_t_size:
+        _array_typecode = _typecode
+        break
+else:
+    # unlikely, but default to Q
+    assert False, "some weird platform"
+    _array_typecode = "Q"
+
+# Py < 3.3 don't support Q, despite it being what we want on most 64-bit systems
+if _array_typecode == "Q":
+    try:
+        array.array("Q")
+    except ValueError:
+        _array_typecode = "L"
+
+
+class DisjointSet(object):
+    """A disjoint set implementation
+
+    Requires members to be hashable objects.
+
+    This implements a disjoint set forest with the union by rank and path
+    compression heuristics.
+
+    :param iterable: initialize disjoint set from iterable's items
+    """
+
+    def __init__(self, iterable=None):
+        # type: (Optional[Iterable]) -> None
+
+        # the elements
+        self._elements = []  # type: List[object]
+
+        # dict mapping element -> index in elements
+        self._idx = {}  # type: Dict[object, int]
+
+        # parent of _elements[i] is stored in _parent[i]
+        self._parent = array.array(_array_typecode)
+
+        # size of the component - correct only for roots
+        self._rank = array.array(_array_typecode)
+
+        if iterable is None:
+            iterable = []
+
+        for el in iterable:
+            self.add(el)
+
+    def __repr__(self):
+        # type: () -> str
+        return '<DisjointSet: %r>' % self.get_disjoint_subsets()
+
+    def __len__(self):
+        # type: () -> int
+        return len(self._elements)
+
+    def __contains__(self, x):
+        # type: (object) -> bool
+        return x in self._idx
+
+    def __iter__(self):
+        # type: () -> Iterator
+        return iter(self._elements)
+
+    def add(self, x):
+        # type: (object) -> None
+        """Add an item to the disjoint set in a new singleton subset
+
+        This is commonly known as MakeSet in the literature. If the item
+        already exists in the disjoint set, this is a no-op.
+        """
+
+        if x in self:
+            return
+
+        new_idx = len(self)
+        self._elements.append(x)
+        assert self._elements[new_idx] is x
+        self._idx[x] = new_idx
+        self._parent.append(new_idx)
+        self._rank.append(0)
+
+    def find(self, x):
+        # type: (object) -> object
+        """Find the representative member of the subset x is a member of
+
+        Raises ValueError is x is not in the set.
+        """
+
+        if x not in self._idx:
+            raise KeyError('%r is not a member' % x)
+
+        # path compression
+        root = self._idx[x]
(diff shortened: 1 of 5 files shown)


gsnedders:
@jgraham review ping

jgraham (changes requested):
> Reviewed 4 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/11345)* status: 4 of 5 files reviewed, 8 unresolved discussions (waiting on @gsnedders and @jgraham)

___
*[tools/manifest/disjointset.py, line 103 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11345#-LJnSlsN8TmNgAZW0K-u:-LJnSlsN8TmNgAZW0K-v:bm51hhy) ([raw file](https://github.com/web-platform-tests/wpt/blob/b106d8651bd500abece5dd5350d105adbde7d412/tools/manifest/disjointset.py#L103)):*
> ```Python
>         assert self._elements[new_idx] is x
>         self._idx[x] = new_idx
>         self._parent.append(new_idx)
> ```

This can overflow, right? Particularly if we happen to have picked a small array type above.
___
*[tools/manifest/disjointset.py, line 155 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11345#-LJnSzYe97S7x66XleBL:-LJnSzYe97S7x66XleBM:bm9unak) ([raw file](https://github.com/web-platform-tests/wpt/blob/b106d8651bd500abece5dd5350d105adbde7d412/tools/manifest/disjointset.py#L155)):*
> ```Python
>             self._parent[y_idx] = x_idx
>             if x_rank == y_rank:
>                 self._rank[x_idx] = x_rank + 1
> ```

(this could also overflow)
___
*[tools/manifest/item.py, line 202 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11345#-LJnOAxkCnMVgx4SqJ8i:-LJnOAxkCnMVgx4SqJ8j:b-svw0m0) ([raw file](https://github.com/web-platform-tests/wpt/blob/b106d8651bd500abece5dd5350d105adbde7d412/tools/manifest/item.py#L202)):*
> ```Python
> 
>     def to_json(self):
>         rv = [self.url, self.references, {}]
> ```

Could call the superclass method?
___
*[tools/manifest/manifest.py, line 167 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11345#-LJnMxD-AFFuCdQiZ7SQ:-LJnMxD-AFFuCdQiZ7SR:babbq8h) ([raw file](https://github.com/web-platform-tests/wpt/blob/b106d8651bd500abece5dd5350d105adbde7d412/tools/manifest/manifest.py#L167)):*
> ```Python
>                 pass_conditions = []
>                 for pass_path in self._get_reftest_pass_paths(reftest):
>                     pass_condition = set()
> ```

This implies we aren't operating in a depth-first order, which seems like it's pretty confusing. In particular randomising the order means that one could see different failures each time that the test runs, which seems like a blocker.
___
*[tools/manifest/manifest.py, line 177 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11345#-LJnM0ZRDvzSsbhoPRNm:-LJnM0ZSDBhegNBoJvKF:b-36pys0) ([raw file](https://github.com/web-platform-tests/wpt/blob/b106d8651bd500abece5dd5350d105adbde7d412/tools/manifest/manifest.py#L177)):*
> ```Python
>                             ref_eq_set.union(url, ref_url)
> 
>                     # iter 2: check all != are sat and add them
> ```

I'm not sure if "sat" is widely used as a term of art, but I assume it's short for "satisfied". Let's make things clearer by not using unnecessary abbreviations.
___
*[tools/manifest/manifest.py, line 185 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11345#-LJnMFEe4Ni0gjCv6lWt:-LJnMFEe4Ni0gjCv6lWu:b80hvdf) ([raw file](https://github.com/web-platform-tests/wpt/blob/b106d8651bd500abece5dd5350d105adbde7d412/tools/manifest/manifest.py#L185)):*
> ```Python
>                         ref_url = ref_eq_set.find(ref_url)
>                         if url == ref_url:
>                             is_sat = False
> ```

Doesn't this mean we have an `==` and `!=` condition for the same pair of URLs? Isn't that an error?
___
*[tools/manifest/manifest.py, line 214 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11345#-LJnHlRv87duoDDvqyNZ:-LJnHlRw7w6-VU--Wg2T:b4socdp) ([raw file](https://github.com/web-platform-tests/wpt/blob/b106d8651bd500abece5dd5350d105adbde7d412/tools/manifest/manifest.py#L214)):*
> ```Python
>             for ref_url, ref_type in node.references:
>                 path[node.url] = (ref_type, ref_url)
>                 if ref_url in path or ref_url not in self.reftest_nodes_by_url:
> ```

This could perhaps do with a comment about why this is the right condition.
___
*[tools/manifest/manifest.py, line 218 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11345#-LJnHy1a5Ao33gMkny9C:-LJnHy1bEqle1tfW0rcl:b-55vrfk) ([raw file](https://github.com/web-platform-tests/wpt/blob/b106d8651bd500abece5dd5350d105adbde7d412/tools/manifest/manifest.py#L218)):*
> ```Python
>                 else:
>                     next_node = self.reftest_nodes_by_url[ref_url]
>                     stack.append((next_node, path.copy()))
> ```

So we could avoid copying here by treating the stack as a linked-list of path dicts, right? I guess that wouldn't actually perform better though, but I'm not sure (probably doesn't make much difference either way? idk).

gsnedders:
Currently abandoned pending resolving https://github.com/web-platform-tests/rfcs/pull/15.

gsnedders:
> Currently abandoned pending resolving [web-platform-tests/rfcs#15](https://github.com/web-platform-tests/rfcs/pull/15).

Following this, it's much simpler to figure out the pass condition (all matches must pass, none of the mismatches can match), thus this probably isn't worthwhile pursuing?

PR #11351, opened 2018-06-05
Make screen-orientation tests fail early if orientation is unsupported

zcorpan:
Part of #11269.

---

The tests still time out if orientation is supported but the event is not fired or so; not sure if anything can be done about that, assuming the tests are correct.

### DIFF
--- a/screen-orientation/onchange-event-subframe.html
+++ b/screen-orientation/onchange-event-subframe.html
@@ -4,11 +4,11 @@
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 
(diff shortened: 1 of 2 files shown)


marcoscaceres (approved):
Looks ok... if it's still being problematic, we can try some refactoring.

zcorpan:
@kaixinjxq does #10701 also happen to address the same issue we're trying to fix here?

kaixinjxq:
@zcorpan, not exactly, but I just added a new commit to fix the timeout issue if `screen.orientation.lock` is not supported.

zcorpan:
OK let's close this in favor of your PR :)

Issue #11343, opened 2018-06-05
POST to an .html resource results in a 404 response in wptserve


zcorpan:
This breaks this test

http://web-platform.test:8000/html/browsers/history/the-location-interface/reload_post_1.html

It seems a bit silly to create a .py resource to handle POST if a test doesn't need to.

@jgraham @gsnedders

zcorpan:
If we decide the current behavior is OK and we should fix the test, that would also resolve this issue.

foolip:
What does apache or nginx do with the same kind of request? Matching that would make sense, fewer surprises.

zcorpan:
In apache at least it just works.

foolip:
What's the response code? 200?

zcorpan:
I think so, not sure where to check right now though.

PR #11354, opened 2018-06-05
 #11343 - Redirect POST requests using .py resource

dheiberg:
Created a simple .py resource that redirects the POST request to `reload_post_1-1.html`. URL is hardcoded, it would be possible to make it a more general resource if necessary?


dhdavvie:
#11343

zcorpan:
I don't think it should redirect, instead it should return a normal response body, and the other helper file can be removed.

dhdavvie:
I've changed it so that the py file does not redirect, but instead returns a response body as suggested. The subsequent reload requires user approval, so I have flagged the test as `manual`.

zcorpan:
Looks good. Can you also remove `reload_post_1-1.html`?

zcorpan (approved):

zcorpan:
Let's leave #11343 open since we may want to change the behavior of the server.

PR #11355, opened 2018-06-05
Update hashchange/popstate event tests

foolip:
Spec: https://html.spec.whatwg.org/#history-traversal

### DIFF
--- a/html/browsers/browsing-the-web/history-traversal/hashchange_event.html
+++ b/html/browsers/browsing-the-web/history-traversal/hashchange_event.html
@@ -22,7 +22,7 @@ window.onload = t.step_func(function () {
     assert_equals(e.target, window, "target");
     assert_equals(e.type, "hashchange", "type");
     assert_true(e instanceof HashChangeEvent, "is HashChangeEvent");
-    assert_true(e.bubbles, "bubbles");
+    assert_false(e.bubbles, "bubbles");
     assert_false(e.cancelable, "cancelable");
     oldURLs.push(e.oldURL);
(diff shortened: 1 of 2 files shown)


foolip:
w3c-test:mirror

foolip:
@sideshowbarker, any idea why this isn't showing up in http://w3c-test.org/submissions/?

foolip:
w3c-test:mirror

foolip:
FWIW, the https://w3c-test.org/sync.py webhook has a red warning triange in the repo admin view. Related?

foolip:
Conclusion from running these tests in Chrome, Edge, Firefox and Safari:
* Neither event has `bubbles` set to true in Chrome, Edge or Safari, but `bubbles` is true for both events (matching https://html.spec.whatwg.org/#history-traversal) in Firefox.
* For all browsers, `cancelable` is false for hashchange, which makes sense and matches the spec.
* For Chrome and Firefox, `cancelable` is true for popstate, which doesn't appear to be per spec.

foolip:
OK, now ready to review together with spec change: https://github.com/whatwg/html/pull/3737

annevk (approved):

foolip:
@cdumez @youennf FYI, I'm fixing the cancelable popstate event in https://chromium-review.googlesource.com/c/chromium/src/+/1087047, which will leave WebKit as the only engine where that event is cancelable. Would you like me to file a bug, or can either of you quickfix it?

cdumez:
@foolip I will try and quick fix it today. Thanks for the heads up.

PR #11364, opened 2018-06-05
[testharness.js] Implement `single_test()` opt-in to single page test

### LINKED ISSUE #10853: Disambiguating uncaught errors
Single page tests are convenient because they shorten test code by two lines (i.e. `test(function(){` and `});`). However, they also introduce an ambiguity which makes differentiating harness errors and harness errors difficult (or maybe impossible).

The ambiguity is apparent any time a test produces an uncaught error without defining any tests. This may represent one of two states:

- a valid "single page test" that has failed
- a test that has been invalidated by exceptional behavior outside of a sub-test

At this moment, it is not possible for testharness.js to differentiate between these two states. Today, it conservatively assumes the former. This is a problem because it means WPT cannot detect (or reject) tests that are invalid in this way. gh-10706 is the most recent example we have identified.

As far as I know, "single page tests" are only offered as a convenience to test authors. The ambiguity described above trumps the convenience, so if the motivation really is so simple, then we should remove the feature.

There may be cases where the semantics under test necessitate that functionality. In that case, we could preserve the feature and eliminate the ambiguity by requiring an explicit "opt-in" pattern. This might take the form of metadata in the test code (e.g. `// META: single=true`) or a filename-based feature pattern (e.g. `.single.html`).

In either case, all "single page tests" defined in WPT today would need to be updated. [My attempts to estimate the number of single-page tests have been woefully misguided.](https://w3.logbot.info/testing/20180504#c403285-c403319) I think I'll need to instrument WPT and do a complete run to get an accurate number. Before investing that time, I'd like to get feedback from the folks who would know better than me.

@jgraham @gsnedders @zcorpan @foolip What do you think about eliminating single page tests? Or requiring explicit opt-in?

zcorpan:
Fixes #10853. Closes #11024.

### DIFF
--- a/FileAPI/url/multi-global-origin-serialization.sub.html
+++ b/FileAPI/url/multi-global-origin-serialization.sub.html
@@ -13,6 +13,7 @@
 
 <script>
 "use strict";
+single();
 document.domain = "{{host}}";
 
 window.onload = () => {
--- a/FileAPI/url/multi-global-origin-serialization.sub.html
+++ b/FileAPI/url/multi-global-origin-serialization.sub.html
@@ -13,7 +13,7 @@
 
 <script>
 "use strict";
-single();
+single_test();
 document.domain = "{{host}}";
 
 window.onload = () => {
--- a/css/css-scroll-snap/snap-at-user-scroll-end-manual.html
+++ b/css/css-scroll-snap/snap-at-user-scroll-end-manual.html
@@ -45,6 +45,7 @@ body {
 </div>
 
 <script>
+single();
 var body = document.body;
 var button = document.getElementById("btn");
 var target = document.getElementById("target");
--- a/css/css-scroll-snap/snap-at-user-scroll-end-manual.html
+++ b/css/css-scroll-snap/snap-at-user-scroll-end-manual.html
@@ -45,7 +45,7 @@ body {
 </div>
 
 <script>
-single();
+single_test();
 var body = document.body;
 var button = document.getElementById("btn");
 var target = document.getElementById("target");
--- a/docs/_writing-tests/testharness-api.md
+++ b/docs/_writing-tests/testharness-api.md
@@ -258,11 +258,9 @@ wrapping everything in functions for isolation becomes
 burdensome. For these cases `testharness.js` support "single page
 tests".
 
-In order for a test to be interpreted as a single page test, then
-it must simply not call `test()` or `async_test()` anywhere on the page, and
-must call the `done()` function to indicate that the test is complete. All
-the `assert_*` functions are available as normal, but are called without
-the normal step function wrapper. For example:
+To write a single page test, call `single_test()`. Call the `done()` function to
+indicate that the test is complete. All the `assert_*` functions are available
+as normal, but are called without the normal step function wrapper. For example:
 
 ```html
 <!doctype html>
@@ -271,12 +269,14 @@ the normal step function wrapper. For example:
 <script src="/resources/testharnessreport.js"></script>
 <body>
   <script>
+    single_test();
     assert_equals(document.body, document.getElementsByTagName("body")[0])
     done()
- </script>
+  </script>
 ```
 
-The test title for single page tests is always taken from `document.title`.
+The test title for single page tests is taken from `document.title` or from the
+argument to `single_test()`.
 
 ## Making assertions ##
 
--- a/dom/nodes/Element-getElementsByTagName-change-document-HTMLNess.html
+++ b/dom/nodes/Element-getElementsByTagName-change-document-HTMLNess.html
@@ -5,6 +5,7 @@
 <script src=/resources/testharnessreport.js></script>
 <iframe src="Element-getElementsByTagName-change-document-HTMLNess-iframe.xml"></iframe>
 <script>
+  single();
   onload = function() {
     var parent = document.createElement("div");
     var child1 = document.createElementNS("http://www.w3.org/1999/xhtml", "a");
--- a/dom/nodes/Element-getElementsByTagName-change-document-HTMLNess.html
+++ b/dom/nodes/Element-getElementsByTagName-change-document-HTMLNess.html
@@ -5,7 +5,7 @@
 <script src=/resources/testharnessreport.js></script>
 <iframe src="Element-getElementsByTagName-change-document-HTMLNess-iframe.xml"></iframe>
 <script>
-  single();
+  single_test();
   onload = function() {
     var parent = document.createElement("div");
     var child1 = document.createElementNS("http://www.w3.org/1999/xhtml", "a");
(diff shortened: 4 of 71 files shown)


zcorpan:
Command to find tests
```
find . -name "*.html" | xargs -I file grep -l "testharness.js" file | xargs -I file grep -l "assert_" file | xargs -I file grep -LE "test\\s*\\(" file | xargs -I file grep -L "generate_tests" file | xargs -I file grep -L "single()" file
```
This might not catch everything; running the tests and finding new `ERROR` statuses could reveal more.

Still to do: 
* add some more functional tests with the new behavior
* docs

jgraham:
I still don't think we should land this feature.

zcorpan:
I think we should. ^_^

zcorpan:
@foolip suggests naming it `single_test()` and ability to set title as argument

foolip:
Yep. I would further suggest that once existing tests have been fixed, calling `single_test()` is what causes `done()` and perhaps some other similar things on the global object, or perhaps that they throw an exception if invoked before `single_test()`.

annevk:
Given @jgraham's objection, perhaps we should reach out to those writing these kind of tests to find out whether this change would be overly burdensome?

foolip:
If we framed it as "you've written a single-page test, would you be burdened by a one-liner opt-in?" I think the answer will universally be "no." @jgraham, if that happens, would that settle the issue, or do you think that's the wrong question to ask, or the wrong people to ask?

zcorpan:
@foolip that's what this PR does (except calling `done()` is OK since that is allowed together with `test()`s).

@annevk, yeah, since you have written a few of the tests, what do you think?

annevk:
I'm fine with it. I didn't even know I wrote such tests.

jgraham:
I think asking people a direct question like that is a pretty terrible metric, because people generally aren't good at modelling the impact a change will have on their behvaiour, so the result you get will almost entirely depend on the framing of the question; people will pick an answer that makes them seem like a reasonable, rational person.

If you ask people about general pain points that occur when writing web-platform-tests and verbosity doesn't come up, then that would be a more compelling indication that we don't have to take steps to reduce test verbosity, and would suggest making different choices compared to the time 5 years ago when verbosity was the #1 compalint I was getting about writing testharness tests compared to mochitests.

jgraham:
(also in https://github.com/web-platform-tests/wpt/issues/10853#issuecomment-394885268 I note that there are two contradictory answers: adding boilerplate is fine, and also that the existing boilerplate is too burdensome. This reflects my belief that asking people about this kind of change directly is not the best way to understand what the actual impact will be)

foolip:
@jgraham, what *would* be a good way to understand what the impact will be, other than pitching our hunches against each other?

zcorpan:
> I'm fine with it. I didn't even know I wrote such tests.

I think this demonstrates a problem with the non-opt-in approach -- it's a mystery-meat feature where it's unclear even to @annevk (who has contributed a lot of tests) that he was using the feature in the first place.

jgraham:
I thought I made a counter-proposal: get feedback about pain points in general and see whether people still bring up verbosity. I don't think it's going to provide an unambiguous yes/no decision on whether this is the right thing to do, and in fact I don't think that such a thing is possible to achieve because the people experiencing the costs and and those getting the benefits are not the same, or at least not on the same time scale (*if* it made tests less buggy on average then maybe there would be some long term benefit to browser developers, but that's just as much an unproven hypothesis as the claim that the verbosity will cause frustration). However it would provide insights into the kind of tradeoffs that will benefit authors and those that might not.

jgraham:
> I think this demonstrates a problem with the non-opt-in approach -- it's a mystery-meat feature where it's unclear even to @annevk (who has contributed a lot of tests) that he was using the feature in the first place.

I don't really know how you would end up writing `done()` at the correct point in a test if you had no idea what you were doing. It does seem plausible that the feature just wasn't that compelling and so @annevk paged it out since writing whichever test it was.

zcorpan:
I've made https://goo.gl/forms/qLJFzy5O0KWKTY6E3 , feel free to spread it to where you think makes sense.

jgraham:
Thanks for doing that. However, I would expect if we are going to do a survey there would be more prior agreement on the questions and distribution method, just because getting people to fill out a lot of questions is a big ask and so something we can't do that often without causing significant fatigue. Therefore it's important to make sure we're making the best use of the oppertunity. For example some programming languages have annual developer surveys for this kind of feedback.

Having said all of that, I'm concerned that I'm being percieved as stop-energying this effort. But from my point of view I'm finding it hard to understand why making changes here is suddenly such a high priority. Is there some time-sensitive goal (or OKR or whatever) that I'm unaware of that's specifically coupled to a metric like the error rate in tests that's driving this work?

zcorpan:
I can still edit the questions, there haven't been any responses yet.

Fixing broken tests and improving tooling is my current focus, and although I'd like to move forward, I'm not in a rush.

jgraham:
I think we should withdraw the survey and get consensus on a set of questions, before sending one out in a more coordinated way in a week or so. I know that's annoying, but since every time we ask for widespread feedback it resets the timer on when we can next get the same kind of feedback, I think it's important to get it right upfront. I'm already slightly concerned about asking Mozilla developers for more input because I sent a survey about wpt integration in Gecko infrastructure in December to set priorities for the year.

zcorpan:
Ok, closed the survey for now.

zcorpan:
> @foolip suggests naming it single_test() and ability to set title as argument

I've implemented this suggestion now.

foolip:
I like it, but take it this is blocked on reaching an agreement. @zcorpan, is some updated survey the next step?

zcorpan:
Yeah, I suppose so.

zcorpan:
Survey sent out, http://lists.w3.org/Archives/Public/public-test-infra/2018AprJun/0015.html

jugglinmike:
Hi @zcorpan! I noticed you assigned this issue to me. I don't have the bandwidth for this right now, so I'm going to un-assign myself. That will make the pull request's status more clear.

zcorpan:
Okey dokey. @foolip, can someone take this? I'm happy to help analyze the survey with someone to inform what to do here.

marcoscaceres (approved):
Code looks ok to me. I don't have a strong opinion about the feature itself tho.

zcorpan:
RFC for this: https://github.com/web-platform-tests/rfcs/pull/28

This needs to be updated to match the RFC

foolip:
I did a tiny amount of work in https://github.com/web-platform-tests/wpt/pull/19202 which could perhaps be used when reviving this. FWIW, I'd recommend closing this and introducing it over many PRs, not least to avoid spamming all of the reviewers with the many rounds of review this will probably require.

foolip:
This is being done in https://github.com/web-platform-tests/wpt/pull/20036 now.

PR #11366, opened 2018-06-06
[css-text-3] Add test for intrinsic sizing and overflow-wrap

frivoal:
This test goes toghether with the spec changes for https://github.com/w3c/csswg-drafts/issues/2682

### DIFF
--- a/css/css-text/overflow-wrap/overflow-wrap-min-content-size-001.html
+++ b/css/css-text/overflow-wrap/overflow-wrap-min-content-size-001.html
@@ -1,0 +1,31 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Text Test: overflow-wrap: break-word and intrinsic sizing</title>
+<link rel="author" title="Florian Rivoal" href="http://florian.rivoal.net/">
+<link rel="help" href="https://drafts.csswg.org/css-text-3/#overflow-wrap-property">
+<meta name="flags" content="">
+<link rel="match" href="reference/overflow-wrap-min-content-size-001-ref.html">
+<meta name="assert" content="Soft wrap opportunities introduced by overflow-wrap:break-word **are** considered when calculating min-content intrinsic sizes.">
+<style>
+table {
+  overflow-wrap: break-word;
+  max-width: 0;
+  border: 0;
+  border-collapse: collapse;
+}
+td {
+  padding: 0;
+  background: green;
+  color: transparent;
+}
+#red {
+  position: absolute;
+  z-index: -1;
+  background: red;
+  color: transparent;
+}
+</style>
+
+<p>Test passes if there is a green box below and no red.
+<div id=red>X<br>X<br>X<br>X</div>
+<table><tr><td>XXXX</table>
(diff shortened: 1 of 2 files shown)


css/css-text/overflow-wrap/overflow-wrap-min-content-size-001.html
```diff
@@ -0,0 +1,31 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Text Test: overflow-wrap: break-word and intrinsic sizing</title>
+<link rel="author" title="Florian Rivoal" href="http://florian.rivoal.net/">
+<link rel="help" href="https://drafts.csswg.org/css-text-3/#overflow-wrap-property">
+<meta name="flags" content="">
```
javifernandez: What does it mean this empty flags tag ?

css/css-text/overflow-wrap/overflow-wrap-min-content-size-001.html
```diff
@@ -0,0 +1,31 @@
...
+<link rel="match" href="reference/overflow-wrap-min-content-size-001-ref.html">
+<meta name="assert" content="Soft wrap opportunities introduced by overflow-wrap:break-word **are** considered when calculating min-content intrinsic sizes.">
+<style>
+table {
```
javifernandez: I think the 'table' object here is a distraction. Perhaps i'm not understanding it properly, but at least in chrome, the max-width properly is not applied and 'table' layout just honors the min-content to determine its width. Wouldn't be better to use an inline-block, a float or just a div with 'min-content' width ?

css/css-text/overflow-wrap/overflow-wrap-min-content-size-001.html
```diff
@@ -0,0 +1,31 @@
...
+  overflow-wrap: break-word;
+  max-width: 0;
```
javifernandez: Is the max-width really relevant for this test ? If the purpose of the test is to evaluate how soft-breaking opportunities account for the min-content width, wouldn't be enough to use width: min-content ?

css/css-text/overflow-wrap/overflow-wrap-min-content-size-001.html
frivoal: it means that none of the possible flags listed in https://web-platform-tests.org/writing-tests/css-metadata.html apply.

Not including the flags meta at all might mean that too, or it might mean that the author forgot to consider the flags. So I prefer to include empty flags rather than leaving them out.

css/css-text/overflow-wrap/overflow-wrap-min-content-size-001.html
frivoal: This is a response to this comment as well as the next one:

Yes, with `width: min-content` on the div directly, without using a table, would achieve the goal of this test, and be simpler. However, `width: min-content` is not yet widely supported, and tests should not depend on features that are not yet stable.

Using a table, and forcing it to be as small as possible will achieve min-content sizing of the div in all browsers.

I'll add a comment to the test saying that.

javifernandez (approved):

Issue #11233, opened 2018-05-29
[Presentation API] Add test for starting presentations on two distinct displays


mfoltzgoogle:
This is to track adding a test case to the controlling-ua suite that alllows start() to succeed twice on the same controlling page, to start two distinct presentations on two presentation displays.

Note that implementation limitations in Chrome may prevent this from currently passing for all display types.

mfoltzgoogle:
CC @tomoyukilabs

tomoyukilabs:
@mfoltzgoogle Okay, I'll submit a PR to address this issue. Thanks!

mfoltzgoogle:
Looks good - I left a couple of minor comments on the tests. I will look forward to verifying them in Chrome.

tomoyukilabs:
Thanks, @mfoltzgoogle!

PR #11370, opened 2018-06-06
[presentation-api] Add a test for starting presentations on distinct displays

tomoyukilabs:
This PR addresses #11233 by adding a test for starting presentations on two distinct displays.

### DIFF
--- a/presentation-api/controlling-ua/reconnectToMultiplePresentations_success-manual.https.html
+++ b/presentation-api/controlling-ua/reconnectToMultiplePresentations_success-manual.https.html
@@ -1,0 +1,120 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Reconnecting presentations on two distinct displays</title>
+<link rel="author" title="Tomoyuki Shimizu" href="https://github.com/tomoyukilabs">
+<link rel="help" href="http://w3c.github.io/presentation-api/#dfn-controlling-user-agent">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="common.js"></script>
+<style>
+#second-step {
+  display: none;
+}
+</style>
+
+<p id="first-step">Click the button below and select the available presentation display, to start the manual test.</p>
+<p id="second-step">Click the button below and select the other available presentation display, to continue the manual test.</p>
+<p>This test asks you to click the button twice, unless the test fails.<br>
+<em>This test requires two or more available displays.<em></p>
+<button id="presentBtn">Start Presentation Test</button>
+
+
+<script>
+  promise_test(async t => {
+    const presentBtn = document.getElementById("presentBtn");
+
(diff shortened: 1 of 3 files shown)


tomoyukilabs:
ping @tidoust @louaybassbouss

tidoust (approved):
That looks good to me. The test goes beyond merely testing starting presentation on distinct displays (e.g. it also tests `reconnect`), but that seems ok and useful. We can split the test later on if that seems needed.

tomoyukilabs:
@tidoust I agree because the reconnecting test looks long enough. Now I've committed the change to split the test. PTAL.

tidoust (approved):
Looks good, thanks!

presentation-api/controlling-ua/reconnectToMultiplePresentations_success-manual.https.html
```diff
@@ -0,0 +1,120 @@
...
+    connection2.close();
+    await eventWatcher2.wait_for('close');
+    assert_equals(connection2.state, 'closed', 'The presentation connection is successfully closed.');
+
+    const c1 = await request1.reconnect(connection1.id);
+    assert_equals(c1, connection1, 'The promise is resolved with the existing presentation connection.');
+
+    // Reconnecting a presentation via a different presentation request with the same presentation
+    // URLs will succeed
+    const c2 = await request1.reconnect(connection2.id);
```
mfoltzgoogle: Note that it should be possible to call request2.reconnect here and above, and achieve the same result.

presentation-api/controlling-ua/reconnectToMultiplePresentations_success-manual.https.html
```diff
@@ -0,0 +1,120 @@
...
+    const presentBtn = document.getElementById("presentBtn");
+
+    const request1 = new PresentationRequest(presentationUrls);
+    const request2 = new PresentationRequest(presentationUrls);
+    const clickWatcher = new EventWatcher(t, presentBtn, 'click');
+    let connection1, connection2, eventWatcher1, eventWatcher2, timer;
+
+    t.add_cleanup(() => {
+      [
+        { connection: connection1, request: request1 },
```
mfoltzgoogle: It looks like the request values are not used in the .forEach below.

tomoyukilabs: The `request` value is used to reconnect a presentation in the `closed` state and then terminate it. Please see the lines 37-38:

```js
          if (p.connection.state == 'closed')
             p.request.reconnect(p.connection.id);
```

presentation-api/controlling-ua/reconnectToMultiplePresentations_success-manual.https.html
tomoyukilabs: I agree. I'll submit another PR to address this. Thanks!

PR #11371, opened 2018-06-06
[presentation-api] Modify the test to check NotFoundError on reconnecting

tomoyukilabs:
This PR modifies controlling-ua/reconnectToPresentation_notfound_error.https.html by:

- Adding a step to reconnect with a presentation ID on a presentation request with a different URL
- Changing the test to a manual test, because starting a presentation is needed to generate an actual presentation ID


tomoyukilabs:
ping @tidoust @louaybassbouss

tidoust (approved):
Thanks, that looks good.

One minor suggestion which you may ignore: I would rename `connection` into `connection1` to make the link with `request1` more explicit. In particular, this would make it more obvious why the first line below is meant to fail and why the second is meant to succeed:

```js
await promise_rejects(t, 'NotFoundError', request2.reconnect(connection1.id), '...');
await request1.reconnect(connection1.id);
```

tomoyukilabs:
@tidoust Okay, I've committed the change for renaming.

presentation-api/controlling-ua/reconnectToPresentation_notfound_error-manual.https.html
```diff
@@ -19,13 +19,13 @@
...
     const request2 = new PresentationRequest('https://www.w3.org');
-    let connection, eventWatcher;
+    let connection1, eventWatcher1;
 
     t.add_cleanup(() => {
-      if (connection) {
-        connection.onconnect = () => { connection.terminate(); }
+      if (connection1) {
+        connection1.onconnect = () => { connection1.terminate(); }
         if (connection.state === 'closed')
```
tidoust: s/connection/connection1

presentation-api/controlling-ua/reconnectToPresentation_notfound_error-manual.https.html
```diff
@@ -19,13 +19,13 @@
...
-          request1.reconnect(connection.id);
+          request1.reconnect(connection1.id);
         else
           connection.terminate();
```
tidoust: s/connection/connection1

tidoust (changes requested):
See inline

tomoyukilabs:
Sorry. I've fixed them.

tidoust (approved):

PR #11375, opened 2018-06-06
[css-text-3] Hanging punctuation is scrollable overflow

frivoal:
Relates to the spec changes decided in
https://github.com/w3c/csswg-drafts/issues/2398


foolip (approved):
I did light review.

PR #11382, opened 2018-06-06
Fix the TemporaryFile mode to prevent crashing.

dheiberg:
In `request.py`, if a remote file is bigger than the given `max_buffer_size`, it will use a [TemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile). The mode for this tempfile caused an error `IOError: [Errno 9] Bad file descriptor` to happen on line 84. Removing this mode and using the default `w+b` fixes this.

### DIFF
--- a/tools/wptserve/tests/functional/test_request.py
+++ b/tools/wptserve/tests/functional/test_request.py
@@ -58,15 +58,11 @@ class TestInputFile(TestUsingServer):
 
         try:
             resp = self.request(route[1], method="POST", body="1"*20)
-        except HTTPError:
-            # Cleanup and fail test
+            self.assertEqual(200, resp.getcode())
+            self.assertEqual(["11", "7", "0", "0"],
+                            resp.read().split(" "))
+        finally:
             InputFile.max_buffer_size = old_max_buf
-            self.assertTrue(False, 'This should not be reached. Server failed.')
-
-        self.assertEqual(200, resp.getcode())
-        self.assertEqual(["11", "7", "0", "0"],
-                         resp.read().split(" "))
-        InputFile.max_buffer_size = old_max_buf
 
     def test_iter(self):
         @wptserve.handlers.handler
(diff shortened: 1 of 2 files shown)


tools/wptserve/tests/functional/test_request.py
```diff
@@ -35,6 +37,37 @@ def handler(request, response):
...
+
+        route = ("POST", "/test/test_seek", handler)
+        self.server.router.register(*route)
+
+        try:
+            resp = self.request(route[1], method="POST", body="1"*20)
+        except HTTPError:
+            # Cleanup and fail test
+            InputFile.max_buffer_size = old_max_buf
+            self.assertTrue(False, 'This should not be reached. Server failed.')
```
jgraham: Can't this just be `raise`?

tools/wptserve/tests/functional/test_request.py
```diff
@@ -47,6 +80,29 @@ def handler(request, response):
...
+
+        route = ("POST", "/test/test_iter", handler)
+        self.server.router.register(*route)
+
+        try:
+            resp = self.request(route[1], method="POST", body="12345\nabcdef\r\nzyxwv")
+        except HTTPError:
+            # Cleanup and fail test
+            InputFile.max_buffer_size = old_max_buf
+            self.assertTrue(False, 'This should not be reached. Server failed.')
```
jgraham: Again, maybe just `raise`?

tools/wptserve/tests/functional/test_request.py
```diff
@@ -47,6 +80,29 @@ def handler(request, response):
...
+
+        self.assertEqual(200, resp.getcode())
+        self.assertEqual(["12345\n", "abcdef\r\n", "zyxwv"], resp.read().split(" "))
+        InputFile.max_buffer_size = old_max_buf
```
jgraham: Maybe rather than having the exception handler in the middle at all, wrap most of the test in a try/finally block for cleanup.

jgraham (changes requested):
This generally looks good, thanks. I think the error handling could use a little cleanup though.

tools/wptserve/tests/functional/test_request.py
dhdavvie: Agreed, that would be a cleaner way to do it. Will push that in a moment

tools/wptserve/tests/functional/test_request.py
```diff
@@ -58,15 +58,11 @@ def handler(request, response):
 
         try:
             resp = self.request(route[1], method="POST", body="1"*20)
-        except HTTPError:
-            # Cleanup and fail test
+            self.assertEqual(200, resp.getcode())
```
jgraham: Move the max_buffer_size change to right before the `try` so that execptions in other places don't result in it not being reset.

jgraham (approved):

PR #11396, opened 2018-06-06
[css-ui-4] link cursor over SVG links

frivoal:
Goes along with https://github.com/w3c/csswg-drafts/issues/2664

### DIFF
--- a/css/css-ui/cursor-pointer-links-001-manual.html
+++ b/css/css-ui/cursor-pointer-links-001-manual.html
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<html lang="en">
+<meta charset="utf-8">
+<title>CSS Basic User Interface Test: pointer cursor on svg links</title>
+<link rel="author" title="Florian Rivoal" href="https://florian.rivoal.net">
+<link rel="help" href="https://drafts.csswg.org/css-ui-4/#cursor">
+<meta name="flags" content="interact">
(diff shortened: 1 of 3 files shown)


AmeliaBR:
Looks good. But why not have HTML links (both a text link and an image map) for completeness' sake?

We'll probably want to have a separate test in the SVG directory that is a stand-alone .svg file once  w3c/svgwg#468 is resolved with normative text, but there's no harm in having an inline SVG test for CSS UI.

Also: Do you want to explicitly test the computed value has changed, rather than the change being part of `auto` flexibility? Or do we only care about the end result? Computed style would have the benefit of being a scriptable test, but should probably still be complemented by a manual test.

frivoal:
> But why not have HTML links (both a text link and an image map) for completeness' sake?

Fair enough. I'll be adding that.

> Do you want to explicitly test the computed value has changed, rather than the change being part of auto flexibility?

It is not part of auto flexibility. `auto` is strictly now allowed to do that, and there are tests for that (See cursor-auto-002.html).

We could also write tests to check that the computed value, not just the appearance, is the right one, but given the above tests, I don't find this especially pressing. I'd be happy to review such a test, but I don't think I'll rush into writing one.

frivoal:
> >  But why not have HTML links (both a text link and an image map) for completeness' sake?

> Fair enough. I'll be adding that.

Done.

AmeliaBR (approved):

svgeesus (approved):
LGTM

PR #11401, opened 2018-06-07
Consolidate label conversion tests + add timeLog

domfarolino:
This completes some `TODO`s for the existing console label conversion tests and also adds necessary label conversion tests for [`console.timeLog`](https://github.com/whatwg/console/pull/138). (This does not satisfy all of `timeLog`'s necessary requests, just the tests of this kind). I've also manually tested that this works etc.


console/console-label-conversion.any.js
```diff
@@ -0,0 +1,37 @@
...
+// https://console.spec.whatwg/org/#counting
+// https://console.spec.whatwg/org/#timing
+
+// toString() console label test strings
+const toStringAssertionEnding = 'must call toString() on label when label is an object';
+const toStringTestNameEnding = 'label gets converted to string via label.toString() when label is an object';
+
+// toString error re-throwing console test strings
+const toStringRethrowAssertionEnding = 'must re-throw any exceptions thrown by label.toString() conversion';
+const toStringRethrowTestNameEnding = 'throws exceptions generated by erroneous label.toString() conversion';
```
domenic: Not sure why these test strings are factored out instead of inlined; it makes things a big harder to read.

domenic (approved):
LGTM with nit

console/console-label-conversion.any.js
domfarolino: Ah I think before using the for loop I had factored them out to avoid duplication, but that's not an issue at all now.

PR #11403, opened 2018-06-07
[testharness.js] Get title from the filename if not given

zcorpan:


Ms2ger:
I'm a little confused about the `get_title()` function; each caller only uses one of the three separate branches, so why not inline the code?

zcorpan:
In the window case, there might not be a title. But can inline "Untitled" for the shell case.

Ms2ger (approved):
Fair point.

zcorpan:
The failing check is because of https://github.com/web-platform-tests/wpt/pull/11405

Ms2ger:
We can test the META feature, can't we?

zcorpan:
Yes, done.

zcorpan:
@Ms2ger OK to merge?

Ms2ger (approved):

PR #11404, opened 2018-06-07
Make set-animation-play-state-to-paused-002 more robust

fred-wang:
This is a follow-up of https://github.com/web-platform-tests/wpt/pull/11200, trying to make it more robust after review comments on another PR.

        - Increase the height of covering rect to avoid anti-aliasing issue with the red lines.
        - Instead of waiting a delay, use the shift returned by getComputedStyle() to know when animated lines are supposed to be hidden by the green rect.
        - Instead of waiting an additional amount of time to ensure red lines will show up again if they were not paused, just use the animation duration to decide when the screenshot should be taken.

### DIFF
--- a/css/css-animations/set-animation-play-state-to-paused-002-ref.html
+++ b/css/css-animations/set-animation-play-state-to-paused-002-ref.html
@@ -4,32 +4,33 @@
     <meta charset="utf-8">
     <title>Dynamically set animation-play-state to paused (reference)</title>
     <style>
-      #container {
+      #containerLinear, #containerSteps {
         position: absolute;
-        left: 0;
-        top: 3em;
+        left: 100px;
+      }
+      #containerLinear {
+        top: 100px;
+      }
+      #containerSteps {
+        top: 200px;
       }
       #coveringRectLinear, #coveringRectSteps {
         position: absolute;
         background: lightgreen;
-        width: 40px;
-        height: 70px;
-        left: 80px;
-        transform-origin: 50% 10%;
-        transform: translate(36px, 0) rotate(144deg);
-      }
-      #coveringRectLinear {
-        top: 50px;
-      }
-      #coveringRectSteps {
-        top: 150px;
+        left: -25px;
+        width: 50px;
+        top: -50px;
+        height: 100px;
+        transform: translate(72px, 0) rotate(144deg);
       }
     </style>
   </head>
   <body>
     <p>This test passes if you see two rotated green rectangles and no red.</p>
-    <div id="container">
+    <div id="containerLinear">
       <div id="coveringRectLinear"></div>
+    </div>
+    <div id="containerSteps">
       <div id="coveringRectSteps"></div>
     </div>
   </body>
--- a/css/css-animations/set-animation-play-state-to-paused-002.html
+++ b/css/css-animations/set-animation-play-state-to-paused-002.html
@@ -9,33 +9,32 @@
                                  when animation-play-state is set to paused">
     <link rel="match" href="set-animation-play-state-to-paused-002-ref.html">
     <style>
-      #container {
+      #containerLinear, #containerSteps {
         position: absolute;
-        left: 0;
-        top: 3em;
+        left: 100px;
+      }
+      #containerLinear {
+        top: 100px;
+      }
+      #containerSteps {
+        top: 200px;
       }
       #lineLinear, #lineSteps  {
         position: absolute;
         background: red;
-        width: 10px;
+        left: -1px;
+        width: 2px;
+        top: -25px;
         height: 50px;
-        left: 95px;
-        transform-origin: 50% 0%;
       }
       #coveringRectLinear, #coveringRectSteps {
         position: absolute;
         background: lightgreen;
-        width: 40px;
-        height: 70px;
-        left: 80px;
-        transform-origin: 50% 10%;
-        transform: translate(36px, 0) rotate(144deg);
-      }
-      #coveringRectLinear, #lineLinear {
-        top: 50px;
-      }
-      #coveringRectSteps, #lineSteps {
-        top: 150px;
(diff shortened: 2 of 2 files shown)


birtles:
For our animation tests that run on CI we normally assume there could be up to 100 seconds between animation frames.

fred-wang:
> For our animation tests that run on CI we normally assume there could be up to 100 seconds between animation frames.

@birtles 100 seconds seem huge, accepting such a margin of error would make the tests pass very easily and hence they would become irrelevant as non-regression tests for the two WebKit bugs mentioned.

With the current set of parameters, taking a horizontal shift error <= 9 is equivalent to accepting up to 100ms between requestAnimationFrame calls. Note that on https://treeherder.mozilla.org/#/jobs?repo=try&revision=07483dcde214241025fc1ad7f0622bdcd3aeea9f none of the failures are related to the red lines being visible but to some anti-aliasing artifact on Linux. See for example https://hg.mozilla.org/mozilla-central/raw-file/tip/layout/tools/reftest/reftest-analyzer.xhtml#logurl=https://queue.taskcluster.net/v1/task/Xi1U4-VcRJa1jm3GnJkYIg/runs/0/artifacts/public/logs/live_backing.log&only_show_unexpected=1

birtles:
> @birtles 100 seconds seem huge, accepting such a margin of error would make the tests pass very easily and hence they would become irrelevant as non-regression tests for the two WebKit bugs mentioned.

Yes, I realize that but we've often observed over 20 seconds between frames on machines in automation. These machines get really loaded down, especially the Android devices.

Can you just set up a long animation with a step timing function using a single step, make the time to the step the minimum time you expect 10 animation frames to take, wait a frame to let the animation run, call `getComputedStyle` and check you are before the step (if not, re-start the test, extending the duration of the animation), add the pause play-state, wait 10 frames, and then snapshot? If the pause worked you'll still be before the step, if not you'll be passed it?

Although we can have long frames, we don't expect to get 10 frames at 50ms each following by one at 50s so I think that should work?

fred-wang:
@birtles  Thanks for your help on this.

Some more context:

The goals of these tests were to verify the rendering of animations after a pause using either a Bézier or a Steps timing functions and for either a simple (-001) or complex (-002) transform.

As I said, these ideas of tests came from particular bugs of WebKit that are only visual (so can not be determined from the getComputedStyle only... although ideally we should also add non-visual css animation tests). Additionally, the first one happens only with the Bézier timing function while the second one only happens with the Steps timing function + a complex transform (e.g. translate + rotation). Since the time I initially wrote the tests, a new regression happened in WebKit for any animation that uses a Steps timing function ( https://bugs.webkit.org/show_bug.cgi?id=186129 ) ; this regression can be detected by any of these two tests.

So coming back to your question, I think your suggestion will work for WebKit and for the bug tested by -002 or the new regression. However, for the bug tested by -001 or for other web engines, it seems a bit artificial to ignore the case of Bézier timing function and to do all of this restart and wait just to adjust the case of Mozilla's CI. And I guess other browser vendors may have similar but different timing issues or regressions/bugs...

So at that point maybe it's not a good idea to write shared WPT tests, I should probably restrict them to WebKit for now if that's too complicated. It looks like the CSS animations tests in the WPT repo are very limited at the moment (most are manual). This discussion about avoiding flakiness and the recent regression in WebKit for steps() make me think internal browser vendors repo don't have good test coverage either for CSS animations. Probably the way automated tests currently check visual rendering is not very adapted to check animations and we would need new approaches :-(

Finally, that's probably not very useful now, but I had retriggered the Linux WPT reftests with transparent covering rect, so we can see the position of the red rect: https://treeherder.mozilla.org/#/jobs?repo=try&revision=ba4676293e635844b71cbde5b82eca4f7f939852&selectedJob=182464474

fred-wang:
Position of the red rects: https://hg.mozilla.org/mozilla-central/raw-file/tip/layout/tools/reftest/reftest-analyzer.xhtml#logurl=https://queue.taskcluster.net/v1/task/HsNyIIk_QhmUTYG2hFZeSw/runs/0/artifacts/public/logs/live_backing.log&only_show_unexpected=1

birtles:
> So at that point maybe it's not a good idea to write shared WPT tests, I should probably restrict them to WebKit for now if that's too complicated. It looks like the CSS animations tests in the WPT repo are very limited at the moment (most are manual). This discussion about avoiding flakiness and the recent regression in WebKit for steps() make me think internal browser vendors repo don't have good test coverage either for CSS animations. Probably the way automated tests currently check visual rendering is not very adapted to check animations and we would need new approaches :-(

I wonder if the case of testing bezier timing on the compositor is too WebKit-specific to bother with wpt. Would be nice if we could do it however. Can we use the bezier but apply it to a discretely-animated property so that it behaves like a step timing function in-effect? I guess none of the properties WebKit animates on the compositor are discretely animated though?

In Gecko we have a test-only function that fetches the animated transform/opacity being used on the compositor and we use that for testing. Obviously we can't land those tests in wpt, however.

fred-wang:
> I wonder if the case of testing bezier timing on the compositor is too WebKit-specific to bother with wpt. Would be nice if we could do it however.

Testing the visual rendering of Bézier or Steps timing functions make sense from the point of view of someone just reading the spec, so I won't say it is WebKit-specific. Also, the transform property seems to be an important use case.

> Can we use the bezier but apply it to a discretely-animated property so that it behaves like a step timing function in-effect? I guess none of the properties WebKit animates on the compositor are discretely animated though?

To be honest, I only checked with the transform property since that's the one reported by users. Someone from Apple mentioned the regression does not happen for non hardware-accelerated animations  ( https://bugs.webkit.org/show_bug.cgi?id=186129#c8 ). So I suspect restricting to specific classes of properties won't allow to cover all the cases (and ideally we should really cover the different classes of properties in WPT...)

PR #11405, opened 2018-06-07
Fix flakiness in test order of worker-dedicated.html

zcorpan:
(An alternative fix may be to make `fetch_tests_from_worker` serialize, like `promise_test`.)

### DIFF
--- a/resources/test/tests/functional/worker-2.js
+++ b/resources/test/tests/functional/worker-2.js
@@ -1,0 +1,10 @@
+importScripts("/resources/testharness.js");
+
+test(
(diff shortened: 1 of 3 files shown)


zcorpan:
Also see #11406

zcorpan:
@jugglinmike suggests adding a passing worker to the non-error test to not regress coverage.

jugglinmike:
[As explained in gh-11344](https://github.com/web-platform-tests/wpt/pull/11344#issuecomment-397771199), this is not necessary.

PR #11414, opened 2018-06-07
[idlharness.js] Support subsetTestByKey in idlharness

### LINKED ISSUE #11413: /dom/interfaces.html is disabled in webkit

Investigate+what's+up+with+this+test:

-- | -- | -- | --
/dom/interfaces.html | webkit | [ Skip ] | https://webkit.org/b/148546

This and html/dom/interfaces.https.html could be split up using `subsetTest`.

zcorpan:
Fixes #11413.


gsnedders:
Will this not easily result in subtests moving, given idlharness.js adding/removing tests will change the number produced? A bit painful for expectation management, IMO.

zcorpan:
Yes. Could split more semantically.

zcorpan:
@gsnedders PTAL

gsnedders (approved):
This LGTM. Would be interested in opinions from others, though?

gsnedders:
One quick thought: I wonder if `subtestByName` is really the best name? To me, that made me think it used `test.name` at first glance, which it doesn't. I wonder if something like `subtestByKey(key, callback, ...)` makes more sense?

zcorpan:
Renamed to key.

PR #11415, opened 2018-06-07
Added tests for InputFile directly

dheiberg:
As suggested, I added some tests that test InputFile directly, without going through a server. I also made some minor changes in `test_request` to satisfy flake8. Let me know if I should make the tests more extensive/change them.

### DIFF
--- a/tools/wptserve/tests/functional/test_input_file.py
+++ b/tools/wptserve/tests/functional/test_input_file.py
@@ -1,0 +1,148 @@
+from wptserve.request import InputFile
+from io import BytesIO
+import pytest
+
+wptserve = pytest.importorskip("wptserve")
+bstr = b'This is a test document\nWith new lines\nSeveral in fact...'
+rfile = ''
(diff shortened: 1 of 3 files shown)


tools/wptserve/tests/functional/test_input_file.py
```diff
@@ -0,0 +1,123 @@
...
+
+        input_file.seek(4)
+        test_file.seek(4)
+        assert input_file.read(1) == test_file.read(1)
+    finally:
+        InputFile.max_buffer_size = old_max_buf
+
+
+def test_readline():
+
```
jgraham: No blank line after the function definition, here and throughout.

jgraham (changes requested):
This mosty looks good. Did you try running this under code coverage to see if there are any obvious missing cases?

jgraham (approved):
In general this looks good, thanks. For future reference, it's helpful to avoid merging branches into pull request histories; generally working with rebase is easier for tooling like the two way sync.

PR #11419, opened 2018-06-07
[idlharness.js] Test exposure set of partial interfaces

lukebjerring:
Resolves https://github.com/web-platform-tests/wpt/issues/10338
Equivalent Chromium CL: https://chromium-review.googlesource.com/c/chromium/src/+/1091210

- Converts the `exposure_set` return value to be a `Set`
- Extracts the 'collapse `partial interface` definitions into original interfaces' code into a method `collapse_partial_interfaces`
- Adds a new `test` for each partial interface with `Exposed`, asserting that its values are a subset of the original interfaces
- Wraps each individual 'partial interface collapse' loop in a test, so that if it's invalid, the remaining setup + testing of the `IdlArray` continues.

### DIFF
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -695,18 +695,15 @@ function exposed_in(globals) {
     }
     if ('DedicatedWorkerGlobalScope' in self &&
         self instanceof DedicatedWorkerGlobalScope) {
-        return globals.has("Worker") ||
-               globals.has("DedicatedWorker");
+        return globals.has("DedicatedWorker");
     }
     if ('SharedWorkerGlobalScope' in self &&
         self instanceof SharedWorkerGlobalScope) {
-        return globals.has("Worker") ||
-               globals.has("SharedWorker");
+        return globals.has("SharedWorker");
     }
     if ('ServiceWorkerGlobalScope' in self &&
         self instanceof ServiceWorkerGlobalScope) {
-        return globals.has("Worker") ||
-               globals.has("ServiceWorker");
+        return globals.has("ServiceWorker");
     }
     throw new IdlHarnessError("Unexpected global object");
 }
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -837,53 +837,56 @@ IdlArray.prototype.collapse_partials = function()
         }
         testedPartials.set(parsed_idl.name, partialTestCount);
 
-        test(function () {
-            if (!(parsed_idl.name in this.members)
-                || !(this.members[parsed_idl.name] instanceof IdlInterface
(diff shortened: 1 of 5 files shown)


lukebjerring:
This fixes a large number of broken tests:
https://test-results.appspot.com/data/layout_results/linux_chromium_rel_ng/112501/site_per_process_webkit_layout_tests%20%28with%20patch%29/layout-test-results/results.html

resources/idlharness.js
```diff
@@ -833,6 +821,76 @@ IdlArray.prototype.test = function()
...
+            const exposureAttr = parsed_idl.extAttrs.find(a => a.name === "Exposed");
+            if (exposureAttr) {
+                if (!parsed_idl.untested) {
+                    test(function () {
+                        const partialExposure = exposure_set(parsed_idl);
+                        const memberExposure = exposure_set(this.members[parsed_idl.name]);
+                        partialExposure.forEach(name => {
+                            if (!memberExposure || !memberExposure.has(name)) {
+                                throw new IdlHarnessError(
+                                    `Partial ${parsed_idl.name} ${parsed_idl.type} is exposed to '${name}', the original ${parsed_idl.type} is not.`);
```
foolip: This combined with how `exposure_set` makes for a bug which isn't currently triggered by any IDL, but could be. If the main interface definition uses `Exposed=(SharedWorker,DedicatedWorker,ServiceWorker)` and a partial interface uses `Exposed=Worker` as a shorthand for the same, this will fail with "Partial Foo interface is exposed to 'Worker', the original interface is not."

Commenting in `exposure_set` with what I think the fix should be.

resources/idlharness.js
```diff
@@ -662,42 +662,50 @@ IdlArray.prototype.is_json_type = function(type)
...
-        set = [ set ];
+    if (exposed && exposed.length) {
+        var set = exposed[0].rhs.value;
+        // Could be a list or a string.
+        if (typeof set == "string") {
+            set = [ set ];
+        }
+        result = new Set(set);
+    }
+    if (result && result.has("Worker")) {
```
foolip: Expanding "Worker" by also *removing* "Worker" would solve two problems:
1. `exposed_in` need not check for "Worker" (actually already unnecessary)
2. Lurking bug in `collapse_partials` is thwarted.

But it's not a very serious problem, so you pick if it's fixed here, in a follow up, or never.

foolip (approved):
LGTM % lurking bug

resources/idlharness.js
lukebjerring: I actually original had the remove (I think you can see it in my first commit), but thought it was better to leave it there; I'll add a unit test for the case you describe and then decide how to fix it.

PR #11424, opened 2018-06-08
Revert "WebKit export of https://bugs.webkit.org/show_bug.cgi?id=1861…

fred-wang:
…27 (#11279)"

This reverts commit 57fea798e1e933e1d5291608d9d12da661185bcb.

According to @birtles, Mozilla CI can have several tens of seconds between two
frames while the test assumes only hundreds of milliseconds. So this is likely
to be flacky.


mrego (approved):
I guess you'd some action on the WebKit side too.

fred-wang:
@mrego I was not imported int WebKit yet. I'll just write WebKit-specific tests instead.

PR #11429, opened 2018-06-08
Ported the wptserve tests over to pytest

dheiberg:


tools/wptserve/tests/functional/base.py
```diff
@@ -36,16 +35,16 @@ def add_data(self, data):
         BaseRequest.add_data(self, data)
 
 
-class TestUsingServer(unittest.TestCase):
-    def setUp(self):
+class TestUsingServer:
```
jgraham: Let's create a `conftest.py` file and use pytest fixtures rather than this kind of class-based setup.

tools/wptserve/tests/functional/test_cookies.py
```diff
@@ -1,9 +1,8 @@
-import unittest
+from .base import TestUsingServer
 
 import pytest
 
 wptserve = pytest.importorskip("wptserve")
-from .base import TestUsingServer
 
 
 class TestResponseSetCookie(TestUsingServer):
```
jgraham: If we move to pytest fixtures, I think we can also remove the class wrappers here and just use normal functions.

jgraham (changes requested):

stephenmcgruer:
The wptserve tests have changed significantly (50+ changes) since June 2018, so I don't think its worth trying to keep this PR alive. If we want to make this change, I think we'd have to start over.

PR #5284, opened 2017-03-31
[css-logical-properties-1] Add test case margin{-block,-inline}

gsnedders:
Originally posted as https://github.com/w3c/csswg-test/pull/1170 by @axinging on 10 Jan 2017, 07:59 UTC:

> Spec is here: https://drafts.csswg.org/css-logical-props-1/#logical-prop.
> This depends on https://github.com/w3c/csswg-test/pull/1165.
> 
>

PR #11431, opened 2018-06-08
[css-logical-properties-1] Add test case margin{-block,-inline}{-start,-end}

gsnedders:
Spec is here: https://drafts.csswg.org/css-logical-props-1/#logical-prop.

[Committer note]: This was split out of a larger mostly obsolete PR, https://github.com/web-platform-tests/wpt/pull/5284


mrego (approved):
Thanks for doing this.
Links could go to css-logical-1 now, but it's fine anyway.

PR #11433, opened 2018-06-08
Add test for WebSocket HTTP Authentication

ricea:
User-agents should support HTTP authentication for WebSockets. Add a
test of basic authentication to verify this.

See https://github.com/whatwg/fetch/issues/565.

### DIFF
--- a/websockets/basic-auth.any.js
+++ b/websockets/basic-auth.any.js
@@ -1,0 +1,18 @@
+// META: script=websocket.sub.js
+// META: global=sharedworker,serviceworker
+
+async_test(t => {
+  const isSecure = new URL(location.href).scheme === 'https';
+  const scheme = isSecure ? 'wss' : 'ws';
+  const port = isSecure ? __SECURE__PORT : __PORT;
+  const url = scheme + '://' + 'foo:bar@' + __SERVER__NAME + ':' + port + '/basic_auth';
+  const ws = new WebSocket(url);
+  ws.onopen = () => {
+    ws.onclose = ws.onerror = null;
+    ws.close();
+    t.done();
+  };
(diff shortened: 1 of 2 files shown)


ricea:
Tested in Chrome, Firefox, Edge and Safari. Unfortunately this test doesn't pass in Edge as it won't accept credentials in the URL. It doesn't pass in Safari because Safari doesn't support HTTP auth for WebSockets: https://bugs.webkit.org/show_bug.cgi?id=80362

annevk:
Safari does support it per https://bugs.webkit.org/show_bug.cgi?id=80362#c7. It's credentials in the URL that are not supported.

I also wonder how this relates to the "use-URL-credentials flag", but I guess that'd be a separate test.

ricea:
I just re-tested on Safari Technology Preview using [pywebsocket](https://github.com/google/pywebsocket) directly:
```
PYTHONPATH=. python mod_pywebsocket/standalone.py --basic-auth -p 8888 \
  --basic-auth-credential=test:test -d ../../example --log-level=fine
```
Then visited http://localhost:8888/console.html

Safari prompts for credentials for the HTML page as expected, and reuses the credentials for other HTTP resources as expected, but doesn't use the credentials when making a WebSocket connection.

So I'm pretty sure that it isn't implemented in Safari, and comment https://bugs.webkit.org/show_bug.cgi?id=80362#c7 is just saying how to work around the issue of not prompting for credentials.

annevk (approved):
Thanks! I checked Fetch again and the main problem I see there is that if the server replied with a 401 here, Fetch currently suggests that would result in prompting, which shouldn't happen. But that's probably generally out-of-date as browsers started prompting less.

PR #11437, opened 2018-06-08
HTML: add a test for :heading

zcorpan:
Part of https://github.com/whatwg/html/pull/3499.

This does not yet test :heading().


annevk (approved):
Thanks! I still wonder if we should have another look at `hgroup` and alternate solutions.

Approving this, but let's wait with landing until we're going ahead with the specification change.

annevk:
Let's close this as it doesn't seem this is happening anymore. 😟

PR #11438, opened 2018-06-08
URL: correct non-special URL's origin

annevk:
Fixup for 0e084a927078d434bcc1952a2ab99fb541f2c25d.


domenic (approved):
LGTM, confirmed passing now (and failing in the expected way before applying the spec change).

PR #11441, opened 2018-06-08
Clean up FileAPI idlharness tests

lukebjerring:
After fixing https://github.com/web-platform-tests/wpt/issues/10338, we can leverage `add_dependency_idls` to import the 'real' untested idl from sources cleanly.

### DIFF
--- a/FileAPI/idlharness.html
+++ b/FileAPI/idlharness.html
@@ -20,37 +20,30 @@
     </form>
 
     <script>
-var file_input;
-setup(function() {
-    var idl_array = new IdlArray();
+      'use strict';
 
-    var request = new XMLHttpRequest();
-    request.open("GET", "/interfaces/FileAPI.idl");
-    request.send();
-    request.onload = function() {
-        var idls = request.responseText;
+      promise_test(async () => {
+        const idl = await fetch('/interfaces/FileAPI.idl').then(r => r.text());
+        const dom = await fetch('/interfaces/dom.idl').then(r => r.text());
+        const html = await fetch('/interfaces/html.idl').then(r => r.text());
+        const url = await fetch('/interfaces/url.idl').then(r => r.text());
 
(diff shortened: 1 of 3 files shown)


lukebjerring:
Note, also, that #11419 caused a test failure to show up for FileAPI (since the untested original interface didn't have an `Exposed` attr). 

> FAIL Partial interface URL: valid exposure set Partial URL interface is exposed to 'DedicatedWorker', the original interface is not.

https://test-results.appspot.com/data/layout_results/linux_chromium_rel_ng/113144/layout-test-results/results.html

This change fixes that.

FileAPI/idlharness.html
```diff
@@ -20,37 +20,30 @@ <h1>File API automated IDL tests</h1>
...
 
-        file_input = document.querySelector("#fileChooser");
         idl_array.add_objects({
-            Blob: ['new Blob(["TEST"])'],
-            File: ['new File(["myFileBits"], "myFileName")'],
-            FileList: ['file_input.files'],
-            FileReader: ['new FileReader()']
+          Blob: ['new Blob(["TEST"])'],
+          File: ['new File(["myFileBits"], "myFileName")'],
+          FileList: [document.querySelector("#fileChooser")],
```
inexorabletash: Doesn't this still need `.files` ?

lukebjerring:
Chromium CI equivalent (except the manual test):
https://chromium-review.googlesource.com/c/chromium/src/+/1094416

foolip (approved):
Fixed indentation, LGTM, not sure if @inexorabletash wants a second look?

inexorabletash (approved):
lgtm; the linked chromium CI's worker test times out; did that get fixed?

lukebjerring:
Re-running now, but not particularly worried since the other test passed fine. Feel free to wait on re-run, or merge if happy as-is.

PR #11444, opened 2018-06-08
Clean up mediasession idl test

lukebjerring:
Follow up from #11253 now that #10338 is fixed.


csnardi (approved):
Generates the same number of passes and fails on Chrome, so seems good to me.

PR #11464, opened 2018-06-11
Fix #11454: Set --verify-log-full to False for ci/check_stability.py

### LINKED ISSUE #11454: check_stability.py is too verbose
After https://github.com/web-platform-tests/wpt/pull/10988 it looks like the stability jobs are more verbose. https://travis-ci.org/web-platform-tests/wpt/jobs/389217237 (for https://github.com/web-platform-tests/wpt/pull/11301) is an example of this, where the raw log ends with "Log reached capacity (3145728 bytes); output disabled" during "Repetition 8 / 10".

The job failed, and it's not possible to tell why.

gsnedders:
I'm not entirely sure why #10988 made this default to on, given that was a change of behaviour from before, so let's turn it back off given it's causing breakage, as documented in #11454.


Hexcles (approved):
r+

PR #11469, opened 2018-06-12
CSS Fonts parsing tests

ewilligers:
Test parsing and serialization of CSS Fonts 3 properties
https://drafts.csswg.org/css-fonts/


css/css-fonts/parsing/font-family-valid.html
```diff
@@ -0,0 +1,24 @@
...
+<script src="/resources/testharnessreport.js"></script>
+<script src="resources/parsing-testcommon.js"></script>
+</head>
+<body>
+<script>
+test_valid_value("font-family", "Helvetica");
+test_valid_value("font-family", "Helvetica, Verdana");
+
+test_valid_value("font-family", "serif");
+test_valid_value("font-family", "serif, sans-serif, cursive, fantasy, monospace");
```
svgeesus: Not a blocker, but would be great to see a related test for Fonts 4 with the larger list of generic families

svgeesus (changes requested):
For all tests, please change

```
<link rel="help" href="https://drafts.csswg.org/css-fonts/
```

to

```
<link rel="help" href="https://www.w3.org/TR/css-fonts-3/
```
This will allow the CSS test harness to pick up these tests, annotate each section of the draft with links to the tests for that section, and update the test reports. 

It also means we can clearly indicate these are Fonts 3 tests (for example Fonts 4 has a larger set of generic font families, among many other changes)

Thanks!

css/css-fonts/parsing/font-variant-alternates-invalid.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>CSS Fonts Level 3: parsing font-variant-alternates with invalid values</title>
```
svgeesus: Please change the font-variant-alternates tests to Fonts 4. The CSS WG resolved to move this feature from Fonts 3 to Fonts 4, as there is currently only one implementation. (the spec edits are pending on me, hope to get them done before the CSS f2f in Sydney).

css/css-fonts/parsing/font-variant-valid.html
```diff
@@ -0,0 +1,80 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>CSS Fonts Level 3: parsing font-variant with valid values</title>
```
svgeesus: Please  remove the font-variant-alternates values from this test (moved from Fonts 3 to Fonts 4)

css/css-fonts/parsing/font-weight-invalid.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>CSS Fonts Level 3: parsing font-weight with invalid values</title>
+<link rel="author" title="Eric Willigers" href="mailto:ericwilligers@chromium.org">
+<link rel="help" href="https://drafts.csswg.org/css-fonts/#font-weight-prop">
+<meta name="assert" content="font-weight supports only the grammar 'normal | bold | bolder | lighter | 100 | 200 | 300 | 400 | 500 | 600 | 700 | 800 | 900'.">
```
svgeesus: Thanks for *not* testing values like font-weight: 345 which are invalid in Fonts 3 but will be valid in Fonts 4.
This test looks good.

svgeesus (changes requested):
Overall, a great set of tests. Please see detailed comments.

css/css-fonts/parsing/font-family-valid.html
ewilligers: People familiar with CSS Fonts 4 can add test cases as a later PR.

Other comments have been addressed.

PR #11478, opened 2018-06-12
Add test run time to wptreport.json format

### LINKED ISSUE #10026: Add test execution time to wptreport.json format
This was suggested in # #9972.

Basically the execution time is the timestamp of the `test_end` event minus the timestamp of the `test_start` event. So to fix this look in `tools/wptrunner/wptrunner/formatters.py`, add a `time` field to the data and populate it with the execution time per test.

gsnedders:
Fixes #10026. Also fix the output when there is no `run_info` provided, given the API allows it, which I hit trying to write a test (given the mozlog docs omit it in their examples).

### DIFF
--- a/tools/wptrunner/wptrunner/tests/test_formatters.py
+++ b/tools/wptrunner/wptrunner/tests/test_formatters.py
@@ -1,0 +1,39 @@
+import json
+import sys
+import time
+from os.path import dirname, join
+from StringIO import StringIO
+
+from mozlog import handlers, structuredlog
+
+sys.path.insert(0, join(dirname(__file__), "..", ".."))
+
+from wptrunner.formatters import WptreportFormatter
+
+
+def test_wptreport_runtime(capfd):
+    # setup the logger
+    output = StringIO()
+    logger = structuredlog.StructuredLogger("test_a")
+    logger.add_handler(handlers.StreamHandler(output, WptreportFormatter()))
+
+    # output a bunch of stuff
+    logger.suite_start(["test-id-1"], run_info={})
+    logger.test_start("test-id-1")
+    time.sleep(0.125)
+    logger.test_end("test-id-1", "PASS")
+    logger.suite_end()
+
+    # check nothing got output to stdout/stderr
+    # (note that mozlog outputs exceptions during handling to stderr!)
+    captured = capfd.readouterr()
+    assert captured.out == ""
+    assert captured.err == ""
+
+    # check the actual output of the formatter
+    output.seek(0)
+    output_obj = json.load(output)
+    # be relatively lax in case of low resolution timers
+    # 62 is 0.125s = 125ms / 2 = 62ms (assuming int maths)
+    # this provides a margin of 62ms, sufficient for even DOS (55ms timer)
+    assert output_obj["results"][0]["duration"] >= 62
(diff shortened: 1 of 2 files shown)


tools/wptrunner/wptrunner/tests/test_formatters.py
```diff
@@ -0,0 +1,61 @@
...
+def test_wptreport_runtime(capfd):
+    # setup the logger
+    output = StringIO()
+    logger = structuredlog.StructuredLogger("test_a")
+    logger.add_handler(handlers.StreamHandler(output, WptreportFormatter()))
+
+    # output a bunch of stuff
+    logger.suite_start(["test-id-1"], run_info={})
+    logger.test_start("test-id-1")
+    time.sleep(0.125)
```
Hexcles: I suppose you've tried to and couldn't mock out the timestamps? Sleeping 0.125 second is *probably* fine but definitely not my favourite.

Hexcles (approved):
Awesome! This looks good to me overall, but please wait for @jgraham , too.

tools/wptrunner/wptrunner/tests/test_formatters.py
gsnedders: Given they're generated internally to `mozlog` I honestly didn't even try.

tools/wptrunner/wptrunner/tests/test_formatters.py
```diff
@@ -0,0 +1,61 @@
...
+    assert captured.err == ""
+
+    # check the actual output of the formatter
+    output.seek(0)
+    output_obj = json.load(output)
+    assert output_obj["results"][0]["runtime"] >= 125
+
+
+def test_wptreport_run_info_optional(capfd):
+    """per the mozlog docs, run_info is optional; check we work without it"""
```
jgraham: FWIW it *is* guaranteed that wptrunner adds `run_info`, so this test is only really important if we upstream this to mozlog.

tools/wptrunner/wptrunner/tests/test_formatters.py
```diff
@@ -0,0 +1,61 @@
...
+    # check nothing got output to stdout/stderr
+    # (note that mozlog outputs exceptions during handling to stderr!)
+    captured = capfd.readouterr()
+    assert captured.out == ""
+    assert captured.err == ""
+
+    # check the actual output of the formatter
+    output.seek(0)
+    output_obj = json.load(output)
+    assert output_obj["results"][0]["runtime"] >= 125
```
jgraham: Sos, is this going to end up intermittent? :)

jgraham (approved):
Thanks!

tools/wptrunner/wptrunner/tests/test_formatters.py
gsnedders: It shouldn't unless `time.sleep` sleeps for less than 125ms at least. Though maybe we'll run into timer resolution issues, actually. Ergh.

gsnedders: Can we just try landing it, though, and see if it actually flaky? I'd hope timer resolution issues would mean we'll just have a longer runtime? (Also `runtime` or `run_time`?)

jgraham: `duration` if you ask me, but I really don't mind.

I'm happy to land this I guess, but I would be more nervous if we were running tests on Windows XP or something. I would also just be happy to check that the value is a number > 0.

gsnedders: It would appear the Windows APIs mostly make no guarantee as to rounding, which is fun. Requiring 62ms should be fine, though, AFAICT.

gsnedders:
@jgraham do you want to look at this again before landing?

jgraham:
It;s not really obvious where `62` comes from. Maybe add a comment.

gsnedders:
@jgraham is this okay?

jgraham (approved):

PR #11480, opened 2018-06-12
Fix c563-list-type-000-ref to match the test

gsnedders:
Currently the introductory text in the ref isn't navy, unlike in the test.


svgeesus (approved):
> 35- Standardized color for the pass/fail conditions sentence

> Some Ian 'Hixie' Hickson tests use unneedlessly blue navy for the pass/fail conditions sentence. 
> -- G. Talbot

Yes, consistency is good and helps automated reftesting

PR #11489, opened 2018-06-13
Rewrite access-control-and-redirects-async test to use promise_test to…

bmac:
…ensure consistent results in WebKit

This test is marked as flaky in WebKit's [test expectation file](https://github.com/WebKit/webkit/blob/f415af3483f4f9136e45a93f69e9d7a511d6018c/LayoutTests/TestExpectations#L286) because the order of results are reported inconsistently. Changing `async_test ` to `promise_test` ensures the tests are run (and reported) in a consistent order. Webkit bug: https://bugs.webkit.org/show_bug.cgi?id=179607.

### DIFF
--- a/cors/status-async.htm
+++ b/cors/status-async.htm
@@ -14,30 +14,31 @@
 <script>
 
 function statusRequest(method, code, text, content, type) {
-    async_test("Status on " + method + " " + code, { timeout: 15000 })
-    .step(function() {
-        var client = new XMLHttpRequest()
-        client.open(method, CROSSDOMAIN + "resources/status.py?code="
-            + code + "&text=" + text + "&content=" + content + "&type=" + type, true)
-        client.onreadystatechange = this.step_func(function() {
-            if (client.readyState != client.DONE)
-                return
+    promise_test(function(t) {
+        return new Promise(resolve => {
+            var client = new XMLHttpRequest()
+            client.open(method, CROSSDOMAIN + "resources/status.py?code="
+                + code + "&text=" + text + "&content=" + content + "&type=" + type, true)
+            client.onreadystatechange = t.step_func(function() {
+                if (client.readyState != client.DONE)
+                    return
 
-            assert_equals(client.status, code, 'response status')
-            assert_equals(client.statusText, text, 'response status text')
-            assert_equals(client.getResponseHeader("X-Request-Method"), method, 'method')
-            if(method != "HEAD") {
-                if(type == "text/xml") {
-                    assert_equals(client.responseXML.documentElement.localName,
+                assert_equals(client.status, code, 'response status')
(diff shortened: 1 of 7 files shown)


cors/status-async.htm
```diff
@@ -60,36 +61,37 @@ <h1>Status returned</h1>
...
-            assert_equals(client.response, "", "response data")
-            assert_equals(client.status, expect_code, "response status")
-            /* Response code 200 forces webserver to send OK(?) */
-            if(expect_code == 200)
-                assert_equals(client.statusText, "OK", "response statusText")
-            else
-                assert_equals(client.statusText, (expect_code == 0 ? "" : "OHAI"), "response statusText")
-            if (client.readyState == client.DONE)
-                this.done()
+    async_test(function() {
```
annevk: This seems wrong? Did you rerun all these tests and verify the results are identical?

xhr/access-control-and-redirects-async-same-origin.htm
```diff
@@ -32,38 +36,38 @@
     // Test simple same origin requests that receive cross origin redirects.
 
     // The redirect response passes the access check.
-    async_test(t => {
-      runTest(t, "/xhr/resources/access-control-basic-allow-star.py",
+    promise_test(t => {
+      return runTest(t, "/xhr/resources/access-control-basic-allow-star.py",
```
annevk: I wonder if we should refactor this such that runTest runs promise_test.

annevk (commented):
This seems like a bug in WebKit's test harness as the order is guaranteed to be deterministic I think. So therefore I'm a little skeptical we should make this change, as it'd argue to never use async_test because WebKit can't handle it. Am I missing something?

cors/status-async.htm
bmac: Yes, I'm seeing the same results on localhost (http://web-platform.test:8000) as I see when I run the tests on http://w3c-test.org/. Which part specifically seems wrong?

#### Local tests using this branch
screen shot 2018-07-17 at 2 07 34 pm


#### w3c-test.org using master
screen shot 2018-07-17 at 2 08 01 pm

youennf:
> This seems like a bug in WebKit's test harness as the order is guaranteed to be deterministic I think. So therefore I'm a little skeptical we should make this change, as it'd argue to never use async_test because WebKit can't handle it. Am I missing something?

async_tests are only running part of the code in a known order.
promise_test will run sub test one after the other is done.
This has impact in particular on outputted console log error messages that can make tests flaky in WebKit. The ordering of these messages is often not guaranteed with async_test.
Also, for debugging purposes, promise_test slightly helps.

annevk:
This was resolved on the WebKit side by ignoring console messages for relevant tests.

PR #11491, opened 2018-06-13
first pass at testing selectionchange

boazsender:
This PR introduces initial coverage for the [`selectionchange` event](https://w3c.github.io/selection-api/#selectionchange-event).

I started looking into `selectionchange` while trying to understand the compat issue that the Facebook React project is working around in its [SelectEventPlugin](https://github.com/facebook/react/blob/v16.3.2/packages/react-dom/src/events/SelectEventPlugin.js#L183). I'll continue to checkin on this topic area while I chase the React SelectEventPlugin issue.

### DIFF
--- a/selection/selectionchange.html
+++ b/selection/selectionchange.html
@@ -1,0 +1,58 @@
+<!doctype html>
+<title>CustomEvent</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div id="log"></div>
+
+<span id="span">outside a content editable</span>
+
+<section id="contenteditableSection" style="border:1px solid" contenteditable>inside a content editable</section>
+
+<textarea id="textarea">inside a textarea</textarea>
+
+<input id="input" type="text" value="inside an input">
+
+<script>
+  window.addEventListener('load', function(){
+    // iterate over all of our different kinds of selectable elements
+    [span, contenteditableSection, textarea, input].forEach(id => {
+      // make a promise test for it
+      promise_test(function(t) {
+        return new Promise(function(resolve){
+          // bind to the doc's selectionchange event
+          document.addEventListener('selectionchange', t.step_func(function(){
+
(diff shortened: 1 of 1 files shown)


gredler:
@boazsender @zcorpan

Hi guys, I've run into a few browser bugs around the selectionchange event (e.g. https://bugs.chromium.org/p/chromium/issues/detail?id=725890 and https://bugzilla.mozilla.org/show_bug.cgi?id=1648944) and figured this might be a good area to test. However, I see that this PR was created a few years ago and never got any traction. I'm wondering whether this is just not an area that we want to test for some reason, or whether I should wait until this PR is merged and then extend it, or...? Please let me know what you think the best way forward is. Thanks!

boazsender:
Hi @gredler :wave: 

I haven't heard anything about this area being intentionally de-prioritized. WPT has a biiig backlog of PRs and not a lot of reviewer bandwidth, so my guess is we're just waiting in line. 

I think the wpt community and this patch would welcome additional tests, and new energy on test writing, responding to a review could very well get new tests in. 

I'll try to take a closer look at those bugs and advise on next steps. In general, I think the thing to do is for us to figure out what browser / spec changes are needed in order to squash the bug, make sure our tests match that behaviour, and then work with implementers from those browser teams and the right spec editor to make those changes.

Thanks so much for pinging this PR.

Here's to a stable and interoperable `selectionchange` event in ~2021!

p.s.
I think there are a handful of frameworks that would benefit from better interop here as well. I know react has code dedicated to papering over inconsistencies in `selectionchange`.

coderabhi5565 (changes requested):
Use document.getElementById(...) instead of relying on id → global variables.
Detect text controls with ('selectionStart' in el && typeof el.selectionStart === 'number') (don't use if (el.value)).
Add the selectionchange listener with { once: true } and a short timeout + cleanup so tests don't hang.
Prefer elem.select() for inputs/textareas; fall back to getSelection().selectAllChildren(elem).
Update the <title> to reflect selectionchange.

PR #11492, opened 2018-06-13
Reformat URL tests

jugglinmike:
Refactor tests so that they may be consumed by non-browser JavaScript runtimes which implement the WHATWG URL standard such as Node.js [1]. Using WPT's `.any.js` convention also extends test coverage in browsers by allowing them to be executed within a Web Worker.

This change is in service of gh-11277.

[1] https://nodejs.org/dist/latest-v8.x/docs/api/url.html#url_the_whatwg_url_api

/cc @jorydotcom @mariestaver @mzgoddard @nomadtechie @spectranaut


domenic:
This does not appear to touch my precious .json files so this should not disrupt my workflow :). I'll leave reviews of the conversion of the wrappers to others.

url/url-tojson.any.js
```diff
@@ -0,0 +1,4 @@
+test(() => {
+  const a = new URL('https://example.com/');
+  assert_equals(JSON.stringify(a), '"https://example.com/"');
```
annevk: Could you undo the quoting changes to see if GitHub then recognizes this as a rename + modifications rather than a new resource?

url/urlencoded-parser.any.js
```diff
@@ -67,5 +60,4 @@
       }
     })
   }, "response.formData() with input: " + val.input)
-})
-</script>
+});
```
annevk: This needs a newline.

url/urlsearchparams-append.any.js
```diff
@@ -1,11 +1,7 @@
...
-<html>
-<head>
-<meta charset="utf8">
-<link rel="help" href="http://url.spec.whatwg.org/#dom-urlsearchparams-append">
-<script src="/resources/testharness.js"></script>
-<script src="/resources/testharnessreport.js"></script>
-<script>
+/*
+    https://url.spec.whatwg.org/#dom-urlsearchparams-append
+ */
```
annevk: No need to keep this. Or any of the links. It's pretty self-explanatory where this is all documented.

url/urlsearchparams-stringifier.any.js
```diff
@@ -128,9 +123,4 @@
 
     assert_equals(url.toString(), 'http://www.example.com/?a=b%2Cc&x=y');
     assert_equals(params.toString(), 'a=b%2Cc&x=y');
-}, 'URLSearchParams connected to URL');
-
-</script>
-</head>
-</html>
-
+}, 'URLSearchParams connected to URL');
```
annevk: Please keep the newline.

url/urlsearchparams-sort.any.js
```diff
@@ -53,5 +47,4 @@
   url.searchParams.sort()
   assert_equals(url.href, "http://example.com/")
   assert_equals(url.search, "")
-}, "Sorting non-existent params removes ? from URL")
-</script>
+}, "Sorting non-existent params removes ? from URL")
```
annevk: Please keep the newline.

url/url-searchparams.any.js
```diff
@@ -0,0 +1,72 @@
...
+    var url2 = bURL('http://example.org/file??a=b&c=d')
+    assert_equals(url2.search, '??a=b&c=d')
+    assert_equals(url2.searchParams.toString(), '%3Fa=b&c=d')
+
+    url2.href = 'http://example.org/file??a=b'
+    assert_equals(url2.search, '??a=b')
+    assert_equals(url2.searchParams.toString(), '%3Fa=b')
+  }, 'URL.searchParams and URL.search setters, update propagation')
+}
+runURLSearchParamTests()
```
annevk: This was strictly copy-and-pasted, right?

annevk (commented):
Thanks for doing this! Couple of nits.

url/urlsearchparams-append.any.js
nomadtechie: Thanks for the feedback @annevk!

I'm curious to hear why you think its obvious and self-explanatory. I agree it might be the case for spec authors and implementors. However, for new WPT contributors without a strong standards background, I think this link would be pretty helpful.

annevk: I was thinking the README had a link to the standard (as a standalone section), but it seems like it doesn't. I think we should add that. Test metadata tends to go out-of-date.

nomadtechie: Sounds good @annevk. Will update.

url/urlencoded-parser.any.js
jugglinmike: Done, though not recognized by GitHub

url/url-tojson.any.js
jugglinmike: @mariestaver Undid the quoting changes, but this is still being interpreted as a new file

url/url-searchparams.any.js
mzgoddard: Yes, this was strictly copy-and-pasted. Its the 68 lines from url/url-constructor.html plus 4 for the copy of `bURL`.

Ms2ger (approved):

Ms2ger:
@jugglinmike could you rebase, please?

url/urlencoded-parser.any.js
annevk: FWIW, I think it is. If there's no newline there's a red indicator at the end of the line.

jugglinmike:
@Ms2ger Sure. The prior version of this branch is available at https://github.com/web-platform-tests/wpt/compare/master...bocoup:liberate-url-3?expand=1 in case that's interesting to anyone.

PR #11073, opened 2018-05-18
Update webidl2.js to v12.1.2

foolip:
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/11073)

PR #9847, opened 2018-03-05
Update the webappsec-referrer-policy IDL file

lukebjerring:

PR #11496, opened 2018-06-14
Update webidl2 to v13.0.3

saschanaz:
Replaces and closes #11073.


saschanaz:
CC'ing @foolip who authored #11073.

resources/test/tests/unit/basic.html
```diff
@@ -24,7 +24,12 @@
         assert_equals(typeof WebIDL2.parse, "function");
     }, 'WebIDL2 namespace should have a parse method');
     test(function() {
-        assert_throws(new TypeError(), function() {  WebIDL2.parse("I'm a syntax error"); });
```
gsnedders: Why rewrite this to not use `assert_throws`?

gsnedders: (Like, yes, the error type has changed, but can we not still use `assert_throws`?)

csnardi:
A couple test changes are needed due to parsing becoming stricter:

- `fetch/api/idl.any.js` cannot reference an empty enum, as that is now invalid. #9847 would pull in that IDL file, which could then be used as a dependency IDL.
- `webaudio/the-audio-api/the-audiobuffer-interface/idl-test.html` and `webaudio/the-audio-api/the-audiodestinationnode-interface/idl-test.html` do not escape angle brackets, creating text that is now rejected as invalid. Really, these tests should be deleted in favor of `webaudio/idlharness.https.html`, but per https://wpt.fyi/results/webaudio/idlharness.https.html Chrome is the only browser that actually can run the test.
- `pointerevents/extension/idlharness.html` also does not escape angle brackets, which could be fixed by moving the IDL to `interfaces/` (but what should it be named? It's an extension of the pointer events spec).

foolip:
>  `pointerevents/extension/idlharness.html` also does not escape angle brackets, which could be fixed by moving the IDL to `interfaces/` (but what should it be named? It's an extension of the pointer events spec).

Spec is https://w3c.github.io/pointerevents/extension.html so I suggest `pointerevents-extension.idl`.

resources/test/tests/unit/basic.html
saschanaz: `assert_throws` seems to allow only DOMException and JS native error objects that have `name` property. Should the function be fixed to allow arbitrary error objects?

saschanaz:
@csnardi 

>creating text that is now rejected as invalid

Should they just be escaped before deletion?

csnardi:
@saschanaz Probably, I'm not sure a good way to make `webaudio/idlharness.https.html` run without failing immediately on the other browsers, so that seems like the only way to do it for now without losing test coverage.

foolip:
Please let me know if you'd like me to try out the change in Chromium's CQ, since that can help find breakage not found by our Travis setup here.

saschanaz:
@foolip That would be a great help! 👍

csnardi:
Since the last run didn't pick up my change to `fetch/api/idl.any.js`, I re-ran the failing jobs and now all tests are passing.

foolip:
Excellent, thanks @csnardi!

gsnedders:
Do they somehow pass despite https://github.com/web-platform-tests/wpt/pull/11496#issuecomment-397358030? does that mean we can land this?

csnardi:
The last two items were fixed inside of this PR (and also included in the try run). I think this could be landed, assuming we can't simplify the `assert_throws` change.

gsnedders (approved):

resources/test/tests/unit/basic.html
gsnedders: Made https://github.com/web-platform-tests/wpt/issues/11556 for this

PR #11511, opened 2018-06-14
[testdriver.js] Implement `bless` method

jugglinmike:
A common use case for the testdriver.js library is to create a task that
was triggered by user interaction [1]. Doing so using the current API
requires a fair amount of logic which is unrelated to the behavior under
test. This has a number of drawbacks:

- it makes tests less descriptive
- it makes tests less uniform both in terms of implementation and in
  terms of user experience (in non-automated scenarios, human operators
  may be prompted to provide the necessary input)
- it makes tests more susceptible to bugs

Define a utility method as a consistent and concise way to facilitate
the common use case. Require descriptive text via the `intent` parameter
in order to encourage test authors to explicitly describe the rationale
for requesting additional privileges.

[1] https://html.spec.whatwg.org/#triggered-by-user-activation

---

@kereliuk @gsnedders This is based on my experience applying testdriver.js to tests that rely on media playback. In every case, I've been using the library for one of two purposes:

- create a trusted task (i.e. to call `element.play()` or `new AudioContext`)
- signal engagement with the document (to ensure the UA observes the `autoplay` attribute)

testharness.js is exactly what's needed to serve these purposes. The thing is, the WebDriver-inspired API it offers is much more precise than what's necessary for this. Every test needs to perform similar setup to get what it needs:

- Create a button
- Add a click handler
- Insert the button into the document
- Request a click interaction
- Create a Promise which wraps a "click" event listener

None of this is particularly challenging, but it adds noise and a source of inconsistency between tests.

My use case is narrowly circumscribed, so I took a look at the existing usages of testdriver.js:

test                                                                                                     | purpose
---------------------------------------------------------------------------------------------------------|----------------------
[bluetooth/resources/bluetooth-helpers.js](https://github.com/web-platform-tests/wpt/blob/cf261625e2d230ab219eec966f4abe26e3401b64/bluetooth/resources/bluetooth-helpers.js)                                                               | create a trusted task
[feature-policy/resources/autoplay.js](https://github.com/web-platform-tests/wpt/blob/cf261625e2d230ab219eec966f4abe26e3401b64/feature-policy/resources/autoplay.js)                                                                   | signal engagement
[fetch/sec-metadata/window-open.tentative.https.sub.html](https://github.com/web-platform-tests/wpt/blob/cf261625e2d230ab219eec966f4abe26e3401b64/fetch/sec-metadata/window-open.tentative.https.sub.html)                                                | create a trusted task
[html/editing/focus/focus-01.html](https://github.com/web-platform-tests/wpt/blob/cf261625e2d230ab219eec966f4abe26e3401b64/html/editing/focus/focus-01.html)                                                                       | simulate a key press
[html/editing/focus/focus-02.html](https://github.com/web-platform-tests/wpt/blob/cf261625e2d230ab219eec966f4abe26e3401b64/html/editing/focus/focus-02.html)                                                                       | simulate a key press
[html/editing/focus/sequential-focus-navigation-and-the-tabindex-attribute/focus-tabindex-negative.html](https://github.com/web-platform-tests/wpt/blob/cf261625e2d230ab219eec966f4abe26e3401b64/html/editing/focus/sequential-focus-navigation-and-the-tabindex-attribute/focus-tabindex-negative.html) | simulate a key press
[html/editing/focus/sequential-focus-navigation-and-the-tabindex-attribute/focus-tabindex-order.html](https://github.com/web-platform-tests/wpt/blob/cf261625e2d230ab219eec966f4abe26e3401b64/html/editing/focus/sequential-focus-navigation-and-the-tabindex-attribute/focus-tabindex-order.html)    | simulate a key press
[html/editing/focus/sequential-focus-navigation-and-the-tabindex-attribute/focus-tabindex-positive.html](https://github.com/web-platform-tests/wpt/blob/cf261625e2d230ab219eec966f4abe26e3401b64/html/editing/focus/sequential-focus-navigation-and-the-tabindex-attribute/focus-tabindex-positive.html) | simulate a key press
[html/editing/focus/sequential-focus-navigation-and-the-tabindex-attribute/focus-tabindex-zero.html](https://github.com/web-platform-tests/wpt/blob/cf261625e2d230ab219eec966f4abe26e3401b64/html/editing/focus/sequential-focus-navigation-and-the-tabindex-attribute/focus-tabindex-zero.html)     | simulate a key press
[payment-request/payment-request-abort-method.https.html](https://github.com/web-platform-tests/wpt/blob/cf261625e2d230ab219eec966f4abe26e3401b64/payment-request/payment-request-abort-method.https.html)                                                | create a trusted task
[picture-in-picture/resources/picture-in-picture-helpers.js](https://github.com/web-platform-tests/wpt/blob/cf261625e2d230ab219eec966f4abe26e3401b64/picture-in-picture/resources/picture-in-picture-helpers.js)                                             | create a trusted task
[uievents/order-of-events/mouse-events/click-cancel.html](https://github.com/web-platform-tests/wpt/blob/cf261625e2d230ab219eec966f4abe26e3401b64/uievents/order-of-events/mouse-events/click-cancel.html)                                                | simulate a click
[webusb/resources/usb-helpers.js](https://github.com/web-platform-tests/wpt/blob/cf261625e2d230ab219eec966f4abe26e3401b64/webusb/resources/usb-helpers.js)                                                                        | create a trusted task

Note how many of those applications are themselves abstractions around the general pattern. I know we're only just beginning to deploy this library, but even now, it seems like an extension to the testdriver.js API is warranted.

If you agree, I'll be happy to finish up this patch, adding tests and documentation and refactoring those tests. What do you think?

### DIFF
--- a/resources/testdriver.js
+++ b/resources/testdriver.js
@@ -1,5 +1,6 @@
 (function() {
     "use strict";
+    var idCounter = 0;
 
     function getInViewCenterPoint(rect) {
         var left = Math.max(0, rect.left);
@@ -46,6 +47,41 @@
      * @namespace
      */
     window.test_driver = {
+        /**
+         * Trigger user interaction in order to grant additional privileges to
+         * a provided function.
+         *
+         * https://html.spec.whatwg.org/#triggered-by-user-activation
+         *
+         * @param {String} intent - a description of the action which much be
+         *                          triggered by user interaction
+         * @param {Function} action - code requiring escalated privileges
+         *
+         * @returns {Promise} fulfilled following user interaction and
+         *                    execution of the provided `action` function;
+         *                    rejected if interaction fails or the provided
+         *                    function throws an error
+         */
+        bless: function(intent, action) {
+            var button = document.createElement("button");
+            button.innerHTML = "This test requires user interaction.<br />" +
+                "Please click here to allow " + intent + ".";
+            button.id = "wpt-test-driver-bless-" + (idCounter += 1);
+            document.body.appendChild(button);
+
+            return new Promise(function(resolve, reject) {
+                    button.addEventListener("click", resolve);
+
+                    test_driver.click(button).catch(reject);
+                }).then(function() {
+                    button.remove();
+
+                    if (typeof action === "function") {
+                        return action();
+                    }
+                });
+        };
+
         /**
          * Triggers a user-initiated click
          *


zcorpan:
I think this seems useful. 👍

resources/testdriver.js
```diff
@@ -46,6 +47,41 @@
...
+            document.body.appendChild(button);
+
+            return new Promise(function(resolve, reject) {
+                    button.addEventListener("click", resolve);
+
+                    test_driver.click(button).catch(reject);
+                }).then(function() {
+                    button.remove();
+
+                    if (typeof action === "function") {
```
gsnedders: Do we need this? If we do need this, can we make the promise reject if it isn't a function?

gsnedders (approved):

gsnedders:
@jugglinmike do you have any idea why I didn't merge this? I don't…

jugglinmike:
I assumed that you were interrupted by a pelican that had wondered into your office. If not, then maybe you just forgot.

gsnedders:
I certainly don't remember a pelican. Oh well.

jugglinmike:
@gsnedders New guess about why you were waiting to merge:

> If you agree, I'll be happy to finish up this patch, adding tests and documentation and refactoring those tests. What do you think?

I'll make good on that offer on Monday.

resources/testdriver.js
```diff
@@ -46,6 +47,41 @@
...
+                        return action();
+                    }
+                });
+        };
```
Hexcles: @jugglinmike are you sure about this semicolon? Shouldn't it be comma instead?

Hexcles (commented):
This has introduced regressions in Chrome: https://chromium-review.googlesource.com/c/chromium/src/+/1120854/4/third_party/WebKit/LayoutTests/external/wpt/infrastructure/testdriver/click-expected.txt

I suspect there's a syntax error. @jugglinmike can you confirm and fix it up? Thanks!

resources/testdriver.js
Hexcles: BTW, @gsnedders do we use things like `eslint` that would catch this kind of errors?

jugglinmike:
@Hexcles I can confirm the syntax error, although it has since been resolved in `master`: https://github.com/web-platform-tests/wpt/pull/11743 . Is the file still giving you trouble since commit e2dcb12445c1e7dd01cf90a7a8302492cc3f3b59 ?

Hexcles:
Ahh sorry I didn't check the head. Yeah the fix has been imported to
Chromium.

jgraham:
So I just saw some tests using this, and it seems a little confusing? It's not really clear to me just reading the code that `bless` means "create a button, click it, and then remove it again", and it's not really clear what the `intent` string acheives (in the example it just rephrases the method in the callback). Given the general difficulty of comments getting outdated, an API requiring a mandatory comment seems like arguably an antipattern.

How would people feel about renaming this `testdriver.generate_trusted_click` or similar and dropping the first parameter?

jugglinmike:
> It's not really clear to me just reading the code that `bless` means "create > a button, click it, and then remove it again"

The method name and description are intentionally generic. This method should be used for functionality that must be "triggered by user interaction." If tests require a "click" event specifically, they likely have different needs and should use the "click" method directly.

Unable to find a precedent, "bless" was the most concise way I could describe this action. That makes documentation particularly important, and this patch is pretty light on docs (and worse, it lacks tests). This was a mistake on my part, so I submitted a follow-up patch to correct it a few weeks ago: gh-11840

> Given the general difficulty of comments getting outdated, an API requiring a > mandatory comment seems like arguably an antipattern.

Part of my... intent for the `intent` string is to encourage better test documentation. I can see how requiring this at the API level might seem too opinionated (bossy, even), but I think it's warranted because these requirements may be allowed but not required by the specification. To me, that makes it important for readers, authors, and editors to be on the same page about why the test uses this API.

The other reason to require that documentation is to ensure the tests are meaningful outside of automation contexts. If gh-11173 lands, then humans will be able to run these tests manually. An "intent" string explains to the user why they are being prompted for input beyond what we can say generically.

zcorpan:
`trusted_click` SGTM

Issue #9599, opened 2018-02-21
Get rid of tools/third_party


gsnedders:
Now we have everything running through `./wpt` which has the ability to manage virtual environments, we have no need to bundle all of our dependencies. We should get rid of tools/third_party and tools/html5lib (see also #7200) and tools/six; note that tools/pywebsocket is a fork of the upstream project and as such we need to keep that.

jgraham:
Pretty sure that doing this naively will break gecko. We *can* install stuff from PyPI in CI, but only from a local mirror that doesn't have everything. That doesn't mean it's a bad idea (it isn't), but it does mean that some coordination is required to ensure that all the required packages are either in that mirror or otherwise vendored into mozilla-central, and that the `mach` and `mozharness` commands are set up to do the right thing.

gsnedders:
How do you cope with the existing virtualenv requirements?

Hexcles:
This will break Blink as well. And it'd require more work than what's needed for Gecko described above. We don't install _anything_ from PyPI, but use a Chromium-only packaging system (cipd & vpython) instead. It's not too complicated to set up, but does require coordination and planning.

gsnedders:
See also https://github.com/pypa/pip/tree/master/src/pip/_vendor for how `pip._vendor` is maintained.

PR #11516, opened 2018-06-15
Fix Python 3 errors when doing ./wpt make-hosts-file

saschanaz:
This does not fix everything, as I was fixing errors by repeating the command and then blocked by pywebsocket. ~~Should we fork it?~~

Edit: Ah, according to https://github.com/web-platform-tests/wpt/issues/9599 the pywebsocket clone is already forked. Maybe it can be modified then :+1: 
Edit 2: Opened #11518.

### DIFF
--- a/tools/wptserve/wptserve/sslutils/__init__.py
+++ b/tools/wptserve/wptserve/sslutils/__init__.py
@@ -1,4 +1,3 @@
-# flake8: noqa (not ideal, but nicer than adding noqa: F401 to every line!)
 from .base import NoSSLEnvironment
 from .openssl import OpenSSLEnvironment
 from .pregenerated import PregeneratedSSLEnvironment
--- a/tools/wptserve/wptserve/sslutils/__init__.py
+++ b/tools/wptserve/wptserve/sslutils/__init__.py
@@ -1,9 +1,7 @@
 # flake8: noqa (not ideal, but nicer than adding noqa: F401 to every line!)
-import openssl
-import pregenerated
-from base import NoSSLEnvironment
-from openssl import OpenSSLEnvironment
-from pregenerated import PregeneratedSSLEnvironment
+from .base import NoSSLEnvironment
+from .openssl import OpenSSLEnvironment
+from .pregenerated import PregeneratedSSLEnvironment
 
 environments = {"none": NoSSLEnvironment,
                 "openssl": OpenSSLEnvironment,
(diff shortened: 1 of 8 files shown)


saschanaz:
Ah, according to https://github.com/web-platform-tests/wpt/issues/9599 the pywebsocket clone is already forked. Maybe it can be modified then :+1:

gsnedders:
@Ms2ger had also just done a fair bit on getting wptserve running under Py3, though I don't know what the status of that is.

tools/wptserve/wptserve/sslutils/__init__.py
```diff
@@ -1,9 +1,7 @@
 # flake8: noqa (not ideal, but nicer than adding noqa: F401 to every line!)
```
Ms2ger: @gsnedders you added this; do you remember why you needed to keep the openssl/pregenerated imports?

Either we should remove the noqa annotation, or we should keep the imports. (I'd merge immediately if you did the latter; need more info to approve the former.)

Ms2ger (changes requested):
LGTM besides one issue.

Ms2ger:
My code is in https://github.com/Ms2ger/web-platform-tests/tree/iteritems-testfiles and https://github.com/Ms2ger/web-platform-tests/tree/py3-resources-test ; feel free to pick up changes from there.

tools/wptserve/wptserve/sslutils/__init__.py
gsnedders: I have no idea. I was trying to change as little as possible.

gsnedders:
How practical would it be to remove the `wptserve`/`serve` exceptions from `tools/conftest.py` to make sure we don't regress this? Even if we have to mark almost every test as xfail on Py3!

tools/wptserve/wptserve/sslutils/__init__.py
saschanaz: >was trying to change as little as possible.

It seems okay to merge then 😁

Ms2ger: I offered two options in my comment; please implement one of them.

saschanaz: Done!

Ms2ger (approved):

gsnedders:
Interestingly `test_make_hosts_file_*` still fail despite the command working!

gsnedders:
Uh, no, that's wrong. I… ran that using Py2.

gsnedders:
#11669 makes us run tests on Py3.

PR #11518, opened 2018-06-15
Fix py3 syntax/import errors on pywebsocket

saschanaz:
#9599 says the wpt clone of pywebsocket is forked so maybe we can do some Py3 patch.

This is a partial fix that does not fix everything.

### DIFF
--- a/tools/pywebsocket/mod_pywebsocket/standalone.py
+++ b/tools/pywebsocket/mod_pywebsocket/standalone.py
@@ -802,7 +802,7 @@ class WebSocketRequestHandler(CGIHTTPServer.CGIHTTPRequestHandler):
 
         # If any Exceptions without except clause setup (including
         # DispatchException) is raised below this point, it will be caught
-        # and logged by Websocketserver.
+        # and logged by WebSocketServer.
 
         try:
             try:
--- a/tools/pywebsocket/mod_pywebsocket/standalone.py
+++ b/tools/pywebsocket/mod_pywebsocket/standalone.py
@@ -143,7 +143,7 @@ configuration file.
 THREADING
 =========
 
-This server is derived from SocketServer.ThreadingMixIn. Hence a thread is
+This server is derived from socketserver.ThreadingMixIn. Hence a thread is
 used for each request.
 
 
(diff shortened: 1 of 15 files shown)


tools/pywebsocket/mod_pywebsocket/standalone.py
```diff
@@ -348,8 +348,8 @@ def recv(self, bufsize, flags=0):
 
         try:
             return self._connection.recv(bufsize)
-        except OpenSSL.SSL.SysCallError, (err, message):
```
Ms2ger: I have been unable to figure out what this syntax meant; would appreciate a second look from someone who knows.

tools/pywebsocket/mod_pywebsocket/standalone.py
```diff
@@ -795,14 +795,14 @@ def parse_request(self):
...
-        except dispatch.DispatchException, e:
+        except dispatch.DispatchException as e:
             self._logger.info('Dispatch failed for error: %s', e)
             self.send_error(e.status)
             return False
 
         # If any Exceptions without except clause setup (including
         # DispatchException) is raised below this point, it will be caught
-        # and logged by WebSocketServer.
+        # and logged by Websocketserver.
```
Ms2ger: Revert

tools/pywebsocket/mod_pywebsocket/util.py
```diff
@@ -68,7 +68,7 @@ def get_stack_trace():
     TODO: Remove this when we only support Python 2.4 and above.
```
Ms2ger: (Hah)

Ms2ger (approved):
That's a lot of exceptions!

tools/pywebsocket/mod_pywebsocket/standalone.py
saschanaz: Python 2 exceptions are like tuples:

```
Exception(0,1,2,3)[2]
// 2
```

Python 3 does not allow this and requires using `.args`, which gives a tuple.

Ms2ger (approved):

PR #11544, opened 2018-06-18
webkit: GTK port should enable acceptInsecureCerts capability

zdobersek:
When using the WebKit product to run the WPT tests, the GTK port of WebKit has
to properly handle the temporary testing-purpose certificate that's used for
loads over SSL/TLS and which can't be validated otherwise.

To allow running tests served over HTTPS, the acceptInsecureCerts capability is
set in the capabilities dict that's then used for the Selenium-based executor.
This is only done for the tesharness tests and reftests but not for the wdspec
tests as that would collide with the tests that are testing this capability.

### DIFF
--- a/tools/wptrunner/wptrunner/browsers/webkit.py
+++ b/tools/wptrunner/wptrunner/browsers/webkit.py
@@ -30,7 +30,7 @@ def browser_kwargs(test_type, run_info_data, **kwargs):
             "webdriver_args": kwargs.get("webdriver_args")}
 
 
-def capabilities_for_port(webkit_port, binary, binary_args):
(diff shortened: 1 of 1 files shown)


andreastt:
@gsnedders Can you look at this?

Technically this patch is fine, but it’s unclear to me how the TLS certificates are meant to work in WPT. Setting this capability will implicitly accept any self-signed or insecure TLS certificates, which may not be what we want?

gsnedders:
We should strongly prefer some way of getting the certificates loaded; we currently pass acceptInsecureCerts with Firefox when no `--certutil-binary` is passed.

zdobersek:
Whitelisting approach taken up in https://github.com/web-platform-tests/wpt/pull/11615.

PR #9755, opened 2018-03-02
Add the InputDeviceCapabilities IDL file with test

lukebjerring:

PR #11549, opened 2018-06-18
Add input-device-capabilities/META.yml

zcorpan:


RByers:
What's the difference between this and the [OWNERS file that already exists](https://github.com/web-platform-tests/wpt/blob/master/input-device-capabilities/OWNERS)?  See #9755 for the discussion of who should be in that.

zcorpan:
Oh, we recently changed OWNERS to META.yml. I can change this to remove the OWNERS file and move over the names.

@RByers would you prefer to be in or out? :)

RByers (approved):
Thanks. LGTM, no need to add me.

PR #11551, opened 2018-06-18
Test setResourceTimingBufferSize event order when set to current buffer size

yoavweiss:
This is to test https://github.com/w3c/resource-timing/issues/96

I integrated @JosephPecoraro's test from the issue and turned it into a promise test. It seems like Chrome is passing it while Safari and Firefox are not.

@JosephPecoraro - Can you PTAL to make sure I didn't botch your test in some way? On the issue, you had different results.


yoavweiss:
Closing this, as I'll try to land a revised version of this test at https://chromium-review.googlesource.com/c/chromium/src/+/1200863

PR #11557, opened 2018-06-18
Run update-built-tests.sh to update built tests

Hexcles:
PR #11362 is failing tools/ci/ci_built_diff.sh which leads me to notice that 2dcontext/ needs to be updated in master.

Disclaimer: I don't really know about the tests. I just ran the script.


Hexcles:
Never mind... The four PNG files are actually ignored by `tools/ci/ci_built_diff.sh` and are not the cause of the failure. Sorry.

PR #11566, opened 2018-06-19
service worker: skipWaiting() should resolve after activation.

mfalken:
See https://github.com/w3c/ServiceWorker/issues/1187 and
https://github.com/w3c/ServiceWorker/pull/1327.


wanderview:
Jake had some questions that probably should be resolved before reviewing this: https://github.com/w3c/ServiceWorker/issues/1187#issuecomment-398385915

PR #11570, opened 2018-06-19
[wptrunner] Reject tests that almost time out in stability check

### LINKED ISSUE #9972: Prevent slow tests from landing
Chromium developers are spending a lot of time triaging test flakiness caused by tests that are simply too slow. I just triaged https://bugs.chromium.org/p/chromium/issues/detail?id=820334 and https://bugs.chromium.org/p/chromium/issues/detail?id=815286 back-to-back.

As a rule of thumb, a test can take 10 times longer on a bot than it takes to run locally.

It would be nice if the CI check rejected new tests that _do_ pass, but take too long to do so. Sometimes the only reliable way to detect failure is via a timeout, so I'm not proposing outlawing tests that timeout.

zcorpan:
Fixes #9972.

~TODO: allow longer time for timeout=long tests.~

### DIFF
--- a/tools/wptrunner/wptrunner/stability.py
+++ b/tools/wptrunner/wptrunner/stability.py
@@ -73,6 +73,10 @@ class LogHandler(reader.LogHandler):
 
         return subtest
 
+    def test_start(self, data):
+        test = self.find_or_create_test(data)
+        test["start_time"] = data["time"]
+
     def test_status(self, data):
         subtest = self.find_or_create_subtest(data)
         subtest["status"][data["status"]] += 1
@@ -82,6 +86,9 @@ class LogHandler(reader.LogHandler):
     def test_end(self, data):
         test = self.find_or_create_test(data)
         test["status"][data["status"]] += 1
+        start_time = test.pop("start_time")
+        test["duration"] = data["time"] - start_time
+        test["timeout"] = data["extra"]["test_timeout"]
 
 
 def is_inconsistent(results_dict, iterations):
@@ -97,13 +104,16 @@ def process_results(log, iterations):
     handler = LogHandler()
     reader.handle_log(reader.read(log), handler)
     results = handler.results
+    longest_duration = 0
     for test_name, test in results.iteritems():
         if is_inconsistent(test["status"], iterations):
             inconsistent.append((test_name, None, test["status"], []))
         for subtest_name, subtest in test["subtests"].iteritems():
             if is_inconsistent(subtest["status"], iterations):
                 inconsistent.append((test_name, subtest_name, subtest["status"], subtest["messages"]))
-    return results, inconsistent
+        if test["duration"] > longest_duration:
+            longest_duration = test["duration"]
+    return results, inconsistent, longest_duration, test["timeout"]
 
 
 def err_string(results_dict, iterations):
@@ -202,8 +212,8 @@ def run_step(logger, iterations, restart_after_iteration, kwargs_extras, **kwarg
     logger._state.suite_started = False
 
     log.seek(0)
-    results, inconsistent = process_results(log, iterations)
-    return results, inconsistent, iterations
+    results, inconsistent, longest_duration, timeout = process_results(log, iterations)
+    return results, inconsistent, longest_duration, timeout, iterations
 
 
 def get_steps(logger, repeat_loop, repeat_restart, kwargs_extras):
@@ -242,6 +252,8 @@ def write_summary(logger, step_results, final_result):
 
     logger.info(':::')
 
+def write_duration(logger, longest_duration):
+    logger.info('::: Longest duration was %sms.' % longest_duration)
 
 def check_stability(logger, repeat_loop=10, repeat_restart=5, chaos_mode=True, max_time=None,
                     output_results=True, **kwargs):
@@ -264,7 +276,7 @@ def check_stability(logger, repeat_loop=10, repeat_restart=5, chaos_mode=True, m
         logger.info(':::')
         logger.info('::: Running test verification step "%s"...' % desc)
         logger.info(':::')
-        results, inconsistent, iterations = step_func(**kwargs)
+        results, inconsistent, longest_duration, timeout, iterations = step_func(**kwargs)
         if output_results:
             write_results(logger.info, results, iterations)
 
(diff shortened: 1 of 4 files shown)


zcorpan:
I was hoping @gsnedders or someone could help me figure out how to check for `timeout=long`

zcorpan:
Let's use 30s for now.

gsnedders:
> I was hoping @gsnedders or someone could help me figure out how to check for timeout=long

I'm not sure there's any easy way to get at that from here, because if I'm not mistaken we've not got as far as loading the manifest at the point of running the stability entry point.

Hexcles:
We almost want to count the longest duration for tests with normal timeout and tests with long timeout separately (instead of using different threshold depending on what the slowest test is), because otherwise the longest running test is very likely one with long timeout and we miss those tests with normal timeout but perhaps closer to timeout.

As for the implementation, @gsnedders is right; it's very hard to consult the manifest here. I'm thinking perhaps the wptrunner can add the timeout of the test as an additional field to its `test_end` messages. This field should be ignored by the other log handlers (to avoid bloating the log or report), but can be used by the stability check.

zcorpan:
>  I'm thinking perhaps the wptrunner can add the timeout of the test as an additional field to its `test_end` messages. 

I've tried to poke around a bit, but can't figure out how to do this. Any pointers?

zcorpan:
Thanks, I think I got it working (but doesn't do anything with `timeout_multiplier` unless that's already baked in).

Hexcles (approved):
Cool! I'm giving my r+, but please also wait for opinions from others.

zcorpan:
@Hexcles can you merge this if there has been no review by June 29? Also, if there *are* review comments, can you address them, please?

gsnedders:
I'd like to see what @jgraham has to say, but he's on holiday till after then IIRC.

zcorpan:
Ok to wait for James.

Hexcles:
@zcorpan alright, so let's wait for more reviews. If it's something easy to fix, I'll help you address the comments and land the PR when you're away. Otherwise, I'll wait till you're back.

Meanwhile, could you take a look at the high-level discussions in #9972 first?

jgraham:
I haven't reviewed this because I'm away, but as noted on the original issue, I'm pretty skeptical of failing travis because of long-running tests. In particular it's hostile to tests for features that aren't implemented in any browser we run in CI which are more likely to time out.

I think we should discuss some mechanism to warn for this kind of thing rather than hard-failing.

ricea:
Indeed, "Failing by timeout" is sometimes the most natural way to write a test (for example, a fetch abort test). I was assuming that reviewers would be able to apply their judgement in these cases.

zcorpan:
Hmm I didn't intend do fail the check for TIMEOUT.

Hexcles:
So to clarify (based on my understanding):

* Tests that consistently time out won't fail the check.
* Tests that time out sometimes will fail the check (which is already the case).
* (NEW) Tests that never timeout (during the 10 reruns) but are very close to timeout will now fail as they are likely to be flaky.

And I think this is a reasonable approach to prevent flakiness with minimal cost of productivity.

foolip:
@jugglinmike, while @zcorpan is away, can you put this on the Bocoup list of things to potentially prioritize?

Hexcles:
@foolip I reviewed the code in detail and am reasonably familiar with related code. I agreed to help @zcorpan to rebase, fix nits and land this PR when he's away. I think this is now still awaiting more review. Ping @gsnedders @jgraham

jugglinmike:
Very good. I'll let you take it from here :)

zcorpan:
I think this has two outstanding todos.

* move logic to test_start
* don't fail for consistent TIMEOUT.

tools/wptrunner/wptrunner/testrunner.py
```diff
@@ -582,6 +582,10 @@ def test_ended(self, test, results):
         if status == "CRASH":
             self.browser.log_crash(test.id)
 
+        if not file_result.extra:
+            file_result.extra = { "test_timeout": test.timeout }
+        else:
+            file_result.extra.test_timeout = test.timeout
```
jgraham: Does this work? I would expect it to be `file_result.extra["test_timeout"]` unless `extra` isn't a dictionary for some reason.

jgraham (changes requested):

tools/wptrunner/wptrunner/stability.py
```diff
@@ -91,9 +106,18 @@ def is_inconsistent(results_dict, iterations):
     return len(results_dict) > 1 or sum(results_dict.values()) != iterations
 
 
+def is_slow(test):
+    # We are interested in tests that almost time out (i.e. likely to be
+    # flaky). Tests that consistently time out are not considered "slow".
+    if test["status"].keys() == ['TIMEOUT']:
```
jgraham: I think `CRASH` should also be excluded because we can change the status from `TIMEOUT` to `CRASH` if we detect evidence of a crash after the test finished. In general I wonder if testing 

Also, it seems arguable that the handling of a test that sometimes times out and sometimes doesn't is wrong; if we get `PASS`, `PASS`, `PASS`, `TIMEOUT`, `PASS` then we'll call the test slow. But if the passes take 1s and the timeout is 10s then we are probably seeing a race condition that causes the test to hang sometimes, and the obvious solution to a slow test (Add a longer timeout) isn't going to help. So I think we should either store the duration per result and filter here, or only update the duration when we see a non-TIMEOUT, non-CRASH result. The latter is easier, but the former could be used to provide more detailed logging if we want.

jgraham (changes requested):

tools/wptrunner/wptrunner/stability.py
Hexcles: You're right about `CRASH`.

As for the second example (inconsistent results), it won't actually happen as we call `is_inconsistent` before `is_slow`, so such tests will be considered as flaky instead of slow. But I do see the code can be confusing and error-prone as the prerequisite isn't obvious. I'll tweak it a bit.

Hexcles:
@jgraham I think I addressed your comment this time. PTAL at `find_slow_status` (the new version of `is_slow`) in particular.

And you can see stability checks rejecting the example slow test in action:
https://travis-ci.org/web-platform-tests/wpt/jobs/422774773#L764
https://travis-ci.org/web-platform-tests/wpt/jobs/422774775#L686

tools/wptrunner/wptrunner/stability.py
```diff
@@ -91,9 +107,29 @@ def is_inconsistent(results_dict, iterations):
...
+    We are interested in tests that almost time out (i.e. likely to be flaky).
+    Therefore, timeout statuses are ignored, including (EXTERNAL-)TIMEOUT &
+    CRASH (tests that time out may be marked as CRASH if crashes are detected).
+
+    Returns:
+        A result status produced by a run that almost times out; None, if no
+        runs almost time out.
+    """
+    threshold = test["timeout"] * FLAKY_THRESHOLD
+    for status in ['PASS', 'FAIL', 'OK', 'ERROR', 'INTERNAL-ERROR']:
```
jgraham: I don't think we can actually get `INTERNAL-ERROR` as a status here; it's alwasy converted to `ERROR`.

tools/wptrunner/wptrunner/stability.py
```diff
@@ -91,9 +107,29 @@ def is_inconsistent(results_dict, iterations):
...
+        if (status in test["longest_duration"] and
+                test["longest_duration"][status] > threshold):
```
jgraham: Weird indentation (yeah I know pyflakes likes it, but I think it's a bad rule which we don't enforce).

jgraham (approved):

PR #11581, opened 2018-06-19
SVG: Check that required properties are supported

ewilligers:
https://svgwg.org/svg2-draft/styling.html#RequiredProperties
The following properties must be supported by all SVG user agents:
...

We check that each property is supported.

### DIFF
--- a/svg/styling/required-properties.svg
+++ b/svg/styling/required-properties.svg
@@ -1,0 +1,147 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg"
+     xmlns:h="http://www.w3.org/1999/xhtml">
+  <metadata>
+    <h:link rel="help" href="https://svgwg.org/svg2-draft/styling.html#RequiredProperties"/>
+    <h:link rel="help" href="https://svgwg.org/svg2-draft/propidx.html"/>
+    <h:meta name="assert" content="All required properties are supported"/>
+    <!-- Note: This test does not verify that the properties are actually applied to SVG rendering. -->
+  </metadata>
+  <g id="target"></g>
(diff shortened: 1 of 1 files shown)


AmeliaBR:
This test isn't really an effective test, because in user agents that support both HTML and SVG (AKA, web browsers), it will only test that the CSS parser recognizes the property, not that it applies the properties to SVG rendering.  E.g., some browsers (thankfully, no longer any of the latest versions) will recognize `transform` as a CSS property name, but not apply it to an SVG element. Even still, most browsers won't apply it to patterns and gradients.

It's a first step, but not sufficient.

Also, this is probably an SVG spec issue: but why is `z-index` not in either of those lists?

ewilligers:
I have added a Note with your concern about the limitations of this test.

The test is effective in identifying properties that an implementation has no support for, e.g. Blink gave
```
FAIL Property "font-variant-position" is supported assert_true: expected true got false
FAIL Property "glyph-orientation-vertical" is supported assert_true: expected true got false
FAIL Property "text-align-all" is supported assert_true: expected true got false
FAIL Property "text-space-collapse" is supported assert_true: expected true got false
```

The test also doesn't check that implementations support the full grammar for each property. I have been writing CSS parsing/serialization tests for all the CSS specs referenced, and contributing them to the relevant CSS test suites.

svgeesus (approved):
With the caveat that this only tests the property is known, not that it gets applied to SVG content, this basic test LGTM

ewilligers:
`Also, this is probably an SVG spec issue: but why is `z-index` not in either of those lists?`

I have uploaded a [test](https://github.com/web-platform-tests/wpt/pull/11583). If z-index has SVG implementation support, we can add it to the list and the tests, otherwise we should remove the z-index section from the 2.0 draft.

PR #11583, opened 2018-06-20
SVG: Test that z-index controls rendering order

ewilligers:
https://svgwg.org/svg2-draft/render.html#ZIndexProperty
The z-index property allows an element to be assigned to a stack level.


ewilligers:
Resolution from 2011: https://www.w3.org/2011/03/03-svg-minutes.html#item03
Open Mozilla bug: https://bugzilla.mozilla.org/show_bug.cgi?id=360148

fsoder (approved):
Sorry for the delay, LGTM.

PR #11584, opened 2018-06-20
Add console timing manual tests

domfarolino:
Adds console timing manual tests. This, as well as https://github.com/web-platform-tests/wpt/blob/master/console/console-label-conversion.any.js should be enough to cover https://github.com/whatwg/console/pull/138.

### DIFF
--- a/console/console-timing-logging-manual.html
+++ b/console/console-timing-logging-manual.html
@@ -1,0 +1,55 @@
+<!DOCTYPE html>
+<html>
+<head>
+<title>Console Timing Methods - Logging Manual Test</title>
+<meta name="author" title="Dominic Farolino" href="mailto:domfarolino@gmail.com">
+<meta name="assert" content="Console timing methods">
+<link rel="help" href="https://console.spec.whatwg.org/#timing">
+</head>
+<body>
+<p>Open the console inside the developer tools. It should contain entries whose contents are:</p>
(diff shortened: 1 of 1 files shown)


console/console-timing-logging-manual.html
```diff
@@ -0,0 +1,55 @@
...
+
+console.time("default");
+console.timeLog("default");
+console.timeLog("default", "extra data");
+console.timeEnd("default");
+
+console.time({toString() {return "default"}});
+console.timeLog({toString() {return "default"}});
+console.timeLog({toString() {return "default"}}, "extra data");
+console.timeEnd({toString() {return "default"}});
```
domenic: It'd be good to test toString() conversion with a non-default label as well, to guard against implementations that fall back to "default" if typeof !== string

console/console-timing-logging-manual.html
```diff
@@ -0,0 +1,55 @@
...
+
+console.time("a label");
+console.timeLog("a label");
+console.timeLog("a label", "extra data");
+console.timeEnd("a label");
+
+console.timeEnd("b"); // should produce a warning
```
domenic: Same for timeLog?

domenic (approved):
LGTM, although you could improve clarity for reviewers by adding newlines in the HTML source to group the expected output in the same way that they are grouped in the JS.

PR #11585, opened 2018-06-20
SVG: transform properties affect rendering

ewilligers:
Test that transform, transform-box and transform-origin
affect rendering of SVG elements.
https://svgwg.org/svg2-draft/styling.html#RequiredProperties


fsoder (approved):
Sorry for the delay, LGTM.

PR #11600, opened 2018-06-21
[sensors] Bring setUpdateSensorReadingFunction from sensor mojo inter…

Honry:
…face

 to provide the mock values for sensors

### DIFF
--- a/accelerometer/Accelerometer.https.html
+++ b/accelerometer/Accelerometer.https.html
@@ -9,8 +9,25 @@
 <div id="log"></div>
 <script>
 
-runGenericSensorTests('Accelerometer');
-runGenericSensorTests('GravitySensor');
-runGenericSensorTests('LinearAccelerationSensor');
+const kDefaultReading = [1.12345, 2.12345, 3.12345];
+const kMappedReading = [-2.12345, 1.12345, 3.12345];
+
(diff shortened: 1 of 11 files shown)


Honry:
@rakuco, this brings `setUpdateSensorReadingFunction` method from blink's own [`sensor-helpers.js`](https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/http/tests/resources/sensor-helpers.js?q=sensor-helper&sq=package:chromium&g=0&l=137), which allows user customization of mock sensor's reading updates, for the purpose of verifying sensor readings exposed to Sensor API are expected.

Tests changes in this PR are referred to [sensor LayoutTests](https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/sensor/resources/generic-sensor-tests.js?q=generic-sensor&sq=package:chromium&g=0&l=1) with some modification, and I've verified it with latest chrome, all tests results are expected.

Changes in `generic-sensor-tests.js` involve all sensors tests, that's why it looks big. :( 
Sorry for the inconvenience. I will try to simplified it, but pls. don't expected there would be a big reduction.

Honry:
This is abandoned and replaced with #18886.

PR #11612, opened 2018-06-21
[compositing] Parsing of compositing properties

ewilligers:
Parsing/serialization tests for
- background-blend-mode
- isolation
- mix-blend-mode


dirkschulze (approved):
Tests look good to me.

Issue #9113, opened 2018-01-20
Write a cssharness.js


gsnedders:
We should have something that can take a CSS property definition (in some format, probably JSON?) and generate a lot of tests based upon it, similar to what we have with idlharness.js.

We have a lot of tests that could be auto-generated from the property definition, especially in the CSS2 testsuite (and these are mostly reftests, which probably costs us a fair bit of time when we could test behaviour through the CSSOM much more quickly en-mass).

Note the one awkward thing here is going to be computing what the computed value will be for any given specified value, but I *hope* we can do that with relative sanity!

 * @tabatkins: Bikeshed uploads propdef info to Shepherd, right? Is this documented anywhere, as to how to get it? What about typedef info?
 * @foolip: WDYT?

emilio:
FWIW Gecko has a `property_database.js` doing something of the sort:

  https://searchfox.org/mozilla-central/source/layout/style/test/property_database.js

Not perfect of course, but maybe a good starting point?

emilio:
Automatic syntax tests from Bikeshed would be awesome, btw.

gsnedders:
@emilio Yeah, I was looking over `property_database.js`, because for some reason I thought it generated the values, rather than having its long explicit lists.

gsnedders:
FWIW: given the number of CSS2 tests we have that essentially test property parsing, I think the MVP here is something that can replace those, and that avoids some of the complexity from more recent specs.

foolip:
I've wanted this too, if it's not too hard we should just do it. Like IDL, probably split per spec. @tabatkins, if one wants to extract all CSS syntax stuff (properties, values, pseudo-things, etc.) would one start from the Bikeshed source, or get it from shepherd?

guest271314:
Interestingly just encountered a bug (unless misinterpreting the specification) for CSS `url()` function implementation at both Chromium and Firefox. According to the specification

> [3.4.2. Empty URLs](https://drafts.csswg.org/css-values-3/#url-empty)
> 
> If the value of the url() is the empty string (like url("") or url()), the url must resolve to an invalid resource (similar to what the url about:invalid does).

where  at HTML

`<div id="test-empty-string-at-css-url-function" style="background-image:url('')">div</div>`

results in `location.href` being set as the argument to the function

```
console.assert(
  !new RegExp(location.href)
  .test(
    window.getComputedStyle(
      document.getElementById("test-empty-string-at-css-url-function")
    ).backgroundImage)
[...]

PR #11613, opened 2018-06-21
[css-writing-modes] Parsing of writing modes properties

ewilligers:
Parsing/serialization tests for
- direction
- text-combine-upright
- text-orientation
- unicode-bidi
- writing-mode

### DIFF
--- a/css/css-writing-modes/parsing/direction-invalid.html
+++ b/css/css-writing-modes/parsing/direction-invalid.html
@@ -1,0 +1,19 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>CSS Writing Modes Level 3: parsing direction with invalid values</title>
+<link rel="author" title="Eric Willigers" href="mailto:ericwilligers@chromium.org">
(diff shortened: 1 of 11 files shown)


upsuper:
While I think the tests look fine, I may argue that there is a better way to test them.

Firefox has a big unified [property_database](https://hg.mozilla.org/mozilla-central/file/tip/layout/style/test/property_database.js), and there are several tests around it to check different aspect of the property behaviors, including parsing-serialization round trip, parsing-computation-serialization round trip, inheriting behavior, shorthand expansion, etc.

I would argue that, that approach might be a better way than writing small tests for each property for each behavior, since it makes it easier to add new properties into the test set, and add new test for some behavior.

emilio (approved):
Yeah, tests looks fine.

The problem with the property database is that the whole test fails if any single property is not supported by the browser, so having multiple tests seems fine to me. Maybe we could auto-generate the tests though? https://github.com/web-platform-tests/wpt/issues/9113 tracks how to improve the situation here.

But I don't see a reason to avoid merging this for now I guess.

PR #11614, opened 2018-06-21
Prevent abort error from breaking test in FF (fixes #10792)

### LINKED ISSUE #10792: /IndexedDB/idbdatabase-createObjectStore-exception-order.htm is disabled in Mozilla and flaky in WebKit

-- | -- | -- | --
/IndexedDB/idbdatabase-createObjectStore-exception-order.htm | mozilla webkit | disabled [ Pass Failure ] | https://bugzilla.mozilla.org/show_bug.cgi?id=1336639 https://webkit.org/b/172243

inexorabletash:
The test intentionally aborts a versionchange transaction, which results in an AbortError being fired at the IDBOpenDBRequest. In browsers that bubble event errors up to `window.error` (i.e. Firefox) this will cause the harness to fail.

Call preventDefault() on the error since it's expected.

### DIFF
--- a/IndexedDB/idbdatabase-createObjectStore-exception-order.htm
+++ b/IndexedDB/idbdatabase-createObjectStore-exception-order.htm
@@ -8,8 +8,12 @@
 <script>
 
 indexeddb_test(
-  (t, db, txn) => {
+  (t, db, txn, rq) => {
     db.createObjectStore('s');
+
+    // Acknowledge the error, to prevent window.error from firing in
+    // browsers that implement that.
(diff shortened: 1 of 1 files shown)


inexorabletash:
Apologies for pushing this branch to web-platform-tests/wpt itself rather than a clone; I forgot to update the origin for my new checkout when the repo moved. :( But I figured it would be noisier to delete this and recreate it - happy to do so if needed, though.

zcorpan (approved):

zcorpan:
Thanks! No problem with creating branches here.

PR #11615, opened 2018-06-21
webkit: GTK port should whitelist the web-platform.test host certificate

zdobersek:
For testing purposes, the TLS certificate for the web-platform.test
host should be whitelisted for the GTK port of WebKit in order to
avoid TLS errors when loading test cases via HTTPS.

This is done by listing the host and certificate file path pair in
the port-specific capabilities. This pair is then used for
whitelisting the certificate for the given host in the
WebDriver-automated browsing session.


zdobersek:
@gsnedders @jgraham review/feedback would be appreciated

gsnedders (approved):

PR #11630, opened 2018-06-22
[idlharness.js] Wrap inheritance chain resolution in tests

lukebjerring:
In a similar vein to https://github.com/web-platform-tests/wpt/pull/11419, this PR reduces the severity of incorrect IDL setup in idlharness tests by making sub-tests for each inheritance/implements resolution check, which fail softly.

Chromium CI for equivalent CL:
https://test-results.appspot.com/data/layout_results/linux_chromium_rel_ng/125105/site_per_process_webkit_layout_tests%20%28with%20patch%29/layout-test-results/results.html

In generally, this adds N passing subtests (where N is the number of interfaces that inherit something).
In cases with harness errors, this fixes broken tests (e.g. html/dom/interfaces.worker.html) and adds all the results of the remaining tests which manage to execute (which is the goal of the change).

@jgraham - wanted to double check that this won't have severe negative impact downstream. Chromium will just take the new results as source-of-truth, an since it only adds PASS results, won't have any bugs filed.


lukebjerring:
Re: Chromium: I literally clicked expand on every single result (82 "regressions") and here's the only one that wasn't just N extra `PASS` results:

image

Hexcles:
I had some in-depth discussions with Luke about this change. In general I think it's a great idea to test the setup (dependency stack in this case) in generated tests like idlharness. It'd prevent some of the most trivial but easy-to-make mistakes and save a lot of time debugging them afterwards.

Luke was a bit concerned about adding a few subtests to all idlharness tests, and mentioned an alternative solution: hide the dependency stack checks by default and only show them when something is wrong, which will minimize the scale of impact. Yet I don't like the idea of dynamically changing the number of (sub)tests depending on some pass/failure conditions, and adding a few subtests to all idlharness tests isn't as bad as it may sound like (at least in Chromium infra, our importer can handle it without issues; but I'd like to hear from @jgraham @jdm if this would be a concern for Gecko & Servo).

Hexcles (approved):
Forgot to give my vote: LGTM.

But please do wait for others to chime in.

lukebjerring:
@foolip @jgraham @gsnedders - ping

resources/idlharness.js
```diff
@@ -841,7 +794,9 @@ IdlArray.prototype.collapse_partials = function()
 //@{
 {
     const testedPartials = new Map();
-    this.partials.forEach(function(parsed_idl)
+    this.partials
+        .sort((p1, p2) => p1.name > p2.name ? 1 : -1)
```
foolip: Unlike all the other added `sort()` bits in this PR, `this.partials` isn't a newly created object, and this line makes we wonder if it's used anywhere else before being sorted. Maybe sort it right after it's populated? Or just leave it in the order it is, since the order will depend on the IDL input and should be reliable?

foolip (approved):
Just one small nit.

lukebjerring:
Going to close this stale PR, and hope that a generic regression detection solution takes shape instead :)

PR #11631, opened 2018-06-22
Auto-update the html IDL file

lukebjerring:
Hello reviewer(s),

Please check that:
All tests which cover the IDL in the spec have been migrated to fetch + use the idl in the interfaces/ directory (instead of inline copies in the test files).


foolip:
@lukebjerring are the test changes from this worth salvaging?

lukebjerring:
Somewhat; they're actually out of date though (not using `idl_test`), so I'd just start from scratch with a cleanup.

PR #11640, opened 2018-06-23
Add a test for text-decoration in tables quirk

zcorpan:


quirks/text-decoration-doesnt-propagate-into-tables/quirks.html
```diff
@@ -0,0 +1,14 @@
+<title>text decoration doesn't propagate into tables quirk</title>
+<link rel=match href=text-decoration-doesnt-propagate-into-tables-ref.html>
```
annevk: This needs to point to quirks-ref.html I think. That's also why the build script failed.

annevk (commented):
This will also require a corresponding PR to the Quirks Mode standard I think to make it build again, right?

quirks/text-decoration-doesnt-propagate-into-tables/quirks.html
zcorpan: Oops. Fixed.

annevk (approved):

PR #11641, opened 2018-06-23
Refactor and split up quirks tests

zcorpan:
The unitless-length test was a flaky timeout in chromium.
https://bugs.chromium.org/p/chromium/issues/detail?id=646133


Hexcles (approved):
LGTM

PR #11645, opened 2018-06-24
Fix EUC-JP reference decoder

peteroupc:
The relevant step fixed by this pull request says "Return error."; thus, the rest of the process should continue with the next iteration, rather than run the rest of the handler for the given byte.


inexorabletash (approved):

peteroupc:
Superseded by #11707.

PR #11647, opened 2018-06-24
[css-box] Parsing tests for CSS Box properties

ewilligers:
Spec:
https://drafts.csswg.org/css-box-3/
    
Omits properties and values that are not yet supported by browsers.

### DIFF
--- a/css/css-box/parsing/overflow-valid.html
+++ b/css/css-box/parsing/overflow-valid.html
@@ -1,0 +1,35 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>CSS basic box model: parsing overflow with valid values</title>
+<link rel="author" title="Eric Willigers" href="mailto:ericwilligers@chromium.org">
+<link rel="help" href="https://drafts.csswg.org/css-box-3/#propdef-overflow">
+<meta name="assert" content="overflow supports the full grammar '[ visible | hidden | scroll | auto | no-display | no-content ]{1,2}'.">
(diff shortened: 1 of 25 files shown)


ewilligers:
In my [first commit](https://github.com/web-platform-tests/wpt/pull/11647/commits/332c508f04d1f9f91bd0ba67b24d95d36bdd410e), I wrote tests based on the spec, before checking implementations.

In my [second commit](https://github.com/web-platform-tests/wpt/pull/11647/commits/c2307ae50fd2d464a0d76641e9d4b2c1a608436e), I have ordered each property's values into those supported by all browsers, those supported by some browsers, and those supported by no browsers. Where a value is supported by no browsers, I have commented out the test. Where a property is supported by no browsers, I have deleted the test.

css/css-box/parsing/overflow-valid.html
```diff
@@ -16,19 +16,20 @@
...
-test_valid_value("overflow", "no-display no-content");
 
 test_valid_value("overflow-x", "visible");
 test_valid_value("overflow-x", "hidden");
 test_valid_value("overflow-x", "scroll");
 test_valid_value("overflow-y", "auto");
-test_valid_value("overflow-y", "no-display");
-test_valid_value("overflow-y", "no-content");
+
+// The following are not yet supported by browsers:-
```
emilio: nit: Why the `-` to the right of the colon?

emilio (approved):
Thanks. Maybe worth squashing the two commits? No big deal in any case.

PR #11648, opened 2018-06-24
Fix ISO-2022-JP reference implementation

peteroupc:
The relevant step in the Encoding Standard says "Prepend _lead_ and _byte_ to _stream_." However, the two bytes are prepended in the wrong order in the reference implementation.  Note that under the Encoding Standard, "[w]hen one or more tokens are prepended to a stream, those tokens must be inserted, _in given order_, before the first token in the stream."  (Note that the current reference implementation, at the time of this request, moves _lead_ to the front of the array, then moves _byte_ to the front of the array.)

There may be other issues like this elsewhere in the multiple-byte encoder reference implementations.

### DIFF
--- a/encoding/legacy-mb-japanese/iso-2022-jp/iso2022jp-decoder.js
+++ b/encoding/legacy-mb-japanese/iso-2022-jp/iso2022jp-decoder.js
@@ -193,8 +193,8 @@ function iso2022jpDecoder(stream) {
                         continue;
                     }
                 }
-                bytes.unshift(lead);
                 bytes.unshift(byte);
+                bytes.unshift(lead);
                 outFlag = false;
                 decState = outState;
                 out += "�";


encoding/legacy-mb-japanese/iso-2022-jp/iso2022jp-decoder.js
```diff
@@ -193,8 +193,8 @@ function iso2022jpDecoder(stream) {
                         continue;
                     }
                 }
-                bytes.unshift(lead);
                 bytes.unshift(byte);
+                bytes.unshift(lead);
```
inexorabletash: This _might_ be more readable as:

`bytes.unshift(lead, byte);`

At least then the code order matches the spec order. On the other hand, I had to look up the behavior of `unshift()` with multiple arguments.

inexorabletash (approved):
Thanks for the fix. Do you know if this failure corresponded to failures in any of the browsers' CIs?

peteroupc:
Superseded by #11707.

PR #11649, opened 2018-06-24
Update Chrome WebDriver capabilities to use vendor prefix

jimevans:
The Chrome WebDriver implementation uses `goog:` for Chrome-specific
settings in its capabilites. The capabilites used in the current test
suite does not use this prefix. This commit resolves that issue.


jimevans:
@kereliuk You might want to take a look at this, and let me know if this is correct.

jgraham (approved):

PR #11650, opened 2018-06-24
Fix bug in EUC-KR reference implementation

peteroupc:
Makes one "if" statement conditional rather than unconditional


inexorabletash (approved):
Thanks.

(It would be nice to refactor all of these to include helpers such as inRange(), isASCIIByte(), etc.)

peteroupc:
Superseded by #11707.

PR #11651, opened 2018-06-25
Fix Shift_JIS reference implementation

peteroupc:
To conform to the most recent Encoding Standard.


encoding/legacy-mb-japanese/shift_jis/sjis-encoder.js
```diff
@@ -73,7 +73,7 @@ function sjisEncoder(stream) {
 		}
 		if (cp >= 0xff61 && cp <= 0xff9f) {
 			temp = cp - 0xff61 + 0xa1;
```
inexorabletash: Not new in this PR, but `var` declarations are missing for: `temp`, `offset`, `leadoffset`, `first`, `second`

inexorabletash (approved):

peteroupc:
Superseded by #11707.

PR #11652, opened 2018-06-25
Fix bug in Big5 reference implementation

peteroupc:
To conform to the Encoding Standard.

### DIFF
--- a/encoding/legacy-mb-tchinese/big5/big5-decoder.js
+++ b/encoding/legacy-mb-tchinese/big5/big5-decoder.js
@@ -62,7 +62,7 @@ function big5Decoder(stream) {
 			}
 			if (ptr == null) cp = null;
 			else cp = big5[ptr];
-			if (cp == null && byte >= 0x00 && byte < 0x7f) {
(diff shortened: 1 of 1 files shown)


encoding/legacy-mb-tchinese/big5/big5-decoder.js
```diff
@@ -46,7 +46,7 @@ function big5Decoder(stream) {
 			if ((byte >= 0x40 && byte <= 0x7e) || (byte >= 0xa1 && byte <= 0xfe))
 				ptr = (lead - 0x81) * 157 + (byte - offset);
 			// "If there is a row in the table below whose first column is pointer, return the two code points listed in its second column"
-			switch (ptr) {
+			switch (ptr + "") {
```
inexorabletash: I'd prefer to convert the `case` statements to use numbers rather than strings, unless there's some reason I'm missing.

inexorabletash (approved):
lgtm with a nit

peteroupc:
Superseded by #11707.

PR #11323, opened 2018-06-04
[css-contain] Improve references on test suite

mrego:
This patch is just a refactoring to reuse reference files and
name them better. Not having a reference file associated to
a particular test (e.g. contain-size-001) and referenced in many other.

This patch tries to reuse reference files from css/reference/
when possible.
In other cases it renames the reference files in reference/
when reused by several tests to make them generic.

This was suggested by @mstensho in a Chromium review:
https://chromium-review.googlesource.com/c/chromium/src/+/1080799

PR #11667, opened 2018-06-26
[css-contain] Use generic references for tests

mrego:
This patch tries to reuse reference files from css/reference/
when possible.

This is a split-off of the patch proposed in PR #11323.


mrego:
@frivoal ping, please could you take a look? Thanks.

frivoal (approved):
Sorry for the delay. Looks all good.

PR #11668, opened 2018-06-26
Add a test to verify that postMessage forwards user gesture.

fred-wang:
I *think* the spec says all versions of postMessage queue a task and so the triggered by user activation flag should be forwarded to the receiver.

At least with window's postMessage, the popup is allowed in all browsers.

I'm not sure if Firefox considers the user activation flag to block popups. It seems to allow them all when using ./wpt run with testdriver's click or blocking them all when running manually.
(--update: Dima opened https://bugzilla.mozilla.org/show_bug.cgi?id=1469422)

On Chromium, the popup is blocked for MessagePort and BroadcastChannel when you run them manually (as expected per https://bugs.chromium.org/p/chromium/issues/detail?id=851493) but not when using ./wpt run with testdriver's click.

On WebKit, BroadcastChannel is not supported ( https://bugs.webkit.org/show_bug.cgi?id=161472 ) and the popup is blocked for MessagePort ( https://bugs.webkit.org/show_bug.cgi?id=186593 ). I tested manually since I'm not sure about the status of ./wpt run or testdriver.

cc' @Ms2ger

### DIFF
--- a/webmessaging/postMessage-triggered-by-user-activation-from-subframe-manual.html
+++ b/webmessaging/postMessage-triggered-by-user-activation-from-subframe-manual.html
@@ -1,0 +1,62 @@
+<!DOCTYPE html>
+<title>postMessage's task triggered by user activation from a subframe</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<!-- This test is manual since we cannot currently test action triggered by user
+     activation with the test runner (blocking popups is always disabled).
+     Moreover, testdriver does not currently support clicking elements from a
+     subframe. -->
+<script src="support/postMessage-triggered-by-user-activation.js"></script>
+<link rel="help" href="https://html.spec.whatwg.org/multipage/web-messaging.html#dom-window-postmessage">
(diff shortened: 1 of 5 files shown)


mkruisselbrink:
Not sure how useful a test like this is, when as you mention running the tests using any of the automated test runners popup blocking is explicitly disabled anyway (as we have many many tests relying on being able to open windows without user gestures).... Not sure what better way there is to test this though. Probably requiring at least some kind of new functionality from the test runner side of things.

fred-wang:
> when as you mention running the tests using any of the automated test runners popup blocking is explicitly disabled anyway 

@mkruisselbrink I was not aware of that, now I understand better what I observed... So I guess for now, the best thing is to convert to a manual test for the WPT repo and/or just have automated version in the browser vendor repos. Anyway, I also wanted to make a stricter version where the buttons are also in an iframe but that's not currently possible with testdriver...

fred-wang:
@mkruisselbrink  I've converted the test to manual for now. And also added a version with a subframe. Hopefully they could be converted in the future or when importing to browser vendors' repositories. For now, the main interest would be to open the discussion of whether forwarding user gesture is the proper behavior as web engines are not consistent here.

PR #11673, opened 2018-06-26
Remove unsupported :has() pseudo-class from Element-closest.html

### LINKED ISSUE #11316: Element-closest.html `:has`-test fails in all browsers
The last test in [/dom/nodes/Element-closest.html](https://github.com/web-platform-tests/wpt/blob/master/dom/nodes/Element-closest.html) makes use of the relational `:has()` pseudo-class which has not been added by any major implementation to date, causing it to [fail in all browsers](https://wpt.fyi/results/dom/nodes/Element-closest.html).

Until there are signs that it will see widespread support, it seems more appropriate to remove it from the general test of the `closest()`-method.

Zirro:
The `:has()` pseudo-class has yet to be supported by any major browser, causing this test to always fail. Until there are signs that it actually will gain support, it seems more appropriate to remove it from the general test of the `closest()`-method.

Resolves: https://github.com/web-platform-tests/wpt/issues/11316

### DIFF
--- a/dom/nodes/Element-closest.html
+++ b/dom/nodes/Element-closest.html
@@ -59,7 +59,6 @@
   do_test(":scope"                     , "test4",  "test4");
   do_test("select > :scope"            , "test4",  "test4");
   do_test("div > :scope"               , "test4",  "");
-  do_test(":has(> :scope)"             , "test4",  "test3");
 function do_test(aSelector, aElementId, aTargetId) {
   test(function() {
     var el = document.getElementById(aElementId).closest(aSelector);


gsnedders:
We shouldn't just remove a test for a new feature because nobody has implemented it (and one failing subtest shouldn't be causing anyone any problems); doing so rather undermines the whole concept of landing tests with spec changes.

We should either:

 * Move the failing subtests into a new file (though I don't like the precedent this sets, given there's an argument we should move lots of tests then, and for things like `idlharness` this is unworkable),
 * File a bug to remove it from the spec due to lack of implementer interest and remove the test (or add a historical test).

Zirro:
Of the two options I favour the latter, as I'd argue that spec changes and tests shouldn't be merged before there is known implementer interest. Since it's already there however I'm not inclined (nor do I have the time) to pursue the removal of a potentially useful feature from the spec itself, given the potential arguments it would involve. I'd rather see it get implemented in that case.

PR #11674, opened 2018-06-26
Auto-update the image-capture IDL file

lukebjerring:
Renames the `mediacapture-image` IDL to be `image-capture`.

Also fixes the IDL test, which currently `ERROR`s for everything except chrome:
https://wpt.fyi/results/mediacapture-image?sha=d093f88ec3


mediacapture-image/idlharness.html
```diff
@@ -17,23 +17,28 @@
...
+    const srcs = ['image-capture', 'mediacapture-main', 'html', 'dom'];
     const [idl, main, html, dom] = await Promise.all(
       srcs.map(i => fetch(`/interfaces/${i}.idl`).then(r => r.text())));
 
-    var canvas = document.getElementById('canvas');
-    var context = canvas.getContext("2d");
-    context.fillStyle = "red";
-    context.fillRect(0, 0, 10, 10);
-    var track = canvas.captureStream().getVideoTracks()[0];
+    let capturer;
```
reillyeon: This variable isn't being assigned into.

reillyeon (changes requested):

mediacapture-image/idlharness.html
lukebjerring: Interesting. I missed this because there's actually currently a typo (`ImageCapturer`) which is the wrong interface name - corrected that too.

https://wpt.fyi/results/mediacapture-image/idlharness.html?sha=d093f88ec3

There's now an extra 7 tests (all passing in Chrome).

reillyeon (approved):

PR #9674, opened 2018-02-26
Update assumptions test to match HTML spec

gsnedders:
Per HTML, div defaults to `unicode-bidi: isolate` (except under ISO-8859-8!).

This makes the test pass in Firefox, though makes it regress in Chromium ([issue 296863](https://bugs.chromium.org/p/chromium/issues/detail?id=296863)).

PR #11677, opened 2018-06-26
Loosen assumptions so actual implementations match them

gsnedders:
This assumption comes from the CSS test suites:

> The X/HTML div element is assigned display: block; and no other property declaration.

Note this already contradicted the informative appendix D in CSS 2.1:

```
div   { unicode-bidi: embed }
```

HTML now contains:

```
div { unicode-bidi: isolate }
```

We changed the test for the assumption to match HTML in 9ff5d62fc664214b86ec8ef4b07aaa8097f7ebed (#9674), but didn't change the documented assumption.

This changes both to allow *any* value of `unicode-bidi`, because both [Blink](https://bugs.chromium.org/p/chromium/issues/detail?id=296863) and [WebKit](https://bugs.webkit.org/show_bug.cgi?id=65617) have `unicode-bidi: initial` (i.e., `normal`) and we want tests to be runnable in all major browsers, so we can't assume things they don't do.

We could also alternatively allow `unicode-bidi` to be `isolate` or `normal`? (Probably not `embed`, because AFAICT no impl does so?)

### DIFF
--- a/docs/_writing-tests/assumptions.md
+++ b/docs/_writing-tests/assumptions.md
@@ -16,8 +16,13 @@ tests can freely rely on these assumptions being true:
  * The user stylesheet is empty (except where indicated by the tests).
  * The device is interactive and uses scroll bars.
  * The device has the Ahem font installed.
- * The HTML `div` element is assigned `display: block;` and no other
-   property declaration.
+ * The HTML `div` element is assigned `display: block;`, the
+   `unicode-bidi` property may be declared, and no other property
+   declarations.
+   <!-- unicode-bidi: isolate should be required; we currently don't
+   assume this because Chrome and Safari are yet to ship this: see
+   https://bugs.chromium.org/p/chromium/issues/detail?id=296863 and
+   https://bugs.webkit.org/show_bug.cgi?id=65617 -->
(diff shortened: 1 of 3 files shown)


gsnedders:
Note WebKit will still fail the assumptions test because of https://bugs.webkit.org/show_bug.cgi?id=187052 (the initial keyword is broken, so it's impossible to test they only set those properties) and Edge will because [they don't support the `all` shorthand](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/cssallshorthand/) (likewise making it impossible to test).

Hexcles (approved):
I know nothing about unicode-bidi, but yikes...

Need to fix the integration test now that this test is passing:
```
 1:18.70 TEST_STATUS: Test OK. Subtests passed 11/11. Unexpected 1
Compare CSS div definitions (only valid if pre-reqs pass)
---------------------------------------------------------
Expected FAIL, got PASS
```

Otherwise, LGTM.

fantasai:
I think it's fair to not care about the value of unicode-bidi on HTML elements in the CSS test suite (other than when dir or bdo are used, but such tests belong in the HTML suite not the CSS one).

PR #11682, opened 2018-06-27
Test for all: initial on unknown :root

gsnedders:
If I'm not wrong about this, this will fail in Safari because of https://bugs.webkit.org/show_bug.cgi?id=187052, and hopefully stop any similar bugs in future.

### DIFF
--- a/css/css-cascade/all-prop-initial-xml.html
+++ b/css/css-cascade/all-prop-initial-xml.html
@@ -1,0 +1,33 @@
+<!doctype html>
+<title>all: initial on unknown XML tree</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<body>
+<script>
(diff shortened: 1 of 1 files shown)


fantasai (approved):
I don't entirely understand the test, because I don't speak JS, but afaict it's creating an iframe pointing at an XML doc whose root element is foo, and which contains an HTML style element  which contains
```
:root { color: blue }
:root { all: initial; direction: initial; unicode-bidi: initial } style { display: none }
```

It then asserts that the serialized gCS() of the foo element is equal to if the style element were not there.

I'm not sure why we care that style is display: none. Can't review the JS for appropriate struture etc. But assuming it all runs as I expect, afaict the test seems correct, i.e. will pass when it should pass, will fail when it should fail.

gsnedders:
> I'm not sure why we care that style is display: none.

Otherwise it affects the height of the root, because it's then rendering the contents of the style element.

PR #11685, opened 2018-06-27
Prepare ReplacementTokenizer for python 3.

Ms2ger:


tools/wptserve/wptserve/pipes.py
```diff
@@ -297,9 +297,10 @@ def index(self, token):
...
     def var(self, token):
         token = token[:-1]
-        return ("var", token)
+        return ("var", token.decode('utf8'))
 
     def tokenize(self, string):
+        assert isinstance(string, binary_type)
         return self.scanner.scan(string)[0]
 
     scanner = re.Scanner([(r"\$\w+:", var),
```
gsnedders: How come these don't need to be binary strings, given the regexes are called from tokenizer, which is binary?

Ms2ger: I don't know.

gsnedders: So it's just `re.Scanner` calling super low-level regexp engine things that makes it work. Let's make these all binary strings, because it's what it should be theoretically, which probably makes it safer with this "experimental" feature.

Ms2ger: Thanks for the comment; I found another bug and missing test coverage for the `arguments` handler.

gsnedders (approved):

PR #11702, opened 2018-06-28
[css-text] Update break-spaces text to white-space property

frivoal:
Relates to https://github.com/w3c/csswg-drafts/issues/2465


css/css-text/overflow-wrap/overflow-wrap-break-word-003.html
```diff
@@ -1,18 +1,18 @@
...
 <meta charset="utf-8">
-<title>CSS Text Test: overflow-wrap: break-word+break-spaces</title>
+<title>CSS Text Test: overflow-wrap:break-word + white-space:break-spaces</title>
 <link rel="author" title="Florian Rivoal" href="https://florian.rivoal.net/">
 <link rel="help" href="https://drafts.csswg.org/css-text-3/#valdef-overflow-wrap-break-word">
-<link rel="help" href="https://drafts.csswg.org/css-text-3/#valdef-overflow-wrap-break-spaces">
+<link rel="help" href="https://drafts.csswg.org/css-text-3/#valdef-white-space-break-spaces">
 <meta name="flags" content="ahem">
 <link rel="match" href="reference/overflow-wrap-break-word-003-ref.html">
 <meta name="assert" content="break-word + break-spaces do allow a break
```
javifernandez: I'm a bit confused by this test case. The spec prose about the break-spaces value states that "a breaking opportunity exists AFTER every preserved white-space character. Where in the spec is stated that it can break BEFORE ? I guess that behavior is part of the overflow-wrap: break-word, but what's the effect of the white-space value in this case ? Is this test verifying a NEW braking opportunity that didin't exist before adding the 'break-spaces' value ?

css/css-text/white-space/pre-wrap-009.html
```diff
@@ -1,23 +1,22 @@
...
+<link rel="help" href="https://drafts.csswg.org/css-text-3/#valdef-white-space-break-spaces">
 <link rel="match" href="reference/pre-wrap-001-ref.html">
-<meta name="assert" content="a single preserved white space at the end of the line is not wrapped when the white-space property is set to pre-wrap and overflow-wrap is break-spaces.">
+<meta name="assert" content="a single preserved white space at the end of the line is not wrapped when the white-space property is set to break-spaces.">
 <style>
 div {
   font-size: 20px;
   font-family: Ahem;
   line-height: 1em;
-  white-space: pre-wrap;
```
javifernandez: it's weird that this test is named pre-wrap-xxx but itsn't testing such value anymore.

css/css-text/white-space/pre-wrap-008.html
```diff
@@ -1,23 +1,22 @@
...
 <link rel="match" href="reference/pre-wrap-001-ref.html">
-<meta name="assert" content="preserved white space at the end of the line is wrapped when the white-space property is set to pre-wrap and overflow-wrap is break-spaces.">
+<meta name="assert" content="preserved white space at the end of the line is wrapped when the white-space property is set to break-spaces.">
 <style>
 div {
   font-size: 20px;
   font-family: Ahem;
   line-height: 1em;
-  white-space: pre-wrap;
+  white-space: break-spaces;
```
javifernandez: it's weird that this test is named pre-wrap-xxx but itsn't testing such value anymore.

css/css-text/white-space/textarea-pre-wrap-010.html
```diff
@@ -1,11 +1,12 @@
...
 <meta charset="utf-8">
-<title>CSS Text level 3 Test: single preserved white space at the end of with white-space:pre-wrap overflow-wrap:break-spaces break-word in a textarea</title>
+<title>CSS Text level 3 Test: single preserved white space at the end of the line with white-space:break-spaces overflow-wrap:break-word in a textarea</title>
 <link rel="author" title="Florian Rivoal" href="http://florian.rivoal.net/">
 <link rel="help" href="https://drafts.csswg.org/css-text-3/#white-space-phase-2">
 <link rel="help" href="https://drafts.csswg.org/css-text-3/#overflow-wrap-property">
+<link rel="help" href="https://drafts.csswg.org/css-text-3/#white-space-property">
 <link rel="match" href="reference/textarea-pre-wrap-001-ref.html">
-<meta name="assert" content="a single preserved white space at the end of the line is wrapped when the white-space property is set to pre-wrap and overflow-wrap is break-spaces break-word in a textarea.">
+<meta name="assert" content="a single preserved white space at the end of the line is wrapped when the white-space property is set to break-spaces and overflow-wrap is break-word in a textarea.">
```
javifernandez: What's the difference between this test and the one defined in overflow-wrap-break-word-003.html ?

css/css-text/white-space/pre-wrap-010.html
```diff
@@ -1,23 +1,24 @@
...
 <meta charset="utf-8">
-<title>CSS Text level 3 Test: single preserved white space at the end of with white-space:pre-wrap overflow-wrap:break-spaces break-word</title>
+<title>CSS Text level 3 Test: single preserved white space at the end of the line with white-space:break-spaces overflow-wrap:break-word</title>
 <link rel="author" title="Florian Rivoal" href="http://florian.rivoal.net/">
 <link rel="help" href="https://drafts.csswg.org/css-text-3/#white-space-phase-2">
 <link rel="help" href="https://drafts.csswg.org/css-text-3/#overflow-wrap-property">
+<link rel="help" href="https://drafts.csswg.org/css-text-3/#white-space-property">
 <link rel="match" href="reference/pre-wrap-001-ref.html">
-<meta name="assert" content="a single preserved white space at the end of the line is wrapped when the white-space property is set to pre-wrap and overflow-wrap is break-spaces break-word.">
+<meta name="assert" content="a single preserved white space at the end of the line is wrapped when the white-space property is set to break-spaces and overflow-wrap is break-word.">
```
javifernandez: What's the difference between this test and the one defined in overflow-wrap-break-word-003.html ?

css/css-text/overflow-wrap/overflow-wrap-break-word-003.html
frivoal: > Is this test verifying a NEW breaking opportunity that didin't exist before adding the 'break-spaces' value ?

Yes.
Also, if you have white-space:pre-wrap + overflow-wrap: break-word, you would not have a breaking opportunity, since the spaces would be collapsed (or hanged). You need white-space:break-spaces to make sure that the spaces are not collapsed/hanged, and you need overflow-wrap:break-word to add the wrapping opportunity before the first space (otherwise, it is just after).

css/css-text/white-space/pre-wrap-008.html
frivoal: good point. I'll rename.

css/css-text/white-space/pre-wrap-009.html
frivoal: good point. I'll rename.

css/css-text/white-space/pre-wrap-010.html
frivoal: I don't think there's a difference. I don't remember exactly, but I think we ended up with two tests because the first one I submitted took so long to get review that I forgot about it, and wrote a new one, and eventually both were merge.

Given that they test the same thing is slightly different ways, I think it is fine to have a little bit of redundancy, even if it isn't really necessary.

css/css-text/white-space/textarea-pre-wrap-010.html
frivoal: It's doing the same thing, but in a textarea instead of in a div. That is relevant because some browsers implement divs and textareas differently (because textarea is a replaced element). So, while both tests check the same statement in the spec, they are likely to catch different bugs.

kojiishi (approved):
I admit I didn't review these tests before, and Javier's comments look reasonable, but renaming look reasonable after the file rename updates, so lgtm. Please ensure Javier is happy either in this PR or in following PRs, looks like reasonable feedback to the original tests.

frivoal:
Merging. @javifernandez, if you want further fixes, can you file new issues and assign them to me?

PR #11706, opened 2018-06-28
Fix Python 3 errors for wpt lint/serve

saschanaz:
`python3 wpt serve` still somehow fails to serve the directories with a strange HTTP 500 message, but at least any Py3 errors that was visible are now fixed.

### DIFF
--- a/tools/wpt/testfiles.py
+++ b/tools/wpt/testfiles.py
@@ -6,6 +6,7 @@ import subprocess
 import sys
 
 from collections import OrderedDict
+from six import iteritems
 
 from ..manifest import manifest, update
 
@@ -67,7 +68,7 @@ def branch_point():
         branch_point = None
 
         # if there are any commits, take the first parent that is not in commits
-        for commit, parents in commit_parents.items():
+        for commit, parents in iteritems(commit_parents):
             for parent in parents:
                 if parent not in commit_parents:
                     branch_point = parent
--- a/tools/wpt/testfiles.py
+++ b/tools/wpt/testfiles.py
@@ -67,7 +67,7 @@ def branch_point():
         branch_point = None
 
         # if there are any commits, take the first parent that is not in commits
-        for commit, parents in commit_parents.iteritems():
+        for commit, parents in commit_parents.items():
             for parent in parents:
                 if parent not in commit_parents:
                     branch_point = parent
--- a/tools/wptserve/tests/functional/base.py
+++ b/tools/wptserve/tests/functional/base.py
@@ -9,6 +9,7 @@ import unittest
 from six.moves.urllib.parse import urlencode, urlunsplit
 from six.moves.urllib.request import Request as BaseRequest
 from six.moves.urllib.request import urlopen
+from six import iteritems
 
 wptserve = pytest.importorskip("wptserve")
 
@@ -57,7 +58,7 @@ class TestUsingServer(unittest.TestCase):
         if headers is None:
             headers = {}
 
-        for name, value in headers.items():
+        for name, value in iteritems(headers):
             req.add_header(name, value)
 
         if body is not None:
--- a/tools/wptserve/tests/functional/base.py
+++ b/tools/wptserve/tests/functional/base.py
@@ -29,7 +29,7 @@ class Request(BaseRequest):
         return self.method
 
     def add_data(self, data):
-        if hasattr(data, "iteritems"):
+        if hasattr(data, "items"):
             data = urlencode(data)
         print(data)
         self.add_header("Content-Length", str(len(data)))
@@ -57,7 +57,7 @@ class TestUsingServer(unittest.TestCase):
         if headers is None:
             headers = {}
 
-        for name, value in headers.iteritems():
+        for name, value in headers.items():
             req.add_header(name, value)
 
         if body is not None:
--- a/tools/wptserve/wptserve/handlers.py
+++ b/tools/wptserve/wptserve/handlers.py
@@ -238,7 +238,7 @@ class PythonScriptHandler(object):
         try:
             environ = {"__file__": path}
             sys.path.insert(0, os.path.dirname(path))
-            with open(path) as f:
+            with open(path, 'rb') as f:
                 exec(compile(f.read(), path, 'exec'), environ, environ)
             if "main" in environ:
                 handler = FunctionHandler(environ["main"])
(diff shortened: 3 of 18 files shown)


tools/wptserve/wptserve/pipes.py
```diff
@@ -478,7 +479,7 @@ def config_replacement(match):
                     "unexpected token type %s (token '%r'), expected ident or arguments" % (ttype, field)
                 )
 
-        assert isinstance(value, (int,) + types.StringTypes), tokens
+        assert isinstance(value, (int, str)), tokens
```
gsnedders: Can you make this `six.string_types`? If we're allowing `str` in Py3, we should allow `unicode` in Py2.

tools/wptserve/wptserve/response.py
```diff
@@ -182,7 +181,7 @@ def iter_content(self, read_file=False):
         True, the entire content of the file will be returned as a string facilitating
         non-streaming operations like template substitution.
         """
-        if isinstance(self.content, types.StringTypes):
+        if isinstance(self.content, str):
```
gsnedders: `six.string_types`.

tools/wptserve/wptserve/response.py
```diff
@@ -426,7 +425,7 @@ def end_headers(self):
 
     def write_content(self, data):
         """Write the body of the response."""
-        if isinstance(data, types.StringTypes):
+        if isinstance(data, (text_type, binary_type)):
```
gsnedders: Can you add a comment saying these are deliberately both allowed and to see `self.encode`, given this is a bit unusual?

tools/wptserve/wptserve/router.py
```diff
@@ -135,7 +134,7 @@ def register(self, methods, path, handler):
                         object and the response object.
 
         """
-        if type(methods) in types.StringTypes or methods in (any_method, "*"):
+        if type(methods) is str or methods in (any_method, "*"):
```
gsnedders: `isinstance(methods, six.string_types)`

tools/wptserve/wptserve/server.py
```diff
@@ -80,7 +79,7 @@ def register(self, methods, input_path, output_path):
         :param output_path: Path to replace the input path with in
                             the request.
         """
-        if type(methods) in types.StringTypes:
+        if type(methods) is str:
```
gsnedders: `isinstance(methods, six.string_types)`

tools/wptserve/wptserve/stash.py
```diff
@@ -42,14 +42,16 @@ def load_env_config():
...
     return address, authkey
 
 def store_env_config(address, authkey):
-    authkey = base64.encodestring(authkey)
-    os.environ["WPT_STASH_CONFIG"] = json.dumps((address, authkey))
+    authkey = base64.b64encode(authkey)
+    os.environ["WPT_STASH_CONFIG"] = json.dumps((address, authkey.decode("ascii")))
 
 def start_server(address=None, authkey=None):
+    if type(authkey) is str:
```
gsnedders: `isinstance(authkey, six.text_type)` (we definitely don't want the next line running on Py2 where str is bytes)

tools/wpt/testfiles.py
```diff
@@ -67,7 +67,7 @@ def branch_point():
         branch_point = None
 
         # if there are any commits, take the first parent that is not in commits
-        for commit, parents in commit_parents.iteritems():
+        for commit, parents in commit_parents.items():
```
gsnedders: Use `six.iteritems` here

tools/wptserve/tests/functional/base.py
```diff
@@ -57,7 +57,7 @@ def request(self, path, query=None, method="GET", headers=None, body=None, auth=
         if headers is None:
             headers = {}
 
-        for name, value in headers.iteritems():
+        for name, value in headers.items():
```
gsnedders: `six.iteritems`

tools/wptserve/wptserve/handlers.py
```diff
@@ -237,7 +237,7 @@ def __call__(self, request, response):
         try:
             environ = {"__file__": path}
             sys.path.insert(0, os.path.dirname(path))
-            execfile(path, environ, environ)
+            exec(compile(open(path).read(), path, 'exec'), environ, environ)
```
gsnedders: Hmm.

This will pass the contents of `path` as a unicode str in Py3 (decoded based on the default system encoding) and a bytes str in Py2 (does `execfile` just match `compile` and treat it as Latin-1 then?). What's our current behaviour here? Do we pay attention to `coding` declarations?

It also will make us inherit `__future__` imports from this file, which is a bit weird (I think all we have to do to there is `dont_inherit=False` on the `compile` call?). What's our current behaviour here?

IMO, we should be following `coding` and `__future__` imports in the handler files, though obviously if we currently don't then we shouldn't block this PR on that.

That all said, I'd strongly much prefer to explicitly close the file object (given this makes a significant behavioural difference on other Python impls, most obviously PyPy), probably using a context manager for it.

tools/wptserve/wptserve/handlers.py
```diff
@@ -375,7 +375,7 @@ def __init__(self, data, content_type, **headers):
         self.data = data
 
         self.resp_headers = [("Content-Type", content_type)]
-        for k, v in headers.iteritems():
+        for k, v in headers.items():
```
gsnedders: `six.iteritems`

tools/wptserve/wptserve/response.py
```diff
@@ -337,7 +336,7 @@ def __setitem__(self, key, value):
         self.set(key, value)
 
     def __iter__(self):
-        for key, values in self.data.itervalues():
+        for key, values in self.data.values():
```
gsnedders: `six.itervalues`

gsnedders (changes requested):
While I know I've written quite a lot of comments here, don't be disheartened!

Most of these relate to iterating over dicts (where we probably don't want to always allocate a list in Py2) and use of `str` in polyglot code (which IMO is rarely a good idea).

tools/wptserve/wptserve/response.py
saschanaz: Done, below this line.

tools/wptserve/wptserve/handlers.py
```diff
@@ -237,7 +238,8 @@ def __call__(self, request, response):
         try:
             environ = {"__file__": path}
             sys.path.insert(0, os.path.dirname(path))
-            execfile(path, environ, environ)
+            with open(path) as f:
```
gsnedders: Can you make this `open(path, "rb")` so we have the same behaviour on Py2 and 3?

saschanaz:
Any idea about the sudden failure? Reverting doesn't help, maybe something happened in the master branch?

gsnedders (approved):

PR #11710, opened 2018-06-28
Add reviewers for css/css-box

gsnedders:
We need some reviewers for css/css-box. This is really still covered by CSS 2, so there's some argument that the tests should get moved into css/CSS2/visudet, though I don't really care too much either way.

Who makes sense as reviewers here? @mstensho @atotic? @dbaron who'd make sense from Mozilla here?

### DIFF
--- a/css/css-box/META.yml
+++ b/css/css-box/META.yml
@@ -1,2 +1,4 @@
 suggested_reviewers:
-  - 
+  - dbaron
+  - dholbert
+  - mstensho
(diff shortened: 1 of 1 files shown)


dbaron:
Probably me and maybe also @dholbert.

dholbert:
Sure, happy to be a reviewer here.

mstensho:
I'm in!

PR #11714, opened 2018-06-28
Sync Mozilla tests as of 2018-06-28

emilio:
Sync Mozilla tests as of https://hg.mozilla.org/mozilla-central/rev/6041c030780420b6205cf2d6640513606609884c

This contains two changes:

 * One by Mihir Iyer from [bug 1398483](https://bugzil.la/1398483), reviewed by
   @dholbert, adding flexbox column-{row,gap} reftests.

 * One by myself from [bug 1464782](https://bugzil.la/1464782), replacing the
   usage of offset-* logical properties by the standard name inset-*, reviewed
   by @upsuper.

PR #11715, opened 2018-06-28
Make Mozilla's sync-tests.sh work with git. 

emilio:
I needed this for #11714.

### DIFF
--- a/css/vendor-imports/mozilla/mozilla-central-reftests/sync-tests.sh
+++ b/css/vendor-imports/mozilla/mozilla-central-reftests/sync-tests.sh
@@ -10,8 +10,15 @@ then
     exit 1
 fi
 
+if [ -e "$MOZTREE/.git" ]
+then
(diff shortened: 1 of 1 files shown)


dbaron (approved):
I tend to put the `then` on its own line rather than sticking a semicolon before it and putting it on the same line as the `if`, but I'm probably a bit unusually in preferring that style for shell scripts...

Otherwise looks great.

dbaron:
Also not sure why this would trigger this lint failure (unless maybe your branchpoint is old?):
```
ERROR:lint:css/vendor-imports/mozilla: Directory is missing META.yml file. (MISSING-META-YML)
```

emilio:
Looks like I'm on top of master, which has 93dceabca4f79ced2a35fa480f78d30d4123ffaa. @zcorpan do you know why?

emilio:
That is, why CI is failing with this patch, and what should be done for it not to complain? :)

zcorpan:
Rebase, something was wrong with the linter.

PR #11727, opened 2018-06-29
Fix make_hosts_file tests in python 3.

Ms2ger:


gsnedders (changes requested):
To me this implies something weirder is going on. Most obviously, what *is* getting imported as `serve` such that this doesn't work?

The answer seems to be that `__init__.py` is importing `serve`, which is found at `/wpt_root/serve.py`, and that's definitely wrong. That's what needs to be fixed, I think?

gsnedders (approved):

PR #11728, opened 2018-06-29
[css-cascade] parsing tests for 'all' property

ewilligers:
Spec:
https://drafts.csswg.org/css-cascade-3/#propdef-all

### DIFF
--- a/css/css-cascade/parsing/support/parsing-testcommon.js
+++ b/css/css-cascade/parsing/support/parsing-testcommon.js
@@ -1,0 +1,39 @@
+'use strict';
+
+// serializedValue can be the expected serialization of value,
+// or an array of permitted serializations,
+// or omitted if value should serialize as value.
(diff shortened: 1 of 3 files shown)


emilio (approved):

css/css-cascade/parsing/support/parsing-testcommon.js
```diff
@@ -0,0 +1,39 @@
...
+
+    test(function(){
+        var div = document.createElement('div');
+        div.style[property] = value;
+        assert_not_equals(div.style.getPropertyValue(property), "", "property should be set");
+
+        var div = document.createElement('div');
+        div.style[property] = value;
+        var readValue = div.style.getPropertyValue(property);
+        if (serializedValue instanceof Array)
```
emilio: (Probably this should use Array.isArray instead, but worth changing everything at the same time)

PR #11730, opened 2018-06-29
[css-flexbox] Flexible Box property parsing tests

ewilligers:
Test parsing of the properties defined by CSS Flexible Box layout
https://drafts.csswg.org/css-flexbox/#property-index

Found bug:
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/18099384/


emilio (commented):
I think most of the align-* and justify-* tests should reference https://drafts.csswg.org/css-align/ itself... I don't know why these properties are defined in two places.

FremyCompany:
The fallback value for stretch in case the element didn't support stretching up to the full available size used to be in the specs, and I can't find when or where it was removed. @ewilligers could you try to find that out?

fantasai:
You can reference both specs, it's possible. :)

FremyCompany (approved):
Disclaimer that I didn't check that other people's comments were addressed, but otherwise this looks good to me

ewilligers:
CSS Box Alignment defines many more values for align-* and justify-* than CSS Flexible Box Layout does, so that will be a better location for those properties' parsing tests.

In a separate PR I will write parsing tests for align-* and justify-* , place them in css-align/parsing and reference both specs with help links.

Issue #11722, opened 2018-06-28
`wpt lint` complains about missing `META.yml` even though the file exists


Hexcles:
In #11720 , `wpt lint` complained: https://travis-ci.org/web-platform-tests/wpt/jobs/397980951

```
ERROR:lint:interfaces: Directory is missing META.yml file. (MISSING-META-YML)
INFO:lint:
INFO:lint:There was 1 error (MISSING-META-YML: 1)
INFO:lint:You must fix all errors; for details on how to fix them, see
INFO:lint:https://web-platform-tests.org/writing-tests/lint-tool.html
INFO:lint:
INFO:lint:However, instead of fixing a particular error, it's sometimes
INFO:lint:OK to add a line to the lint.whitelist file in the root of the
INFO:lint:web-platform-tests directory to make the lint tool ignore it.
INFO:lint:
INFO:lint:For example, to make the lint tool ignore all 'MISSING-META-YML'
INFO:lint:errors in the interfaces file,
INFO:lint:you could add the following line to the lint.whitelist file.
INFO:lint:
INFO:lint:MISSING-META-YML: interfaces
```

But `interfaces/META.yml` does exist. @gsnedders can you take a look?

gsnedders:
INTERESTING.

gsnedders:
Closed by reverting. Relanding will be tracked in the original bug (#10552).

Issue #9171, opened 2018-01-24
Lint can regress on master because CSS-COLLIDING-* lints don't get all paths


gsnedders:
```
ERROR:lint:css/css-grid/alignment/support/100x100-green.png: The filename support/100x100-green.png is shared by: css/css-flexbox/support/100x100-green.png, css/css-grid/alignment/support/100x100-green.png, css/css-grid/grid-items/support/100x100-green.png (CSS-COLLIDING-SUPPORT-NAME)
```

This seems to have started failing with 1f69b4225e6c424b85b7a90b3851a72bf8915b60 from https://github.com/w3c/web-platform-tests/pull/9089 (where CI passed, including lint).

The underlying cause of this is that in CI we call lint with only the changed paths, and the CSS-COLLIDING-* lints need to know all filenames to consider whether they collide or not.

cc/ @jgraham

gsnedders:
One obvious solution could just be to make `css/` all get linted any time a thing in `css/` changes.

belaeifert:
Hello, I could not replicate this error.
Could you tell how to replicate?

Thank you

gsnedders:
This really is just a special case of #10021?

PR #11734, opened 2018-06-29
Lint that META.yml files exist

### LINKED ISSUE #10552: Idea for new lint check: missing OWNERS for new top-level directory

gsnedders:
Fixes #10552. Previously landed in #11092, reverted due to #11722.

TODO: The failure that caused #11722 is the same underlying issue as #9171: as an optimisation, we only pass changed files to lint, which doesn't suffice for the things that need to consider more than one file.


tools/manifest/sourcefile.py
```diff
@@ -264,6 +264,8 @@ def name_is_non_test(self):
                 self.name_prefix("MANIFEST") or
                 self.filename == "META.yml" or
                 self.filename.startswith(".") or
+                self.filename == "OWNERS" or
```
jugglinmike: When is this value necessary?

tools/manifest/sourcefile.py
```diff
@@ -264,6 +264,8 @@ def name_is_non_test(self):
...
+                self.filename == "META.yml" or
```
jugglinmike: This value is already included in the expression

gsnedders: Hah, I never noticed that when I reviewed this before. I suck.

jgraham (changes requested):

PR #11741, opened 2018-06-29
[fetch] Do not return value from cleanup fns

jugglinmike:
Today, the return value of functions provided to the `Test#add_cleanup`
function has no effect on the behavior of the test runner. Despite this,
some existing tests have already been authored to return the `null`
value.

An upcoming feature addition to testharness.js will cause the return
value to influence test results [1]. To allow the new test harness
feature to land without introducing harness errors, refactor existing
tests to omit a return value.

[1] https://github.com/web-platform-tests/wpt/issues/6075


domfarolino (approved):
That seems reasonable to me

PR #11743, opened 2018-06-30
fix(resources/testdriver): correct syntax

saschanaz:
Introduced by #11511 😁

cc: @jugglinmike


jugglinmike (approved):

jugglinmike:
Thanks!

gsnedders:
…how did this not break everything so significantly that it was obvious?

PR #11745, opened 2018-07-01
Fix wptserve test_stash for Python 3

saschanaz:


tools/wptserve/tests/functional/base.py
```diff
@@ -34,7 +34,7 @@ def add_data(self, data):
             data = urlencode(data)
         print(data)
         self.add_header("Content-Length", str(len(data)))
-        BaseRequest.add_data(self, data)
+        self.data = data
```
saschanaz: This is because `add_data` was deprecated in 3.3 and removed in 3.4.

PR #11746, opened 2018-07-01
Fix wptserve test_request on Python 3

saschanaz:


tools/wptserve/tests/functional/base.py
```diff
@@ -32,9 +32,11 @@ def get_method(self):
     def add_data(self, data):
         if hasattr(data, "items"):
             data = urlencode(data)
+        if isinstance(data, text_type):
+            data = data.encode("ascii")
         print(data)
         self.add_header("Content-Length", str(len(data)))
-        BaseRequest.add_data(self, data)
+        self.data = data
```
saschanaz: This is because `add_data` was deprecated in 3.3 and removed in 3.4.

tools/wptserve/wptserve/response.py
```diff
@@ -181,7 +181,7 @@ def iter_content(self, read_file=False):
         True, the entire content of the file will be returned as a string facilitating
         non-streaming operations like template substitution.
         """
-        if isinstance(self.content, string_types):
+        if isinstance(self.content, binary_type):
```
saschanaz: I think we can just use `bytes` which is in both Py2 and 3, but not sure whether reviewers agree.

gsnedders:
@foolip there are tests there, just marked xfail on Py3!

gsnedders (commented):
One single thought, mostly wondering about the API surface. Otherwise LGTM, but does need rebasing.

saschanaz:
Seems that those fixes are already done by another PR(s), so just removing some more xfails.

gsnedders (approved):

PR #11750, opened 2018-07-02
Add a test for getRootNode({composed:true})

### LINKED ISSUE #3712: Test getRootNode({composed:true})

xiuqijix:
Fix #3712, @Ms2ger, PTAL.

### DIFF
--- a/dom/nodes/rootNode.html
+++ b/dom/nodes/rootNode.html
@@ -11,6 +11,20 @@
 <script>
 
 test(function () {
+    var shadowHost = document.createElement('div');
+    document.body.appendChild(shadowHost);
+
+    var shadowRoot = shadowHost.attachShadow({mode: 'open'});
+    shadowRoot.innerHTML = '<div class="shadowChild">content</div>';
+
+    var shadowChild = shadowRoot.querySelector('.shadowChild');
+    assert_equals(shadowChild.getRootNode({composed: true}), shadowRoot, "getRootNode() must return context object's shadow-including root if options's composed is true");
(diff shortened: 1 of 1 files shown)


dom/nodes/rootNode.html
```diff
@@ -10,6 +10,20 @@
...
 
+test(function () {
+    var shadowHost = document.createElement('div');
+    document.body.appendChild(shadowHost);
+
+    var shadowRoot = shadowHost.attachShadow({mode: 'open'});
+    shadowRoot.innerHTML = '<div class="shadowChild">content</div>';
+
+    var shadowChild = shadowRoot.querySelector('.shadowChild');
+    assert_equals(shadowChild.getRootNode({composed: true}), shadowRoot, "getRootNode() must return context object's shadow-including root if options's composed is true");
```
annevk: This would return document, no?

annevk: I think you got the argument inversed.

kaixinjxq: @annevk, according to [spec](https://dom.spec.whatwg.org/#dom-node-getrootnode) `The getRootNode(options) method, when invoked, must return context object’s shadow-including root if options’s composed is true, and context object’s root otherwise`, so if `composed: true` should return `shadowRoot`

annevk: When it's true, it must return the shadow-including root, which can never be a `ShadowRoot` object if you read the definition of that.

kaixinjxq: My mistake, I've addressed it in new commit, PTAL.

annevk (approved):

PR #11765, opened 2018-07-03
Auto-update the appmanifest IDL file

lukebjerring:
Hello reviewer(s),

Please check that:

### DIFF
--- a/appmanifest/META.yml
+++ b/appmanifest/META.yml
@@ -1,0 +1,7 @@
+suggested_reviewers:
+  - anssiko
+  - kenchris
+  - marcoscaceres
+  - mgiuca
+  - mounirlamouri
+  - RobDolin
--- a/appmanifest/idlharness.window.js
+++ b/appmanifest/idlharness.window.js
@@ -16,6 +16,7 @@ promise_test(async () => {
   idl_array.add_dependency_idls(dom);
   idl_array.add_objects({
     Window: ['window'],
+    BeforeInstallPromptEvent: ['new BeforeInstallPromptEvent("type")'],
   });
   idl_array.test();
 }, 'appmanifest interfaces');
--- a/appmanifest/idlharness.window.js
+++ b/appmanifest/idlharness.window.js
@@ -1,0 +1,21 @@
+// META: script=/resources/WebIDLParser.js
+// META: script=/resources/idlharness.js
+
+// https://w3c.github.io/manifest/
+
+'use strict';
+
+promise_test(async () => {
+  const srcs = ['appmanifest', 'dom', 'html'];
+  const [idl, dom, html] = await Promise.all(
+    srcs.map(i => fetch(`/interfaces/${i}.idl`).then(r => r.text())));
+
+  const idl_array = new IdlArray();
+  idl_array.add_idls(idl);
+  idl_array.add_dependency_idls(html);
+  idl_array.add_dependency_idls(dom);
+  idl_array.add_objects({
+    Window: ['window'],
+  });
+  idl_array.test();
+}, 'appmanifest interfaces');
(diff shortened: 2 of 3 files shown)


marcoscaceres (commented):
I'm still not super comfortable with this... it's all just dictionaries and there is nothing to actually test.

marcoscaceres:
(except the event)

marcoscaceres:
Oops, wrong button.

lukebjerring:
> I'm still not super comfortable with this...

re: "still" - Was this test proposed at another time (I can't find any closed PRs), or are you referring to a similar test?

I realize it doesn't have a large number of tests, but not testing an API because the API is small isn't really a great argument, so only testing a small portion of the IDL is worth it for the same reason IMHO. There should also be other (functional) tests, which will flesh out the dir.

marcoscaceres:
> re: "still" - Was this test proposed at another time (I can't find any closed PRs), or are you referring to a similar test?

Apologies, we've been discussing elsewhere how we are actually going to test this spec. 

>  but not testing an API because the API is small isn't really a great argument,

Sorry, that's not the argument I'm making. I'm saying that most of the dictionaries are not part of an API at all.   

> so only testing a small portion of the IDL is worth it for the same reason IMHO.

I agree. It's why I think we should only include the IDL associated with the BIP event. I'd be totally ok with that and think that's the right thing to do here.

lukebjerring:
Feel free to link to the thread from here where relevant.

By pulling in all the IDL, there are just some basic assertions made about `partial` dictionaries (e.g. that the original dictionary exists somewhere, exposure set is valid, ...), which I think have value.

marcoscaceres:
> Feel free to link to the thread from here where relevant.

https://github.com/w3c/manifest/issues/666

> By pulling in all the IDL, there are just some basic assertions made about partial dictionaries (e.g. that the original dictionary exists somewhere, exposure set is valid, ...), which I think have value.

Ok, but only in as far as they relate to the BIP event (if at all), or am I missing something?

lukebjerring:
Yeah you're totally right, nothing special about these dicts.
Though, side-note, I am missing a `BeforeInstallPromptEvent` object.

lukebjerring:
Test which fails in Chrome is fixed by https://github.com/w3c/manifest/pull/698

mgiuca:
I think there is value to this, but as Marcos says, most of this IDL is not actually exposed and so probably shouldn't be tested. Just the `BeforeInstallPromptEvent` and `Window` parts should be tested.

Note that this not only tests `BeforeInstallPromptEvent` object, but also the presence of the two attributes `onbeforeinstallprompt` and `onappinstalled`, so that's valuable.

Luke, will we have to manually update the WPT copy of the IDL when we change the manifest spec, or will there be some automated process? How will the automated process know not to copy the non-exposed dictionaries? (Perhaps "`Exposed=Window`" is explicitly required, and nothing without that is exposed.)

lukebjerring:
This is a PR that moves __towards__ automatically scraping and updating the IDL, but that flow isn't ready yet, so updates like this will be irregular. I wouldn't be concerned about manually updating the IDL in this case - the scrape will eventually keep it up to date.

Not testing the dictionaries is how idlharness behaves as-is so I wouldn't think anything special needs to be done. The only idlharness tests are that if it's a partial dict, the original dict exists, and if it's a dict that extends another dict, the other dict exists. Neither applies here so those dicts are ignored in the test output.

mgiuca (approved):
OK, so you're saying that it's *correct* to import all the dictionaries into the WPT file, even though idlharness will ignore them? Sounds fine to me.

If this isn't actually establishing automatically scraping, perhaps reword the title (removing "Auto-").

LGTM, though wait for @marcoscaceres to chime in also.

lukebjerring:
Yep that's what I'm saying! 
Good point - "auto" is the scrape part. Update is still manual.

marcoscaceres (approved):
I can live with it... but maybe we need so way to tell the scraper "not this IDL".

PR #11767, opened 2018-07-03
Streams Object.prototype.then interception

ricea:
Add tests to verify that modifying Object.prototype.then does not make the
internals of pipeTo() and tee() visible.

Equivalent standard change is whatwg/streams#936.

### DIFF
--- a/streams/piping/then-interception.js
+++ b/streams/piping/then-interception.js
@@ -1,0 +1,64 @@
+'use strict';
+
+if (self.importScripts) {
+  self.importScripts('/resources/testharness.js');
+  self.importScripts('../resources/test-utils.js');
+  self.importScripts('../resources/recording-streams.js');
+}
+
+function interceptThen() {
+  const intercepted = [];
+  const callCount = 0;
+  Object.prototype.then = function(resolver) {
+    if (!this.done) {
+      intercepted.push(this.value);
+    }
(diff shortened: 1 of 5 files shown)


streams/piping/then-interception.js
```diff
@@ -0,0 +1,64 @@
+'use strict';
+
+if (self.importScripts) {
+  self.importScripts('/resources/testharness.js');
+  self.importScripts('../resources/test-utils.js');
+  self.importScripts('../resources/recording-streams.js');
+}
+
+function interceptThen() {
```
domenic: Seems a bit simpler to just make then() throw? The test will then fail if it's accessed.

domenic (approved):
Should we also add tests for fetch()?

streams/piping/then-interception.js
ricea: Yes, but less fun. "I can make pipeTo throw" is far less compelling than "I can make pipeTo write arbitrary data". Please indulge my desire to make this test exciting.

ricea:
> Should we also add tests for fetch()?

My plan is to add tests for fetch() when I fix this in Blink, since we can't test fetch() with the streams reference implementation anyway. Is that okay?

domenic:
Sounds like a plan!

Issue #7583, opened 2017-10-04
Add support for namespaces in idlharness


tobie:
See: #7551.

bzbarsky:
At this point is this more or less blocking tests being created for https://github.com/w3c/csswg-drafts/pull/437 and to some extent UAs updating to that spec change...

lukebjerring:
Blocking https://github.com/web-platform-tests/wpt/pull/11768

lukebjerring:
FYI, I'm taking a stab at this.

PR #11768, opened 2018-07-03
Auto-update the console IDL file

lukebjerring:
Hello reviewer(s),

This PR is intended to consolidate the spec’s IDL definition with the WPT test suite’s copy, and any idlharness tests.

The up-to-date copy of the IDL file was automatically extracted from the reffy-reports repo (https://github.com/tidoust/reffy-reports/tree/master/whatwg/idl) which scrapes known specs automatically + regulary.

This PR is part of a migration project which will eventually be automatically updating and creating PRs for changes in spec IDL.

Please check that:
The spec (and its source) is correct and up-to-date
All tests which cover the IDL in the spec have been migrated to fetch + use the idl in the `interfaces/` directory (instead of inline copies in the test files).

PR #11776, opened 2018-07-03
Handle (+ test) 'namespace' in idlharness

lukebjerring:
Fixes https://github.com/web-platform-tests/wpt/issues/7583

Runs some basic tests on `namespace` definitions in IDL.
- Tests partial definitions are valid
- Tests exposure set of partial interface are subset of original namespace
- Tests the `length` property of `namespace` `operation` definitions (`optional` params is handled incorrectly in Chrome, it seems..)
- Tests the type of the return value for the methods

### DIFF
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -3033,9 +3033,6 @@ function IdlNamespace(obj)
 
     /** An array of IdlInterfaceMembers. */
     this.members = obj.members.map(m => new IdlInterfaceMember(m));
-    if (this.has_extended_attribute("Unforgeable")) {
-        this.members.forEach(m => { m.isUnforgeable = true; });
-    }
 }
 //@}
 
@@ -3062,13 +3059,6 @@ IdlNamespace.prototype.do_member_operation_asserts = function (memberHolderObjec
         typeof memberHolderObject[member.name],
         "function",
          "property must be a function");
-
-    var args = member.arguments.map(function(a) {
-        return create_suitable_object(a.idlType);
-    });
-    this.array.assert_type_is(
-        memberHolderObject[member.name].call(null, args),
-        member.idlType);
 
     assert_equals(
         memberHolderObject[member.name].length,
@@ -3125,14 +3115,8 @@ IdlNamespace.prototype.test_member_attribute = function (member)
             member.name,
             this.name + ' does not have property ' + format_value(member.name));
 
-        // "The attribute setter is undefined if the attribute is declared
-        // readonly and has neither a [PutForwards] nor a [Replaceable]
-        // extended attribute declared on it."
-        if (!member.has_extended_attribute("PutForwards")
-            && !member.has_extended_attribute("Replaceable")) {
-            var desc = Object.getOwnPropertyDescriptor(self[this.name], member.name);
-            assert_equals(desc.set, undefined, "setter must be undefined for readonly attributes");
-        }
+        var desc = Object.getOwnPropertyDescriptor(self[this.name], member.name);
+        assert_equals(desc.set, undefined, "setter must be undefined for namespace members");
         a_test.done();
     }.bind(this));
 };
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -1321,7 +1321,7 @@ IdlInterface.prototype.get_inheritance_stack = function() {
     while (idl_interface.base) {
         var base = this.array.members[idl_interface.base];
         if (!base) {
-            throw new IdlHarnessError(`${idl_interface.name} base ${idl_interface.base} not found`);
+            throw new Error(idl_interface.type + " " + idl_interface.base + " not found (inherited by " + idl_interface.name + ")");
         } else if (stack.indexOf(base) > -1) {
             stack.push(base);
             let dep_chain = stack.map(i => i.name).join(',');
@@ -2267,11 +2267,12 @@ IdlInterface.prototype.do_member_operation_asserts = function(memberHolderObject
     assert_equals(typeof memberHolderObject[member.name], "function",
                   "property must be a function");
 
-    const ctors = this.members
-        .filter(m => m.type == "operation" && m.name == member.name);
+    const ctors = this.members.filter(function(m) {
+        return m.type == "operation" && m.name == member.name;
+    });
     assert_equals(
-        minOverloadLength(ctors),
         memberHolderObject[member.name].length,
+        minOverloadLength(ctors),
         "property has wrong .length");
 
     // Make some suitable arguments
@@ -3018,7 +3019,7 @@ function IdlTypedef(obj)
 }
 //@}
 
-IdlNamespace.prototype = Object.create(IdlObject.prototype);
+IdlTypedef.prototype = Object.create(IdlObject.prototype);
 
 /// IdlNamespace ///
 function IdlNamespace(obj)
@@ -3038,23 +3039,23 @@ function IdlNamespace(obj)
 }
 //@}
 
+IdlNamespace.prototype = Object.create(IdlObject.prototype);
+
 IdlNamespace.prototype.do_member_operation_asserts = function (memberHolderObject, member, a_test)
 //@{
 {
-    var done = a_test.done.bind(a_test);
-    var operationUnforgeable = member.has_extended_attribute("Unforgeable");
     var desc = Object.getOwnPropertyDescriptor(memberHolderObject, member.name);
 
     assert_false("get" in desc, "property should not have a getter");
     assert_false("set" in desc, "property should not have a setter");
     assert_equals(
         desc.writable,
-        !operationUnforgeable,
+        !member.isUnforgeable,
         "property should be writable if and only if not unforgeable");
     assert_true(desc.enumerable, "property should be enumerable");
     assert_equals(
         desc.configurable,
-        !operationUnforgeable,
+        !member.isUnforgeable,
         "property should be configurable if and only if not unforgeable");
 
     assert_equals(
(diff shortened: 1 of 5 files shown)


lukebjerring:
Chromium equivalent CL (which also includes the IDL test from #11768):
https://chromium-review.googlesource.com/c/chromium/src/+/1125206

Note: Both FF and Chrome are misreporting `.length` on a bunch of `console` methods.
e.g. 
> console namespace: operation clear()
> assert_equals: operation has wrong .length expected 0 but got 1

Indeed, `console.clear.length` is `1` in Chrome.

resources/idlharness.js
```diff
@@ -1310,7 +1321,7 @@ IdlInterface.prototype.get_inheritance_stack = function() {
     while (idl_interface.base) {
         var base = this.array.members[idl_interface.base];
         if (!base) {
-            throw new Error(idl_interface.type + " " + idl_interface.base + " not found (inherited by " + idl_interface.name + ")");
+            throw new IdlHarnessError(`${idl_interface.name} base ${idl_interface.base} not found`);
```
Ms2ger: I know backtick-strings are nice, but let's keep this file working in older browsers.

resources/idlharness.js
```diff
@@ -2255,15 +2266,12 @@ IdlInterface.prototype.do_member_operation_asserts = function(memberHolderObject
...
-    // ". . .
-    // "Return the length of the shortest argument list of the
-    // entries in S."
-    assert_equals(memberHolderObject[member.name].length,
-        minOverloadLength(this.members.filter(function(m) {
-            return m.type == "operation" && m.name == member.name;
-        })),
+
+    const ctors = this.members
+        .filter(m => m.type == "operation" && m.name == member.name);
```
Ms2ger: Same here (`const`, arrow function).

resources/idlharness.js
```diff
@@ -2255,15 +2266,12 @@ IdlInterface.prototype.do_member_operation_asserts = function(memberHolderObject
...
+    assert_equals(
+        minOverloadLength(ctors),
+        memberHolderObject[member.name].length,
```
Ms2ger: Actual value first, expected value second.

resources/idlharness.js
```diff
@@ -3010,7 +3018,143 @@ function IdlTypedef(obj)
 }
 //@}
 
-IdlTypedef.prototype = Object.create(IdlObject.prototype);
```
Ms2ger: Keep this

resources/idlharness.js
```diff
@@ -3010,7 +3018,143 @@ function IdlTypedef(obj)
...
+        this.members.forEach(m => { m.isUnforgeable = true; });
+    }
+}
+//@}
+
+IdlNamespace.prototype.do_member_operation_asserts = function (memberHolderObject, member, a_test)
+//@{
+{
+    var done = a_test.done.bind(a_test);
+    var operationUnforgeable = member.has_extended_attribute("Unforgeable");
```
Ms2ger: Should this check the `isUnforgeable` property instead?

resources/idlharness.js
```diff
@@ -3010,7 +3018,143 @@ function IdlTypedef(obj)
...
+    if (this.has_extended_attribute("Unforgeable")) {
+        this.members.forEach(m => { m.isUnforgeable = true; });
+    }
+}
+//@}
+
+IdlNamespace.prototype.do_member_operation_asserts = function (memberHolderObject, member, a_test)
+//@{
+{
+    var done = a_test.done.bind(a_test);
```
Ms2ger: This seems moderately pointless; why not call `a_test.done()` below?

resources/idlharness.js
```diff
@@ -3010,7 +3018,143 @@ function IdlTypedef(obj)
...
+        "property should be configurable if and only if not unforgeable");
+
+    assert_equals(
+        typeof memberHolderObject[member.name],
+        "function",
+         "property must be a function");
+
+    var args = member.arguments.map(a => create_suitable_object(a.idlType));
+    this.array.assert_type_is(
+        memberHolderObject[member.name].call(null, ...args),
```
Ms2ger: Arrow function, rest arguments.

resources/idlharness.js
```diff
@@ -3010,7 +3018,143 @@ function IdlTypedef(obj)
...
+        assert_own_property(
+            self[this.name],
+            member.name,
+            `${this.name} does not have property ${format_value(member.name)}`);
+
+        var desc = Object.getOwnPropertyDescriptor(self[this.name], member.name);
+        // "The attribute setter is undefined if the attribute is declared
+        // readonly and has neither a [PutForwards] nor a [Replaceable]
+        // extended attribute declared on it."
+        assert_equals(desc.set, undefined, "setter must be undefined for readonly attributes");
```
Ms2ger: This unconditional check does not seem to match the quoted spec text.

Ms2ger (changes requested):

resources/idlharness.js

resources/idlharness.js

resources/idlharness.js
lukebjerring: Conditions added.

lukebjerring (commented):
@Ms2ger thanks for the thorough review.

Note that arrow functions and consts have been used throughout idlharness.js already (L56, L260, ...). Some of those were for new functionality I added
(e.g. #10381) and some pre-date me. What browser set are we intending to support? I was under the impression that we're running against the latest stable browsers?

lukebjerring:
Chromium CI dry run looks promising.
https://test-results.appspot.com/data/layout_results/linux_chromium_rel_ng/133564/site_per_process_webkit_layout_tests%20%28with%20patch%29/layout-test-results/results.html

jgraham:
@bzbarsky Do you think you are a good person to review idl harness changes, or should I stop suggesting it? :)

bzbarsky:
I  feel comfortable doing it.  The question is priority (on my end) and hence turnaround time...

So far it hasn't been too much of a burden so I'm happy to try to get them done as needed.

resources/idlharness.js
```diff
@@ -3012,5 +3021,151 @@ function IdlTypedef(obj)
...
+    {
+        assert_own_property(
+            self[this.name],
+            member.name,
+            this.name + ' does not have property ' + format_value(member.name));
+
+        // "The attribute setter is undefined if the attribute is declared
+        // readonly and has neither a [PutForwards] nor a [Replaceable]
+        // extended attribute declared on it."
+        if (!member.has_extended_attribute("PutForwards")
```
foolip: Note that [PutForwards] is for interfaces only and I filed https://github.com/heycam/webidl/issues/572 about [Replaceable] probably being nonsensical for namespace members. Given that namespace members have to be readonly, we could assert as much of the parsed IDL, and here simply assert that there is no setter unconditionally.

resources/idlharness.js
```diff
@@ -3012,5 +3021,151 @@ function IdlTypedef(obj)
...
+};
+//@}
+
+IdlNamespace.prototype.test = function ()
+//@{
+{
+    /**
+     * TODO(lukebjerring): Assert:
+     * - "Note that unlike interfaces or dictionaries, namespaces do not create types."
+     * - "Of the extended attributes defined in this specification, only the
```
foolip: Ah, so this is in support of dropping the Unforgeable bits commented on.

foolip (commented):
Great tests there, yay!

resources/idlharness.js
```diff
@@ -3012,5 +3021,151 @@ function IdlTypedef(obj)
...
+    assert_equals(
+        typeof memberHolderObject[member.name],
+        "function",
+         "property must be a function");
+
+    var args = member.arguments.map(function(a) {
+        return create_suitable_object(a.idlType);
+    });
+    this.array.assert_type_is(
+        memberHolderObject[member.name].call(null, args),
```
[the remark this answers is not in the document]
lukebjerring: Sure; removed.

resources/idlharness.js
lukebjerring: OK; changed to be unconditional that setter isn't defined.

resources/test/tox.ini
```diff
@@ -13,4 +13,4 @@ deps =
   selenium
   requests
 
-commands = pytest {posargs} -vv tests
+commands = pytest {posargs} -vv tests/functional/idlharness
```
foolip: This is local debugging accidentally committed, right?

foolip (approved):
LGTM % tox.ini

resources/test/tox.ini
lukebjerring: Yah, my bad.

lukebjerring:
@Ms2ger - did you have any further comments?

resources/idlharness.js
```diff
@@ -3012,5 +3021,151 @@ function IdlTypedef(obj)
...
+{
+    this.name = obj.name;
+    this.extAttrs = obj.extAttrs;
+    this.untested = obj.untested;
+    /** A back-reference to our IdlArray. */
+    this.array = obj.array;
+
+    /** An array of IdlInterfaceMembers. */
+    this.members = obj.members.map(m => new IdlInterfaceMember(m));
+    if (this.has_extended_attribute("Unforgeable")) {
```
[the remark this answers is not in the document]
bzbarsky: [Unforgeable] used to be able to appear on interfaces.  `Location` was the only consumer.  Then the IDL spec got changed to disallow that and do `Location` via hand-coded ES bits.

resources/idlharness.js
bzbarsky: [PutForwards] is only allowed for readonly attributes, so in theory it could be allowed on namespaces.  I don't think we should, though; will follow up in https://github.com/heycam/webidl/issues/572

Ms2ger (approved):

Ms2ger:
If @foolip's happy, I'm happy

PR #11781, opened 2018-07-04
[css-grid]  Parse various grid properties

ewilligers:
Spec: https://drafts.csswg.org/css-grid/

These tests highlight the following 5 issues:

Spec
[cssom] [css-grid] Serialization of custom identifiers
https://github.com/w3c/csswg-drafts/issues/2858

Spec
[css-grid] grid-line custom identifier auto?
https://github.com/w3c/csswg-drafts/issues/2856

Edge
grid-area should reject non-integer numbers
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/18112694/

Firefox
grid-auto-columns/rows should accept multiple track-size values
https://bugzilla.mozilla.org/show_bug.cgi?id=1339672

WebKit
[CSS Grid Layout] Fix grid-{row, column, area} shorthand CSSOM serialization
https://bugs.webkit.org/show_bug.cgi?id=149890

Note that some properties such as grid and grid-template are not yet tested.


css/css-grid/parsing/display-invalid.html
```diff
@@ -0,0 +1,19 @@
...
+<link rel="help" href="https://drafts.csswg.org/css-grid/#propdef-display">
+<meta name="assert" content="display supports the new values 'grid | inline-grid'.">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="support/parsing-testcommon.js"></script>
+</head>
+<body>
+<script>
+test_invalid_value("display", "grid inline-grid");
+test_invalid_value("display", "none grid");
```
mrego: Not a strong opinion but I don't see the need to have two separated tests valid vs invalid, instead of doing just one test checking both things.

That test could be name something like `grid-model/display-parsing-001.html` (the 001 suffix is always nice in case someone wants to add a new test about the same things in the future and won't need to do any rename).

css/css-grid/parsing/display-invalid.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>CSS Grid Layout Module Level 1: parsing display with invalid values</title>
```
mrego: Nit: On grid layout test suite titles usually start with: "CSS Grid Layout Test: ".

css/css-grid/parsing/display-invalid.html
```diff
@@ -0,0 +1,19 @@
...
+<link rel="author" title="Eric Willigers" href="mailto:ericwilligers@chromium.org">
+<link rel="help" href="https://drafts.csswg.org/css-grid/#propdef-display">
```
mrego: This should link to `css-grid-1`, as we're testing Level 1 of the spec.

Note that the link is broken the right one would be something like:
https://drafts.csswg.org/css-grid-1/#ref-for-propdef-display%E2%91%A0
Which is quite strange, maybe this is an issue in the spec. :smile:

css/css-grid/parsing/display-invalid.html
```diff
@@ -0,0 +1,19 @@
...
+<meta name="assert" content="display supports the new values 'grid | inline-grid'.">
```
mrego: This is not being checked in this test we should reword this sentence or move these checks to the other test (like I'm suggesting in the next comment).

css/css-grid/parsing/display-invalid.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
```
mrego: On Grid Layout suite, we have a subfolder for each spec section (using the anchor of the spec as folder name).
For example this test should go under `grid-model`.

css/css-grid/parsing/grid-area-invalid.html
```diff
@@ -0,0 +1,45 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>CSS Grid Layout Module Level 1: parsing grid-area with invalid values</title>
+<link rel="author" title="Eric Willigers" href="mailto:ericwilligers@chromium.org">
+<link rel="help" href="https://drafts.csswg.org/css-grid/#propdef-grid-area">
+<meta name="assert" content="grid-area supports only the grammar '<grid-line> [ / <grid-line> ]{0,3}'.">
```
mrego: This is testing more things than grid-area.

css/css-grid/parsing/grid-area-valid.html
```diff
@@ -0,0 +1,67 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>CSS Grid Layout Module Level 1: parsing grid-area with valid values</title>
+<link rel="author" title="Eric Willigers" href="mailto:ericwilligers@chromium.org">
+<link rel="help" href="https://drafts.csswg.org/css-grid/#propdef-grid-area">
+<meta name="assert" content="grid-area supports the full grammar '<grid-line> [ / <grid-line> ]{0,3}'.">
```
mrego: Ditto.

css/css-grid/parsing/grid-area-invalid.html
```diff
@@ -0,0 +1,45 @@
...
+test_invalid_value("grid-area", "auto 2 auto 4");
+test_invalid_value("grid-row", "33 -A0 auto");
+test_invalid_value("grid-row", "auto i 2 j span 3 k");
+
+// The following values are rejected by Blink/WebKit and accepted by Edge/Firefox
+// https://github.com/w3c/csswg-drafts/issues/2856
+// test_invalid_value("grid-row-end", "1 auto");
+// test_invalid_value("grid-row-end", "span 1 auto");
+// test_invalid_value("grid-row-end", "span auto 1");
+// test_invalid_value("grid-row-end", "1 auto span");
```
mrego: I won't comment out these cases if yout think it's the right thing. If the CSSWG resolves something different on the issue we would just need to update the tests later.

Also I'm not sure if adding comments about current behavior in browsers on the different tests make sense. When the implementations change the comments will get outdated.

css/css-grid/parsing/grid-area-valid.html
```diff
@@ -0,0 +1,67 @@
...
+test_valid_value("grid-area", "auto / i / 2 j", "auto / i / 2 j / i");
+test_valid_value("grid-area", "auto / i / 2 j / span 3 k");
+test_valid_value("grid-row", "auto / i");
+test_valid_value("grid-column", "2 j / span 3 k");
+
+
+// https://github.com/w3c/csswg-drafts/issues/2858
+// '\\31 st' in Blink, Firefox, '1st' in Edge, '"1st"' in Safari.
+test_valid_value("grid-column-end",  "\\31st", ["\\31 st", "1st", '"1st"']);
+test_valid_value("grid-column-end",  "\\31 st", ["\\31 st", "1st", '"1st"']);
```
mrego: For example here you're not commenting out these cases.

css/css-grid/parsing/grid-auto-columns-valid.html
```diff
@@ -0,0 +1,54 @@
+<!DOCTYPE html>
```
mrego: I guess it'd be nice to have also the `grid-auto-rows` version. I guess implementations would share the parsing logic, but having a test covering the other property seems like a good idea.

css/css-grid/parsing/grid-auto-flow-valid.html
```diff
@@ -0,0 +1,24 @@
...
+</head>
+<body>
+<script>
+test_valid_value("grid-auto-flow", "row");
+test_valid_value("grid-auto-flow", "column");
+test_valid_value("grid-auto-flow", "row dense");
+test_valid_value("grid-auto-flow", "dense column", "column dense");
+
+// Blink/WebKit "dense", Edge/Firefox "row dense"
+test_valid_value("grid-auto-flow", "dense", ["dense", "row dense"]);
```
mrego: Are Edge/Firefox right here? The spec says:
> Computed value: specified value

css/css-grid/parsing/grid-auto-rows-valid.html
```diff
@@ -0,0 +1,48 @@
+<!DOCTYPE html>
```
mrego: Again probably checking `grid-auto-columns` too would be nice.

css/css-grid/parsing/grid-template-areas-valid.html
```diff
@@ -0,0 +1,29 @@
...
+<script>
+test_valid_value("grid-template-areas", "none");
+test_valid_value("grid-template-areas", '"first"');
+test_valid_value("grid-template-areas", '"first second"');
+test_valid_value("grid-template-areas", '"1st 2nd 3rd"');
+test_valid_value("grid-template-areas", '"first second" "third fourth"');
+test_valid_value("grid-template-areas", '"first second" "third ." "1st 2nd" "3rd 4th"');
+
+// Firefox preserves specific whitespace and full stop sequences.
+// Other browsers consolidate to ' ' and '.'
```
mrego: Again spec says:
> Computed value: specified value

Is Firefox right or I'm missing something else.

css/css-grid/parsing/grid-template-areas-valid.html
```diff
@@ -0,0 +1,29 @@
...
+test_valid_value("grid-template-areas", '"  a  \t  b  "', '"a b"'); // Fails in Firefox
+test_valid_value("grid-template-areas", '"c\td"', ['"c d"', '"c\\9 d"']);
```
mrego: I don't get this `\\9` stuff.

mrego (changes requested):
Thanks for working on this and sorry for the long review but I have a bunch of things to comment. :innocent:

css/css-grid/parsing/support/parsing-testcommon.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
```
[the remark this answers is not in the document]
ewilligers: See https://github.com/web-platform-tests/wpt/pull/12465#issuecomment-413506239

Problems like https://github.com/web-platform-tests/wpt/issues/12419 are avoided by having the support directory adjacent to the tests.

css/css-grid/parsing/display-valid.html
```diff
@@ -0,0 +1,38 @@
...
+test_valid_value("display", "inline-table");
+test_valid_value("display", "table-row-group");
+test_valid_value("display", "table-header-group");
+test_valid_value("display", "table-footer-group");
+test_valid_value("display", "table-row");
+test_valid_value("display", "table-column-group");
+test_valid_value("display", "table-column");
+test_valid_value("display", "table-cell");
+test_valid_value("display", "table-caption");
+test_valid_value("display", "none");
```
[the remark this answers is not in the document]
ewilligers: Agreed. Moving.

css/css-grid/parsing/display-invalid.html
ewilligers: I find the separate files helpful when some browsers don't support a property.

css/css-grid/parsing/grid-auto-flow-valid.html
ewilligers: These tests check the serialization of the specified value, not the computed value.

I suspect the serialization should be "dense":
https://drafts.csswg.org/cssom/#serializing-css-values
> If component values can be omitted or replaced with a shorter representation without changing the meaning of the value, omit/replace them.

css/css-grid/parsing/grid-template-areas-valid.html
ewilligers: `\\9` is how some browsers are serializing the tab character.

css/css-grid/parsing/grid-auto-columns-valid.html
ewilligers: In css/css-backgrounds/parsing/border-radius-valid.html I test border-top-left-radius...border-bottom-left-radius as it is obvious where to look for the test.

For grid-auto-rows and grid-auto-columns, it seemed clearer to have separate files.

css/css-grid/parsing/grid-template-areas-valid.html
ewilligers: I tried to include a Firefox serialization, but the following error occurs:

```
test_valid_value("grid-template-areas", '"  a  \t  b  "', '" a \\9 b "');
test_valid_value("grid-template-areas", '"  a  \t  b  "', '"  a  \\9  b  "');

assert_equals: serialization should be canonical expected "\" a \\9 b \"" but got "\" a \\9 b \""
assert_equals: serialization should be canonical expected "\" a \\9 b \"" but got "\" a \\9 b \""
```

css/css-grid/parsing/grid-area-valid.html
ewilligers: I left these while waiting for a CSS WG decision. I suspect this will become

test_valid_value("grid-column-end",  "\\31st"); 

when we have a decision.

ewilligers:
Now using /css/support/parsing-testcommon.js

Please review.

css/css-grid/parsing/grid-auto-columns-valid.html
mrego: Ok, I didn't realize we have two files to cover columns and rows. Sorry about that.

css/css-grid/parsing/grid-auto-columns-valid.html
```diff
@@ -0,0 +1,54 @@
...
+test_valid_value("grid-auto-columns", "fit-content(1px)");
+test_valid_value("grid-auto-columns", "fit-content(2em)");
+test_valid_value("grid-auto-columns", "fit-content(calc(2em + 3ex))");
+test_valid_value("grid-auto-columns", "fit-content(4%)");
+
+test_valid_value("grid-auto-columns", "0px");
+test_valid_value("grid-auto-columns", "0%");
+test_valid_value("grid-auto-columns", "0fr");
+test_valid_value("grid-auto-columns", "minmax(auto, 0%)");
+test_valid_value("grid-auto-columns", "fit-content(0px)");
```
mrego: These 5 lines are not in the `grid-auto-rows` test, I guess we can copy there too.

css/css-grid/parsing/grid-area-valid.html
```diff
@@ -0,0 +1,67 @@
...
+// <grid-line> [ / <grid-line> ]{0,3}
+test_valid_value("grid-area", "auto / i", "auto / i / auto / i");
+test_valid_value("grid-area", "auto / i / 2 j", "auto / i / 2 j / i");
+test_valid_value("grid-area", "auto / i / 2 j / span 3 k");
+test_valid_value("grid-row", "auto / i");
+test_valid_value("grid-column", "2 j / span 3 k");
+
+
+// https://github.com/w3c/csswg-drafts/issues/2858
+// '\\31 st' in Blink, Firefox, '1st' in Edge, '"1st"' in Safari.
```
mrego: I'm not sure about this kind of comments about current results in different browsers,
because when browsers change the comment is going to get outdated.

css/css-grid/parsing/grid-auto-flow-valid.html
mrego: Then IMHO we shouldn't check both things and some browsers will see a failure and eventually fix the issue (or re-discuss the topic if they don't agree).

mrego (approved):
Thanks for your work on this. LGTM!

I added some minor comments inline.

PR #11798, opened 2018-07-05
Test that will-change:will-change has no effect.

heycam:


dbaron (approved):
This is fine, although one could test this in simpler ways since this is actually *syntactically* invalid.

PR #11801, opened 2018-07-05
[css-backgrounds] don't serialize as currentColor

ewilligers:
CSS keywords serialize as lower case.

https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/16905526/


mrego (approved):
LGTM.

Issue #7470, opened 2017-09-25
Document what tests in the multicolum test suite are about


frivoal:
Some tests in the multicol test suite have no indication of what it is they are testing, neither in the form of comments, nor of an explicit assertion.

This makes them hard to work with, and would be nice to fix. 

There's at least this one:
* https://github.com/w3c/web-platform-tests/blob/5b857bf79ede0c99761aca3d5562e86ba979e49f/css/css-multicol-1/multicol-rule-shorthand-2.xht

gsnedders:
There's a very slim chance https://github.com/operasoftware/presto-testo/tree/b5d5d519321bb6da4bc825f92c3f4e84db31be6f/css/multicolumn might be helpful, given they're ultimately derived from those tests (and specifically those in the reftest subdirectory).

TalbotG:
> Some tests in the multicol test suite have no indication of what it is they are testing, neither in the form of comments, nor of an explicit assertion.

Some tests in the multicol test suite give some indication (hint) of what they are testing in their filenames. Normally, unless the test is really basic, all the multicolumn tests that I reviewed and approved have useful comments or explicit assertion.

> There's at least this one:

    https://github.com/w3c/web-platform-tests/blob/5b857bf79ede0c99761aca3d5562e86ba979e49f/css/css-multicol-1/multicol-rule-shorthand-2.xht

I did not create that multicol-rule-shorthand-2 test.
I did not review that multicol-rule-shorthand-2 test but I can explain it.

The important lines in the test are:
line 28	`column-rule: solid blue 1em;`
line 29	`column-rule: normal red 1em;`
line 30	`column: normal red 1em;`

where lines 29 and 30 are invalid: 'normal' is not a border-style. 'normal' should only be used for 'column-gap'. So, there should be no red. There should be black (glyphs), gray (border), some yellow (background) and 3 short vertical stripes of blue (separators of column boxes)). So, on top of a few tweakings, I would add the invalid flag and add:

line 29	`column-rule: normal red 1em; /* invalid; 'normal' is not a 'border-style' */`
line 30	`column: normal red 1em; /* invalid; 'normal' can only apply to 'column-gap' */`

and **I would remove lines 32 to 45** since they are pointless and useless in the test. Giving something like:

http://www.gtalbot.org/BrowserBugsSection/CSS3Multi-Columns/Opera/multicol-rule-shorthand-2-GT.xht

[...]

Issue #7475, opened 2017-09-25
Surface test regressions in PRs


foolip:
Currently, Travis will check if tests are flaky and if so fail the build. Otherwise, it is considered a pass and surfaced as a comment linking to details. Example: https://github.com/w3c/web-platform-tests/pull/7472

In order to allow test writers to more easily understand the impact of their changes, we should also clearly surface when a new test is failing, or when an existing test goes from passing to failing.

To do this, Travis also has to run the tests once without the changes applied.

RByers:
+1, this has always been a key piece of what we were looking for from the PR tooling.  @bobholt discusses it a little [here](https://docs.google.com/document/d/1HK5roexqU2Nd2iTMKO4UM5Gc3b5vuFm06J3E06iGcxM/edit#), and I think we had some earlier docs about this but I can't find them at the moment.

When a test goes from failing to passing do we also want to surface it in the same way?  Eg. I may make some test changes specifically to get the test passing on a number of browsers, it should be easy to tell at a glance if I succeeded at that.

foolip:
I think that we want to easily see the old and new status of any test that has been touched, including if the old or new status is "doesn't exist".

Probably we will run into the GitHub comment size limit, and the full view will have to be on pulls.web-platform-tests.org. That should always be linked in a comment like https://github.com/w3c/web-platform-tests/pull/7506#issuecomment-332517463

Then, we should decide which kinds of changes warrant another comment pointing out that something unusual is happening. That includes at least going passing to failing in some browser, and of course going from non-flaky to flaky.

foolip:
I'm preparing a presentation for https://webengineshackfest.org/ and made this mockup:
<img width="788" alt="screen shot 2017-10-01 at 2 36 43 pm" src="https://user-images.githubusercontent.com/498917/31054602-480e417c-a6b6-11e7-814f-15cb4597d347.png">

jgraham:
I have some concerns about this.

The easiest solution would just be to run all the tests twice on travis. But we already have capacity issues and I don't fancy adding an extra job, or additional work to the existing jobs. I'm also not sure that this actually helps in most cases. In particular who will see these results? For gecko developers we want something like this, but on the bug we generate when the PR is imported. In theory if the data is generated upstream it might be possible to use that, but certainly having it as a GH comment alone isn't going to get visibility in the right places.

I can see three possible parts to this:

[...]

Issue #5319, opened 2017-04-03
Stability checker should run for reference changes


gsnedders:
cc/ @jgraham  @jugglinmike

bobholt:
@gsnedders: can you elaborate on this? Sounds like it was parrt of a discussion I missed.

gsnedders:
Given a reftest a.html with `<link rel=match href=a-ref.html>`, the stability checker will run the test if there is a change to a.html, but won't run the test if there is a change to a-ref.html.

gsnedders:
(Of course the problem is harder than that because of the possibility of chaining references, where a-ref.html would have another `<link rel=match>`.)

foolip:
@lukebjerring is this something you could look into as part of your work on https://github.com/w3c/web-platform-tests/issues/7475? I'm calling this roadmap on the assumption that will make sense.

JKereliuk:
@lukebjerring any update on this?

lukebjerring:
No, I haven't looked at it yet, will assign to self though (it's in the pile)

JKereliuk:
@lukebjerring have you had a chance to look at this yet?

lukebjerring:
Not as yet.

alijuma:
Re-ping about an update since this has prioirty:roadmap.

lukebjerring:
Dropping that label; this is a side-issue barely related to a P2 OKR.

foolip:
@gsnedders given changes in the manifest format since this issue was filed, is this now easier to implement?

gsnedders:
Not sure. It's been a long time since I've touched the code that finds how we find affected tests.

foolip:
It looks like this has been fixed actually. Here's how I tried it:
```sh
echo >> css/css-grid/table-grid-item-dynamic-001-ref.html
git commit -a -m bla
./wpt tests-affected HEAD~
```

That lists `css/css-grid/table-grid-item-dynamic-001.html` as affected!

PR #11802, opened 2018-07-05
Multicol width tests

rachelandrew:
A set of cleaned up tests for multicol:

1. Added asset where it was missing.
2. Added a description to the page on tests to explain what a pass was.
3. Removed CSS that wasn't required (orphans and widows properties)
4. Tidied up multicol-width-003 as discussed in https://github.com/web-platform-tests/wpt/issues/7470
5. Changed multicol-width-count-002 so that it fails if column-count is ignored and column-width used otherwise it appears to test the same thing as multicol-width-count-001


mstensho (approved):

gsnedders:
Is there any reason this hasn't landed?

mstensho:
There's a mistake in this change; got automatically reported in Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=880084

I'll fix it shortly (via the Chromium repo). But my question is: Aren't automatic tests run before landing pull requests?

gsnedders:
@mstensho we don't currently do any differential comparison between how browsers pass/fail before/after a change (that's #7475), mostly because we don't store the results in the Travis environment anywhere to do a comparison against the merge base, and we don't even do the stability checking properly for reference changes (#5319).

mstensho:
@gsnedders That's a pity. Such mistakes are so easy to make, and could be so easy to detect as well.

Landing a fix now: https://chromium-review.googlesource.com/c/chromium/src/+/1203250

PR #11806, opened 2018-07-05
[WIP] Prototype HTTP/2.0 Server

dheiberg:
Hey, so this is a prototype/proof of concept of how to integrate HTTP/2.0 into the current wptserve infrastructure.

I'm attempting to use as much of the existing infrastructure as possible, as it gives me functionality for free (such as handling TCP connections and threading client connections), and also means I don't have to re-invent the wheel and risk making mistakes in the various bits of infrastructure I might have to redo. 

Currently, this prototype is able to serve directories, handle multiple h2 connections simultaneously, as well as h2 connections alongside h1 requests. Obviously it is going to have many bugs, but please point out anything glaring that doesn't have a note in the comments. 

So the general idea here is to launch a H2 server on a port, 9000 currently, using the existing `WebTestServer` in `server.py` by wrapping the socket with an ssl context that sets up the ALPN protocol to negotiate H2. Because of this, the server has to be run on python 2.7.15+, as it requires OpenSSL 1.0.2+. When the server receives the first request/data from the client, it will enter the `handle_one_request` method in `WebTestRequestHandler`, and in here I simply move to a H2 handler: `handle_h2`. 

In `handle_one_request`, I have moved the logic that handles the parsed request to another method: `finish_handling`. This is because the H2 Handler will be reusing that logic, so by placing this in a separate method, both H1 and H2 handlers are able to use that. 

The `handle_h2` method is fairly straightforward, it initiates the HTTP/2.0 connection, and then enters a loop that continuously reads directly from the socket and parses the H2 frames. This changes the way that the original handler works, as it now handles the entire connection from start to finish, instead of a single request at a time in a one and done fashion. 

The `Request` object did not currently need to be changed much, just adding a `h2_stream_id` attribute. On the Response side, I created a `H2Response` object which mostly just changes which writer methods are called. The `write_content` method in here is a bit hacky, uses a `try/except` to check for the last iteration of the generator. This is because you need to know what the last item from the generator is, as you need to flag the end of the stream. Please let me know if you know of a better way!  The `H2ResponseWriter` handles the construction of the H2 frames and sending them via the socket. 

Please let me know what you think. This currently feels like the right approach to implementing this solution, but would love some extra eyes!

P.S. The large additions and file changes are due to adding the dependencies (h2, hyperframe, hpack, enum and certifi) in the third_party folder. To view the actual changes that implement the H2 server, view the changes made by the first commit. Also sorry for mini WOT.

Edit: I have made it so the h2 server is no longer started by default, since it is not ready for use and should be explicitly started. It is now started by using `./wpt serve --h2`

---
This change is [Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/11806)


jugglinmike:
@sideshowbarker @bmac This may be relevant to your efforts on w3c-test.org

jgraham (commented):
> Reviewed 11 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/11806)* status: 3 of 184 files reviewed, 10 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/serve/serve.py, line 499 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGjp8201gYDSsOejFVV:-LGjp8221Kz819IjFaSD:b-tj70s2) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/serve/serve.py#L499)):*
> ```Python
> 
>         # Check that python 2.7.15 is being used for HTTP/2.0
>         if scheme == 'http2':
> ```

This obviously isn't a good long term solution, but it's fine for now. But maybe add a `TODO` comment about removing it?
___
*[tools/serve/serve.py, line 738 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGjqNQ-BI_ULZGOoiyR:-LGjqNQ0DOlAV1rVPB0P:b6jef43) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/serve/serve.py#L738)):*
> ```Python
>             "ws": ["auto"],
>             "wss": ["auto"],
>             "http2":[9000],
> ```

I would like us to not support this by default and require a command line flag or something similar in order to get the server to launch durning development.
___
*[tools/serve/serve.py, line 763 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGjq_ekEQQu0JAwGcup:-LGjq_el1WR8S16Y421-:b-3y4oo7) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/serve/serve.py#L763)):*
> ```Python
>     def __init__(self, *args, **kwargs):
>         # Skip if not version 2.7.15
>         if not (sys.version_info[0] == 2 and sys.version_info[1] == 7 and sys.version_info[2] == 15):
> ```

Can we create a `supports_http2` function to encapsulate this check since it's used in multiple places?
___
*[tools/wptserve/wptserve/response.py, line 73 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGjrAKZB0u0UaKqfbg2:-LGjrAKZB0u0UaKqfbg3:basm6m3) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/response.py#L73)):*
> ```Python
> 
>         self.logger = get_logger()
>         if h2:
> ```

I am not sure, but I feel like having a base class and HTTP/1 and HTTP/2 variants would be better than baking the knoledge into the constructor.
___
*[tools/wptserve/wptserve/response.py, line 392 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkkiUyChWWoNemu6Ms:-LGkkiUz7bv3gC0v5WoF:bajab5c) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/response.py#L392)):*
> ```Python
>         self.logger = response.logger
> 
>     def write_headers(self, headers, status_code, status_message=None):
> ```

We probably need to make this more explicit, right? Like it seems useful to test how things work if we send headers in the "wrong" order &c.
___
*[tools/wptserve/wptserve/server.py, line 191 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkWJwT8rZcZSLUvyCl:-LGkWJwT8rZcZSLUvyCm:b-7e2fzz) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/server.py#L191)):*
> ```Python
>                 ssl_context.set_alpn_protocols(['h2'])
>                 self.socket = ssl_context.wrap_socket(self.socket,
>                                             server_side=True)
> ```

This indentation is wrong (also elsewhere)
___
*[tools/wptserve/wptserve/server.py, line 382 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkVavv7OwCeRjTzHns:-LGkVavxEQGn1qwT2Xb-:b-omb3a2) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/server.py#L382)):*
> ```Python
>                 data = self.request.recv(65535)
> 
>                 # self.logger.debug('(%s) Data: ' % (str(uid)) + str(data.encode('hex')))
> ```

Remove this unless it can be enabled
___
*[tools/wptserve/wptserve/server.py, line 387 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkWUtd8m8ojD3-RZLb:-LGkWUtd8m8ojD3-RZLc:b-kekjzk) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/server.py#L387)):*
> ```Python
> 
>                 for event in events:
>                     if isinstance(event, RequestReceived):
> ```

So, this only supports a single simulaneous stream and a single set of headers (no body) at the moment, right? It seems like we are going to want ot make the Request object take a queue of frames for the associated stream and/or have a handler type that can react directly to events rather than just being a single function call.
___
*[tools/wptserve/wptserve/server.py, line 410 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkPaRQ--opKKXJi16-:-LGkPaRR9ta59lxI_qbM:bceleuk) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/server.py#L410)):*
> ```Python
> 
> 
> class H2Headers(dict):
> ```

Does this mean we don't have any way to access the raw headers? Seems like we should probably find a way to expose that.
___
*[tools/wptrunner/wptrunner/environment.py, line 140 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkPQ7H0D2VrbLxeLRH:-LGkPQ7IBiNJTDXdewZn:b-gmyooy) ([raw file](https://github.com/web-platform-tests/wpt/blob/5c05fc07d58b1352de26f62434103c8ec3941749/tools/wptrunner/wptrunner/environment.py#L140)):*
> ```Python
>             "https": [8443],
>             "ws": [8888],
>             "wss": [8889],
> ```

Seems like an unnecessary change.

dheiberg (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/11806)* status: 3 of 184 files reviewed, 10 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/serve/serve.py, line 499 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGjp8201gYDSsOejFVV:-LGyYsFM6EhSfmbcLb5J:b-ul8jrf) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/serve/serve.py#L499)):*

This obviously isn't a good long term solution, but it's fine for now. But maybe add a `TODO` comment about removing it?
</blockquote></details>

Good idea
___
*[tools/serve/serve.py, line 738 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGjqNQ-BI_ULZGOoiyR:-LGyWwQz9sG8jdjwBard:bu3fo4t) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/serve/serve.py#L738)):*

I would like us to not support this by default and require a command line flag or something similar in order to get the server to launch durning development.
</blockquote></details>

This has now been changed, the HTTP/2.0 server is started using a command line flag
___
*[tools/serve/serve.py, line 763 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGjq_ekEQQu0JAwGcup:-LGyX3Sn-aaY2IPbwubr:bqud23f) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/serve/serve.py#L763)):*

Can we create a `supports_http2` function to encapsulate this check since it's used in multiple places?
</blockquote></details>

Yup, I'll find somewhere useful to put that and refactor this
___
*[tools/wptserve/wptserve/response.py, line 73 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGjrAKZB0u0UaKqfbg2:-LGyXFLZBjA2y-eir594:bqa7r20) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/response.py#L73)):*

I am not sure, but I feel like having a base class and HTTP/1 and HTTP/2 variants would be better than baking the knoledge into the constructor.
</blockquote></details>

I agree, so something along the lines of making this class the base class, and move the handle functions into specific classes for each protocol version? 
___
*[tools/wptserve/wptserve/response.py, line 392 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkkiUyChWWoNemu6Ms:-LGyXZQI6B3RRsE1W-WT:b-c08wci) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/response.py#L392)):*

We probably need to make this more explicit, right? Like it seems useful to test how things work if we send headers in the "wrong" order &c.
</blockquote></details>

So that is something that requires further testing. I believe headers can be split up and sent in different frames, using CONTINUATION frames. This is some needed functionality so I will look into how to introduce this. 
___
*[tools/wptserve/wptserve/server.py, line 191 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkWJwT8rZcZSLUvyCl:-LGyV8wS3sRde4Qbcdg6:b-4p0xjo) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/server.py#L191)):*

This indentation is wrong (also elsewhere)
</blockquote></details>

Noted, will fix on next run through
___
*[tools/wptserve/wptserve/server.py, line 382 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkVavv7OwCeRjTzHns:-LGyVGIf-5-Vlpa8ZSND:bebrt1w) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/server.py#L382)):*

Remove this unless it can be enabled
</blockquote></details>

Will do
___
*[tools/wptserve/wptserve/server.py, line 387 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkWUtd8m8ojD3-RZLb:-LGyW-Au6jbkhTAdLmdn:bh2lt3l) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/server.py#L387)):*

So, this only supports a single simulaneous stream and a single set of headers (no body) at the moment, right? It seems like we are going to want ot make the Request object take a queue of frames for the associated stream and/or have a handler type that can react directly to events rather than just being a single function call.
</blockquote></details>

Yes, so I am currently working on multithreading the request handling, and need to do some more testing to see how POST requests will behave. I currently don't believe that the Request object needs to be altered dramatically, as the server knows when it has received a full request when the client signals the end of a stream, but this is something to consider as I continue expanding the functionality.
___
*[tools/wptserve/wptserve/server.py, line 410 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkPaRQ--opKKXJi16-:-LGyVVNsEq6lAR9eYpeR:b-b7z91g) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/server.py#L410)):*

Does this mean we don't have any way to access the raw headers? Seems like we should probably find a way to expose that.
</blockquote></details>

They are exposed, but they are in HTTP/2.0 format. For example, 'path' in HTTP/1.1 becomes ':path' in HTTP/2.0. This class is used as a way to convert ':' prefixed headers into their HTTP/1.1 format, so that the rest of the system can utilize them without knowing the difference. 
___
*[tools/wptrunner/wptrunner/environment.py, line 140 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkPQ7H0D2VrbLxeLRH:-LGyUQdk1IMwWtKihQob:b-86tmdr) ([raw file](https://github.com/web-platform-tests/wpt/blob/5c05fc07d58b1352de26f62434103c8ec3941749/tools/wptrunner/wptrunner/environment.py#L140)):*

Seems like an unnecessary change.
</blockquote></details>

I believe this was leftover from when I had the "http2" key entered, must've forgotten to remove this change. Will touchup in next push

jgraham (commented):
> Reviewed 1 of 12 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/11806)* status: 5 of 184 files reviewed, 7 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/server.py, line 387 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkWUtd8m8ojD3-RZLb:-LGzdrOv-hcsQwHXrgEs:b4f0o5j) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/server.py#L387)):*

Yes, so I am currently working on multithreading the request handling, and need to do some more testing to see how POST requests will behave. I currently don't believe that the Request object needs to be altered dramatically, as the server knows when it has received a full request when the client signals the end of a stream, but this is something to consider as I continue expanding the functionality.
</blockquote></details>

OK, let's address that in a followup.
___
*[tools/wptserve/wptserve/server.py, line 410 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkPaRQ--opKKXJi16-:-LGzdc940I41CRYce3X6:bptqckr) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/server.py#L410)):*

They are exposed, but they are in HTTP/2.0 format. For example, 'path' in HTTP/1.1 becomes ':path' in HTTP/2.0. This class is used as a way to convert ':' prefixed headers into their HTTP/1.1 format, so that the rest of the system can utilize them without knowing the difference. 
</blockquote></details>

Right, the point is sometimes we might care about the difference. This seems like reasonable default behaviour, but we need a way to get the actual string that was sent over the wire.

dheiberg (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/11806)* status: 5 of 184 files reviewed, 7 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/server.py, line 410 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkPaRQ--opKKXJi16-:-LGzeNUXBaGiJiWMAe21:bbuahsh) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/server.py#L410)):*

Right, the point is sometimes we might care about the difference. This seems like reasonable default behaviour, but we need a way to get the actual string that was sent over the wire.
</blockquote></details>

I see your point. I can add some sort of `raw_header` attribute that contains the actual header keys as sent by the client?

jgraham (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/11806)* status: 5 of 184 files reviewed, 7 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/server.py, line 410 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkPaRQ--opKKXJi16-:-LGzgjKH-r5FmgLNqSUA:brxp9sq) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/server.py#L410)):*

I see your point. I can add some sort of `raw_header` attribute that contains the actual header keys as sent by the client?
</blockquote></details>

Sounds good.

jgraham (commented):
> Reviewed 22 of 188 files at r4, 2 of 2 files at r5, 184 of 184 files at r6.
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/11806)* status: all files reviewed, 10 unresolved discussions (waiting on @dhdavvie, @jgraham, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/response.py, line 11 at r6](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LHYOaeD0y092b84ShMf:-LHYOaeE5MSUxuZT0YlP:bxm73qc) ([raw file](https://github.com/web-platform-tests/wpt/blob/985440fc8b60bcc17e38f76a3326c4c05e347f38/tools/wptserve/wptserve/response.py#L11)):*
> ```Python
> 
> try:
>     from StringIO import StringIO
> ```

Do we actually just want `from io import BytesIO`? I feel like that should work everywhere?
___
*[tools/wptserve/wptserve/response.py, line 409 at r6](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LHYP0Sz0fLCZgi45PyJ:-LHYP0Sz0fLCZgi45PyK:b-214z2u) ([raw file](https://github.com/web-platform-tests/wpt/blob/985440fc8b60bcc17e38f76a3326c4c05e347f38/tools/wptserve/wptserve/response.py#L409)):*
> ```Python
>         formatted_headers.extend(secondary_headers)
> 
>         self.h2_lock.acquire()
> ```

So, I asusme the deal here is that `h2conn` is single threaded?

Instead of having a lock used directly like this, my favourite Python pattern is to have a guard object that acts as a context manager, aquiring the lock, and giving back the lock-protected object, like

```
class ConnectionGuard(object):
    lock = threading.Lock

    def __init__(self, obj):
        self.obj = obj

    def __enter__(self):
        self.lock.acquire()
        return self.obj

    def __exit__(self, *args):
        self.lock.release()
```

Then you do something like:
```
with self.connection as connection:
    connection.send_headers([…])
# Maybe allow a connection to be passed in to self.write if there's a performance issue
# having to re-acquire the  lock
self.write()
```
___
*[tools/wptserve/wptserve/response.py, line 434 at r6](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LHYSbNb9DV6Sz09HBYv:-LHYSbNb9DV6Sz09HBYw:bgmvssb) ([raw file](https://github.com/web-platform-tests/wpt/blob/985440fc8b60bcc17e38f76a3326c4c05e347f38/tools/wptserve/wptserve/response.py#L434)):*
> ```Python
>             self._write_content(data.read(payload_size), False)
>             data_len -= payload_size
>             payload_size = self.get_max_payload_size()
> ```

Why do we care about the maxmum payload size changing, but not the data changing length?
___
*[tools/wptserve/wptserve/response.py, line 438 at r6](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LHYSRRu-fnn0TI6pFcb:-LHYSRRu-fnn0TI6pFcc:b-tzdmb4) ([raw file](https://github.com/web-platform-tests/wpt/blob/985440fc8b60bcc17e38f76a3326c4c05e347f38/tools/wptserve/wptserve/response.py#L438)):*
> ```Python
>         self._write_content(data.read(), last)
> 
>     def _write_content(self, data, last):
> ```

Can we call this `write_content_frame` and make it public, so that a user could e.g. write a too-long frame or something as test.
___
*[tools/wptserve/wptserve/server.py, line 212 at r6](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LHYSug7-Tfi08wBpwzf:-LHYSug7-Tfi08wBpwzg:bsg7zgf) ([raw file](https://github.com/web-platform-tests/wpt/blob/985440fc8b60bcc17e38f76a3326c4c05e347f38/tools/wptserve/wptserve/server.py#L212)):*
> ```Python
> 
> 
> class _WebTestRequestHandler(BaseHTTPServer.BaseHTTPRequestHandler):
> ```

Use `Base` rather than `_`
___
*[tools/wptserve/wptserve/server.py, line 307 at r6](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LHYTvmCBMPVGjl1lViS:-LHYTvmDCv1FAhPH5e_o:be1hziw) ([raw file](https://github.com/web-platform-tests/wpt/blob/985440fc8b60bcc17e38f76a3326c4c05e347f38/tools/wptserve/wptserve/server.py#L307)):*
> ```Python
> 
> class Http2WebTestRequestHandler(_WebTestRequestHandler):
> ```

Unnecessary empty line.
___
*[tools/wptserve/wptserve/server.py, line 344 at r6](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LHYUckT6sTVVKLFR5kq:-LHYUckT6sTVVKLFR5kr:b-ap69pd) ([raw file](https://github.com/web-platform-tests/wpt/blob/985440fc8b60bcc17e38f76a3326c4c05e347f38/tools/wptserve/wptserve/server.py#L344)):*
> ```Python
>                         self.logger.debug('(%s) Parsing RequestReceived' % (uid))
>                         self._h2_parse_request(event)
>                         t = threading.Thread(target=_WebTestRequestHandler.finish_handling, args=(self, True, H2Response))
> ```

I think we want one thread per `stream_id` and a queue to send the data to that thread, right? Is that coming in the next part? Maybe add a TODO for now.
___
*[tools/wptserve/wptserve/server.py, line 370 at r6](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LHYUMcE6F5eTkAO2sW0:-LHYUMcE6F5eTkAO2sW1:b-fqnklb) ([raw file](https://github.com/web-platform-tests/wpt/blob/985440fc8b60bcc17e38f76a3326c4c05e347f38/tools/wptserve/wptserve/server.py#L370)):*
> ```Python
> class H2Headers(dict):
>     def __init__(self, headers):
>         self.raw_headers = {}
> ```

`OrderedDict`, perhaps.

dheiberg (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/11806)* status: all files reviewed, 10 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/response.py, line 11 at r6](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LHYOaeD0y092b84ShMf:-LHYYUOu1G6GZGZ79Mz2:b-pk2ztn) ([raw file](https://github.com/web-platform-tests/wpt/blob/985440fc8b60bcc17e38f76a3326c4c05e347f38/tools/wptserve/wptserve/response.py#L11)):*

Do we actually just want `from io import BytesIO`? I feel like that should work everywhere?
</blockquote></details>

Yup, that's a much better solution, thanks!
___
*[tools/wptserve/wptserve/response.py, line 409 at r6](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LHYP0Sz0fLCZgi45PyJ:-LHYYxgGEOBl90x7oLW7:b-hob8ba) ([raw file](https://github.com/web-platform-tests/wpt/blob/985440fc8b60bcc17e38f76a3326c4c05e347f38/tools/wptserve/wptserve/response.py#L409)):*

So, I asusme the deal here is that `h2conn` is single threaded?

Instead of having a lock used directly like this, my favourite Python pattern is to have a guard object that acts as a context manager, aquiring the lock, and giving back the lock-protected object, like

```
class ConnectionGuard(object):
    lock = threading.Lock

    def __init__(self, obj):
        self.obj = obj

    def __enter__(self):
        self.lock.acquire()
        return self.obj

    def __exit__(self, *args):
        self.lock.release()
```

Then you do something like:
```
with self.connection as connection:
    connection.send_headers([…])
# Maybe allow a connection to be passed in to self.write if there's a performance issue
# having to re-acquire the  lock
self.write()
```
</blockquote></details>

This looks very neat, I will look into implementing something like this. Should I push that to this PR or save it for the next?
___
*[tools/wptserve/wptserve/response.py, line 434 at r6](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LHYSbNb9DV6Sz09HBYv:-LHYZ7IwE1PIR8kIdM2F:bg0c6ly) ([raw file](https://github.com/web-platform-tests/wpt/blob/985440fc8b60bcc17e38f76a3326c4c05e347f38/tools/wptserve/wptserve/response.py#L434)):*

Why do we care about the maxmum payload size changing, but not the data changing length?
</blockquote></details>

`payload_size` is the largest size the next payload can be, which can change depending on stream capacity and WindowUpdate events received from the client. So this checks how big the next chunk of data to be sent down the pipe will be, sends that chunk, and decreases the counter keeping track of how much data is left. 
___
*[tools/wptserve/wptserve/response.py, line 438 at r6](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LHYSRRu-fnn0TI6pFcb:-LHYZoY55OklZRkllDKl:bvljwg2) ([raw file](https://github.com/web-platform-tests/wpt/blob/985440fc8b60bcc17e38f76a3326c4c05e347f38/tools/wptserve/wptserve/response.py#L438)):*

Can we call this `write_content_frame` and make it public, so that a user could e.g. write a too-long frame or something as test.
</blockquote></details>

I don't see why not
___
*[tools/wptserve/wptserve/server.py, line 382 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LGkVavv7OwCeRjTzHns:-LHYe9TM-x2kLHRYHuit:b-896fix) ([raw file](https://github.com/web-platform-tests/wpt/blob/2bbc07de9a3db615f390c1786938276e7e100c6e/tools/wptserve/wptserve/server.py#L382)):*

Will do
</blockquote></details>

Done.
___
*[tools/wptserve/wptserve/server.py, line 344 at r6](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LHYUckT6sTVVKLFR5kq:-LHY_OYkDeRO0LCajpcU:bgkhzld) ([raw file](https://github.com/web-platform-tests/wpt/blob/985440fc8b60bcc17e38f76a3326c4c05e347f38/tools/wptserve/wptserve/server.py#L344)):*

I think we want one thread per `stream_id` and a queue to send the data to that thread, right? Is that coming in the next part? Maybe add a TODO for now.
</blockquote></details>

For the time being, it seems to make more sense to have a thread per entire request-response interaction and kill the thread after the interaction is done. One reasoning behind this is that one interaction could happen over multiple streams if PUSH frames are involved for example. It seems like it would add more complications to have threads per stream waiting for something to happen on that stream (which may never happen, Firefox often avoid reusing stream_ids within a certain timeframe by the looks of it). This may be a thing to consider in the future, but it currently seems like keeping the threads to a request-response scenario makes the most sense. 
___
*[tools/wptserve/wptserve/server.py, line 370 at r6](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LHYUMcE6F5eTkAO2sW0:-LHYdwz23PATHRBy55oe:b-ul8jrf) ([raw file](https://github.com/web-platform-tests/wpt/blob/985440fc8b60bcc17e38f76a3326c4c05e347f38/tools/wptserve/wptserve/server.py#L370)):*

`OrderedDict`, perhaps.
</blockquote></details>

Good idea

jgraham (commented):
> Reviewed 3 of 186 files at r7, 184 of 184 files at r8.
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/11806)* status: all files reviewed, 2 unresolved discussions (waiting on @jgraham, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/response.py, line 409 at r6](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LHYP0Sz0fLCZgi45PyJ:-LHmAu2JA3jZBABdrdYQ:bg8emls) ([raw file](https://github.com/web-platform-tests/wpt/blob/985440fc8b60bcc17e38f76a3326c4c05e347f38/tools/wptserve/wptserve/response.py#L409)):*

This looks very neat, I will look into implementing something like this. Should I push that to this PR or save it for the next?
</blockquote></details>

I think it's a relatively small change to add it to this PR.

dheiberg (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/11806)* status: all files reviewed, 2 unresolved discussions (waiting on @jgraham, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/response.py, line 409 at r6](https://reviewable.io/reviews/web-platform-tests/wpt/11806#-LHYP0Sz0fLCZgi45PyJ:-LHmsjD087gbyIK4cY7I:b-896fix) ([raw file](https://github.com/web-platform-tests/wpt/blob/985440fc8b60bcc17e38f76a3326c4c05e347f38/tools/wptserve/wptserve/response.py#L409)):*

I think it's a relatively small change to add it to this PR.
</blockquote></details>

Done.

jgraham (approved):
> Reviewed 4 of 188 files at r9, 184 of 184 files at r10.
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/11806)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @AutomatedTester, @annevk, @gsnedders, and @andreastt)

PR #11815, opened 2018-07-05
Update the longtask-timing IDL file

lukebjerring:
Hello reviewer(s),

Please check that:


lukebjerring:
@kereliuk - an example of an API which is a bit more complicated to get the objects for `add_object`.

longtask-timing/idlharness.window.js
```diff
@@ -0,0 +1,46 @@
...
+
+  const taskTime = 100;
+  const longTask = () => {
+    var begin = self.performance.now();
+    while (self.performance.now() < begin + taskTime);
+  }
+  t.step_timeout(longTask, 0);
+
+  const timeout = new Promise(
+      (_, reject) => t.step_timeout(reject, 2 * taskTime));
```
npm1: Is this needed? This seems to assume that the part of the test that does not involve the long task takes less than 100ms.

npm1 (approved):

longtask-timing/idlharness.window.js
lukebjerring: It's supposed to just be a fast-fail really. Yeah, it does assume that, though `step_timeout` is enforced so that the settings scale it for slower environments:

> The difference between setTimeout and step_timeout is that the latter takes account of the timeout multiplier when computing the delay; e.g., in the above case a timeout multiplier of 2 would cause a pause of 4000ms before calling the callback. This makes it less likely to produce unstable results in slow configurations.

https://web-platform-tests.org/writing-tests/testharness-api.html#timeouts-in-tests

That said, this could be a case of "works on my box" at its finest.. 50ms is the threshold for a "long" task. Doubled it and .. doubled that. lol. Want me to bump it up to a higher value?

npm1: The assumption that a 100ms task is considered a longtask seems fine. The assumption that a valid test run will take less than 200ms less so. I'd prefer some simple check to fail right away when longtasks are not supported at all. With this, not sure this early rejection is needed. Anyways, this is still better than the other longtask tests that look like they'll just Timeout in unsupported browsers. So I'm ok with an increase in the rejection time, which is currently 200 ms (or just start the rejection clock after executing the long task).

lukebjerring: kk - In https://github.com/web-platform-tests/wpt/pull/11815/commits/72262b4f8bb72a5353fda441ada7124492167b4b I've try-catched the object creation, changed the timeout to 1000ms, and run the (partial) IdlArray.test on rejection. 

In FF, it's 23 tests (18 fails, 5 passes) and Chrome it's 32 (all passes).

PR #11818, opened 2018-07-05
Updating OS identifier in WebDriver control+click tests

jimevans:
The identifier for the platform/OS should be "windows", not "windows_nt".

### DIFF
--- a/webdriver/tests/actions/control_click.py
+++ b/webdriver/tests/actions/control_click.py
@@ -18,7 +18,7 @@ def test_control_click(session, test_actions_page, key_chain, mouse_chain, modif
     outer = session.find.css("#outer", all=False)
     mouse_chain.click(element=outer)
     session.actions.perform([key_chain.dict, mouse_chain.dict])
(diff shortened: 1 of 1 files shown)


andreastt (approved):
This is already fixed in a downstream patch in central, but I don’t suppose there will be any merge conflicts.

cc @jgraham

whimboo:
There are even more places which have the wrong platformName listed in the wdspec tests. All of those got fixed via https://hg.mozilla.org/integration/autoland/rev/16ba5f9ae84e which hopefully gets synced upstream soon. I would just close this PR.

andreastt:
It appears that Chrome is flipping between error and fail here, which I’m not sure is a blocker for landing this as it is mostly likely an implementation problem. https://bugzilla.mozilla.org/show_bug.cgi?id=1470646 that @whimboo referred to would presumably exhibit the same problem.

jimevans:
Closing as subsequent changes have made this PR obsolete.

PR #11819, opened 2018-07-06
Update the cssom IDL file

lukebjerring:
Hello reviewer(s),

Please check that:

### DIFF
--- a/css/cssom/interfaces.html
+++ b/css/cssom/interfaces.html
@@ -35,7 +35,7 @@ const waitForLoad = new Promise(r => { addEventListener('load', r); });
 idl_test(
   ['cssom'],
   ['SVG', 'uievents', 'html', 'dom'],
-  idlArray => {
+  async idlArray => {
     idlArray.add_objects({
       Document: ['document', 'new Document()'],
       StyleSheetList: ['document.styleSheets'],
@@ -65,6 +65,7 @@ idl_test(
       SVGElement: ['svg_element'],
     });
 
+    await waitForLoad;
     self.style_element = document.getElementById('styleElement');
     self.sheet = style_element.sheet;
     self.svg_element = document.getElementById('svgElement');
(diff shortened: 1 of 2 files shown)


lukebjerring:
Note that the `namespace` chunk is manually commented out, as it's blocked by https://github.com/web-platform-tests/wpt/issues/7583

lukebjerring:
This is now ready for review.

foolip:
Tried the changes locally and it looks good, there are slightly fewer tests but it seems to be because tests like "Unscopable handled correctly for escape(CSSOMString) on CSS" no longer exist. ("CSS namespace: operation escape(CSSOMString)" does.) The number of failing tests is the same.

I expected the switch from `interface CSS` to `namespace CSS` to break other tests because css-conditional.idl and css-paint-api.idl still have `partial interface CSS`, but that didn't happen for either test, the test names just changed and still work. Filed: https://github.com/web-platform-tests/wpt/issues/12123

Also updated some specs:
https://github.com/w3c/css-houdini-drafts/pull/784
https://github.com/WICG/CSS-Parser-API/pull/12

@lukebjerring, do you think we should just merge this and let the IDL files be in an inconsistent state for a while as long as it doesn't break things, or line up a bunch of PRs and merge them at once?

lukebjerring:
Probably a good idea for this to update all of the `partial namespace` defs at the same time.

foolip:
The two spec PRs have landed. https://github.com/w3c/css-houdini-drafts/pull/784 has propagated to reffy in https://github.com/tidoust/reffy-reports/commit/ee634c9a84ed4f0ae281cdbdf118dbafec2ac8e8 but https://github.com/WICG/CSS-Parser-API/pull/12 will need @tabatkins to regenerate index.html first.

tabatkins:
Regenerated CSS-Parser-API

foolip:
Cool, in 24 hours or so https://github.com/tidoust/reffy-reports/blob/master/whatwg/idl/CSS-Parser-API.idl ought to be updated.

foolip (approved):

foolip:
Tested locally, it works!

PR #11822, opened 2018-07-06
Test fetch abort with keepalive

ricea:
Aborting a fetch with keepalive: true should normally work. However, if
the context has been detached then it should not work. Test these cases.

### DIFF
--- a/fetch/api/abort/keepalive.html
+++ b/fetch/api/abort/keepalive.html
@@ -1,0 +1,83 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/common/utils.js"></script>
+<script>
+// This controller must be on the window so it is visible to the iframe.
+window.sharedController = new AbortController();
+
+async function fetchJson(url) {
+  const response = await fetch(url);
+  assert_true(response.ok, 'response should be ok');
+  return response.json();
+}
+
+promise_test(async () => {
+  const stateKey = token();
+  const controller = new AbortController();
+  await fetch(`../resources/infinite-slow-response.py?stateKey=${stateKey}`,
+              {
+                signal: controller.signal,
+                keepalive: true
+              });
+  const before = await fetchJson(`../resources/stash-take.py?key=${stateKey}`);
+  assert_equals(before, 'open', 'connection should be open');
+
+  controller.abort();
+
+  // Spin until the abort completes.
+  while (true) {
+    const after = await fetchJson(`../resources/stash-take.py?key=${stateKey}`);
+    if (after) {
+      assert_equals(after, 'closed', 'connection should have closed');
+      break;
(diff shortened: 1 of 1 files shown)


ricea:
@yutakahirano PTAL.

fetch/api/abort/keepalive.html
```diff
@@ -0,0 +1,83 @@
...
+  });
+
+  // Detach the context of the fetch.
+  iframeEl.remove();
+
+  sharedController.abort();
+
+  // The abort should not do anything. The connection should stay open. Wait 1
+  // second to give time for the fetch to complete.
+  await new Promise(resolve => t.step_timeout(() => resolve(), 1000));
```
domfarolino: Substituting `() => resolve()` for `resolve` might be more succinct?

fetch/api/abort/keepalive.html
```diff
@@ -0,0 +1,83 @@
...
+
+<iframe srcdoc="
+                <!DOCTYPE html>
+                <meta charset=utf-8>
+                <script>
+                onmessage = async event => {
+                  const key = event.data;
+                  await fetch(
+                    `../resources/redirect.py?delay=500&amp;location=` +
+                    `../resources/stash-put.py%3fkey=${key}%26value=on`,
```
domfarolino: Is there a reason for %26 over `&`? If it's preference that's cool, just curious.

domfarolino (approved):

fetch/api/abort/keepalive.html
```diff
@@ -0,0 +1,83 @@
...
+  const before = await fetchJson(`../resources/stash-take.py?key=${stateKey}`);
+  assert_equals(before, 'open', 'connection should be open');
+
+  controller.abort();
+
+  // Spin until the abort completes.
+  while (true) {
+    const after = await fetchJson(`../resources/stash-take.py?key=${stateKey}`);
+    if (after) {
+      assert_equals(after, 'closed', 'connection should have closed');
```
yutakahirano: Is this deterministic? Is there any chance that the abort operation is slow and we see "open" here?

fetch/api/abort/keepalive.html
ricea: I didn't know at the time that redirect.py copies all its parameters to the target URL.

Even now that I do know that, I think it's still nicer to make the value explicitly part of the target URL

fetch/api/abort/keepalive.html
ricea: stash-take.py has the surprising side-effect of removing the key from the dictionary, so yes, it is deterministic*.

* It can still loop until it times out in theory, but the risk is tiny.

fetch/api/abort/keepalive.html
ricea: Nice.

ricea:
@yutakahirano PTAL

yutakahirano (approved):
Sorry for the delay.

PR #11823, opened 2018-07-06
sensors: Stop passing undefined as an invalid SensorOptions value

rakuco:
While the undefined -> double conversion in WebIDL should indeed throw a
TypeError, we cannot forget we have a WebIDL dictionary (SensorOptions) in
the middle:

    {frequency: undefined}

is essentially equivalent to

    {}

so we do not actually throw a TypeError in this case at all.


Honry (approved):
LGTM, thanks!

PR #11833, opened 2018-07-06
Add idl_test helper to idlharness.js

lukebjerring:
`pointerlock` example attached to the PR demonstrates the removed boilerplate (common across almost all idlharness tests).

### DIFF
--- a/pointerlock/interfaces.window.js
+++ b/pointerlock/interfaces.window.js
@@ -5,22 +5,14 @@
 
 // https://w3c.github.io/pointerlock/
 
-promise_test(async () => {
-  const html = await fetch('/interfaces/html.idl').then(r => r.text());
-  const dom = await fetch('/interfaces/dom.idl').then(r => r.text());
-  const uievents = await fetch('/interfaces/uievents.idl').then(r => r.text());
-  const idl = await fetch('/interfaces/pointerlock.idl').then(r => r.text());
-
-  const idl_array = new IdlArray();
-  idl_array.add_idls(idl);
-  idl_array.add_dependency_idls(uievents);
-  idl_array.add_dependency_idls(dom);
-  idl_array.add_dependency_idls(html);
-
-  idl_array.add_objects({
-    Document: ["window.document"],
-    Element: ["window.document.documentElement"],
-    MouseEvent: ["new MouseEvent('foo')"]
-  });
-  idl_array.test();
-}, 'pointerlock interfaces.');
+idl_test(
+  ['pointerlock'],
+  ['uievents', 'html', 'dom'],
+  idl_array => {
+    idl_array.add_objects({
+      Document: ["window.document"],
+      Element: ["window.document.documentElement"],
+      MouseEvent: ["new MouseEvent('foo')"]
+    });
+  },
+  'pointerlock interfaces.');
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -3013,4 +3013,41 @@ function IdlTypedef(obj)
 IdlTypedef.prototype = Object.create(IdlObject.prototype);
 
 }());
+
+/**
+ * idl_test is a promise_test wrapper that handles the fetching of the IDL,
+ * avoiding repetitive boilerplate.
+ *
+ * @param {String|String[]} srcs Spec name(s) for source idl files (fetched from
+ *      /interfaces/{name}.idl).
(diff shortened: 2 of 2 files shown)


resources/idlharness.js
```diff
@@ -3013,4 +3013,41 @@ function IdlTypedef(obj)
...
 
 }());
+
+/**
+ * idl_test is a promise_test wrapper that handles the fetching of the IDL,
+ * avoiding repetitive boilerplate.
+ *
+ * @param {String|String[]} srcs Spec name(s) for source idl files (fetched from
+ *      /interfaces/{name}.idl).
+ * @param {String|String[]} deps Spec name(s) for dependency idl files (fetched
```
kereliuk: Do these need to have a specific order for when they are added in the for loop below?
If so, we should document that order.

pointerlock/interfaces.window.js
```diff
@@ -5,22 +5,14 @@
...
-  idl_array.add_objects({
-    Document: ["window.document"],
-    Element: ["window.document.documentElement"],
-    MouseEvent: ["new MouseEvent('foo')"]
-  });
-  idl_array.test();
-}, 'pointerlock interfaces.');
+idl_test(
+  ['pointerlock'],
+  ['uievents', 'html', 'dom'],
```
kereliuk: The dependency idls are added in a different order here than in the previous revision. Does this make a difference?

kereliuk (commented):
Made some ordering comments, otherwise looks good!

resources/idlharness.js
lukebjerring: Yes, and done.

pointerlock/interfaces.window.js
lukebjerring: Actually, if you look at https://github.com/web-platform-tests/wpt/pull/11833/files/e3f9f7b704f6f50514f1db4dfaf16747cb5f7c3d#diff-0375a5322ffb1328cfc6a75c965c8ea2L16 `uievents` was first. I double-checked and all the tests are happily passing with this new order.

kereliuk: uievents is still first but dom and html are in reverse order, although if all tests pass in the new order should be good

kereliuk (approved):

resources/idlharness.js
```diff
@@ -3013,4 +3013,43 @@ function IdlTypedef(obj)
...
+ *      each source will only be included if they're already know to be a
+ *      dependency (i.e. have already been seen).
+ * @param {Function} setup_func Function for extra setup of the idl_array, such
+ *      as adding objects. Do not call idl_array.test() in the setup; it is
+ *      called by this function (idl_test).
+ */
+function idl_test(srcs, deps, setup_func, test_name) {
+    return promise_test(function (t) {
+        var idl_array = new IdlArray();
+        srcs = (srcs instanceof Array) ? srcs : [srcs] || [];
```
jgraham: `Array.isArray`?

jgraham: Also the bit after the `||` seems like a use case for default parameters? Unless we are keen to make this work with older ES engines, which I guess is not a bad idea.

resources/idlharness.js
```diff
@@ -3013,4 +3013,43 @@ function IdlTypedef(obj)
...
+        deps = (deps instanceof Array) ? deps : [deps] || [];
+        return Promise.all(
+            srcs.concat(deps).map(function(i) {
```
jgraham: `i` is not a super-clear name

resources/idlharness.js
```diff
@@ -3013,4 +3013,43 @@ function IdlTypedef(obj)
...
+                return fetch('/interfaces/' + i + '.idl').then(function(r) {
```
jgraham: Isn't the most common style to line up then conditions on `.`, like

```
foo
  .then(()=>{})
  .then(()=>{})
```

?

resources/idlharness.js
lukebjerring: yeah, recently a PR was flagged by @Ms2ger as it being a better outcome to support older engines, so this PR is written attempting to be backward-friendly.

resources/idlharness.js
lukebjerring: Followed up style issues in https://github.com/web-platform-tests/wpt/pull/11836

PR #11834, opened 2018-07-06
Update the media-source IDL file

lukebjerring:
Hello reviewer(s),

Please check that:


foolip (approved):
Tested manually, fewer failures now because `changeType` went away.

wolenetz:
LGTM. Thanks! This is much cleaner, solves also crbug 633669 (though many of those ancient failures may have been fixed in interim).

The SourceBuffer.changetype() method is in WICG incubation (see WICG/media-source spec repo's codec-switching branch: https://github.com/WICG/media-source/tree/codec-switching). At what point can/should it be added back to the IDL checks in wpt?

foolip:
@wolenetz, publishing https://github.com/WICG/media-source/tree/codec-switching as a separate spec would be the most straightforward way to get it tested, since we get the IDL files from https://github.com/tidoust/reffy-reports, which is scraped from spec.

If you want some proper test coverage, adding tentative tests for it, and you could do a manual idlharness.js test as well.

wolenetz:
@foolip Thanks - I'd already added some functional tests (not just IDL) for codec-switching. They're in media-source/mediasource-changetype*.html.  I suppose I could add some manual IDL check until the spec goes to a WG beyond incubation, though perhaps such IDL reverification could be added then.

foolip:
It looks like the media-source/mediasource-changetype*.html tests already cover about as much as an idlharness test would. idlharness.js does very shallow testing, helpful in revealing things which aren't implemented and which have no other tests, or small mistakes in IDL bindings, but for this case I think you wouldn't gain much by adding back that coverage ahead of it being included in the main spec.

PR #11844, opened 2018-07-07
SVG ellipse supports rx/ry properties

ewilligers:
Spec: https://svgwg.org/svg2-draft/shapes.html#EllipseElement

When only rx is not auto, rx is used for both dimensions.
When only ry is not auto, ry is used for both dimensions.

### DIFF
--- a/svg/shapes/ellipse-02-ref.svg
+++ b/svg/shapes/ellipse-02-ref.svg
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" width="100" height="100">
+  <ellipse cx="50" cy="50" rx="40" ry="40" fill="none" stroke="blue" stroke-width="5"/>
+</svg>
--- a/svg/shapes/ellipse-02.svg
+++ b/svg/shapes/ellipse-02.svg
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" width="100" height="100">
+  <metadata>
+    <link xmlns="http://www.w3.org/1999/xhtml" rel="help" href="https://www.w3.org/TR/SVG2/shapes.html#EllipseElement"/>
+    <link xmlns="http://www.w3.org/1999/xhtml" rel="match" href="ellipse-02-ref.svg"/>
+    <meta xmlns="http://www.w3.org/1999/xhtml" name="assert" content="ry auto means rx is used"/>
+  </metadata>
+  <style>
+    ellipse {
+      rx: 40px;
+      ry: auto;
+    }
+  </style>
+  <ellipse cx="50" cy="50" fill="none" stroke="blue" stroke-width="5"/>
+</svg>
--- a/svg/shapes/ellipse-03-ref.svg
+++ b/svg/shapes/ellipse-03-ref.svg
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" width="100" height="100">
+  <ellipse cx="50" cy="50" rx="30" ry="30" fill="none" stroke="blue" stroke-width="5"/>
+</svg>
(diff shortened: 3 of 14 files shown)


svg/shapes/ellipse-02-ref.svg
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" width="100" height="100">
+  <ellipse cx="50" cy="50" rx="40" ry="40" fill="none" stroke="red" stroke-width="5"/>
```
fsoder: stroke=orange

svg/shapes/ellipse-02.svg
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" width="100" height="100">
+  <metadata>
+    <link xmlns="http://www.w3.org/1999/xhtml" rel="help" href="http://www.w3.org/TR/SVG2/shapes.html#RectElement"/>
```
fsoder: Same comment as in -01.

svg/shapes/ellipse-03-ref.svg
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" width="100" height="100">
+  <ellipse cx="50" cy="50" rx="30" ry="30" fill="none" stroke="red" stroke-width="5"/>
```
fsoder: stroke=orange

svg/shapes/ellipse-03.svg
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" width="100" height="100">
+  <metadata>
+    <link xmlns="http://www.w3.org/1999/xhtml" rel="help" href="http://www.w3.org/TR/SVG2/shapes.html#RectElement"/>
```
fsoder: URL

svg/shapes/ellipse-04.svg
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" width="100" height="100">
+  <metadata>
+    <link xmlns="http://www.w3.org/1999/xhtml" rel="help" href="http://www.w3.org/TR/SVG2/shapes.html#RectElement"/>
```
fsoder: URL

svg/shapes/ellipse-04.svg
```diff
@@ -0,0 +1,15 @@
...
+    <link xmlns="http://www.w3.org/1999/xhtml" rel="match" href="ellipse-04-ref.svg"/>
+    <meta xmlns="http://www.w3.org/1999/xhtml" name="assert" content="rx and ry auto disables rendering."/>
+  </metadata>
+  <style>
+    ellipse {
+      rx: auto;
+      ry: auto;
+    }
+  </style>
+  <ellipse cx="50" cy="50" fill="none" stroke="orange" stroke-width="5"/>
```
fsoder: Maybe here it'd make sense to actually use 'red' as the stroke color (to clearly show if there's an error.)

svg/shapes/ellipse-04-ref.svg
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" width="100" height="100">
+  <ellipse cx="50" cy="50" rx="0" ry="0" fill="none" stroke="red" stroke-width="5"/>
```
fsoder: Seem like this ref could trivially be an "empty" SVG - otherwise there's a risk that '0' & 'auto' is mishandled, but the test still passes.

fsoder (changes requested):

AmeliaBR:
These look good, thanks Eric.

Other things to test:

- for `auto` in one direction and percentage value in the other, the computed value of the percentage is used for the dimension in which it was specified (you'll need a non-square SVG to test this).

- check that an `auto` value is used as the default for a missing attribute/unspecified property (not defaulting to 0).

ewilligers:
Tests added.

fsoder (approved):

PR #11847, opened 2018-07-08
Test animatedPoints during a SMIL animation

ewilligers:
In SVG 2, animatedPoints represents the current non-animated value.

Spec:
https://svgwg.org/svg2-draft/shapes.html#InterfaceSVGAnimatedPoints


svgeesus (approved):
Yes, this looks correct.
Important to test, as this is a change in SVG2 wrt SVG 1.1SE

PR #11850, opened 2018-07-09
Revise HTTP caching tests

mnot:
General cleanup and refactoring of the HTTP caching tests, with several improvements to reliability.


fetch/http-cache/cc-request.html
```diff
@@ -202,7 +202,7 @@
               ["Cache-Control", "only-if-cached"]
             ],
             expected_status: 504,
-            expected_response_text: ""
+            expected_response_text: null
```
jdm: Why do we not want to verify the contents of the body?

fetch/http-cache/status.html
```diff
@@ -51,7 +51,7 @@
     [
       [200, "OK"],
       [203, "Non-Authoritative Information"],
-      [204, "No Content", ""],
+      [204, "No Content", null],
```
jdm: Why do we not want to verify that there actually is no content?

fetch/http-cache/http-cache.js
```diff
@@ -210,7 +190,7 @@ function makeTest (rawRequests) {
       .then(function () {
         // Now, query the server state
         return serverState(uuid)
-      }).then(function (state) {
+      }).then(function (step_timeoutate) {
```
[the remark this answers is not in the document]
mnot: yes; fixed in e9120f5a

fetch/http-cache/http-cache.js
```diff
@@ -252,6 +210,57 @@ function fetchInit (config) {
   return init
 }
 
+function makeCheckResponse (idx, config) {
+  return function checkResopnse (response) {
```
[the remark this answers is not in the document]
mnot: Fixed in ea2214eb

fetch/http-cache/cc-request.html
mnot: Since the 504 is generated by the cache, not by the origin, it can contain anything; it's not under control of the test suite.

(I'm also starting to use these test cases for intermediary cache testing, and some of them do put things in there)

jdm (approved):

fetch/http-cache/status.html
mnot: AIUI in fetch a 204's body should not be an empty string.

PR #11851, opened 2018-07-09
Remove .gitkeep files in 2dcontext/ and html/

foolip:
These were added in commit 818bcab4ba58d2bc8dba94276c8b944c76570a19
and reflect the structure of the HTML spec as it was in 2012.

The structure largely remains the same, but some directories like
webappapis/user-prompts/dialogs-implemented-using-separate-documents/
are gone from the spec (https://github.com/whatwg/html/pull/374) and
others like introduction/ are non-normative and should never be
populated.

But mainly, a lot of empty directories is a nuisance when exploring
the directory in a file manager or on GitHub.

The number of directories in 2dcontext/ and html/ combined decreases
by almost 200, from 653 to 456.


foolip:
Reviewers, if any of you have found these directories useful, please say something. They've mostly annoyed me, and trying to vet which ones are still correct seems like a lot of work.

domenic (approved):

Issue #11521, opened 2018-06-15
META.yml: Replace ayg in suggested_reviewers


zcorpan:
@ayg has not accepted the invite, but appears in a number of META.yml files. Consider removing and adding someone else.

```
$ git grep "  - ayg"
dom/META.yml:  - ayg
editing/META.yml:  - ayg
html/dom/META.yml:  - ayg
html/editing/META.yml:  - ayg
html/infrastructure/META.yml:  - ayg
html/obsolete/META.yml:  - ayg
html/semantics/META.yml:  - ayg
resources/META.yml:  - ayg
selection/META.yml:  - ayg
```
files that include only ayg
```
$ git grep -l "  - ayg" | xargs wc -l | grep "2 "
       2 editing/META.yml
       2 html/dom/META.yml
       2 html/editing/META.yml
       2 html/infrastructure/META.yml
       2 html/obsolete/META.yml
       2 html/semantics/META.yml
       2 selection/META.yml
```
Since html/META.yml exists, the html ones could be removed. That leaves `editing` and `selection`.

zcorpan:
@rniwa can you take `selection`?

@johanneswilm can you take `editing`?

foolip:
I can take html/obsolete, just my kind of thing.

foolip:
html/semantics is very broad, is there really anyone who would be a good reviewer for all of it?

zcorpan:
If nobody steps up, we could split html/semantics in some manner between you, me, @annevk and @domenic. WDYT?

foolip:
Yeah, I'd be fine with with most of html/semantics/embedded-content/. Here are the numbers of files in the directories:
```
2 disabled-elements
2 the-root-element
6 common-idioms
9 edits
15 sections
22 links
26 selectors
27 interactive-elements
44 tabular-data
72 grouping-content
86 document-metadata
88 text-level-semantics
253 forms
495 scripting-1
779 embedded-content
```

zcorpan:
Here are the 5 most recent commit authors for each dir under semantics:
```
semantics zcorpan$ find . -type d -depth 1 | while read line ; do echo $line; git log -r -5 --format="%an %ae" $line ; done
./the-root-element
Robin Berjon robin@berjon.com
./text-level-semantics
Bhagirathi Satpathy bhagirathi.s@samsung.com
Bhagirathi Satpathy bhagirathi.s@samsung.com
Bhagirathi Satpathy bhagirathi.s@samsung.com
[...]

PR #11852, opened 2018-07-09
Replace/remove ayg in suggested_reviewers in META.yml

foolip:
Responsibilities distributed as discussed in
https://github.com/web-platform-tests/wpt/issues/11521.

In html/, because there aren't obvious candidates for some of the
subdirectories, instead expand html/META.yml to include all present
editors of the HTML standard.

In the past year, most of the action has been in html/semantics/, and
the busiest of the directories within do all have more granular
reviewers, so hopefully this is an improvement.

Fixes https://github.com/web-platform-tests/wpt/issues/11521.

### DIFF
--- a/dom/META.yml
+++ b/dom/META.yml
@@ -1,5 +1,4 @@
 suggested_reviewers:
-  - ayg
   - jdm
   - zqzhang
   - annevk
--- a/editing/META.yml
+++ b/editing/META.yml
@@ -1,2 +1,2 @@
 suggested_reviewers:
-  - ayg
+  - johanneswilm
--- a/html/META.yml
+++ b/html/META.yml
@@ -1,4 +1,9 @@
 suggested_reviewers:
+  - annevk
+  - domenic
+  - foolip
   - jdm
   - jgraham
+  - sideshowbarker
+  - zcorpan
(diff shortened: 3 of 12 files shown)


foolip:
@annevk @domenic @sideshowbarker @zcorpan, seeking your review in particular, since the html/META.yml change wasn't discussed in the issue, but seemed like a good idea as I was making the edits.

domenic:
I guess I'm OK with it, although it kind of defeats the point of assigning individuals to more granular subdirectories, if we're just going to get pinged for all of HTML anyway.

It's a shame there's no way to know when asked for a review whether it's because you're a "primary" reviewer or "fallback" reviewer.

foolip:
> it kind of defeats the point of assigning individuals to more granular subdirectories, if we're just going to get pinged for all of HTML anyway.

It's not the union of suggested reviewers which are pinged, but the ones in the closest META.yml file. Given that, is this setup better? (A drawback is that if someone is on vacation, the PRs with just a single reviewer will sit unreviewed.)

domenic:
Ah, awesome, in that case, I'm totally on board.

johanneswilm (approved):

PR #11853, opened 2018-07-09
Replace kristijanburnik in some META.yml files

foolip:
No longer active on GitHub.

### DIFF
--- a/mixed-content/META.yml
+++ b/mixed-content/META.yml
@@ -1,2 +1,2 @@
 suggested_reviewers:
-  - kristijanburnik
+  - andypaicu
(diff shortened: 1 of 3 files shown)


andypaicu:
I am somewhat familiar with mixed content tests but don't know too much about referrer-policy or websockets so I doubt I'd be a good reviewer.

Is there necessarily a need to replace @kristijanburnik, I see there are other people already in the review list?

foolip:
No, mixed-content/ is the only one that triggered this. I'll remove in the other two instead.

andypaicu (approved):

PR #11855, opened 2018-07-09
Update the navigation-timing IDL file

lukebjerring:
Hello reviewer(s),

Please check that:


igrigorik (commented):
Sanity check: we no longer have to define PerformanceEntry and PerformanceResourceTiming here?

navigation-timing/idlharness.window.js
```diff
@@ -0,0 +1,22 @@
+// META: script=/resources/WebIDLParser.js
+// META: script=/resources/idlharness.js
+
+'use strict';
+
+// http://www.w3.org/TR/navigation-timing/
+
+idl_test(
+  ['navigation-timing'],
+  ['resource-timing', 'performance-timeline', 'hr-time', 'dom'],
```
lukebjerring: @igrigorik - correct, `PerformanceEntry` and `PerformanceResourceTiming` are defined in `hr-time` + `performance-timeline` specs. Previously, they were duplicated, under the `<pre id='untested_idl' style='display:none'>` tag (untested).

This is now handled by `add_dependency_idls` which only calls `add_untested_idls` for any IDL which it has detected to be a dependency of the main IDL (in this case, `navigation-timing`). This line here is the `dependency specs` param of the `idl_test` helper, meaning that each spec in this second array is passed to `add_dependency_idls`.

igrigorik: @lukebjerring cool, thanks!

igrigorik (approved):
👍

PR #11865, opened 2018-07-09
Update the html-media-capture IDL file

lukebjerring:
Hello reviewer(s),

Please check that:


html-media-capture/idlharness.window.js
```diff
@@ -0,0 +1,20 @@
...
+// https://w3c.github.io/html-media-capture/
+
+idl_test(
+  ['html-media-capture'],
+  ['html', 'dom'],
+  idl_array => {
+    const input = document.createElement('input');
+    input.type = 'file';
+    input.accept = 'image/*';
+    input.capture = 'user';
```
foolip: This line will mask if the capture attribute is not there at all I think, or at least lead to failures that are misleading? A plain new input element with nothing changed should suffice here.

lukebjerring: Done.

foolip (approved):
Tested locally and the important tests (for the capture attribute) fail in both Chrome, Firefox and Safari. Still, will be useful down the line if for nothing else than to show lack of support.

PR #11868, opened 2018-07-09
Update the css-conditional IDL file.

lukebjerring:
Hello reviewer(s),

Please check that:

### DIFF
--- a/interfaces/css-conditional.idl
+++ b/interfaces/css-conditional.idl
@@ -1,6 +1,7 @@
 // GENERATED CONTENT - DO NOT EDIT
-// Content of this file was automatically extracted from the CSS Conditional Rules spec.
-// See https://drafts.csswg.org/css-conditional/
+// Content of this file was automatically extracted from the
+// "CSS Conditional Rules Module Level 3" spec.
+// See: https://drafts.csswg.org/css-conditional-3/
 
 partial interface CSSRule {
     const unsigned short SUPPORTS_RULE = 12;
@@ -27,7 +28,7 @@ interface CSSMediaRule : CSSConditionRule {
 interface CSSSupportsRule : CSSConditionRule {
 };
 
-partial interface CSS {
-  static boolean supports(CSSOMString property, CSSOMString value);
-  static boolean supports(CSSOMString conditionText);
+partial namespace CSS {
+  boolean supports(CSSOMString property, CSSOMString value);
+  boolean supports(CSSOMString conditionText);
 };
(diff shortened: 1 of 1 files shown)


gsnedders:
@lukebjerring why is this using the levelled spec?

lukebjerring:
@gsnedders could you clarify what you mean by the levelled spec, and what you would expect to be used instead?

gsnedders:
@lukebjerring we've gone from using https://drafts.csswg.org/css-conditional/ (without level in the URL) to https://drafts.csswg.org/css-conditional-3/ (with level 3 in the URL), and renamed the file per the old levelled format (to css3-conditional).

lukebjerring:
Filed https://github.com/tidoust/reffy-reports/issues/11

lukebjerring:
Renamed back to css-conditional. PTAL.

css/css-conditional/idlharness.html
```diff
@@ -14,18 +14,31 @@
...
+        idl_array.add_objects({
+          CSSRule: ['rule'],
+          CSSMediaRule: ['media'],
+          CSSSupportsRule: ['supports'],
+        });
+        idl_array.prevent_multiple_testing('CSSRule');
+
+        self.media = document.styleSheets[0].cssRules[0];
+        self.supports = document.styleSheets[0].cssRules[1];
+        self.rule = media.cssRules[0];
```
foolip: Does this actually work, is this object a CSSRule and not some more specific type?

interfaces/css-conditional.idl
```diff
@@ -27,7 +28,7 @@ interface CSSMediaRule : CSSConditionRule {
 interface CSSSupportsRule : CSSConditionRule {
 };
 
-partial interface CSS {
-  static boolean supports(CSSOMString property, CSSOMString value);
-  static boolean supports(CSSOMString conditionText);
+partial namespace CSS {
```
foolip: cssom.idl still has interface CSS, so this will break, right?

foolip (commented):
Fixed two little things and wonder about more, can you run and verify this still works?

interfaces/css-conditional.idl
lukebjerring: Not any more :)

lukebjerring:
Note that this has a single failing test because the IDL in WPT says that `CSSConditionRule.conditionText` is not readonly, but Chrome's currently IDL says it is readonly.

https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/css/css_condition_rule.idl?l=8&rcl=0f0fa2e588f06f76c425651f608fe0e50c80e3ef

foolip (approved):
Tested in Chrome and Firefox. A different test fails in Firefox, looks like `insertRule` has a 2nd optional argument which shouldn't be optional.

PR #11872, opened 2018-07-09
Update the WebXR IDL file

lukebjerring:
Hello reviewer(s),

Please check that:


interfaces/webxr.idl
```diff
@@ -53,15 +63,16 @@ callback XRFrameRequestCallback = void (DOMHighResTimeStamp time, XRFrame frame)
...
 };
 
 [SecureContext, Exposed=Window] interface XRCoordinateSystem : EventTarget {
   Float32Array? getTransformTo(XRCoordinateSystem other);
 };
 
 enum XRFrameOfReferenceType {
-  "head-model",
-  "eye-level",
+  "headModel",
```
klausw: Why is this reverting back to the camelCase variants? https://github.com/immersive-web/webxr/blob/master/explainer.md#appendix-b-proposed-idl uses `head-model` etc

lukebjerring: Looks like the scrape was out of date, thanks for reviewing thoroughly!

foolip (approved):

PR #11880, opened 2018-07-10
Don't assume opacity computed values are exactly the same as specified on keyframe input

graouts:
The CSS Color Module Level 3 specification defines the [opacity](https://drafts.csswg.org/css-color-3/#opacity) property as taking <number> values. Meanwhile, the CSS Values and Units Module Level 3 specification [discusses the precision of numeric data types](https://drafts.csswg.org/css-values-3/#numeric-types) like so:

> CSS theoretically supports infinite precision and infinite ranges for all value types; however in reality implementations have finite capacity. UAs should support reasonably useful ranges and precisions.

This test fails in WebKit because it requires the provided value to be round-tripped through the computed style even though they cannot all be represented as precisely as the test entails. We update the test to compare values with a small alpha.

### DIFF
--- a/web-animations/animation-model/keyframe-effects/effect-value-overlapping-keyframes.html
+++ b/web-animations/animation-model/keyframe-effects/effect-value-overlapping-keyframes.html
@@ -25,7 +25,7 @@ test(t => {
                              { offset: 1, opacity: 1 } ],
                            { duration: 1000,
                              easing: 'cubic-bezier(0.5, -0.5, 0.5, 1.5)' });
-  assert_opacity_value(getComputedStyle(div).opacity, '0.2',
+  assert_opacity_value(getComputedStyle(div).opacity, 0.2,
                 'When progress is zero the last keyframe with offset 0 should'
                 + ' be used');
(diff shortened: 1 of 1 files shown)


web-animations/animation-model/keyframe-effects/effect-value-overlapping-keyframes.html
```diff
@@ -21,7 +25,7 @@
                              { offset: 1, opacity: 1 } ],
                            { duration: 1000,
                              easing: 'cubic-bezier(0.5, -0.5, 0.5, 1.5)' });
-  assert_equals(getComputedStyle(div).opacity, '0.2',
+  assert_opacity_value(getComputedStyle(div).opacity, '0.2',
```
birtles: Don't we want to pass in a number for the expected value here? Or call `parseFloat` on the expected value?

graouts: Somehow this works, but it would likely be better to use numbers. I will update this in an additional commit.

birtles (approved):
Thanks!

PR #11885, opened 2018-07-10
Remove CSS Speech tests

foolip:
https://www.w3.org/TR/css3-speech/ is published as Working Group Note,
which ought to mean to further work planned.

Most recent mentions found on www-style:
https://lists.w3.org/Archives/Public/www-style/2017Feb/0004.html
https://lists.w3.org/Archives/Public/www-style/2017Feb/0069.html
https://lists.w3.org/Archives/Public/www-style/2017Sep/0056.html

All of the tests are of the "This should be read in [some voice]"
variety, manual tests. Thus they do not show up in wpt.fyi.


foolip:
This is motivated by issue 11293 (not linked to avoid pinging 70+ people), and that @danielweck is not reacting to GitHub notifications. The directory would be left without reviewers when I finish that issue, and since the spec seems to be unmaintained I thought removing the tests is the best option.

Folks listed as authors in the tests themselves: @johnfoliot @musingvirtual @eilders @arronei.

If anyone wants to be in META.yml to get reviews instead of removing the tests, please shout.

musingvirtual:
Hello! I am still active in the WAI with Silver Project (although haven't looked at this spec in some time). I would be happy to maintain the tests if that ever becomes useful, but it sounds like because the spec is unmaintained the best thing for now is to drop these tests out of Web Platform. Someone can ping me to improve upon them should the spec ever return to active development.

foolip:
@musingvirtual there's no need to remove the tests other than to avoid them becoming orphaned (without reviewer), so if you're happy to get reviews, that's great, thank you! I've sent you an invite to the web-platform-tests org, if you accept it at https://github.com/web-platform-tests and then update the META.yml file, then I'll will close this PR. Thanks again!

PR #11910, opened 2018-07-11
Add objects to ResizeObserver IDL test

lukebjerring:
Adds a `ResizeObserverEntry` to test, obtained using the ResizeTestHelper resource.


resize-observer/idlharness.window.js
```diff
@@ -0,0 +1,26 @@
...
+
+// https://wicg.github.io/ResizeObserver/
+idl_test(
+  ['ResizeObserver'],
+  ['dom', 'geometry'],
+  idl_array => {
+    try {
+      self.observer = new ResizeObserver(entries => {});
+      self.entry = new ResizeObserverEntry(document);
+      self.observation = new ResizeObservation(document);
```
atotic: ResizeObservation is not supposed to be visible in the browser according to the spec.

lukebjerring: Oh - I was under the impression the default exposure set was `Window`, and I'm pretty sure that assumption exists in [idlharness](https://github.com/web-platform-tests/wpt/blob/master/resources/idlharness.js#L681):

> let result = default_set || ["Window"];

That doesn't seem to be the case in https://heycam.github.io/webidl/#Exposed though.
Note that a bunch of tests (11, to be precise) fail for precisely that reason:

> assert_own_property: self does not have own property "ResizeObservation" expected property "ResizeObservation" missing

lukebjerring: Found where I got that impression from:
From https://www.w3.org/TR/WebIDL-1/#Exposed

> If the [Exposed] extended attribute does not appear on a construct, then its exposure set is defined implicitly, depending on the type of construct:
> 
> interface
> dictionary
> The exposure set of the interface or dictionary only contains the primary global interface.

lukebjerring: @heycam - could you clarify what expected behaviour here is? Has defaulting to `Window` changed from WebIDL v1?

interfaces/ResizeObserver.idl
```diff
@@ -11,7 +11,7 @@ interface ResizeObserver {
     void disconnect();
 };
 
-callback ResizeObserverCallback = void (sequence<ResizeObserverEntry> entries, ResizeObserver observer);
+callback ResizeObserverCallback = void (sequence<ResizeObserverEntry> entries, ResizeObserver observer)
```
foolip: Where did the semicolon go?

resize-observer/idlharness.window.js
```diff
@@ -0,0 +1,40 @@
...
+    "ResizeObserverEntry creator",
+    [
+      {
+        setup: observer => {
+          self.observer = observer;
+          observer.observe(div);
+          div.style.width = "5px";
+        },
+        notify: entries => {
+          self.entry = entries[0];
```
foolip: Turn this stuff into async setup now?

foolip (changes requested):

interfaces/ResizeObserver.idl
lukebjerring: It's been updated in the IDL (https://github.com/WICG/ResizeObserver/pull/56) but not pushed, I think.

resize-observer/idlharness.window.js
lukebjerring: Done.

foolip (approved):

PR #11911, opened 2018-07-11
Update the performance-timeline IDL file + test

lukebjerring:
Hello reviewer(s),

Please check that:

### DIFF
--- a/performance-timeline/idlharness.any.js
+++ b/performance-timeline/idlharness.any.js
@@ -6,21 +6,22 @@
 
 'use strict';
 
-promise_test(async () => {
-  const idl_array = new IdlArray();
-  const idl = await fetch("/interfaces/performance-timeline.idl").then(r => r.text());
-  const dom = await fetch("/interfaces/dom.idl").then(r => r.text());
-  const hrtime = await fetch("/interfaces/hr-time.idl").then(r => r.text());
+idl_test(
+  ['performance-timeline'],
+  ['hr-time', 'dom'],
+  idl_array => {
+    try {
+      const callback = (e, o) => {};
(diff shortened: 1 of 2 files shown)


performance-timeline/idlharness.any.js
```diff
@@ -6,21 +6,22 @@
...
-  idl_array.add_objects({
-    Performance: ["performance"],
-    PerformanceMark: [self.performance.getEntriesByName("mark")[0]],
-  });
-  idl_array.test();
-}, "Test IDL implementation of performance-timeline API");
+    idl_array.add_objects({
+      Performance: ['performance'],
+      // PerformanceEntry implicitly tested in user-timing.
+      ResizeObserverEntry: ['observerEntry'],
```
lukebjerring: @npm1 - I'd like to cover `PerformanceObserverEntryList` but I'm not sure where to get an instance.
Am I missing something, or is it the same method set as `PerformanceEntryList`, so probably implemented by `Performance`, but not stated as such in the IDL?

npm1: PerformanceObserverEntryList is the first parameter that a callback to the PerformanceObserver receives. So you can get such an object by creating a PerformanceObserver and calling observe().

It is not quite the same as PerformanceEntryList. PerformanceObserverEntryList is a wrapper interface that allows calling getEntries() and a couple other methods, whose return values are a sequence of entries, i.e. a PerformanceEntryList.

igrigorik: @lukebjerring is this PR blocked on the above, or?

igrigorik (approved):
👍

PR #11913, opened 2018-07-11
Add tests for element setCustomValidity

### LINKED ISSUE #993: Test for <output>.setCustomValidity()
Proposed test:

http://software.hixie.ch/utilities/js/live-dom-viewer/saved/3026

Checks that <output> can't match :invalid, get an 'invalid' event, or block submission.

xiuqijix:
Add following tests (Fix #993):

- Add tests for output/object/select/textarea setCustomValidity
- Optimize tests for input/button/fieldset setCustomValidity by using element.validity.customError to check if the element has a custom error

@foolip, PTAL.

### DIFF
--- a/html/semantics/forms/the-button-element/button-setcustomvalidity.html
+++ b/html/semantics/forms/the-button-element/button-setcustomvalidity.html
@@ -1,43 +1,17 @@
 <!DOCTYPE HTML>
-<html>
- <head>
-  <title>Forms</title>
-  <script src="/resources/testharness.js"></script>
-  <script src="/resources/testharnessreport.js"></script>
- </head>
- <body>
-  <p>
-    <h3>button_setCustomValidity</h3>
-  </p>
+<title>button setCustomValidity</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
 
-  <hr>
+<button id='test'></button>
 
-  <div id="log"></div>
+<script>
 
-  <form method="post"
-      enctype="application/x-www-form-urlencoded"
-      action=""
-      id="input_form">
-    <p><button id='button_id'>button</button></p>
-  </form>
-  <script>
+test(() => {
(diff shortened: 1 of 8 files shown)


html/semantics/forms/the-button-element/button-setcustomvalidity.html
```diff
@@ -1,43 +1,17 @@
...
- <body>
-  <p>
-    <h3>button_setCustomValidity</h3>
-  </p>
+<title>button setCustomValidity</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
 
-  <hr>
+<button id='test'></button>
```
tkent-google: nit: Please avoid the ID 'test'.  Both of this element and testharness's test() can be 'test' property of 'window'.  It's confusing.

html/semantics/forms/the-object-element/object-setcustomvalidity.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE HTML>
```
tkent-google: We should not create the-object-element/ directory under semantics/forms/.  It should be in semantics/embedded-content/.  The directory structure in html/ should be matched to the structure of the HTML standard.

tkent-google (changes requested):

html/semantics/forms/the-button-element/button-setcustomvalidity.html
kaixinjxq: @tkent-google, many thanks for your comments, I've addressed it, PTAL.

html/semantics/forms/the-object-element/object-setcustomvalidity.html
kaixinjxq: Updated

tkent-google (approved):

PR #11922, opened 2018-07-11
Update the pointerevents IDL file + tests

lukebjerring:
Hello reviewer(s),

Please check that:


foolip:
@lukebjerring, changes from another PR somehow got entangled here, I merged to resolve conflicts and made those changes go away. This PR should definitely be squashed when landed though.

foolip:
http://w3c-test.org/pointerevents/idlharness.html is entirely broken now so this is an improvement for sure, but I see a problem. Subtest 'GlobalEventHandlers interface: window must inherit property "ongotpointercapture" with the proper type' and others like it are failing because `ongotpointercapture` is an own property of the global object, and there's apparently no special casing to avoid that check. @lukebjerring, this probably affects other tests you've recently updated. I'll hold off merging this until you can react.

foolip (approved):
LGTM % GlobalEventHandlers problem

PR #11926, opened 2018-07-11
Rename 'white-space' tests about the 'break-spaces' value instead of 'pre-wrap'

javifernandez:
The tests modified by this PR are verifying the 'break-space' value, so they shouldn't have pre-wrap in their name. Hence, I renamed them so that they have 'break'-spaces' instead.


kojiishi (approved):
Nice, thank you for doing this!

frivoal (approved):

PR #11929, opened 2018-07-11
[html] Add tests for document.open() after window.stop()

TimothyGu:
Passes on Chrome and Safari, but fails on Firefox and Edge.


domenic (approved):
Test LGTM, but it would probably be worth documenting why it's expected to pass. Is this a specific line in the spec that we're testing? Or is it more that the spec says document.open() should work just fine, and browsers have bugs because they reuse the parser, or...?

TimothyGu:
@domenic Done.

PR #11930, opened 2018-07-11
Reformat Encoding tests

jugglinmike:
Refactor tests so that they may be consumed by non-browser JavaScript runtimes which implement the WHATWG Encoding standard such as Node.js [1]. Using WPT's `.any.js` convention also extends test coverage in browsers by allowing them to be executed within a Web Worker.

This change is in service of gh-11277.

[1] https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_whatwg_supported_encodings

/cc @leobalter @mariestaver @nomadtechie


jdm (approved):

inexorabletash (approved):
Looks good, thanks!

PR #11931, opened 2018-07-11
Non protocol specific part of pointer actions in testdriver

kereliuk:
This PR adds a `pointer_action_sequence` method to testdriver aimed to automate pointerevent tests and more. It contains the basic non-protocol specific code and allows for vendors to add their own routines in their own tests. It does **not** contain protocol specific implementations. This will be in a follow up CL.

Please check the docs and function parameters of the API surface. I think a lot of the checking and logic should be left to the protocol specific executors.

cc/ @NavidZ

### DIFF
--- a/docs/_writing-tests/testdriver.md
+++ b/docs/_writing-tests/testdriver.md
@@ -50,15 +50,9 @@ For example, to send the tab key you would send "\uE004".
 
 ### `test_driver.pointer_action_sequence(actions)`
  - `actions` <[Array]<[Object]>> an array of Action objects`
-  - `action` <[Object]> A single action
-   - `id` The ID of the pointer to use. If one with this id does not exist a new one will be used
-   - `pointerType` The type of pointer to use, an enum of `mouse`, `touch` and `pen`.
-   - `subtype` The type of action to perform, an enum of `pointerUp`, `pointerMove`, `pointerDown` and `pause`
-   - `coordinates` An Object with exactly two integer attrbutes `x` and `y`.
-   - `duration` The duration in ms before the next action in the sequence should start
+  - `action` <[Object]> A single action. See [spec](https://www.w3.org/TR/webdriver/#actions) for format
 
-
-This function causes a sequence of pointer actions to be sent to the browser. It returns a `Promise` that
+This function causes a sequence of actions to be sent to the browser. It is based of the [WebDriver API](https://www.w3.org/TR/webdriver/#actions) The action can be a keyboard event, a pointer event or a pause. It returns a `Promise` that
 resolves after the actions have been sent or rejects if an error was thrown.
 
 ### `test_driver.send_keys(element, keys)`
--- a/docs/_writing-tests/testdriver.md
+++ b/docs/_writing-tests/testdriver.md
@@ -48,4 +48,34 @@ between the function being called and the promise settling.
 To send special keys, one must send the respective key's codepoint. Since this uses the WebDriver protocol, you can find a [list for code points to special keys in the spec](https://w3c.github.io/webdriver/webdriver-spec.html#keyboard-actions).
 For example, to send the tab key you would send "\uE004".
 
+### `test_driver.pointer_action_sequence(actions)`
+ - `actions` <[Array]<[Object]>> an array of Action objects`
+  - `action` <[Object]> A single action
+   - `id` The ID of the pointer to use. If one with this id does not exist a new one will be used
+   - `pointerType` The type of pointer to use, an enum of `mouse`, `touch` and `pen`.
+   - `subtype` The type of action to perform, an enum of `pointerUp`, `pointerMove`, `pointerDown` and `pause`
+   - `target` Either a DOM Element object or an Object with exactly two integer attrbutes `x` and `y`.
+   - `duration` The duration in ms before the next action in the sequence should start
+
+
+This function causes a sequence of pointer actions to be sent to the browser. It returns a `Promise` that
+resolves after the actions have been sent or rejects if an error was thrown.
+
+### `test_driver.send_keys(element, keys)`
+#### `element: a DOM Element object`
+#### `keys: string to send to the element`
+
+This function causes the string `keys` to be send to the target
+element (an `Element` object), potentially scrolling the document to
+make it possible to send keys. It returns a `Promise` that resolves
+after the keys have been send or rejects if the keys cannot be sent
+to the element.
+
+Note that if the element that's keys need to be send to does not have
+a unique ID, the document must not have any DOM mutations made
+between the function being called and the promise settling.
+
+To send special keys, one must send the respective key's codepoint. Since this uses the WebDriver protocol, you can find a [list for code points to special keys in the spec](https://w3c.github.io/webdriver/webdriver-spec.html#keyboard-actions).
+For example, to send the tab key you would send "\uE004".
+
 [testharness]: {{ site.baseurl }}{% link _writing-tests/testharness.md %}
--- a/docs/_writing-tests/testdriver.md
+++ b/docs/_writing-tests/testdriver.md
@@ -54,7 +54,7 @@ For example, to send the tab key you would send "\uE004".
    - `id` The ID of the pointer to use. If one with this id does not exist a new one will be used
    - `pointerType` The type of pointer to use, an enum of `mouse`, `touch` and `pen`.
    - `subtype` The type of action to perform, an enum of `pointerUp`, `pointerMove`, `pointerDown` and `pause`
-   - `target` Either a DOM Element object or an Object with exactly two integer attrbutes `x` and `y`.
+   - `coordinates` An Object with exactly two integer attrbutes `x` and `y`.
    - `duration` The duration in ms before the next action in the sequence should start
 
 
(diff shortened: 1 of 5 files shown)


docs/_writing-tests/testdriver.md
```diff
@@ -48,4 +48,34 @@ between the function being called and the promise settling.
 To send special keys, one must send the respective key's codepoint. Since this uses the WebDriver protocol, you can find a [list for code points to special keys in the spec](https://w3c.github.io/webdriver/webdriver-spec.html#keyboard-actions).
 For example, to send the tab key you would send "\uE004".
 
+### `test_driver.pointer_action_sequence(actions)`
+ - `actions` <[Array]<[Object]>> an array of Action objects`
+  - `action` <[Object]> A single action
+   - `id` The ID of the pointer to use. If one with this id does not exist a new one will be used
+   - `pointerType` The type of pointer to use, an enum of `mouse`, `touch` and `pen`.
+   - `subtype` The type of action to perform, an enum of `pointerUp`, `pointerMove`, `pointerDown` and `pause`
+   - `target` Either a DOM Element object or an Object with exactly two integer attrbutes `x` and `y`.
```
NavidZ: I wonder if we should allow the float coordinates as the events may have float coordinates in high dpi environments.

resources/testdriver.js
```diff
@@ -170,6 +170,22 @@
...
+         * to perform. It is modeled after the behaviour of {@link
+         * https://w3c.github.io/webdriver/#actions|WebDriver Actions Command}
+         *
+         * @param {Array} actions - an array of action objects to perform as
+         *                          described in docs
+         * @returns {Promise} fufiled after the actions are performed, or rejected in
+         *                    the cases the WebDriver command errors
+         */
+        pointer_action_sequence(actions) {
+            return window.test_driver_internal.pointer_action_sequence(actions);
```
NavidZ: Do we want to do similar logic to click that we did before like calculating the x,y from the given element with the same logic or scrollIntoView here?

docs/_writing-tests/testdriver.md
```diff
@@ -48,4 +48,34 @@ between the function being called and the promise settling.
...
+
+
+This function causes a sequence of pointer actions to be sent to the browser. It returns a `Promise` that
+resolves after the actions have been sent or rejects if an error was thrown.
+
+### `test_driver.send_keys(element, keys)`
+#### `element: a DOM Element object`
+#### `keys: string to send to the element`
+
+This function causes the string `keys` to be send to the target
```
NavidZ: Will you call .focus() programmatically in js to get the target element focused? If not how do you want the browser to send the event to that element? If not maybe remove the element altogether and then browser decides the given key will to which element (which will be the focused element).

docs/_writing-tests/testdriver.md
kereliuk: The webdriver spec has an error thrown if these aren't integers, so for now I think we should keep this with that spec; however, this is a good point and maybe the spec should change.

resources/testdriver.js
kereliuk: Yep, I think if the input is an element we need to do that conversion here. Good catch!

jgraham:
Why pointer actions rather than actions in general? It should be the same API basically.

resources/testdriver.js
kereliuk: Given a bit more thought, I think I want to discuss this a bit more.
What do you think about the providing coordinates vs. providing an element for the API surface.

One approach would be to keep the surface as it is now and infer whether an Element or an object with x, y attributes is being passed in and react accordingly.

Another is to have it always be x, y coordinates and maybe providing a util function (that could live in testdriver?) to calculate these given an element for ease of use.

WYT?

docs/_writing-tests/testdriver.md
kereliuk: I'm not sure I have context on what focus does, and I am also confused with the line you are referring to.

kereliuk:
@jgraham I think it is basically the same API, but might be different at the protocol/executor level, (for example, I'm not sure how marionette would handle this with its API). Also I don't see too much use for keyboard actions in WPT since we have send_keys, but I could be wrong with this.

resources/testdriver.js
NavidZ: I like the idea of having x, y only. I just wanted to follow the click API and also support the element thing but if we can keep the API surface simple and have util functions as you suggested IMO it is clearer.

docs/_writing-tests/testdriver.md
NavidZ: I'm referring to this API as whole and how it is implemented. Although in this PR you only added this documentation but the SendKeysAction doesn't seem to call focus on the given element. So I was wondering how it would possibly route a key event to a particular given element as explained in this paragraph.

resources/testdriver.js
kereliuk: That sounds good to me, I will push a change :)

docs/_writing-tests/testdriver.md
kereliuk: I think this is a strange git diff error since I didn't change this part of the doc but only seem to have moved it. I will try and update my PR so it doesn't think I moved it.
Can you open up an issue for that and tag me in it?

kereliuk:
I think it makes more sense to have them as separately named API calls from a user standpoint, but I'm open to having them bundled together as they are in the WebDriver spec. @NavidZ WYT?

jgraham:
I think it makes sense to expose all the flexibility that we worked hard to get into the actions spec here, rather than subsetting it.

My imagined design here is a builder API that allows bundling together click, key, pause, and eventually, scroll actions, so that one can generate sets of events like dragging an element whilst holding down a key, pausing over a target element, and then continuing to drag it away. I don't think separate API calls will allow for the necessary composition of inputs to model rich user interactions (this is exactly why selenium moved from separate API calls to the current model). But perhaps I'm not understanding what you have in mind?

NavidZ:
I don't have any strong preferences either way. I'm fine with both approaches. Just a few points that come to my mind.

Having them in one API might be confusing in the sense that the format of the input is slightly different. In both of the key and pointer actions we are going to have _action_ but for example _pointerId_ only can appear in the pointer actions. Some people might not like that. Not very convincing point though.

Regarding what @jgraham said I believe even with different APIs that use case is possible. The builder just queues a bunch of promises together. Basically it calls send_keys on _then_ function of pointer_action_sequence and then maybe a few other functions and at the end, the returned promise will get resolved after all that input is injected.

NavidZ:
Any updates on this PR?

kereliuk:
Sorry on the delay on this, was OOO.

We can change to follow the WebDriver spec and allow the protocol to parse the input to make sense of the key/pointer/none difference. Will push an update.

tools/wptrunner/wptrunner/testdriver-extra.js
```diff
@@ -70,4 +70,13 @@
         window.opener.postMessage({"type": "action", "action": "send_keys", "selector": selector, "keys": keys}, "*");
         return pending_promise;
     };
+
+    window.test_driver_internal.pointer_action_sequence = function(actions) {
```
NavidZ: This needs to be action_sequence as well. Right?

resources/testdriver.js
```diff
@@ -170,6 +170,22 @@
...
+         * to perform. It is modeled after the behaviour of {@link
+         * https://w3c.github.io/webdriver/#actions|WebDriver Actions Command}
+         *
+         * @param {Array} actions - an array of action objects to perform as
+         *                          described in docs
+         * @returns {Promise} fufiled after the actions are performed, or rejected in
+         *                    the cases the WebDriver command errors
+         */
+        action_sequence(actions) {
+            return window.test_driver_internal.action_sequence(actions);
```
NavidZ: I believe we should do some target->coordinate translation here similar to click. However, we only need to do it for pointer actions as they are the only actions that may have target. @jgraham @gsnedders what do you think?

NavidZ (commented):
As we are taking ownership of Chromedriver and the related changes I'm going to iterate over this change. @jgraham @gsnedders let me know what you think about the current state.

NavidZ:
FYI @smaug----  

We are going to use this API to automate instructions in pointerevent manual tests for the start.

docs/_writing-tests/testdriver.md
```diff
@@ -48,4 +48,28 @@ between the function being called and the promise settling.
 To send special keys, one must send the respective key's codepoint. Since this uses the WebDriver protocol, you can find a [list for code points to special keys in the spec](https://w3c.github.io/webdriver/webdriver-spec.html#keyboard-actions).
 For example, to send the tab key you would send "\uE004".
 
+### `test_driver.pointer_action_sequence(actions)`
```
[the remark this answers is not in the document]
gsnedders: I think that's fine.

resources/testdriver.js
NavidZ: I retract this suggestion. As scrolling the target into view is not quite possible as there might be different targets within one sequence and it is not possible to scrollintoview on different targets within one function call. So I changed the input of this function to only expect coordinates.

NavidZ (changes requested):
I'm not a maintainer in this repo to be able to push the changes on top of this pull request. I create another PL for this.
https://github.com/web-platform-tests/wpt/pull/12726

foolip:
Closing as per https://github.com/web-platform-tests/wpt/pull/11931#pullrequestreview-150249219

PR #9796, opened 2018-03-02
Update payment-handler IDL file

lukebjerring:
Hello reviewer(s),

This PR is intended to consolidate the spec’s IDL definition with the WPT test suite’s copy, and any idlharness tests.

The up-to-date copy of the IDL file was automatically extracted from the reffy-reports repo (https://github.com/tidoust/reffy-reports/tree/master/whatwg/idl) which scrapes known specs automatically + regulary.

This PR is part of a migration project which will eventually be automatically updating and creating PRs for changes in spec IDL.

Please check that:
The Spec (and its source) is correct and up-to-date
All tests which cover the IDL in the spec have been migrated to fetch + use the idl in the `interfaces/` directory (instead of inline copies in the test files).

PR #11932, opened 2018-07-11
Update payment-handler IDL and tests

lukebjerring:
Hello reviewer(s),

Please check that:


lukebjerring:
@marcoscaceres  - I've taken a stab at adding some instance objects to the test, but I'm not too familiar with `ServiceWorker` and the registration process.

Is there a way of getting a `ServiceWorkerRegistration` object from a SharedWorker and a DedicatedWorker? 
Also, `PaymentManager` doesn't seem to be available on `self` for _any_ of the workers; is that expected current-state? (that's in Chrome Version 68.0.3440.42 (Official Build) beta (64-bit))

lukebjerring:
I think I accidentally 'added' an extra test, instead of renaming the original test, in #9796.

interfaces/payment-handler.idl
```diff
@@ -4,83 +4,87 @@
 // See: https://w3c.github.io/payment-handler/
 
 partial interface ServiceWorkerRegistration {
-    [SameObject]
-    readonly attribute PaymentManager paymentManager;
+  [SameObject] readonly attribute PaymentManager paymentManager;
```
foolip: Changes here confirmed to be all whitespace with `git diff -w --word-diff-regex=.`

foolip (approved):
IDL changes LGTM, updated the test enough that it requires review from you, @lukebjerring

PR #11935, opened 2018-07-11
[webaudio] Generalize convolver test tolerances for numerical error

karlt:
Gecko uses FFTs for convolution, even with short response buffers, and so
precision would be expected to be similar to that with longer buffers in Blink
and Webkit.

In Gecko, the actual calculations performed depend on what instructions are
available on the processor, and different code is used on some platforms.  The
precision in the answers is similar with different processors and code, but
the slightly larger errors may occur in different tests.

This patch picks a single expected precision for similar tests and uses that
for all such tests.


karlt:
@rtoy, thank you for writing these tests and adding tolerances in
https://chromium.googlesource.com/chromium/src/+/c47304410d65ea7f031681fea8c5c3d8b9ec7683%5E%21/

I'm working on making Gecko conform to more of these tests.  The tests which
now almost pass don't have tolerances and even some of the tests that already
pass on some platforms are reporting small errors on others.

Trying to keep tolerances as tight as absolutely possible for each individual
subtest would mean ongoing small changes as new optimizations are applied
to the FFT algorithm, and so I'd prefer to aim for what we consider a
reasonable tolerance and apply that to all similar tests.

The tolerances selected here are the minimum that allow Gecko to pass each
test file with two different algorithms and two different processors.  They seem
reasonable to me for floating point arithmetic with 23-bit fractions, and
tolerances are very similar with two different algorithms.

Would you be happy with this change?

(Gecko uses 16-bit arithmetic on Android, but I haven't checked results there
because our infrastructure doesn't actually run wpt on Android, and the
spec-compliance of 16-bit arithmetic may be questionable.)

karlt:
@padenot: I haven't heard back from @rtoy.  Would you be able to look at this change, please?

webaudio/the-audio-api/the-convolvernode-interface/convolver-response-1-chan.html
```diff
@@ -13,7 +13,9 @@
     <script id="layout-test-code">
       // Test various convolver configurations when the convolver response has
       // one channel (mono).
-      //
+
+      const absoluteThreshold = 4.7684e-7; // 2^-21
```
padenot: Can we write it like you do below:

> Math.pow(2, -21);

so it's a bit more clear ?

webaudio/the-audio-api/the-convolvernode-interface/convolver-response-4-chan.html
```diff
@@ -14,6 +14,8 @@
       // Test various convolver configurations when the convolver response has
       // a four channels.
 
+      const absoluteThreshold = 3 * Math.pow(2, -22);
```
padenot: Is this empirical? Can we say so, or provide a rationale for picking this value and multiplier?

padenot (approved):
This looks good to me. Maybe a couple things can be clarified.

PR #11941, opened 2018-07-12
Fix metadata on css-writing-modes/ch-unit and logical-props tests.

fantasai:


frivoal (approved):

frivoal:
I think we could move these files to out of the writing modes test suite directory, and into css-logical and css-values, since that's where they do indeed belong. Approved, so you can merge as-is, but feel free to move the files if you think that's better.

gsnedders:
This should either be merged or worked on further; it lying around in a PR isn't great.

PR #11949, opened 2018-07-12
Update the payment-request IDL file

lukebjerring:
Hello reviewer(s),

Please check that:

### DIFF
--- a/payment-request/idlharness.https.window.js
+++ b/payment-request/idlharness.https.window.js
@@ -23,8 +23,8 @@ idl_test(
 
     idlArray.add_objects({
       PaymentRequest: ['paymentRequest'],
-      PaymentMethodChangeEvent: ['new PaymentMethodChangeEvent("type")'],
-      PaymentRequestUpdateEvent: ['new PaymentRequestUpdateEvent("type")'],
+      PaymentMethodChangeEvent: ['new PaymentMethodChangeEvent("paymentmethodchange")'],
+      PaymentRequestUpdateEvent: ['new PaymentRequestUpdateEvent("paymentrequestupdate")'],
     });
   },
   'Setup for Payment Request API IDL tests.'
--- a/payment-request/idlharness.https.window.js
+++ b/payment-request/idlharness.https.window.js
@@ -10,7 +10,10 @@ idl_test(
   ['dom', 'html'],
   idlArray => {
     try {
-      const methods = [{supportedMethods: 'basic-card'}];
+      const methods = [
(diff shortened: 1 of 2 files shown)


marcoscaceres (approved):
Thanks for updating. This is great. More updates hopefully landing this week - though I think they are just dictionaries.

payment-request/idlharness.https.window.js
```diff
@@ -10,16 +10,21 @@ idl_test(
...
+      const details = {total: {label: 'label', amount: amount} };
       window.paymentRequest = new PaymentRequest(methods, details);
     } catch (e) {
-      // Will be surfaced in idlharness.js's test_object below.
+      // Surfaced below when paymentRequest is undefined.
     }
 
     idlArray.add_objects({
       PaymentRequest: ['paymentRequest'],
+      PaymentMethodChangeEvent: ['new PaymentMethodChangeEvent("type")'],
```
marcoscaceres: I guess the "type" doesn't really matter, right?

marcoscaceres: (if it does, "paymentmethodchange" would be the right type here)

lukebjerring: I don't think it would cause an issues, but doesn't hurt to use real values.

lukebjerring:
np - I actually took a stab at adding a `PaymentResponse` object too, until I noticed there's no way to automate the user's input of the payment (all the other tests are manual).. let me know if you make progress on that front. I think @kereliuk will have some authority on what's possible in terms of webdriver stuffs.

lukebjerring:
@marcoscaceres - my general (limited) understanding is that, prior to a standardized webdriver API/implementation, we'd want some `testdriver.js` methods which trigger the user-controlled flow of the payment request (cancel, approve, add a card... , idk what else ). That `testdriver.js` method's default behaviour is 'reject as unimplemented'.

Then, each browser will write its `testdriver-vendor.js` specific implementation of said API, which isn't maintained it WPT, and may only work for that browser when it's in a test build. We benefit from sharing the tests, even if they all fail when being run in WPT.

https://github.com/web-platform-tests/wpt/blob/master/docs/_writing-tests/testdriver-tutorial.md

So, knowing very little about payment-requests - is there any (payment-request specific) automation behaviour available in any of the browsers?

marcoscaceres:
> So, knowing very little about payment-requests - is there any (payment-request specific) automation behaviour available in any of the browsers?

No, there is none. It's too hard to automate. Best we can do is these clicks.

PR #11957, opened 2018-07-12
Update the FileAPI IDL tests

lukebjerring:
Context: having some timeout problems.
https://bugs.chromium.org/p/chromium/issues/detail?id=856601

`idl_test` helper executes the fetches of the IDL files in parallel, so we should at least not be hanging around waiting for each request. Not convinced this will be enough, but it's cleaner none-the-less.

### DIFF
--- a/FileAPI/idlharness.html
+++ b/FileAPI/idlharness.html
@@ -22,28 +22,22 @@
     <script>
       'use strict';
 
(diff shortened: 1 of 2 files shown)


Hexcles (approved):
Simple refactoring without substantial changes. LGTM.

PR #11958, opened 2018-07-12
Update the IndexedDB tests to cover all workers

lukebjerring:
Uses the new `idl_test` helper and adds 

    // META: global=window,worker

which defaults to all 3 workers, instead of just the default of only the dedicatedworker ([citation](https://web-platform-tests.org/writing-tests/testharness.html#multi-global-tests)).


inexorabletash (approved):
Wow, nice and simple now.

PR #11959, opened 2018-07-12
Update the cors-rfc1918 IDL test

lukebjerring:
Use the `idl_test` helper, which parallelizes the fetches, and 

    // META: global=window,worker

which tests all workers (not just dedicatedworker, the default behaviour).


Hexcles (approved):
LGTM

domfarolino (commented):
Thanks!

PR #11960, opened 2018-07-12
Clean up the service-workers IDL test

lukebjerring:
Deletes some of the untested IDL copies in favor of `add_dependency_idls`.


foolip (approved):
Tried to run locally but requires HTTPS and how to get that working locally isn't documented. @lukebjerring, did you run them locally, and were there sufficient docs for how?

lukebjerring:
AFAIK, You don't do anything special to get `https` tests to work on `localhost` (considered a secure origin?)

> Found 95 tests
> 93 Pass
> 2 Fail

foolip:
Right, localhost is considered a secure context, problem is I was using http://web-platform.test:8000, and https://web-platform.test:8443 doesn't work unless you fiddle with the certs of your real browser, which I didn't.

PR #11976, opened 2018-07-13
Convert config to a builder pattern

jgraham:
This solves several problems with the curent design, notably:
* Config is created once and then shared, so multiple processses/threads can't race when creating the configuration data
* It unbreaks the SSL environments that expect to be used as context managers

---
This change is [Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/11976)

### DIFF
--- a/tools/ci/make_hosts_file.py
+++ b/tools/ci/make_hosts_file.py
@@ -3,14 +3,19 @@ import os
 
 from ..localpaths import repo_root
 
-from ..serve.serve import load_config, make_hosts_file
+from ..serve.serve import build_config, make_hosts_file
+
 
 def create_parser():
     parser = argparse.ArgumentParser()
-    parser.add_argument("address", default="127.0.0.1", nargs="?", help="Address that hosts should point at")
+    parser.add_argument("address", default="127.0.0.1", nargs="?",
+                        help="Address that hosts should point at")
     return parser
 
-def run(**kwargs):
-    config = load_config(os.path.join(repo_root, "config.json"))
 
-    print(make_hosts_file(config, kwargs["address"]))
+def run(**kwargs):
+    config_builder = build_config(os.path.join(repo_root, "config.json"),
+                                  ssl={"type": "none"})
+
+    with config_builder as config:
+        print(make_hosts_file(config, kwargs["address"]))
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -430,7 +430,6 @@ class ServerProc(object):
 def check_subdomains(config):
     paths = config.paths
     bind_address = config.bind_address
-    ssl_config = config.ssl_config
     aliases = config.aliases
 
     host = config.server_host
@@ -656,8 +655,8 @@ def iter_procs(servers):
             yield server.proc
 
 
-def load_config(override_path=None, **kwargs):
-    rv = Config()
+def build_config(override_path=None, **kwargs):
+    rv = ConfigBuilder()
 
     if override_path and os.path.exists(override_path):
         with open(override_path) as f:
@@ -695,10 +694,10 @@ _subdomains = {u"www",
 _not_subdomains = {u"nonexistent"}
 
 
-class Config(config.Config):
+class ConfigBuilder(config.ConfigBuilder):
     """serve config
 
-    this subclasses wptserve.config.Config to add serve config options"""
+    this subclasses wptserve.config.ConfigBuilder to add serve config options"""
 
     _default = {
         "browser_host": "web-platform.test",
@@ -735,29 +734,30 @@ class Config(config.Config):
         "aliases": []
     }
 
+    computed_properties = ["ws_doc_root"] + config.ConfigBuilder.computed_properties
+
     def __init__(self, *args, **kwargs):
-        super(Config, self).__init__(
+        super(ConfigBuilder, self).__init__(
             subdomains=_subdomains,
             not_subdomains=_not_subdomains,
             *args,
             **kwargs
         )
 
-    @property
-    def ws_doc_root(self):
-        if self._ws_doc_root is not None:
-            return self._ws_doc_root
+    def _get_ws_doc_root(self, data):
+        if data["ws_doc_root"] is not None:
+            return data["ws_doc_root"]
         else:
-            return os.path.join(self.doc_root, "websockets", "handlers")
+            return os.path.join(data["doc_root"], "websockets", "handlers")
 
-    @ws_doc_root.setter
     def ws_doc_root(self, v):
         self._ws_doc_root = v
 
-    @property
-    def paths(self):
-        rv = super(Config, self).paths
-        rv["ws_doc_root"] = self.ws_doc_root
+    ws_doc_root = property(None, ws_doc_root)
+
+    def _get_paths(self, data):
+        rv = super(ConfigBuilder, self)._get_paths(data)
+        rv["ws_doc_root"] = data["ws_doc_root"]
         return rv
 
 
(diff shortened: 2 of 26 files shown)


tools/wptserve/wptserve/config.py
```diff
@@ -57,14 +104,34 @@ class Config(Mapping):
...
 
     def __init__(self,
                  logger=None,
                  subdomains=set(),
                  not_subdomains=set(),
+                 config_cls=None,
                  **kwargs):
 
-        self.log_level = kwargs.get("log_level", "DEBUG")
+        self._data = self._default.copy()
```
Hexcles: Nit: perhaps we can make `self._default` a `frozendict` for an extra layer of safety.

tools/serve/serve.py
```diff
@@ -695,7 +685,7 @@ def load_config(override_path=None, **kwargs):
 _not_subdomains = {u"nonexistent"}
 
 
-class Config(config.Config):
+class ConfigBuilder(config.ConfigBuilder):
     """serve config
 
     this subclasses wptserve.config.Config to add serve config options"""
```
Hexcles: ...subclasses `wptserv.config.ConfigBuilder`*...

tools/ci/make_hosts_file.py
```diff
@@ -3,14 +3,16 @@
...
     parser.add_argument("address", default="127.0.0.1", nargs="?", help="Address that hosts should point at")
     return parser
 
 def run(**kwargs):
-    config = load_config(os.path.join(repo_root, "config.json"))
+    config_builder = build_config(os.path.join(repo_root, "config.json"),
+                                  ssl={"type":"none"})
 
-    print(make_hosts_file(config, kwargs["address"]))
+    with config_builder.get() as config:
```
Hexcles: I might have missed something but where is `ConfigBuilder.get` defined?

jgraham: That's just a bug; the initial iteration had a `.get` method. I'm surprised the tests didn't catch this.

tools/wptserve/wptserve/sslutils/openssl.py
```diff
@@ -361,7 +362,7 @@ def host_cert_path(self, hosts):
 
         hosts must be a list of all hosts to appear on the certificate, with
         the primary hostname first."""
-        hosts = tuple(sorted(hosts, key=lambda x:-len(x)))
+        hosts = tuple(sorted(hosts, key=lambda x:len(x)))
```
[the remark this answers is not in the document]
jgraham: I think this is expained in the commit message; we sort in order to get the top-level domain first. That wasn't what was happening ;)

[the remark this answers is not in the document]
Hexcles: Ahh I didn't read the commit messages. Not sure if you plan to squash or rebase merge this, but as long as the message goes to the final history I'm happy.

dheiberg (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/11976)* status: 0 of 22 files reviewed, 8 unresolved discussions (waiting on @Hexcles, @jgraham, @dhdavvie, and @gsnedders)

___
*[tools/wptrunner/wptrunner/environment.py, line 107 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11976#-LHXdQkU0r6zfgB7eBHF:-LHXdQkU0r6zfgB7eBHG:bol5vkf) ([raw file](https://github.com/web-platform-tests/wpt/blob/3f02f9d31f819be60a06783b868a602784c00c41/tools/wptrunner/wptrunner/environment.py#L107)):*
> ```Python
>         self.cache_manager.__exit__(exc_type, exc_val, exc_tb)
>         self.stash.__exit__()
>         self.config_ctx.__exit__(exc_type, exc_val, exc_tb)
> ```

Is there any reason for this now being after `self.stash.__exit__()`? Or is the order arbitrary?
___
*[tools/wptrunner/wptrunner/wptrunner.py, line 39 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11976#-LHXdkmtCvCKAy5tz_I4:-LHXdkmtCvCKAy5tz_I5:bhpqx9s) ([raw file](https://github.com/web-platform-tests/wpt/blob/3f02f9d31f819be60a06783b868a602784c00c41/tools/wptrunner/wptrunner/wptrunner.py#L39)):*
> ```Python
> """
> 
> print sslutils.__file__
> ```

stray print statement by the looks of it

tools/wptserve/wptserve/config.py
jgraham: Not sure it's worth implementing this or importing a third party package just for this.

tools/ci/make_hosts_file.py
Hexcles: Sorry please disregard. I wrote too much code in blinkpy (which uses `frozendict` a lot) to the point I somehow thought PEP 416 was implemented...

tools/serve/serve.py
```diff
@@ -735,29 +734,30 @@ class Config(config.Config):
...
             not_subdomains=_not_subdomains,
             *args,
             **kwargs
         )
 
-    @property
-    def ws_doc_root(self):
-        if self._ws_doc_root is not None:
-            return self._ws_doc_root
+    def _get_ws_doc_root(self, data):
```
gsnedders: I'm confused by this; we appear to have a private getter never called from anywhere?

jgraham: OK, so what do you suggest? The requirements are
* Properties must be computed in a defined order, since later properties often depend on earlier ones.
* Subclasses must be able to add to the list of properties in a relatively straightforward manner.

We could just have a list of methods to call on the class, but that's a bit ugly because they have to be defined at the bottom of the class definition whereas it's more usual to define data properties at the top.

gsnedders: A comment pointing at how it gets called and turned into an actual property would probably be useful?

jgraham: Added something to the docstring for the class describing the protocol.

Hexcles (changes requested):
> Reviewed 17 of 22 files at r1, 9 of 9 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/11976)* status: 25 of 26 files reviewed, 3 unresolved discussions (waiting on @Hexcles and @jgraham)

___
*[tools/wpt/run.py, line 100 at r2](https://reviewable.io/reviews/web-platform-tests/wpt/11976#-LHe1hi22LoSF7-d1jW9:-LHe1hi22LoSF7-d1jWA:b-9lmuyd) ([raw file](https://github.com/web-platform-tests/wpt/blob/9bf86d7e3dbc8874f507da766d8f88bf079c13cf/tools/wpt/run.py#L100)):*
> ```Python
>         with config_builder as config:
>             expected_hosts = set(config.domains_set)
>             is_windows = platform.uname()[0] == "Windows"
> ```

Could you move this statement back (out of the `with` block)?
___
*[tools/wptrunner/wptrunner/wptrunner.py, line 39 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/11976#-LHXdkmtCvCKAy5tz_I4:-LHiLy2t0WXejEiadMte:bkml4sv) ([raw file](https://github.com/web-platform-tests/wpt/blob/3f02f9d31f819be60a06783b868a602784c00c41/tools/wptrunner/wptrunner/wptrunner.py#L39)):*

stray print statement by the looks of it
</blockquote></details>

ping @jgraham

Hexcles (approved):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/11976)* status: 23 of 26 files reviewed, 2 unresolved discussions (waiting on @Hexcles, @jgraham, and @dhdavvie)

PR #11990, opened 2018-07-14
Adding custom pytest marker registration to WebDriver tests

jimevans:
The WebDriver tests can now use a custom pytest marker to specify the
capabilities for a session in a specific test. The marker must be
registered with the system to be valid, however.


gsnedders (approved):

whimboo:
Thanks @jimevans for fixing it. Looks like I totally missed that when adding the feature.

PR #11993, opened 2018-07-15
Create multiple files in single temp directory for upload tests

jimevans:
Some user agents do not support multiple files in input elements in the
file upload state if the files selected are in different directories.
This commit adds the ability to create multiple temporary files in a
single temp directory instead of creating each file in its own
directory. Additionally, it adds a test uploading multiple files in
different directories for coverage of that case.

### DIFF
--- a/webdriver/tests/element_send_keys/conftest.py
+++ b/webdriver/tests/element_send_keys/conftest.py
@@ -12,7 +12,7 @@ def create_file(tmpdir_factory):
     inner.__name__ = "create_file"
     return inner
 
-	
+
 @pytest.fixture
 def create_files(tmpdir_factory):
     def inner(filenames):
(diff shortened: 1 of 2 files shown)


andreastt (approved):

whimboo:
@jimevans, wouldn't it have been easier to just update the `create_file` fixture? I don't see a reason why we have to keep both now. I'm fine with using `create_files` with just a single entry in the array. It's a bit confusing right now.

jimevans:
@whimboo My change here was intended to be minimally disruptive, so I opted for additive, rather than transformative. This is in part due to my lack of knowledge about other parts of the web platform (e.g. whether it’s imperative that the `multiple` attribute allow selection of files from different directories, and that the platform for which I’m “responsible” is just non-conforming). My reasearch on this particular topic was inconclusive. If you want to revert and/or refactor the fixtures to be more in line with a cleaner vision, feel free to do so; I’ve no objection whatever to that.

andreastt:
Testing if the UA supports upload from different directories is beyond the scope of this specification.

whimboo:
Thanks @jimevans. Now that your change has been downstreamed to our repository I will work on removing the `create_file` fixture via https://bugzilla.mozilla.org/show_bug.cgi?id=1479677.

PR #11994, opened 2018-07-15
Updating Safari executor to be able to execute wdspec tests

jimevans:


jimevans:
@burg Can you see if this is the right way to enable Safari's driver for `wdspec` tests?

jgraham (approved):

burg:
@jimevans What's the simplest way to test that out from wpt trunk?

jimevans:
@burg Easiest is to do something like this:

    ./wpt run safari webdriver --test-type wdspec --webdriver-binary /path/to/safaridriver 

If things run in the browser, you’re good.

burg:
It seems to work:

Ran 1146 checks (89 tests, 1057 subtests)
Expected results: 833
Unexpected results: 313
  test: 1 (1 timeout)
  subtest: 312 (308 fail, 4 error)

PR #11996, opened 2018-07-16
Add tests for cookieEnabled

### LINKED ISSUE #3622: Test for cookieEnabled
Doesn't seem like there are any.

xiuqijix:
Fix #3622, @Ms2ger, PTAL.

### DIFF
--- a/infrastructure/assumptions/cookie.html
+++ b/infrastructure/assumptions/cookie.html
@@ -6,7 +6,12 @@
 <script src="/resources/testharnessreport.js"></script>
 
 <script>
-  test(() => {
+  test(t => {
+    t.add_cleanup(() => {
+      let date = new Date();
+      date.setTime(date.getTime() - 10000);
+      document.cookie = "name=''; expires=" + date.toGMTString();
+    });
     document.cookie = "name=test_cookie";
     assert_not_equals(document.cookie.match(/name=test_cookie/), null);
   });
--- a/infrastructure/assumptions/cookie.html
+++ b/infrastructure/assumptions/cookie.html
@@ -1,0 +1,13 @@
+<!doctype html>
+<title>cookies work in default browse settings</title>
+<link rel="author" title="Intel" href="http://www.intel.com">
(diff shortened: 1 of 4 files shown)


gsnedders:
We really prefer avoiding manual tests; it would be much better to just come up with the expected value for `cookieEnabled` based on whether we can set cookies from the server.

foolip:
@gsnedders I think the trouble is that there is no way to disable cookies for all sites, https://w3c.github.io/webdriver/#cookies only has APIs to get/set/delete cookies. It's not a given that browsers even have such a setting internally, since disabling it would break a lot of stuff, although it appears that Chrome does.

@kaixinjxq, how about adding a basic cookie test in https://github.com/web-platform-tests/wpt/tree/master/infrastructure/assumptions that asserts that cookies work in the default WPT setup, another test that just asserts that `cookieEnabled` is true, and then keeping the manual test for the false case?

kaixinjxq:
@foolip, thanks for your comments, I've addressed it in new commit, PTAL.

infrastructure/assumptions/cookie.html
```diff
@@ -0,0 +1,13 @@
...
+<link rel="author" title="Intel" href="http://www.intel.com">
+<link rel="help" href="https://html.spec.whatwg.org/#dom-document-cookie">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<script>
+  test(() => {
+    document.cookie = "name=test_cookie";
+    assert_not_equals(document.cookie.match(/name=test_cookie/), null);
+  });
```
gsnedders: Can you add a cleanup function here to remove all cookies?

Arguably we actually want something that ensures that cookies work across navigation, but I'll let others decide whether that's worth having here?

kaixinjxq: @gsnedders, I've addressed the suggestion in new commit, PTAL.

foolip (approved):
Thank for the follow up! As you can tell, I missed it in my inbox. Please poke over and over if this happens :)

PR #12004, opened 2018-07-16
Adding explicit call to session.end for WebDriver delete session tests

jimevans:
For the "delete session" command tests, the recent changes to the fixture
to create and delete sessions are bypassed. This means that at the end of
each delete session test, the harness thinks a valid session still exists.
Making an explicit call to `session.end()` at the end of each test will
put the tracking variables in the correct state for the harness to start a
new session with the next test.

### DIFF
--- a/webdriver/tests/delete_session/delete.py
+++ b/webdriver/tests/delete_session/delete.py
@@ -13,6 +13,9 @@ def test_null_response_value(session):
     response = delete_session(session)
     value = assert_success(response)
     assert value is None
+    # Need an explicit call to session.end() to notify the test harness
+    # that a new session needs to be created for subsequent tests.
+    session.end()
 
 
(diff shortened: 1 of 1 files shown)


andreastt (approved):

jimevans:
@whimboo I don’t know what to tell you, except that both tests in that source file now pass for the IE implementation. I freely admit that I don’t understand the WebDriver WPT test harness and how it does session management. If you feel a refactor is appropriate, make whatever changes you think are required, as long as it doesn’t cause spurious failures for the IE implementation.

andreastt:
From an API standpoint `session.end()` could be made to swallow the `SessionNotCreatedException` since it should be deteremental to call this API subsequently.

I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1477978 about this.

PR #12008, opened 2018-07-16
WIP: Add basic blending tests.

Tavmjong:
Some basic test for blending (part of w3c/svgwg#492).

Note: In Chrome there is a slight visual difference in colors between the test and reference.


svgeesus (approved):
Straightforward, lgtm

Issue #10330, opened 2018-04-05
Say something useful when someone tries to run a testdriver test with wpt serve


kereliuk:
Testdriver tests can only be run using wpt run. I think we should provide some sort of message indicating this is the case when someone tries to run the test using wpt serve.

cc/ @foolip @jgraham WYT?

foolip:
@kereliuk, what do you think we should do, concretely?

jugglinmike:
In gh-11173, I'm proposing that we don't force the test to fail but instead allow authentic human input and react accordingly. As described in the patch, that would allow test authors and visitors to http://w3c-test.org to validate testdriver.js-powered tests.

lukebjerring:
Even if Mike's tweak lands (👍 by the way, Mike), we should have some docs @ https://web-platform-tests.org/writing-tests/testdriver.html which explain that there are differences between `run` and `serve` for this case.

foolip:
@kereliuk, can you look into adding such documentation?

jugglinmike:
Since gh-11173 introduces the new behavior for `wpt serve` contexts, and since that hasn't been merged yet, it's probably best for me to add the documentation that @lukebjerring has requested to that patch. I'll do that later today.

jugglinmike:
Actually, it looks like that patch already updates the documentation:

```diff
--- a/docs/_writing-tests/testdriver.md
+++ b/docs/_writing-tests/testdriver.md
@@ -5,7 +5,9 @@ order: 8.5
 ---
 
 testdriver.js provides a means to automate tests that cannot be
-written purely using web platform APIs.
+written purely using web platform APIs. Outside of automation
+contexts, it allows human operators to provide expected input
+manually.
 
 It is currently supported only for [testharness.js][testharness]
 tests.
```

I'll follow up about potential improvements on the pull request.

foolip:
Included in [Ecosystem Infra 2018 Q4 OKRs](https://docs.google.com/document/d/1UKuPaQm2YGHYEfbrrYBHoCUGAVhyQZEp8nE09qT7vYg/edit#heading=h.jwokykbtd9cq).

I think we can use `navigator.webdriver` for this, somehow.

mdittmer:
_Ping from your friendly neighbourhood ecosystem infra rotation_

Any updates on this, @gsnedders?

gsnedders:
@jugglinmike worked on this a while ago in #11173, and that's had some more activity recently.

[...]

PR #12010, opened 2018-07-16
Request pip via trusted click

lukebjerring:
Fixes a bug in the picture-in-picture IDL test, which intends to obtain a `PictureInPictureWindow` object for testing IDL.

### DIFF
--- a/picture-in-picture/idlharness.window.js
+++ b/picture-in-picture/idlharness.window.js
@@ -9,8 +9,7 @@
 
 promise_test(async () => {
   try {
-    const video = await loadVideo();
-    self.video = video;
+    self.video = await loadVideo();
     self.pipw = await requestPictureInPictureWithTrustedClick(video);
   } catch (e) {
     // Will be surfaced when video/pipw are undefined below.
--- a/picture-in-picture/idlharness.window.js
+++ b/picture-in-picture/idlharness.window.js
@@ -10,7 +10,6 @@
 promise_test(async () => {
   try {
     const video = await loadVideo();
-    document.body.appendChild(video);
     self.video = video;
     self.pipw = await requestPictureInPictureWithTrustedClick(video);
   } catch (e) {
--- a/picture-in-picture/idlharness.window.js
+++ b/picture-in-picture/idlharness.window.js
@@ -1,6 +1,7 @@
 // META: script=/resources/WebIDLParser.js
 // META: script=/resources/idlharness.js
 // META: script=/resources/testdriver.js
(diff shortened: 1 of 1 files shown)


beaufortfrancois (approved):

picture-in-picture/idlharness.window.js
```diff
@@ -11,7 +12,8 @@ promise_test(async () => {
     const video = await loadVideo();
     document.body.appendChild(video);
     self.video = video;
-    self.pipw = await video.requestPictureInPicture();
+    video.autoplay = true;
```
beaufortfrancois: Nit: `video.autoplay` shoud not be needed.

picture-in-picture/idlharness.window.js
```diff
@@ -11,7 +12,8 @@ promise_test(async () => {
     const video = await loadVideo();
     document.body.appendChild(video);
```
beaufortfrancois: Nit: This should not be needed as well.

picture-in-picture/idlharness.window.js
```diff
@@ -11,7 +12,8 @@ promise_test(async () => {
     const video = await loadVideo();
```
beaufortfrancois: It could be as simple as `self.video = await loadVideo();`

lukebjerring:
@kereliuk - this actually fails, since `click` doesn't seem to be implemented. Where are we at with simulating user clicks via `testdriver.js` ?

EDIT: Calling `click` fails when running the test using `./wpt serve` and loading the URL locally:

    Uncaught (in promise) Error: unimplemented
        at Object.click (testdriver.js:185)
        at Object.click (testdriver.js:119)
        at requestAnimationFrame (picture-in-picture-helpers.js:14)

It fails differently using `./wpt run chrome picture-in-picture/idlharness.window.html`:

    Uncaught TypeError: Cannot read property 'postMessage' of null
        at Object.window.test_driver_internal.click (testdriver.js:269)
        at Object.click (testdriver.js:119)
        at requestAnimationFrame (picture-in-picture-helpers.js:14)

kereliuk:
@lukebjerring testdriver cannot be used with wpt serve, only with wptrunner

lukebjerring:
OK so before I get too tetchy about it not being documented, there's already an issue open for that:
https://github.com/web-platform-tests/wpt/issues/10330

lukebjerring:
OK so if I run this locally with `wpt run`, in its current state, testdriver.js behaves as expected (the click event happens).

PR #12013, opened 2018-07-16
WebDriver: fix expected key code for 'Unidentified' keys.

burg:
Per the UI Events specification (https://w3c.github.io/uievents-code/#key-legacy),
keys such as \u00e0 (i.e., à) should have a code of "Unidentified". Some tests in
key.py expect an empty value, which seems to be an oversight.

### DIFF
--- a/webdriver/tests/actions/key.py
+++ b/webdriver/tests/actions/key.py
@@ -26,11 +26,11 @@ def test_lone_keyup_sends_no_events(session, key_reporter, key_chain):
     ("a", "KeyA",),
     (u"\"", "Quote"),
     (u",", "Comma"),
-    (u"\u00E0", ""),
-    (u"\u0416", ""),
+    (u"\u00E0", "Unidentified"),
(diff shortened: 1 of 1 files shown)


burg:
I don't quite understand this CI failure. Did Chrome previously pass these tests?

ERROR:/home/travis/build/web-platform-tests/wpt/tools/ci:| `/webdriver/tests/actions/key.py` | `test_single_printable_key_sends_correct_events["-Quote]`                   | **ERROR, FAIL** | `teardown error;UnknownCommandException: unknown command (404): (Session infochrome=69.0.3486.0)` |

jgraham:
The CI failure is a bug in the harness that makes it look like Chrome gets multiple results for tests it fails. If this PR is still valid I can merge it.

whimboo:
The PR would need an update given that the tests have been relocated and refactored.

andreastt:
@burg Do you want to have a stab at updating the PR? I think it’s a worthwhile fix to have.

PR #11836, opened 2018-07-06
Update the paint-timing IDL file

lukebjerring:
Hello reviewer(s),

This PR is intended to consolidate the spec’s IDL definition with the WPT test suite’s copy, and any idlharness tests.

The up-to-date copy of the IDL file was automatically extracted from the reffy-reports repo (https://github.com/tidoust/reffy-reports/tree/master/whatwg/idl) which scrapes known specs automatically + regulary.

This PR is part of a migration project which will eventually be automatically updating and creating PRs for changes in spec IDL.

Please check that:
The spec (and its source) is correct and up-to-date
All tests which cover the IDL in the spec have been migrated to fetch + use the idl in the `interfaces/` directory (instead of inline copies in the test files).

PR #12014, opened 2018-07-16
Expand coverage of performance-timeline IDL test

lukebjerring:
Follow up from https://github.com/web-platform-tests/wpt/issues/11911

Adds code for grabbing an instance of `PerformanceObserverEntryList` and `PerformanceObserverEntry` (correctly).

Passes (67/67) in Chrome 67.

### DIFF
--- a/performance-timeline/idlharness.any.js
+++ b/performance-timeline/idlharness.any.js
@@ -9,7 +9,6 @@
 promise_test(async t => {
   const observe = new Promise((resolve, reject) => {
     try {
-      const callback = (e, o) => { };
       self.observer = new PerformanceObserver((entries, observer) => {
         self.entryList = entries;
         resolve();
--- a/performance-timeline/idlharness.any.js
+++ b/performance-timeline/idlharness.any.js
@@ -6,11 +6,16 @@
 
 'use strict';
 
+function cast(i, t) {
+  return Object.assign(Object.create(t.prototype), i);
+}
+
 promise_test(async t => {
   const observe = new Promise((resolve, reject) => {
     try {
       self.observer = new PerformanceObserver((entries, observer) => {
         self.entryList = entries;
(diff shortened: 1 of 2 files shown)


lukebjerring:
@foolip - this is another example of a change where #11836 would help (by handling the resolve/reject outcomes of the `Promise.race`).

performance-timeline/idlharness.any.js
```diff
@@ -6,22 +6,44 @@
...
+    idl_test(
+      ['performance-timeline'],
+      ['hr-time', 'dom'],
+      idl_array => {
+        idl_array.add_objects({
+          Performance: ['performance'],
+          PerformanceObserver: ['observer'],
+          // NOTE: PerformanceEntry is implicitly tested in user-timing.
+          PerformanceObserverEntryList: ['entryList'],
+          PerformanceObserverEntry: ['entryList[0]'],
```
npm1: I don't think there's an interface called PerformanceObserverEntry, this is a PerformanceEntry right? Not sure if the name given here matters or not.

lukebjerring: Yep, the name definitely matters. Corrected the test, by pulling in `user-timing` for a class that extends `PerformanceEntry`. PTAL.

performance-timeline/idlharness.any.js
```diff
@@ -6,11 +6,16 @@
 
 'use strict';
 
+function cast(i, t) {
```
npm1: Hmm where is this used?

npm1 (approved):

performance-timeline/idlharness.any.js
lukebjerring: Oh - at first I was casting `PerformanceMark` as `PerformanceEntry`, but that didn't work as intended.

PR #12037, opened 2018-07-17
Add test for document.open() in custom element constructor

TimothyGu:
Blink and Gecko pass both tests. WebKit has a [bug](https://bugs.webkit.org/show_bug.cgi?id=187319) open, as well as a [FIXME](https://github.com/WebKit/webkit/blob/e50db3db8d808b34f1e37655aa8a371d9f891dcc/Source/WebCore/dom/Document.cpp#L2615-L2616) in the source code. Edge doesn't yet implement custom elements.


domenic (approved):
Flow is a bit strange with the global variable storage and the like, but I can't think of anything better. LGTM.

PR #12039, opened 2018-07-17
Update the webauthn IDL file + test

lukebjerring:
Hello reviewer(s),

Please check that:


webauthn/idlharness.https.any.js
```diff
@@ -0,0 +1,60 @@
...
+
+promise_test(() => {
+  const execute_test = () => idl_test(
+    ['webauthn'],
+    ['credential-management'],
+    idlArray => {
+      idlArray.add_objects({
+        WebAuthentication: ['navigator.authentication'],
+        PublicKeyCredential: ['cred'],
+        AuthenticatorAssertionResponse: ['assertion']
```
lukebjerring: @jcjones @apowers313 - I'm not sure how to check whether I've set up these tests correctly.

In short, these two fields `self.cred` and `self.assertion` are supposed to be a `PublicKeyCredential` and a `AuthenticatorAssertionResponse` respectively. Is that what you would expect to result from the code below? I can see the prompt in FF61, but how do I add a security token to Firefox?

apowers313:
cc: @christiaanbrand @kpaulh

kpaulh:
@lukebjerring it definitely is on the radar. We have a virtual authenticator for our own layout tests (https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/http/tests/credentialmanager/resources/virtual-navigator-credentials.js), but Firefox, MS, and Chrome would have to agree on a uniform testing API to do the same for WPTs. (Similar to WebBluetooth's testing API)

apowers313 (approved):
I reviewed against the current spec and it looks good to me. Need to make sure that browsers are planning on implementing the `tokenBinding` "not-supported" change since this is based against the latest editorial spec and not the CR version and it may impact server implementations.

cc: @akshayku @kpaulh @jcjones

apowers313:
@lukebjerring we had talked about creating a mock authenticator framework, but it hasn't come to fruition yet. Right now, I think each vendor has it's own soft authentication (e.g. - FF has a flag for enabling it's internal soft authenticator).

lukebjerring:
Took a stab at an automated version of the same test:
https://chromium-review.googlesource.com/c/chromium/src/+/1144133

@kpaulh can you take a look at that and we'll continue the discussion there?

PR #12066, opened 2018-07-19
wai-aria: add tests for aria-expanded on menuitem and its subclasses

joanmarie:


wai-aria/menuitem_expanded_false-manual.html
```diff
@@ -0,0 +1,78 @@
...
+    <title>aria-expanded=false on a menuitem</title>
+    <meta content="text/html; charset=utf-8" http-equiv="Content-Type"/>
+    <link rel="stylesheet" href="/wai-aria/scripts/manual.css">
+    <script src="/resources/testharness.js"></script>
+    <script src="/resources/testharnessreport.js"></script>
+    <script src="/wai-aria/scripts/ATTAcomm.js"></script>
+    <script>
+    setup({explicit_timeout: true, explicit_done: true });
+
+    var theTest = new ATTAcomm(
```
fred-wang: Not specific to this test, but I think it is not really nice to have an apparently unused "theTest" variable... I guess it is actually known by ATTAcomm.js but probably the API should be improved to make things more explicit.

fred-wang (approved):
Please check the spacing of the test.

I checked the structure of the tests and it looks good, but I'm not familiar with platform a11y APIs to confirm the tests are correct so I'll trust you ;-)

wai-aria/menuitem_expanded_false-manual.html
halindrome: ATTAcomm is an object, so (as far as I know) you need to tie it to a variable when you instantiate it.   In theory the generated test can use this variable to access aspects of the test (including the pending promises).   In reality, we have not found a use for them yet.  But I am not actually sure what would happen if just called the ATTAcomm method...  Anyway, thanks for the comment.  I will noodle on that.

joanmarie: Thanks for noodling, @halindrome. :smile: Related aside, if you do change this and the associated test generator, could you also either please modify the tests or let me know so that I can do so?

PR #12079, opened 2018-07-19
Change webauthn test to window test only

lukebjerring:
`webauthn` interfaces are only exposed on `Window`, so `.any.js` doesn't make sense (and, as expected, doesn't pass..)


Hexcles (approved):
LGTM

PR #12118, opened 2018-07-21
[permissions] Add tests for "Set Permission"

jugglinmike:
https://w3c.github.io/permissions/#automation

### DIFF
--- a/webdriver/tests/permissions/set.py
+++ b/webdriver/tests/permissions/set.py
@@ -1,0 +1,135 @@
+from tests.support.asserts import assert_error, assert_success
+import pytest
+
+def query(session, name):
+    script = """
+        var done = arguments[0];
+        navigator.permissions.query({ name: '%s' })
+          .then(function(value) {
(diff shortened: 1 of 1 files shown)


andreastt:
@jugglinmike @burg 

Is the WebDriver extension in Permissions in sync with the implementation in Safari? Are these tests in sync with the specification and/or reality?

jugglinmike:
These tests are in sync with the specification.

andreastt:
Tests are failing in Chrome. Retriggered them to see if that makes a difference.

andreastt (approved):
Thanks for your feedback @jugglinmike!

I don’t have a particular opinion on this API, but if this is defined in Permissions we should have tests for it in WPT. It would be good if we could get some feedback from @burg before we land it, but I’m OK with it.

burg:
These tests look good to me. Thanks!

andreastt:
@jugglinmike Do you want to do the honours when you have time?

jugglinmike:
Thanks for the help!

PR #12124, opened 2018-07-22
Add a test for Gecko bug 1472169.

emilio:
It's not clear that we want the fix over there given the performance
implications, but landing the test is still worth it.


tkent-google (approved):

emilio:
Thanks!

PR #12125, opened 2018-07-22
Use generic hasOwnProperty call for asserted objects

zenmumbler:
This fixes issues in the streams tests that call assert_equal and friends on objects explicitly created with a null prototype.  Only the calls to hasOwnProperty on external objects are replaced.

Changed in main harness as it (likely?) is not a bad idea to assume other oddly-shaped objects will be passed to the assert functions.

The commit making the changes to streams that breaks these fixes is: https://github.com/whatwg/streams/commit/a27a1fd0f8640d6054e0cbc8b6d7f5464dd096ab

### DIFF
--- a/resources/testharness.js
+++ b/resources/testharness.js
@@ -1007,7 +1007,7 @@ policies and contribution forms [3].
 
              var p;
              for (p in actual) {
-                 assert(expected.hasOwnProperty(p), "assert_object_equals", description,
+                 assert(Object.prototype.hasOwnProperty.call(expected, p), "assert_object_equals", description,
                                                     "unexpected property ${p}", {p:p});
 
                  if (typeof actual[p] === "object" && actual[p] !== null) {
@@ -1021,7 +1021,7 @@ policies and contribution forms [3].
                  }
              }
              for (p in expected) {
-                 assert(actual.hasOwnProperty(p),
+                 assert(Object.prototype.hasOwnProperty.call(actual, p),
                         "assert_object_equals", description,
                         "expected property ${p} missing", {p:p});
              }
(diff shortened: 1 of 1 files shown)


domenic:
It is indeed worth investigating what's going wrong here. It should never be possible for JavaScript to observe null-prototype objects created by a streams implementation, so if yours is doing that, and you think it's following the spec to the letter, then either you're mistaken, or the spec is wrong.

zenmumbler:
Not sure if I was clear, I'm not saying the spec or the ref impl is wrong or unclear, but the WPT tests simply cannot function with a per-spec implementation of the new hidden promises…… By design, they create the objects with a null prototype, but the wpt tests are keen to test these…  with the normal assert calls, which assume a normal operation… Hm. hm.

Given how the WPT tests like to rummage around in certain streams internals to really test details I assumed that it may access internal requests, but I did not verify this assumption. If the WPT tests only touch public / "forAuthorCode" objects they should just work normally indeed.

Ok, looked again and I was indeed far too concerned with my aforementioned assumption. The cause was the default/optional parameter `forAuthorCode` to `ReadableStreamDefaultReaderRead` set to `false` that hid the fact that I did not actually specify `true` for that parameter in the `ReadableStreamDefaultReader.read()` method.

TIL: do not doubt large test codebases; especially not on a Sunday in 28°C heat.
Thanks @domenic for saying the right things to show me the error of my thinking, sorry for taking your time this weekend, I will be do better next time. Retracting this PR.

gsnedders:
@zenmumbler While not necessary for streams, as I said I still think it would be a good idea to land this, probably with the array assert tests using `Object.prototype.hasOwnProperty.call` because the assertion functions probably should all cope with weirdness.

zenmumbler:
Hi @gsnedders, my concern is that if these changes were already in the test framework, they would not have caught the actual error I made. Having had to retrace my steps now I worry that this change may mask other subtle problems in the future. Maybe when needed a separate assert can be added for objects with an expected null prototype, but only once that becomes a issue. WPT is large though and I've only looked at the streams tests so there may already be potential test cases that can benefit from this, but this is beyond my purview.

gsnedders:
The idlharness tests should catch things being wrong in the prototype chain; I don't think we need to guard against this in every test.

PR #11980, opened 2018-07-13
Update the css-images IDL file

lukebjerring:
Hello reviewer(s),

This PR is intended to consolidate the spec’s IDL definition with the WPT test suite’s copy, and any idlharness tests.

The up-to-date copy of the IDL file was automatically extracted from the reffy-reports repo (https://github.com/tidoust/reffy-reports/tree/master/whatwg/idl) which scrapes known specs automatically + regulary.

This PR is part of a migration project which will eventually be automatically updating and creating PRs for changes in spec IDL.

Please check that:
The spec (and its source) is correct and up-to-date
All tests which cover the IDL in the spec have been migrated to fetch + use the idl in the `interfaces/` directory (instead of inline copies in the test files).

PR #12144, opened 2018-07-23
[idlharness.js] Assert correct type for original definition of partials

### LINKED ISSUE #12123: Check for incorrect mixing of interface and namespace
While investigating a surprising lack of regression in https://github.com/web-platform-tests/wpt/pull/11819, it turns out that mixing `interface CSS` + `partial namespace CSS` and `namespace CSS` + `partial interface CSS` work fine, and don't result in the extra failed tests.

Note that trying to define a partial namespace of some random name results in "Original namespace should be defined" so there is some checking in place which just isn't as strict as it should be.

lukebjerring:
Found during authorship of #11980 - The original definition of `CSS` is (currently) an `interface`, but having a `partial namespace CSS` didn't fail any tests.

This fixes that.

Fixes #12123

### DIFF
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -870,6 +870,16 @@ IdlArray.prototype.collapse_partials = function()
 
             test(function () {
                 assert_true(originalExists, `Original ${parsed_idl.type} should be defined`);
+
+                var expected = IdlInterface;
+                switch (parsed_idl.type) {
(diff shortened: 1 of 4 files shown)


Ms2ger (approved):

lukebjerring:
Fixes #12123

Issue #11764, opened 2018-07-03
Should response.writer.write_content implicitly write status, headers?


Ms2ger:
Right now, a handler like
```python
         @wptserve.handlers.handler
         def handler(request, response):
             response.writer.write_content("PASS")
```

will write the exact four bytes for "PASS", no status, no headers. The quoted handler came from a unit test that's broken in python3 because its urllib is stricter about responses like that.

Hexcles:
From an API design perspective (and my experience of using various HTTP libraries/web frameworks), I'd say it should, and we could have a separate `write_raw_content` API to write raw responses, but I'm not familiar with the use case of this particular API.

jgraham:
Yes, this is just a bug, I think. Maybe @dhdavvie will get a chance to have a look?

dhdavvie:
Sure I can take a crack at it. @Ms2ger so are you saying the expected functionality should be that if you call `write_content` before having sent headers/status, it should send some default headers and status?

I'm slightly worried about changing the nature of `write_content` as other tests/handlers may already be using it in the way that @Hexcles described, but will do some digging.

dhdavvie:
So it seems like you can just add a check for `content_written`, like so:

    def write_content(self, data):
        """Write the body of the response."""
        if not self.content_written:
            self.write_status(200)
            self.write_default_headers()
            self.end_headers()
        
        #continue

I'll setup this fix in a branch and write some tests for it

PR #12160, opened 2018-07-24
Ensure that `write_content` performs checks before writing

dheiberg:
As mentioned in #11764, the python3 HTTP lib doesn't support simply writing HTTP body without writing status and headers. To help with this, I've changed the way `write_content` behaves by no longer blindly writing the data given to it, but instead performing checks to see if a status has been written and if required headers have been written (if at all needed). 

I've moved the original behaviour of `write_content` to `write_raw_content` which blindly tries to write the data given to it to the `wfile`.

### DIFF
--- a/tools/wptserve/wptserve/response.py
+++ b/tools/wptserve/wptserve/response.py
@@ -470,11 +470,13 @@ class ResponseWriter(object):
         self._wfile = handler.wfile
         self._response = response
         self._handler = handler
+        self._status_written = False
         self._headers_seen = set()
         self._headers_complete = False
         self.content_written = False
         self.request = response.request
         self.file_chunk_size = 32 * 1024
+        self.default_status = 200
 
     def write_status(self, code, message=None):
         """Write out the status line of a response.
@@ -489,13 +491,18 @@ class ResponseWriter(object):
                 message = ''
         self.write("%s %d %s\r\n" %
                    (self._response.request.protocol_version, code, message))
+        self._status_written = True
 
     def write_header(self, name, value):
         """Write out a single header for the response.
 
+        If a status has not been written, a default status will be written (currently 200)
+
         :param name: Name of the header field
         :param value: Value of the header field
         """
+        if not self._status_written:
+            self.write_status(self.default_status)
         self._headers_seen.add(name.lower())
         self.write("%s: %s\r\n" % (name, value))
         if not self._response.explicit_flush:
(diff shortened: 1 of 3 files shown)


Hexcles (commented):
Can you add just some tests for `write_raw_content`?

On a related note, you mentioned that Python 3 would not allow requests without minimal headers. I take that as it'd raise an exception, and Python 2 doesn't. If that's the case, I feel like we should examine the argument of `write_raw_content` ourselves to make sure invalid responses will always fail hard in both Python 2 and 3.

Ms2ger:
Note that this was *fetching* the incomplete response with python3 urllib that failed, not creating the response on the server side.

dhdavvie:
@Ms2ger Ahh yes now I see! Sorry I misinterpreted the issue. Is the intention to have python3 behave like python2 in this case? As in when the response is not valid HTTP, it seems to auto assign status code 200 and just allow you to read the message

Hexcles:
@Ms2ger thanks. I also misunderstood the original issue. That said, I think this change is still good to have (but please disregard my comment about raising exceptions).

And as the PR is right now, it's possible to use `write_raw_content` to produce invalid responses (on both Python 2 and 3), which I think is fine and might in fact be useful if someone wants to test the handling of invalid HTTP responses of browsers (I believe we have some of these tests in Chromium). However, in unit tests of `wptserve` itself, we'd need to make sure the test case for `write_raw_content`  produce valid responses otherwise the test case would behave differently between Python 2 and 3.

dhdavvie:
I've updated the PR with a few tests for `write_raw_content`

tools/wptserve/wptserve/response.py
```diff
@@ -489,13 +490,16 @@ def write_status(self, code, message=None):
                 message = ''
         self.write("%s %d %s\r\n" %
                    (self._response.request.protocol_version, code, message))
+        self._status_written = True
 
     def write_header(self, name, value):
         """Write out a single header for the response.
```
Hexcles: Document the default status.

Hexcles (changes requested):
In addition to inline comments, I think `AsIsHandler` needs to be changed to use `write_raw_content` instead. I'm also surprised no test has caught the bug.

tools/wptserve/wptserve/response.py
dhdavvie: Oh yes I forgot to add `self.default_status = 200` in `__init__` to make it clearer instead of using magic numbers. Is that enough or do you think I should also add comments to document it?

tools/wptserve/wptserve/response.py
```diff
@@ -530,7 +534,21 @@ def end_headers(self):
         self._headers_complete = True
 
     def write_content(self, data):
-        """Write the body of the response."""
+        """
+        Write the body of the response. Performs checks to ensure status and headers
+        have been written.
```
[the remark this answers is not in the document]
dhdavvie: Good idea, I'll use this instead

dhdavvie:
I have already switched `AsIsHandler` to use `write_raw_content`, it's just a small file change so easy to miss

tools/wptserve/wptserve/response.py
Hexcles: Oh sorry, I meant to request to document the *behaviour* that a default status would be set rather than the *value* of the default status, but `self.default_status` is also great to have.

Hexcles (approved):
LGTM % a request for documentation.

I still don't know why Travis didn't catch the bug in AsIsHandler in the last iteration. I tested locally and without changing from `write_content` to `write_raw_content`, a unit test would fail. Yet since the PR was updated by force pushing I can't find the build history any more and don't want to dig any deeper. At least we know for sure there's a working test case.

dhdavvie:
@Hexcles just checking I can go ahead and merge this?

Hexcles:
@dhdavvie merged it fo ya. :)

PR #12170, opened 2018-07-24
[wptrunner] Fix regression in Sauce Labs browser

jugglinmike:
A recent commit [1] updated the invocation of the `start` method for
"Browser" instances by introducing a new keyword argument. This
interfered with the use of the "Sauce Labs" browser because its
implementation of the `start` method did not accommodate any keyword
arguments.

Update the `start` method to support (but ignore) all keyword arguments.

[1] 91803e3a25afd8e0c201bc57cd05fffd5d543551

---

@jgraham do you have any thoughts on how we could test for this?

### DIFF
--- a/tools/wptrunner/wptrunner/browsers/sauce.py
+++ b/tools/wptrunner/wptrunner/browsers/sauce.py
@@ -233,7 +233,7 @@ class SauceBrowser(Browser):
         Browser.__init__(self, logger)
         self.sauce_config = sauce_config
 
-    def start(self):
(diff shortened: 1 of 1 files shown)


jgraham (approved):

jgraham:
Not really good thoughts about how to test, other than actually running sauce. I suppose the other approach involves mocking out the actual sauce parts and just pretending to the code that we're running tests. Or, you know, using a language with a type system that can catch trivial errors…

Issue #11554, opened 2018-06-18
Support multiple document roots in wpt serve


jdm:
Servo uses a [config](https://github.com/servo/servo/blob/master/tests/wpt/config.ini) that has multiple document roots. I have not figured out a way to replicate this setup when using the `./wpt serve` tool, which only supports using a single document root.

foolip:
@jdm, can you say more about what this is used for? Are there kinds of tests that can't be written because of this?

jdm:
Servo has non-upstreamed tests in a separate document root. There was some reason I was using `./wpt serve` to run the server rather than our usual test harness when I filed this issue, but I can't remember why at this time.

foolip:
@jdm, should we close this issue, then? :)

jgraham:
I think this issue is reasonable to keep open.

jdm:
What I needed was the ability to run non-upstreamed tests in other browsers to verify compatibility. It would be nice if the wpt server supported this

foolip:
Ah, to graft some directory into what `wpt serve` runs, to fix stuff before really moving them into wpt? FWIW, we've had a somewhat related request in Chromium, which is to enable wptserve in directories outside of our wpt import directory. @Hexcles FYI.

PR #12172, opened 2018-07-24
Exposed support for multiple document roots

dheiberg:
As mentioned in #11554, some teams want to be able to use multiple document roots with `wpt serve`. The functionality was already there but was not really exposed in an easy way. I've added a command line argument `--alias_file` which takes the path of a file containing a list of aliases that looks like this:

    /test/,/home/usr/foo
    /other/,/home/usr2/bar

And invoked like this:

    $ ./wpt serve --alias_file aliases.txt

This will serve the main wpt doc root under `/`, the foo doc root under `/test/` and the bar doc root under `/other/`.

I hope this is the functionality that the issue was trying to find!

### DIFF
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -787,6 +787,8 @@ def get_parser():
                         help="Path to document root. Overrides config.")
     parser.add_argument("--ws_doc_root", action="store", dest="ws_doc_root",
                         help="Path to WebSockets document root. Overrides config.")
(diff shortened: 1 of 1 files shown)


foolip:
@jdm, can you take this one? I've never used this functionality.

jdm:
I tried this out and it looks like I can use relative paths for the alias sources. This fits my use case perfectly; thanks!

jdm (approved):

dhdavvie:
No worries, happy to help 😄

PR #12174, opened 2018-07-24
Fix resources/test and prevent infra changes from breaking the suite again

Hexcles:
The suite was broken by PR #11976 where the config class was changed.
The suite uses a hack to get the default config of wptserve. This commit
fixes the hack, but unfortunately does not find a less hacky way to get
the default config.

To prevent changes in tools (especially `wpt serve` and its
dependencies) from breaking the resources/test suite again, this commit
adds "tools/" to the watchlist of the suite.

### DIFF
--- a/resources/test/wptserver.py
+++ b/resources/test/wptserver.py
@@ -9,12 +9,15 @@ import urllib2
 class WPTServer(object):
     def __init__(self, wpt_root):
         self.wpt_root = wpt_root
+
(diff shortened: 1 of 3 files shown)


foolip (approved):
This is a bit like https://github.com/web-platform-tests/wpt/pull/10127 the last time around :)

jgraham:
Thanks for doing this; sorry for the bustage.

Issue #11388, opened 2018-06-06
Improve caching for manifest and lint


gsnedders:
Split out from https://github.com/web-platform-tests/wpt/issues/11356

One thing I've wanted to do for a long time is to massively improve caching for manifest and lint.

The status quo for manifest is we store a sha1 hash of the contents of each file and then compare it with a cached copy; this is expensive given it means we have to read every file to check if the cache is valid (this is especially notable on Windows due to https://github.com/web-platform-tests/wpt/issues/11386).

For lint we do all kinds of crazy magic to try and lint only a subset of files, based on git history.

We should instead:

 * use git blob IDs (in combination with the path where needed)
 * use file's mtime
 * maybe use hg node ids (or just the last revision, given I don't think hg has any public API or file format stability guarantees) for the sake of mozilla-central?

We need some combination of the above, given different things make sense in different cases (most obviously with untracked files but also the mozilla-central case).

We should absolutely be able to make manifest take almost no time at all when we have a cache (because we just need to verify that the tree still has the same blobs at the same paths) and make it possible for lint to always lint the entire tree mostly returning cached results (similarly in almost no time), though the simplest implementation would still to run all the `all_files_lints`.

Obviously any mtime-based caching would be filesystem specific, and hence not sharable. I'd like to suggest we add a local cache file, and make it optional to add a built-in cache to the generated manifest for the shared case. The local cache file could also easily store cached values for multiple revisions of a file, which would be useful.

gsnedders:
To add:

This might be a good opportunity to do a larger refactoring where we can make things like `--work` and similar be consistent (e.g., `--work` should generate a manifest for the working tree; `--head` should generate one for HEAD rather than the current default of "files in HEAD but the copies currently in the working tree"; `--index` should generate one for the index).

jgraham:
I have thoughts about this, which I'll add in the future, but for now @Cactusmachete who is likely to work on at least the manifest part of this.

jgraham:
So in general this sounds pretty sensible. I agree that reading all the file contents to generate a hash is slow and would be better as a fallback path.

My concern with using the version control bits is that it's harder to make the cache distributed in the face of multiple environments and multiple version control systems. For example a requirement we have for gecko is that we are able to download a manifest into a tree that might be `hg` and might be `git` and immediately get good performance (i.e. we need to be able to download at least some level of caching). The file contents works well for this because it's independent of any other considerations like the particular VCS system or the specific history that we've built up.

[...]

PR #12182, opened 2018-07-25
Rely on git hashes for manifest update

Cactusmachete:
We earlier hashed all in-tree files when updating the manifest.
Relying on git hashes instead gives us a speedup of ~50%
when running |wpt manifest|

With reference to https://github.com/web-platform-tests/wpt/issues/11388

### DIFF
--- a/tools/manifest/sourcefile.py
+++ b/tools/manifest/sourcefile.py
@@ -157,7 +157,7 @@ class SourceFile(object):
                          ("css", "CSS2", "archive"),
                          ("css", "common")}
 
-    def __init__(self, tests_root, rel_path, url_base, contents=None):
+    def __init__(self, tests_root, rel_path, url_base, hash=None, contents=None):
         """Object representing a file in a source tree.
 
         :param tests_root: Path to the root of the source tree
@@ -188,6 +188,7 @@ class SourceFile(object):
         self.meta_flags = self.name.split(".")[1:]
 
         self.items_cache = None
+        self._hash = hash
 
     def __getstate__(self):
         # Remove computed properties if we pickle this class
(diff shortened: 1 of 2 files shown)


tools/manifest/vcs.py
```diff
@@ -56,18 +56,21 @@ def _show_file(self, path):
...
-                    contents = None
-                yield SourceFile(self.root,
-                                 rel_path,
-                                 self.url_base,
-                                 contents=contents)
+          cmd = ["ls-tree", "-r", "-z", "HEAD"]
+          local_changes = self._local_changes()
+          for result in self.git(*cmd).split("\0")[:-1]:
+              rel_path = result.split("\t")[-1]
+              hash = result.split()[2]
```
jgraham: `hash` is also a bult-in function so maybe change the variable name.

tools/manifest/sourcefile.py
```diff
@@ -237,8 +238,12 @@ def url(self):
 
     @cached_property
     def hash(self):
-        with self.open() as f:
-            return hashlib.sha1(f.read()).hexdigest()
+        if not self._hash:
+            with self.open() as f:
+                content = f.read()
+            data = "blob " + str(len(content)) + "\0" + content
```
jgraham: So, I wonder if you profiled `"".join(["blob ", str(len(content)), "\0", content])` and `"blob %d\0%s" % (len(content), content)` vs this approach where the manifest isn't being generated from git (e.g. if you pass `--work` to the update). I don't know which is fastest, but it seems worth trying.

jgraham (approved):
This is great! Happy to land this as-is, but there are a couple of comments which might be worth looking at.

tools/manifest/sourcefile.py
Cactusmachete: Profiled them and seems like joins over tuples are the fastest of dealing with _very_ long strings. On my system, at least.

gsnedders: And they'll definitely be quicker on PyPy, given the lack of refcounting means they can't do the ugly CPython hack.

jgraham (approved):

PR #12187, opened 2018-07-25
Add support for fuzzy matching in reftests.

jgraham:
This allows fuzzy matching in reftests in which a comparison can
succeed if the images are different within a specified tolerance. It
is useful in the case of antialiasing, and in other scenarios where
it's not possible to make an exact match in all cases.

Differences between tests are characterised by two values:

* The maximum difference for any pixel on any color channel (in the
  range 0 to 255)

* The maximum total number of differing pixels

The fuzziness can be supplied in two places, according to whether it's
a property of the test or of the implementation:

* In the reftest itself, using a <meta name=fuzzy> tag

* In the expectation metadata file using a fuzzy: key that takes a
  list

The general format of the fuzziness specifier is

range = [digits, "-"], digits
fuzziness = [ url, "-" ], range, ";", range

The first range represents the maximum difference of any channel per
pixel and the second represents the total number of pixel
differences. So for example a specifier could be:

* "10;300" - meaning a difference of up to 10 per color channel and up to
300 pixels different in total (all ranges are inclusive).

* "5-10;200-300" - meaning a maximum difference of between 5 and 10
  per color channel and between 200 and 300 pixels differing in total

The definition of url is a little different between the meta element
and the expecation metadata. In the first case the url is resolved
against the current file, and applies to any reference in the current
file with that name. So for example

<meta name="fuzzy" content="option-1-ref.html:5;200">

would allow a fuzziness of up to 5 on a specific channel and up to 200
opixels different for comparisons involving the file containing the
meta element and option-1-ref.html.

In the case of expectation metadata, the metadata is always associated
with the root test, so urls are always resolved relative to that. In
the case as above where only a single URL is supplied, any reference
document with that URL will have the fuzziness applied for whatever
comparisons it's involved in e.g.

[test1.html]
  fuzzy: option-1-ref.html:5;200

would apply the fuziness to any comparison involving option-1-ref.html
whilst running the set of reftests rooted on test1.html. To specify an
exact comparison for the fuzziness, one can also supply a full
reference pair e.g.

[test1.html]
  fuzzy: subtest.html==option-1-ref.html:5;200

in which case the fuzziness would only apply to "match" comparison
involving subtest.html on the lhs and option-1-ref.html on the
rhs (both resolved relative to test1.html).

---
This change is [Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12187)

### DIFF
--- a/docs/_writing-tests/reftests.md
+++ b/docs/_writing-tests/reftests.md
@@ -112,10 +112,67 @@ only guaranteed to be after those events.
 ## Fuzzy Matching
 
 In some situations a test may have subtle differences in rendering
-compared to the reference due to, e.g., anti-aliasing. This may cause
-the test to pass on some platforms but fail on others. In this case
-some affordance for subtle discrepancies is desirable. However no
-mechanism to allow this has yet been standardized.
+compared to the reference due to, e.g., anti-aliasing. To allow for
+these small differences, we allow tests to specify a fuzziness
+characterised by two parameters, both of which must be specified:
+
+ * A maximum difference in the per-channel color value for any pixel.
+ * A number of total pixels that may be different.
+
+The maximum difference in the per pixel color value is formally
+defined as follows: let <code>T<sub>x,y,c</sub></code> be the value of
+colour channel `c` at pixel coordinates `x`, `y` in the test image and
+<code>R<sub>x,y,c</sub></code> be the corresponding value in the
+reference image, and let <code>width</code> and <code>height</code> be
+the dimensions of the image in pixels. Then <code>maxDifference =
+max<sub>x=[0,width) y=[0,height), c={r,g,b}</sub>(|T<sub>x,y,c</sub> -
+R<sub>x,y,c</sub>|)</code>.
+
+To specify the fuzziness in the test file one may add a `<meta
+name=fuzzy>` element (or, in the case of more complex tests, to any
+page containing the `<link rel=[mis]match>` elements). In the simplest
+case this has a `content` attribute containing the parameters above,
+separated by a colon e.g.
+
+```
+<meta name=fuzzy content="maxDifference=15;totalPixels=300">
+```
+
+would allow for a  difference of exactly 15 / 255 on any color channel
+and 300 exactly pixels total difference. The argument names are optional
+and may be elided; the above is the same as:
+
+```
+<meta name=fuzzy content="15;300">
+```
+
+The values may also be given as ranges e.g.
+
+```
+<meta name=fuzzy content="maxDifference=10-15;totalPixels=200-300">
+```
+
+or
+
+```
+<meta name=fuzzy content="10-15;200-300">
+```
+
+In this case the maximum pixel difference must be in the range
+`10-15` and the total number of different pixels must be in the range
+`200-300`.
+
+In cases where a single test has multiple possible refs and the
+fuzziness is not the same for all refs, a ref may be specified by
+prefixing the `content` value with the relative url for the ref e.g.
+
+```
+<meta name=fuzzy content="option1-ref.html:10-15;200-300">
+```
+
+One meta element is required per reference requiring a unique
+fuzziness value, but any unprefixed value will automatically be
+applied to any ref that doesn't have a more specific value.
 
 ## Limitations
 
--- a/tools/wptrunner/wptrunner/executors/base.py
+++ b/tools/wptrunner/wptrunner/executors/base.py
@@ -1,12 +1,15 @@
 import base64
 import hashlib
 import httplib
+import io
 import os
 import threading
 import traceback
 import socket
 import urlparse
 from abc import ABCMeta, abstractmethod
+
+from PIL import Image, ImageChops, ImageStat
 
 from ..testrunner import Stop
 from protocol import Protocol, BaseProtocolPart
@@ -286,8 +289,7 @@ class RefTestImplementation(object):
 
             screenshot = data
             hash_value = hash_screenshot(data)
-
-            self.screenshot_cache[key] = (hash_value, None)
+            self.screenshot_cache[key] = (hash_value, screenshot)
 
             rv = (hash_value, screenshot)
         else:
@@ -299,11 +301,32 @@ class RefTestImplementation(object):
     def reset(self):
         self.screenshot_cache.clear()
 
-    def is_pass(self, lhs_hash, rhs_hash, relation):
+    def is_pass(self, hashes, screenshots, relation, fuzzy):
         assert relation in ("==", "!=")
-        self.message.append("Testing %s %s %s" % (lhs_hash, relation, rhs_hash))
-        return ((relation == "==" and lhs_hash == rhs_hash) or
-                (relation == "!=" and lhs_hash != rhs_hash))
+        if not fuzzy or fuzzy == ((0,0), (0,0)):
+            equal = hashes[0] == hashes[1]
+        else:
+            max_per_channel, pixels_different = self.get_differences(screenshots)
+            allowed_per_channel, allowed_different = fuzzy
+            self.logger.info("Allowed %s pixels different, maximum difference per channel %s" %
+                             ("-".join(str(item) for item in allowed_different),
+                              "-".join(str(item) for item in allowed_per_channel)))
+            equal = (allowed_per_channel[0] <= max_per_channel <= allowed_per_channel[1] and
+                     allowed_different[0] <= pixels_different <= allowed_different[1])
+        return equal if relation == "==" else not equal
+
+    def get_differences(self, screenshots):
+        lhs = Image.open(io.BytesIO(base64.b64decode(screenshots[0]))).convert("RGB")
+        rhs = Image.open(io.BytesIO(base64.b64decode(screenshots[1]))).convert("RGB")
+        diff = ImageChops.difference(lhs, rhs)
+        minimal_diff = diff.crop(diff.getbbox())
+        mask = minimal_diff.convert("L", dither=None)
+        stat = ImageStat.Stat(minimal_diff, mask)
+        per_channel = max(item[1] for item in stat.extrema)
+        count = stat.count[0]
+        self.logger.info("Found %s pixels different, maximum difference per channel %s" %
+                         (count, per_channel))
+        return per_channel, count
 
     def run_test(self, test):
         viewport_size = test.viewport_size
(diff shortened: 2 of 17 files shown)


jgraham:
Seems to work, but missing testing etc. The PR is to get feedback on the general approach.

dbaron:
For what it's worth, I think the original design of fuzziness in Gecko was broken, because it was done in a way that never produced "unexpected pass" results to cause removal of no-longer-needed annotations.  I started to fix this by [introducing ranges](https://bugzilla.mozilla.org/show_bug.cgi?id=1252361https://bugzilla.mozilla.org/show_bug.cgi?id=1252361) a while ago, but many of the annotations haven't been fixed yet.

A concrete change that I would have done differently from the current Gecko design is that I would prefer that a number without an explicit range (say "10") mean the same as "10-10" rather than meaning the same as "0-10".  In the Gecko reftest harness I had to do the latter, at least temporarily (although I'd like to change to the former after changing all the annotations), but I think the former (meaning "10-10") is a better default since it encourages annotations that are more likely to produce unexpected-pass results when the problem is fixed.

jgraham:
@gsnedders you wanted to be mentioned here

jgraham:
@dbaron What's the primary situation in which expected fuzziness is reduced, browser-side changes or test-side changes?

The reason I ask is that automatic sync typically requires us to update metadata in a conservative way i.e. if we have an expectation of `8-10` and get `6` we will probably update the metadata to be `6-10` (or maybe `5-10` or something), because we don't really have any way to automatiically differentiate between "this test changed in a way that makes the difference consistently smaller" and "this test changed in a way that increases the variance in possible results" and assuming the former results in frequent backouts and manual work and assuming the latter doesn't.

The proposal seems reasonable either way and it's not difficult to change of course, but it may not entirely solve the problem.

jgraham:
Actually I guess it's an open question how we will deal with out-of-range fuzziness; maybe we will just mark the test as expected fail in that case any rely on a human to notice that actually the amount of fuzz decreased. But in that case the tradeoff is more clear; assuming `10` means `0-10` would mean that the sync would not set a bad test status on `8` but assuming it meant `10-10` would cause it to make the test `EXPECTED-FAIL`

tools/wptrunner/wptrunner/executors/base.py
```diff
@@ -267,11 +271,31 @@ def get_hash(self, test, viewport_size, dpi):
...
+
+    def get_differences(self, screenshots):
+        diff = ImageChops.difference(Image.open(io.BytesIO(screenshots[0])),
+                                     Image.open(io.BytesIO(screenshots[1])))
+        minimal_diff = diff.crop(diff.getbbox())
+        mask = minimal_diff.convert("L", dither=None)
+        stat = ImageStat.Stat(minimal_diff, mask)
+        per_channel = max(item[1] for item in stat.extrema)
+        count = stat.count[0]
+        self.logger.info("Found %s pixels different, mzximum difference per channel %s" %
```
gsnedders: maximum

docs/_writing-tests/reftests.md
```diff
@@ -112,10 +112,47 @@ only guaranteed to be after those events.
...
+compared to the reference due to, e.g., anti-aliasing. To allow for
+these small differences, we allow tests to specify a fuzziness
+characterised by two parameters:
+
+ * A maximum difference in the per-channel color value for any pixel.
+ * A maximum number of total pixels that may be different
+
+To specify the fuzziness in the test file one may add a `<meta
+name=fuzzy>` element (or, in the case of more complex tests, to any
+page containing the `<href rel=[mis]match>` elements). In the simplest
```
gsnedders: So if there are multiple `<link rel=match>` elements, what happens? Is the fuzziness applied to all alternations? That's another change from Gecko, right?

(also `<link`, not `<href`)

docs/_writing-tests/reftests.md
```diff
@@ -112,10 +112,47 @@ only guaranteed to be after those events.
...
+case this has a `content` attribute containing the parameters above,
+separated by a colon e.g.
+
+
```
gsnedders: Too much whitespace

docs/_writing-tests/reftests.md
gsnedders: Oh this is covered below.

dbaron:
> @dbaron What's the primary situation in which expected fuzziness is reduced, browser-side changes or test-side changes?

My concern is browser-side changes that fix the bugs covered by annotations.

I'd also note that part of the underlying problem here is that fuzzy matching is used for two different things:
* tests that are not actually expected to match perfectly
* browser bugs where the test only "fails just a little bit", and you want the test harness to ensure that it doesn't get worse

So to put it another way, my concern is about the second category:  I want browser fixes to yield "unexpected pass" results so that we can remove the "only fails a little bit" annotations when the bugs are fixed, and so that we don't then regress back within that tolerance.

jgraham:
I changed this to treat non-ranges as exact matches rather than 0-value.

Hexcles:
cc @foolip FWIW, I'm looking into the general approach here and see how Blink infra could adapt to it.

Hexcles:
I carefully read through the explanation (not the implementation yet) and it looks good to me (apart from one question; see below). In Blink, we will probably need to implement this in our image differ, which isn't too hard. (Of course, there's also the long-term possibility/discussion of using (more pieces from) wptrunner).

The question: why do we have a *range* of *maximum* allowed differences? What does the lower bound of a maximum value mean?

Hexcles:
And I'd like to introduce Ned, @nedn . He's on the Chrome Core Automation team and will also be working on Blink layout test runner. Ned, this is the fuzzy matching mechanism for reftests currently being implemented in upstream WPT.

jgraham:
Or greater than the upper bound, yes.

Hexcles:
> Or greater than the upper bound, yes.

This is intuitive, but the lower bound isn't IMHO. Would it be better to use mismatch & an upper bound instead?

jgraham:
So I think this is why the lower bound originally defaulted to 0. But per the feedback from dbaron above it's useful to have a lower bound so that you can catch the case where the test requires less fuzz than you thought so that you can update the condition to match the change in implemenatation (and therefore have more chance of catching unexpected deviations).

jgraham:
@Hexcles Are you intending to do a more detailed review here?

Hexcles (commented):
@jgraham I skimmed through the PR, and specifically looked at `tools/manifest/sourcefile.py` and `tools/wptrunner/wptrunner/executors/base.py`, both of which look good to me (though some tests would be great). I'm not familiar with the WPT expectation at all so I didn't review that.

jgraham:
This now passes CI and has some somewhat more reasonable tests. In my mind the biggest open question is whether the syntax is OK or if it's too hard to understand. And if the latter, what it ought to look like. In any case it would be very helpful to get this landed soon since we know this helps with a major issue in result correctness.

foolip:
@Hexcles, will you take another review pass?

jgraham:
I made the syntax a bit more friendly by allowing the arguments to be named.

Hexcles (commented):
A bit confused by the updated doc

docs/_writing-tests/reftests.md
```diff
@@ -112,10 +112,52 @@ only guaranteed to be after those events.
...
+<meta name=fuzzy content="15;300">
+```
+
+The values may also be given as ranges e.g.
+
+```
+<meta name=fuzzy content="maxDifference=10-15;totalPixels=200-300">
+```
+
+In this case the maximum pixel difference must be in the range
```
Hexcles: Also, "the maximum pixel difference" could also use some clarification. Based on our discussion, I think you mean the difference of the colour value of the channel with the largest difference for a pixel. I really struggle to concisely define this without ambiguity, so perhaps it's worthwhile to formally define the terminology at the top of the document, e.g.

Actual(p, c) and Expected(p, c) are the actual and expected colour values of channel c at pixel p. D(p) = max{|Actual(p, c) - Expected(p, c)| for c in (R, G, B)}.

docs/_writing-tests/reftests.md
```diff
@@ -112,10 +112,52 @@ only guaranteed to be after those events.
...
+page containing the `<link rel=[mis]match>` elements). In the simplest
+case this has a `content` attribute containing the parameters above,
+separated by a colon e.g.
+
+```
+<meta name=fuzzy content="maxDifference=15;totalPixels=300">
+```
+
+would allow for a  difference of exactly 15 / 255 on any color channel
+and 300 exactly pixels total difference. The argument names are optional
```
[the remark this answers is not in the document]
jgraham: It's exact. That was the change requested by dbaron based on experience with gecko reftests. I apparently forgot to update the commit message. `maxDifference` is a little confusing, but it is a maximum (it's the maximum difference between colour channel values in any pixel i.e. `maxDifference = max(abs(img1[x,y][channel] - img2[x,y][channel] for x in range(img1.width) for y in range(img1.height) for channel in ('r', 'g', 'b'))`. We then specify a range of possible values that maximum can have. If you have a better name please let me know.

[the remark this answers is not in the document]
Hexcles: Yeah the name is a bit confusing and I can't come up with a better one at the moment.

That said, personally I'm quite happy with the formal definition you just added, so I'm going to say the syntax looks good to me.

jgraham:
So, does anyone want to review the actual code changes? :)

foolip:

gsnedders:
To note, the Marionette internal reftest runner needs https://bugzilla.mozilla.org/show_bug.cgi?id=1500158 for this, which is riding the trains for Firefox 65 (which should ship this month).

docs/_writing-tests/reftests.md
```diff
@@ -112,10 +112,67 @@ only guaranteed to be after those events.
...
 In some situations a test may have subtle differences in rendering
-compared to the reference due to, e.g., anti-aliasing. This may cause
-the test to pass on some platforms but fail on others. In this case
-some affordance for subtle discrepancies is desirable. However no
-mechanism to allow this has yet been standardized.
+compared to the reference due to, e.g., anti-aliasing. To allow for
+these small differences, we allow tests to specify a fuzziness
+characterised by two parameters:
+
+ * A maximum difference in the per-channel color value for any pixel.
```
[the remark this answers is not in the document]
jgraham: It's a bullet list with two items. I don't think you need and/or at the end.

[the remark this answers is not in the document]
gsnedders: The rest of the documentation doesn't seem to really handle the "or" case? Like, the argument names are optional which implies I can do "200", right? What does that mean? Heck, what does "maxDifference=15" mean? (15 / 255 on any colour channel with 0 pixels differing?)

gsnedders (approved):

Issue #7693, opened 2017-10-12
Full-duplex streaming


annevk:
HTTP, both 1 and 2, supports full-duplex in principle whereby the response arrives before the request is fully transmitted, as discussed in https://github.com/whatwg/fetch/issues/229 and "implemented" in the Fetch Standard.

However, the testing infrastructure does not allow for testing this setup. This is problematic as it means browser support may be/remain spotty.

In that issue @Lukasa suggests hyper-h2 might be able to help with this (presumably only for HTTP/2).

This was filed as https://github.com/w3c/wptserve/issues/80 before by @louiscryan, but I'm rather wary of that repository still being tracked as a source of suggestions.

jakearchibald:
The latest Node.js seems to support duplex.

Here's a little echo server: https://gist.github.com/jakearchibald/d21b5b070d265efa3b16c6a566ae33fa

JKereliuk:
Are there existing manual tests that need this?

annevk:
@JKereliuk manual tests have the same limitation with this missing...

PR #12193, opened 2018-07-26
[WIP] Extending the capabilities of the HTTP/2.0 server

dheiberg:
Setting up this PR now to get some feedback on this. I have extended the functionality of the the H2 server by improving the multithreading aspect mostly. To summarise, when the `Http2WebTestRequestHandler` receives a frame on a new stream, it creates a thread for that stream specifically and uses a `Queue` to pass frames from the h2 handler to the individual stream handlers.  

In the stream threads, when it receives a `RequestReceived` frame it creates a `H2Request` straight away along with a `H2Response` object and begins handling those in a new thread. This is because at this point there is enough information to begin parsing and checking for invalid requests, mostly giving the ability to check for invalid routes without needing to wait for Data frames (#7693). In the instance of POST requests where Data frames are expected, I set up an `os.pipe()` pair of file-like objects, and use this to write the data received in the stream handler to the `rfile` that the request handler will read POST data from. Reason for using `os.pipe()` is that it ended up being the most appropriate file-like object for the scenario, as other ones like `BytesIO` didn't provide the blocking and EOF characteristics in a way that allows me to leave the request handler side untouched.

The main reason for this PR is to get some feedback about what the API should look like. In theory (tests to come), someone should currently be able to write a test where a handler can create frames using the H2 Connection object and write them. I want to add something where the handler could create bogus frames (HPACK encoding might make this tricky for headers) and write those. Please let me know what you think this might look like, and any other functionality that should be exposed.

TL;DR:
* Made multithreading more robust, and made it so each stream gets its
own thread

* Requests start getting parsed immediately and do not wait for the
entire request to get there. This helps with 404 detection early on and
other use cases.

* Created H2Request object

* Implemented basic Push Promise functionality

---
This change is [Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)


jgraham (changes requested):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: all files reviewed, 8 unresolved discussions (waiting on @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/response.py, line 445 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMcSqlBf18a1v7JunF:-LIMcSqlBf18a1v7JunG:bev8i1u) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/response.py#L445)):*
> ```Python
>     def write_push(self, promise_headers, response_headers, status, data=None, stream_id=None):
>         with self.h2conn as connection:
>             connection.push_stream(self.request.h2_stream_id, stream_id, promise_headers)
> ```

Does it make sense that `stream_id` can be `None` here?
___
*[tools/wptserve/wptserve/response.py, line 448 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMcG7F4krV4EGiDJu0:-LIMcG7H5SkfsUJMJmaU:b-dhiyng) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/response.py#L448)):*
> ```Python
>             self.write(connection)
> 
>         has_data = data is not None
> ```

Does it make sense to automatically send the actual response here? 
___
*[tools/wptserve/wptserve/server.py, line 15 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMcciY40unpf7nKi4v:-LIMcciY40unpf7nKi4w:b-k4kg7v) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/server.py#L15)):*
> ```Python
> from collections import OrderedDict
> 
> if sys.version[0] == '2':
> ```

You can use `six.moves` here instead of the explicit conditional.
___
*[tools/wptserve/wptserve/server.py, line 361 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMdcv2EHXGT9VtZP4M:-LIMdcv2EHXGT9VtZP4N:b-u85k0f) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/server.py#L361)):*
> ```Python
> 
>                 for frame in frames:
>                     if isinstance(frame, (RequestReceived, DataReceived, StreamEnded, StreamReset)):
> ```

Do we actually just want to special case `ConnectionTerminated` and otherwise take this branch as an `else`?
___
*[tools/wptserve/wptserve/server.py, line 364 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMi-5A30d3y3WNmJJ-:-LIMi-5CEjvByE_w3qQ_:b-fb86da) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/server.py#L364)):*
> ```Python
>                         if frame.stream_id not in self.stream_queues:
>                             self.stream_queues[frame.stream_id] = Queue()
>                             self.start_stream_thread(frame)
> ```

Pass the `Queue` in here and maybe the stream id.
___
*[tools/wptserve/wptserve/server.py, line 407 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMjULDD4eFucMo_TpA:-LIMjULEC1L7rJvvzgRL:b-6c46q9) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/server.py#L407)):*
> ```Python
>             if isinstance(frame, RequestReceived):
>                 # Create a shallow copy of this instance, to avoid issues with multiple threads editing the same object
>                 self_copy = copy(self)
> ```

So I'm not super-happy about this. It might be better to have a special method that returns an object we can pass into this thread with just the relevant data. OR create the Request and Response objects in the parent thread and pass it in
___
*[tools/wptserve/wptserve/server.py, line 411 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMfVNv1wo3aj9XTFQs:-LIMfVNx4g6FDktAJhbf:b1brkwm) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/server.py#L411)):*
> ```Python
> 
>                 # If its a POST request, we expect data frames and open the pipe to the request object
>                 if self_copy.command == 'POST':
> ```

So I don't really like the idea of specialising on POST here. Surely we can just handle the data frame when we actually get it rather than making asumptions based on the method.
___
*[tools/wptserve/wptserve/server.py, line 423 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMpcky27puG88ibGZt:-LIMpckzET5bKY9YFzp6:bxtrp2z) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/server.py#L423)):*
> ```Python
>                 # Begin processing the response in another thread,
>                 # and continue listening for more data here in this one
>                 t = threading.Thread(
> ```

So I'm not clear why this happens in a separate thread. Can't it just be in the same thread as the request handler for this stream?

dheiberg (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: all files reviewed, 8 unresolved discussions (waiting on @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/response.py, line 445 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMcSqlBf18a1v7JunF:-LIMukKu7xHj-gm_2LJ1:bc4cyej) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/response.py#L445)):*

Does it make sense that `stream_id` can be `None` here?
</blockquote></details>

Nope, leftover from a previous iteration, will remove
___
*[tools/wptserve/wptserve/response.py, line 448 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMcG7F4krV4EGiDJu0:-LIMupMZ1vdtAscFz_Iu:bduem0c) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/response.py#L448)):*

Does it make sense to automatically send the actual response here? 
</blockquote></details>

So I think there should be a default behaviour, which may be to send the push promise along with the actual content, but I agree that it should also be a possibility to delay and choose when to actually send the data
___
*[tools/wptserve/wptserve/server.py, line 15 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMcciY40unpf7nKi4v:-LIMv95nC3EASzSNV7nu:buljm7i) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/server.py#L15)):*

You can use `six.moves` here instead of the explicit conditional.
</blockquote></details>

Good to know, will look into that
___
*[tools/wptserve/wptserve/server.py, line 361 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMdcv2EHXGT9VtZP4M:-LIMy-ob3y5tBLR-RYLl:bx1tr7k) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/server.py#L361)):*

Do we actually just want to special case `ConnectionTerminated` and otherwise take this branch as an `else`?
</blockquote></details>

That makes much more sense, it was like this due to previously having multiple branches at this level, now those sit in the stream handler
___
*[tools/wptserve/wptserve/server.py, line 364 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMi-5A30d3y3WNmJJ-:-LIMy8Jl0u426uV5wmij:bne44k4) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/server.py#L364)):*

Pass the `Queue` in here and maybe the stream id.
</blockquote></details>

Is there a need to pass the stream id if it is in the frame?
___
*[tools/wptserve/wptserve/server.py, line 407 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMjULDD4eFucMo_TpA:-LIMvOjA6q964Ot57SzZ:biqg9fm) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/server.py#L407)):*

So I'm not super-happy about this. It might be better to have a special method that returns an object we can pass into this thread with just the relevant data. OR create the Request and Response objects in the parent thread and pass it in
</blockquote></details>

I agree this is super messy and not at all ideal, was just something to get it working quickly and come back to. I like the idea of having a method that creates a pseudo handler to pass along.
___
*[tools/wptserve/wptserve/server.py, line 411 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMfVNv1wo3aj9XTFQs:-LIMveYyAfjb0PMiosAb:brtrnf9) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/server.py#L411)):*

So I don't really like the idea of specialising on POST here. Surely we can just handle the data frame when we actually get it rather than making asumptions based on the method.
</blockquote></details>

I suppose it is unnecessary, I guess I was just trying to avoid creating a pipe unless it was needed, but either way works.
___
*[tools/wptserve/wptserve/server.py, line 423 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMpcky27puG88ibGZt:-LIMwStw8c_wKtnWfCUq:b-xf1bcu) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/server.py#L423)):*

So I'm not clear why this happens in a separate thread. Can't it just be in the same thread as the request handler for this stream? 
</blockquote></details>

Its mainly to avoid having to alter other bits of code further on. If it were to happen in the same thread, you would need to check the queue again for Data frames (in the scenario of a POST request) somewhere down the line. Here, the stream thread is in charge of checking for frames in the queue and forwards them to the request using the pipe, allowing the request to function without needed special checks for H2.

jgraham (changes requested):
> Reviewed 59 of 63 files at r2, 5 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: 64 of 67 files reviewed, 7 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/handlers.py, line 250 at r3](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJJgP7S0BU1BziBdkdF:-LJJgP7T6nxQQlgO-nr8:bzaslfy) ([raw file](https://github.com/web-platform-tests/wpt/blob/d9c89b745292c5d455c499d97821816afb9f790c/tools/wptserve/wptserve/handlers.py#L250)):*
> ```Python
>                 wrap_pipeline(path, request, response)
>             elif "h2main" in environ:
>                 handler = environ["h2main"](H2ResponseHandler)
> ```

I'm not entirely sure why this is being called with a `H2ResponseHandler`class as the argument.
___
*[tools/wptserve/wptserve/request.py, line 355 at r3](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJJfOIVDrR5-Z3HCEQm:-LJJfOIVDrR5-Z3HCEQn:bvhh7oy) ([raw file](https://github.com/web-platform-tests/wpt/blob/d9c89b745292c5d455c499d97821816afb9f790c/tools/wptserve/wptserve/request.py#L355)):*
> ```Python
>     def __init__(self, request_handler):
>         self.h2_stream_id = request_handler.h2_stream_id
>         self.frames = Queue()
> ```

Making the frames here a queue rather than a list seems a bit unfortunate since it's harder to consider multiple frames at once. 

I think my imagined approach here is that the request handling thread in server.py introspects the handler somehow to figure out if it's using the request/response model or the multi-frame model. In the first case it just builds up the request object and at the end of the request, hands over a complete request to the hander. In the latter case it updates the request per-frame and then directly calls the per-frame methods on the handler
___
*[tools/wptserve/wptserve/server.py, line 423 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMpcky27puG88ibGZt:-LJJaMSC3RPxFHYBeiOR:b-6n0dqw) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/server.py#L423)):*

Its mainly to avoid having to alter other bits of code further on. If it were to happen in the same thread, you would need to check the queue again for Data frames (in the scenario of a POST request) somewhere down the line. Here, the stream thread is in charge of checking for frames in the queue and forwards them to the request using the pipe, allowing the request to function without needed special checks for H2.
</blockquote></details>

I'm not sure I entirely follow, and I'm a little concerned that we may end up leaking this thread, although I'm not entirely sure. It seems like on the first request on a stream, we are starting a thread that will create and run the handler, and attempt to write the response. But that is happening before handling any later frames (or at least races with that), so presumably something is stopping `finish_handling` from actually finishing handling the request.

It *seems* like what we want to do is break out part of the logic from `finish_handling` into some kind of `get_handler` method which gives us a handler in this thread, which we can use as required in response to later frames, rather than appearing to go through a request-response cycle in yet another thread. What am I missing?
___
*[tools/wptserve/wptserve/server.py, line 427 at r3](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJJf-1AA27Mx6aHpAwH:-LJJf-1AA27Mx6aHpAwI:b-9827sn) ([raw file](https://github.com/web-platform-tests/wpt/blob/d9c89b745292c5d455c499d97821816afb9f790c/tools/wptserve/wptserve/server.py#L427)):*
> ```Python
>             elif isinstance(frame, (StreamReset, ConnectionTerminated)):
>                 # assert self.stream_queues[stream_id].empty()
>                 del self.stream_queues[stream_id]
> ```

I think the parent thread can/should do this just for the simplicity of only accessing the object in the main thread.
___
*[tools/wptserve/wptserve/utils.py, line 116 at r3](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJJTyBc5yNyLfdNaR-w:-LJJTyBeA9gYh66PvqBg:b-llu50z) ([raw file](https://github.com/web-platform-tests/wpt/blob/d9c89b745292c5d455c499d97821816afb9f790c/tools/wptserve/wptserve/utils.py#L116)):*
> ```Python
>     import ssl
>     ssl_v = ssl.OPENSSL_VERSION_INFO
>     return (sys.version_info[0] == 2 and sys.version_info[1] == 7 and sys.version_info[2] >= 10) \
> ```

Don't need a continuation line when you are using parens.

dheiberg (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: 60 of 69 files reviewed, 7 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/handlers.py, line 250 at r3](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJJgP7S0BU1BziBdkdF:-LJKQKzmBCE74yD7qkG1:b-rtb2kk) ([raw file](https://github.com/web-platform-tests/wpt/blob/d9c89b745292c5d455c499d97821816afb9f790c/tools/wptserve/wptserve/handlers.py#L250)):*

I'm not entirely sure why this is being called with a `H2ResponseHandler`class as the argument.
</blockquote></details>

This has now been simplified. In this version, the way of declaring h2 custom handlers was overly complex. I have addressed this alongside the changes proposed
___
*[tools/wptserve/wptserve/request.py, line 355 at r3](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJJfOIVDrR5-Z3HCEQm:-LJKR2KYE6eMT1xiZR8K:bqvsbce) ([raw file](https://github.com/web-platform-tests/wpt/blob/d9c89b745292c5d455c499d97821816afb9f790c/tools/wptserve/wptserve/request.py#L355)):*

Making the frames here a queue rather than a list seems a bit unfortunate since it's harder to consider multiple frames at once. 

I think my imagined approach here is that the request handling thread in server.py introspects the handler somehow to figure out if it's using the request/response model or the multi-frame model. In the first case it just builds up the request object and at the end of the request, hands over a complete request to the hander. In the latter case it updates the request per-frame and then directly calls the per-frame methods on the handler
</blockquote></details>

I agree with this way of doing it, I have pushed a WIP commit to show these new changes
___
*[tools/wptserve/wptserve/server.py, line 423 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LIMpcky27puG88ibGZt:-LJKRH8a9DExmwW0eNCm:b-6g29a) ([raw file](https://github.com/web-platform-tests/wpt/blob/f2654f3642d291a6de3f80bf4ff37062b9722ca4/tools/wptserve/wptserve/server.py#L423)):*

I'm not sure I entirely follow, and I'm a little concerned that we may end up leaking this thread, although I'm not entirely sure. It seems like on the first request on a stream, we are starting a thread that will create and run the handler, and attempt to write the response. But that is happening before handling any later frames (or at least races with that), so presumably something is stopping `finish_handling` from actually finishing handling the request.

It *seems* like what we want to do is break out part of the logic from `finish_handling` into some kind of `get_handler` method which gives us a handler in this thread, which we can use as required in response to later frames, rather than appearing to go through a request-response cycle in yet another thread. What am I missing?
</blockquote></details>

Indeed, this setup was overly complex, and was originally implemented like this to maintain backward compatability as well as bring some H2 optimisations to the existing infrastructure. I have now simplified this and removed the extra thread that added the complications. 
___
*[tools/wptserve/wptserve/server.py, line 427 at r3](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJJf-1AA27Mx6aHpAwH:-LJKPrPO3TFiKUi9Zw6q:b-i0mkg9) ([raw file](https://github.com/web-platform-tests/wpt/blob/d9c89b745292c5d455c499d97821816afb9f790c/tools/wptserve/wptserve/server.py#L427)):*

I think the parent thread can/should do this just for the simplicity of only accessing the object in the main thread.
</blockquote></details>

Does it not make sense for the child thread to do this, since otherwise the parent thread would send the connection terminated frame, and might delete the queue before the child thread processes it, unless I make the parent thread sit and wait for the queue to empty/join?
___
*[tools/wptserve/wptserve/utils.py, line 116 at r3](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJJTyBc5yNyLfdNaR-w:-LJKPW2wDC2ex6qh1E5A:b-fd7s0j) ([raw file](https://github.com/web-platform-tests/wpt/blob/d9c89b745292c5d455c499d97821816afb9f790c/tools/wptserve/wptserve/utils.py#L116)):*

Don't need a continuation line when you are using parens.
</blockquote></details>

I believe I do since the continuation is not actually within a set of brackets. My editor yells at me otherwise

jgraham (changes requested):
> Reviewed 2 of 7 files at r4, 3 of 9 files at r5, 1 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: 65 of 71 files reviewed, 5 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/handlers.py, line 235 at r5](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJTjy2o0XtSeM8ePUs2:-LJTjy2p9qghOL9RTZLX:b-789jft) ([raw file](https://github.com/web-platform-tests/wpt/blob/59ee4f103b27e5383edac0548a61f0ccb76e06a7/tools/wptserve/wptserve/handlers.py#L235)):*
> ```Python
>         return "<%s base_path:%s url_base:%s>" % (self.__class__.__name__, self.base_path, self.url_base)
> 
>     def read_file_and_run(self, request, response, func):
> ```

 I think this doesn't work, because at the time we call the handler we've reset sys.path and sys.modules.
___
*[tools/wptserve/wptserve/handlers.py, line 266 at r5](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJTkNFKB6QLSR3v68Aa:-LJTkNFM6VtHP3TG3mc9:bxk3sd) ([raw file](https://github.com/web-platform-tests/wpt/blob/59ee4f103b27e5383edac0548a61f0ccb76e06a7/tools/wptserve/wptserve/handlers.py#L266)):*
> ```Python
> 
> 
>     def generate_handler(self, request):
> ```

So, I was thinking that we could make this look like

```
class PythonScripHandler(object):
    def __init__(self, base_path, url_base):
        self.handler_obj = None

   def __call__(self, request, response):
        # This finlizes the response
        if self.handler_obj is None:
            self.handle_setup(request, response)
        with self.handler_env:
            self.handler_obj(request, resp)

   def handle_setup(self, request, resp):
       # Load the .py file, and look for a function or class called main or Handler
      # Store this in self.handler_obj

   def handle_headers(self, request, resp):
       if hasattr(self.handler_obj, "handle_data"):
           with self.handler_env:
               self.handler_obj.handle_headers(request, response)
[…]
```

Where `handler_env` is a context manager that sets sys.path and handles the possibility of an `IOError` (although it occurs to me now that this sys.path manipulation is very non-threadsafe).

What do you think?
___
*[tools/wptserve/wptserve/server.py, line 427 at r3](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJJf-1AA27Mx6aHpAwH:-LJTLONT7cREQxRUMaRr:bls1m3l) ([raw file](https://github.com/web-platform-tests/wpt/blob/d9c89b745292c5d455c499d97821816afb9f790c/tools/wptserve/wptserve/server.py#L427)):*

Does it not make sense for the child thread to do this, since otherwise the parent thread would send the connection terminated frame, and might delete the queue before the child thread processes it, unless I make the parent thread sit and wait for the queue to empty/join?
</blockquote></details>

Oh right, so I was also thinking that it would just store the queue as a local variable in the thread rather than accessing it through the dict, so it can't be deleted by the parent.
___
*[tools/wptserve/wptserve/server.py, line 418 at r5](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJTKa9T9p_kE423ZRot:-LJTKa9VBHETAhgda1bF:b-rf7z7) ([raw file](https://github.com/web-platform-tests/wpt/blob/59ee4f103b27e5383edac0548a61f0ccb76e06a7/tools/wptserve/wptserve/server.py#L418)):*
> ```Python
>                 req_handler = stream_handler.server.router.get_handler(request)
> 
>                 if isinstance(req_handler, PythonScriptHandler):
> ```

Instead of using `isinstance` here use `if hasattr(req_handler, "generate_handler")` so we're defining a protocol that any handler can implement. I'm also not suer-sold on the name and might perfer something like `init` or `create` or `handle_new_stream` or something.
___
*[tools/wptserve/wptserve/utils.py, line 116 at r3](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJJTyBc5yNyLfdNaR-w:-LJYAUrX04HEz9MTCcjj:b-egv6u9) ([raw file](https://github.com/web-platform-tests/wpt/blob/d9c89b745292c5d455c499d97821816afb9f790c/tools/wptserve/wptserve/utils.py#L116)):*

I believe I do since the continuation is not actually within a set of brackets. My editor yells at me otherwise
</blockquote></details>

Oh, then add one more set of parens :) Continuation lines are kind of terrible and can usually be avoided.

dheiberg (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: 65 of 71 files reviewed, 5 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/handlers.py, line 235 at r5](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJTjy2o0XtSeM8ePUs2:-LJYD-pW-rutbY3ayu3j:b-mi6fqd) ([raw file](https://github.com/web-platform-tests/wpt/blob/59ee4f103b27e5383edac0548a61f0ccb76e06a7/tools/wptserve/wptserve/handlers.py#L235)):*

 I think this doesn't work, because at the time we call the handler we've reset sys.path and sys.modules.
</blockquote></details>

I believe it works, because the `func` is run whilst the path is in this modified state, and the path is only reset after `func` is finished. 
___
*[tools/wptserve/wptserve/handlers.py, line 266 at r5](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJTkNFKB6QLSR3v68Aa:-LJYE02K4g9fHOy2pJtJ:b-pyvtb6) ([raw file](https://github.com/web-platform-tests/wpt/blob/59ee4f103b27e5383edac0548a61f0ccb76e06a7/tools/wptserve/wptserve/handlers.py#L266)):*

So, I was thinking that we could make this look like

```
class PythonScripHandler(object):
    def __init__(self, base_path, url_base):
        self.handler_obj = None

   def __call__(self, request, response):
        # This finlizes the response
        if self.handler_obj is None:
            self.handle_setup(request, response)
        with self.handler_env:
            self.handler_obj(request, resp)

   def handle_setup(self, request, resp):
       # Load the .py file, and look for a function or class called main or Handler
      # Store this in self.handler_obj

   def handle_headers(self, request, resp):
       if hasattr(self.handler_obj, "handle_data"):
           with self.handler_env:
               self.handler_obj.handle_headers(request, response)
[…]
```

Where `handler_env` is a context manager that sets sys.path and handles the possibility of an `IOError` (although it occurs to me now that this sys.path manipulation is very non-threadsafe).

What do you think?
</blockquote></details>

The issue I see with this is that it can cause issues when running multiple tests since PythonScriptHandler is an object shared between all tests and is a generator for FunctionHandler objects. It would feel like bad practice to turn it from a generator to now storing handler attributes for a single handler instance. For example, what does the server do if two different connections request .py resources at the same time? In your proposal, the second .py resource called would overwrite the `handle_headers` attribute, or see that there is already a handler object and use those.

To implement this, PythonScriptHandler would need to be changed to be a class that you instantiate a new object for every .py resource called.
___
*[tools/wptserve/wptserve/server.py, line 427 at r3](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJJf-1AA27Mx6aHpAwH:-LJYGYsn9GmfOv94B7xb:b-vi2sku) ([raw file](https://github.com/web-platform-tests/wpt/blob/d9c89b745292c5d455c499d97821816afb9f790c/tools/wptserve/wptserve/server.py#L427)):*

Oh right, so I was also thinking that it would just store the queue as a local variable in the thread rather than accessing it through the dict, so it can't be deleted by the parent.
</blockquote></details>

That's a very good point, I'll change it so that the stream thread has its own local variable of the Queue! 
___
*[tools/wptserve/wptserve/server.py, line 418 at r5](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJTKa9T9p_kE423ZRot:-LJYFoyH4zU6bPqUHf8m:b8onizv) ([raw file](https://github.com/web-platform-tests/wpt/blob/59ee4f103b27e5383edac0548a61f0ccb76e06a7/tools/wptserve/wptserve/server.py#L418)):*

Instead of using `isinstance` here use `if hasattr(req_handler, "generate_handler")` so we're defining a protocol that any handler can implement. I'm also not suer-sold on the name and might perfer something like `init` or `create` or `handle_new_stream` or something.
</blockquote></details>

That's a good idea! I believe the name is fairly suitable since PythonScriptHandler is a class for generating handlers, but the name can obviously be changed if need be. 
___
*[tools/wptserve/wptserve/utils.py, line 116 at r3](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJJTyBc5yNyLfdNaR-w:-LJYBWQdA2gZ8b_alHGE:b-896fix) ([raw file](https://github.com/web-platform-tests/wpt/blob/d9c89b745292c5d455c499d97821816afb9f790c/tools/wptserve/wptserve/utils.py#L116)):*

Oh, then add one more set of parens :) Continuation lines are kind of terrible and can usually be avoided.
</blockquote></details>

Done.

jgraham (changes requested):
> Reviewed 1 of 8 files at r3, 3 of 9 files at r5, 1 of 2 files at r6, 4 of 4 files at r7.
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: all files reviewed, 4 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/tests/functional/test_response.py, line 339 at r7](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJnaiYlDxUUyOb_lf2F:-LJnaiYm4eZFFse0XPVs:b17vb3a) ([raw file](https://github.com/web-platform-tests/wpt/blob/20e4e2ff41c2df0d8fa8e8453d67fa6ca9254a91/tools/wptserve/tests/functional/test_response.py#L339)):*
> ```Python
> 
>     def test_raw_header_frame(self):
> ```

Remove all the blank lines between the signature and the method body.
___
*[tools/wptserve/wptserve/handlers.py, line 277 at r7](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJnYdvE7zx4QovwTwPm:-LJnYdvF80CX4JG_ZFIJ:be9zex8) ([raw file](https://github.com/web-platform-tests/wpt/blob/20e4e2ff41c2df0d8fa8e8453d67fa6ca9254a91/tools/wptserve/wptserve/handlers.py#L277)):*
> ```Python
> 
> 
>     def convert_to_h2_handler(self, request):
> ```

Can we rename this something that doesn't specifically mention `h2` since it seems at least plasible we'll reuse this infrastructure for some other protcol in the future. `frame_handler` or something would be fine.
___
*[tools/wptserve/wptserve/handlers.py, line 294 at r7](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJn_xeFA0Bn2d0uCnGZ:-LJn_xeFA0Bn2d0uCnG_:b-ttvayy) ([raw file](https://github.com/web-platform-tests/wpt/blob/20e4e2ff41c2df0d8fa8e8453d67fa6ca9254a91/tools/wptserve/wptserve/handlers.py#L294)):*
> ```Python
>                 handler.func = environ["main"]
>             if "handle_headers" in environ:
>                 handler.handle_headers = environ["handle_headers"]
> ```

So, API-wise this means that there's no way to just return data from one of these functions and have it do something. Maybe that's OK, I'm not sure (don't change it in this PR though).
___
*[tools/wptserve/wptserve/server.py, line 427 at r3](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJJf-1AA27Mx6aHpAwH:-LJnZqnu-YiaY1ZynLhn:bo6m0r7) ([raw file](https://github.com/web-platform-tests/wpt/blob/d9c89b745292c5d455c499d97821816afb9f790c/tools/wptserve/wptserve/server.py#L427)):*

That's a very good point, I'll change it so that the stream thread has its own local variable of the Queue! 
</blockquote></details>

Are you going to move the removal from `stream_queues` to the parent?
___
*[tools/wptserve/wptserve/server.py, line 344 at r7](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJn_Eu7EO-26boVtwwX:-LJn_Eu7EO-26boVtwwY:b-s6zhs7) ([raw file](https://github.com/web-platform-tests/wpt/blob/20e4e2ff41c2df0d8fa8e8453d67fa6ca9254a91/tools/wptserve/wptserve/server.py#L344)):*
> ```Python
>         self.request.sendall(data)
> 
>         self.request_threads = []
> ```

This and `stream_queues` could be local variables, I think. I know we aren't always great at it, but reducing shared state is good!

dheiberg (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: all files reviewed, 4 unresolved discussions (waiting on @dhdavvie, @jgraham, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/tests/functional/test_response.py, line 339 at r7](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJnaiYlDxUUyOb_lf2F:-LJnkVwRBQ_fMbkm3OAa:b32w1t) ([raw file](https://github.com/web-platform-tests/wpt/blob/20e4e2ff41c2df0d8fa8e8453d67fa6ca9254a91/tools/wptserve/tests/functional/test_response.py#L339)):*

Remove all the blank lines between the signature and the method body.
</blockquote></details>

Ok
___
*[tools/wptserve/wptserve/handlers.py, line 277 at r7](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJnYdvE7zx4QovwTwPm:-LJnkcnn-ODg3Z3MQjxn:byr6ftq) ([raw file](https://github.com/web-platform-tests/wpt/blob/20e4e2ff41c2df0d8fa8e8453d67fa6ca9254a91/tools/wptserve/wptserve/handlers.py#L277)):*

Can we rename this something that doesn't specifically mention `h2` since it seems at least plasible we'll reuse this infrastructure for some other protcol in the future. `frame_handler` or something would be fine.
</blockquote></details>

Sure, making it more general would be a good idea
___
*[tools/wptserve/wptserve/handlers.py, line 294 at r7](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJn_xeFA0Bn2d0uCnGZ:-LJnkqLa3JLZB_I0S9rI:bmo14en) ([raw file](https://github.com/web-platform-tests/wpt/blob/20e4e2ff41c2df0d8fa8e8453d67fa6ca9254a91/tools/wptserve/wptserve/handlers.py#L294)):*

So, API-wise this means that there's no way to just return data from one of these functions and have it do something. Maybe that's OK, I'm not sure (don't change it in this PR though).
</blockquote></details>

I'm not sure what you mean by this, mind elaborating what you had in mind?
___
*[tools/wptserve/wptserve/server.py, line 427 at r3](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJJf-1AA27Mx6aHpAwH:-LJnqxO79xBfma2VSe2K:b2nhtan) ([raw file](https://github.com/web-platform-tests/wpt/blob/d9c89b745292c5d455c499d97821816afb9f790c/tools/wptserve/wptserve/server.py#L427)):*

Are you going to move the removal from `stream_queues` to the parent?
</blockquote></details>

Yup
___
*[tools/wptserve/wptserve/server.py, line 344 at r7](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJn_Eu7EO-26boVtwwX:-LJnl4m39UGt3KsVx27C:b-gfpky4) ([raw file](https://github.com/web-platform-tests/wpt/blob/20e4e2ff41c2df0d8fa8e8453d67fa6ca9254a91/tools/wptserve/wptserve/server.py#L344)):*

This and `stream_queues` could be local variables, I think. I know we aren't always great at it, but reducing shared state is good!
</blockquote></details>

Yup, moving the last bit of logic out of the stream thread and then making this local

jgraham (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: 67 of 71 files reviewed, 4 unresolved discussions (waiting on @jgraham, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/handlers.py, line 294 at r7](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJn_xeFA0Bn2d0uCnGZ:-LJnzr1Q7mWg0k0GC1Fs:bf44nx8) ([raw file](https://github.com/web-platform-tests/wpt/blob/20e4e2ff41c2df0d8fa8e8453d67fa6ca9254a91/tools/wptserve/wptserve/handlers.py#L294)):*

I'm not sure what you mean by this, mind elaborating what you had in mind?
</blockquote></details>

You can't write a function like

```
def handle_data(request, response):
     return "hello world"
```

and have the server return `hello world` as the response body.

dheiberg (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: 67 of 71 files reviewed, 4 unresolved discussions (waiting on @jgraham, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/handlers.py, line 294 at r7](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJn_xeFA0Bn2d0uCnGZ:-LJo-hBv8aCRFav0Dzfa:b-u5pljz) ([raw file](https://github.com/web-platform-tests/wpt/blob/20e4e2ff41c2df0d8fa8e8453d67fa6ca9254a91/tools/wptserve/wptserve/handlers.py#L294)):*

You can't write a function like

```
def handle_data(request, response):
     return "hello world"
```

and have the server return `hello world` as the response body. 
</blockquote></details>

Ahh yes, now I see what you mean. I don't see why this shouldn't be possible by using similar code to the `__call__` function in `FunctionHandler` to parse return values from the handler functions. I will look into doing this for the final PR :)

jgraham (changes requested):
> Reviewed 4 of 4 files at r8.
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: all files reviewed, 3 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/handlers.py, line 277 at r7](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJnYdvE7zx4QovwTwPm:-LJo-cs54Q1F4DUBOWeZ:b-l0ax9l) ([raw file](https://github.com/web-platform-tests/wpt/blob/20e4e2ff41c2df0d8fa8e8453d67fa6ca9254a91/tools/wptserve/wptserve/handlers.py#L277)):*

Sure, making it more general would be a good idea
</blockquote></details>

Sorry this, is somewhat irrational, but I really dislike `convert_to`. It sounds like it mutates the object itself, but it doesn't. So I really think we should drop that from the method name.
___
*[tools/wptserve/wptserve/server.py, line 370 at r8](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJo02Hm4BVvbohQMN03:-LJo02Hm4BVvbohQMN04:b4ios95) ([raw file](https://github.com/web-platform-tests/wpt/blob/38993c9cf70462239b339b6c21bfe83d6b679b03/tools/wptserve/wptserve/server.py#L370)):*
> ```Python
>                         if frame.stream_id not in stream_queues:
>                             stream_queues[frame.stream_id] = Queue()
>                             request_threads.append(self.start_stream_thread(frame, stream_queues[frame.stream_id]))
> ```

This list leaks, right? That's maybe not a huge problem since the number of threads per connection might be limited, but we should perhaps fix it anyway, maybe by having a single data structure like `{stream_id: (thread, Queue)}`

dheiberg (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: all files reviewed, 2 unresolved discussions (waiting on @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/handlers.py, line 277 at r7](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJnYdvE7zx4QovwTwPm:-LJo1Ke97UmsGikJhw0J:b-1huowq) ([raw file](https://github.com/web-platform-tests/wpt/blob/20e4e2ff41c2df0d8fa8e8453d67fa6ca9254a91/tools/wptserve/wptserve/handlers.py#L277)):*

Sorry this, is somewhat irrational, but I really dislike `convert_to`. It sounds like it mutates the object itself, but it doesn't. So I really think we should drop that from the method name.
</blockquote></details>

Fair enough, the implication is definitely there. should the function just be `frame_handler(.....)` then?
___
*[tools/wptserve/wptserve/server.py, line 370 at r8](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJo02Hm4BVvbohQMN03:-LJo1WzdE8u2zQjne-4j:b9f4jzt) ([raw file](https://github.com/web-platform-tests/wpt/blob/38993c9cf70462239b339b6c21bfe83d6b679b03/tools/wptserve/wptserve/server.py#L370)):*

This list leaks, right? That's maybe not a huge problem since the number of threads per connection might be limited, but we should perhaps fix it anyway, maybe by having a single data structure like `{stream_id: (thread, Queue)}`
</blockquote></details>

Yes, I realised that this does leak, but put it on the backburner to fix in the next PR. Want me to fix it in this one instead?

jgraham (changes requested):
> Reviewed 2 of 2 files at r9.
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: all files reviewed, 3 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/server.py, line 365 at r9](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJsYKt-7cn8b4MlQE81:-LJsYKt-7cn8b4MlQE82:b-lwhef1) ([raw file](https://github.com/web-platform-tests/wpt/blob/98ce16f20df3d1d252f857026d80f96cbee53c14/tools/wptserve/wptserve/server.py#L365)):*
> ```Python
>                         for stream_id, (thread, queue) in stream_queues.items():
>                             queue.put(frame)
>                             del stream_queues[stream_id]
> ```

So, a question. Here we put a message that should cause the threads to terminate, but don't join them.
___
*[tools/wptserve/wptserve/server.py, line 379 at r9](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJsYwEy8EE6saYPxYdN:-LJsYwEy8EE6saYPxYdO:bat2qok) ([raw file](https://github.com/web-platform-tests/wpt/blob/98ce16f20df3d1d252f857026d80f96cbee53c14/tools/wptserve/wptserve/server.py#L379)):*
> ```Python
>                 self.logger.debug('(%s) ERROR - Closing Connection - \n%s' % (self.uid, str(e)))
>                 self.close_connection = True
>                 for stream_id, (thread, queue) in stream_queues:
> ```

But here we wait for the threads to exit without sending them any message. That seems wrong?

dheiberg (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: all files reviewed, 3 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/server.py, line 379 at r9](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJsYwEy8EE6saYPxYdN:-LJsb-zFDiTS00m-eS6Z:bcbi12l) ([raw file](https://github.com/web-platform-tests/wpt/blob/98ce16f20df3d1d252f857026d80f96cbee53c14/tools/wptserve/wptserve/server.py#L379)):*

But here we wait for the threads to exit without sending them any message. That seems wrong?
</blockquote></details>

So what I could do is move the join code back 2 indents, so that it will run after the while loop, which will happen in the case of ConnectionTerminated and a socket exception. I agree that there should be some communication to the threads that they should finish up. Would creating dummy ConnectionTerminated frames to trigger them to finish make sense?

jgraham (changes requested):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: all files reviewed, 3 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/server.py, line 379 at r9](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJsYwEy8EE6saYPxYdN:-LJscSll9BNrp1wmJZnB:bggansg) ([raw file](https://github.com/web-platform-tests/wpt/blob/98ce16f20df3d1d252f857026d80f96cbee53c14/tools/wptserve/wptserve/server.py#L379)):*

So what I could do is move the join code back 2 indents, so that it will run after the while loop, which will happen in the case of ConnectionTerminated and a socket exception. I agree that there should be some communication to the threads that they should finish up. Would creating dummy ConnectionTerminated frames to trigger them to finish make sense?
</blockquote></details>

So I might just make a message of `None` act like `ConnectionTerminated` in the theads, rather than trying to really create a fake message. Then put the `join` in a `finally` block outside the `while` loop.

dheiberg (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: all files reviewed, 3 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/server.py, line 379 at r9](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJsYwEy8EE6saYPxYdN:-LJsiBrOCiBM6HLTcNo8:bret1e1) ([raw file](https://github.com/web-platform-tests/wpt/blob/98ce16f20df3d1d252f857026d80f96cbee53c14/tools/wptserve/wptserve/server.py#L379)):*

So I might just make a message of `None` act like `ConnectionTerminated` in the theads, rather than trying to really create a fake message. Then put the `join` in a `finally` block outside the `while` loop.
</blockquote></details>

How does this look? It seems easier to move the `join` to outside the while loop, instead of putting a `try/finally` wrap around the `while` loop, unless I've misunderstood what you meant

```--- tools/wptserve/wptserve/server.py	(revision 98ce16f20df3d1d252f857026d80f96cbee53c14)
+++ tools/wptserve/wptserve/server.py	(date 1534254666000)
@@ -362,7 +362,6 @@
                         # Flood all the streams with connection terminated, this will cause them to stop
                         for stream_id, (thread, queue) in stream_queues.items():
                             queue.put(frame)
-                            del stream_queues[stream_id]
 
                     elif hasattr(frame, 'stream_id'):
                         if frame.stream_id not in stream_queues:
@@ -376,8 +375,11 @@
             except (socket.timeout, socket.error) as e:
                 self.logger.debug('(%s) ERROR - Closing Connection - \n%s' % (self.uid, str(e)))
                 self.close_connection = True
-                for stream_id, (thread, queue) in stream_queues:
-                    thread.join()
+                for stream_id, (thread, queue) in stream_queues.items():
+                    queue.put(None)
+
+        for stream_id, (thread, queue) in stream_queues:
+            thread.join()
 
     def start_stream_thread(self, frame, queue):
         t = threading.Thread(
@@ -433,7 +435,7 @@
 
                 if frame.stream_ended:
                     wfile.close()
-            elif isinstance(frame, (StreamReset, StreamEnded, ConnectionTerminated)):
+            elif isinstance(frame, (StreamReset, StreamEnded, ConnectionTerminated, type(None))):
                 self.logger.debug('(%s - %s) Stream Reset, Thread Closing' % (self.uid, stream_id))
                 break```
```

jgraham (changes requested):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: all files reviewed, 3 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/server.py, line 379 at r9](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJsYwEy8EE6saYPxYdN:-LJslJra2vKBtSCNxVhy:b86xe1p) ([raw file](https://github.com/web-platform-tests/wpt/blob/98ce16f20df3d1d252f857026d80f96cbee53c14/tools/wptserve/wptserve/server.py#L379)):*

How does this look? It seems easier to move the `join` to outside the while loop, instead of putting a `try/finally` wrap around the `while` loop, unless I've misunderstood what you meant

```--- tools/wptserve/wptserve/server.py	(revision 98ce16f20df3d1d252f857026d80f96cbee53c14)
+++ tools/wptserve/wptserve/server.py	(date 1534254666000)
@@ -362,7 +362,6 @@
                         # Flood all the streams with connection terminated, this will cause them to stop
                         for stream_id, (thread, queue) in stream_queues.items():
                             queue.put(frame)
-                            del stream_queues[stream_id]
 
                     elif hasattr(frame, 'stream_id'):
                         if frame.stream_id not in stream_queues:
@@ -376,8 +375,11 @@
             except (socket.timeout, socket.error) as e:
                 self.logger.debug('(%s) ERROR - Closing Connection - \n%s' % (self.uid, str(e)))
                 self.close_connection = True
-                for stream_id, (thread, queue) in stream_queues:
-                    thread.join()
+                for stream_id, (thread, queue) in stream_queues.items():
+                    queue.put(None)
+
+        for stream_id, (thread, queue) in stream_queues:
+            thread.join()
 
     def start_stream_thread(self, frame, queue):
         t = threading.Thread(
@@ -433,7 +435,7 @@
 
                 if frame.stream_ended:
                     wfile.close()
-            elif isinstance(frame, (StreamReset, StreamEnded, ConnectionTerminated)):
+            elif isinstance(frame, (StreamReset, StreamEnded, ConnectionTerminated, type(None))):
                 self.logger.debug('(%s - %s) Stream Reset, Thread Closing' % (self.uid, stream_id))
                 break```
</blockquote></details>

I would just make the condition `frame is None or isinstance(…)`. 

The difference here is that other kinds of exceptions aren't handled, which might actually be right, but now I think about it I wonder if we correctly wrap all the user code in exception handlers so that handlers for one frame can't cause us to stop processing other streams.


<!-- Sent from Reviewable.io -->

dheiberg (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12193)* status: all files reviewed, 3 unresolved discussions (waiting on @jgraham, @dhdavvie, @AutomatedTester, @annevk, @gsnedders, and @andreastt)

___
*[tools/wptserve/wptserve/server.py, line 379 at r9](https://reviewable.io/reviews/web-platform-tests/wpt/12193#-LJsYwEy8EE6saYPxYdN:-LJsrxgA7eIUP0-mJTAE:b-dn8y4) ([raw file](https://github.com/web-platform-tests/wpt/blob/98ce16f20df3d1d252f857026d80f96cbee53c14/tools/wptserve/wptserve/server.py#L379)):*

I would just make the condition `frame is None or isinstance(…)`. 

The difference here is that other kinds of exceptions aren't handled, which might actually be right, but now I think about it I wonder if we correctly wrap all the user code in exception handlers so that handlers for one frame can't cause us to stop processing other streams.
</blockquote></details>

Exceptions raised in on stream thread should not affect the other streams AFAIK

jgraham (approved):
Thanks for all the fixesd!

Want to rewrite the history to remove [WIP] from the commit messages and document the changes made?

dhdavvie:
Doing that now

PR #12196, opened 2018-07-26
Update the webgl1 IDL file

lukebjerring:
Hello reviewer(s),

Please check that:


kenrussell (approved):
Looks good to me.

The WebGL working group already scrapes their own spec into https://github.com/KhronosGroup/WebGL/blob/master/specs/latest/1.0/webgl.idl ; we were asked to add a license header to this a little while ago by a different project. If it makes it easier to copy that file rather than re-scrape the spec please feel free to.

lukebjerring:
Thanks - we're leveraging a quite general scraper, so it's not a problem.

foolip (approved):

PR #12197, opened 2018-07-26
Update the webgl2 IDL file

lukebjerring:
NOTE: Builds on https://github.com/web-platform-tests/wpt/pull/12196 - see https://github.com/lukebjerring/web-platform-tests/pull/7 for diff

---

Hello reviewer(s),

Please check that:

### DIFF
--- a/webgl/idlharness.window.js
+++ b/webgl/idlharness.window.js
@@ -13,3 +13,12 @@ idl_test(
   },
   'webgl1 interfaces'
 );
+
+idl_test(
+  ['webgl2'],
+  ['webgl1', 'dom'],
+  idl_array => {
+    // TODO: objects
+  },
+  'webgl2 interfaces'
+);
--- a/interfaces/webgl2.idl
+++ b/interfaces/webgl2.idl
@@ -1,0 +1,574 @@
+// GENERATED CONTENT - DO NOT EDIT
+// Content of this file was automatically extracted from the
+// "WebGL 2.0 Specification" spec.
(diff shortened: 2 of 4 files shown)


kenrussell (approved):
Looks good to me.

Per your other pull request, might want to consider grabbing the already-scraped IDL directly from the spec's repository:
https://github.com/KhronosGroup/WebGL/blob/master/specs/latest/2.0/webgl2.idl

foolip:
Now rebased after https://github.com/web-platform-tests/wpt/pull/12196 was merged.

webgl/idlharness.window.js
```diff
@@ -13,3 +13,12 @@ idl_test(
   },
   'webgl1 interfaces'
 );
+
+idl_test(
+  ['webgl2'],
```
foolip: Maybe you also tried this, but I put `['webgl1', 'webgl2']` as the first argument here, since it seems like exactly the use case for more than one source, but that causes "20 duplicate test names", seemingly all involving "WebGL2RenderingContext" even though that interface isn't in webgl1.idl. Simply concatenating the two IDL files into one doesn't change matters, so it's not because of something `idl_test` does wrong.

Can you make sense of the problem?

Note that when using two `idl_test`s like this, the subtests are interleaved, in particular the first two are "webgl1 interfaces" and "webgl2 interfaces". Trying to work out which IDL file a problem is from might be tricky. So maybe two separate test files would be better?

foolip (approved):

PR #12201, opened 2018-07-26
HTML: tests for new event handler spec

TimothyGu:
- Add tests for event handler removal when the IDL attribute is set to null or content attribute is removed
- Add tests for event handler re-registration even if content attribute   is set to the same value
- Add tests for event handlers on the body element that are not shadowed  to its `Window`
- Use IDL parser for list of event handlers
- Use .window.js instead of .html for some files

See: https://github.com/whatwg/html/pull/3850


domenic:
w3c-test:mirror (probably doesn't work, but worth trying...)

domenic (commented):
I hate to do this, but could you separate out the changes which are refactoring/style changes to the existing tests, to the ones that introduce new tests? With all the renaming/.window.js-ification, it's hard to tell which is which.

TimothyGu:
@domenic Comments should be addressed.

domenic (approved):
This is super-great, each individual commit is an improvement.

PR #12202, opened 2018-07-26
Fix test case which relied on outdated Name/QName production

cscott:
The latest version of the DOM spec references the Name and QName productions from XML (https://dom.spec.whatwg.org/#validate).  In errata 9 from XML 1.0 these productions were changed to be much more inclusive ( https://www.w3.org/XML/xml-V10-4e-errata#E09 ) which means these old test cases are overly restrictive.
Changed to mark \u0BC6 as valid in a qname, and added a new test case using \u037E for the invalid case.


cscott:
I updated my patch with a few more problematic `0BC6` cases.  See also https://github.com/whatwg/dom/issues/671 for some brief discussion of the exception type used (InvalidCharacterError vs NamespaceError).

annevk:
Thanks, I'm guessing some browsers fail these tests given they implement XML 1.0 4th edition generally?

gsnedders:
AFAIK, every browser *deliberately* implements XML 1.0 4th ed (and not the 5th ed including this errata), so DOM should just cite 4th ed and be clear that you cannot substitute that citation for a later edition?

cscott:
My link is to XML 4th edition, not 5th edition, and the errata linked in the bug description is E09 to the 4th edition (as is clear from the URL).

It is confusing because the 5th edition includes all the errata to the 4th edition.  But a note saying "4th edition of XML" is not clarifying; you'd have to say "4th edition of XML including errata E01 through E08 but no later errata" or something like that.

I did verify that Firefox is still throwing an InvalidCharacterError for `document.createElement('\u0BC6foo')` and the like.  I don't know whether this is "deliberate" or not.

annevk:
@gsnedders that's no longer true, Chrome and WebKit have 5th at least for their parser.

cscott:
@gsnedders in theory Firefox's behavior was harmonized by https://bugzilla.mozilla.org/show_bug.cgi?id=1358104 but it appears they made it match the exception type but didn't update their xml character class. It appears that https://bugzilla.mozilla.org/show_bug.cgi?id=501837 is their open bug for upgrading to the latest version of the expat library, which would fix the issue.

annevk:
So I checked and U+037E and U+0300 are allowed per https://www.w3.org/TR/xml/#NT-NameChar. I think we want to require the 5th edition at this point. That's also what the DOM Standard references.

cscott:
They are allowed as NameChar but not as NameStartChar

annevk (approved):
Sorry, the a: threw me off.

PR #12203, opened 2018-07-26
servo: Use config property instead of kwarg for browser kwargs.

jdm:
This addresses a regression from 3b02e29929bb0062d2c7b5a4ccfaae242bbc4c75 that broke Servo's ability to run tests.


jgraham (approved):

jgraham:
We should really add servo to the smoketests…

PR #12211, opened 2018-07-27
Fold "Unscopable handled correctly" tests into the attribute/operation tests

foolip:
The `[Unscopable]` checks were added here:
https://github.com/web-platform-tests/wpt/pull/9490

However, this extended attribute is very rarely used, currently only
in DOM and Fullscreen. And yet, every property and operation generates
a test like this, which is normally passing, example:
https://wpt.fyi/results/compat/interfaces.any.html

Just fold these into the existing tests for attributes/operations like
the many other aspects already covered. Because of the
"do_interface_attribute_asserts must be the last thing" problem,
change of structure for the attribute test.

### DIFF
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -2109,7 +2109,12 @@ IdlInterface.prototype.test_member_attribute = function(member)
                 "The interface object must have a property " +
                 format_value(member.name));
             a_test.done();
-        } else if (this.is_global()) {
+            return;
+        }
+
+        this.do_member_unscopable_asserts(member);
+
+        if (this.is_global()) {
             assert_own_property(self, member.name,
                 "The global object must have a property " +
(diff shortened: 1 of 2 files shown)


foolip:
Trying it in Chromum:
https://chromium-review.googlesource.com/c/chromium/src/+/1152981

I already tested it a lot locally and didn't see anything other than "PASS Unscopable handled correctly" lines being removed.

foolip:
The Chromium CQ run looks good, just like locally all see is a lot of "PASS Unscopable handled correctly" lines being removed.

lukebjerring (approved):

lukebjerring:
A `resources/test` unit test covering the negative case might be justified?

lukebjerring:
Still LGTM

PR #12217, opened 2018-07-27
Don't test clone of obsolete <font> and <dir> elements

cscott:
HTMLFontElement and HTMLDirectoryElement are obsolete and not included in the HTML5 specs.  Don't include them in the Node#cloneNode tests.


cscott:
You're right of course.  I must have been looking at some old/strange version of the HTML5 spec which omitted these entirely.  Thanks for the clarification.

PR #12225, opened 2018-07-30
Clean up IDL tests which reference SVG

lukebjerring:
Follow-up from #11796 to import the `SVGElement` interface properly.


pwnall (approved):
cookie-store/ change LGTM

html/dom/interfaces.https.html
```diff
@@ -33,210 +33,192 @@ <h1>HTML IDL tests</h1>
   return input;
 }
 
-function doTest([html, dom, cssom, uievents, touchevents]) {
-  var idlArray = new IdlArray();
-  var svg = "interface SVGElement : Element {};";
-  idlArray.add_untested_idls(dom + svg + cssom + uievents + touchevents);
-  idlArray.add_idls(html);
+promise_test(async () => {
+  await new Promise(resolve => { addEventListener('load', resolve); })
```
foolip: Whether this is needed I don't know, but to avoid the extra test (now implicitly called "HTML IDL tests"?) can the await part be moved into the async setup of idl_test? (The even listener itself needs to be added here, since otherwise the fetches and the load event will be racing.)

foolip (approved):
LGTM % load event thing

html/dom/interfaces.https.html
lukebjerring: Done.

PR #12230, opened 2018-07-30
webrtc: throw InvalidAccessError on mismatching rtcpmuxPolicy in setRemoteDescription

fippo:
throws InvalidAccessError when setting a remote description
without rtcp-mux and rtcpMuxPolicy is set to 'require'.

test for https://github.com/w3c/webrtc-pc/pull/1942

### DIFF
--- a/webrtc/RTCConfiguration-rtcpMuxPolicy.html
+++ b/webrtc/RTCConfiguration-rtcpMuxPolicy.html
@@ -140,4 +140,56 @@
       Tested    2
       Total     2
    */
+  const FINGERPRINT_SHA256 = '00:00:00:00:00:00:00:00:00:00:00:00:00' +
+      ':00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00';
+  const ICEUFRAG = 'someufrag';
+  const ICEPWD = 'somelongpwdwithenoughrandomness';
+
+  promise_test(async t => {
+    // audio-only SDP offer without BUNDLE and rtcp-mux.
+    const sdp = 'v=0\r\n' +
+        'o=- 166855176514521964 2 IN IP4 127.0.0.1\r\n' +
+        's=-\r\n' +
+        't=0 0\r\n' +
+        'm=audio 9 UDP/TLS/RTP/SAVPF 111\r\n' +
+        'c=IN IP4 0.0.0.0\r\n' +
+        'a=rtcp:9 IN IP4 0.0.0.0\r\n' +
+        'a=ice-ufrag:' + ICEUFRAG + '\r\n' +
+        'a=ice-pwd:' + ICEPWD + '\r\n' +
+        'a=fingerprint:sha-256 ' + FINGERPRINT_SHA256 + '\r\n' +
+        'a=setup:actpass\r\n' +
+        'a=mid:audio1\r\n' +
+        'a=sendonly\r\n' +
+        'a=rtcp-rsize\r\n' +
+        'a=rtpmap:111 opus/48000/2\r\n';
+    const pc = new RTCPeerConnection({rtcpMuxPolicy: 'require'});
+    t.add_cleanup(() => pc.close());
+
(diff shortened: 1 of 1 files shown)


webrtc/RTCConfiguration-rtcpMuxPolicy.html
```diff
@@ -140,4 +140,56 @@
       Tested    2
       Total     2
    */
+  const FINGERPRINT_SHA256 = '00:00:00:00:00:00:00:00:00:00:00:00:00' +
+      ':00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00';
+  const ICEUFRAG = 'someufrag';
+  const ICEPWD = 'somelongpwdwithenoughrandomness';
+
+  promise_test(async t => {
+    const sdp = 'v=0\r\n' +
```
fippo: main reason for static SDP is that bundle implies rtcp-mux so we need to do unbundled

webrtc/RTCConfiguration-rtcpMuxPolicy.html
```diff
@@ -140,4 +140,56 @@
...
+        's=-\r\n' +
+        't=0 0\r\n' +
+        'm=audio 9 UDP/TLS/RTP/SAVPF 111\r\n' +
+        'c=IN IP4 0.0.0.0\r\n' +
+        'a=rtcp:9 IN IP4 0.0.0.0\r\n' +
+        'a=ice-ufrag:' + ICEUFRAG + '\r\n' +
+        'a=ice-pwd:' + ICEPWD + '\r\n' +
+        'a=fingerprint:sha-256 ' + FINGERPRINT_SHA256 + '\r\n' +
+        'a=setup:active\r\n' +
+        // 'a=mid:AUDIOMID\r\n' +
```
fippo: omitting this is easier than parsing it -- or depending on getTransceivers to avoid parsing it from the sdp

webrtc/RTCConfiguration-rtcpMuxPolicy.html
```diff
@@ -140,4 +140,56 @@
...
+        'm=audio 9 UDP/TLS/RTP/SAVPF 111\r\n' +
+        'c=IN IP4 0.0.0.0\r\n' +
+        'a=rtcp:9 IN IP4 0.0.0.0\r\n' +
+        'a=ice-ufrag:' + ICEUFRAG + '\r\n' +
+        'a=ice-pwd:' + ICEPWD + '\r\n' +
+        'a=fingerprint:sha-256 ' + FINGERPRINT_SHA256 + '\r\n' +
+        'a=setup:actpass\r\n' +
+        'a=mid:audio1\r\n' +
+        'a=sendonly\r\n' +
+        // 'a=rtcp-mux\r\n' +
```
alvestrand: this looks like commented-out that might get commented back in - can you instead put a comment on top of the SDP saying "a=rtcp-mux is NOT present"?

webrtc/RTCConfiguration-rtcpMuxPolicy.html
alvestrand: agreed. is the spec clear on which goes first of the mid check and the rtcp-mux check?
(I'm not clear on whether interworking with non-mid-supporting endpoints is actually required, but if it is, then omitting it is a Good Thing - need to test that too)

alvestrand (commented):
Mostly good.

webrtc/RTCConfiguration-rtcpMuxPolicy.html
fippo: done, on both places. Also for the a=mid

webrtc/RTCConfiguration-rtcpMuxPolicy.html
fippo: the spec has provisions for this at least:
```
Set transceiver's mid value to the mid of the corresponding media description. If the media description has no MID, and transceiver's mid is unset, generate a random value as described in [JSEP] (section 5.10.).
```
I smell another test but separate from this :-)

alvestrand (approved):
LGTM

PR #12240, opened 2018-07-31
HTML: Add bailout order checks for document.open()

TimothyGu:


html/webappapis/dynamic-markup-insertion/opening-the-input-stream/bailout-order.window.js
```diff
@@ -0,0 +1,29 @@
+function openHasTakenEffect(doc) {
+  return !doc.childNodes.length;
+}
+
+async_test(t => {
+  const iframe = document.body.appendChild(document.createElement("iframe"));
+  t.add_cleanup(() => { try { iframe.contentDocument.close(); } catch (e) {} });
```
domenic: When would close throw? Would doing so be OK per spec? Swallowing an exception always deserves a comment, I think.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/bailout-order.window.js
```diff
@@ -0,0 +1,29 @@
...
+  iframe.src = "/common/dummy.xhtml";
+  iframe.onload = t.step_func_done(() => {
+    const origURL = iframe.contentDocument.URL;
+    assert_throws("InvalidStateError", () => {
+      iframe.contentDocument.open();
+    }, "document.open() should throw on XML documents");
+    assert_false(openHasTakenEffect(iframe.contentDocument), "document.open() should not clear a XML document's nodes");
```
domenic: I might abstract these into an `assertOpenHasNotTakenEffect(document)`

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/bailout-order.window.js
```diff
@@ -0,0 +1,29 @@
...
+  const iframe = document.body.appendChild(document.createElement("iframe"));
+  t.add_cleanup(() => { try { iframe.contentDocument.close(); } catch (e) {} });
+  self.testSynchronousScript = t.step_func_done(() => {
+    const origURL = iframe.contentDocument.URL;
+    iframe.contentDocument.open();
+    assert_false(openHasTakenEffect(iframe.contentDocument), "document.open() should not clear a document nodes when executed as a synchronous script");
+    assert_equals(iframe.contentDocument.URL, origURL, "document.open() should keep the original URL on a document when executed as a synchronous script");
+  });
+  iframe.src = "resources/bailout-order-synchronous-script-frame.html";
+}, "URL should not be changed on a document.open bailout (synchronous script)");
```
domenic: I would use the spec phrasing so it's more clear what you're testing: e.g. "script nesting level > 0" or similar.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/bailout-order.window.js
```diff
@@ -0,0 +1,29 @@
+function openHasTakenEffect(doc) {
```
domenic: From this filename, I would expect this test to be complete, but the rest of the checks are in other files.

Since the security origin check needs its own file, maybe the right thing is to move `assertOpenHasNotTakenEffect(document)` into a shared helper file, and also add references to that file to this one. Or, rename the other file something like "bailout-order-same-origin". (And this one "bailout-order-xml-and-sync-script"?)

I also believe we have similar tests for the custom elements thing and the active parser thing. Can we similarly rename or consolidate those?

domenic (commented):
These tests are pretty good themselves, but I'd like to do a bit of work to make it clearer our coverage is complete.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/bailout-order.window.js
TimothyGu: It would throw in this case with an XML document, since `document.close()` is spec'd to throw. However, I don't think every browser does that.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/bailout-order.window.js
TimothyGu: For custom elements it's more difficult to test, in fact I haven't quite found a way to test custom elements' interactions with URl. The catch is that the check would only be triggered if the custom element is created by the parser.

TimothyGu:
@domenic Comments should all be addressed.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/bailout-side-effects-same-origin-domain.sub.window.js
```diff
@@ -0,0 +1,14 @@
...
+
+document.domain = "{{host}}";
+
+testInIFrame("http://{{host}}:{{ports[http][1]}}/common/domain-setter.sub.html", (ctx) => {
+  const iframe = ctx.iframes[0];
+  const origURL = iframe.contentDocument.URL;
+  assertDocumentIsReadyForSideEffectsTest(iframe.contentDocument, "same origin-domain (but not same origin) document");
+  assert_throws("SecurityError", () => {
+    ctx.iframes[0].contentDocument.open();
+  }, "Opening a same origin-domain (but not same origin) document doesn't throw.");
```
domenic: Assertions probably shouldn't say what failed to happen. Instead use the "should" form you use elsewhere.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/origin-check-in-document-open-same-origin-domain.sub.html
```diff
@@ -10,12 +10,14 @@
 testInIFrame("http://{{host}}:{{ports[http][1]}}/html/webappapis/dynamic-markup-insertion/opening-the-input-stream/resources/set-document-domain.html", (ctx) => {
   document.domain = document.domain;
   var doc = ctx.iframes[0].contentDocument;
+  const origURL = doc.URL;
```
domenic: Leftovers

domenic (approved):
LGTM with nits

PR #12243, opened 2018-07-31
Extend coverage reporting to wptserve

dheiberg:
So for some reason `wptserve` was omitted from the pytest coverage. I suspect that this is accidental/remnants of an old system, as it would make sense to have coverage reports of wptserve. If I am mistaken and this is intentional, please let me know!


jgraham (approved):
I don't rememeber why this was like that, but it doesn't make sense to me to omit it so let's try removing it.

PR #12248, opened 2018-07-31
[URL] Add additional tests for URLSearchParams.sort()

TRowbotham:
This adds the following tests for:

* Parameter names with more than one character.
* Parameter names with identical characters, but differing lengths.
* Paramter names that are the empty string.
* Parameter names that contain characters with the same number of code units.


url/urlsearchparams-sort.html
```diff
@@ -24,6 +24,24 @@
   {
     "input": "z=z&a=a&z=y&a=b&z=x&a=c&z=w&a=d&z=v&a=e&z=u&a=f&z=t&a=g",
     "output": [["a", "a"], ["a", "b"], ["a", "c"], ["a", "d"], ["a", "e"], ["a", "f"], ["a", "g"], ["z", "z"], ["z", "y"], ["z", "x"], ["z", "w"], ["z", "v"], ["z", "u"], ["z", "t"]]
+  },
+  {
+    "input": "bbb&bb&aaa&aa=x&aa=y",
+    "output": [
+        ["aa", "x"],
```
TimothyGu: Could you use the same indentation style as everywhere else?

TimothyGu (commented):
Generally looks good, one nit.

TimothyGu (approved):
I'll let @annevk take a look at this as well but LGTM.

annevk:
Thanks!

PR #12267, opened 2018-08-02
[web-locks] Detect errors to avoid timeouts

jugglinmike:
Ensure that uncaught errors within workers cause test failure. This
makes non-conforming browsers fail more quickly and with more
descriptive messages.


web-locks/resources/helpers.js
```diff
@@ -54,6 +54,7 @@
         resolve(event.data);
       };
       worker.addEventListener('message', listener);
+      worker.addEventListener('error', (e) => reject(e && e.message));
```
pwnall: nit: `(e)` -> `e` to match the style in the rest of the test

pwnall (approved):
LGTM, seems like a good step forward.

Might be worth adding a comment to `postToFrameAndWait` stating that iframes don't have an `error` event, for readers wondering why that function doesn't take the same approach.

inexorabletash (approved):
Thanks!

web-locks/resources/helpers.js
jugglinmike: Sure thing!

jugglinmike:
> Might be worth adding a comment to postToFrameAndWait stating that iframes
> don't have an error event, for readers wondering why that function doesn't
> take the same approach.

This conflicted with my experience with Firefox, and after further research, I
learned that may be due to a bug in that browser. In case you're curious, the
result is at gh-12283.

In the mean time, I've updated the comment with the note you've suggested.

jugglinmike:
@pwnall @inexorabletash The build on TravisCI is failing because his change causes `web-locks/frames.tentative.https.html` to behave inconsistently in Chrome. Any idea why?

jugglinmike:
Since our CI infrastructure has changed so drastically over the past year, I've merged `master` into this branch to trigger a new trial on the new system. The good news is that `web-locks/frames.tentative.https.html` behaves consistently in Firefox and Chrome, now. The bad news is that today, `web-locks/signal.tentative.https.any.js` is the flaky one. That's actually a problem in `master`, so I've filed a separate pull request to address it: gh-20244.

PR #12269, opened 2018-08-02
[css-contain] Containment and stacking contexts

mrego:
This patch adds tests to verify when layout and paint containment
create a stacking context.


frivoal (approved):
LGTM

PR #12270, opened 2018-08-02
[css-contain] Apply different changes to improve tests

mrego:
Gérard Talbot shared some suggestions to improve css-contain tests suite:
http://www.gtalbot.org/BrowserBugsSection/CSS3Contain/CSS3Contain-ReviewedTests.txt

This patch applies the changes suggested in section:
APPROVED by me with some MODIFICATIONS, IMPROVEMENTS.

### DIFF
--- a/css/css-contain/contain-layout-001.html
+++ b/css/css-contain/contain-layout-001.html
@@ -9,11 +9,15 @@
   <link rel=help href="https://drafts.csswg.org/css-contain-1/#containment-layout">
 
 <style>
-span {
+#pa {
   contain: layout;
   height: 100vh; /*If layout containment applies, the span becomes a BFC, height applies, and knocks SS off the page */
+}
+
+#ss {
+  vertical-align: bottom;
 }
 </style>
 
 <p>Test passes if there is the word "PASS" below.</p>
-<div><span>PA</span>SS</div>
+<div><span id="pa">PA</span><span id="ss">SS</span></div>
--- a/css/css-contain/contain-size-breaks-001.html
+++ b/css/css-contain/contain-size-breaks-001.html
@@ -6,7 +6,7 @@
   <meta name=flags content="ahem">
   <meta name=assert content="size containment makes element monolithic">
   <link rel="match" href="reference/contain-size-breaks-001-ref.html">
-  <link rel=help href="https://drafts.csswg.org/css-contain-1/#containment-style">
+  <link rel=help href="https://drafts.csswg.org/css-contain-1/#containment-size">
   <link rel=help href="https://drafts.csswg.org/css-break-3/#monolithic">
 
 <style>
@@ -15,7 +15,6 @@ article {
   column-gap: 0;
   columns: 3 1em;
   width: 3em;
-  column-fill: auto;
 
   font-size: 40px;
   font-family: ahem;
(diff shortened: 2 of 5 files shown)


css/css-contain/contain-size-breaks-001.html
```diff
@@ -31,18 +30,36 @@
...
-<div>A A A A A</div>
+<div>A B C D E</div>
 </article>
 
 <!--
-If the element is monolythic, it may:
-* overflow and be a 1x5 em rectangle
-* be sliced and be 2 contigious 1x2.5 em rectangles, appearing as a single
-  2x2.5 em rectangle.
+As the element is monolithic, it would be sliced in two contiguous 2.5em rectangles.
```
frivoal: The ASCII Art is good, but the comment isn't quite right. should be

> As the element is monolithic, it should either:
> * Overflow vertically, appearing as a signle 1x5 em rectangle
>  [... diagram...]
> * Be sliced, with one slice in each column, appearing as  a single 2x2.5 em rectangle
>  [... diagram...]

In other words, in the case where it overflows, it isn't being sliced.

css/css-contain/contain-layout-001.html
```diff
@@ -11,9 +11,14 @@
 <style>
 span {
   contain: layout;
+  display: inline-block;
   height: 100vh; /*If layout containment applies, the span becomes a BFC, height applies, and knocks SS off the page */
 }
+
+#ss {
+  vertical-align: bottom;
```
frivoal: I disagree with Gérard that this is necessary, as the initial value is baseline, which should be sufficient. 
That said, `vertical-align: bottom` isn't wrong, and maybe it is more explicit, so OK.

css/css-contain/contain-layout-001.html
```diff
@@ -11,9 +11,14 @@
 <style>
 span {
   contain: layout;
+  display: inline-block;
```
frivoal: Delete the `display: inline-block;` line. we're trying to test the absence of effect on non-atomic inlines, and this makes it into an atomic inline. Maybe an accidental leftover from debugging?

css/css-contain/contain-layout-001.html
```diff
@@ -11,9 +11,14 @@
 <style>
 span {
```
frivoal: This selector should not be `span`, but should only select the first of the two spans. Whether you want to add an id or use `:first-of-type` or whatever is up to you, but it should only select the first span.

frivoal (changes requested):
Overall, it's an improvement (tests simplified, comments improved), but there are a couple of mistakes. Please fix first.

css/css-contain/contain-size-breaks-001.html
mrego: Ok thanks for the clarification, I've updated the comment.

css/css-contain/contain-layout-001.html
mrego: Ok.

css/css-contain/contain-layout-001.html
mrego: Ooops, yeah sorry I was checking some things there and I forgot to delete it later.

css/css-contain/contain-layout-001.html
mrego: Done.

mrego:
Thanks for the review, uploaded new version fixing the mistakes.

frivoal (approved):

PR #12274, opened 2018-08-02
Fix webusb IDL test

lukebjerring:
Cleans up the IDL test by inverting the containment, such that failing to set up a fake device does not prevent the rest of the IDL from being tested.

Also fixes a bug in the way that we check whether `MojoInterfaceInterceptor` is defined (when it's not).


reillyeon:
Why did you rename the test back from idlharness.https.any.js to idlharness.any.js?

lukebjerring:
@reillyeon - Was confused about `idlharness.https.https.any.serviceworker.html` and misinterpreted this to mean that the `https` was redundant. Reverted. PTAL.

reillyeon (approved):

PR #12281, opened 2018-08-02
Ensure that in-repo files that match gitignore rules are synced.

jdm:
This addresses an issue where `tools/third_party/attrs/.coveragerc` was added to that package's gitignore file but the but the file remains in the vendored package.

### DIFF
--- a/tools/wptrunner/wptrunner/update/tree.py
+++ b/tools/wptrunner/wptrunner/update/tree.py
@@ -1,6 +1,7 @@
 import os
 import re
 import subprocess
+import tempfile
 
 from .. import vcs
 from ..vcs import git, hg
@@ -38,6 +39,9 @@ class NoVCSTree(object):
         return True
 
     def add_new(self, prefix=None):
+        pass
+
+    def add_ignored(self, sync_tree, prefix):
         pass
 
     def create_patch(self, patch_name, message):
@@ -89,6 +93,9 @@ class HgTree(object):
         else:
             args = ()
         self.hg("add", *args)
+
+    def add_ignored(self, sync_tree, prefix):
+        pass
 
     def create_patch(self, patch_name, message):
         try:
@@ -178,10 +185,26 @@ class GitTree(object):
                        add all files under that path.
         """
         if prefix is None:
-            args = ("-a",)
+            args = ["-a"]
         else:
-            args = ("--no-ignore-removal", prefix)
+            args = ["--no-ignore-removal", prefix]
         self.git("add", *args)
+
+    def add_ignored(self, sync_tree, prefix):
+        """Add files to the staging area that are explicitly ignored by git.
+
+        :param prefix: None to include all files or a path prefix to
+                       add all files under that path.
+        """
+        with tempfile.TemporaryFile() as f:
+            sync_tree.git("ls-tree", "-r", "--name-only", "HEAD", stdout=f)
+            f.seek(0)
+            ignored_files = sync_tree.git("check-ignore", "--no-index", "--stdin", "-z", stdin=f)
+        args = []
+        for entry in ignored_files.split('\0'):
+            args.append(os.path.join(prefix, entry))
+        if args:
+            self.git("add", "--force", *args)
 
     def list_refs(self, ref_filter=None):
         """Get a list of sha1, name tuples for references in a repository.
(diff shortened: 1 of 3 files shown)


jdm:
This should address https://github.com/servo/servo/issues/21319; I've verified that the code runs without any exceptions, but it's extremely difficult to test the behaviour of sync-related code which gets overwritten by the upstream copy during the actual sync.

tools/wptrunner/wptrunner/update/tree.py
```diff
@@ -183,6 +189,20 @@ def add_new(self, prefix=None):
             args = ("--no-ignore-removal", prefix)
         self.git("add", *args)
 
+    def add_ignored(self, prefix=None):
+        """Add files to the staging area that are explicitly ignored by git.
+
+        :param prefix: None to include all files or a path prefix to
+                       add all files under that path.
+        """
+        ignored_files = self.git("status", "--short", "--ignored", prefix)
```
jgraham: I don't think this does what you want since it includes files that actually ought to be ignored. So far I only found `git ls-tree -r --name-only HEAD | git check-ignore --no-index --stdin` to list files that would be ignored but aren't.

jdm: I'm not exactly clear on what "actually ought to be ignored" means. The command in this PR only returns .coveragerc; the command you suggest returns that as well as a couple testharness_runner.html files.

jdm: Mmm, when I run the status-based command in my web-platform-tests clone, I get a bunch of pyc files. In the sync tree those don't exist, though.

jgraham: Yeah, it's the latter I'm worried about because if you start causing those to exist (e.g. by running the manifest update at a different point during sync or similar) then you end up accidentially checking them in. I think the files you want is the intersection of files in the working tree that match a gitignore rule and files that are in the HEAD commit.

jdm:
I tested this locally and the .coveragerc file was correctly added to the synced git repository.

tools/wptrunner/wptrunner/update/tree.py
```diff
@@ -178,11 +185,27 @@ def add_new(self, prefix=None):
...
+    def add_ignored(self, sync_tree, prefix):
+        """Add files to the staging area that are explicitly ignored by git.
+
+        :param prefix: None to include all files or a path prefix to
+                       add all files under that path.
+        """
+        with tempfile.TemporaryFile() as f:
+            sync_tree.git("ls-tree", "-r", "--name-only", "HEAD", stdout=f)
+            f.seek(0)
+            ignored_files = sync_tree.git("check-ignore", "--no-index", "--stdin", stdin=f).split()
```
jgraham: I think this would be `splitlines()` but really you want to pass `-z` and split on `\0`

tools/wptrunner/wptrunner/update/tree.py
```diff
@@ -178,11 +185,27 @@ def add_new(self, prefix=None):
...
+        args = []
+        for entry in ignored_files:
+            args.append(os.path.join(prefix, entry))
+        if args:
+            self.git("add", *(["--force"] + args))
```
jgraham: `self.git("add", "--force", *args)`

tools/wptrunner/wptrunner/update/tree.py
```diff
@@ -178,11 +185,27 @@ def add_new(self, prefix=None):
...
+            args = ["--no-ignore-removal", prefix]
         self.git("add", *args)
 
+    def add_ignored(self, sync_tree, prefix):
+        """Add files to the staging area that are explicitly ignored by git.
+
+        :param prefix: None to include all files or a path prefix to
+                       add all files under that path.
+        """
+        with tempfile.TemporaryFile() as f:
```
jgraham: Using a temp file seems potentially unnecessary? You could pipe the commands together, but it might be simpler to use a `io.BytesIO` here? Up to you.

jdm: io.BytesIO does not appear to be supported as a stdin or stdout argument for subprocess.check_output/subprocess.check_call in Python <3.0. Using subprocess.PIPE with these functions is called out as dangerous by the docs; I would rather stick with the implementation that exists and works than continue to try to find something else that I can use instead.

jdm:
I've pushed changes to address the other comments.

jgraham (approved):

PR #12283, opened 2018-08-02
[html] Add test for iframe onerror event

jugglinmike:
Firefox emits an "onerror" event for uncaught exceptions within iframes whose
location matches the parent. I believe this is incorrect, but I'm not certain.
That's because saying so requires confidence in the absence of specification
text, and it's possible that I am simply not looking in the right place.

It is strange this behavior is only exhibited when the iframe's location
matches that of the parent. That makes me slightly more confident that this is
unintentional.

@bz I think you could speak on this much more authoritatively; would you mind
weighing in?


zcorpan:
I think you meant @bzbarsky

jugglinmike:
You're right, @zcorpan, I do. Thanks!

bzbarsky:
> Firefox emits an "onerror" event for uncaught exceptions within iframes whose
location matches the parent. 

That's not what the testcase shows.

What the testcase shows is that in Firefox is a page with some url tries to load a subframe with the same exact url, the load is blocked.  This is specified at https://html.spec.whatwg.org/multipage/iframe-embed-object.html#otherwise-steps-for-iframe-or-frame-elements step 2.  When Firefox does the blocking, we fire an error event on the `<iframe>`.  The spec doesn't say to fire any events, whether load or error, but that seems like a pretty weird behavior to me; the load is not being done as requested, and firing error seems appropriate.

Chrome doesn't implement this part of the spec correctly, as far as I can tell; it loads the subframe (though not inside itself; for that latter case it doesn't fire either load or error).

Anyway, none of this has to do with the exceptions thrown inside the iframe, as you can tell if you just comment out the `if (isChild) { throw ... }` bit.  None of that code is even running in this case in the child.

jugglinmike:
> The spec doesn't say to fire any events, whether load or error, but that
> seems like a pretty weird behavior to me; the load is not being done as
> requested, and firing error seems appropriate.

I agree. [I've forwarded your observation to the relevant discussion in HTML](https://github.com/whatwg/html/issues/1518#issuecomment-410118468).

> Anyway, none of this has to do with the exceptions thrown inside the iframe,

To maintain the focus of this pull request, I re-implemented the test to only assert that the event is not fired in response to uncaught errors. As far as I know, that's not actually disputed by implementers, though. I don't think there's value in maintaining such a test. So I'll close this pull request.

As for the behavior that *is* disputed, I've opened a new pull request: gh-12290

PR #12288, opened 2018-08-02
Compress wpt_report.json on Taskcluster

### LINKED ISSUE #12265: Compress the JSON output in Taskcluster tasks
For unknown reasons, Taskcluster disabled the [server-side auto compression](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/ServingCompressedFiles.html#compressed-content-S3-origin) for the task artifacts (hosted on taskcluster-artifacts.net).

For each run we have tens of large JSON files; each of them is about tens of MB. Based on my observations in the results receiver, although the full report can take up to nearly 300MB, the gzipped form is only ~10MB (due to lots of duplicated strings). We can really benefit from compressing the artifacts.

In addition to performance gains, compressing the reports will also give me a much larger safety margin against various AppEngine runtime limits in the results receiver (e.g. max fetch size, timeout, ...). It's much easier to change the task definition to compress the output than to hack around the limits using async queues etc.

Again, this is mostly about `wpt_report.json`. I can either compress the report in place, or produce another compressed copy of the report. I'll try the former first as I'll probably be the only consumer of these artifacts in the near future, unless anyone objects my PR :)

Hexcles:
Use gzip to compress the output artifcat wpt_report.json in place.

Fixes #12265.

I think no one else is currently fetching the artifacts, so I chose in-place compression. If it breaks anything and/or causes any concern, I'm fine with preserving the original file.

### DIFF
--- a/tools/ci/ci_taskcluster.sh
+++ b/tools/ci/ci_taskcluster.sh
@@ -7,3 +7,4 @@ if [ $1 == "firefox" ]; then
 elif [ $1 == "chrome" ]; then
     ./wpt run chrome --log-tbpl=../artifacts/log_tbpl.log --log-tbpl-level=info --log-wptreport=../artifacts/wpt_report.json --log-mach=- --this-chunk=$3 --total-chunks=$4 --test-type=$2 -y  --no-pause --no-restart-on-unexpected --install-fonts --no-fail-on-unexpected
 fi
+gzip ../artifacts/wpt_report.json


jgraham (approved):

jgraham:
I assume we're OK with the fact that if the job fails this never runs? I'll merge it for now but perhaps we should consider something more robust.

Hexcles:
@jgraham could you explain the failure mode? As long as `wpt_report.json` is produced, regardless of whether the `wpt run` command succeeds or not, the file will be compressed in place (there's no `set -e` in the script).

jgraham:
Oh, I was assuming there was without checking. Don't mind me!

Issue #11529, opened 2018-06-15
Merge .window.js/.worker.js/.any.js


zcorpan:
I think it's a bit confusing how the three kinds of tests are subtly different, and I imagine that new contributors can be similarly confused.

I have two ideas for possible improvement:

* Use only .any.js and update the tests accordingly.
* Still use the 3 filename conventions, but make it more consistent how to write each kind of test (e.g. include `GLOBAL` in .window.js and .worker.js).

The first would simplify the implementation, the second could be slightly more convenient for test authors.

annevk:
I kinda like `window.js`. It seems silly that `worker.js` currently requires `importScripts("/resources/testharness.js");` at the top though.

inexorabletash:
`worker.js` also requires a `done();` at the end.

(Came here to file a bug proposing that we fix worker.js and update the tests.)

PR #12293, opened 2018-08-03
Use a .worker.js test for WorkerGlobalScope_requestAnimationFrame.

Ms2ger:


workers/WorkerGlobalScope_requestAnimationFrame.tentative.worker.js
```diff
@@ -0,0 +1,19 @@
+importScripts("/resources/testharness.js");
```
annevk: Is this needed? Seems weird that it's required for worker.js and not window.js.

annevk (approved):

workers/WorkerGlobalScope_requestAnimationFrame.tentative.worker.js
Ms2ger: It is. One could file an issue about that; or perhaps we should remove `.worker.js`.

annevk: Commented on #11529.

PR #12296, opened 2018-08-03
[Streams] Add a test for piping to errored

ricea:
The test 'Piping from an errored readable stream to an errored writable
stream' was misnamed, as the destination writable was erroring, not
errored. Rename it. Add a variant of the test where pipeTo() isn't
called until the destination is errored.

### DIFF
--- a/streams/piping/multiple-propagation.js
+++ b/streams/piping/multiple-propagation.js
@@ -12,6 +12,24 @@ error1.name = 'error1';
 const error2 = new Error('error2!');
 error2.name = 'error2';
 
+function createErroredWritableStream(t) {
+  return Promise.resolve().then(() => {
+    const ws = recordingWritableStream({
+      start(c) {
+        c.error(error2);
+      }
+    });
+
+    const writer = ws.getWriter();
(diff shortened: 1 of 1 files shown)


streams/piping/multiple-propagation.js
```diff
@@ -35,6 +35,32 @@ promise_test(t => {
...
+    }
+  });
+  const ws = recordingWritableStream({
+    start(c) {
+      c.error(error2);
+    }
+  });
+
+  // Trying to abort a stream that was errored will give that error back
+  return Promise.resolve()
```
domenic: The test might be more obviously correct if you checked the errored state of `ws` before doing the piping, instead of checking it at the end.

domenic (approved):
LGTM with potential improvement.

streams/piping/multiple-propagation.js
ricea: Done. I also split a couple more tests into erroring / errored variants, although in these cases the asserts don't change. PTAL.

domenic (approved):
Very nice. Will let you squash and merge, or rebase and merge, as desired.

PR #12303, opened 2018-08-03
Fix a t.assert_unreached typo in a CSP test helper

foolip:


foolip:
I've confirmed this fixes the harness in Chrome, Firefox and Safari.

andypaicu (approved):
This piece of code seems eerily familiar so I'm going to assume this is my fault in the first place. Therefore thank you for fixing it.

PR #12304, opened 2018-08-03
Wrap a stray assert (causing harness error) in a test

foolip:


domfarolino (approved):
Thank you! Good catch

PR #12305, opened 2018-08-03
Wrap an assert in step_func in a CSP test

foolip:
Previously, this test caused a harness error in Firefox and Safari, now
the test itself fails correctly instead.


andypaicu (approved):
LGTM

PR #12310, opened 2018-08-04
Add default prompt value for user prompt WebDriver tests

jimevans:
When the JavaScript `prompt()` function is called without a default value,
some browsers (like IE) return "undefined" for the default prompt value if
the argument is omitted. This commit refactors the user prompt handler
tests so that prompt() is called with an appropriate value for the default
value argument, providing a consistent behavior across browsers.

### DIFF
--- a/webdriver/tests/support/fixtures.py
+++ b/webdriver/tests/support/fixtures.py
@@ -253,12 +253,16 @@ def create_dialog(session):
     function to validate that the dialog has been "handled" (either accepted or
     dismissed) by returning some value."""
 
-    def create_dialog(dialog_type, text=None):
+    def create_dialog(dialog_type, text=None, prompt=None):
         assert dialog_type in ("alert", "confirm", "prompt"), (
             "Invalid dialog type: '%s'" % dialog_type)
 
         if text is None:
             text = ""
+
+        dialog_params = "text"
+        if prompt is not None:
+            dialog_params += ", prompt"
 
         assert isinstance(text, basestring), "`text` parameter must be a string"
 
@@ -266,11 +270,12 @@ def create_dialog(session):
         session.execute_async_script("""
             let dialog_type = arguments[0];
             let text = arguments[1];
+            let prompt = arguments[2];
 
-            setTimeout(function() {
-              window.dialog_return_value = window[dialog_type](text);
-            }, 0);
-            """, args=(dialog_type, text))
+            setTimeout(function() {{
+              window.dialog_return_value = window[dialog_type]({});
+            }}, 0);
+            """.format(dialog_params), args=(dialog_type, text, prompt))
 
         wait(session,
              lambda s: s.alert.text == text,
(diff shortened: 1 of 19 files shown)


andreastt:
I think @whimboo should have a look at this.

whimboo:
Hi @jimevans. Thank you for the PR. I had a quick glance over it and two questions came up for me:

1) Which versions of IE is your driver still supporting? As I read on the cross-browser [MDN documentation for `prompt`](https://developer.mozilla.org/en-US/docs/Web/API/Window/prompt) this specific `default value` behavior is only necessary for IE7 and IE8. So I assume those are still on your compatibility list?

2) I don't think that we should expose the default value all the way down to the tests. Also because it adds a huge amount of boiler-plate. Instead why can't we just hide it's usage inside of the `create_dialog` fixture? I feel it's a proper place, and we can determine it by the `dialog_type` argument.

jimevans:
@whimboo

1. The docs are wrong. I tested this against IE11, and the same behavior happens even outside the WPT test harness. 

2. That’s a perfectly fine approach. I’ll refactor to make that work.

webdriver/tests/support/fixtures.py
```diff
@@ -262,15 +262,20 @@ def create_dialog(dialog_type, text=None):
 
         assert isinstance(text, basestring), "`text` parameter must be a string"
 
+        dialog_params = "text"
+        if dialog_type == "prompt":
+            # For prompt() dialogs, add a value for the 'default' argument
```
whimboo: Please note that IE needs that.

webdriver/tests/support/fixtures.py
```diff
@@ -262,15 +262,20 @@ def create_dialog(dialog_type, text=None):
...
+            dialog_params += ", ''"
```
whimboo: Mind, completely moving this into the script code so that we can get rid of the call to `format()`?

whimboo (changes requested):

webdriver/tests/support/fixtures.py
```diff
@@ -263,12 +263,19 @@ def create_dialog(dialog_type, text=None):
         assert isinstance(text, basestring), "`text` parameter must be a string"
 
         # Script completes itself when the user prompt has been opened.
+        # For prompt() dialogs, add a value for the 'default' argument,
+        # as some user agents (IE, for example) do not produce consistent
```
jugglinmike: Kudos for resisting "(i.e. IE)" :P

jimevans: @jugglinmike True confession time, I very nearly did "(e.g., IE)".

whimboo (approved):
What a difference to the first version. Thanks a lot for this addition @jimevans! Would you mind to update the MDN article?

whimboo:
Oh, and please don't forget to squash before merge. Thanks.

jgraham:
wdspec timeouts.

PR #12313, opened 2018-08-06
Add tests for dynamic changes to base urls

### LINKED ISSUE #2122: Write tests for dynamic changes to <base> elements
Base elements are tested, and dynamic changes to xml:base attributes in XHTML documents is tested, but dynamic changes to base elements aren't.

xiuqijix:
Fix #2122, @nox, @anneke, PTAL.

### DIFF
--- a/html/infrastructure/urls/dynamic-changes-to-base-urls/dynamic-urls.sub.html
+++ b/html/infrastructure/urls/dynamic-changes-to-base-urls/dynamic-urls.sub.html
@@ -1,0 +1,42 @@
+<!DOCTYPE html>
+<meta charset=utf-8>
+<title>HTML Test: Dynamic changes to base URLs</title>
+<link rel="author" title="Intel" href="http://www.intel.com/"/>
+<link rel="help" href="https://html.spec.whatwg.org/multipage/#dynamic-changes-to-base-urls" />
+<base href="" id="base">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<div id="div" style="display:none"></div>
+<script>
+      var div = document.getElementById("div"),
+          base = document.getElementById("base"),
+          url =  document.location.href;
+
+      var testData = [
+        {elements: ["a", "link", "area"], set: "href", get: "href"},
+        {elements: ["a", "area"], set: "ping", get: "ping"},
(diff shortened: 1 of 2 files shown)


html/infrastructure/urls/dynamic-changes-to-base-urls/dynamic-urls.sub.html
```diff
@@ -0,0 +1,42 @@
...
+
+<div id="div" style="display:none"></div>
+<script>
+      var div = document.getElementById("div"),
+          base = document.getElementById("base"),
+          url =  document.location.href;
+
+      var testData = [
+        {elements: ["a", "link", "area"], set: "href", get: "href"},
+        {elements: ["a", "area"], set: "ping", get: "ping"},
```
foolip: I ran the tests in Chrome and Firefox, and the ping tests fail in both. (All other tests pass.)

That's because the [ping attribute](https://html.spec.whatwg.org/multipage/links.html#ping) is a space-separated list of URLs, not a single URL. So the reflection is verbatim. The base URL presumably still has some effect for ping URLs, but it's not possible to test it just by reflection. I'd suggest just removing this line.

foolip (changes requested):
All looks good except for the ping problem.

html/infrastructure/urls/dynamic-changes-to-base-urls/dynamic-urls.sub.html
kaixinjxq: @foolip, thanks for your comments, I've addressed it in new commit, PTAL.

foolip (approved):

foolip (changes requested):
Looking to see if there might actually be some tests for this already, I found dynamic-urls.sub.xhtml, which this is based on.

AFAICT, dynamic-urls.sub.xhtml is wrong, as `xml:base` shouldn't be supported. I'd suggest deleting it, but Firefox passes and so there's a problem to be fixed.

Can you rename that test to historical.xhtml (in the same directory) and invert the pass conditions so that only Firefox fails? A link to https://github.com/whatwg/html/pull/84 in the test would be good, that's where `xml:base` was dropped from HTML.

foolip (approved):
I've checked that historical.sub.xhtml is now failing in Firefox and passing in Chrome, which seems right as Firefox still supports xml:base.

PR #12314, opened 2018-08-06
[css-contain] Update contain-paint-001 reference

mrego:
In #12270 we did some changes to the test suite
and modified contain-paint-001 test.
However we forget to modify the reference too.


fred-wang (approved):
LGTM

PR #12315, opened 2018-08-06
service worker: Add a couple of tests

wanghongjuan:
Add 5 checkpoints for attributes active/installing/waiting/scriptURL/navigationPreload and format relevant tests using es6:
- Verify the scriptURL property with query
- The ServiceWorker objects returned from installing attribute getter that represent the same service worker are the same objects.
- The ServiceWorker objects returned from waiting attribute getter that represent the same service worker are the same objects.
- The ServiceWorker objects returned from active attribute getter that represent the same service worker are the same objects.
- The navigationPreload attribute must return service worker registration's NavigationPreloadManager object.


service-workers/service-worker/active.https.html
```diff
@@ -5,51 +5,55 @@
...
+
 // "active" is set
-async_test(function(t) {
-    var step = t.step_func.bind(t);
-    var url = 'resources/empty-worker.js';
-    var scope = 'resources/blank.html';
-    var frame;
-    var registration;
+promise_test(async t => {
+    let step = t.step_func.bind(t);
```
mattto: Is step used?

service-workers/service-worker/active.https.html
```diff
@@ -5,51 +5,55 @@
...
+    let frame;
+
+    await service_worker_unregister(t, SCOPE);
+    frame = await with_iframe(SCOPE);
+    registration =
+        await navigator.serviceWorker.register(SCRIPT, {scope: SCOPE});
```
mattto: how about removing lines 11 and 16 and saying const frame = and const registration = here?

service-workers/service-worker/active.https.html
```diff
@@ -5,51 +5,55 @@
...
+    await wait_for_state(t, registration.installing, 'activating');
+    const container = await frame.contentWindow.navigator.serviceWorker;
+    assert_equals(
+        container.controller,
+        null,
+        'On activating state a document should not have a controller');
```
mattto: This description doesn't seem accurate. Even when the worker becomes activated the document should still not have a controller.

I would just make the description 'controller', so it's easy to know what assert fails.

service-workers/service-worker/active.https.html
```diff
@@ -5,51 +5,55 @@
...
+    assert_equals(
+        registration.active.scriptURL,
+        normalizeURL(SCRIPT),
+        'On activating state a document should have an active worker ');
```
mattto: Similarly, documents don't have acive/waiting/installing workers (that was an old version of the spec). I should just make the desription 'registration.active'. Ditto below.

service-workers/service-worker/active.https.html
```diff
@@ -5,51 +5,55 @@
...
-          assert_equals(
-            registration.installing,
-            null,
-            'On activating state a document should not have an installing ' +
-                'worker');
+    // FIXME: Add a test for a frame created after installation.
+    // Should the existing frame ("frame") block activation?
+    t.add_cleanup(() => frame.remove());
+    await registration.unregister();
+}, 'active is set');
```
mattto: service worker tests tend to use 2 space indents and 4 spaces for wrapping. Could we follow that here?

service-workers/service-worker/active.https.html
```diff
@@ -5,51 +5,55 @@
...
-      .then(step(function() {
-          frame.remove();
-          return service_worker_unregister_and_done(t, scope);
-        }))
-      .catch(unreached_rejection(t));
-  }, 'active is set');
+// Tests that the ServiceWorker objects returned from active attribute getter
+// that represent the same service worker are the same objects.
+promise_test(async t => {
+    registration =
```
mattto: const

service-workers/service-worker/active.https.html
```diff
@@ -5,51 +5,55 @@
...
+        await service_worker_unregister_and_register(t, SCRIPT, SCOPE);
+    await registration.active;
```
mattto: This await looks strange. active is not a promise.

service-workers/service-worker/active.https.html
```diff
@@ -5,51 +5,55 @@
...
+    const r = await navigator.serviceWorker.getRegistration(SCOPE);
+    assert_equals(r.scriptURl, registration.scriptURl, 'getRegistration ' +
+                  'should return the same object');
```
mattto: This isn't testing what the comments claim. It should be assert_equals(registration.active, r.active).

Ditto below.

service-workers/service-worker/serviceworkerobject-scripturl.https.html
```diff
@@ -7,25 +7,21 @@
 'use strict';
 
 function url_test(name, url) {
-  var scope = 'resources/scope/' + name;
-  var expectedURL = normalizeURL(url);
+    let scope = 'resources/scope/' + name;
+    let expectedURL = normalizeURL(url);
```
mattto: const?

service-workers/service-worker/navigationpreload.https.html
```diff
@@ -0,0 +1,18 @@
...
+    const SCRIPT = 'resources/empty-worker.js';
+    const SCOPE = 'resources/navigationpreload';
+    let registration =
+        await service_worker_unregister_and_register(t, SCRIPT, SCOPE);
+    let np = await registration.navigationPreload;
+    assert_true(np instanceof NavigationPreloadManager, 'navigationPreload ' +
+                'attribute return NavigationPreloadManager object');
+}, "The navigationPreload attribute must return service worker " +
+    "registration's NavigationPreloadManager object.");
+</script>
```
mattto: Could you move this to the service-worker/navigation-preload/ directory?

service-workers/service-worker/navigationpreload.https.html
```diff
@@ -0,0 +1,18 @@
...
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="resources/test-helpers.sub.js"></script>
+<script>
+promise_test(async t => {
+    const SCRIPT = 'resources/empty-worker.js';
+    const SCOPE = 'resources/navigationpreload';
+    let registration =
+        await service_worker_unregister_and_register(t, SCRIPT, SCOPE);
+    let np = await registration.navigationPreload;
```
mattto: I think registration.navigationPreload is not a promise

service-workers/service-worker/waiting.https.html
```diff
@@ -5,36 +5,40 @@
...
+        await navigator.serviceWorker.register(SCRIPT, {scope: SCOPE});
+    await wait_for_state(t, registration.installing, 'installed');
+    const controller =
+        await frame.contentWindow.navigator.serviceWorker.controller;
+
+    assert_equals(controller, null, '.controller');
+    assert_equals(registration.active, null, '.active');
+    assert_equals(registration.waiting.state, 'installed', '.waiting');
+    assert_equals(registration.installing, null, '.installing');
+    t.add_cleanup(() => frame.remove());
```
mattto: It's better to remove the frame before calling unregister, to help the agent clean up.

mfalken (changes requested):
Thanks for the tests!

service-workers/service-worker/navigationpreload.https.html
wanghongjuan: @mattto Many thanks for your review, I submit a new commit to address your comments, please take a look.

service-workers/service-worker/installing.https.html
```diff
@@ -5,33 +5,38 @@
...
+  await service_worker_unregister_and_done(t, SCOPE);
 }, 'installing is set');
+
+// Tests that The ServiceWorker objects returned from installing attribute getter
+// that represent the same service worker are the same objects.
+promise_test(async t => {
+  const registration =
+      await service_worker_unregister_and_register(t, SCRIPT, SCOPE);
+  const r = await navigator.serviceWorker.getRegistration(SCOPE);
+  assert_equals(registration.installing, r.installing, 'getRegistration ' +
```
mattto: Btw, I think the spec requires registration1 == registration2, which makes this test not so valuable. But we can keep it for now.

service-workers/service-worker/installing.https.html
```diff
@@ -5,33 +5,38 @@
...
+                'should return the same object');
```
mattto: I think the description should be like `ServiceWorkerRegistration.active should return the same object`. getRegistration() should return the same object implies the test is register() == getRegistration().

service-workers/service-worker/installing.https.html
```diff
@@ -5,33 +5,38 @@
...
+  await service_worker_unregister_and_done(t, SCOPE);
+}, 'The ServiceWorker objects returned from installing attribute getter that ' +
```
mattto: installing -> active

service-workers/service-worker/navigation-preload/navigationpreload.https.html
```diff
@@ -0,0 +1,20 @@
...
+<script src="resources/helpers.js"></script>
+<script>
+promise_test(async t => {
+  const SCRIPT = '../resources/empty-worker.js';
+  const SCOPE = '../resources/navigationpreload';
+  const registration =
+      await service_worker_unregister_and_register(t, SCRIPT, SCOPE);
+  const np = registration.navigationPreload;
+  assert_true(np instanceof NavigationPreloadManager, 'navigationPreload ' +
+              'attribute return NavigationPreloadManager object');
```
mattto: return -> should return

service-workers/service-worker/navigationpreload.https.html
```diff
@@ -0,0 +1,18 @@
...
+<meta charset="utf-8">
+<title>ServiceWorker: navigator.serviceWorker.navigationPreload</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="resources/test-helpers.sub.js"></script>
+<script>
+promise_test(async t => {
+    const SCRIPT = 'resources/empty-worker.js';
+    const SCOPE = 'resources/navigationpreload';
+    let registration =
```
mattto: const

service-workers/service-worker/navigationpreload.https.html
mattto: I'm a little confused by github's review tool, but I think you haven't yet fixed the `await` here. It's not needed since navigationPreload isn't a promise. In fact I'd just remove `np` and test registration.navigationPreload directly (I tend to avoid abbreviations).

service-workers/service-worker/navigationpreload.https.html
mattto: This one too.

service-workers/service-worker/serviceworkerobject-scripturl.https.html
```diff
@@ -7,25 +7,21 @@
...
-            assert_equals(worker.scriptURL, expectedURL,
-                          'Returned ServiceWorker object should have scriptURL');
-          });
-    }, 'Verify the scriptURL property: ' + name);
+  promise_test(async t => {
+    const registration =
+        await service_worker_unregister_and_register(t, url, scope);
+    const worker = registration.installing;
+    assert_equals(worker.scriptURL, expectedURL,
+                  'Returned ServiceWorker object should have scriptURL');
```
mattto: I'm not sure this description explains things. Maybe just call it 'scriptURL'.

service-workers/service-worker/waiting.https.html
```diff
@@ -5,36 +5,37 @@
...
+  frame.remove();
+  await service_worker_unregister_and_done(t, SCOPE);
+}, 'waiting is set after installation');
+
+// Tests that the ServiceWorker objects returned from waiting attribute getter
+// that represent the same service worker are the same objects.
+promise_test(async t => {
+  const registration =
+      await service_worker_unregister_and_register(t, SCRIPT, SCOPE);
+  const r = await navigator.serviceWorker.getRegistration(SCOPE);
```
mattto: ditto with registration vs r

service-workers/service-worker/waiting.https.html
```diff
@@ -5,36 +5,37 @@
...
+  assert_equals(registration.waiting, r.waiting, 'getRegistration ' +
+                'should return the same object');
```
mattto: ditto with description here

mfalken (changes requested):
Thank you, some more comments.

service-workers/service-worker/installing.https.html
wanghongjuan: Yes, but in this case, the spec requires the `installing` attribute should return `ServiceWorker` object instead of `ServiceWokerRegistration` object, how about using `registration1.installing == registration2.installing`?

service-workers/service-worker/installing.https.html
wanghongjuan: Thanks for your clarification, updated to `ServiceWorkerRegistration.installing` in assertion as we have a complete description in test body.

service-workers/service-worker/installing.https.html
wanghongjuan: This is installing.https.html, so we don't need change this, am i right?

service-workers/service-worker/navigationpreload.https.html
wanghongjuan: This test should be move to directory `navigation-preload` and this problem should not exist

service-workers/service-worker/navigation-preload/navigationpreload.https.html
wanghongjuan: Updated, please review my latest commit, thanks a lot.

service-workers/service-worker/navigationpreload.https.html
mattto: Oh I see. I think you need to remove this file from the pull request maybe?

service-workers/service-worker/navigation-preload/navigationpreload.https.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
```
mattto: Can we call this navigationPreload.https.html.

service-workers/service-worker/waiting.https.html
```diff
@@ -5,36 +5,40 @@
...
+  const controller = frame.contentWindow.navigator.serviceWorker.controller;
+
+  assert_equals(controller, null, 'controller');
+  assert_equals(registration.active, null, 'registration.active');
+  assert_equals(registration.waiting.state, 'installed',
+                'registration.waiting');
+  assert_equals(registration.installing, null, 'registration.installing');
+  t.add_cleanup(async() => {
+    frame.remove();
+    await service_worker_unregister_and_done(t, SCOPE);
```
mattto: could we make the cleanup function like this:

t.add_cleanup(async() => {
  if (frame)
    frame.remove();
  if (registration)
    await registration.unregister();
});

It'd be better if it's moved to the top of the function so in case an assert fails, the cleanup is still run. I'm not sure if it would have to go after registration and frame are defined.

service-workers/service-worker/waiting.https.html
```diff
@@ -5,36 +5,40 @@
...
+// Tests that the ServiceWorker objects returned from waiting attribute getter
+// that represent the same service worker are the same objects.
+promise_test(async t => {
+  const registration1 =
+      await service_worker_unregister_and_register(t, SCRIPT, SCOPE);
+  const registration2 = await navigator.serviceWorker.getRegistration(SCOPE);
+  assert_equals(registration1.waiting, registration2.waiting,
+                'ServiceWorkerRegistration.waiting should return the same ' +
+                'object');
+  await service_worker_unregister_and_done(t, SCOPE);
```
mattto: await registration1.unregister() is good enough

service-workers/service-worker/serviceworkerobject-scripturl.https.html
```diff
@@ -7,25 +7,20 @@
...
-            assert_equals(worker.scriptURL, expectedURL,
-                          'Returned ServiceWorker object should have scriptURL');
-          });
-    }, 'Verify the scriptURL property: ' + name);
+  promise_test(async t => {
+    const registration =
+        await service_worker_unregister_and_register(t, url, scope);
+    const worker = registration.installing;
+    assert_equals(worker.scriptURL, expectedURL, 'scriptURL');
+    await service_worker_unregister_and_done(t, scope);
```
mattto: await registration.unregister()

service-workers/service-worker/installing.https.html
```diff
@@ -5,33 +5,40 @@
...
+  assert_equals(container.controller, null, 'controller');
+  assert_equals(registration.active, null, 'registration.active');
+  assert_equals(registration.waiting, null, 'registration.waiting');
+  assert_equals(registration.installing.scriptURL, normalizeURL(SCRIPT),
+                'registration.installing.scriptURL');
+  // FIXME: Add a test for a frame created after installation.
+  // Should the existing frame ("frame") block activation?
+  t.add_cleanup(async() => {
+    frame.remove();
+    await service_worker_unregister_and_done(t, SCOPE);
```
mattto: await registration.unregister()

service-workers/service-worker/installing.https.html
```diff
@@ -5,33 +5,40 @@
...
+// Tests that The ServiceWorker objects returned from installing attribute getter
+// that represent the same service worker are the same objects.
+promise_test(async t => {
+  const registration1 =
+      await service_worker_unregister_and_register(t, SCRIPT, SCOPE);
+  const registration2 = await navigator.serviceWorker.getRegistration(SCOPE);
+  assert_equals(registration1.installing, registration2.installing,
+                'ServiceWorkerRegistration.installing should return the ' +
+                'same object');
+  await service_worker_unregister_and_done(t, SCOPE);
```
mattto: await registration1.unregister()

service-workers/service-worker/active.https.html
```diff
@@ -5,51 +5,42 @@
...
+// Tests that the ServiceWorker objects returned from active attribute getter
+// that represent the same service worker are the same objects.
+promise_test(async t => {
+  const registration1 =
+      await service_worker_unregister_and_register(t, SCRIPT, SCOPE);
+  const registration2 = await navigator.serviceWorker.getRegistration(SCOPE);
+  assert_equals(registration1.active, registration2.active,
+                'ServiceWorkerRegistration.active should return the same ' +
+                'object');
+  await service_worker_unregister_and_done(t, SCOPE);
```
mattto: await registration1.unregister()

service-workers/service-worker/active.https.html
```diff
@@ -5,51 +5,42 @@
...
-          assert_equals(
-            registration.installing,
-            null,
-            'On activating state a document should not have an installing ' +
-                'worker');
+  // FIXME: Add a test for a frame created after installation.
+  // Should the existing frame ("frame") block activation?
+  t.add_cleanup(async() => {
+    frame.remove();
+    await service_worker_unregister_and_done(t, SCOPE);
```
mattto: if (frame)
  frame.remove();
if (registration)
  await registration.unregister()

can the cleanup move to the top of function so it's still executed if an assert fails? I'm not sure if it has to go after frame and registration are defined.

mfalken (changes requested):
Thanks! lgtm once these comments are addressed

service-workers/service-worker/active.https.html
wanghongjuan: I tried as your suggestion, moved the cleanup to the top of function, it's still executed if an assertion fails.  All of comments are addressed in my latest commit.

service-workers/service-worker/active.https.html
```diff
@@ -5,51 +5,46 @@
...
-async_test(function(t) {
-    var step = t.step_func.bind(t);
-    var url = 'resources/empty-worker.js';
-    var scope = 'resources/blank.html';
-    var frame;
-    var registration;
+promise_test(async t => {
+
+  t.add_cleanup(async() => {
+    if(frame)
```
mattto: add a space between `if` and the opening paren (ditto throughout)

service-workers/service-worker/installing.https.html
```diff
@@ -5,33 +5,44 @@
...
-        var container = frame.contentWindow.navigator.serviceWorker;
-        assert_equals(container.controller, null);
-        assert_equals(registration.active, null);
-        assert_equals(registration.waiting, null);
-        assert_equals(registration.installing.scriptURL, normalizeURL(url));
+  t.add_cleanup(async() => {
+    if(frame)
+      frame.remove();
+    if(registration)
+    await registration.unregister();
```
mattto: indent this line so it's inside the if()

mfalken (approved):
Thank you! Approved with some formatting nits.

service-workers/service-worker/installing.https.html
wanghongjuan: Thanks for your correction, fixed the formatting error in newest commit.

PR #12316, opened 2018-08-06
Test for 304 updates on Content-* headers

mnot:
Some browsers seem to exempt Content-* headers from 304 updates; this is to see how widespread that is.

### DIFF
--- a/fetch/http-cache/304-update.html
+++ b/fetch/http-cache/304-update.html
@@ -68,6 +68,59 @@
         ]
       },
       {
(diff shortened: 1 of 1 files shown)


foolip:
@mnot looks like you didn't get any review on this and now it conflicts. Can you rebase? If you still don't get review you'll need to ping people, there's no fallback that you can depend on kicking in after a while.

mnot:
This is overcome by events; see https://cache-tests.fyi/#update304

PR #12327, opened 2018-08-06
SpeechSynthesis.speak requires activation

csharrison:
To be merged with https://github.com/w3c/speech-api/pull/28 if it ends up with normative changes involving user activation.

CC @foolip


speech-api/SpeechSynthesis-speak-ownership.html
```diff
@@ -1,10 +1,14 @@
 <!doctype html>
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
+<script src="/resources/testdriver.js"></script>
+<script src="/resources/testdriver-vendor.js"></script>
 <iframe></iframe>
 <script>
 // using an utterance for different SpeechSynthesis instances should throw
-test(t => {
+var t = async_test(
```
foolip: The previous `test(f(t), name)` structure works for `async_test` as well, just change `test` to `async_test`.

speech-api/SpeechSynthesis-speak-ownership.html
```diff
@@ -1,10 +1,14 @@
...
+  'Using the same SpeechSynthesisUtterance with two SpeechSynthesis instances');
+test_driver.click(document.body).then(t.step_func(() => {
```
foolip: If you use `test_driver.bless`, this test will be easier to run manually. (Caveat: I haven't tried it myself.)

speech-api/SpeechSynthesis-speak-without-activation-fails.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<p>Some body text.
+<script>
+var t = async_test('speechSynthesis.speak requires user activation');
+const utter = new SpeechSynthesisUtterance('1');
+utter.onerror = t.step_func_done((e) => {
+  assert_equals(e.name, "not-allowed");
```
foolip: Great test, thank you!

speech-api/SpeechSynthesis-speak-without-activation-fails.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<p>Some body text.
+<script>
+var t = async_test('speechSynthesis.speak requires user activation');
+const utter = new SpeechSynthesisUtterance('1');
```
foolip: Can you use the `async_test(f(n), name)` structure here as well? The reason is that if `new SpeechSynthesisUtterance('1')` throws an exception in the current structure, it would be a harness error plus a test that times out, instead of a failing test. Putting all the interesting code inside the first callback or in `t.step_func` for callbacks avoids this kind of problem.

speech-api/SpeechSynthesis-speak-without-activation-fails.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
```
foolip: As discussed in https://github.com/w3c/speech-api/pull/28#issuecomment-410660492, maybe make this test tenatative (name it "SpeechSynthesis-speak-without-activation-fails.tentative.html") and link to an open issue in the speech-api repo.

speech-api/SpeechSynthesis-speak-ownership.html
csharrison: Done.

speech-api/SpeechSynthesis-speak-ownership.html
csharrison: Done.

speech-api/SpeechSynthesis-speak-without-activation-fails.html
csharrison: :)

speech-api/SpeechSynthesis-speak-without-activation-fails.html
csharrison: Done.

csharrison:
Thanks for the review, Philip!

speech-api/SpeechSynthesis-speak-ownership.html
```diff
@@ -1,19 +1,23 @@
...
-  const iframe = document.querySelector('iframe');
-  assert_not_equals(speechSynthesis, iframe.contentWindow.speechSynthesis);
-  speechSynthesis.speak(utter);
-  // the spec doesn't say what exception to throw:
-  // https://github.com/w3c/speech-api/issues/8
-  assert_throws(null, () => {
-    iframe.contentWindow.speechSynthesis.speak(utter);
-  });
+async_test(t => {
+  test_driver.bless('speechSynthesis.speak', t.step_func(() => {
```
foolip: Oops, looks like this test will never finish, using `t.step_func_done` here would fix it.

speech-api/SpeechSynthesis-speak-twice.html
```diff
@@ -1,17 +1,22 @@
 <!doctype html>
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
+<script src="/resources/testdriver.js"></script>
+<script src="/resources/testdriver-vendor.js"></script>
+<p>Some body text.
```
foolip: You probably don't need this text now that you're using `test_driver.bless` instead. Ditto in following tests.

foolip (approved):
LGTM % nits

speech-api/SpeechSynthesis-speak-twice.html
csharrison: Actually we still need it, since bless requires the document body to be around.

csharrison:
Hey Philip, is there anything else blocking merging this and the associated spec PR?

speech-api/SpeechSynthesis-speak-twice.html
foolip: Ah, if that's all you need then `<body>` would be clearer. Plenty of tests to that ahead of doing something with `document.body`.

foolip:
I made the `<body>` changes? Can you check that the tests still work? Then I'll merge.

csharrison:
I confirmed the tests are working properly, thanks for the quick fixes.

PR #12330, opened 2018-08-07
Fix worklet referrer tests to match spec

domfarolino:
This PR fixes worklet referrer tests to match the spec. Basically this reverts a good bit of https://github.com/web-platform-tests/wpt/commit/50b0943d82d36bb92977002bebd5a7990e0ed388#diff-f5dd3dfe0ce7dd53814287bc30b9a197 (the WPTs bit of https://chromium-review.googlesource.com/c/chromium/src/+/1111743) to ensure that we actually do test that the referrer policy of the ["outer settings object"](https://drafts.css-houdini.org/worklets/#dom-worklet-addmodule) is honored when fetching worklet scripts. Related PR for module worker referrer tests: https://github.com/web-platform-tests/wpt/pull/12321.

/cc @nhiroki @nyaxt @hiroshige-g (Note: when this is merged and auto-rolls into Blink, I'll submit my CL that allows Blink to conform to these tests).

### DIFF
--- a/worklets/resources/referrer-checker.py
+++ b/worklets/resources/referrer-checker.py
@@ -2,14 +2,22 @@
 # |expected_referrer|.
 def main(request, response):
     referrer = request.headers.get("referer", "")
+    referrer_policy = request.GET.first("referrer_policy")
     expected_referrer = request.GET.first("expected_referrer", "")
-
     response_headers = [("Content-Type", "text/javascript"),
                         ("Access-Control-Allow-Origin", "*")]
 
-    # The expected referrer doesn't contain query params for simplification, so
-    # we check the referrer by startswith() here.
-    if (expected_referrer != "" and
-        referrer.startswith(expected_referrer + "?")):
-        return (200, response_headers, "")
+    if referrer_policy == "no-referrer" or referrer_policy == "origin":
+        if referrer == expected_referrer:
+            return (200, response_headers, "")
+        return (404, response_headers)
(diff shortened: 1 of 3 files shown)


worklets/resources/referrer-checker.py
```diff
@@ -2,14 +2,23 @@
...
+    if referrer_policy == "no-referrer" or referrer_policy == "origin":
+        if referrer == expected_referrer:
+            return (200, response_headers, "")
+        return (404, response_headers)
+
+    if referrer_policy == "same-origin":
+        if referrer == expected_referrer:
+            return (200, response_headers, "")
+        # The expected referrer doesn't contain query params for simplification,
+        # so we check the referrer by startswith() here.
```
nyaxt: Optional (since this isn't really introduced by this PR). I'm not a big fan of this logic. How about gating this on request.GET.first("expected_referrer_prefix")?

nyaxt (approved):

domfarolino:
Ok so the reason for that logic was that the referrer header might contain query params and the expected referrer does not. What I've done is made the `referrer` variable equal only to the portion of the referrer header without query params.

When we compare it later on, we don't have to do any "`expected_referrer + ?`"ing. This is only relevant to the (same-origin referrer policy && same-origin request) case, so I've simplified the logic there. This has no effect on other cases, since `.split("?")[0]` is safe when "?" is not present.
Merging; if any additional feedback on this change please LMK.

PR #12341, opened 2018-08-07
Don't try to re-encode already-encoded bytes when serializing.

jdm:
This unbreaks Servo's syncing process when strings like https://github.com/web-platform-tests/wpt/blob/5c219ef55a4b2099c0f4c53b9669df10d73e6f3a/url/urlsearchparams-sort.any.js#L31 are encountered during manifest serialization.


jgraham (approved):

jgraham:
Thanks!

PR #12343, opened 2018-08-07
[html] Verify synchronicity of beforeunload event

jugglinmike:
After reviewing all the files that reference "beforeunload" (as reported by `git grep -l beforeunload`), I can't find coverage for this behavior. There are a number of ways to observe this behavior. The choice of `location.href` is arbitrary, and I'm open to suggestion on using some other API either in place of `location.href` or in addition to it.

Browser     | Result
------------|-------
Chrome 68   | :x: 
Edge 17     | :x: 
Firefox 61  | :heavy_check_mark: 
IE 11       | :heavy_check_mark: 
Safari 11.1 | :x:

This came up in the context of gh-12290, where I authored a test to support asynchronous behavior. A number of tests in WPT are written in this way, but they generally concern other details of the specification, so I suppose it's acceptable for them to be loose in this regard.

### DIFF
--- a/html/browsers/browsing-the-web/unloading-documents/beforeunload-synchronous.html
+++ b/html/browsers/browsing-the-web/unloading-documents/beforeunload-synchronous.html
@@ -1,0 +1,33 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>beforeunload event is emitted synchronously</title>
+<link rel="help" href="https://html.spec.whatwg.org/multipage/webappapis.html#the-event-handler-processing-algorithm">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<body>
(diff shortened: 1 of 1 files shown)


bzbarsky (approved):

jugglinmike:
@foolip This test demonstrates a discrepancy between the spec and Chrome's behavior. Would you mind taking a look?

zcorpan (approved):

jugglinmike:
@foolip I was hoping to get a Chromium dev's perspective before landing this, but the spec is clear, so maybe that was overly conservative of me. In any event, I'm still interested in learning whether this is an intentional violation and seeing it fixed (or the spec patched), so I've followed up with a report on Chromium's issue tracker:

https://bugs.chromium.org/p/chromium/issues/detail?id=873683

PR #12350, opened 2018-08-08
[Fetch] Object.prototype shouldn't interfere with fetch streams

ricea:
Test that setting Object.prototype.start to a function that throws
doesn't interfere with fetch operations that create a stream. Similarly,
verify that setting throwing accessors for 'type', 'size', and
'highWaterMark' doesn't cause problems.

See https://github.com/whatwg/fetch/pull/781 for background.

### DIFF
--- a/fetch/api/basic/stream-safe-creation.any.js
+++ b/fetch/api/basic/stream-safe-creation.any.js
@@ -1,3 +1,5 @@
+// META: global=worker
+
 // These tests verify that stream creation is not affected by changes to
 // Object.prototype.
 
@@ -19,7 +21,10 @@ for (creationCase of Object.keys(creationCases)) {
         get() { throw Error(`Object.prototype.${accessorName} was accessed`); },
         configurable: true
       });
-      t.add_cleanup(() => delete Object.prototype[accessorName]);
+      t.add_cleanup(() => {
+        delete Object.prototype[accessorName];
+        return Promise.resolve();
+      });
       await creationCases[creationCase]();
     }, `throwing Object.prototype.${accessorName} accessor should not affect ` +
        `stream creation by '${creationCase}'`);
@@ -28,7 +33,10 @@ for (creationCase of Object.keys(creationCases)) {
       // -1 is a convenient value which is invalid, and should cause the
       // constructor to throw, for all four fields.
       Object.prototype[accessorName] = -1;
-      t.add_cleanup(() => delete Object.prototype[accessorName]);
+      t.add_cleanup(() => {
+        delete Object.prototype[accessorName];
+        return Promise.resolve();
+      });
       await creationCases[creationCase]();
     }, `Object.prototype.${accessorName} accessor returning invalid value ` +
        `should not affect stream creation by '${creationCase}'`);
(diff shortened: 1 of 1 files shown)


fetch/api/basic/stream-safe-creation.any.js
```diff
@@ -0,0 +1,33 @@
...
+  consumeNonEmpty: () => new Response(new Uint8Array([64])).text()
+};
+
+for (creationCase of Object.keys(creationCases)) {
+  promise_test(async t => {
+    Object.prototype.start = () => {
+      throw Error('Object.prototype.start was called');
+    };
+    t.add_cleanup(() => delete Object.prototype.start);
+    await creationCases[creationCase];
```
jyasskin: Should this and the same line below be `creationCases[creationCase]()`?

fetch/api/basic/stream-safe-creation.any.js
```diff
@@ -0,0 +1,33 @@
...
+  fetch: async () => fetch(location.href),
+  request: () => new Request(location.href),
+  response: () => new Response('hi'),
+  consumeEmpty: () => new Response().text(),
+  consumeNonEmpty: () => new Response(new Uint8Array([64])).text()
+};
+
+for (creationCase of Object.keys(creationCases)) {
+  promise_test(async t => {
+    Object.prototype.start = () => {
```
jyasskin: Why make this one throw on call, when .size() throws on get()? It seems like .start shouldn't be gotten by the `CreateReadableStream()` version either.

fetch/api/basic/stream-safe-creation.any.js
```diff
@@ -0,0 +1,33 @@
+// These tests verify that stream creation is not affected by changes to
+// Object.prototype.
+
+const creationCases = {
+  fetch: async () => fetch(location.href),
+  request: () => new Request(location.href),
```
jyasskin: I think I'm missing where [Request(url)](https://fetch.spec.whatwg.org/#dom-request) constructs a ReadableStream.

jyasskin (approved):

fetch/api/basic/stream-safe-creation.any.js
ricea: This is an important problem! Fixed.

fetch/api/basic/stream-safe-creation.any.js
ricea: That was indeed not very useful. I replaced it with a version which errors the stream, which is nice because the side-effect is delayed, so if the implementation is (accidentally?) eating exceptions from the constructor, it will still break if it tries to read from the stream later.

fetch/api/basic/stream-safe-creation.any.js
ricea: IIRC, it actually does in Blink. But I changed the test case to one which always constructs a ReadableStream by the standard.

gsnedders:
AFAIK, @jyasskin's feedback has been addressed. Can this be re-reviewed/landed as needed?

annevk (approved):

PR #12359, opened 2018-08-08
HTML: Add checks for the return value of document.open()

TimothyGu:
Edge currently fails some of these as they return `null` instead of the document sometimes.


annevk (approved):
This is great, thanks! Please file an issue against Edge.

TimothyGu:
@annevk Edge is already aware of this problem but will do!

TimothyGu:
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/18605326/

PR #12360, opened 2018-08-08
[wptrunner] Make test skipping explicit

jugglinmike:
Explicitly report on skipped "jsshell" tests during execution. This
gives users an indication that the behavior is intentional and promotes
parity with the output generated by the `--list-tests` flag of the `wpt
run.  command.

---

@jgraham @Ms2ger This removes a comment that states "we silently skip them," so
I expect there is a good reason they aren't being documented as "SKIP". In
that case, I can look in to updating the `--list-tests` feature to correctly
support the new feature. But could we also extend this comment to explain why
this needs to be done "silently"?

### DIFF
--- a/tools/wptrunner/wptrunner/wptrunner.py
+++ b/tools/wptrunner/wptrunner/wptrunner.py
@@ -246,7 +246,7 @@ def run_tests(config, test_paths, product, **kwargs):
                         run_tests = {"testharness": []}
                         for test in test_loader.tests["testharness"]:
                             if (test.testdriver and not executor_cls.supports_testdriver) or (
-                                test.jsshell and not executor_cls.supports_jsshell):
+                                    test.jsshell and not executor_cls.supports_jsshell):
                                 logger.test_start(test.id)
                                 logger.test_end(test.id, status="SKIP")
                             else:
--- a/tools/wptrunner/wptrunner/wptrunner.py
+++ b/tools/wptrunner/wptrunner/wptrunner.py
@@ -245,15 +245,10 @@ def run_tests(config, test_paths, product, **kwargs):
                     if test_type == "testharness":
                         run_tests = {"testharness": []}
                         for test in test_loader.tests["testharness"]:
-                            if test.testdriver and not executor_cls.supports_testdriver:
+                            if (test.testdriver and not executor_cls.supports_testdriver) or (
+                                test.jsshell and not executor_cls.supports_jsshell):
                                 logger.test_start(test.id)
                                 logger.test_end(test.id, status="SKIP")
-                            elif test.jsshell and not executor_cls.supports_jsshell:
-                                # We expect that tests for JavaScript shells
-                                # will not be run along with tests that run in
-                                # a full web browser, so we silently skip them
-                                # here.
-                                pass
                             else:
                                 run_tests["testharness"].append(test)
                     else:


tools/wptrunner/wptrunner/wptrunner.py
```diff
@@ -245,15 +245,10 @@ def run_tests(config, test_paths, product, **kwargs):
                     if test_type == "testharness":
                         run_tests = {"testharness": []}
                         for test in test_loader.tests["testharness"]:
-                            if test.testdriver and not executor_cls.supports_testdriver:
+                            if (test.testdriver and not executor_cls.supports_testdriver) or (
+                                test.jsshell and not executor_cls.supports_jsshell):
```
Hexcles: This line needs to be intended one more level.

Hexcles (commented):
Consistency is key.

If we want to *silently* skip jshell tests, we also need to silently omit them in `--list-tests`, i.e. to make `--list-tests` aware of the executor. Otherwise, we need to skip jshell tests *explicitly* as what this PR does.

Since I don't have enough context on the use case of jsshell (there's no executor with `supports_jsshell = True`, so I assume it's used out of the tree), I don't know which approach should be preferred.

tools/wptrunner/wptrunner/wptrunner.py
jugglinmike: That's for clarity, right? It's functional on my system as written

Hexcles: That's for PEP8 compliance and to make the linter happy :)

jugglinmike: I would have been satisfied if it was just to make Robert happy :P

jugglinmike:
I agree about the importance of consistency. However, I'm starting to think that "SKIP" may not be the appropriate semantic here. It looks like @Ms2ger initially implemented this feature to report these tests as "SKIP", but during the review, @jgraham [wrote](https://github.com/web-platform-tests/wpt/pull/10769#discussion_r185803558):

> Let's just ignore these rather than print a "skip" message; we expect a disjoint set of jsshell and non-jsshell executors.

You could argue that in the absence of a suitable executor, these tests don't exist from WPT's perspective. That may be why this was implemented as a "silent" skip. If so, I believe we'll need to implement this at a lower level, specifically within the test loader.

[This is important for wpt.fyi](https://github.com/web-platform-tests/results-collection/issues/590), so I've submitted a patch along those lines as an alternative to this one: gh-12368. Separately (and a bit out of order), I've filed an issue report about the problem (gh-12366). That way we can track the issue independently of any particular solution.

Ms2ger (approved):
Since this was my initial approach, this is fine by me. Let's give @jgraham a chance to respond, though.

jgraham:
My argument for not showing them as skip is that they *can't* run. For example showwing them as skipped in wpt.fyi is misleading because it's not a kind of test that we can enable in browsers.

This is obviously all dancing around the fundamental problem here which is that they should just be a new test type that isn't supported by browser implementations, but that's a bit hard. In the meantime I think the alternate approach is more helpful.

jgraham:
(GitHub doesn't have an "I think this patch is a bad idea" button, so I'm going to go ahead and close the issue, on the assumption that people won't disagree with me ;)

PR #12367, opened 2018-08-09
[wptrunner] Remove unused code

jugglinmike:
This file is not structured for direct execution, and it does not appear
to be reference by any other file in the wptrunner project. In addition,
it includes incorrect references to the `TestLoader` and `TestFilter`
classes, and it invokes their constructors with arguments that do not
match their supported parameters.


jugglinmike:
Amended to add:

    Because no other code invokes the `wptrunner.run_tests` function with
    the `test_loader` keyword argument, the associated code path may also be
    removed.

jgraham:
So this is accessible from a `mach` command in mozilla-central. But as you note the code bitrotted and no one noticed, so I think it makes sense to remove this.

jgraham (approved):

gsnedders:
@jgraham any reason not to have merged this?

PR #12368, opened 2018-08-09
[wptrunner] Consistently hide tests

jugglinmike:
Ensure the "jsshell" tests are not reported by any of the commands that
interface with the `TestLoader`.

---

@jgraham The `wpt_run` function optionally accepts a `TestLoader` instance as a keyword argument. I was only able to find one case where that is used: `reduce.py`. This patch breaks that usage, but I think it may be dead code, so I'm making a case for its removal in gh-12367.

This is intended to resolve gh-12366

### DIFF
--- a/tools/wptrunner/wptrunner/executors/base.py
+++ b/tools/wptrunner/wptrunner/executors/base.py
@@ -108,6 +108,14 @@ class TestExecutor(object):
     test_type = None
     convert_result = None
     supports_testdriver = False
+
+    # `supports_jsshell` determines whether a given test is enumerated in the
+    # various test-traversal commands exposed by the wptrunner CLI (see the
+    # `ignores` method). It is intended to be overridden by subclasses which
+    # are defined outside of WPT.
+    #
+    # Note that this interpretation of "supports" is distinct from that of the
+    # `supports_testdriver` attribute.
     supports_jsshell = False
 
     def __init__(self, browser, server_config, timeout_multiplier=1,
@@ -182,6 +190,16 @@ class TestExecutor(object):
 
     def test_url(self, test):
         return urlparse.urljoin(self.server_url(test.environment["protocol"]), test.url)
+
+    @classmethod
+    def ignores(cls, test):
+        """Determine if a given test should be omitted during test
+        traversal."""
+
+        if cls.supports_jsshell:
+            return False
+
+        return test.test_type == "testharness" and test.jsshell
 
     @abstractmethod
     def do_test(self, test):
--- a/tools/wptrunner/wptrunner/testloader.py
+++ b/tools/wptrunner/wptrunner/testloader.py
@@ -467,7 +467,8 @@ class TestLoader(object):
                  total_chunks=1,
                  chunk_number=1,
                  include_https=True,
-                 skip_timeout=False):
+                 skip_timeout=False,
+                 executor_classes=None):
 
         self.test_types = test_types
         self.run_info = run_info
@@ -480,6 +481,7 @@ class TestLoader(object):
         self.disabled_tests = None
         self.include_https = include_https
         self.skip_timeout = skip_timeout
+        self.executor_classes = executor_classes
 
         self.chunk_type = chunk_type
         self.total_chunks = total_chunks
@@ -562,6 +564,11 @@ class TestLoader(object):
                  "disabled":defaultdict(list)}
 
         for test_path, test_type, test in self.iter_tests():
+            executor_class = (self.executor_classes and
+                              self.executor_classes.get(test_type))
+            if executor_class and executor_class.ignores(test):
+                continue
+
             enabled = not test.disabled()
             if not self.include_https and test.environment["protocol"] == "https":
                 enabled = False
--- a/tools/wptrunner/wptrunner/wptrunner.py
+++ b/tools/wptrunner/wptrunner/wptrunner.py
@@ -41,7 +41,8 @@ def setup_logging(*args, **kwargs):
     logger = wptlogging.setup(*args, **kwargs)
 
 
-def get_loader(test_paths, product, debug=None, run_info_extras=None, **kwargs):
+def get_loader(test_paths, product, executor_classes, debug=None,
+               run_info_extras=None, **kwargs):
     if run_info_extras is None:
         run_info_extras = {}
 
@@ -74,16 +75,19 @@ def get_loader(test_paths, product, debug=None, run_info_extras=None, **kwargs):
                                         total_chunks=kwargs["total_chunks"],
                                         chunk_number=kwargs["this_chunk"],
                                         include_https=ssl_enabled,
-                                        skip_timeout=kwargs["skip_timeout"])
+                                        skip_timeout=kwargs["skip_timeout"],
+                                        executor_classes=executor_classes)
     return run_info, test_loader
 
 
 def list_test_groups(test_paths, product, **kwargs):
     env.do_delayed_imports(logger, test_paths)
 
-    run_info_extras = products.load_product(kwargs["config"], product)[-1](**kwargs)
+    product_data = products.load_product(kwargs["config"], product)
+    executor_classes = product_data[3]
+    run_info_extras = product_data[-1](**kwargs)
 
-    run_info, test_loader = get_loader(test_paths, product,
+    run_info, test_loader = get_loader(test_paths, product, executor_classes,
                                        run_info_extras=run_info_extras, **kwargs)
 
     for item in sorted(test_loader.groups(kwargs["test_types"])):
@@ -95,9 +99,11 @@ def list_disabled(test_paths, product, **kwargs):
 
     rv = []
 
-    run_info_extras = products.load_product(kwargs["config"], product)[-1](**kwargs)
+    product_data = products.load_product(kwargs["config"], product)
+    executor_classes = product_data[3]
+    run_info_extras = product_data[-1](**kwargs)
 
-    run_info, test_loader = get_loader(test_paths, product,
+    run_info, test_loader = get_loader(test_paths, product, executor_classes,
                                        run_info_extras=run_info_extras, **kwargs)
 
     for test_type, tests in test_loader.disabled_tests.iteritems():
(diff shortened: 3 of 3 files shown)


jgraham (approved):

tools/wptrunner/wptrunner/testloader.py
```diff
@@ -467,7 +467,8 @@ def __init__(self,
                  total_chunks=1,
                  chunk_number=1,
                  include_https=True,
-                 skip_timeout=False):
+                 skip_timeout=False,
+                 executor_classes={}):
```
Ms2ger: Mutable objects in default arguments are not great.

tools/wptrunner/wptrunner/testloader.py
```diff
@@ -562,6 +564,11 @@ def _load_tests(self):
                  "disabled":defaultdict(list)}
 
         for test_path, test_type, test in self.iter_tests():
+            if test.jsshell:
```
Ms2ger: This doesn't work:

```
    wpt.main()
  File "/home/travis/build/web-platform-tests/wpt/tools/wpt/wpt.py", line 132, in main
    rv = script(*args, **kwargs)
  File "/home/travis/build/web-platform-tests/wpt/tools/wpt/run.py", line 453, in run
    rv = run_single(venv, **kwargs) > 0
  File "/home/travis/build/web-platform-tests/wpt/tools/wpt/run.py", line 460, in run_single
    return wptrunner.start(**kwargs)
  File "/home/travis/build/web-platform-tests/wpt/tools/wptrunner/wptrunner/wptrunner.py", line 334, in start
    return not run_tests(**kwargs)
  File "/home/travis/build/web-platform-tests/wpt/tools/wptrunner/wptrunner/wptrunner.py", line 170, in run_tests
    **kwargs)
  File "/home/travis/build/web-platform-tests/wpt/tools/wptrunner/wptrunner/wptrunner.py", line 79, in get_loader
    executor_classes=executor_classes)
  File "/home/travis/build/web-platform-tests/wpt/tools/wptrunner/wptrunner/testloader.py", line 500, in __init__
    self._load_tests()
  File "/home/travis/build/web-platform-tests/wpt/tools/wptrunner/wptrunner/testloader.py", line 567, in _load_tests
    if test.jsshell:
AttributeError: 'ReftestTest' object has no attribute 'jsshell'
```

tools/wptrunner/wptrunner/testloader.py
```diff
@@ -467,7 +467,8 @@ def __init__(self,
                  total_chunks=1,
                  chunk_number=1,
                  include_https=True,
-                 skip_timeout=False):
+                 skip_timeout=False,
+                 executor_classes={}):
 
         self.test_types = test_types
```
Ms2ger: This class should have docs

tools/wptrunner/wptrunner/wptrunner.py
```diff
@@ -74,16 +75,19 @@ def get_loader(test_paths, product, debug=None, run_info_extras=None, **kwargs):
...
 def list_test_groups(test_paths, product, **kwargs):
     env.do_delayed_imports(logger, test_paths)
 
-    run_info_extras = products.load_product(kwargs["config"], product)[-1](**kwargs)
+    product_data = products.load_product(kwargs["config"], product)
+    executor_classes = product_data[3]
+    run_info_extras = product_data[-1](**kwargs)
 
-    run_info, test_loader = get_loader(test_paths, product,
+    run_info, test_loader = get_loader(test_paths, product, executor_classes,
```
Ms2ger: At least common up some of this code?

tools/wptrunner/wptrunner/testloader.py
```diff
@@ -562,6 +564,11 @@ def _load_tests(self):
                  "disabled":defaultdict(list)}
 
         for test_path, test_type, test in self.iter_tests():
+            if test.jsshell:
+                executor_class = self.executor_classes.get(test_type)
```
Ms2ger: It doesn't feel right to me that teastloader needs to know about executors. Can we avoid that?

jgraham (changes requested):
(just wondering if this affects the overall status; all Ms2ger's comments are right and should be addressed).

tools/wptrunner/wptrunner/testloader.py
jugglinmike: Got it

tools/wptrunner/wptrunner/testloader.py
jugglinmike: Agreed. I've filed an issue so that can be addressed independently: https://github.com/web-platform-tests/wpt/issues/12384

jugglinmike:
Responding here in order to avoid confusion with collapsed threads.

> This doesn't work:

Fixing this involves extending the heuristic for "ignoring" still further, and it now feels too heavy to declare in-line within `TestLoader`. I've implemented it as a class method on the base Executor class. I'm proposing the method name "ignores" as opposed to "supports" because (as per [James' feedback elsewhere](https://github.com/web-platform-tests/wpt/pull/12360#issuecomment-411702208)) the semantics for `supports_jsshell` are different than the existing use of the term "supports" (as in `supports_webdriver`). This clashes with the `supports_jsshell` property, but I don't want to cause problems downstream, so I've maintained (and documented) that interface.

> It doesn't feel right to me that teastloader needs to know about executors.
> Can we avoid that?

That detail is fundamental to this patch. As far as I can tell, it falls out from the following observations:

- The `TestLoader` is responsible for generating lists of tests
- The `TestExecutor` is responsible for determining if the test exists in WPT
  (as of gh-10769)
- The common interface to all WPT commands that interact with lists of tests is
  `TestLoader`

While we could address this with some other abstraction that mediates `TestLoader` and the CLI commands, I'd worry about future changes reintroducing the kind of inconsistency we're trying to avoid with this change.

That said, I'm not the best person to be designing this solution because I'm missing critical information:

- The implementation details in the external project that uses this feature
- The delineation of wptrunner's public API
- The long-term vision for "jsshell" tests

For instance, if `TestLoader` was considered "public," we could implement this via a flag/method on `TestLoader`, and the consumer could subclass that. (all the more reason to add [the documentation you requested](https://github.com/web-platform-tests/wpt/issues/12384)). If "jsshell" tests are expected to never be supported in WPT, then this might be a more fitting abstraction overall. That's because there are many executor subclasses, and seeing a flag like `supports_jsshell` in the base class might lead one to conclude that there is an in-tree subclass which specified an alternate value.

This is to say: I am not wedded to the current delegation of responsibilities.  I'm most interested in fixing the regression as soon as possible without breaking the downstream consumers.

As you pointed out elsewhere in your review, the current implementation is deficient. Maybe the encapsulation provided by the new `ignores` method is enough to address your concern. If not, do you have any more concrete suggestions for a refactoring you would prefer?

> At least common up some of this code?

There are many keyword arguments and return values involved, and I'm reluctant to tie this patch to such a sensitive refactoring. Less repetition would be a good thing, but would you mind if I addressed that in a follow-up patch?

jgraham:
One consideration about making the `TestLoader` itself depend on the executors is that https://github.com/web-platform-tests/wpt/pull/12227 proposes moving the `TestLoader` into the manifest module so it can be used by `serve`. That obviously doesn't know anything about the executor classes.

That said, since there's a real regression here I''m prepared to accept a temporary fix, and I think the real fix will be allowing a single test file to generate multiple test types, and moving the jsshell tests to a separate type.

jugglinmike:
gh-12367 introduced a conflict, so I resolved that and squashed the "fixup" commit implementing the changes suggested by @Ms2ger. I force-pushed the result to this branch, and [the original version of the branch is available here](https://github.com/web-platform-tests/wpt/compare/master...bocoup:jsshell-hide-tests-prebase)

@jgraham Would you mind merging this if it looks good to you?

tools/wptrunner/wptrunner/testloader.py
```diff
@@ -562,6 +564,11 @@ def _load_tests(self):
                  "disabled":defaultdict(list)}
 
         for test_path, test_type, test in self.iter_tests():
+            executor_class = (self.executor_classes and
+                              self.executor_classes.get(test_type))
+            if executor_class and executor_class.ignores(test):
```
Ms2ger: Have you checked the performance impact of this addition?

tools/wptrunner/wptrunner/executors/base.py
```diff
@@ -108,6 +108,14 @@ class TestExecutor(object):
     test_type = None
     convert_result = None
     supports_testdriver = False
+
+    # `supports_jsshell` determines whether a given test is enumerated in the
+    # various test-traversal commands exposed by the wptrunner CLI (see the
+    # `ignores` method). It is intended to be overridden by subclasses which
+    # are defined outside of WPT.
```
Ms2ger: That's not quite true in practice; [my code](https://searchfox.org/mozilla-central/rev/2466b82b729765fb0a3ab62f812c1a96a7362478/js/src/tests/jstests.py#353) doesn't use executors at all.

jugglinmike:
> Have you checked the performance impact of this addition?

I have not. I presumed that the new operation (a function call and six property retrievals) would not make a measurable difference in this context. We can be more precise, but this is not something I've needed to verify before. Could you point me in the direction of the tools you used to load test your work?

> That's not quite true in practice; [my code](https://searchfox.org/mozilla-central/rev/2466b82b729765fb0a3ab62f812c1a96a7362478/js/src/tests/jstests.py#353) doesn't use executors at all.

Thanks for the link! This is the first time I'm seeing the downstream code, believe it or not. It looks like no one uses the `supports_jsshell`. If that's right, then we can remove it to simplify the `ignores` method and limit the API surface area. Will that work in the project you are maintaining?

---

We can iterate on this solution until it is perfect, but please remember that this is an effort to correct a regression. By way of due diligence, I've offered two patches to fix the feature in-place. However, we are approaching one week since the regression, so we should consider reverting the troublesome patch. Here's a pull request which does just that:

https://github.com/web-platform-tests/wpt/pull/12482

Merging that would allow this patch (or [the even better one that James envisions](https://github.com/web-platform-tests/wpt/pull/12368#issuecomment-412040724)) to be developed without [the pressure caused by broken systems](https://github.com/web-platform-tests/wpt.fyi/issues/432).

Ms2ger:
Re perf: I used cProfile, IIRC. The reason I ask is that the code you're adding here is run for every test in the manifest, so that magnifies any effect.

I only got the executor classes involved in the setup to match `supports_webdriver`, but I agree there's no real reason to do that.

In the meantime, I landed https://github.com/web-platform-tests/wpt/pull/12360 as a trivial temporary fix for the regression. I don't know if we want to keep iterating on this PR, or we should instead implement the multiple-test-type approach immediately. (That does seem to be a better design overall.)

jugglinmike:
Closing as gh-12360 was accepted as the solution to this problem

Issue #12378, opened 2018-08-09
secure-contexts idlharness test is incorrect for Window scope


bzbarsky:
The test fails in every single implementation in the same way; see https://wpt.fyi/results/secure-contexts/idlharness.any.html

That's because the test is actually wrong for the Window scope: it's claiming the property should be on some prototype, but it should be on the object itself.

@lukebjerring @foolip

gsnedders:
(I'm assuming this is an idlharness.js bug?)

bzbarsky:
Unclear.  I'm not sure what the 

    WindowOrWorkerGlobalScope: ["self"],

bit in `add_objects` is supposed to mean.  The test also never defines that `Window` implements `WindowOrWorkerGlobalScope`, which seems like crucial information if all the idlharness bits were correct.  That's clearly a test  bug.

foolip:
There is definitely something broken here, I've discussed it with @lukebjerring on a few PRs and https://github.com/web-platform-tests/wpt/pull/11862 works around something like this, with reference to https://github.com/web-platform-tests/wpt/issues/8053.

In this test, changing to `Window: ["self"]` just makes the failing test go away, so it appears that idlharness.js in this context just isn't able to make the connection between `WindowOrWorkerGlobalScope` or `Window` in any way, whether the object given was `WindowOrWorkerGlobalScope` or `Window`.

lukebjerring:
So, @foolip you're mis-matching a couple of separate issues here; in #11862 I had [originally](https://github.com/web-platform-tests/wpt/pull/11862/commits/8a117ad29869b01682d2486dfb1c4ec29b9b33c2) manually handled the imports when #8053 hadn't been addressed (as a workaround that `Window` wasn't being picked up as a dependency of `WindowOrWorkerGlobalScope`.) It's still needed to check window vs worker and add `Window` or `WorkerGlobalScope` as `"self"`.

@bzbarsky, I think the reason for the problem is that
> The interface has to be the primary interface of all the objects provided. For example, don’t pass {Node: ["document"]}, but rather {Document: ["document"]}.

(from https://web-platform-tests.org/writing-tests/idlharness.html#add_objectsdict)

So, #12386 does indeed fix this.

PR #12386, opened 2018-08-09
Rewrite secure-contexts/idlharness.any.js to use idl_test

foolip:
Fixes https://github.com/web-platform-tests/wpt/issues/12378, by adding
the right depencies and the right global object depending on context.


lukebjerring (approved):
Rubber-stamped (haven't tested locally) but code LGTM

PR #12389, opened 2018-08-09
Fix typo "internatl" in contain-size-002.html

dholbert:
s/internatl/internal/


foolip (approved):
If you'd like to have ago, the same typo is also in:
css/css-contain/contain-size-003.html
css/css-contain/contain-size-004.html
css/css-contain/contain-size-005.html

dholbert:
Good point! I'll fix all the tests at once. I don't think I can do that with github's in-browser "pencil" tool, so I'll fork and make the edits locally and make a "real" pull request from my fork.

PR #12390, opened 2018-08-09
Rewrite Budget API IDL test to use idl_test

foolip:
Also add more objects.


budget-api/idlharness.any.js
```diff
@@ -0,0 +1,24 @@
...
+  ['budget-api'],
+  ['html'],
+  async idl_array => {
+    idl_array.add_objects({ BudgetService: ['navigator.budget'] });
+    if (self.Window) {
+      idl_array.add_objects({ Navigator: ['navigator'] });
+    } else {
+      idl_array.add_objects({ WorkerNavigator: ['navigator'] });
+    }
+    const budgetStates = await navigator.budget.getBudget();
```
foolip: Hmm, actually https://wicg.github.io/budget-api/#dom-budgetservice-getbudget says "If the origin is not a secure context, reject promise with a SecurityError and terminate these steps."

But the interfaces themselves aren't decorated with `[SecureContext]`. So ideally we should verify that the interfaces exist in non-secure contexts too, which switching to .https.any.js would prevent. Thoughts, @lukebjerring?

foolip: Found https://github.com/WICG/budget-api/pull/22. I guess we should just go with .https.any.js then.

lukebjerring: I'm not sure there's a way of testing both http and https (except that serviceworker forces it). Is there a meta tag or anything?

lukebjerring (approved):

budget-api/idlharness.any.js
foolip: There isn't. Discussed the possibility a bit on IRC yesterday: https://w3.logbot.info/testing/20180809#c431719 (about fixing another issue)

PR #11840, opened 2018-07-06
[testdriver.js] Document and test `bless` method

jugglinmike:
This new API was first introduced via [1]. Complete that initial work by
adding documentation and automated tests.

[1] 8b8082f30267410e4b3729439be23df8b803dde8

PR #12399, opened 2018-08-10
fix(resources/testdriver): wait for a document.body

marcoscaceres:
`bless()` is racy with document.body. It should actually wait for there to be a body.

### DIFF
--- a/resources/testdriver.js
+++ b/resources/testdriver.js
@@ -62,24 +62,25 @@
          *                    rejected if interaction fails or the provided
          *                    function throws an error
          */
-        bless: function(intent, action) {
-            var button = document.createElement("button");
+        async bless(intent, action) {
+            const button = document.createElement("button");
             button.innerHTML = "This test requires user interaction.<br />" +
                 "Please click here to allow " + intent + ".";
             button.id = "wpt-test-driver-bless-" + (idCounter += 1);
+            if (document.readyState === "loading") {
+                await new Promise(
+                    resolve => document.addEventListener("DOMContentLoaded", resolve)
+                );
+            }
             document.body.appendChild(button);
-
-            return new Promise(function(resolve, reject) {
-                    button.addEventListener("click", resolve);
-
-                    test_driver.click(button).catch(reject);
-                }).then(function() {
-                    button.remove();
-
-                    if (typeof action === "function") {
-                        return action();
-                    }
-                });
+            await new Promise((resolve, reject) => {
+                button.addEventListener("click", resolve);
+                test_driver.click(button).catch(reject);
+            });
+            button.remove();
+            if (typeof action === "function") {
+                return action();
+            }
         },
 
         /**
--- a/resources/testdriver.js
+++ b/resources/testdriver.js
@@ -62,25 +62,25 @@
          *                    rejected if interaction fails or the provided
          *                    function throws an error
          */
-        async bless(intent, action) {
-            const button = document.createElement("button");
+        bless: function(intent, action) {
+            var button = document.createElement("button");
             button.innerHTML = "This test requires user interaction.<br />" +
                 "Please click here to allow " + intent + ".";
             button.id = "wpt-test-driver-bless-" + (idCounter += 1);
-            if (document.readyState === "loading") {
-                await new Promise(
-                    resolve => document.addEventListener("DOMContentLoaded", resolve)
-                );
-            }
-            document.body.appendChild(button);
-            await new Promise((resolve, reject) => {
-                button.addEventListener("click", resolve);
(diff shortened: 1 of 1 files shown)


gsnedders:
We could also in principle not be an HTML document and not have a body (or have the body removed by a script); we should probably more generally handle the case when there is no body.

I'm a bit skeptical about whether we actually want to block tests on DOMContentLoaded, though?

jugglinmike:
This patch introduces ES2017 syntax which bucks the trend in the testdriver.js utility and in the `resources/` directory at large. I don't think that's necessary; is that right?

> We could also in principle not be an HTML document and not have a body (or have the body removed by a script); we should probably more generally handle the case when there is no body.

What do you think about waiting for a test to require support for that?

> I'm a bit skeptical about whether we actually want to block tests on DOMContentLoaded, though?

Can you say more?

gsnedders:
> This patch introduces ES2017 syntax which bucks the trend in the testdriver.js utility and in the resources/ directory at large. I don't think that's necessary; is that right?

Oh that's true, and as long as https://github.com/servo/servo/issues/20585 is yet to land it'll break it on Servo (though AFAIK there's no Servo support for it), so idk how much that matters.

> What do you think about waiting for a test to require support for that?

FWIW, I know we've had previous cases where people have avoided features because "oh that doesn't work unless you're HTML" before without telling us for a long time, so I'd rather have a small workaround for it (`(document.body || document.documentElement)` should do).

> Can you say more?

Normally tests start running before DOMContentLoaded, so this would be a subtle difference when using `bless`, which doesn't seem ideal.

jugglinmike:
> Oh that's true, and as long as servo/servo#20585 is yet to land it'll break it on Servo (though AFAIK there's no Servo support for it), so idk how much that matters.

Consistency matters, but I think the troubling part is we currently advocate for consistency derived from precedence. Without something like gh-6333, all we can really say is, "please don't rock the boat." That's only going to get more discouraging as TC-39 adds more features, but we should talk about that in gh-6333.

> FWIW, I know we've had previous cases where people have avoided features because "oh that doesn't work unless you're HTML" before without telling us for a long time, so I'd rather have a small workaround for it ((document.body || document.documentElement) should do).

That's worth a lot! I'm convinced that we should do this. And test for it.

> Normally tests start running before DOMContentLoaded, so this would be a subtle difference when using `bless`, which doesn't seem ideal.

That would be more concerning if this were testharness-level behavior. As it stands, though, tests will still start running before the event. The delay isn't so much a "subtle difference" among tests of the same type. It's more that the delay is an intentional feature of the API which tests will be explicitly invoking.

I [snuck](https://www.youtube.com/watch?v=q51ld-scMI8) "intentional feature" in there, though. Since this is intended to be used for features that require user activation, then I think it's appropriate to wait for DOMContentLoaded. (Hopefully there's no API that requires users to click prior to document ready :P) If you buy that, then we can update the docs accordingly.

marcoscaceres:
@gsnedders wrote
> We could also in principle not be an HTML document 

I'm a bit confused... I'm not sure how then creating a "button" would work if it's not HTML, as the expectation is that it's a `HTMLButtonElement` and will behave as such?  

> This patch introduces ES2017 syntax which bucks the trend in the testdriver.js utility and in the resources/ directory at large. I don't think that's necessary; is that right?

One could definitely write this using lots of promises, `.then()`s, and the old syntax. However,  that make the code much less maintainable and readable, so seems somewhat irresponsible to do that. As you state, "consistency matters, but I think the troubling part is we currently advocate for consistency derived from precedence." 

I agree is not a good reason to keep using the current style: It be nice to modernize our development practices - there is nothing particularly radical about using async/await or anything else in this patch. For new things, why not just use async/await and new object syntax?

jgraham:
A button will work as a button even if the root is SVG or XHTML or some other type that doesn't necessarily have a body.

We are more conservative in features for this mind of shared infrastructure than we are in features for tests; for tests there's a clear rule which is "latest Blink / EdgeHTML / Gecko / WebKit must support all features", for stuff in `resources` it's more fuzzy, but I would hold off from using `async`/`await` syntax at this time since it doesn't seem worthwhile to break e.g. Servo (I would also avoid discussing this in terms of moral positions, which is unhelpful when it's a question of technical tradeoffs).

marcoscaceres:
Ok, will rewrite it ☺️

Ms2ger (approved):
This seems reasonable enough.

jugglinmike:
[A separate pull request introduces tests for this method](https://github.com/web-platform-tests/wpt/pull/11840), but it's been overlooked for over a month. I've updated it to include a test for this fix. Hopefully we'll see it in `master` soon!

PR #12403, opened 2018-08-10
Update interfaces/*.idl boilerplate from reffy-reports

foolip:
The boilerplate has changed:
https://github.com/tidoust/reffy/pull/114

This change updates the files where *only* the boilerplate would
differ, files with other differences are left untouched.


lukebjerring (approved):
Skimmed through, as these are Auto generated.

Issue #12289, opened 2018-08-02
xhr/overridemimetype-unsent-state-force-shiftjis is now at odds with the XHR spec.


wisniewskit:
With the recent overrideMimeType changes, the WPT xhr/overridemimetype-unsent-state-force-shiftjis is now invalid (since the mimetype override is cleared as the XHR is opened in step 12).

I'm not sure if this means that the spec is now no longer web-compatible, however.

foolip:
@annevk

annevk:
Sigh, I guess that's why we never cleared `overrideMimeType` state? Does any implementation follow the specification? I thought at least Edge did.

foolip:
https://wpt.fyi/results/xhr/overridemimetype-unsent-state-force-shiftjis.htm?labels=experimental isn't too illuminating, but does suggest *some* difference between Edge and the others.

annevk:
I verified that Edge indeed behaves this way, but in retrospect it seems better to align with the other browsers, especially given the situation on mobile. I'll create some PRs to fix the tests and specification.

annevk:
PRs created.

PR #12404, opened 2018-08-10
XHR: overrideMimeType() is not reset by open()

annevk:
Also fix the number of replacement characters.

See also #12289.


wisniewskit (approved):
LGTM

PR #12405, opened 2018-08-10
Convert compat+console+fullscreen+storage+url+xhr to idl_test

foolip:
Also add more objects in some of these tests.

XHR will now be tested against workers as well, and local tested
revealed a test will be failing due to an old TODO in Blink:
https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/xmlhttprequest/xml_http_request.idl?l=79&rcl=9c356a9697625b3c25856529908af0c3253dfa2c

A failing test for that is good :)


annevk:
compat shouldn't be tested in workers?

foolip:
Nope, it only has partial interfaces for things in exposed on window.

annevk (approved):

PR #12415, opened 2018-08-10
Test that there's no help cursor on help links

domenic:
Follows https://github.com/whatwg/html/pull/3902.

### DIFF
--- a/html/rendering/the-css-user-agent-style-sheet-and-presentational-hints/no-help-cursor-on-links.historical.html
+++ b/html/rendering/the-css-user-agent-style-sheet-and-presentational-hints/no-help-cursor-on-links.historical.html
@@ -1,0 +1,38 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>link with rel="help" cursor tests</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<link rel="help" href="https://html.spec.whatwg.org/#phrasing-content-3">
+<link rel="help" href="https://github.com/whatwg/html/pull/3902">
+
+<div id="log"></div>
+
+<a href="/common/blank.html?unvisited" rel="help" id="unvisited">unvisited</a>
+<a href="/common/blank.html?willbevisited" rel="help" id="willbevisited">will be visited</a>
+
+<script>
+"use strict";
+
+
+test(() => {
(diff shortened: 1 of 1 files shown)


html/rendering/the-css-user-agent-style-sheet-and-presentational-hints/no-help-cursor-on-links.historical.html
```diff
@@ -0,0 +1,38 @@
...
+},"Unvisited help links must have pointer cursor, not help cursor");
+
+async_test(t => {
+  const el = document.querySelector("#willbevisited");
+
+  const iframe = document.createElement("iframe");
+  iframe.src = el.href;
+  iframe.onload = t.step_func_done(() => {
+    const style = window.getComputedStyle(el);
+    assert_equals(style.cursor, "pointer");
```
foolip: Is there also a test that checks that loading a URL in an iframe causes it to be considered visited? I guess asserting `el.matches(':visited')` here doesn't work because privacy-preserving lies are told, but did you try to check?

foolip (approved):

html/rendering/the-css-user-agent-style-sheet-and-presentational-hints/no-help-cursor-on-links.historical.html
domenic: There isn't any way to distinguish unvisited from visited in script, so it's hard to tell. In Firefox and Edge running the test manually makes the links visually purple; in Chrome it does not. This whole test is kind of dubious in these regards, but I figured it doesn't hurt. Maybe worth a comment....

html/rendering/the-css-user-agent-style-sheet-and-presentational-hints/no-help-cursor-on-links.historical.html
```diff
@@ -0,0 +1,50 @@
...
+
+test(() => {
+  const el = document.querySelector("#unvisited");
+  const style = window.getComputedStyle(el);
+
+  assert_equals(style.cursor, "pointer");
+},"Unvisited help links must have pointer cursor, not help cursor");
+
+
+// This test is kind of dubious. Browsers don't allow you to distinguish visited and unvisited links
```
foolip: Thanks, comment LGTM.

foolip (approved):

PR #12422, opened 2018-08-13
[Battery] Correct IDL file name

Honry:
The IDL file name is [`battery-status.idl`](https://github.com/web-platform-tests/wpt/blob/master/interfaces/battery-status.idl)

PR #12427, opened 2018-08-13
Fix broken Battery Status API idlharness test (due to typo)

foolip:
'battery' => 'battery-status' is the fix, but also rename to
idlharness.js for consistency with most other idlharness tests.


foolip:
There are now 29 tests, all passing in Chrome, and 13 failing in a sensible way in Firefox.

foolip:
Dupe of https://github.com/web-platform-tests/wpt/pull/12422

Issue #12428, opened 2018-08-13
idl_test uses Promise.finally, which isn't available in Edge 17


foolip:
In https://github.com/web-platform-tests/wpt/pull/12272 a dependency on Promise.finally was added to `idl_test`:
https://github.com/web-platform-tests/wpt/blame/e9f3244421c353a63142dd9742e0ed6edc509cc8/resources/idlharness.js#L3224

Unfortunately, that's causing tests to fail in Edge 17:
https://wpt.fyi/results/encoding/idlharness.any.html?sha=eea0b54014&labels=stable

We need to achieve the same thing by some other means.

PR #12429, opened 2018-08-13
Fix dependency order in subresource-integrity/idlharness.window.js

foolip:
With these 3 dependencies, there is only one order which doesn't
result in an error during setup, and this is it. The previous error
was "ProcessingInstruction includes LinkStyle, but
ProcessingInstruction is undefined.":
https://wpt.fyi/results/subresource-integrity/idlharness.window.html?sha=eea0b54014&labels=stable

Separate issue filed about the Edge failure:
https://github.com/web-platform-tests/wpt/issues/12428

### DIFF
--- a/subresource-integrity/idlharness.window.js
+++ b/subresource-integrity/idlharness.window.js
@@ -7,7 +7,7 @@
 
 idl_test(
   ['SRI'],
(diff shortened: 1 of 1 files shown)


joelweinberger (approved):
lgtm, but it's been a while since I've had context

PR #12430, opened 2018-08-13
Tests for TextEncoderStream and TextDecoderStream

ricea:
The standard change that adds these classes is
https://github.com/whatwg/encoding/pull/149.


encoding/streams/readable-writable-properties.any.js
```diff
@@ -0,0 +1,22 @@
+// META: global=worker
+
+// This just tests that the "readable" and "writable" properties pass the brand
+// checks. All other relevant attributes are covered by idlharness.html.
```
annevk: This is already out-of-date (it's idlharness.any.js), but it seems better to say "IDL tests" here given future refactoring in that area.

encoding/streams/idlharness.any.js
```diff
@@ -0,0 +1,57 @@
...
+`;
+
+var idl_array = new IdlArray();
+idl_array.add_untested_idls(untested_idl);
+idl_array.add_idls(idl);
+idl_array.add_objects({
+  TextEncoderStream: ['new TextEncoderStream()'],
+  TextDecoderStream: ['new TextDecoderStream()']
+});
+idl_array.test();
```
annevk: Now this is in PR form it seems we can make the changes directly to encoding.idl.

annevk:
* It seems you need to rebase due to IDL changes on master
* Turn encoding/streams/realms.html into encoding/streams/realms.window.js?
* Why do you restrict many any.js tests to workers?

annevk:
@foolip should we just not make IDL changes in this PR? Still not entirely sure how the new setup is supposed to work.

ricea:
> It seems you need to rebase due to IDL changes on master

Done.

> Turn encoding/streams/realms.html into encoding/streams/realms.window.js?

Done.

> Why do you restrict many any.js tests to workers?

Confusingly `META: global=worker` [means "run everywhere"](https://web-platform-tests.org/writing-tests/testharness.html#multi-global-tests).

lukebjerring:
@ricea - wow, that is surprising. FWIW I've seen a style of explicitly doing `window,worker` regardless of the `window` being redundant and IMHO that's more predictable for the reader.

encoding/streams/decode-bad-chunks.any.js
```diff
@@ -0,0 +1,73 @@
...
+      const ab = u8.buffer;
+      const mc = new MessageChannel();
+      mc.port1.postMessage(ab, [ab]);
+      return ab;
+    })()
+  }
+];
+
+let badShared = [];
+if (typeof SharedArrayBuffer !== 'undefined') {
```
annevk: I guess this is here so the test doesn't fail completely in user agents that have it disabled? This kind of conditional testing does feel a little bit problematic. Perhaps we can put the constructor in a callback so only the actual test fails?

interfaces/encoding.idl
```diff
@@ -3,6 +3,12 @@
 // (https://github.com/tidoust/reffy-reports)
 // Source: Encoding Standard (https://encoding.spec.whatwg.org/)
 
+interface mixin TextDecoderCommon {
```
foolip: Hmm, we have an interesting problem here. As the boilerplate says this file is generated, and if these changes are made in the spec they'd eventually propagate here. And yet making the changes in the wpt PR linked to the spec PR is great, makes it easy to understand the source of the changes.

But, if this change is merged before the spec is scraped, an interfaces/ IDL auto-update might try to effectively revert the changes again, and later re-apply them. That would be silly.

Still, please carry on, we should just make tooling handle a situation like this.

encoding/streams/decode-bad-chunks.any.js
ricea: Yes. I found a simple way to make only those specific tests fail if SharedArrayBuffer is absent. PTAL.

annevk (approved):

PR #12432, opened 2018-08-13
Avoid using Promise.prototype.finally in idlharness.js

foolip:
In this context, using .then() achieves the same thing because of the
preceding .catch().

Fixes https://github.com/web-platform-tests/wpt/issues/12428.

### DIFF
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -3221,7 +3221,7 @@ function idl_test(srcs, deps, idl_setup_func) {
                 }
             })
             .catch(function(e) { setup_error = e || 'IDL setup failed.'; })
-            .finally(function () {
+            .then(function () {
                 var error = setup_error;
                 try {
                     idl_array.test(); // Test what we can.


foolip:
I've verified that this fixes the setup error in https://github.com/web-platform-tests/wpt/issues/12428.

I've also checked that Promise.prototype.finally is in Edge Insider Preview, so that the root problem is going to be resolved. The only way to prevent this from happening in the first place would be to run stable versions of browsers on PRs and have some sort of smoke test for idlharness.js which can be assumed to pass 100%. Second best would be noticing the regression automatically once a new run happens in wpt.fyi. cc @lukebjerring.

@thejohnjansen FYI that this has happened, in case you're also noticing these failures on the imported tests.

jgraham (approved):

foolip:
https://github.com/web-platform-tests/wpt/issues/6266 is about writing down the policy that I've implicitly followed here, and https://github.com/web-platform-tests/wpt/pull/6333 the PR to do so.

PR #12433, opened 2018-08-13
Rewrite Vibration API IDL test to use idl_test

foolip:
Tests still pass on Chrome and Firefox.


lukebjerring (approved):
Rubber stamp (untested locally)

PR #12436, opened 2018-08-13
Rewrite WebVTT IDL tests to use idl_test

foolip:
The results are unchanged on Chrome and Firefox. Chrome fails 4/55
tests and Firefox passes them all.


lukebjerring (approved):
Rubber stamp (untested locally)

PR #12439, opened 2018-08-13
Update interfaces/animation-worklet.idl

autofoolip:
https://github.com/tidoust/reffy-reports/blob/f14a522/whatwg/idl/animation-worklet.idl


lukebjerring (approved):

foolip:
This one required manual changes, so copied it over to https://github.com/web-platform-tests/wpt/pull/12444. (My plan is that PRs like this would be auto-updated daily, so modifying them wouldn't be safe.)

PR #12442, opened 2018-08-13
Update interfaces/webxr.idl

autofoolip:
https://github.com/tidoust/reffy-reports/blob/f14a522/whatwg/idl/webxr.idl


lukebjerring (approved):

foolip (approved):
Verified that the number of test decreases in Chrome, but the number of failures stays constant.

Issue #12443, opened 2018-08-13
Support worklets in .any.js and idlharness.js


foolip:
Writing idlharness.js tests for https://wicg.github.io/animation-worklet/ (https://github.com/web-platform-tests/wpt/pull/12439) is not possible because we don't support `global=animationworklet` in .any.js tests. That support would have to be added to wptserve.

In addition, idlharness.js would need to know about this new global scope, or alternatively the handling of global scopes be generalized to understand from `[ Exposed=AnimationWorklet, Global=AnimationWorklet ] interface AnimationWorkletGlobalScope` and the existence of `self.AnimationWorkletGlobalScope` that interfaces marked `[Exposed=AnimationWorklet]` should be tested.

PR #12444, opened 2018-08-13
Update the Animation Worklet IDL file + test

foolip:


foolip:
The test didn't work at all before and now mostly works. Filed https://github.com/WICG/animation-worklet/issues/108 about one failure, and https://github.com/web-platform-tests/wpt/issues/12443 about the fact that we're not actually testing anything inside the worklet.

lukebjerring (approved):

PR #12448, opened 2018-08-13
Test length of `selectedOptions` after change

Zirro:
This adds a test for the length of `select.selectedOptions` after an option element was changed through `option.selected`. jsdom would previously fail to update the value under these conditions.

The test passes consistently in Chrome and Safari. It passes in Firefox on the first run, but fails if the page is reloaded with cache as that retains the previous state of the `select` element.

### DIFF
--- a/html/semantics/forms/the-select-element/select-selectedOptions.html
+++ b/html/semantics/forms/the-select-element/select-selectedOptions.html
@@ -114,12 +114,15 @@ test(() => {
 test(() => {
   const select = document.getElementById("select-same-object-change");
   const before = select.selectedOptions;
+  assert_equals(before.length, 3);
 
   select.selectedOptions[1].selected = false;
 
   const after = select.selectedOptions;
 
   assert_equals(before, after);
+  assert_equals(before.length, 2);
+  assert_equals(after.length, 2);
 
 }, ".selectedOptions should return the same object after selection changes - [SameObject]");
 </script>


zcorpan (approved):
Can you file an issue for Firefox? Have you tested Edge?

Zirro:
The Firefox behaviour could potentially be regarded as a feature (unless these tests are explicitly meant to survive a reload, but I figured I'd mention it anyway). Edge is not available to me right now.

zcorpan:
The test should pass in all browsers, if it fails in Firefox then there should be a bug for them to track it.

The relevant parts of the spec is

> The selectedOptions IDL attribute must return an HTMLCollection rooted at the select node, whose filter matches the elements in the list of options that have their selectedness set to true.

https://html.spec.whatwg.org/multipage/form-elements.html#dom-select-selectedoptions

> The selected IDL attribute, on getting, must return true if the element's selectedness is true, and false otherwise. On setting, it must set the element's selectedness to the new value, set its dirtiness to true, and then cause the element to ask for a reset.

https://html.spec.whatwg.org/multipage/form-elements.html#dom-option-selected

Zirro:
It passes in Firefox on the first run, and when the test is reloaded without caching. The failure appears only with a regular reload, as the previous value of the `select` is kept. It's not clear to me whether these tests are meant to account for reloads.

zcorpan:
Oh I see. Sorry. Yeah, I suppose that is intentional; not sure if the spec needs to say anything to account for remembering form state across reloads. @annevk any thoughts?

PR #12450, opened 2018-08-13
Ensure templated tests do set a Content-Type: text/html HTTP header

youennf:


gsnedders (approved):
This LGTM, though I'm sad about the lack of tests. If it's blocking you I guess we shouldn't block everything on someone figuring out how to test this…  🙁

PR #12458, opened 2018-08-13
Ensure `x` is declared before use in test

cscott:
This suppresses spurious warnings when running the platform tests in some test runners.


foolip (approved):
Thanks for the fix! May I ask which runner/browser you got a warning in?

cscott:
@foolip The [domino](https://github.com/fgnass/domino/) DOM implementation for node uses web-platform-tests as part of its test suite.

foolip:
Alright, cool!

PR #12462, opened 2018-08-13
Revert "Partly fix #8581: add extra subdomains for cookie tests"

### LINKED ISSUE #8581: Isolated domain for cookie tests?

### LINKED ISSUE #12461: Trying to run any WebDriver test fails in Edge
Everything fails like:

```
ERROR /webdriver/tests/switch_to_window/switch.py - the environment variable is longer than 32767 bytes
  File "C:\Users\gsnedders\Documents\web-platform-tests\tools\wptrunner\wptrunner\executors\base.py", line 405, in _run
    self.result = True, self.func(self.session, self.path, self.timeout)
  File "C:\Users\gsnedders\Documents\web-platform-tests\tools\wptrunner\wptrunner\executors\base.py", line 371, in do_wd
spec
    timeout=timeout)
  File "C:\Users\gsnedders\Documents\web-platform-tests\tools\wptrunner\wptrunner\executors\pytestrunner\runner.py", lin
e 47, in run
    os.environ["WD_SERVER_CONFIG"] = json.dumps(server_config.as_dict())
  File "C:\Users\gsnedders\Documents\wpt-env\lib\os.py", line 422, in __setitem__
    putenv(key, item)
ValueError: the environment variable is longer than 32767 bytes
```

bisect points at ff4e72e977d0776510aea28ddfb497c7fc12e848 from #10953

gsnedders:
This reverts commit ff4e72e977d0776510aea28ddfb497c7fc12e848.

Fixes #12461.

### DIFF
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -16,7 +16,6 @@ import traceback
 from six.moves import urllib
 import uuid
 from collections import defaultdict, OrderedDict
-from itertools import chain, product
 from multiprocessing import Process, Event
 
 from localpaths import repo_root
@@ -702,9 +701,6 @@ def build_config(override_path=None, **kwargs):
 
     return rv
 
-def _make_subdomains_product(s, depth=3):
-    return set(u".".join(x) for x in chain(*(product(s, repeat=i) for i in range(1, depth+1))))
-
 _subdomains = {u"www",
                u"www1",
                u"www2",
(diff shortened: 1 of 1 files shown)


Hexcles (approved):
Revert LGTM

Issue #12419, opened 2018-08-11
parsing-testcommon.js not fetched


TalbotG:
Many writing-modes tests fail just because functions being called are not fetched. Examples given:
http://test.csswg.org/harness/test/css-writing-modes-3_dev/direction-invalid/
http://test.csswg.org/harness/test/css-writing-modes-3_dev/direction-valid/
http://test.csswg.org/harness/test/css-writing-modes-3_dev/text-combine-upright-invalid/
http://test.csswg.org/harness/test/css-writing-modes-3_dev/text-combine-upright-valid/
http://test.csswg.org/harness/test/css-writing-modes-3_dev/text-orientation-invalid/
http://test.csswg.org/harness/test/css-writing-modes-3_dev/text-orientation-valid/
and there are many more tests like those

**ReferenceError: Can't find variable: test_invalid_value**
can be read in the Message column and also be found in browsers' javascript consoles.

foolip:
@ewilligers can you take a look?

gsnedders:
This is due to them being in `resources` which the build system doesn't know about; they'd need to be in `support` for the build system to copy them around.

TalbotG:
Philip,
How do you add a label to an issue? Do you have to be on a privilege list? As the reporter of this issue, I can not do this.

gsnedders:
@TalbotG You have to have write access to the repo, IIRC.

foolip:
@gsnedders I checked before and that's not the case, and probably wasn't the case at the time of the repo move either, since I pinged a lot of people who had write access then in issue 11568.

gsnedders:
@foolip I mean to add labels to issues, which per https://help.github.com/articles/repository-permission-levels-for-an-organization/ needs write access.

TalbotG:
Philip,

I received your invitation to join @web-platform-tests . Thank you. I should now be able to add label to issues that I report.

foolip:
I'm not sure if org membership is enough, let me know if it doesn't work.

TalbotG:
@ewilligers 
Eric, has this issue really been fixed? I ask because I still get a reference error:
**ReferenceError: test_invalid_value is not defined**
when I try the following 10 tests:

http://test.csswg.org/harness/test/css-writing-modes-3_dev/direction-valid/
http://test.csswg.org/harness/test/css-writing-modes-3_dev/direction-invalid/
[...]

PR #12465, opened 2018-08-14
Move parsing-testcommon.js to support/

ewilligers:
The file parsing-testcommon.js was previously in parsing/resources/

By moving it to css/*/support/ we enable scripted replication.

Requested in #12419


ewilligers:
@foolip @mrego please review.

foolip:
Context for my own review: identical copies of parsing-testcommon.js previously existed here:
```
css/css-transforms/parsing/resources/parsing-testcommon.js
css/css-masking/parsing/resources/parsing-testcommon.js
css/filter-effects/parsing/resources/parsing-testcommon.js
css/css-color/parsing/resources/parsing-testcommon.js
css/css-writing-modes/parsing/resources/parsing-testcommon.js
css/css-box/parsing/resources/parsing-testcommon.js
css/css-images/parsing/resources/parsing-testcommon.js
css/css-ui/parsing/support/parsing-testcommon.js
css/motion/parsing/resources/parsing-testcommon.js
css/css-cascade/parsing/support/parsing-testcommon.js
css/css-backgrounds/parsing/resources/parsing-testcommon.js
css/compositing/parsing/resources/parsing-testcommon.js
css/css-shapes/parsing/resources/parsing-testcommon.js
```

foolip:
And afterwards, the same (unmodified) file exists here:
```
css/css-transforms/support/parsing-testcommon.js
css/css-masking/support/parsing-testcommon.js
css/filter-effects/support/parsing-testcommon.js
css/css-color/support/parsing-testcommon.js
css/support/parsing-testcommon.js
css/css-writing-modes/support/parsing-testcommon.js
css/css-box/support/parsing-testcommon.js
css/css-images/support/parsing-testcommon.js
css/css-ui/support/parsing-testcommon.js
css/motion/support/parsing-testcommon.js
css/css-cascade/support/parsing-testcommon.js
css/css-backgrounds/support/parsing-testcommon.js
css/compositing/support/parsing-testcommon.js
css/css-shapes/support/parsing-testcommon.js
```

In other words, this is just renaming, not consolidating.

foolip (approved):

foolip:
This seems like it would fix the problem to me, but I don't know if it's necessary to keep so many copies. @gsnedders, can you explain how this works, is it possible to have just a single copy in css/support/ and use relative URLs for that?

gsnedders:
@foolip The CSS build system finds tests (based on `<link rel=help>`), moves them to the destination direction (roughly one per spec), and then also moves any files in an adjacent `support` directory. To make relative URLs work, one would need to actually find the set of files linked to, move the files, and then rewrite the relative URL (given the directory structures differ).

foolip:
Huh. So no tests can reference https://github.com/web-platform-tests/wpt/tree/master/css/support by relative URL and still work in the CSS build system, right? If so, is there any point in having that directory at all?

gsnedders:
Historically it wasn't possible to run the unbuilt tests, so maybe it was useful then, given the readme says, "these are sync'ed into the support directories of all our test suites"? (And I don't think that's been true since the build system was rewritten years ago.)

Certainly nothing in there can be referenced and work in both unbuilt and built test suites.

ewilligers:
@gsnedders,
For the tests in `css/css-writing-modes/parsing/*.html`
I was hoping that #12419 would be fixed by the use of moving
`css/css-writing-modes/parsing/resources/parsing-testcommon.js`
to
`css/css-writing-modes/support/parsing-testcommon.js`

"... and then also moves any files in an adjacent support directory."
suggests that I should have instead moved the js file to
`css/css-writing-modes/parsing/support/parsing-testcommon.js`

Would that be accurate?

gsnedders:
@ewilligers Yes.

It literally finds tests and moves foo.html to something like css-writing-modes/{html4,xhtml1}/foo.html, and moves ./support/* (from `foo.html`) to css-writing-modes/{html4,xhtml1}/support/*.

mrego:
> Huh. So no tests can reference https://github.com/web-platform-tests/wpt/tree/master/css/support by relative URL and still work in the CSS build system, right? If so, is there any point in having that directory at all?

The option would be to use absolute paths for that folder.
Then we could have just one copy of `parsing-testcommon.js` in `css/support/` and use it in the tests with `<script src="/css/support/parsing-testcommon.js"></script>`.
I believe that's much better than keep N copies of `parsing-testcommon.js` manually synced on the repository.

ewilligers:
OK, absolute paths in PR https://github.com/web-platform-tests/wpt/pull/12812

gsnedders:
I'm surprised to see that that actually already works on `test.csswg.org`; certainly it's still the case that things like `/fonts` don't work (see https://github.com/w3c/csswg-drafts/issues/3076)!

mrego:
> I'm surprised to see that that actually already works on test.csswg.org; certainly it's still the case that things like /fonts don't work (see w3c/csswg-drafts#3076)!

I believe `/fonts/` is working fine, if you use absolute paths too.
Several tests use that already, for example:
https://test.csswg.org/harness/test/css-writing-modes-3_dev/single/text-orientation-010/

gsnedders:
Oh, um, `/common` doesn't. Bleh. Some do, obviously.

PR #12469, opened 2018-08-14
Update interfaces/css-paint-api.idl

autofoolip:
https://github.com/tidoust/reffy-reports/blob/bd06a9e/whatwg/idl/css-paint-api.idl


foolip (approved):
Tested manually, this fixes the "Partial interface CSS: original interface defined" failure seen in http://web-platform-tests.live/css/css-paint-api/idlharness.html

PR #12470, opened 2018-08-14
Update interfaces/SVG.idl

autofoolip:
https://github.com/tidoust/reffy-reports/blob/bd06a9e/whatwg/idl/SVG.idl


foolip (approved):
Run manually to confirm that this increases the number of tests and decreases the number of failing ones in both Chrome and Firefox. Until we have regression detection that lines up and compares subtests, this is about as good as we can do.

PR #12471, opened 2018-08-14
Update interfaces/background-fetch.idl

autofoolip:
https://github.com/tidoust/reffy-reports/blob/bd06a9e/whatwg/idl/background-fetch.idl


foolip (approved):
Run manually to see that the test still works in Chrome and Safari, and that the numbers of failures have decreased. Presumably that means the spec got closer to implemented reality.

PR #12473, opened 2018-08-14
HTML: fieldset should have min-inline-size: min-content

zcorpan:
See https://github.com/whatwg/html/issues/2078.


annevk:
This doesn't test what happens if you change it. Presumably that should also be tested here?

zcorpan:
There is a test already for min-width, but I could change it to be part of this test instead so overriding in the vertical case is also covered.

annevk:
@bzbarsky or @emilio should probably review this as I'm not really familiar with these new CSS properties.

emilio (approved):

B3ebec89:
@bzbarsky on @emilio should probably review this as I'm not really familiar with these new properties

PR #12482, opened 2018-08-14
Revert "Add some tests for the Wasm JS API; +bbouvier"

jugglinmike:
This reverts commit 5c219ef55a4b2099c0f4c53b9669df10d73e6f3a.

This resolves gh-12366. See also gh-12360 and gh-12368.


Ms2ger:
I landed #12360 as a less harmful workaround.

PR #12485, opened 2018-08-14
[css-contain] Fix antialiasing issue

### LINKED ISSUE #12362: contain-size-012.html legitimately fails if antialiased serifs are clipped by its `overflow:hidden`
The test [contain-size-012.html](https://github.com/web-platform-tests/wpt/blob/master/css/css-contain/contain-size-012.html) fails on Firefox for Windows, with the only difference being on the left fringes of serifs on the "P" of "Pass", as reported here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1474848

I think this is a test bug -- the testcase has the word "Pass" wrapped in `overflow:hidden`, and it looks like the font happens to have serifs which create antialiased ink-overflow off the left side of the content box, and those antialiased serifs get correctly clipped by the `overflow:hidden` wrapper.  In contrast, the reference case doesn't have `overflow:hidden` so nothing is clipped.

So: the test's expectations of pixel-for-pixel accuracy are incorrect here. CC test-author @mrego

frivoal:
Closes #12362

### DIFF
--- a/css/css-contain/contain-size-001.html
+++ b/css/css-contain/contain-size-001.html
@@ -5,7 +5,7 @@
   <link rel="author" title="Florian Rivoal" href="https://florian.rivoal.net">
   <meta name=flags content="">
   <meta name=assert content="size containment does not apply to non atomic inlines">
-  <link rel="match" href="../reference/pass_if_pass_below.html">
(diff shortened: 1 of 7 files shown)


dholbert (approved):
Thanks! Looks good to me.

PR #12486, opened 2018-08-14
[css-contain] use overflow:hidden on elements where it actually works

### LINKED ISSUE #12388: contain-size-007.html through 010.html use "overflow:hidden" on a box that doesn't support it
The tests `contain-size-007.html` through `contain-size-010.html` have `overflow:hidden` in a context where it has no effect, which is somewhat problematic. Test links:
https://github.com/web-platform-tests/wpt/blob/master/css/css-contain/contain-size-007.html
https://github.com/web-platform-tests/wpt/blob/master/css/css-contain/contain-size-008.html
https://github.com/web-platform-tests/wpt/blob/master/css/css-contain/contain-size-009.html
https://github.com/web-platform-tests/wpt/blob/master/css/css-contain/contain-size-010.html

They're trying to test that `contain:size` has no effect on internal table parts (e.g. table-row & table-row-group), and it looks like they're trying to test that by seeing whether they collapse to nothing when given `overflow:hidden`.  But this is ineffective because `overflow:hidden` doesn't actually do anything useful on these boxes.  (You can easily see this by using `overflow:scroll` instead -- if the `overflow` property had any effect in this context, then that'd make scrollbars appear. But none appear; hence, the `overflow` property seems to be ignored on these boxes.)

To really test this, I think we should put `overflow:hidden` on an *ancestor element*, like we do for e.g. https://github.com/web-platform-tests/wpt/blob/master/css/css-contain/contain-size-001.html and https://github.com/web-platform-tests/wpt/blob/master/css/css-contain/contain-size-002.html .  That would still let us accurately find out whether the `contain:size` thing actually collapsed away or not, by virtue of whether it props up the size of our auto-sized `overflow:hidden` ancestor.

CC @frivoal

frivoal:
Closes #12388

<del>This PR reuses the ref file introduced in #12485, so it is based on that branch. Should be merged (or rebased) after #12485 is landed.</del>

**edit: this is now rebased, and ready to review / merge**

### DIFF
--- a/css/css-contain/contain-size-007.html
+++ b/css/css-contain/contain-size-007.html
@@ -3,15 +3,19 @@
 <title>CSS Containment Test: Size containment on table-row-group</title>
 <link rel="author" title="Manuel Rego Casasnovas" href="mailto:rego@igalia.com">
 <link rel="help" href="https://drafts.csswg.org/css-contain-1/#containment-size">
-<link rel="match" href="../reference/pass_if_pass_below.html">
+<link rel="match" href="reference/pass_if_pass_below_clipped.html">
 <meta name=assert content="Size containment doesn't apply to table-row-group elements.">
 <style>
 div {
   display: table-row-group;
   contain: size;
+}
+section {
   overflow: hidden;
 }
 </style>
 
 <p>Test passes if there is the word "PASS" below.</p>
+<section>
 <div>PASS</div>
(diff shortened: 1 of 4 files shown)


frivoal:
this is now rebased, and ready to review / merge

dholbert (approved):

PR #12489, opened 2018-08-15
Fieldset should support CSS Flexbox

zcorpan:
Bugs:
https://bugs.chromium.org/p/chromium/issues/detail?id=375693
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/4511145/


annevk (approved):
@emilio might wanna double check, but this looks good to me.

emilio (approved):
LGTM, thanks!

PR #12490, opened 2018-08-15
XHR: statusText when using H2

### LINKED ISSUE #7274: HTTP/2 tests for statusText
HTTP/2 doesn't have a HTTP's reason phrase concept and doesn't really define what it should be in legacy APIs either... So in https://github.com/whatwg/fetch/issues/599 I decided to side with Chrome and make it the empty byte sequence. However, that cannot really be tested at the moment.

annevk:
Fixes #7274.


annevk:
This cannot land as H2 is not enabled by default.

andreubotella (commented):
Why two different files? They seem identical other than the comment, and `.window 2.js` won't generate an HTML boilerplate.

annevk:
Domenic at some point expressed a preference not to have fetch and XHR in the same file. I wouldn't be opposed to merging things more, but that'd be some work.

andreubotella:
> Domenic at some point expressed a preference not to have fetch and XHR in the same file. I wouldn't be opposed to merging things more, but that'd be some work.

The two files are identical other than the initial comment, though.

annevk:
My apologies, I blame iCloud and not paying close enough attention.

andreubotella (approved):
LGTM

PR #12491, opened 2018-08-15
Shadow DOM: test more non-safelisted elements in attachShadow

zcorpan:
See https://github.com/whatwg/html/issues/3920.


rniwa (approved):
r=me.

Issue #12371, opened 2018-08-09
Image not found in CSS built test suite


mrego:
Hi @plinss, we have a css-containt test that fails to load the image in the built test suite:
http://test.csswg.org/harness/test/css-contain-1_dev/single/contain-size-013/format/html5/

It uses an image with the following absoulte path: `/css/support/60x60-red.png`.

Could we serve that folder `/css/support/` so we don't need to duplicate the image once more in a subfolder under `/css/css-contain/`.
There are 39 copies of that image right now in the repository.

Thanks!

TalbotG:
+1

plinss:
Done.

Sorry for the delay, I’ve been swamped with work.

PR #12492, opened 2018-08-15
[css-contain] Add support file in local directory

frivoal:
It is very annoying that we have to keep on replicating the support files all around, but for now, given how the build system works, we do have to.

I'll be the first to be happy when this annoyance is solved, but until it is, we still need functional tests.

http://test.csswg.org/suites/css-contain-1_dev/nightly-unstable/html/contain-size-013.htm needs this file.


ewilligers (approved):

mrego:
I'm not sure this is actually useful as the test still uses an absolute path `/css/support/` to reference that file.

Apart from that, the issue with this particular test has been fixed in #12371 by serving that folder.

Should we revert this change?

frivoal:
I think the path was to the relative file when I made this patch, so maybe another fix went in parallel, making this one useless. Or maybe I was just not paying attention :)

Either way, yes, we can revert.

PR #12494, opened 2018-08-15
HTML: fieldset used value for CSS 'display'

zcorpan:
See https://github.com/whatwg/html/issues/3915.


annevk (approved):
Some kind of note somewhere that the computed value is not the used value would help.

Especially "display: x is not supported" might give folks the impression that they need to support x for fieldset.

zcorpan:
I pushed a commit with an explanation.

annevk:
Thanks!

PR #12501, opened 2018-08-15
HTML: default style for fieldset

zcorpan:
See https://github.com/whatwg/html/issues/3921.

### DIFF
--- a/html/rendering/non-replaced-elements/the-fieldset-element-0/fieldset-default-style.html
+++ b/html/rendering/non-replaced-elements/the-fieldset-element-0/fieldset-default-style.html
@@ -20,7 +20,7 @@
   padding-inline-start: 0.75em;
   */
   min-width: min-content;
-  /* TODO change the above to min-inline-size per https://github.com/whatwg/html/pull/3914 */
+  /* TODO change the above to min-inline-size when it's widely supported. */
 }
 </style>
 <fieldset id=test></fieldset>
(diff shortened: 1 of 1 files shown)


zcorpan:
@annevk the spec change was merged. Can you review this also?

html/rendering/non-replaced-elements/the-fieldset-element-0/fieldset-default-style.html
```diff
@@ -0,0 +1,61 @@
...
+  border: groove 2px ThreeDFace;
+  padding: 0.35em 0.75em 0.625em 0.75em;
+  /* TODO replace above declarations with these when they are widely supported.
+  padding-block-start: 0.35em;
+  padding-inline-end: 0.75em;
+  padding-block-end: 0.625em;
+  padding-inline-start: 0.75em;
+  */
+  min-width: min-content;
+  /* TODO change the above to min-inline-size per https://github.com/whatwg/html/pull/3914 */
```
annevk: Should this be changed already? I think Fx is close to shipping this at least.

annevk (approved):

html/rendering/non-replaced-elements/the-fieldset-element-0/fieldset-default-style.html
zcorpan: min-inline-size is not yet supported in Safari or Edge, but I can update the comment.

PR #12502, opened 2018-08-15
Add MediaStreamTrack Content Hints IDL file + test

foolip:


foolip:
These tests pass on Chrome Dev with experimental web platform features enabled.

mst-content-hint/idlharness.window.js
```diff
@@ -0,0 +1,19 @@
...
+idl_test(
+  ['mst-content-hint'],
+  ['mediacapture-streams', 'dom'],
+  async idl_array => {
+    idl_array.add_objects({
+      MediaStreamTrack: ['audioTrack', 'videoTrack'],
+    });
+
+    const stream = await getNoiseStream({ audio: true, video: true });
+    self.audioTrack = stream.getAudioTracks()[0];
```
lukebjerring: Nit: Pull `stream.getAudioTracks()[0]` into a variable.

mst-content-hint/idlharness.window.js
```diff
@@ -0,0 +1,19 @@
...
+    self.videoTrack = stream.getAudioTracks()[0];
```
lukebjerring: The above nit assumes it was intended to re-use the same value. Is it? Or is there a getVideoTracks equivalent?

lukebjerring (approved):

mst-content-hint/idlharness.window.js
foolip: But it has to be on the global object for `add_objects` to work, I presume using `eval()`. Am I misunderstanding?

mst-content-hint/idlharness.window.js
foolip: Oops, copypasta, will fix.

mst-content-hint/idlharness.window.js
lukebjerring: DW - Irrelevant given that there should have been 2 different values :)

alvestrand (approved):

PR #12505, opened 2018-08-15
[HTTP/2.0] Extra safety checks to fix crash during test

dheiberg:
Developing tests for the H2 server revealed some bugs. Also notably the server did not catch unexpected exceptions and would not log them.

### DIFF
--- a/tools/wptserve/wptserve/server.py
+++ b/tools/wptserve/wptserve/server.py
@@ -369,17 +369,19 @@ class Http2WebTestRequestHandler(BaseWebTestRequestHandler):
                             stream_queues[frame.stream_id] = (self.start_stream_thread(frame, queue), queue)
                         stream_queues[frame.stream_id][1].put(frame)
 
-                        if isinstance(frame, StreamEnded) or frame.stream_ended:
+                        if isinstance(frame, StreamEnded) or (hasattr(frame, "stream_ended") and frame.stream_ended):
                             del stream_queues[frame.stream_id]
 
         except (socket.timeout, socket.error) as e:
-            self.logger.debug('(%s) ERROR - Closing Connection - \n%s' % (self.uid, str(e)))
+            self.logger.error('(%s) Closing Connection - \n%s' % (self.uid, str(e)))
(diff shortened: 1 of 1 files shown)


tools/wptserve/wptserve/server.py
```diff
@@ -378,8 +378,10 @@ def handle_one_request(self):
                 self.close_connection = True
                 for stream_id, (thread, queue) in stream_queues.items():
                     queue.put(None)
+        except Exception as e:
+            self.logger.debug('(%s) ERROR - Unexpected Error - \n%s' % (self.uid, str(e)))
```
jgraham: Why isn't this just `self.logger.error` ?

dhdavvie: oh yes that would make more sense...

dhdavvie: Updated this, and also changed the other exception catch to `self.logger.error` aswell

tools/wptserve/wptserve/server.py
```diff
@@ -369,17 +369,19 @@ def handle_one_request(self):
                             stream_queues[frame.stream_id] = (self.start_stream_thread(frame, queue), queue)
                         stream_queues[frame.stream_id][1].put(frame)
 
-                        if isinstance(frame, StreamEnded) or frame.stream_ended:
+                        if isinstance(frame, StreamEnded) or (hasattr(frame, "stream_ended") and frame.stream_ended):
                             del stream_queues[frame.stream_id]
 
         except (socket.timeout, socket.error) as e:
-            self.logger.debug('(%s) ERROR - Closing Connection - \n%s' % (self.uid, str(e)))
+            self.logger.error('(%s) ERROR - Closing Connection - \n%s' % (self.uid, str(e)))
```
jgraham: The formatter will add `ERROR` to the string in this case, so you don't need to.

dhdavvie: Ahh yes, brain is elsewhere

jgraham (approved):

PR #12525, opened 2018-08-16
HTML: fieldset containing block when it has a legend

zcorpan:
Bugs:
https://bugzilla.mozilla.org/show_bug.cgi?id=942341
https://bugs.webkit.org/show_bug.cgi?id=10821

### DIFF
--- a/html/rendering/non-replaced-elements/the-fieldset-element-0/fieldset-containing-block-ref.html
+++ b/html/rendering/non-replaced-elements/the-fieldset-element-0/fieldset-containing-block-ref.html
@@ -2,7 +2,7 @@
 <title>Reference for fieldset containing block</title>
 <style>
 p { margin: 0; height: 100px }
-div { position: absolute; top: 108; width: 100px; height: 100px; background: lime; }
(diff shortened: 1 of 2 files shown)


html/rendering/non-replaced-elements/the-fieldset-element-0/fieldset-containing-block-ref.html
```diff
@@ -0,0 +1,8 @@
+<!DOCTYPE html>
+<title>Reference for fieldset containing block</title>
+<style>
+p { margin: 0; height: 100px }
+div { position: absolute; top: 108; width: 100px; height: 100px; background: lime; }
```
annevk: Presumably this should be 108px? If this worked regardless, something seems off?

zcorpan: It gets positioned where it would be if it was statically positioned. But it should say px of course.

zcorpan: fixed

annevk (approved):

mstensho (approved):

PR #12527, opened 2018-08-16
HTML: fieldset's legend and position: relative

zcorpan:
The legend should mask the border at its static position, but
Edge moves the mask if it is relatively positioned.


zcorpan:
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/18695616/

html/rendering/non-replaced-elements/the-fieldset-element-0/legend-position-relative-ref.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<title>Reference for legend position: relative</title>
+<style>
+div { display: inline-block; width: 100px; background: lime; }
```
annevk: Non-blocking nit: It seems you can omit the width property here.

annevk (approved):

PR #12537, opened 2018-08-16
Remove the Budget API tests

beverloo:
The specification has been made obsolete with no existing
implementations. We can revive these tests from history if that changes.

https://wicg.github.io/budget-api/


foolip (approved):

foolip:
Thanks for the cleanup! I've sent https://github.com/tidoust/reffy/pull/122 so that we don't try to add the IDL test back :)

PR #12557, opened 2018-08-18
[serve] Correct error detection and add more log messages

jugglinmike:
A more advanced solution might involve attempting to re-start the failed
subprocess(es). I think we'd need to understand the failure conditions before
committing to something like that, though. Otherwise, we'll risk enabling a
"continual restart" state if the problem can't be resolved by creating a new
sub-process.

I wrote the new test to a dedicated file because as an integration-level test,
it's quite a bit slower than the existing tests. It's also a little messy
because it stubs out a module method. I'm open to suggestions about better ways
to test this!

### DIFF
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -827,7 +827,7 @@ def run(**kwargs):
             servers = start(config, build_routes(config["aliases"]), **kwargs)
 
             try:
-                while any(item.is_alive() for item in iter_procs(servers)):
+                while all(item.is_alive() for item in iter_procs(servers)):
                     for item in iter_procs(servers):
                         item.join(1)
             except KeyboardInterrupt:
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -378,15 +378,17 @@ def build_routes(aliases):
 
 
 class ServerProc(object):
-    def __init__(self):
+    def __init__(self, scheme=None):
         self.proc = None
         self.daemon = None
         self.stop = Event()
+        self.scheme = scheme
 
     def start(self, init_func, host, port, paths, routes, bind_address, config, **kwargs):
         self.proc = Process(target=self.create_daemon,
                             args=(init_func, host, port, paths, routes, bind_address,
                                   config),
+                            name='%s on port %s' % (self.scheme, port),
                             kwargs=kwargs)
         self.proc.daemon = True
         self.proc.start()
@@ -506,7 +508,7 @@ def start_servers(host, ports, paths, routes, bind_address, config, **kwargs):
                          "ws":start_ws_server,
                          "wss":start_wss_server}[scheme]
 
-            server_proc = ServerProc()
+            server_proc = ServerProc(scheme=scheme)
             server_proc.start(init_func, host, port, paths, routes, bind_address,
                               config, **kwargs)
             servers[scheme].append((port, server_proc))
@@ -830,6 +832,13 @@ def run(**kwargs):
                 while all(item.is_alive() for item in iter_procs(servers)):
                     for item in iter_procs(servers):
                         item.join(1)
+                exited = [item for item in iter_procs(servers) if not item.is_alive()]
+                subject = "subprocess" if len(exited) == 1 else "subprocesses"
+
+                logger.info("%s %s exited:" % (len(exited), subject))
+
+                for item in iter_procs(servers):
+                    logger.info("Status of %s:\t%s" % (item.name, "running" if item.is_alive() else "not running"))
             except KeyboardInterrupt:
                 logger.info("Shutting down")
 
--- a/tools/serve/test_functional.py
+++ b/tools/serve/test_functional.py
@@ -1,0 +1,77 @@
+try:
+    from importlib import reload
+except ImportError:
+    pass
+import json
+import os
+try:
+    import Queue as queue
+except ImportError:
+    import queue
+import tempfile
+import threading
+
+import pytest
+
+from . import serve
+from wptserve import logger
+
+
+class ServerProcSpy(serve.ServerProc):
+    instances = None
+
+    def start(self, *args, **kwargs):
+        result = super(ServerProcSpy, self).start(*args, **kwargs)
+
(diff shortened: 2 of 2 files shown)


jugglinmike:
The new test interacts with the existing tests because the server implements logging as a singleton:

>     Exception in thread Thread-1:
>     Traceback (most recent call last):
>       File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
>         self.run()
>       File "/usr/lib/python2.7/threading.py", line 754, in run
>         self.__target(*self.__args, **self.__kwargs)
>       File "/home/mike/projects/bocoup/google-wpt/web-platform-tests/tools/serve/test_functional.py", line 49, in target
>         serve.run(config_path=tempfile_name)
>       File "/home/mike/projects/bocoup/google-wpt/web-platform-tests/tools/serve/serve.py", line 809, in run
>         set_logger(logger)
>       File "/home/mike/projects/bocoup/google-wpt/web-platform-tests/tools/wptserve/wptserve/logger.py", line 23, in set_logger
>         raise Exception("Logger must be set at most once")
>     Exception: Logger must be set at most once

I'm out of time for the day, but I'll try to find a workaround by Monday.

tools/serve/serve.py
```diff
@@ -378,15 +378,17 @@ def build_routes(aliases):
 
 
 class ServerProc(object):
-    def __init__(self):
+    def __init__(self, name=None):
```
foolip: If any other review wonders, yes there is another case `wrapper = ServerProc()` where `name` isn't given, and that seems fine.

tools/serve/serve.py
```diff
@@ -827,12 +829,15 @@ def run(**kwargs):
...
             try:
-                while any(item.is_alive() for item in iter_procs(servers)):
+                while all(item.is_alive() for item in iter_procs(servers)):
                     for item in iter_procs(servers):
                         item.join(1)
             except KeyboardInterrupt:
                 logger.info("Shutting down")
 
+        for item in iter_procs(servers):
+            logger.info("Status of %s:\t%s" % (item.name, "running" if item.is_alive() else "not running"))
```
foolip: If `item.name` is `None` here the message won't be great. Is that impossible? If not, maybe just require `name` after all?

tools/serve/serve.py
```diff
@@ -827,12 +829,15 @@ def run(**kwargs):
...
 
             try:
-                while any(item.is_alive() for item in iter_procs(servers)):
+                while all(item.is_alive() for item in iter_procs(servers)):
                     for item in iter_procs(servers):
                         item.join(1)
             except KeyboardInterrupt:
                 logger.info("Shutting down")
 
+        for item in iter_procs(servers):
```
foolip: A sample of the output I see:
```
^CINFO:web-platform-tests:Shutting down
INFO:web-platform-tests:Status of ws on port 61486:	running
INFO:web-platform-tests:Status of wss on port 61487:	running
INFO:web-platform-tests:Status of http on port 8000:	not running
INFO:web-platform-tests:Status of http on port 61488:	not running
INFO:web-platform-tests:Status of https on port 8443:	not running
```

As a user of `./wpt serve` manually, I'm not sure what to do with this on shutdown, if it's cause for concern or not. Do you think these make more sense as debug messages?

Alternatively, if the reason we shut down is that some subprocess has died, can we log that fact clearly, and just in that case?

foolip (commented):
I've reviewed without the final "DEBUG" commit, assuming nothing in that commit should stay.

tools/serve/serve.py
jugglinmike: Thanks for pointing this out. Including the port number in the name is a litle redundant since that's already defined via an instance property. Because that property *is* available on all instances, we can name the process with something slightly useful even when additional information (i.e. the scheme) is not specified.

tools/serve/serve.py
jugglinmike: I prefer the second alternative because this information does seem relevant to the typical user of `wpt serve`, just not so much in the "Ctrl-C" case. Here's what it looks like now (printed only in response to subprocess exit):

>     INFO:web-platform-tests:2 subprocesses exited:
>     INFO:web-platform-tests:Status of ws on port 44283:     running
>     INFO:web-platform-tests:Status of wss on port 44819:    running
>     INFO:web-platform-tests:Status of http on port 8000:    not running
>     INFO:web-platform-tests:Status of http on port 40161:   running
>     INFO:web-platform-tests:Status of https on port 8443:   not running

tools/serve/serve.py
```diff
@@ -827,9 +829,17 @@ def run(**kwargs):
...
                     for item in iter_procs(servers):
                         item.join(1)
+                exited = filter(lambda item: not item.is_alive(),
+                                iter_procs(servers))
+                subject = "subprocess" if len(exited) == 1 else "subprocesses"
+
+                logger.info("%s %s exited:" % (len(exited), subject))
+
+                for item in iter_procs(servers):
+                    logger.info("Status of %s:\t%s" % (item.name, "running" if item.is_alive() else "not running"))
```
foolip: Ah, when removed from the ctrl+c case, I agree this is useful info. Thanks!

foolip (approved):

jugglinmike:
Regarding the trouble with the logger: I initially tried to resolve this by running the test in a subprocess, but [the particulars of multiprocessing in Python](https://kaushikghose.wordpress.com/2016/08/26/python-global-state-multiprocessing-and-other-ways-to-hang-yourself/) mean that "use a subprocess" is not a silver bullet.

Instead, I'm using `reload` to forcibly reset the logging module's internal state prior to testing. Just like overriding `ServerProc`, this seems like an unhealthy coupling between the tests and the implementation, but I can't think of a more hygienic solution.

This branch accrued a number of debugging commits because working this out took a bit of trial-and-error. My apologies to the folks who received notifications for all that. I've rebased and squashed everything back down (including the changes @foolip recommended). [The original version of this branch is available here.](https://github.com/web-platform-tests/wpt/compare/master...bocoup:serve-shutdown-prebase) This branch should be ready to merge.

Issue #12559, opened 2018-08-18
Broken .https.any.serviceworker. for non-https .any.js remain in MANIFEST.json


foolip:
After https://github.com/web-platform-tests/wpt/pull/12381, the MANIFEST.json from `./wpt manifest-download` has not been correctly updated.

This was in fact anticipated by @jgraham in https://github.com/web-platform-tests/wpt/issues/11105#issuecomment-411788257:
> One possible solution is just to make `.serviceworker.` imply that the test is loaded over https directly without also having to add `.https.` to the url. Assuming that works the way I think it ought to, the "only" problem is that it breaks incremental manifest updates because the tests generated are changed but the file content remains the same.

Diff of MANIFEST.json from `./wpt manifest-download` and `./wpt manifest --no-download` at 6aef6676d2f95c29de17666cc70d61b517939fbf:
<details>
<pre>
+--- MANIFEST-download.json	2018-08-19 00:24:25.477679434 +0200
+++ MANIFEST-gen.json	2018-08-19 00:27:03.738704485 +0200
@@ -340944,15 +340944,15 @@
      {}
     ],
     [
-     "/encoding/idlharness.any.sharedworker.html",
+     "/encoding/idlharness.any.serviceworker.html",
      {}
     ],
     [
-     "/encoding/idlharness.any.worker.html",
+     "/encoding/idlharness.any.sharedworker.html",
      {}
     ],
     [
-     "/encoding/idlharness.https.any.serviceworker.html",
+     "/encoding/idlharness.any.worker.html",
      {}
     ]
    ],
@@ -349726,15 +349726,15 @@
      {}
     ],
     [
-     "/fetch/api/abort/general.any.sharedworker.html",
+     "/fetch/api/abort/general.any.serviceworker.html",
      {}
     ],
     [
-     "/fetch/api/abort/general.any.worker.html",
+     "/fetch/api/abort/general.any.sharedworker.html",
      {}
     ],
     [
-     "/fetch/api/abort/general.https.any.serviceworker.html",
+     "/fetch/api/abort/general.any.worker.html",
      {}
     ]
    ],
@@ -351484,15 +351484,15 @@
      {}
     ],
     [
-     "/hr-time/idlharness.any.sharedworker.html",
+     "/hr-time/idlharness.any.serviceworker.html",
      {}
     ],
     [
-     "/hr-time/idlharness.any.worker.html",
+     "/hr-time/idlharness.any.sharedworker.html",
      {}
     ],
     [
[...]

PR #12560, opened 2018-08-18
Use --no-download when uploading manifest from Travis

foolip:
This will be a lot slower than using the previous manifest, but it will mean
that issues like https://github.com/web-platform-tests/wpt/issues/12559 can't
persist over time.

### DIFF
--- a/tools/ci/ci_manifest.sh
+++ b/tools/ci/ci_manifest.sh
@@ -7,7 +7,7 @@ cd $WPT_ROOT
 mkdir -p ~/meta
 
 python tools/ci/tag_master.py
-./wpt manifest -p ~/meta/MANIFEST.json
+./wpt manifest --no-download -p ~/meta/MANIFEST.json
 cp ~/meta/MANIFEST.json $WPT_MANIFEST_FILE
 # Force overwrite of any existing file
 gzip -f $WPT_MANIFEST_FILE


gsnedders:
Do we not want this anyway? It avoids another remote dependency, and with the manifest getting quicker it's ever less of a performance hit?

foolip:
How long does it take on Travis, and do you expect it to become quicker?

gsnedders:
#12182 made it much faster, assuming we have some baseline (which I think we have from the Travis cache anyway?), and #12553 will make it faster still.

foolip:
If it can be made reasonable fast then I would like to run with `--no-download` in Travis. If nothing else, then to check that it gives the same result as an incremental update, and failing the build if it doesn't.

jgraham:
So there are three options for Travis:
* Full manifest rebuild always
* Invremental based on the downloaded manifest
* Incremental based on what happens to be in the Travis cache

It's unclear to me that the second and third options are meaningfully different from the point of view of testing that the result is correct. But they do differ in that with a download we implicitly test the download codepath. If GitHub is down I think we a) have bigger problems and b) fall back to doing an incremental update anyway, so I don't think there's much difference from a robustness point of view.

If we want to test that the incremental update is doing the right thing, we should compare the rebuilt manifest with the updated one. But that has a significant time penalty (8 minutes on my machine to rebuild the whole thing) and probably relies on moving away from `html5lib` to get significant performance improvements, which afaik no one is planning to work on (I tried replacing it with `html5ever` and `html5-parser` and got big performance wins and bogus output, whereupon I stopped investigating since neither is in a state that makes them easy to distribute).

foolip:
Oh @gsnedders was saying it's faster if we have some baseline, i.e. an existing MANIFEST.json, full rebuild hasn't gotten faster.

Maybe we could run a daily Travis job to do some more expensive checks like verifying that a full rebuild works, but I guess it's not very important.

foolip:
https://travis-ci.org/web-platform-tests/wpt/jobs/418141503 is the job that ran when https://github.com/web-platform-tests/wpt/pull/12563 was merged and it did have to rebuild the manifest as expected. The whole job took 5 min 28 sec, which is faster than I thought.

PR #12561, opened 2018-08-19
Add two selector tests for :nth-child selector with whitespace in the argument

jkillian:
Over in [nwsapi](https://github.com/dperini/nwsapi) (which powers [jsdom](https://github.com/jsdom/jsdom)), we found [an issue](https://github.com/dperini/nwsapi/issues/20) where having space in an `nth-child` selector (e.g. `:nth-child(2n - 1)`) was causing the selector to fail.

This PR adds two lines of tests to ensure that this scenario is covered. I verified that selectors with spaces like these two work in FF and Chrome and also that [the spec](https://drafts.csswg.org/selectors-3/#nth-child-pseudo) allows whitespace here.

Note that this PR is most easily reviewed by hiding whitespace changes: https://github.com/web-platform-tests/wpt/pull/12561/files?utf8=%E2%9C%93&diff=unified&w=1


annevk (approved):
Should we test different kinds of whitespace?

JKillian:
Updated the tests to include more forms of whitespace, good idea! I believe the build is failing due to https://github.com/web-platform-tests/wpt/issues/12589, I can retrigger things after that is fixed.

annevk:
Thanks, let me know when that's fixed and I'll land this.

JKillian:
Thanks @annevk, looks like some build caches got cleared over in https://github.com/web-platform-tests/wpt/issues/12635 and things are working now. Lmk if you need any other changes, if not, should be good to merge!

Issue #12549, opened 2018-08-17
/html/semantics/scripting-1/the-script-element/script-charset-02.html fails in all browsers


Zirro:
The [`script-charset-02.html`](https://github.com/web-platform-tests/wpt/blob/master/html/semantics/scripting-1/the-script-element/script-charset-02.html) test fails the same way [in all browsers](https://wpt.fyi/results/html/semantics/scripting-1/the-script-element/script-charset-02.html) and is likely incorrect.

foolip:
@njskalski who added the test

njskalski:
Hi. I recall adding it, but I was not (and am not right now) a specialist in this area, it was just marked as "easy" in Servo project if I remember correctly. I am not able to contribute in any reasonable time because I would have to make sure it does not collide with other agreements I signed and that takes time. Sorry :(

foolip:
Looks from the test like it should be pretty easy to debug and make it pass instead. Marking as good first bug.

foolip:
Fixed in https://github.com/web-platform-tests/wpt/pull/12562, yay!

PR #12562, opened 2018-08-19
[FIXED] /html/semantics/scripting-1/the-script-element/script-charset-02.html Passed

aaqaishtyaq:
Reference issue: #12549 
Tested on Chrome/Firefox/Safari
@foolip @annevk  Please review.

### DIFF
--- a/html/semantics/scripting-1/the-script-element/script-charset-02.html
+++ b/html/semantics/scripting-1/the-script-element/script-charset-02.html
@@ -9,24 +9,23 @@
   <div id="log"></div>
   <!-- to avoid conflating tests for script encoding declaring the encoding at the top of file. i.e, windows-1250-->
   <meta charset="windows-1250">
-
+  <script>
+  test(function() {
+    assert_equals(document.characterSet, "windows-1250")
+  }, "assumption: document encoding is windows-1250");
+  </script>
 
   <!-- in this case, neither response's Content Type nor charset attribute bring correct charset information.
   -->
   <script type="text/javascript"
     src="serve-with-content-type.py?fn=external-script-windows1250.js&ct=text/javascript">
   </script>
-  <script>
-  test(function() {
-    //assumption: document encoding is windows-1250
-    assert_equals(document.characterSet, "windows-1250")
-  });
-  </script>
+
   <script>
   test(function() {
     //these string should match since, windows-1250 is the fallback encoding.
     assert_equals(window.getSomeString(), "\u015b\u0107\u0105\u017c\u017a");
-  });
+  }, "windows-1250 script decoded using document encoding (also windows-1250)");
   </script>
 
   <script type="text/javascript"
@@ -36,7 +35,7 @@
   //these strings should match, since this string is the result of decoding the utf-8 text as windows-1250.
   test(function() {
     assert_equals(window.getSomeString(), "\u0139\u203a\xc4\u2021\xc4\u2026\u0139\u013d\u0139\u015f");
-  });
+  }, "UTF-8 script decoded using document encoding (windows-1250)");
   </script>
 
 </head>
--- a/html/semantics/scripting-1/the-script-element/script-charset-02.html
+++ b/html/semantics/scripting-1/the-script-element/script-charset-02.html
@@ -23,7 +23,7 @@
   <script>
   test(function() {
     //these strings should not match, since the tested file is in windows-1250, and fallback is defined as utf-8
-    assert_not_equals(window.getSomeString().length, 5);
+    assert_not_equals(window.getSomeString().length, 10);
   });
   </script>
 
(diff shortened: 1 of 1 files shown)


foolip:
Thanks for working on this! This makes the test pass, but I noticed some of a wrong. This is testing this part of the spec: https://html.spec.whatwg.org/#fetch-a-classic-script

And the encoding is determined at step 15 of https://html.spec.whatwg.org/#prepare-a-script.

What it boils down to is that if `<script charset="valid-encoding">` is used, then that encoding is used, otherwise it's the document's encoding. That in turn is defined in https://html.spec.whatwg.org/#determining-the-character-encoding and is usually "windows-1252".

Surprisingly there aren't many tests for this, so thanks for improving this one. Based on the above, I'll suggest a few edits for this test in particular.

html/semantics/scripting-1/the-script-element/script-charset-02.html
```diff
@@ -23,7 +23,7 @@
   <script>
   test(function() {
     //these strings should not match, since the tested file is in windows-1250, and fallback is defined as utf-8
-    assert_not_equals(window.getSomeString().length, 5);
+    assert_not_equals(window.getSomeString().length, 10);
```
foolip: Rather than checking the string length, could you check the exact value? `assert_equals(window.getSomeString(), "\u015b\u0107\u0105\u017c\u017a");` in script-charset-03.html is an example.

html/semantics/scripting-1/the-script-element/script-charset-02.html
```diff
@@ -23,7 +23,7 @@
   <script>
   test(function() {
     //these strings should not match, since the tested file is in windows-1250, and fallback is defined as utf-8
```
foolip: This comment and the TODO at the top of this file is wrong about the fallback encoding. It's windows-1252 for this test on my locale, but can be other values. To avoid conflating tests for script encoding with the default document encoding, I suggest simply declaring the encoding at the top of the file. I suggest `<meta charset="windows-1250">` (rather than 1252) so that one of the scripts is decoded correctly, and here, `assert_equals(document.characterSet, "windows-1250")` as a precondition for the test.

With those changes, the main value of this test will that even if the script can be decoded as UTF-8, it will still be decoded using the document's encoding.

html/semantics/scripting-1/the-script-element/script-charset-02.html
```diff
@@ -33,7 +33,7 @@
   <script>
   //these strings should match, since fallback utf-8 is the correct setting.
```
foolip: This comment is also wrong, but the assert itself is good, but even better would be to replace it with `"\u0139\u203a\xc4\u2021\xc4\u2026\u0139\u013d\u0139\u015f"` which should be what the UTF-8 decodes to if interpreted as windows-1250.

foolip:
Oh, right, if you can also point to https://html.spec.whatwg.org/multipage/#fetch-a-classic-script instead of https://html.spec.whatwg.org/multipage/#scriptingLanguages that'd be good.

html/semantics/scripting-1/the-script-element/script-charset-02.html
aaqaishtyaq: should I delete the comment or you suggest.
by the way thanks for guiding me.

aaqaishtyaq:
@foolip tests passed on my machine locally. what should I do now?

foolip (commented):
Almost there, thanks for the updates!

html/semantics/scripting-1/the-script-element/script-charset-02.html
```diff
@@ -33,7 +42,7 @@
...
   test(function() {
-    assert_equals(window.getSomeString().length, 5);
+    assert_equals(window.getSomeString(), "\u0139\u203a\xc4\u2021\xc4\u2026\u0139\u013d\u0139\u015f");
   });
```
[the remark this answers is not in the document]
foolip: For this comment I meant adding "UTF-8 script decoded using document encoding (windows-1250)" as the title, i.e. the second argument of `test(fn, title)`. I'll push a commit to show what I mean for this and the other tests.

aaqaishtyaq:
Okay now I get it. Thanks.
Now I need to change anything or not?

foolip (approved):
Tried this locally and it looks good to me!

foolip:
Thank you @aaqaishtyaq, I've merged this now. I squashed into a single commit with the title of the PR. (If you'd like to control exactly what the commit message is, what you can do is squash everything to a single commit and force push when the PR is ready to be merged, or leave a comment on the PR.)

Also, if you thought this was fun, you'll find that there's some very similar problems in script-charset-01.html and script-charset-03.html, where the test titles in particular could use improving. If those are fixed to describe what is really being tested, it might become apparent that some interesting combinations are still untested.

PR #12565, opened 2018-08-20
[wptserver] Correct value in unit test

jugglinmike:
This correction was previously submitted via gh-12459. Since the patch which that pull request intended to fix has since been reverted, it's unclear if/how the pull request can be accepted. I'm re-submitting the commit in a dedicated pull request because it corrects an unrelated mistake.

Since this concerns a Windows-specific codepath, TravisCI won't be able to validate the change. The following code may be manually inserted into the test module to verify expected behavior using a non-Windows system:

```python
def uname():
    return ('Windows',)
platform.uname = uname
```

### DIFF
--- a/tools/serve/test_serve.py
+++ b/tools/serve/test_serve.py
@@ -35,7 +35,7 @@ def test_make_hosts_file_windows():
                        browser_host="foo.bar",
                        alternate_hosts={"alt": "foo2.bar"},
                        subdomains={"a", "b"},
-                       not_subdomains={"x, y"}) as c:
+                       not_subdomains={"x", "y"}) as c:
         hosts = serve.make_hosts_file(c, "192.168.42.42")
         lines = hosts.split("\n")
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
@thejohnjansen FYI that we have no checks at all running on Windows, making problems like this a lot more likely. If we wanted to fix this, is there any other option than https://github.com/marketplace/appveyor for Windows CI that you know of?

jugglinmike:
Thanks, @foolip 

For this particular case, we *could* extend `make-hosts-file` with a `--platform` option. That would allow us to test both branches from any machine. I'm not a an of adding features only for testability's sake, but there's at least a theoretical use case for doing this: a consumer might want to generate a file in their dev environment and upload it to a production environment. Then again, `wpt serve` and `wpt make-hosts-file` are coupled, so there may not actually be a situation where you can run the server but not the hosts file generator.

PR #12566, opened 2018-08-20
Add tests for cross-origin redirects getting back to the original origin

yutakahirano:
This is discussed at https://github.com/whatwg/fetch/issues/737, but
currently implementors have a somewhat consistent behavior. I will
update the test if we change the behavior.

### DIFF
--- a/fetch/api/redirect/redirect-back-to-original-origin.any.js
+++ b/fetch/api/redirect/redirect-back-to-original-origin.any.js
@@ -1,0 +1,37 @@
+// META: script=/common/get-host-info.sub.js
+
+const BASE = location.href;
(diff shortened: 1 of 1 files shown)


jdm (approved):

annevk:
I think `fetch/api/redirect/redirect-mode.any.js` already tests this, but I guess it's okay to have more.

yutakahirano:
I don't think so. At least, when I tried to change Chrome behavior no wpt test detected the change: https://chromium-review.googlesource.com/c/chromium/src/+/1179107

annevk:
Ah, indeed, my test doesn't cross the origin boundary due to redirects. Weird that we didn't have any tests for that...

PR #12567, opened 2018-08-20
HTML: test fieldset's legend with float and position

zcorpan:
cc @MatsPalmgren


html/rendering/non-replaced-elements/the-fieldset-element-0/legend-float-abspos.html
```diff
@@ -0,0 +1,87 @@
...
+  <legend id=legend2>legend2</legend>
+  <legend id=legend3>legend3</legend>
+  <legend id=legend4>legend4</legend>
+  <legend id=legend5>legend5</legend>
+</fieldset>
+<script>
+ const fieldset = document.getElementById('fieldset');
+ const legends = document.getElementsByTagName('legend');
+ const [legend1, legend2, legend3, legend4, legend5] = legends;
+ const expectedTop = 8;
```
emilio: This can be `0` if you add a `body { margin: 0 }` rule, which could make this clearer. Otherwise please add a comment to explain it?

emilio (approved):

PR #12568, opened 2018-08-20
Test if speechSynthesis.speak() works with SSML

foolip:
This test takes a roundabout way of checking if SSML appears to be
supported, but should work with reasonable assumptions made.

### DIFF
--- a/speech-api/SpeechSynthesis-speak-SSML.html
+++ b/speech-api/SpeechSynthesis-speak-SSML.html
@@ -1,0 +1,48 @@
+<!DOCTYPE html>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/resources/testdriver.js"></script>
+<script src="/resources/testdriver-vendor.js"></script>
+<script>
+// The shortest possible valid SSML document to speak "one":
+const ssmlText = `<speak version="1.1"
+       xmlns="http://www.w3.org/2001/10/synthesis" xml:lang="en-US">
+  one
+</speak>`;
+
+// A precondition test to ensure the SSML can be parsed.
+// The resulting document isn't used in the actual test.
+test(() => {
+  const parser = new DOMParser();
+  const doc = parser.parseFromString(ssmlText, 'text/xml');
+  assert_equals(doc.documentElement.textContent.trim(), 'one');
+}, 'precondition: valid XML document');
+
+async_test(t => {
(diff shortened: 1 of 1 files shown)


foolip:
Tested manually in Chrome and Firefox where SSML apparently is not supported.

foolip:
Now also tested manually in Edge and Safari and failing there. In every case, something like "speech version one point one" was spoken, i.e. the raw XML document read aloud.

speech-api/SpeechSynthesis-speak-SSML.html
```diff
@@ -0,0 +1,48 @@
...
+// The resulting document isn't used in the actual test.
+test(() => {
+  const parser = new DOMParser();
+  const doc = parser.parseFromString(ssmlText, 'text/xml');
+  assert_equals(doc.documentElement.textContent.trim(), 'one');
+}, 'precondition: valid XML document');
+
+async_test(t => {
+  // The assumption of this test is that if SSML is *not* supported at all,
+  // then "less than speak version equals one point one ..." will be spoken,
```
saschanaz: Currently [the spec says](https://w3c.github.io/speech-api/#utterance-attributes):

>For speech synthesis engines that do not support SSML, or only support certain tags, the user agent or speech engine must strip away the tags they do not support and speak the text.

I think this means browsers without SSML support should also just speak "one", assuming that the *tags* also include `<speak>`.

foolip: That would be a form of support of SSML, really. One could pass this test with such tag-stripping support, it's true.

foolip:
Test folded into https://github.com/web-platform-tests/wpt/pull/12689.

guest271314:
@foolip See https://github.com/WICG/speech-api/issues/10#issuecomment-619589378


> This looks to me like a chicken-and-egg issue.
> 
> If nothing in the spec says whether .text is ssml or not, I don't see how implementations would dare telling the synth that it's ssml when it's not even sure whether it is or not.
> 
> The backends are ready: espeak-ng supports it, speech-dispatcher supports it, it's a one-line change in speech-dispatcher-based browser backends to support it. But if the API does not tell when it should be enabled, we can not dare enabling it.

Am still banned, evidently indefinitely, from WICG. 

The technology is available to achieve the requirement of parsing "**either**" text or SSML. 

As pointed out by @sthibaul the specification needs to be addressed to move this forward.

PR #12569, opened 2018-08-20
HTML: test inline margin on fieldset's legend

zcorpan:
See https://github.com/whatwg/html/issues/3929


annevk (commented):
So the top margin ends up affecting the position of the fieldset, but bottom margin affects the height of the fieldset?

zcorpan:
Yeah, that's the effect of it.

annevk (approved):
Seems weird, but if that's what implementations do...

zcorpan:
It's what Edge does, and what I think is most useful. WebKit and Chromium ignore margin-top, which is not so useful. Gecko positions the fieldset's top border centered around the legend's margin box (as opposed to border box), which is not so visually pleasing as the legend ends up hanging inside or outside the fieldset.

html/rendering/non-replaced-elements/the-fieldset-element-0/legend-margin-block.html
```diff
@@ -0,0 +1,23 @@
...
+ #top { margin-top: 100px; }
+ #bottom { margin-bottom: 100px; }
+ #behind {
+   position: absolute;
+   z-index: -1;
+   background: red;
+   left: 0;
+   right: 0;
+   top: 150px;
+   height: 220px;
```
mstensho: Is this really behavior we want to standardize on? Looks like this test expects the legend's top margin to collapse with that of the fieldset, even though the fieldset has a border, AND both elements establish a BFC.

annevk: FWIW, I'd personally be okay with ignoring margin altogether (or just ignoring margin-top) as well.

zcorpan: I guess we can ignore margin in the block flow direction. @whatwg/fieldset any opinions?

mstensho: I don't find the LEGEND block margins useful at all. Since interoperability is so lousy in this regard anyway, it would be tempting to say that we just ignore them.

zcorpan: OK.

zcorpan: Bug where margin-bottom was implemented for WebKit
https://bugs.webkit.org/show_bug.cgi?id=35981

mstensho: Thanks. There's nothing in that bug report that points out the usefulness of supporting margin-bottom on legend, though, is there? Might as well use padding-top on the fieldset, right?

zcorpan: Yeah, I agree.

dholbert: > Thanks. There's nothing in that bug report that points out the usefulness of supporting margin-bottom on legend, though, is there? Might as well use padding-top on the fieldset, right?

Seems like it -- though in Firefox, you can use margin-bottom values to move the legend outside of the fieldset's border entirely (which you can't do with padding-top, but you could achieve with relative positioning instead).  That special margin-bottom behavior doesn't seem to happen in other browsers, so I'm happy to see that behavior go, anyway.

zcorpan: One could use relative positioning to achieve that if desired.

MatsPalmgren:
Using relative positioning or padding to a simulate a legend margin seems like a hack to me.

I tend to think we should support block-axis margins on the rendered legend. It's the most logical property to reach for when you want to create some distance to the surrounding content and it would be inconsistent if it doesn't work on legends when [it applies to all other boxes (except inlines)](https://www.w3.org/TR/CSS22/box.html#margin-properties). Exceptions like this makes the web platform harder to understand and teach IMO. Margin properties are working fine on `<caption>` and I see no reason why we can't make it work similarly on `<legend>`.

Also, `margin-bottom` works fine in both Firefox and Chrome currently and I see no reason to change that.

dholbert:
> margin-bottom works fine in both Firefox and Chrome currently and I see no reason to change that.

It does something quite different in Firefox vs Chrome right now, FWIW.  Testcase: https://jsfiddle.net/28xhe3sj/  

In Firefox, the legend is way outside the fieldset border there, for reasons that @zcorpan [alluded to above](https://github.com/web-platform-tests/wpt/pull/12569#issuecomment-414391871)
In Chrome, on the other hand, it just looks like there's extra padding at the top of the fieldset's content box.

> Using relative positioning or padding to a simulate a legend margin seems like a hack to me.

The `margin-bottom` usage/behavior feels like more of a hack to me than padding/relpos, FWIW. The effect of `<legend style="margin-bottom:Npx">` in Firefox right now is pretty bizarre -- it offsets the fieldset-border downwards by `N/2 px`, and then it offsets the fieldset's *contents* downward by the remaining `N/2 px`.  If a site really wants the "legend detatched from the fieldset" look that Firefox (and only Firefox) gives them, it seems more direct for them to use relative positioning to shift the legend upwards by a precise amount, and to not arbitrarily offset the fieldset's contents downwards unless they really want that as well.

zcorpan:
@MatsPalmgren can we land this?

zcorpan:
@MatsPalmgren @mstensho PTAL

mstensho:
I can't build the spec anymore, caused by the last commit:

$ ../html-build/build.sh
[...]
Generating HTML variant...
Error: Unused reference: [COMPAT]
Error count: 1
Saving index-html
Splitting...
Saving index.html
Saving introduction.html
[...]
Saving acknowledgements.html
Generating DEV variant...
Warning: Could not find ID selector-read-only for annotation that uses URLs:
   https://caniuse.com/#feat=css-read-only-write
Splitting...
Saving index.html
Saving introduction.html
[...]
Saving acknowledgements.html
Generating SNAP variant...
Error: Unused reference: [COMPAT]
Error count: 1
Saving index-snap

There were errors. Stopping.

zcorpan:
Oops, fixed.

zcorpan:
I had forgotten about this and made https://github.com/web-platform-tests/wpt/pull/13040 so now I've removed the block margin tests here.

mstensho:
Can you rename the commit comment to say "inline margins", then?

PR #12571, opened 2018-08-20
Remove me from the suggested reviewers from XHR

emilio:
Truth is it's been a long time since I last looked at any XHR-related stuff, so I'm unlikely to be useful there.


annevk (approved):

annevk:
Thanks for keeping it updated!

PR #12572, opened 2018-08-20
Convert Input Events IDL test to use idl_test

foolip:


foolip:
Tested locally and found to pass all 9 tests on Chrome, while Firefox fails 6 as this isn't implemented. So the tests now work.

foolip:
Uh, disregard the scroll-animations link given, I was working on multiple conversions and got them mixed up. This test was already working:
https://staging.wpt.fyi/results/input-events/idlharness.html?sha=6aef6676d2&label=stable

johanneswilm:
@foolip Can staging.wpt.fyi be used in a way to check what the test results are if a PR like this one is merged? Or should I be running this locally?

foolip:
@johanneswilm unfortunately there is no easy way to see what the results after a change are, even for those tests that do run on Travis currently. We're still working on https://github.com/web-platform-tests/wpt/issues/7475 to make this better.

johanneswilm (approved):

PR #12575, opened 2018-08-20
Convert Scroll-linked Animations IDL test to use idl_test

foolip:
The current test is broken because of missing dependencies:
https://staging.wpt.fyi/results/scroll-animations/idlharness.html?label=stable&sha=6aef6676d2


birtles (approved):
Thanks!

PR #12576, opened 2018-08-20
Convert Input Device Capabilities IDL test to use idl_test

foolip:
After these changes, Chrome still fails two tests as before:
https://staging.wpt.fyi/results/input-device-capabilities/interfaces.html?sha=6aef6676d2&label=stable


NavidZ:
That's correct. We never implemented the pointerMovementScrolls. I remember nobody asking for it yet but we specced it for the completion of the spec. @mustaqahmed do you have anything to add here? Do we know if there are people needing it?

foolip:
Ping @mustaqahmed. Note that if the spec changes, then the IDL file in wpt will also be updated. @lukebjerring and I are working on making that more automatic.

lukebjerring (approved):

PR #12584, opened 2018-08-21
Add some missing 'var's to urlsearchparams-constructor.any.js.

Ms2ger:


zcorpan (commented):
Can you also remove the multiple `var params` in some tests?

zcorpan (approved):

PR #12587, opened 2018-08-21
HTML: test fieldset's border gap

zcorpan:
Part of https://github.com/whatwg/html/pull/3934

cc @MatsPalmgren


mstensho (approved):
Nice! :)

PR #12593, opened 2018-08-21
Test inheritance of CSS Inline Layout properties

ewilligers:
Check that properties in
https://drafts.csswg.org/css-inline-3/#property-index
inherit or not, as per spec.

e.g. dominant-baseline inherits
https://github.com/w3c/csswg-drafts/issues/2926

### DIFF
--- a/css/css-inline/META.yml
+++ b/css/css-inline/META.yml
@@ -1,0 +1,4 @@
+spec: https://drafts.csswg.org/css-fonts/
+suggested_reviewers:
+  - dauwhe
+  - fantasai
--- a/css/css-inline/inheritance.html
+++ b/css/css-inline/inheritance.html
@@ -6,6 +6,7 @@
 <link rel="author" title="Eric Willigers" href="mailto:ericwilligers@chromium.org">
 <link rel="help" href="https://drafts.csswg.org/css-inline-3/#property-index">
 <meta name="assert" content="Properties inherit or not according to the spec.">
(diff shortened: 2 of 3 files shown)


gsnedders:
@ericwilligers we don't have a lint for it yet, but it would be great if you could add a META.yml with some relevant reviewers!

ewilligers:
Added @dauwhe and @fantasai to META.yml

FremyCompany (commented):
Test looks fine to me, but I wouldn't mind more consistency in the checks being done to better help the reader understand the boilerplate, or move that boilerplate to its own file (either or both would work, just trying to give options) or if not its own file, move the functions to below the asserts, since they are hosted they don't need to be in order; if that unease is just me feel free not to change, I don't want to block anything, just trying to provide some feedback ;)

ewilligers:
For each property, we will want a test to check if it inherits and also want a test to check its initial value.

ewilligers:
The [second commit](https://github.com/web-platform-tests/wpt/pull/12593/commits/fd15cc5478a285bab7373941b1994cc2726c3217) has the changes requested by @FremyCompany. The [third commit](https://github.com/web-platform-tests/wpt/pull/12593/commits/0d0afba85c2d4e02c770d8db3b63f8ca2e1294de) adds tests for initial values.

FremyCompany (approved):
Great :) 
Thanks!

PR #12598, opened 2018-08-21
[Shadow DOM] Fix test using assert_array_equals

pmdartus:
Fix occurences of assert_array_equals where it is used to compare the array length instead of the array itself.

https://staging.wpt.fyi/results/shadow-dom/leaktests/get-elements.html?sha=0398567bac
https://staging.wpt.fyi/results/shadow-dom/Extensions-to-Event-Interface.html?sha=0398567bac

### DIFF
--- a/shadow-dom/Extensions-to-Event-Interface.html
+++ b/shadow-dom/Extensions-to-Event-Interface.html
@@ -54,7 +54,6 @@ function testComposedEvent(mode) {
 
         var expectedPath = ['A1a', 'A1-SR', 'A1', 'A-SR', 'A'];
         assert_array_equals(log.eventPath, expectedPath);
-        assert_array_equals(log.eventPath.length, log.pathAtTargets.length);
         assert_array_equals(log.pathAtTargets[0], expectedPath);
         assert_array_equals(log.pathAtTargets[1], expectedPath);
         assert_array_equals(log.pathAtTargets[2], mode == 'open' ? expectedPath : ['A1', 'A-SR', 'A'],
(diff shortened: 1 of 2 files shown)


shadow-dom/Extensions-to-Event-Interface.html
```diff
@@ -54,7 +54,6 @@
 
         var expectedPath = ['A1a', 'A1-SR', 'A1', 'A-SR', 'A'];
         assert_array_equals(log.eventPath, expectedPath);
-        assert_array_equals(log.eventPath.length, log.pathAtTargets.length);
```
rniwa: We need to replace `assert_array_equals` with `assert_equals`, not remove it.
r- because we clearly need this assertion.

rniwa (changes requested):

shadow-dom/Extensions-to-Event-Interface.html
pmdartus: The `assert_array_equals` is already asserting the length of the array. https://github.com/web-platform-tests/wpt/blob/master/resources/testharness.js#L1052-L1055

pmdartus: My bad, I added the `assert_equals` back.

rniwa (approved):
Thanks!

rniwa:
Why are some checks still pending? It has been so many days, and I don't see anything failing or pending in the travis-ci page.

rniwa:
@jgraham @foolip any idea?

zcorpan:
I'm guessing there was a hickup in communicating back the result. I've restarted the build.

foolip:
I've seen this happen twice before and contacted Travis support about it, see https://github.com/web-platform-tests/wpt/pull/12750#issuecomment-417678899.

Those flukes were 11 days ago, while this seems to have been longer ago. @rniwa, have you seen it on other PRs too? Restarting the build, or just force merging if it was passing, is the immediate fix. As quoted in the linked issue, Travis thinks this will be resolved when we migrate to travis-ci.com.

PR #12599, opened 2018-08-21
Remove duplicate importScripts and testharness.js script in CORP test

youennf:
The WPT server is handling that itself.


annevk:
We don't need fetch-in-service-worker.html either, pushed a fixup for that.

annevk (approved):

PR #12617, opened 2018-08-22
HTML: test that floating legends do not disappear

zcorpan:
Bugs:
https://bugs.chromium.org/p/chromium/issues/detail?id=675676

### DIFF
--- a/html/rendering/non-replaced-elements/the-fieldset-element-0/legend-float.html
+++ b/html/rendering/non-replaced-elements/the-fieldset-element-0/legend-float.html
@@ -1,0 +1,12 @@
+<!doctype html>
+<title>legend and float</title>
+<link rel=match href=legend-float-ref.html>
+<style>
+  fieldset { margin: 0; padding: 0; border: none; width: 100px; height: 50px; background: red; }
+  legend { width: 100px; height: 50px; background: lime; padding: 0; }
+  .left { float: left; }
+  .right { float: right; }
+</style>
+<p>There should be no red.</p>
+<fieldset><legend class=left></legend></fieldset>
+<fieldset><legend class=right></legend></fieldset>
--- a/html/rendering/non-replaced-elements/the-fieldset-element-0/legend-float.html
+++ b/html/rendering/non-replaced-elements/the-fieldset-element-0/legend-float.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<title>legend and float</title>
+<title>floated legend should not disappear</title>
 <link rel=match href=legend-float-ref.html>
 <style>
   fieldset { margin: 0; padding: 0; border: none; width: 100px; height: 50px; background: red; }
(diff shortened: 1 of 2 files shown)


html/rendering/non-replaced-elements/the-fieldset-element-0/legend-float.html
```diff
@@ -0,0 +1,12 @@
+<!doctype html>
+<title>legend and float</title>
```
mstensho: "floated legend" would be a better title?

html/rendering/non-replaced-elements/the-fieldset-element-0/legend-float.html
```diff
@@ -0,0 +1,12 @@
...
+<link rel=match href=legend-float-ref.html>
+<style>
+  fieldset { margin: 0; padding: 0; border: none; width: 100px; height: 50px; background: red; }
+  legend { width: 100px; height: 50px; background: lime; padding: 0; }
+  .left { float: left; }
+  .right { float: right; }
+</style>
+<p>There should be no red.</p>
+<fieldset><legend class=left></legend></fieldset>
```
mstensho: Not sure how you expect this to fail in an imaginary faulty browser (all pass this, right?), since you're removing the potential effects a rendered legend could have (no border or padding on the fieldset).

zcorpan: The faulty browser would be Chrome M55 per https://bugs.chromium.org/p/chromium/issues/detail?id=675676 ("Legend float in fieldset makes legend disappear").

html/rendering/non-replaced-elements/the-fieldset-element-0/legend-float.html
zcorpan: Maybe, but the title doesn't matter much and I think it's clear enough as-is. :)

html/rendering/non-replaced-elements/the-fieldset-element-0/legend-float.html
mstensho: Oh, I see. I misguessed what you were trying to test. I thought you wanted to test that a floated legend isn't accepted as a rendered legend. But maybe you *should* write the test so that it also renders differently depending on whether it becomes a rendered legend or not.

zcorpan: There's a separate test for that at https://github.com/web-platform-tests/wpt/pull/12567

html/rendering/non-replaced-elements/the-fieldset-element-0/legend-float.html
zcorpan: Actually, changed to "floated legend should not disappear" now.

html/rendering/non-replaced-elements/the-fieldset-element-0/legend-float.html
mstensho: Ok, fine then.

mstensho (approved):

PR #12620, opened 2018-08-22
HTML: test fieldset and transform: translateZ(0)

zcorpan:
Bugs:
https://bugs.chromium.org/p/chromium/issues/detail?id=661020


zcorpan:
Apparently this fails in Chromium on Windows and Linux only. I have macOS so I don't know if it reproduces with this test, but I tried to not change things from the attached test case.

mstensho:
Doesn't fail in Chrome for Linux here. https://bugs.chromium.org/p/chromium/issues/attachmentText?aid=310812 fails.

zcorpan:
How about now?

mstensho:
Yep, fails now.

mstensho (approved):

PR #12634, opened 2018-08-22
HTML: document.open() and history entry creation

TimothyGu:
No browser currently passes the 010.html test, and the particular concept it is
trying to test (salvageability) has been removed from the `document.open()`
spec. The test has been removed.

For whatwg/html#3946.


html/webappapis/dynamic-markup-insertion/opening-the-input-stream/010.html
```diff
@@ -1,22 +0,0 @@
-<!doctype html>
```
domenic: Let's expand on the reasons for removing this. Seems to be something like:

- No browser currently passes this
- It's unclear what, if anything, it's testing; there are no interesting asserts
- The title, regarding "salvagability", does not appear connected to the test contents. And we will add better tests for salvagability ourselves.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/resources/history-frame.html
```diff
@@ -0,0 +1,15 @@
+<script>
+function queueTest() {
+  // The timeout is necessary to avoid the parser still being active when
+  // `document.open()` is called and becoming a no-op.
+  setTimeout(() => {
```
domenic: You should be able to grab opener.t.step_timeout or something like that. That would help catch any exceptions, and allow you to avoid the lint whitelist.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/resources/history-frame.html
```diff
@@ -0,0 +1,15 @@
+<script>
+function queueTest() {
+  // The timeout is necessary to avoid the parser still being active when
+  // `document.open()` is called and becoming a no-op.
```
domenic: Explain more why setTimeout(,0) doesn't work (task queues etc.)

domenic (changes requested):

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/history.window.js
```diff
@@ -0,0 +1,25 @@
+// Historically, document.open() created an entry in the session history so
+// that the original page could be seen by going back. Test that this behavior
+// no longer occurs.
```
domenic: We should add a comment saying why this doesn't use an iframe. It's totally fine if the comment is "This could probably use an iframe, but I just wrote a lot of iframe tests, so for variety let's test window.open(). If the difference matters we can convert to or add an iframe test."

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/resources/history-frame.html
TimothyGu: I'm reluctant to do that, as it will cause the entry settings object to be that of the opener's, which has the potential of opening a can of worms.

domenic: Blehhh. Can you include testharness.js and use the global step_timeout?

PR #12639, opened 2018-08-23
Fetch/XHR/Beacon: test locked/disturbed ReadableStream

annevk:
See https://github.com/whatwg/fetch/pull/801 for context.

### DIFF
--- a/beacon/beacon-readablestream.window.js
+++ b/beacon/beacon-readablestream.window.js
@@ -1,0 +1,23 @@
+function assert_beacon(stream) {
+  assert_throws(new TypeError(), () => navigator.sendBeacon("...", stream));
+}
+
+test(() => {
+  const stream = new ReadableStream();
+  stream.getReader();
+  assert_beacon(stream);
+}, "sendBeacon() with a stream on which getReader() is called");
+
+test(() => {
+  const stream = new ReadableStream();
+  stream.getReader().read();
+  assert_beacon(stream);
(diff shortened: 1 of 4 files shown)


igrigorik (commented):
LGTM from this end for the sendBeacon portion. Although, I'd love to get @yutakahirano sanity check too.

beacon/beacon-readablestream.window.js
```diff
@@ -0,0 +1,23 @@
+function assert_beacon(stream) {
```
yutakahirano: Beacon with streams is never going to work because it's keepalive + ReadableStream which is rejected by another reason.

yutakahirano (changes requested):

annevk:
@yutakahirano good point, fixed.

fetch/api/request/request-init-stream.any.js
```diff
@@ -0,0 +1,51 @@
+// META: global=window,worker
+
+"use strict";
+
+function assert_request(input, init) {
+  assert_throws(new TypeError(), () => new Request(input, init), "new Request()");
+  assert_throws(new TypeError(), () => fetch(input, init), "fetch()");
```
yutakahirano: await missing?

annevk: Oh my, thanks for spotting that.

annevk:
I think I fixed that too.

yutakahirano (approved):

PR #12645, opened 2018-08-23
HTML: more tests for document.open bailout order

TimothyGu:
A continuation of https://github.com/web-platform-tests/wpt/pull/12240, but this treats cases when two bailout criteria compete against each other.

- Chrome passes all tests.
- Safari fails 5 out of 6 tests in bailout-exception-vs-return-origin.sub.window.js as they don't implement the same origin check yet, and because Safari Tech Preview hasn't gotten the [fix](https://bugs.webkit.org/show_bug.cgi?id=187319) for `document.open()` in custom element constructor yet. It passes all four tests in bailout-exception-vs-return-xml.window.js.
- Firefox passes 3 out of 6 tests in bailout-exception-vs-return-origin.sub.window.js due to internal bailout ordering differences and all four tests in bailout-exception-vs-return-xml.window.js.
- Edge only implements the active parser running script bailout so there isn't any meaningful ordering to test.

See https://github.com/whatwg/html/issues/3818.

### DIFF
--- a/html/webappapis/dynamic-markup-insertion/opening-the-input-stream/bailout-exception-vs-return-origin.sub.window.js
+++ b/html/webappapis/dynamic-markup-insertion/opening-the-input-stream/bailout-exception-vs-return-origin.sub.window.js
@@ -1,0 +1,82 @@
+document.domain = "{{host}}";
+
+async_test(t => {
+  const iframe = document.body.appendChild(document.createElement("iframe"));
+  t.add_cleanup(() => { iframe.remove(); });
+  iframe.onload = t.step_func(() => {
+    // Here, the entry settings object is still the iframe's. Delay it in such
+    // a way that makes the entry settings object the top-level page's, but
+    // without delaying too much that the parser becomes inactive. A microtask
+    // is perfect as it's executed in "clean up after running script".
+    Promise.resolve().then(t.step_func_done(() => {
+      assert_throws("InvalidStateError", () => {
+        iframe.contentDocument.open();
+      }, "opening an XML document should throw an InvalidStateError");
+    }));
+  });
+  const frameURL = new URL("resources/bailout-order-xml-with-domain-frame.sub.xhtml", document.URL);
+  frameURL.port = "{{ports[http][1]}}";
+  iframe.src = frameURL.href;
+}, "document.open should throw an InvalidStateError with XML document even if it is cross-origin");
+
+async_test(t => {
+  const iframe = document.body.appendChild(document.createElement("iframe"));
+  t.add_cleanup(() => { iframe.remove(); });
+  window.onCustomElementReady = t.step_func(() => {
+    window.onCustomElementReady = t.unreached_func("onCustomElementReady called again");
+    // Here, the entry settings object is still the iframe's. Delay it in such
+    // a way that makes the entry settings object the top-level page's, but
+    // without delaying too much that the throw-on-dynamic-markup-insertion
+    // counter gets decremented. A microtask is perfect as it's executed in
+    // "clean up after running script".
+    Promise.resolve().then(t.step_func_done(() => {
+      assert_throws("InvalidStateError", () => {
+        iframe.contentDocument.open();
+      }, "opening a document from a custom element constructor should throw an InvalidStateError");
+    }));
+  });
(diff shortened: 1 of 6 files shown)


html/webappapis/dynamic-markup-insertion/opening-the-input-stream/bailout-exception-vs-return-origin.sub.window.js
```diff
@@ -0,0 +1,82 @@
+document.domain = "{{host}}";
+
+async_test(t => {
+  const iframe = document.body.appendChild(document.createElement("iframe"));
+  t.add_cleanup(() => { iframe.remove(); });
+  iframe.onload = t.step_func(() => {
+    // Here, the entry settings object is still the iframe's. Delay it in such
+    // a way that makes the entry settings object the top-level page's, but
+    // without delaying too much that the parser becomes inactive. A microtask
```
domenic: Maybe "...parser becomes inactive (which would trigger a separate bailout from the two that this test pits against each other)" or similar.

Also, extract this comment block into a function with a nice name, like `changeEntrySettingsObjectButWithoutWaitingTooLong(t.step_func_done(...))`, that you can reuse.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/bailout-exception-vs-return-origin.sub.window.js
```diff
@@ -0,0 +1,82 @@
...
+    // is perfect as it's executed in "clean up after running script".
```
domenic: In addition to the factored-out common comment, it's good to explain in each case when "clean up after running script" occurs. Here, it occurs in the `</script>` tag in the subframe.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/bailout-exception-vs-return-origin.sub.window.js
```diff
@@ -0,0 +1,82 @@
...
+  const iframe = document.body.appendChild(document.createElement("iframe"));
+  t.add_cleanup(() => { iframe.remove(); });
+  window.onCustomElementReady = t.step_func(() => {
+    window.onCustomElementReady = t.unreached_func("onCustomElementReady called again");
+    // Here, the entry settings object is still the iframe's. Delay it in such
+    // a way that makes the entry settings object the top-level page's, but
+    // without delaying too much that the throw-on-dynamic-markup-insertion
+    // counter gets decremented. A microtask is perfect as it's executed in
+    // "clean up after running script".
+    Promise.resolve().then(t.step_func_done(() => {
```
domenic: Here, it occurs via "constructing" in "create an element for a token", because the element is created by the parser and thus the stack is empty.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/bailout-exception-vs-return-origin.sub.window.js
```diff
@@ -0,0 +1,82 @@
...
+      assert_throws("InvalidStateError", () => {
+        iframe.contentDocument.open();
+      }, "opening a document from a custom element constructor should throw an InvalidStateError");
```
domenic: "from" a custom element constructor isn't quite right... it's more like "while the incrementing of the counter caused by create an element for a token is still incremented". No need to be super-precise, but best to be a bit less misleading.

domenic (approved):
Some clarity suggestions, but actual test contents look good.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/resources/bailout-order-xml-with-domain-frame.sub.xhtml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml">
+  <head><title>Dummy XHTML document</title></head>
```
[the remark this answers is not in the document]
TimothyGu: Ah yeah this was just copied from dummy.xhtml. I should change this.

TimothyGu:
Comments should all be addressed.

domenic (approved):

PR #12647, opened 2018-08-23
HTML: test dynamically updating the text content of a legend

zcorpan:
Bugs:
https://bugs.chromium.org/p/chromium/issues/detail?id=449299

### DIFF
--- a/html/rendering/non-replaced-elements/the-fieldset-element-0/legend-dynamic-update.html
+++ b/html/rendering/non-replaced-elements/the-fieldset-element-0/legend-dynamic-update.html
@@ -1,0 +1,10 @@
+<!DOCTYPE html>
+<title>legend and dynamic update</title>
+<link rel=fieldset-foo-ref.html>
+<p>There should be a normal fieldset below with the legend "Foo".</p>
+<fieldset>
+  <legend>F</legend>
+</fieldset>
+<script>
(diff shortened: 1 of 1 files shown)


mstensho (approved):

annevk:
@zcorpan should that bug be reopened?

html/rendering/non-replaced-elements/the-fieldset-element-0/legend-dynamic-update.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<title>legend and dynamic update</title>
+<link rel=fieldset-foo-ref.html>
+<p>There should be a normal fieldset below with the legend "Foo".</p>
+<fieldset>
+  <legend>F</legend>
+</fieldset>
+<script>
+  document.querySelector('legend').textContent += "oo";
```
mstensho: Oops, forgot one thing here! Should trigger layout before making this change, to make sure that we need to re-layout afterwards. Add a document.body.offsetTop or something in front of this.

zcorpan: Oops good catch. https://github.com/web-platform-tests/wpt/pull/12665

PR #12650, opened 2018-08-23
HTML: add test for history.state after document.open()

TimothyGu:
Chrome, Firefox, and Safari pass this.

For https://github.com/whatwg/html/pull/3946.

### DIFF
--- a/html/webappapis/dynamic-markup-insertion/opening-the-input-stream/history-state.window.js
+++ b/html/webappapis/dynamic-markup-insertion/opening-the-input-stream/history-state.window.js
@@ -22,4 +22,4 @@ async_test(t => {
     iframe.contentDocument.open("", "replace");
     assert_equals(iframe.contentWindow.history.state, "state");
   });
-}, "history.state is kept by document.open() (with legacy replace parameter set)");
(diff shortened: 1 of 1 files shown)


html/webappapis/dynamic-markup-insertion/opening-the-input-stream/history-state.window.js
```diff
@@ -0,0 +1,25 @@
...
+  t.add_cleanup(() => iframe.remove());
+  iframe.src = "/common/blank.html";
+  iframe.onload = t.step_func_done(() => {
+    assert_equals(iframe.contentWindow.history.state, null);
+    iframe.contentWindow.history.replaceState("state", "");
+    assert_equals(iframe.contentWindow.history.state, "state");
+    iframe.contentDocument.open("", "replace");
+    assert_equals(iframe.contentWindow.history.state, "state");
+  });
+}, "history.state is kept by document.open() (with legacy replace parameter set)");
```
domenic: s/legacy/historical. Legacy sounds like something that might still be in the spec, just treated as old.

domenic (approved):

PR #12654, opened 2018-08-23
Convert webdriver pytest fixtures to use pytest.fixture as a decorator.

burg:
Starting in pytest 3.7.0, calling the decorator pytest.fixture directly
and passing a function object does not work. Since this is a weird way to
set up fixtures, just migrate fixture implementations from fixtures.py
to the top-level conftest.py and use a decorator to register each fixture.

Move non-fixture helpers to tests/support/helpers.py. Move constant values
used as defaults to tests/support/defaults.py and adjust imports and uses.
Some default values are used both by fixtures in conftest.py and by helpers
in helpers.py.


burg:
We can't upgrade the vendored pytest until this issue is resolved in some way.

webdriver/tests/conftest.py
```diff
@@ -1,18 +1,22 @@
...
+import os
+import urlparse
+
 import pytest
+import webdriver
+
+from tests.support import defaults
+from tests.support.http_request import HTTPRequest
+from tests.support.wait import wait
+from tests.support.helpers import cleanup_session
```
whimboo: nit: move up before `http_requests`.

webdriver/tests/conftest.py
```diff
@@ -1,18 +1,22 @@
...
+
+default_host = "http://127.0.0.1"
+default_port = "4444"
```
whimboo: You moved both defaults to `defaults.py`. We should get rid of them here.

webdriver/tests/support/helpers.py
```diff
@@ -0,0 +1,118 @@
+from __future__ import print_function
+import sys
```
whimboo: nit: I think the linter wants an empty line here.

webdriver/tests/support/helpers.py
```diff
@@ -0,0 +1,118 @@
...
+import webdriver
+
+from tests.support import defaults
+
+def ignore_exceptions(f):
+    def inner(*args, **kwargs):
+        try:
+            return f(*args, **kwargs)
+        except webdriver.error.WebDriverException as e:
+            print("Ignored exception %s" % e, file=sys.stderr)
```
whimboo: I don't understand the linter failure:
> ERROR:lint:webdriver/tests/support/helpers.py:13: Print function used (PRINT STATEMENT)

You are already using the print function, or do I miss something?

whimboo (changes requested):
Lovely. Those are the changes which I wanted to see for a long time. But when I asked I was told that the fixtures were on purpose managed that way (AFAIR @jugglinmike introduced them). 

Good to see that this hard-to-follow implementation will not be supported with the next minor release of pytest.

I looked over the patch and only noticed some small issues. Otherwise I assume it's only copy and pasting of code anyway.

webdriver/tests/support/helpers.py
burg: This code is not new.

whimboo: Correct. That's why I don't understand why the lint job is failing on it.

burg: According to the docs (https://web-platform-tests.org/writing-tests/lint-tool.html),

PRINT STATEMENT: A server-side python support file contains a print statement; fix: remove the print statement or replace it with something else that achieves the intended effect (e.g., a logging call).


So, we need to figure out the correct logger to use here. I'm not really familiar enough to know the answer. @gsnedders ?

whimboo: Actually, this call to print was whitelisted for the linter at https://github.com/web-platform-tests/wpt/blob/master/lint.whitelist#L103. Now that you moved it you also want to update its location.

burg: Got it!

whimboo:
@jgraham in the Travis log I can see some Unicode decoding errors for mozlog like:
> File "/home/travis/build/web-platform-tests/wpt/tools/ci/check_stability.py", line 77, in write
>    encoded = msg.encode("utf8", "backslashreplace").decode("utf8")

I don't think that this comes from this PR, but I would like to know if that is known, and not the problem why the stability tests fail. I assume it's the same timeout problem again?

whimboo (approved):

PR #12657, opened 2018-08-23
Validate pull requests in TaskCluster

### LINKED ISSUE #10503: Run PRs on Taskcluster
If https://github.com/w3c/web-platform-tests/pull/9226 works out, it's an obvious next step to also run some subset of the tests for PRs, probably replacing the Travis runs that so often time out: https://github.com/w3c/web-platform-tests/issues/7660

jugglinmike:
Extend the configuration for the TaskCluster service to generate tasks in
response to GitHub pull requests. Each pull request should create one cluster
containing four tasks, all of which operate on the affected tests as identified
by `wpt tests-affected`:

- run the tests in Chrome and publish the results as a gzipped artifact
- run the tests in Firefox and publish the results as a gzipped artifact
- verify the stability of the tests in Chrome
- verify the stability of the tests in Firefox

The former two tasks are intended for future use in comparing test results with
data available on https://wpt.fyi. This will be informative only. The latter
two tasks are intended to verify that the affected tests are stable. This will
influence whether or not the pull request may be merged.

A demonstration of expected functionality is available on Bocoup's fork of WPT:

- `master` branch (this functionality was implemented previously, and this pull
  request is not intended to influence the behavior we see today)
  - [`master` branch of WPT fork](https://github.com/bocoup/wpt/commits/master)
  - [Task cluster](https://tools.taskcluster.net/groups/dSGgOEa3S9y5ebTLBKIVMw) 
- Pull request validation
  - [pull request](https://github.com/bocoup/wpt/pull/4)
  - [Task cluster](https://tools.taskcluster.net/groups/ZhRXqeYgTZS3C4TifTRzFg)

It seems wise to vet this in WPT for a bit before allowing the results to
control whether pull requests may be merged. The configuration proposed here
will permit unstable pull requests, but it stores the status of the check in a
dedicated artifact. After some time, we can use this to compare the behavior
between the current TravisCI-powered checks and this new one. If they align, we
can revert the second commit.

Due to some changes to indentation, using `-w` or the [`?w=1` query string
parameter](?w=1) may make this change set easier to review.

[fixes #10503]


jugglinmike:
I had to iterate on this a bit after changing the upstream repository, but this patch is functioning as expected. I've rebased and pushed up a couple of commits to demonstrate how the CI handles patches with no affected tests and how it handles patches with one or more affected tests (and unstable ones, at that). Those final two commits should not be merged to `master`.

jgraham (changes requested):
Very very excited to see this land. But I think the implementation got a little hard to maintain by trying to force everything to go via a shell script.

gsnedders:
From #10503:

> In TravisCI today, we script unnecessary jobs to exit early. This avoids unnecessary work, but it also makes task reports a little noisy (since they include meaningless results for those no-op jobs).

> A nice-to-have improvement would be to use TaskCluster's "decision tasks" feature to avoid scheduling unnecessary tasks.

At the same time, I'm unconvinced we should block doing this on having that working, because it's not a regression currently.

tools/ci/taskcluster-run.py
```diff
@@ -0,0 +1,101 @@
...
+
+import argparse
+import gzip
+import logging
+import os
+import shutil
+import subprocess
+
+browser_specific_args = {
+    "firefox": ["--install-browser", "--reftest-internal"]
```
jgraham: `--reftest-internal` is the default, I think.

tools/ci/taskcluster-run.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python
```
jgraham: So, I think I'm OK with landing this as-is, but I wonder what the effect would be of moving the logic in this file into `wpt run` directly? If one added `--commit-range` as an argument to that function then the only things that wouldn't directly fit would be getting the arguments right per-browser and gzipping artifacts. Those could perhaps be moved out into the task definitions.

tools/ci/taskcluster-run.py
```diff
@@ -0,0 +1,101 @@
...
+import os
+import shutil
+import subprocess
+
+browser_specific_args = {
+    "firefox": ["--install-browser", "--reftest-internal"]
+}
+
+def tests_affected(commit_range):
+    output = subprocess.check_output([
```
jgraham: You could, of course, just import and use the function directly rather than going via a process.

jgraham (approved):

tools/ci/taskcluster-run.py
```diff
@@ -0,0 +1,101 @@
...
+
+    logger = logging.getLogger("tc-run")
+    logger.setLevel(logging.INFO)
+    handler = logging.StreamHandler()
+    handler.setFormatter(
+        logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+    )
+    logger.addHandler(handler)
+
+    child = subprocess.Popen(['python', './wpt', 'manifest-download'])
```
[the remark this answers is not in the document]
Hexcles: My 2 cents for removing this.

tools/ci/taskcluster-run.py
Hexcles: @jgraham wouldn't that require hacking sys.path? If so, I'm in favour of forking.

Hexcles (approved):
The implementation LGTM. I didn't look into the Travis & TC failures though. Could you take a look?

tools/ci/taskcluster-run.py
jgraham: Yes. I think it makes more long term sense to move all of this into `wpt run` and not have another wrapper script at all. So I don't object to defering that change here.

tools/ci/taskcluster-run.py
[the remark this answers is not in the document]
jugglinmike: Will do

tools/ci/taskcluster-run.py
jugglinmike: I think you're right

    tools/wptrunner/wptrunner/wptcommandline.py:    if kwargs["reftest_internal"] is None:
    tools/wptrunner/wptrunner/wptcommandline.py-        # Default to the internal reftest implementation on Linux and OSX
    tools/wptrunner/wptrunner/wptcommandline.py:        kwargs["reftest_internal"] = sys.platform.startswith("linux") or sys.platform.startswith("darwin")

I wanted to verify with the person who added this flag, and [it turns out that was you](https://github.com/web-platform-tests/wpt/commit/1e51e1722a294830c85866462757d6924de7bac4#diff-57f0e26181ff24ebd57b4bf3976b6260):

> Explicitly set Firefox to use the fast reftest runner.
>
> This should happen by default on Linux, but it doesn't hurt to be explicit.

Have you had a change of heart? Or should we keep the flag for the sake of explicitness?

jugglinmike:
@jgraham Instead of installing Firefox via `wpt run`, we could install it via `wpt install firefox browser` from within `start.sh`. Admittedly, that script is a little more difficult to maintain since it's built into the Docker image.  However, refactoring like this will make the script's behavior more consistent, and ([depending on how you feel about the technically-unnecessary argument](https://github.com/web-platform-tests/wpt/pull/12657#discussion_r213793025)), it could allow us to remove all browser-specific `run` arguments.

jugglinmike:
> Yes. I think it makes more long term sense to move all of this into `wpt run` and not have another wrapper script at all. So I don't object to defering that change here.

This aspect of the wrapper is motivated by my naive attempt to print the affected tests to standard out. Without that, it would be a simple matter of piping `wpt tests-affected` into `wpt run`. Then again, after reviewing the `xargs` docs, it looks like the `--verbose` flag would get us the functionality without the need for all the encoding nonsense:

    $ ./wpt tests-affected --null master... | xargs --verbose --null ./wpt run firefox
    INFO:manifest:Updating manifest
    ./wpt run firefox dom/events/CustomEvent.html 
    Using certutil /usr/bin/certutil
    (etc.)

That serves the same purpose as the logging statement in this patch:

    logger.info("Executing command: %s" % " ".join(command))

Considering that [we may not need browser-specific arguments, either](https://github.com/web-platform-tests/wpt/pull/12657#issuecomment-417066436), we may be able to compose these tasks in the `.taskcluster.yml` file, after all. Look ma, no wrapper. I'd be happy to keep iterating in this direction, but since we're also considering the long-term, it would be helpful for me to understand your design sensibilities.

Do either of you see value in maintaining separation between the various sub-commands? I don't mean to push the UNIX philosophy just for the sake of it, and I know that technically speaking, these are all implemented in the same application. That said, as someone who's been collecting results for a while, I appreciate the ability to compose tasks without patching WPT. I also find it easier to find features that are exposed in their own terms rather than as arguments to the increasingly-large `wpt run` subcommand. This could also allow us to be more strict in the contract we offer to consumers--if we were able to formally document a stable Python API (or remove it altogether), then I think the contribution experience would be even better.

jgraham:
> @jgraham Instead of installing Firefox via wpt run, we could install it via wpt install firefox browser from within start.sh. Admittedly, that script is a little more difficult to maintain since it's built into the Docker image. However, refactoring like this will make the script's behavior more consistent, and (depending on how you feel about the technically-unnecessary argument), it could allow us to remove all browser-specific run arguments.

I was thinking about the opposite; making `wpt install chrome browser` (and hence `wpt run --install-browser chrome`) actually work in the case that you're on Ubuntu/Debian Linux and can get root (and maybe adding more configurations in the future). Ideally I'd like to have less in the `start.sh` script becase updating the docker image is such a pain.

> Do either of you see value in maintaining separation between the various sub-commands? I don't mean to push the UNIX philosophy just for the sake of it, and I know that technically speaking, these are all implemented in the same application. That said, as someone who's been collecting results for a while, I appreciate the ability to compose tasks without patching WPT. I also find it easier to find features that are exposed in their own terms rather than as arguments to the increasingly-large wpt run subcommand. This could also allow us to be more strict in the contract we offer to consumers--if we were able to formally document a stable Python API (or remove it altogether), then I think the contribution experience would be even better.

So, I think there are just a series of tradeoffs here and no clearly obvious right answer. Having a series of independent subcommands is great because, as you say, we can script different things together using common utilities, and we get a cleaner separation between concerns for testing. But on the other hand it has real and serious disadvantages:

* It makes it harder to reproduce results locally. Having a single command that can just be copied and pasted locally is a big advanatage for reproducing results, especially compared to something that requies CI-specific setup (gecko CI has an egregious case of this where the way things run on CI is via a wrapper application that's very difficult to get running on a local machine, and it's awful. Nothing we are doing is that bad, but it's a concern).

* It's not very portable. Bash scripts aren't going to work on Windows unless people are using WSL, so it becomes particuarly hard for Windows users to figure out how to reproduce what they saw in CI.

* We end up with a collection of single-purpose shell scripts that themselves represent significant complexity; although I like the idea of lots of small composable utilities (and indeed that's how the wpt cli is structured in large part), it is often overlooked that the composition itself is a program that has to be maintained.

So I don't have a single answer at the moment, but my inclination such as it is is to keep building composable utilities for flexibility and experimentation, but to make the most common patterns baked in features (particularly of `wpt run` which is largely a wrapper around other functionaility anyway) so that they are portable and easy to run locally.

> The simplest solution would be to perform a full repository fetch for every pull request. I don't know what effect this would have on build times, though maybe we could use TaskCluster's caching feature to optimize that.

So, the current `start.sh` script is supposed to perform a shallow clone and then if we don't have the commit we are looking for extend it to a full clone. But I think that for PRs we probably want to modify the script to pull from the `refs/pulls` namespace directly rather than first pulling in some of master and then extending it with the relevant PR. We can get the number of commits in the PR from the GH API so can figure out how much we need to pull to get the whole PR plus the merge base.

jugglinmike:
> So I don't have a single answer at the moment, but my inclination such as it
> is is to keep building composable utilities for flexibility and
> experimentation, but to make the most common patterns baked in features
> (particularly of wpt run which is largely a wrapper around other
> functionaility anyway) so that they are portable and easy to run locally.

Sounds good to me. Thanks for taking the time to write all of that up!

> We can get the number of commits in the PR from the GH API so can figure out
> how much we need to pull to get the whole PR plus the merge base.

I've pushed up a commit to do this. It necessitated a change to `start.sh`
because we can't use `git clone` to initialize a repository from an arbitrary
reference such as `refs/pull/1/head` (the command only supports branch names
and tag names). I renamed the variable `REV` to `REVISION` in order to avoid
confusion with the new `REF` variable.

That said, the strategy may be incomplete. If we test from the tip of the pull
request branch, then we will not be including any commits that may have landed
in `master` since the pull request was opened. That means the result reported
by TaskCluster might not reflect what we'd see after the patch was merged.

Contributors could get a more accurate picture by rebasing their patch, but I
don't know if many people would expect this. We could perform a merge in
TaskCluster (or maybe use the `merge_commit_sha` from GitHub). Would it be an
issue that we couldn't validate unmergable pull requests?

That idea highlights an underlying problem: we will only run these tasks in
response to a pull request event. Even with a "test merge" strategy, the
results will fall out of date if `master` advances and no new commits are
pushed. So staleness may be unavoidable to some extent.

These are all questions that I'm sure TravisCI et. al have worked out already,
but at the moment, I can't bring to mind any evidence of how they operate. I'll
look into that tomorrow, but I thought I'd leave some rambling here in case
@jgraham (or anyone else) feels like this issue is worth discussing.

jugglinmike:
> I think wpt run will do this be default now, if possible.

`wpt run` does attempt to download a manifest file by default, but `wpt tests-affected` (which precedes `run` in this script) does not. That seems appropriate for a lower-level utility like `wpt tests-affected`, so I've persisted the invocation of `wpt manifest-download`.

Regarding revision selection: a quick review of some TravisCI logs showed that they use a GitHub-provided reference I was not aware of: `refs/pull/{pull request number}/merge`. Since TravisCI has precedent in this project and on GitHub at large, I've updated the TaskCluster configuration to mimic this behavior. It means that the results will tend to reflect how the patch will behave after being merged to `master`, regardless of whether the author has rebased. There are a few other benefits, too. The diffing logic is simplified since we can just compare the first parent. Also, because the git repository now includes a commit present in `master`, we are downloading the generated manifest (rather than creating it from scratch).

I'm removing the "do not merge yet label"; this should be ready for another review. @jgraham and/or @Hexcles: would you mind?

jugglinmike:
@jgraham @Hexcles In `master` today, [TravisCI is configured to verify stability using the command `./wpt check-stability`](https://github.com/web-platform-tests/wpt/blob/04bc9c5667911cce7361ee33244dfc8d7ed61212/tools/ci/ci_stability.sh#L12).  This patch configures TaskCluster to verify stability using the command `wpt run --verify`. I chose this command because `check-stability` performs extra git/GitHub operations and because it includes TravisCI-specific code. I assumed validation heuristics of the commands were equivalent, but while researching [an issue filed against the results-collection project](https://github.com/web-platform-tests/results-collection/issues/599), I learned that they differ.

I walked through an example in that issue report, but to summarize:

- `./wpt check-stability` interleaves the tests
- `./wpt run --verify` executes all iterations of each test separately

Is this difference intentional? The former allowed an unstable test to pass through undetected, but that doesn't necessarily mean the latter is technically superior.

jgraham:
The difference is semi-intentional in that the `--verify` behaviour matches what other suites do on mozilla-central where the approach to test verification is to verify every test one at a time until some time limit is reached. This is an alternative apporach to the problem of how to handle time limits in the case that many tests need to be validated. I guess it also means that if one test is leaking state the others aren't affected (although that's bad in itself; doing this really really well would require running whole directories which has a significant time penalty).

Anyway, my general feeling is that the difference is OK and we should try out this patch as is without trying to change the semantics.

.taskcluster.yml
```diff
@@ -47,31 +106,39 @@ tasks:
...
             metadata:
-              name: wpt-${browser.name}-${browser.channel}-${chunk[0]}-${chunk[1]}
-              description: >-
-                A subset of WPT's "${chunk[0]}" tests (chunk number ${chunk[1]}
-                of ${chunk[2]}), run in the ${browser.channel} release of
-                ${browser.name}.
-              owner: ${event.pusher.email}
+              name: ${operation.name}
+              description: ${operation.description}
+              owner: ${event.pull_request.user.login}@users.noreply.github.com
```
jgraham: I meant to ask what the thinking was behind this change (not necessarily a problem; just wondering).

tools/ci/taskcluster-run.py
jgraham: Let's remove the flag and aim for a situation where we can use the same flags for all browsers.

tools/docker/start.sh
```diff
@@ -12,23 +12,28 @@
...
-# Initially we just fetch 50 commits in order to save several minutes of fetching
-git clone ${REMOTE} --single-branch --branch ${BRANCH} --no-checkout -q --depth=50 web-platform-tests
+mkdir web-platform-tests
 cd web-platform-tests
-if [[ ! `git rev-parse --verify -q ${REV}` ]];
+git init
+git remote add origin ${REMOTE}
+
+# Initially we just fetch 50 commits in order to save several minutes of fetching
+git fetch --quiet --depth=50 origin ${REF}
```
jgraham: Technically we could pass in the base ref or the number of commits in the branch to ensure that the full PR is pulled in the unlikley event there are > 50 commits.

tools/ci/taskcluster-run.py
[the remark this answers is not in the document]
jgraham: So the conclusion was not to remove this because of the tests-affected, but it should probably be accompanied by a manifest update or something?

.taskcluster.yml
jugglinmike: git has incorrectly interpreted this section as a change to existing code. Actually, this is configuration for the new pull request tasks, so the "owner" specified is not a change but an addition (the "owner" for push-triggered tasks is unchanged). The two differ because in the context of a pull request, [the `event` object doesn't have a `pusher` property](https://developer.github.com/v3/activity/events/types/#pullrequestevent). TaskCluster requires a value here, so I cribbed this e-mail address from [an example in the TaskCluster docs](https://docs.taskcluster.net/docs/reference/integrations/taskcluster-github/docs/taskcluster-yml-v1#example).

tools/ci/taskcluster-run.py
jugglinmike: You got it.

tools/ci/taskcluster-run.py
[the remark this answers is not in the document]
jugglinmike: I don't think that's necessary:

- `run_tests_affected` calls `affected_testfiles`: https://github.com/web-platform-tests/wpt/blob/9f63e45e03bb836556466ba43082c4a1c464b14e/tools/wpt/testfiles.py#L344
- `affected_testfiles` calls `load_manifest`: https://github.com/web-platform-tests/wpt/blob/9f63e45e03bb836556466ba43082c4a1c464b14e/tools/wpt/testfiles.py#L212)
- `load_manifest` calls `update.update`: https://github.com/web-platform-tests/wpt/blob/9f63e45e03bb836556466ba43082c4a1c464b14e/tools/wpt/testfiles.py#L183-L202

tools/docker/start.sh
jugglinmike: The initial fetch is a provision for tasks run against pushes to `master`. In those cases, we don't have another base reference or a specific history length to work with. The good news is that this isn't a concern for pull requests [now that this patch uses GitHub's `ref/pull/XXX/merge` reference](https://github.com/web-platform-tests/wpt/pull/12657#issuecomment-417786800).  Thanks to the merge commit, we only need to fetch at a depth of 2 to generate a complete diff, regardless of the number of commits in the pull request.

jugglinmike:
I've included another commit to avoid a runtime exception for pull requests that have zero affected tests. We can validate that prior to merging when we remove the intentional instability.

tools/ci/taskcluster-run.py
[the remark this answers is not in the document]
jgraham: So I think the conclusion is that this should automatically try to download the manifest where that makes sense. But I'll make a different PR for that.

tools/docker/start.sh
jgraham: You're right, but this is a little different to what we do on Travis. The difference isn't very important and doesn't block this PR but does impose some limitations if we ever want to use the contents of the actual PR commits in the future.

For travis we use the GitHub-provided branch base to get the whole list of commits on the branch, and use that to figure out the tests that changed. Here we are falling back on the fact that the merge commit isn't on the master branch and the diff between that and master will produce the same list of tests. Which is fine, as long as we don't actually want the PR commits themselves for any reason.

jugglinmike:
@jgraham I've resolved the conflicts introduced by gh-12679 and triggered both types of jobs:

- `github-push` (via https://github.com/bocoup/wpt/commits/master): https://tools.taskcluster.net/task-group-inspector/#/K3hP_1FMSC-qyjJ0FQTtTg
- `github-pull-request` (via this pull request): https://tools.taskcluster.net/task-group-inspector/#/UF5mqKDYRYWT5q_TtWkaXQ

Could you take another look?

jugglinmike:
gh-12878 was recently merged to `master`. That introduced a conflict in this branch: it modified `tools/ci/ci_taskcluster.sh` which this patch removes in favor of a new Python script, `taskcluster-run.py`. The new script includes the semantics that gh-12878 introduced in the old script, so no change was necessary to resolve the conflict.

jgraham (approved):

.taskcluster.yml
```diff
@@ -2,81 +2,155 @@ version: 1
...
+                wpt-docker-worker
+              else:
+                github-worker
+            metadata:
+              name: ${operation.name}
+              description: ${operation.description}
+              owner: ${event.pull_request.user.login}@users.noreply.github.com
+              source: ${event.repository.url}
+            payload:
+              image: jugglinmike/web-platform-tests:0.21
```
Hexcles: A somewhat unrelated question: how can we create a public/shareable DockerHub account?

Hexcles (approved):
LGTM

.taskcluster.yml
jugglinmike: I don't know, but [I think we all want that](https://github.com/web-platform-tests/wpt/pull/12657#discussion_r212583279)

Issue #2314, opened 2015-11-10
Expand event-handler-attributes-body-window.html


Ms2ger:
The test at `html/webappapis/scripting/events/event-handler-attributes-body-window.html` could be expanded. For example, currently it tests
```js
    document.body['on' + handler] = f;
    assert_equals(window['on' + handler], f);
```
while it could also test the opposite direction:
```js
    window['on' + handler] = f;
    assert_equals(document.body['on' + handler], f);
```

Also, it would be useful to test with:

- a `frameset` as the body
- a `frameset` or `body` element that's not [*in a document*](https://dom.spec.whatwg.org/#in-a-document)
- a `frameset` or `body` element that's not [*the body element*](https://html.spec.whatwg.org/multipage/#the-body-element-2) (for example, with multiple body elements in the DOM).

gsnedders:
It would also be useful to test behaviour with setting to `null`. (Also that both are initially null.)

martiansideofthemoon:
@Ms2ger , @gsnedders : Could I work on this one, since I did the PR for #2258 ?
Do I need to add a separate test for testing in the opposite direction?

> Also, it would be useful to test with:
> 
> a frameset as the body
> a frameset or body element that's not in a document
> a frameset or body element that's not the body element (for example, with multiple body elements in  the DOM).

Didn't quite get you @Ms2ger .

PR #12664, opened 2018-08-24
Add tests for event handler when body is a frameset

xiuqijix:
Supplement test point `a frameset as the body` in #2314.


foolip:
@kaixinjxq it looks like these are copied / based on these tests:
https://github.com/web-platform-tests/wpt/blob/master/html/webappapis/scripting/events/event-handler-attributes-body-window.html
https://github.com/web-platform-tests/wpt/blob/master/html/webappapis/scripting/events/event-handler-attributes-window-body.html

Rather than duplicating, what do you think of generalizing the existing tests to test more combinations?

kaixinjxq:
@foolip, apologies, this got lost. I just submit a new commit to address your comments, PTAL.

foolip (approved):
I've run the tests locally in Chrome and Firefox to verify no obvious regressions.

PR #12670, opened 2018-08-24
HTML: fixup for legend position: relative test

zcorpan:
This test fails in Firefox because it doesn't leave a gap for the
legend, but this test isn't intended to test that.


emilio (approved):
Whops.

PR #12679, opened 2018-08-24
Run stable and beta releases on TaskCluster

jgraham:


jgraham:
@jugglinmike if you happen to already have a staging repo configured to run TaskCluster I would appreciate some testing of this branch. If you don't I can set one up myself.

gsnedders:
Also can we document how to set something up to test TC changes?

tools/wpt/browser.py
```diff
@@ -84,9 +84,27 @@ def platform_string_geckodriver(self):
...
+            "beta": "release",
+            "stable": "release"
+        }
+        version = {
+            "stable": "latest",
+            "beta": "latest-beta",
+            "nightly": "latest",
+        }
+        if channel not in branch:
+            raise ValueError
```
jugglinmike: Same as above regarding error message

jugglinmike (commented):
@jgraham in wpt.fyi, we use the terms "stable" and "experimental" to refer to
various browser releases generically. Internally, we map the latter to Chrome's
"dev" channel and Firefox's "nightly" channel. Taking that approach here would
allow us to expose a much more uniform interface. The new terminology will have
to be documented, but it may be easier to explain what we mean by
"experimental" versus having to interpret input like "Chrome nightly" or
"Firefox dev". What do you think?

> @jugglinmike if you happen to already have a staging repo configured to run
> TaskCluster I would appreciate some testing of this branch. If you don't I
> can set one up myself. 

Sure. I had to apply some of my suggestions to get it to run, but here's a push
to "epochs/daily": https://tools.taskcluster.net/groups/PS2oLaLRRZOl0M94VKHLZQ

jugglinmike:
> Also can we document how to set something up to test TC changes?

The `.taskcluster.yml` file currently includes the following:

    # Contributors interested in configurating TaskCluster to run
    # against their fork of WPT should change the `workerType` to
    # "github-worker".
    workerType: wpt-docker-worker

I'm open to suggestion for improving this. Do you think more of an explanation
is necessary? Maybe instruction to enable the TaskCluster GitHub app? Or maybe
we should move this elsewhere--the top of the file, or a dedicated page in the
`docs` directory?

jgraham:
Regarding using experimental/stable, I don't think that quite works because we have three states; nightly, beta and stable. In fact for Chrome there are four states; canary, dev, beta and stable. Running stable and beta gives us the ability to test version N+1 and version N simultaneously (assuming that beta and stable are both using the same configuration which is more or less true for gecko at least).

I think that latest/beta/stable would be a sufficient ontology to cover our needs (although I'm not exactly sure how Edge and WebKit/Safari fit in there). In terms of actual command line arguments, if we expect people to run this generally I would prefer to be liberal in what we accept rather than forcing people to learn a specific set of nomenclature (e.g. I would happily accept all of experimetal, nightly, dev, preview, latest to mean whatever the most up to date version of the browser is (dev is perhaps the least appealing since Firefox has a developer edition that isn't quite the same thing). And equally I would happily accept stable, release, and other synonyms used for the stable channel.

.taskcluster.yml
```diff
@@ -4,10 +4,19 @@ policy:
...
   then:
-    $if: event.ref == "refs/heads/master"
+    $if: event.ref == "refs/heads/master" || event.ref == "refs/heads/epochs/daily" || event.ref == "refs/heads/epochs/weekly"
     then:
       $flattenDeep:
-        $map: [{name: firefox, channel: nightly}, {name: chrome, channel: dev}]
+        $map:
+          $if: event.ref == 'refs/heads/master'
+          then:
+            {config: [{name: firefox, channel: nightly}, {name: chrome, channel: dev}]}
```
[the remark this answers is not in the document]
jgraham: I didn't do that because it's not exhaustive, but `if`/`else` is. I agree that `$match` is more readable, but I don't know what the right tradeoff is. Maybe I shouldn't care about handling every case since hitting a case that isn't handled here resulting in a runtime error would be fine.

jgraham:
https://tools.taskcluster.net/groups/KkY09Cj_QKO4A2JPHUyn_Q is a run starting from a `epochs/daily' push on my fork.

jgraham:
@foolip created some infrastructure for pushing to the `epochs` branches; see e.g. https://github.com/web-platform-tests/wpt/tree/epochs/daily

.taskcluster.yml
```diff
@@ -4,10 +4,16 @@ policy:
 tasks:
   $if: tasks_for == "github-push"
   then:
-    $if: event.ref == "refs/heads/master"
+    $if: event.ref == "refs/heads/master" || event.ref == "refs/heads/epochs/daily" || event.ref == "refs/heads/epochs/weekly"
```
jugglinmike: With the introduction of `$match`, this condition is no longer necessary.

jgraham:
https://tools.taskcluster.net/groups/NTBwJqX_TNiRKL-FEPegPw is a run of this.

foolip:
Exciting! I guess that wasn't triggered by a branch update, but otherwise is the same as what we want?

jgraham:
It wasn't but I can do that too if you like. I had an earlier run that was triggered by a branch update so I'm pretty sure that works, I was mostly checking my refactoring didn't break the yml file entirely.

jgraham:
https://tools.taskcluster.net/groups/CT4NJclrTG-UKAg8FC0Kvw is a beta run on the daily branch.

foolip:
Very cool! @Hexcles, I presume that we're still a ways away from ingesting results into wpt.fyi? Do we have some scripts that'd allow us to do a one-off comparison between that run and something close for stable?

gsnedders (approved):
I think this LGTM, but I don't exactly know much about TC. :)

gsnedders (approved):
I think this LGTM, but I don't exactly know much about TC. :)

.taskcluster.yml
```diff
@@ -4,71 +4,76 @@ policy:
...
+          - [reftest, 9, 10]
+          - [reftest, 10, 10]
+          - [wdspec, 1, 1]
+        each(chunk):
+          taskId: {$eval: 'as_slugid(browser.name + browser.channel + chunk[0] + str(chunk[1]))'}
+          taskGroupId: {$eval: 'as_slugid("task group")'}
+          created: {$fromNow: ''}
+          deadline: {$fromNow: '24 hours'}
+          provisionerId: aws-provisioner-v1
+          workerType:
```
jugglinmike: Good idea!

tools/wpt/install.py
```diff
@@ -2,20 +2,57 @@
...
+}
+
+channel_by_name = {
+    'stable': 'stable',
+    'release': 'stable',
+    'beta': 'beta',
+    'nightly': latest_channels,
+    'dev': latest_channels,
+    'preview': latest_channels,
+    'experimental': latest_channels,
```
jugglinmike: The terms "experimental", "release" and "preview" aren't used by either browser. If we're going to support both "beta" and "newer than beta," I'm afraid those may confuse folks. Even if the confusion is rare, it seems better to omit them since they are all aliases for more familiar terms.

Alternatively, we could document their meaning in the `--help` text.

jgraham:
@jugglinmike Added some more help text.

tools/wpt/browser.py
```diff
@@ -202,8 +238,14 @@ def get_profile_bundle_url(self, version, channel):
 
         return "%s/archive/%s.zip/testing/profiles/" % (repo, tag)
 
-    def install_prefs(self, binary, dest=None):
-        version, channel = self.get_version_and_channel(binary)
+    def install_prefs(self, binary, dest=None, channel=None):
+        version, channel_ = self.get_version_and_channel(binary)
+        if channel is not None and channel != channel_:
+            # Beta doesn't always seem to have the b in the version string, so allow the
+            # manually supplied value to override the one from the binary
```
Hexcles: Wow, really? This is good to know. It's yet another reason why we need to include channels in `wptreport.json`. Currently the receiver infers the channel from the version string, which according to this comment is apparently unreliable.

tools/wpt/install.py
Hexcles: I suppose "preview" might be used for the future Safari support, perhaps?

tools/wptrunner/wptrunner/wptcommandline.py
```diff
@@ -192,6 +192,12 @@ def create_parser(product_choices=None):
                               help="Path to directory containing extra json files to add to run info")
     config_group.add_argument("--product", action="store", choices=product_choices,
                               default=None, help="Browser against which to run tests")
+    config_group.add_argument("--browser-version", action="store",
+                              default=None, help="Informative string detailing the browser "
+                              "release version. This is included in the run_info data.")
```
Hexcles: In my earlier change, I added this to `wpt/run.py`:

https://github.com/web-platform-tests/wpt/blob/1de1550e95396f979da97620a289c213acaf55e2/tools/wpt/run.py#L439

Would it always override the command line argument?

Hexcles (commented):
Just got one question regarding the `--browser-version` argument.

Hexcles (approved):

jugglinmike (approved):

foolip:
Oh cool, this is merged! epochs/daily currently points to https://github.com/web-platform-tests/wpt/commit/9a5d71b326166e12784bdd9d161772e20f87c1fd and I see that we had beta results:
https://tools.taskcluster.net/groups/dMjbBGL-TPu9z6TuZYzonw

However, I can't find the regular (experimental) results. Did beta run *instead* of experimental?

.taskcluster.yml
```diff
@@ -4,71 +4,76 @@ policy:
...
-                - -c
-                - "~/start.sh ${event.repository.url} ${event.ref[len('refs/heads/'):]} ${event.after} ${browser.name}-${browser.channel} &&
-                  cd ~/web-platform-tests &&
-                  ./tools/ci/ci_taskcluster.sh ${browser.name} ${chunk[0]} ${chunk[1]} ${chunk[2]}"
-    else: []
+    $flattenDeep:
+      $map:
+        $flatten:
+          $match: {
+            event.ref == "refs/heads/master": [{name: firefox, channel: nightly}, {name: chrome, channel: dev}],
```
foolip: The answer to my question is probably here, but I can't interpret it. Pushes to master and pushes to epochs/* will be separate events, but they will be the same sha. Does Taskcluster do something to avoid retesting the same sha? Or is it perhaps just task group triggered by push to master is now hidden in GitHub's UI? @Hexcles, if that is the case, will your integration code be able to find both after the fact, or will it only work if we react to the webhooks as the runs are happening?

jgraham: The answer is actually at https://api.github.com/repos/web-platform-tests/wpt/commits/9a5d71b326166e12784bdd9d161772e20f87c1fd/statuses There are two statuses with context `Taskcluster (push)`, one with the URL https://tools.taskcluster.net/groups/dMjbBGL-TPu9z6TuZYzonw and one with the URL https://tools.taskcluster.net/groups/FFjitCXpQ2C4fD_7O7gPeA GitHub groups by context so it only displays one push. AFAIK there isn't a way to have the TaskCluster tasks get different GitHub context values (possibly we could add that feature?) so we can still get at all the results but can't see them all in the GH UI.

GEEKGIRL36:
Pppppp

On Sep 10, 2018 3:41 AM, "jgraham" <notifications@github.com> wrote:

*@jgraham* commented on this pull request.
------------------------------

In .taskcluster.yml
<https://github.com/web-platform-tests/wpt/pull/12679#discussion_r216275262>
:

> -                public/results:
-                  path: /home/test/artifacts
-                  type: directory
-              command:
-                - /bin/bash
-                - --login
-                - -c
-                - "~/start.sh ${event.repository.url}
${event.ref[len('refs/heads/'):]} ${event.after}
${browser.name}-${browser.channel} &&
-                  cd ~/web-platform-tests &&
-                  ./tools/ci/ci_taskcluster.sh ${browser.name}
${chunk[0]} ${chunk[1]} ${chunk[2]}"
-    else: []
+    $flattenDeep:
+      $map:
+        $flatten:
+          $match: {
+            event.ref == "refs/heads/master": [{name: firefox,
channel: nightly}, {name: chrome, channel: dev}],

The answer is actually at
https://api.github.com/repos/web-platform-tests/wpt/commits/9a5d71b326166e12784bdd9d161772e20f87c1fd/statuses
There are two statuses with context Taskcluster (push), one with the URL
https://tools.taskcluster.net/groups/dMjbBGL-TPu9z6TuZYzonw and one with
the URL https://tools.taskcluster.net/groups/FFjitCXpQ2C4fD_7O7gPeA GitHub
groups by context so it only displays one push. AFAIK there isn't a way to
have the TaskCluster tasks get different GitHub context values (possibly we
could add that feature?) so we can still get at all the results but can't
see them all in the GH UI.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/web-platform-tests/wpt/pull/12679#discussion_r216275262>,
or mute the thread

.taskcluster.yml
foolip: Cool, thanks! The epochs/weekly branch currently points to 9614def367ae8af8247142c34b00f755863d2fbb, and [the statuses API](https://api.github.com/repos/web-platform-tests/wpt/commits/9614def367ae8af8247142c34b00f755863d2fbb/statuses) does include three runs: [stable](https://tools.taskcluster.net/groups/Dh4Iu9GTRNaSjRLNZ3Sr9w), [beta](https://tools.taskcluster.net/groups/bXctOOvuQriKNr5T-oKERg) and [dev](https://tools.taskcluster.net/groups/TZSm9Pa8TlaP-CCdaA_KPQ).

@Hexcles, is that the API you use to find Taskcluster runs?

Hexcles: My code listens to the "status" webhook events, whose payload is similar to that API. There'll be two separate updates in this scenario and my code will run twice. Therefore, all Taskcluster runs will be uploaded as long as the status is "success", regardless of how GitHub UI displays them and whether they are on the same revision.

whimboo:
@jgraham are we creating artifacts especially for the logs? If not yet, what is the follow-up issue to get this integrated?

jugglinmike:
Log files are available for all builds. They can be downloaded via the TaskCluster UI or via the command `./wpt tc-download` (more detail on that is available from the `--help` argument).

whimboo:
@jugglinmike I tried but I cannot find any log for https://tools.taskcluster.net/groups/ccZFTpXgQ1CqBIHft81eXQ which was triggered by https://github.com/web-platform-tests/wpt/pull/12968. Taskcluster doesn't list any artifacts nor logs.

whimboo:
Oh, wait. That is actually a task group! Navigating to a single test job I can see the artifacts now. So all is fine.

Hexcles:
@whimboo https://tools.taskcluster.net/groups/ccZFTpXgQ1CqBIHft81eXQ/tasks/DuJQTji0QomkUiTD3q3CCg/runs/0/logs/public%2Flogs%2Flive.log

You need to select a task in the group first. Then you'll see some logs in the "Run Logs" dropdown.

By the way, this particular example had no "affected tests", so no tests ran.

PR #12680, opened 2018-08-24
Service worker: redirects and URL fragments

annevk:
See https://github.com/whatwg/fetch/pull/696.


annevk:
Note this is WIP, it needs more tests.

annevk:
@davidben it seems that Chrome fails both redirect tests here. Firefox only fails the first. But I suppose that's to be expected as Chrome also fails a bunch of tests in https://github.com/web-platform-tests/wpt/pull/27575 (non-synthetic redirects with fragments for subresources).

annevk:
I also checked navigations, fragments, and redirects and it seems everyone is doing those as expected already.

service-workers/redirects/fragment.https.window.js
```diff
@@ -0,0 +1,34 @@
...
+      });
+      const img = frame.contentDocument.body.appendChild(document.createElement("img"));
+      img.src = `/images/colors.svg${input}`;
+      await new Promise(resolve => img.onload = resolve);
+      ctx.drawImage(img, 0, 0);
+      _assertPixelApprox(canvas, 40, 40, red, green, blue, alpha, undefined, undefined, 4);
+    }, "Forward response fragments: " + subtitle);
+  });
+
+  // HOW? t.add_cleanup(async () => await reg.unregister());
```
annevk, line 33: If someone has ideas for how to address this they'd be much appreciated. I could probably make add_cleanup await a promise that the individual subtests end up resolving by decrementing a counter, but is that the pattern?

davidben:
> @davidben it seems that Chrome fails both redirect tests here. Firefox only fails the first. But I suppose that's to be expected as Chrome also fails a bunch of tests in #27575 (non-synthetic redirects with fragments for subresources).

This is just a random guess (we've got today off, so I'll need to actually poke at it tomorrow), but maybe it's not the redirect per se but something funny with SVGs not picking up the fragments? Our redirect fragment logic isn't anywhere navigation-specific, so it'd be odd for it to only work in navigations.

annevk:
Could be, but it works if redirects are not involved. See also https://github.com/web-platform-tests/wpt/pull/27575.

I'm not aware of any format apart from SVG exposing fragments and we don't (yet) expose them on responses so unfortunately there's no other way to test.

annevk:
@wanderview @davidben could you review this one as well please? Thanks!

service-workers/redirects/resources/fragment-serviceworker.js
```diff
@@ -0,0 +1,21 @@
...
+  ].forEach(({ input, output, type="fetch" }) => {
+    if (e.request.url.endsWith(`/images/colors.svg${input}`)) {
+      const url = `/images/colors.svg${output}`;
+      let response;
+      if (type === "redirect") {
+        response = Response.redirect(url);
+      } else {
+        response = fetch(url);
+      }
+      e.respondWith(response);
```
wanderview, line 18: Maybe return after the respondWith() to avoid continuing the loop iteration?

service-workers/redirects/resources/navigate-fragment-serviceworker.js
```diff
@@ -0,0 +1,21 @@
...
+  ].forEach(({ input, output, type="fetch" }) => {
+    if (e.request.url.endsWith(`${path}${input}`)) {
+      const url = `${path}${output}`;
+      let response;
+      if (type === "redirect") {
+        response = Response.redirect(url);
+      } else {
+        response = fetch(url);
+      }
+      e.respondWith(response);
```
wanderview, line 18: Same here.  Return after respondWith() to terminate loop.

service-workers/redirects/fragment.https.window.js
```diff
@@ -0,0 +1,35 @@
+// META: script=/html/canvas/resources/canvas-tests.js
+// META: script=../service-worker/resources/test-helpers.sub.js
```
wanderview, line 2: Why are these tests in `service-workers/redirects` instead of `service-workers/service-workrer`?  We have other redirect tests in `service-workers/service-worker` so it doesn't make sense for this to be split off.  I personally would prefer to have all the service worker tests together.

service-workers/redirects/fragment.https.window.js
```diff
@@ -0,0 +1,35 @@
...
+  // Not add_cleanup as we want this to run after all tests
+  add_completion_callback(async () => await reg.unregister());
+  await wait_for_state(t, reg.installing, 'activated');
+  const frame = await with_iframe("resources/dummy.html");
+  const canvas = frame.contentDocument.body.appendChild(document.createElement("canvas"));
+  canvas.width = 100;
+  canvas.height = 50;
+  const ctx = canvas.getContext("2d");
+
+  // Changing this list requires corresponding changes in resources/fragment-serviceworker.js
```
wanderview, line 15: Instead of requiring reasoning across multiple files and keeping things in sync you could have a separate json data file.  That would let you see the entire test input, output, and expectations in a single place.  And things wouldn't get out of sync.  You would just need to have the document and service workers load the json data file using fetch(), etc.

service-workers/redirects/fragment.https.window.js
```diff
@@ -0,0 +1,35 @@
...
+  [
+    { input: "?1#green", subtitle: "control" },
+    { input: "?2#green", subtitle: "request overrides" },
```
wanderview, line 18: It still feels wrong that this is not consistent with redirects.  It really feels like it should work the same way as:

https://github.com/web-platform-tests/wpt/blob/8602e9c9a168c25838bc4b808e3d213ce94ccac1/fetch/redirects/subresource-fragments.html#L19

I know we did this partly to avoid the issue with cache_storage, but I think that may be a mistake.  I'm starting to believe we should match redirect behavior here and sites would just need to be careful not to store responses with fragments in cache_storage if they care about multi-fragment to the same resource.

wanderview (changes requested):

service-workers/redirects/fragment.https.window.js
annevk: I might not understand the file structure of the service workers tests. Should I just move the directory to be a subdirectory of that one?

wanderview: That would be better, yes.  We don't really use many subdirectories in there, though.  I think there is only one other case of subdirectories in service-workers.  I guess I would recommend just putting the test in service-workrers/service-worker directly without a subdirectory.

annevk: Sounds good. Once you all reach some conclusions in the new issue I'd be happy to update these tests.

PR #12683, opened 2018-08-25
Update webkit-pseudo-element test with correct spec link

upsuper:
To reflect that this was added into selectors spec rather than compat spec, and update the link correspondingly.

Related to w3c/csswg-drafts#3051.


upsuper:
I didn't remove anyone... It seems GitHub doesn't handle conflict editing very well... anyway...

foolip (approved):

upsuper:
Thanks, anyway.

PR #12689, opened 2018-08-27
Add historical tests for SSML in Speech API

foolip:
The onmark event handler attribute should be removed.

SpeechSynthesis-speak-SSML.html also checks for the non-support of
SSML based on the time it takes to speak an SSML document, making
reasonable assumptions about the time it would take if supported.

Follows https://github.com/w3c/speech-api/pull/38.

### DIFF
--- a/speech-api/SpeechSynthesis-speak-SSML.html
+++ b/speech-api/SpeechSynthesis-speak-SSML.html
@@ -1,0 +1,50 @@
+<!DOCTYPE html>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/resources/testdriver.js"></script>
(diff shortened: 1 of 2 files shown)


saschanaz (approved):

foolip:
@saschanaz I brought in the test from https://github.com/web-platform-tests/wpt/pull/12568 and inverted the pass condition, manually verified it passes in Chrome and Firefox because the utterance takes too long. Can you review the "invert pass condition" commit?

saschanaz (approved):
👍

foolip:
As indicated in https://lists.w3.org/Archives/Public/public-speech-api/2018Aug/0001.html, I'll let this sit until the end of the week before merging.

foolip:
https://github.com/w3c/speech-api/pull/38 wasn't merged, closing this too.

PR #12695, opened 2018-08-27
Make reftests 800x600 everywhere

### LINKED ISSUE #7598: Consistently use one viewport size for reftests
We in various places use 800x600 and 600x600 for the reftest screenshot size. We should stop being inconsistent and use one of them everywhere. I think we had agreement to go back to 800x600 after the WPT change to 600x600 for B2G, @jgraham?

https://lists.w3.org/Archives/Public/public-test-infra/2014JanMar/0014.html was the previous discussion about this in 2014.

### LINKED ISSUE #12578: Selenium executor uses wrong viewport size for reftests
From executorselenium.py:

```
self.protocol.webdriver.set_window_size(600, 600)
```

This is the selenium.webdriver method, which sets the outer size of the browser window, and not the size of the viewport (including any scrollbars).

gsnedders:
Fixes #7598 and fixes #12578.

This makes everything consistent with 800x600, instead of the current mix. Note the default executor mode (the internal reftest runner) in Firefox is broken as a result of this, but that's part of Marionette and not WPT.

### DIFF
--- a/tools/wptrunner/wptrunner/executors/executorselenium.py
+++ b/tools/wptrunner/wptrunner/executors/executorselenium.py
@@ -372,7 +372,7 @@ class SeleniumRefTestExecutor(RefTestExecutor):
             """return [window.outerWidth - window.innerWidth,
                        window.outerHeight - window.innerHeight];"""
         )
-        self.protocol.webdriver.set_window_rect(0, 0, 600 + width_offset, 600 + height_offset)
+        self.protocol.webdriver.set_window_rect(0, 0, 800 + width_offset, 600 + height_offset)
 
         result = self.implementation.run_test(test)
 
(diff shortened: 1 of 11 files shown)


jgraham (approved):

tools/wptrunner/wptrunner/executors/executorselenium.py
```diff
@@ -336,7 +334,11 @@ def is_alive(self):
     def do_test(self, test):
         self.logger.info("Test requires OS-level window focus")
 
-        self.protocol.webdriver.set_window_size(600, 600)
+        width_offset, height_offset = self.protocol.webdriver.execute_script(
+            """return [window.outerWidth - window.innerWidth,
+                       window.outerHeight - window.innerHeight];"""
+        )
+        self.protocol.webdriver.set_window_size(800 + width_offset, 600 + height_offset)
```
gsnedders: This needs changed similarly in executorwebdriver; we should probably just rebase this on top of #14922 so we make everything correct everywhere then change the viewport size.

PR #12712, opened 2018-08-27
 Add reftests for vector effects

satakagi:


svgeesus (changes requested):
Thanks for producing these.
In terms of the metadata, the reviewer links are not used anymore (because we use GitHub for reviewing) so should preferably be removed. More importantly the reference files should not have rel=help links. Please remove them.
In terms of the tested functionality, this looks correct to me.
Oh and [Travis is complaining](https://travis-ci.org/web-platform-tests/wpt/jobs/421301092) about tabs instead of spaces, and trailing whitespace on some lines. I know, it seems trivial, but please fix this so the build system does not produce errors.

PR #12380, opened 2018-08-09
Implement WebDriver executor for wptrunner

gsnedders:
Taken over #10197. Fixes #12374.

- [x] Decide what to do with capabilities and how to pass W3C-style capabilities in (see https://github.com/web-platform-tests/wpt/pull/12380#pullrequestreview-146796077)
- [x] Test Edge stable to see if this works with that
- [x] Figure out how to deal with Safari (stable needs the legacy protocol; TP works fine with W3C)
- [x] Test this (somehow)

---
This change is [Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/12380)

PR #12726, opened 2018-08-28
[test_driver] Add WebDriver-style actions support

NavidZ:
Following web driver spec for pointer action
sequence and key actions this change adds this
API to the test_driver interface.

### DIFF
--- a/resources/testdriver-actions.js
+++ b/resources/testdriver-actions.js
@@ -1,0 +1,391 @@
+(function() {
+  let sourceNameIdx = 0;
+
+  /**
+   * Builder for creating a sequence of actions
+   */
+  function Actions() {
+    this.sourceTypes = new Map([["key", KeySource],
+                                ["pointer", PointerSource],
+                                ["general", GeneralSource]]);
+    this.sources = new Map();
+    this.sourceOrder = [];
+    for (let sourceType of this.sourceTypes.keys()) {
+      this.sources.set(sourceType, new Map());
+    }
+    this.currentSources = new Map();
+    for (let sourceType of this.sourceTypes.keys()) {
+      this.currentSources.set(sourceType, null);
+    }
+    this.createSource("general");
+    this.tickIdx = 0;
+  }
+
+  Actions.prototype = {
+    /**
+     * Generate the action sequence suitable for passing to
+     * test_driver.action_sequence
+     *
+     * @returns {Array} Array of WebDriver-compatible actions sequences
+     */
+    serialize: function() {
+      let actions = [];
+      for (let [sourceType, sourceName] of this.sourceOrder) {
+        let source = this.sources.get(sourceType).get(sourceName);
+        let serialized = source.serialize(this.tickIdx + 1);
+        if (serialized) {
+          serialized.id = sourceName;
+          actions.push(serialized);
+        }
+      }
+      return actions;
+    },
+
+    /**
+     * Generate and send the action sequence
+     *
+     * @returns {Promise} fulfilled after the sequence is executed,
+     *                    rejected if any actions fail.
+     */
+    send: function() {
+      let actions;
+      try {
+        actions = this.serialize();
+      } catch(e) {
+        return Promise.reject(e);
+      }
+      return test_driver.action_sequence(actions);
+    },
+
+    /**
+     * Get the action source with a particular source type and name.
+     * If no name is passed, a new source with the given type is
+     * created.
+     *
+     * @param {String} type - Source type ('general', 'key', or 'pointer')
+     * @param {String?} name - Name of the source
+     * @returns {Source} Source object for that source.
+     */
+    getSource: function(type, name) {
+      if (!this.sources.has(type)) {
+        throw new Error(`${type} is not a valid action type`);
+      }
+      if (name === null || name === undefined) {
+        name = this.currentSources.get(type);
+      }
+      if (name === null || name === undefined) {
+        return this.createSource(type, null);
+      }
+      return this.sources.get(type).get(name);
+    },
+
+    setSource: function(type, name) {
+      if (!this.sources.has(type)) {
+        throw new Error(`${type} is not a valid action type`);
+      }
+      if (!this.sources.get(type).has(name)) {
+        throw new Error(`${name} is not a valid source for ${type}`);
+      }
+      this.currentSources.set(type, name);
+      return this;
+    },
+
+    /**
+     * Add a new key input source with the given name
+     *
+     * @param {String} name - Name of the key source
+     * @param {Bool} set - Set source as the default key source
+     * @returns {Actions}
+     */
+    addKeyboard: function(name, set=true) {
+      this.createSource("key", name, true);
+      if (set) {
+        this.setKeyboard(name);
+      }
+      return this;
+    },
+
+    /**
+     * Set the current default key source
+     *
+     * @param {String} name - Name of the key source
+     * @returns {Actions}
+     */
+    setKeyboard: function(name) {
+      this.setSource("key", name);
+      return this;
+    },
+
+    /**
+     * Add a new pointer input source with the given name
+     *
+     * @param {String} type - Name of the key source
+     * @param {String} pointerType - Type of pointing device
+     * @param {Bool} set - Set source as the default key source
+     * @returns {Actions}
+     */
+    addPointer: function(name, pointerType="mouse", set=true) {
+      this.createSource("pointer", name, true, {pointerType: pointerType});
+      if (set) {
+        this.setPointer(name);
+      }
+      return this;
+    },
+
+    /**
+     * Set the current default pointer source
+     *
+     * @param {String} name - Name of the pointer source
+     * @returns {Actions}
+     */
+    setPointer: function(name) {
+      this.setSource("pointer", name);
+      return this;
+    },
+
+    createSource: function(type, name, parameters={}) {
+      if (!this.sources.has(type)) {
+        throw new Error(`${type} is not a valid action type`);
+      }
+      let sourceNames = new Set();
+      for (let [_, name] of this.sourceOrder) {
+        sourceNames.add(name);
+      }
+      if (!name) {
+        do {
+          name = "" + sourceNameIdx++;
+        } while (sourceNames.has(name))
+      } else {
+        if (sourceNames.has(name)) {
+          throw new Error(`Alreay have a source of type ${type} named ${name}.`);
+        }
+      }
+      this.sources.get(type).set(name, new (this.sourceTypes.get(type))(parameters));
+      this.currentSources.set(type, name);
+      this.sourceOrder.push([type, name]);
+      return this.sources.get(type).get(name);
+    },
+
+    /**
+     * Insert a new actions tick
+     *
+     * @param {Number?} duration - Minimum length of the tick in ms.
+     * @returns {Actions}
+     */
+    addTick: function(duration) {
+      this.tickIdx += 1;
+      if (duration) {
+        this.pause(duration);
+      }
+      return this;
+    },
+
+    /**
+     * Add a pause to the current tick
+     *
+     * @param {Number?} duration - Minimum length of the tick in ms.
+     * @returns {Actions}
+     */
+    pause: function(duration) {
+      this.getSource("general").addPause(this, duration);
+      return this;
+    },
+
+    /**
+     * Create a keyDown event for the current default key source
+     *
+     * @param {String} key - Key to press
+     * @param {String?} sourceName - Named key source to use or null for the default key source
+     * @returns {Actions}
+     */
+    keyDown: function(key, {sourceName=null}={}) {
+      let source = this.getSource("key", sourceName);
+      source.keyDown(this, key);
+      return this;
+    },
+
+    /**
+     * Create a keyDown event for the current default key source
+     *
+     * @param {String} key - Key to release
+     * @param {String?} sourceName - Named key source to use or null for the default key source
+     * @returns {Actions}
+     */
+    keyUp: function(key, {sourceName=null}={}) {
+      let source = this.getSource("key", sourceName);
+      source.keyUp(this, key);
+      return this;
+    },
+
+    /**
+     * Create a pointerDown event for the current default pointer source
+     *
+     * @param {String} button - Button to press
+     * @param {String?} sourceName - Named pointer source to use or null for the default
+     *                               pointer source
+     * @returns {Actions}
+     */
+    pointerDown: function({button=0, sourceName=null}={}) {
+      let source = this.getSource("pointer", sourceName);
+      source.pointerDown(this, button);
+      return this;
+    },
+
+    /**
+     * Create a pointerUp event for the current default pointer source
+     *
+     * @param {String} button - Button to release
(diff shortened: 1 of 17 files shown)


NavidZ:
@gsnedders @jgraham I applied all the comments on @kereliuk review and upload this PL. One thing I couldn't find anywhere in the WebDriver spec is a well defined format for the list of actions. I put something in the comment of the function here but let me know if there is a format defined with samples and what not that I can use that. I prefer to set a format before the API starts being widely used.

jgraham:
I don't fully understand the concern here. The format of actions messages is fully specified in the WebDriver spec, albeit largely by way of the implementation behaviour.

In any case I think support for this should include a JS API for building the actions sequence, not just allowing people to construct the sequence in raw js objects.

NavidZ:
I'm totally with you on that. A builder helper to build that sequence is something we should do after this. But the question I asked was for myself to at least understand the raw format (as we need to override the API in the testdriver-vendor.js file) and also to put a sample input format as a comment for the function action_sequence to make it more clear in a single look.

I see some hints to different attributes here and there in [webdriver spec](https://w3c.github.io/webdriver/#pointer-actions) but it is not consolidated into one format. I tried to gather stuff into the comment of the function. Did I get it right or do we have some sort of nested dictionaries in the format?

NavidZ:
@jgraham @gsnedders Could you take another look at this? I'd like to make sure to document an example input format in the document of the API.

jgraham:
I have a patch locally that actually adds a useful JS API here.

NavidZ:
Do you want me to land this as is and then you send that patch or do you want to push that on top of this PR?

jgraham:
I'll push on top of this PR if that's possible or create a replacement if it isn't.

NavidZ:
You should have access to do that I believe. But either way I'm fine.

tools/wptrunner/wptrunner/testdriver-extra.js
```diff
@@ -76,6 +76,15 @@
             pending_resolve = resolve;
             pending_reject = reject;
         });
+        for (let actionSequence of actions) {
+            if (actionSequence.type == "pointer") {
+                for (let action of actionSequence.actions) {
+                    if (action.type == "pointerMove" && action.origin instanceof Element) {
+                        action.origin = {selector: get_selector(action.origin)};
```
NavidZ: How do we handle when an origin element is out of the screen? I mean for click case we were scrolling into view to container the target. How about here?

resources/testdriver-actions.js
```diff
@@ -0,0 +1,383 @@
...
+    /**
+     * Generate the action sequence suitable for passing to
+     * test_driver.action_sequence
+     *
+     * @returns {Array} Array of WebDriver-compatible actions sequences
+     */
+    serialize: function() {
+      let actions = [];
+      for (let [_, sources] of this.sources) {
+        for (let [sourceName, source] of sources) {
```
NavidZ: Based on what I understood the way we store/serialized these actions is basically a few parallel lists. I was wondering why we chose that. Correct me if I'm wrong. But in one tick we might have key/pointer actions from different names. Do you leave it to the implementation to choose which one to inject first? Because I don't think we can inject anything at literally the same time. We can inject them back to back and in the same frame but not at the same time. Also this comes into play as one of the actions in one tick could be pause and the question is whether we should pause before injecting other actions in that tick or after.

resources/testdriver-actions.js
```diff
@@ -0,0 +1,383 @@
...
+    /**
+     * Insert a new actions tick
+     *
+     * @param {Number?} duration - Minimum length of the tick in ms.
+     * @returns {Actions}
+     */
+    addTick: function(duration) {
+      this.tickIdx += 1;
+      if (duration) {
+        this.addPause(duration);
```
NavidZ: Is this correct? did you mean to call this.pause(duration) as there is no addPause function in Actions.

resources/testdriver-actions.js
```diff
@@ -0,0 +1,383 @@
...
+     *
+     * @returns {Array} Array of WebDriver-compatible actions sequences
+     */
+    serialize: function() {
+      let actions = [];
+      for (let [_, sources] of this.sources) {
+        for (let [sourceName, source] of sources) {
+          let serialized = source.serialize(this.tickIdx + 1);
+          if (serialized) {
+            serialized.id = sourceName;
```
NavidZ: Just to double check here since the current implementation does seem so support this scenario and as long as the driver supports this is should work. The developer can call pointerdown for say a touch source with name "finger 1" in one call to Actions and send it. Then they should be able to create the same source (with the same name and pointertype) in the next call and this time add a pointerup. Correct?

resources/testdriver-actions.js
jgraham: The spec processes events in a specific order; basically in the order the input devices are added (we work top to bottom on the per-device lists of events). Honestly I'm not sure why we didn't just have a single list rather than multiple parallel lists; it seems like a bad design to me (and I"m at least particually responsible), but it is what it is.

resources/testdriver-actions.js
jgraham: Fixed, thanks.

resources/testdriver-actions.js
jgraham: I just implemented this.

NavidZ: I believe what you already have is perfect. Particularly I believe most of the use cases don't care about multiple sources at the same time. So what we have is fine and concise and is capable of representing everything already. As you suggested we can always enhance this later if we really think it is needed.

tools/wptrunner/wptrunner/testdriver-extra.js
jgraham: Actions don't do that. There was a plan to add explicit scrolling to the actions spec, so you could (in terms of this api) do something like

var elem = document.getElementById("element");
Actions()
  .scroll(0, 0, origin=elem)
  .pointerDown(elem)
  .pointerUp()
  .send()
```

But at the moment that doesn't work and you need to call `scrollIntoView` before you send actions.

File: resources/testdriver-actions.js (hunk 2 above)
Comment (NavidZ) [reply resources/testdriver-actions.js]: I feel like this will bite us down the road when drivers and possibly the test-driver-vendor.js (which could override how this command is routed/executed by vendors) treat this ordering differently. But again I feel like this multi source things at the same time is not a common usecase and definitely not worth blocking this PR for.

File: resources/testdriver.js [hunk 5]
```diff
@@ -171,6 +171,30 @@
...
+        },
+
+        /**
+         * Send a sequence of actions
+         *
+         * This function sends a sequence of actions to the top level window
+         * to perform. It is modeled after the behaviour of {@link
+         * https://w3c.github.io/webdriver/#actions|WebDriver Actions Command}
+         *
+         * @param {Array} actions - an array of action objects to perform as
```
NavidZ: This comment needs to be updated as well as the format of the serialization is not this anymore.

tools/wptrunner/wptrunner/testdriver-extra.js
NavidZ: One thing that makes our job a little harder and possibly inconsistent across different implementation is calculating "in-view center point of element" which I believe now we should do in testdriver-vendor.js as well (at least in Chrome side) . I was hoping that this calculation can be done in this layer (which we already have functions for and used them in click) and only x's/y's are passed over.

resources/testdriver-actions.js
jgraham: I mean the deal here is clearly that the semantics are defined in the WebDriver spec, and that does specify an order. I guess we can allow only one action per tick. Although I should have pointed out that pause doesn't work like that; the pause is asynchronous from the point of view of the rest of the actions in the tick. The algorithm is something like

```
let tickEnd;
if (tick.duration) {
   tickEnd = new Promise(resolve => setTimeout(resolve, duration));
} else {
   tickEnd = new Promise(resolve => resolve())
}
for (action of tick.actions) {
  action.dispatch();
}
await tickEnd;
```

tools/wptrunner/wptrunner/testdriver-extra.js
jgraham: *process* doesn't do anything, as such, it builds a data stucture that's passed to *dispatch*.

Not supporting `origin` is probably a bad idea since the 80% use case here is clicking on a specific element. We could move the support into the client, but notice that there isn't a roundtrip and the semantics are equivalent. For example if I arrange a set of actions that click on element a, which causes scrolling, then click on element b, the coordinates of the element change between the command being sent and the actual action executing. But because we resolve the coordinates right before the click, things  work with the WebDriver method but don't with the precompute method. Obviously there are ways to break both cases e.g. animating the elment.

NavidZ: I agree with the ease of use in all the scenarios.
I believe if the page wants to move around the layout during the actions their best bet is to break off the actions into two calls as they need to make sure the action is indeed done and they verity the effect (e.g. scrolling) before injecting the next action. Otherwise they are just writing a racy test and they are not sure what action is done in what state of the page.

Aside from that what is the driver supposed to do if it gets an element that is outside of the view? There is no scrollIntoView at this point in any steps at this point and not a clear error message. I'm just trying to make sure not to leave these steps up to each implementation as it is just another avenue for interop problems.

NavidZ: Just to be clear I'm not saying webdriver spec should change or anything. I just think that if we expose some minimal set in test_driver that is less prune to differences in the implementation of vendors in their drivers it makes it so that wpt is easier to develop. Any builder or wrapper api like calculating the coordinate from a target and what not that to makes the life of the test developer easier could all live here in wpt repo and shared across browsers so we don't run into interop issues on that part.
This can also apply to that one action in on tick. It is still possible to do anything across multiple ticks since ticks have duration.

jgraham: Well the steps aren't up to each implementation; the reason that this is following WebDriver is that we're fuly reusing the WebDriver-specified semantics. I agree that there's a chance of errors in edge cases when people [re]implement the WebDriver logic.

It seems like the request here is effectively to disallow `origin=pointer` in this API and force everyone to call a js method for getting the element coordinates; I would be very *very* reluctant to allow passing an element in with the actions sequence but that element having its coordinates evaluated at a different time to that specified in the WD spec; it would be extremely confusing. So then the question is whether it's more reasonable to make authors write `.pointerMove(...testdriver.elementCoordinates(element), origin="viewport")` or allow them to write `.pointerMove(0, 0, origin=element)` and take increased risk of some implemenation having an error.

Re ticks, I just realised why that can't work (i.e. why webdriver has the design it has). If you want to simulate a pinch action, you require two fingers to move simultaneously. That means multiple `pointerMove` events in a single tick (the implementation will act to make the two movements interleave in the way you would expect).

jgraham: Re: origin, I made the same suggestion; I was just pointing out that there is a serious burden on test authors to having to do all the position compoutations vs letting the implementation do it. Typically we don't prioritise implementation simplicity over authoring experience, so I"m reluctant to do so in this case, but if you are claiming that Chrome can't implement the API if we allow elements to be passed in, I guess we have to put the burden on authors. The alternative is we come up with a testsuite that ensures the 

The way it works in WebDriver, the move duration is not orthogonal to the tick duration; the length of the tick is extended to be the maximum duration of any action in that tick. Otherwise you could have nonsense like a single pointer doing two different overlapping moves. Forcing all guestures to be build out of a series of moves corresponding to different ticks seems like it doesn't work because you have to spin the event loop enough at the end of a tick to ensure that events generated in that tick have a chance to be processed, so you can't guarantee timing.

Generally I feel like this conversation is going through the same process that led to the current WebDriver actions API. Whilst there is some complexity in that API it is motivated by real requirements and I'm not sure we're going to find out that there's a simpler yet compatible model that can do all the same things.

jgraham:
I added some more tests that alternate implementations can use to verify they have the correct semantics. Does this address any of the concerns?

tools/wptrunner/wptrunner/testdriver-extra.js
NavidZ: Sure. Sounds good then. Let's go with your suggestion. @foolip  is it possible for us to run these tests added here on try bots to make sure we did things correctly in testdriver-vendor.js? These are a little different from the py tests for webdriver and are very similar to our layout tests.

NavidZ (commented):
Apparently, I don't have any permission to choose the approve here. But the change looks good to me.

tools/wptrunner/wptrunner/testdriver-extra.js
foolip: You would have to copy over the changes manually and see how they work, we don't have a way of created a Chromium CL from a WPT PR, the only thing we do in that direction is import master continuously, *after* the PR is merged.

foolip:
> Apparently, I don't have any permission to choose the approve here. But the change looks good to me.

I'll add you to the reviewers team. (We don't have any system to make sure everyone in META.yml also is added to that team.)

foolip:
Oh, you are already in that team. But this is your own PR, and you can't approve it yourself. When someone pushes a change to you're PR, you just have to say "that looks OK" as you just did.

jgraham:
Because GitHub is weird, I *can* approve this PR. So if you are happy with my changes I can go ahead and land everything.

jgraham (approved):
Approving the parts of the PR I didn't write.

NavidZ:
It seems that SeleniumProtocol still doesn't have that send actions. @gsnedders I thought #12380 would be adding that. What should we do for that?

foolip:
@NavidZ now that https://github.com/web-platform-tests/wpt/pull/12726 has landed, is some extra change needed here? @gsnedders, ETA for making chrome_webdriver the default?

jgraham:
Isn't supporting this in Chromium blocked on https://bugs.chromium.org/p/chromedriver/issues/detail?id=1897 ?

foolip:
Yes, but that's assigned to @SummerLW who works with @NavidZ.

jgraham:
OK, but I'm confused about the immediate relevance of #12380. Since Chromedriver doesn't support this yet, the choice of WebDriver protocol library doesn't make much difference either, right?

foolip:
I'm assuming that this won't work with the Selenium executor for any browser ever, so that it's better to just change the WebDriver executor. For it to really work in Chrome it's blocked on a ChromeDriver change, and for Edge and Safari I think we're blocked on new releases before ExecutorWebDriver will work.

jgraham:
I'm not really sure why it wouldn't be supported by Selenium (you can just call the relevant command directly if necessary), but at the same time I don't mind moving away from a large external dependency.

I think you're correct that Edge and Safari only work in prerelease versions (c.f. https://twitter.com/jimevansmusic/status/1039988439465320449 ) although I thought we were testing those anyway.

foolip:
OK, so maybe it would work with Selenium on Edge and Safari. In that case I guess let's just leave that bit in the ExecutorSelenium and also add it to ExecutorWebDriver. For Chrome it ought to not work in either case until the chromedriver support is there.

foolip:
Wonderful, thanks @NavidZ and @jgraham!

PR #12729, opened 2018-08-28
Test that important flags are checked when serializing shorthand with "initial" values

Loirooriol:
See https://bugs.webkit.org/show_bug.cgi?id=188984

### DIFF
--- a/css/cssom/shorthand-serialization.html
+++ b/css/cssom/shorthand-serialization.html
@@ -58,6 +58,20 @@
             assert_equals(testElem.style.margin, "20px");
             assert_equals(testElem.style.cssText, "margin: 20px;")
         }, "Shorthand serialization after setting");
+
+        test(function() {
+            const testElem = document.getElementById("test");
+            testElem.style.cssText = "margin: initial;";
+            assert_equals(testElem.style.margin, "initial");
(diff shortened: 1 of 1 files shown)


css/cssom/shorthand-serialization.html
```diff
@@ -58,6 +58,20 @@
             assert_equals(testElem.style.margin, "20px");
             assert_equals(testElem.style.cssText, "margin: 20px;")
         }, "Shorthand serialization after setting");
+
+        test(function() {
+            var testElem = document.getElementById("test");
```
javifernandez: Please, use 'let' or 'const' here.

css/cssom/shorthand-serialization.html
```diff
@@ -58,6 +58,20 @@
...
+            testElem.style.cssText = "margin: initial;";
+            assert_equals(testElem.style.margin, "initial");
+            assert_equals(testElem.style.cssText, "margin: initial;");
+        }, "Shorthand serialization with 'initial' value.");
+
+        test(function() {
+            var testElem = document.getElementById("test");
```
javifernandez: ditto

javifernandez (approved):

Issue #4772, opened 2017-02-09
Media element autoplay cannot be tested because supporting it is optional


foolip:
In https://html.spec.whatwg.org/multipage/embedded-content.html#ready-states the requirements around autoplay are all "may" and there is this note:

> User agents do not need to support autoplay, and it is suggested that user agents honor user preferences on the matter. Authors are urged to use the autoplay attribute rather than using script to force the video to play, so as to allow the user to override the behavior if so desired.

Still, when UAs *do* support autoplay, it is very much possible to test that it has the right timing and that e.g. calling `video.pause()` causes autoplay to never happen.

To test this, we would need to either strengthen the requirements in HTML, or have some kind of testing API to set the user preference.

@jgraham

robwalch:
`video.play()` Promise coverage would be nice :)

foolip:
@robwalch, do you mean tests for the play promise generically, or specifically around autoplay?

@mounirlamouri FYI

foolip:
This could be tested now: https://web-platform-tests.org/writing-tests/file-names.html?highlight=.optional

PR #12732, opened 2018-08-29
Add test for "Allow delaying autoplay until video elements become visible

### LINKED ISSUE #3225: Update tests for "Allow delaying autoplay until video elements become visible"
See https://github.com/whatwg/html/pull/1409#issuecomment-227775512

@foolip: 

> There is one thing that's still needed, and that's to look for event order tests in web-platform-tests and update them, or to write them if they don't exist. Any takers?

xiuqijix:
Fix #3225, @zcorpan, PTAL.


zcorpan:
This behavior is optional per spec. @foolip how do we want to handle this?

foolip:
We don't have a way to mark optional tests. A .optional. naming convention is the first thing that comes to mind, or perhaps in-file metadata.

foolip:
It is possible to structure the test so that it passes if the optional behavior isn't supported too but with different subtest names depending on support status. @jugglinmike has done this in some fairly recent case I think.

zcorpan:
There's in-file metadata for optional features for css/ tests, see
https://web-platform-tests.org/writing-tests/css-metadata.html ("may")

gsnedders:
https://github.com/web-platform-tests/wpt/issues/4772 is an issue about this specifically being untestable because it is optional, rather the more general optional discussion (in #12766).

foolip:
Let's treat this as blocked on https://github.com/web-platform-tests/wpt/issues/12766.

foolip:
@kaixinjxq https://github.com/web-platform-tests/wpt/pull/13586 has been resolved now, can you rename to `.optional.`?

kaixinjxq:
@foolip, this is odd on github, I push the commit locally and receive the change mail, but the change didn't reflect on github. :(

foolip:
@kaixinjxq there's been a pretty serious incident on GitHub today:
https://blog.github.com/2018-10-21-october21-incident-report/

If you just retry it might sort itself out.

kaixinjxq:
@foolip, it works now, please check again.

foolip (approved):
Thanks! Have you check if any browser passes this test?

kaixinjxq:
Unfortunately, Chrome 72 and FireFox 63 on Android are both failed, can't fire `pause` event after `video.hidden = true`.

foolip:
Oh, that's odd. Do they actually pause the video and not fire any events? How about the paused event with a d at the end?

kaixinjxq:
> Do they actually pause the video and not fire any events? 

Actually, when `video.hidden = true`, the video isn't paused, the `playing` event fired again(As the `video.loop` is true).

> How about the paused event with a d at the end?

Sorry, what do you mean by `d`?

gsnedders:
> > How about the paused event with a d at the end?
> 
> Sorry, what do you mean by `d`?


`pause` event v. `paused` event

foolip:
The event type "paused" is indeed what I meant, but I see in https://html.spec.whatwg.org/multipage/media.html#ready-states that actually "pause" is the event that would be fired.

Rather than hiding the video, maybe try scrolling it out of view, which is more like the normal case this autopausing feature is supposed to handle?

foolip:
If the test doesn't pass on any browser, then it seems likely to be wrong. Have you tried on Chrome for Android?

kaixinjxq:
Yes, I tried on Chrome 72 and FireFox 63 for Android. It still fails.
And another findings:
on Chrome for Android: switch tab and run background will fired the pause event
on FireFox for Android: switch tab and run background the video always playing
on Chrome for Windows: switch tab and run background the video always playing

foolip:
@mounirlamouri can you help? What are the conditions under which we'll pause video for being out of view?

html/semantics/embedded-content/media-elements/ready-states/autoplay-hidden.optional.html
```diff
@@ -0,0 +1,31 @@
...
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/common/media.js"></script>
+<div id="log"></div>
+<script>
+
+promise_test(async t => {
+  let video = document.createElement("video");
+  video.src = getVideoURI("/media/movie_5");
+  video.autoplay = true;
```
mounirlamouri:
```suggestion
  video.autoplay = true;
  video.muted = true;
```

mounirlamouri:
I put a suggestion in the code review. In Safari and Chrome, the video needs to be muted in order to be paused when hidden.

kaixinjxq:
Thanks @mounirlamouri, the test passes with `video.muted=true`.
Could you explain why this is necessary? And I don't see the spec requires this.

mounirlamouri:
> Thanks @mounirlamouri, the test passes with `video.muted=true`.
> Could you explain why this is necessary? And I don't see the spec requires this.

This isn't really per-spec. Both Safari and Chrome decided to do this in order to save resources when a video goes out of view and isn't expected to make any sound. Pausing a video that makes sound will have an audible impact so it's unlikely the right call.

kaixinjxq:
@foolip, can we land this PR now?

foolip:
@kaixinjxq how about @mounirlamouri's suggested edits, don't you want to apply those?

kaixinjxq:
Yes, I had addressed @mounirlamouri's suggested in my last two commits.

foolip:
@kaixinjxq I see, I couldn't tell because the suggestion is still open. I guess GitHub isn't clever enough to notice you pushed the same change.

foolip (approved):

kaixinjxq:
Thank you @foolip, I will try GitHub Suggestion next time. :)

foolip:
https://wpt.fyi/results/html/semantics/embedded-content/media-elements/ready-states/autoplay-hidden.optional.html?label=experimental has results from Chrome, Firefox and Safari now, and Safari does pass the test. Neat :)

mounirlamouri:
Chrome should pass the "Allow delaying autoplay until video elements become
visible.". Any idea why it doesn't? Could the test be constructed in a way
that breaks on Chrome?

kaixinjxq:
I test pass on Chrome canary 72.0.3624.2 for Windows and Chrome dev 72.0.3622.0 for Linux.

foolip:
https://wpt.fyi/results/html/semantics/embedded-content/media-elements/ready-states/autoplay-hidden.optional.html?label=experimental does still show Chrome dev 72 for Linux failing. It does pass for me locally, but when passing `--autoplay-policy=no-user-gesture-required` the test fails. @kaixinjxq can you try that and see if you can work out why it fails?

foolip:
Here's where that flag is passed for `./wpt run`:
https://github.com/web-platform-tests/wpt/blob/f13770f6053eb26328adf487c7f80431da5c4509/tools/wpt/run.py#L277-L278

mounirlamouri:
Oh well, indeed, Chrome only do this when the autoplay policy isn't permissive. Hopefully, this will be resolved if and when we have a Web Driver API to override the autoplay policy.

PR #12733, opened 2018-08-29
More properties do not accept number

ewilligers:
Test that various additional properties do not accept <number>
in quirks mode.


ewilligers:
These tests cover the red cells in 
https://docs.google.com/spreadsheets/d/1HRzb_-S28xq7GqYKhHbMP2YQlqWLvOWSS_twLXy-I40/edit#gid=1297517027

The yellow cells are being discussed in the CSS or SVG working groups.

quirks/unitless-length/excluded-properties-002.html
```diff
@@ -0,0 +1,41 @@
...
+    'mask',
+    'object-position',
+    'perspective-origin',
+    'text-shadow',
+    'transform-origin'
+];
+
+for (let property of properties) {
+  test(() => {
+    if (!getComputedStyle(target)[property])
```
zcorpan: I think this should fail the test instead, e.g. using assert_unreached

zcorpan (changes requested):

quirks/unitless-length/excluded-properties-002.html
ewilligers: Now fails the test if the property is not known.

zcorpan (approved):

PR #12743, opened 2018-08-29
Add a test for WebKit bug 189089.

emilio:


frivoal (approved):
I would have written this as a reftest, there's no need to invoke JavaScript here. At the same time, the test is correct, so you can merge as is if you don't want to humor me.

emilio:
> I would have written this as a reftest, there's no need to invoke JavaScript here. At the same time, the test is correct, so you can merge as is if you don't want to humor me.

Every time I write a test I get conflicting advice in this regard, fwiw, we should all agree on some guidelines :P 

In any case, I don't think it's worth spending time on rewriting the test.

frivoal:
> Every time I write a test I get conflicting advice in this regard, fwiw, we should all agree on some guidelines :P

My guideline, which I think is broadly shared within the CSSWG, is don't use JS unless it is necessary, because JS isn't actually a dependency of CSS in general, so tests shouldn't fail just because JS is off, especially given that there are useful implementations of CSS that don't have JS support (or limited JS support), such as the various print formatters.

I guess the conflicting advice comes from the perspective that wpt is primarily a browser interop project, and browsers have JS, so even if JS isn't a dependency of the thing you're testing, it is a part of what web browsers support, and can always be assumed to be there. It's fine to use whenever you feel like it, and some people might even have a preference for js based tests.

> In any case, I don't think it's worth spending time on rewriting the test.

Which is why I approved regardless.

PR #12758, opened 2018-08-30
Fix bugs in the network-error-logging lock handler.

Ms2ger:
It's not clear to me if this is a good way to use the stash, but at least it doesn't throw anymore.

### DIFF
--- a/network-error-logging/support/lock.py
+++ b/network-error-logging/support/lock.py
@@ -1,4 +1,6 @@
-_LOCK_KEY = "network-error-logging:lock"
+import time
+
+_LOCK_KEY = "67966d2e-a847-41d8-b7c3-5f6aee3375ba"
 _TIMEOUT = 5  # seconds
 
 def wait_for_lock(request):


dcreager (approved):
Thanks! I had not realized that a UUID was a hard requirement.

zcorpan (approved):

jgraham (changes requested):
A hardcoded UUID misses the point of the stash which is to prevent different tests conflicting whilst having a shared data source. There's a function `token` in `common/utils.js` that you can use instead,

dcreager:
> prevent different tests conflicting

We're using the stash for a different purpose here.  These tests are affected by which NEL policies are cached in the user agent.  The cache is global, not per browser context, and so we need to guarantee that all the tests in this subdirectory run serially to ensure that the cache doesn't leak across test cases.  We're using a hard-coded stash key to implement a shared lock that ensures that only one test can proceed at a time.

I realize that's not the usual way to use the stash.  If there's another way to implement this kind of lock I'm happy to switch to that.  There's [some text](/web-platform-tests/wpt/tree/master/network-error-logging#test-parallelism) in the README explaining all of this, but we could add a comment here explaining why we're using a hard-coded key — @jgraham, would that work?

foolip (approved):

foolip:
@dcreager, can you add a comment in a separate PR? (Or in Chromium CL.) @jgraham, can you undo you change request for this PR?

dcreager:
>  can you add a comment in a separate PR?

Done: #12760

PR #12767, opened 2018-08-30
Capitalize WebdriverSpecTest => WebDriverSpecTest

foolip:
This was the only class capitalized like this, and the only occurences
of "Webdriver" anywhere in the repo. For consistency with classes
WebDriverException, WebDriverProtocol and WebDriverServer.

### DIFF
--- a/tools/manifest/item.py
+++ b/tools/manifest/item.py
@@ -234,7 +234,7 @@ class Stub(URLManifestItem):
     item_type = "stub"
 
 
-class WebdriverSpecTest(URLManifestItem):
+class WebDriverSpecTest(URLManifestItem):
     item_type = "wdspec"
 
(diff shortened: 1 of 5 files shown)


gsnedders (approved):
LGTM, but should check with others in case this might break any downstream code (@Hexcles? @jgraham?)

Ms2ger:
https://searchfox.org/mozilla-central/search?q=WebdriverSpecTest&path= shows no occurrences in m-c outside wpt proper.

jgraham (approved):

gsnedders:
https://cs.chromium.org/search/?q=WebDriverSpecTest&sq=package:chromium&type=cs shows nothing in Chromium either.

Hexcles:
Confirmed this shouldn't affect Chromium. And this change will be manually rolled into Chromium together with other wpt tooling changes next time.

PR #12770, opened 2018-08-30
document.open(): Test non-fully active documents

TimothyGu:
Also fix a missing file in the related `url.js` test.

For https://github.com/whatwg/html/pull/3977.

----

All six tests pass in Chrome and Safari.

The "active but not fully active" and "non-active with associated Window; navigated away" cases fail in Edge with a "Permission denied" error, but all others pass.

All but the first one fail in Firefox, which implements the active document bailout

### DIFF
--- a/html/webappapis/dynamic-markup-insertion/opening-the-input-stream/active.window.js
+++ b/html/webappapis/dynamic-markup-insertion/opening-the-input-stream/active.window.js
@@ -3,19 +3,21 @@ function assertOpenIsEffective(doc, initialNodeCount) {
 
   // Test direct document.open() call.
   assert_equals(doc.open(), doc);
-  assert_equals(doc.childNodes.length, 0);
+  assert_equals(doc.childNodes.length, 0, "after open: no nodes in document");
   doc.write("<!DOCTYPE html>");
-  assert_equals(doc.childNodes.length, 1);
+  assert_equals(doc.childNodes.length, 1, "after write: doctype node in document");
   doc.close();
-  assert_equals(doc.childNodes.length, 2);
+  assert_equals(doc.childNodes.length, 2, "after parser close: doctype node and an html element in document");
 
-  // Test implicit document.open() call through write().
+  // Test implicit document.open() call through write(). Since we called
+  // doc.close() above, which sets the insertion point of the parser to
+  // undefined, document.write() will run the document open steps.
   doc.write();
-  assert_equals(doc.childNodes.length, 0);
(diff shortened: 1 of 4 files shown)


html/webappapis/dynamic-markup-insertion/opening-the-input-stream/active.window.js
```diff
@@ -0,0 +1,90 @@
...
+  const doc = frame.contentDocument.implementation.createHTMLDocument();
+  assertOpenIsEffective(doc, 2);
+}, "document.open() does not change document's URL (non-active document without an associated Window object; createHTMLDocument)");
+
+test(t => {
+  const frame = document.body.appendChild(document.createElement("iframe"));
+  t.add_cleanup(() => frame.remove());
+  const doc = new frame.contentWindow.DOMParser().parseFromString("", "text/html");
+  assertOpenIsEffective(doc, 1);
+}, "document.open() does not change document's URL (non-active document without an associated Window object; DOMParser)");
```
domenic: I think Boris would appreciate a test of an XHR responseDocument too.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/active.window.js
```diff
@@ -0,0 +1,90 @@
...
+    });
+
+    frame.src = "/common/blank.html";
+  });
+}, "document.open() does not change document's URL (non-active document with an associated Window object; navigated away)");
+
+test(t => {
+  const frame = document.body.appendChild(document.createElement("iframe"));
+  t.add_cleanup(() => frame.remove());
+  const doc = frame.contentDocument.implementation.createHTMLDocument();
```
domenic: Any reason to use a frame for these two tests? If so we should comment on it; if not, probably best to avoid it.

domenic (commented):
Some clarity improvements and one request for another test.

domenic:
Also need (as a separate PR if you like) tests that event listeners don't get erased in these cases. And ideally tests that aborting doesn't happen, although I know aborting is still WIP.

html/webappapis/dynamic-markup-insertion/opening-the-input-stream/active.window.js
TimothyGu: Hmm, I don't remember a reason. Removed.

TimothyGu:
@domenic Added some tests for event listeners as well. PTAL.

domenic (approved):
LGTM. You should be able to remove the initUIEvent("focus"), and doing so is probably a good idea.

PR #12783, opened 2018-08-31
[wptrunner] Parameterize Sauce Connect timeout

jugglinmike:
Parameterize the maximum delay so that consumers may select an
appropriate value according to their connection and the current
performance of the Sauce Labs service.


gsnedders (approved):

jugglinmike:
Thanks!

PR #12786, opened 2018-08-31
webrtc: fix no-media-call

fippo:
the legacy offer options need to be passed to createOffer, not the then.
regression from #12715, caused chrome to fail on import

@youennf can you please take a look? Whether this test makes sense is another question...
@henbos: you can close https://bugs.chromium.org/p/chromium/issues/detail?id=879021 once this is resolved


alvestrand (approved):
This fix looks right.

PR #12810, opened 2018-09-03
HTML: test fieldset content before legend

zcorpan:


zcorpan:
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/18881637/
https://bugs.webkit.org/show_bug.cgi?id=189244
https://bugs.chromium.org/p/chromium/issues/detail?id=880062

mstensho (approved):

PR #12811, opened 2018-09-03
HTML: test how <legend align> does not map to text-align

zcorpan:

### DIFF
--- a/html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/legend-align-text-align.html
+++ b/html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/legend-align-text-align.html
@@ -1,0 +1,26 @@
+<!doctype html>
+<title>legend align to text-align</title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<style>
+ legend { width: 13em }
+</style>
+<fieldset><legend>foo bar abcdefghijklmnopqrstuvwxyz</legend></fieldset>
+<fieldset><legend align=left>foo bar abcdefghijklmnopqrstuvwxyz</legend></fieldset>
+<fieldset><legend align=center>foo bar abcdefghijklmnopqrstuvwxyz</legend></fieldset>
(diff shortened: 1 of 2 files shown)


zcorpan:
https://bugzilla.mozilla.org/show_bug.cgi?id=1488228
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/18881842/

emilio (approved):

emilio:
Well, let's wait for an answer for @MatsPalmgren's question about why should we spec WebKit's behavior instead of Gecko / Edge's.

html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/legend-align-text-align.html
```diff
@@ -0,0 +1,26 @@
...
+<fieldset><legend align=justify>foo bar abcdefghijklmnopqrstuvwxyz</legend></fieldset>
+<script>
+function test_align(selectorTest, expectedAlign) {
+  const testElm = document.querySelector(selectorTest);
+  test(() => {
+    assert_equals(getComputedStyle(testElm).textAlign, expectedAlign);
+  }, selectorTest);
+}
+
+test_align('legend', 'start');
```
emilio: Just `for (const legend of document.querySelectorAll('legend'))` instead of testing per-selector, maybe? Anyway not a big deal.

html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/legend-align-text-align.html
```diff
@@ -0,0 +1,26 @@
+<!doctype html>
+<title>legend align to text-align</title>
```
emilio: It'd be good to have a more descriptive title, and a link to the spec PR / discussions, maybe.

emilio (approved):

html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/legend-align-text-align.html
zcorpan: Yeah, though the selector is also used for the test name, so I think I'll leave it as-is.

PR #12817, opened 2018-09-04
[cookies] Remove cookies when tests complete

jugglinmike:
Avoid test interactions by removing cookies at the completion of each
test. Use the `add_cleanup` feature of testharness.js to ensure that
this occurs regardless of the passing/failing status of each test.
Persist the existing "setup" logic which defensively removes cookies as
a precaution against still other tests which do not properly restore
global state.

This resolves a known instability caused by
`cookies/prefix/__secure.document-cookie.https.html`. That test sets a
cookie with a "Domain" attribute which
`cookies/prefix/__secure.header.https.html` is not written to remove.


jdm (approved):

mikewest:
Looks totally reasonable to me, thank you for addressing it!

jugglinmike:
Happy to help :)

PR #12818, opened 2018-09-04
Remove duplicated copies of 60x60-red.png file

mrego:
Update paths to use the only copy that is kept under "/css/support/".


mrego:
Now that `/css/support/` is served (see #12371) we can remove duplicated files and keep just one copy there.

frivoal:
Can we use relative paths instead of absolute ones? Currently, most ref tests can be run directly from `file://` without starting up a server, and for things that don't actually need the server, this is very convenient, so I'd like to keep it that way if possible.

I know how to start the server, but I dislike the idea that this knowledge is a prerequisite for doing anything with the test suite.

dirkschulze (approved):

foolip:
@mrego do you want to try rebasing this PR to get it merged? It's approved :)

mrego:
Somehow I missed the approval and totally forgot about this, thanks @foolip for the ping. :)

Just rebased it, let's wait for the checks and if everything is fine I'll merge it.

PR #12824, opened 2018-09-04
Install nightly geckodriver when testing Firefox nightly.

jgraham:
This (ab)uses mozdownload to grab the common.tests.tar.gz package from the
latest nightly build and extracts geckodriver from that. The main disadvantage
is that it's rather slow since the package itself is rather large and we want
only a small part of it. The solution to this would be to package up
geckodriver by itself on taskcluster, but that will require changes in the
gecko buildsystem.

### DIFF
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -333,19 +333,23 @@ class Firefox(Browser):
             package_path = s.download()
         except mozdownload.errors.NotFoundError:
             return
-        exe_suffix = ".exe" if uname[0] == "Windows" else ""
-        with tarfile.open(package_path, "r") as f:
-            try:
-                member = f.getmember("bin%sgeckodriver%s" % (os.path.sep,
-                                                             exe_suffix))
-            except KeyError:
-                return
-            # Remove bin/ from the path.
-            member.name = os.path.basename(member.name)
-            f.extractall(members=[member], path=dest)
-            path = os.path.join(dest, member.name)
-        logger.info("Extracted geckodriver to %s" % path)
-        os.unlink(package_path)
+
+        try:
+            exe_suffix = ".exe" if uname[0] == "Windows" else ""
+            with tarfile.open(package_path, "r") as f:
+                try:
+                    member = f.getmember("bin%sgeckodriver%s" % (os.path.sep,
+                                                                 exe_suffix))
+                except KeyError:
+                    return
+                # Remove bin/ from the path.
+                member.name = os.path.basename(member.name)
+                f.extractall(members=[member], path=dest)
+                path = os.path.join(dest, member.name)
+            logger.info("Extracted geckodriver to %s" % path)
+        finally:
+            os.unlink(package_path)
+
         return path
 
     def version(self, binary=None, channel=None):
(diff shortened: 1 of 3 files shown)


tools/wpt/browser.py
```diff
@@ -315,6 +322,33 @@ def install_webdriver(self, dest=None):
             untar(get(url).raw, dest=dest)
         return find_executable(os.path.join(dest, "geckodriver"))
 
+    def install_geckodriver_nightly(self, dest):
+        import tarfile
+        import mozdownload
```
whimboo: If you are using mozdownload here, where is the update for a requirements.txt, or setup.py file?

jgraham: It's already used to download Firefox.

whimboo: Oh, cool! Didn't know that. Good that there is another customer of the project I own, and that our changes are still useful.

With all that the changes look fine to me. But would be good to get a final review from a person who has more knowledge on wpt than me.

tools/wpt/browser.py
```diff
@@ -394,7 +428,7 @@ def find_binary(self, venv_path=None, channel=None):
     def find_webdriver(self):
         return find_executable("chromedriver")
 
-    def install_webdriver(self, dest=None):
+    def install_webdriver(self, dest=None, channel=None):
```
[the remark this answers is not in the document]
jgraham: I think that's a little pedantic (but I see your point) :)

I agree that the WebDriver releases aren't always coupled with the browser releases (although I think that for Dafari and Edge they actually are). So yeah, there's a little bit of ambiguity in what "channel" means in this case. I can maybe just fix the documentation.

jugglinmike:
One last thing--a discrepancy in the `--destination` argument:

    $ ./wpt install --destination foo firefox webdriver
    Binary installed as foo/bin/geckodriver
    $ ./wpt install --destination foo --channel stable firefox webdriver
    Binary installed as foo/geckodriver

I don't know the use case for that argument, so this might be acceptable. If it's intended for human use, then variation should be fine as long as standard output is accurate. This could cause issues in automation, though.

That said, when `--destination` is *not* specified, both versions of the binary end up at `_venv/bin/geckodriver`.

jgraham:
> One last thing--a discrepancy in the --destination argument:

Fixed, with the help of stackoverflow.

jugglinmike:
Great! I just now realized that due to the early `return` we don't always clean
up the downloaded artifact. If this were a hard failure, we might want to leave
it around for forensics. But because we fall back to the stable release, we
should probably remove the file.

I spoke with @jgraham about this on IRC, and he agreedw with the change. In the
interest of transparency, I've pushed it up as a distinct commit, and I'll
merge it to the `geckodriver_nightly` feature branch in that form. We can
rebase and squash however we like prior to merging that branch to WPT's
`master`.

whimboo:
Btw. when will this be live on wpt.fyi and we can results via geckodriver nightly?

jugglinmike:
As of today, the Firefox results on wpt.fyi are collected by [web-platform-tests/results-collection](https://github.com/web-platform-tests/results-collection).

[That project uses a separate procedure to install Geckodriver because reproducibility is important there.](https://github.com/web-platform-tests/results-collection/issues/586) If we want to use nightly builds of Geckodriver in that project, we'll either have to make this command reproducible or update that other procedure.

PR #12826, opened 2018-09-04
[SVG] Scripts are valid content for SVG+HTML

ewilligers:
This is for SVG intro chapter tests #488

### DIFF
--- a/svg/interact/script-content.svg
+++ b/svg/interact/script-content.svg
@@ -1,0 +1,239 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg"
+     xmlns:h="http://www.w3.org/1999/xhtml">
+  <h:link rel="help" href="https://svgwg.org/svg2-draft/interact.html#ScriptElement"/>
+  <h:meta name="assert" content="The script element is in the content model of all elements."/>
+  <h:script src="/resources/testharness.js"/>
+  <h:script src="/resources/testharnessreport.js"/>
+
+<script>function s0() { return s1(); }</script>
+
+<a>
+  <script>function s1() { return s2(); }</script>
+</a>
+
+<g>
+  <script>function s2() { return s3(); }</script>
+</g>
+
+<animate>
+  <script>function s3() { return s4(); }</script>
+</animate>
+
+<animateMotion>
+  <script>function s4() { return s5(); }</script>
+  <mPath>
+    <script>function s5() { return s6(); }</script>
+  </mPath>
+</animateMotion>
+
+<animateTransform>
+  <script>function s6() { return s7(); }</script>
+</animateTransform>
+
+<circle>
+  <script>function s7() { return s8(); }</script>
+  <clipPath>
+    <script>function s8() { return s9(); }</script>
+  </clipPath>
+</circle>
+
+<defs>
+  <script>function s9() { return s10(); }</script>
+</defs>
+
+<desc>
+  <script>function s10() { return s11(); }</script>
+</desc>
+
+<!-- Becomes https://svgwg.org/svg2-draft/struct.html#UnknownElement
+     if the implementation does not support <discard> -->
+<discard>
+  <script>function s11() { return s12(); }</script>
+</discard>
+
+<ellipse>
+  <script>function s12() { return s13(); }</script>
+</ellipse>
+
+<filter>
+  <script>function s13() { return s14(); }</script>
+  <feBlend>
+    <script>function s14() { return s15(); }</script>
+  </feBlend>
+  <feColorMatrix>
+    <script>function s15() { return s16(); }</script>
+  </feColorMatrix>
+  <feComponentTransfer>
+    <script>function s16() { return s17(); }</script>
+  </feComponentTransfer>
+  <feComposite>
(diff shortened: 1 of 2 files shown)


svg/intro/script-content.html
```diff
@@ -0,0 +1,56 @@
...
+  </span>
+</div>
+
+<script>
+test(function(){
+  assert_equals(s0(), 'same');
+  assert_equals(total, 123);
+
+  for (let element of [spot, box]) {
+    element.style.opacity = '0.75';
```
AmeliaBR: I'm a little worried that this bit of code could get spurious fails in non-browser environments, from

- expecting elements with `id` values to be available in the global scope (not really sure if this is defined anywhere that SVG normatively references)
- expecting the `style` property to be accessible as a dictionary. This _is_ normative in the versions of CSSOM that SVG 2 references, but it wasn't in the specs referenced by SVG 1.1. So, for example, it would fail in Batik last I checked. And that's really a separate test/failure from the main purpose of this test (scripts are working and the entire HTML+SVG document shares a global scope).

svg/intro/script-content.html
```diff
@@ -0,0 +1,56 @@
...
+<svg>
+  <script>function s1() { total += 20; return s2(); }</script>
+  <g>
+    <script>function s2() { return s3(); }</script>
+    <circle id="spot">
+      <script>function s3() { return s4(); }</script>
+    </circle>
+  </g>
+  <desc>
+    <script>function s4() { return s5(); }</script>
```
AmeliaBR: Fun side note: if the document is parsed by an HTML parser, this is going to be an HTMLScriptElement, not an SVGScriptElement. But regardless, the script content should be processed the same!

svg/intro/script-content.html
ewilligers: I can use `getElementById`.

With regards to property, the purpose of the code is to check that the same scripts can operate on SVG and CSS elements. I can use `lastChild` / `removeChild()` / `firstChild` / `insertBefore()` to achieve this.

svg/intro/script-content.html
```diff
@@ -0,0 +1,56 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Test: </title>
+<link rel="help" href="https://svgwg.org/svg2-draft/intro.html#AboutSVG">
```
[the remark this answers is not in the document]
ewilligers: OK, I will also add all the other SVG elements to check that script is valid inside all of them. I won't add dozens of help links.

svg/intro/script-content.html
AmeliaBR: Yeah, something simple that is defined in core DOM. Could just be set & get an attribute.

svg/intro/script-content.html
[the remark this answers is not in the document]
AmeliaBR: Yeah, probably a help link to the script definition is enough.

I didn't go as far as recommending including _every_ SVG element, because I think you've got a pretty good cross-section of the different types of elements (shapes, effects, groups, etc.). But then again: you never know which one might have been overlooked in an implementation.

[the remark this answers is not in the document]
svgeesus: Agree with @AmeliaBR that the rel=help should point to https://svgwg.org/svg2-draft/interact.html#ScriptElement

svg/intro/script-content.html
```diff
@@ -0,0 +1,56 @@
...
+<meta name="assert" content="The same scripts can work on both HTML and SVG elements.">
+<meta name="assert" content="The script element is in the content model of all elements.">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<script>var total = 0;</script>
+<script>function s0() { total += 100; return s1(); }</script>
+<svg>
+  <script>function s1() { total += 20; return s2(); }</script>
```
svgeesus: GitHub syntax highlights script content in HTML and not in SVG, lol

svg/intro/script-content.html
svgeesus: Since the root element is html and given the purpose of the test, not working in non-browser implementations is fine here.

A similar test with a chain of script elements as children of assorted svg-namespace elements, and with an svg root, would also be a useful test. (Not a blocker on this one, obviously).

svgeesus (dismissed):
Apart from the help link, this looks great

ewilligers:
I have made the requested changes.

svg/interact/script-content.svg
```diff
@@ -0,0 +1,237 @@
...
+
+<defs>
+  <script>function s9() { return s10(); }</script>
+</defs>
+
+<desc>
+  <script>function s10() { return s11(); }</script>
+</desc>
+
+<discard>
```
AmeliaBR: The `<discard>` element is new in SVG Animations Level 2; I don't think it's supported anywhere.

But I don't think that should affect the function of a script element that it wraps.  Inside a .svg file, unknown elements won't change the parsing context.

AmeliaBR (approved):
Looks Good To Me.

Thanks for the fixes & extra test. See one comment about the `<discard>` element, which maybe shouldn't be in the SVG 2 element index anymore...

ewilligers:
@svgeesus PTAL

PR #12835, opened 2018-09-04
[cookies] Correct utility function and tests

jugglinmike:
The `cookie-helper.sub.js` utility script includes
`set_prefixed_cookie_via_http_test`, a function that defines sub-tests
using the `promise_test` API. Previously, it included the following
code:

    promise_test(t => {
      var postDelete = _ => {
        // (elided)
      };

      if (!options.origin) {
        return postDelete;
      } else {
        // (elided)
      }
    });

The `promise_test` function does not recognize return values which are
functions, so returning the `postDelete` method had no effect, and as a
result, the generated tests performed zero assertions. Because none of
the consumers of `set_prefixed_cookie_via_http_test` specified a value
for the `origin` option, all invocations were effected by this bug.

Correcting the problem surfaced a number of errors in the tests. In the
interest of atomicity, this patch attempts to address them all:

- The logic intended to defensively remove cookies prior to testing was
  implemented using `document.cookie`. Because some tests create cookies
  which include the `HttpOnly` attribute, the DOM API cannot remove
  cookies in all cases. This patch refactors the solution to remove
  such cookies via an HTTP request. It also assumes the environment is
  initially clean and instead expresses the concern via an asynchronous
  "cleanup" function. (This change necessitated an extension to the
  `set.py` script so that it could be used to expire cookies.)
- The test name `__secure.header.html` incorrectly asserted that a
  cookie set with the `Secure` attribute could be observed in a
  non-secure context. This patch corrects the expectation.
- The test named `__secure.header.https.html` incorrectly asserted that
  a cookie set with a foreign `Origin` attribute could be observed from
  the current origin. This patch corrects the expectation.

---

This patch does not affect the test results in Firefox or Chrome. The commands:

    $ xvfb-run --auto-servernum ./wpt run firefox cookies/prefix
    $ xvfb-run --auto-servernum ./wpt run chrome cookies/prefix

...both continue to report the following:

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 95 checks (9 tests, 86 subtests)
    Expected results: 95

### DIFF
--- a/cookies/prefix/__host.header.html
+++ b/cookies/prefix/__host.header.html
@@ -8,6 +8,7 @@
     set_prefixed_cookie_via_http_test({
       prefix: "__Host-",
       params: "Path=/;" + extraParams,
+      origin: self.origin,
       shouldExistInDOM: false,
       shouldExistViaHTTP: false,
       title: "__Host: Non-secure origin: Does not set 'Path=/;" + extraParams + "'"
@@ -17,6 +18,7 @@
     set_prefixed_cookie_via_http_test({
       prefix: "__Host-",
       params: "Secure; Path=/;" + extraParams,
+      origin: self.origin,
       shouldExistInDOM: false,
       shouldExistViaHTTP: false,
       title: "__Host: Non-secure origin: Does not set 'Secure; Path=/;" + extraParams + "'"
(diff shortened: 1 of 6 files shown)


mikewest (approved):
Thank you, this patch looks correct to me. I very much appreciate you digging in.

LGTM.

PR #12844, opened 2018-09-05
CSS animations and transitions property parsing

ewilligers:
Test the parsing and serialization of the longhand properties from
CSS Animations and CSS Transitions.

Relevant browser bugs:

Edge animation-duration should reject negatives
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/18347049/

Edge and Safari incorrectly accept unitless 0 on `<time>` values
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/18348525/
https://bugs.webkit.org/show_bug.cgi?id=113230

### DIFF
--- a/css/css-animations/parsing/animation-delay-valid.html
+++ b/css/css-animations/parsing/animation-delay-valid.html
@@ -2,8 +2,7 @@
 <html>
 <head>
 <meta charset="utf-8">
-<title>CSS Animations Level 1: parsing animation-delay with valid values</title>
-<link rel="author" title="Eric Willigers" href="mailto:ericwilligers@chromium.org">
+<title>CSS Animations: parsing animation-delay with valid values</title>
 <link rel="help" href="https://drafts.csswg.org/css-animations/#propdef-animation-delay">
 <meta name="assert" content="animation-delay supports the full grammar '<single-animation-play-state> #'.">
 <script src="/resources/testharness.js"></script>
--- a/css/css-animations/parsing/animation-delay-valid.html
+++ b/css/css-animations/parsing/animation-delay-valid.html
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>CSS Animations Level 1: parsing animation-delay with valid values</title>
+<link rel="author" title="Eric Willigers" href="mailto:ericwilligers@chromium.org">
+<link rel="help" href="https://drafts.csswg.org/css-animations/#propdef-animation-delay">
+<meta name="assert" content="animation-delay supports the full grammar '<single-animation-play-state> #'.">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/css/support/parsing-testcommon.js"></script>
+</head>
+<body>
+<script>
+test_valid_value("animation-delay", "-5ms");
+test_valid_value("animation-delay", "0s");
+test_valid_value("animation-delay", "10s");
+test_valid_value("animation-delay", "20s, 10s");
+</script>
+</body>
+</html>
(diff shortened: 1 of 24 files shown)


css/css-animations/parsing/animation-name-invalid.html
```diff
@@ -0,0 +1,19 @@
...
+<link rel="author" title="Eric Willigers" href="mailto:ericwilligers@chromium.org">
+<link rel="help" href="https://drafts.csswg.org/css-animations/#propdef-animation-name">
+<meta name="assert" content="animation-name supports only the grammar '[ none | <keyframes-name> ]#'.">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/css/support/parsing-testcommon.js"></script>
+</head>
+<body>
+<script>
+test_invalid_value("animation-name", '12');
```
birtles: Can we add a test that 'none' is invalid. As per:
"The <custom-ident> additionally excludes the none keyword."
(https://drafts.csswg.org/css-animations/#keyframes)
And CSS-wide keywords like "inherit" and "default": As per:
"The CSS-wide keywords are not valid <custom-ident>s."
(https://drafts.csswg.org/css-values-4/#custom-idents)

css/css-animations/parsing/animation-name-valid.html
```diff
@@ -0,0 +1,24 @@
...
+</head>
+<body>
+<script>
+test_valid_value("animation-name", 'none');
+test_valid_value("animation-name", 'foo');
+test_valid_value("animation-name", 'both');
+test_valid_value("animation-name", 'ease-in');
+test_valid_value("animation-name", 'infinite');
+test_valid_value("animation-name", 'paused');
+test_valid_value("animation-name", 'first, second, third');
```
birtles: Can we add some tests for "strings"?
And escaped spaces in custom idents?
And otherwise-forbidden CSS-wide keywords which are allowed as strings, e.g. "initial"?
You can see some examples here: https://hg.mozilla.org/mozilla-central/rev/f41a419197de

css/css-animations/parsing/animation-delay-valid.html
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>CSS Animations Level 1: parsing animation-delay with valid values</title>
```
birtles: We can probably drop "Level 1" here.

css/css-animations/parsing/animation-delay-valid.html
```diff
@@ -0,0 +1,21 @@
...
+<link rel="author" title="Eric Willigers" href="mailto:ericwilligers@chromium.org">
```
birtles: (For what it's worth, we've been dropping the author annotations in files we've been submitting since it just seems to grow over time, but if you think it's useful feel free to leave it. We've also been dropping `<html>` and `<head>` tags but that's up to you too.)

birtles (approved):
Looks good to me with the extra `animation-name` tests added.

css/css-animations/parsing/animation-name-invalid.html
ewilligers: none is a valid value for the property, it simply isn't a keyword.

For CSS-wide keywords like inherit and initial, I'd prefer a separate test to check that all properties of a spec support them. Like testing inheritance/non-inheritance
https://github.com/web-platform-tests/wpt/pull/12593

css/css-animations/parsing/animation-name-valid.html
ewilligers: Tests added. They pass in Firefox.

birtles:
This is great, thanks Eric!

PR #12848, opened 2018-09-05
Add a script to download all the logs for a run from TaskCluster.

jgraham:
This currently has rather basic error handling and perhaps does more
HTTP requests than necessary, but it at least allows retriving all the
logs for a specified commit.

### DIFF
--- a/tools/ci/tcdownload.py
+++ b/tools/ci/tcdownload.py
@@ -1,0 +1,95 @@
+import argparse
+import os
+import logging
+
+import requests
+
+import github
+
+
+logging.basicConfig()
+logger = logging.getLogger("tc-download")
+
+
+def get_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--ref", action="store", default="master",
+                        help="Branch (in the GitHub repository) or commit to fetch logs for")
+    parser.add_argument("--artifact-name", action="store", default="wpt_report.json.gz",
+                        help="Log type to fetch")
+    parser.add_argument("--repo-name", action="store", default="web-platform-tests/wpt",
+                        help="GitHub repo name in the format owner/repo. "
+                        "This must be the repo from which the TaskCluster run was scheduled "
+                        "(for PRs this is the repo into which the PR would merge)")
+    parser.add_argument("--token-file", action="store",
+                        help="File containing GitHub token")
+    parser.add_argument("--out-dir", action="store", default=".",
+                        help="Path to save the logfiles")
+    return parser
+
+
+def get_json(url, key=None):
+    resp = requests.get(url)
+    resp.raise_for_status()
+    data = resp.json()
+    if key:
+        data = data[key]
+    return data
+
+
+def get(url, dest, name):
+    resp = requests.get(url)
+    resp.raise_for_status()
+    path = os.path.join(dest, name)
+    with open(path, "w") as f:
+        f.write(resp.content)
+    return path
+
+
+def run(*args, **kwargs):
+    if not os.path.exists(kwargs["out_dir"]):
+        os.mkdir(kwargs["out_dir"])
+
+    if kwargs["token_file"]:
+        with open(kwargs["token_file"]) as f:
+            gh = github.Github(f.read().strip())
+    else:
+        gh = github.Github()
+
+    repo = gh.get_repo(kwargs["repo_name"])
+    commit = repo.get_commit(kwargs["ref"])
+    statuses = commit.get_statuses()
+    taskgroups = set()
+
+    for status in statuses:
+        if not status.context.startswith("Taskcluster "):
+            continue
+        if status.state == "pending":
+            continue
+        taskgroup_id = status.target_url.rsplit("/", 1)[1]
+        taskgroups.add(taskgroup_id)
+
+    if not taskgroups:
+        logger.error("No complete TaskCluster runs found for ref %s" % kwargs["ref"])
+        return
+
+    for taskgroup in taskgroups:
+        taskgroup_url = "https://queue.taskcluster.net/v1/task-group/%s/list"
+        artifacts_list_url = "https://queue.taskcluster.net/v1/task/%s/artifacts"
+        tasks = get_json(taskgroup_url % taskgroup, "tasks")
+        for task in tasks:
+            task_id = task["status"]["taskId"]
+            url = artifacts_list_url % (task_id,)
+            for artifact in get_json(url, "artifacts"):
+                if artifact["name"].endswith(kwargs["artifact_name"]):
+                    filename = "%s-%s-%s" % (task["task"]["metadata"]["name"],
+                                             task_id,
+                                             kwargs["artifact_name"])
+                    path = get("%s/%s" % (url, artifact["name"]), kwargs["out_dir"], filename)
+                    logger.info(path)
+
+
+def __main__():
+    kwargs = get_parser().parse_args()
+
+    run(None, vars(kwargs))
--- a/tools/ci/commands.json
+++ b/tools/ci/commands.json
@@ -1,7 +1,42 @@
 {
-    "test-jobs": {"path": "jobs.py", "script": "run", "parser": "create_parser", "help": "List test jobs that should run for a set of commits",
-                  "virtualenv": false},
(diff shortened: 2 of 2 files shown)


tools/ci/tcdownload.py
```diff
@@ -0,0 +1,84 @@
...
+
+def get_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--ref", action="store", default="master",
+                        help="branch or commit to fetch logs for")
+    parser.add_argument("--artifact-name", action="store", default="wpt_report.json.gz",
+                        help="Log type to fetch")
+    parser.add_argument("--repo-name", action="store", default="web-platform-tests/wpt",
+                        help="GitHub repo name in the format owner/repo")
+    parser.add_argument("--token-file", action="store",
```
gsnedders: Can we either make this work without a token or make this required? Currently if you fail to to pass `--token-file` you get a TypeError, which isn't a great UX.

tools/ci/tcdownload.py
```diff
@@ -0,0 +1,84 @@
...
+import os
+
+import requests
+
+import github
+
+
+def get_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--ref", action="store", default="master",
```
gsnedders: What is this argument? Passing a ref like `gsnedders/exec-webdriver` produces `github.GithubException.GithubException: 422 {u'documentation_url': u'https://developer.github.com/v3/repos/commits/#get-a-single-commit', u'message': u'No commit found for SHA: gsnedders/exec-webdriver'}` here. Does this actually have to be a commit object?

jgraham: https://developer.github.com/v3/repos/statuses/#list-statuses-for-a-specific-ref

> The ref can be a SHA, a branch name, or a tag name

Is that a branch name? I suspect you want `--repo-name=gsnedders/web-platform-tests --ref=exec-webdriver` or something?

tools/ci/tcdownload.py
jgraham: The thinking was that in the future it could read the token from an environment variable or somewhere else so making it a required argument will result in API churn. But sure, the error should be nicer.

gsnedders: I mean, in the common case, we shouldn't hit the rate limit here even without a token, no?

tools/ci/tcdownload.py
gsnedders: Oh, I was expecting it to be a local refish.

Given GitHub exposes all commits in the entire network through any repo (given they don't track which commits are reachable from refs in a given repo), it'd be nice to just resolve refs to commit objects locally and pass that up to the GH API, given then it would work more like most other git CLIs?

tools/ci/tcdownload.py
jgraham: I though it required authentication, but apparently I jsut typoed the URL. Fixed.

tools/ci/tcdownload.py
jgraham: Sure, but it doesn't seem like a necessary feature to land something here. The current implementation doesn't depend on actually having git or anything, which can be useful in some scenarios.

gsnedders: We should clarify what the arg is and that it must be something resolvable on GitHub within that repo in the `--help` string at least, IMO.

jgraham: Done.

jugglinmike (commented):
While testing this, I got myself all confused about the semantics of the `--repo-name` flag.

gh-12657 was submitted from [bocoup/wpt](https://github.com/bocoup/wpt) and includes d9a12c3b3484278b5, a commit with "complete" TaskCluster status. Trying to download from that repository fails:

    $ ls *.json.gz
    ls: cannot access '*.json.gz': No such file or directory
    $ ./wpt tc-download --repo-name bocoup/wpt --ref d9a12c3b3484278b5
    No complete TaskCluster runs found for ref d9a12c3b3484278b5
    $ ls *.json.gz
    ls: cannot access '*.json.gz': No such file or directory

...but if I omit the `--repo-name` argument, the download succeeds:

    $ ls *.json.gz
    ls: cannot access '*.json.gz': No such file or directory
    $ ./wpt tc-download --ref d9a12c3b3484278b5
    $ ls *.json.gz
    wpt-chrome-dev-results-0-wpt_report.json.gz  wpt-firefox-nightly-results-0-wpt_report.json.gz

I can reproduce this by querying the GitHub API directly:

    $ curl -s https://api.github.com/repos/bocoup/wpt/commits/d9a12c3b3484278b5/statuses | python -c 'import sys, json; print len(json.loads(sys.stdin.read()))'
    0
    $ curl -s https://api.github.com/repos/web-platform-tests/wpt/commits/d9a12c3b3484278b5/statuses | python -c 'import sys, json; print len(json.loads(sys.stdin.read()))'
    17

My best guess is that GitHub "statuses" are associated with commits via the repository that triggered the status check (and that this is the "target" repo in the case of a pull request). I suppose I would only need to specify `--repo-name` if I were dealing with code I pushed to a fork directly.

Is that surprising to anyone else? Is it something we can document in simple terms?

tools/ci/tcdownload.py
```diff
@@ -0,0 +1,87 @@
...
+                        filename = "%s-%s-%s" % (task["task"]["metadata"]["name"],
+                                                 run_id,
+                                                 kwargs["artifact_name"])
+                        get("%s/%s" % (url, artifact["name"]), kwargs["out_dir"], filename)
+
+
+def __main__():
+    kwargs = get_parser().parse_args()
+
+    run(None, vars(kwargs))
```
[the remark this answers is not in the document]
jgraham: It's the virtualenv argument which we always pass in for wpt commands that run in in a virtualenv. But we aren't actually using it anywhere so we can pass `None` in quite safely.

gsnedders:
> My best guess is that GitHub "statuses" are associated with commits via the repository that triggered the status check (and that this is the "target" repo in the case of a pull request). I suppose I would only need to specify --repo-name if I were dealing with code I pushed to a fork directly.

That's correct.

tools/ci/tcdownload.py
```diff
@@ -0,0 +1,97 @@
...
+        logger.error("No complete TaskCluster runs found for ref %s" % kwargs["ref"])
+        return
+
+    for taskgroup in taskgroups:
+        taskgroup_url = "https://queue.taskcluster.net/v1/task-group/%s/list"
+        artifacts_list_url = "https://queue.taskcluster.net/v1/task/%s/runs/%s/artifacts"
+        tasks = get_json(taskgroup_url % taskgroup, "tasks")
+        for task in tasks:
+            task_id = task["status"]["taskId"]
+            for run in task["status"]["runs"]:
```
Hexcles: When will we have multiple runs for a task? I didn't find any. Retries? And in those cases, do we need results from all the runs, or just the latest run?

jgraham: So jobs can be retriggered, which I believe will lead to a second run. But I'm certainly not sure this is the right logic; it just seemed like a reasonable conservative thing to do.

jugglinmike: [Jobs cannot *currently* be re-triggered.](https://bugzilla.mozilla.org/show_bug.cgi?id=1484044) That's being worked on, so it shouldn't effect this decision; this just seemed like a good time to point that out.

jugglinmike (approved):

tools/ci/tcdownload.py
```diff
@@ -0,0 +1,97 @@
...
+                        "This must be the repo from which the TaskCluster run was scheduled "
+                        "(for PRs this is the repo into which the PR would merge)")
+    parser.add_argument("--token-file", action="store",
+                        help="File containing GitHub token")
+    parser.add_argument("--out-dir", action="store", default=".",
+                        help="Path to save the logfiles")
+    return parser
+
+
+def get_json(url, key=None):
```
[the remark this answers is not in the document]
Hexcles: This would be a nice-to-have optimization, but I don't think it's blocking. I'll leave it to @jgraham to decide whether to implement it in this PR :)

Hexcles (approved):
This now LGTM

PR #12867, opened 2018-09-06
[css-contain] the contain property is not animatable

frivoal:
Tests for https://github.com/w3c/csswg-drafts/issues/2737

### DIFF
--- a/css/css-contain/contain-animation-001.html
+++ b/css/css-contain/contain-animation-001.html
@@ -1,0 +1,31 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Containment Test: contain is not animatable</title>
+<link rel="author" title="Florian Rivoal" href="https://florian.rivoal.net/">
+<link rel="help" href="https://drafts.csswg.org/css-contain/#contain-property">
+<link rel="match" href="../reference/ref-filled-green-100px-square.xht">
+<meta name=assert content="the contain property is not animatable">
+<style>
+div {
+  border: 50px solid green;
+  background: red;
+  position: absolute; /* for shrinkwrap */
+  contain: strict;
+
+  animation-duration: 1s;
+  animation-name: bad;
+  animation-play-state: paused;
+
+  font-size: 100px;
+}
+
+@keyframes bad {
+  from {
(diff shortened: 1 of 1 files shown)


birtles (approved):
I'm not so familiar with css-contains but from what I understand it looks good to me.

Also I'm not sure how well the w3c reftest harness waits for animations (I always just write testharness tests), so might it be worth adding `reftest-wait` then dropping it in `requestAnimationFrame` callback?

For css-logical we just used the animation API:
https://github.com/web-platform-tests/wpt/blob/3c6c9350988f98c8fac94fab8174bc4729cd6755/css/css-logical/animation-001.html#L302-L321

frivoal:
@birtles 

> Also I'm not sure how well the w3c reftest harness waits for animations

Is there anything to wait for? the animation is paused, and the initial state is supposed to apply immediately.

The reftest system wouldn't wait for an animation to complete, so if that's what we wanted, we'd need `refttest-wait` followed signaling when it is ready, but with this test, I believe the initial state should already be correct.

Am I missing something?

birtles:
> Is there anything to wait for? the animation is paused, and the initial state is supposed to apply immediately.

Right, the `animation-*` properties will be resolved immediately generating a new animation. In Gecko at least we'll restyle again immediately applying the starting values of the animation even while the animation has yet to begin running. The spec doesn't strictly require this (only the opposite, that an animation does _not_ affect computed style outside of when it is applied) but hopefully everyone does that. There was certainly something in this area where Edge was different but I suspect it was specific to compositor animations.

> The reftest system wouldn't wait for an animation to complete, so if that's what we wanted, we'd need refttest-wait followed signaling when it is ready, but with this test, I believe the initial state should already be correct.
> 
> Am I missing something?

No, my initial reaction came from working on SMIL reftests where the animation doesn't start until onload (hence you need to remove `reftest-wait` on onload). I wasn't suggesting we wait for the animation to end, just that if not all user agents are applying the animated style on the same frame, we might need to wait for the next animation frame.

frivoal:
I think it's fine. If it turns out someone does it differently and fail the test due to that, they can open a bug at that point and we'll look into it then.

PR #12870, opened 2018-09-06
Don't try to parse the contents of server errors in idl_test().

Ms2ger:


resources/idlharness.js
```diff
@@ -3251,7 +3251,11 @@ function idl_test(srcs, deps, idl_setup_func) {
  * fetch_spec is a shorthand for a Promise that fetches the spec's content.
  */
 function fetch_spec(spec) {
-    return fetch('/interfaces/' + spec + '.idl').then(function (r) {
+    var url = '/interfaces/' + spec + '.idl';
+    return fetch(url).then(function (r) {
+        if (!r.ok) {
+            throw new IdlHarnessError("Error fetching " + url + ".");
```
foolip: Thanks, I've been meaning to to something like this myself, as it happens when you make typos :)

foolip (approved):

PR #12875, opened 2018-09-06
Don't pass --install-browser to Chrome in stability check

jgraham:


tools/wpt/run.py
```diff
@@ -447,8 +447,11 @@ def setup_wptrunner(venv, prompt=True, install_browser=False, **kwargs):
     del kwargs["channel"]
 
     if install_browser:
-        logger.info("Installing browser")
-        kwargs["binary"] = setup_cls.install(venv, channel=channel)
+        try:
+            logger.info("Installing browser")
+            kwargs["binary"] = setup_cls.install(venv, channel=channel)
+        except NotImplementedError:
+            logger.warning("Installing browser not supported for %s" % kwargs["product"])
```
gsnedders: If the user has asked to install the browser and we can't, that should be a fatal error, not a warning.

gsnedders (changes requested):

gsnedders (approved):

PR #10998, opened 2018-05-14
 [wptrunner] Recover from malformed test manifest 

jugglinmike:
Ensure that the WPT CLI exits successfully when the `--list-tests`
option is specified (including the case where no test manifest is
present).

---

If the process of updating the test manifest is interrupted
unexpectedly, the file written to disk may not contain valid JSON.
Update that WPT CLI to recognize this condition and recover by
re-generating the manifest file.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/web-platform-tests/10998)

PR #12877, opened 2018-09-06
Don't try to normalise the channel if we don't have data for a product.

jgraham:


gsnedders (changes requested):
Please add tests so we don't regress `wpt run`. We should make sure `setup_wptrunner` can work with all known products.

gsnedders:
@jgraham any idea what caused this to fail CI? the tests that fail are all from @jugglinmike's #10998.

tools/wpt/tests/test_run.py
```diff
@@ -16,3 +22,22 @@ def test_check_environ_fail(platform):
                 run.check_environ("foo")
 
     assert "wpt make-hosts-file" in excinfo.value.message
+
+
+@pytest.mark.parametrize("product", product_list)
+def test_setup_wptrunner(product):
```
gsnedders: This test seems to change some global state, making later tests fail.

`tox -e py27 -- --no-cov -k 'test_run_firefox or test_setup_wptrunner'` is sufficient to observe a failure; Travis has many more.

gsnedders (changes requested):

jgraham:
I made the requested changes some time ago. @gsnedders please re-review.

gsnedders (approved):

Hexcles (commented):
Found two bugs caused by this PR.

foolip:
Given the regression in https://github.com/web-platform-tests/wpt/pull/13567, what sort of test coverage are we missing here?

PR #12883, opened 2018-09-06
Improve the formatting of testdriver documentation

foolip:
In https://web-platform-tests.org/writing-tests/testdriver.html it is currently
hard to tell where the different sections begin and end because the formatting
of headings looks just like a line of code.


Hexcles (approved):
This looks clearer to me, so LGTM.

PR #12885, opened 2018-09-06
HTML: Fix test when run through ./wpt run

TimothyGu:
When run through `./wpt run`, the test window is created by another window through `window.open()`, so `opener` is not `null` in the main page (when we expect it to be `null`).

@foolip, can you double check this looks correct?

### DIFF
--- a/html/webappapis/dynamic-markup-insertion/opening-the-input-stream/reload.window.js
+++ b/html/webappapis/dynamic-markup-insertion/opening-the-input-stream/reload.window.js
@@ -15,13 +15,13 @@
 // as the test file and as part of the test file. The `if (!opener)` condition
 // controls what role this file plays.
 
-if (!opener) {
+if (window.name !== "opened-dummy-window") {
   async_test(t => {
     const testURL = document.URL;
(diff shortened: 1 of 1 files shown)


foolip (approved):
Looks good, `./wpt run` does open new windows and this fix looks correct.

domenic:
@foolip it's worrying to me that test authors have to be aware of this. Is there anything we could do?

foolip:
@domenic I think the code in question is here:
https://github.com/web-platform-tests/wpt/blob/master/tools/wptrunner/wptrunner/executors/testharness_webdriver.js#L18

I don't think we could use `window.open("%(abs_url)s", "%(window_id)s", "noopener")` here, because the two windows communicate to make testdriver.js work:
https://github.com/web-platform-tests/wpt/blob/master/tools/wptrunner/wptrunner/testdriver-extra.js

Not sure if there are other places that depend on it as well. @gsnedders, do you know? Any ideas for how we could make this trap go away?

domenic:
Personally I think playing in the global scope of the test is much less dangerous than messing with web platform features that might themselves be under test.

PR #12893, opened 2018-09-07
Add tests for mediacapture-image

xiuqijix:
Cover following checkpoints:
- throw TypeError when ImageCapture parameter is not mediaStramTrack
- ImageCapture track attribute is readonly
- getPhotoCapabilities() of an ended Track should throw InvalidStateError
- getPhotoSettings() of an ended Track should throw InvalidStateError
- takePhoto() of an ended Track should throw InvalidStateError
- the parameter of takePhoto() is nullable

### DIFF
--- a/mediacapture-image/ImageCapture-track.html
+++ b/mediacapture-image/ImageCapture-track.html
@@ -16,17 +16,16 @@ test(t => {
 
   let stream = canvas.captureStream();
   let videoTrack = stream.getVideoTracks()[0];
-  assert_equals(videoTrack.readyState, "live");
-  assert_true(videoTrack.enabled);
 
   let capturer = new ImageCapture(videoTrack);
   assert_true(capturer.track instanceof MediaStreamTrack);
   assert_equals(capturer.track, videoTrack);
 
-  capturer.track.enabled = false;
-  assert_true(videoTrack.enabled, "track enabled");
-  capturer.track.stop();
-  assert_equals(videoTrack.readyState, "live", "track readyState");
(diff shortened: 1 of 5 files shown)


mediacapture-image/ImageCapture-track.html
```diff
@@ -0,0 +1,32 @@
...
+  let stream = canvas.captureStream();
+  let videoTrack = stream.getVideoTracks()[0];
+  assert_equals(videoTrack.readyState, "live");
+  assert_true(videoTrack.enabled);
+
+  let capturer = new ImageCapture(videoTrack);
+  assert_true(capturer.track instanceof MediaStreamTrack);
+  assert_equals(capturer.track, videoTrack);
+
+  capturer.track.enabled = false;
```
reillyeon: Is this really the proper definition of `readonly`. The `MediaStreamTrack` returned by the attribute getter should be the same (and just as modifiable through mutable attributes and exposed methods) as `videoTrack` is above.

kaixinjxq: Yes, you are right, my mistake. Test the `readonly` attribute with a clone `MediaStreamTrack` now, PTAL.

reillyeon (approved):

kaixinjxq:
@reillyeon, thanks for your review!
`InvaildStateError` does not thrown when invoking `getPhotoCapabilities()/getPhotoSettings()` with an ended Track, I've filed a bug to track it https://bugs.chromium.org/p/chromium/issues/detail?id=882730

PR #12924, opened 2018-09-10
url: Added test cases for IDN-encoded input, valid and invalid.

mgiuca:
In particular, exposes implementation bugs around URLs with "xn--" prefix but
containing non-ASCII characters (https://crbug.com/804688).

Tested against:
- https://jsdom.github.io (all tests pass)
- Google Chrome 70. These tests fail:
  - `sc://xn--fa-hia.ExAmPlE/` and `sc://xn--abcd.com/` (due to unrelated issue)
  - `https://xn--abcd.com/` (expected to throw)
  - `https://ab--abcdé.com/` (throws, due to https://crbug.com/804688)
- Firefox 62. These tests fail:
  - `sc://xn--fa-hia.ExAmPlE/` and `sc://xn--abcd.com/` (due to unrelated issue)
  - `https://xn--abcd.com/` (expected to throw)
  - `https://xn--abcdé.com/` (expected to throw)

Re the case `https://xn--abcd.com/`: Given that both Chrome and Firefox agree that an invalid IDN-encoded domain should pass through unmodified, it's possible the spec should be updated to allow this. This would probably entail modifying [domain to ASCII](https://url.spec.whatwg.org/#concept-domain-to-ascii), replacing "If *result* is a failure value, validation error, return failure." with "If *result* is a failure value, validation error, then if *result* is an ASCII string, return *result*, else return failure."

But for the time being, I think we should land a test matching the specified text.

**Edit**: Updated with Firefox results and paragraph about changing the spec.


GPHemsley:
w3c-test:mirror

TimothyGu:
As I commented on the crbug, there’s a lot of history in this. See https://github.com/whatwg/url/commit/dc9d83106cada9af507bf37dee3973de97b020fd and the linked tickets. For this PR, you might want to dedupe the tests against the existing ones in https://github.com/web-platform-tests/wpt/blob/master/url/resources/toascii.json, and add the tests to that file.

GEEKGIRL36:
Pppoo

On Sep 9, 2018 10:00 PM, "Timothy Gu" <notifications@github.com> wrote:

As I commented on the crbug, there’s a lot of history in this. See
whatwg/url@dc9d831
<https://github.com/whatwg/url/commit/dc9d83106cada9af507bf37dee3973de97b020fd>
and the linked tickets. For this PR, you might want to dedupe the tests
against the existing ones in
https://github.com/web-platform-tests/wpt/blob/master/url/resources/toascii.json,
and add the tests to that file.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/web-platform-tests/wpt/pull/12924#issuecomment-419787734>,
or mute the thread

mgiuca:
@TimothyGu Thanks. I didn't realise there were tests for this in `toascii.json`. Looks like the right place to have these (though maybe there should be at least one end-to-end test in url-constructor as well).

However, I can't find any way to *run* the `toascii` tests in a browser. It doesn't seem to have an associated .html file like `url-constructor.html`. How do you run these tests?

FYI I've updated the PR description with some more analysis including Firefox.

annevk:
`toascii.window.js` can be run as `toascii.window.html`. See https://web-platform-tests.org/writing-tests/testharness.html#auto-generated-test-boilerplate for documentation.

GPHemsley:
w3c-test:mirror

annevk:
I think we have sufficient IDN coverage at this point.

PR #12932, opened 2018-09-10
HTML: test rendering of legend with different display values

zcorpan:

### DIFF
--- a/html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/legend-display-rendering-ref.html
+++ b/html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/legend-display-rendering-ref.html
@@ -1,0 +1,34 @@
+<!doctype html>
+<title>Reference for rendered legend and CSS display rendering</title>
+<style>
+body { margin: 0; }
+div { width: 600px; box-sizing: border-box; background: #ddd; border: 1px solid; }
+.padding { padding-left: 5em; }
+.margin { margin-left: 5em; }
+.cell { display: inline-block; width: 50%; }
+.ruby { display: block; }
+</style>
+<div><span class=cell>table</span><span class=cell>table</span></div>
+<div class=padding>table-row-group</div>
+<div class=padding>table-header-group</div>
+<div class=padding>table-footer-group</div>
+<div class=padding>table-row</div>
+<div class=margin>table-cell</div>
+<div class=padding>table-column-group</div>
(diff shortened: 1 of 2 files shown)


html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/legend-display-rendering-ref.html
```diff
@@ -0,0 +1,34 @@
...
+<div class=padding>table-column</div>
+<div>table-caption</div>
+<div>flow</div>
+<div>flow-root</div>
+<div>run-in</div>
+<div>inline</div>
+<div>inline-block</div>
+<div><span class=cell>inline-table</span><span class=cell>inline-table</span></div>
+<div><span class=ruby>ruby</span><span class=ruby>ruby</span></div>
+<div>ruby-base</div>
```
mstensho: What's these four lines supposed to test? Blink passes, although it has no support for ruby display types.

zcorpan: That they blockify. Blink fails the "ruby" test.

mstensho: Right. Guess there's no real point in making these fail for engines (like Blink) that don't support these display property values at all. So, fine.

zcorpan: Yeah. There's a separate test (legend-display.html) that fails when the value is not supported.

html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/legend-display-rendering.html
```diff
@@ -0,0 +1,119 @@
+<!doctype html>
+<title>rendered legend and CSS display rendering</title>
+<link rel=match href=legend-display-rendering-ref.html>
+<style>
+body { margin: 0; }
+fieldset { margin: 0; padding: 0; border: none; }
+legend { width: 600px; box-sizing: border-box; padding: 0; background: #ddd; border: 1px solid; }
+[style="display: table"] span,
+[style="display: inline-table"] { display: table-cell; }
```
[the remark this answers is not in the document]
zcorpan: oops, fixed.

mstensho (approved):

Issue #12935, opened 2018-09-10
tools/wpt testsuite failures on Windows


gsnedders:
Excluding the tests mentioned in #12934, running this gets the following:

```
(wpt-env) C:\Users\geoff\Documents\projects\wpt\tools\wpt [master ≡]> tox -e py27 -- --no-cov -vv -k 'not test_tests_aff
ected'
py27 installed: asn1crypto==0.24.0,atomicwrites==1.2.1,attrs==18.2.0,blessings==1.7,certifi==2018.8.24,cffi==1.11.5,chardet==3.0.4,colorama==0.3.9,coverage==4.5.1,cryptography==2.3.1,enum34==1.1.6,funcsigs==1.0.2,html5lib==1.0.1,hypothesis==3.71.3,idna==2.7,ipaddress==1.0.22,marionette-driver==2.6.0,mock==2.0.0,mohawk==0.3.4,more-itertools==4.3.0,mozcrash==1.0,mozdebug==0.1,mozdevice==0.52,mozdownload==1.24,mozfile==1.2,mozinfo==0.10,mozInstall==1.16.0,mozleak==0.1,mozlog==3.8,moznetwork==0.27,mozprocess==0.26,mozprofile==1.1.0,mozrunner==7.0.0,mozterm==1.0.0,mozversion==1.5,pathlib2==2.3.2,pbr==4.2.0,pluggy==0.7.1,progressbar2==3.38.0,py==1.6.0,pycparser==2.18,pyOpenSSL==18.0.0,pytest==3.8.0,pytest-cov==2.6.0,python-utils==2.3.0,redo==1.6,requests==2.19.1,requests-hawk==1.0.0,scandir==1.9.0,selenium==3.14.0,six==1.11.0,treeherder-client==4.0.0,urllib3==1.22,webencodings==0.5.1
py27 runtests: PYTHONHASHSEED='112'
py27 runtests: commands[0] | pytest --cov --no-cov -vv -k 'not test_tests_affected'
============================= test session starts =============================
platform win32 -- Python 2.7.15, pytest-3.8.0, py-1.6.0, pluggy-0.7.1 -- c:\users\geoff\documents\projects\wpt\tools\wpt\.tox\py27\scripts\python.exe
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('C:\\Users\\geoff\\Documents\\projects\\wpt\\tools\\wpt\\.hypothesis\\examples')
[...]

PR #12936, opened 2018-09-10
Mark tools/wpt Windows test failures as xfail

gsnedders:
See https://github.com/web-platform-tests/wpt/issues/12935


jgraham (approved):

jugglinmike (commented):
The first two tests are failing because `wpt run` forks a process which attempts to install `requests` with PIP. That's as much as I can say without a dedicated Windows machine, so maybe it's okay to simply skip those tests for now. (Thanks for filing a separate issue to track this, by the way.)

The third test fails because it expects a manifest file to be present, implicitly relying on the passing status of one of the preceding two tests.  This is wrong on any platform, and it only circumstantially passes in GNU/Linux today. I'd prefer if we addressed it by inserting an explicit call to `wpt manifest` instead. What do you think?

PR #12940, opened 2018-09-10
HTML: -webkit-appearance: fieldset/legend: auto on replaced elements

zcorpan:

### DIFF
--- a/html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/fieldset-replaced-element-ref.html
+++ b/html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/fieldset-replaced-element-ref.html
@@ -1,0 +1,12 @@
+<!doctype html>
+<title>Reference for fieldset on replaced elements</title>
+<img src=/images/pattern.png>
+<video>
+ <source src=/images/pattern.mp4 type=video/mp4>
+ <source src=/images/pattern.ogv type=video/ogg>
+</video>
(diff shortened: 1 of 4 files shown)


mstensho (commented):
General note: I think it's wrong to use -webkit- prefixed properties.

annevk:
@mstensho is there a good place to discuss that? I think we should go with -webkit-appearance for now since that's what implementations use. As I mentioned elsewhere I don't think we can remove it. We might at some point support an unprefixed alias as well, but if it's an alias it shouldn't matter that the prefixed version is tested here.

mstensho:
No, we can't remove it. But I'd like to have a proper (unprefixed, of course) spec for this first. I'd prefer to live with the fieldset / legend magic until we have that spec. We're doing things in the wrong order here.

annevk:
How does it matter if the end state is the same?

mstensho:
Because in the meantime the web will have been polluted with even more -webkit- stuff, and that can be avoided.

zcorpan:
You can keep it behind a flag in the meantime, though?

I concur with Anne, FWIW.

zcorpan:
@mstensho @MatsPalmgren do you think it helps to unprefix 'appearance' now (with -webkit-appearance as a simple alias, ignoring css-ui for now)?

zcorpan:
See https://github.com/whatwg/html/pull/4035#issuecomment-1476330098

PR #12944, opened 2018-09-11
Use document.scrollingElement to access scroll properties of the view…

fred-wang:
…port

See https://drafts.csswg.org/cssom-view/#dom-document-scrollingelement

Please do not merge until someone confirms document.scrollingElement is supported by Microsoft Edge.

https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/17744174/ seems to suggest it always returns the body.

### DIFF
--- a/conformance-checkers/html-aria/_functional/tree/js/prototype.js
+++ b/conformance-checkers/html-aria/_functional/tree/js/prototype.js
@@ -2705,8 +2705,8 @@ document.viewport = {
 
   getScrollOffsets: function() {
     return Element._returnOffset(
-      window.pageXOffset || document.documentElement.scrollLeft || document.body.scrollLeft,
-      window.pageYOffset || document.documentElement.scrollTop || document.body.scrollTop);
+      window.pageXOffset || document.scrollingElement.scrollLeft,
+      window.pageYOffset || document.scrollingElement.scrollTop);
   }
 };
 /* Portions of the Selector class are derived from Jack Slocumâ€™s DomQuery,
@@ -3751,10 +3751,8 @@ Event.Methods = (function() {
 
     pointer: function(event) {
       return {
-        x: event.pageX || (event.clientX +
-          (document.documentElement.scrollLeft || document.body.scrollLeft)),
-        y: event.pageY || (event.clientY +
-          (document.documentElement.scrollTop || document.body.scrollTop))
+        x: event.pageX || (event.clientX + document.scrollingElement.scrollLeft),
+        y: event.pageY || (event.clientY + document.scrollingElement.scrollTop)
       };
     },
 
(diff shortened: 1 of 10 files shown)


foolip:
@fred-wang is you rebase this on master I could trigger a full run on Edge Dev or Canary to see what the results are. For Chrome, Firefox and Safari we run the affected tests automatically, but not for Edge yet.

@mustjab FYI in case you'd like to set up per-PR runs of Edge Dev or Canary and the custom wpt.fyi checks.

fred-wang:
I have completely forgotten about this one. I think it would be worth looking again in all WPT tests where document.documentElement / document.body are used to access scroll properties and use scrollingElement instead.

fred-wang:
> @fred-wang is you rebase this on master I could trigger a full run on Edge Dev or Canary

I expect they use recent versions of Blink so they should support document.scrollingElement

foolip:
That would be great, yeah.

foolip:
Yes, I'd expect that too, just addressing the concern as stated a year ago :)

fred-wang:
I updated the PR with more changes. In most cases, documentElement is called on a document that is not in quirks mode, so this is equivalent to document.scrollingElement. 

The only behavior change is really in element-timing/scroll-to-text.html where document.body.scrollHeight is the offset used to scroll to the end of the document. Testing in Chromium, document.documentElement.scrollHeight gives a larger value for me and I believe this is really what the author wants, but we should probably check with him.

foolip:
element-timing/scroll-to-text.html  came from https://github.com/web-platform-tests/wpt/pull/19525. @npm1 can you review that change?

andreastt (dismissed):
Overall this looks good and more concise, but we can’t make this change to any JS under webdriver/ because IE7 is a target and according to https://developer.mozilla.org/en-US/docs/Web/API/document/scrollingElement `Document.scrollingElement` does not exist there.

fred-wang:
> Overall this looks good and more concise, but we can’t make this change to any JS under webdriver/ because IE7 is a target and according to https://developer.mozilla.org/en-US/docs/Web/API/document/scrollingElement `Document.scrollingElement` does not exist there.

fair enough. I removed that part.

> element-timing/scroll-to-text.html came from #19525. @npm1 can you review that change?

thanks, I had already pinged him on the chromium issue but coudn't find his github nickname

npm1 (approved):
element-timing lgtm

fred-wang:
> @fred-wang is you rebase this on master I could trigger a full run on Edge Dev or Canary to see what the results are. For Chrome, Firefox and Safari we run the affected tests automatically, but not for Edge yet.

@foolip Are you still planning to do that full run on Edge Dev & Canary?

foolip:
@fred-wang no I don't think that's necessary, and also the Edge setup is having a bit of trouble now

foolip:
I've filed https://github.com/web-platform-tests/wpt/issues/19975 about the current outage of Edge Dev results.

PR #12956, opened 2018-09-11
Test resolved value for inset CSS properties

Loirooriol:
See https://bugs.webkit.org/show_bug.cgi?id=188711

### DIFF
--- a/css/cssom/getComputedStyle-insets-absolute.html
+++ b/css/cssom/getComputedStyle-insets-absolute.html
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSSOM: resolved values of the inset properties for absolute positioning</title>
+<link rel="help" href="https://drafts.csswg.org/cssom/#resolved-value">
+<link rel="help" href="https://drafts.csswg.org/css-position/#pos-sch">
+<link rel="author" title="Oriol Brufau" href="mailto:obrufau@crisal.io">
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<script type="module">
+import {runTests, containerForAbspos} from "./support/getComputedStyle-insets.js";
+runTests({
+  style: "position: absolute",
+  containingBlockElement: containerForAbspos,
+  containingBlockArea: "padding",
+  preservesPercentages: false,
+  preservesAuto: false,
+  canStretchAutoSize: true,
+  staticPositionY: 1 + 2 + 4 + 8,
+  staticPositionX: 2 + 4 + 8 + 16,
+});
+</script>
--- a/css/cssom/getComputedStyle-insets-fixed.html
+++ b/css/cssom/getComputedStyle-insets-fixed.html
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSSOM: resolved values of the inset properties for fixed positioning</title>
+<link rel="help" href="https://drafts.csswg.org/cssom/#resolved-value">
+<link rel="help" href="https://drafts.csswg.org/css-position/#pos-sch">
+<link rel="author" title="Oriol Brufau" href="mailto:obrufau@crisal.io">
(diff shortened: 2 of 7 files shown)


Loirooriol:
No browser passes all tests:

 - Firefox
   - Fails some sticky positioning tests because it resolves percentages in a way that I didn't understand, so I just tested what Chromium and WebKit do.
   - Fails some absolute and fixed positioning tests because when there is overconstraintment, the resolved value is the used value. According to https://drafts.csswg.org/cssom/#resolved-value-special-case-property-like-top it should be the computed value (which Firefox respects for relative positioning).
 - Chromium, WebKit
   - Fail some absolute and fixed positioning tests because the static position is wrong when the writing mode of the element is different from its containing block. https://bugs.webkit.org/show_bug.cgi?id=189513
 - Chromium
   - Fails some relative, sticky, absolute and fixed positioning tests because the resolved value of a percentage is calculated with respect to the wrong containing block size in orthogonal flows (the resolved value is wrong, the used one is right).
   - Fails some fixed positioning tests because percentages are resolved ignoring the padding of the containing block (the resolved value is wrong, the used one is right).
 - WebKit
   - Fails some relative, absolute and fixed positioning tests because when there is overconstraintment, the resolved value is the used value. I plan to fix this in https://bugs.webkit.org/show_bug.cgi?id=188711
   - Fails some sticky tests because the resolved value is the raw computed value without absolutizing percentages or `calc()`. I plan to fix this in https://bugs.webkit.org/show_bug.cgi?id=188711
   - Fails all remaining fixed positioning tests because the resolved value is calculated using the border area as the containing block (but the effect of a specified value is with respect to the padding area as expected). I plan to fix this for non-`auto` values in https://bugs.webkit.org/show_bug.cgi?id=188711, for `auto` values it's https://bugs.webkit.org/show_bug.cgi?id=189518

@emilio Can you please review the patch, since you are a CSSOM editor?

emilio:
Thanks so much for writing these tests! I'll take a look, sure. I think I want to understand the different bugs here.

emilio:
> Fails some absolute and fixed positioning tests because when there is overconstraintment, the resolved value is the used value. According to https://drafts.csswg.org/cssom/#resolved-value-special-case-property-like-top it should be the computed value (which Firefox respects for relative positioning).

I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1490619 for this.

emilio:
> Fails some sticky positioning tests because it resolves percentages in a way that I didn't understand, so I just tested what Chromium and WebKit do.

I think what Firefox is doing is the right thing. It's resolving percentages against the thing you stick and position against, which is what layout does, and makes sense I'd think.

In the test-case the thing you stick against is the viewport, since `#container-for-inflow` has `overflow: visible`.

Do you agree with that? If so, the test should be changed, and a bug should be filed against Blink and WebKit.

emilio (commented):
I think we need to sort out the sticky test, other than that it looks great!

A way to sort out the sticky test may be to put `overflow: hidden` on the `#container-for-inflow` rule and file a spec issue, but I think what Firefox does is unambiguously better. Perhaps a standalone test for that would be useful.

Thanks a lot for writing the tests!

Loirooriol:
> Do you agree with that? If so, the test should be changed, and a bug should be filed against Blink and WebKit.

I don't know the details of sticky positioning but I guess it makes sense.

In fact only Blink is currently affected by this. I planned to copy its behavior into WebKit in https://bugs.webkit.org/show_bug.cgi?id=188711, but if Blink is wrong I will leave the current behavior for now (the resolved value is the raw computed value without absolutizing percentages).

Loirooriol:
@emilio Does it look good now?

emilio (approved):
Assuming the only change apart from nits and such is adding overflow: hidden to `#container-for-inflow` then yes, it looks great to me, thanks!

And thanks for the ping, github doesn't always send mail when you push apparently.

Loirooriol:
Yes, the changes were
 - Added `overflow: hidden` to `#container-for-inflow`
 - Added comment with link to spec
 - Removed newline character

PR #12958, opened 2018-09-12
[testharness.js] Disallow `test` return value

jugglinmike:
As a follow up to gh-12898, I wondered about similar test authoring mistakes
that we could avoid from within testharness.js. Are there any synchronous tests
that return a value? This would generally be harmless, but it that value were a
thenable, then it could be an unintentional use of `test` instead of
`promise_test`. TaskCluster again helped to answer the question:

    $ ./wpt tc-download --ref suspicious-sync --repo bocoup/wpt --artifact-name suspicious.txt --out-dir suspicious
    INFO:tc-download:suspicious-2/wpt-firefox-nightly-testharness-4-DlCidQw0SvOYFb_9dvqWLg-suspicious.txt
    INFO:tc-download:suspicious-2/wpt-chrome-dev-testharness-4-P0O_249FTdGyXlloowzNRQ-suspicious.txt
    cat suspicious/*
    http://web-platform.test:8000/dom/events/CustomEvent.html - this is likely a falsey mistake
    http://web-platform.test:8000/web-animations/timing-model/animations/setting-the-target-effect-of-an-animation.html - Setting the target effect to null causes a pending playback rate to be applied
    http://web-platform.test:8000/dom/events/CustomEvent.html - this is likely a falsey mistake
    http://web-platform.test:8000/web-animations/timing-model/animations/setting-the-target-effect-of-an-animation.html - Setting the target effect to null causes a pending playback rate to be applied

---

The `test` function does not recognize the return value of the provided
test body. Authors may mistakenly use this API instead of
`promise_test`, a function which is designed to respond to promise
values. Previously, mistakes like this would produce tests that
spuriously passed.

Update the `test` function to fail immediately in response to a body
that returns a value. Also correct a test which exhibited the
programming mistake that this change is intended to discourage (a survey
using both the Chrome and Firefox browsers indicated that this is the
only test in WPT which would be affected by this change).

### DIFF
--- a/resources/testharness.js
+++ b/resources/testharness.js
@@ -548,7 +548,7 @@ policies and contribution forms [3].
                 "\" inappropriately returned a value";
 
             try {
-                if (value && value.hasOwnProperty('then')) {
+                if (value && value.hasOwnProperty("then")) {
                     msg += ", consider using `promise_test` instead";
                 }
             } catch (err) {}
--- a/resources/testharness.js
+++ b/resources/testharness.js
@@ -543,12 +543,19 @@ policies and contribution forms [3].
         var test_obj = new Test(test_name, properties);
         var value = test_obj.step(func, test_obj, test_obj);
 
-        test_obj.step(function() {
-            assert(value === undefined,
-                   "test", null,
-                   "tests defined via `test` may not return a value",
-                   null);
-        });
+        if (value !== undefined) {
+            var msg = "test named \"" + test_name +
+                "\" inappropriately returned a value";
+
+            try {
+                if (value && value.hasOwnProperty('then')) {
+                    msg += ", consider using `promise_test` instead";
+                }
+            } catch (err) {}
+
+            tests.status.status = tests.status.ERROR;
+            tests.status.message = msg;
+        }
 
         if (test_obj.phase === test_obj.phases.STARTED) {
             test_obj.done();
--- a/resources/testharness.js
+++ b/resources/testharness.js
@@ -541,7 +541,15 @@ policies and contribution forms [3].
         var test_name = name ? name : test_environment.next_default_test_name();
         properties = properties ? properties : {};
         var test_obj = new Test(test_name, properties);
-        test_obj.step(func, test_obj, test_obj);
+        var value = test_obj.step(func, test_obj, test_obj);
+
+        test_obj.step(function() {
+            assert(value === undefined,
+                   "test", null,
+                   "tests defined via `test` may not return a value",
+                   null);
+        });
+
         if (test_obj.phase === test_obj.phases.STARTED) {
             test_obj.done();
         }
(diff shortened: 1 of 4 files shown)


jugglinmike:
Almost forgot: here's the branch I used to conduct the survey https://github.com/bocoup/wpt/commits/suspicious-sync

resources/testharness.js
```diff
@@ -541,7 +541,15 @@ policies and contribution forms [3].
         var test_name = name ? name : test_environment.next_default_test_name();
         properties = properties ? properties : {};
         var test_obj = new Test(test_name, properties);
-        test_obj.step(func, test_obj, test_obj);
+        var value = test_obj.step(func, test_obj, test_obj);
+
+        test_obj.step(function() {
+            assert(value === undefined,
```
jgraham: This is OK, but if I was being picky I might do

```
var msg = "Test object returned a value " + value";
if value.hasOwnProperty("then") {
  msg += " Consider writing promise_test instead";
}
test_obj.set_status(test_obj.ERROR, msg);
test_obj.phase = test_obj.phases.HAS_RESULT;
test_obj.done();
```

gsnedders: +1 for @jgraham's suggestion.

jugglinmike:
Thanks for the review! @jgraham's example referenced a non-existent property, `test_obj.ERROR` instead of `test_obj.FAIL`. I can guess why they made that mistake: this condition really is a harness error, not a test failure. By communicating it in those terms, the problem will be harder to miss if it crops up in a test that is expected to fail.

I've updated the patch to report via a harness error. In order to limit the number of test files, I'm using a different approach for managing the tests. I've also implemented the hint that @jgraham originally requested and included some extra code to protect from the wacky values that tests might return.

The original commit message is no longer accurate, but I'd be happy to update it once this patch is approved.

jgraham (approved):
The implementation looks fine. The selftest looks a little over complex?

jugglinmike:
The test for this needs to verify that a harness error is reported, so it can't be expressed using testharness.js alone. [When I first automated this test suite](https://github.com/web-platform-tests/wpt/pull/5627), I extended the manual tests with JSON that describes how the harness is expected to behave.  This seemed like the lowest-friction way to get us up and running, even if it did require specialized test execution logic.

Since then, I've recognized a few deficiencies with the approach. For one, the tests are difficult to author because the process involves manually writing JSON into an HTML file, because error reports are difficult to interpret, and because there is some ambiguity with how the unordered subtest results should be encoded in an array. Secondly, there can only be one harness error per file, so when you are writing many tests for harness errors (as in the case here), you are forced to express expectations across many files.

These problems were particularly frustrating in gh-8748, so I experimented with this approach there--see [`resources/test/tests/unit/exceptional-cases.html`](https://github.com/web-platform-tests/wpt/blob/445544dcf920e46d158bfe82b1e45e30360ed278/resources/test/tests/unit/exceptional-cases.html).  I'm continuing that trend with these tests because I feel it's much more technically sound. I'm interested in converting all the existing tests to use this approach, and that should make things a little better still through the introduction of a shared utility function and the removal of the JSON-scraping interpretation logic. If you have specific recommendations for further simplification, I'm all ears!

PR #12975, opened 2018-09-12
[css-contain] Split contain-paint-013 in three smaller tests

mrego:


tabatkins:
Note that inline-blocks are *already* IFCs by definition; there's no way to strip them of it without just reverting them fully to inlines.  Is the 002 test just trying to ensure that `contain:paint` doesn't randomly change some IFCs into other random display types?

frivoal:
right, I guess that test-002 verifies that requiring an inline-box to establish and ifc doesn't do anything funky to it. It's probably not the most useful test, but it doesn't seem wrong.

tabatkins:
You're right that it's not wrong. I'll go ahead and r+ then.

tabatkins (approved):

mrego:
Sorry for the late answer, it's true that the test is not strictly needed.
Somehow it's similar to `contain-paint-independent-formatting-context-003.html` that doesn't change if you remove `contain: paint`.
I remember some weird issues with this combinations that I was fixing in Chrome (I don't remember the exact details now) and that was the origin of `contain-paint-013.html` test.
@TalbotG and @frivoal suggested to split it in three so it was clearer, and that's what I did in this PR.

PR #12980, opened 2018-09-12
Make tools/ci/jobs.py work on Windows

gsnedders:
We called os.path.normcase on the path to be compared, and not the
rules themselves. This fixes this by calling normcase on both the rule
and the path to be compared.

We also tried to do separator normalization before calling normcase,
which itself does slash normalization.

### DIFF
--- a/tools/ci/jobs.py
+++ b/tools/ci/jobs.py
@@ -35,15 +35,21 @@ job_path_map = {
 }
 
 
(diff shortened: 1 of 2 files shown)


Ms2ger (approved):
Can you extract the normcase + separator replacement into a helper function to make sure they stay in sync?

PR #12993, opened 2018-09-13
Update test for serializing shorthand with "initial" values and different important flags

Loirooriol:
The initial test was added in #12729 but some changes were requested in https://bugs.webkit.org/show_bug.cgi?id=188984


mrego (approved):
LGTM.

In the future it'd be nice to recover the assert for cssText too, maybe in two different `test()` functions if needed.

PR #12997, opened 2018-09-13
Unlock the lock in the logging module instead of reloading it

Eijebong:
Since 2ae9668bd0f23edf15d0591aeb90060dbd2565c2, the logging module is
reload in both `start_ws_server` and `start_wss_server`. This was done
to fix failing tests in gecko caused by a locked lock in the logging
module being inherited in new processes. Unfortunately this caused servo
tests to start failing.

This commits unlocks the logging global lock if it's still held instead
of reloading the whole module limiting the side effects and fixing the
tests in servo while keeping the gecko tests green.

The function `_releaseLock` is not public but has been there for a long
time and is still there in python 3.8 so I don't think we should worry
about it disappearing.

### DIFF
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -19,7 +19,6 @@ from collections import defaultdict, OrderedDict
 from multiprocessing import Process, Event
 
 from localpaths import repo_root
-from six.moves import reload_module
 
 from manifest.sourcefile import read_script_metadata, js_meta_re, parse_variants
 from wptserve import server as wptserve, handlers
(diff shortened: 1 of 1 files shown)


Eijebong:
CC @jdm @jgraham 

Link to the servo tests https://github.com/servo/servo/pull/21695 we had 10 try runs without seeing the websockets failures.
Link to the geckotry: https://treeherder.mozilla.org/#/jobs?repo=try&revision=576d26db8b0659ab536db6ada3126f69f2ee218f

jgraham (approved):

PR #12999, opened 2018-09-14
[css-contain] Split contain-layout-008 in three smaller tests

mrego:
Following what we did in #12975 for contain-paint-013,
this patch does the same for contain-layout-008
as they were sharing the reference file that was removed by mistake.


frivoal:
Are you sure you posted the right commits to this pull request? you say it's about contain-layout-008, but all it does it delete contain-layout-003

mrego:
> Are you sure you posted the right commits to this pull request? you say it's about contain-layout-008, but all it does it delete contain-layout-003

Sorry I don't know what I did before, probably it was too early in the morning.
The new patch should be better.

TimothyGu (approved):

TimothyGu:
Merged to fix `./wpt lint` failure on master.

PR #13006, opened 2018-09-14
Fetch/XHR: test parsing of WWW-Authenticate response header

annevk:
Context: https://bugzilla.mozilla.org/show_bug.cgi?id=1491010.

### DIFF
--- a/fetch/api/basic/header-value-combining.any.js
+++ b/fetch/api/basic/header-value-combining.any.js
@@ -1,0 +1,11 @@
+// META: global=window,worker
+
+promise_test(async t => {
+  const response = await fetch("../../../xhr/resources/headers-basic.asis");
+  assert_equals(response.headers.get("foo-test"), "1, 2, 3");
+}, "response.headers.get('foo-test')");
+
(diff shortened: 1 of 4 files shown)


annevk:
Curiously Chrome fails the fetch() test in the same manner as Firefox, but passes the XMLHttpRequest tests.

jobisoft:
Have you run this test already? The tests for XHR will probably fail, because XHR returns them "as is"

1\n2\n3,4

annevk:
Yeah, that's a bug in Firefox. Other browsers don't have that problem.

annevk:
There's no standard that supports the behavior you want and I don't think it'd be a good idea to special case headers or pass through the server data as-is (as proxies and such could also have meddled with it).

jobisoft:
Ok, but then you need to get this in the official MDN documentation of XHR as fast as possible, so that coders like me do not rely on the current behaviour. 

I was actually thinking about switching from fetch() to XHR because of the *current* behaviour.

youennf (approved):

PR #12317, opened 2018-08-06
Add Web Speech API IDL file + test

foolip:
IDL from https://github.com/tidoust/reffy-reports/tree/master/whatwg/idl.

PR #13017, opened 2018-09-15
Use the voiceschanged event to wait for SpeechSynthesisVoice instance

foolip:
In https://github.com/web-platform-tests/wpt/pull/12317 I noticed that
speechSynthesis.getVoices() could return an empty list and is allowed to
do so by spec, but not that there's also an event that's fired when the
voices are updated:
https://w3c.github.io/speech-api/#eventdef-speechsynthesis-voiceschanged

Using this improves coverage somewhat.

### DIFF
--- a/speech-api/idlharness.window.js
+++ b/speech-api/idlharness.window.js
@@ -3,10 +3,27 @@
 
 'use strict';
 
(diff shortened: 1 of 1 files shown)


lukebjerring (approved):
Nice.

PR #13018, opened 2018-09-15
Test new SpeechSynthesisUtterance(null/undefined)

foolip:
Spec: https://w3c.github.io/speech-api/#tts-section

This is targeting a mundane difference in `new
SpeechSynthesisUtterance(undefined)` depending on whether the IDL definition of
the construtor uses a single constructor with an optional argument, or two
constructor overloads with zero and one argument respectively.

This is not equivalent because
https://heycam.github.io/webidl/#dfn-overload-resolution-algorithm first
considers the number of arguments, so given one argument the
`Constructor(DOMString text)` overload is used. With optional arguments,
however, trailing undefined arguments are ignored so that `foo(undefined)` will
be indistinguishable from `foo()`.

In other words, `new SpeechSynthesisUtterance(undefined)` should result in a
`text` of "undefined" because the spec uses two contructor overloads, but
Blink/WebKit have a single constructor. Edge and Firefox pass the added test, so
align with them rather than changing the spec. (It doesn't really matter.)

### DIFF
--- a/speech-api/SpeechSynthesisUtterance-basics.https.html
+++ b/speech-api/SpeechSynthesisUtterance-basics.https.html
@@ -31,6 +31,16 @@ test(function() {
 }, 'new SpeechSynthesisUtterance("hello") text and defaults');
 
 test(function() {
+  const utt = new SpeechSynthesisUtterance(null);
+  assert_equals(utt.text, 'null');
+}, 'new SpeechSynthesisUtterance(null)');
+
+test(function() {
+  const utt = new SpeechSynthesisUtterance(undefined);
+  assert_equals(utt.text, 'undefined');
+}, 'new SpeechSynthesisUtterance(undefined)');
+
+test(function() {
   const utt = new SpeechSynthesisUtterance();
   utt.text = 'word';
   assert_equals(utt.text, 'word');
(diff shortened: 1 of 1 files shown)


saschanaz:
Personally an IDL attribute that returns `undefined` is a bit weird, and AFAIK the Web IDL spec cannot represent that behavior (https://github.com/heycam/webidl/issues/60).

And it seems the attribute is [not marked as nullable](https://w3c.github.io/speech-api/#utterance-attributes), maybe a spec bug?

foolip:
It doesn't return undefined, it's the string "undefined" (and "null"). That's because the argument to the constructor is of type DOMString and is thus converted to a string.

saschanaz:
Oh, that's right. Thank you for making things clear.

But can we choose the shorter one (the optional argument one) instead of current overloads, when browsers should anyway change?

foolip:
Yep, it doesn't really matter which form we pick, but I thought that since Edge and Gecko have already matched the spec, it's more fair to leave that be. I need to make a change around the same place in the Chromium code anyway, which is how I noticed this.

foolip:
Although, there is one additional consideration. If we'd ever want to add a  dictionary and have that as an optional argument for the constructor, then we'd probably want to change to a single `Constructor(optional (DOMString or SpeechSynthesisUtteranceInit) init)` overload, in which case the case tested in this PR would change.

Thoughts?

saschanaz:
So one more +1 for optional argument 😄 to ease future changes.

foolip:
@saschanaz can you review this PR as well? Now updated to match https://github.com/w3c/speech-api/pull/48.

saschanaz (approved):

PR #13020, opened 2018-09-16
HTML: Test rendering of fieldset with a tall legend

zcorpan:
Fixes https://github.com/whatwg/html/issues/4014

### DIFF
--- a/html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/legend-tall-ref.html
+++ b/html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/legend-tall-ref.html
@@ -1,0 +1,20 @@
+<!DOCTYPE html>
+<title>Reference for tall legend</title>
+<style>
+body { margin: 0; }
+.legend { position: absolute; height: 160px; margin-left: 20px; inline-size: fit-content; background: lime }
+.fieldset {
+  position: absolute;
+  z-index: -1;
+  margin-top: calc((/* half legend height */ 160px / 2) - (/* half top border */ 20px / 2));
+  background: green;
+  height: 40px;
+  left: 0;
+  right: 0;
+}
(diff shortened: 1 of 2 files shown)


zcorpan:
I don't understand why "hello" is 176 px down, and not 180px, but it seems to match browsers...

html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/legend-tall-ref.html
```diff
@@ -0,0 +1,20 @@
...
+.fieldset {
+  position: absolute;
+  z-index: -1;
+  margin-top: calc((/* half legend height */ 160px / 2) - (/* half top border */ 20px / 2));
+  background: green;
+  height: 40px;
+  left: 0;
+  right: 0;
+}
+.hello { margin-top: 176px; margin-left: 20px; }
```
mstensho: Assuming that 1em == 16px here? But maybe that's ok?

mstensho (dismissed):

html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/legend-tall-ref.html
mstensho: I misunderstood here. It should be 180px, not 176px. I really think we need to figure out why.

zcorpan:
@MatsPalmgren do you know why the content is pushed down by this particular amount?

mstensho:
In the ref, the top margin of .hello collapses with the bottom margin of the p. So 16px goes away there. So my original comment about assuming that 1em == 16px was right after all. :)

zcorpan:
Aah, yes. It should be 160px (the height of the legend). Thanks.

mstensho (approved):

PR #13021, opened 2018-09-16
[css-animations] Test inheritance, initial values

ewilligers:
Test that animations and transitions properties are not inherited.

Verify the initial values of the properties.


ewilligers:
See also spec typo fix https://github.com/w3c/csswg-drafts/pull/3119

tabatkins (approved):

PR #13023, opened 2018-09-16
[css-transforms] Transforms initial values

ewilligers:
Test the initial values and non-inheritance of
CSS Transforms and CSS Motion Path properties.

https://drafts.csswg.org/css-transforms/#property-index
https://drafts.csswg.org/css-transforms-2/#property-index
https://drafts.fxtf.org/motion-1/#property-index

Browsers currently report 'flat' as the initial value of
transform-style, contrary to spec.

### DIFF
--- a/css/css-transforms/inheritance.html
+++ b/css/css-transforms/inheritance.html
@@ -1,0 +1,41 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Inheritance of CSS Transforms properties</title>
+<link rel="help" href="https://drafts.csswg.org/css-transforms/#property-index">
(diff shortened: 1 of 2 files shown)


css/css-transforms/inheritance.html
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Inheritance of CSS Transfoms properties</title>
```
birtles:
```suggestion
<title>Inheritance of CSS Transforms properties</title>
```

ewilligers: Thanks, fixed.

ewilligers:
@dirkschulze  please review

dirkschulze (approved):

PR #13027, opened 2018-09-17
Added line-breaking-013 and reference test

faceless2:
Added test for line breaking behaviour discussed in https://github.com/w3c/csswg-drafts/issues/3085#issuecomment-421714901


css/css-text/line-breaking/line-breaking-013.xht
```diff
@@ -0,0 +1,90 @@
...
+<html xmlns="http://www.w3.org/1999/xhtml">
+<head>
+  <title>CSS Test: atomic inlines are treated as ideographic characters for line-breaking</title>
+  <meta charset="utf-8" />
+  <link rel="author" title="Mike Bremford" href="http://bfo.com/" />
+  <link rel="help" href="https://www.w3.org/TR/css-text-3/#line-break-details" />
+  <meta name="assert" content="Atomic inlines are treated as ideographic characters for line-breaking" />
+  <link rel="match" href="reference/line-breaking-013-ref.xht" />
+  <style>
+    p {
```
frivoal: Styling the `<p>` doesn't seem essential for the test. I suggest dropping it to simplify, and to avoid misleading people into thinking that it might matter somehow.

frivoal (changes requested):
Overall, this looks good, but:
* please split each block (1/2/3/4) into separate tests files. They are testing related but distinct things, and standalone tests make that easier to work with. Adjust the assert meta accordingly (for instance by combining the existing assert meta with the relevant comment for each block). 
* please further split each sub-case of each block (image vs inline block) into separate test files. 
* is the third subcase of each block (with 国) meant to be a test, or just a handy reference? If it is a test (which seems reasonable), it should have its own test file and test assertion.
* Please also address the comments made inline
* using html instead of xhtml is preferred. It allows skipping a number of closing tags (`</p>`), elements (`<head>`, `<body>`, `<html>`), or attributes (`xmlns="http://www.w3.org/1999/xhtml"`), which makes the whole file shorter and easier to read and review. Using xhtml is not forbidden, so I'll still accept the test if it stays that way, but html would be nicer.

PR #13032, opened 2018-09-17
Improve the wasm idlharness test.

Ms2ger:
CC @foolip

### DIFF
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -1289,6 +1289,15 @@ IdlInterface.prototype.get_interface_object = function() {
     return this.get_interface_object_owner()[this.name];
 };
 
+IdlInterface.prototype.get_qualified_name = function() {
+    // https://heycam.github.io/webidl/#qualified-name
+    var legacyNamespace = this.get_legacy_namespace();
+    if (legacyNamespace) {
+        return legacyNamespace + "." + this.name;
+    }
+    return this.name;
+};
+
 IdlInterface.prototype.has_to_json_regular_operation = function() {
     return this.members.some(function(m) {
         return m.is_to_json_regular_operation();
@@ -1752,20 +1761,20 @@ IdlInterface.prototype.test_self = function()
         }
 
         // "The class string of an interface prototype object is the
-        // concatenation of the interface’s identifier and the string
+        // concatenation of the interface’s qualified identifier and the string
         // “Prototype”."
 
         // Skip these tests for now due to a specification issue about
         // prototype name.
         // https://www.w3.org/Bugs/Public/show_bug.cgi?id=28244
 
-        // assert_class_string(this.get_interface_object().prototype, this.name + "Prototype",
+        // assert_class_string(this.get_interface_object().prototype, this.get_qualified_name() + "Prototype",
         //                     "class string of " + this.name + ".prototype");
 
         // String() should end up calling {}.toString if nothing defines a
         // stringifier.
         if (!this.has_stringifier()) {
-            // assert_equals(String(this.get_interface_object().prototype), "[object " + this.name + "Prototype]",
+            // assert_equals(String(this.get_interface_object().prototype), "[object " + this.get_qualified_name() + "Prototype]",
             //         "String(" + this.name + ".prototype)");
         }
     }.bind(this), this.name + " interface: existence and properties of interface prototype object");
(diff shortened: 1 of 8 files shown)


foolip (commented):
Alright, I've done it, I've reviewed!

Nice new helper functions that make it more obvious in many places what's going on than with `self[this.name]`, just some naming quibbles really.

Also, with a change this large, it seems like something will break. Can you run it through Gecko's CQ to see if any tests are broken by it?

Ms2ger:
Fixed your comments, checked in Firefox CI, and fixed the bug I introduced in `test_member_iterable()`. Thanks for making me do that :)

foolip:
@Ms2ger, I can't review the delta between last time and now because you force pushed the branch (I think) but I'll pay special attention in `test_member_iterable()` to see if I can spot the change.

resources/test/tests/unit/IdlInterface/get_interface_object.html
```diff
@@ -0,0 +1,22 @@
...
+test(function() {
+    window.A = {};
+    var i = interfaceFrom('interface A { };');
+    assert_equals(i.get_interface_object(), window.A);
+}, 'Interface does not have LegacyNamespace.');
+
+test(function() {
+    window.Foo = { A: {} };
+    var i = interfaceFrom('[LegacyNamespace=Foo] interface A { }; namespace Foo { };');
+    assert_equals(i.get_interface_object(), window.Foo.A);
```
foolip: Nice succinct tests!

foolip (approved):
Nah, couldn't spot the difference in `test_member_iterable()` from memory, but I'm sure it's good.

I thought I'd try running this through Chromium's CQ as well, which I can do if I get https://chromium-review.googlesource.com/c/chromium/src/+/1262716 landed first quickly. If it takes me beyond Friday, just land this :)

Ms2ger:
(I forgot to add the `bind(this)` calls initially.)

foolip:
Actually, I don't need to wait for https://chromium-review.googlesource.com/c/chromium/src/+/1262716, I confused myself. I'll try it now.

foolip:
https://chromium-review.googlesource.com/c/chromium/src/+/1264163 (if it's all green then no need to wait for me to report back)

resources/idlharness.js
```diff
@@ -1514,10 +1550,10 @@ IdlInterface.prototype.test_self = function()
             // "If I was not declared with a [Constructor] extended attribute,
             // then throw a TypeError."
             assert_throws(new TypeError(), function() {
-                self[this.name]();
+                this.get_interface_object()();
             }.bind(this), "interface object didn't throw TypeError when called as a function");
             assert_throws(new TypeError(), function() {
-                new self[this.name]();
+                new this.get_interface_object()();
```
foolip: There's a bug here. `DetectedBarcode` doesn't have a constructor in shape-detection-api.idl, but it does in Blink. A test used to fail for that but no longer does:
```diff
--- /b/s/w/ioYUXUDo/layout-test-results/external/wpt/shape-detection/idlharness.any-expected.txt
+++ /b/s/w/ioYUXUDo/layout-test-results/external/wpt/shape-detection/idlharness.any-actual.txt
@@ -7,7 +7,7 @@
 PASS FaceDetector interface: existence and properties of interface prototype object's "constructor" property
 PASS FaceDetector interface: existence and properties of interface prototype object's @@unscopables property
 PASS FaceDetector interface: operation detect(ImageBitmapSource)
-FAIL DetectedFace interface: existence and properties of interface object assert_throws: interface object didn't throw TypeError when called as a constructor function "function () { [native code] }" did not throw
+PASS DetectedFace interface: existence and properties of interface object
 PASS DetectedFace interface object length
 PASS DetectedFace interface object name
 PASS DetectedFace interface: existence and properties of interface prototype object
@@ -23,7 +23,7 @@
 PASS BarcodeDetector interface: existence and properties of interface prototype object's @@unscopables property
 FAIL BarcodeDetector interface: attribute supportedFormats assert_true: The prototype object must have a property "supportedFormats" expected true got false
 PASS BarcodeDetector interface: operation detect(ImageBitmapSource)
-FAIL DetectedBarcode interface: existence and properties of interface object assert_throws: interface object didn't throw TypeError when called as a constructor function "function () { [native code] }" did not throw
+PASS DetectedBarcode interface: existence and properties of interface object
 PASS DetectedBarcode interface object length
 PASS DetectedBarcode interface object name
 PASS DetectedBarcode interface: existence and properties of interface prototype object
```

The same has happened with the `RTCIceTransport` interface.

Unfortunately neither of those are shipped yet so you can't try against a vanilla Chrome. But could you check if perhaps the change in the condition has just made this code unreachable in this case? That would explain the incorrect pass.

foolip (commented):
https://chromium-review.googlesource.com/c/chromium/src/+/1264163 has results now and it looks like there's a real regression (manifesting as fail->pass) and the message for assert_interface_object_exists could do with some tweaking.

resources/idlharness.js
Ms2ger: The issue was operator precedence; this fixed it:
```diff
-                new this.get_interface_object()();
+                new (this.get_interface_object())();
```

but I used a slightly different fix in the end.

(I notice that this test isn't particularly useful; it only catches a constructor without required arguments. Oh well.)

foolip:
Post-merge review, the fixup commits in https://github.com/web-platform-tests/wpt/compare/master...Ms2ger:wasm-idl-review LGTM.

The changes are being imported into Chromium now and no unexpected changes AFAICT:
https://chromium-review.googlesource.com/c/chromium/src/+/1278537

Note that one failures message went from "self.SVGPoint should be the same value as self.DOMPoint" to "self.SVGPoint should be the same value as the interface object", and the former was more clear/accurate. @Ms2ger, would you care to try to fix that?

PR #13033, opened 2018-09-17
Correctly handle ScriptTimeoutException in the WebDriver executor

jgraham:


Hexcles (approved):
r+

I wonder why this wasn't noticed before.

PR #13034, opened 2018-09-17
Delete goal-parameter.htm

domenic:
This test was committed without the corresponding spec change being agreed upon.


zcorpan:
Is there a pointer to the suggested spec change?

domenic:
https://github.com/whatwg/html/pull/3205

zcorpan (approved):

PR #13043, opened 2018-09-18
[service-workers] Refactor tautological tests

jugglinmike:
Previously, these tests created Promise values which would be fulfilled
even when expectations were violated. This would cause the test harness
to interpret them as "passing" regardless of implementation status.

Remove the use of `Promise.prototype.catch` to ensure that promise
rejections which occur during test execution lead to test failure.

### DIFF
--- a/service-workers/service-worker/worker-interception.https.html
+++ b/service-workers/service-worker/worker-interception.https.html
@@ -69,16 +69,9 @@ promise_test(function(t) {
           return wait_for_state(t, r.installing, 'activated');
         })
       .then(function() {
-          return new Promise(function(resolve, reject) {
-              var w = new Worker(worker_url);
-              w.onmessage = function() {
-                reject(new Error('intercepted cors response to a same-origin ' +
-                                 'mode worker load should fail'));
-              }
-
-              // The DOM error is expected.
-              w.onerror = resolve;
-            });
+          var w = new Worker(worker_url);
+          var watcher = new EventWatcher(t, w, ['message', 'error']);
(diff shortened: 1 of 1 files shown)


mkruisselbrink:
Doesn't this still have the problem that some errors (for example in service_worker_unregister_and_register or wait_for_state) are silently swallowed? It seems like the better fix would be to wrap whatever operation is actually expected to fail in a promise_rejects() call, rather than adding explicit assert_unreached/promise rejection handlers like the test is doing today.

jugglinmike:
This test actually creates the Promise that is expected to reject. I think a
simpler solution would be to reserve the "fulfillment" path for expected
behavior, even when that involves a DOM error. I pushed up a commit along these
lines, taking care to document the suspicious-looking code as intentional. What
do you think?

mkruisselbrink:
> This test actually creates the Promise that is expected to reject. I think a
> simpler solution would be to reserve the "fulfillment" path for expected
> behavior, even when that involves a DOM error. I pushed up a commit along these
> lines, taking care to document the suspicious-looking code as intentional. What
> do you think?

Ah, now I see what the test is trying to do. Maybe somehow using `EventWatcher` could make the code nicer? I.e. something like
```js
var w = new Worker(worker_url);
var watcher = new EventWatcher(t, w, ['message', 'error']);
return watcher.wait_for('error');
```

jugglinmike:
Works for me! That's implemented by the latest commit

mkruisselbrink (approved):
Thanks, lgtm!

PR #13044, opened 2018-09-18
[css-align] Property inheritance, initial values

ewilligers:
CSS Box Alignment properties are not inherited.

The initial values match the spec
https://drafts.csswg.org/css-align-3/#property-index

### DIFF
--- a/css/css-align/inheritance.html
+++ b/css/css-align/inheritance.html
@@ -1,0 +1,33 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Inheritance of CSS Box Alignment properties</title>
+<link rel="help" href="https://drafts.csswg.org/css-align-3/#property-index">
+<meta name="assert" content="Properties inherit or not according to the spec.">
+<meta name="assert" content="Properties have initial values according to the spec.">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/css/support/inheritance-testcommon.js"></script>
+</head>
+<body>
+<div id="container">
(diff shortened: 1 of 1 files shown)


ewilligers:
Note that inheritance with justify-items 'legacy' is already tested in
default-alignment/justify-items-legacy-001.html

css/css-align/inheritance.html
```diff
@@ -0,0 +1,33 @@
...
+</style>
+<script>
+assert_not_inherited('align-content', 'normal', 'last baseline');
+assert_not_inherited('align-items', 'normal', 'last baseline');
+assert_not_inherited('align-self', 'auto', 'last baseline');
+assert_not_inherited('column-gap', 'normal', '10px');
+assert_not_inherited('justify-content', 'normal', 'space-evenly');
+assert_not_inherited('justify-items', 'legacy center', 'last baseline');
+assert_not_inherited('justify-self', 'auto', 'last baseline');
+assert_not_inherited('row-gap', 'normal', '10px');
```
javifernandez: Shouldn't we test here column-gap as well ?

javifernandez (commented):
The change looks good, but before approve it, I'd like to know more about my comment on the inheritance-testcommon.js file here:

css/css-align/inheritance.html
ewilligers: I sorted the properties alphabetically, matching the index. Thus column-gap appears above.

In future there may be multiple levels of the spec, and I think having all the properties listed alphabetically works better than having all the properties in one level, in spec listing order, followed by all the properties in another level, etc.

javifernandez (approved):

PR #13045, opened 2018-09-18
[css-grid] Property initial values and inheritance

ewilligers:
Initial values of CSS Box Alignment properties match spec
https://drafts.csswg.org/css-grid/#property-index

None of the properties are inherited.

### DIFF
--- a/css/css-grid/inheritance.html
+++ b/css/css-grid/inheritance.html
@@ -1,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Inheritance of CSS Grid Layout properties</title>
+<link rel="help" href="https://drafts.csswg.org/css-grid/#property-index">
+<meta name="assert" content="Properties inherit or not according to the spec.">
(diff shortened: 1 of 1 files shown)


css/css-grid/inheritance.html
```diff
@@ -0,0 +1,30 @@
...
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/css/support/inheritance-testcommon.js"></script>
+</head>
+<body>
+<div id="container">
+<div id="target"></div>
+</div>
+<script>
+assert_not_inherited('grid-auto-columns', 'auto', '10px');
```
javifernandez: Shouldn't we test grid-auto-rows here as well ?

javifernandez (commented):
The change looks good, but before approve it, I'd like to know more about my comment on the inheritance-testcommon.js file here:

ewilligers:
@javifernandez Your comment is correct.
Fixing in https://github.com/web-platform-tests/wpt/pull/13058

javifernandez (approved):

PR #13049, opened 2018-09-18
Add several tests to check scroll behavior smooth/instant

fred-wang:


fred-wang:
cc @alijuma @rwlbuis @smfr @wangxianzhu

css/cssom-view/scroll-behavior-005.html
```diff
@@ -0,0 +1,141 @@
...
+<script>
+  ["scroll", "scrollTo", "scrollBy", "scrollIntoView"].forEach(function(scrollFunction) {
+    promise_test(() => {
+      return new Promise(function(resolve) {
+        resetScroll(overflowNode);
+        var finalLeft = 500;
+        var finalTop = 250;
+        // Note: exact smooth scrolling behavior is not specified by CSSOM View. This test relies on the minimal assumption that the element scrolls less than 1/4 of the total distance between two frames.
+        var maxDeltaLeft = finalLeft / 4;
+        var maxDeltaTop = finalTop / 4;
```
alijuma: I wonder if these constraints will cause flakiness when run as layout tests, particularly on debug builds where you might occasionally have a long gap between frames. You might need to relax these to 1/2 the distance or even 3/4 of the distance.

css/cssom-view/scroll-behavior-006.html
```diff
@@ -0,0 +1,69 @@
...
+        </div>
+      <div class="gradient" style="width: 100px; height: 500px;"></div>
+      </div>
+      <div class="gradient" style="width: 100px; height: 500px;"></div>
+    </div>
+  </div>
+</div>
+<script>
+  promise_test(() => {
+    // CSSOM seems to suggest revealing happens from innermost to outermost but some browsers do the opposite.
```
alijuma: Yeah, I think the issue with innermost to outermost is that the innermost scroller might be fully invisible to start, so you if you animate that scroller first, the user sees none of that animation. If you do outer-to-inner, each step is visible. So the test approach below seems the right way to go until there's more consistency in implementations.

alijuma (commented):
These tests look great overall! I know this is a really tricky spec to write good non-flaky tests and the approach taken here seems solid to me.

One request: can we give the tests more descriptive names? So scroll-behavior-default.html instead of scroll-behavior-001.html, scroll-behavior-subframe.html instead of scroll-behavior-003.html, etc.? It just makes it easier to understand a list of passing/failing tests without having to go read each one.

css/cssom-view/scroll-behavior-006.html
fred-wang: Actually, the spec, Gecko and Chromium have different approaches. I opened https://github.com/w3c/csswg-drafts/issues/3127 for now.

I can probably make the test slightly stricter by observing scroll positions and verifying each coordinate is non-decreasing, but probably not much more (in particular not checking smooth VS instant behaviors).

css/cssom-view/scroll-behavior-005.html
fred-wang: Yes, I was wondering that too. I guess I can change to 3/4 or even just remove that constraint. BTW the current uploaded patch for WebKit fails the test but passes for 1/2, I have a local version to make the scroll smoother.

fred-wang:
@alijuma I've addressed the review comments. I improved the scrollIntoView tests and added some tests for Window.scroll* and Element.scroll() too.

alijuma (approved):
Thanks!

PR #13058, opened 2018-09-18
[css-text-decor] Inheritance and initial values

ewilligers:
Test that CSS Text Decoration properties inherit or not
according to spec. Test that they have the expected
initial values.

https://drafts.csswg.org/css-text-decor-3/#property-index

Fix asserts in inheritance-testcommon.js

### DIFF
--- a/css/css-text-decor/inheritance.html
+++ b/css/css-text-decor/inheritance.html
@@ -1,0 +1,33 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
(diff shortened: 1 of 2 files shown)


ewilligers:
I hadn't previously detected the `assert_inherited` bug as `css/css-inline/inheritance.html` was previously the only client. It has three inherited properties: `initial-letters-align` and `initial-letters-wrap` are not supported by the browsers I tested, and `dominant-baseline` isn't supported correctly.

javifernandez (approved):

PR #13067, opened 2018-09-19
[css-position] Inheritance and initial values

ewilligers:
Test that the CSS Positioned Layout properties do not inherit.
Test their initial values against the spec.
https://drafts.csswg.org/css-position/#property-index

### DIFF
--- a/css/css-position/inheritance.html
+++ b/css/css-position/inheritance.html
@@ -1,0 +1,35 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Inheritance of CSS Positioned Layout properties</title>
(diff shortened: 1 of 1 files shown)


FremyCompany:
Didn't we have an agreement to rename `offset` properties to `outset`? Other than that, this looks good, I'm just trying to make sure this spec is still up to date and reflecting the correct information.

FremyCompany (approved):
Approving, but it would be good to check the status of the spec.

ewilligers:
The spec was [updated](https://github.com/w3c/csswg-drafts/commit/6516a86d202a92f1f83a9f45d26ac6766f686df3#diff-9d7ce4c29507451bef88e3db07585b98) to use inset-*

Tests updated to match.

PR #13070, opened 2018-09-19
Update list-marker-inside-whitespace-001.xml

ivanha09:
whitespace change


Ms2ger:
The whitespace change is unnecessary, and the color change is unexplained. Closing.

PR #13076, opened 2018-09-19
Add a headless mode command line option

jgraham:
For browsers which support headless mode starting wptrunner with
--headless will run the tests in headless mode; otherwise it will be
ignored.

--no-headless is avaiable to explicitly disable the mode when
frontends change the defaults.

For firefox switch the default for wpt run to be headless; for Chrome
changing the default broke some smoketests, so don't switch until that
is resolved. In all cases continue to run in a normal windowed mode in CI.


tools/wptrunner/wptrunner/wptcommandline.py
```diff
@@ -205,6 +205,10 @@ def create_parser(product_choices=None):
                               help="Allow the wptrunner to install fonts on your system")
     config_group.add_argument("--font-dir", action="store", type=abs_path, dest="font_dir",
                               help="Path to local font installation directory", default=None)
+    config_group.add_argument("--headless", action="store_true",
+                              help="Run browser in headless mode", default=None)
+    config_group.add_argument("--no-headless", action="store_false", dest="headless",
+                              help="Don't run browser in headless mode")
```
Hexcles: Can we clarify what the default behaviour is in the help text?

jgraham: Weirdly, not easily. Because`wpt run` sets its own per-browser default. So we'd have to override the help text in that case.

jgraham:
Fixes https://github.com/web-platform-tests/wpt/pull/13076 and related to https://github.com/web-platform-tests/results-collection/issues/603

gsnedders:
> For firefox switch the default for wpt run to be headless; for Chrome
changing the default broke some smoketests, so don't switch until that
is resolved. In all cases continue to run in a normal windowed mode in CI.

Is there a bug filed for the Chrome issue?

jgraham:
It's much easier to file a bug if we merge this because it gives simpler STR.

Hexcles (approved):
LGTM.

Enabling by default on Firefox but not on Chrome for now seems fine (assuming you've run the tests and there's no difference on Firefox w/ and w/o headless mode).

Issue #10588, opened 2018-04-23
encoding/utf-32.html seems wrong


annevk:
In particular not defaulting to UTF-8 for XML doesn't seem like something that standard supports.

cc @inexorabletash

inexorabletash:
Without paging all the algorithms in... I believe that the content is not being interpreted as either XML or HTML - it ends up treated as plain text. (Which is non- or under- specified?)

* Big Endian BOM: `00 00 FE FF` - no UTF-8 or -16 BOM, doesn't sniff, so plain text, windows-1252
* Little Endian BOM: `FF FE 00 00` - UTF-16LE BOM, so UTF-16, doesn't sniff?, so plain text
* Big Endian no BOM: `00 00 00 3C` - no UTF-8 or -16 BOM, doesn't sniff, so plain text, windows-1252
* Little Endian no BOM: `3C 00 00 00` - no UTF-8 or -16 BOM, doesn't sniff, so plain text, windows-1252

Content types served with the tests probably make this more complicated, though - again, would need to go through the algorithms. Unsure if we'd even make it as far as trying to parse as UTF-8 XML, or what we do if that path is taken and we fail super early.

inexorabletash:
i.e. agreed that the expectations encoded in the file are non-obvious, and more investigation is needed

annevk:
(This issue was supposed to be closed from 315b16be16977448d33165f0dc538f00dac0ecd5 but that did not work, at least not as quick as I wanted.)

PR #13082, opened 2018-09-19
Encoding: Verify UTF-32 tests aren't parsing as (X)HTML

inexorabletash:
One approach for fixing https://github.com/web-platform-tests/wpt/issues/10588

This modifies the .html/.xml samples to include a data-parsed attribute on the HTML element, and probe for it. Since the files should have failed to parse as HTML/XHTML the resulting iframe documents should not contain the attribute.


inexorabletash:
@annevk - please take a look? 

Not urgent, just picking off an easy item with my name on it.

I tried adding .txt samples, but in Chrome at least navigating an iframe to a .txt file triggers a download instead. The sniffing exposed in the other cases here may be implementation specific and/or just way out of scope for this directory. *shrug*

annevk:
1. XML defaults to UTF-8, so I think that would make sense to expect here.
2. XML would result in a parse error. I'm not sure if you should be able to access the resulting document, I don't think that's defined.

For HTML this is probably okay though.

hsivonen:
> XML defaults to UTF-8, so I think that would make sense to expect here.

Indeed. I'm curious where the notion of expecting windows-1252 for XML came from.

inexorabletash:
In Chrome, at least, the files with `.xml` suffix are not parsing as XML/XHTML or HTML; presumably due to sniffing (I haven't traced through) they are being treated as plain text, hence the _"such content will be sniffed as plain text and interpreted in the default encoding (windows-1252)"_ comment in the file. So "expecting windows-1252 for XML" is not what the test asserts. (Suggestions for ways to make that clearer are appreciated!)

I haven't gone through [MIMESNIFF](https://mimesniff.spec.whatwg.org) to see if the sniffing is standardized. And I think implementations differ on the default encoding for text/plain.

annevk:
Per https://mimesniff.spec.whatwg.org/#determining-the-computed-mime-type-of-a-resource you basically never sniff resources with an XML MIME type.

hsivonen:
Ping? Any chance we could get the XML tests fixed?

inexorabletash:
Firefox and Chrome have differing behavior.

If the containing doc has `<meta charset=utf-8>` then both agree (everything is UTF-8 or UTF-16LE). If not, or if `<meta charset=windows-1252>` then behavior differs, and I still haven't dug into why yet.

inexorabletash:
Okay, these pass in Firefox, and I'd call the Chrome behavior a bug (or let others have that argument). Take a peek. 

Also feel free to push changes/take over/land if you want.

annevk (approved):
Thanks, looks great now! I pushed two minor nits and will land this with those applied.

inexorabletash:
Thanks! (and nit fixes lgtm too)

PR #13084, opened 2018-09-19
WebKit export of https://bugs.webkit.org/show_bug.cgi?id=189516

woaiwyhty:
WebKit export from bug: [MediaStream-MediaElement-srcObject.https.html expects video duration for a MediaStream to be Infinity and is currently NaN](https://bugs.webkit.org/show_bug.cgi?id=189516)


youennf (approved):

youennf:
Test might be flaky according bot since it now runs more tests while it was previously failing early.
This is a progression so I will merge the request.
@gsnedders, let me know whether this procedure is ok.

gsnedders:
@youennf It's not ideal, but there's downsides to all the various options here:

 * Do nothing, have the test be flaky fail if the test does run to completion anywhere
 * Fix the test to actually test the correct behaviour, exposing the flakiness to more browsers
 * Disable the test everywhere due to flakiness

Ideally we'd fix the flakiness, but if we don't know where that is, it's easier said than done. :)

PR #13085, opened 2018-09-19
Replace deprecated pytest marker 'skip_if' with 'skipif'.

burg:
This no longer works in pytest >= 3.7, so migrate to the supported name.


andreastt:
This has already been fixed by @whimboo in https://bugzilla.mozilla.org/show_bug.cgi?id=1491975 and https://github.com/web-platform-tests/wpt/pull/13074.

whimboo:
That's correct. No need for this PR anymore.

PR #13087, opened 2018-09-20
[css-ruby] Parsing and inheritance tests

ewilligers:
Test that ruby-align and ruby-position are inherited and support
the full grammar.

https://drafts.csswg.org/css-ruby-1/#property-index


ewilligers:
I skipped ruby-merge as it currently lacks browser support.

upsuper (approved):

PR #13095, opened 2018-09-20
[css-ui] Inheritance and initial value

ewilligers:
Test that CSS Basic User Interface properties inherit.
https://drafts.csswg.org/css-ui/#property-index

### DIFF
--- a/css/css-ui/inheritance.html
+++ b/css/css-ui/inheritance.html
@@ -1,0 +1,47 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Inheritance of CSS Basic User Interface properties</title>
+<link rel="help" href="https://drafts.csswg.org/css-ui/#property-index">
+<meta name="assert" content="Properties inherit or not according to the spec.">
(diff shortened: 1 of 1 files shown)


foolip:

svgeesus (commented):
LGTM

ewilligers:
@svgeesus can you please merge this. GitHub tells me "Review required".

foolip (approved):
Glossy review LGTM

frivoal:
Hmmm. Sorry for the slow review. I think we should actually revert / fix this:
* For color properties, getComputedStyle returns the used value, not the computed value. They are different, but your asserts are written assuming the computed value. Hence why you get `rgb(…)` when you were expecting `auto`.
* For the `invert` color of the outline-color property, you should check if the `invert` keyword is supported at all. If it is, it should be the initial value, but if it is not, having `currentColor` be the initial value is spec compliant, and should not cause the test to fail.

frivoal:
Just noticed you had filed a follow up in #13748 will continue discussing there.

PR #13105, opened 2018-09-20
[filter-effects] Inheritance and initial values

ewilligers:
Test that Filter Effects properties inherit.
https://drafts.fxtf.org/filter-effects/#property-index

### DIFF
--- a/css/filter-effects/inheritance.html
+++ b/css/filter-effects/inheritance.html
@@ -1,0 +1,25 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Inheritance of Filter Effects properties</title>
+<link rel="help" href="https://drafts.fxtf.org/filter-effects/#property-index">
+<meta name="assert" content="Properties inherit according to the spec.">
+<meta name="assert" content="Properties have initial values according to the spec.">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/css/support/inheritance-testcommon.js"></script>
+</head>
+<body>
+<div id="container">
+  <div id="target"></div>
+</div>
+<script>
+assert_inherited('color-interpolation-filters', 'auto', 'srgb'); // SVG 1.1 had linearrgb as initial value.
+assert_not_inherited('filter', 'none', 'blur(10px)');
+assert_not_inherited('flood-color', 'rgb(0, 0, 0)', 'rgb(42, 53, 64)');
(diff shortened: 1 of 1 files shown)


ewilligers:
All browsers fail on the initial value of `color-interpolation-filters` - see https://github.com/w3c/csswg-drafts/issues/3131

foolip:

css/filter-effects/inheritance.html
```diff
@@ -0,0 +1,25 @@
...
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/css/support/inheritance-testcommon.js"></script>
+</head>
+<body>
+<div id="container">
+  <div id="target"></div>
+</div>
+<script>
+assert_inherited('color-interpolation-filters', 'auto', 'srgb'); // SVG 1.1 had linearrgb as initial value.
```
svgeesus: I think https://github.com/w3c/csswg-drafts/issues/3131 shows that this was an inadvertent change, and should be reverted. This line in the test would then need to be altered, and browsers would pass again :)

svgeesus (dismissed):
Once https://github.com/w3c/csswg-drafts/issues/3131 is resolved, one change needed on this test. Apart from that, LGTM.

ewilligers:
Test updated. Spec change in https://github.com/w3c/fxtf-drafts/pull/312

css/filter-effects/inheritance.html
```diff
@@ -0,0 +1,25 @@
...
+<script src="/css/support/inheritance-testcommon.js"></script>
+</head>
+<body>
+<div id="container">
+  <div id="target"></div>
+</div>
+<script>
+assert_inherited('color-interpolation-filters', 'linearrgb', 'srgb');
+assert_not_inherited('filter', 'none', 'blur(10px)');
+assert_not_inherited('flood-color', 'rgb(0, 0, 0)', 'rgb(42, 53, 64)');
```
dirkschulze: Wha do the 2 functions test?

ewilligers: See comments in https://github.com/web-platform-tests/wpt/blob/master/css/support/inheritance-testcommon.js

The first function tests that `color-interpolation-filters` is an inherited property, with `linearrgb` the initial value, and `srgb` being another accepted value (a value other than the initial value is used to test inheritance).

The second function is used to test that `filter` and `flood-color` do not inherit, and have the initial values `none` and `rgb(0, 0, 0)` respectively.

Each function creates 2 tests (initial value and inheritance).

dirkschulze (approved):

PR #13109, opened 2018-09-20
[css-overflow] Inheritance and initial values

ewilligers:
Test that CSS Overflow properties inherit.
Test that they have the initial values given in the spec.
https://drafts.csswg.org/css-overflow/#property-index

### DIFF
--- a/css/css-overflow/inheritance.html
+++ b/css/css-overflow/inheritance.html
@@ -1,0 +1,28 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Inheritance of CSS Overflow properties</title>
+<link rel="help" href="https://drafts.csswg.org/css-overflow/#property-index">
+<meta name="assert" content="Properties inherit or not according to the spec.">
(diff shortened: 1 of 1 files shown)


foolip:

ewilligers:
@dbaron  Please review

frivoal (changes requested):
The test looks correct based on the spec, but the spec seems to have an issue. how can overflow-block and overflow-inline have `auto` as the initial value, when overflow-x and overflow-y have `visible` as the initial value? That made no sense (and I had introduced that mistake), so I've fixed in in https://github.com/w3c/csswg-drafts/commit/60fc9bf33b3d8e8c5f7c8e37b9ed9aaf194f0286. If you can update the test to match that, then it will be good to merge.

ewilligers:
@frivoal I have updated the test.

frivoal (approved):

frivoal:
Looks good. Ready to merge, but Taskcluster seems stuck.
@gsnedders @foolip @Ms2ger

Issue #13112, opened 2018-09-20
All new PRs are blocked on failing Chrome stability job


foolip:
Lots of recently opened PRs are blocked on failing Travis checks.

This is the earliest I can find, from https://travis-ci.org/web-platform-tests/wpt/jobs/430913594:
```
Traceback (most recent call last):
  File "./wpt", line 5, in <module>
    wpt.main()
  File "/home/travis/build/web-platform-tests/wpt/tools/wpt/wpt.py", line 129, in main
    rv = script(*args, **kwargs)
  File "/home/travis/build/web-platform-tests/wpt/tools/ci/check_stability.py", line 276, in run
    wpt_kwargs = setup_wptrunner(venv, **wpt_kwargs)
  File "/home/travis/build/web-platform-tests/wpt/tools/wpt/run.py", line 458, in setup_wptrunner
    check_environ(kwargs["product"])
  File "/home/travis/build/web-platform-tests/wpt/tools/wpt/run.py", line 140, in check_environ
    raise WptrunError(message)
tools.wpt.run.WptrunError: Missing hosts file configuration. Run
./wpt make-hosts-file | sudo tee -a /etc/hosts
```

Affected PRs:
https://github.com/web-platform-tests/wpt/pull/13092
https://github.com/web-platform-tests/wpt/pull/13093
https://github.com/web-platform-tests/wpt/pull/13094
https://github.com/web-platform-tests/wpt/pull/13095
https://github.com/web-platform-tests/wpt/pull/13103
https://github.com/web-platform-tests/wpt/pull/13104
https://github.com/web-platform-tests/wpt/pull/13105
https://github.com/web-platform-tests/wpt/pull/13106
https://github.com/web-platform-tests/wpt/pull/13109
https://github.com/web-platform-tests/wpt/pull/13110

Failing status checks for other reasons:
https://github.com/web-platform-tests/wpt/pull/13107
https://github.com/web-platform-tests/wpt/pull/13108

Not affected because "No tests changed" in log:
https://github.com/web-platform-tests/wpt/pull/13096
https://github.com/web-platform-tests/wpt/pull/13097
https://github.com/web-platform-tests/wpt/pull/13098
https://github.com/web-platform-tests/wpt/pull/13099
https://github.com/web-platform-tests/wpt/pull/13100
https://github.com/web-platform-tests/wpt/pull/13101
https://github.com/web-platform-tests/wpt/pull/13102

foolip:
[...]

PR #13113, opened 2018-09-20
Handle case where mozlog is not installed.

jgraham:
In case that mozlog is present we need to ensure any lock it holds is
released in subprocesses. In case it's not present we don't need to do
anything, so don't fail if we can't import it.


foolip:
Would this fix https://github.com/web-platform-tests/wpt/issues/13112?

foolip (approved):

jgraham:
Yes, I believe so.

foolip:
Whether it does or not, I reviewed and the changes seem sensible, so I will merge as soon as Travis passes.

PR #13116, opened 2018-09-20
[css-lists] Inheritance and initial values

ewilligers:
Test that the properties below have the initial values
given in the spec, and inherit or not according to spec.

- counter-increment
- counter-reset
- list-style-image
- list-style-position
- list-style-type

https://drafts.csswg.org/css-lists/#property-index


ewilligers:
I skipped properties that are not currently supported by browsers.

tabatkins (approved):

PR #13119, opened 2018-09-20
Run all unit tests for resources/.

Ms2ger:


resources/test/conftest.py
```diff
@@ -107,7 +107,8 @@ def __init__(self, filename, test_type, parent):
         elif self.type == 'unit' and self.expected:
             raise ValueError('Unit tests must not specify expected report data')
 
-        super(HTMLItem, self).__init__(name, parent)
+        pytest.Item.__init__(self, name, parent)
```
jgraham: Add a comment about what this is doing, please.

jgraham (commented):
We could possibly just disable the pytest cache?

jgraham (approved):

Ms2ger:
@jugglinmike: heads up: I needed to remove a subtest of yours here; the test never actually ran in CI, but you might want to take a look anyway.

Issue #5289, opened 2017-03-31
duplicated docs


gsnedders:
Originally posted as https://github.com/gsnedders/wpt-docs/issues/1 by @gsnedders on 21 Dec 2016, 18:29 UTC:

> docs/configuration.md should really somehow be magically extracted from README.md

gsnedders:
Originally posted as https://github.com/gsnedders/wpt-docs/issues/1#issuecomment-271297883 by @gsnedders on 09 Jan 2017, 14:30 UTC:

> I think realistically this isn't going to happen

foolip:
@gsnedders, what do you think should be the 1 place where /etc/hosts other installation instructions are? Now they're forked and presumably slightly diverged.

gsnedders:
@foolip Last we spoke about it, @jgraham was opposed to removing them from the README and I was opposed to removing them from the documentation.

gsnedders:
Note that https://github.com/jekyll/jekyll/issues/5728 or something similar could allow us to include the README.md (in its entirety), which is probably not ideal (given we don't want, e.g., the header).

zcorpan:
I think the README has too much information, and it would be easier to follow if things moved to docs.

PR #13125, opened 2018-09-20
[docs] Reorganize information on running the tests

jugglinmike:
Previously, information was organized as follows:

- Section: Introduction
  - Page: Introduction
    - Introductory material
    - Instructions for running the tests manually
    - Abbreviated instructions for `wpt run`
- Section: Running the Tests
  - Page: Running the Tests
    - Instructions for `wpt run`
    - Instructions for the web runner
    - Notes on writing a custom runner

Dispersing the information about "test running" across multiple sections
tended to dilute the meaning of each section. It also made the
introduction lengthier and possibly more intimidating (especially for
readers who have no intention of running the tests manually).

Restructure the information as follows:

- Section: Introduction
  - Page: Introduction
    - Introductory material
- Section: Running the Tests
  - Page: Running the Tests
    - Instructions for running the tests manually
    - Reference to "Running the Tests in Automation"
  - Page: Running the Tests in Automation
    - Instructions for `wpt run`
    - Instructions for the web runner
    - Notes on writing a custom runner


Hexcles:
IMHO, I think we should put "running tests automatically" before "manually". I believe conventionally the earlier in docs the more supported an approach is.

Hexcles:
There's a lot of duplication between these docs and the top-level `README.md`. I'm sure this has been discussed somewhere (perhaps multiple times). What's the plan here? Shall we just trim down the top-level `README.md` and refer to docs whenever possible?

jugglinmike:
> Shall we just trim down the top-level `README.md` and refer to docs whenever possible?

That's what I had in mind

gsnedders:
> There's a lot of duplication between these docs and the top-level `README.md`. I'm sure this has been discussed somewhere (perhaps multiple times). What's the plan here? Shall we just trim down the top-level `README.md` and refer to docs whenever possible?

#5289

docs/_running-tests/automation.md
```diff
@@ -0,0 +1,83 @@
...
+chosen browser, and extract the test results. For example to run the
+`dom/historical.html` tests in a local copy of Chrome:
+
+    ./wpt run chrome dom/historical.html
+
+Or to run in a specified copy of Firefox:
+
+    ./wpt run --binary ~/local/firefox/firefox firefox dom/historical.html
+
+`./wpt run --help` lists the supported products.
```
Ms2ger: This is also in the next paragraph.

docs/_running-tests/index.md
```diff
@@ -2,79 +2,77 @@
 layout: page
 title: Running Tests
```
Ms2ger: This isn't really about running tests anymore, but only about setting up to run the server.

docs/_running-tests/automation.md
```diff
@@ -0,0 +1,83 @@
+---
+layout: page
+title: Running Tests in Automation
```
Ms2ger: If we want to suggest `wpt run` as the easiest approach for newcomers, I think "in Automation" gives the wrong impression. It sounds more like information about setting up CI systems.

I feel it would be better to put the first two paragraphs here should become `_running-tests/index.md`, followed by links to two subpages: one for the "Writing Your Own Runner" section, and one for setting up `./wpt serve`.

Does `./wpt run` need the hosts file setup as well? I suspect it isn't necessary for Gecko/Servo, but maybe it is for Chromium/WebKit? In that case, we need to find a way to weasel that into the front page too.

Ms2ger (commented):
Thanks! I have some Opinions(TM); see below :)

docs/_running-tests/automation.md
gsnedders: I think this is the fundamental question that needs to be addressed here: how do we expect tests to be run in the normal individual user case?

Hexcles: Re the "fundamental question", my vote goes that `wpt run` should be the default, recommended way to run tests, given how much setup we do in `wpt run` (not just dependencies, but also browser configurations).

gsnedders: `wpt run` is less convenient when you want to attach a JS debugger, for example, though, which isn't that unusual of a case when writing tests.

docs/_running-tests/automation.md
jugglinmike: Yup. Doesn't hurt to fix that in this patch

jugglinmike:
Thanks for the feedback, folks!

Since there are two divergent use cases and we lack data (or even subjective opinion) on which is more common, I think the overhead of one more level in the hierarchy is justified.

I first tried to incorporate this with two pages, "From the Command Line" (`wpt run`) and "From the Browser" (w3c-test.org and `wpt server`). The idea was to split according to the interface used to trigger test execution.

It didn't take long to realize that this distinction wouldn't be relevant for most readers. It's also hard to organize relevant information along those lines (i.e. details on setting up the hosts file would be necessary on both pages, for Chrome, anyway).

That's why I'm now proposing: "From the web" (w3c-test.org) and "From a local system" (`wpt run` and `wpt serve`).

- Section: Introduction
  - Page: Introduction
    - Introductory material
- Section: Running the Tests
  - Page: Running the Tests
    - Reference to "Running the Tests from the Web"
    - Reference to "Running the Tests from a Local System"
    - Reference to "Writing Your Own Runner"
  - Page: Running the Tests from the Web
    - Link to w3c-test.org
    - Instructions for the web runner
  - Page: Running the Tests from a Local System
    - Instructions for `wpt serve` (with reference to "Running the Tests from
      the Web")
    - Instructions for `wpt run`
  - Page: Writing Your Own Runner"
    - Notes on writing a custom runner

This invalidates the original commit message, but I'll be happy to amend the commit when/if we get some agreement on this patch.

Hexcles (approved):
Thanks, @jugglinmike ! This LGTM.

I don't really have a strong opinion over which one should be promoted in the docs. And as you pointed out, the arguments about which is the most common use case lack data to support. 

Regardless, this is a vast improvement over the status quo IMHO.

docs/_running-tests/automation.md
jgraham: That sounds like a possible feature request rather than a fundamental limitation; we could support js debuggers like we support system-lvel debuggers (i.e. we add a --js-debug option that causes the browser to start with the debugger open and disables timeouts).

docs/_running-tests/index.md
```diff
@@ -2,79 +2,17 @@
...
 
-The simplest way to run tests is to use the `wpt run` command from the
-root of the repository. This will automatically load the tests in the
-chosen browser, and extract the test results. For example to run the
-`dom/historical.html` tests in a local copy of Chrome:
+Contributors who are interested in modifying and creating tests should refer to
+[Running Tests from the Local System][from-local-system].
 
-    ./wpt run chrome dom/historical.html
+Advanced use cases may call for a customized method of executing the tests.
```
jgraham: I think this should say

> Users are strongly reccommended to use the test runners provided with web-platform-tests. These are tested in CI and kept up to date with changes in the test formats. However if this is impossible for some reason, people implementing custom test runners may find these [notes on writing a runner][custom-runner] helpful.

Ms2ger (approved):

jugglinmike:
@jgraham Today, the docs endorse the development of custom runners without stipulation. Introducing discouraging language like that is a substantive change, one that approaches policy-level decision making. Could we discuss it separately?

sideshowbarker:
@jgraham This seems to be waiting on feedback from you. See https://github.com/web-platform-tests/wpt/pull/13125#issuecomment-430025721

jugglinmike:
Commit 2b3e49da896ad67c8e337304a07a8c77f93b4c3e introduced a trivial merge conflict, so I've rebased this patch and force-pushed the result. [The original version is available here.](https://github.com/web-platform-tests/wpt/compare/master...bocoup:docs-running-2-orig)

@Hexcles @Ms2ger would you mind merging this? The change requested by @jgraham is orthogonal to this patch, and I'd like to avoid further conflicts.

PR #13142, opened 2018-09-21
IdlHarness: Improve the formatting of generic types.

Ms2ger:


foolip:
@Ms2ger can you rebase that and trigger som full runs so that it's easy to see what effect this will have?

Ms2ger:
@foolip rebased, but I don't know how to trigger full runs.

foolip:
@Ms2ger see https://web-platform-tests.org/running-tests/from-ci.html

Chrome Dev, Firefox Nightly and Safari TP should be enough to catch anything interesting in the changes.

foolip:
Oh wait, the affected tests already includen the right tests, so no need to trigger full runs.

foolip:
https://wpt.fyi/results/?diff&filter=ADC&q=is%3Adifferent&run_id=349900001&run_id=340780001 is a good view to start from. Is it intentional that "<undefined>" should be in so many names, or is that a bug?

Ms2ger:
No, that's a bug, probably caused by https://github.com/w3c/webidl2.js/commit/f4292a7a21008b80c1a2ee17acd7d825d1fec5dc. I attempted a fix.

CC @saschanaz

Ms2ger:
Finally got around to this, and improved the output a bit more. re-r? @foolip

foolip (approved):
Yay! In https://wpt.fyi/results/?diff&filter=ADC&q=status%3Amissing&run_id=407600001&run_id=432290001 one can spot plenty of improved test names.

foolip:
wpt-firefox-nightly-stability timed out because there were too many tests to run. If it hadn't timed out it probably would have failed because of https://github.com/web-platform-tests/wpt/issues/16175. Admin merging.

PR #13143, opened 2018-09-21
Add more referrer-policy css-intergration tests

tungnh28:


tungmangtdh3:
Close as it fails in travis

PR #13144, opened 2018-09-21
HTML: test stylesheet loading text/css requirement

annevk:


annevk:
I wrote this for https://bugzilla.mozilla.org/show_bug.cgi?id=562377. It seems this could use some cleanup in HTML as well around the requirements.

annevk:
(I would have thought we test this elsewhere too, but I couldn't find anything quickly.)

zcorpan (approved):

annevk:
@zcorpan thanks, do you happen to know if we have quirks tests for this? If not, I'll add more tests Monday or so.

html/semantics/document-metadata/the-link-element/link-style-error-01.html
```diff
@@ -44,5 +45,44 @@
   elt.rel = "stylesheet";
   elt.href = "../../../../../common/css-red.txt";
   document.getElementsByTagName("head")[0].appendChild(elt);
-})
+});
```
domfarolino: nit: Would you be up for adding semicolons above as well to match here, and maybe use t.step_func_done() in the above preexisting tests?

domfarolino (approved):

html/semantics/document-metadata/the-link-element/link-style-error-01.html
annevk: I try not to make stylistic changes at the same time as that messes with blame too much in my opinion. And then doing it as a separate commit often seems like too much effort... I might not be making the right tradeoffs though.

annevk:
https://github.com/whatwg/html/issues/3457 is probably the best HTML Standard issue to track improvements at.

annevk:
I don't think we should block landing this on those happening though.

gsnedders:
"limitedd" (in the commit message) isn't how you spell limited. :)

Also: 🎉 for having tests for this.

annevk:
* https://bugs.chromium.org/p/chromium/issues/detail?id=888508
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/19085222/
* https://bugs.webkit.org/show_bug.cgi?id=189913

html/semantics/document-metadata/the-link-element/link-style-error-quirks.html
```diff
@@ -0,0 +1,48 @@
+<!DOCTYPE HTML PUBLIC "-//Sun Microsystems Corp.//DTD HotJava Strict HTML//" "">
```
zcorpan: Can you replace this with a comment saying "quirks mode"?

html/semantics/document-metadata/the-link-element/link-style-error-limited-quirks.html
```diff
@@ -0,0 +1,48 @@
...
+<title>link: error events in limited quirks mode</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src=/common/get-host-info.sub.js></script>
+<div id="log"></div>
+<script>
+[
+ ["same-origin", "resources/css.py"],
+ ["cross-origin", get_host_info().HTTP_REMOTE_ORIGIN + "/html/semantics/document-metadata/the-link-element/resources/css.py"]
+].forEach(originType => {
```
zcorpan: Please deduplicate the code by moving it into a separate JS file.

zcorpan (changes requested):

html/semantics/document-metadata/the-link-element/link-style-error-quirks.html
annevk: That's a lot less fun.

html/semantics/document-metadata/the-link-element/link-style-error-limited-quirks.html
annevk: I guess that's reasonable. Do you think it's fine to make it depend on document.compatMode or do you think it should be something you call externally?

The other thing I was thinking of also testing is `@import`, but how that affects load/error events isn't really defined yet.

zcorpan: Using `document.compatMode` should be fine.

zcorpan (approved):

zcorpan (approved):

annevk:
Thanks @zcorpan. I might wait a bit for @bzbarsky to take a look if he wants, since I'm not entirely sure about the `@import` stuff, but if that doesn't come to pass I need to at least file bugs.

bzbarsky:
The test is loading urls like "resources/css.py...." from a sheet that's loaded from a data: URL.  Those relative URIs are resolved relative to the data: URL, which fails, so the load is not performed at all.  Whether this should trigger the error event is an interesting question, but at the moment the test is not testing what it thinks it's testing.

PR #13147, opened 2018-09-21
[taskcluster] Run stable daily and beta weekly

Hexcles:
This PR simply switches the two configurations around. I'll let @foolip to explain the rationale.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -9,8 +9,8 @@ tasks:
           $flatten:
             $match: {
               event.ref == "refs/heads/master": [{name: firefox, channel: nightly}, {name: chrome, channel: dev}],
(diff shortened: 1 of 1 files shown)


foolip:
The most mundane explanation is that we currently run stable builds daily via https://github.com/web-platform-tests/results-collection and for parity should keep doing that.

The value of beta runs is in comparing them to stable runs to detect regressions not caught by any other CI before they reach stable. For that use case the we have weekly alignment before and after this change.

Also, between stable and beta, stable is the only thing we can plausibly get daily runs for across all browsers, as Edge and Safari don't have anything we'd call beta as distinct from experimental.

foolip (approved):

jugglinmike (approved):

PR #13154, opened 2018-09-21
[wptrunner] Fix support for `--binary-arg` in Fx

jugglinmike:
In order to automate Firefox, wptrunner requires the browser's internal
module "Marionette" to be enabled. This is a requirement regardless of
any custom configuration specified by the user, including additional
command-line flags via the `--binary-arg` argument.

Previously, this project would omit the `-marionette` argument in the
presence of any user-specified arguments. Because this interferes with
normal operation (and because the source code is formatted in a way that
conflicts with the order of operations), this is likely an unintentional
behavior.

Modify the code which derives the command-line arguments to always
include the `-marionette` argument.

### DIFF
--- a/tools/wpt/tests/test_wpt.py
+++ b/tools/wpt/tests/test_wpt.py
@@ -189,6 +189,30 @@ def test_run_firefox(manifest_dir):
 
 
 @pytest.mark.slow
+@pytest.mark.remote_network
+@pytest.mark.xfail(sys.platform == "win32",
+                   reason="Tests currently don't work on Windows for path reasons")
+def test_run_firefox_binary_args(manifest_dir):
+    if is_port_8000_in_use():
+        pytest.skip("port 8000 already in use")
+
+    if sys.platform == "darwin":
+        fx_path = os.path.join(wpt.localpaths.repo_root, "_venv", "browsers", "nightly", "Firefox Nightly.app")
+    else:
+        fx_path = os.path.join(wpt.localpaths.repo_root, "_venv", "browsers", "nightly", "firefox")
+    if os.path.exists(fx_path):
+        shutil.rmtree(fx_path)
+    with pytest.raises(SystemExit) as excinfo:
+        wpt.main(argv=["run", "--no-pause", "--install-browser", "--yes",
+                       "--metadata", manifest_dir,
+                       "--binary-arg=-headless",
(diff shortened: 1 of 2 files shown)


tools/wpt/tests/test_wpt.py
```diff
@@ -188,6 +188,30 @@ def test_run_firefox(manifest_dir):
         del os.environ["MOZ_HEADLESS"]
 
 
+@pytest.mark.slow
+@pytest.mark.remote_network
+@pytest.mark.xfail(sys.platform == "win32",
+                   reason="Tests currently don't work on Windows for path reasons")
+def test_run_firefox_binary_args(manifest_dir):
```
jgraham: I think this test is pretty heavyweight for what it does. Can we fold it into another test, or just add an assert to the code?

tools/wptrunner/wptrunner/browsers/firefox.py
```diff
@@ -251,9 +251,11 @@ def start(self, group_metadata=None, **kwargs):
         if self.ca_certificate_path is not None:
             self.setup_ssl()
 
+        args = self.binary_args[:] if self.binary_args else []
+        args += [cmd_arg("marionette"), "about:blank"]
```
jgraham: I think we possibly want to check for duplicate arguments. It doesn't make much difference, but is a little cleaner. 
```
If not any(item.strip("-/") == "marionette" for item in args):
  args += "--marionette"
```

Just appending `about:blank` is probably fine, although we might worry about args not starting with `--`, `-` or `/`, which would be interpreted as URLs and could in some future hypothetically start opening multiple tabs.

jugglinmike: Since we have a utility that's designated for creation of CLI arguments, I'd rather incorporate that:

    marionette_flag = cmd_arg("marionette")
    if not any(item == marionette_flag for item in args):
        args.append(marionette_flag)

...but I don't think this discussion should delay the bug fix. I'm also not sure how to relate the new behavior to the fix in the commit message. What do you think about deferring this for a follow-up patch?

jgraham: That doesn't work because multiple variants are supported. The Firefox command line parser is something quite… special.

I'm OK on this not delaying the fix though.

jgraham (approved):

PR #13158, opened 2018-09-21
Also ignore ERROR in slow test check

Hexcles:
Tweak the docstring accordingly to explain why.

@jgraham this is the bug you pointed out to me earlier. PTAL, especially to see if my docstring makes sense.

### DIFF
--- a/tools/wptrunner/wptrunner/stability.py
+++ b/tools/wptrunner/wptrunner/stability.py
@@ -115,8 +115,9 @@ def find_slow_status(test):
     """Check if a single test almost times out.
 
     We are interested in tests that almost time out (i.e. likely to be flaky).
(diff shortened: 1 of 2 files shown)


jgraham (approved):

jgraham:
The other option I had in mind was to check if the recorded time was greater than the expected timeout, and not fail the check if it was. But let's see if this works well first.

Hexcles:
@jgraham I thought about that, too, but decided to go with this way because of the reasons explained in the docstring. Even if the run time of a crashy test fluctuates around the timeout, as long as the test consistently crashes, it won't be flaky as the result is always CRASH. Also, this is a more conservative approach in general so there'll less likely be false positives.

foolip:
This was follow up to https://github.com/web-platform-tests/wpt/pull/11570.

PR #13159, opened 2018-09-21
Fix the path of mplus-1p-regular.woff in all tests

Hexcles:
Following #9718, the font is now in /fonts/ instead of /fonts/CSSTest/.
Globally fix the path using sed.


mrego (approved):
Thanks for fixing this.

PR #13163, opened 2018-09-21
[service-workers] Return value in cleanup

jugglinmike:
Previously, many tests un-registered service workers using the
testharness.js API `add_cleanup` without returning the promise which
tracked the operation. While this ensured the operation was started at
the completion of the test, it did not guarantee that the operation
would be completed before the test was considered "complete."

In automation scenarios where many tests are executed in rapid
succession, this enabled a race condition: the navigation could
interrupt the un-registration process. In order to account for the
possibility of registrations that persisted from previous test failures,
the tests included "setup" code to defensively un-register such workers.

asynchronous "clean up" operations [1]. Use that API to ensure tests are
not considered "complete" until after service worker un-registration is
done.

---

Chromium on `master`:

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 96 checks (8 tests, 88 subtests)
    Expected results: 85
    Unexpected results: 11
      subtest: 11 (11 fail)

Chromium with patch applied:

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 96 checks (8 tests, 88 subtests)
    Expected results: 85
    Unexpected results: 11
      subtest: 11 (11 fail)

Firefox on `master`:

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 96 checks (8 tests, 88 subtests)
    Expected results: 88
    Unexpected results: 8
      subtest: 8 (8 fail)

Firefox with patch applied:

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 96 checks (8 tests, 88 subtests)
    Expected results: 88
    Unexpected results: 8
      subtest: 8 (8 fail)

### DIFF
--- a/service-workers/service-worker/fetch-request-xhr-sync.https.html
+++ b/service-workers/service-worker/fetch-request-xhr-sync.https.html
@@ -14,7 +14,7 @@ promise_test(function(t) {
     return service_worker_unregister_and_register(t, url, scope)
       .then(function(registration) {
           t.add_cleanup(function() {
-              registration.unregister();
+              return registration.unregister();
             });
 
           return wait_for_state(t, registration.installing, 'activated');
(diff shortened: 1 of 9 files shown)


mkruisselbrink (approved):
Thanks for seeing this through and making these tests better!

jugglinmike:
My pleasure :)

PR #13164, opened 2018-09-21
[service-workers] Use asynchronous cleanup

jugglinmike:
Previously, many tests un-registered service workers only after all
assertions had been satisfied. This meant that failing tests would not
un-register workers. In order to account for workers that persisted from
previous test failures, the tests included "setup" code to defensively
un-register such workers.

asynchronous "clean up" operations [1]. Use that API to schedule service
worker un-registration so that it occurs regardless of the result of
each test.

---

Chromium on `master`:

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 145 checks (32 tests, 113 subtests)
    Expected results: 120
    Unexpected results: 25
      test: 2 (2 timeout)
      subtest: 23 (15 fail, 6 notrun, 2 timeout)

Chromium with patch applied:

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 145 checks (32 tests, 113 subtests)
    Expected results: 120
    Unexpected results: 25
      test: 2 (2 timeout)
      subtest: 23 (15 fail, 6 notrun, 2 timeout)

Firefox on `master`:

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 145 checks (32 tests, 113 subtests)
    Expected results: 134
    Unexpected results: 11
      test: 2 (2 timeout)
      subtest: 9 (6 fail, 1 notrun, 2 timeout)

Firefox with patch applied:

     web-platform-test
     ~~~~~~~~~~~~~~~~~
     Ran 145 checks (32 tests, 113 subtests)
     Expected results: 134
     Unexpected results: 11
       test: 2 (2 timeout)
       subtest: 9 (6 fail, 1 notrun, 2 timeout)

### DIFF
--- a/service-workers/service-worker/claim-fetch.https.html
+++ b/service-workers/service-worker/claim-fetch.https.html
@@ -32,9 +32,13 @@ promise_test(function(t) {
 
     // Register a service worker.
     .then(() => service_worker_unregister_and_register(t, script, scope))
-    .then(r => worker = r.installing)
-    .then(() => wait_for_state(t, worker, 'activated'))
+    .then(r => {
+        t.add_cleanup(() => service_worker_unregister(t, scope));
 
+        worker = r.installing;
+
+        return wait_for_state(t, worker, 'activated');
+      })
     // Let the service worker claim the iframe.
     .then(() => {
       var channel = new MessageChannel();
@@ -62,7 +66,6 @@ promise_test(function(t) {
 
     // Cleanup this testcase.
     .then(() => frame.remove())
-    .then(() => service_worker_unregister_and_done(t, scope));
 }, 'fetch() should be intercepted after the client is claimed.')
 
 </script>
--- a/service-workers/service-worker/claim-shared-worker-fetch.https.html
+++ b/service-workers/service-worker/claim-shared-worker-fetch.https.html
@@ -30,9 +30,13 @@ promise_test(function(t) {
                                          'fetch() should not be intercepted.'))
     // Register a service worker.
     .then(() => service_worker_unregister_and_register(t, script, scope))
-    .then(r => worker = r.installing)
-    .then(() => wait_for_state(t, worker, 'activated'))
+    .then(r => {
+        t.add_cleanup(() => service_worker_unregister(t, scope));
 
+        worker = r.installing;
+
+        return wait_for_state(t, worker, 'activated')
+      })
     // Let the service worker claim the iframe and the shared worker.
     .then(() => {
       var channel = new MessageChannel();
@@ -60,8 +64,7 @@ promise_test(function(t) {
                         'fetch() in the shared worker should be intercepted.'))
 
     // Cleanup this testcase.
-    .then(() => frame.remove())
-    .then(() => service_worker_unregister_and_done(t, scope));
+    .then(() => frame.remove());
 }, 'fetch() in SharedWorker should be intercepted after the client is claimed.')
 
 </script>
--- a/service-workers/service-worker/clients-matchall-order.https.html
+++ b/service-workers/service-worker/clients-matchall-order.https.html
@@ -123,6 +123,8 @@ function matchAllOrderTest(t, opts) {
   let frameResultList;
   let extraWindowResult;
   return service_worker_unregister_and_register(t, script, opts.scope).then(swr => {
+    t.add_cleanup(() => service_worker_unregister(t, opts.scope));
+
     worker = swr.installing;
     return wait_for_state(t, worker, 'activated');
   }).then(_ => {
@@ -143,8 +145,6 @@ function matchAllOrderTest(t, opts) {
   }).then(_ => {
     frameResultList.forEach(result => result.top.remove());
     extraWindowResult.top.remove();
-  }).then(_ => {
-    return service_worker_unregister_and_done(t, opts.scope);
   }).catch(e => {
     if (frameResultList) {
       frameResultList.forEach(result => result.top.remove());
--- a/service-workers/service-worker/performance-timeline.https.html
+++ b/service-workers/service-worker/performance-timeline.https.html
@@ -17,7 +17,11 @@ promise_test(t => {
   let slowURL = url + '&slow';
   let frame;
   return service_worker_unregister_and_register(t, script, scope)
-    .then(reg => wait_for_state(t, reg.installing, 'activated'))
+    .then(reg => {
+        t.add_cleanup(() => service_worker_unregister(t, scope));
+
+        return wait_for_state(t, reg.installing, 'activated');
+      })
     .then(_ => with_iframe(scope))
     .then(f => {
       frame = f;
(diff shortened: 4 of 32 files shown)


service-workers/service-worker/clients-matchall-order.https.html
```diff
@@ -142,8 +144,8 @@
   }).then(_ => {
     frameResultList.forEach(result => result.top.remove());
     extraWindowResult.top.remove();
-  }).then(_ => {
-    return service_worker_unregister_and_done(t, opts.scope);
+
+    t.done();
```
mkruisselbrink: is there actually a need to call t.done()? It always seems suspicious to me if a promise_test calls done(), since the promise resolving itself should have that same effect?

service-workers/service-worker/claim-fetch.https.html
```diff
@@ -32,9 +32,13 @@
 
     // Register a service worker.
     .then(() => service_worker_unregister_and_register(t, script, scope))
-    .then(r => worker = r.installing)
-    .then(() => wait_for_state(t, worker, 'activated'))
+    .then((r) => {
```
mkruisselbrink: nit: not sure why you're adding () around r? it seems more consistent with the rest of the file to not add those.

service-workers/service-worker/claim-shared-worker-fetch.https.html
```diff
@@ -30,9 +30,13 @@
                                          'fetch() should not be intercepted.'))
     // Register a service worker.
     .then(() => service_worker_unregister_and_register(t, script, scope))
-    .then(r => worker = r.installing)
-    .then(() => wait_for_state(t, worker, 'activated'))
+    .then((r) => {
```
mkruisselbrink: nit: here too, no () for consistency with rest of file

service-workers/service-worker/performance-timeline.https.html
```diff
@@ -17,7 +17,11 @@
   let slowURL = url + '&slow';
   let frame;
   return service_worker_unregister_and_register(t, script, scope)
-    .then(reg => wait_for_state(t, reg.installing, 'activated'))
+    .then((reg) => {
```
mkruisselbrink: nit: here too no ()

mkruisselbrink (approved):
I think this looks good, modulo some nits and the thing I raised on IRC where it appears that using service_worker_unregister() from a cleanup function will result in any failures during that unregister to be silently ignored (as rather than returning a rejected promise service_worker_unregister() runs some code inside a t.step() to handle errors). Not sure how bad it is if those are ignored though, so it might be okay to leave it like this until somebody has time for more extensive cleanups of this code?

mkruisselbrink:
(the test that is flaky in chrome that is causing travis to fail is https://crbug.com/831509)

service-workers/service-worker/clients-matchall-order.https.html
jugglinmike: You're right: that's superfluous.

service-workers/service-worker/claim-fetch.https.html
jugglinmike: Yup, my personal style leaked through while re-writing these callbacks. Should be fixed, now.

service-workers/service-worker/claim-shared-worker-fetch.https.html
jugglinmike: Done

service-workers/service-worker/performance-timeline.https.html
jugglinmike: Done

jugglinmike:
Thanks for the review, @mkruisselbrink. Regarding cleanup failures: I wouldn't mind extending this patch to trigger harness errors. To me, it's more a question of how much change we should tolerate in one patch.

We could define a new function that does everything `service_worker_unregister` does up until the `t.step` invocation. We should probably keep `service_worker_unregister` as is, but that makes this new function hard to name. `service_worker_unregister_simple` is the best I have right now. Pretty dopey, right?

We could overload `service_worker_unregister`:

```diff
 function service_worker_unregister(test, scope) {
   var absoluteScope = (new URL(scope, window.location).href);
+  var fail = test === null ?
+    null : unreached_rejection(test, 'unregister should not fail')
   return navigator.serviceWorker.getRegistration(scope)
     .then(function(registration) {
         if (registration && registration.scope === absoluteScope)
           return registration.unregister();
       })
-    .catch(unreached_rejection(test, 'unregister should not fail'));
+    .catch(fail);
 }
```

That's an awkward API, but intentionally so--we wouldn't want folks accidentally opting in to this behavior.

My favorite option is more drastic. I've had my eyes on removing `service_worker_unregister_and_register` because as far as I can tell, it's obviated by proper cleanup. I've held off on that for these patches because they're large enough as it is. If we wanted to start that here, we could define a replacement helper that would consolidate registration and cleanup:

```javascript
function service_worker_hygienic_register(test, url, scope) {
  if (!scope)
    return Promise.reject(new Error('tests must define a scope'));

  var options = { scope: scope };
  return navigator.serviceWorker.register(url, options)
    .then(function(registration) {
      test.add_cleanup(function() {
          return registration.unregister();
        });

      return registration;
    });
}
```

mkruisselbrink:
I think I'd prefer not to have too many unrelated changes in one PR, so probably not the change to `service_worker_hygienic_register` (as an aside, even if proper cleanup should alleviate the need for an unregister before register, it might still be a good idea to keep that unregister anyway, just in case... At least personally I'm a fan of tests of course cleaning up everything they do, but still also being somewhat resistant to other tests not doing so where it doesn't complicate stuff too much).

I'd be fine with just landing this as is, or something like that overload your suggesting/a separate method (`unregister_service_worker`? or `unregister_service_worker_by_scope`?)

jugglinmike:
@mattto Do you have a preference between the options @mkruisselbrink has selected?

mattto:
Not totally sure I've followed everything in the thread, but I'd say land this as-is and then add service_worker_hygienic_register (just call it service_worker_register?) in another patch. I'd also prefer to bundle unregister at the beginning of service_worker_register just to make good tests resilient to bad tests.

BTW we can probably remove that "tests must define a scope" thing. I added that back when everything was an async_test. You had to give each test a unique scope so the tests didn't stomp on each other, so that warning helped ensure you added a scope.

jugglinmike:
@mattto reminder that [you can merge things](https://github.com/web-platform-tests/wpt/pull/13165#issuecomment-430914326) :)

Also, the instability that TravisCI has reported is unrelated--see gh-14083

mfalken (approved):
I'm very sorry for the delay. I think I finally have a system that filters the "review requested" mails to a folder that I'll notice.

Unfortunately I can't merge on the web UI due to the CI error, so I'm looking into doing this by the command line.

jugglinmike:
You heard right, but this branch predated [the change which allowed instability](https://github.com/web-platform-tests/wpt/pull/14096). I've rebased to incorporate that change, so this pull request is no longer blocked by unrelated instability (though the instability is still present).

The prior version of this branch is available on Bocoup's fork of wpt:

https://github.com/web-platform-tests/wpt/compare/master...bocoup:service-workers-async-cleanup-use-orig

PR #13165, opened 2018-09-21
[service-workers] Refactor to use async cleanup

jugglinmike:
Previously, many tests un-registered service workers only after all
assertions had been satisfied. This meant that failing tests would *not*
un-register workers. In order to account for workers that persisted from
previous test failures, the tests included "setup" code to defensively
un-register such workers.

asynchronous "clean up" operations [1], but the functionality is only
available to tests declared using `promise_test`. Update tests which
were previously declared with `async_test` and use the new "async
cleanup" API to schedule service worker un-registration so that it
occurs regardless of the result of each test.

---

Chromium on `master`:

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 96 checks (32 tests, 64 subtests)
    Expected results: 93
    Unexpected results: 3
      subtest: 3 (3 fail)

Chromium with patch applied:

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 96 checks (32 tests, 64 subtests)
    Expected results: 93
    Unexpected results: 3
      subtest: 3 (3 fail)

Firefox on `master`:

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 96 checks (32 tests, 64 subtests)
    Expected results: 89
    Unexpected results: 7
      test: 1 (1 timeout)
      subtest: 6 (4 fail, 2 timeout)

Firefox with patch applied:

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 96 checks (32 tests, 64 subtests)
    Expected results: 89
    Unexpected results: 7
      test: 1 (1 timeout)
      subtest: 6 (4 fail, 1 notrun, 1 timeout)

The discrepancy in Firefox is expected. In `master`, two of the four
`async_tests` in `fetch-request-css-images.https.html` time out in Firefox.
This branch refactors all four into `promise_test`s. `promise_test`s run in
series (unlike the parallel `async_test`s), so the harness stops running the
tests following the first timeout and reports the final test as "NOT RUN".

I've attempted to ease the review process by splitting the work across a few
pull requests, but this patch is still pretty onerous. I'd be happy to break it
up further if desired.

### DIFF
--- a/service-workers/service-worker/activation-after-registration.https.html
+++ b/service-workers/service-worker/activation-after-registration.https.html
@@ -5,25 +5,24 @@
 <script src="resources/test-helpers.sub.js"></script>
 <body>
 <script>
-var t = async_test('activation occurs after registration');
-t.step(function() {
+promise_test(function(t) {
     var scope = 'resources/blank.html';
     var registration;
 
-    service_worker_unregister_and_register(
+    return service_worker_unregister_and_register(
         t, 'resources/empty-worker.js', scope)
       .then(function(r) {
+          t.add_cleanup(function() {
+              return service_worker_unregister(t, scope);
+            });
+
           registration = r;
           assert_equals(
               r.installing.state,
               'installing',
               'worker should be in the "installing" state upon registration');
           return wait_for_state(t, r.installing, 'activated');
-        })
-      .then(function() {
-          service_worker_unregister_and_done(t, scope);
-        })
-      .catch(unreached_rejection(t));
-});
+        });
+}, 'activation occurs after registration');
 </script>
 </body>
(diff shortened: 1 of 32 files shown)


mfalken (approved):
Sorry for the review delay. I didn't have an effective system for receiving GitHub reviews in my inbox. This is great. Thank you for all your work adding async cleanup and for converting these tests.

This patch is fine as-is. One observation is we sometimes remove frames inline and sometimes in add_cleanup. I think add_cleanup is better to isolate promise_tests from each other. We've also tried to remove the frames before calling unregister, as it makes it more likely for the service worker to actually get deactivated, since otherwise it waits for all clients to close. But that's sometimes inconvenient to code.

jugglinmike:
Happy to help out! Would you mind merging on my behalf, @mattto?

I noticed some inconsistency with iframe management while writing this patch. I'd be happy to normalize things a bit in a follow up.

mattto:
Thanks! Sorry for the delay. I didn't know I could merge things.

PR #13174, opened 2018-09-23
Avoid relying on ResourceTiming L2 for preload tests

yoavweiss:
This change makes sure that preload tests do not fail when ResourceTiming transferSize is not supported, as it is not essential for the tests.

### DIFF
--- a/preload/resources/preload_helper.js
+++ b/preload/resources/preload_helper.js
@@ -14,7 +14,7 @@ function verifyNumberOfDownloads(url, number)
 {
     var numDownloads = 0;
     performance.getEntriesByName(getAbsoluteURL(url)).forEach(entry => {
-        if (entry.transferSize > 0) {
+        if (entry.transferSize === undefined || entry.transferSize > 0) {
             numDownloads++;
         }
     });


yoavweiss:
The purpose here is to have https://wpt.fyi/results/preload?label=experimental better reflect preload support when it comes to Safari (which supports preload, but not `transferSize`)

yoavweiss:
@ddragana - seems like firefox nightly flakes, but I'm not sure if it's the test change (it consistently fails locally for me).
Could you take a look at the flakes? Is the nightly implementation behind a flag I should turn on to see the flakes locally?

preload/resources/preload_helper.js
```diff
@@ -14,7 +14,7 @@ function verifyNumberOfDownloads(url, number)
 {
     var numDownloads = 0;
     performance.getEntriesByName(getAbsoluteURL(url)).forEach(entry => {
-        if (entry.transferSize > 0) {
+        if (entry.transferSize === undefined || entry.transferSize > 0) {
```
igrigorik: Sanity check: is transferSize even necessary? Is the intent to explicitly count number of resources fetched over the network, as opposed to fetched from HTTP cache?

yoavweiss: I'm *think* that was the original intent, but not 100% sure. @ddragana introduced it as part of https://github.com/web-platform-tests/wpt/commit/e7327c5ba23c244c1e98358f9622cd1e84202ba2

@ddragana - Can you confirm?

Issue #12693, opened 2018-08-27
WindowsError does not exist on non-Windows


Ms2ger:
https://github.com/web-platform-tests/wpt/pull/9652 added an `except WindowsError` clause, but if there's an error raised in the `try` block on non-Windows, you get

```
  File "/tools/manifest/vcs.py", line 19, in git
    except WindowsError:
NameError: global name 'WindowsError' is not defined
```

CC @csnardi

gsnedders:
Oh I thought I'd fixed this. Obviously not. :(

PR #13177, opened 2018-09-23
Check if platform is Windows before catching WindowsError

Cactusmachete:
Check if the platform is Windows and the error is a WindowsError before
trying to run git.bat, as WindowsError is not defined in non-Windows.

Closes https://github.com/web-platform-tests/wpt/issues/12693


jgraham (approved):

jgraham:
Thanks!

PR #13178, opened 2018-09-23
AZZ replaced with A9Z

### LINKED ISSUE #13066: [css-contain] quote-scoping-002 "AZZ" to be replaced with "A9Z"
@frivoal 

In [quote-scoping-002](http://test.csswg.org/suites/css-contain-1_dev/nightly-unstable/html/quote-scoping-002.htm) test, "AZZ" should be replaced with "A9Z" . 
The [quote-scoping-002-ref](http://test.csswg.org/suites/css-contain-1_dev/nightly-unstable/html/reference/quote-scoping-002-ref.htm) reference file is correct.

kritisingh1:
Fixes #13066 
@TalbotG please review. Thanks!

### DIFF
--- a/css/css-contain/quote-scoping-002.html
+++ b/css/css-contain/quote-scoping-002.html
@@ -24,5 +24,5 @@ span {
 }
 </style>
 
-<p>Test passes if the text below is "AZZ" (not including the quotation marks).<p>
+<p>Test passes if the text below is "A9Z" (not including the quotation marks).<p>
 <div><span></span></div>


TalbotG:
@frivoal is the appropriate and **best person to review and merge** your pull request, kritisingh1. It is really preferable that Florian reviews and merges this. Florian is author of the css-contain spec and author of the quote-scoping-002 test.

frivoal:
Thanks. We should indeed fix the difference between the test and the reference. However, it seems to me that the reference is the correct one, not the test, so they should both be fixed to read `AZZ`.

From the spec:

> The effects of the content property’s open-quote, close-quote, no-open-quote and no-close-quote must be scoped to the element’s sub-tree.
> 
> Note: This implies that the depth of quote nesting in the subtree is unchanged and starts at the value that its context normally implies, but that changes to the depth of quote nesting by these values inside the subtree do not affect the depth of quote nesting outside the subtree.

There's only 1 `open-quote`, so it must yield `A`. Then the `close-quote` on `span::after` yields the corresponding `Z`. But since it is inside of the style containment, it does not affect what `div::after` sees, so that one two yields a `Z`.

Do you think I am getting this wrong?

TalbotG:
> There's only 1 open-quote, so it must yield A. Then the close-quote on span::after yields the corresponding Z. But since it is inside of the style containment, it does not affect what div::after sees, so that one too yields a Z. Do you think I am getting this wrong?

I think you get this correct.

What "bugs" me is that the span's close-quote value of the content property does not require 'contain: style' anyway in order to be scoped to the element's subtree ...

TalbotG:
Argh.... I am still confused. I need to understand the note from the spec you quoted...

PR #13179, opened 2018-09-23
[resource-timing] Add TAO test for multiple wildcards

yoavweiss:
This PR adds a test for multiple TAO wildcards, to make sure it passes the timing allow check.
It also adds asserts to make sure TAO tests fail if the relevant RT entry is missing.


igrigorik (approved):
👍

PR #13185, opened 2018-09-24
Fix firefox find_binary_path to accomodate channels

Cactusmachete:
Firefox find_binary_path returned None since browser-channels were introduced
due to path differences. This fixes the same, along with a minor change to the
Nightly path on Mac as it contains a space.

The default for |wpt run firefox| without a channel specified is set to nightly,
along with a log message when this happens.

### DIFF
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -102,6 +102,11 @@ class Firefox(Browser):
             "beta": "latest-beta",
             "nightly": "latest"
         }
+        application_name = {
+            "stable": "Firefox.app",
+            "beta": "Firefox.app",
+            "nightly": "Firefox Nightly.app"
+        }
         if channel not in branch:
             raise ValueError("Unrecognised release channel: %s" % channel)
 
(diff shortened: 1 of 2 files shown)


tools/wpt/browser.py
```diff
@@ -134,15 +139,15 @@ def install(self, dest=None, channel="nightly"):
             if platform == "mac" and os.path.exists(os.path.join(dest, "Firefox Nightly.app")):
```
gsnedders: Don't we want this to be application_name[channel] too?

Cactusmachete: Yep! Missed that earlier. Thanks for pointing it out :)

gsnedders (approved):
LGTM, but I'd rather @jgraham had a look too. 🙂

tools/wpt/browser.py
```diff
@@ -151,9 +156,15 @@ def find_binary_path(self, path=None):
             "Darwin": "mac"
         }.get(uname[0])
 
+        application_name = {
```
jgraham: I feel this ought to be factored into a single place, but let's do that in a followup.

tools/wpt/browser.py
```diff
@@ -163,22 +174,24 @@ def find_binary_path(self, path=None):
...
 
     def find_binary(self, venv_path=None, channel=None):
         if venv_path is None:
             venv_path = os.path.join(os.getcwd(), "_venv")
 
-        path = os.path.join(venv_path, "browsers")
-        if channel is not None:
-            path = os.path.join(path, channel)
-        binary = self.find_binary_path(path)
+        if channel is None:
```
jgraham: I think ideally this would be done in the caller, since that's where we log it.

jgraham (approved):
I think this is an improvement so I'll land it now, but there are a couple of pieces of possible cleanup to do if you have some time.

PR #13196, opened 2018-09-25
[presentation-api] Revise the test for reconnecting multiple presentations

tomoyukilabs:
This PR addresses https://github.com/web-platform-tests/wpt/pull/11370#pullrequestreview-157510255. A minor revision has been made so that a couple of presentation connections are checked symmetrically.

### DIFF
--- a/presentation-api/controlling-ua/reconnectToMultiplePresentations_success-manual.https.html
+++ b/presentation-api/controlling-ua/reconnectToMultiplePresentations_success-manual.https.html
@@ -94,13 +94,31 @@
     await eventWatcher2.wait_for('close');
     assert_equals(connection2.state, 'closed', 'The presentation connection is successfully closed.');
 
-    const c1 = await request1.reconnect(connection1.id);
-    assert_equals(c1, connection1, 'The promise is resolved with the existing presentation connection.');
+    const c11 = await request1.reconnect(connection1.id);
+    assert_equals(c11, connection1, 'The promise is resolved with the existing presentation connection.');
+
+    const c22 = await request2.reconnect(connection2.id);
+    assert_equals(c22, connection2, 'The promise is resolved with the existing presentation connection.');
+
+    await Promise.all([
+      eventWatcher1.wait_for('connect'),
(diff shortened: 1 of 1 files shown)


tidoust:
So, that all looks good to me, but I'm not sure I understand the comment in https://github.com/web-platform-tests/wpt/pull/11370#pullrequestreview-157510255 and what this is meant to test on top of what was already being tested. @mfoltzgoogle, could you have a look?

In the initial test, calling `const c2 = await request1.reconnect(connection2.id);` checked that a connection ID does not have strong ties to a `PresentationRequest` object: `reconnect` can be called on any `PresentationRequest` object, even if the connection ID has been created through another `PresentationRequest` object. The update seems to add more of these situations, but why is that needed?

tomoyukilabs:
This update tests reconnection of two connections in two rounds. In round 1, both connections are reconnected through their tied presentation requests, as a regular pattern. Round 2 represents an irregular pattern where each connection is reconnected through a presentation request tied to other connection. Note that round 1 would fail on Chrome 70 as far as I've tried.

Anyway, I can agree to close this PR without merging if this update seems redundant and unnecessary.

mfoltzgoogle:
@tidoust I think we should test `reconnect()` on both `request1` and `request2` for the multiple-presentation scenario, which this PR does.  Testing that the ID is bound to the URL and not the request object seems useful as well.

I'll look into the Round 1 failure.  Chrome currently only supports a single 1-UA presentation at a time per profile which may be an issue with passing this test.

tomoyukilabs:
@mfoltzgoogle Is a single 1-UA restriction applicable to HDMI-connected displays? Actually, there are one HDMI display and one Chromecast in my testing environment.

tomoyukilabs:
I talked to @mfoltzgoogle at TPAC, and we confirmed that the test for closing and reconnecting multiple connections simultaneously is necessary, as this PR does.

@tidoust if you don't have any concern, could you merge this PR?

tidoust (approved):

PR #13197, opened 2018-09-25
Run resources unit tests on python 3.

Ms2ger:


.travis.yml
```diff
@@ -64,6 +64,8 @@ matrix:
       env: JOB=tools_unittest TOXENV=py36 HYPOTHESIS_PROFILE=ci SCRIPT=tools/ci/ci_tools_unittest.sh
     - python: 2.7
       env: JOB=resources_unittest TOXENV=py27 SCRIPT=tools/ci/ci_resources_unittest.sh
+    - python: 3.6
```
gsnedders: Is this actually worth the performance cost? Ultimately the goal here is to the test the JS rather the Python?

wpt
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
```
gsnedders: Breaks on macOS out-of-the-box. We can't do this. :(

If you want to be super-evil, you could write something that started python2 || python.

gsnedders (commented):
Do we not need to add `six` as a dep in tox.ini? I guess this probably works given it's already an indirect dep, but aside from that LGTM.

gsnedders (approved):

PR #13200, opened 2018-09-25
Make wptrunner chrome always use the WebDriver executor

gsnedders:


jgraham (approved):

tools/wptrunner/wptrunner/browsers/chrome.py
```diff
@@ -30,34 +30,37 @@ def browser_kwargs(test_type, run_info_data, config, **kwargs):
...
+                }
+            },
+            "w3c": True
         }
     }
+
     for (kwarg, capability) in [("binary", "binary"), ("binary_args", "args")]:
         if kwargs[kwarg] is not None:
             capabilities["goog:chromeOptions"][capability] = kwargs[kwarg]
-    if kwargs["headless"]:
```
jgraham: This should have beeen added to `chrome_webdriver`, please [re-]add it before landing.

gsnedders: Added that in #13201, let's land that first

gsnedders:
cc @jugglinmike @Hexcles you guys should know this has landed so you can revert if it breaks wpt.fyi

Hexcles:
@gsnedders thanks for the heads up! This is exciting!

PR #13211, opened 2018-09-26
[css-text] word-break:keep-all with U+3000 (space) and U+3001 (punctuation)

frivoal:
At the time of writing, Chrome passes, but Firefox and Safari fail.

Related to https://github.com/w3c/csswg-drafts/issues/3142


foolip (approved):
I did light review.

PR #13218, opened 2018-09-26
Revert "Make wptrunner chrome always use the WebDriver executor"

gsnedders:
Reverts web-platform-tests/wpt#13200.

Regressions on TC: https://staging.wpt.fyi/results/?product=chrome@1e7408fcd3&product=chrome@9bc71d422a&diff=true


gsnedders:
Um, that's not that PR GitHub.

gsnedders:
And most of those commits are empty, WTF.

PR #13221, opened 2018-09-26
Simplify MultipartContent.

Ms2ger:
The class is only used once, so change it to simplify that case.


jgraham:
If multipart supports multiple content types then it seems a shame to go back to only supporting one, even if we aren't really taking advantage yet.

Ms2ger:
I think this is a YAGNI situation, but feel free to close.

jgraham:
I feel like changing it is worse than just leaving it alone.

PR #13228, opened 2018-09-26
Added whitespace test and template system fix for UTF-8 whitespace

andypaicu:
Hi @foolip 
This is in relation to the whitespace issue in https://github.com/webcompat/web-bugs/issues/18902

It contains a test (with some support files) and a small fix for the template system.

Normally I would have fixed this is chrome and added a test and let it upstream naturally but unfortunately the test also requires a tiny fix to the template system to allow non-ASCII characters in the request GET dictionary so I had to submit a manual PR.

I have run this test in chrome and firefox and the results are almost as expected. Interestingly enough, firefox allows NBSP (U+00A0) in between the directive name and value, but not inside the directive value which I believe will need fixing as it can't be right.

Firefox

PASS Should load image without any CSP - meta tag
PASS Should load image without any CSP - HTTP header
PASS Should not load image with 'none' CSP - meta tag
PASS Should not load image with 'none' CSP - HTTP header
PASS U+0009 TAB   should be properly parsed between directive name and value - meta tag
PASS U+0009 TAB   should be properly parsed between directive name and value - HTTP header
PASS U+000C FF    should be properly parsed between directive name and value - meta tag
PASS U+000C FF    should be properly parsed between directive name and value - HTTP header
PASS U+000A LF    should be properly parsed between directive name and value - meta tag
PASS U+000D CR    should be properly parsed between directive name and value - meta tag
PASS U+0020 SPACE should be properly parsed between directive name and value - meta tag
PASS U+0020 SPACE should be properly parsed between directive name and value - HTTP header
PASS U+0009 TAB   should be properly parsed inside directive value - meta tag
PASS U+0009 TAB   should be properly parsed inside directive value - HTTP header
PASS U+000C FF    should be properly parsed inside directive value - meta tag
PASS U+000C FF    should be properly parsed inside directive value - HTTP header
PASS U+000A LF    should be properly parsed inside directive value - meta tag
PASS U+000D CR    should be properly parsed inside directive value - meta tag
PASS U+0020 SPACE should be properly parsed inside directive value - meta tag
PASS U+0020 SPACE should be properly parsed inside directive value - HTTP header
FAIL U+00A0 NBSP  should be properly parsed between directive name and value - meta tag 
FAIL U+00A0 NBSP  should be properly parsed between directive name and value - HTTP header 
PASS U+00A0 NBSP  should be properly parsed inside directive value - meta tag
PASS U+00A0 NBSP  should be properly parsed inside directive value - HTTP header

Chrome
PASS Should load image without any CSP - meta tag
PASS Should load image without any CSP - HTTP header
PASS Should not load image with 'none' CSP - meta tag
PASS Should not load image with 'none' CSP - HTTP header
PASS U+0009 TAB   should be properly parsed between directive name and value - meta tag
PASS U+0009 TAB   should be properly parsed between directive name and value - HTTP header
PASS U+000C FF    should be properly parsed between directive name and value - meta tag
PASS U+000C FF    should be properly parsed between directive name and value - HTTP header
PASS U+000A LF    should be properly parsed between directive name and value - meta tag
PASS U+000D CR    should be properly parsed between directive name and value - meta tag
PASS U+0020 SPACE should be properly parsed between directive name and value - meta tag
PASS U+0020 SPACE should be properly parsed between directive name and value - HTTP header
PASS U+0009 TAB   should be properly parsed inside directive value - meta tag
PASS U+0009 TAB   should be properly parsed inside directive value - HTTP header
PASS U+000C FF    should be properly parsed inside directive value - meta tag
PASS U+000C FF    should be properly parsed inside directive value - HTTP header
PASS U+000A LF    should be properly parsed inside directive value - meta tag
PASS U+000D CR    should be properly parsed inside directive value - meta tag
PASS U+0020 SPACE should be properly parsed inside directive value - meta tag
PASS U+0020 SPACE should be properly parsed inside directive value - HTTP header
FAIL U+00A0 NBSP  should be properly parsed between directive name and value - meta tag 
FAIL U+00A0 NBSP  should be properly parsed between directive name and value - HTTP header 
FAIL U+00A0 NBSP  should be properly parsed inside directive value - meta tag
FAIL U+00A0 NBSP  should be properly parsed inside directive value - HTTP header

### DIFF
--- a/content-security-policy/generic/only-valid-whitespaces-are-allowed.html
+++ b/content-security-policy/generic/only-valid-whitespaces-are-allowed.html
@@ -1,0 +1,66 @@
+<!DOCTYPE HTML>
+<html>
+<head>
+    <script src='/resources/testharness.js'></script>
+    <script src='/resources/testharnessreport.js'></script>
+</head>
+<body>
+  <script>
+    var tests = [
+      // Make sure that csp works properly in normal situations
+      { "csp": "", "expected": true, "name": "Should load image without any CSP" },
+      { "csp": "img-src 'none';", "expected": false, "name": "Should not load image with 'none' CSP" },
+      // Ensure ASCII whitespaces are properly parsed
+      // ASCII whitespace is U+0009 TAB, U+000A LF, U+000C FF, U+000D CR, or U+0020 SPACE.
+
+      // between directive name and value
+      { "csp": "img-src\u0009'none';", "expected": false, "name": "U+0009 TAB   should be properly parsed between directive name and value" },
+      { "csp": "img-src\u000C'none';", "expected": false, "name": "U+000C FF    should be properly parsed between directive name and value" },
+      { "csp": "img-src\u000A'none';", "expected": false, "name": "U+000A LF    should be properly parsed between directive name and value" },
+      { "csp": "img-src\u000D'none';", "expected": false, "name": "U+000D CR    should be properly parsed between directive name and value" },
+      { "csp": "img-src\u0020'none';", "expected": false, "name": "U+0020 SPACE should be properly parsed between directive name and value" },
+
+      // inside directive value
+      { "csp": "img-src http://example.com\u0009http://example2.com;", "expected": false, "name": "U+0009 TAB   should be properly parsed inside directive value" },
+      { "csp": "img-src http://example.com\u000Chttp://example2.com;", "expected": false, "name": "U+000C FF    should be properly parsed inside directive value" },
+      { "csp": "img-src http://example.com\u000Ahttp://example2.com;", "expected": false, "name": "U+000A LF    should be properly parsed inside directive value" },
+      { "csp": "img-src http://example.com\u000Dhttp://example2.com;", "expected": false, "name": "U+000D CR    should be properly parsed inside directive value" },
+      { "csp": "img-src http://example.com\u0020http://example2.com;", "expected": false, "name": "U+0020 SPACE should be properly parsed inside directive value" },
+
+      // Ensure nbsp (U+00A0) is not considered a valid whitespace
+      { "csp": "img-src\u00A0'none';", "expected": false, "name": "U+00A0 NBSP  should be properly parsed between directive name and value" },
+      { "csp": "img-src http://example.com\u00A0http://example2.com;", "expected": false, "name": "U+00A0 NBSP  should be properly parsed inside directive value" },
+    ];
+
+    tests.forEach(test => {
+      async_test(t => {
+        var url = "support/load_img_and_post_result_meta.sub.html?csp=" + encodeURIComponent(test.csp);
+        test_image_loads_as_expected(test, t, url);
+      }, test.name + " - meta tag");
+
+      // We can't test csp delivered in an HTTP header if we're testing CR/LF characters
+      if (test.csp.indexOf("\u000A") == -1 && test.csp.indexOf("\u000D") == -1) {
+        async_test(t => {
+          var url = "support/load_img_and_post_result_meta.sub.html?csp=" + encodeURIComponent(test.csp);
+          test_image_loads_as_expected(test, t, url);
+        }, test.name + " - HTTP header");
+      }
+    });
+
+    function test_image_loads_as_expected(test, t, url) {
+      var i = document.createElement('iframe');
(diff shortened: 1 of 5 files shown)


andypaicu:
Agh the 00A0 is actually incorrect, I've flipped the expected result, it should *not* be parsed.

That means that chrome is actually passing all these tests, and firefox is failing two of them.

I will try to figure out why that is.

content-security-policy/generic/only-valid-whitespaces-are-allowed.html
```diff
@@ -0,0 +1,66 @@
...
+      { "csp": "img-src\u0020'none';", "expected": false, "name": "U+0020 SPACE should be properly parsed between directive name and value" },
+
+      // inside directive value
+      { "csp": "img-src http://example.com\u0009http://example2.com;", "expected": false, "name": "U+0009 TAB   should be properly parsed inside directive value" },
+      { "csp": "img-src http://example.com\u000Chttp://example2.com;", "expected": false, "name": "U+000C FF    should be properly parsed inside directive value" },
+      { "csp": "img-src http://example.com\u000Ahttp://example2.com;", "expected": false, "name": "U+000A LF    should be properly parsed inside directive value" },
+      { "csp": "img-src http://example.com\u000Dhttp://example2.com;", "expected": false, "name": "U+000D CR    should be properly parsed inside directive value" },
+      { "csp": "img-src http://example.com\u0020http://example2.com;", "expected": false, "name": "U+0020 SPACE should be properly parsed inside directive value" },
+
+      // Ensure nbsp (U+00A0) is not considered a valid whitespace
```
foolip: Can you link to https://github.com/webcompat/web-bugs/issues/18902 to explain why this is of particular interest?

tools/wptserve/wptserve/pipes.py
```diff
@@ -471,21 +471,24 @@ def config_replacement(match):
             raise Exception("Undefined template variable %s" % field)
 
         while tokens:
-            ttype, field = tokens.popleft()
+            ttype, ffield = tokens.popleft()
```
foolip: The `ttype` convention is a bit odd but this follows it at first glance. But it doesn't seem like it's to avoid shadowing a `type` variable, so I take it it's actually t as in token. Going by that, let me suggest `tfield` here instead.

tools/wptserve/wptserve/pipes.py
```diff
@@ -471,21 +471,24 @@ def config_replacement(match):
...
-                    "unexpected token type %s (token '%r'), expected ident or arguments" % (ttype, field)
+                    "unexpected token type %s (token '%r'), expected ident or arguments" % (ttype, ffield)
                 )
 
         assert isinstance(value, (int, (binary_type, text_type))), tokens
 
         if variable is not None:
             variables[variable] = value
 
+        if field == "GET" and not isinstance(value, str):
```
foolip: I'm not sure about the ramifications this might have. Would any case where this code path is hit previous be totally broken, or is it some more subtle change of behavior? If subtle, then we should look at differences in test results after this lands.

Also, I guess we'll need to import this change into Chromium ASAP after it lands here, or the tests won't work? (Tests and tools are imported separately, which is a bit of a problem in cases like this.)

foolip: Alright, sounds like we can try landing this without running all the tests against the change first.

foolip (approved):

tools/wptserve/wptserve/pipes.py
Ms2ger: This doesn't make sense to me at all, and I can't reproduce it either. There's already
```py
if isinstance(value, binary_type):
    value = value.decode("utf-8")
```

below, which I can only imagine makes your change pointless.

Can you explain what exactly is going wrong?

gsnedders: And is there any reasons that we don't have any tests added for whatever the failure was?

gsnedders: @andypaicu Oh, I thought we had unit tests for this, but no, only the replacement tokenizer.

karlcow:
Firefox issue for this https://bugzilla.mozilla.org/show_bug.cgi?id=1497742

PR #13236, opened 2018-09-27
Add test for iframe without src attribute

upsuper:


upsuper:
Gecko currently fails on the two tests here, and there is a [Gecko bug](https://bugzilla.mozilla.org/show_bug.cgi?id=543435) probably for this.

zcorpan (approved):

PR #13246, opened 2018-09-27
[wptserve] Rework header encoding/decoding in Request

### LINKED ISSUE #13204: Many tests in /cookies/http-state consistently time out on Taskcluster but not on BuildBot
https://staging.wpt.fyi/results/cookies/http-state?sha=b040cb2a07&product=firefox[experimental,taskcluster]&product=firefox[buildbot,experimental]&diff=true
https://staging.wpt.fyi/results/cookies/http-state?sha=b040cb2a07&product=chrome[experimental,taskcluster]&product=chrome[buildbot,experimental]&diff=true

Presumably, the tests are a bit slow. I'll try to give them long timeout.

Hexcles:
This PR relands the reverted #13248 from @Ms2ger , but with some tweaks on how to encode/decode the headers. See individual commits for details.

Fixes #13204 .

### DIFF
--- a/tools/wptserve/tests/functional/base.py
+++ b/tools/wptserve/tests/functional/base.py
@@ -75,7 +75,7 @@ class TestUsingServer(unittest.TestCase):
             req.add_data(body)
 
         if auth is not None:
-            req.add_header("Authorization", "Basic %s" % base64.b64encode('%s:%s' % auth))
+            req.add_header("Authorization", b"Basic %s" % base64.b64encode(("%s:%s" % auth).encode("utf-8")))
 
         return urlopen(req)
 
--- a/tools/wptserve/tests/functional/base.py
+++ b/tools/wptserve/tests/functional/base.py
@@ -75,7 +75,7 @@ class TestUsingServer(unittest.TestCase):
             req.add_data(body)
 
         if auth is not None:
-            req.add_header("Authorization", b"Basic %s" % base64.b64encode(("%s:%s" % auth).encode("utf-8")))
+            req.add_header("Authorization", b"Basic %s" % base64.b64encode((b"%s:%s" % auth)))
 
         return urlopen(req)
 
--- a/tools/wptserve/tests/functional/test_pipes.py
+++ b/tools/wptserve/tests/functional/test_pipes.py
@@ -115,7 +115,6 @@ server: http://localhost:{0}""".format(self.server.port).encode("ascii")
     def test_sub_var(self):
         resp = self.request("/sub_var.sub.txt")
         port = self.server.port
-        print(port, type(port))
         expected = b"localhost %d A %d B localhost C" % (port, port)
         self.assertEqual(resp.read().rstrip(), expected)
 
--- a/tools/wptserve/tests/functional/test_pipes.py
+++ b/tools/wptserve/tests/functional/test_pipes.py
@@ -57,40 +57,35 @@ class TestSlice(TestUsingServer):
         self.assertEqual(resp.read(), expected[:10])
 
 class TestSub(TestUsingServer):
-    @pytest.mark.xfail(sys.version_info >= (3,), reason="wptserve only works on Py2")
     def test_sub_config(self):
         resp = self.request("/sub.txt", query="pipe=sub")
-        expected = "localhost localhost %i" % self.server.port
+        expected = b"localhost localhost %i" % self.server.port
         self.assertEqual(resp.read().rstrip(), expected)
 
     @pytest.mark.xfail(sys.platform == "win32",
                        reason="https://github.com/web-platform-tests/wpt/issues/12949")
-    @pytest.mark.xfail(sys.version_info >= (3,), reason="wptserve only works on Py2")
     def test_sub_file_hash(self):
         resp = self.request("/sub_file_hash.sub.txt")
-        expected = """
+        expected = b"""
 md5: JmI1W8fMHfSfCarYOSxJcw==
 sha1: nqpWqEw4IW8NjD6R375gtrQvtTo=
 sha224: RqQ6fMmta6n9TuA/vgTZK2EqmidqnrwBAmQLRQ==
 sha256: G6Ljg1uPejQxqFmvFOcV/loqnjPTW5GSOePOfM/u0jw=
 sha384: lkXHChh1BXHN5nT5BYhi1x67E1CyYbPKRKoF2LTm5GivuEFpVVYtvEBHtPr74N9E
-sha512: r8eLGRTc7ZznZkFjeVLyo6/FyQdra9qmlYCwKKxm3kfQAswRS9+3HsYk3thLUhcFmmWhK4dXaICz
-JwGFonfXwg=="""
+sha512: r8eLGRTc7ZznZkFjeVLyo6/FyQdra9qmlYCwKKxm3kfQAswRS9+3HsYk3thLUhcFmmWhK4dXaICzJwGFonfXwg=="""
         self.assertEqual(resp.read().rstrip(), expected.strip())
 
     def test_sub_file_hash_unrecognized(self):
         with self.assertRaises(urllib.error.HTTPError):
             self.request("/sub_file_hash_unrecognized.sub.txt")
 
-    @pytest.mark.xfail(sys.version_info >= (3,), reason="wptserve only works on Py2")
     def test_sub_headers(self):
         resp = self.request("/sub_headers.txt", query="pipe=sub", headers={"X-Test": "PASS"})
-        expected = "PASS"
+        expected = b"PASS"
         self.assertEqual(resp.read().rstrip(), expected)
 
     @pytest.mark.xfail(sys.platform == "win32",
                        reason="https://github.com/web-platform-tests/wpt/issues/12949")
-    @pytest.mark.xfail(sys.version_info >= (3,), reason="wptserve only works on Py2")
     def test_sub_location(self):
         resp = self.request("/sub_location.sub.txt?query_string")
         expected = """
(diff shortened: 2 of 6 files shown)


tools/wptserve/tests/functional/test_request.py
```diff
@@ -113,6 +114,17 @@ def handler(request, response):
...
         self.assertEqual(b"some route", resp.read())
 
+    def test_unicode_in_headers(self):
+        @wptserve.handlers.handler
+        def handler(request, response):
+            return request.headers["foo"]
+
+        route = ("GET", "/test/test_unicode_in_headers", handler)
+        self.server.router.register(*route)
+        encoded_text = u"你好".encode("utf-8")
```
gsnedders: I think we should test more than just UTF-8 (given this would still pass with `header.decode('utf-8')`).

tools/wptserve/wptserve/ranges.py
```diff
@@ -3,6 +3,7 @@
 
 class RangeParser(object):
     def __call__(self, header, file_size):
+        header = header.decode("ascii")
```
gsnedders: Let's put this in a try block, with `except UnicodeDecodeError: raise HTTPException(400, "Non-ASCII range header value")`, rather than returning a somewhat mysterious error.

tools/wptserve/wptserve/request.py
```diff
@@ -365,11 +365,30 @@ def maybedecode(s):
     raise TypeError("Unexpected value in RequestHeaders: %r" % s)
 
 
+def maybeencode(s):
```
gsnedders: Can you write a docstring to actually make it clear what this is and what this is for?

tools/wptserve/wptserve/request.py
```diff
@@ -308,7 +308,7 @@ def POST(self):
             self.raw_input.seek(0)
             fs = cgi.FieldStorage(fp=self.raw_input,
                                   environ={"REQUEST_METHOD": self.method},
-                                  headers=self.headers,
+                                  headers=self.raw_headers,
```
gsnedders: Why the behaviour change here?

tools/wptserve/wptserve/request.py
```diff
@@ -613,14 +632,18 @@ def __init__(self, headers):
...
 
     def decode_basic(self, data):
-        assert isinstance(data, text_type)
-        decoded_data = base64.decodestring(data.encode("utf-8"))
-        return decoded_data.decode("utf-8").split(":", 1)
+        assert isinstance(data, binary_type)
+        # RFC7617 Section 2.1 does not specify the encoding for username &
+        # passsword (as long it's compatible with ASCII). UTF-8 should be a
+        # relatively safe choice as Firefox and other browsers use it.
+        decoded_data = base64.b64decode(data).decode("utf-8")
```
gsnedders: Given the historic differences here, I think we should give the username/password as bytes? Certainly we shouldn't do something irreversible (and throw in the invalid UTF-8 case).

tools/wptserve/wptserve/request.py
Hexcles: Because `cgi.FieldStorage` needs to parse the headers and expect the values to be string not bytes. Since `self.readers` is read-only, I figure it would be safe to pass in the original `raw_headers` here.

Hexcles:
A few clarifications from the discussions on IRC regarding whether we should encode the header values back to bytes (i.e. `headers = [maybeencode(items[header])]`): https://w3.logbot.info/testing/20180928#c444317

First of all, regardless of the approach (whether to encode and which encoding to use), this WILL NOT affect existing tests (namely custom handlers) at all. Currently, WPT only really works with Python 2, and in Python 2 headers (both names and values) have `str` type, which is a "binary" type, so `maybeencode` is a no-op.

People reasonably want `wptserve` itself to work in Python 3, and this is where the change has effects. I would hope Python 3 still gives us headers in binary type (i.e. `bytes`), as Python 2, but Python 3 decides to decode the headers in iso-8859-1 perhaps for better compatibility. I think the sane choice here is to return `bytes`, which requires re-encoding, as HTTP RFC doesn't really say what encoding headers *must* use and encoding non-ASCII characters in iso-8859-1 is broken. The downside is that potentially more places that manipulate headers need to be changed (e.g. literal operands of string operations need to be `bytes`).

Now, since many custom handlers don't work in Python 3 in the first place, I'm inclined to to `maybeencode` the header values (i.e. return `bytes` in Python 3).

Hexcles:
@gsnedders This PR is ready for review again. I addressed all your comments but one (the `cgi.FieldStorage` one; please see my reply there).

annevk:
> and encoding non-ASCII characters in iso-8859-1 is broken

It's not? Using bytes is perhaps purer in a way, but if you make them strings that is the way to go. Anything else would lead to information loss.

Ms2ger:
I suspect that supporting only `binary_type` in this API would be the most straightforward approach.

Hexcles:
@Ms2ger you mean keys should also be binary type? Yeah I can do that. It might be break a few more things (in Python 3 only) but it's probably more consistent.

gsnedders:
@Ms2ger @Hexcles I have no strong preference; in principle header names are strictly ASCII only, though I don't think impls enforce that.

Hexcles:
@Ms2ger @gsnedders there's one complication here. Technically, a byte string doesn't have the `lower` method. Yet we usually wish header names to be case-insensitive and we implement it by always converting the keys to lowercase both during storage and lookup.

gsnedders:
@Hexcles Yes it does? `bytes.lower` does ASCII lowercasing on both Py2/3.

Hexcles:
@gsnedders duh my bad. Yeah let me update the PR.

tools/wptserve/tests/functional/base.py
```diff
@@ -75,7 +75,7 @@ def request(self, path, query=None, method="GET", headers=None, body=None, auth=
             req.add_data(body)
 
         if auth is not None:
-            req.add_header("Authorization", "Basic %s" % base64.b64encode('%s:%s' % auth))
+            req.add_header("Authorization", b"Basic %s" % base64.b64encode(("%s:%s" % auth).encode("utf-8")))
```
gsnedders: I wonder if we want to not encode there, so we can test with non-UTF-8 values?

tools/wptserve/tests/functional/test_pipes.py
```diff
@@ -101,30 +96,27 @@ def test_sub_location(self):
...
         resp = self.request("/sub_uuid.sub.txt")
-        self.assertRegexpMatches(resp.read().rstrip(), r"Before [a-f0-9-]+ After")
+        self.assertRegexpMatches(resp.read().rstrip(), b"Before [a-f0-9-]+ After")
 
-    @pytest.mark.xfail(sys.version_info >= (3,), reason="wptserve only works on Py2")
     def test_sub_var(self):
         resp = self.request("/sub_var.sub.txt")
         port = self.server.port
-        expected = "localhost %s A %s B localhost C" % (port, port)
+        print(port, type(port))
```
gsnedders: I presume this is residual from debugging? :)

tools/wptserve/tests/functional/test_request.py
gsnedders: I'd rather this be done with `pytest.mark.parametrize` so they run as separate tests, FWIW.

tools/wptserve/wptserve/pipes.py
```diff
@@ -490,7 +491,11 @@ def config_replacement(match):
 
         #Should possibly support escaping for other contexts e.g. script
         #TODO: read the encoding of the response
-        return escape_func(text_type(value)).encode("utf-8")
+        if isinstance(value, binary_type):
+            value = value.decode("utf-8")
```
gsnedders: ~~Can we just avoid the roundtrip here?~~ No, we can't, can you add a comment saying *why* we can't? (i.e., the type of escape_func)

tools/wptserve/wptserve/request.py
gsnedders: `self.raw_headers` is the headers as `bytes`, no? What type does cgi.FieldStorage expect? `str` on both Py2/3, no?

tools/wptserve/wptserve/request.py
```diff
@@ -355,11 +355,34 @@ def __init__(self, request_handler):
...
+
+    raise TypeError("Unexpected value in RequestHeaders: %r" % s)
+
+
 class RequestHeaders(dict):
-    """Dictionary-like API for accessing request headers."""
+    """Read-only dictionary-like API for accessing request headers.
+
+    Unlike BaseHTTPRequestHandler.headers, this class always returns all
+    headers with the same name (separated by commas). Besides, this class
```
gsnedders: "Besides" seem a bit clumsy here. Probably just make this "and"?

tools/wptserve/tests/functional/test_pipes.py
```diff
@@ -101,30 +96,27 @@ def test_sub_location(self):
 port: {0}
 query: ?query_string
 scheme: http
-server: http://localhost:{0}""".format(self.server.port)
+server: http://localhost:{0}""".format(self.server.port).encode("ascii")
```
gsnedders: Is there any reason not to use a binary string here? format still works, no?

tools/wptserve/wptserve/pipes.py
```diff
@@ -273,8 +274,9 @@ def slice(request, response, start, end=None):
                 (spelled "null" in a query string) to indicate the end of
                 the file.
     """
-    content = resolve_content(response)
-    response.content = content[start:end]
+    content = resolve_content(response)[start:end]
+    response.content = content
+    response.headers.set("Content-Length", len(content))
```
gsnedders: This is a separate fix, yes? Do we have any tests for this? ^_^

Also: don't we want both of the header args to be binary strings? I can't see what goes on here looking quickly? But I guess the Response side probably makes sense to do so separately.

tools/wptserve/tests/functional/test_pipes.py
Hexcles: Oops. Removed.

tools/wptserve/wptserve/pipes.py
Hexcles: I didn't modify this file; the change was cherry-picked from #11769 by @Ms2ger .

tools/wptserve/tests/functional/test_pipes.py
Hexcles: No, `bytes` doesn't have `format` method. We could use `%`, but it'd be a bit clumsy as we need to repeat the port number a few times.

tools/wptserve/tests/functional/test_request.py
Hexcles: This is a neat feature! However, it looks like `pytest.mark.parametrize` [does not work](https://docs.pytest.org/en/latest/unittest.html#pytest-features-in-unittest-testcase-subclasses) with `unittest.TestCase`.

tools/wptserve/tests/functional/base.py
Hexcles: Done. Also added a test case (in Shift JIS).

andypaicu:
Just as an fyi guys I'm also waiting on this re-land as I suspect the revert broke the "wpt / content-security-policy / generic / only-valid-whitespaces-are-allowed.html" test.

There hasn't been any movement on this for a week and I just want to make sure that this will eventually get merged.

https://wpt.fyi/results/content-security-policy/generic/only-valid-whitespaces-are-allowed.html?label=experimental

tools/wptserve/tests/functional/test_request.py
gsnedders: Oh, I missed your response here!

Ergh, another reason to move away from `unittest` entirely, but that probably shouldn't block this PR.

gsnedders (approved):

gsnedders:
@Hexcles @Ms2ger do one of you want to rewrite history to something saner?

Ms2ger:
If @Hexcles hasn't done it by tomorrow, I'll squash them all together.

Hexcles:
Thanks for the review, @gsnedders . (The comment threads do look quite confusing...)

I just cleaned up the history and will merge the rebase-merge PR once Travis passes.

Issue #12880, opened 2018-09-06
`foo-manual.html.headers` is considered as a test by manifest


Hexcles:
It should be a support file.

https://github.com/web-platform-tests/wpt/blob/2c6612a9b0f67c95fc8cc772cfecc2c1d435d75d/tools/manifest/sourcefile.py#L184

https://github.com/web-platform-tests/wpt/blob/2c6612a9b0f67c95fc8cc772cfecc2c1d435d75d/tools/manifest/sourcefile.py#L604

This logic above essentially treats anything "-manual" as a manual test.

gsnedders:
Hahahahahaha, this is fantastic. A++ us.

foolip:
@Hexcles how did you discover this? priority:backlog or roadmap?

Hexcles:
Found the issue when helping @clelland to troubleshoot a feature-policy test.

It was worked around by automating the test (using `testdriver.js`, woohoo!), so the test isn't manual any more. However, I can't think of any viable workaround to tell `wptserve` to add headers to a manual test, so let's call it roadmap.

gsnedders:
Fundamentally, I don't think many people are relying on what the manifest lists as manual tests, so it's arguably not even `roadmap` but should be `backlog`.

It shouldn't stop the `.headers` from working, though? `wptserve` doesn't look at the type the manifest gives it (it doesn't have any knowledge of the manifest).

clelland:
> It shouldn't stop the .headers from working, though?

When running `foo-manual.html`, the `foo-manual.html.headers` file *does* get correctly found and used as the headers for the test.

The trouble is that `foo-manual.html.headers` is then *also* run as if it were a test of its own, and of course that test fails.

gsnedders:
FWIW: I think the fix is literally just special-casing ".headers" in the code in the original comment, if anyone wants to try fixing this.

qiuzhong:
Hi, I'm interested in taking this task.

Here's my proposal:

Just treat all ".headers" files as non-test items. 
```
    @property
    def name_is_non_test(self):
        """Check if the file name matches the conditions for the file to
        be a non-test file"""
        return (self.is_dir() or
                self.name_prefix("MANIFEST") or
                self.filename == "META.yml" or
                self.filename.startswith(".") or
                self.filename.endswith(".headers") or
                self.type_flag == "support" or
                self.in_non_test_dir())
```

Then they'll be considered as support files (as **name_is_non_test** checking comes first).

```
[...]

PR #13251, opened 2018-09-28
Consider .headers files as non tests

qiuzhong:
Consider all the .headers files as non tests so the files like
"foo-manual.html.headers" will be classified into the support ones.

Fix: #12880


jgraham (approved):

Hexcles:
Thanks!

Issue #2613, opened 2016-02-23
Test importScripts with multiple arguments in service workers


Ms2ger:
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1249351

Ms2ger:
https://github.com/web-platform-tests/wpt/pull/13271

PR #13271, opened 2018-09-30
[service-workers] Add test for method importScripts to fix issue #2613

wanghongjuan:
- Add test with two different script urls when invoke importScripts(urls) to fix issue #2613
- Simplified test with async/await


service-workers/service-worker/import-scripts-resource-map.https.html
```diff
@@ -10,14 +10,23 @@
...
+  const SCRIPT = 'resources/import-scripts-resource-map-worker.js'
+  await service_worker_unregister(t, SCOPE);
+  const registration =
+      await navigator.serviceWorker.register(SCRIPT, {scope: SCOPE});
+  await registration.unregister();
+}, 'import the same script URL multiple times')
+
+promise_test(async t => {
+  const SCOPE = 'resources/import-scripts-diff-resource-map';
+  const SCRIPT = 'resources/import-scripts-diff-resource-map-worker.js';
```
Ms2ger: You're loading this script from the `service-worker/resources/` subdir, but you put it directly in `service-worker/`.

Ms2ger (changes requested):

service-workers/service-worker/import-scripts-resource-map.https.html
wanghongjuan: Oops, that's my mistake, fix it in new commit.

service-workers/service-worker/import-scripts-resource-map.https.html
```diff
@@ -10,14 +10,23 @@
...
+  await service_worker_unregister(t, SCOPE);
+  const registration =
+      await navigator.serviceWorker.register(SCRIPT, {scope: SCOPE});
+  await registration.unregister();
+}, 'import two different script URLs');
```
mattto: nit: I think the point of the test is rather 'call importScripts() with multiple arguments'

service-workers/service-worker/resources/import-scripts-diff-resource-map-worker.js
```diff
@@ -0,0 +1,12 @@
+importScripts('/resources/testharness.js');
+
+let version = null,
+    echo_output = null;
+importScripts('import-scripts-version.py');
+// Once imported, the stored script should be loaded for subsequent importScripts.
+const expected_version = version;
+
+version = null;
+importScripts('import-scripts-version.py', 'import-scripts-echo.py?msg=test');
```
mattto: It kind of seems like an accident of history that this test is both testing 'import the same script URL multiple times' and 'importScripts() with multiple arguments' that #2613 is about. A simpler test for #2613 would just be a service worker that does a single importScripts(script1, script2); I assume the reason you're using "version" and "echo" is that they are the only scripts that write to global variables so you can do assert_equals() after the call.

I guess that's fine... just a bit hard to understand why this test is written the way it is. Another option could be to extend import-scripts-echo.py to accept a string for the name of the output variable.

mfalken (approved):
I think this looks good to me, though I was a bit puzzled at first.

service-workers/service-worker/resources/import-scripts-diff-resource-map-worker.js
wanghongjuan: @mattto Thanks for your comment. Did your means changing the line 10 to e.g. `importScripts('import-scripts-echo.py?msg=test1', 'import-scripts-echo.py?msg=test2');`?

mattto: I was thinking the following is more clearly testing what #2613 asks for:

```
echo1  = null;
echo2 = null;
importScripts('import-scripts-echo.py?output=echo1&msg=test1', 'import-scripts-echo.py?output=echo2&msg=test2');?
assert_equals(echo1, 'test1');
assert_equals(echo2, 'test2');
```

It'd have to extend import-scripts-echo.py to have an `output` param.

service-workers/service-worker/import-scripts-resource-map.https.html
wanghongjuan: Updated

service-workers/service-worker/resources/import-scripts-diff-resource-map-worker.js
wanghongjuan: Updated in newest commit

service-workers/service-worker/resources/import-scripts-echo.py
```diff
@@ -3,4 +3,4 @@ def main(req, res):
         ('Cache-Control', 'no-cache, must-revalidate'),
         ('Pragma', 'no-cache'),
         ('Content-Type', 'application/javascript')],
-      'echo_output = "%s";\n' % req.GET['msg'])
+        req.GET['output'] + ' = "%s";\n' % req.GET['msg'])
```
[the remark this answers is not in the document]
wanghongjuan: Updated the code.

wanghongjuan:
@mattto PTAL

wanghongjuan:
@Ms2ger PTAL

service-workers/service-worker/resources/import-scripts-diff-resource-map-worker.js
```diff
@@ -0,0 +1,10 @@
+importScripts('/resources/testharness.js');
+
+let echo1 = null;
+    echo2 = null;
```
Ms2ger: Add `let`, or use commas rather than semicolons.

service-workers/service-worker/resources/import-scripts-get.py
```diff
@@ -0,0 +1,6 @@
+def main(req, res):
+    return ([
+        ('Cache-Control', 'no-cache, must-revalidate'),
+        ('Pragma', 'no-cache'),
+        ('Content-Type', 'application/javascript')],
+        req.GET['output'] + ' = "%s";\n' % req.GET['msg'])
```
Ms2ger: Style nit: please use string interpolation twice, rather than both concatenation and interpolation. (Possibly also move the output out of the return statement and into a local variable for readability.)

service-workers/service-worker/import-scripts-resource-map.https.html
```diff
@@ -10,14 +10,23 @@
...
-        })
-      .then(r => r.unregister());
-  }, 'import the same script URL multiple times');
+
+promise_test(async t => {
+  const SCOPE = 'resources/import-scripts-resource-map';
+  const SCRIPT = 'resources/import-scripts-resource-map-worker.js'
+  await service_worker_unregister(t, SCOPE);
+  const registration =
+      await navigator.serviceWorker.register(SCRIPT, {scope: SCOPE});
```
Ms2ger: Style nit: please use 2-space or 4-space indentation consistently. (Twice in this file.)

Ms2ger (approved):

service-workers/service-worker/resources/import-scripts-diff-resource-map-worker.js
wanghongjuan: Updated in latest commit.

service-workers/service-worker/resources/import-scripts-get.py
wanghongjuan: Done

service-workers/service-worker/import-scripts-resource-map.https.html
wanghongjuan: Done

PR #13272, opened 2018-09-30
Fix #8581: add extra subdomains for cookie tests

### LINKED ISSUE #8581: Isolated domain for cookie tests?

gsnedders:
This relands the second commit from #10953, which was reverted as a result of #12461 in #12462. See also the discussion in #12459. Fixes #8581.

### DIFF
--- a/tools/serve/serve.py
+++ b/tools/serve/serve.py
@@ -16,6 +16,7 @@ import traceback
 from six.moves import urllib
 import uuid
 from collections import defaultdict, OrderedDict
+from itertools import chain, product
 from multiprocessing import Process, Event
 
 from localpaths import repo_root
@@ -720,6 +721,9 @@ def build_config(override_path=None, **kwargs):
 
     return rv
 
+def _make_subdomains_product(s, depth=2):
+    return set(u".".join(x) for x in chain(*(product(s, repeat=i) for i in range(1, depth+1))))
+
 _subdomains = {u"www",
                u"www1",
                u"www2",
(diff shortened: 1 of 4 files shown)


jgraham:
I have lost the thread of what's happening here. Why do we think that all possible domains to a depth of 2 is the right tradeoff here?

foolip:
@gsnedders what is the status of this?

gsnedders:
@jgraham there was some discussion on IRC a while back with @annevk that suggested two sufficed for everything, and the reason to duplicate everything was simply that it seemed like the path of least resistance as to what the set should be.

jgraham (approved):

jugglinmike:
I still think [this is too heavy-handed for the use case we're trying to satisfy](https://github.com/web-platform-tests/wpt/pull/12459#issuecomment-415195762), but if I'm alone in that, we should at least update the documentation on the available subdomains:

- [`docs/_running-tests/chrome_android.md`](https://github.com/web-platform-tests/wpt/blob/master/docs/_running-tests/chrome_android.md)
- [`docs/_writing-tests/server-features.md`](https://github.com/web-platform-tests/wpt/blob/d5a33af8ef598019bc84515add78e030b54e13ed/docs/_writing-tests/server-features.md)

An explicit contract will help contributors discover the full depth of alternative which are available to them and give them confidence that their new tests aren't relying on accidental behavior. It will also help maintainers understand what they are responsible for maintaining and deploying.

Issue #13276, opened 2018-09-30
TestRunnerManager AssertionError assert isinstance(self.state, RunnerManagerState.restarting)


gsnedders:
[wpt-chrome-dev-testharness-6](https://tools.taskcluster.net/groups/HuC12DhIShShqh7UcovreA/tasks/Pn2YWHFYRuWDpl3y1bWEAA/runs/0/logs/public%2Flogs%2Flive.log) ends with:

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/testrunner.py", line 359, in run
    new_state = self.wait_event()
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/testrunner.py", line 434, in wait_event
    return f(*data)
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/testrunner.py", line 653, in error
    self.restart_runner()
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/testrunner.py", line 644, in restart_runner
    assert isinstance(self.state, RunnerManagerState.restarting)
AssertionError
```

This is just after we've had the browser fail to respond on several occasions. This looks like we aren't cleaning up properly when exiting abnormally.

foolip:
Per @gsnedders this was the cause behind https://github.com/web-platform-tests/wpt/pull/17219#issuecomment-501272482 so perhaps this issue is more important than we thought :)

@LukeZielinski it seems that a lot of issues around TestRunnerManager might be incidentally fixed by refactoring the executor to live in another process and communicate over some protocol. Would you guess this is one of them?

LukeZielinski:
Yeah, seems likely that this would have to be rewritten if those were split into different processes. But we'd likely need to handle such a scenario (abnormal exit + cleanup) in the split world too.

foolip:
Changing to roadmap since @gsnedders says this a bug in the handling of the state machine, fix not obvious but it's affecting runs. @jgraham might have ideas.

Not assigning to anybody though, yet.

gsnedders:
Fundamentally, `TestRunnerManager.error` will always throw because it calls `restart_runner` which asserts we're in the restart state. I'm not entirely sure if the right fix is just:

```
diff --git a/tools/wptrunner/wptrunner/testrunner.py b/tools/wptrunner/wptrunner/testrunner.py
[...]

PR #13277, opened 2018-09-30
Enable WebDriver executor by default for Chrome

### LINKED ISSUE #13225: Re-enable ExecutorWebDriver in Chrome
#13200 enabled it, but this was reverted in #13220.

See [wpt-chrome-dev-testharness-5](https://tools.taskcluster.net/groups/HuC12DhIShShqh7UcovreA/tasks/J9HbAmKwQSuyfn-3f-TEmw/runs/0/logs/public%2Flogs%2Flive.log) and [wpt-chrome-dev-testharness-6](https://tools.taskcluster.net/groups/HuC12DhIShShqh7UcovreA/tasks/Pn2YWHFYRuWDpl3y1bWEAA/runs/0/logs/public%2Flogs%2Flive.log) logs both of which contain a fair number of tracebacks.

https://staging.wpt.fyi/results/?product=chrome@1e7408fcd3&product=chrome@9bc71d422a&diff=true shows a lot of tests vanishing due to these tracebacks. This suggests there is at least one of the following bugs here:

 * We aren't exiting with non-zero when we fail to run all the tests in the job (because, e.g., of #13073)
 * We aren't setting results to ERROR/CRASH in some cases

@jgraham pointed out that in many implementations when the content process crashes the WebDriver "execute script" command just returns `null`, and this is indistinguishable from `wd.execute_script("return null")`.

gsnedders:
Fixes #13225.

Note this doesn't deal with all the exceptions we got last time, notably it doesn't solve #13276, but that's a generic error-case cleanup failure I believe.


foolip (approved):
Alright, let's merge and compare before/after results from Taskcluster.

gsnedders:
Let's just go ahead and try this again.

foolip:
Good, we have results for the immediately preceding commit:
https://wpt.fyi/results/?sha=f95086e8eb

When https://tools.taskcluster.net/groups/EIuVf04JSsOCsg2xd7ufaQ is done, hopefully results are identical :)

foolip:
@gsnedders can you list each test with changes and why those changes are probably OK?

https://staging.wpt.fyi/results/storage/storagemanager-persisted.https.any.html?product=chrome@f95086e8eb&product=chrome@caf4aa367e&diff=true is a change that seems a bit worrying, but I didn't check them all.

gsnedders:
> @gsnedders can you list each test with changes and why those changes are probably OK?

It's almost entirely flake, if you compare against more recent builds (e.g., https://staging.wpt.fyi/results/?product=chrome@f95086e8eb&product=chrome@aab176b97b&diff=true).

> 
> https://staging.wpt.fyi/results/storage/storagemanager-persisted.https.any.html?product=chrome@f95086e8eb&product=chrome@caf4aa367e&diff=true is a change that seems a bit worrying, but I didn't check them all.

That's just flake (it goes back to passing in later builds).

foolip:
I see at least two directories where the number doesn't change. Can you look into those?

gsnedders:
* WebCryptoAPI/: this is some `testharness.js` tests running to completion but returning fewer results, and it seems very unlikely to be executor related
* offscreen-canvas/: this has some more tests that run to completion than before; I don't know why they've started completing now.
* web-animations/: this is just timing related in the test.
* service-workers: again, tests running to completion rather than ERRORing.

Anything else?

foolip:
The only other I can see is https://staging.wpt.fyi/results/paint-timing/first-image-child.html?product=chrome@f95086e8eb&product=chrome@aab176b97b&diff=true which is now timing out, but that wasn't in the original diff so probably also flaky.

Looks like we should be good, then!

@jugglinmike, please shout if this causes trouble for Buildbot results collection.

PR #13285, opened 2018-10-01
Replace setTimeout with step_timeout

plehegar:
Minor change on a manual test, to promote best practices

### DIFF
--- a/pointerevents/pointerevent_boundary_events_at_implicit_release_hoverable_pointers-manual.html
+++ b/pointerevents/pointerevent_boundary_events_at_implicit_release_hoverable_pointers-manual.html
@@ -43,9 +43,9 @@
             event_log.push(event.type + '@' + event.target.id);
             if (event.type == 'lostpointercapture') {
               test_pointer_event.step_timeout(function () {
-                  var expected_events = "pointerup, lostpointercapture, pointerout, pointerleave";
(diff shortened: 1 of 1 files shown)


pointerevents/pointerevent_boundary_events_at_implicit_release_hoverable_pointers-manual.html
```diff
@@ -42,12 +42,10 @@
             detected_pointertypes[event.pointerType] = true;
             event_log.push(event.type + '@' + event.target.id);
             if (event.type == 'lostpointercapture') {
-              setTimeout(function() {
-                test_pointer_event.step(function () {
+              test_pointer_event.step_timeout(function () {
                   var expected_events = "pointerup, lostpointercapture, pointerout, pointerleave";
```
Steditor: I think we should then adjust the indentation in 46 and 47.

plehegar: good catch. fixed.

Steditor (changes requested):
Currently this results in `TypeError: test_pointer_event.step_timeout is not a function` for me - and as far as I can see, `setup_pointerevent_test` returns a `MultiPointerTypeTest` from `pointerevent_support.js`, which is never equipped with a `step_timeout`-function.

plehegar:
you're right. this is more work than worth it to remove that one setTimeout, and for a manual test. abandoning this.

PR #13292, opened 2018-10-01
Test for <form> in quirks mode should have margin-block-end: 1em

zcorpan:
WebKit bug https://bugs.webkit.org/show_bug.cgi?id=157788


zcorpan:
Moved to https://github.com/web-platform-tests/wpt/pull/13293

PR #13300, opened 2018-10-01
Give all Travis jobs human-readable names

foolip:
Also make travis.yml more consistent around `os` and `python`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,7 +76,7 @@ matrix:
       os: linux
       python: "2.7"
       env: JOB=resources_unittest TOXENV=py27 SCRIPT=tools/ci/ci_resources_unittest.sh
-    - name: "flake8"
+    - name: "tools/wpt/ unittests"
       os: linux
       python: "2.7"
       addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,7 @@ matrix:
       os: linux
       python: "3.6"
       env: JOB=tools_unittest TOXENV=py36 HYPOTHESIS_PROFILE=ci SCRIPT=tools/ci/ci_tools_unittest.sh
-    - name: "resources/ unittests"
+    - name: "resources/ tests"
       os: linux
       python: "2.7"
       env: JOB=resources_unittest TOXENV=py27 SCRIPT=tools/ci/ci_resources_unittest.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ install:
 matrix:
   fast_finish: true
   include:
-    - name: "manifest upload (no-op for PR)"
+    - name: "tag master + upload manifest"
       os: linux
       python: "2.7"
       env:
@@ -72,10 +72,6 @@ matrix:
       os: linux
       python: "3.6"
       env: JOB=tools_unittest TOXENV=py36 HYPOTHESIS_PROFILE=ci SCRIPT=tools/ci/ci_tools_unittest.sh
-    - name: "resources/ tests"
-      os: linux
-      python: "2.7"
-      env: JOB=resources_unittest TOXENV=py27 SCRIPT=tools/ci/ci_resources_unittest.sh
     - name: "tools/wpt/ unittests"
       os: linux
       python: "2.7"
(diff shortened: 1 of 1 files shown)


foolip:
I did this because I've been unable to understand what some of the jobs are from a view like https://travis-ci.org/web-platform-tests/wpt/builds/435826707.

https://travis-ci.org/web-platform-tests/wpt/builds/435843332 shows what this would look like.

.travis.yml
```diff
@@ -20,7 +20,8 @@ install:
 matrix:
   fast_finish: true
   include:
-    - os: linux
+    - name: "manifest upload"
```
foolip: Should I add "(no-op for PR)" or something to this description?

.travis.yml
```diff
@@ -58,19 +64,28 @@ matrix:
...
+    - name: "flake8"
+      os: linux
+      python: "2.7"
       addons:
         apt:
           packages:
             - libnss3-tools
       env: JOB=wpt_integration TOXENV=py27,py27-flake8 SCRIPT=tools/ci/ci_wpt.sh
-    - os: linux
+    - name: "wptrunner infrastructure"
```
foolip: Is this a good name? Are these unit tests, or what?

jgraham (approved):
Oh, I didn't know about this.

.travis.yml
jgraham: This runs pytest in `tools/wpt`

foolip: Hmm, that doesn't seem quite right. The only place pytest appears in tools/ci is in tools/ci/tests/test_jobs.py, and this job will run tools/ci/ci_wptrunner_infrastructure.sh. Looks like it will run infrastructure/ in chrome and firefox, using in-tree expectations. Am I reading that right?

If so, maybe call it "infrastructure/ tests"? I'll throw on a slash on some of the other names too for consistency.

.travis.yml
```diff
@@ -58,19 +64,28 @@ matrix:
...
+      os: linux
+      python: "3.6"
       env: JOB=tools_unittest TOXENV=py36 HYPOTHESIS_PROFILE=ci SCRIPT=tools/ci/ci_tools_unittest.sh
-    - python: 2.7
+    - name: "resources unittests"
+      os: linux
+      python: "2.7"
       env: JOB=resources_unittest TOXENV=py27 SCRIPT=tools/ci/ci_resources_unittest.sh
-    - python: 2.7
+    - name: "flake8"
```
foolip: I'm now pretty sure I've misnamed this one. This does involve `tools/wpt` and is probably what @jgraham referred to in https://github.com/web-platform-tests/wpt/pull/13300/files#r221878898. This does run flake8, but also other things it seems. @gsnedders, what should it be called? "tools/wpt/ pytest + flake8"?

.travis.yml
foolip: I'll go with that.

.travis.yml
Ms2ger: "wpt unittests", for consistency with all the other unittests jobs? (Or "wpi cli unittests", if that's unclear?)

foolip: I'll go with tools/wpt/ unittests, if they are indeed accurate described as unit tests.

foolip:
Does anyone know why we run flake8 the way we do, and not as part of the lint, say? Would it be reasonable to restructure the jobs to run flake8 as part of the lint job? (I'd just file a follow-up issue, not here.)

foolip:
So, it seems like we run flake8 as part of 3 tox.ini files, run in 2 different jobs:
tools/tox.ini (run by tools/ci/ci_tools_unittest.sh)
tools/wpt/tox.ini (run by ci_wpt.sh)
tools/wptrunner/tox.ini (run by tools/ci/ci_tools_unittest.sh)

That's kind of weird, but I'll just leave it alone in this PR.

However, poking around I noticed that "resources/ unittests" is not pedantically correct, as resources/test/tests/ has both functional and unit subdirectories. I'll just call it "resources/ tests".

foolip:
OK, I've tinkered a bit more, changed the order to make it read better. Enough tinkering, now merging. Please shout at me if you don't like the names :)

PR #13302, opened 2018-10-02
These are reftests of vector effects of svg 2 other than non scaling stroke.

satakagi:
This is a pull request that solved issues pointed out on https://github.com/web-platform-tests/wpt/pull/12712

### DIFF
--- a/svg/coords/vectorEffects/reftests/ve-fixedPosition1.svg
+++ b/svg/coords/vectorEffects/reftests/ve-fixedPosition1.svg
@@ -1,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:html="http://www.w3.org/1999/xhtml" width="1000" height="600" viewBox="0,0,1000,600">
+    <g id="testmeta">
+        <title>Vector Effects fixedPosition - 1</title>
+        <html:link rel="author"
+            title="Satoru Takagi"
+            href="mailto:sa-takagi@kddi.com"/>
+        <html:link rel="help"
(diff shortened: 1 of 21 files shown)


svg/coords/vectorEffects/reftests/ve-fixedPosition1.svg
```diff
@@ -0,0 +1,67 @@
...
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:html="http://www.w3.org/1999/xhtml" width="1000" height="600" viewBox="0,0,1000,600">
+    <g id="testmeta">
+        <title>Vector Effects fixedPosition - 1</title>
+        <html:link rel="author"
+            title="Satoru Takagi"
+            href="mailto:sa-takagi@kddi.com"/>
+        <html:link rel="help"
+            href="https://svgwg.org/svg2-draft/coords.html#VectorEffects"/>
+        <html:link rel="match"  href="ve-fixedPosition1-ref.svg" />
+        <metadata class="flags">TOKENS</metadata>
```
svgeesus:
```suggestion
        
```
Either use specific tokens for the flags or, if none are needed, remove the metadata element

svg/coords/vectorEffects/reftests/ve-fixedPosition1.svg
```diff
@@ -0,0 +1,67 @@
...
+        <desc class="assert">TEST ASSERTION</desc>
```
svgeesus: Similarly, either remove the assertion or (better) describe in prose what specifically each test is testing. In this case the title is pretty descriptive so I would just remove the assertion.

svgeesus (changes requested):
Apart from the flags and assertion comments (which apply to all the tests) these look good to me.

svgeesus (approved):

PR #13317, opened 2018-10-02
XHR: enforce particular header line serialization

### LINKED ISSUE #7567: setrequestheader-content-type.htm is erroneously sending back an extra blank space back to the browser for empty headers.
This web platform test is checking that an empty Content-Type header is being sent by the UA, which both Firefox and Edge are actually doing. However, the test fails because it's returning a responseText with an extra space.

That is, when I have the test's python script `XMLHttpRequest/resources/inspect-headers.py` print out the `raw_headers.headers` it actually receives, the results are what the test expects: 'Content-Type: \r\n'

However, when I ask that same script to spit out the byte sequence (in decimal) it's going to pass back as the responseText it will be checking on the client-side, it gives:  67, 111, 110, 116, 101, 110, 116, 45, 84, 121, 112, 101, 58, 32, 32, 10

Note that extra space at the end. The test is accidentally fooling itself into thinking that Firefox (and probably Edge as well) are failing.

annevk:
See https://github.com/httpwg/http-core/issues/137 about getting this properly standardized.

Closes #7567.

### DIFF
--- a/xhr/resources/inspect-headers.py
+++ b/xhr/resources/inspect-headers.py
@@ -5,11 +5,9 @@ def get_response(raw_headers, filter_value, filter_name):
             return "Syntax error: missing CRLF: " + line
         line = line[:-2]
 
-        if ':' not in line:
-            return "Syntax error: no colon found: " + line
-        name, value = line.split(':', 1)
-        if len(value) > 1 and value[0] == ' ':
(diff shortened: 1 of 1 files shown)


wisniewskit (approved):
Looks good to me, thanks!

PR #13335, opened 2018-10-03
Fix import of serve.serve in servo runner

Manishearth:
This fixes `./wpt run servo <..> --install-browser`

### DIFF
--- a/tools/wptrunner/wptrunner/executors/executorservo.py
+++ b/tools/wptrunner/wptrunner/executors/executorservo.py
@@ -8,7 +8,7 @@ import uuid
 
 from mozprocess import ProcessHandler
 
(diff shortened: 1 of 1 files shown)


boazsender:
Confirming that this is a bug for me, and that this patch fixes it.

Thanks @Manishearth!

jgraham (approved):

jgraham:
Out of interest, did you verify that this fix also works in upstream servo? I wonder how it ever worked there if the previous value was just wrong.

Manishearth:
I made this change within servo's vendoring first, and was able to `./mach test-wpt webaudio` from the top level just fine. I guess the paths get set up differently there causing this bug.

PR #13337, opened 2018-10-03
Python 3: Encode text content in Response.iter_content().

Ms2ger:


foolip (approved):
Looks safe, we'll see :)

If you make these tests fail, do they fail in Travis, i.e. are we running the right tests with Python 3 do make this not regress?

PR #13339, opened 2018-10-03
Switch WebKit browser product to WebDriver executors

zdobersek:
Switch away from Selenium executors to WebDriverTestharnessExecutor and
WebDriverRefTestExecutor for the WebKit browser product.

For the moment this requires manually mapping Selenium-style capability names
to the W3C equivalents that can be handled by WebKit's WebDriver implementation.
The browserVersion value is also hard-coded to the 2.20 release series for the
GTK+ port of WebKit as that's when the WebDriver functionality was introduced.

### DIFF
--- a/tools/wptrunner/wptrunner/browsers/webkit.py
+++ b/tools/wptrunner/wptrunner/browsers/webkit.py
@@ -50,8 +50,8 @@ def capabilities_for_port(server_config, **kwargs):
         if capabilities.has_key("version"):
             capabilities["browserVersion"] = capabilities.pop("version")
 
-        # Force the 2.x release series to be used as the required version.
-        capabilities["browserVersion"] = "2.0.0"
+        # Force the 2.20 release series to be used as the required version.
+        capabilities["browserVersion"] = "2.20"
 
         return capabilities
 
--- a/tools/wptrunner/wptrunner/browsers/webkit.py
+++ b/tools/wptrunner/wptrunner/browsers/webkit.py
@@ -31,27 +31,20 @@ def browser_kwargs(test_type, run_info_data, config, **kwargs):
 
 
 def capabilities_for_port(server_config, **kwargs):
-    from selenium.webdriver import DesiredCapabilities
-
     if kwargs["webkit_port"] == "gtk":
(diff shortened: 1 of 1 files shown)


gsnedders:
What testing have you done of this? Have you run at least `/infrastructure` and checked for regressions? (Or, better still, run everything?)

tools/wptrunner/wptrunner/browsers/webkit.py
```diff
@@ -43,6 +43,16 @@ def capabilities_for_port(server_config, **kwargs):
                  "certificateFile": kwargs["host_cert_path"]}
             ]
         }
+
+        # Manually map Selenium's key names to the corresponding W3C versions.
```
jgraham: Couldn't you just statically define a set of capabilities rather than still depending on the selenium defaults? Looks like the [DesiredCapabilites.WEBKITGTK capabilities](https://github.com/SeleniumHQ/selenium/blob/07a18746ff756e90fd79ef253a328bd7dfa9e6dc/py/selenium/webdriver/common/desired_capabilities.py#L124) are rather simple anyway:

jgraham (changes requested):
I like the idea here, but it would be good to remove the dependency on Selenium entirely.

zdobersek:
A small set of tests is run currently. Nothing changes there.

Things in `/infrastructure` also remain without big changes. Before:
```
web-platform-test
~~~~~~~~~~~~~~~~~
Ran 114 checks (46 tests, 68 subtests)
Expected results: 84
Unexpected results: 30
  test: 8 (4 fail, 3 timeout, 1 error)
  subtest: 22 (20 fail, 1 notrun, 1 error)
```

Now:
```
web-platform-test
~~~~~~~~~~~~~~~~~
Ran 114 checks (46 tests, 68 subtests)
Expected results: 85
Unexpected results: 29
  test: 8 (4 fail, 1 crash, 3 timeout)
  subtest: 21 (20 fail, 1 notrun)
```

tools/wptrunner/wptrunner/browsers/webkit.py
zdobersek: Yeah I'll do that.

jgraham (approved):

PR #13365, opened 2018-10-04
XHR: no headers received event for a network error

### LINKED ISSUE #8049: XMLHttpRequest/event-error-order.sub.html is incorrect per spec (and browser-compat).
The test expects a "headers received" event, but I see no way for that to be fired given the spec-text.
The only place it would be fired is in the "process response" steps of the `send()` method, and because the response is an error we will never reach that step.

According to the [wpt.fyi](https://wpt.fyi/results/xhr/event-error-order.sub.html?label=stable&aligned=true), Firefox is currently firing the event, but Edge, Blink, and Safari do not (I confirmed manually in Blink and Safari).

annevk:
Closes #8049.

### DIFF
--- a/xhr/event-error-order.sub.html
+++ b/xhr/event-error-order.sub.html
@@ -22,7 +22,7 @@
             xhr.addEventListener("loadend", function() {
                 test.step(function() {
                     // no progress events due to CORS failure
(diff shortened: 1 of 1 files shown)


annevk:
Browsers still fail a number of things here so probably need to file bugs...

wisniewskit (approved):
Seems fine to me.

annevk:
Bugs:
* https://bugs.chromium.org/p/chromium/issues/detail?id=945095
* https://bugzilla.mozilla.org/show_bug.cgi?id=1538456
* https://bugs.webkit.org/show_bug.cgi?id=196174

PR #13367, opened 2018-10-04
Refactor the testharness/testdriver messaging protocol implementation

jgraham:
This causes most of the code to be run once when the runner window is
opened rather than being run on each resume. This should be a little
more efficient. However the main motivation is to have a single
long-lived event handler rather than having to ensure that the event
handler is correctly cleaned up after each test. Previously this was
not done carefully and it was possible for events from one test to
affect subsequent tests.


tools/wptrunner/wptrunner/executors/runner.js
```diff
@@ -0,0 +1,59 @@
...
+     is present we are able to handle the next event and when is is not
+     present we must wait. Therefore to drive the event processing, this
+     function must be called in two circumstances:
+       * Every time there is a new event that we may be able to handle
+       * Every time we set the callback function
+     This function unsets the callback, so no further testdriver actions
+     will be run until it is reset, which wptrunner does after it has
+     completed handling the current action.
+   */
+  if(!window.testdriver_callback) {
```
Ms2ger: Nit: space after `if`

gsnedders (approved):
% @Ms2ger's nit

PR #13368, opened 2018-10-04
Test StorageEvent constructor and initStorageEvent()

foolip:
Follows https://github.com/whatwg/html/pull/4063.


foolip:
A little report on failures in various browsers:
* Chrome+Safari have no required arguments, and `key` isn't nullable.
* Edge has 8 (!) required arguments, arguments `key`, `oldValue` and `newValue` aren't nullable, and `storageArea` is weird because it can take the value `undefined` in the last test.
* Firefox: `url` *is* nullable, otherwise everything passes

The "initStorageEvent with 8 sensible arguments" test passes on all, suggesting that the code path that matters most is actually pretty fine, interop-wise. (Unsurprising.)

webstorage/event_constructor.html
```diff
@@ -9,21 +9,18 @@
     <h1>event_session_Constructor</h1>
     <div id="log"></div>
     <script>
-        test(function() {
-            var t = async_test("storageeventinit test");
+        async_test(function(t) {
```
foolip: This was a drive-by change to make the test more similar event_constructor_eventinit.html. All the business with `dispatchEvent` actually isn't necessary to test the constructor, but I just wanted to get of the extra no-op test.

annevk: This can be a synchronous test as you're not actually waiting for anything. You still need to wrap the callbacks as you've done though.

annevk (approved):
initstorageevent.html could have been .any.js, but this will do.

webstorage/event_constructor.html
foolip: I know, but the whole `dispatchEvent` thing is also unnecessary. This made it consistent with other tests, and I don't want to rewrite them all since it's pretty harmless.

webstorage/event_constructor.html
```diff
@@ -1,32 +1,81 @@
 <!DOCTYPE HTML>
 <html>
  <head>
-  <title>WebStorage Test: StorageEvent - init value</title>
+  <title>WebStorage Test: StorageEvent - constructor</title>
```
foolip: I've run these tests manually. They pass 5/5 in Chrome, Firefox in Safari. In Edge on real hardware and on BrowserStack however, the test seems to crash Edge, saying "This page is having a problem loading". @thejohnjansen FYI, there's probably a real bug here.

foolip: I've narrowed it down to `new StorageEvent('storage', { url: null })`. I'll file a bug.

foolip: I've filed https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/19216299/ with this test case: http://software.hixie.ch/utilities/js/live-dom-viewer/saved/6254

inexorabletash (approved):
looks great, thanks!

foolip:
I've merged the spec change with the intention of merging this right after, but then saw that Travis was failing. Looks like a network flake, so I've restarted.

webstorage/event_initstorageevent.html
```diff
@@ -0,0 +1,72 @@
...
+        test(() => {
+            const event = new StorageEvent('storage');
+            event.initStorageEvent(undefined, undefined, undefined, undefined, undefined, undefined, undefined, undefined);
+            assert_equals(event.type, 'undefined', 'event.type');
+            assert_equals(event.bubbles, false, 'event.bubbles');
+            assert_equals(event.cancelable, false, 'event.cancelable');
+            assert_equals(event.key, null, 'event.key');
+            assert_equals(event.oldValue, null, 'event.oldValue');
+            assert_equals(event.newValue, null, 'event.newValue');
+            assert_equals(event.url, 'undefined', 'event.url');
```
Zirro: @foolip I'm implementing `initStorageEvent()` in jsdom and got an error on this line. Should this be the empty string rather than `'undefined'`, given the default value for `url`?

foolip: Yes, this was a mistake. Only the non-optional argument (type) will be the string 'undefined'. Sending fix.

PR #13383, opened 2018-10-05
Fix typo: 'funtion'.

Ms2ger:


foolip:
Mid-air collision: https://github.com/web-platform-tests/wpt/pull/13382

PR #13385, opened 2018-10-05
[css-text] text-transform must not disturb shaping in arabic

frivoal:
Related to https://github.com/w3c/csswg-drafts/issues/2135


r12a:
I'm not sure why you go to the trouble of left-aligning the text rather than the more natural right-alignment. I don't have any other observations, however.

frivoal:
I did that because I wrote it on a large screen, and opening it my first reaction was "where is the text?" It's a short string, way off to the right, so I though it be nicer to bring it closer. I'll make it less extreme, to make it look more appropriate for arabic.

frivoal:
@r12a Updated in place, sounds be better now.

PR #13389, opened 2018-10-05
[css-text] Control characters should be visible

frivoal:
Tests for https://github.com/w3c/csswg-drafts/issues/1990 and https://github.com/w3c/csswg-drafts/issues/855

### DIFF
--- a/css/css-text/white-space/control-chars-000.html
+++ b/css/css-text/white-space/control-chars-000.html
@@ -1,0 +1,18 @@
+<!doctype html>
+<meta charset=utf-8>
+<title>Control charcters must be visible: U+0000</title>
+<link rel=author title="Florian Rivoal" href="https://florian.rivoal.net">
+<link rel=help href="https://drafts.csswg.org/css-text-3/#white-space-processing">
+<link rel=mismatch href="reference/control-chars-000-ref.html">
+<meta name=flags content="">
+<meta name=assert content="U+0000, which is in the unicode category CC, must be visible">
+<style>
+div {
+  font-size: 4em;
+}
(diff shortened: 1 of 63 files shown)


FremyCompany (approved):

foolip:
@frivoal, the control-chars-00D.html test is failing in both Chrome Dev and Firefox Nightly:
https://wpt.fyi/results/css/css-text/white-space?sha=434ca47448&label=experimental

No results from Edge or Safari yet, but is that known and intentional?

foolip:
Ah, 0x0D is precisely the topic of https://github.com/w3c/csswg-drafts/issues/855.

I've tested http://web-platform-tests.live/css/css-text/white-space/control-chars-00D.html and it's not visible in Chrome, Edge, Firefox or Safari.

Anyway, I've filed https://crbug.com/893490 to track it.

frivoal:
> is that known and intentional?

Yes. Thanks for filing the bug.

FremyCompany:
I was about to say that the character was visible in the latest Edge builds where our new text engine is enabled... but apparently we "fixed" (or "regressed" I guess) this at some point during the release and it no longer shows up. Sigh...

I don't have a very strong opinion but I think being consistent with our general decision with control characters makes sense, and we agreed to go that road at the last meeting, maybe we should stick with this decision except if we find a compelling reason to switch back.

frivoal:
I think consistency is the way to go. This is not a pressing issue, since it only applies to U+000D injected by script or CSS (those in the markup get converted by the HTML parser), but given some time to fix it, one less oddity in the platform sounds nice.

eaenet:
Agreed, consistency here seems like the best way forward.

Issue #13024, opened 2018-09-17
Update affected tests heuristic to handle changes in interfaces/*.idl


foolip:
https://github.com/web-platform-tests/wpt/pull/13015 is an example of a PR that changed an IDL file, but where no test was taken to be affected. The [Taskcluster log](https://taskcluster-artifacts.net/B4EejmHXSwiaz9GPkLlGZg/0/public/logs/live_backing.log) said:
```
2018-09-16 06:12:40,213 - tc-run - INFO - Identifying tests affected in range 'HEAD^'...
2018-09-16 06:12:50,278 - tc-run - INFO - Identified 0 affected tests
2018-09-16 06:12:50,278 - tc-run - INFO - Quitting because no tests were affected.
```

This needs to be fixed to avoid manually testing all PRs created from reffy-reports.

PR #13392, opened 2018-10-05
Find affected idlharness tests

lukebjerring:
Fixes https://github.com/web-platform-tests/wpt/issues/13024

Example output after tweaking `css-font-loading.idl` locally, and running `./wpt tests-affected --modified HEAD..master`

> INFO:manifest:Updating manifest
> css/css-font-loading/idlharness.https.html

### DIFF
--- a/tools/wpt/testfiles.py
+++ b/tools/wpt/testfiles.py
@@ -265,7 +265,7 @@ def affected_testfiles(files_changed, skip_tests, manifest_path=None):
             if 'idlharness.js' in file_contents:
                 for interface in interfaces_changed_names:
                     regex = '[\'"]' + interface + '(\\.idl)?[\'"]'
-                    if re.match(regex, file_contents):
+                    if re.search(regex, file_contents):
                         affected_testfiles.add(test_full_path)
                         break
 
--- a/tools/wpt/testfiles.py
+++ b/tools/wpt/testfiles.py
@@ -219,6 +219,11 @@ def affected_testfiles(files_changed, skip_tests, manifest_path=None):
     test_files = {os.path.join(wpt_root, path)
                   for _, path, _ in wpt_manifest.itertypes(*test_types)}
 
+    interface_dir = os.path.join(wpt_root, 'interfaces')
+    interfaces_files = {os.path.join(wpt_root, 'interfaces', filename)
+                        for filename in os.listdir(interface_dir)}
+
+    interfaces_changed = interfaces_files.intersection(nontests_changed)
     nontests_changed = nontests_changed.intersection(support_files)
 
     tests_changed = set(item for item in files_changed if item in test_files)
@@ -237,6 +242,9 @@ def affected_testfiles(files_changed, skip_tests, manifest_path=None):
             full_path = os.path.join(wpt_root, repo_path[1:].replace("/", os.path.sep))
         nontest_changed_paths.add((full_path, repo_path))
 
+    interface_name = lambda x: os.path.splitext(os.path.basename(x))[0]
+    interfaces_changed_names = map(interface_name, interfaces_changed)
+
     def affected_by_wdspec(test):
         affected = False
         if test in wdspec_test_files:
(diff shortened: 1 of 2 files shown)


foolip:
@lukebjerring Travis is still not pleased with these changes, should I review anyway?

lukebjerring:
Yeah I can fix up the tests asynchronously, the heuristic can already be assessed on it's merits.

tools/wpt/testfiles.py
```diff
@@ -237,6 +242,9 @@ def affected_testfiles(files_changed, skip_tests, manifest_path=None):
             full_path = os.path.join(wpt_root, repo_path[1:].replace("/", os.path.sep))
         nontest_changed_paths.add((full_path, repo_path))
 
+    interface_name = lambda x: re.match('.*/([^/]*)\.idl$', x).group(1)
```
foolip: `os.path.splitext(os.path.basename(x))[0]` would be clearer here I think. Possibly not exactly the same for every input, but I can't think of one where it would matter.

tools/wpt/testfiles.py
```diff
@@ -252,6 +260,14 @@ def affected_by_wdspec(test):
                     break
         return affected
 
+    def affected_by_interfaces(file_contents):
+        if 'idlharness.js' in file_contents:
```
foolip: If you compare the time taken to compute affected tests before/after this, is there a difference? Won't this look through the contents of all files? If so, can you exit early if `interfaces_changed_names` is empty?

tools/wpt/testfiles.py
```diff
@@ -252,6 +260,14 @@ def affected_by_wdspec(test):
...
+            for interface in interfaces_changed_names:
+                if '\'' + interface + '\'' in file_contents \
```
foolip: This will work for `idl_test`, but tests still fetching /interfaces/something.idl would not be considered affected. Maybe also look for `interface + '.idl'`?

tools/wpt/testfiles.py
lukebjerring: Testing the new hotness:

```suggestion
    interface_name = os.path.splitext(os.path.basename(x))[0]
```

lukebjerring: Neeeeeat! You should use that next time @foolip (hit the little `add suggestion` button)

tools/wpt/testfiles.py
lukebjerring: Done.

tools/wpt/testfiles.py
foolip: What, what's this new feature?! I will use it!

foolip: Hmm, I don't see it. Maybe it's an A/B test? Where can I find it?

tools/wpt/testfiles.py
```diff
@@ -252,6 +260,15 @@ def affected_by_wdspec(test):
                     break
         return affected
 
+    def affected_by_interfaces(file_contents):
+        if len(interfaces_changed_names) > 0:
+            if 'idlharness.js' in file_contents:
+                for interface in interfaces_changed_names:
```
foolip: Old comment was resolved by changing indent, so again: This will work for `idl_test`, but tests still fetching /interfaces/something.idl would not be considered affected. Maybe also look for `interface + '.idl'`?

foolip (approved):
LGTM % remaining comment

tools/wpt/testfiles.py
```diff
@@ -252,6 +260,15 @@ def affected_by_wdspec(test):
...
+                    regex = '[\'"]' + interface + '(\\.idl)?[\'"]'
```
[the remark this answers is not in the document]
lukebjerring: image

It's the little +/- icon... maybe I'm in an A/B test?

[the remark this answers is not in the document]
lukebjerring: I'm actually gonna veto the `r''` suggestion here because I still have to escape the single-quote, and the only gain is the double-backslash of the `\\.idl` which is not very hard to read.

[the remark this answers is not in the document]
foolip: I think that @Hexcles might veto your veto, but I'm happy to back down :)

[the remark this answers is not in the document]
foolip: Actually, I see the icon too, so I'll use that next time.

Issue #13407, opened 2018-10-07
`Command left in command_queue during cleanup: u'runner_teardown', ()` warning always emitted from `./wpt run`


foolip:
Taking https://github.com/web-platform-tests/wpt/pull/13405 as a recent example, you can find `WARNING u'runner_teardown': ()` is the logs of everything that involves `./wpt run`:
https://travis-ci.org/web-platform-tests/wpt/builds/438107535 (the stability jobs)
https://tools.taskcluster.net/groups/aPHmZD1kRSGHHPHE8S1iTg (all 4 results/stability tasks)

The source must be from here:
https://github.com/web-platform-tests/wpt/blob/08d0c7d0de6c135b8f11364f6d0175981524f530/tools/wptrunner/wptrunner/testrunner.py#L78

It appears the message doesn't go where it's supposed to go, and probably does nothing except generate this warning.

Edit: https://github.com/web-platform-tests/wpt/pull/13408 changed the warning message.

foolip:
This is how the string ends up emitted as a warning:
https://github.com/web-platform-tests/wpt/blob/08d0c7d0de6c135b8f11364f6d0175981524f530/tools/wptrunner/wptrunner/testrunner.py#L720

foolip:
With https://github.com/web-platform-tests/wpt/pull/13408 merged, the warning will be `WARNING Command left in command_queue during cleanup: u'runner_teardown', ()` instead.

Issue #13274, opened 2018-09-30
mozdownload sometimes fails on Taskcluster and Travis


foolip:
At least twice some of the Firefox tasks for pushes to master have failed in mozdownload:

https://tools.taskcluster.net/groups/AGupAeh6TrSdyNlDxcqZXw (for commit 2df7f9ff620cbdaa2928464892fb1dfb880fd6c6, now only discoverable via [API](https://api.github.com/repos/web-platform-tests/wpt/commits/2df7f9ff620cbdaa2928464892fb1dfb880fd6c6/statuses))
https://tools.taskcluster.net/groups/eRtqwYxjTfeob0g4hHcOlw (for commit 91491deb7dcfe4a0d0ece61f086345abad4d46a4)

The most recent failure was:
```
Traceback (most recent call last):
  File "./wpt", line 5, in <module>
    wpt.main()
  File "/home/test/web-platform-tests/tools/wpt/wpt.py", line 129, in main
    rv = script(*args, **kwargs)
  File "/home/test/web-platform-tests/tools/wpt/run.py", line 510, in run
    **kwargs)
  File "/home/test/web-platform-tests/tools/wpt/run.py", line 488, in setup_wptrunner
    kwargs["binary"] = setup_cls.install(venv, channel=channel)
  File "/home/test/web-platform-tests/tools/wpt/run.py", line 165, in install
    return self.browser.install(venv.path, channel)
  File "/home/test/web-platform-tests/tools/wpt/browser.py", line 134, in install
    destination=dest).download()
  File "/home/test/web-platform-tests/_venv/lib/python2.7/site-packages/mozdownload/factory.py", line 121, in __init__
    scraper_types[scraper_type].__init__(self, **kwargs)
  File "/home/test/web-platform-tests/_venv/lib/python2.7/site-packages/mozdownload/scraper.py", line 346, in __init__
    Scraper.__init__(self, *args, **kwargs)
  File "/home/test/web-platform-tests/_venv/lib/python2.7/site-packages/mozdownload/scraper.py", line 135, in __init__
    self._retry_check_404(self.get_build_info)
  File "/home/test/web-platform-tests/_venv/lib/python2.7/site-packages/mozdownload/scraper.py", line 150, in _retry_check_404
    self._retry(func, **retry_kwargs)
  File "/home/test/web-platform-tests/_venv/lib/python2.7/site-packages/mozdownload/scraper.py", line 141, in _retry
[...]

PR #13409, opened 2018-10-07
Remove runner_teardown message/method in TestRunner/Manager

foolip:
The relationship between `TestRunnerManager` and `TestRunner` is
described in //tools/wptrunner/docs/design.rst.

In a new process (in `start_runner`) a `TestRunner` instance is created
as `with TestRunner(...) as runner`, and leaving the scope of this
`with` statement is what causes `TestRunner. __exit__` and `teardown()`
to be invoked, sending the "runner_teardown" message. At this point,
`start_runner` is just about to return and the process exit.

There was originally a warning around this which was silenced:
https://github.com/web-platform-tests/wpt/issues/13407
https://github.com/web-platform-tests/wpt/pull/13417

This reverts https://github.com/web-platform-tests/wpt/pull/13417.

The warning occured when the `run` loop had already ended and the
message was still in the queue during `TestRunnerManager.cleanup`.
(See https://github.com/web-platform-tests/wpt/issues/13413 on naming.)

If the message arrived before the cleanup stage, the `runner_teardown`
handler would call `ensure_runner_stopped()` and change runner manager
state to stopped in the `run` loop, which is what would happen anyway.

### DIFF
--- a/tools/wptrunner/wptrunner/testrunner.py
+++ b/tools/wptrunner/wptrunner/testrunner.py
@@ -75,7 +75,6 @@ class TestRunner(object):
 
     def teardown(self):
         self.executor.teardown()
-        self.send_message("runner_teardown")
         self.result_queue = None
         self.command_queue = None
         self.browser = None
@@ -395,7 +394,6 @@ class TestRunnerManager(threading.Thread):
             RunnerManagerState.error: {},
             RunnerManagerState.stop: {},
             None: {
-                "runner_teardown": self.runner_teardown,
                 "log": self.log,
                 "error": self.error
             }
@@ -697,10 +695,6 @@ class TestRunnerManager(threading.Thread):
         else:
             self.logger.debug("Runner process exited with code %i" % self.test_runner_proc.exitcode)
 
-    def runner_teardown(self):
-        self.ensure_runner_stopped()
-        return RunnerManagerState.stop()
-
     def send_message(self, command, *args):
         self.remote_queue.put((command, args))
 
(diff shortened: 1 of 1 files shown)


foolip:
Review welcome, but I'm not yet entirely confident about this fix, seeing what happens in Taskcluster/Travis is part of the reason for submitting.

foolip:
Alright, I've written an elaborate commit message with my understanding of the situation. @jgraham, review appreciated now.

Note that a more conservative approach would be to just not log a warning if this message arrives late, but given that it doesn't seem to serve a real purpose simplifying the code is my first attempt. At a high level, it seems odd that the subprocess would need to tell the manager that it's done, when the manager already knows when it expects it to be done and also waits for it to be done then.

If my fix is wrong, I suspect it'd manifest as a hung test run.

foolip:
In the case of the `TestRunner` process crashing, this message will never be sent, as it's basically the last thing that happens during the orderly exit of that process. FWIW, I did try to make the process exit with a non-zero exit code while investigating all of this, and found that `TestRunnerManager` doesn't immediately do anything about this, but the `self.test_runner_proc.is_alive()` check will cause it to be slowly restarted over and over until some restart limit is reached.

foolip:
I've rebased this to also remove what was added in https://github.com/web-platform-tests/wpt/pull/13417. Review still appreciated, but now there's just risk and nothing noticeable being fixed.

It seems to me that unlike the overall process isolation scheme, this bit would not be a guard against a misbehaving browser or executor, but rather would be smoothing over bugs in `TestRunnerManager` itself, which somehow does not realize that the `TestRunner` has nothing more to do, or has exited.

If there is a case where this code is operative and necessary, then I suspect it could be replaced by checking if the process is alive in more places, as this message will only be sent when the process is just about to exit. It's *possible* that Python could somehow get stuck in the final cleanup, but then that'd get seen as a hung process, which it is.

gsnedders:
Travis failure due to #13274.

foolip:
@jgraham would you like me to close this PR, or is it sitting in your low-prio queue?

jgraham:
Let's just close it; I think you're basically right that the design here should be refactored but I don't think that just deleting the ability for the runner to notify that it's shutting down is the right thing.

foolip:
Alright, leaving it alone :) Is there an issue tracking some refactoring that would make this code more approachable, or do we just hope that we almost never have to make changes here?

PR #13411, opened 2018-10-07
Expect InvalidStateError instead of InvalidAccessError in document.open

dguenther:
While working on https://github.com/servo/servo/pull/21882, I noticed that the [spec for `document.open(url, name, features)`](https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#dom-document-open-window) states that it should throw an InvalidStateError, but tests expect an InvalidAccessError. I changed the expectation to align with the spec, but that may not be correct -- here's some additional information:

I believe the error was changed in this PR: https://github.com/whatwg/html/pull/2672

Chrome, Safari, Firefox throw `InvalidAccessError`: https://wpt.fyi/results/html/webappapis/dynamic-markup-insertion/opening-the-input-stream/document.open-02.html

InvalidAccessError is deprecated: https://heycam.github.io/webidl/#invalidaccesserror

### DIFF
--- a/html/webappapis/dynamic-markup-insertion/opening-the-input-stream/document.open-02.html
+++ b/html/webappapis/dynamic-markup-insertion/opening-the-input-stream/document.open-02.html
@@ -20,7 +20,7 @@ test(function() {
   var parser = new DOMParser();
   var doc = parser.parseFromString("", "text/html");
   assert_equals(doc.defaultView, null);
-  assert_throws("INVALID_ACCESS_ERR", function() {
+  assert_throws("INVALID_STATE_ERR", function() {
     doc.open("/resources/testharness.js", "", "");
   });
 }, "document.open should throw when it has no window and is called with three arguments");


foolip:
You've found the right definition in the spec, and per that the spec is wrong. However, given that 3 implementations already throw InvalidAccessError I think it'd be unfortunate to try to change it. Better to change the spec back to use InvalidAccessError I think.

foolip:
Thanks for reporting this, @dguenther! I've sent https://github.com/whatwg/html/pull/4066 to change the spec back, and have a question for you there.

domenic:
I'm surprised we didn't catch this during the course of @TimothyGu's work on this test suite, and I'm surprised this doesn't affect more tests. Would love to hear his thoughts.

domenic:
I realize now why @TimothyGu and I didn't catch this is because we weren't concerned about the (url, name, features) overload.

domenic:
Closing since it was the spec that was wrong, not the tests. Merged https://github.com/whatwg/html/pull/4066 so now they match :).

PR #13412, opened 2018-10-07
Move a comment in TestRunnerManager.wait_finished to make more sense

foolip:
It's certaintly not the logging that does post-stop processing, but
`after_test_end` might, so the comment is accurate if moved.

The comment was added in commit 7f422ef4b1b1ecf09ff7ffc94b60c6b7bf7a41b4.


foolip:
This isn't me being pedantic, I genuinely suspected that the line the comment was for had been removed, before going back to look where the comment came from.

jgraham (approved):

PR #13417, opened 2018-10-08
Ignore "runner_teardown" messages in `TestRunnerManager.cleanup()`

foolip:
This is an alternative to https://github.com/web-platform-tests/wpt/pull/13409,
which adds more handling around this rather than removing it.

Fixes https://github.com/web-platform-tests/wpt/issues/13407.

### DIFF
--- a/tools/wptrunner/wptrunner/testrunner.py
+++ b/tools/wptrunner/wptrunner/testrunner.py
@@ -716,6 +716,11 @@ class TestRunnerManager(threading.Thread):
             else:
                 if cmd == "log":
                     self.log(*data)
+                elif cmd == "runner_teardown":
+                    # It's OK for the "runner_teardown" message to be left in
+                    # the queue during cleanup, as we will already have tried
+                    # to stop the TestRunner in `stop_runner`.
+                    pass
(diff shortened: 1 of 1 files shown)


foolip:
This is an alternative to https://github.com/web-platform-tests/wpt/pull/13409 and I'm OK with merging it to fix https://github.com/web-platform-tests/wpt/issues/13407 ASAP. However, I would still greatly review of https://github.com/web-platform-tests/wpt/pull/13409, as per my deep dive it doesn't seem like the mechanism is doing anything.

foolip:
Looks like the warnings are gone in https://travis-ci.org/web-platform-tests/wpt/jobs/438580296.

jgraham (approved):

gsnedders:
So when exactly is this left in the queue? When will the TestRunner fail to stop?

foolip:
@gsnedders see the description of https://github.com/web-platform-tests/wpt/pull/13409 for the chain of events. I haven't observed `TestRunner` failing to stop, but if it does fail to show any sign of life for long enough `TestRunnerManager` will try to restart it.

PR #13421, opened 2018-10-08
Use conditionals to avoid running no-op jobs on Travis

foolip:
See documentation:
https://docs.travis-ci.com/user/conditional-builds-stages-jobs/
https://docs.travis-ci.com/user/conditions-v1

Note that the `branch` attribute isn't useful since it will be `master`
for PRs with master as the base branch, i.e. almost all PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,9 +18,13 @@ before_install:
 install:
   - ./tools/ci/install.sh
 matrix:
+  # The use of `if` conditionals to exclude jobs from master should align with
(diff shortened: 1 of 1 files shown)


foolip:
Yes indeed, https://docs.travis-ci.com/user/conditional-builds-stages-jobs/ has a pretty bad example using `if: branch = master` for a deploy stage, which will be triggered for PRs that are against the master branch, like this one. So let's *not* use that construct :)

jgraham (approved):

jgraham (approved):

PR #13424, opened 2018-10-08
Limit the use of `sudo: required` to Travis jobs that need it

foolip:
These are the jobs that do `source tools/ci/lib.sh`. Note that the
"stability (Chrome Dev)" job already (redundantly) specified
`sudo: required`.

The hosts addon functionality isn't needed. The hosts list is out of
date, and tools/ci/lib.sh does the hosts fixup when needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ matrix:
   fast_finish: true
   include:
     - name: "tag master + upload manifest"
-      if: type = push AND branch = master
+      if: type = pull_request
       os: linux
       python: "2.7"
       env:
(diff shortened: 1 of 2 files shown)


foolip:
https://travis-ci.org/web-platform-tests/wpt/builds/438648140 can be used to evaluate if this made things faster the the jobs that ran fully in both, which is only lint. There, it's actually slower, not faster.

Even though sudo is no longer in the job config, it seems like it's the same sort of VM/container/whatever in both cases. So probably best to just leave `sudo: required` globally for simplicity.

PR #13425, opened 2018-10-08
Drop the use of the `hosts` addon in Travis

foolip:
The hosts list is out of date and this isn't needed, as
tools/ci/lib.sh does the hosts fixup when needed.
    
Drive-by: remove a redundant `sudo: required`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,14 +4,6 @@ language: python
 branches:
   only:
     - master
-addons:
-  hosts:
-    - web-platform.test
-    - www.web-platform.test
-    - www1.web-platform.test
-    - www2.web-platform.test
-    - xn--n8j6ds53lwwkrqhv28a.web-platform.test
-    - xn--lve-6lad.web-platform.test
 before_install:
   # This needs be sourced as it sets various env vars
   - . ./tools/ci/before_install.sh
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -64,7 +56,6 @@ matrix:
     - name: "stability (Chrome Dev)"
       if: type = pull_request
       os: linux
-      sudo: required
```
gsnedders: Surely this requires sudo to write `/etc/hosts`?

gsnedders: But Chrome needs `/etc/hosts` which `hosts_fixup` will try to write to, no? What description says it's redundant? (I see it manages to write it, so I don't know what description you're referring to here.)

foolip: I mean the description of this PR. That doesn't say *why* it's redundant, I left that implied: because it's already at the top-level scope of `.travis.yml`.

foolip:
From https://travis-ci.org/web-platform-tests/wpt/builds/438685080 it looks like serve/test_functional.py inadvertently depend on /etc/hosts :)

gsnedders:
@jugglinmike this is your test, from 1f17212bcf6/#12557. Let me try and fix this…

gsnedders:
Passing now, yay!

gsnedders (approved):

gsnedders:
Would rather someone else reviewed this given my commit here. Also note it needs rebased to remove the REVERTME commit.

jgraham (approved):

Issue #13183, opened 2018-09-24
Support testdriver.js in reftests


gsnedders:
Previously, when testdriver.js we decided to punt on supporting it in reftests; we now have an actual use for it (#9346), so this needs done.

This will need changes in all the executors: the Python code needs to handle various messages to interact with the document before eventually taking a screenshot, and the JavaScript callbacks needs to deal with passing the testdriver.js messages. The right way to do this is probably by way of comparison of both bits of reftest code with the test harness code.

gsnedders:
Myself and @jgraham have separately on various occasions remembered that this is going to be somewhat awkward for Firefox the internal Marionette reftest runner, given we'll need to reimplement the testdriver.js implementation within Marionette for that.

Perhaps the right solution here is to just put some "testdriver" flag in the manifest and then use the external reftest runner for those tests.

jgraham:
Yeah, that seems acceptable for a v1 implemenation; we would probably want to fix this to work inside the harness later, but that could be a seperate project.

gsnedders:
I'm concerned about the ongoing maintenance cost of having yet another implementation, especially one that wptrunner relies on (and which exists in an external repo). I'm not convinced the benefits of having the internal reftest runner support testdriver.js are actually big enough to outweigh that cost.

foolip:
@gsnedders any chance of getting this fixed before the end of the quarter?

gsnedders:
Is there any reason to believe this is so high priority? Does it affect more tests than other untestable issues?

foolip:
@Hexcles having dug into test writing patterns in Chromium a bit, do you have a hunch for how often this would be useful? I'm inclined to agree with @gsnedders that maybe this isn't very high prio, but don't know.

foolip:
I've downgraded the priority. We'll probably want this to convert manual tests in the future, but at this point the request for it isn't coming from outside, it's just those of us working on wpt infra that think it's a good idea.

Hexcles:
My research only looked at existing tests / test history. And of course, there is zero testdriver usage in reftests as it doesn't work, but I don't know how many people attempted to do that.

gsnedders:
I think there was some discussion at TPAC suggesting this was needed for a whole bunch of tests, but I think it's pretty low priority still.

stephenmcgruer:
[...]

PR #13428, opened 2018-10-09
[css-text] plain-text copy & paste does not preserve text-transform

frivoal:
This is a test for https://github.com/w3c/csswg-drafts/issues/627

### DIFF
--- a/css/css-text/text-transform/text-transform-copy-paste-001-manual.html
+++ b/css/css-text/text-transform/text-transform-copy-paste-001-manual.html
@@ -1,0 +1,31 @@
+<!doctype html>
+<meta charset=utf-8>
+<title>CSS Text 3 test: effects of text-transform on plain text copy&paste</title>
+<link rel=author title="Florian Rivoal" href="https://florian.rivoal.net">
+<link rel=help href="https://drafts.csswg.org/css-text-3/#text-transform-property">
+<meta name=flags content="interact">
+<meta name=assert content="text-transform must not affect the content of a plain text copy and paste operation.">
+<!--
+I wish this test could be written automatedly, but I don't think it can.
+We can work around the fact that  document.execCommand(‘copy’) only works when triggered by user interactions
+by using test_driver.bless,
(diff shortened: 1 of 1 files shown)


css/css-text/text-transform/text-transform-copy-paste-001-manual.html
```diff
@@ -0,0 +1,25 @@
+<!doctype html>
+<meta charset=utf-8>
+<title>CSS Text 3 test: effects of text-transform on plain text copy&paste</title>
+<link rel=author title="Florian Rivoal" href="https://florian.rivoal.net">
+<link rel=help href="https://drafts.csswg.org/css-text-3/#text-transform-property">
+<meta name=flags content="interact">
+<meta name=assert content="text-transform must not affect the content of a plain text copy and paste operation.">
+<!--
+I wish this test could be written automatedly, but I don't think it can,
```
foolip: If you could rewrite this as testharness.js, then you can use testdriver.js and `test_driver.bless()` to automate this. And I think you can, since you just need to read back the value?

Note that @gsnedders will also be working on making testdriver.js work with reftests in https://github.com/web-platform-tests/wpt/issues/13183.

frivoal: Well, `document.execCommand(‘copy’)` would benefit form `test_driver.bless()`, but also it's not quite clear whether it is standard or not. It is documented in https://w3c.github.io/editing/execCommand.html#the-copy-command-0, but that spec has a big "no done & not making progress" warning, and MDN calls it deprecated.

Also, even if you get past that, I then need to check the content of the clipboard, and I don't think here's a workable cross browser way to do that:
* `document.execCommand(‘paste’)` isn't supported in regular web pages by anyone but IE
* The alternative, the Clipboard API, is not supported across the board, and Firefox only supports reading the clipboard in browser extensions.

This may change eventually, but as far as I can tell, a manual test is still all we can do.

PR #13430, opened 2018-10-09
[css-text][css-fonts] Testing writing system dependent rules

frivoal:
Related to https://github.com/w3c/csswg-drafts/issues/2015

### DIFF
--- a/css/css-text/writing-system/reference/writing-system-line-break-002-ref.html
+++ b/css/css-text/writing-system/reference/writing-system-line-break-002-ref.html
@@ -1,0 +1,11 @@
+<!doctype html>
+<html lang=en>
+<meta charset=utf-8>
+<title>CSS test reference</title>
+<link rel=author title="Florian Rivoal" href="https://florian.rivoal.net">
+<style>
+</style>
+
+<p>The test passes if the question mark in the phrase below is on alone on the second line.
+
+<div lang=en-Hrkt>ハロー、ハウアーユー<br>？</div>
--- a/css/css-text/writing-system/reference/writing-system-line-break-002-ref.html
+++ b/css/css-text/writing-system/reference/writing-system-line-break-002-ref.html
@@ -6,6 +6,6 @@
 <style>
 </style>
 
-<p>The test passes if the question mark in the phrase below is on alone on the second line.
+<p>The test passes if the question mark in the phrase below is alone on the second line.
 
 <div lang=en-Hrkt>ハロー、ハウアーユー<br>？</div>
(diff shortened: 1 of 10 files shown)


css/css-text/writing-system/writing-system-font-001.html
```diff
@@ -0,0 +1,25 @@
...
+<html lang=en>
+<meta charset=utf-8>
+<title>CSS test: font selection and writing sytem</title>
+<link rel=author title="Ambrose Li" href="http://port.ambroseli.ca/">
+<link rel=author title="Florian Rivoal" href="https://florian.rivoal.net">
+<link rel=help href="https://drafts.csswg.org/css-text-3/#languages">
+<link rel=help href="https://drafts.csswg.org/css-text-3/#script-tagging">
+<link rel=help href="https://drafts.csswg.org/css-fonts-4/#language-specific-support">
+<link rel=match href="reference/writing-system-font-001-ref.html">
+<meta name=assert content="An explicit writing system takes precedence over the content language for font selection">
```
foolip: This comment doesn't seem to match the test, which asserts (effectively) that `lang=foo` has no effect at all in this case.

frivoal: The test does not assert that `lang=foo` has no effect, but that if you have `lang=foo-bar` where `foo` is a language and `bar` is a writing system, it's `bar` that wins, not `foo`.

So unless I'm missing something, the comment is correct.

foolip: Oh, so the `-latn` bit is the important part. I've just revealed my ignorance about the area, but if nobody else will review, I will :)

css/css-text/writing-system/reference/writing-system-line-break-002-ref.html
```diff
@@ -0,0 +1,11 @@
+<!doctype html>
+<html lang=en>
+<meta charset=utf-8>
+<title>CSS test reference</title>
+<link rel=author title="Florian Rivoal" href="https://florian.rivoal.net">
+<style>
+</style>
+
+<p>The test passes if the question mark in the phrase below is on alone on the second line.
```
foolip:
```suggestion
<p>The test passes if the question mark in the phrase below is alone on the second line.
```

foolip (approved):

frivoal:
> I'll merge as soon as w3c/csswg-drafts#3202 is merged.

That's done, so merging this as well.

foolip:
@frivoal I noticed via https://bugs.chromium.org/p/chromium/issues/detail?id=905629 that my fixup commit ended up in the commit history, which I didn't intend. When merging PRs, please try to use the "squash" option in cases like this. (Nothing really bad happens if you forget, which I've done.)

frivoal:
@foolip Yeah, I try to use squash when the commits are not meaningful, and regular merge (or rebase) when they are, but occasionally I forget. Sorry

PR #13434, opened 2018-10-09
[compat] Manual test for all -webkit-appearance values

zcorpan:
See https://github.com/whatwg/compat/issues/6

---

cc @jwatt @tkent-google


tkent-google:
I'm not sure whether this should be merged to WPT because all-manual.html looks just a file to investigate behavior.  It's not a test.

zcorpan:
Yeah I think you're right. I can put it in a jsfiddle or something.

zcorpan:
http://jsfiddle.net/h09jz2qr/show

PR #13435, opened 2018-10-09
Test Location sans browsing context

annevk:
For https://github.com/whatwg/html/issues/3959.

### DIFF
--- a/html/browsers/history/the-location-interface/no-browsing-context.window.js
+++ b/html/browsers/history/the-location-interface/no-browsing-context.window.js
@@ -49,3 +49,8 @@ test(() => {
     }, "Invoking `" + method + "` with `" + value + "` on a `Location` object sans browsing context is a no-op");
   });
 });
+
+test(() => {
+  const loc = bcLessLocation();
+  assert_array_equals(loc.ancestorOrigins, []);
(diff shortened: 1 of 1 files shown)


foolip (approved):
Reviewed for test structure and general reasonableness of what is being asserted, but please name `.tentative.` if this is merged before the spec change. (Doing that might make it easier to see cross-browser results.)

html/browsers/history/the-location-interface/no-browsing-context.window.js
```diff
@@ -0,0 +1,56 @@
...
+
+function bcLessLocation() {
+  const frame = document.body.appendChild(document.createElement("iframe")),
+        win = frame.contentWindow,
+        loc = win.location;
+  frame.remove();
+  return loc;
+}
+
+[
```
domenic: This is pretty hard to read. It's more work for you but I'd suggest

```js
[
  {
    prop: "href",
    expected: "about:blank",
    valuesToTry: ["https://example.com/", "/", "http://test:test/", "test test", "test:test", "chrome:fail"]
  },
  ...
}
```

domenic (approved):
Tests seem good but I have a readability improvement I'd really like to see if possible :)

PR #13443, opened 2018-10-09
PaymentItem.type member was removed from the spec

marcoscaceres:
Removes tests as part of https://github.com/w3c/payment-request/pull/794


marcoscaceres:
Will send this through Mozilla-central instead.

PR #13460, opened 2018-10-10
Support `./wpt run --channel=preview safari`

foolip:
This mostly to avoid the lengthy argument currently required:
`--webdriver-binary "/Applications/Safari Technology Preview.app/Contents/MacOS/safaridriver"`

Using `--browser-channel=preview` and the product `safari_webdriver`
also works.

### DIFF
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -41,7 +41,7 @@ class Browser(object):
         return NotImplemented
 
     @abstractmethod
-    def find_webdriver(self):
+    def find_webdriver(self, channel=None):
         """Find the binary of the WebDriver."""
         return NotImplemented
 
@@ -211,7 +211,7 @@ class Firefox(Browser):
             return None
         return path
 
-    def find_webdriver(self):
+    def find_webdriver(self, channel=None):
         return find_executable("geckodriver")
 
     def get_version_and_channel(self, binary):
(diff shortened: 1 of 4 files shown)


foolip:
Stern suggestions about how to test this appreciated.

foolip:
(I have tested it locally, and it works!)

tools/wpt/browser.py
```diff
@@ -635,8 +635,11 @@ def install(self, dest=None, channel=None):
     def find_binary(self, venv_path=None, channel=None):
         raise NotImplementedError
 
-    def find_webdriver(self):
-        return find_executable("safaridriver")
+    def find_webdriver(self, channel=None):
+        path = None
+        if channel == "preview":
+            path = "/Applications/Safari Technology Preview.app/Contents/MacOS"
```
jgraham: Bleh. How sure are we that this is the only possible path?

jgraham (approved):

tools/wpt/browser.py
foolip: It's where the dmg is installed and the installer doesn't give any option to put it elsewhere. I think you could well install it elsewhere at the command line. But then you'll have to use `--webdriver-binary`.

jgraham: OK.

PR #13461, opened 2018-10-11
[css-text] Test for text-tranform:full-size-kana

frivoal:
tests for https://github.com/w3c/csswg-drafts/issues/3143

### DIFF
--- a/css/css-text/text-transform/reference/text-transform-full-size-kana-002-ref.html
+++ b/css/css-text/text-transform/reference/text-transform-full-size-kana-002-ref.html
@@ -1,225 +1,26 @@
 <!DOCTYPE html>
-<html lang="en">
-<head>
-<meta charset="utf-8"/>
-<title>CSS3 Text, text transform: full-size-kana, full-size kanas</title>
-<meta name="assert" content="For full-size kanas, text-transform: full-size-kana leaves all kanas unaltered.">
+<meta charset="utf-8">
+<title>CSS Text level 3 Test Reference</title>
+<link rel="author" title="Florian Rivoal" href="https://florian.rivoal.net/">
 <link rel='author' title='Diego Pino Garcia' href='mailto:dpino@igalia.com'>
-<link rel='help' href='https://drafts.csswg.org/css-text-3/#text-transform'>
-<style type='text/css'>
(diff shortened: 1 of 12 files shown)


r12a:
This is just a suggestion: I changed the format for these kinds of tests recently to make it easier for manual reviewers to do the comparisons (and slightly easier to create the ref tests). You can see an example at https://w3c.github.io/i18n-tests/run?base=css-text&batch=text-transform&test=text-transform/text-transform-fullwidth-005.html  Basically, results are superimposed such that divergeances appear in red.  Makes it very quick to spot fails. Just mentioning it in case you think it's a good idea, since your tests are similar.

frivoal:
I considered doing something like this, and decided against because it's really hard to decide if the test is passing of failing when you have sub-pixel or single pixel differences, and just a tiny little bit of red showing. In the full size kana case, a real failure would involve a lot more red than that, but the reviewer does not know that, so when font rendering artifact show a little bit of red, it's not clear.

Case in point, with the test you shared, on my machine, chrome is a clear failure: 
screenshot 2018-10-11 20 51 19

Firefox, on the other hand, is maybe a failure, maybe not. I think it is meant to be a pass, but I do see some red:
screenshot 2018-10-11 20 52 56


A human reviewer would be unsure, and an automated ref checker would fail it, even though it is fine.

r12a:
Yes. I thought a lot about that too, but came to the opposite conclusion ;-).  I didn't think the subpixel shadow would be picked up on the ref test - i could be wrong.  The instructions on some of the other tests i did like this actually mention that subpixel shadows don't count.

frivoal:
To be rebased on top of https://github.com/web-platform-tests/wpt/pull/13504, as discussed in https://bugzilla.mozilla.org/show_bug.cgi?id=1498148#c41

frivoal:
@dpino Rebased on top of your tests from #13504 to try and get the best of both worlds:
* kept your warning about missing glyphs
* kept the test checking that already full-size kana don't get changed
* used by format to make it easier to review by just looking at it
* split into multiple files to fit a single screen
* simplified the markup a bit
* made the test that check that nothing happens on already full-size kana only apply the (ineffective) transform to one character in each pair, as otherwise it isn't really testing anything.

Review appreciated.

PR #13467, opened 2018-10-11
Extract Safari version from safaridriver

foolip:


tools/wpt/browser.py
```diff
@@ -644,10 +643,21 @@ def find_webdriver(self, channel=None):
...
+    def version(self, binary=None, webdriver_binary=None):
+        if webdriver_binary is not None:
+            try:
+                output = call(webdriver_binary, "--version").strip()
+                # TODO: explain
+                if output.startswith('Included with '):
+                    output = output[14:]
+                return output
+            except subprocess.CalledProcessError:
+                # TODO stderr gets printed :'(
```
foolip: I've found that `call` doesn't suppress stderr and just prints it instead. This isn't great because this is expected to fail on every `safaridriver` before STP 67, notable including Safari 12 stable. Should I just change `call` to return a stdout,stderr tuple?

tools/wpt/browser.py
```diff
@@ -644,10 +643,21 @@ def find_webdriver(self, channel=None):
     def install_webdriver(self, dest=None, channel=None):
         raise NotImplementedError
 
-    def version(self, binary):
+    def version(self, binary=None, webdriver_binary=None):
+        if webdriver_binary is not None:
+            try:
+                output = call(webdriver_binary, "--version").strip()
+                # TODO: explain
```
foolip: What needs explaining is why this string begins with 'Included with '.

But more importantly, are there any requirements on the shape of this string? See https://github.com/web-platform-tests/results-collection/issues/617 for the ouput and what it's presented like in UI.

tools/wpt/browser.py
```diff
@@ -644,10 +643,21 @@ def find_webdriver(self, channel=None):
...
+                if output.startswith('Included with '):
+                    output = output[14:]
+                return output
+            except subprocess.CalledProcessError:
+                # TODO stderr gets printed :'(
+                pass
+        # TODO: this message doesn't show up anywhere
```
foolip: I don't know why this, is the logger not wired up correctly?

foolip (commented):
This sort of works but is only half done. Sent out with comments hoping that @jgraham or @jugglinmike might have ideas. For what's left a mystery I will dig in myself :)

tools/wpt/browser.py
jgraham: That's surprising; it looks like it calls `subprocess.checkou_output` with no arguments. I would suggest making it take `**kwargs` and passing those down to the caller so you can redirect stderr to `/dev/null` or whatever.

tools/wpt/browser.py
jgraham: I think the logger handler gets wired up late, but we should probably fix that.

tools/wpt/browser.py
```diff
@@ -364,8 +364,7 @@ def install_geckodriver_nightly(self, dest):
             os.unlink(package_path)
 
         return path
-
-    def version(self, binary=None, channel=None):
+    def version(self, binary=None, webdriver_binary=None, channel=None): # !!
```
Hexcles: ? # !!

foolip: Yes, this line was irregular, `channel` is never passed and there's dead code. I left the comment to remember simplifying it first in a separate commit.

foolip:
Needs to be rebased on top of https://github.com/web-platform-tests/wpt/pull/13642 now.

Hexcles:
@foolip could your rebase this and fix the unit tests?

foolip:
Alright, this finally works! A snippet from a wpt_report.json produced locally:
```
    "run_info": {
        "bits": 64,
        "browser_channel": "preview",
        "browser_version": "Technology Preview (Release 68, 13607.1.11)",
        "debug": false,
        "has_sandbox": true,
        "os": "mac",
        "os_version": "10.13",
        "processor": "x86_64",
        "product": "safari_webdriver",
        "revision": "eb59233bf287cd7391926719bcfc9bf9f4e02a20",
        "verify": false,
        "version": "OS X 10.13.6",
        "wasm": false
    }
```

(It works when using `--webdriver-binary` instead of `--channel` too, but then `"browser_channel"` won't be present. That's why it's a good idea to include "Technology Preview" in the output.)

jgraham (approved):

PR #13471, opened 2018-10-11
Fetch: combining header values

annevk:
Tests for both request and response combining.

Needed for https://github.com/whatwg/fetch/issues/757.


annevk:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1499356 against Firefox.

annevk:
And https://bugs.chromium.org/p/chromium/issues/detail?id=897026 against Chrome for a subtle `setRequestHeader()` issue.

foolip:
@MattMenke2 can you review this too? :)

MattMenke2:
Happy to.  I probably won't get to it until Monday, though.

foolip:
I'd be no faster, so thanks! :)

xhr/setrequestheader-combining.window.js
```diff
@@ -0,0 +1,11 @@
+test(() => {
+  const client = new XMLHttpRequest();
+  client.open("POST", "resources/inspect-headers.py?filter_name=test-me", false);
+  client.setRequestHeader("test-me", "");
+  client.setRequestHeader("test-me", "");
+  client.setRequestHeader("test-me", " ");
```
MattMenke2: Maybe a tab, too?

xhr/resources/headers-some-are-empty.asis
```diff
@@ -0,0 +1,6 @@
+HTTP/1.0 200 MEH
+HEYA:	 
+HEYA: 1
+HEYA:
+HEYA:
```
MattMenke2: Suggest adding trailing whitespace either after the 1 or 2, but not both (At least I think neither of them has trailing whitespace)

xhr/resources/headers-some-are-empty.asis
```diff
@@ -0,0 +1,6 @@
+HTTP/1.0 200 MEH
+HEYA:	 
+HEYA: 1
+HEYA:
```
MattMenke2: Should we have a test that \C is not removed, since apparently some rules are going to have special handling of it (At least if we can't get rid of that)?

fetch/api/basic/header-value-combining.any.js
```diff
@@ -1,11 +1,14 @@
...
-  const response = await fetch("../../../xhr/resources/headers-basic.asis");
-  assert_equals(response.headers.get("foo-test"), "1, 2, 3");
-}, "response.headers.get('foo-test')");
-
-promise_test(async t => {
-  const response = await fetch("../../../xhr/resources/headers-www-authenticate.asis");
-  assert_equals(response.headers.get("www-authenticate"), "1, 2, 3, 4");
-}, "response.headers.get('www-authenticate')");
+[
+  ["content-length", "0", "header-content-length"],
```
MattMenke2: Suggest also a:

content-length: 0
content-length: 0

Test, though maybe that should go elsewhere.

MattMenke2 (commented):
Think this looks pretty good.  A couple suggestions to slightly beef up the tests.

xhr/resources/headers-some-are-empty.asis
annevk: Yeah good call, Safari fails that looks like.

annevk:
Safari fails the 0x0B & 0x0C test, good call. Firefox fails the double `Content-Length` test. The others you suggested aren't catching browsers out it seems, but still seem good additions.

MattMenke2 (approved):

annevk:
Filed https://bugs.webkit.org/show_bug.cgi?id=191102 against Safari.

PR #13474, opened 2018-10-11
Python 3: Fix failing tests in test_handlers.py.

Ms2ger:


tools/wptserve/wptserve/handlers.py
```diff
@@ -175,13 +175,14 @@ def load_headers(self, request, path):
...
         try:
-            with open(headers_path) as headers_file:
+            with open(headers_path, "rb") as headers_file:
                 data = headers_file.read()
         except IOError:
             return []
         else:
             if use_sub:
                 data = template(request, data, escape_type="none")
+            data = data.decode("ascii")
```
gsnedders: Hmm. I suspect we actually want ISO-8859-1 here? We should definitely be able to have non-ASCII in `.headers`. (I presume there's some reason why we need unicode strings here?)

Ms2ger: Because `template()` works on unicode strings, though maybe that was the wrong call.

Ms2ger: No, I misread; `template()` works on bytes. `ResponseHeaders` handles both, but its callers expect text.

gsnedders:
@Ms2ger can you rebase this onto master?

gsnedders (approved):

PR #13486, opened 2018-10-12
IdlHarness: Improve the error message for alias values.

Ms2ger:


foolip (approved):
Ah, such a simple fix, thanks!

PR #13489, opened 2018-10-12
Fix handling of about:blank in lint.

jgraham:
about:blank should be unconditionally allowed as a reference. Also add
an infrastructure test that actually uses this to ensure it keeps working.

### DIFF
--- a/infrastructure/assumptions/blank.html
+++ b/infrastructure/assumptions/blank.html
@@ -1,0 +1,2 @@
+<title>Blank Document</title>
+<link rel=match href="about:blank">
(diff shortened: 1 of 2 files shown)


emilio (approved):
Ah, I see what went wrong, thanks for fixing this!

gsnedders:
Can you add a test to `tools/lint/test_lint.py` too?

gsnedders:
Wait, we already have one that passes. So what's the case that fails? Did it trigger NON-EXISTENT-REF? Why didn't the existing test catch that?

emilio:
@gsnedders see https://bugzilla.mozilla.org/show_bug.cgi?id=1494685#c13

emilio:
(No idea why the existing test didn't catch it)

Issue #11120, opened 2018-05-23
Remove testharness.js test-level timeouts (often used for async_test)


foolip:
See discussion at https://github.com/w3c/web-platform-tests/issues/11029#issuecomment-389499369

https://chromium-review.googlesource.com/c/chromium/src/+/1061753 shows what tests currently use this timeout.

Steps to this issue:
 * Remove the use of `{ timeout: 1234 }` from all tests
 * Run https://chromium-review.googlesource.com/c/chromium/src/+/1061753 through Chromium's CQ again (or Gecko's)
 * Remove support for the timeout entirely.

@jgraham, plan SGTY?

foolip:
What would remain of the `Test` constructors second argument `properties` is the "assert" member, AFAICT only used in the output. @jgraham, should we want to get rid of that as well?

jgraham:
Yeah, I'm happy with all of that plan, including removing the `assert` support.

If you make a PR it should automatically run all those tests through the gecko CI.

foolip:
I've sent https://chromium-review.googlesource.com/c/chromium/src/+/1070155 to see if anything uses that.

foolip:
It looks like these strings show up in their own column in the report. Just having the information in the test name or assertion message would work. Needing to pick one means this will require some human judgement to do.

qiuzhong:
I'm interested in working on this issue. 

@foolip, I didn't find any strings like `{timeout: 1234}` in the html tests above. How did you get this list?

foolip:
@qiuzhong, that's great! Try `git grep -P "timeout:\s*\d+"` in the root of the repository, that matches 223 files, most of which look like this: `async_test("getCurrentPosition success callback tests", {timeout:20000})`. The first step is to update those tests to look like this instead: `async_test("getCurrentPosition success callback tests")`.

foolip:
It would be a good idea to split those changes into multiple PRs since otherwise a lot of reviewers will get pulled in. Maybe one PR for each of html, IndexedDB, websockets, mediacapture-streams and cookies, and then one for everything else?

Once we think all uses are gone, I'll run a change similar to https://chromium-review.googlesource.com/1070155 through Chromium's CQ to see if anything was missed. Then we can just delete timeouts.

qiuzhong:
@foolip, thanks for your good advice. It's really a good idea to split this update into several PRs. 

[...]

Issue #11029, opened 2018-05-16
Declare timeout in test and extract it into manifest


foolip:
https://bugs.chromium.org/p/chromium/issues/detail?id=843213 is about how the information in `async_test(undefined, { timeout: 10000 })` isn't machine readable, and so the test runner still can't increase the external timeout based on it.

To make that work, I think the information would have to be in `<meta something=somethin>` and we should extract it into the manifest for `./wpt run` to use.

foolip:
@jgraham, how do you deal with slow tests in Gecko?

foolip:
OK, so we already have `<meta name=timeout content=long>`. Is that information extracted into the manifest and used by the runner already?

jgraham:
Yes.

foolip:
Should the per-test timeout be nuked? Nothing about `var timeout = properties.timeout ? properties.timeout : settings.test_timeout` looks deprecated.

jgraham:
The indicaton that it's deprecated is that that it's not in the documentation. I admit it's not a *great* signal, as there are lots of other things that aren't deprecated and aren't in the documentation. Anyway I would welcome a PR to remove the feature and all the instances of per-test timeouts in existing tests.

jgraham:
wptrunner makes a distinction between an internal timeout ("browser still respodng enough to signal a timeout occurred; keep using the same instance for the next test") and an external timeout ("things are so broken we should kill the browser and restart"). It's possible to just have an external timeout everywhere, but the distinction is meaningful. Also it is helpful for people running tests directly in a browser to understand if they are taking too long.

foolip:
OK, I agree that's a meaningful distinction and useful when running manually. I presume the external timeout is always a bit longer, to not make it entirely racy?

jgraham:
Yes. We set the selenium/marionette timeout to the test timeout + 5s and the "give up waiting for the driver to return" timeout to the test timeout + 10s.

foolip:
OK, nothing to be done here.

Hexcles:
This matches my understanding ("timeouts in tests" are discouraged).

Shall we remove that section from https://web-platform-tests.org/writing-tests/testharness-api.html, or at least mark it as deprecated?

Issue #11009, opened 2018-05-15
Revisit LICENSE.md?


plehegar:
The current [LICENSE.md](https://github.com/w3c/web-platform-tests/blob/master/LICENSE.md) comes from history in W3C around test suites and hasn't been looked at in the past 4 years at least.

We should look at it again and see if it still makes sense. If not, we need to figure out what would make sense. and then a transition path from the current one to a new one.

jgraham:
(I am not an expert, etc.)

The only transition path in terms of changing licences would be to adopt the only the BSD variant currently allowed. Relicensing wpt is effectively impossible, since there is no copyright assignment and there are such a large number of contributors.

I'm all in favour of dropping the W3C Test License part though; I don't think anyone uses that in practice.

foolip:
There is also [CONTRIBUTING.md](https://github.com/w3c/web-platform-tests/blob/master/CONTRIBUTING.md), which I do not fully understand. @plehegar, can you explain in lay terms what that means, and how it relates to the license, if at all?

plehegar:
The first is the grant of license, which basically gives us the right to relicense under the original BSD license, effectively replacing the current LICENSE.md if we'd like to. I could check to see if we still need to have it but I would guess so, except that we could give the right to relicense to more than just W3C.
The second part is a disclaimer. not sure if we need this.

foolip:
@plehegar, by "the original BSD license", do you mean something other than the "W3C 3-clause BSD License" in LICENSE.md?

wseltzer:
@foolip yes, I believe we could switch to just the 3-clause BSD license instead of the current dual-license. (Anyone who was using tests under the W3C Test Suite License would be able to continue to do so.)

foolip:
@wseltzer, that sounds great, do you want to prepare a PR to trim LICENSE.md so that we get it exactly right?

annevk:
I looked into providing the PR for this, taking https://opensource.org/licenses/BSD-3-Clause as inspiration. However, that suggests you need to start with

> Copyright <YEAR> <COPYRIGHT HOLDER>

which is something that we currently lack. Would

> Copyright 2018 web-platform-tests contributors

work?

foolip:
I think the idea is to trim LICENSE.md to just contain https://github.com/w3c/web-platform-tests/blob/master/LICENSE.md#w3c-3-clause-bsd-license verbatim, but waiting for PR from @wseltzer or @plehegar instead of trying it myself :)

[...]

PR #13499, opened 2018-10-13
More Tests for WebRTC Data Channels

lgrahl:
### Description

This PR updates all data channel related tests to reflect the current spec (at the time of writing, that was [revision 1cc5bfc3ff18741033d804c4a71f7891242fb5b3](https://rawgit.com/w3c/webrtc-pc/1cc5bfc3ff18741033d804c4a71f7891242fb5b3/webrtc.html)).

### Updated/Fixed Tests

* RTCDataChannel-id
  - *DTLS client uses odd data channel IDs*
  - *DTLS server uses even data channel IDs*
* RTCDataChannel-send-receive (renamed from RTCDataChannel-send)
  - (\<type> is of *In-band negotiated channel*, *Negotiated channel* or *Asymmetric negotiated channel*)
  - *\<type>: Calling send() when not open should throw InvalidStateError* (renamed from *Calling send() when data channel is in connecting state should throw InvalidStateError*)
  - *\<type>: Calling send() with null should throw TypeError*
  - *\<type>: Calling send() with undefined should throw TypeError*
  - *\<type>: Should be able to send (local) simple string and receive (remote) as string* (based on *Data channel should be able to send simple string and receive as string*)
  - *\<type>: Should be able to send (remote) simple string and receive (local) as string* (based on *Data channel should be able to send simple string and receive as string*)
  - *\<type>: Should be able to send unicode string and receive as unicode string* (renamed from *Data channel should be able to send unicode string and receive as unicode string*)
  - *\<type>: Should ignore binaryType and always receive string message as string* (renamed from *Data channel should ignore binaryType and always receive string message as string*)
  - *\<type>: Should be able to send Uint8Array message and receive as ArrayBuffer* (renamed from *Data channel should be able to send Uint8Array message and receive as ArrayBuffer*)
  - *\<type>: Should be able to send ArrayBuffer message and receive as ArrayBuffer* (renamed from *Data channel should be able to send ArrayBuffer message and receive as ArrayBuffer*)
  - *\<type>: Should be able to send Blob message and receive as ArrayBuffer* (renamed from *Data channel should be able to send Blob message and receive as ArrayBuffer*)
  - *\<type>: Should be able to send ArrayBuffer message and receive as Blob* (renamed from *Data channel should be able to send ArrayBuffer message and receive as Blob*)
  - *\<type>: Should be able to transmit an empty string* (moved from and based on datachannel-emptystring *Can send empty strings across a WebRTC data channel.*)
  - *\<type>: Should receive messages as Blob by default* (renamed from *Data channel binaryType should receive message as Blob by default*)
  - *\<type>: Sending multiple messages with different types should succeed* (based on *Sending multiple messages with different types should succeed and be received*)
  - *\<type>: Receiving multiple messages with different types should succeed* (based on *Sending multiple messages with different types should succeed and be received*)
* RTCPeerConnection-createDataChannel
  - *createDataChannel attribute default values*
  - *createDataChannel with provided parameters should initialize attributes to provided values*
  - *createDataChannel with both maxPacketLifeTime and maxRetransmits should throw TypeError* (renamed from *createDataChannel with both maxPacketLifeTime and maxRetransmits should throw SyntaxError*)
  - *createDataChannel with too long label should throw TypeError* (renamed from *createDataChannel with negotiated false and long label should throw TypeError*)
  - *createDataChannel with too long protocol should throw TypeError* (renamed from *createDataChannel with negotiated false and long protocol should throw TypeError*)
  - *Channels created (after setRemoteDescription) should have id assigned* (renamed from *Channels created after SCTP transport is established should have id assigned*)
* RTCPeerConnection-ondatachannel
  - *Data channel event should fire when new data channel is announced to the remote peer* (renamed from *datachannel event should fire when new data channel is announced to the remote peer*)
  - *In-band negotiated channel created on remote peer should match the same configuration as local peer* (renamed from *Data channel created on remote peer should match the same configuration as local peer*)
* RTCSctpTransport-constructor
  - *setRemoteDescription() with answer containing data media should initialize pc.sctp*
  - *setLocalDescription() with answer containing data media should initialize pc.sctp*
* RTCSctpTransport-maxMessageSize
  - *Remote offer SDP missing max-message-size attribute*
  - *max-message-size with a (non-zero) value provided by the remote peer*
  - *Renegotiate max-message-size with various values provided by the remote peer* (based on *Renegotiate max-message-size with a (non-zero) value provided by the remote peer*)
  - *max-message-size with a (non-zero) value larger than canSendSize provided by the remote peer*

### New Tests

* RTCDataChannel-binaryType
  - *Setting binaryType to 'blob' should succeed*
  - *Setting binaryType to 'arraybuffer' should succeed*
  - *Setting invalid binaryType 'jellyfish' should throw SyntaxError*
  - *Setting invalid binaryType 'arraybuffer ' should throw SyntaxError*
* RTCDataChannel-bufferedAmount
  - *bufferedAmount initial value should be 0 for both peers*
  - *bufferedAmount should not decrease immediately after initiating closure*
  - *bufferedAmount should not decrease after closing the peer connection*
* RTCDataChannel-bufferedAmountLowThreshold
  - (\<type> is of `DOMString` or `Uint8Array`)
  - *\<type> of 4 bytes: bufferedamountlow event should fire once bufferedAmount <= 0 (default)*
  - *\<type> of 4 bytes: bufferedamountlow event should fire once bufferedAmount <= 2 (custom)*
  - *\<type> of 4 bytes: bufferedamountlow event should fire once bufferedAmount <= 4 (custom)*
  - *\<type> of 4 bytes: bufferedamountlow event should fire once bufferedAmount <= 8 (custom)*
  - *\<type> of 4 bytes x 10: bufferedamountlow event should fire once bufferedAmount <= 2 (custom)*
  - *\<type> of 4 bytes x 10: bufferedamountlow event should fire once bufferedAmount <= 4 (custom)*
  - *\<type> of 4 bytes x 10: bufferedamountlow event should fire once bufferedAmount <= 8 (custom)*
  - *\<type> of 4 bytes: bufferedamountlow event should not fire on other channels*
* RTCDataChannel-close
  - *In-band data channel closing before connection establishment*
  - *In-band data channel closing after connection establishment*
  - *In-band data channels x10 closing on local side should be synchronized across peers*
  - *In-band data channels x10 closing on remote side should be synchronized across peers*
  - *Negotiated data channels x10 closing should be synchronized across peers*
  - *Closing peer connection locally should close all 10 channels for the remote peer*
* RTCDataChannel-dcep
  - (\<value> is of `undefined`, `null`, `0`, `65535` or `4294967295`)
  - *Ordered and reliable channel should be created via DCEP*
  - *Unordered and reliable channel should be created via DCEP*
  - *Ordered and unreliable (maxRetransmits=\<value>) channel should be created via DCEP*
  - *Unordered and unreliable (maxRetransmits=\<value>) channel should be created via DCEP*
  - *Ordered and unreliable (maxPacketLifeTime=\<value>) channel should be created via DCEP*
  - *Unordered and unreliable (maxPacketLifeTime=\<value>) channel should be created via DCEP*
  - *Channel with priority set to high should be created via DCEP*
  - *Channel IDs should be synchronized when created via DCEP*
  - *Zero length label and protocol option should be transmitted via DCEP*
  - *Maximum length label and protocol option should be transmitted via DCEP*
  - *Maximum length label and protocol option (3 byte unicode) should be transmitted via DCEP*
* RTCDataChannel-id
  - *In-band negotiation with a specific ID should not be allowed*
  - *Odd/even role should not be violated when mixing with negotiated channels*
  - *Creating a channel after exhausting the maximum number of channels should throw OperationError (after connection establishment)*
  - *ID reuse should not be possible before the former channel with the same ID closed*
  - *ID reuse should be possible once the former channel with the same ID closed*
  - *Channel IDs should be reassigned after exhaustion in case a channel has been closed*
  - *Exhausting channels with one peer should not violate the odd/even rule*
  - *Channel ID exhaustion handling (before and after connection establishment)*
* RTCDataChannel-onopen
  - *In-band channel: Open event should be fired (local) when the data channel opens*
  - *In-band channel: Open event should be fired (remote) when the data channel opens*
  - *Negotiated channel: Open event should be fired when the channels open*
* RTCDataChannel-send-receive
  - (\<type> is of *In-band negotiated channel*, *Negotiated channel* or *Asymmetric negotiated channel*)
  - (\<size> is of `1024`, `16384`, `65536`, `131072`, `262144`, `524288`, `1048576`, `16777216` or `33554432`)
  - *\<type>: Calling send() with 42 should throw TypeError*
  - *\<type>: Should be able to send Int8Array message and receive as ArrayBuffer*
  - *\<type>: Should be able to send Uint8ClampedArray message and receive as ArrayBuffer*
  - *\<type>: Should be able to send Int16Array message and receive as ArrayBuffer*
  - *\<type>: Should be able to send Uint16Array message and receive as ArrayBuffer*
  - *\<type>: Should be able to send Int32Array message and receive as ArrayBuffer*
  - *\<type>: Should be able to send Uint32Array message and receive as ArrayBuffer*
  - *\<type>: Should be able to send Float32Array message and receive as ArrayBuffer*
  - *\<type>: Should be able to send Float64Array message and receive as ArrayBuffer*
  - *\<type>: Should be able to send DataView message and receive as ArrayBuffer*
  - *\<type>: Should be able to send Uint8Array (with offset) message and receive as ArrayBuffer*
  - *\<type>: Should be able to send DataView (with offset) message and receive as ArrayBuffer*
  - *\<type>: Should be able to send Blob (with offset) message and receive as ArrayBuffer*
  - *\<type>: Should be able to transmit an empty Uint8Array*
  - *\<type>: Should be able to transmit an empty Blob*
  - *\<type>: Sending and receiving \<size> bytes should succeed or raise TypeError*
  - *\<type>: Sending and receiving \<size> +1 bytes should raise TypeError*
  - *\<type>: Sending and receiving 16 KiB x64 should succeed*
  - *\<type>: Sending and receiving 16 KiB x256 on both peer simultaneously should succeed*
  - *\<type>: Sending and receiving 2048 messages should succeed and keep order*
  - *\<type>: Closing a channel (local) with pending data should transfer that data before becoming closed*
  - *\<type>: Sending after the channel has been closed (local) should raise InvalidStateError*
  - *\<type>: Sending after the channel has been closed (remote) should raise InvalidStateError*
  - *\<type>: Sending after the peer connection has been closed (local) should raise InvalidStateError*
  - *\<type>: Sending after the peer connection has been closed (remote) should raise InvalidStateError*
* RTCDataChannel-send-receive-stress
  - *Stress-test with multiple channels sending and receiving using various data sources*
* RTCPeerConnection-close
  - *Closing the peer connection should close all channels (before connection establishment)*
  - *Closing the local peer connection should close all channels (after connection establishment)*
  - *Closing the remote peer connection should close all channels (after connection establishment)*
  - *Closing the peer connection should set the SCTP transport to closed*
* RTCPeerConnection-createDataChannel
  - *createDataChannel with both maxPacketLifeTime and maxRetransmits null should succeed*
  - *createDataChannel with too long label (2 byte unicode) should throw TypeError*
  - *createDataChannel with same label used twice should not throw*
  - *createDataChannel with too long protocol (2 byte unicode) should throw TypeError*
  - *createDataChannel with maximum length label and protocol should succeed*
  - *createDataChannel with negotiated false should succeed*
  - *createDataChannel with negotiated false and id 42 should ignore the id*
  - *createDataChannel with negotiated true and id not defined should throw TypeError*
  - *createDataChannel with negotiated true and id null should throw TypeError*
  - *Reusing a data channel id that is in use should throw OperationError*
  - *Reusing a data channel id that is in use (after setRemoteDescription) should throw OperationError*
  - *Reusing a data channel id that is in use (after setRemoteDescription, negotiated via DCEP) should throw OperationError*
  - *New data channel should be in the connecting state after creation (after connection establishment)*
* RTCPeerConnection-ondatachannel
  - *Should be able to send data in a datachannel event handler*
  - *Open event should not be raised when closing the channel in the datachannel event*
  - *Open event should be raised when closing the channel in the datachannel event after enqueuing a task*
  - *Open event should not be raised when sending and immediately closing the channel in the datachannel event*
  - *In-band negotiated channel created on remote peer should match the same (default) configuration as local peer*
* RTCSctpTransport-constructor
  - *setRemoteDescription() with answer not containing data media should not initialize pc.sctp*
  - *setLocalDescription() with answer not containing data media should not initialize pc.sct*
* RTCSctpTransport-onstatechange
  - *RTCSctpTransport.onstatechange should fire when the state changes*
* historical
  - *RTCDataChannel member maxRetransmitTime should not exist*

### Comments

Only few helper functions have been changed but everything is either backwards compatible or the affected tests have been updated. FWIW, I have carefully tested new test cases (if that was possible) on Firefox and Chrome by temporarily commenting out asserts and in some cases even by throwing in adapter (`RTCSctpTransport.maxMessageSize` for example).

Towards WebRTC implementations: A bunch of new issues are revealed by these tests. I'm at a tight time schedule, so I can't file them for you at the moment. This listing here should help you in finding new or updated test cases, so you can track and file new issues in your bug trackers. /cc @nils-ohlmeier @jan-ivar. Please also forward this to whoever is responsible for Google's, Safari's and Edge's (:joy_cat:) data channel implementation.

And I apologise for creating this rather large PR.


lgrahl:
Harald, can I ping you for another review? The changes to testharness have been removed and the tests have been rebased against master.

/cc @alvestrand

foolip:
@lgrahl thank you for contributions! Unfortunately, I have to ask that you not add a license file to the repo beyond the existing https://github.com/web-platform-tests/wpt/blob/master/LICENSE.md. There's an ongoing discussion about changing it in https://github.com/web-platform-tests/wpt/issues/11009 but for now the license is what it is.

If you don't yourself have the rights to all of this code, then please work with the original authors to approve contribution under the license of this repo.

lgrahl:
@foolip All authors approve to publish under the W3C license but I would like to keep the attribution to these three guys. Can I change the license file to an attribution file?

foolip:
If you'd like to acknowledge the authors, I'd suggest listing them in the tests, like this:
https://github.com/web-platform-tests/wpt/blob/7c50c216081d6ea3c9afe553ee7b64534020a1b2/css/css-transforms/transform-compound-001.html

lgrahl:
I would like to do that but associating the authors with the individual test cases (and files) is near impossible by now (I can explain why if you want me to). Adding them to all data channel files is also a bit weird but I guess I wouldn't mind. What do you think?

agouaillard-cosmo:
The license and CLA does not allow you to touch license and copyright, so In my humble opinion, foolip is right, and you should put their name in every single file you think deserve the mention.

foolip:
Oh, all right, well in that case perhaps you can put something in a README.md. https://github.com/web-platform-tests/wpt/commit/dd299f45326e12acf453a14db5b582eba6d5db63 is another README that explains the origins of something, as an example.

agouaillard-cosmo:
if you put something in the readme, you will need to put the list of file those names are credited in the readme. The readme then becomes dependent on those files, and you need to know to update it when you modify the files. It's brittle in my opinion. Better to keep the credits in corresponding files.

foolip:
Getting attribution exactly right and having it stay right over time is a pain, so I'd just say "some tests". Sort of like some of the attributions after the long list of names in https://html.spec.whatwg.org/multipage/acknowledgements.html, which aren't very precise about exactly what came from whom.

lgrahl:
I have removed the license file and added a readme (with attribution).

I must say I agree with @foolip here as the attribution cannot be tied to specific files, so we just shouldn't try. Just to explain why this happened: First of all, I have updated all existing data channel tests to be up to date. Then, to maximise diversity and not spoil myself, I have written my own tests based on the specification and on existing examples I've written a while ago. After that, I have added missing tests from the [data channel conformance test suite][nplab-webrtc-dc-playground]. And I don't remember which ones that were because I didn't think it would be important. Anyhow, because of this, saying that "some data channel tests are based on ..." seems fair to me.

[nplab-webrtc-dc-playground]: https://github.com/nplab/WebRTC-Data-Channel-Playground/tree/master/conformance-tests

webrtc/README.md
```diff
@@ -0,0 +1,12 @@
+# WebRTC
+
+This directory contains the WebRTC test suite.
+
+## Attribution
```
foolip: This looks great! To avoid the possible impression that this is fulfilling a license requirement for attribution (such as with CC-by) may I suggest "Acknowledgments" instead?

webrtc/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+Some data channel tests are based on the [data channel conformance test
+suite][nplab-webrtc-dc-playground] of the Network Programming Lab of the Münster
+University of Applied Sciences. We would like to thank Peter Titz, Felix Weinrank and Timo
+Völker for allowing us to make their test cases available under the W3C license.
```
foolip: Note that per https://github.com/web-platform-tests/wpt/issues/11009 the license may change, as allowed for by the terms in CONTRIBUTING.md.

To not have to update this text later, may I suggest not mentioning the license but instead sometime like "agreeing to contribute their test cases to this repository"?

webrtc/RTCDataChannel-bufferedAmount.html
```diff
@@ -5,165 +5,239 @@
...
+  const pc2 = new RTCPeerConnection();
+  t.add_cleanup(() => pc1.close());
+  t.add_cleanup(() => pc2.close());
+
+  const [dc1, dc2] = await createDataChannelPair(pc1, pc2);
+
+  assert_equals(dc1.bufferedAmount, 0, 'Expect bufferedAmount to be 0');
+  assert_equals(dc2.bufferedAmount, 0, 'Expect bufferedAmount to be 0');
+}, 'bufferedAmount initial value should be 0 for both peers', {
+  timeout: 5000
```
youennf: Even for async tests, I am not sure this is worth the effort to add specific timeouts.
In some configurations, 5 seconds will be what is expected to run the tests of the whole file for instance.

If we really want to do something there, maybe createDataChannelPair should actually reject after some timeout?
That way, if we have several await in a promise test, we will know which one is timing out.

webrtc/RTCPeerConnection-helper.js
```diff
@@ -191,93 +191,242 @@ function test_never_resolve(testFunc, testName) {
...
+//
+// IMPORTANT: Due to a bug in Safari which leads to 'id' being 'null', we send
+//            the remote channel a message that requests the ID to be sent back
+//            to the local channel in order to identify the pair.
+const remoteChannels = {};
 function createDataChannelPair(
   pc1=new RTCPeerConnection(),
-  pc2=new RTCPeerConnection())
+  pc2=new RTCPeerConnection(),
+  options={})
```
youennf: s/=/ = /

youennf (commented):
This looks great, thanks a lot!

I haven't looked at whether the tests are matching the spec or not.
Given the number of tests, we might want to concentrate on the tests that show differences/bugs between implementation and spec. If you already have a list, that would be great, otherwise it should be ok to repro it locally.

Also, I am not a fan of explicit handling of timeouts, I would tend to remove most of them.

alvestrand:
@lgrahl can we find some time when you'll be available for updating the tests? Having to wait multiple days after a review before seeing the responses increases the cognitive load on reviewers - if we can find a day to sit down and ping-pong on these tests, it should be possible to get the review done.
Also, Chrome tests have trouble - see bot run.

lgrahl:
@alvestrand @youennf Sorry for the delay. I'm not sure what the issue is on Travis since the Taskcluster seems to run fine. Looking at the log, I'm speculating that Travis takes too long to process the large changeset... but I might be wrong. In any way, I was not able to pin it down so far.

I'm available today for *ping-pong*. The upcoming weeks are packed.

webrtc/RTCDataChannel-binaryType.html
```diff
@@ -0,0 +1,69 @@
...
+ */
+for (const binaryType of validBinaryTypes) {
+  test((t) => {
+    const pc = new RTCPeerConnection();
+    const dc = pc.createDataChannel('test-binary-type');
+
+    dc.binaryType = binaryType;
+    assert_equals(dc.binaryType, binaryType, `dc.binaryType should be '${binaryType}'`);
+  }, `Setting binaryType to '${binaryType}' should succeed`, {
+    timeout: 3000
```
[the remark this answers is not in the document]
lgrahl: You're right, in this file the tests are not async so the timeouts can be removed. Done!

webrtc/RTCDataChannel-binaryType.html
```diff
@@ -0,0 +1,69 @@
...
+<title>RTCDataChannel.prototype.binaryType</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="RTCPeerConnection-helper.js"></script>
+<script>
+'use strict';
+
+// Disable global timeout
+// IMPORTANT: You need to add a timeout to *every test* of this file!
+setup({ explicit_timeout: true });
```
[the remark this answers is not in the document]
lgrahl: Removed (in this file).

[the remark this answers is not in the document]
lgrahl: However, in general this exists to allow a per-test timeout instead of the inflexible global per-file timeout which can be either 10 seconds or 60 seconds. This is insufficient for a test file with many promise-based test cases that can only run sequentially. Therefore, these files provide a timeout for each test case and disable the global timeout.

webrtc/RTCDataChannel-bufferedAmount.html
lgrahl: If we don't specify the local timeout, the test case will block running all other test cases. The alternative is to carefully wrap all possible awaitables with timeouts but I don't want to go that way for several reasons:

- it's quite a lot of boilerplate,
- makes the test cases brittle since people (including me) will tend to forget an awaitable here and there, and
- not every awaitable that could in theory time out is likely to time out.

Adding failure context to some awaitables by using timeouts is a good idea, though (and I have done that in vital places). Rejecting `createDataChannelPair` after some timeout could be done but since that isn't part of the test case definition, the question is when the timeout should fire...

foolip:
Travis wasn't timing out, the limit is 50 minutes and the stability jobs took much less than that. Probably the tests actually are flaky.

lgrahl:
@foolip Do you have any hints? Frankly, I'm having a hard time to figure out what Travis is doing.

webrtc/RTCDataChannel-binaryType.html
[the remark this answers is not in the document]
lgrahl: > If we fear that timing out subtests is an issue, we should simply split subtests in different files.

Don't forget that a `promise_test` blocks all subsequent tests. This means, if a `promise_test` does not have a local timeout and it does time out, it prevents all other tests from running. If we don't want local timeouts, then we have to move each test case into its own file.

webrtc/RTCDataChannel-bufferedAmount.html
lgrahl: What about the tests that share 99% of their code?

webrtc/RTCDataChannel-binaryType.html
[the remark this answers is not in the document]
youennf: In practice, timing out subtests do happen.
Browser implementors tend to try fixing the corresponding issues sooner than regular failures.

webrtc/RTCDataChannel-bufferedAmount.html
youennf: Sure, if some tests share 99%, it is ok to group them (and incur the timeout risk).

The alternative is of course to have some helper functions in some JS files.
That way, you keep the meat of each subtest in each promise_test function.
And you can then split tests as desired.

In general, the more tests on a given file, the more flaky.
And the more difficult to debug as well...

youennf:
I do not see the logs.
There seems to be some conflicts with the recent WebRTC patches we uploaded during today hackathon.

lgrahl:
There's an older build that failed (only cosmetic changes have been done in the meantime): https://travis-ci.org/web-platform-tests/wpt/builds/444438000

webrtc/RTCDataChannel-binaryType.html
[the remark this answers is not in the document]
lgrahl: I very much like the thought of browser vendors fixing their implementations due to tests timing out but don't consider that approach a good thing. Anyhow, I have moved this discussion further down into the comments since I want to hear some more opinions on that.

youennf:
For the record, the tests that seem to fail are:
1. "/webrtc/RTCDataChannel-bufferedAmountLowThreshold.html#DOMString of 4 bytes x 10: bufferedamountlow event should ...".
In WebKit, we had issues getting a consistent bufferedamountlow testing, we ended up writing a large enough message. Maybe this is a bug in implementations...
Isolating a flaky subtest from other subtests is good. Otherwise it might end up making other subtest results difficult to use.

2. "/webrtc/RTCDataChannel-close.html#Negotiated data channels x10 closing should be synchronized across peers"
Not sure why it fails. If nChannels is equal to 1, then nCloses will be 1 but will be compared to 0?
Should it be compared to nChannels? Ditto in some other places.

3. "/webrtc/RTCDataChannel-send-receive.html#In-band negotiated channel..."
Not sure why it fails either.
This test has both a step_timeout and an explicit timeout value, is there a particular reason for that.

The fact that there is a comment stating that a 10 seconds step_timeout is required worries me a bit.

I also spotted some very big timeouts at least in webrtc/RTCDataChannel-send-receive-stress.html
and webrtc/RTCDataChannel-send-receive.html (69 seconds, 420 seconds).
I would try either rewriting these tests or putting them aside in their own file.
These tests might otherwise be skipped in browsers CI, at least in slow configurations(think debug+asan with multiple tests run in parallel).

youennf:
For the record, WebKit default timeout for running a file is 30 seconds.
We can extend it to 150 seconds but we usually prefer not to have to do that.

lgrahl:
> 1. "/webrtc/RTCDataChannel-bufferedAmountLowThreshold.html#DOMString of 4 bytes x 10: bufferedamountlow event should ...".
> In WebKit, we had issues getting a consistent bufferedamountlow testing, we ended up writing a large enough message. Maybe this is a bug in implementations...

That bug is explained in my thesis but this being flaky is new to me. This should only become flaky if one reuses an existing SCTP association (`RTCPeerConnection` instance, or creating more than one data channel) in another test.

> [...] Isolating a flaky subtest from other subtests is good. Otherwise it might end up making other subtest results difficult to use.

I don't understand why it would make the results difficult to use. They don't depend on each other. If some test case breaks another, let me know because that would be a bug.

Splitting them however would duplicate a lot of code. I find the idea awkward to split test cases depending on the implementation status in browsers.

>     2. "/webrtc/RTCDataChannel-close.html#Negotiated data channels x10 closing should be synchronized across peers"
>        Not sure why it fails. If nChannels is equal to 1, then nCloses will be 1 but will be compared to 0?
>        Should it be compared to nChannels? Ditto in some other places.

Looks like a bug. I'll look into it.

>     3. "/webrtc/RTCDataChannel-send-receive.html#In-band negotiated channel..."
>        Not sure why it fails either.
>        This test has both a step_timeout and an explicit timeout value, is there a particular reason for that.

The only reason is being able to get the failure context.

> The fact that there is a comment stating that a 10 seconds step_timeout is required worries me a bit.

Mhh, but what's the alternative? Let it time out early to get Chrome people look into it?

> I also spotted some very big timeouts at least in webrtc/RTCDataChannel-send-receive-stress.html
> and webrtc/RTCDataChannel-send-receive.html (69 seconds, 420 seconds).
> I would try either rewriting these tests or putting them aside in their own file.

I'm not sure how to rewrite them... it just can take a while on a slow machine to send 32 MiB and receive it. But I absolutely think the test is necessary.

The stress test is already in its own file.

> These tests might otherwise be skipped in browsers CI, at least in slow configurations(think debug+asan with multiple tests run in parallel).

Is there a meta tag to mark it as a *long* test to be skipped by slow configurations?

youennf:
> Is there a meta tag to mark it as a _long_ test to be skipped by slow configurations?

Yes: '<meta name="timeout" content="long">'
But stress tests will probably not work well in the way WebKit runs WPT tests in its CI.

youennf:
> > [...] Isolating a flaky subtest from other subtests is good. Otherwise it might end up making other subtest results difficult to use.
> 
> I don't understand why it would make the results difficult to use. They don't depend on each other. If some test case breaks another, let me know because that would be a bug.

In WebKit CI (Chrome might still be fairly similar), we generate a text file containing all the subtest results (usually a file containing one PASS/FAIL line per subtest). If one of the line is different, the whole test is marked as failed. If there are variations run after run in the subtest results, we mark the whole file as flaky, even though only one subtest is actually flaky. We do not have per subtest expectation.

youennf:
> > I also spotted some very big timeouts at least in webrtc/RTCDataChannel-send-receive-stress.html
> > and webrtc/RTCDataChannel-send-receive.html (69 seconds, 420 seconds).
> > I would try either rewriting these tests or putting them aside in their own file.
> 
> I'm not sure how to rewrite them... it just can take a while on a slow machine to send 32 MiB and receive it. But I absolutely think the test is necessary.

What exactly does it test? Is it the robustness of the implementation or the compliance of it?

webrtc/RTCDataChannel-send-receive-stress.html is just one test so I am not sure how feasible splitting is. I would go with just meta/long there.

lgrahl:
> > > [...] Isolating a flaky subtest from other subtests is good. Otherwise it might end up making other subtest results difficult to use.
> > 
> > 
> > I don't understand why it would make the results difficult to use. They don't depend on each other. If some test case breaks another, let me know because that would be a bug.
> 
> In WebKit CI (Chrome might still be fairly similar), we generate a text file containing all the subtest results (usually a file containing one PASS/FAIL line per subtest). If one of the line is different, the whole test is marked as failed. If there are variations run after run in the subtest results, we mark the whole file as flaky, even though only one subtest is actually flaky. We do not have per subtest expectation.

I see. No disrespect, but that seems like a design flaw of the *flaky* evaluation. Again, I'm really unhappy with that approach... it feels wrong to move test cases out of their test group because some implementation is incorrect. Moving test case definitions out of the actual test case to reuse it in both files doesn't feel any less wrong.

But I suppose this isn't going to change in just a few days, so let's go through each test case that fails for Chrome (since I can't see which ones fail for Firefox as the log blows up the log buffer, /cc @jan-ivar):

- RTCDataChannel.bufferedAmountLowThreshold.html, *multiple*: I'm fine with the whole file being marked as flaky. There aren't many tests in this and they are mostly variants of the same test. If one is flaky, then the attribute is flaky.
- RTCDataChannel-close.html, *Negotiated data channels x10 closing should be synchronized across peers*: Could be grudgingly moved into a new file unless it also fails in Firefox which might indicate that there's a bug in the test.
- RTCDataChannel-send-receive.html, *In-band negotiated channel: Sending and receiving 262144 bytes should succeed or raise TypeError*: This could also be (grudgingly) moved into a different file but I find it misleading. How is it going to be displayed in the wpt.fyi results?
- RTCDataChannel-send-receive.html, *multiple marked as NOTRUN*: This should not happen with the global timeout disabled. I have no idea why it does.

> What exactly does it test? Is it the robustness of the implementation or the compliance of it?

Compliance.

> webrtc/RTCDataChannel-send-receive-stress.html is just one test so I am not sure how feasible splitting is. I would go with just meta/long there.

I can mark it as a *long* test but it will only serve as meta-information and not apply the global timeout of `60 * multiplier` seconds. Perhaps there should be a different meta tag for the purpose of marking it to not run in specific configurations.

foolip:
It looks like the Travis job has gotten entirely stuck here, most likely a Travis infrastructure issue since there's not even a link to follow. @lgrahl can you rebase to resolve conflicts? That'll also trigger a new run.

In https://travis-ci.org/web-platform-tests/wpt/builds/444438000, for Chrome Dev the answer is hidden in the log as a Markdown table. This was originally written for a bot to extract and paste as a comment, and here it is:

|                           Test                           |                                                 Subtest                                                 |             Results             |                                                          Messages                                                         |
|----------------------------------------------------------|---------------------------------------------------------------------------------------------------------|---------------------------------|---------------------------------------------------------------------------------------------------------------------------|
| `/webrtc/RTCDataChannel-bufferedAmountLowThreshold.html` | `DOMString of 4 bytes x 10: bufferedamountlow event should fire once bufferedAmount <=\n  2 (custom)`   | **FAIL: 8/10, TIMEOUT: 2/10**   | `assert_unreached: bufferedamountlow event did not fire in time Reached unreachable code;Test timed out`                  |
| `/webrtc/RTCDataChannel-bufferedAmountLowThreshold.html` | `DOMString of 4 bytes x 10: bufferedamountlow event should fire once bufferedAmount <=\n  4 (custom)`   | **NOTRUN: 2/10, TIMEOUT: 8/10** | `Test timed out`                                                                                                          |
| `/webrtc/RTCDataChannel-close.html`                      | `Negotiated data channels x10 closing should be synchronized across peers`                              | **PASS: 6/10, TIMEOUT: 4/10**   | `Test timed out`                                                                                                          |
| `/webrtc/RTCDataChannel-send-receive.html`               | `In-band negotiated channel: Sending and receiving 262144 bytes should succeed or raise\n    TypeError` | **PASS: 2/10, TIMEOUT: 8/10**   | `Test timed out`                                                                                                          |
| `/webrtc/RTCDataChannel-send-receive.html`               | `In-band negotiated channel: Sending and receiving 262144 +1 bytes should raise TypeError`              | **FAIL: 2/10, NOTRUN: 8/10**    | `assert_throws: Should throw TypeError in case the message is too large function "() => dc1.send(message)" did not throw` |
| `/webrtc/RTCDataChannel-send-receive.html`               | `In-band negotiated channel: Sending and receiving 524288 bytes should succeed or raise\n    TypeError` | **NOTRUN: 8/10, TIMEOUT: 2/10** | `Test timed out`                                                                                                          |

Looks like flaky timeouts across the board. Are the tests fairly slow, how long do they take to run locally?

For Firefox, the logs ends like this:
```
Log reached capacity (3145728 bytes); output disabled.

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```

This was the topic of https://github.com/web-platform-tests/wpt/issues/11454, which is closed, but clearly it's still a problem.

The stability checks on Taskcluster don't have this problem. @jugglinmike, do you have an ETA on turning off stability checks in Travis?

lgrahl:
> Are the tests fairly slow, how long do they take to run locally?

`RTCDataChannel-send-receive.html` takes ~45s for Chrome and `RTCDataChannel-bufferedAmountLowThreshold` ~30s, mainly because of the failures due to timeouts (an event should be fired but it isn't being fired, so we'll need to wait for it for a while).
`RTCDataChannel-close.html` takes only a few seconds for Chrome and the particular test mentioned above was never flaky for me locally. I'm also not seeing anything wrong with it but perhaps someone else wants to take a closer look: https://github.com/web-platform-tests/wpt/blob/99d792a605af57c390a39f7f817dad8135d0c7d2/webrtc/RTCDataChannel-close.html#L152:L184

The `NOTRUN` markings are puzzling me. This should never happen with a local timeout for each test in combination with the global timeout being disabled. Until I know why that happens, I don't think it's useful to run the tests again.

If no one else can look into it, then I will have to put this PR on hold until I have some more spare time.

lgrahl:
We can disable the flaky ones but I need to see which ones are flaky for Firefox first. There must be some flaky ones since Travis is failing. @jan-ivar can you or someone else from Mozilla take a look? I am not able to reproduce any *flakyness* on my machine.

> If we start splitting, we could also split modified existing tests/files from new ones.

We can split tests into logical blocks if necessary. I'm not expecting a lot of conflicts and trying to figure out which ones existed prior to this changeset would be a Sisyphean task.

nils-ohlmeier:
I found why WPT is logging so much ICE logging: because it uses the default log variables from mozrunner.

I'm currently trying to find a way to change the logging variables in WPT to produce less verbose logging when executed in the WPT environment.

youennf:
Looking at https://github.com/web-platform-tests/wpt/issues/11029#issuecomment-389499369 and https://github.com/web-platform-tests/wpt/issues/11120, WPT seems to go away from setting timeout in individual tests.

lgrahl:
Note to self: Needs updating to some tests due to https://github.com/w3c/webrtc-pc/pull/1999

alvestrand:
@youennf will try to get the impractical (multi-minute) tests disabled or removed; we should merge-and-iterate on the rest.

lgrahl:
@alvestrand The main problem seems to be the design choice now (local timeouts for each test, disabled global timeout). Since I have a little bit of time, a suggestion: I'll refactor one of the larger test groups and you folks tell me if it looks good to you. Then I'll proceed with refactoring the rest. Sounds good?

youennf:
Current plan is to split this PR as smaller pieces:
- 1 PR for RTCPeerConnection-helper.js changes
- 1 or more PRs for updated test files
- 1 or more PRs for new test files
Some of the new test files may take a very long time.
I am not sure we will be able to land all of them if it cannot be easily run in browsers CI.

lgrahl:
To do notes (mostly for myself):

- [ ]  PR for `RTCSctpTransport` changes

webrtc/RTCSctpTransport-onstatechange.html
```diff
@@ -0,0 +1,56 @@
...
+// Test is based on the following revision:
+// https://rawgit.com/w3c/webrtc-pc/1cc5bfc3ff18741033d804c4a71f7891242fb5b3/webrtc.html
+
+/*
+  4.4.2.  RTCPeerConnection.close
+    7.    Set the [[SctpTransportState]] slot of connection's [[SctpTransport]] to "closed".
+ */
+
+promise_test(async (t) => {
+  const resolver = new Resolver();
```
jan-ivar: `Resolver` is the deferred anti-pattern, which I'd like to get rid of.

webrtc/RTCSctpTransport-onstatechange.html
```diff
@@ -0,0 +1,56 @@
...
+  const pc1 = new RTCPeerConnection();
+  const pc2 = new RTCPeerConnection();
+  t.add_cleanup(() => pc1.close());
+  t.add_cleanup(() => pc2.close());
+
+  // We use the queue to hold back candidates, so no connection can be established
+  const pc1Queue = new IceCandidateQueue(pc1);
+  const pc2Queue = new IceCandidateQueue(pc2);
```
jan-ivar: This being the lone use of *IceCandidateQueue*, I propose using a promise as a simple queue here instead:
```js
promise_test(async (t) => {
  const pc2 = new RTCPeerConnection();
  t.add_cleanup(() => pc2.close());
  // We use a queue to hold back candidates, so no connection can be established
  let flush;
  const queue = new Promise(r => flush = r);
  pc1.onicecandidate = e => queue.then(() => pc2.addIceCandidate(e.candidate));
  pc2.onicecandidate = e => queue.then(() => pc1.addIceCandidate(e.candidate));

  // Negotiate SCTP transport
  pc1.createDataChannel('test');
  await doSignalingHandshake(pc1, pc2);

  for (const pc of [pc1, pc2]) {
    assert_equals(pc.sctp.state, 'connecting',
      'RTCSctpTransport should be in the connecting state');
  }
  flush(); // Let the peers connect
  await Promise.all([pc1, pc2].map(async pc => {
    await new Promise(r => pc.sctp.onstatechange = r);
    assert_equals(pc.sctp.state, 'connected');
  }));
}
```
This should also fix that the test doesn't bail before both `pc`s have had their `statechange` event tested.

jan-ivar: @lgrahl Odd, looks like github had some old review comments cached here. The last comment is the one I wanted you to look at.

jan-ivar: @youennf ^

jan-ivar: The first unresolved comment thread is the one about splitting this up. Presumably that's still happening? I just commented here because I couldn't find this test in the newer PRs. If we split it up and avoid modifying *exchangeIceCandidates* to take a queue, then I'm happy.

alvestrand:
Status Dec 18: Waiting for split-up PRs to land, then we'll close this one.

alvestrand:
Closing this PR on the assumption that split-up PRs have landed.

PR #13509, opened 2018-10-14
 Remove Tip note in text-transform/ tests

### LINKED ISSUE #13505: Incorrect "tip" in text transform tests
As discussed in https://bugzilla.mozilla.org/show_bug.cgi?id=1498148#c40, a [bunch of text transform tests](https://searchfox.org/mozilla-central/search?q=Tip%3A%20To%20identify%20the%20characters%20where%20differences%20occur) give the suggestion to copy and paste the result of the transform to in some unicode tool to check if the correct characters have been produced. However, text-transform is not supposed to affect copy and paste, so that's wrong, and that "tip" should be removed. Maybe also the tests that cannot be evaluated without relying on that tip, if any, should be rewritten.

dpino:
Fixes #13505.

The PR removes the Tip note entirely in all the `text-transform/` tests that include such note. As commented, in #13505 there might be a part of the tip that is relevant. I can reintroduce it if it's valuable. I think in case that part of the tip needs to be rephrased.

In addition, the PR removes some extra whitespace in the `html` tag of many tests.


frivoal (approved):

frivoal:
Looks good to me, approved.

However, I'm not sure why the travis task failed when testing stability on firefox. I've restarted it for now, but if it keeps on failing we'll have to look into it.

sideshowbarker:
> Still failing. Looks like an infrastructure problem to me, but I'm not sure.
> @gsnedders @foolip @Ms2ger Any idea how to interpret this failure (and what to do about it)? https://travis-ci.org/web-platform-tests/wpt/jobs/441391903

Raised https://github.com/web-platform-tests/wpt/issues/13513

foolip:
Thanks for reporting! Let me admin merge this, it's still possible to investigate by creating an equivalent PR from the same point in history.

r12a:
See my comment at https://github.com/web-platform-tests/wpt/issues/13505#issuecomment-430294917

PR #13510, opened 2018-10-15
[css-text] Rework some tests to avoid anti aliasing issues

frivoal:
As described in https://bugzilla.mozilla.org/show_bug.cgi?id=1498698#c1
The tests introduced in https://github.com/web-platform-tests/wpt/pull/11373
fail due to anti aliasing issues that let some red show through at the edges
even when things are correctly aligned and the red should be hidden.

Rework these tests to avoid the problem.


foolip:
@frivoal looks like there are conflicts. I tried to resolve them but they're not trivial enough. Can you resolve and I'll do light review?

PR #13516, opened 2018-10-15
Add URL parsing test for "ssh://example.com/foo/bar.git"

foolip:
Exercised in url-constructor.html and url/url-origin.html.

Points of disagreement:
 * `protocol` should be "ssh:" because it is simply the concatenation of
   scheme and ":": https://url.spec.whatwg.org/#dom-url-protocol. Set to
   "ssh" here: https://url.spec.whatwg.org/#scheme-state
 * `origin` should be "null" because the URL will be get an opaque
   origin in https://url.spec.whatwg.org/#concept-url-origin.
 * `host`, `hostname` and `pathname` require stepping through the parser
   to see why the result is what it is:
   https://url.spec.whatwg.org/#host-state
   https://url.spec.whatwg.org/#path-state

Based on https://felixfbecker.github.io/whatwg-url-custom-host-repro/
for https://github.com/webcompat/web-bugs/issues/19792.

### DIFF
--- a/url/resources/urltestdata.json
+++ b/url/resources/urltestdata.json
@@ -1433,6 +1433,22 @@
     "search": "",
     "hash": ""
   },
+  "# Based on https://felixfbecker.github.io/whatwg-url-custom-host-repro/",
+  {
+    "input": "ssh://example.com/foo/bar.git",
+    "base": "http://example.org/",
+    "href": "ssh://example.com/foo/bar.git",
+    "origin": "null",
+    "protocol": "ssh:",
+    "username": "",
+    "password": "",
+    "host": "example.com",
+    "hostname": "example.com",
+    "port": "",
+    "pathname": "/foo/bar.git",
+    "search": "",
+    "hash": ""
+  },
   "# Based on http://trac.webkit.org/browser/trunk/LayoutTests/fast/url/file.html",
   {
     "input": "file:c:\\foo\\bar.html",


foolip:
Browser results:
* Chrome Canary passes url-origin.html but fails url-constructor.html because `host` is "". (I'll find a bug.)
* Edge 17 passes url-constructor.html but fails url-origin.html because `origin` is "ssh://" (@thejohnjansen, can you find a contact?)
* Firefox Nightly status is the same as Chrome Canary (`host` is "") (@annevk, who to poke?)
* Safari Technology Preview 67 passes url-constructor.html but fails url-origin.html because `origin` is "ssh://example.com". (@achristensen07 FYI)

annevk:
As I noted in https://github.com/whatwg/url/issues/418 to @felixfbecker I think this is already tested by equivalent tests, though I'm happy to add this specific variant. I'm also happy with other people caring about this and poking browsers.

For Firefox it's @valenting, but I think he's aware and trying to get the general infrastructure in such a shape it can eventually support a different URL parser. Not really sure what the status is though.

foolip:
@domenic as mentioned on IRC, your review on this would be much appreciated.

domenic (approved):
Passes with jsdom/whatwg-url.

achristensen07:
I am aware that Safari's behavior differs from the spec in the case of non-special origins.  Thanks for the ping, and I'm glad to see such a wave of interest in spec compliance.

felixfbecker:
Filed on Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=902253
Filed on Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1504962

foolip:
Thanks @felixfbecker!

PR #13518, opened 2018-10-15
[Resource Timing] Test XO redirection sandwich with and without TAO

### LINKED ISSUE #152: `origin-or-null` definition and TAO processing
I'm failing to find if and where `origin-or-null` is defined. The "null" case is also not addressed in the "timing allow check" processing model.

IIUC, we need to:
* Add a definition for `origin-or-null`. I'm not sure if the `origin` part should be a token or there are more specific syntax definitions at https://tools.ietf.org/html/rfc7230#appendix-B or elsewhere (CCing @reschke for his opinion)
* Point the origin comparison in the processing model to https://tools.ietf.org/html/rfc6454#section-5 as https://tools.ietf.org/html/rfc6454#section-4 doesn't really return a string.
* Add tests that make sure that origin comparison in TAO also works for non-standard ports, doesn't work for different schemes, etc.
* Add a `null` related condition in the processing model.

Thoughts?

yoavweiss:
Add a test to make sure that a Same-Origin=>Cross-Origin=>Same-origin redirection chain is not exposing timing information unless Timing-Allow-Origin is set.

Partially fixes https://github.com/w3c/resource-timing/issues/152

### DIFF
--- a/common/get-host-info.sub.js
+++ b/common/get-host-info.sub.js
@@ -3,9 +3,12 @@ function get_host_info() {
   var HTTP_PORT = '{{ports[http][0]}}';
   var HTTP_PORT2 = '{{ports[http][1]}}';
   var HTTPS_PORT = '{{ports[https][0]}}';
+  var PROTOCOL = window.location.protocol;
+  var IS_HTTPS = (PROTOCOL == "https:");
   var HTTP_PORT_ELIDED = HTTP_PORT == "80" ? "" : (":" + HTTP_PORT);
   var HTTP_PORT2_ELIDED = HTTP_PORT2 == "80" ? "" : (":" + HTTP_PORT2);
   var HTTPS_PORT_ELIDED = HTTPS_PORT == "443" ? "" : (":" + HTTPS_PORT);
+  var PORT_ELIDED = IS_HTTPS ? HTTPS_PORT_ELIDED : HTTP_PORT_ELIDED;
   var ORIGINAL_HOST = '{{host}}';
   var REMOTE_HOST = (ORIGINAL_HOST === 'localhost') ? '127.0.0.1' : ('www1.' + ORIGINAL_HOST);
   var OTHER_HOST = '{{domains[www2]}}';
@@ -18,10 +21,12 @@ function get_host_info() {
     ORIGINAL_HOST: ORIGINAL_HOST,
     REMOTE_HOST: REMOTE_HOST,
 
+    ORIGIN: PROTOCOL + "//" + ORIGINAL_HOST + PORT_ELIDED,
     HTTP_ORIGIN: 'http://' + ORIGINAL_HOST + HTTP_PORT_ELIDED,
     HTTPS_ORIGIN: 'https://' + ORIGINAL_HOST + HTTPS_PORT_ELIDED,
     HTTPS_ORIGIN_WITH_CREDS: 'https://foo:bar@' + ORIGINAL_HOST + HTTPS_PORT_ELIDED,
     HTTP_ORIGIN_WITH_DIFFERENT_PORT: 'http://' + ORIGINAL_HOST + HTTP_PORT2_ELIDED,
+    REMOTE_ORIGIN: PROTOCOL + "//" + REMOTE_HOST + PORT_ELIDED,
     HTTP_REMOTE_ORIGIN: 'http://' + REMOTE_HOST + HTTP_PORT_ELIDED,
     HTTP_NOTSAMESITE_ORIGIN: 'http://' + NOTSAMESITE_HOST + HTTP_PORT_ELIDED,
     HTTP_REMOTE_ORIGIN_WITH_DIFFERENT_PORT: 'http://' + REMOTE_HOST + HTTP_PORT2_ELIDED,
--- a/resource-timing/crossorigin-sandwich-TAO.sub.html
+++ b/resource-timing/crossorigin-sandwich-TAO.sub.html
@@ -3,7 +3,6 @@
 <head>
 <meta charset="utf-8" />
 <title>This test validates resource timing information for a same-origin=>cross-origin=>same-origin redirect chain with Timing-Allow-Origin.</title>
-<link rel="author" title="Intel" href="http://www.intel.com/" />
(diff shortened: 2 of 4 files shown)


annevk:
Going from the description, do you require Timing-Allow-Origin on each response in the chain?

yoavweiss:
> Going from the description, do you require Timing-Allow-Origin on each response in the chain?

The "with TAO" test does indeed have TAO on each response other than the last one (which is same origin). Is there value in making sure that *only* cross-origin responses have TAO?

annevk:
The last one also needs to have it, if you ever went cross-origin. Otherwise you have a different design from CORS which seems bad for security (as I tried to explain in the corresponding issue and maybe also some other PR in  that repo).

resource-timing/crossorigin-sandwich-TAO.sub.html
```diff
@@ -0,0 +1,48 @@
+<!DOCTYPE HTML>
+<html>
+<head>
+<meta charset="utf-8" />
+<title>This test validates resource timing information for a same-origin=>cross-origin=>same-origin redirect chain with Timing-Allow-Origin.</title>
+<link rel="author" title="Intel" href="http://www.intel.com/" />
```
npm1: Not the right author

resource-timing/crossorigin-sandwich-TAO.sub.html
```diff
@@ -0,0 +1,48 @@
...
+
+<script>
+    setup({explicit_done: true});
+    test_namespace('getEntriesByName');
+    const pageOrigin = get_host_info()['ORIGIN'];
+    const crossOrigin = get_host_info()['REMOTE_ORIGIN'];
+
+    function onload_test()
+    {
+        const context = new PerformanceContext(performance);
```
npm1: Do we need this? Could we replace |context| with |performance| below?

resource-timing/crossorigin-sandwich-no-TAO.sub.html
```diff
@@ -0,0 +1,47 @@
+<!DOCTYPE HTML>
+<html>
+<head>
+<meta charset="utf-8" />
+<title>This test validates resource timing information for a same-origin=>cross-origin=>same-origin redirect chain with Timing-Allow-Origin.</title>
+<link rel="author" title="Intel" href="http://www.intel.com/" />
```
npm1: ditto same comments

resource-timing/crossorigin-sandwich-no-TAO.sub.html
```diff
@@ -0,0 +1,47 @@
+<!DOCTYPE HTML>
+<html>
+<head>
+<meta charset="utf-8" />
+<title>This test validates resource timing information for a same-origin=>cross-origin=>same-origin redirect chain with Timing-Allow-Origin.</title>
```
npm1: without Timing-Allow-Origin

npm1:
Test looks good to me, we should have tests for these sandwiches now instead of waiting on the integration with fetch which will make it consistent with CORS. Do you mind adding a comment in multi_redirect.py to be precise about what it is doing?

common/get-host-info.sub.js
```diff
@@ -18,10 +21,12 @@ function get_host_info() {
     ORIGINAL_HOST: ORIGINAL_HOST,
     REMOTE_HOST: REMOTE_HOST,
 
+    ORIGIN: PROTOCOL + "//" + ORIGINAL_HOST + PORT_ELIDED,
```
npm1: WPTs are served on HTTP by default. These don't seem to be needed because there is no ambiguity.

resource-timing/crossorigin-sandwich-TAO.sub.html
yoavweiss: As this is a copy of the existing tests, not sure what "author" should say, tbh. Seems best remove it entirely.

resource-timing/crossorigin-sandwich-TAO.sub.html
yoavweiss: We definitely could!

resource-timing/crossorigin-sandwich-no-TAO.sub.html
yoavweiss: Good catch!

resource-timing/crossorigin-sandwich-no-TAO.sub.html
yoavweiss: yup

common/get-host-info.sub.js
yoavweiss: When run on the bots, sure. But on w3c-test.org, they can be accessed from both http and https. This makes the test more robust, so I prefer to keep it (unless WPT folks object for some reason)

yoavweiss:
@annevk - PTAL. This tests the current behavior that's specified and implemented for TAO, which is to not require TAO on same-origin after a cross-origin redirect. I plan to try and align the behavior with CORS as part of L3, but want to first document and test what's implemented today.

resource-timing/resources/multi_redirect.py
```diff
@@ -13,27 +13,31 @@ def main(request, response):
...
     if "timing_allow" in request.GET:
         timing_allow = request.GET.first("timing_allow")
 
-    redirect_url = "/resource-timing/resources/multi_redirect.py?"
-    redirect_url += "page_origin=" + page_origin
-    redirect_url += "&cross_origin=" + cross_origin
-    redirect_url += "&timing_allow=" + timing_allow
-    redirect_url += "&step="
+    redirect_url_path = "/resource-timing/resources/multi_redirect.py?"
+    redirect_url_path+= "page_origin=" + page_origin
```
npm1: nit: space before +=

npm1 (approved):
LGTM

resource-timing/resources/multi_redirect.py
yoavweiss: yup

Hexcles:
Force-merging because a file in `common/` was modified, affecting too many tests and causing stability checks to time out.

PR #13519, opened 2018-10-15
Fix encoding/resources/two-boms-utf-16[be|le].html which were corrupted

hsivonen:
These were imported by moz-wptsync-bot in
changeset 3d778629a3948660e510f8b4f2b22c26fa4ad727 and were corrupted in
the process. This is a manual copy of the two files from the Gecko repo
copy of wpt.


hsivonen:
@jgraham, where should the moz-wptsync-bot bug be handled?

annevk (approved):
I'm gonna trust you did this correctly, but presumably we need something to prevent this from happening again?

hsivonen:
> I'm gonna trust you did this correctly

Thanks.

> where should the moz-wptsync-bot bug be handled?
...
> presumably we need something to prevent this from happening again

It appears that there are a couple of moz-wptsync-bot bugs in Bugzilla, so I [filed a bug there](https://bugzilla.mozilla.org/show_bug.cgi?id=1499036).

PR #13524, opened 2018-10-15
HTTP: message parsing with CR

annevk:
See https://github.com/httpwg/http-core/issues/31 for context.

### DIFF
--- a/fetch/h1-parsing/lone-cr.window.js
+++ b/fetch/h1-parsing/lone-cr.window.js
@@ -10,8 +10,11 @@
   "HTTP/1.1 200 OK\n\rHeader: Value\n\nBODY",
   "HTTP/1.1 200 OK\nHeader\r: Value\n\nBODY",
   "HTTP/1.1 200 OK\nHeader:\r Value\n\nBODY",
+  "HTTP/1.1 200 OK\nHeader: Value\r\n\nBody",
   "HTTP/1.1 200 OK\nHeader: Value\r\r\nBODY",
   "HTTP/1.1 200 OK\nHeader: Value\rHeader2: Value2\n\nBODY",
+  "HTTP/1.1 200 OK\nHeader: Value\n\rBODY",
+  "HTTP/1.1 200 OK\nHeader: Value\n\r"
 ].forEach(input => {
   promise_test(t => {
     const message = encodeURIComponent(input);
--- a/fetch/h1-parsing/lone-cr.window.js
+++ b/fetch/h1-parsing/lone-cr.window.js
@@ -1,0 +1,21 @@
+// These tests expect that a network error is returned if there's a CR that is not immediately
+// followed by LF before reaching message-body.
+//
+// No browser does this currently, but Firefox does treat it equivalently to a space which gives
+// hope.
+
+[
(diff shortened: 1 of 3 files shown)


foolip:
@annevk can you rebase this to trigger new CI checks and ping a reviewer? I can review as a fallback.

annevk:
I have rebased this, but taking this further would also require figuring out what behavior we want to align on.

foolip:
I see, spec issue is still open. Sorry for the noise then :)

annevk:
Looking at these tests, browsers, and what the HTTP WG concluded they do not really seem to match up much. Firefox is pretty close in treating CR as if it was SP, but doesn't actually transform it. Chrome and Safari seem a lot more lenient as the current (probably wrong) pass conditions assume.

@mnot @MattMenke2 @ddragana @youennf I'd appreciate thoughts on how we can get closer to interoperability here.

MattMenke2:
My preference for alignment would be to reject responses with CR.  The fact that FireFox treats these like LWS characters gives me hope the fallout won't be too bad.  I think filing bugs and adding tests is the way to go here.  I can't make any guarantees about how fast Chrome will be to change behavior here, since I'm not on the network team, and we're being increasingly strict about the launch process around breaking changes.

annevk:
If you see a lone CR (i.e., not followed immediately by LF) before the message-body in particular (using RFC7230 parlance), is what you mean? And otherwise CRLF and LF would be treated the same? That sounds reasonable to me as a target.

MattMenke2:
Yes, exactly.

annevk:
Bugs:
* https://bugs.chromium.org/p/chromium/issues/detail?id=1174095
* https://bugs.webkit.org/show_bug.cgi?id=221332
* https://bugzilla.mozilla.org/show_bug.cgi?id=1690591

fetch/h1-parsing/lone-cr.window.js
```diff
@@ -0,0 +1,20 @@
...
+// No browser does this currently, but Firefox does treat it equivalently to a space which gives
+// hope.
+
+[
+  "HTTP/1.1\r200 OK\n\nBODY",
+  "HTTP/1.1 200\rOK\n\nBODY",
+  "HTTP/1.1 200 OK\n\rHeader: Value\n\nBODY",
+  "HTTP/1.1 200 OK\nHeader\r: Value\n\nBODY",
+  "HTTP/1.1 200 OK\nHeader:\r Value\n\nBODY",
+  "HTTP/1.1 200 OK\nHeader: Value\r\r\nBODY",
```
MattMenke2, line 13: Suggest the following:

* \r before the body, since the final linebreak is potentially special.  e.g.
  "HTTP/1.1 200 OK\nHeader: Value\n\rBODY"
* \r without a body, which is also potentially a bit special.
   "HTTP/1.1 200 OK\nHeader: Value\n\r"
* One with just \r after a value (first line is potentially special, so best to have a \r at a reasonable place after a header line)
  "HTTP/1.1 200 OK\nHeader: Value\r\nBody"

MattMenke2 (approved):

PR #13525, opened 2018-10-15
Fixed the insertAdjacentHTML tests.

koto:
Firefox merges the text nodes when afterbegin / beforeend is used. This doesn't seem to be specced, and the domparsing tests don't capture that behavior; removed this Chrome behavior assertions.


mikewest (approved):
LGTM.

PR #13530, opened 2018-10-15
[docs] Deprecate `async_test`

jugglinmike:
This may be a little too aggressive. [We know that test contributors dislike the `step`-based API of `async_test`](http://lists.w3.org/Archives/Public/public-test-infra/2018JulSep/0011.html), but [using `promise_test` for a callback-driven API carries it's own overhead](http://lists.w3.org/Archives/Public/public-test-infra/2018OctDec/0000.html). The web platform may not be sufficiently Promise-based to make `promise_test` an ergonomic default. There may also be other reasons to keep `async_test` around (e.g. cases where parallel testing is a hard requirement). A pull request seemed like the best way to prompt the discussion.

### DIFF
--- a/docs/_writing-tests/testharness-api.md
+++ b/docs/_writing-tests/testharness-api.md
@@ -66,7 +66,7 @@ The function passed in is run in the `test()` call.
 test. Currently it is only used to provide test-specific
 metadata, as described in the [metadata](#metadata) section below.
 
-## Asynchronous Tests ##
+## **DEPRECATED** Asynchronous Tests ##
 
 Testing asynchronous features is somewhat more complex since the result of
 a test may depend on one or more events or other callbacks. The API provided
@@ -152,6 +152,11 @@ document.documentElement.addEventListener("DOMContentLoaded",
 
 Keep in mind that other tests could start executing before an Asynchronous
 Test is finished.
+
+This is deprecated because the `step`-based API introduces overhead which is
+difficult to maintain. New asynchronous tests should be defined using
+`promise_test` instead (see also the `EventWatcher` interface for succinctly
+testing evented APIs with promises).
 
 ## Promise Tests ##
 


jgraham (changes requested):
I am strongly opposed to this change for the reasons set out in my comment.

foolip:
This is ambitious, well worth considering I think.

This is a bit of a detour, but bear with me. We now have `promise_test` which you can use as `promise_test(async t => { ... })`, an `async_test` that *can't* be used with "async functions", and a `test` which also doesn't allow any value to be returned from the callback.

This makes it not too obvious how to write async tests I think.

youennf:
FWIW, in WebKit, I am pushing people to use one async_test in a file, otherwise use promise_test.
We have some issues with unreliable ordering of console logging with several async_test in a given file.

youennf:
To be clear, I am not pushing for async_test, I am pushing against more than one async_test in a file...

foolip:
Oops, posted too early.

It would be quite nice if we had `test` and a new `parallel_test` both able to accept thenable-vending callbacks. Details to be worked out, but basically the current flavor of `async_test` would be one of the modes of `parallel_test`, the other being parallel promise tests which currently cannot be done.

I think we shouldn't really be deprecating things but instead doing the work to change all the call sites if we think it's important enough. The name `async_test` is very entrenched, but it's also very unfortunate how it collides with modern JS nomenclature, and that might be fixable with some effort.

foolip:
> We have some issues with unreliable ordering of console logging with several async_test in a given file.

We have this in Blink too. We didn't really solve it I'm afraid, just marked those few tests as flaky. There weren't enough in the end to justify doing more we thought. @Hexcles.

jugglinmike:
@youennf For [single page tests](https://github.com/web-platform-tests/wpt/blob/4ba3cd24c69a859d8c7864d15958d2656ef16173/docs/_writing-tests/testharness-api.md), there's no need to use `t.step`, `t.step_func`, etc.  because uncaught exceptions are interpreted as errors. Given that you have other reasons to encourage a 1-to-1 test-to-file ratio, advocating for single page tests instead would reduce the overhead for test authors. That might make it a convenient benefit.

Pardon me if you're aware of this; Simon and Jory's survey found that not everyone knows about the feature.

youennf:
Right!
I heard about single page tests recently by Simon and agrees this looks nicer.
I actually started advertising it and will continue doing so.

foolip:
I realize I've confused two different issues: subtest ordering and console output. For console output I think we ended up redirecting it to stderr so that it's not part of the baseline, and for subtest order we considered fixing it, but it just doesn't affect very many tests so we left it be. I don't think either of these concerns are reasons to change/replace `async_test`, especially log output which affects all test types.

stephenmcgruer:
This never got much traction, so closing. If anyone was interested in pushing for this, nowadays [an RFC](https://github.com/web-platform-tests/rfcs) would be required.

Issue #8368, opened 2017-11-20
Creation of a handler which doesn't set a string as response content hangs the server


wpt-issue-mover:
Originally posted as https://github.com/w3c/wptserve/issues/105 by @whimboo on 28 Oct 2016, 12:22 UTC:

> The following handler will cause the server to hang while delivering the page content:
> 
> > ```
> >   def handler(request, response):
> >         response.content = 1
> >         return ()
> > ```
> 
> This happens because a non-string value gets set as the content.

jgraham:
I can't see why this should hang, but possibly in `tools/wptserve/wptserve/response.py::Response.iter_content` we should have a check that the content actually is iterable before we try to iterate over it and in case it is not raise a `ValueError`.

teh-f00l:
Hope y'all don't mind me working on this one!

qiuzhong:
Is anyone still working on this issue?

In `tools/wptserve/wptserve/response.py::Response.iter_content` it doesn't check if `self.content` is iterable. However, the following code then uses it to iterate directly:
https://github.com/web-platform-tests/wpt/blob/434ca4744845966d5f3f87355f41ccc6f777376f/tools/wptserve/wptserve/response.py#L214

If a bad handler assigns inappropriate type that is not iterable to the response content, above iteration will raise a `TypeError`. 

However, I didn't encounter server hanging. In my tries, this error was caught in `tools/wptserve/wptserve/server.py::Http1WebTestRequestHandler.handle_one_request` and returned a `500` HTTP error immediately.

PR #13534, opened 2018-10-16
Add iterable checking for response content

qiuzhong:
We need check if the response content is iterable before we iterable
over it. Otherwise, it might cause the server hanging. See #8368

### DIFF
--- a/tools/wptserve/wptserve/response.py
+++ b/tools/wptserve/wptserve/response.py
@@ -1,4 +1,4 @@
-from collections import OrderedDict
+from collections import OrderedDict, Iterable
 from datetime import datetime, timedelta
 from six.moves.http_cookies import BaseCookie, Morsel
 import json
@@ -183,6 +183,9 @@ class Response(object):
         True, the entire content of the file will be returned as a string facilitating
         non-streaming operations like template substitution.
         """
+        if not isinstance(self.content, Iterable) and not hasattr(self.content, "read"):
+            raise TypeError("Response content must be an iterable or file-like object!")
+
         if isinstance(self.content, binary_type):
             yield self.content
         elif isinstance(self.content, text_type):
--- a/tools/wptserve/wptserve/response.py
+++ b/tools/wptserve/wptserve/response.py
@@ -1,4 +1,4 @@
-from collections import OrderedDict, Iterable
+from collections import OrderedDict
 from datetime import datetime, timedelta
 from six.moves.http_cookies import BaseCookie, Morsel
 import json
@@ -94,6 +94,15 @@ class Response(object):
         else:
             self._status = (int(value), None)
 
+    def can_iter_content(self):
+        try:
+            iter(self.content)
+            return True
+        except TypeError:
+            if hasattr(self.content, "read"):
+                return True
+            return False
+
     def set_cookie(self, name, value, path="/", domain=None, max_age=None,
                    expires=None, secure=False, httponly=False, comment=None):
         """Set a cookie to be sent with a Set-Cookie header in the
@@ -183,9 +192,6 @@ class Response(object):
         True, the entire content of the file will be returned as a string facilitating
         non-streaming operations like template substitution.
         """
-        if not isinstance(self.content, Iterable) and not hasattr(self.content, "read"):
-            raise TypeError("Response content must be an iterable or file-like object!")
-
         if isinstance(self.content, binary_type):
             yield self.content
         elif isinstance(self.content, text_type):
--- a/tools/wptserve/tests/functional/test_handlers.py
+++ b/tools/wptserve/tests/functional/test_handlers.py
@@ -193,17 +193,29 @@ class TestFunctionHandler(TestUsingServer):
         assert "Content-Length" not in resp.info()
         assert resp.read() == b""
 
-    def test_non_iterable_content(self):
+    def test_non_iterable_rv(self):
         @wptserve.handlers.handler
         def handler(request, response):
-            response.content = 1234
-            return response
+            response.content = 1
+            return ()
 
         route = ("GET", "/non_iterable_rv", handler)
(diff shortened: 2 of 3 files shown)


tools/wptserve/wptserve/response.py
```diff
@@ -183,6 +183,9 @@ def iter_content(self, read_file=False):
         True, the entire content of the file will be returned as a string facilitating
         non-streaming operations like template substitution.
         """
+        if not isinstance(self.content, Iterable) and not hasattr(self.content, "read"):
```
Ms2ger: I don't think the `Iterable` check works in general.

tools/wptserve/wptserve/response.py
```diff
@@ -183,6 +183,9 @@ def iter_content(self, read_file=False):
...
+            raise TypeError("Response content must be an iterable or file-like object!")
```
Ms2ger: Raising an exception here doesn't work; that only works inside the `handler()` call.

```
________________ TestFunctionHandler.test_non_iterable_content _________________
self = <functional.test_handlers.TestFunctionHandler testMethod=test_non_iterable_content>
    def test_non_iterable_content(self):
        @wptserve.handlers.handler
        def handler(request, response):
            response.content = 1234
            return response
    
        route = ("GET", "/non_iterable_rv", handler)
        self.server.router.register(*route)
        resp = self.request(route[1])
>       self.assertEqual(500, resp.getcode())
E       AssertionError: 500 != 200
wptserve/tests/functional/test_handlers.py:205: AssertionError
----------------------------- Captured stderr call -----------------------------
DEBUG:web-platform-tests:Route pattern: ^/(.*)$
DEBUG:web-platform-tests:Route pattern: ^/(.*\.asis)$
DEBUG:web-platform-tests:Route pattern: ^/(.*\.py)$
DEBUG:web-platform-tests:Using default configuration
INFO:web-platform-tests:Starting http server on localhost:35117
DEBUG:web-platform-tests:Route pattern: ^/non\_iterable\_rv$
DEBUG:web-platform-tests:GET /non_iterable_rv
DEBUG:web-platform-tests:Found handler FunctionHandler
DEBUG:web-platform-tests:200 GET /non_iterable_rv (None) 0
ERROR:web-platform-tests:Traceback (most recent call last):
  File "/home/travis/build/web-platform-tests/wpt/tools/wptserve/wptserve/server.py", line 544, in handle_one_request
    self.finish_handling_h1(request_line_is_valid)
  File "/home/travis/build/web-platform-tests/wpt/tools/wptserve/wptserve/server.py", line 243, in finish_handling_h1
    self.finish_handling(request, response, handler)
  File "/home/travis/build/web-platform-tests/wpt/tools/wptserve/wptserve/server.py", line 278, in finish_handling
    response.write()
  File "/home/travis/build/web-platform-tests/wpt/tools/wptserve/wptserve/response.py", line 223, in write
    self.write_content()
  File "/home/travis/build/web-platform-tests/wpt/tools/wptserve/wptserve/response.py", line 217, in write_content
    for item in self.iter_content():
  File "/home/travis/build/web-platform-tests/wpt/tools/wptserve/wptserve/response.py", line 187, in iter_content
    raise TypeError("Response content must be an iterable or file-like object!")
TypeError: Response content must be an iterable or file-like object!
```

tools/wptserve/wptserve/response.py
Ms2ger: Also, this check should go in the `else` block below, IMO.

tools/wptserve/wptserve/response.py
Ms2ger: Also, this yields a pretty useless stack trace. Can we turn the response attribute into a getter/setter and check in the setter?

Ms2ger (changes requested):

tools/wptserve/wptserve/response.py
gsnedders: No, it definitely doesn't. The vast majority of iterables don't descend from it.

qiuzhong: [Python documentation](https://docs.python.org/2.7/library/collections.html#collections.Iterable) says 
> `collections.Iterable`: ABC for classes that provide the `__iter__()` method. 

`isinstance(xxx, collections.Iterable)` returns `True` for all the sequence types and `dict` types and `file` objects. 
It's not the best way for checking iterable. Maybe I need to remove this `if` statement and add a `try...catch` statement whenever calling the `item_content` method.

tools/wptserve/wptserve/response.py
qiuzhong: What I thought was any non-interable response content will raise a `TypeError` and an HTTP request gets a `500` status error. I assigned a number to `response.content` in handler on purpose. But obviously, it didn't work. Any good solution suggestion?

Sorry for this PR that didn't pass the unit test. I failed to set up CI environment before. My local CI environment for tools unit tests is set up successfully now. I'll run more unit test before I file any PRs.

qiuzhong:
This PRs derives from #8638. In its comments, @jgraham suggested that  
> we should have a check that the content actually is iterable before we try to iterate over it and in case it is not raise a `ValueError`.

Indeed, an inappropriate handler (on purpose) like this
```
def handler(request, response):
	response.content = 1
	return response.content
```
sets the response content an int type, which is not be used for iteration. Such case was not handled well by `tools/wptserve/wptserve/server.py::BaseWebTestRequestHandler.finish_handling`: calling `response.write()` will finally raise a `TypeError` and the testing server crashes. This PR tries to check if the response content iterable before the response writes its content. 
I added the iterable checking in `iter_content()` method before, but it didn't work. So I removed it completely  and used a new way to check if the response content is an iterable.

qiuzhong:
@Ms2ger, @gsnedders, @jgraham , would you please take a look at it and maybe give some suggestion?

PR #13559, opened 2018-10-17
Fetch: test nosniff parsing better

annevk:


annevk:
@MattMenke2 I guess what you noticed in Chrome about empty headers (which I hadn't noticed since they are serialized) is indeed true. E.g., in this nosniff test Chrome ends up ignoring empty values altogether. Firefox only does it when they are separate header lines.

MattMenke2:
Except for certain headers, Chrome unconditionally splits header lines around commas into multiple header values, unconditionally allowing commas between both single and double quoted strings.  This code also ignores empty values.  A lot of parsers use the values output by this method, instead of the actual header lines we received, to avoid needing redundant code to split headers around commas.

It does *mostly* allow retrieving the original header lines, unmodified, though even there, believe it will remove the leading and trailing commas (Since it goes from the start of the first parsed value to the end of the last, but leading/trailing commas don't add a value).

annevk:
That kind of generic handling makes sense to me, but it seems wrong to drop empty values as you argued elsewhere (though before discovering Chrome did so). It also seems wrong to have single quoted strings as they are not a thing in HTTP I think?

Irrespective of what Chrome does today, what do you think the future should look like here?

MattMenke2:
The problem with generic quote handling is that Content-Type, at least, specifies its own quote handling, which isn't compatible with it:

Content-Type: text/html; foo= ",";  charset=GBK

If you split around commas with quotes and without quote handling, you get different results.  Content-Type rules don't consider that a quoted string.  Of course, it could be that Content-Type should be considered a non-coalescing header, which resolves the issue, but anything that only allows quoted strings in certain parts of the header will potentially have similar problems., if commas should also separate multiple values.

I'd really like to be able to have some unambiguous rules for how to handle these situations - merge all instances of a header, separated by commas, works, though it then means that we either need a general way to separate in values around commas, independent of header, or Chrome's approach doesn't work.  Such rules don't currently exist, and I feel that specifying quote handling on a per-header basis isn't compatible with rules of that type, making everything essentially a non-coalescing header.  If common code to split lines doesn't work, that means a lot more code to handle each individual header type, which seems more likely to result in bugs and inconsistent handling of headers.  It has been looking to me like this is the direction we're going in (Which may be necessary, and I can live with, but does mean Chrome's current approach here is wrong).

I'd also like to have only a few parsing rules apply to all headers.  For instance, one rule for everything of the form "Foo: a; b=c; d=e, x; y=z, s, t" - consistent rules for quote handling (Which would apply to everything from Accept-Language to Content-Type, etc).

annevk:
Thanks! I agree with these goals. I was hoping to get to consistency by sorting out ideal parsing for a bunch of headers and then abstracting the common patterns as I go.

I do think we need to do combining and then parsing as otherwise there can be subtle differences if there's an intermediary that combines, which seems really bad for robustness.

Splitting on commas except for those quoted and then removing whitespace around the resulting tokens might be good enough, but it'll need a bunch of tests for various headers to show it's not obviously broken somehow.

To make progress here it would help if some of the refactoring PRs I created for Fetch and some of the tests against wpt, such as this one, can land. I could create a tracking bug against Chrome and other browsers to track this effort so we don't miss anything as we go, but also don't require anything to be patched before we are more convinced about the approach? Thoughts on that appreciated.

cc @ddragana @domenic @foolip @mnot

foolip:
@annevk, if this is a large effort that will span many PRs, and keeping track of order/dependency is hard, then you might consider using a GitHub project. We currently have just one in the web-platform-tests org :)

foolip:
@MattMenke2, can you review this PR?

fetch/nosniff/parsing-nosniff.html
```diff
@@ -3,7 +3,7 @@
 <div id=log></div>
 <script>
```
MattMenke2: Mind adding a comment, while you're here, on just what this test is doing?  It looks to me like an accurate description would be something like:

Tests whether variations of a "X-Content-Type-Options: nosniff" header can be parsed as such. When the response header fails to be parsed as a valid nosniff header, the advertised Content-Type is not recognized as a possible script, and the script's onerror handler is invoked.  When the nosniff header is not parsed as a valid nosniff header, the MIME type of the response is sniffed as text, and the onload handler is run.

fetch/nosniff/resources/nosniff-weird-tokens.asis
```diff
@@ -0,0 +1,6 @@
+HTTP/1.1 200 YOU HAVE NO POWER HERE
+Content-Length: 22
+Content-Type: x/x
+X-Content-Type-options: nosniff,,@#$#%%&^&^*()()11!
```
MattMenke2: Could also do a weird tokens test with the "@#$#%%&^&^*()()11!" first, to make sure their headers with weird tokens aren't just ignored.  Could additionally / alternatively do something similar for the tabulation test as well, adding another nosniff after the comma.

fetch/nosniff/resources/nosniff-line-tabulation.asis
```diff
@@ -0,0 +1,6 @@
+HTTP/1.1 200 YOU HAVE NO POWER HERE
+Content-Length: 22
+Content-Type: x/x
+X-Content-Type-options: nosniff,
```
MattMenke2: Maybe call this vertical-tab instead of line-tabulation?  At least I've always seen it caller vertical tab when looking at ascii tables and the like.

fetch/nosniff/resources/nosniff-first-empty-double.asis
```diff
@@ -0,0 +1,7 @@
+HTTP/1.1 200 YOU HAVE NO POWER HERE
+Content-Length: 22
+Content-Type: x/x
+X-Content-Type-options:
```
MattMenke2: Is options deliberately lowercase in these tests?  Shouldn't hurt, but as that's not what they're aimed at testing, should probably capitalize it.

fetch/nosniff/resources/nosniff-line-tabulation.asis
MattMenke2: Do we get anything from the comma in this one?

MattMenke2 (commented):
Tests look pretty thorough to me.  One other test we could add is one with a null at the end of nosniff.

annevk:
Thanks, those are good cases to test. I should probably rewrite this and instead of using of using `.asis`, change this to use a Python script so there's less duplication and more test information in a single resource.

annevk:
Thanks again for the review. I've now restructured the tests to be similar to the Content-Length tests and presumably all future header tests. I also think that this new setup negates the need for a comment, but let me know if you want one anyway.

I haven't tested 0x00 because of https://github.com/whatwg/xhr/issues/165. Implementations haven't been updated (some tests did land), but there we reached the conclusion that such a byte in the header block should result in a network error.

fetch/nosniff/resources/x-content-type-options.json
```diff
@@ -0,0 +1,58 @@
...
+  {
+  	"input": "X-Content-Type-Options: nosniff,,@#$#%%&^&^*()()11!",
+  	"nosniff": true
+  },
+  {
+  	"input": "X-Content-Type-Options: @#$#%%&^&^*()()11!,nosniff",
+  	"nosniff": false
+  },
+  {
+    "input": "X-Content-Type-Options: nosniff\r\nX-Content-Type-Options: no",
```
MattMenke2: nit:  Should probably use consistent indentation for all of these (x3)

fetch/nosniff/parsing-nosniff.window.js
```diff
@@ -0,0 +1,21 @@
+promise_test(() => fetch("resources/x-content-type-options.json").then(res => res.json()).then(runTests), "Loading JSON…");
+
+function runTests(allTestData) {
+  for (let i = 0; i < allTestData.length; i++) {
+    const testData = allTestData[i],
+          input = encodeURIComponent(testData.input);
+    async_test(t => {
+      const script = document.createElement("script");
+      t.add_cleanup(() => script.remove());
+      if (testData.nosniff) {
```
MattMenke2: I live completely in network land, so am far from knowledgeable about Javascript or the web platform, beyond HTTP (Admittedly, in Chrome, that also includes the MIME sniffing logic), but looking at the updated test, I still find it non-obvious why the nosniff results in errors.  The sniffed as text -> treated as valid Javascipt connection isn't obvious to me.

That having been said, I do think that this makes it much clearer that the onerror cases correspond to nosniff being respected.

Maybe just something along the lines of "The resource must be sniffed as test to be successfully loaded as a script."

annevk:
I fixed the formatting issues, flipped the expectation for 0x0C (see https://github.com/whatwg/fetch/pull/818#issuecomment-434283476 though), and added a comment explaining how the `script` element is used here.

MattMenke2 (approved):

PR #13561, opened 2018-10-17
 Make logging on taskcluster more space efficient

jgraham:
This fixes two issues

 * Remove the uncompressed wptreport artifact after creating a
   compressed one.

 * Use the tbpl format as the stdout logger.

Originally we didn't upload tbpl-format logs at all since they include
screenshot information and can be large. However this was
accidentially enabled when TaskCluster checks for PRs were introduced,
and the screenshots are useful. However the logs are rather large and
uncompressed. Explicitly compressing them like we do for
wptreport.json doesn't work well with the reftest analyzer since it
expects an uncompressed log url. However the logs written to stdout
are automatically compressed and served with the appropriate headers,
so using tbpl logging there gives us both the screenshots and also
more efficient storage. The tradeoff is that these logs are more
verbose than the mach logs (and don't include e.g. a summary) so
there's more to wade through reading the logs.

### DIFF
--- a/tools/ci/taskcluster-run.py
+++ b/tools/ci/taskcluster-run.py
@@ -31,10 +31,12 @@ def find_wptreport(args):
     return parser.parse_known_args(args)[0].log_wptreport
 
 
-def gzip_file(filename):
+def gzip_file(filename, delete_original=True):
     with open(filename, 'rb') as f_in:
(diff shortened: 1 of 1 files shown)


Hexcles (approved):
LGTM

PR #13571, opened 2018-10-17
[cssom-view] Inheritance of CSSOM View properties

ewilligers:
https://drafts.csswg.org/cssom-view/#property-index


ewilligers:
@zcorpan  Please review

zcorpan (approved):

zcorpan:
Seems like we should be able to auto generate this sort of tests for all css properties? Cc @foolip

PR #13572, opened 2018-10-17
[css-rhythm] Inheritance, initial values

ewilligers:
https://drafts.csswg.org/css-rhythm/#property-index


ewilligers:
@kojiishi  please review

kojiishi (approved):
Can you split block and line to separate tests? We haven't heard anyone interested in block property yet, unfortunately.

ewilligers:
We can omit the properties that lack implementation, no problem.

PR #13575, opened 2018-10-17
[compositing] Inheritance, initial values

ewilligers:
https://drafts.fxtf.org/compositing-1/#property-index


ewilligers:
@plinss  Please review.

svgeesus (approved):
LGTM

PR #13578, opened 2018-10-17
Setup logging earlier in the wpt run startup.

jgraham:
Curerntly we wait until we're setting up wptrunner kwargs to create the logger.
But that's pretty silly; we should set up the logger first thing and then use it
consistently to log during the setup phase.

### DIFF
--- a/tools/wpt/run.py
+++ b/tools/wpt/run.py
@@ -197,7 +197,7 @@ Install Firefox or use --binary to set the binary path""")
                 logger.info("""Can't find certutil, certificates will not be checked.
 Consider installing certutil via your OS package manager or directly.""")
             else:
-                print("Using certutil %s" % certutil)
+                logger.info("Using certutil %s" % certutil)
 
             kwargs["certutil_binary"] = certutil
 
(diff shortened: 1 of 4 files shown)


tools/wpt/run.py
```diff
@@ -510,6 +514,7 @@ def setup_wptrunner(venv, prompt=True, install_browser=False, **kwargs):
 
 def run(venv, **kwargs):
     # Remove arguments that aren't passed to wptrunner
```
foolip: This comment applied to the following line, can you move it down?

foolip (approved):
Excellent, thank you!

foolip:
I see there's still "STDOUT: INFO:manifest:Manifest downloaded" in the output, so there's still something needed to fix that particular problem.

jgraham:
`STDOUT: INFO:manifest:Manifest downloaded` comes from the manifest code, which installs its own stdlib logger and adds a handler to it (not the root logger) to write to stdout. The `wptrunner` mozlog setup wraps stdout (so that the output to stdout can be pure mozlog json, which is used in mozilla-central to set the job status).

I'm pretty sure the solution is just for the manifest code to not do funky things with the logging.

sideshowbarker:
unittests failing in Travis with:

```
ERROR: unknown environment 'py27-flake8'
```

PR #13544, opened 2018-10-16
XHR/MIME Sniffing: adjust Content-Type manipulation and MIME type ser…

annevk:
…ialization

Tests for https://github.com/whatwg/xhr/pull/224 and https://github.com/whatwg/mimesniff/pull/86. They are combined in a single commit as otherwise there would be two rounds of changes to the XMLHttpRequest tests, while they are really supposed to be coupled.

PR #13580, opened 2018-10-17
Update data URL test case for empty MIME parameter

domenic:
This was missed in #12925.


annevk (approved):
Took me a while, but this is the error handling case. Nice find!

annevk:
It seems however that these need to be updated in #13544 as well. To add the space.

Issue #12766, opened 2018-08-30
Annotating tests for optional features


jugglinmike:
It's currently difficult to contribute tests for functionality that is specified as optional. (See gh-12732 for a recent example.)

In some cases, test authors can wrap their sub-tests tests in feature-detection logic so that they are only executed in browsers that advertise support. This results in "empty" tests for non-supporting browsers, and that's not ideal because it complicates the meaning of cross-browser comparisons. Worse, [this approach isn't always an option because support cannot be detected.](https://github.com/web-platform-tests/wpt/pull/10840)

Authors could simply write the test as if the feature was mandatory. Consumers generally are able to allow test failures, so they can tolerate tests that they do not intend to satisfy. That's a bit of a hassle, though, since consumers will need to triage the relevance of each new failure whenever they update WPT.  People outside of the browser projects don't necessarily have access to these lists, so they'll be on their own when it comes to interpreting failures.

A machine-readable annotation for these tests would allow implementers to explicitly opt-in on a case-by-case basis, and it would help others to interpret and compare test results.

[As @zcorpan points out](https://github.com/web-platform-tests/wpt/pull/12732#issuecomment-416956116), [the CSS test suite already supports this via the "may" flag](https://web-platform-tests.org/writing-tests/css-metadata.html). We might promote that to be used across all of WPT. However, it seems likely that consumers will implement some subset of optional behavior. If all optional tests were annotated generically, consumers would still have to pick through them on a case-by-case basis. They may be able to use the directory structure to automate this. Rules like "allow optional tests in ` html/semantics/embedded-content/media-elements` to fail" may be sufficient. It also might be too coarse. It might be better to support a rule like "allow optional tests for autoplay to fail." That would require additional metadata (specifying both that the feature is optional and a name for the optional feature), but I don't know enough about the prevalence of optional features to say if the trade-off is worth it.

foolip:
Another feature which has been frustrating testing-wise is [`<video preload>`](https://html.spec.whatwg.org/multipage/media.html#attr-media-preload). It is merely a hint, even for `preload=none`, see "Optionally, run the following substeps" in the [resource fetch algorithm](https://html.spec.whatwg.org/multipage/media.html#concept-media-load-resource).

@jugglinmike, I think you're right that merely marking tests as optional won't do much, each user of WPT would still need to look at each such test and decide whether they want to make it pass or not. However, I'm not sure how much optional tests could be grouped together to alleviate this problem.

[...]

PR #13586, opened 2018-10-17
Add a `.optional.` file name flag for optional behavior

foolip:
Fixes https://github.com/web-platform-tests/wpt/issues/12766.

### DIFF
--- a/docs/_writing-tests/file-names.md
+++ b/docs/_writing-tests/file-names.md
@@ -57,7 +57,7 @@ themselves precede any test type flag, but are otherwise unordered.
 `.optional`
  : Indicates that a test makes assertions about optional behavior in a
    specification, typically marked by the [RFC 2119] "MAY" or "OPTIONAL"
-   keywords. This flag should not be used for "SHOULD", such requirements
+   keywords. This flag should not be used for "SHOULD"; such requirements
    can be tested with regular tests, like "MUST".
 
 `.tentative`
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):

docs/_writing-tests/file-names.md
```diff
@@ -54,6 +54,11 @@ themselves precede any test type flag, but are otherwise unordered.
  : (js files only) Indicates that the file generates tests in which it
     is [run in multiple scopes][multi-global-tests].
 
+`.optional`
+ : Indicates that a test makes assertions about optional behavior in a
+   specification, typically marked by the [RFC 2119] "MAY" or "OPTIONAL"
```
domenic: What about "should"?

foolip: I think in practice we're just adding tests for such requirements, or at least I've never seen a discussion about whether it's okay to add such tests.

domenic: Hmm, for some reason I thought that's what spawned this discussion. Maybe call that out explicitly here, to reduce confusion?

foolip: I've added a bit about should and must.

zcorpan (approved):

Ms2ger (approved):

docs/_writing-tests/file-names.md
gsnedders: That was my understanding of @jgraham's view (that adding such tests often has little benefit).

docs/_writing-tests/file-names.md
```diff
@@ -54,6 +54,12 @@ themselves precede any test type flag, but are otherwise unordered.
...
+   keywords. This flag should not be used for "SHOULD", such requirements
```
domenic: Nit: comma -> semicolon

domenic (approved):

PR #13600, opened 2018-10-18
Add tests for simple dialog newline normalization

domenic:
Follows https://github.com/whatwg/html/pull/4105.


html/webappapis/user-prompts/newline-normalization.manual.html
```diff
@@ -0,0 +1,34 @@
...
+<meta charset="utf-8">
+<title>Newline normalization in simple dialogs</title>
+<link rel="help" href="https://html.spec.whatwg.org/multipage/#concept-dialog-newline-normalization">
+
+<p>The dialogs should all contain text looking like:</p>
+
+<pre>First line
+Second line</pre>
+
+<p>or:</p>
```
zcorpan: Don't you want to require one or the other?

domenic: Yes, this is "inclusive or"?

domenic:
Travis lint says

```
INFO:lint:There was 1 error (CONTENT-VISUAL: 1)
INFO:lint:You must fix all errors; for details on how to fix them, see
INFO:lint:https://web-platform-tests.org/writing-tests/lint-tool.html
```

but that URL does not explain CONTENT-VISUAL.

html/webappapis/user-prompts/newline-normalization.manual.html
zcorpan: I mean for each test, say if 1 or 2 newlines are expected

domenic: I guess we could, but it's not possible for a test that was meant to output "line" to instead say "paragraph", so I'm not sure what value it would add.

annevk:
It's for `Visual test whose filename doesn't end in '-visual'`, but I'd expect `Manual test whose filename doesn't end in '-manual'` (you used a `.`) or some such.

cc @gsnedders @jgraham

gsnedders:
@domenic @annevk Basically, having a `<link rel=help>` makes a file eligible to be a test file when it otherwise wouldn't be, for the sake of compatibility with the old CSS tests. These explicitly mark manual tests with `<meta name=flags content=interact>` (and testharness.js/reftests are categorised by appropriate `script`/`link` elements), leaving everything else to be manual/visual.

It would probably make a decent amount of sense to limit these cases (both CONTENT-VISUAL and CONTENT-MANUAL) to `css/`.

html/webappapis/user-prompts/newline-normalization-manual.html
```diff
@@ -0,0 +1,20 @@
...
+<meta charset="utf-8">
+<title>Newline normalization in simple dialogs</title>
+<link rel="help" href="https://html.spec.whatwg.org/multipage/#simple-dialogs">
+
+<p>The dialogs should all contain text looking like:</p>
+
+<pre>Line 1.1
+Line 1.2
+Line 1.3
+
```
annevk: You need to add "Line 1.4" here followed by a newline.

annevk (approved):

PR #13616, opened 2018-10-19
[css-multicol] Adding assert statements to tests

rachelandrew:
As discussed in https://github.com/web-platform-tests/wpt/issues/7470 I have added assert lines to a bunch of these tests, identified while inlining the tests into the multicol spec. If I've misrepresented any test please let me know.


mstensho (approved):

css/css-multicol/multicol-span-all-002.xht
```diff
@@ -7,6 +7,7 @@
   <link rel="help" href="http://www.w3.org/TR/css3-multicol/#column-span" title="6.1. 'column-span'" />
   <link rel="match" href="multicol-span-all-002-ref.xht" />
   <meta name="flags" content="ahem" />
+  <meta name="assert" content="Test of the situation where there is not sufficient space for the spanning element, UA may treat the element as 'column-span: none'; in which case, 'overflow: hidden' will take care of overflowed content out of multi-column box."
```
Ms2ger:
```suggestion
  <meta name="assert" content="Test of the situation where there is not sufficient space for the spanning element, UA may treat the element as 'column-span: none'; in which case, 'overflow: hidden' will take care of overflowed content out of multi-column box." />
```

PR #13638, opened 2018-10-20
Add expectations for Safari in for infrastructure/ tests

foolip:
In preparation for running these tests in CI.

Verified locally with Safari Technology Preview 67:
```
export no_proxy='*'
./wpt run --metadata infrastructure/metadata/ --channel=preview safari infrastructure/
./wpt run --metadata infrastructure/metadata/ --channel=preview safari_webdriver infrastructure/
```

With Safari 12 there are still failures, but we'll not use that in CI.

### DIFF
--- a/infrastructure/metadata/infrastructure/assumptions/allowed-to-play.html.ini
+++ b/infrastructure/metadata/infrastructure/assumptions/allowed-to-play.html.ini
@@ -1,0 +1,8 @@
+[allowed-to-play.html]
+  expected:
+    if product == "safari" or product == "safari_webdriver": ERROR # https://bugs.webkit.org/show_bug.cgi?id=190775
+
+
+  [<audio> autoplay]
+    expected:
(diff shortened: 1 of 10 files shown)


foolip:
Note that https://bugs.webkit.org/show_bug.cgi?id=190775 is newly filed for this PR.

gsnedders (approved):

foolip:
I didn't check too closely, some service worker tests I think. If it's possible to use the browser version as part of the condition I could set it up, but is it? Then it's depend on my PR to get the version...

foolip:
No merge_pr_13638tag was created for this because https://travis-ci.org/web-platform-tests/wpt/jobs/444636306 failed, which is because the GitHub API doesn't realize that commit 40e92668df0d05414b58ad5fd58b6ad96e2e53aa is for this PR. Given the timing I assume it's because of https://blog.github.com/2018-10-21-october21-incident-report/ so I've manually pushed the tag.

PR #13639, opened 2018-10-20
Avoid the use of `readlink -f` in shell scripts

### LINKED ISSUE #7452: build-css-testsuites.sh fails on macOS
a77f076e54ef0f20838f64a7b16bfb2b13f72f87 from #6631 seems to have made it impossible to run css/build-css-testsuites.sh on macOS; it fails with:

```
pnin:css gsnedders$ ./build-css-testsuites.sh 
+++ readlink -f ./build-css-testsuites.sh
readlink: illegal option -- f
usage: readlink [-n] [file ...]
++ dirname
usage: dirname path
+ SCRIPT_DIR=
```

As far as I can tell, `readlink` is some non-standard GNU tool with some limited support elsewhere.

foolip:
This doesn't work on macOS. Replacement taken from:
https://serverfault.com/a/406371

Fixes #7452

### DIFF
--- a/css/build-css-testsuites.sh
+++ b/css/build-css-testsuites.sh
@@ -1,8 +1,7 @@
 #!/usr/bin/env sh
 set -ex
 
-SCRIPT_DIR=$(dirname $(readlink -f "$0"))
-WPT_ROOT=$(readlink -f $SCRIPT_DIR/..)
+WPT_ROOT=$(cd $(dirname "$0")/.. && pwd -P)
 cd $WPT_ROOT
 
 main() {
--- a/css/build-css-testsuites.sh
+++ b/css/build-css-testsuites.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/env sh
 set -ex
 
-WPT_ROOT=$(cd $(dirname "$0")/.. && pwd -P)
+SCRIPT_DIR=$(cd $(dirname "$0") && pwd -P)
+WPT_ROOT=$SCRIPT_DIR/..
 cd $WPT_ROOT
 
 main() {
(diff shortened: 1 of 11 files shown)


css/build-css-testsuites.sh
```diff
@@ -1,8 +1,7 @@
 #!/usr/bin/env sh
 set -ex
 
-SCRIPT_DIR=$(dirname $(readlink -f "$0"))
-WPT_ROOT=$(readlink -f $SCRIPT_DIR/..)
+WPT_ROOT=$(cd $(dirname "$0")/.. && pwd -P)
```
gsnedders: In general, this is wrong, though I expect we can assume the `css` directory isn't a symlink.

Really what we want is `SCRIPT_DIR=$(cd $(dirname "$0") && pwd -P)` and then keep WPT_ROOT as is.

foolip: Can you give an example of when the result would be different, confirned on command line?

foolip: Note that leaving `WPT_ROOT` unchanged would mean still using `readlink -f`, which would make this change pointless :)

foolip: Alright, so I tested and if `x` is a symlink to `y`, then `cd x/..` will end up in the parent directory of the symlink `x` rather than `y`'s parent directory. If `css` for some odd reason were a symlink in this repo that would actually be the right thing, but if someone has a symlink outside of the checkout pointing to the `css` directory they can't invoke the script via that symlink. So, uh, I guess I'll first establish a symlink-free `SCRIPT_DIR` and then use that...

gsnedders: I think having `WPT_ROOT=$SCRIPT_DIR/..` should be fine?

gsnedders (approved):
@jgraham, can you check this too?

css/build-css-testsuites.sh
foolip: That's what I went with

jgraham (approved):

PR #13647, opened 2018-10-22
[css-scrollbars] Inheritance, initial values

ewilligers:
https://drafts.csswg.org/css-scrollbars/#property-index

### DIFF
--- a/css/css-scrollbars/inheritance.html
+++ b/css/css-scrollbars/inheritance.html
@@ -1,0 +1,22 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Inheritance of CSS Scrollbars properties</title>
+<link rel="help" href="https://drafts.csswg.org/css-scrollbars/#property-index">
+<meta name="assert" content="Properties inherit according to the spec.">
+<meta name="assert" content="Properties have initial values according to the spec.">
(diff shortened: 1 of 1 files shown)


ewilligers:
The inheritance tests were failing in Firefox when
scrollbar-color was light and scrollbar-width was 10px.
    
Changed the values used for inheritance testing, to allow the
inheritance tests to pass with the current Firefox implementation.
    
(The scrollbar-color initial value test still fails in Firefox.)

upsuper:
Filed w3c/csswg-drafts#3237 for that.

upsuper:
The latest nightly should pass all tests here now.

ewilligers:
The spec is clear that `auto` is the expected computed value for `scrollbar-color` initial value:

> Initial: auto

>  Computed value: keyword or two computed colors

> If this property computes to value other than auto

@tantek Please review.

upsuper:
But the result of gCS is resolved value, which is not necessarily computed value. Anyway the group has resolved on have keywords keep being keywords in resolved value.

ewilligers:
@upsuper  Please review.  I  have updated the commit message so it no longer mentions Firefox.

upsuper (approved):
LGTM. `auto` should work now given the resolution a while back.

PR #13649, opened 2018-10-22
W3C 3-clause BSD License

wseltzer:
Change from dual licensing to the W3C 3-clause BSD.

Ref https://github.com/web-platform-tests/wpt/issues/11009


foolip:
In https://github.com/web-platform-tests/rfcs/pull/18 we resolved to go with https://github.com/web-platform-tests/wpt/pull/13650 instead.

PR #11191, opened 2018-05-28
Use the The 3-Clause BSD License from opensource.org

annevk:
Fixes #11009.

PR #13650, opened 2018-10-22
Change to 3 clause BSD

wseltzer:
Derived from https://opensource.org/licenses/BSD-3-Clause
changing Copyright <YEAR> <COPYRIGHT HOLDER> to Copyright 2018 web-platform-tests contributors

Ref https://github.com/web-platform-tests/wpt/issues/11009

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,33 +1,13 @@
-# Dual-License for W3C Test Suites
+# 3-Clause BSD License
 
-All documents in this Repository are licensed by contributors to be distributed under both the [W3C Test Suite License](#w3c-test-suite-license) and the [W3C 3-clause BSD License](#w3c-3-clause-bsd-license), reproduced below. The choice of license is up to the licensee. For more information, see [Licenses for W3C Test Suites](https://www.w3.org/Consortium/Legal/2008/04-testsuite-copyright.html)
-
-# W3C Test Suite License
-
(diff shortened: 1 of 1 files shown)


LICENSE.md
```diff
@@ -1,33 +1,13 @@
-# Dual-License for W3C Test Suites
+# 3-Clause BSD License
```
annevk: This PR seems nearly equivalent to #11191 except that this doesn't say "The" at the start and the "Redistribution ..." line appears modified in the diff, whereas it doesn't seem to be.

foolip: I've compared it again and the only difference is indeed "The".

I also compared to the license file that GitHub gives you if you create a new repo with a license, and the only non-whitespace difference was that they call it "BSD 3-Clause License".

I've pushed @jugglinmike's changes to this PR and restored the documentation link and intend to merge this PR once checks have passed.

foolip (approved):

foolip:
Sorry for the noise on this PR, I pushed some changes from https://github.com/web-platform-tests/wpt/pull/11191 with the intention of merging this, but https://github.com/web-platform-tests/rfcs/pull/18 references https://github.com/web-platform-tests/wpt/pull/11191 so I've undone that. This PR is pointing to the same commit as when it was created.

foolip:
This has effectively been merged via https://github.com/web-platform-tests/wpt/pull/11191, which was almost identical apart from a "The" and whitespace, which is now the cause of conflicts. I'm going to close this, but thank you @wseltzer for creating it to make it clear that you were OK with this change.

PR #13651, opened 2018-10-22
[css-multicol] Initial values and inheritance

ewilligers:
Test that CSS Multi-column Layout properties have the initial values
given in the spec. None of the properties inherit.

https://drafts.csswg.org/css-multicol/#property-index

Identified bugs:
- https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/19367855/
- https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/19367975/
- https://bugs.webkit.org/show_bug.cgi?id=190702

### DIFF
--- a/css/css-multicol/inheritance.html
+++ b/css/css-multicol/inheritance.html
@@ -1,0 +1,45 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Inheritance of CSS Multi-column Layout properties</title>
+<link rel="help" href="https://drafts.csswg.org/css-multicol/#property-index">
+<meta name="assert" content="Properties not not inherit.">
+<meta name="assert" content="Properties have initial values according to the spec.">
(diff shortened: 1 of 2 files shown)


css/css-multicol/inheritance.html
```diff
@@ -0,0 +1,45 @@
...
+  }
+  #target {
+    column-rule-style: dotted;
+  }
+</style>
+<script>
+const mediumWidth = getComputedStyle(reference).borderTopWidth; // e.g. 3px
+reference.style.display = 'none';
+
+assert_not_inherited('column-count', 'auto', '2');
```
mstensho: These are generally hard to read. Probably because I'm not used to these functions. How about documenting the functions in /css/support/inheritance-testcommon.js ?

mstensho (approved):

css/css-multicol/inheritance.html
ewilligers: Comments added.

PR #13660, opened 2018-10-22
Remove reference loop from test clip-path-content-syling.svg.

dholbert:
This test currently has a reference loop (`clip1` is currently specified as being its own `clip-path`).

This is bogus and seems to be accidental -- `clip1` is also used lower down for the actual user-visible business of the test (and nothing about the test's naming or metadata/prose seem to suggest that it intended to exercise clip-path reference loops).

Let's remove the bogus reference loop.

### DIFF
--- a/css/css-masking/clip-path-svg-content/clip-path-content-syling.svg
+++ b/css/css-masking/clip-path-svg-content/clip-path-content-syling.svg
@@ -9,7 +9,7 @@
 	<desc class="assert">Style properties on content elements of clipPath
 	must be ignored. A green square should be visible.</desc>
 </g>
-<clipPath id="clip1" clip-path="url(#clip1)">
+<clipPath id="clip1">
 	<rect x="50" y="50" width="100" height="100" stroke="black" stroke-wdith="10" stroke-dasharray="10 10" fill="none"/>
 </clipPath>
 <rect width="200" height="200" fill="green" clip-path="url(#clip1)"/>
--- a/css/css-masking/clip-path-svg-content/clip-path-content-syling.svg
+++ b/css/css-masking/clip-path-svg-content/clip-path-content-syling.svg
@@ -10,7 +10,7 @@
 	must be ignored. A green square should be visible.</desc>
 </g>
 <clipPath id="clip1">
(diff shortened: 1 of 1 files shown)


css/css-masking/clip-path-svg-content/clip-path-content-syling.svg
```diff
@@ -9,8 +9,8 @@
 	<desc class="assert">Style properties on content elements of clipPath
 	must be ignored. A green square should be visible.</desc>
 </g>
-<clipPath id="clip1" clip-path="url(#clip1)">
+<clipPath id="clip1">
 	<rect x="50" y="50" width="100" height="100" stroke="black" stroke-wdith="10" stroke-dasharray="10 10" fill="none"/>
```
fsoder: Maybe fix the typo'd stroke-width (which could be relevant to this test based on the description.)

```suggestion
	<rect x="50" y="50" width="100" height="100" stroke="black" stroke-width="10" stroke-dasharray="10 10" fill="none"/>
```

fsoder (approved):
Looks reasonable to me, FWIW.

dholbert:
(This edit lets the test pass in Firefox, BTW -- otherwise it fails, because Firefox currently seems to deactivate clipPath elements that participate in reference loops. I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1501023 on that.)

css/css-masking/clip-path-svg-content/clip-path-content-syling.svg
dholbert: Good point! might as well fix that while we're here.

birtles (approved):

dholbert:
Hmm, looks like this got 2 approvals, but it's still pending some "required statuses"... Anyone know what that means?

(I don't post a lot of pull requests so it might be something obvious I'm unaware of - or I'm wondering if it's a symptom of the fact that there was a (accepted) suggested-edit here, since suggested edits are a brand-new "beta" github feature which maybe might not play well with everything yet [dunno]...

longsonr:
Would be nice if we fixed the typo in the file name. Presumably styling is meant rather than syling.

PR #13667, opened 2018-10-22
Fix #13656: make percentage-heights-003 test definiteness

### LINKED ISSUE #13656: external/wpt/css/css-flexbox/percentage-heights-003.html does not test the right thing
https://github.com/web-platform-tests/wpt/blob/master/css/css-flexbox/percentage-heights-003.html

@gregwhitworth fyi

This test has a column flexbox with an indefinite size, but expects that the flex item size is considered definite. I am fairly sure this is incorrect, per https://drafts.csswg.org/css-flexbox/#definite-sizes item 2 the flexbox needs a definite main size.

gsnedders:
Fix #13656. This was meant to be a test for https://github.com/w3c/csswg-drafts/issues/1679, so test both definite and indefinite flex container cases.

It seems nothing actually now passes this: Blink/WebKit always treat it as indefinite, Gecko always definite. Or I've misunderstood the spec.

### DIFF
--- a/css/css-flexbox/percentage-heights-003.html
+++ b/css/css-flexbox/percentage-heights-003.html
@@ -22,13 +22,13 @@
 }
 
 .flexbox > div {
-    background: red;
+    background: teal;
     flex-grow: 1;
 }
 
-span {
+.flexbox span {
     height: 100%;
-    background: green;
+    background: orange;
     display: block;
 }
 </style>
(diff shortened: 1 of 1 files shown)


cbiesinger (approved):

cbiesinger:
I finally realized what's been bothering me about this -- should there also be tests making sure that flex-basis: auto is not considered definite?

cbiesinger:
FYI the reason Blink is failing is due to https://bugs.chromium.org/p/chromium/issues/detail?id=784059. Hope to reland a fix soon.

PR #13671, opened 2018-10-23
Cross origin workers should fail to fetch

domfarolino:
/cc @domenic @bzbarsky @nhiroki 

Tackle low-hanging fruit in https://github.com/web-platform-tests/wpt/issues/13426.

Prior to this PR, testing that cross-origin workers fail was pretty [grim](https://wpt.fyi/results/workers/Worker_cross_origin_security_err.htm?aligned&label=stable), in that the test basically broke for everyone but Firefox, who seems to be the only conforming implementation here.

It is true that cross-origin workers fail to fetch in all or most implementations, but it seems that every failure is some synchronous security error, as opposed to letting the Fetch naturally result in a failure, and going through the normal means of firing ["an event named error"](https://html.spec.whatwg.org/multipage/workers.html#worker-processing-model:event-error).

This PR fixes the test for non-Firefox-like impls, and extends it to include module workers. I guess I should follow-up with vendors though, to see if they'd be willing to remove their immediate-erroring behavior in favor of letting Fetch naturally fail the request. Is it possible that vendors see cross-origin workers as a big enough security concern such that it's too big a risk to Fetch the doomed request, lest it somehow slip through the cracks and got sent? I just think its odd that a lot of impls are immediately failing cross-origin worker construction, when the spec does never mentions this. Here's where we do it [in Chrome](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/workers/abstract_worker.cc?sq=package:chromium&g=0&l=52).

### DIFF
--- a/workers/Worker_cross_origin_security_err.htm
+++ b/workers/Worker_cross_origin_security_err.htm
@@ -11,7 +11,7 @@ async_test(function(t) {
       assert_true(e instanceof Event);
     });
   } catch (e) {
-    assert_throws("SecurityError", () => {throw e}, "DOMExceptions thrown on cross-origin worker construction must be SecurityErrors");
+    assert_throws("SecurityError", () => {throw e}, "DOMExceptions thrown on cross-origin Worker construction must be SecurityErrors");
     t.done();
   }
 }, "Cross-origin classic workers should fail to fetch");
@@ -23,7 +23,7 @@ async_test(function(t) {
       assert_true(e instanceof Event);
     });
   } catch (e) {
-    assert_throws("SecurityError", () => {throw e}, "DOMExceptions thrown on cross-origin module worker construction must be SecurityErrors");
+    assert_throws("SecurityError", () => {throw e}, "DOMExceptions thrown on cross-origin module Worker construction must be SecurityErrors");
     t.done();
   }
 }, "Cross-origin module workers should fail to fetch");
--- a/workers/Worker_cross_origin_security_err.htm
+++ b/workers/Worker_cross_origin_security_err.htm
@@ -11,6 +11,7 @@ async_test(function(t) {
       assert_true(e instanceof Event);
     });
   } catch (e) {
+    assert_throws("SecurityError", () => {throw e}, "DOMExceptions thrown on cross-origin worker construction must be SecurityErrors");
     t.done();
   }
 }, "Cross-origin classic workers should fail to fetch");
@@ -22,6 +23,7 @@ async_test(function(t) {
       assert_true(e instanceof Event);
     });
   } catch (e) {
+    assert_throws("SecurityError", () => {throw e}, "DOMExceptions thrown on cross-origin module worker construction must be SecurityErrors");
     t.done();
   }
 }, "Cross-origin module workers should fail to fetch");
--- a/workers/Worker_cross_origin_security_err.htm
+++ b/workers/Worker_cross_origin_security_err.htm
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<title> Worker cross-origin URL </title>
+<title>Worker cross-origin URL</title>
 <script src="/resources/testharness.js"></script>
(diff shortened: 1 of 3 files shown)


domfarolino:
Clearly the original author of this test anticipated cross-origin Worker construction to fail in some implementations, hence the `try {} catch() {}`. The reason this test was timing out is because the `step_func_done` in `catch` was never being invoked. I've changed this to an `assert_unreached`, because though this means the cross-origin worker will not be fetched, it does not error out in the way the spec mandates.

bzbarsky (approved):
Looks good, but it would be good to fix the two same-origin.html tests too.

bzbarsky:
So in terms of the  spec...  https://html.spec.whatwg.org/multipage/workers.html#dom-worker step 1 says:

> The user agent may throw a "SecurityError" DOMException if the request violates a policy decision (e.g. if the user agent is configured to not allow the page to start dedicated workers).

which seems like the clause the other UAs are using; "policy decision" is pretty broad.

In theory, if implementing the spec word for word one would have to do a bunch of work including setting up the worker global before discovering that the load failed and erroring.  In practice, of course, one can just queue a task to fire error instead of throwing and not do anything else.

workers/Worker_cross_origin_security_err.htm
```diff
@@ -11,7 +11,18 @@
       assert_true(e instanceof Event);
     });
   } catch (e) {
```
Ms2ger: You can remove the try/catch entirely; `async_test` will handle any exceptions better than you can.

domfarolino: Thanks! But it looks like we're actually going to need to keep the try/catch after all, since the test can still pass in the catch

domfarolino:
Good catch @bzbarsky. I've updated the tests to still pass in the case where cross-origin workers fail to construct, given the spec text you found. I've also fixed the {Worker, SharedWorker}/same-origin.html tests. PTAL

bzbarsky (changes requested):

workers/Worker_cross_origin_security_err.htm
```diff
@@ -11,7 +11,18 @@
...
-    t.step_func_done(function(e) { assert_true(true); });
+    t.done();
```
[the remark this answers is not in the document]
domfarolino: Could also example `e.code` and assert that it is `18` (Security Error), but yeah, your assert_throws is nicer, I did not know that `assert_throws` could take in the error's `name` as a param

workers/constructors/SharedWorker/same-origin.html
```diff
@@ -19,7 +19,7 @@
       assert_true(e instanceof Event);
     });
   } catch (e) {
-    t.step_func_done(function(e) { assert_true(true); });
+    t.done();
```
bzbarsky: You want the assert_throws here too, and in the third test.

bzbarsky (changes requested):

workers/constructors/SharedWorker/same-origin.html
domfarolino: Sigh...of course, whoops

workers/constructors/Worker/same-origin.html
```diff
@@ -10,14 +10,15 @@
 // not propogate to the window before the tests finish
 setup({allow_uncaught_exception: true});
 
-function testSharedWorkerHelper(t, script) {
+function testWorkerHelper(t, script) {
   try {
-    var worker = new SharedWorker(script, '');
+    var worker = new Worker(script);
```
domfarolino: I also noticed that the Worker constructor test was still testing the SharedWorker constructor, probably as a result of copy paste. This meant all of these tests failed in Safari, a UA that doesn't support SharedWorkers.

bzbarsky: Yikes.   Good catch!

bzbarsky (approved):

PR #13676, opened 2018-10-23
Update docker image to Ubuntu 18.04

jgraham:
This is the latest LTS release, so it makes sense to test on it. Also it is a requirement
to get servo builds running in the docker image since 16.04 has a too-old gstreamer.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -61,7 +61,7 @@ tasks:
               owner: ${event.pusher.email}
               source: ${event.repository.url}
             payload:
-              image: jugglinmike/web-platform-tests:0.21
+              image: harjgam/web-platform-tests:0.22
               maxRunTime: 7200
               artifacts:
                 public/results:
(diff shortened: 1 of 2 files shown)


gsnedders:
Is there any way to kick this into doing a full run on TC?

Ms2ger (approved):

gsnedders:
https://wpt.fyi/results/?product=firefox%407a90fe4444&product=firefox%40fbc56f388a&diff is FF.

The css/CSS2/selectors changes concern me.

Hexcles:
FWIW, the result was just being processed (it took a bit longer than usual because of the backlog caused by the GitHub incident yesterday).

And the selectors changes indeed seem deterministic.

jgraham:
The selectors changes seem to be a bogus fail in the previous setup that is now correctly passing in the new setup (based on examining the screenshots).

PR #13677, opened 2018-10-23
[css-overscroll-behavior] Inheritance, initial

ewilligers:
https://drafts.csswg.org/css-overscroll-behavior/#property-index

### DIFF
--- a/css/css-overscroll-behavior/META.yml
+++ b/css/css-overscroll-behavior/META.yml
@@ -1,0 +1,3 @@
+spec: https://drafts.csswg.org/css-overscroll-behavior/
+suggested_reviewers:
+  - majido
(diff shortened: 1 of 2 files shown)


majido (approved):
lgtm % nit

nit: Do we need to also have an assertion for the longhand syntax e.g.,
`assert_not_inherited('overscroll-behavior-y', 'auto', 'contain');`

css/css-overscroll-behavior/META.yml
```diff
@@ -0,0 +1,6 @@
+spec: https://drafts.csswg.org/css-overscroll-behavior/
+suggested_reviewers:
+  - bgirard
```
majido: Not sure if @bgirard has signed up to do reviews. May want to confirm with them first.

ewilligers:
@frivoal @plinss  Are you happy to be reviewers?

PR #13684, opened 2018-10-23
service workers: Add long timeout to matchall test

mfalken:
clients-matchall-order.https.html is a slow test because it does a lot of navigations.


jdm (approved):

wanderview (approved):
LGTM

Issue #5504, opened 2017-04-07
Check for wide Python build in build-css-testsuites.sh


wpt-issue-mover:
Originally posted as https://github.com/w3c/csswg-test/issues/1245 by @gsnedders on 22 Mar 2017, 17:48 UTC:

> At least according to <https://wiki.csswg.org/test/css2.1/harness> the build system needs this.

foolip:
Since csswg-test is now merged, I assume this is no longer relevant.

gsnedders:
Nah, we still have the build system here, and we probably want this for the sake of people who care about the CSS test harness.

gsnedders:
(It's also a few lines if someone sits down and fixes it!)

PR #13696, opened 2018-10-24
Add UCS-4 support check for Python in build-css-testsuites.sh

qiuzhong:
CSS test suite build system requires Python UCS-4 support.
Add this check in css/build-css-testsuites.sh.
Fix: #5504

### DIFF
--- a/css/build-css-testsuites.sh
+++ b/css/build-css-testsuites.sh
@@ -35,6 +35,11 @@ main() {
             exit 1
         fi
 
+        if [ `$PYTHON -c 'import sys; print(sys.maxunicode)'` != "1114111" ]; then
+            echo "UCS-4 support for Python is required"
+            exit 1
+        fi
+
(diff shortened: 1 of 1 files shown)


css/build-css-testsuites.sh
```diff
@@ -35,6 +35,11 @@ main() {
             exit 1
         fi
 
+        if [ `$PYTHON -c 'import sys; print(sys.maxunicode)'` != "1114111" ]; then
```
foolip:
```suggestion
        # The maximum Unicode code point is U+10FFFF = 1114111
        if [ `$PYTHON -c 'import sys; print(sys.maxunicode)'` != "1114111" ]; then
```

foolip (approved):
Suggest a comment to add, otherwise looks good to me. I notice the Python on my Mac doesn't support UCS-4.

foolip (approved):

foolip:
@gsnedders can you sign off too?

gsnedders (approved):

PR #13701, opened 2018-10-24
Add a wpt create command

jgraham:
wpt create makes a file with the required elements for a test, and
optionally opens it in an editor.

### DIFF
--- a/tools/wpt/create.py
+++ b/tools/wpt/create.py
@@ -1,0 +1,131 @@
+import subprocess
+import os
+
+here = os.path.dirname(__file__)
+
+template_prefix = """<!doctype html>
+%(documentElement)s<meta charset=utf-8>
+"""
+template_long_timeout = "<meta name=timeout content=long>\n"
+
+template_body_th = """<title></title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<script>
+
+</script>
+"""
+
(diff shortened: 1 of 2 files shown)


tools/wpt/create.py
```diff
@@ -0,0 +1,143 @@
...
+    if path is None:
+        return
+
+    abs_path = os.path.normpath(os.path.abspath(path))
+    return os.path.relpath(abs_path, tests_root)
+
+
+def rel_url(rel_path):
+    upstream_path = os.path.join("testing", "web-platform", "tests")
+    local_path = os.path.join("testing", "web-platform", "mozilla", "tests")
```
Ms2ger: These two don't really belong upstream, do they?

tools/wpt/create.py
```diff
@@ -0,0 +1,143 @@
...
+    ref_path = rel_path(kwargs["ref"], kwargs["tests_root"])
+
+    if kwargs["ref"]:
+        kwargs["reftest"] = True
+
+    if ".." in path:
+        print("""Test path %s is not under wpt root.""" % path)
+        return 1
+
+    if ref_path and ".." in ref_path is None:
```
Ms2ger: `is None`?

Ms2ger (approved):
Can we unit test this?

sideshowbarker (approved):

PR #13706, opened 2018-10-24
Remove alexshalamov from the suggested reviewers lists

rakuco:
Alex is no longer working on those things, so drop him from those lists to
avoid confusion.


Honry (approved):
lgtm

PR #13738, opened 2018-10-26
Add tests for element/attribute name selectors in HTML

gsnedders:
Inspired by https://github.com/whatwg/html/issues/3372. This tests the current spec.

### DIFF
--- a/html/semantics/selectors/case-sensitivity-of-selectors/attr-name-selectors.html
+++ b/html/semantics/selectors/case-sensitivity-of-selectors/attr-name-selectors.html
@@ -1,0 +1,126 @@
+<!doctype html>
+<meta charset=utf8>
+<title>CSS attribute name selector case-sensitivity</title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<body>
+<script>
+const xhtmlns = "http://www.w3.org/1999/xhtml";
+const svgns = "http://www.w3.org/2000/svg";
+const fakens = "http://example.com/i.made.this.up";
+
+const xhtml_doc = document.implementation.createDocument(xhtmlns, "html", null);
+const svg_doc = document.implementation.createDocument(svgns, "svg", null);
+
+function pretty_print_name(ns, ln) {
+  if (ns === xhtmlns)
+    return "xhtml:" + ln;
+  if (ns === svgns)
+    return "svg:" + ln;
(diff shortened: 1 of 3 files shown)


gsnedders:
Oh actually because they're only lowercased for attributes in the null namespace the attribute case isn't interesting here, duh.

html/semantics/selectors/case-sensitivity-of-selectors/selector-namespace-prefix.html
```diff
@@ -0,0 +1,24 @@
+<!doctype html>
+<title>Test whether namespace prefixes are lowercased</title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
```
lilles: Doesn't this test really belong in css/css-namespaces[1], and isn't this already covered by css/css-namespaces/prefix-001.xml ?

[1] https://drafts.csswg.org/css-namespaces-3/#prefixes

html/semantics/selectors/case-sensitivity-of-selectors/element-selectors.html
```diff
@@ -0,0 +1,116 @@
+<!doctype html>
+<meta charset=utf8>
+<title>CSS element selector case-sensitivity</title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
```
lilles: Could you add a `<link rel=help>` with the relevant spec-link?

html/semantics/selectors/case-sensitivity-of-selectors/attr-name-selectors.html
```diff
@@ -0,0 +1,126 @@
+<!doctype html>
+<meta charset=utf8>
+<title>CSS attribute name selector case-sensitivity</title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
```
lilles: Could you add a `<link rel=help>` with the relevant spec-link?

html/semantics/selectors/case-sensitivity-of-selectors/element-selectors.html
gsnedders: if you don't know, for HTML you can always take the last path segment as a fragment (i.e., #case-sensitivity-of-selectors).

PR #13741, opened 2018-10-26
[testdriver.js] Relax test for Actions

jugglinmike:
The sequence of actions scheduled within the same tick is undefined.
Relax the test concerning this behavior to reflect the non-determinism.

### DIFF
--- a/infrastructure/testdriver/actions/eventOrder.html
+++ b/infrastructure/testdriver/actions/eventOrder.html
@@ -11,13 +11,17 @@
 <button id="b">Button b</button>
 
 <script>
-// Pointer 1 is added before Pointer 2 so it comes first in the list of sources
-// Therefore its actions happen first
-let events = [];
+// Pointer 1 is added before Pointer 2 so it comes first in the list of
+// sources. The execution order of actions within a tick is undefined, so this
+// detail does not inform the ordering of parallel events.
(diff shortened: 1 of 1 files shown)


jugglinmike:
testdriver.js's "actions" feature is intended to match [the corresponding behavior in the W3C WebDriver
specification](https://w3c.github.io/webdriver/#actions). Today, that text is slightly incorrect. I've proposed a fix via https://github.com/w3c/webdriver/pull/1323, and this patch aligns with the new text. We may want to wait to merge this until that pull request is accepted.

jgraham:
Eh, I'm concerned by this. Inside an event loop everything has defined ordering and it is fully intended that the actions are scheduled in the order resulting from a transpose of the input array i.e. so that the actions from the first defined device are placed on the event loop first).

jgraham (changes requested):
The ordering of actions within a tick is intended to be well-defined and should remain so. The one consession that the webdriver spec makes is that `pointerMove` of non-zero duration may wait before starting to move, and after the move starts the generated events need not be deterministically ordered, but there's no defined duration in this case, so the order is again well-defined.

jugglinmike:
We addressed this via a change to the spec: https://github.com/w3c/webdriver/pull/1323

PR #13742, opened 2018-10-26
Add some missing window properties

Zirro:
This test is meant to include all properties specified on `window`, but some from the [WindowOrWorkerGlobalScope mixin](https://html.spec.whatwg.org/multipage/webappapis.html#windoworworkerglobalscope) and [CSSOM View](https://drafts.csswg.org/cssom-view/#extensions-to-the-window-interface) are missing.

The added tests all pass in Chrome Canary and Safari TP, while Firefox Nightly fails one due to a missing `queueMicrotask()`-implementation.

### DIFF
--- a/html/browsers/the-window-object/window-properties.https.html
+++ b/html/browsers/the-window-object/window-properties.https.html
@@ -54,6 +54,7 @@ var replaceableAttributes = [
   "parent",
   "external",
   "length",
+  "origin",
 
   // CSSOM-View
   "screen",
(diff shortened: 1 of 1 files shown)


zcorpan (approved):

zcorpan:
Is there a bug about queueMicrotask?

Zirro:
Yes, https://bugzilla.mozilla.org/show_bug.cgi?id=1480236.

PR #13748, opened 2018-10-28
[css-ui] expect resolved color in inheritance test

ewilligers:
getComputedStyle returns the resolved color or 'invert', not 'auto'
https://drafts.csswg.org/cssom/#resolved-value

### DIFF
--- a/css/css-ui/inheritance.html
+++ b/css/css-ui/inheritance.html
@@ -32,7 +32,6 @@
 <script>
 const mediumWidth = getComputedStyle(reference).borderTopWidth; // e.g. 3px
 const currentColor = getComputedStyle(reference).color;
-const invertBackgroundColor = 'rgb(255, 255, 0)';
 
 assert_not_inherited('appearance', 'auto', 'none');
 assert_inherited('caret-color', currentColor, 'rgba(42, 53, 64, 0.75)');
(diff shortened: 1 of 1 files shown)


frivoal:
For the invert color of the outline-color property, you should check if the invert keyword is supported at all. If it is, it should be the initial value, but if it is not, having currentColor be the initial value is spec compliant.

frivoal (dismissed):
Looks good, except for the comment I made in https://github.com/web-platform-tests/wpt/pull/13748#issuecomment-435116605.

Once that's fixed, this should be merged.

ewilligers:
outline-color test now checks if 'invert' is supported. 

Blink/Firefox/Safari pass the outline-color test by not supporting 'invert', and returning the current rgb() color.

Edge fails the outline-color test by accepting 'invert' but returning 'invert' from getComputedStyle() instead of the inverted background color.

frivoal:
> Edge fails the outline-color test by [...] returning 'invert' from getComputedStyle() instead of the inverted background color.

That's actually a pass. returning the inverted background color isn't actually possible, since background may not be a single color. returning 'invert' as a keyword is the right thing to do.

frivoal:
> Does that mean the UA should check whether the background is a single, opaque solid color (and if so, return that) or should it always return invert?

It should always return `invert`. Happy to go clarify that in level 4, but that's the intention behind level 3, and is what microsoft & presto implemented.

svgeesus:
@ewilligers waiting for your changes to address comments before we merge.

svgeesus (commented):
Look good except for the point already raised by Florian

ewilligers:
Test updated, thank you for explaining.

ewilligers:
@frivoal's comments have been addressed.

ewilligers:
Ping. I believe this is ready for merging.

frivoal (approved):

PR #13749, opened 2018-10-28
[svg] circle radius 0 disables rendering

ewilligers:
A value of 0 for circle r, ellipse rx/ry, rect width/height
disables rendering.

https://www.w3.org/TR/SVG2/shapes.html#CircleElement
https://www.w3.org/TR/SVG2/shapes.html#EllipseElement
https://www.w3.org/TR/SVG2/shapes.html#RectElement


ewilligers:
Accompanies spec PR https://github.com/w3c/svgwg/pull/568

dirkschulze (approved):

PR #13750, opened 2018-10-28
[filter-effects] filter arguments optional

ewilligers:
grayscale() etc may be used without arguments.

https://github.com/w3c/fxtf-drafts/issues/1


ewilligers:
I was looking at an old version of the spec when I initially wrote the parsing test.

dirkschulze (approved):

Issue #13299, opened 2018-10-01
Run affected tests in Edge and Safari for PRs


foolip:
One upon a time we had Chrome, Edge, Firefox and Safari all running some tests for PRs. It was never robust enough to be truly useful, but we had it until we had to remove Edge and Safari in https://github.com/web-platform-tests/wpt/issues/9903.

We are now in the process (https://github.com/web-platform-tests/wpt/issues/13194) of moving Chrome and Firefox to Taskcluster and we'll use the results in https://github.com/web-platform-tests/wpt/issues/7475.

That still leaves Edge and Safari without any coverage on PRs.

For Edge, there doesn't appear to exist an CI integrated with GitHub where we could run Edge, much less Edge Insider Preview. @thejohnjansen for ideas.

For Safari, it's possible we could use Travis's macOS support, but nobody has tested this. It would also be a fairly different setup from what @jugglinmike has set up. @youennf for ideas.

The value of this would be preventing accidental regressions of tests in Edge and Safari, just like for Chrome and Firefox.

foolip:
https://github.com/travis-ci/travis-ci/issues/7187 as some discussion on Safari in Travis.

foolip:
In https://github.com/foolip/web-platform-tests/pull/3 I've experimented with macOS on Travis to the point where I got Safari TP installed, do nothing with it, and then run a lint that fails:
https://travis-ci.org/foolip/web-platform-tests/builds/435885997

(For Safari 12 stable, the Travis macOS image wasn't new enough, "[This update requires macOS version 10.13.6 or later](https://travis-ci.org/foolip/web-platform-tests/builds/435880537)", but Travis's latest is 10.13.3.)

Put together with some of @jugglinmike's amazing tricks for macOS in https://github.com/web-platform-tests/results-collection it would probably be possible to run affected tests on Safari TP on Travis.

Whether we run them 10 times for stability or just once for results, this will time out if too many tests are affected. We'd also have to find a way to get the results into wpt.fyi. So it's not a trivial undertaking, but at least the OS images exist and probably work for this.

foolip:
I've also experimented briefly with Cirrus CI. Their macOS support is ultimately backed by https://www.macstadium.com/, which I recall @jugglinmike has mentioned at some point.

Cirrus CI has macOS 10.13.5, and in https://github.com/foolip/web-platform-tests/pull/4 I got it as far as with Travis, namely I got STP 65 installed.

[...]

PR #13769, opened 2018-10-30
Add configuration for Azure Pipelines

foolip:
This just runs infrastructure/ tests on relevant PRs, something minimal
and useful before trying to enable Azure Pipelines for the first time.

Part of https://github.com/web-platform-tests/wpt/issues/13299.


foolip:
The result of exactly this configuration on a fork can be seen in https://github.com/foolip/wpt/pull/7/checks?check_run_id=27262041.

foolip:
https://dev.azure.com/web-platform-tests/wpt is the Azure DevOps org/project where I'll add a pipeline to point at this file once landed. There is some minimal configuration that is done in that UI, like whether PRs trigger builds, but nothing too interesting.

foolip:
https://github.com/foolip/wpt/pull/7/checks?check_run_id=27372593 / https://dev.azure.com/foolip/wpt/_build/results?buildId=118&view=logs is a run using this configuration, but with the `condition` lines deleted to make everything unconditional.

.azure-pipelines.yml
```diff
@@ -0,0 +1,60 @@
...
+    displayName: 'List test jobs'
+
+  - script: |
+      sudo easy_install pip
+      sudo pip install -U virtualenv
+    displayName: 'Install Python packages'
+    condition: variables.run_wptrunner_infrastructure
+
+  - script: |
+      sudo cp fonts/Ahem.ttf /Library/Fonts
```
jgraham: Is this actually needed? Shouldn't we pass `--install-fonts` to `wpt run` instead?

.azure-pipelines.yml
```diff
@@ -0,0 +1,60 @@
...
+    condition: variables.run_wptrunner_infrastructure
+
+  - script: |
+      # https://github.com/web-platform-tests/results-collection/blob/master/src/scripts/trust-root-ca.sh
+      sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain tools/certs/cacert.pem
+    displayName: 'Install web-platform.test certificate'
+    condition: variables.run_wptrunner_infrastructure
+
+  - script: |
+      # Pin to STP 67, as SafariDriver isn't working in 68.
```
jgraham: How will we remember to update this?

.azure-pipelines.yml
foolip: This was after that failed to work once, font installation seems to be racy or something. I guess I could debug it more, how bad do you think this is?

.azure-pipelines.yml
foolip: I'll need to file a bug and bother @burg about it.

.azure-pipelines.yml
jgraham: It's an issue we've had in buildbot and in gecko CI, so I would like to understand what's going on. If you tried it and it doesn't work out of the box, this is OK for now but we should certainly follow up with the investigations, because `--install-fonts` always works locally on a Mac.

foolip:
To make review easier if there are more changes, I've created https://github.com/foolip/wpt/pull/10 from the very same branch, so that PR will be in sync with this one, but you can see Azure Pipelines results there.

.azure-pipelines.yml
foolip: Nope, still broken: https://dev.azure.com/foolip/wpt/_build/results?buildId=122&view=logs

I'll file an issue and link it from this file.

foolip:
This PR now builds on https://github.com/web-platform-tests/wpt/pull/13801, to be able to verify the config with a minimal change.

jgraham (approved):

PR #13781, opened 2018-10-30
[webdriver]: add checks for execute_script tests recognising promises

christian-bromann:
I've been come across point 4, 5, 6 and 7 of the `[execute_script](https://w3c.github.io/webdriver/#execute-script)` recognising promises if returned as such and apparently Geckodriver does resolve/reject a promise properly but don't recognise script timeouts for it (see https://github.com/web-platform-tests/wpt/pull/13781#issuecomment-434319629). Chromedriver always returns with status code 200 and `value: {}`.

This is my first stab at this and would love to get initial feedback as it seems that drivers behave differently/inconsistent here which can be either a bug in the driver or in the protocol definition.

### DIFF
--- a/webdriver/tests/execute_script/promise.py
+++ b/webdriver/tests/execute_script/promise.py
@@ -1,5 +1,3 @@
-# META: timeout=long
-
 import pytest
 
 from tests.support.asserts import assert_dialog_handled, assert_error, assert_success
@@ -16,44 +14,67 @@ def execute_script(session, script, args=None):
         body)
 
 
-def set_timeouts(session, timeouts):
-    return session.transport.send(
-        "POST", "session/{session_id}/timeouts".format(**vars(session)),
-        timeouts)
-
-
 def test_promise_resolve(session):
     response = execute_script(session, """
-        return new Promise(
-            (resolve) => setTimeout(
-                () => resolve('foobar'),
-                1
-            )
-        );
+        return Promise.resolve('foobar');
         """)
     assert_success(response, "foobar")
 
 
-def test_promise_reject(session):
-    response = execute_script(session, """
-        return new Promise(
-            (resolve, reject) => setTimeout(
-                () => reject('some_error'),
-                1
-            )
-        );
-        """)
-    assert_error(response, "some_error")
-
-
-def test_promise_timeout(session):
-    response = set_timeouts(session, {"script": 100})
+def test_promise_resolve_delayed(session):
     response = execute_script(session, """
         return new Promise(
             (resolve) => setTimeout(
-                () => resolve('timeout_error'),
-                200
+                () => resolve('foobar'),
+                50
             )
         );
         """)
-    assert_error(response, "timeout_error")
+    assert_success(response, "foobar") 
+
+
+def test_promise_reject(session):
+    response = execute_script(session, """
+        return Promise.reject(new Error('my error'));
+        """)
+    assert_error(response, "unknown error")
+    assert response.body["value"]["message"] == "Error: my error"
+
+
+def test_promise_reject_delayed(session):
+    response = execute_script(session, """
+        return new Promise(
+            (resolve, reject) => setTimeout(
+                () => reject(new Error('my error')),
+                50
+            )
+        );
+        """)
+    assert_error(response, "unknown error")
+    assert response.body["value"]["message"] == "Error: my error" 
+
+
+def test_promise_resolve_timeout(session):
+    session.timeouts.script = .1
+    response = execute_script(session, """
+        return new Promise(
+            (resolve) => setTimeout(
+                () => resolve('success'),
+                1000
+            )
+        );
+        """)
+    assert_error(response, "timeout error")
+
+
+def test_promise_reject_timeout(session):
+    session.timeouts.script = .1
+    response = execute_script(session, """
+        return new Promise(
+            (resolve, reject) => setTimeout(
+                () => reject(new Error('my error')),
+                1000
+            )
+        );
+        """)
+    assert_error(response, "timeout error")
--- a/webdriver/tests/execute_script/promise.py
+++ b/webdriver/tests/execute_script/promise.py
@@ -1,0 +1,59 @@
+# META: timeout=long
+
+import pytest
+
+from tests.support.asserts import assert_dialog_handled, assert_error, assert_success
+
+
+def execute_script(session, script, args=None):
+    if args is None:
+        args = []
+    body = {"script": script, "args": args}
+
+    return session.transport.send(
+        "POST", "/session/{session_id}/execute/sync".format(
+            session_id=session.session_id),
+        body)
+
+
+def set_timeouts(session, timeouts):
+    return session.transport.send(
+        "POST", "session/{session_id}/timeouts".format(**vars(session)),
+        timeouts)
+
+
+def test_promise_resolve(session):
+    response = execute_script(session, """
+        return new Promise(
+            (resolve) => setTimeout(
+                () => resolve('foobar'),
+                1
+            )
+        );
+        """)
+    assert_success(response, "foobar")
+
+
+def test_promise_reject(session):
+    response = execute_script(session, """
+        return new Promise(
+            (resolve, reject) => setTimeout(
+                () => reject('some_error'),
+                1
+            )
+        );
+        """)
+    assert_error(response, "some_error")
+
+
+def test_promise_timeout(session):
+    response = set_timeouts(session, {"script": 100})
+    response = execute_script(session, """
+        return new Promise(
+            (resolve) => setTimeout(
+                () => resolve('timeout_error'),
+                200
+            )
+        );
+        """)
+    assert_error(response, "timeout_error")
--- a/webdriver/tests/execute_script/promise.py
+++ b/webdriver/tests/execute_script/promise.py
@@ -30,7 +30,7 @@ def test_promise_resolve_delayed(session):
             )
         );
         """)
-    assert_success(response, "foobar") 
+    assert_success(response, "foobar")
 
 
 def test_promise_all_resolve(session):
@@ -69,6 +69,7 @@ def test_promise_reject_delayed(session):
         """)
     assert_error(response, "javascript error")
 
+
 def test_promise_all_reject(session):
     response = execute_script(session, """
         return Promise.all([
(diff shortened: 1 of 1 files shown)


whimboo:
In case of the rejection of a promise a `Javascript error` has to be thrown. Given the [error table](https://w3c.github.io/webdriver/#dfn-javascript-error) it needs a HTTP response of 500. If geckodriver responds with 200, then this is clearly a bug in our implementation. @christian-bromann can you please raise a [geckodriver bug](https://bugzilla.mozilla.org/enter_bug.cgi?product=Testing&component=geckodriver)?

webdriver/tests/execute_script/promise.py
```diff
@@ -0,0 +1,59 @@
...
+    response = set_timeouts(session, {"script": 100})
+    response = execute_script(session, """
+        return new Promise(
+            (resolve) => setTimeout(
+                () => resolve('timeout_error'),
+                200
+            )
+        );
+        """)
+    assert_error(response, "timeout_error")
```
whimboo: This should be a `timeout` error and not `timeout_error`.

whimboo (dismissed):
Thank you @christian-bromann for having a look at those steps of `Execute Script`. Your contribution is welcome! I added a couple of inline comments you want to have a look at. Also I think it might be ok, to just add those tests into `execute.py` directly.

Once all issues have been clarified maybe you also want to add similar tests for `Execute Async Script`?

webdriver/tests/execute_script/promise.py
```diff
@@ -0,0 +1,59 @@
...
+def set_timeouts(session, timeouts):
+    return session.transport.send(
+        "POST", "session/{session_id}/timeouts".format(**vars(session)),
+        timeouts)
+
+
+def test_promise_resolve(session):
+    response = execute_script(session, """
+        return new Promise(
+            (resolve) => setTimeout(
```
andreastt: Why do we use a `setTimeout` for this test?

webdriver/tests/execute_script/promise.py
```diff
@@ -0,0 +1,59 @@
...
+        """)
+    assert_error(response, "some_error")
+
+
+def test_promise_timeout(session):
+    response = set_timeouts(session, {"script": 100})
+    response = execute_script(session, """
+        return new Promise(
+            (resolve) => setTimeout(
+                () => resolve('timeout_error'),
```
andreastt: Remove the argument value to resolve.

webdriver/tests/execute_script/promise.py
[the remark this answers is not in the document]
andreastt: I don’t think IE supports promises in the first place, so keeping the fat arrows here is probably fine.

andreastt (changes requested):
Thanks for writing these tests! 👍🏻

We have https://bugzilla.mozilla.org/show_bug.cgi?id=1398095 tracking the implementation of this in Gecko, and we are committed to implementing this soon-ish.

Notifying @jleyba of this change, since he added promises to the WebDriver spec in the first place.

christian-bromann:
@whimboo @andreastt I applied your feedback and realised that Geckodriver is actually already handling rejected promises well as `unknown error`. However that I find confusing is that `execute_script` is not recognising `script` timeouts when returning a promise. I guess this should be mentioned in the spec or fixed in the driver. A script can therefor stale the command when the promise is never resolved. Will update PR text.

christian-bromann:
Also question: the reason why I initial thought that Geckodriver doesn't handle rejected promises correctly is because I rejected it with a string `reject('error')` instead an error `reject(new Error('error'))`. Do we want to have a check for this in the spec to transform it to an error or resolve it and treat it as resolved promises which definitely can cause confusion.

Running a small test showed that Chrome is still rejecting the promise whatsoever:
```js
Promise.reject('hi').then(
  () => console.log("a"),
  () => console.log("b"))
> "b"
```

So we might want to also return with 500 and whatever was put into the reject method as payload.

webdriver/tests/execute_script/promise.py
andreastt: OK.

webdriver/tests/execute_script/promise.py
```diff
@@ -0,0 +1,80 @@
...
+        """)
+    assert_success(response, "foobar") 
+
+
+def test_promise_reject(session):
+    response = execute_script(session, """
+        return Promise.reject(new Error('my error'));
+        """)
+    assert_error(response, "unknown error")
+    assert response.body["value"]["message"] == "Error: my error"
```
andreastt: The error message is entirely implementation-defined, so we can’t test this.

webdriver/tests/execute_script/promise.py
```diff
@@ -0,0 +1,80 @@
...
+        """)
+    assert_error(response, "timeout error")
+
+
+def test_promise_reject_timeout(session):
+    session.timeouts.script = .1
+    response = execute_script(session, """
+        return new Promise(
+            (resolve, reject) => setTimeout(
+                () => reject(new Error('my error')),
```
andreastt: Why an `Error`  type here? Is it significant? I’d prefer if we called `reject` without any arguments.

webdriver/tests/execute_script/promise.py
```diff
@@ -0,0 +1,80 @@
...
+    assert_error(response, "unknown error")
+    assert response.body["value"]["message"] == "Error: my error" 
+
+
+def test_promise_resolve_timeout(session):
+    session.timeouts.script = .1
+    response = execute_script(session, """
+        return new Promise(
+            (resolve) => setTimeout(
+                () => resolve('success'),
```
andreastt: The "success" string is not significant to the test, so let’s call `resolve` without arguments.

webdriver/tests/execute_script/promise.py
```diff
@@ -0,0 +1,80 @@
...
+    response = execute_script(session, """
+        return new Promise(
+            (resolve, reject) => setTimeout(
+                () => reject(new Error('my error')),
+                50
+            )
+        );
+        """)
+    assert_error(response, "unknown error")
+    assert response.body["value"]["message"] == "Error: my error"
```
andreastt: The message is entirely implementation-defined, so we can’t test for this.

andreastt (changes requested):
One thing that occurred to me is that we also want to test if awaiting promises internally in a script works:

```python
execute_script("""
    await new Promise(…);
    // do something else
    return "foobar";
    """)
```

This means the wrapping function that WebDriver applies internally needs to be marked as an asynchronous function, but this may not be called out in the specification currently. Perhaps this should be solved in a separate patch as it likely requires prose changes as well.

As @whimboo pointed out, I’d also like you to address changes to the Execute Async Script command.

andreastt:
@christian-bromann In your latest update, making a distinction between delayed and timed out promises is good.

> Also question: the reason why I initial thought that Geckodriver doesn't handle rejected promises correctly is because I rejected it with a string reject('error') instead an error reject(new Error('error')). Do we want to have a check for this in the spec to transform it to an error or resolve it and treat it as resolved promises which definitely can cause confusion.

What the spec says is that the rejected value should be placed in a `data` field, so we should have a test for that.

What is _not_ called out in the spec is actually how to serialise the rejected data. For example, how do we know what to conver the `Error` into? I think this may be a different bug that we should address separately of this PR.

Does that answer your question?

webdriver/tests/execute_script/promise.py
```diff
@@ -0,0 +1,80 @@
...
+        );
+        """)
+    assert_success(response, "foobar") 
+
+
+def test_promise_reject(session):
+    response = execute_script(session, """
+        return Promise.reject(new Error('my error'));
+        """)
+    assert_error(response, "unknown error")
```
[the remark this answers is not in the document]
christian-bromann: fixed

webdriver/tests/execute_script/promise.py
christian-bromann: removed

webdriver/tests/execute_script/promise.py
christian-bromann: This is what I wanted to address with https://github.com/web-platform-tests/wpt/pull/13781#issuecomment-434323660. `Promise.reject` expects a reason as to why the promise was rejected. It doesn't and probably can't specify the type of this reason. For debugging purposes [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject#Description) recommends to use an error instance. I can't find anything specific in the EcmaScript [spec](https://www.ecma-international.org/ecma-262/6.0/#sec-promise.reject) though.

webdriver/tests/execute_script/promise.py
christian-bromann: removed

webdriver/tests/execute_script/promise.py
christian-bromann: removed

christian-bromann:
@andreastt I added two more test types to this:

- unveiling of Promise.all
- proper internal wrapping into `async` function (fails in Geckodriver)

I see the following things need to get spec'ed out:

- wrapping reject reasons into an error, I suggest the following:
    - if reason instance of `Error` return with `javascript error` and WebDriver error message containing error message
    - if reason is type of `string` return with `javascript error` and WebDriver error message as that string
    - anything else return `unknown error`
- function body of script argument has to be executed within an `async` context (https://w3c.github.io/webdriver/#dfn-extract-the-script-arguments-from-a-request needs enhancement)

I would love to create an issue for both in the Webdriver repo if you think that is the right way to move forward.

webdriver/tests/execute_script/promise.py
andreastt: OK.

webdriver/tests/execute_script/promise.py
```diff
@@ -0,0 +1,111 @@
...
+        ]);
+        """)
+    assert_success(response, [1, 2])
+
+
+def test_await_promise_resolve(session):
+    response = execute_script(session, """
+        const res = await Promise.resolve('foobar');
+        return res;
+        """)
```
andreastt: I would like to see a similar test where the promise rejects, to test that the return value _isn’t_ what is being returned. But happy to see a follow-up PR for that!

andreastt (approved):
This looks good now, with two caveats:

- Missing test for `data` field in error object.
- Test that "foo" isn’t returned in `await Promise.reject(…); return "foo";`.

@christian-bromann Do you want to take on fixing these up as well in follow-up PRs?

andreastt:
Before we can land this, you first need to fix the lints: https://travis-ci.org/web-platform-tests/wpt/jobs/448776034

christian-bromann:
@andreastt I fixed lint error and added a return value to the `await Promise.reject()` test we already had. I am not sure though what you mean by:

> Missing test for data field in error object.

christian-bromann:
@andreastt yes

PR #13784, opened 2018-10-30
Move getDisplayMedia to MediaDevices

youennf:
Tests for PR https://github.com/w3c/mediacapture-screen-share/pull/86

### DIFF
--- a/screen-capture/getdisplaymedia.https.html
+++ b/screen-capture/getdisplaymedia.https.html
@@ -1,13 +1,15 @@
 <!doctype html>
 <meta charset=utf-8>
-<title>RTCDTMFSender.prototype.insertDTMF</title>
+<title>getDisplayMedia</title>
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <script>
   'use strict';
+test(() => {
+  assert_idl_attribute(navigator.mediaDevices, 'getDisplayMedia');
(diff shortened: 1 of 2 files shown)


interfaces/screen-capture.idl
```diff
@@ -3,7 +3,7 @@
 // (https://github.com/tidoust/reffy-reports)
 // Source: Screen Capture (https://w3c.github.io/mediacapture-screen-share/)
 
-partial interface Navigator {
+partial interface MediaDevices {
```
lukebjerring: @foolip - (assuming this change has been made to the spec), this IDL change will be scraped at a later date, but is it still better to bundle it here with the changing tests to keep them behaving as expected?

foolip: The change matches https://github.com/w3c/mediacapture-screen-share/pull/86 which hasn't been merged yet, but I think it's fine to merge spec and test update together like this. It can result in some confusion if a PR is created to revert the change due to slower propagation of the spec changes through reffy-reports, but we should just fix our tooling if it keeps happening.

youennf: spec PR was merged, let's merge this one too.

alvestrand:
Redundant with https://github.com/web-platform-tests/wpt/pull/13855 - let's see if there's anything left to merge after that one lands.

alvestrand (approved):
The move itself has been done by another PR, but the remaining updates are good stuff.
Approving and merging.

PR #13789, opened 2018-10-30
Make Server-Timing tests more resilient

cvazac:
Per spec, the performance timeline is not guaranteed to have entries for blocking resources during `load` callbacks - which has led to intermittent failures. Adding `setTimeout` of zero should clean those up.


server-timing/server_timing_header-parsing.https.html
```diff
@@ -7,7 +7,8 @@
 <head>
     <meta charset='utf-8' />
     <script src="/resources/testharness.js"></script>
-    <script src='/resources/testharnessreport.js'></script>
+    <script src="/resources/testharnessreport.js"></script>
+    <script src="/common/performance-timeline-utils.js"></script>
```
yoavweiss: Including this once seems enough :)

yoavweiss (approved):
LGTM % nit

server-timing/server_timing_header-parsing.https.html
cvazac: whoops! thx - fixed.

common/performance-timeline-utils.js
```diff
@@ -42,3 +42,10 @@ function test_entries(actualEntries, expectedEntries) {
     }
   })
 }
+
+function delayedLoadListener(callback) {
+  window.addEventListener('load', function() {
+    // TODO(cvazac) Remove this setTimeout when spec enforces sync entries.
```
gsnedders: Can you file a bug for this to track it?

cvazac: https://github.com/web-platform-tests/wpt/issues/13874

PR #13795, opened 2018-10-30
Enable unified plan by default for Chrome Dev

foolip:
As documented in https://webrtc.org/web-apis/chrome/unified-plan/.

The plan is to enable this by default in Chrome 72, which is the current Chrome
Dev version. Enabling it now will make wpt.fyi more accurate reflect the likely
status of Chrome 72. (It's expected this will cause more tests to pass.)

### DIFF
--- a/tools/wpt/run.py
+++ b/tools/wpt/run.py
@@ -264,6 +264,7 @@ class Chrome(BrowserSetup):
         if kwargs["browser_channel"] == "dev":
             logger.info("Automatically turning on experimental features for Chrome Dev")
             kwargs["binary_args"].append("--enable-experimental-web-platform-features")
+            # TODO(foolip): remove after unified plan is enabled on Chrome stable
             kwargs["binary_args"].append("--enable-features=RTCUnifiedPlanByDefault")
 
         # Allow audio autoplay without a user gesture.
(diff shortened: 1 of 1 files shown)


foolip:
Bah, racing with @wpt-pr-bot for reviewer assignment. I'll add back.

Hexcles (approved):
LGTM

Nit: please add a TODO to remove this when Chrome 72 becomes stable.

henbos (approved):
LGTM!

foolip:
I've added a TODO!

foolip:
The Taskcluster run for this is finished: https://tools.taskcluster.net/groups/dH86-aNxSTKOHx9iERDEQQ

@Hexcles, what's the expected delay from a finished run to seeing it in wpt.fyi? (I don't see it yet.)

foolip:
@lukebjerring, crazy idea, but when we have a custom status check, we could update it post merging to also point to a diff between the full runs :) That's the view I'm trying to find manually now.

Hexcles:
@foolip https://wpt.fyi/results/?product=chrome&max-count=2&diff

foolip:
Neat! @henbos is that about what you expected?

lukebjerring:
No, that's not a crazy idea at all. Might even be a good starting place for the status check.

foolip:
Removed again in https://github.com/web-platform-tests/wpt/pull/15672.

PR #13799, opened 2018-10-30
Remove ES6 construct from WebDriver element click center point tests

jimevans:
Older user agents (read: "Internet Explorer") do not support the
ES6 "fat arrow" construct. Changing the test event handler to the
older function-based syntax.


whimboo (approved):
Sorry, that this happened again. The change looks fine.

andreastt:
So I don’t feel particularly sorry about landing a change that passes CI. If we want to block changes that don’t work in IE, I’m sorry to say IE needs to be added to CI.

PR #13802, opened 2018-10-30
Use `test-jobs --includes` to simplify tools/ci/ scripts

foolip:


lukebjerring (approved):
Looks reasonable. Merge if you've verified it locally.

foolip:
The way I can tell this works is from https://travis-ci.org/web-platform-tests/wpt/builds/448613710, that some jobs were run and that some were not, and in particular that tools/ unittests ran.

PR #13809, opened 2018-10-31
Adds Self Describing section to rendering docs

### LINKED ISSUE #5294: Additional hyperlinks
Originally posted as https://github.com/gsnedders/wpt-docs/issues/11 by @frivoal on 09 Jan 2017, 01:49 UTC:

> Once #10 is solved, it would be good for the [be-self-describing](https://gsnedders.github.io/wpt-docs/writing-tests/general-guidelines#be-self-describing) section to link to these guidelines.

akansha2608:
Fixes #5294

### DIFF
--- a/docs/_writing-tests/rendering.md
+++ b/docs/_writing-tests/rendering.md
@@ -13,7 +13,6 @@ given test passed or not.
 
 Tests should make it obvious when they pass and when they fail. It shouldn't be necessary to consult the specification to figure out whether a test has passed of failed.
 
-
 ## Indicating success
 
 Success is largely indicated by the color green; typically in one of
--- a/docs/_writing-tests/rendering.md
+++ b/docs/_writing-tests/rendering.md
@@ -9,7 +9,7 @@ these are especially using for [visual][] tests which need to be manually
 judged and following common patterns makes it easier to correctly tell if a
 given test passed or not.
 
-## Make Tests Self-Describing
+## Make tests self-describing
 
 Tests should make it obvious when they pass and when they fail. It shouldn't be necessary to consult the specification to figure out whether a test has passed of failed.
 
--- a/docs/_writing-tests/rendering.md
+++ b/docs/_writing-tests/rendering.md
@@ -93,4 +93,3 @@ colors!
 [visual]: {{ site.baseurl }}{% link _writing-tests/visual.md %}
 
 [ref-filled-green-100px-square]: https://github.com/w3c/csswg-test/blob/master/reference/ref-filled-green-100px-square.xht
-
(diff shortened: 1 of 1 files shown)


docs/_writing-tests/rendering.md
```diff
@@ -84,8 +84,14 @@ Obviously, none of this can be taken as absolute rules, as testing of
 color-related functionality will necessitate using some of these
 colors!
 
+## Self Describing
+
+Tests should described the situations when they and when they fail. Read more about it [here][].
+
```
sideshowbarker:
```
+ Tests should should make it obvious when they pass and when they fail. Read more about it [here][].
```

I suggest change that wording to this:

*Tests should make it obvious when they pass and when they fail. It shouldn't be necessary to consult the specification to figure out whether a test has passed of failed.*

…and completely remove the *Read more about it [here][]* part — because those two sentences above are the entire verbatim contents of what’s at https://github.com/web-platform-tests/wpt/blob/master/docs/_writing-tests/general-guidelines.md#be-self-describing (so the guidance is short enough to be copied over to here entirely, so there’d no longer be any need to link to the other file).

And to be clear, IMHO at least it’s not bad to have the exact same text verbatim in two places in the docs. It’s all about context and providing readers with useful information at any point of use where they might need it — and readers might have use for the same information in multiple places.

sideshowbarker (changes requested):

docs/_writing-tests/rendering.md
```diff
@@ -84,8 +84,14 @@ Obviously, none of this can be taken as absolute rules, as testing of
 color-related functionality will necessitate using some of these
 colors!
 
+## Self Describing
```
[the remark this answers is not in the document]
championshuttler: Done. :100:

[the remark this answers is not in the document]
akansha2608: Hi @sideshowbarker thanks for the suggestions, made the requested changes

docs/_writing-tests/rendering.md
akansha2608: That makes sense to me. Thank you.
Made the requested changes.

Please take a look.

docs/_writing-tests/rendering.md
```diff
@@ -9,6 +9,10 @@ these are especially using for [visual][] tests which need to be manually
 judged and following common patterns makes it easier to correctly tell if a
 given test passed or not.
 
+## Make Tests Self-Describing
```
sideshowbarker:
```
+ ## Make Tests Self-Describing
```

One final request: Change this so that only the first word of the heading is capitalized:

    ## Make tests self-describing

…because that’s the heading style used in the rest of this file (and I think throughout the rest of the docs.

(Sorry I should have caught that initially…)

sideshowbarker (changes requested):

docs/_writing-tests/rendering.md
akansha2608: Sorry my bad i should also kept in mind.

Fixes now though :smile:

sideshowbarker (approved):
Thanks!

PR #13812, opened 2018-10-31
add a warning about runner issues

### LINKED ISSUE #10843: Add big red warning to /tools/runner/index.html
I've heard from multiple people who have tried to use http://w3c-test.org/tools/runner/index.html and run into various problems, without realizing it's not a well-supported way to run WPT.

We should discourage its use with a warning, so that people only use it if they really want to.

akansha2608:
Fixes #10843 
cc: @foolip

### DIFF
--- a/tools/runner/index.html
+++ b/tools/runner/index.html
@@ -21,6 +21,13 @@
 </header>
 
 <div class="container">
+  <div class="alert alert-warning">
+        This runner does not fully support all test types and has a number of
+        <a href="https://github.com/web-platform-tests/wpt/labels/runner">known issues</a>.
+        <code>./wpt run</code> is a more well-supported runner, see the documentation on
+        <a href="https://web-platform-tests.org/running-tests/">running tests</a>.
+  </div>
+
   <div id="testControl" class="panel panel-default">
     <div class="panel-body">
       <form id='options' class='horizontal-form' onsubmit='return false;'>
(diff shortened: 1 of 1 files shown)


tools/runner/index.html
```diff
@@ -91,6 +91,12 @@
     </div>
   </div>
 
+  <div class="warning">
```
foolip: Thanks for working on this! I had a look at https://getbootstrap.com/docs/3.3/components/#panels and the bootstrap.css in this repo, and think this structure might result in a more warning-looking styling:

```
  <div class="alert alert-warning">
      This runner does not fully support all test types and has a number of
      <a href="https://github.com/web-platform-tests/wpt/labels/runner">known issues</a>.
      <code>./wpt run</code> is a more well-supported runner, see the documentation on
      <a href="https://web-platform-tests.org/running-tests/">running tests</a>.
  </div>
```

WDYT?

If you like that, can you put it before big controls box, first thing in `<div class="container">`?

akansha2608: Hi @foolip thanks for the review and suggestions :smile: it makes sense to me , let me make changes and share the screenshot for the same :100:

akansha2608: Its looking like this after making the suggested changes.

image

foolip: That looks great, thanks!

foolip (approved):

PR #13817, opened 2018-10-31
Remove cors/allow-headers.htm

annevk:
It's a duplicate of an earlier cors/origin.htm and doesn't actually test Access-Control-Allow-Headers.


annevk:
Seems this basically wasn't noticed in #112.

jdm (approved):

PR #13820, opened 2018-10-31
[motion] offset-rotate computed value

ewilligers:
https://drafts.fxtf.org/motion-1/#offset-rotate-property
reverse is the same as specifying auto 180deg.

https://drafts.csswg.org/css-values-4/#angles
All <angle> units are compatible, and deg is their canonical unit.


jihyerish (approved):
Thanks! Looks good to me.

PR #13823, opened 2018-10-31
Add nested-context resource-timing tests

yoavweiss:


bzbarsky (dismissed):

resource-timing/embed-reload.sub.html
```diff
@@ -0,0 +1,28 @@
...
+<script>
+    let t = async_test("Test that HTML embed reloads are not observable by the parent");
+    function verify_iframe_reload_not_observable() {
+        let entries = performance.getEntriesByType("resource");
+        let found_first_iframe = false;
+        for (entry of entries) {
+            if (entry.name == "{{location[scheme]}}://{{host]}}:{{ports[http][0]}}/resource-timing/resources/iframe-that-reloads.html") {
+                found_first_iframe = true;
+            }
+            assert_not_equals(entry.name, "{{location[scheme]}}://{{host]}}:{{ports[http][0]}}/resource-timing/resources/reloaded_iframe.html", "Reloaded iframe should not be observable");
```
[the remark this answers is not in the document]
yoavweiss: Indeed! Good catch

> I wonder whether it's feasible to have a complete whitelist of resource timing names that _should_ be present, and fail the test if they're not all present or if something else is present. That would avoid issues like this...
> 

There seem to be differences between browsers on that front (which is not great...), so better not go the whitelist route.

resource-timing/embed-reload.sub.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8" />
+<title>Resource Timing embed reload</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+    let t = async_test("Test that HTML embed reloads are not observable by the parent");
```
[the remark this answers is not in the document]
yoavweiss: Indeed, it's testing location.href assignment. Should we also test `location.reload()`? Won't these URLs already be exposed? Or maybe href assignment and then reload?

resource-timing/embed-reload-back.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8" />
+<title>Resource Timing embed reload - back button navigation</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+    let t = async_test("Test that object reloads are not observable by the parent, even after history navigations");
```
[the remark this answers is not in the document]
yoavweiss: ack

resource-timing/embed-refresh.sub.html
```diff
@@ -0,0 +1,28 @@
...
+<title>Resource Timing embed refresh</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+    let t = async_test("Test that an HTML embed refreshes are not observable by the parent");
+    function verify_iframe_refresh_not_observable() {
+        let entries = performance.getEntriesByType("resource");
+        let found_first_iframe = false;
+        for (entry of entries) {
+            if (entry.name == "{{location[scheme]}}://{{host]}}:{{ports[http][0]}}/resource-timing/resources/iframe-that-refreshes.html") {
```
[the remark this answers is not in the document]
yoavweiss: Did that and it's indeed cleaner. Thanks

resource-timing/embed-refresh.sub.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8" />
+<title>Resource Timing embed refresh</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+    let t = async_test("Test that an HTML embed refreshes are not observable by the parent");
+    function verify_iframe_refresh_not_observable() {
```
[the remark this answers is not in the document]
yoavweiss: ack

resource-timing/embed-reload.sub.html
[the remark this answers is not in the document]
bzbarsky: Hmm.  What are the differences?  Could we have a list of must-haves and a list of may-haves, maybe?  And file bugs on specs and/or browsers as needed to eliminate the may-haves?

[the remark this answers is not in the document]
yoavweiss: One difference ATM is failed fetches (See them in Firefox, but not Chrome), which is currently a MAY in the spec.
Once we'll define a stricter processing model, I think we can also align the implementations on a stricter behavior, where the entries are predictable in all engines.

yoavweiss:
@bzbarsky - friendly ping :)

yoavweiss:
@bzbarsky - post holidays ping :)

annevk (commented):
It also seems like a lot of the JavaScript could be abstracted somehow as the files look very similar. That would also help in reviewing them (and presumably when adding more tests at some point).

I hope this high-level review is useful.

I haven't run these yet, but if there's browser failures, are there bugs filed on those?

resource-timing/embed-navigate-back.html
```diff
@@ -0,0 +1,16 @@
...
+<html>
+<head>
+<meta charset="utf-8" />
+<title>Resource Timing embed navigate - back button navigation</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+    let t = async_test("Test that embed navigates are not observable by the parent, even after history navigations");
+    window.open("resources/embed-navigate-back.html");
+    addEventListener("message", t.step_func(e => {
```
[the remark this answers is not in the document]
yoavweiss: yup

lint.whitelist
```diff
@@ -186,6 +186,8 @@ SET TIMEOUT: paint-timing/resources/subframe-painting.html
 SET TIMEOUT: payment-request/allowpaymentrequest/setting-allowpaymentrequest-timing.https.sub.html
 SET TIMEOUT: preload/single-download-preload.html
 SET TIMEOUT: resize-observer/resources/iframe.html
+SET TIMEOUT: resource-timing/resources/*-reload-back.sub.html
```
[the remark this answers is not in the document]
yoavweiss: Indeed

resource-timing/embed-navigate-back.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8" />
+<title>Resource Timing embed navigate - back button navigation</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+    let t = async_test("Test that embed navigates are not observable by the parent, even after history navigations");
+    window.open("resources/embed-navigate-back.html");
```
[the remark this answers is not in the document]
yoavweiss: Did not know that. Thanks! :)

yoavweiss:
> It also seems like a lot of the JavaScript could be abstracted somehow as the files look very similar. That would also help in reviewing them (and presumably when adding more tests at some point).
> 
> I hope this high-level review is useful.
> 
> I haven't run these yet, but if there's browser failures, are there bugs filed on those?

Thanks for reviewing! :)

I tidied up and coalesced the JS.

Haven't filed bugs yet, but I will.

resource-timing/embed-reload.sub.html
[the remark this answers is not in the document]
annevk: Are there follow-up issues for this?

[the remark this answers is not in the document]
yoavweiss: https://github.com/w3c/resource-timing/issues/39 is the umbrella issue for that. #12 was the more specific one that covered failed fetches

resource-timing/embed-navigate-back.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8" />
+<title>Resource Timing embed navigate - back button navigation</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+    async_test(t => {
+        window.open("resources/embed-navigate-back.html");
```
annevk: Based on running the tests locally, it'd help a lot to add something like
```js
const openee = window.open(...);
t.add_cleanup(() => openee.close());
```
here and elsewhere where this pattern is used.

resource-timing/embed-navigate.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8" />
+<title>Resource Timing embed navigate</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="resources/nested-contexts.js"></script>
```
annevk: Add `<div id=log></div>` here so the `<embed>` doesn't show up at the top?

resource-timing/embed-refresh.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8" />
+<title>Resource Timing embed refresh</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="resources/nested-contexts.js"></script>
```
annevk: `<div id=top></div>` again.

annevk (commented):
A thing I noticed is that there's no cross-origin tests anymore. Is that intentional? `/common/get-host-info.sub.js` should make those relatively easy to add to add.

Without cross-origin it's not really clear to me why you'd even need the `URL` object here.

annevk:
Firefox appears to pass all these tests, FWIW.

resource-timing/embed-navigate-back.html
yoavweiss: Nice! Thanks :)

resource-timing/embed-navigate.html
yoavweiss: cool

resource-timing/embed-refresh.html
yoavweiss: yup

yoavweiss:
> A thing I noticed is that there's no cross-origin tests anymore. Is that intentional? `/common/get-host-info.sub.js` should make those relatively easy to add to add.

I didn't have them initially, but will add just to be on the safe side.


> Without cross-origin it's not really clear to me why you'd even need the `URL` object here.

We need the full URL for the `PerformanceEntry.name` comparison.

resource-timing/resources/nested-contexts.js
```diff
@@ -0,0 +1,95 @@
+let destination = location;
+if (location.search == "?crossorigin") {
+    host_info = get_host_info()
```
annevk: const host_info*

yoavweiss: oops

annevk (approved):

resource-timing/resources/nested-contexts.js
```diff
@@ -0,0 +1,95 @@
...
+        }
+        if (!found_first_document) {
+            opener.postMessage("FAIL - first document not exposed", "*");
+            return;
+        }
+        opener.postMessage("PASS", "*");
+    }
+    window.addEventListener("message", e=>{
+        if (e.data == "navigated") {
+            if (sessionStorage.navigated) {
```
bzbarsky: It might have been good to document somewhere the tight coupling between this and the exact test-serialization behavior we get from using promise_test for everything, thus ensuring that we never touch sessionStorage from two different tests in a racy way.

resource-timing/resources/nested-contexts.js
```diff
@@ -0,0 +1,95 @@
+let destination = location;
+if (location.search == "?crossorigin") {
+    const host_info = get_host_info()
+    destination = location.protocol + "//" + host_info["REMOTE_HOST"] + ":" + location.port;
```
bzbarsky: As a followup, it might be good to have not-samesite tests here too, just to test the cross-process case.  hostinfo does not expose NOTSAMESITE_HOST at the moment, so that would need to get fixed too...  I filed https://github.com/web-platform-tests/wpt/issues/15819 on that.

bzbarsky (commented):
@yoavweiss @annevk Thank you!   This is very readable now and quite clear in terms of what it's doing.  My sincere apologies for being so slow on reviewing this...  :( 

@yoavweiss I really appreciate the work you put into this and you pushing this over the finish line in spite of me timing out on you.  Thank you, again.

resource-timing/resources/nested-contexts.js
noamr: But this test is already cross-origin between localhost and some other host, or am I missing something?

yoavweiss: IIRC, @bzbarsky was asking for cross-origin and cross-site tests (so using hosts that have different domains)

noamr: Right but this test is both cross-origin and cross-site if runs from localhost. Maybe what's missing is same-site cross-origin?

bzbarsky: This test does not normally run "from localhost".  It runs from one of the define test hosts in the test configuration.

bzbarsky: And for details, see the `get_host_info` function's implementation.

noamr: Gotcha. I wasn't aware of these details. Thanks!

PR #13828, opened 2018-10-31
Ensure that lone surrogates don't appear in wptreport json

jgraham:
Python will happily handle "string" data containing codepoints in the surrogate
range. In the JSON output these are always escaped and appear as \uD800 or
similar. That's pretty unfortunate for interop since other languages don't
necessarily support such codepoints not representing a character in the
unicode set in their normal string type. Although it's possible to clean this
up on import it seems more reasonable to ensure that our primary interchange
format is unicode-clean. Therefore this change starts escaping lone surrogates
using the U+ notation so that \uD800 becomed U+D800.

Because Python, this requires two implementations. In UCS4 Python we know that
only lone surrogates should appear in string data; characters outside the BMP
will be represented as a single internal character (*technically* this is
untrue since someone may manually create a string containing a sequence of
surrogates that happens to form a pair; we don't care about handling this
case other than as two unpaired surrogates that happen to be adjacent). In
UCS2 Python surrogates appear in the string data literally, so we need to
check each time we encounter one whether it's a low surrogate followed by
a high surrogate and not escape in that case.

### DIFF
--- a/tools/wptrunner/wptrunner/formatters.py
+++ b/tools/wptrunner/wptrunner/formatters.py
@@ -5,7 +5,7 @@ import sys
 from mozlog.structured.formatters.base import BaseFormatter
 
 
-LONE_SURROGATE_RE = re.compile(u"[\uD800-\uDFFFF]")
+LONE_SURROGATE_RE = re.compile(u"[\uD800-\uDFFF]")
 
 
 def surrogate_replacement_ucs4(match):
--- a/tools/wptrunner/wptrunner/formatters.py
+++ b/tools/wptrunner/wptrunner/formatters.py
@@ -1,6 +1,53 @@
 import json
+import re
+import sys
 
 from mozlog.structured.formatters.base import BaseFormatter
+
+
+LONE_SURROGATE_RE = re.compile(u"[\uD800-\uDFFFF]")
+
+
+def surrogate_replacement_ucs4(match):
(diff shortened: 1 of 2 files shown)


gsnedders:
@jgraham did you solve the issues with this? or is it still broken on macOS, or wherever uses UCS2?

tools/wptrunner/wptrunner/formatters.py
```diff
@@ -1,8 +1,55 @@
 import json
+import re
+import sys
 
 from mozlog.structured.formatters.base import BaseFormatter
 
 
+LONE_SURROGATE_RE = re.compile(u"[\uD800-\uDFFFF]")
```
gsnedders:
```suggestion
LONE_SURROGATE_RE = re.compile(u"[\uD800-\uDFFF]")
```

tools/wptrunner/wptrunner/formatters.py
```diff
@@ -1,8 +1,55 @@
...
+
+
+def surrogate_replacement_ucs4(match):
+    return "U+" + hex(ord(match.group()))[2:]
```
gsnedders: This always outputs six characters, right? Given the match is always a single surrogate code point?

gsnedders (changes requested):

gsnedders (approved):

PR #13831, opened 2018-10-31
[Azure Pipelines] Split manifest update into a separate step

foolip:
The infrastructure/ step is suspiciously slow, and possibly this is
because the manifest is updating. Do that separate to see.

Drive-by: don't use multi-line script where not needed

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -31,8 +31,7 @@ jobs:
 
   # Installig Ahem in /Library/Fonts instead of using --install-fonts is a
   # workaround for https://github.com/web-platform-tests/wpt/issues/13803.
-  - script: |
-      sudo cp fonts/Ahem.ttf /Library/Fonts
+  - script: sudo cp fonts/Ahem.ttf /Library/Fonts
     displayName: 'Install Ahem font'
     condition: variables.run_wptrunner_infrastructure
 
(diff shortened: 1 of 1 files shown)


foolip:
https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=21&view=logs finished in just over 3 minutes, so this worked.

Somewhat concerning/puzzling is that https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=10&view=results (before this change) took over 5 minutes in the infrastructure/ step, with 4+ minutes spent waiting for the tests to start running. That was probably a full manifest regen. @jgraham, any idea why splitting out the manifest generation like this could result in a speedup?

foolip:
https://github.com/web-platform-tests/wpt/pull/13832 provides some clues about what's going on, from the logs:
```
Updating test manifest /Users/vsts/agent/2.141.1/work/1/s/MANIFEST.json
Unable to find test manifest
Creating test manifest
STDOUT: WARNING:manifest:No generated manifest found
STDOUT: INFO:manifest:Updating manifest
```

@jgraham, does `./wpt run` use a different code path for updating manifest than `./wpt manifest`? Or  might this be a bug with `--manifest MANIFEST.json`?

In either case I'd still like to do this PR, since the manifest will also be needed to determine affected tests, and splitting into more steps provides insight into which steps are slow.

jgraham (approved):
With required revert.

I think the bug you see is that we don't try very hard to find a recent manifest to download if there isn't yet one for the most recent commit on master. That seems fixable.

foolip:
Reverted by force pushing to drop the commit.

> I think the bug you see is that we don't try very hard to find a recent manifest to download if there isn't yet one for the most recent commit on master. That seems fixable.

Is this different for `./wpt manifest` and the code path that `./wpt run` uses? If not then I don't understand why this change makes it fast.

I'll merge this, but would like to know what sort of issue to file about this.

jgraham:
It could be timing dependent or something? I mean if we're not looking for an older manifest at all there certainly *is* a timing-dependent bug here since you'll get different behaviour when the task to create the manifest has finished and when it hasn't.

foolip:
Oh, yeah, that race condition when there are recent commits to master does exist. Let me try to revert this change in a PR to test that hypothesis. If it doesn't get slow again, then the hypothesis might be true.

foolip:
Filed https://github.com/web-platform-tests/wpt/issues/13853.

PR #13832, opened 2018-10-31
[Azure Pipelines] Enable TBPL logging for infrastructure/ tests

foolip:
This matches tools/ci/ci_wptrunner_infrastructure.sh.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -17,7 +17,7 @@ jobs:
 
   - script: |
       echo "Test jobs:"
-      ./wpt test-jobs | while read job; do
+      ./wpt test-jobs --all | while read job; do
         echo "$job"
         echo "##vso[task.setvariable variable=run_$job]true";
       done
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -59,6 +59,6 @@ jobs:
     displayName: 'Update manifest'
     condition: variables.run_wptrunner_infrastructure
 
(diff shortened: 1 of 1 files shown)


foolip:
I did this PR mostly to debug https://github.com/web-platform-tests/wpt/pull/13831, I don't really like the output format. @jgraham, do you think we should change tools/ci/ci_wptrunner_infrastructure.sh instead?

jgraham:
I mean I don't love the tbpl format either but it's designed for CI use cases and is the one supported by the reftest analyzer so I think it makes more sense to land this than change the other thing.

.azure-pipelines.yml
```diff
@@ -17,7 +17,7 @@ jobs:
 
   - script: |
       echo "Test jobs:"
-      ./wpt test-jobs | while read job; do
+      ./wpt test-jobs --all | while read job; do
```
jgraham: Need to revert this change, right?

jgraham (approved):
Assuming that you're reverting the `--all` change.

foolip:
> I mean I don't love the tbpl format either but it's designed for CI use cases and is the one supported by the reftest analyzer so I think it makes more sense to land this than change the other thing.

I tried downloading the log from this PR and putting it into the reftest-analyzer, but it doesn't work because there's a bunch of extra stuff, timestamps and a header at the top. I tried to trim it manually but seems to have failed.

I could instead log to a file and put that in artifacts for debugging. Would you like me to do that, or wait until there's a problem to debug?

jgraham:
Wait until there's a problem to debug.

PR #13842, opened 2018-11-01
Add test for document wg color

### LINKED ISSUE #2391: Add tests for Document#fgColor with special body elements.
<https://html.spec.whatwg.org/multipage/#dom-document-fgcolor> states:

> The attributes of the Document object listed in the first column of the following table must reflect the content attribute on the body element with the name given in the corresponding cell in the second column on the same row, if the body element is a body element (as opposed to a frameset element). When there is no body element or if it is a frameset element, the attributes must instead return the empty string on getting and do nothing on setting.

In particular, the following should be tested:
- [x] getting with no body or frameset element
- [ ] setting with no body or frameset element
- [ ] getting with both only a frameset element
- [ ] setting with both only a frameset element
- [ ] getting with both a body and a frameset element
- [ ] setting with both a body and a frameset element

See some existing tests in `html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/`.

akansha2608:
Fixes #2391

### DIFF
--- a/html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
+++ b/html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
@@ -1,0 +1,25 @@
+<!DOCTYPE html>
+<title>document wg color</title>
+<link rel="author" title="akansha2608" href="mailto:akanshasinghal999@gmail.com">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-fgcolor">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div id="log"></div>
+<script>
+test(function() {
+var body = document.createElement("body");
+document.documentElement.appendChild(body);
+document.fgColor = "red";
+document.bgColor = "red";
+document.linkColor = "red";
+document.vlinkColor = "red";
+document.alinkColor = "red";
+assert_equals(document.fgColor,"");
+assert_equals(document.bgColor,"");
+assert_equals(document.linkColor,"");
+assert_equals(document.vlinkColor,"");
+assert_equals(document.alinkColor,"");
+document.documentElement.removeChild(body);
+});
+</script>
--- a/html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
+++ b/html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
@@ -1,25 +1,30 @@
 <!DOCTYPE html>
 <title>document wg color</title>
-<link rel="author" title="akansha2608" href="mailto:akanshasinghal999@gmail.com">
 <link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-fgcolor">
 <link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html">
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
-<div id="log"></div>
+<body>
 <script>
 test(function() {
 var body = document.createElement("body");
 document.documentElement.appendChild(body);
 document.fgColor = "red";
-document.bgColor = "red";
-document.linkColor = "red";
-document.vlinkColor = "red";
-document.alinkColor = "red";
 assert_equals(document.fgColor,"");
+
+document.bgColor = "red";
 assert_equals(document.bgColor,"");
+
+document.linkColor = "red";
 assert_equals(document.linkColor,"");
+
+document.vlinkColor = "red";
 assert_equals(document.vlinkColor,"");
+
+document.alinkColor = "red";
 assert_equals(document.alinkColor,"");
+
 document.documentElement.removeChild(body);
 });
 </script>
+</body>
--- a/html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
+++ b/html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
@@ -8,46 +8,70 @@
 <link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html">
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
-<body text="red" linkColor="green" vlink="blue" alink="orange" bgColor="yellow">
+<body text="red" linkColor="red" vlink="red" alink="red"  bgColor="red">
 <script>
 test(function() {
-  assert_equals(document.fgColor, document.body.text);
 
-  assert_equals(document.linkColor, document.body.link);
-
-  assert_equals(document.vlinkColor, document.body.vLink);
-
-  assert_equals(document.alinkColor, document.body.aLink);
-
-  assert_equals(document.bgColor, document.body.bgColor);
-})
-test(function() {
-  document.body.text = red;
-
+  document.linkColor = "red";
+  
   assert_equals(document.fgColor, "red");
+
+  assert_equals(document.linkColor, "red");
+
+  assert_equals(document.vlinkColor, "red");
+
+  assert_equals(document.alinkColor, "red");
+
+  assert_equals(document.bgColor, "red");
+
+}, "testing every attribute is red")
+
+test(function() {
+  document.body.text = "blue";
+
+  assert_equals(document.fgColor, "blue");
 }, "document's fgColor IDL attribute reflects the body's text content attribute");
 
 test(function() {
-  document.body.link = green;
+  document.body.link = "green";
 
   assert_equals(document.linkColor, "green");
 }, "document's linkColor IDL attribute reflects the body's link content attribute");
 
 test(function() {
-  document.body.vLink = blue;
+  document.body.vLink = "blue";
 
   assert_equals(document.vlinkColor, "blue");
 }, "document's vlinkColor IDL attribute reflects the body's vLink content attribute");
 
 test(function() {
-  document.body.aLink = orange;
+  document.body.aLink = "orange";
 
   assert_equals(document.alinkColor, "orange");
 }, "document's alinkColor IDL attribute reflects the body's aLink content attribute");
 
 test(function() {
-  document.body.bgColor = yellow;
+  document.body.bgColor = "yellow";
 
   assert_equals(document.bgColor, "yellow");
 }, "document's bgColor IDL attribute reflects the body's bgColor content attribute");
+
+test(function(){
+   document.body.removeAttribute("text");
+   assert_equals(document.fgColor, "");
+
+   document.body.removeAttribute("link");
+   assert_equals(document.linkColor, "");
+
+   document.body.removeAttribute("vlink");
+   assert_equals(document.vlinkColor, "");
+
+   document.body.removeAttribute("alink");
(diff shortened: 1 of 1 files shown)


html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE html>
+<title>document wg color</title>
+<link rel="author" title="akansha2608" href="mailto:akanshasinghal999@gmail.com">
```
marcoscaceres: Git history provides authorship ;) this allows multiple people to work on different parts of tests over time.

html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
```diff
@@ -0,0 +1,25 @@
...
+<link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-fgcolor">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div id="log"></div>
```
marcoscaceres: Nit: I don’t think you need this div. The test harness adds it automatically.

html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
```diff
@@ -0,0 +1,25 @@
...
+<script>
```
marcoscaceres: Please add an explicit body tag before this script element. Otherwise, the test might run before a body is created causing it to randomly fail.

html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
```diff
@@ -0,0 +1,25 @@
...
+test(function() {
+var body = document.createElement("body");
+document.documentElement.appendChild(body);
+document.fgColor = "red";
```
marcoscaceres: Please  each assert after each declaration. That way, if one fails, it’s quicker to potentially see why something failed. Also, please add a blank new line after each assert. Helps make things visually a bit easier to read.

html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
```diff
@@ -0,0 +1,25 @@
...
+document.linkColor = "red";
+document.vlinkColor = "red";
+document.alinkColor = "red";
+assert_equals(document.fgColor,"");
+assert_equals(document.bgColor,"");
+assert_equals(document.linkColor,"");
+assert_equals(document.vlinkColor,"");
+assert_equals(document.alinkColor,"");
+document.documentElement.removeChild(body);
+});
```
marcoscaceres: Please add a description as a second argument to the test().

marcoscaceres (changes requested):
Couple of nits, but great start.

html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
marcoscaceres: To be clear, please remove this line.

html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
marcoscaceres: Oh, also, I think adding the div also creates a body automagically.

html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
```diff
@@ -0,0 +1,25 @@
...
+<link rel="author" title="akansha2608" href="mailto:akanshasinghal999@gmail.com">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-fgcolor">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div id="log"></div>
+<script>
+test(function() {
+var body = document.createElement("body");
+document.documentElement.appendChild(body);
```
marcoscaceres: I’m a bit unsure about why you are creating and removing this body. If it’s essential to the test, it would be good to add a comment.

marcoscaceres (changes requested):

html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
akansha2608: Removed it. :dancer:

html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
+<title>document wg color</title>
+<link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-fgcolor">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html">
```
akansha2608: @marcoscaceres i think we do not need to add one more link for help. What do you think about it?

marcoscaceres: Actually, we probably want to link all of the values:

https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-linkcolor
https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-vlinkcolor
https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-alinkcolor

And so on... as this test file is for each of them.

html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
zcorpan: div implies body though

marcoscaceres: Right. Working with Akansha on this in Slack. Stay tuned, better body coming:)

html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
```diff
@@ -0,0 +1,53 @@
...
+  assert_equals(document.fgColor, document.body.text);
+
+  assert_equals(document.linkColor, document.body.link);
+
+  assert_equals(document.vlinkColor, document.body.vLink);
+
+  assert_equals(document.alinkColor, document.body.aLink);
+
+  assert_equals(document.bgColor, document.body.bgColor);
+})
```
marcoscaceres: Nit: missing description.

html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
```diff
@@ -0,0 +1,53 @@
...
+<title>document wg color</title>
+<link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-fgcolor">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-linkcolor">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-vlinkcolor">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-alinkcolor">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-bgcolor">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/obsolete.html">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<body text="red" linkColor="green" vlink="blue" alink="orange" bgColor="yellow">
```
marcoscaceres: Let’s initially set all these to “red”. Later, we can change them dynamically to “green”. That will prevent false positives.

html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
```diff
@@ -0,0 +1,53 @@
...
+<script>
+test(function() {
+  assert_equals(document.fgColor, document.body.text);
+
```
marcoscaceres: I this test, let’s confirm that each of these is “red”. 
assert_equal(document.fgColor, “red”);

And so on...

html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
```diff
@@ -0,0 +1,53 @@
...
+  assert_equals(document.vlinkColor, document.body.vLink);
+
+  assert_equals(document.alinkColor, document.body.aLink);
+
+  assert_equals(document.bgColor, document.body.bgColor);
+})
+test(function() {
+  document.body.text = red;
+
```
marcoscaceres: Set this to “green”. And then check if it’s “green”. And do the samr for the rest.

marcoscaceres (changes requested):
Almost there! These are starting to look good.

html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/document-color-fg.html
```diff
@@ -0,0 +1,53 @@
...
+  assert_equals(document.alinkColor, "orange");
+}, "document's alinkColor IDL attribute reflects the body's aLink content attribute");
+
+test(function() {
+  document.body.bgColor = yellow;
+
+  assert_equals(document.bgColor, "yellow");
+}, "document's bgColor IDL attribute reflects the body's bgColor content attribute");
+</script>
```
marcoscaceres: I think we need one more set of tests, where we finally remove the attributes from `body` to make sure that they get reflected properly in `document`. So:

```JS
test(function(){
   document.body.removeAttribute("test");
   assert_equals(document.fgColor, "");
   // and so on...
}, "Removing the attributes causes them to be reflected as the empty string");
```

domenic: Aren't all these reflection tests already done in an automated fashion by https://github.com/web-platform-tests/wpt/blob/master/html/dom/elements-sections.js ?

marcoscaceres: Indeed... we don't need these tests :) @akansha2608 was learning how to read/process spec text. 

So, I think we should abandon this and do the "TODO" in https://github.com/web-platform-tests/wpt/blob/master/html/dom/elements-sections.js

marcoscaceres: So yeah, looking at the above, we could basically run everything above in a frameset and on a new body. 

@akansha2608, ping me on Slack on Monday and I’ll help you with it. In the meantime, read over the original bug to get a more clear sense of exactly what we need to test.

akansha2608:
Closing it as its already in https://github.com/web-platform-tests/wpt/blob/master/html/dom/elements-sections.js :)

Thanks for everyone to helping me :smile:

PR #13849, opened 2018-11-01
Fetch: Access-Control-Expose-Headers parsing

annevk:
See https://github.com/whatwg/fetch/issues/814 for context.

### DIFF
--- a/cors/access-control-expose-headers-parsing.window.js
+++ b/cors/access-control-expose-headers-parsing.window.js
@@ -2,7 +2,7 @@ promise_test(() => fetch("resources/access-control-expose-headers.json").then(re
 
 function runTests(allTestData) {
   allTestData.forEach(testData => {
-  	const encodedInput = encodeURIComponent(testData.input)
+    const encodedInput = encodeURIComponent(testData.input);
     promise_test(() => {
       const relativeURL = "resources/expose-headers.py?expose=" + encodedInput,
             url = new URL(relativeURL, location.href).href.replace("://", "://élève.");
--- a/cors/access-control-expose-headers-parsing.window.js
+++ b/cors/access-control-expose-headers-parsing.window.js
@@ -1,13 +1,15 @@
-function exposeTest(resource, desc) {
-  const url = new URL("resources/" + resource, location.href).href.replace("://", "://élève.");
+promise_test(() => fetch("resources/access-control-expose-headers.json").then(res => res.json()).then(runTests), "Loading JSON…");
 
-  promise_test(() => {
-    return fetch(url).then(res => {
-      assert_equals(res.headers.get("content-language"), "sure")
-      assert_equals(res.headers.get("x-custom"), null);
-    })
-  }, "Access-Control-Expose-Headers parsing: " + desc);
+function runTests(allTestData) {
+  allTestData.forEach(testData => {
+  	const encodedInput = encodeURIComponent(testData.input)
+    promise_test(() => {
+      const relativeURL = "resources/expose-headers.py?expose=" + encodedInput,
+            url = new URL(relativeURL, location.href).href.replace("://", "://élève.");
+      return fetch(url).then(res => {
+        assert_equals(res.headers.get("content-language"), "mkay");
+        assert_equals(res.headers.get("bb-8"), (testData.exposed ? "hey" : null));
+      });
+    }, "Parsing: " + encodedInput);
+  })
 }
-
-exposeTest("access-control-expose-headers-parsing.asis", "#1");
-exposeTest("access-control-expose-headers-parsing-2.asis", "#2")
--- a/cors/resources/access-control-expose-headers.json
+++ b/cors/resources/access-control-expose-headers.json
@@ -1,0 +1,62 @@
+[
+  {
+    "input": "access-control-expose-headers: BB-8",
+    "exposed": true
+  },
+  {
+    "input": "Access-Control-Expose-Headers: bb-8,,@#$#%%&^&^*()()11!",
+    "exposed": false
+  },
+  {
+    "input": "Access-Control-Expose-Headers: bb-8, no no",
+    "exposed": false
+  },
+  {
+    "input": "Access-Control-Expose-Headers: @#$#%%&^&^*()()11!,bb-8",
+    "exposed": false
+  },
+  {
+    "input": "Access-Control-Expose-Headers: bb-8\r\nAccess-Control-Expose-Headers: no",
+    "exposed": true
+  },
+  {
+    "input": "Access-Control-Expose-Headers: bb-8\r\nAccess-Control-Expose-Headers: no no",
+    "exposed": false
+  },
+  {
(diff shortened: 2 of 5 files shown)


cors/access-control-expose-headers-parsing.window.js
```diff
@@ -1,13 +1,15 @@
...
 
-  promise_test(() => {
-    return fetch(url).then(res => {
-      assert_equals(res.headers.get("content-language"), "sure")
-      assert_equals(res.headers.get("x-custom"), null);
-    })
-  }, "Access-Control-Expose-Headers parsing: " + desc);
+function runTests(allTestData) {
+  allTestData.forEach(testData => {
+  	const encodedInput = encodeURIComponent(testData.input)
```
jdm: nit: indentation.

cors/access-control-expose-headers-parsing.window.js
```diff
@@ -1,13 +1,15 @@
...
+    promise_test(() => {
+      const relativeURL = "resources/expose-headers.py?expose=" + encodedInput,
+            url = new URL(relativeURL, location.href).href.replace("://", "://élève.");
```
jdm: Shouldn't this use the configurable domain rather than hardcoding it?

jdm (changes requested):

cors/access-control-expose-headers-parsing.window.js
odinho: Many years ago `*.<whatever>` was indeed required for wpt tests. Because we'd test any random domain, and that would be different for different tests.

cors/access-control-expose-headers-parsing.window.js
```diff
@@ -1,13 +1,15 @@
...
-  promise_test(() => {
-    return fetch(url).then(res => {
-      assert_equals(res.headers.get("content-language"), "sure")
-      assert_equals(res.headers.get("x-custom"), null);
-    })
-  }, "Access-Control-Expose-Headers parsing: " + desc);
+function runTests(allTestData) {
+  allTestData.forEach(testData => {
+    const encodedInput = encodeURIComponent(testData.input);
+    promise_test(() => {
```
odinho: So these tests would be defined quite late (not after first-run). We wanted to not really do those if we could earlier. Often we'd define the tests, and only later do the running things.

But I can understand that things might have changed.

Why are you defining the tests in JSON:
```
[
  {
    "input": "access-control-expose-headers: BB-8",
    "exposed": true
  },
  {
    "input": "Access-Control-Expose-Headers: bb-8,,@#$#%%&^&^*()()11!",
    "exposed": false
  },
```
Instead of JS (where the tests would not have been runtime-defined):
```
runTest({input: "access-control-expose-headers: BB-8", exposed: true});
runTest({input: "Access-Control-Expose-Headers: bb-8,,@#$#%%&^&^*()()11!", exposed: false});
```
Or maybe there's some other system reading and defining test from the JSON? I see there's some very few tests defined like this already, but the other pattern is much more used still.

---
Could even inline the JSON in the test to get around "tests only defined after fetch has finished".

odinho (commented):
Shouldn't the spec PR actually go in first? I'm not quite sure what specification steps I should review this against as it is.

I think the test itself is improved :+1:.
I have a few questions on implementation, but that's all good.
I can review properly if you can tell me what to review against (your latest comment on https://github.com/whatwg/fetch/issues/827 says you'll think about what to do, so ... :P ).

cors/access-control-expose-headers-parsing.window.js
odinho: Okay :)

cors/access-control-expose-headers-parsing.window.js
odinho: I was wrong on the `*.<whatever>`, I see there's just a small list of required domains, and élève.` is one of them. :)

cors/access-control-expose-headers-parsing.window.js
```diff
@@ -1,13 +1,15 @@
...
+      const relativeURL = "resources/expose-headers.py?expose=" + encodedInput,
+            url = new URL(relativeURL, location.href).href.replace("://", "://élève.");
+      return fetch(url).then(res => {
+        assert_equals(res.headers.get("content-language"), "mkay");
+        assert_equals(res.headers.get("bb-8"), (testData.exposed ? "hey" : null));
+      });
+    }, "Parsing: " + encodedInput);
```
odinho: I much prefer reading the test output if you don't encode it here. Reads so much nicer, and the descriptions are mostly meant for human consumption, and it's easier to see what it actually tests. If you want to debug the thing, you can figure out the exact thing anyway.
```suggestion
    }, "Parsing: " + testData.input);
```

cors/resources/access-control-expose-headers.json
```diff
@@ -0,0 +1,62 @@
...
+  },
+  {
+    "input": "Access-Control-Expose-Headers: 'bb-8',bb-8",
+    "exposed": true
+  },
+  {
+    "input": "Access-Control-Expose-Headers: \"bb-8\",bb-8",
+    "exposed": false
+  }
+]
```
odinho: I've been reading the linked HTTP ABNF RFC (https://tools.ietf.org/html/rfc7230#section-3.2.6), which does seem define about the thing browsers do, expect the `"` DQUOTE wrapped content should allow you to send any-random thing through.

So proposing adding these tests, one positive test for all valid text:
```
  {
    "input": "Access-Control-Expose-Headers: valid_!#$&'*+-.^`|~1a%40z,bb-8",
    "exposed": true
  },
```
Two on the browser-specific (checked gecko and chromium (Opera + Chrome)) ignore of what the spec (as I read it which is probably wrong) allows:
```
 {
    "input": "Access-Control-Expose-Headers: \"bb-8\"",
    "exposed": false
  },
  {
    "input": "Access-Control-Expose-Headers: bb-8\r\nAccess-Control-Expose-Headers: \"no no\"",
    "exposed": false
  },
```
If we *would* want to follow the spec (at least as I read it), we could just flip `exposed: true` on those two (the tests existing would still fail as I read it).

odinho (approved):
LGTM.

I have two suggestions (1 + 2 more tests, human-readable descriptions).

cors/resources/access-control-expose-headers.json
annevk: What makes you think this header allows quoted strings?

odinho: Hm. I was confused. But I can explain.

Spec links to https://tools.ietf.org/html/rfc7230#section-3.2
I was reading `3.2.6 Field Value Components`: https://tools.ietf.org/html/rfc7230#section-3.2.6
It has the section on `quoted-string`, but Fetch doesn't actually specify that, and the "default" 3.2 also does not specify it.

The tests could still be included as a "don't do this" though, so no-one does the same error as me. But other people might be less likely to do such wrong readings though :)

jdm (approved):

PR #13850, opened 2018-11-01
Request/Response's clone() realm

annevk:
For https://github.com/whatwg/fetch/issues/825.

### DIFF
--- a/WebIDL/current-realm.html
+++ b/WebIDL/current-realm.html
@@ -50,6 +50,9 @@
 
        const cloneObj = obj.clone();
        assert_global(cloneObj);
+
+       const involvedCloneObj = self[val].prototype["clone"].call(cloneObj);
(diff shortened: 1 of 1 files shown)


WebIDL/current-realm.html
```diff
@@ -43,6 +43,16 @@
      }, val[0])
    })
 
+   ;["Request", "Response"].forEach(val => {
```
domenic: You can just loop over `[Request, Response]`?

domenic (approved):

WebIDL/current-realm.html
annevk: How would that work obtaining them from a different global?

domenic: I misread `self[0]` as `self`, nevermind.

domenic (approved):

PR #13851, opened 2018-11-01
[Azure Pipelines] Remove separate manifest step

foolip:
This reverts the important part of #13831.


foolip:
This is to investigate if manifest update is always slow when implicitly done in `./wpt run`.

foolip:
Oh, it became fast again. I'll file a bug about the race condition that probably caused the slowness.

foolip:
Filed https://github.com/web-platform-tests/wpt/issues/13853, closing this.

PR #13854, opened 2018-11-01
location.pathname should return "blank" when there is no browsing con…

cdumez:
…text

The pathname of "about:blank" is "blank", not "".


domenic (approved):
This looks correct to me per https://jsdom.github.io/whatwg-url/#url=YWJvdXQ6Ymxhbms=&base=YWJvdXQ6Ymxhbms= ; sorry for missing it while reviewing.

PR #13861, opened 2018-11-02
change test order

marcoscaceres:
Switching order of tests so these click ones can run before canMakePayment() gets upset.


marcoscaceres:
Actually, I might split this into two tests... we really want to make sure the first two tests pass. The other tests are optional.

rsolomakhin (approved):

PR #13862, opened 2018-11-02
webkit: provide port name as run info

zdobersek:
For the WebKit product, include the port name inside the run info, under the
'wk_port' key. This will come in handy when disabling tests in metadata files
for specific ports.

### DIFF
--- a/tools/wptrunner/wptrunner/browsers/webkit.py
+++ b/tools/wptrunner/wptrunner/browsers/webkit.py
@@ -71,7 +71,7 @@ def env_options():
 
 
 def run_info_extras(**kwargs):
-    return {"wk_port": kwargs["webkit_port"]}
+    return {"webkit_port": kwargs["webkit_port"]}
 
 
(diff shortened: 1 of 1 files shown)


tools/wptrunner/wptrunner/browsers/webkit.py
```diff
@@ -69,6 +70,10 @@ def env_options():
     return {}
 
 
+def run_info_extras(**kwargs):
+    return {"wk_port": kwargs["webkit_port"]}
```
jgraham: Can we use `webkit_port` rather than `wk_port`?

jgraham (changes requested):

tools/wptrunner/wptrunner/browsers/webkit.py
zdobersek: Yes.

jgraham (approved):

PR #13867, opened 2018-11-02
Rename the stability test job to affected_tests

foolip:
Based on tests being affected, we can run both stability checks and
just collect results, both of which are happening on Taskcluster now.

Taskcluster doens't use `./wpt test-jobs`, but Azure Pipelines will.


jgraham:
I don't really understand the argument here. I think stability is still the thing that this job can fail for, so calling it that seems more helpful for now.

PR #13869, opened 2018-11-02
Check for AudioContext to enable generating audio tracks with it

youennf:
WebKit exposes webkitAudioContext which is slightly different from AudioContext.
For WebRTC tests, it is simpler to rely on getUserMedia tests with mock active.

### DIFF
--- a/webrtc/RTCPeerConnection-helper.js
+++ b/webrtc/RTCPeerConnection-helper.js
@@ -358,7 +358,7 @@ const trackFactories = {
    */
   canCreate(requested) {
     const supported = {
-      audio: !!window.MediaStreamAudioDestinationNode,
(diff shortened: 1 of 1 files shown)


alvestrand (approved):
It's good to make explicit that this way is only supposed to be used on standards-compliant WebAudio.

alvestrand:
@nils-ohlmeier can you figure out what failed on Firefox? (if it still fails)

nils-ohlmeier:
@foolip @jgraham does either of you know where the Firefox logging variables get set? I think we should remove the ICE logging, as they are not very useful on WPT. But the logs are way to big to be useful.

foolip:
I defer to @jgraham on that.

PR #13880, opened 2018-11-02
[webdriver] Test serialization of script values

jugglinmike:


webdriver/tests/execute_async_script/execute_async.py
```diff
@@ -26,6 +26,82 @@ def test_no_browsing_context(session, closed_window):
...
+# >
+# >     instance of NodeList
+# >     instance of HTMLCollection
+# >     instance of Array
+# >         A new Array which length property is equal to the result of getting
+# >         the property length of value.
+# >     Otherwise
+# >         A new Object.
+#
+# https://w3c.github.io/webdriver/#dfn-clone-an-object
```
andreastt: Please leave out all quotes of the specification. We removed most of these some time ago because we weren’t able to keep them up with the changing spec.

webdriver/tests/execute_async_script/execute_async.py
```diff
@@ -26,6 +26,82 @@ def test_no_browsing_context(session, closed_window):
...
 
 
+@pytest.mark.parametrize("values", [
+      {"primitive": "null", "expected": None},
+      {"primitive": "undefined", "expected": None},
+      {"primitive": "true", "expected": True},
+      {"primitive": "false", "expected": False},
+      {"primitive": "23", "expected": 23},
+      {"primitive": "'a string'", "expected": "a string"}
+    ])
```
andreastt: nit: Instead of parametrizing on a dictionary with `primitive` and `expected` keys, I feel it would read better if you just exposed both values as arguments:

```python
@pytest.mark.parametrize("primitive,expected", [("null", None), ("undefined", None), ("true", True), …])
def test_primitive_serialization(session, primitive, expected):
    …
```

andreastt (changes requested):
It seems the specification in this area is not in sync with implementations. For example, in addition to NodeList, HTMLCollection, and Array, we expect WebDriver to also serialise the following collection types:

* Arguments
* FileList
* HTMLAllCollection
* HTMLFormControlsCollection
* HTMLOptionsCollection

There are some tests for this in https://github.com/web-platform-tests/wpt/blob/master/webdriver/tests/execute_script/collections.py, which this PR partially duplicates in `test_array_serialization` and `test_foreign_array_serialization`. The other tests for primitives look absolutely fine.

We have two options: either we land this as it is, seeing it tests the actual wording in the spec, and fix up the tests later after addressing the missing collection types, or we take out the collection tests and land only the primitives tests.

Let me know what you’d prefer!

gsnedders:
> We have two options: either we land this as it is, seeing it tests the actual wording in the spec, and fix up the tests later after addressing the missing collection types, or we take out the collection tests and land only the primitives tests.

We also have a third: make sure we have rough consensus that the spec is wrong, and test what we want everyone to do. Obviously in a case like this, there's little reason not to also land the spec changes quickly (which @andreastt already has).

andreastt:
Fair point @gsnedders. Seeing as https://github.com/w3c/webdriver/pull/1340, let’s land tests that are to its intent. There are already collection tests in https://github.com/web-platform-tests/wpt/blob/master/webdriver/tests/execute_script/collections.py so I’m leaning towards dropping all the collection tests from this PR + addressing the parametrization nitpick.

gsnedders:
@jugglinmike can we have at least one test that Edge fails for now?

jugglinmike:
This patch was originally motivated by the need for tests of collection serialization. That's why I wanted to delay merging until the corresponding specification text was clear.

However, I've just today come across a conformance issue in Edge/Edgedriver that concerns serialization of primitive values. In order to get that issue documented, I've limited the scope of this patch to primitive values (including a test case for Edge's bug).

I'll use https://github.com/w3c/webdriver/issues/1347 to continue the discussion about collections.

@andreastt mind taking another look?

jugglinmike:
@brendyna & @thejohnjansen The conformance issue I mentioned above is this: if a string value returned from "Execute Script" or "Execute Async Script" contains the null byte, Edgedriver truncates the string to the first occurrence of that character.

brendyna:
Thanks for reporting that issue, @jugglinmike. @thejohnjansen and I will take a look and report back.

Is the issue blocking this PR at all or will we simply pass a test once this bug is fixed on our end?

jugglinmike:
Thanks for the quick response! The bug isn't blocking this pull request, though it may motivate a change to the testharness.js framework: gh-14245

whimboo:
@jugglinmike as it looks like this PR was forgotten. Do you still have interest to update it to the current state on master?

jugglinmike:
Hey there, @whimboo! I have the interest but not the time.

whimboo:
@jugglinmike thanks for the reply! Let me actually add those tests while I'm working on https://bugzilla.mozilla.org/show_bug.cgi?id=1819029. As such they will be upstreamed later this week.

PR #13419, opened 2018-10-08
Don't switch back to the harness window during testharness tests.

jgraham:
Previously we kept switching back and forth between the harness window
and the window containing the test as we processed WebDriver
callbacks. But this is unnecessary; we can use the fact that we have
access to window.opener to set the callback on the opener window and
run it there, yes still have it block a WebDriver call in the test
window. This means that during the test we can keep the focus on the
test window, but put all the complex logic in the parent and retain
the ability to close the test window after each test to create a clean
environment.

Issue #13876, opened 2018-11-02
Move wptrunner testharness implementation into a single browsing context


jgraham:
We have some tests like /html/browsers/the-windowproxy-exotic-object/windowproxy-prototype-setting-goes-cross-origin-domain.sub.html that alter the browser window in such a way that `window.opener` appears to be cross-origin. These tests end up throwing a security error. To fix this we probably need to put all the harness code in a single browsing context, mostly using `testharnessreport.js` to ensure that there isn't a race during load.

Issue #13263, opened 2018-09-28
Ability to run all tests for some PRs


foolip:
In https://github.com/web-platform-tests/wpt/pull/11769 a handful of tests were broken accidentally. This was reported in https://github.com/web-platform-tests/wpt/issues/13204 and also some of the same tests were found to fail in an [attempt to roll into Chromium](https://chromium-review.googlesource.com/c/chromium/src/+/1249141/1).
<details>
<pre>
 cookie-store/http_cookie_and_set_cookie_headers.tentative.https.html
 cookies/http-state/charset-tests.html
 eventsource/format-field-id-2.htm
 eventsource/format-field-id.htm
 fetch/api/headers/header-values.html
</pre>
</details>

Relatedly, many changes in resources/ by @lukebjerring, @jugglinmike and me have been put through Chromium's CQ first to see if they would break existing tests.

To help prevent accidental regressions and make it easier to look for anticipated regressions, the ability to run *all* tests in Taskcluster for some PRs would be very good.

Strawman: including the magic string "Taskcluster-Full-Run" somewhere would cause Taskcluster to treat all tests as affected.

Lots of similar solutions might be possible.

@jgraham

jgraham:
This is probably implementable right away, but reasonably we should make a proper decision task if we're going to introduce this kind of complexity.

foolip:
Ah. And that requires switching away from .taskcluster.yml and using code to create tasks, I guess?

gsnedders:
Note that if we want to alert people to regressing tests (#7475) this may end up noisy for tools changes as long as there are flaky tests (and noisy to a far bigger degree than for normal PRs, given it'll expose flake in *any* test).

That said, I think this is probably worthwhile doing, given it's mostly the same work as needed to migrate the rest of our testing infrastructure to TC.

foolip:
This would have come very much in handy to avoid https://github.com/web-platform-tests/wpt/issues/14495, if it were easy to do a full run we would do it.

@jgraham does this still depend on moving to a decision task? How much work would that be?

We should also give some consideration to how the feature would work on Azure Pipelines, if we want to also get full runs of Edge or Safari for PRs down the line.

foolip:
Renaming to not be Taskcluster-specific, we want this for Azure Pipelines too.

foolip:
On Azure Pipelines it would be fairly easy to do this based on something in the commit message.

[...]

PR #13881, opened 2018-11-02
[wptrunner] Transmit test results via postMessage

jugglinmike:
A recent improvement to wptrunner modified the way data structures are
transmitted from the browser to the WebDriver server [1]. One
consequence of this change is that the data structures are created in a
JavaScript realm which differs from the realm in which the "Execute
Script" WebDriver command is running. As of version 17, Microsoft Edge
does not correctly serialize cross-realm values, so referenced
improvement unintentionally precluded testing in that browser.

Transmit data structures between realms via the `postMessage` API to
avoid this problem.

[1] https://github.com/web-platform-tests/wpt/pull/13419

---

This is the reason we've recently been reporting such [laughably incomplete results for Edge on wpt.fyi](https://wpt.fyi/results/?sha=8295368c82). I've also submitted [WebDriver spec tests for the expected serialization behavior](https://github.com/web-platform-tests/wpt/pull/13880)

### DIFF
--- a/tools/wptrunner/wptrunner/executors/runner.js
+++ b/tools/wptrunner/wptrunner/executors/runner.js
@@ -1,23 +1,35 @@
 document.title = '%(title)s';
 
 var message_queue = [];
+var source = null;
 
 /**
  * In preparation to execute a test, testdriver issues a number of messages
  * which should be queued until execution time. During execution, testdriver
  * consumes the queued messages by repeatedly "resuming" via `do_testharness`.
+ * The sequence of these events is variable, so the handler must account for
+ * cases where "resume" requests arrive before any messages are available (by
+ * caching the request's source) and where messages arrive before "resume"
+ * requests (by queuing the messages).
  */
 window.addEventListener("message", function(event) {
-  if (!event.data || !event.data.type) {
+  if (!event.data) {
     return;
   }
 
   if (event.data.type === "testdriver-resume") {
-    var next_message = message_queue.shift();
-    reply(event.source, next_message);
+    source = event.source;
   } else {
     message_queue.push(event.data);
   }
+
+  if (!source || !message_queue.length) {
+    return;
+  }
+
+  reply(source, message_queue[0]);
+  source = null;
+  message_queue.shift();
 }, false);
 
 function reply(source, data) {
--- a/tools/wptrunner/wptrunner/executors/runner.js
+++ b/tools/wptrunner/wptrunner/executors/runner.js
@@ -1,42 +1,32 @@
 document.title = '%(title)s';
 
-window.addEventListener(
-  "message",
-  function(event) {
-    window.message_queue.push(event);
-    window.process_next_event();
-  },
-  false
-);
+var message_queue = [];
 
-
-window.process_next_event = function() {
-  /* This function handles the next testdriver event. The presence of
-     window.testdriver_callback is used as a switch; when that function
-     is present we are able to handle the next event and when is is not
-     present we must wait. Therefore to drive the event processing, this
-     function must be called in two circumstances:
-       * Every time there is a new event that we may be able to handle
-       * Every time we set the callback function
-     This function unsets the callback, so no further testdriver actions
-     will be run until it is reset, which wptrunner does after it has
-     completed handling the current action.
-   */
-  if (!window.testdriver_callback) {
+/**
+ * In preparation to execute a test, testdriver issues a number of messages
+ * which should be queued until execution time. During execution, testdriver
+ * consumes the queued messages by repeatedly "resuming" via `do_testharness`.
+ */
+window.addEventListener("message", function(event) {
+  if (!event.data || !event.data.type) {
     return;
   }
-  var event = window.message_queue.shift();
-  if (!event) {
-    return;
-  }
-  var data = event.data;
 
+  if (event.data.type === "testdriver-resume") {
+    var next_message = message_queue.shift();
+    reply(event.source, next_message);
+  } else {
+    message_queue.push(event.data);
+  }
+}, false);
+
+function reply(source, data) {
   var payload = undefined;
 
   switch(data.type) {
   case "complete":
-    var tests = event.data.tests;
-    var status = event.data.status;
+    var tests = data.tests;
+    var status = data.status;
 
     var subtest_results = tests.map(function(x) {
       return [x.name, x.status, x.message, x.stack];
@@ -53,7 +43,11 @@ window.process_next_event = function() {
   default:
     return;
   }
-  var callback = window.testdriver_callback;
-  window.testdriver_callback = null;
-  callback([window.url, data.type, payload]);
-};
+
+  source.postMessage({
+      type: "testdriver-next message",
+      message: [window.url, data.type, payload]
+    },
+    "*"
+  );
+}
--- a/tools/wptrunner/wptrunner/executors/testharness_webdriver_resume.js
+++ b/tools/wptrunner/wptrunner/executors/testharness_webdriver_resume.js
@@ -1,3 +1,24 @@
 var callback = arguments[arguments.length - 1];
-window.opener.testdriver_callback = callback;
-window.opener.process_next_event();
+
+window.addEventListener("message", function handle(event) {
+  if (!event.data || event.data.type !== "testdriver-next message") {
+    return;
+  }
+
+  window.removeEventListener("message", handle);
+
+  callback(event.data.message);
+});
+
+/**
+ * The current window and its opener belong to the same domain, making it
+ * technically possible for data structures to be shared directly. Some
+ * browser/WebDriver implementations [1] do not correctly serialize Arrays from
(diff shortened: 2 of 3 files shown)


jugglinmike:
[I posted some more context to the relevant issue in the results-collection project.](https://github.com/web-platform-tests/results-collection/issues/563#issuecomment-435522838)

foolip:
@jugglinmike, infrastructure/ tests are failing on Travis, can you see about fixing those?

jugglinmike:
Looks like I was a little too aggressive in trying to remove global state. I think I've resolved this by caching the `origin` in the realm of the message recipient. This is a crucial section of code, though, so vetting it through all of WPT seems prudent:

https://tools.taskcluster.net/groups/Mmj2u8piTuyoDlvdad2gGg

This branch is currently based on ffeba23ebd16b9adb2e0e2b5fe6923fef9bb9e5e, so we should compare with the Taskcluster results for that commit:

https://tools.taskcluster.net/groups/dFx-WCDHRzSMGzj4wuxF6Q

foolip:
@lukebjerring @Hexcles how can we get https://tools.taskcluster.net/groups/Mmj2u8piTuyoDlvdad2gGg into wpt.fyi for a comparison?

This is a perfect case for https://github.com/web-platform-tests/wpt/issues/13263 :)

lukebjerring:
The Taskcluster webhook takes all the task groups and turns them into artifact URLs, and then posts to the results receiver with a full list of all of those URLs.. might be worth extracting that code, and exposing it into another api endpoint which is Taskcluster specific.

However, as a one off, you could reproduce the same list of urls and send the POST event yourself?

lukebjerring:
So, with a little scrape of the UI:

    Array.from(document.querySelectorAll('a')).map(a => /groups\/[^\/]+\/tasks\/([^\/]+)\//.exec(a.href)).filter(m => !!m).map(m => m[1]).filter(m => m !== "null")

Looks like the task ID is in the URL you gave (`Mmj2u8piTuyoDlvdad2gGg`) and the group IDs are:

    ["ZVChn1bdRcSX5I56CabVbQ", "ViEYc7BHSVCJi-CXO6cqaw", "XWoTo0t7TZqOWoeCvBcOWg", "IEOm-11CR82h2FwGpmosBg", "Yd_5Hnt-S4qsGeHPJhTBig", "R6GiWrXzTg-kejTnU30zvA", "Fau84efbRaCGblDHiGFHcA", "MwmJtqrKRNS77YBOVd5XGQ", "BAeb_zRBS0aFh0iIIlyCKA", "b4Hp03NYRRSekLSnCYn5qg", "bp0Gqv5WRZG0DBrj63HBhA", "DOPVhU7JQ92aPrcp7ppExw", "NRh6qioERYOqz6bmUG4ItQ", "HYtXlm4hQ6yU_nMvdIWoPQ", "Iea3Cya4RWm29-jrA6e1zQ", "OkivyodfR6K5hOA7k7Vfsw", "GiBI8IU4QsGaTMp0L26_2Q", "ToArMA6BShylgk2RBSj4zA", "QY-4hYBNSfKoQFKj2zCvIw", "J3apkKfVTZCOpf8yqvGICw", "PrNFDkkwTGW7Z8R8q8sQYw", "KmyFIDOmQ_yy5PZqCAOKjw", "dL_1ex-DSzmoGykBrbpVlA", "UVNAkAbMRYiKOdqBQIyaDA", "RQzufNthTeORh7EvgCe76w", "FGaW2YD0RV-VbMibuB2Gbg", "CtzBh4pVTbWQgK1lCW5xDg", "SYEXrRsvRWG-4l2u5hXFuw", "ZIF0d4wuTpe4FKeb_L0jJg", "OdplewErRBePy9Ip16EyNw", "Pth9wUmHSPGQRNJOauGayA", "ZnT3IIhXS4KxYiRTM1ly6w", "APC22dKoTOi6M8RcaOSBFg", "YZZOZ4lCQfyibIYOAbsXsg", "dMVwLtBAR8COF0nKk7wKTA", "NXoFXs8fQRqDinQ7HwO_BA", "NYS5CXqmR5e577VxRC0K2g", "Iq6I4zZCSzm7yhDYVgHGFg", "N2I9IpteRE6WnekVg7TunQ", "a5lDXY5dRcWRwLScsvPz6Q", "Okpv13iQSDyUUAj4vwcRbQ", "M_gtodHiSW-OhMnjqNYMvQ", "OlbqEI_ATSq1cmUgif-QtA", "O0Xj9tcbQoyPxhx3Jmg0pw", "d3l7NyrAQQeeZbwU2a7gXA", "MC0-dASJTNS0csqv5kZ3dQ", "AIGlcx0gSJ-dTsC92liuug", "AmAOmo5rT4qBG2XEpnmEvw", "KG_kQA2HTqmrhsIMGOsbUg", "RXc3as_fTTG4VSKyc7tibg", "LiL-UsQ_TGi_30EImKhgJQ", "dKVNfAaxS9G5_bXfNYz4VQ"]

lukebjerring:
Wait, nvm, we only need the TaskID:

    https://queue.taskcluster.net/v1/task/Mmj2u8piTuyoDlvdad2gGg/artifacts/public/results/wpt_report.json.gz"

lukebjerring:
In theory, the following scrape of URLs, and a hack for /admin/results/upload, might result in the output we want:
https://gist.github.com/lukebjerring/d4dd09930c4959d9c6611241da4dfc8e

I (cheekily) ran it as `jugglinmike`, so https://wpt.fyi/test-runs?label=jugglin-mike&max-count=100 will show if it's successful (and revision was set to `ffeba23ebd`)

foolip:
Amazing, thanks @lukebjerring!

@jugglinmikes, looks like there are regressions to investigate here.

gsnedders (changes requested):
@jugglinmike plz look through the regressions @lukebjerring linked

jugglinmike:
I took the weekend off, but this is on the docket for today.

jgraham:
FWIW we eventually need to work on getting this to work without `window.opener` being used at all, either via `postMessage` or otherwise. That's because there's a feature coming to allow nulling out the `opener` on `window.open` and testing that probably requires the harness to run tests in a top level browsing context. I started working on that, but it's a bigger change and has a few tricky parts.

tools/wptrunner/wptrunner/executors/runner.js
```diff
@@ -1,42 +1,44 @@
...
+ */
+window.addEventListener("message", function(event) {
+  if (!event.data) {
     return;
   }
-  var event = window.message_queue.shift();
-  if (!event) {
+
+  if (event.data.type === "testdriver-resume") {
+    source = event.source;
```
jgraham: I don't really understand the point of this; it seems like the identity of the test window is statically known here; it's `window.win`. Why do we need to keep updating it?

tools/wptrunner/wptrunner/executors/runner.js
```diff
@@ -53,7 +55,11 @@ window.process_next_event = function() {
   default:
     return;
   }
-  var callback = window.testdriver_callback;
-  window.testdriver_callback = null;
-  callback([window.url, data.type, payload]);
-};
+
+  source.postMessage({
+      type: "testdriver-next message",
```
jgraham: This type string doesn't match anything else.

tools/wptrunner/wptrunner/executors/testharness_webdriver_resume.js
```diff
@@ -1,3 +1,24 @@
 var callback = arguments[arguments.length - 1];
-window.opener.testdriver_callback = callback;
-window.opener.process_next_event();
+
+window.addEventListener("message", function handle(event) {
```
jgraham: I don't like the pattern where we keep adding and removing message listeners; we've had bags with such code before. Ideally we define the message handler once in `testharnessreport.js` and just have this script update the current callback and resume the test.

jgraham (approved):
I think what you're doing here is OK, but it's a larger than necessary change from the current design which makes it harder to reason about the correctness.

jgraham (changes requested):
Wrong button, sorry!

jugglinmike:
On Friday, we learned that in addition to tripping up Edge, gh-13419 interferes with the semantics of some tests (see gh-13876). That prompted @jgraham to begin work on an alternate solution which reduces the number of browsing contexts to one. [As James noted above](https://github.com/web-platform-tests/wpt/pull/13881#issuecomment-435948983), that approach will also enable contributors to test more parts of the platform.

We discussed simplifying this patch to use as a short-term fix. @jgraham shared some details about what they had in mind, but I'm having trouble interpreting them. What I have locally is no less complicated than this patch, and it's still not quite right, somehow.

I'm time-boxing my effort on this because we expect the change to be short-lived. That's why I'm proposing an even simpler change: gh-13938. That will not fix support for the invalidated tests, but it will correct the regression to Microsoft Edge. That may be preferable to reverting gh-13419 altogether.

jugglinmike:
Superseded by gh-13938.

PR #13886, opened 2018-11-03
Rewrite RTC ICE and DTLS transport tests with alternative dependencies (Rebased)

soareschen:
This is a rebased version of PR #9424.

- The PR is updated to make use of async/await.
- The code is simplified based on reviews in #9424.


webrtc/RTCDtlsTransport-getRemoteCertificates.html
```diff
@@ -4,104 +4,97 @@
...
+
+  function validateConnectedDtlsTransport(dtlsTransport) {
+    const certs = dtlsTransport.getRemoteCertificates();
+
+    assert_greater_than(certs.length, 0,
+      'Expect DTLS transport to have at least one remote certificate when connected');
+
+    for(const cert of certs) {
+      assert_true(cert instanceof ArrayBuffer,
+        'Expect certificate elements be instance of ArrayBuffer');
```
alvestrand: English: "Certificates should be represented as ArrayBuffers".

webrtc/RTCDtlsTransport-getRemoteCertificates.html
```diff
@@ -4,104 +4,97 @@
...
+    }
+  }
+
+  function validateConnectingDtlsTransport(dtlsTransport) {
+    if (dtlsTransport.state !== 'connected') {
+      assert_array_equals(dtlsTransport.getCertificates(), [],
+        'Expect DTLS certificates be initially empty until become connected');
```
alvestrand: English: "The DTLS certificate list should be empty before the transport is connected".

webrtc/RTCDtlsTransport-getRemoteCertificates.html
```diff
@@ -4,104 +4,97 @@
...
+      'Expect DTLS transport to have at least one remote certificate when connected');
+
+    for(const cert of certs) {
+      assert_true(cert instanceof ArrayBuffer,
+        'Expect certificate elements be instance of ArrayBuffer');
+    }
+  }
+
+  function validateConnectingDtlsTransport(dtlsTransport) {
+    if (dtlsTransport.state !== 'connected') {
```
alvestrand: This will still succeed if the state is already connected when you get here - so you're not necessarily checking what the test looks like it's checking.
Better: Assert that the state is not "connected"

soareschen: We can't make sure that the dtlsTransport does not become connected when it reach this part of the code. There is a race condition of the transport becoming connected in parallel with the async tests. Since the tests are running locally, there is a chance that the whole connection is connected before the code resume from `await doSignalingHandshake(...)`.

If we want the test to run consistently, the alternative is to remove `validateConnectingDtlsTransport()` altogether.

alvestrand: state changes will happen only while you await something. the only place where you await stuff before this is in doSignalingHandshake. But it's possible that the connection might connect while you're awaiting doSignalingHandshake. OK, let's do it this way.

alvestrand (approved):

webrtc/RTCDtlsTransport-getRemoteCertificates.html
```diff
@@ -4,104 +4,97 @@
...
-        transitions to "connected".
-   */
-  async_test(t => {
+  //  exchangeIceCandidates
+  //  doSignalingHandshake
+
+  // The following helper functions are called from RTCTransport-helper.js:
+  //  waitForConnectedState
+  //  getDtlsTransportFromSctpTransport
+  //  getDtlsTransportsFromSenderReceiver
```
youennf: Do we need those comments?
I am fearing that they will be obsolete when/if we start updating the tests and forget to update these comments.

webrtc/RTCDtlsTransport-getRemoteCertificates.html
```diff
@@ -4,104 +4,97 @@
...
+  }
+
+  promise_test(async t => {
     const pc1 = new RTCPeerConnection();
     t.add_cleanup(() => pc1.close());
+
+    const pc2 = new RTCPeerConnection();
+    t.add_cleanup(() => pc2.close());
+
+    const [track, mediaStream] = await getTrackFromUserMedia('audio')
```
youennf: missing ;

webrtc/RTCTransport-helper.js
```diff
@@ -0,0 +1,124 @@
...
+    doSignalingHandshake
+*/
+
+// Wait for all DTLS and ICE transport of a peer connection
+// to turn to connected or completed state. When the RTCPeerConnectionState
+// is "connected", all RTCIceTransports and RTCDtlsTransports are in the
+// "connected", "completed" or "closed" state and at least one of them is in the
+// "connected" or "completed" state.
+async function waitForConnectedState(pc) {
+  if (pc.connectionState === 'connected') return;
```
youennf: return in its own line?

webrtc/tools/.eslintrc.js
```diff
@@ -111,6 +114,7 @@ module.exports = {
     getUserMediaTracksAndStreams: true,
     performOffer: true,
     Resolver: true,
+    addTrackOrTransceiver: true,
```
youennf: Ah, I did not know about that file.
Should I have updated it when adding some helper functions in RTCPeerConnection-helpers.js?

youennf (approved):
Nice!
Some nits and one question about webrtc/tools/.eslintrc.js below.

webrtc/RTCDtlsTransport-getRemoteCertificates.html
soareschen: Since the helper functions are exposed as global variables, I thought it might be helpful for readers to find out where they come from. If people don't find it useful, we may remove them in new PRs.

webrtc/tools/.eslintrc.js
soareschen: The lint tool was intended as quick way to find out common errors such as undefined variable without being flooded by warnings on the global helpers. It has not been updated for some time so running lint now is not very useful anyway. We probably should either add all global variables to it, or remove the lint tools from the repo.

youennf: Either we should enforce running the lint tool for every PR or we should remove it.
Do you think we should do with one or the other?

sideshowbarker:
stability checker failing https://travis-ci.org/web-platform-tests/wpt/jobs/450733046

youennf (approved):

webrtc/RTCIceTransport.html
```diff
@@ -4,86 +4,43 @@
...
+  //  getIceTransportsFromDtlsTransports
+
+  function validateIceCandidates(candidates) {
     assert_greater_than(candidates.length, 0,
       'Expect at least one ICE candidate returned from get*Candidates()');
 
     for(const candidate of candidates) {
-      assert_true(candidate instanceof RTCIceCandidate,
-        'Expect candidate elements to be instance of RTCIceCandidate');
+     assert_true(candidate instanceof RTCIceCandidate,
```
alvestrand: Here and below, indentation seems to have turned odd. Please fix.

webrtc/RTCPeerConnection-helper.js
```diff
@@ -460,6 +460,37 @@ async function exchangeOfferAndListenToOntrack(t, caller, callee) {
   return ontrackPromise;
 }
 
+// Feature detect whether we can call add transceiver on the target browser.
+// We cannot simply test if the addTransceiver method exist, as Chrome
+// for example would have it implemented but only usable in unified plan,
+// which means it would fail the tests when using default peer connection config
+function canAddTransceiver() {
+  const pc = new RTCPeerConnection();
```
alvestrand: Please close this PeerConnection in cleanup.

alvestrand (changes requested):
Found a few more nits that need fixing. The PC leak is important.

soareschen:
All issues fixed. Is there any reason the CI keep failing with Chrome and Firefox Nightly? Any way we can get this merged?

webrtc/RTCPeerConnection-helper.js
```diff
@@ -460,6 +460,39 @@ async function exchangeOfferAndListenToOntrack(t, caller, callee) {
...
+  } finally {
+    pc.close()
+  }
+}
+
+// A dependency agnostic way to add a track track depending
+// on the supported method on a browser.
+//   - If addTransceiver(track) is usable, use that.
+//   - If addTrack(track, mediastream) is available, use that.
+function addTrackOrTransceiver(pc, track, mediaStream) {
```
youennf: Do we need this one?
It makes sense to support older tests for older browsers but for new tests/new constructs, we can probably hope to have transceivers available. Maybe we do want to test these new constructs with plan B as well, I guess this is fine.

Maybe we should rename it to addTransceiverOrTrack since we first try to add a transceiver and fall back to tracks otherwise.

webrtc/RTCPeerConnection-helper.js
```diff
@@ -460,6 +460,39 @@ async function exchangeOfferAndListenToOntrack(t, caller, callee) {
...
+  if (canAddTransceiver()) {
+    return pc.addTransceiver(track);
```
youennf: We should probably pass the mediaStream to addTransceiver so as to be consistent with addTrack

webrtc/RTCPeerConnection-helper.js
```diff
@@ -460,6 +460,39 @@ async function exchangeOfferAndListenToOntrack(t, caller, callee) {
...
+  } else if (typeof pc.addTrack === 'function') {
+    // Note: The mediaStream argument is optional in the spec, but we require
+    // it for now so that we can test older versions of Firefox that only
+    // implements addTrack with compulsory mediaStream argument.
+    return pc.addTrack(track, mediaStream);
+  } else {
+    throw new Error('This test requires either addTrack or addTransceiver implemented');
+  }
```
youennf: Maybe we can simplify and just call addTrack without check. It will anyway throw.

webrtc/RTCTransport-helper.js
```diff
@@ -0,0 +1,126 @@
...
+
+function getDtlsTransportFromSctpTransport(sctpTransport) {
+  assert_true(sctpTransport instanceof RTCSctpTransport,
+    'Expect pc.sctp to be instantiated from RTCSctpTransport');
+
+  const dtlsTransport = sctpTransport.transport;
+  assert_true(dtlsTransport instanceof RTCDtlsTransport,
+    'Expect sctp.transport to be an RTCDtlsTransport');
+
+  return dtlsTransport;
```
jan-ivar: I confess I'm personally not a fan of tests hidden in helpers and helpers that otherwise merely substitute JS for English—I prefer helpers that encapsulate whole sub-problems, usually with some output.

I find it much easier to read tests that call spec-documented APIs directly. Less to learn.

Others may disagree, but I'd rather see repetition in the tests themselves than too many helpers.

webrtc/RTCTransport-helper.js
```diff
@@ -0,0 +1,126 @@
...
+}
+
+function getIceTransportFromSctpTransport(sctpTransport) {
+  const dtlsTransport = getDtlsTransportFromSctpTransport(sctpTransport);
+
+  const { iceTransport } = dtlsTransport;
+  assert_true(iceTransport instanceof RTCIceTransport,
+    'Expect dtlsTransport.iceTransport to be an RTCIceTransport');
+
+  return iceTransport;
```
jan-ivar: ditto with this one.

webrtc/RTCTransport-helper.js
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// Get one or more RTCDtlsTransports from RTCRtpSender/Receiver
+// in pc.There may be more than one DTLSTransport if there are
+// multiple underlying transports, such as when not using max-bundle,
+// or when not using RTCP multiplexing.
+function getDtlsTransportsFromSenderReceiver(pc) {
```
jan-ivar: These helpers collectively abstract away the peer connection APIs we're testing. That may be the right approach for application development, but for writing API tests, I prefer repetition operating on the APIs we're actually testing.

This applies to the remaining helpers here as well.

My 2 cents.

jan-ivar (commented):
This may be subjective, but I'd prefer not organizing around helper functions the way this does.

jan-ivar:
These are tests meant to enforce that all browsers behave the same way. We want them readable like prose.

soareschen:
Added suggested changes. Just want to point out that the motivation for addTrackOrTransceiver came out when we were back in the "old" days when addTransceiver was not well supported in most browsers. At that time people were concerned of the tests on interfaces such as RTCRtpSender are failing because they are using addTransceiver, even though browsers have implemented alternative methods addTrack or addStream.

The tests here are neither testing addTrack nor addTransceiver. I prefer not have helpers like addTrackOrTransceiver just to make more tests passing. But I also prefer people not raising this or similar issues because different browsers have different methods implemented, and they all want the tests to pass.

Let's hope history never repeat again for issues similar to whether to use addStream / addTrack(track) / addTrack(track, mediaStream) / addTransceiver to make tests pass, and everyone can move on to just assume addTransceiver / addTrack are there.

webrtc/RTCTransport-helper.js
soareschen: I think it depends on how complicated it is to figure how to perform an operation correctly. There are many things in WebRTC that are not as straightforward as "I want to do X, therefore I'd just call method Y".

I have removed the helpers and make them inline. The inline transport extractors are simplified and not doing any thorough checking or assertions anymore. Hope this will be more readable, easier to verify correctness, and easy for copy/paste for new test cases.

webrtc/RTCIceTransport.html
```diff
@@ -114,81 +65,128 @@
...
+    await doSignalingHandshake(pc1, pc2);
+
+    const iceTransports = [
+      pc1.sctp.transport,
+      pc2.sctp.transport
+    ]
+    .filter(dtlsTransport => dtlsTransport) // filter null/undefined
+    .map(dtlsTransport => dtlsTransport.iceTransport);
+
+    assert_greater_than_equal(iceTransports.length, 2, 'expect to find at least 2 RTCIceTransports');
```
jan-ivar: Nit: Wouldn't it be fewer lines to just assert each of the two `pcN.sctp.transport.iceTransport`s?

webrtc/RTCIceTransport.html
```diff
@@ -114,81 +65,128 @@
...
-  }, 'Unconnected iceTransport should have empty remote candidates and selected pair');
+    // pc1.setRemoteDescription(answer) without pc2 setting answer to it's
+    // localDescription. This is because pc1.sctp is only set after setting
+    // answer description.
+    const offer = await pc1.createOffer()
+
+    await pc1.setLocalDescription(offer)
+    await pc2.setRemoteDescription(offer)
+    const answer = await pc2.createAnswer()
+    await pc1.setRemoteDescription(answer)
```
jan-ivar: Tip: could be inlined, for fewer names polluting the cognitive stack space:
```js
await pc1.setRemoteDescription(await pc2.createAnswer());
```
Also, missing `;` semicolons at EOLs here.

webrtc/RTCTransport-helper.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  return new Promise((resolve, reject) => {
+    pc.addEventListener('connectionstatechange', () => {
+      const { connectionState } = pc;
+
+      if (connectionState === 'connected') {
+        resolve();
+      } else if (['closed', 'failed'].includes(connectionState)) {
+        reject(new Error(`one of DTLS/ICE transports transition to unexpected state ${connectionState}`));
```
jan-ivar: I advocate [wrapping non-promise code as tightly as possible](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises#Creating_a_Promise_around_an_old_callback_API), which means avoiding any kind of application code inside promise constructors.

The WPT test harness adds a second reason: the awkward `t.step_func()` wrapper is needed to correlate spurious event handlers with a particular test, and it's missing here. Unfortunately, this would mean adding `t` as an additional argument to `waitForConnectedState`, which gets ugly.

Instead, I would recommend something like this:
```js
while (pc.connectionState != 'connected') {
  await new Promise(r => pc.addEventListener('connectionstatechange', r, {once: true}));
  if (['closed', 'failed'].includes(pc.connectionState)) {
    throw new Error(`DTLS/ICE transition to unexpected state ${pc.connectionState}`);
  }
}
```

jan-ivar (changes requested):
Thanks for addressing this, I like this direction! Some nits and comments, particularly around the tests around number of transports seems wrong.

alvestrand:
@soareschen will you update?

webrtc/RTCTransport-helper.js
soareschen: I agree with your proposition that inside the promise constructor the code should be as minimal as possible. However I do not agree with the proposed solution. Given that the timing for promise resolution is non-deterministic, you may miss an event firing if after the first event, a second event fires before the continuation following the await is executed. Instead of keeping the promise constructor absolute minimal, I think the better way to interpret is that code inside the promise constructor should be more carefully scrutinized, and having less code is a way for verifying the correctness.

I have added a try catch inside the callback to make sure the promise does not get unresolved forever due to exceptions. There are definitely better ways to do this, but given the many conflicting constraints we have for writing tests, I think that should be enough for this particular case.

webrtc/RTCDtlsTransport-getRemoteCertificates.html
```diff
@@ -4,104 +4,102 @@
...
+    .filter(dtlsTransport => dtlsTransport); // filter null/undefined
+
+    assert_greater_than_equal(dtlsTransports.length, 2, 'expect to find at least 2 RTCDtlsTransports');
+
+    for (const dtlsTransport of dtlsTransports) {
+      validateConnectingDtlsTransport(dtlsTransport);
+    }
+
+    await waitForConnectedState(pc1);
+    await waitForConnectedState(pc2);
```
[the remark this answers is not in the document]
soareschen: Changed it to use Promise.all(). I don't think it matter in this case, as the operations are just waiting for another concurrent operation that are already running.

Async functions always return a promise. So the caller should still resume asynchronously unless they are ignoring the promise.

> So while this is fine as is, I worry about people new to async/await copying this pattern. Tests often serve as examples. My 2 cents.

IMO most of the test code are pretty bad examples for writing good _application_ code, even though they may be considered good _test_ code. If we want WPT to serve as examples for using the spec APIs, they should probably written in a much different way.

webrtc/RTCDtlsTransport-getRemoteCertificates.html
```diff
@@ -4,104 +4,102 @@
...
+    const dtlsTransports = [
+      ...pc1.getSenders(),
+      ...pc1.getReceivers(),
+      ...pc2.getSenders(),
+      ...pc2.getReceivers()
+    ]
+    .map(senderReceiver => senderReceiver.transport)
+    .filter(dtlsTransport => dtlsTransport); // filter null/undefined
+
+    assert_greater_than_equal(dtlsTransports.length, 2, 'expect to find at least 2 RTCDtlsTransports');
```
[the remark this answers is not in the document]
soareschen: Yes the semantics kind of changed a bit, as I tried to avoid copy-pasting too much code when inlining. Though the original test don't really test that the transport objects in different `pc`s are different though, and validating that is a bit complicated and redundant.

I have simplified the code to just skip all the checkings. They just assume the browser is spec compliant and create a transport object after setting local description. The consequence is mainly error messages in tests is more obscure, for trying to access fields of a `null` or `undefined` object for current implementations.

webrtc/RTCDtlsTransport-getRemoteCertificates.html
```diff
@@ -4,104 +4,102 @@
...
+    await doSignalingHandshake(pc1, pc2);
+
+    const dtlsTransports = [
+      ...pc1.getSenders(),
+      ...pc1.getReceivers(),
+      ...pc2.getSenders(),
+      ...pc2.getReceivers()
+    ]
+    .map(senderReceiver => senderReceiver.transport)
+    .filter(dtlsTransport => dtlsTransport); // filter null/undefined
```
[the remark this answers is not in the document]
soareschen: I probably misremembered or referred to older spec where some transport may be null and still spec-compliant. Removed that assumption now.

foolip:

alvestrand:
There's still a lot of dependency on sctp.transport.
Given the pattern of implementation, it seems better to get the transport via senders/receivers.

alvestrand:
Closing as "likely obsolete".

PR #13887, opened 2018-11-03
Basic test for generated SDP

alvestrand:
This adds a protocol test that SDP is generated by createOffer
conforming to JSEP requirements.

### DIFF
--- a/webrtc/protocol/jsep-initial-offer.https.html
+++ b/webrtc/protocol/jsep-initial-offer.https.html
@@ -21,7 +21,12 @@
     // to detect if anyone ever uses something different.
     assert_regexp_match(offer_lines[1], /^o=- \d+ \d+ IN IP4 127.0.0.1$/);
     const fields = RegExp(/^o=- (\d+) (\d+)/).exec(offer_lines[1]);
-    assert_less_than(Number(fields[1]), 2**62);
+    // Per RFC 3264, the sess-id should be representable in an uint64
+    // Note: JSEP -24 has this wrong - see bug:
+    // https://github.com/rtcweb-wg/jsep/issues/855
+    assert_less_than(Number(fields[1]), 2**64);
+    // Per RFC 3264, the version should be less than 2^62 to avoid overflow
+    assert_less_than(Number(fields[2]), 2**62);
     // Note: using - in s=- is a SHOULD in JSEP, not a MUST.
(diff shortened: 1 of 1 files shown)


webrtc/protocol/jsep-initial-offer.https.html
```diff
@@ -0,0 +1,30 @@
...
+    const offer = await pc.createOffer({offerToReceiveVideo: true});
+    let offer_lines = offer.sdp.split('\r\n');
+    // The first 3 lines are dictated by JSEP.
+    assert_equals(offer_lines[0], "v=0");
+    assert_equals(offer_lines[1].slice(0, 2), "o=");
+    // JSEP says that the address part SHOULD be a meaningless address
+    // "such as" IN IP4 127.0.0.1. We do strict matching here in order
+    // to detect if anyone ever uses something different.
+    assert_regexp_match(offer_lines[1], /^o=- \d+ \d+ IN IP4 127.0.0.1$/);
+    const fields = RegExp(/^o=- (\d+) (\d+)/).exec(offer_lines[1]);
```
soareschen: This fails on Firefox as it has the username field value "mozilla...THIS_IS_SDPARTA". JSEP says "`The value of the <username> field SHOULD be "-"`". Since it is optional it should not result in a test failure.

webrtc/protocol/jsep-initial-offer.https.html
```diff
@@ -0,0 +1,30 @@
...
+    assert_less_than(Number(fields[1]), 2**62);
+    // Note: using - in s=- is a SHOULD in JSEP, not a MUST.
```
soareschen: Should being not complying to "SHOULD" requirements in JSEP result in test failure?

soareschen (changes requested):

webrtc/protocol/jsep-initial-offer.https.html
alvestrand: It seems that we should (SHOULD) do that....

webrtc/protocol/jsep-initial-offer.https.html
alvestrand: @nils-ohlmeier says that Firefox expects to change this.
SHOULD is a problem for tests; if we don't test them, will they ever be tested?

soareschen (approved):

PR #13889, opened 2018-11-03
Move identity tests to separate directory

alvestrand:
I am not sure whether these tests are executing in the same way after being moved. Authors need to verify.


soareschen (approved):
We should also add a META.yml to the new directory.

martinthomson (approved):

PR #13893, opened 2018-11-04
 Rename generateOffer to generateDataChannelOffer and remove use of legacy optionsRename generateOffer to generateDataChannelOffer and remove use of legacy options

youennf:


alvestrand (approved):
Good catch that generateOffer() is only used to generate data channel offers!
Short helpers are good helpers.

webrtc/idlharness.https.window.js
```diff
@@ -5,7 +5,7 @@
 'use strict';
 
 // The following helper functions are called from RTCPeerConnection-helper.js:
-//   generateAnswer()
+//   generateDataChannelAnswer()
```
soareschen: `generateAnswer()` is not renamed to `generateDataChannelAnswer()`.

soareschen (changes requested):

webrtc/idlharness.https.window.js
youennf: Fixed in above commit

soareschen (approved):

youennf:
Firefox bot issue is probably unrelated, merging.

PR #13894, opened 2018-11-04
Use transceiver instead of offerToReceiveVideo in no-media-call

youennf:


alvestrand (approved):

youennf:
Actually some changes to use more generateVideoReceiveOnlyOffer are actually making some tests to fail. Will push PR to fix these tonight.

PR #13902, opened 2018-11-04
An incoming offer can generate simulcast

alvestrand:
This test needs discussion - my reading of draft-ietf-mmusic-simulcast is that an incoming offer with two RIDs should cause a transceiver to be created with two encodings avaialble, and an answer with two rid lines with send direction, indicating that the answerer is capable of sending 2 layers of simulcast.

This currently fails on both Chrome and Firefox.


alvestrand:
Actually JSEP says:
```
"  In addition, JSEP does not provide a mechanism to handle an incoming
   offer requesting simulcast from the JSEP endpoint.  This means that
   setting up simulcast in the case where the JSEP endpoint receives the
   initial offer requires out-of-band signaling or SDP inspection.
   However, in the case where the JSEP endpoint sets up simulcast in its
   in initial offer, any established simulcast streams will continue to
   work upon receipt of an incoming re-offer.  Future versions of this
   specification may add additional APIs to handle the incoming initial
   offer scenario.
"
```
The SDP in the test would be reasonable if it was supposed to work, but apparently it's not supposed to.

nils-ohlmeier:
In Firefox this should work, but you need to set the Encodings before setting the Offer.

fippo:
You might want to call getUserMedia, otherwise the m-line will be inactive...

http://jsfiddle.net/xcmq4vbn/1/ works in Firefox and has less boilerplate than the Firefox test. Calling setParameters works after setting the offer.

webrtc/protocol/simulcast-negotiate.html
```diff
@@ -0,0 +1,50 @@
...
+a=simulcast:recv 1;2
+a=rid:1 recv max-fps=60
+a=rid:2 recv max-fps=30
+`;
+      const pc = new RTCPeerConnection();
+      await pc.setRemoteDescription({type: 'offer', sdp: offer_sdp});
+      const answer = await pc.createAnswer();
+      const rid_lines = answer.sdp.split('\r\n')
+            .filter(line => line.match('^a=rid'));
+      assert_equals(rid_lines.length, 2, 'Answer shold have 2 RIDs');
```
youennf: s/shold/should/

webrtc/protocol/simulcast-negotiate.html
```diff
@@ -0,0 +1,50 @@
+<!doctype html>
+<meta charset=utf-8>
+<title>Negotiate Simulcast</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="../RTCPeerConnection-helper.js"></script>
```
youennf: Is it used?

webrtc/protocol/simulcast-negotiate.html
```diff
@@ -0,0 +1,50 @@
...
+<script>
+'use strict';
+
+
+
```
youennf: Two unneeded empty lines.

youennf (commented):
> In Firefox this should work, but you need to set the Encodings before setting the Offer.
> 
> Test code https://searchfox.org/mozilla-central/source/dom/media/tests/mochitest/test_peerConnection_simulcastAnswer.html#59

In such a case, do you need to have simulcast enabled in the offer?

nils-ohlmeier:
> In such a case, do you need to have simulcast enabled in the offer?

I'm not sure I get what you mean. Yes the offer needs to contain a=simulcast and a=rid attributes to be able to turn on simulcast in the Answer.

aboba:
> Yes the offer needs to contain a=simulcast and a=rid attributes to be able to turn on simulcast

[BA] Yes, but other than the number of encodings and the rids are there expectations that SDP attributes in the test's Offer (max-fr) will be translated into encoding parameters on the Answering side when the transceiver is evented?

webrtc/protocol/simulcast-negotiate.html
aboba: Are we expecting the max-fr attributes to translate into maxFramerate values within the encoding parameters?

alvestrand: If they have any effect at all, that's the effect that they should have.

webrtc/protocol/simulcast-negotiate.html
alvestrand: Not yet.

alvestrand:
The number of encodings is really all I was looking for with this test. Nils, you're saying that you can do
<pre>
pc.setRemoteDescription(offer);
pc.getSenders().setParameters(.... some simulcast stuff ...);
</pre>
and get simulcast, right?
According to spec, this should be pc.getSenders().setParameters(modifyParameters(pc.getSenders().getParameters())), but mozilla doesn't check this - Chrome does.
Do you have getParameters() at all at the moment?

alvestrand:
(I see that getParameters on Firefox ESR returns a rather empty thing. Has it improved?)

youennf:
> > In such a case, do you need to have simulcast enabled in the offer?
> 
> I'm not sure I get what you mean. Yes the offer needs to contain a=simulcast and a=rid attributes to be able to turn on simulcast in the Answer.

I was wondering whether just setParameters in Firefox would trigger simulcast (not as per spec but hey...). From your answer, it seems not.

My main question is to understand whether this is nice to have or required to support SDP simulcast negotiation, especially with the trend to move out of SDP negotiation.

It seems a SFU needs to send some simulcast parameters to the web page so that setParameters is called.
The SFU could instead send the number of participants to let the web page create an offer with a sendonly simulcast section and the necessary recvonly sections.
It would then receive an offer that matches what it wants. This is racy but no more than the alternative.
What are we loosing in doing that?

fippo:
> The SFU could instead send the number of participants to let the web page create an offer with a sendonly simulcast section and the necessary recvonly sections.

This directly affects things like ICE and DTLS roles and also changes behaviour in weird ways, see e.g. [here](https://bugs.chromium.org/p/webrtc/issues/detail?id=9931) for an example.

alvestrand:
Status: Needs a spec update to make sense.

foolip:

alvestrand:
I think the spec has caught up to this one now.
@youennf can you take another look?

alvestrand:
Closing in favor of a PR submitted via the chromium pipeline.

alvestrand:
Chromium review: https://chromium-review.googlesource.com/c/chromium/src/+/1540306

foolip:
Back on July 2 I got a reply from GitHub support about this:

> We were wondering if you can try removing 14 labels from the PR and giving the close another go? There's a maximum label limit our engineering team think is causing the PR not to close.
>
> That limitation is a recent addition, which means it won't be possible to apply that many labels to a PR going forward. It seems this change may have some unintended side effects for older PRs however.

I'll try that now.

PR #13903, opened 2018-11-04
Move RTCRtpTransceiver tests related to OfferToReceive legacy options to webrtc/legacy

youennf:


alvestrand (approved):
I don't love the shape of these tests, but it doesn't seem unreasonable to segregate these.

youennf:
> I don't love the shape of these tests, but it doesn't seem unreasonable to segregate these.

Agreed, and I did not want to touch these in this patch :)
Some of these tests might be redundant with other webrtc/legacy tests.
If we want to do some clean-up there, it would best be done as a follow-up.

PR #13908, opened 2018-11-04
Fix regressions introduced in https://github.com/web-platform-tests/wpt/pull/13893

youennf:


alvestrand (approved):
As your reviewer, I should have seen that one :-(

youennf:
I do not know how I missed it in the first place, I tried running the modified tests in the 3 browsers I have installed before uploading the PRs.
Let's go then.

PR #13909, opened 2018-11-04
Move oneremovetrack tests to webrtc/legacy

youennf:
Note that these tests are different than other webrtc/legacy tests since onremovetrack is unspecified AFAIK.


alvestrand (changes requested):
I'm afraid I'll have to disagree with this move. Step 2.2.12 of the "set an RTCSessionDescription" in section 4.4.1.6 is not marked as legacy.

youennf:
> I'm afraid I'll have to disagree with this move. Step 2.2.12 of the "set an RTCSessionDescription" in section 4.4.1.6 is not marked as legacy.

Ah, good point, let's close it for now. I'll file a bug in webkit

PR #13911, opened 2018-11-04
[Azure Pipelines] Run Chrome + Firefox infrastructure/ on macOS

foolip:


foolip:
@gsnedders maybe we'll want to make the Windows setup look similar, testing both Chrome and Firefox? I have some thoughts on how to generalize this using job/step templates, which we can have a look at when you do the Windows CI stuff.

jgraham (approved):

PR #13920, opened 2018-11-05
XMLHttpRequest: manipulate Content-Type a little more

annevk:
99f4fdcd906411c036e96c055e28b3e6ca84a1cc was not entirely correct unfortunately.

### DIFF
--- a/xhr/send-content-type-charset.htm
+++ b/xhr/send-content-type-charset.htm
@@ -89,7 +89,11 @@
       request(
         "text/plain;charset= utf-8",
         "text/plain;charset=UTF-8",
-        "charset with space that is UTF-8 does change")
+        "charset with leading space that is UTF-8 does change")
+      request(
+        "text/plain;charset=utf-8 ;x=x",
+        "text/plain;charset=utf-8 ;x=x",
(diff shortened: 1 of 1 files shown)


wisniewskit (commented):
It's probably worth having two tests, one for leading and one for trailing space (since leading spaces aren't stripped by MimeSniff.serial, but trailing ones are, so charsets ending with spaces will be normalized, but ones beginning with spaces will not).

annevk:
Fair, pushed a commit.

wisniewskit (changes requested):
Heh, sorry, this is the inverse of what I meant :)

>"charset= utf-8" --> will not change because leading spaces are not stripped by serialize.
>"charset=utf-8 " --> will become "charset=UTF-8", because trailing spaces *are* stripped.

annevk:
If it is an ASCII case-insensitive match for UTF-8, we won't change the value. So if the spaces are stripped, it'll be a match, meaning the value won't change, right?

wisniewskit (approved):
Ah, right. I was getting myself mixed up (busy day). This should be fine.

PR #13940, opened 2018-11-06
Update RTCPeerConnection-setRemoteDescription-tracks.https.html as MediaStream has no constraint on the order of the tracks

youennf:


henbos (approved):
The order that tracks are added to the stream is well-defined, but I see that getTracks() isn't:
"The conversion from the track set to the sequence is User Agent defined and the order does not have to be stable between calls."
LGTM

PR #13943, opened 2018-11-06
[svg] Parsing of the pointer-events CSS property

ewilligers:

PR #13951, opened 2018-11-06
SVG Geometry Properties - parsing

ewilligers:
The properties cx cy r rx ry x y have the specified initial values,
grammar and inheritance behavior.
https://svgwg.org/svg2-draft/geometry.html

### DIFF
--- a/svg/geometry/inheritance.svg
+++ b/svg/geometry/inheritance.svg
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg"
+     xmlns:h="http://www.w3.org/1999/xhtml"
+     width="800px" height="8000px">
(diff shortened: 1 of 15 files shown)


ewilligers:
Builds on https://github.com/web-platform-tests/wpt/pull/13943

dirkschulze (approved):
Tests look good. Could we get additional tests for the computed value? `auto` computes to different things in different properties. Same for geometry properties with percentage values on paint servers.

ewilligers:
I will create a separate PR for computed values.

foolip:
@ewilligers, the stability checks on Travis create rather a lot of trouble and I'd like to disable them sooner rather than later, see https://github.com/web-platform-tests/wpt/pull/14033.

I don't know if it's expected that there's no output at all while this script is running, or if "travis_wait 30" would just lead to it failing in the same way after 30 minutes.

For now please just ignore the check, I'll override and admin merge this.

PR #13958, opened 2018-11-06
The computed value of float with absolute positioning when there is no box should be "none"

FremyCompany:
https://github.com/w3c/csswg-drafts/issues/1436

### DIFF
--- a/css/CSS2/floats/computed-float-position-absolute.html
+++ b/css/CSS2/floats/computed-float-position-absolute.html
@@ -17,14 +17,12 @@ div[id] {
 <div style="display: none"><div id="test3"></div></div>
 
 <script>
-var test_description = document.title;
 promise_test(
 	t => {
 		return new Promise(test => addEventListener('load', e=>test()))
 		.then(test => assert_equals(getComputedStyle(test1)['float'], "none", "[display:none] Invalid gCS(test1)['float'];"))
 		.then(test => assert_equals(getComputedStyle(test2)['float'], "none", "[display:contents] Invalid gCS(test2)['float'];"))
 		.then(test => assert_equals(getComputedStyle(test3)['float'], "none", "[in a display:none] Invalid gCS(test3)['float'];"))
-	},
-	test_description
+	}
 );
 </script>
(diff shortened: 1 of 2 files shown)


css/CSS2/floats/computed-float-position-absolute.html
```diff
@@ -0,0 +1,30 @@
...
+
+<div id="test1" style="display: none"></div>
+<div id="test2" style="display: contents"></div>
+<div style="display: none"><div id="test3"></div></div>
+
+<script>
+var test_description = document.title;
+promise_test(
+	t => {
+		return new Promise(test => addEventListener('load', e=>test()))
```
gsnedders: There's no need for the test to wait for the load event; this can be a normal test().

css/CSS2/floats/computed-float-position-absolute.html
```diff
@@ -0,0 +1,30 @@
...
+}
+
+</style>
+
+<div id="test1" style="display: none"></div>
+<div id="test2" style="display: contents"></div>
+<div style="display: none"><div id="test3"></div></div>
+
+<script>
+var test_description = document.title;
```
gsnedders: No need for this, it defaults to `document.title` unless I'm sorely mistaken.

gsnedders (commented):
(minor nits only)

css/CSS2/floats/computed-float-position-absolute.html
FremyCompany: I guess, but the test is autogenerated, and it is a generally impossible to resolve problem to decide that there won't be dependencies like font loading that are required in the general case; is there a problem if the test loads for the load event? I wouldn't assume this to change test execution time given there isn't anything load will be waiting on in this case.

css/CSS2/floats/computed-float-position-absolute.html
FremyCompany: Oh, didn't know that. 
In such case, yeah, could be removed.

gsnedders (approved):

PR #13966, opened 2018-11-07
 Make testharness tests run in a top-level browsing context

### LINKED ISSUE #13418: Avoid window.open in the test runner
From https://github.com/web-platform-tests/wpt/pull/12885#issuecomment-419240394 (@domenic, @TimothyGu).

See also discussion in https://github.com/web-platform-tests/wpt/issues/10398#issuecomment-427525644.

Note this is complicated for executors using WebDriver by the lack of any window management (beyond `close`) in WebDriver; see https://github.com/w3c/webdriver/issues/1138.

jgraham:
fixes #13418

---
This change is [Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/13966)


tools/wptrunner/wptrunner/executors/executormarionette.py
```diff
@@ -673,29 +679,18 @@ def do_test(self, test):
...
-        script = self.script % format_map
-
-        protocol.base.execute_script(script, async=True)
-        test_window = protocol.testharness.get_test_window(self.window_id, parent_window)
+        format_map = {"url": strip_server(url)}
 
+        protocol.base.execute_script("window.open('about:blank', '%s', 'noopener')" % self.window_id)
+        test_window = protocol.testharness.get_test_window(self.window_id, parent_window,
+                                                           timeout=10*self.timeout_multiplier)
+        self.protocol.base.set_window(test_window)
```
gsnedders: This seems a bit needlessly complex now we're opening a new window. Can we not: navigate to about:blank (to avoid the window potentially opening anything else), close everything open, then navigate that (with `<a rel=noopener>` probably, rather than opening another new window?) to the test?

tools/wptrunner/wptrunner/executors/executorwebdriver.py
```diff
@@ -316,27 +321,49 @@ def do_test(self, test):
...
-                      "timeout_multiplier": self.timeout_multiplier,
-                      "timeout": timeout * 1000}
+        format_map = {"url": strip_server(url)}
 
         parent_window = protocol.testharness.close_old_windows()
         # Now start the test harness
-        protocol.base.execute_script(self.script % format_map, async=True)
-        test_window = protocol.testharness.get_test_window(self.window_id, parent_window)
-
+        protocol.base.execute_script("window.open('about:blank', '%s', 'noopener')" % self.window_id)
```
gsnedders: Note that Edge will shortly be using this and doesn't support `noopener`.

tools/wptrunner/wptrunner/products.py
```diff
@@ -26,28 +28,39 @@ def product_module(config, product):
...
-    executor_classes = {}
-    for test_type, cls_name in data["executor"].iteritems():
-        cls = getattr(module, cls_name)
-        executor_classes[test_type] = cls
-
-    return (check_args,
-            browser_cls, browser_kwargs,
-            executor_classes, executor_kwargs,
-            env_options, env_extras, run_info_extras)
+class Product(object):
```
gsnedders: Oh, I was planning on doing this!

gsnedders (commented):
By no means a complete review, but from a quick look through.

lukebjerring:
@jgraham 
Started the scrapes, you'll eventually see them @ https://wpt.fyi/test-runs?label=testdriver-message-passing

lukebjerring:
Whoops - @jgraham I didn't realise affected tests would be like, 19 css tests.
@jugglinmike how did you trigger a complete run for https://github.com/web-platform-tests/wpt/pull/13881 ?

jugglinmike:
@lukebjerring I pushed [a separate branch to Bocoup's fork](https://github.com/web-platform-tests/wpt/compare/master...bocoup:testdriver-message-passing-trial) which included [a commit to enable results
collection](https://github.com/web-platform-tests/wpt/commit/9057744610084d6358dd4b23340e1230da87cd44).

lukebjerring:
Err, my bad on pushing 3d0c723, reverted.

lukebjerring:
Diff URLs:
- https://staging.wpt.fyi/?products=chrome[master],chrome@79d09131ba&diff
- https://staging.wpt.fyi/?products=firefox[master],firefox@79d09131ba&diff

tools/wptrunner/wptrunner/executors/executorwebdriver.py
jgraham: I"m not super worried about this; any test that depends on the opener being null at the moment is already broken. I only used noopener to prove that we aren't depending on the opener any more,

In the medium term I hope that everyone will implement the `Create Window` webdriver endpoint, which is inteded to have exactly the behaviour we require here and also returns the handle of the new window so we can delete the window finding code. I'm happy to do the more complex solution in the interim if people think that it's important, but otherwise I would tend to argue that this good enough for now.

tools/wptrunner/wptrunner/executors/executormarionette.py
jgraham: That seems far more complex? What's the benefit?

tools/wptrunner/wptrunner/executors/executorwebdriver.py
```diff
@@ -316,27 +321,49 @@ def do_test(self, test):
...
+  callback(false);
+} else if (document.readyState !== "loading") {
+  callback(true);
+} else {
+  document.addEventListener("DOMContentLoaded", () => callback(true));
+}""", async=True)
+            except client.JavascriptErrorException:
+                if seen_error:
+                    raise
+                seen_error = True
```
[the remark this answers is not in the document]
jgraham: So! There's a race condition. Sometimes this script is run in the initial `about:blank` document as it's undergoing navigation. In that case you get a "Document was unloaded" error. Maybe there should be a comment to that effect.

jgraham:
The diffs are a little hard to read, but after ensuring that the two Chrome versions match, the changes I can see look mostly like flaky tests rather than anything concerning. The tests that were broken by the previous change are indeed fixed: https://staging.wpt.fyi/results/html/browsers/the-windowproxy-exotic-object?product=firefox%5Bmaster%5D&product=firefox%4079d09131ba&diff

tools/wptrunner/wptrunner/executors/executorwebdriver.py
gsnedders: I don't think we need to worry about older versions of Safari; it's almost exclusively Edge that's my concern here.

tools/wptrunner/wptrunner/executors/executormarionette.py
gsnedders: It avoids guessing what window we need to keep open. Which is mostly a hypothetical concern it turns out, admittedly.

tools/wptrunner/wptrunner/executors/executorwebdriver.py
gsnedders: let's try explicitly setting window.open().opener = null, because I don't think that can make anything worse?

tools/wptrunner/wptrunner/executors/executorwebdriver.py
```diff
@@ -316,27 +321,52 @@ def do_test(self, test):
...
+        while not loaded:
+            try:
+                loaded = protocol.base.execute_script("""
+var callback = arguments[arguments.length - 1];
+if (location.href === "about:blank") {
+  callback(false);
+} else if (document.readyState !== "loading") {
+  callback(true);
+} else {
+  document.addEventListener("DOMContentLoaded", () => callback(true));
```
gsnedders: Let's make this `readyStateChange` being `ready` to match WebDriver.

tools/wptrunner/wptrunner/executors/executorselenium.py
```diff
@@ -102,29 +103,34 @@ def close_old_windows(self):
         self.webdriver.switch_to_window(self.runner_handle)
         return self.runner_handle
 
-    def get_test_window(self, window_id, parent):
+    def get_test_window(self, window_id, parent, timeout=5):
```
gsnedders: Can you add a comment for all the executors which now have a timeout here explaining *why*?

tools/wptrunner/wptrunner/executors/executorwebdriver.py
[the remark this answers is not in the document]
jugglinmike: Got it. Thanks

jgraham:
This seems to fix `/html/dom/usvstring-reflection.html` in Chrome.

tools/wptrunner/wptrunner/executors/executorwebdriver.py
gsnedders: @jgraham https://gist.github.com/fb3941b207c8d9ff3f3343a7bf302165 (this is with Selenium)

gsnedders: @jgraham https://gist.github.com/c939fd3a4276f3bc056e175e483e57a5 (edge_webdriver)

jugglinmike:
[It also fixes `FileAPI/url/sandboxed-iframe.html`.](https://github.com/web-platform-tests/wpt/pull/14132)

gsnedders:
@jgraham so what's the status of this now?

jgraham:
Works on Firefox and Chrome on Linux and macOS. Status for Edge and Safari is unknown.

foolip:
Safari looks happy on Azure Pipelines. If you trigger Azure Pipelines again even more stuff will run.

gsnedders:
In Edge almost all (all?) the testharness.js tests now give 

```
JavascriptErrorException: javascript error (500): Object doesn't support property or method '__wptrunner_process_next_event'
```

gsnedders (commented):
> Reviewed 27 of 27 files at r1, 3 of 3 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/13966)* status: all files reviewed, 5 unresolved discussions (waiting on @jgraham, @frivoal, @annevk, @zqzhang, @zcorpan, @dbaron, @jdm, @foolip, and @domenic)

___
*[tools/wptrunner/wptrunner/products.py, line 34 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/13966#-LTOQD6a44YuHGTqUm6l:-LTOQD6a44YuHGTqUm6m:b-ia8yas) ([raw file](https://github.com/web-platform-tests/wpt/blob/ec633de484ac2690150d9a728234a5058b158940/tools/wptrunner/wptrunner/products.py#L34)):*
> ```Python
>     def __init__(self, config, product):
>         module = product_module(config, product)
>         data = module.__wptrunner__
> ```

What would it take to have a function (or just a variable) in that module that returned a Product?
___
*[tools/wptrunner/wptrunner/products.py, line 52 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/13966#-LTOQYfO8Lny5A7Egb8k:-LTOQYfO8Lny5A7Egb8l:br1w2cl) ([raw file](https://github.com/web-platform-tests/wpt/blob/ec633de484ac2690150d9a728234a5058b158940/tools/wptrunner/wptrunner/products.py#L52)):*
> ```Python
> 
> 
> def load_product(config, product, load_cls=False):
> ```

What's the point of the `load_cls` arg here? When does it make sense to call this with it being `True`; why wouldn't you just construct the class yourself?
___
*[tools/wptrunner/wptrunner/browsers/firefox.py, line 103 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/13966#-LTOt7h78xEFSpCAlMJf:-LTOt7h78xEFSpCAlMJg:bfzwi8v) ([raw file](https://github.com/web-platform-tests/wpt/blob/ec633de484ac2690150d9a728234a5058b158940/tools/wptrunner/wptrunner/browsers/firefox.py#L103)):*
> ```Python
>     capabilities = {}
>     if test_type == "testharness":
>         capabilities["pageLoadStrategy"] = "eager"
> ```

Can we not make this default to eager, somehow? The fact we have all of this is to deal with ChromeDriver not supporting it, right?
___
*[tools/wptrunner/wptrunner/executors/base.py, line 525 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/13966#-LTPB7TyE36ac1LNj5qa:-LTPB7TyE36ac1LNj5qb:b-g1n2xn) ([raw file](https://github.com/web-platform-tests/wpt/blob/ec633de484ac2690150d9a728234a5058b158940/tools/wptrunner/wptrunner/executors/base.py#L525)):*
> ```Python
> 
>     def process_complete(self, url, payload):
>         rv = [strip_server(url)] + payload
> ```

Is this an earlier underlying bug? What caught this?
___
*[tools/wptrunner/wptrunner/executors/executormarionette.py, line 221 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/13966#-LTPBmfYCYoweuXDNkwB:-LTPBmfYCYoweuXDNkwC:b-59lgj2) ([raw file](https://github.com/web-platform-tests/wpt/blob/ec633de484ac2690150d9a728234a5058b158940/tools/wptrunner/wptrunner/executors/executormarionette.py#L221)):*
> ```Python
>                 return test_window
> 
>             time.sleep(0.1)
> ```

This is addressing the fact that closing windows is async, right?
___
*[tools/wptserve/wptserve/stash.py, line 175 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/13966#-LTPGY8I8qw1BYgT9Jtb:-LTPGY8I8qw1BYgT9Jtc:b-3h8715) ([raw file](https://github.com/web-platform-tests/wpt/blob/ec633de484ac2690150d9a728234a5058b158940/tools/wptserve/wptserve/stash.py#L175)):*
> ```Python
> 
>         return value
> ```

Given you're adding loads of whitespace, can you enable whichever flake8 check checks for them, or does that still have lots of failures?

jgraham (commented):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/13966)* status: all files reviewed, 5 unresolved discussions (waiting on @gsnedders, @frivoal, @annevk, @zqzhang, @zcorpan, @dbaron, @jdm, @foolip, and @domenic)

___
*[tools/wptrunner/wptrunner/products.py, line 34 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/13966#-LTOQD6a44YuHGTqUm6l:-LTT6cR19Ybl3IZQ1VPl:b-2sbof5) ([raw file](https://github.com/web-platform-tests/wpt/blob/ec633de484ac2690150d9a728234a5058b158940/tools/wptrunner/wptrunner/products.py#L34)):*

What would it take to have a function (or just a variable) in that module that returned a Product?
</blockquote></details>

A new PR to implement that.
___
*[tools/wptrunner/wptrunner/products.py, line 52 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/13966#-LTOQYfO8Lny5A7Egb8k:-LTT6eW77EjtIxDHyoD6:bsansit) ([raw file](https://github.com/web-platform-tests/wpt/blob/ec633de484ac2690150d9a728234a5058b158940/tools/wptrunner/wptrunner/products.py#L52)):*

What's the point of the `load_cls` arg here? When does it make sense to call this with it being `True`; why wouldn't you just construct the class yourself?
</blockquote></details>

Fair point. Should that block this PR?
___
*[tools/wptrunner/wptrunner/browsers/firefox.py, line 103 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/13966#-LTOt7h78xEFSpCAlMJf:-LTT70rv01clvPqQhl_9:b-v2d049) ([raw file](https://github.com/web-platform-tests/wpt/blob/ec633de484ac2690150d9a728234a5058b158940/tools/wptrunner/wptrunner/browsers/firefox.py#L103)):*

Can we not make this default to eager, somehow? The fact we have all of this is to deal with ChromeDriver not supporting it, right?
</blockquote></details>

Chromedriver doesn't support it and it's not clear that it's right for other test types. We could set this in the executor, but you'd still need to set it once per executor.
___
*[tools/wptrunner/wptrunner/executors/base.py, line 525 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/13966#-LTPB7TyE36ac1LNj5qa:-LTT7RDY4SD18PYYej9d:b-ibncmd) ([raw file](https://github.com/web-platform-tests/wpt/blob/ec633de484ac2690150d9a728234a5058b158940/tools/wptrunner/wptrunner/executors/base.py#L525)):*

Is this an earlier underlying bug? What caught this?
</blockquote></details>

I think this is a change because we aren't returning the same URL as before.
___
*[tools/wptrunner/wptrunner/executors/executormarionette.py, line 221 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/13966#-LTPBmfYCYoweuXDNkwB:-LTT7g3N8k5DI_kzYOwZ:b-a0xx9e) ([raw file](https://github.com/web-platform-tests/wpt/blob/ec633de484ac2690150d9a728234a5058b158940/tools/wptrunner/wptrunner/executors/executormarionette.py#L221)):*

This is addressing the fact that closing windows is async, right?
</blockquote></details>

s/closing/opening/ and yes.
___
*[tools/wptserve/wptserve/stash.py, line 175 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/13966#-LTPGY8I8qw1BYgT9Jtb:-LTT7rZO-Ah4keLtalTJ:b9mxyt) ([raw file](https://github.com/web-platform-tests/wpt/blob/ec633de484ac2690150d9a728234a5058b158940/tools/wptserve/wptserve/stash.py#L175)):*

Given you're adding loads of whitespace, can you enable whichever flake8 check checks for them, or does that still have lots of failures?
</blockquote></details>

I didn't check but I would expect lots of failures. We should do that as a seperate PR.

gsnedders (approved):
> *[Reviewable](https://reviewable.io/reviews/web-platform-tests/wpt/13966)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @frivoal, @annevk, @zqzhang, @zcorpan, @dbaron, @jdm, @foolip, and @domenic)

___
*[tools/wptrunner/wptrunner/products.py, line 34 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/13966#-LTOQD6a44YuHGTqUm6l:-LTTEbGsBb5Td7Vmwf_8:bnjvwe8) ([raw file](https://github.com/web-platform-tests/wpt/blob/ec633de484ac2690150d9a728234a5058b158940/tools/wptrunner/wptrunner/products.py#L34)):*

A new PR to implement that.
</blockquote></details>

Given you've done half the work here, it's probably worth actually prioritising this now…
___
*[tools/wptrunner/wptrunner/products.py, line 52 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/13966#-LTOQYfO8Lny5A7Egb8k:-LTTEkZM2osmsMYwnYiD:b-46vg46) ([raw file](https://github.com/web-platform-tests/wpt/blob/ec633de484ac2690150d9a728234a5058b158940/tools/wptrunner/wptrunner/products.py#L52)):*

Fair point. Should that block this PR?
</blockquote></details>

Let's actually keep this as is on the assumption we'll refactor so the module itself gives the `Product` and this provides an abstraction to make this easier, and remove this then.
___
*[tools/wptrunner/wptrunner/browsers/firefox.py, line 103 at r1](https://reviewable.io/reviews/web-platform-tests/wpt/13966#-LTOt7h78xEFSpCAlMJf:-LTTF9C0AfrzWi93uuYs:b-c1reqi) ([raw file](https://github.com/web-platform-tests/wpt/blob/ec633de484ac2690150d9a728234a5058b158940/tools/wptrunner/wptrunner/browsers/firefox.py#L103)):*

Chromedriver doesn't support it and it's not clear that it's right for other test types. We could set this in the executor, but you'd still need to set it once per executor.
</blockquote></details>

I guess both of these suck. Meh.

foolip:
@jgraham @gsnedders I see that regressions were looked for in https://github.com/web-platform-tests/wpt/pull/13966#issuecomment-436848048, but that was a month before this was merged. It would have been great to do another trial right before landing.

PR #13975, opened 2018-11-07
[Payment Request] Include both valid and invalid Apple Pay payment methods in payment-request-canmakepayment-method.https.html.

aestes:
Apple Pay has a number of required fields in its payment method data. If these fields are missing,
PaymentRequest.canMakePayment() will resolve to false.

Updated defaultMethods to include a valid Apple Pay payment method. Updated unsupportedMethods to include an invalid Apple Pay payment method.


marcoscaceres (approved):

marcoscaceres:
@aestes and I talked about this test. We decided not to merge, as it didn't quite match what was in the spec.

PR #13988, opened 2018-11-08
webkitRelativePath preserves common ancestors

inexorabletash:
Drop requirement that webkitRelativePath only provides the shortest common ancestor, which matches behavior in Edge/Safari/Firefox, and soon Chrome.

Spec PR: https://github.com/WICG/entries-api/pull/27
Tracking issue: https://github.com/WICG/entries-api/issues/18


aliams (approved):
LGTM

Issue #13989, opened 2018-11-08
Taskcluster Chrome runs for commit 71fe0a6c3d (epochs/daily) failed


foolip:
Commit 71fe0a6c3d286faafc6c5c9849d7f28d2f14ee20 was the previous epochs/daily commit, and Taskcluster did run for it:
https://tools.taskcluster.net/groups/FJTQgT1GQ1qVCsAwYXVKzg

However, it failed, so no Taskcluster results are available in wpt.fyi:
https://wpt.fyi/test-runs?sha=71fe0a6c3d

The Chrome Dev runs (found [via API](https://api.github.com/repos/web-platform-tests/wpt/commits/71fe0a6c3d286faafc6c5c9849d7f28d2f14ee20/statuses)) also failed:
https://tools.taskcluster.net/groups/RoFpb0S_SB2-KBrk-6wGcg

In both cases it was testharness chunk 13 that failed, which seems suspicious.

Looking at https://github.com/web-platform-tests/wpt/commits/master, it looks like Taskcluster reliability has been very, very bad recently.

@Hexcles @jgraham @jugglinmike can any of you looks into this?

@Hexcles, what sort of monitoring should we have in place to notice this immediately when it begins to happen? Quite likely there's some change that was landed days ago that should be reverted.

jgraham:
This is the same as https://github.com/web-platform-tests/results-collection/issues/625

Looking at the logs it seems like we are seeing Chrome crash on startup (historically that's where I've seen the BadStatusLine error) and then we aren't handling it properly. But there are also some worrying errors earlier in the log that look like maybe the stash is failing or something. It's not impossible that the environment is broken in some way (e.g. OOM), but we should certainly ensure that we handle the failure more gracefully rather than just stopping.

jgraham:
https://taskcluster-artifacts.net/OFUJ8juCTeW4ZJ8xksovXw/0/public/logs/live_backing.log is a Firefox failure, which suspiciously comes in the same CSP tests. It looks a lot like the server died in that case. Which could be a bug in the server or, again, could be OOM or similar.

jugglinmike:
This appears to be a race condition in wptserver's "Stash" feature. [I've filed a pull request to fix the bug](https://github.com/web-platform-tests/wpt/pull/14024) and included mode detail there.

foolip:
Results for commit fc1a5b73230225faf32bfb85f3928ce154211b0a are also missing. It was triggered 3 times:
[...]

PR #14024, opened 2018-11-12
[wptserve] Eliminate race condition

jugglinmike:
This race condition was expressed during testing sessions where the
first test to use the Stash feature issued did so with multiple requests
made in parallel.

---

The tests are brittle, but given the conditions we need to model, brittleness seems unavoidable.

The server code hasn't changed recently, so it's a little surprising that a race condition should suddenly become an issue [in Taskcluster](https://github.com/web-platform-tests/wpt/issues/13989) and [in the results-collection project](https://github.com/web-platform-tests/results-collection/issues/625).  Reviewing the status for each commit in `master` shows that [this intermittent error first surfaced on November 5](https://github.com/web-platform-tests/wpt/commits/master?after=7c568fdbd98ccdb92152c14521d3cc8c729a45a2+74).  Guy Fawkes is the obvious suspect, but the trail's gone cold. I turned to more technical explanations.

[The most recent Chromedriver release is version 2.43](https://chromedriver.storage.googleapis.com/index.html), and [that was published on October 16](https://chromedriver.storage.googleapis.com/2.43/notes.txt), ruling out a regression there. Chrome itself may have regressed, but I'm not set up to do any sort of bisecting, so I'm proceeding under the assumption that this is an issue in WPT.

A small incongruity between the Buildbot and Taskcluster setup is useful here.  Each system experiences the issue on a different "chunk" of WPT, but they define the segments in different terms. Buildbot uses 20 "chunks" containing tests of all types (failing on number 3) while Taskcluster uses 15 "chunks" for the testharness.js tests (failing on 13). We can narrow the set of suspect files by using the union of the following two lists:

    ./wpt run --list-tests --this-chunk 3 --total-chunks 20 chrome
    ./wpt run --list-tests --this-chunk 13 --total chunks 15 --test-type testdriver -- chrome

That returns 187 test files to consider. Feeding those files into `git log` shows that only one of them changed on that day:

    $ cat both.txt | xargs git log --after=2018-11-04 --before=2018-11-06 --oneline --
    7e66d13 Ensure eval flag is properly transfered to context from CSPRO

[`content-security-policy/script-src/eval-allowed-in-report-only-mode-and-sends-report.html` doesn't appear to be doing anything invalid.](https://github.com/web-platform-tests/wpt/blob/89188af97522b15cfe7ab6cc1c9cdb58007f0967/content-security-policy/script-src/eval-allowed-in-report-only-mode-and-sends-report.html) It does make parallel requests to a Stash-enabled endpoint, though (via [`checkReport.sub.js`](https://github.com/web-platform-tests/wpt/blob/89188af97522b15cfe7ab6cc1c9cdb58007f0967/content-security-policy/support/checkReport.sub.js)). The race condition fixed by this patch would only be expressed when the first test to use Stash did so via two parallel requests; my guess is that's uncommon but became true on November 5.

### DIFF
--- a/tools/wptserve/tests/test_stash.py
+++ b/tools/wptserve/tests/test_stash.py
@@ -68,7 +68,7 @@ def test_delayed_lock(add_cleanup):
 
     queue = multiprocessing.Queue()
 
-    def handle_lock_request():
+    def mutex_lock_request():
         """This request handler allows the caller to delay execution of a
         thread which has requested a proxied representation of the `lock`
         property, simulating a "slow" interprocess communication channel."""
@@ -78,7 +78,7 @@ def test_delayed_lock(add_cleanup):
         return threading.Lock()
 
     SlowLock.register("get_dict", callable=lambda: {})
-    SlowLock.register("Lock", callable=handle_lock_request)
+    SlowLock.register("Lock", callable=mutex_lock_request)
 
     slowlock = SlowLock(("localhost", 4543), b"some key")
     slowlock.start()
@@ -106,7 +106,7 @@ def test_delayed_dict(add_cleanup):
 
     # This request handler allows the caller to delay execution of a thread
     # which has requested a proxied representation of the "get_dict" property.
-    def handle_dict_request():
+    def mutex_dict_request():
         """This request handler allows the caller to delay execution of a
         thread which has requested a proxied representation of the `get_dict`
         property, simulating a "slow" interprocess communication channel."""
@@ -114,7 +114,7 @@ def test_delayed_dict(add_cleanup):
         response_lock.acquire()
         return {}
 
-    SlowDict.register("get_dict", callable=handle_dict_request)
+    SlowDict.register("get_dict", callable=mutex_dict_request)
     SlowDict.register("Lock", callable=lambda: threading.Lock())
 
     slowdict = SlowDict(("localhost", 4543), b"some key")
--- a/tools/wptserve/tests/test_stash.py
+++ b/tools/wptserve/tests/test_stash.py
@@ -19,8 +19,8 @@ def add_cleanup():
         fn()
 
 def test_delayed_lock(add_cleanup):
-    '''Ensure that delays in proxied Lock retrieval do not interfere with
-    initialization in parallel threads.'''
+    """Ensure that delays in proxied Lock retrieval do not interfere with
+    initialization in parallel threads."""
 
     class SlowLock(BaseManager):
         pass
@@ -38,13 +38,13 @@ def test_delayed_lock(add_cleanup):
     SlowLock.register("get_dict", callable=lambda: {})
     SlowLock.register("Lock", callable=handle_lock_request)
 
-    slowlock = SlowLock(('localhost', 4543), 'some key')
+    slowlock = SlowLock(("localhost", 4543), b"some key")
     slowlock.start()
     add_cleanup(lambda: slowlock.shutdown())
 
     def run(process_queue, request_lock, response_lock):
         def target(thread_queue):
-            stash = Stash('/', ('localhost', 4543), 'some key')
+            stash = Stash("/", ("localhost", 4543), b"some key")
             thread_queue.put(stash.lock is None)
 
         thread_queue = multiprocessing.Queue()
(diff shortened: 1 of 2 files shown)


gsnedders:
At first glance this looks good, but I definitely need to have a closer look while more awake; thanks for tracking this down!

jgraham (approved):
Excellent job diagnosing this issue and providing a fix.

I am slightly concerned about the tests; it took a lot longer to understand the test code than it did to understand the fix. If we want to keep them I think we should factor out the 90% common code, and comment it thoroughly so that it's easier to understand what is supposed to be happening at each point; a test involving two processes and two threads is inherently complex and difficult to understand.

tools/wptserve/tests/test_stash.py
```diff
@@ -0,0 +1,131 @@
...
+    yield add
+
+    for fn in fns:
+        fn()
+
+def test_delayed_lock(add_cleanup):
+    """Ensure that delays in proxied Lock retrieval do not interfere with
+    initialization in parallel threads."""
+
+    class SlowLock(BaseManager):
```
gsnedders: What makes this slow?

tools/wptserve/tests/test_stash.py
```diff
@@ -0,0 +1,131 @@
...
+    add_cleanup(lambda: parallel.terminate())
+
+    assert [queue.get(), queue.get()] == [False, False], (
+        "both instances had valid locks")
+
+def test_delayed_dict(add_cleanup):
+    """Ensure that delays in proxied `dict` retrieval do not interfere with
+    initialization in parallel threads."""
+
+    class SlowDict(BaseManager):
```
gsnedders: What makes this slow?

tools/wptserve/tests/test_stash.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def handle_lock_request():
+        request_lock.release()
+        response_lock.acquire()
+        return threading.Lock()
+
+    SlowLock.register("get_dict", callable=lambda: {})
+    SlowLock.register("Lock", callable=handle_lock_request)
+
+    slowlock = SlowLock(("localhost", 4543), b"some key")
```
gsnedders: I hate to be stupid, but what effect does this manager have?

gsnedders (commented):
(Mostly just agree with @jgraham)

jugglinmike:
Thanks for the review, @jgraham and @gsnedders. I've attempted to address @jgraham's feedback by answering @gsnedders' questions via inline documentation. I've also reduced duplication by factoring out the multiprocessing target (`run`) and sharing it between both tests.

tools/wptserve/tests/test_stash.py
```diff
@@ -0,0 +1,130 @@
...
+    # Create a second thread in order to inspect the behavior of the Stash
+    # constructor at this moment.
+
+    second.start()
+
+    # Allow the `first` thread to proceed
+
+    response_lock.release()
+
+    # Wait for both threads to complete and report their stateto the test
```
jgraham: type: `stateto`

tools/wptserve/tests/test_stash.py
```diff
@@ -0,0 +1,130 @@
...
+    def handle_lock_request():
+        """This request handler allows the caller to delay execution of a
+        thread which has requested a proxied representation of the `lock`
+        property, simulating a "slow" interprocess communication channel."""
+
+        request_lock.release()
+        response_lock.acquire()
+        return threading.Lock()
+
+    SlowLock.register("get_dict", callable=lambda: {})
```
jgraham: It still somewhat seems like you could write these tets as as a parameterised test in which your parameters are `get_dict` and `Lock`, with one case having the parameters `(lambda :{}, mutex_lock_request)` and one having the parameters `(mutex_get_dict, lambda: threading.Lock())`. I also *think* the names with "mutex" in make more sense than the generic `handle_` names.

jgraham (approved):
This looks better, and I don't want to block it landing, but there is possibly scope for one more round of cleanup.

tools/wptserve/tests/test_stash.py
jgraham: Or maybe `blockable` rather than `mutex`

jugglinmike:
Those functions can't be parameterized like that because they need to reference
locks (currently named `request_lock` and `response_lock`) which are shared
with the child process created by the test.

We could also parameterize the locks, and maybe make some sort of
mutex-function-producing factory function, but that kind of abstraction is
contrary to our goal of making the test easier to understand.

I agree that `mutex_` is more descriptive than `handle_`. I've updated the
function names accordingly.

jgraham:
I mean you could make the locks globals and set them at the start of the test and unset them in the cleanup. But I'm going to merge this PR now for the correctness fix, and if we want to clean up the tests more we can do that in a later PR.

jugglinmike:
Thanks!

PR #14025, opened 2018-11-12
Find manifest for download by tags instead of commits

foolip:
This makes it possible to use `./wpt manifest` on a checkout that is
more than 50 commits behind master. The downside is that it will require
more network requests if the most recent tag/tags don't have releases
uploaded.

Add more warnings to make it clearer why it goes wrong when it does.

### DIFF
--- a/tools/manifest/download.py
+++ b/tools/manifest/download.py
@@ -33,46 +33,53 @@ def should_download(manifest_path, rebuild_time=timedelta(days=5)):
     return False
 
 
-def git_commits(repo_root):
+def merge_pr_tags(repo_root, max_count=50):
     git = Git.get_func(repo_root)
-    return [item for item in git("log", "--format=%H", "-n50").split("\n") if item]
+    tags = []
+    for line in git("log", "--format=%D", "--max-count=%s" % max_count).split("\n"):
+        for ref in line.split(", "):
+            if ref.startswith("tag: merge_pr_"):
+                tags.append(ref[5:])
+    return tags
 
 
-def github_url(commits):
-    try:
-        resp = urlopen("https://api.github.com/repos/web-platform-tests/wpt/releases")
-    except Exception:
-        return None
+def github_url(tags):
+    for tag in tags:
+        url = "https://api.github.com/repos/web-platform-tests/wpt/releases/tags/%s" % tag
+        try:
+            resp = urlopen(url)
+        except Exception:
+            logger.warning("Fetching %s failed" % url)
+            continue
 
-    if resp.code != 200:
-        return None
+        if resp.code != 200:
+            logger.warning("Fetching %s failed; got HTTP status %d" % (url, resp.code))
+            continue
 
-    try:
-        releases = json.load(resp.fp)
-    except ValueError:
-        logger.warning("Response was not valid JSON")
-        return None
+        try:
+            release = json.load(resp.fp)
+        except ValueError:
+            logger.warning("Response was not valid JSON")
+            return None
 
-    fallback = None
-    for release in releases:
-        for commit in commits:
-            for item in release["assets"]:
-                if item["name"] == "MANIFEST-%s.json.gz" % commit:
-                    return item["browser_download_url"]
-                elif item["name"] == "MANIFEST.json.gz" and not fallback:
-                    fallback = item["browser_download_url"]
-    if fallback:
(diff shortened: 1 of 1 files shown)


tools/manifest/download.py
```diff
@@ -33,46 +33,53 @@ def should_download(manifest_path, rebuild_time=timedelta(days=5)):
...
-                elif item["name"] == "MANIFEST.json.gz" and not fallback:
-                    fallback = item["browser_download_url"]
-    if fallback:
-        logger.info("Can't find a commit-specific manifest so just using the most recent one")
-        return fallback
-
-
-def download_manifest(manifest_path, commits_func, url_func, force=False):
+    tags = []
+    for line in git("log", "--format=%D", "--max-count=%s" % max_count).split("\n"):
```
gsnedders: `git describe --tags --match 'merge_pr_*'` does basically this (this is supported back to ancient times)

gsnedders: or maybe that doesn't work if we need more than one? hmm

foolip: It won't get just the tags reachable from HEAD, or even the tags in tree order.

foolip: And yeah, also just one.

foolip: We could call it again using old_tag^1 of course. Possibly that would look nicer?

gsnedders: > It won't get just the tags reachable from HEAD, or even the tags in tree order.

I thought that's exactly what it did with no refish?

> We could call it again using old_tag^1 of course. Possibly that would look nicer?

For non-linear trees I'm pretty sure that has the wrong behaviour. And given this is often called from PRs I'm pretty sure that's even more wrong.

foolip: `git describe --tags --match 'merge_pr_*'` just tries to describe HEAD, man page says "Commit-ish object names to describe. Defaults to HEAD if omitted." Please provide exact command if you think there's a way :)

If we did want to call `git describe` many times to get from merge_pr_14054 (a merge commit) to merge_pr_14137 (the previous tag) then `^1` is the right thing unless an admin disables the master branch protection and pushes a merge commit to master with old master as the non-first parent *and* that push closes a PR such that it becomes tagged with merge_pr_*. And then all that would happen is that maybe an older version of the manifest were found. The only improvement would be `git describe --tags` all the parents and then pick the most recent tag. But then you're better off using `git log` to walk commits in date order as I've already done.

gsnedders: @foolip according to the man page I have here:

> The command finds the most recent tag that is reachable from a commit. If the tag
       points to the commit, then only the tag is shown. Otherwise, it suffixes the tag
       name with the number of additional commits on top of the tagged object and the
       abbreviated object name of the most recent commit.

On `wptrunner_tlbc` for example it currently gives `merge_pr_13862-3-g2ed77c929e`. Adding `--abbrev=0` gets just the tag: `merge_pr_13862`.

See also the [Search Strategy](https://www.git-scm.com/docs/git-describe#_search_strategy).

jgraham: I think what's here is fine.

gsnedders: Right, indeed, which is why this doesn't work.

tools/manifest/download.py
```diff
@@ -33,46 +33,53 @@ def should_download(manifest_path, rebuild_time=timedelta(days=5)):
...
+            # there's no longer a reason to include the commit sha.
+            if item["name"].startswith("MANIFEST-") and item["name"].endswith(".json.gz"):
+                return item["browser_download_url"]
+            elif item["name"] == "MANIFEST.json.gz":
+                return item["browser_download_url"]
+
+    return None
+
+
+def download_manifest(manifest_path, tags_func, url_func, force=False):
```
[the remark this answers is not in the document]
jgraham: Well the right indicrection would probably be a single function that did the equivalent of `f = lambda x:url_func(tags_func())`, although in practice we built the gecko equivalent of this at a higher level (perhaps incorrectly) so I don't mind. In any case, followup.

jgraham (approved):

tools/manifest/download.py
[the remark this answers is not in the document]
foolip: Alrighty, thanks!

PR #14031, opened 2018-11-13
SVG Painting: inheritance and initial values

ewilligers:
Painting properties inherit, and have the expected initial values.
https://svgwg.org/svg2-draft/painting.html

### DIFF
--- a/svg/painting/inheritance.svg
+++ b/svg/painting/inheritance.svg
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg"
+     xmlns:h="http://www.w3.org/1999/xhtml"
+     width="800px" height="8000px">
+  <title>Inheritance of SVG painting properties</title>
+  <metadata>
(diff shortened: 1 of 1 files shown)


svg/painting/inheritance.svg
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg"
+     xmlns:h="http://www.w3.org/1999/xhtml"
+     width="800px" height="8000px">
```
svgeesus: 8000px?

Does this assume the UA paints into a buffer the non-visible part?

svgeesus (approved):
LGTM (assuming the 8000px height is not an issue)

svg/painting/inheritance.svg
ewilligers: If there are lots of test results, we can scroll down to read them all. There was a similar test I wrote (not uploaded) and I couldn't read all the results when I opened the test .svg in Safari unless I increased the height.

PR #14032, opened 2018-11-13
SVG Text: inheritance, initial values

ewilligers:
Test that SVG text properties inherit according to spec.
Test that they have the expected initial values.
https://svgwg.org/svg2-draft/text.html


svgeesus (approved):
LGTM, straightforward

Issue #7660, opened 2017-10-10
When many tests are affected, CI stability jobs will time out


foolip:
In https://github.com/w3c/web-platform-tests/pull/7654 I changed 778 files, and as a result all of the stability jobs failed, with messages like "No output has been received in the last 10m0s".

This is similar to https://github.com/w3c/web-platform-tests/issues/7073.

jgraham:
There are two issues here. One is that we should have a periodic keepalive message when the tests are still running but not producing output. The other is that if we get no output for 10 minutes we're pretty much guaranteed to hit the hard timout (since that implies a minimum runtime of 100 minutes which is about twice the threshold).

foolip:
Changed the title a bit because in https://github.com/w3c/web-platform-tests/pull/9641 not that many files were changed, but many included the changed file, and that's what matters. Also, not *all* browsers necessarily fail.

gsnedders:
Does it make sense to run with `--no-restart-on-unexpected` in CI?

foolip:
@jgraham, what kinds of things would you expect to blow up if we do that? And would that still restart between each of the 10 runs?

jgraham:
Nothing should blow up; the default assumes that if you have one failing test then you don't want it to infect subsequent tests if it happens to set some bad state. In this case what you might see is that one unstable test would be more likely to make subsequent tests on the run look unstable. It won't affect restarting between runs at all.

An alternate apporach would be to set the expectation data after the first run and use that to avoid restarting on subequent runs unless something actually is unstable. At the moment the way we update the metadata is really slow, so that will itself have a performance impact.

foolip:
@jgraham, as you can see, this is quite a common occurrence, requiring regular manual intervention, and our current setup isn't robust enough to build https://github.com/w3c/web-platform-tests/issues/7475 on top of. Are you open to revisiting the idea of running fewer iterations if we know we're going to exceed the timeout?

jgraham:
So I'm not opposed to that, but it is a limitation of the current system that we should look to address in a better way.

But this issue seems to be about the "10m without output" problem which is different to the entire run timing out and requires  a different solution.

foolip:
I think most of the PRs I've linked to here are were actual timeouts close to 50 minutes, at least https://travis-ci.org/w3c/web-platform-tests/jobs/352958109 ended with just "The job exceeded the maximum time limit for jobs, and has been terminated".

[...]

PR #14033, opened 2018-11-13
Remove Travis stability checks (Chrome Dev + Firefox Nightly)

foolip:
Fixes https://github.com/web-platform-tests/wpt/issues/7660.


foolip:
I'm sending this PR now because these stability checks create seemingly endless trouble, just today on a bunch of Chromium export PRs and also on https://github.com/web-platform-tests/wpt/pull/13951. Some of the problems it catches are real, but enough are just timeouts and the like that I find myself distrusting the checks and not wanting to ask anyone else to look into failures either.

Disabling the checks will make it harder to verify that the Taskcluster equivalent is working, and at this point that is the main value of these checks. Perhaps we're at a point where we can flip things around to make the Taskcluster checks blocking and keep running Travis in a non-blocking way for a while to verify?

jgraham:
We need to make the TC checks actually fail when there's a problem.

jugglinmike:
[Here is some more analysis](https://github.com/web-platform-tests/wpt/issues/13194#issuecomment-438500438)

sideshowbarker:
I’m not sure what this PR is waiting on, but short of merging this, is there any reason not to go ahead and update .travis.yml to move the stability checks to being allow_failures cases?

foolip:
Sent https://github.com/web-platform-tests/wpt/pull/14096 instead.

foolip:
@jugglinmike, can any information be extracted by comparing the results in the period where this was an allowed failure?

I've seen way more PRs blocked on flakiness lately, so I strongly suspect the new setup is just better at finding flakiness. If nobody suspects otherwise I think it's fine to not spend much time looking for cases where Travis failed but Taskcluster passed.

gsnedders (approved):
Though we should probably wait for more feedback than just from me. 🙂

jgraham (approved):

PR #14040, opened 2018-11-13
Replace direct use of offerToReceiveAudio by generateAudioReceiveOnlyOffer

youennf:
This allows to use non legacy APIs by default if supported.


youennf:
@alvestrand, @soareschen, PTAL.

soareschen (approved):

PR #14061, opened 2018-11-14
[testharness.js] Honor test config in workers

### LINKED ISSUE #14041: allow_uncaught_exception flag does not work as expected for worker.html promise_test in Chrome
In Chrome 72 it appears that the allow_uncaught_exception flag does not behave as expected during promise_test execution for worker.html tests, but rather still allows the testharness to error as a result of an uncaught exception.

To test this, copy the following small test case into a .any.js file and and run it as a worker.html test:

```javascript
// META: global=window,worker
 
setup({allow_uncaught_exception:true});
 
promise_test(async testCase => {
  const error = new Error("boo");
  const eventReceived = new Promise((resolve,reject) => {
    self.addEventListener("error", ev => {
      assert_equals(ev.error, error);
      resolve();
    });
  });
  queueMicrotask(() => { throw error; });
  await eventReceived;
}, 'error catching test');
```

The output result on chrome 72 should be similar to the following:
"
This is a testharness.js-based test.
Harness Error. harness_status.status = 1 , harness_status.message = Error in remote http://web-platform.test:8001/IndexedDB/promise-test-error.any.js: Uncaught Error: boo
PASS error catching test
Harness: the test ran to completion.
"

As can be seen, the test passes (as it should) but the testharness still errors because of the uncaught exception. My understanding is that by setting the allow_uncaught_exception flag this should not be the case. The issue appears to be unique to worker.html promise_tests (the same issue was not replicable for async tests or promise_tests run normally, on a serviceworker, or on a sharedworker).

jugglinmike:
Ensure that uncaught exceptions originating from workers do not
influence harness status when the `allow_uncaught_exception` setting has
been enabled.

This is intended to resolve gh-14041.

### DIFF
--- a/lint.whitelist
+++ b/lint.whitelist
@@ -311,6 +311,7 @@ SET TIMEOUT: resources/test/tests/functional/add_cleanup_async_rejection.html
 SET TIMEOUT: resources/test/tests/functional/add_cleanup_async_rejection_after_load.html
 SET TIMEOUT: resources/test/tests/functional/api-tests-1.html
 SET TIMEOUT: resources/test/tests/functional/worker.js
+SET TIMEOUT: resources/test/tests/functional/worker-uncaught-allow.js
 SET TIMEOUT: resources/test/tests/unit/exceptional-cases.html
 SET TIMEOUT: resources/testharness.js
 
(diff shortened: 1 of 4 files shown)


jgraham (approved):
Thanks!

Issue #7281, opened 2017-09-07
Automating fullscreen tests


gsnedders:
Even after #6897 lands we won't be able to automate fullscreen tests; we run testharness.js tests with two windows, where WebDriver has the parent window as the current browsing context and the test window is an auxiliary browsing context of that.

To enter fullscreen, we need a user-initiated click. With #6897, we do this by switching the current browsing context to the window that wants the click, clicking, then switching back to the parent window. However, in current implementations the OS focus follows the current top-level browsing context (AFAICT), and therefore when we change back to the parent window we move out of fullscreen (because on at least some platforms you can't have another window having focus while something is fullscreen). This makes testing fullscreen currently impossible.

cc/ @foolip

foolip:
@JKereliuk, this is the issue I mentioned.

foolip:
What would a fix for this have to look like? What is the fundamental reason that it matters for WebDriver which window has focus?

JKereliuk:
Whoops I missed this, I am going to take closer look at these tests

foolip:
After https://github.com/web-platform-tests/wpt/pull/13419 it's probably possible to use `test_driver.bless` to automate some fullscreen tests, at least the ones that don't involve iframes.

PR #14072, opened 2018-11-15
Automate one fullscreen test using `test_driver.bless`

foolip:
Part of https://github.com/web-platform-tests/wpt/issues/7281.

### DIFF
--- a/fullscreen/api/element-request-fullscreen.html
+++ b/fullscreen/api/element-request-fullscreen.html
@@ -21,6 +21,7 @@ async_test(function(t)
         t.done();
     }));
 
-    test_driver.bless('fullscreen', () => div.requestFullscreen());
+    test_driver.bless("fullscreen", () => div.requestFullscreen())
+        .catch(t.step_func(assert_unreached));
 });
 </script>
--- a/fullscreen/api/element-request-fullscreen.html
+++ b/fullscreen/api/element-request-fullscreen.html
@@ -2,11 +2,14 @@
 <title>Element#requestFullscreen()</title>
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
-<script src="../trusted-click.js"></script>
+<script src="/resources/testdriver.js"></script>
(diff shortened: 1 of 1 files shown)


foolip:
I'd like to make this single change, and if it works everywhere convert more tests.

upsuper:
Is there any document for `test_driver.bless`?

fullscreen/api/element-request-fullscreen.html
```diff
@@ -18,6 +21,7 @@
         t.done();
     }));
 
-    trusted_request(t, div);
+    test_driver.bless("fullscreen", () => div.requestFullscreen())
+        .catch(t.step_func(assert_unreached));
```
Ms2ger: `t.unreached_func(…)`

foolip: Unfortunately, to include the reason it would have to be `.catch(reason => t.unreached_func(reason)())` which I thought was a lot worse.

foolip:
@upsuper see https://web-platform-tests.org/writing-tests/testdriver.html#bless

foolip:
Ping @upsuper @Ms2ger.

mustaqahmed (approved):

upsuper:
Sorry for the delay. It looks good to me.

foolip:
Results are in https://wpt.fyi/results/fullscreen/api/element-request-fullscreen.html?label=experimental

@upsuper it didn't work for Firefox, can you tell why?

foolip:
Yeah, `div.requestFullscreen()` must have rejected.

jgraham:
I strongly agree about the warnings, but I don't know how to arrange that. We don't have access to the logs in marionette/WebDriver, so it would have to be browser-specific e.g. a setting to dump all console output to stdout. Does that exist for Firefox?

foolip:
There's also just a problem with this test. I tried lots of ways in this PR, trying another now: https://github.com/web-platform-tests/wpt/pull/14501

PR #14073, opened 2018-11-15
Allow `test_driver.bless` to work when running manually

foolip:
Tests using `test_driver.bless` now currently always fail when running
them manually, because the default `test_driver_internal.click`
rejects with "unimplemented".

Mute the rejection if the session isn't controlled by WebDriver, which
we check with `navigator.webdriver`.

### DIFF
--- a/resources/testdriver.js
+++ b/resources/testdriver.js
@@ -73,7 +73,11 @@
             return new Promise(function(resolve, reject) {
                     button.addEventListener("click", resolve);
 
-                    test_driver.click(button).catch(reject);
+                    test_driver.click(button).catch(reason => {
+                        if (navigator.webdriver) {
+                            reject(reason);
+                        }
+                    });
                 }).then(function() {
                     button.remove();
 


foolip:
This is an alternative to https://github.com/web-platform-tests/wpt/pull/11173, and only fixes `test_driver.bless`.

jgraham:
I'm somewhat concerned about the use of the `webdriver` property here; although it's possibly fine it's unclear how future-proof it is to assume that all automation is webdriver-driven.

foolip:
I'd be happy to use any other signal we have that we're being controlled by wptrunner. Do we have something else we could use?

jugglinmike:
[That's the question which is blocking gh-11173.](https://github.com/web-platform-tests/wpt/pull/11173#issuecomment-400029951)

foolip:
The answer is straightforward I think: use `navigator.webdriver`, add a test to /infrastructure/assumptions/ for it, and refine if there's a problem. The consequence of getting it wrong is just to change the failure mode from fast fail to timeout.

gsnedders:
@foolip does this work for `run-webkit-tests` and Marionette and CDP and other non-WebDriver runners?

foolip:
@gsnedders, no, but there we'll replace `test_driver_internal.click` anyway. (Moving this sort of thing into `test_driver_internal.*` might make sense.)

foolip:
@jugglinmike, do you intend to get https://github.com/web-platform-tests/wpt/pull/11173 landed? If that's not imminent I'd like to land something simple like this PR that makes manual testing possible.

jugglinmike:
I'd like to land gh-11173. I've just updated it in an attempt to satisfy the remaining concerns.

foolip:
All right, reviewing that.

PR #14074, opened 2018-11-15
Add test for response headers

christian-bromann:
The WebDriver spec defines the response header in the [processing modal section](https://www.w3.org/TR/webdriver1/#processing-model) as follows:

> 3. Set the response’s header with name and value with the following values:
>
> Content-Type
>    "application/json; charset=utf-8"
> Cache-Control
>    "no-cache"

This patch adds tests for this case.

__Question__: the RFC defines header names as case insensitive. Should be lowercase header keys and values by default?

__Note:__: Geckodriver returns a wrong response value for content-type, so I created a bug here: https://bugzilla.mozilla.org/show_bug.cgi?id=1507428

### DIFF
--- a/webdriver/tests/new_session/headers.py
+++ b/webdriver/tests/new_session/headers.py
@@ -1,4 +1,4 @@
-def test_missing_first_match(new_session, add_browser_capabilities):
+def test_response_headers(new_session, add_browser_capabilities):
     response, _ = new_session({"capabilities": {"alwaysMatch": add_browser_capabilities({})}})
     assert 'cache-control' in response.headers
     assert 'no-cache' == response.headers['cache-control']
--- a/webdriver/tests/new_session/headers.py
+++ b/webdriver/tests/new_session/headers.py
@@ -1,0 +1,6 @@
+def test_missing_first_match(new_session, add_browser_capabilities):
+    response, _ = new_session({"capabilities": {"alwaysMatch": add_browser_capabilities({})}})
+    assert 'cache-control' in response.headers
+    assert 'no-cache' == response.headers['cache-control']
+    assert 'content-type' in response.headers
+    assert 'application/json; charset=utf-8' == response.headers['content-type']
--- a/webdriver/tests/new_session/headers.py
+++ b/webdriver/tests/new_session/headers.py
@@ -1,6 +1,0 @@
-def test_response_headers(new_session, add_browser_capabilities):
-    response, _ = new_session({"capabilities": {"alwaysMatch": add_browser_capabilities({})}})
-    assert 'cache-control' in response.headers
-    assert 'no-cache' == response.headers['cache-control']
-    assert 'content-type' in response.headers
-    assert 'application/json; charset=utf-8' == response.headers['content-type']
--- a/tools/webdriver/webdriver/client.py
+++ b/tools/webdriver/webdriver/client.py
@@ -470,7 +470,7 @@ class Session(object):
                 self.session_id = None
 
             raise err
-        
+
         self._assert_response_headers(response.headers)
 
         if "value" in response.body:
@@ -508,11 +508,11 @@ class Session(object):
         """
         url = urlparse.urljoin("session/%s/" % self.session_id, uri)
         return self.send_command(method, url, body)
-    
+
     def _assert_response_headers(self, headers):
         """
         Method to assert response headers for WebDriver requests
-        
+
         :param headers: dict with header data
         """
         assert 'cache-control' in headers
--- a/tools/webdriver/webdriver/client.py
+++ b/tools/webdriver/webdriver/client.py
@@ -471,8 +471,6 @@ class Session(object):
 
             raise err
 
-        self._assert_response_headers(response.headers)
-
         if "value" in response.body:
             value = response.body["value"]
             """
@@ -508,17 +506,6 @@ class Session(object):
         """
         url = urlparse.urljoin("session/%s/" % self.session_id, uri)
         return self.send_command(method, url, body)
-
-    def _assert_response_headers(self, headers):
-        """
-        Method to assert response headers for WebDriver requests
-
-        :param headers: dict with header data
-        """
-        assert 'cache-control' in headers
-        assert 'no-cache' == headers['cache-control']
-        assert 'content-type' in headers
-        assert 'application/json; charset=utf-8' == headers['content-type']
 
     @property
     @command
--- a/tools/webdriver/webdriver/client.py
+++ b/tools/webdriver/webdriver/client.py
@@ -470,6 +470,8 @@ class Session(object):
                 self.session_id = None
 
             raise err
+        
+        self._assert_response_headers(response.headers)
 
         if "value" in response.body:
             value = response.body["value"]
@@ -506,6 +508,17 @@ class Session(object):
         """
         url = urlparse.urljoin("session/%s/" % self.session_id, uri)
         return self.send_command(method, url, body)
+    
+    def _assert_response_headers(self, headers):
+        """
+        Method to assert response headers for WebDriver requests
+        
+        :param headers: dict with header data
+        """
+        assert 'cache-control' in headers
+        assert 'no-cache' == headers['cache-control']
+        assert 'content-type' in headers
+        assert 'application/json; charset=utf-8' == headers['content-type']
 
     @property
     @command
--- a/tools/webdriver/webdriver/transport.py
+++ b/tools/webdriver/webdriver/transport.py
@@ -16,9 +16,10 @@ class Response(object):
     body has been read and parsed as appropriate.
     """
 
-    def __init__(self, status, body):
+    def __init__(self, status, body, headers):
         self.status = status
         self.body = body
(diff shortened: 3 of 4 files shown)


webdriver/tests/new_session/headers.py
```diff
@@ -0,0 +1,6 @@
+def test_missing_first_match(new_session, add_browser_capabilities):
+    response, _ = new_session({"capabilities": {"alwaysMatch": add_browser_capabilities({})}})
+    assert 'cache-control' in response.headers
```
jgraham: I think you should lowercase the recieved values here.

webdriver/tests/new_session/headers.py
```diff
@@ -0,0 +1,6 @@
+def test_missing_first_match(new_session, add_browser_capabilities):
```
jgraham: This isn't really what this is testing. You could possibly even just use the `status` endpoint to test this since it doesn't require a session.

jgraham (dismissed):

webdriver/tests/new_session/headers.py
christian-bromann: oh dang, copy paste issue

webdriver/tests/new_session/headers.py
christian-bromann: @jgraham see PR question: shall we by default lowercase key and value pairs since headers are case insensitive?

christian-bromann:
@jgraham given the fact that this test should be applied to every response I moved the assertion into the `Session` class. I also converted header key/value pairs into lowercase per default.

andreastt:
There also appears to be some test failures after one of the recent fixups:
https://travis-ci.org/web-platform-tests/wpt/builds/455470716?utm_source=github_status&utm_medium=notification

christian-bromann:
@andreastt addressed outstanding issues

andreastt (approved):

andreastt:
Retriggered some of the jobs and dismissed @jgraham’s old review.

christian-bromann:
Azure Pipeline is failing due to
- Chrome not having `cache-control` in command response header
- FF Nightly returning `application/json; charset=utf8` in response header

continuous-integration/travis-ci/pr failing due to
- FF Nightly returning `application/json; charset=utf8` in response header

andreastt:
@jgraham Aren’t those jobs supposed to run stability tests, like Travis?

andreastt:
Or we could keep the checks in the client and emit a warning when the headers do not conform. I support the idea of moving the hard assertions to a separate test.

andreastt (changes requested):
See comment above.

christian-bromann:
@andreastt @jugglinmike I moved the assertions into the `assert_error` and `assert_success` assertion helpers which are only used in the webdriver test harness. What do you think?

jugglinmike (approved):
Good idea! Unlike the alternatives we've been considering, this may allow some responses to go untested. Those cases seem rare, so I think the simplicity of this approach wins out.

As for the timeout reported by Taskcluster: [unlike with the previous solution](https://github.com/web-platform-tests/wpt/pull/14074#issuecomment-439517580), the system is now capable of identifying the affected tests:

>     2018-11-19 12:17:10,747 - tc-run - INFO - Identifying tests affected in range 'HEAD^'...
>     2018-11-19 12:17:26,483 - tc-run - INFO - Identified 138 affected tests

That's a lot of tests, possibly enough to exceed our 2-hour limit, though the logs for Chrome are suspiciously short...

andreastt (approved):

andreastt:
It looks like both the Chrome and Firefox stability tests timed out after two hours, whereas the regular test results passed. Is the timeout here too short when lots of tests are affected, @jgraham?

whimboo:
@jgraham mind having a look at this issue? You may have missed @andreastt's question from Nov 20th. Thanks.

andreastt:
Just to be sure, let’s retrigger the tests once more.

whimboo:
There are still various job failures for TaskCluster, TravisCI, and Azure Pipelines.

gsnedders:
Most of these look like all the brokenness we temporarily had on Friday… :/

whimboo:
The newly added assertion has failures:
https://tools.taskcluster.net/groups/cZgXyDz3T0S0m1Sh4X024w/tasks/Ub-qkzjxRoCtmPoz6V_PNg/runs/1/logs/public%2Flogs%2Flive.log#L96762

> TypeError: assert_response_headers() takes exactly 2 arguments (1 given)

@christian-bromann can you please fix that? Thanks

christian-bromann:
@whimboo done

- FF still fails because content type is `application/json; charset=utf8` instead of `application/json; charset=utf-8`
- Chrome still fails as it doesn't have a cache control header

whimboo:
I don't know why it fails with Firefox given that we shipped the fix for `utf-8` with geckodriver 0.24.0. @jugglinmike do we still have to manually change the version number of the driver?

jugglinmike:
[Nope](https://github.com/web-platform-tests/wpt/commit/480fd85811ed2e3449d48bc0c8b8ec09d0e20412)

whimboo:
Oh wait. The Firefox job with the patch applied doesn't fail! What's failing is the verify job, which is killed by taskcluster after a runtime of 7200s.

christian-bromann:
@whimboo would closing and reopening the PR help?

whimboo:
Not sure. Running all the tests simply take that long. Someone might take a bit of time to check how long those verify tasks run in general, and maybe we have to bump up the maximum allowed time.

gsnedders:
@whimboo I'm not sure we can while we're using the GitHub Taskcluster integration, but @jgraham would know better. There are a few cases where we end up running lots of tests, but I think the longer term plan is to do something smarter and dynamically shard them.

In either case, if you think this looks good, we should just force-merge this.

mustjab:
@jgraham I see WPT WebDriver tests failing in Chrome on Windows and Linux runs with this change, and pass after reverting it locally. Anyone else seeing this?

```
  Γöé webdriver\tests\add_cookie\add.py:17:
  Γöé _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  Γöé webdriver\tests\support\asserts.py:55: in assert_error
  Γöé     assert_response_headers(response.headers)
  Γöé _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  Γöé
  Γöé headers = {'content-length': '3243', 'content-type': 'application/json; charset=utf-8'}
  Γöé
  Γöé     def assert_response_headers(headers):
  Γöé         """
  Γöé         Method to assert response headers for WebDriver requests
  Γöé
  Γöé         :param headers: dict with header data
  Γöé         """
  Γöé >       assert 'cache-control' in headers
  Γöé E       AssertionError: assert 'cache-control' in {'content-length': '3243', 'content-type': 'application/json; charset=utf-8'}
  Γöé
  Γöé headers    = {'content-length': '3243', 'content-type': 'application/json; charset=utf-8'}
  Γöé
  Γöö webdriver\tests\support\asserts.py:82: AssertionError
```

andreastt:
@mustjab I think that would be expected, since chromedriver doesn’t return the mandated `Cache-Control` header.

See [_send a response_](https://w3c.github.io/webdriver/#dfn-send-a-response) in the spec.

mustjab:
@andreastt So should we change the way we run these tests? We run them on a daily basis and at this moment most of the tests fail due to this error.

andreastt:
Right, the assertion was added as a step to `assert_success()`, causing most—if not all!—of the wdspec tests to fail because the driver doesn’t return the mandated headers.

In order to avoid numerous workarounds, I think the best fix here is to insert a new line into https://cs.chromium.org/chromium/src/chrome/test/chromedriver/server/http_handler.cc?l=1125 and fix chromedriver. Alternatively we could back this change out temporarily to give you time to fix this.

mustjab:
Thank you! I'll outreach to ChromeDriver team to fix this and for now, we'll run with this fix internally.

jgraham:
I should have read the patch before merging it. I fully agree that this shouldn't be tested for every test but should be a specific set of tests or otherwise isolated from testing other features.

christian-bromann:
@jgraham I will tweak the patch and make it so that it is handled in a single test instead of being part of all assertions.

whimboo:
@jgraham, so we want to backout the already landed patch for a clean base? I don't think patching on top of the current patch is that great.

andreastt:
If @mustjab is fixing chromedriver, I don’t necessarily see the need to move the assertion to separate test, nor a reason to back this out.

Hexcles:
https://chromium-review.googlesource.com/c/chromium/src/+/1501997

PR #14076, opened 2018-11-15
Revert PR 13640

jgraham:
This broke the infrastructure tests by trying to change the testdriver API to something nonstandard.


Ms2ger (approved):

gsnedders:
This reverted #13640.

NavidZ:
That's fair. Thanks for taking care of that. For now we stick to updating a few tests that doesn't hit that lack of the functionality and in parallel we can work on that piece in the standard.

PR #14095, opened 2018-11-16
EventSource: test U+0000 in ID better

annevk:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1387355 for context.


eventsource/format-field-id-null.window.js
```diff
@@ -0,0 +1,25 @@
...
+    const source = new EventSource("resources/last-event-id.py?idvalue=" + encodedIdValue);
+    let seenhello = false;
+    source.onmessage = t.step_func(e => {
+      if (e.data == "hello" && !seenhello) {
+        seenhello = true;
+        assert_equals(e.lastEventId, "");
+      } else if(seenhello) {
+        assert_equals(e.data, "hello");
+        assert_equals(e.lastEventId, "");
+        source.close();
```
foolip: If this really matters, then doing it in cleanup would be more robust.

foolip (approved):

PR #14096, opened 2018-11-16
Allow stability-checker failures on Travis and report them on Taskcluster

foolip:
Fixes https://github.com/web-platform-tests/wpt/issues/7660.

https://github.com/web-platform-tests/wpt/issues/13194 remains open to
verify that Taskcluster isn't failing to detect flakiness that Travis
would.


foolip:
I didn't get the Travis change right, trying again...

jgraham (approved):

foolip:
Alright, I appear to have succeeded! Review please :)

Ms2ger (approved):
Please update the commit message and remove the flaky test before landing :)

foolip:
Alright, merged, I feel like celebrating on such a great Friday. I'll make Taskcluster a required check, that's plenty celebratory.

jugglinmike:
> @jugglinmike other than the 3% failure rate, which I think we can tolerate if it's being improved

3% only describes the performance over the past seven days. We have no reason to expect any improvements yet, so a more accurate picture should include the previously-reported data. That indicates a 6% failure rate.

> is there anything else that would favor continuing with Travis instead of making the Taskcluster stability runs blocking?

I'm not convinced that silently allowing 6% of commits to go unverified is acceptable. The acceptance of gh-14096 shows a difference of opinion. I was hoping to get some clarity around this from the new project governance (it's why [I requested a status update earlier this week](https://lists.w3.org/Archives/Public/public-test-infra/2018OctDec/0015.html)), but it looks like we need to move on this before that's active.

jugglinmike:
Arg, I posted to the wrong thread. My previous message was intended for gh-13194. I'll repost there. Sorry for the confusion!

PR #14110, opened 2018-11-17
[paint-timing] Avoid race condition

jugglinmike:
Previously, this test implicitly relied on "paint" timing entries that
were created by the test harness. In some circumstances, the test may
execute after these entries have passed. Introduce an explicit document
modification that will reliably produce a new entry for the purpose of
WebIDL conformance testing.

### DIFF
--- a/paint-timing/idlharness.window.js
+++ b/paint-timing/idlharness.window.js
@@ -19,6 +19,9 @@ idl_test(
         resolve();
       });
       observer.observe({ entryTypes: ['paint'] });
+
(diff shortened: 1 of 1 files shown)


foolip:
@jugglinmike this conflicts now, can you rebase if there's still a problem on master?

jugglinmike:
[An equivalent change from the Chromium project was accepted earlier this year](https://github.com/web-platform-tests/wpt/pull/15239), so this patch is no longer necessary.

PR #14111, opened 2018-11-17
SVG Text Parsing

ewilligers:
Test that properties support the full grammar in
https://svgwg.org/svg2-draft/text.html

Blink/Firefox/Edge/Safari do not support
- shape-inside
- shape-subtract
- text-decoration-fill
- text-decoration-stroke

Safari/Edge do not support
- inline-size

Edge does not support
- shape-margin


svgeesus (approved):
Good tests for supported properties

PR #14112, opened 2018-11-17
Load an image with redirect: initially same-origin, it is redirected to a cross origin URL which redirects to the first origin.

youennf:


domfarolino:
We'll also eventually want tests for other resources right? Maybe getting a complete list together would be good (though perhaps that already exists somewhere/is easily computable and I just don't know)?

annevk:
I don't think we have a good no-CORS list. `img` (can you read decoded pixels), `script` (can you see exception details), fonts (in some browsers, but arguably that's a bug in those browsers; can you observe glyph details, such as height), style sheets (can you see the rules, iirc), `fetch()` (do you get an opaque response).

And then maybe `embed` and `object` for plugins.

And then there's navigation which is `iframe` and also `embed` and `object` (when they create a browsing context), which is like no-CORS, but I'm pretty sure we only look at the final URL.

annevk:
@bzbarsky @mikewest so Safari manages to throw for no-CORS A -> B -> A, for ``. I suspect you can workaround this with `<iframe>`, but perhaps it's worth doing anyway? Note that `fetch()` also treats it as "opaque" and nobody wanted to change that.

bzbarsky:
> Safari manages to throw for no-CORS A -> B -> A, for ``

Throw when doing what?

annevk:
@bzbarsky cross-origin to same-origin redirect with no CORS involved. So the final URL is same-origin, but it was tainted along the way.

bzbarsky:
So the throw is when trying to read back from a canvas that the resulting image is painted into it or something like that?  Because clearly the load itself is not what's going to throw.

annevk:
Oh sorry, yes, as per the test.

bzbarsky:
So in theory we could keep track of all the redirects involved on the response, or just whether there were cross-site redirects...  I don't have an opinion on whether this is actually useful.

domfarolino:
> or just whether there were cross-site redirects

I think we have this in the form of the [tainted origin](https://fetch.spec.whatwg.org/#concept-request-tainted-origin) flag right?

annevk:
Yeah, I suspect bz was talking about the implementation, which might be different.

Unless someone feels strongly about merging this now I suggest we wait for more tests to figure out the overall story for https://github.com/whatwg/fetch/issues/737 first.

domfarolino:
Chromium [see here](https://github.com/whatwg/fetch/issues/737#issuecomment-444385589) and WebKit on board with this behavior, and currently are implementing it. Are we comfortable moving forward merging this test?

domfarolino (approved):

PR #14132, opened 2018-11-19
[FileAPI] Remove reference to non-existent file

jugglinmike:
The file `FileAPI/url/url-with-tags.window.js` is not currently present
in the repository nor was it present when the reference was introduced
[1]. Remove it to avoid triggering spurious request failures during test
execution.

[1] 9d8a1855092e90b2974453d6181c7f985170deba

### DIFF
--- a/FileAPI/url/sandboxed-iframe.html
+++ b/FileAPI/url/sandboxed-iframe.html
@@ -11,8 +11,9 @@
 const iframe_scripts = [
   'resources/fetch-tests.js',
   'url-format.any.js',
+  'url-in-tags.window.js',
   'url-with-xhr.any.js',
-  'url-with-fetch.any.js'
+  'url-with-fetch.any.js',
 ];
 
 let html = '<!doctype html>\n<meta charset="utf-8">\n<body>\n';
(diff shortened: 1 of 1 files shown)


mkruisselbrink (commented):
ah, good catch. I think I meant that to be url-in-tags.window.js

jugglinmike:
Got it. I've added a reference to that file via another commit. The commit message will have to be updated, but before that, we have an instability on our hands.

Bisecting indicates that it's a wptrunner regression from 31664a8d285f2df3c86f7051c8e2a45a75bbd339, which was merged via gh-13419. [A newer wptrunner patch is currently under review](https://github.com/web-platform-tests/wpt/pull/13966), and as luck would have it, it fixes this problem.

This change should pass the stability checks once that patch is merged.

Ms2ger (approved):

jugglinmike:
@Ms2ger This branch has a commit prefixed with `fixup!`; mind squashing those prior to landing in the future? I'd also be happy to do that and force-push to the branch if that's your preference.

Ms2ger:
Sure; I overlooked the commit here.

Issue #12256, opened 2018-08-01
[HTTP/2.0] H2 Server tests can't be run on Travis


dhdavvie:
HTTP/2.0 requires the ALPN TLS extension, which in turn requires OpenSSL 1.0.2+. Unfortunately, Travis [does not currently support](https://docs.travis-ci.com/user/languages/python/) OpenSSL 1.0.2+ (under the 'Development releases support' heading), which means that the HTTP/2.0 server functional tests cannot be run on Travis currently. I don't really know what can be done about this, suggestions?

jgraham:
So one obvious option is to gate running these tests on moving CI to TaskCluster where we get to choose our own environment.

foolip:
@jugglinmike will be experimenting a bit with running the equivalent of ci_stability on Taskcluster. Mike, is there an issue for that?

stephenmcgruer:
@foolip Can this be closed since we no longer use Travis, or is it still an issue with Taskcluster?

jgraham:
Not an issue with TC, but we haven't updated the default configuration to actually enable H2 yet.

stephenmcgruer:
Ack. I think that is covered by https://github.com/web-platform-tests/wpt/issues/13532, so closing this one. Please re-open if I'm wrong.

PR #14133, opened 2018-11-19
[Azure Pipelines] Run pytest tests (and publish results) on macOS

foolip:
Two changes in this PR:

The decision job running in a Linux VM will decide whether or not jobs
running on other OSes start at all. Most templates are now only used
once, but most ought to be reused when running affected tests.

And run pytest tests and public results using this nifty task:
https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -4,6 +4,7 @@
 # https://docs.microsoft.com/en-us/azure/devops/pipelines/process/multiple-phases
 # https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates
 # https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables
+# https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/index
 #
 # In addition to this configuration file, the "Build pull requests from forks
 # of this repository" setting must also be enabled in the Azure DevOps project:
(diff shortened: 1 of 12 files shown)


foolip:
Pardon the rename, I hit buttons too fast. Also want to see Azure Pipelines go green for the first commit before I push the next.

foolip:
See https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=848&view=logs for proof that the first commit works. (I intend for this PR to be rebased with separate commits.)

foolip:
@Hexcles @lukebjerring requesting your review since it's probably a good idea for more of us to understand this setup when I add affected tests to it.

lukebjerring:
Can confirm I read through, but any review would merely be a rubber stamp. 

Is that publish results tool flexible enough for custom results ( WPT reports ) in future?

foolip:
> Is that publish results tool flexible enough for custom results ( WPT reports ) in future?

If we could convert our results into the JUnit XML format then we could see something cool for `./wpt run` as well. However, I suspect we'd run into some trouble with the assumption that each test is a method and we just have a bunch of strings as our test names. Totally worth experimenting with though, there might be workarounds.

.azure-pipelines.yml
```diff
@@ -10,76 +13,81 @@
...
 
 jobs:
-- job: macOS
-
+- job: root
+  displayName: './wpt test-jobs'
   pool:
-    vmImage: 'macOS-10.13'
-
+    vmImage: 'ubuntu-16.04'
```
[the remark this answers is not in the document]
foolip: Nope, not available: https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=vsts&tabs=yaml#use-a-microsoft-hosted-agent

tools/ci/azure/tox_pytest.yml
```diff
@@ -0,0 +1,18 @@
...
+
+steps:
+- template: pip_install.yml
+  parameters:
+    packages: tox virtualenv
+
+- script: tox -c ${{ parameters.directory }} -e ${{ parameters.toxenv }} -- --junitxml=results.xml
+  displayName: 'Run tests'
+
+- task: PublishTestResults@2
```
Hexcles: Out of curiosity (and I have to admit I haven't read the docs): is the `@2` a magic suffix?

tools/ci/azure/pip_install.yml
```diff
@@ -0,0 +1,9 @@
+parameters:
+  packages: ''
+
+steps:
+- script: |
+    sudo easy_install pip
+    pip install --user ${{ parameters.packages }}
+    echo "##vso[task.prependpath]$HOME/Library/Python/2.7/bin"
```
Hexcles: We are already modifying a bunch of system configs and installing packages to global. What about we just `sudo pip install --system` here? That will save us the trouble of modifying PATH.

tools/ci/azure/tox_pytest.yml
```diff
@@ -0,0 +1,18 @@
+parameters:
+  directory: ''
+  toxenv: 'ALL'
+
+steps:
+- template: pip_install.yml
+  parameters:
+    packages: tox virtualenv
```
Hexcles: Nit: `tox` is enough. It depends on `virtualenv`.

Hexcles (approved):
LGTM % some minor nits & questions

.azure-pipelines.yml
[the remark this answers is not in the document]
gsnedders: Bah. 🙁

For the decision task this doesn't really matter, but for reasons of things like #12256 we probably do want to use the same(-ish?) environment as on TC.

tools/ci/azure/tox_pytest.yml
foolip: It's just the version of the tool, presumably so they can change it without existing content breaking.

tools/ci/azure/pip_install.yml
foolip: Unfortunately tox has a dependency which isn't possible to upgrade on the system because it's somehow protected. setuptools I think. I'll try to make the suggested change and then document why it doesn't work.

tools/ci/azure/tox_pytest.yml
foolip: Nice, thanks!

tools/ci/azure/pip_install.yml
gsnedders: It'll be System Integrity Protection of the system Python.

Hexcles: Oh right... I brew Python on my mac so I forgot there's that...

PR #14144, opened 2018-11-20
Fetch: change Response's statusText's default

annevk:
See https://github.com/whatwg/fetch/pull/836 for context.

This also cleans up the code a bit.


jdm (approved):

annevk:
@jdm thanks for the review!

@Ms2ger I should have added a label, but maybe I can also ask to not merge my PRs as a general rule? I usually try to time them together with a change to the standard, which often awaits some kind of implementer feedback.

PR #14156, opened 2018-11-20
[Azure Pipelines] Run affected tests on Safari Technology Preview

foolip:


tools/wpt/run.py
```diff
@@ -493,6 +495,26 @@ def setup_wptrunner(venv, prompt=True, install_browser=False, **kwargs):
     setup_cls = product_setup[kwargs["product"]](venv, prompt, sub_product)
     setup_cls.install_requirements()
 
+    if kwargs["affected"]:
```
jgraham: `if kwargs.pop("affected")`

tools/wpt/run.py
```diff
@@ -493,6 +495,26 @@ def setup_wptrunner(venv, prompt=True, install_browser=False, **kwargs):
...
+        if not test_list:
+            # TODO: make something show up in all loggers, and also still write wpt_report.json?
+            logger.info("no affected tests")
+            sys.exit(1)
+        test_list = [os.path.relpath(item, wpt_root) for item in test_list]
+        test_list.sort()
+        if len(kwargs["test_list"]):
+            logger.warning("ignoring test list because --affected")
+        kwargs["test_list"] = test_list
+    del kwargs["affected"]
```
jgraham: Remove this with the change to `pop`

tools/wpt/run.py
```diff
@@ -493,6 +495,26 @@ def setup_wptrunner(venv, prompt=True, install_browser=False, **kwargs):
...
+            set(["conformance-checkers", "docs", "tools"]), # TODO: dedupe
+        )
+        test_list = list(tests_changed | tests_affected)
+        if not test_list:
+            # TODO: make something show up in all loggers, and also still write wpt_report.json?
+            logger.info("no affected tests")
+            sys.exit(1)
+        test_list = [os.path.relpath(item, wpt_root) for item in test_list]
+        test_list.sort()
+        if len(kwargs["test_list"]):
```
jgraham: I was thinking it should be the union of the test list and the affected tests. I think that makes some sense for interactive use.

tools/wpt/testfiles.py
```diff
@@ -162,9 +162,8 @@ def exclude_ignored(files, ignore_rules):
 
 
 def files_changed(revish, ignore_rules=None, include_uncommitted=False, include_new=False):
-    """Get and return files changed since current branch diverged from master,
-    excluding those that are located within any path matched by
-    `ignore_rules`."""
+    """Get and return files changed since `revish`, excluding those that are
```
jgraham: This isn't quite right; if revish is a range it will return the files changed in that range rather than since that range.

jgraham (dismissed):

tools/wpt/run.py
foolip: Rather `revish = kwargs.pop("affected")` and `if revish` I guess?

tools/wpt/run.py
foolip: Yeah, that's good, saves me the trouble of writing a good error message.

tools/wpt/testfiles.py
foolip: Yeah, you're right. It doesn't make sense to have anything but HEAD as the end of the range because affected tests are determined using the checkout (working tree = HEAD + modified + new) but you're still right. I'll revert this and leave it alone, again. What I'd prefer is https://github.com/web-platform-tests/wpt/pull/14093 and changing everything that suggests that revish could be a range to taking about a single revision, and possibly even asserting that it is so even though A..B will happen to work with `git diff`.

foolip:
In order to verify that the "affected tests without changes" job even works before merging this PR, the `--affected` flag has to be merged first. I'll split that into a separate PR when it seems like it will work.

tools/wpt/run.py
```diff
@@ -487,6 +489,19 @@ def setup_wptrunner(venv, prompt=True, install_browser=False, **kwargs):
...
     setup_cls.install_requirements()
 
+    affected_revish = kwargs.pop("affected", None)
+    if affected_revish is not None:
+        files_changed, _ = testfiles.files_changed(affected_revish, include_uncommitted=True, include_new=True)
+        # TODO: honor --no-manifest-update and other manifest args
+        tests_changed, tests_affected = testfiles.affected_testfiles(files_changed)
+        test_list = tests_changed | tests_affected
+        if not test_list:
+            # TODO: make something show up in all loggers, and also still write wpt_report.json?
```
foolip: @jgraham this is the main outstanding part that I need advice on. This code with the default logger doesn't print anything or create a wpt_report.json if --log-wptreport is used. Where is the code that decides to run all tests if none are given? I think there, is `--affected` is used, would be the place to terminate in a way that achieves what the TODO asks for.

And how do I make the default logger print things?

gsnedders: `--log-mach -` should make us get the default output again, at least

foolip: That's fine for CI, but when people run `./wpt run --affected HEAD^1` or similar locally, I'd like there to be clear if no tests are run, instead of silently exiting.

jgraham: I think this is a generic problem with the default logger. I think the right solution is to allow log messages until a flag is set in the logger and then turn them off once the run starts (since the point of the logger is to be as non-verbose as possible). I don't know if it's worthwhile to do something special here instead of just fixing the generic problem.

.azure-pipelines.yml
```diff
@@ -26,6 +25,71 @@ jobs:
...
+  - template: tools/ci/azure/pip_install.yml
+    parameters:
+      packages: virtualenv
+  - template: tools/ci/azure/install_fonts.yml
+  - template: tools/ci/azure/install_certs.yml
+  - template: tools/ci/azure/install_safari.yml
+  - template: tools/ci/azure/update_hosts.yml
+  - script: no_proxy='*' ./wpt run --yes --no-pause --no-fail-on-unexpected --no-restart-on-unexpected --log-wptreport $(Build.ArtifactStagingDirectory)/wpt_report.json --channel preview safari --affected HEAD@{1}
+    displayName: 'Run tests'
+  - task: PublishBuildArtifacts@1
```
Hexcles: `tools/ci/azure/tox_pytest.yml` uses `PublishTestResults@2`. Is the difference intentional?

tools/wpt/testfiles.py
```diff
@@ -199,9 +199,10 @@ def load(manifest_path=None):
 
 load_manifest = _init_manifest_cache()
```
Hexcles: Nit: two blank lines before top-level class/function definitions is recommended (by PEP8 I think).

tools/wpt/run.py
```diff
@@ -487,6 +489,19 @@ def setup_wptrunner(venv, prompt=True, install_browser=False, **kwargs):
     setup_cls = product_setup[kwargs["product"]](venv, prompt, sub_product)
     setup_cls.install_requirements()
 
+    affected_revish = kwargs.pop("affected", None)
+    if affected_revish is not None:
+        files_changed, _ = testfiles.files_changed(affected_revish, include_uncommitted=True, include_new=True)
```
Hexcles: This is different from the current behaviour of `check_stability`. Namely,

1. `include_uncommitted=True`: why? I think this won't make a difference on CI, but is this an intentional change for local interactive use?
2. `include_new=True`
3. `ignore_changes` isn't taken into into account: this might be a separate concern (the tests *are* affected, but we are unable to run them all because of resource constraints). Besides, IIUC, although the setup in this PR doesn't run stability checks, it'll still attempt to run all testharness.js tests if `testharness.js` is modified, which will still take too long even w/o `--verify`.

tools/wpt/testfiles.py
Hexcles: The original docstring is confusing (inaccurate?), too. If we fail to summarize the behaviour of `git diff` in various different cases, shall we just expose the implementation detail by explicitly saying `git diff [revish]` is what we do under the hood? I think that'd be less bad.

Hexcles (commented):
I very much agree that the `--affected` part of your PR should be split into a separate PR. I can run affected tests without the patch using your split out PR on Taskcluster, which will not only help to verify that part of your PR (with fewer variables) but also reduce duplicated code in my work towards https://github.com/web-platform-tests/wpt.fyi/issues/708.

.azure-pipelines.yml
foolip: Yeah, different tasks are versioned independently.

tools/wpt/run.py
foolip: Yes, I the `=True` args are for local use, although I didn't test them.

testharness.js is excluded by the affected test logic so modifying that actually doesn't cause all those tests to be run. idlharness.js however is not excluded.

tools/wpt/testfiles.py
foolip: Any idea why flake8 doesn't complain about this?

gsnedders: Because we disabled that, because @jgraham objected to it, viewing it as unnecessary noise.

.azure-pipelines.yml
```diff
@@ -26,6 +25,71 @@ jobs:
...
+  - template: tools/ci/azure/pip_install.yml
+    parameters:
+      packages: virtualenv
+  - template: tools/ci/azure/install_fonts.yml
+  - template: tools/ci/azure/install_certs.yml
+  - template: tools/ci/azure/install_safari.yml
+  - template: tools/ci/azure/update_hosts.yml
+  - script: no_proxy='*' ./wpt run --yes --no-pause --no-fail-on-unexpected --no-restart-on-unexpected --log-wptreport $(Build.ArtifactStagingDirectory)/wpt_report.json --channel preview safari --affected HEAD^1
+    displayName: 'Run tests'
+  - task: PublishBuildArtifacts@1
```
Hexcles: According to the [doc](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/publish-build-artifacts?view=vsts), we can explicitly specify `pathtoPublish` to a *file* instead of the default directory. Can we try that?

foolip: Sure, I'll try and see what happens

.azure-pipelines.yml
```diff
@@ -26,6 +25,71 @@ jobs:
...
+    displayName: 'Publish results'
```
lukebjerring:
```suggestion
    displayName: 'Publish results'
    pathtoPublish: '$(Build.ArtifactStagingDirectory)/wpt_report.json'
    artifactName: 'report'
```

.azure-pipelines.yml
```diff
@@ -26,6 +25,77 @@ jobs:
...
+  - script: git checkout HEAD^1
+    displayName: 'Checkout target branch'
+  - template: tools/ci/azure/pip_install.yml
+    parameters:
+      packages: virtualenv
+  - template: tools/ci/azure/install_fonts.yml
+  - template: tools/ci/azure/install_certs.yml
+  - template: tools/ci/azure/install_safari.yml
+  - template: tools/ci/azure/update_hosts.yml
+  - script: no_proxy='*' ./wpt run --yes --no-pause --no-fail-on-unexpected --no-restart-on-unexpected --affected 'HEAD@{1}' --log-wptreport $(Build.ArtifactStagingDirectory)/wpt_report.json --channel preview safari
```
Hexcles: I can run this command just fine locally (of course, not using Safari), but on Azure it's complaining about invalid arguments:

```
$ no_proxy='*' ./wpt run --yes --no-pause --no-fail-on-unexpected --no-restart-on-unexpected --affected 'HEAD@{1}' --log-wptreport /Users/vsts/agent/2.142.1/work/1/a/wpt_report.json --channel preview safari
wpt run: error: unrecognized arguments: --affected safari
```

I don't really understand why this fails (while the "with change" counterpart succeeds).

foolip: That's because the argument didn't exist before this PR :) Will work when the argument is landed on master first.

Hexcles:
Lol of course, somehow "without changes" didn't cross my mind :)

foolip:
This seems to work now. One weird thing is that when no tests are affected, an empty file is published as the report, but IIUC that's also the case for Taskcluster. @Hexcles @lukebjerring are either of you tracking a fix for that?

The other issue is that the manifest is generated from scratch, so the job is not as fast as it could be. Making it fast actually does matter, as the results show up on wpt.fyi and being able to cycle in tens of seconds or at least less than one minute seems both doable and very valuable. Adding 3+ minutes to that means one has to switch context for each try.

lukebjerring:
Yah I noticed the artifacts get created either way. The receiver handles empty files cleanly though, so it doesn't break anything (see https://github.com/web-platform-tests/wpt.fyi/pull/876)

jgraham (approved):

jgraham:
I'm not sure why the manifest isn't being downloaded; running `wpt manifest` should try to download it?

foolip:
@jgraham I can add a `./wpt manifest` step first, and see if that results in an incremental update in `./wpt run`. I think the code path taken there just doesn't have the "download or generate" bits? Is there an issue somewhere about consolidating the entry points to ensuring manifest existence?

jgraham:
So ` - template: tools/ci/azure/update_manifest.yml` should run it; I missed that this job doesn't call that. The run frontend should end up doing the same thing, but it's possibly more broken there? Anyway I think just adding the other template to the steps will fix this?

Hexcles (changes requested):

Hexcles:
Confirmed that it is WAI: https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=1862&view=results

The artifacts also look good. Woohoo!

Hexcles (approved):
LGTM

foolip:
Thanks everyone! Merging now.

foolip:
Oops, blocked by https://github.com/web-platform-tests/wpt/issues/14448.

foolip:
Now fixed, merging!

lukebjerring:
@foolip - for Azure it's fine, but not for Taskcluster - we should open an issue for a report being produced when no tests are affected, as this causes permanently-pending wpt.fyi checks.

PR #14172, opened 2018-11-21
[streams] Refactor to multi-global test format

### LINKED ISSUE #10597: Clean up our custom multi-global solution for streams tests
There's a new way to do things!!

See http://web-platform-tests.org/writing-tests/testharness.html

/cc @ricea @MattiasBuelens

We'll probably need to update my https://github.com/domenic/wpt-runner to preserve our Node.js workflow, but that's a small price to pay.

MattiasBuelens:
This changes all tests for streams to the multi-global `.any.js` format, and removes the old test wrappers.

Previously, the streams tests had to use custom test wrappers, because one of the tools used by the spec authors did not support auto-generating the necessary test boilerplate. This has been fixed in domenic/wpt-runner#12, so we can finally migrate the tests to the standard format. 😄 

Fixes #10597


MattiasBuelens:
Hmm, I guess I should have run `wpt lint` first. Travis complains about lint errors:
```
ERROR:lint:streams/piping/error-propagation-forward.any.js:143: setTimeout used; step_timeout should typically be used instead (SET TIMEOUT)
```
I guess we can no longer use `setTimeout` inside tests? 😕

jgraham:
Presumably you moved files and a rule in lint.whitelist that previously matched now doesn't. But it would be good to remove the setTimeout use if possible.

MattiasBuelens:
@jgraham You're right, the paths changed, so I had to update the whitelist.

I'll see if switching to `step_timeout` "just works", then we can probably remove these from the whitelist. 🤞

MattiasBuelens:
Good news, it just works! 😄

streams/piping/close-propagation-forward.any.js
```diff
@@ -166,7 +163,7 @@ promise_test(() => {
 
   const pipePromise = rs.pipeTo(ws);
 
-  setTimeout(() => rs.controller.close());
+  step_timeout(() => rs.controller.close());
```
ricea: Technically since this is not a single-page test, you should use t.step_timeout(). See https://web-platform-tests.org/writing-tests/testharness-api.html#timeouts-in-tests.

However, it doesn't seem to make any actual difference.

ricea (approved):
lgtm

You no longer need the call to `done()` at the bottom of each js file. But it's not doing any harm.

streams/piping/close-propagation-forward.any.js
MattiasBuelens: If I understand correctly, the `step_*` functions ensure that any `assert_*` calls are scoped to the correct test. However, for these tests, we don't do any assertions inside these timeout callbacks, so it doesn't really matter.

For good measure, I'll update the tests to use `t.step_timeout` anyway. There's one usage in `resources/rs-utils.js` where there's no obvious reference to a `Test` object, so there I'll keep the global `step_timeout`.

MattiasBuelens:
Something went wrong with Taskcluster for Firefox Nightly? 😕 Looks like `git fetch` failed, [see log](https://tools.taskcluster.net/groups/eoQ_bG6xS7i1ieoACVGGJQ/tasks/T5WjoAYiSkqsBOqWXMgTKw/runs/0/logs/public%2Flogs%2Flive.log#L22):
```
++ git remote add origin https://github.com/web-platform-tests/wpt.git
++ git fetch --quiet --depth=50 origin refs/pull/14172/merge
fatal: The remote end hung up unexpectedly
[taskcluster 2018-11-22 10:26:23.174Z] === Task Finished ===
[taskcluster 2018-11-22 10:26:23.281Z] Unsuccessful task run with exit code: 128 completed in 3.829 seconds
```

jgraham:
I closed and reopened as that's the only way to retrigger taskcluster until https://bugzilla.mozilla.org/show_bug.cgi?id=1509083 is fixed (reruns have complex semantics in general and people tried to deprecate them as a result, but they're what we need in this case, so…).

PR #14182, opened 2018-11-22
[cookies] Add test for default path calculation

jugglinmike:
@mikewest This is the cause of many of [the failures reported by wpt.fyi for `cookies/http-state` in Firefox](https://wpt.fyi/results/cookies/http-state?sha=e340f910d3&labels=taskcluster,firefox), but it's difficult to understand the problem in those terms.

browser    | status
-----------|-------
Chrome 70  | :heavy_check_mark:
Edge 15    | :x:
Firefox 63 | :x:
IE 11      | :x:
Safari 12  | :heavy_check_mark:

### DIFF
--- a/cookies/http-state/resources/cookie-http-state-template.js
+++ b/cookies/http-state/resources/cookie-http-state-template.js
@@ -86,6 +86,13 @@ CookieManager.prototype.resetCookies = () => {
     expireCookie(cookies_to_delete[i].replace(/=.*$/, ""),
                  /*expiry_date=*/null,
                  /*path=*/'/');
+    // Some browsers incorrectly include the final "forward slash" character
+    // when calculating the default path. The expected behavior for default
+    // path calculation is verified elsewhere; this utility accommodates the
+    // non-standard behavior in order to improve the focus of the test suite.
+    expireCookie(cookies_to_delete[i].replace(/=.*$/, ""),
+                 /*expiry_date=*/null,
+                 /*path=*/getLocalResourcesPath() + "/");
   }
 }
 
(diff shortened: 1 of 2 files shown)


jugglinmike:
We should probably accommodate the aberrant behavior in the `http-state` tests since they are not intended to verify default path calculation. I've added a commit to do that.

Here are the results for that directory in Firefox on `master` today (i.e. `./wpt run --no-restart-on-unexpected firefox cookies/http-state`)

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 232 checks (11 tests, 221 subtests)
    Expected results: 19
    Unexpected results: 213
      subtest: 213 (213 fail)

Restarting after failures alleviates the problem to some degree, though not completely and at the cost of a slower time-to-results (`./wpt run firefox cookies/http-state`)

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 232 checks (11 tests, 221 subtests)
    Expected results: 77
    Unexpected results: 155
      subtest: 155 (155 fail)

Firefox fares much better with this patch applied (`./wpt run firefox cookies/http-state`)

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 232 checks (11 tests, 221 subtests)
    Expected results: 169
    Unexpected results: 63
      subtest: 63 (63 fail)

...and those improvements persist even when running without restarting after failures (`./wpt run --no-restart-on-unexpected firefox cookies/http-state`)

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 232 checks (11 tests, 221 subtests)
    Expected results: 169
    Unexpected results: 63
      subtest: 63 (63 fail)

mikewest (approved):
LGTM, thank you for digging into this!

PR #14193, opened 2018-11-22
ReadableStream: modify tests for pipeThrough() having a brand check

ricea:
Remove tests that ReadableStream.prototype.pipeThrough operates
generically on its arguments. Add tests that it performs brand checks on
|this|, |readable| and |writable|, and throws for other precondition
failures.

Also verify that it *doesn't* call pipeTo().

Also verify that preventClose, preventCancel and preventAbort work,
since we can no longer test that it passes through to pipeTo().

Also change the expected length of the function from 2 to 1 to reflect
that the options argument is optional.

### DIFF
--- a/streams/piping/pipe-through.any.js
+++ b/streams/piping/pipe-through.any.js
@@ -192,7 +192,7 @@ test(() => {
   rs.getReader();
   assert_throws(new TypeError(), () => rs.pipeThrough(uninterestingReadableWritablePair()),
                 'pipeThrough should throw');
-}, 'pipeThrough should throw if |this| is locked');
+}, 'pipeThrough should throw if this is locked');
 
 test(() => {
   const rs = new ReadableStream();
--- a/streams/piping/pipe-through.any.js
+++ b/streams/piping/pipe-through.any.js
@@ -85,7 +85,7 @@ test(t => {
   });
   ReadableStream.prototype.pipeTo = () => {
     calledFakePipeTo = true;
-      };
+  };
   const rs = new ReadableStream();
   const readable = new ReadableStream();
   const writable = new WritableStream();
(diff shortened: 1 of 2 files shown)


streams/piping/pipe-through.any.js
```diff
@@ -56,167 +57,97 @@ promise_test(() => {
...
-  methodThrows(ReadableStream.prototype, 'pipeThrough', thisValue, [undefined, {}]);
-  methodThrows(ReadableStream.prototype, 'pipeThrough', thisValue, [null, {}]);
+  assert_false(calledFakePipeTo, 'the monkey-patched pipeTo should not have been called');
+  assert_equals(result, readable, 'return value should be the passed readable property');
 
-}, 'pipeThrough should throw when its first argument is not convertible to an object');
+}, 'pipeThrough should not call pipeTo on the ReadableStream prototype');
 
-test(() => {
+const badReadables = [null, undefined, 0, NaN, true, 'ReadableStream', Object.create(ReadableStream.prototype)];
```
domenic: May make more sense to put this in brand-checks.js. Especially if we plan on eventually switching to idlharness.js, that makes it easier to nuke that file all at once.

streams/piping/pipe-through.any.js
```diff
@@ -235,22 +166,90 @@ test(() => {
...
-    pipeTo() {
-      ++count;
-    }
-  };
+test(() => {
+  const rs = new ReadableStream();
+  rs.getReader();
+  assert_throws(new TypeError(), () => rs.pipeThrough(uninterestingReadableWritablePair()),
+                'pipeThrough should throw');
+}, 'pipeThrough should throw if |this| is locked');
```
domenic: Nit, I don't think we use |this| notation much in the tests

domenic (approved):

PR #14200, opened 2018-11-22
[selectors] :enabled and :disabled CSS selectors

ewilligers:
:enabled matches UI elements that are in an enabled state.
:disabled matches UI elements that are in a disabled state.
https://drafts.csswg.org/selectors-4/#enableddisabled

### DIFF
--- a/css/selectors/invalidation/enabled-disabled.html
+++ b/css/selectors/invalidation/enabled-disabled.html
@@ -1,0 +1,51 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>CSS Selectors Invalidation: :enabled and :disabled</title>
+    <link rel="help" href="https://drafts.csswg.org/selectors-4/#enableddisabled">
+    <link rel="help" href="https://html.spec.whatwg.org/#enabling-and-disabling-form-controls:-the-disabled-attribute">
+    <meta name="assert" content="This tests that the :enabled and :disabled selectors are effective, and that the enabled/disabled status of an element is changed when updating its 'disabled' attribute by script.">
+    <script src="/resources/testharness.js"></script>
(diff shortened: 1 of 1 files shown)


frivoal (approved):
LGTM.

In addition to testing that the selectors work, this test is also testing that the IDL attribute reflects the content attribute. That is required by the HTML spec, so the test is correct, but since that's about half of what the test does, I think saying so in the meta assert, and/or linking to the relevant bit of the HTML spec (https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#enabling-and-disabling-form-controls:-the-disabled-attribute) would be good.

That said, that doesn't change the validity of the test, so approving anyway.

css/selectors/invalidation/enabled-disabled.html
```diff
@@ -0,0 +1,50 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>CSS Selectors Invalidation: :enabled and :disabled</title>
+    <link rel="help" href="https://drafts.csswg.org/selectors-4/#enableddisabled">
```
frivoal:
```suggestion
    <link rel="help" href="https://drafts.csswg.org/selectors-4/#enableddisabled">
    <link rel="help" href="https://html.spec.whatwg.org/#enabling-and-disabling-form-controls:-the-disabled-attribute">
```

css/selectors/invalidation/enabled-disabled.html
```diff
@@ -0,0 +1,50 @@
...
+    <meta name="assert" content="This tests that the :enabled and :disabled selectors are effective">
```
frivoal:
```suggestion
    <meta name="assert" content="This tests that the :enabled and :disabled selectors are effective, and that the enabled/disabled status of an element is changed when updating its 'disabled' attribute by script.">
```

Issue #13465, opened 2018-10-11
Arbitrary subdomains


annevk:
In order to test what code points are allowed in domain name labels (and how they end up in `Host`, `Origin`, and `Referrer` headers) I need arbitrary subdomains. (See https://github.com/httpwg/http-extensions/issues/702.)

This would also make #4880, #8581, and #11219 easier as we'd simply route all requests.

The flipside is that this would make deployment of wpt more involved as it requires running dnsmasq (or equivalent) plus configuring some OS-level routing rules.

It does seem very important to test this, but I can see an argument for testing this in isolation.

(If https://tools.ietf.org/html/draft-west-let-localhost-be-localhost became a thing however this would be very easy as we could switch to localhost for all things and not even require `/etc/hosts` manipulation.)

cc @mikewest @mnot

foolip:
@jugglinmike

jgraham:
The localhost thing wouldn't help with running this on w3c-test.org or equivalent. Setting up custom DNS in browser CI infrastructure also seems daunting given that typically test jobs don't have root access. Maybe DoH could help here if that gets widespead adoption? I'm not sure.

jgraham:
I think it solves the problem for Chromium but not for other browsers :) It's probably possible to make development builds of Firefox do something similar, but I'm not sure that there's an equivalent workaround available for e.g. testing stable releases of Safari, or mobile browsers where you actually don't want localhost but the address of the host machine.

mikewest:
I wasn't at all clear, apologies. In my head, I added a sentence like "Other browsers should be able to do the same without much effort."

But I didn't think about testing against stable releases, and you're entirely correct that that would require more work. We can either put that work on browser vendors (by designing and implementing a webdriver flag for this kind of wildcarding), or put it on folks who host testing infrastructure (e.g. by requiring W3C to add a wildcard DNS record to `w3c-test.org`).

jugglinmike:
@foolip mentioned me because I'm maintaining http://web-platform-tests.live, an installation we'd like to replace w3c-test.org. I can speak to some of the deployment concerns, but I'm new to this domain (note the fantastic pun), so please bear with me.

[...]

PR #14206, opened 2018-11-23
Add name constraints to CA

### LINKED ISSUE #11075: Is the Certificate Authority name-constrained somehow?
I wonder if the Certificate Authority we have to add to root stores to work with web-platform-tests is name-constrained somehow or if an attacker can use it mint certificates.

franziskuskiefer:
Something like this should fix issue #11075 (fixes #11075).
I don't know how to run this so I can't tell for sure. But adding something of the form `nameConstraints=permitted;DNS:.example.com,permitted;DNS:.example.org` will give you what you want.


annevk:
Thanks @franziskuskiefer!

@jgraham @gsnedders @jugglinmike would you be willing to iterate on this further?

tools/wptserve/wptserve/sslutils/openssl.py
```diff
@@ -213,6 +218,7 @@ def get_config(root_dir, hosts, duration=30):
 subjectKeyIdentifier=hash
 authorityKeyIdentifier=keyid:always,issuer:always
 keyUsage = keyCertSign
+%(constraints_line)s
```
mozfreddyb: Drive-by comment: This line above uses a variable that is local to the formatted string, which needs to be defined when formatting the string. This is named as the same variable as in the outer code block, but not implicitly linked.
To make this work, the in-line dictionary after the formatted string would need an additional key/value pair of

`"constraints_line": constraints_line,`

foolip:
@gsnedders, I assigned to you since https://github.com/web-platform-tests/wpt/issues/11075 is on our list of important infra issues.

jugglinmike:
We recently received [a feature request to support arbitrary subdomains](https://github.com/web-platform-tests/wpt/issues/13465). We're not doing that (yet), but it might be wise to account for the possibility in this patch by relaxing the constraints a bit. As far as I know (which is not very far), it will be no less secure to constrain to:

- `web-platform.test`
- `.web-platform.test`
- `not-web-platform.test`
- `.not-web-platform.test`

Though now I'm wondering if we could simply constrain to `.test`.

gsnedders:
@jugglinmike 

> Though now I'm wondering if we could simply constrain to `.test`.

I'd strongly oppose this, given I (and probably others) have other local dev environments served on `.test`.

jgraham:
I don't think the arbitary subdomains thing needs to be considered here; if we change the range of supported hosts it will in any case be necessary to update the SAN line on the certs so there is barely any advantage in preempting that change for the constraints line only.

foolip:
Ping @gsnedders for review.

tools/wptserve/wptserve/sslutils/openssl.py
```diff
@@ -114,16 +114,18 @@ def make_subject(common_name,
...
+def make_name_constraints(hosts):
+    return ",".join("permitted;DNS:%s" % host for host in hosts)
 
 def get_config(root_dir, hosts, duration=30):
     if hosts is None:
         san_line = ""
+        constraints_line = ""
     else:
         san_line = "subjectAltName = %s" % make_alt_names(hosts)
+        constraints_line = "nameConstraints = " + make_name_constraints(hosts)
```
gsnedders: This will lead to trying to put `nameConstraints` on the server cert, rather than the CA cert.

From `_generate_ca` we call `get_config` with `hosts=None`:

```
  web-platform-tests/tools/wptserve/wptserve/sslutils/openssl.py(399)_generate_host_cert()
-> self._generate_ca()
 web-platform-tests/tools/wptserve/wptserve/sslutils/openssl.py(342)_generate_ca()
-> with self._config_openssl(None) as openssl:
  web-platform-tests/tools/wptserve/wptserve/sslutils/openssl.py(43)__enter__()
-> f.write(get_config(self.base_path, self.hosts, self.duration))
```

(interestingly `pdb` seems to be omitting `_config_openssl` in that!)

As such, this doesn't actually fix the underlying issue! I think we need to have a `is_ca` arg here?

gsnedders (changes requested):
@foolip I actually started on this yesterday, but couldn't figure out why it wasn't working. :)

@jgraham @franziskuskiefer This PR should probably also update `tools/certs`, preferably also providing some documentation on how to update them!

gsnedders:
@jgraham what do I need to do to update `tools/certs`? I presume I also need to make it have longer validity?

jgraham (approved):

jgraham:
Updating the certs is a bit of a black art at the moment. You need to change the wptserve config to use openssl as the config type and add a duration key to the openssl part of the config with the desired duration. Then if you run the server you can copy the generated files out of the temporary location and into the irectory containing the pregenerated certs.

foolip:

gsnedders (approved):

tools/certs/.gitignore
```diff
@@ -0,0 +1,3 @@
+/0*.pem
```
jgraham: Add this to the root .gitignore please (some tooling we have only works with the root).

jgraham (approved):
Subject to .gitignore change.

jgraham (approved):

Issue #14207, opened 2018-11-23
Taskcluster failured due to memory "Cannot allocate memory"


foolip:
The [7 most recent commits to master](https://github.com/web-platform-tests/wpt/commits/bf2bef64b0056d8641ba1caf65315fa28221add9) at the time of writing all have failed Taskcluster runs:
https://tools.taskcluster.net/groups/f9h2rnpNSGSZtf8XqKsI7g
https://tools.taskcluster.net/groups/PEZ79dY9R2GhxheZKTXC3Q
https://tools.taskcluster.net/groups/MEBnx5kARYKxgH42Ot3zsQ
https://tools.taskcluster.net/groups/HiCTerPsTsGZXTaF_Gz4qQ
https://tools.taskcluster.net/groups/Ekt9SdPFT2umvUiNaiTaWQ (for master)
https://tools.taskcluster.net/groups/DWgMz-OFQOKp5iktSe1gSg (for epochs/daily)
https://tools.taskcluster.net/groups/BotCYMVpT--o_Kz1D99puQ
https://tools.taskcluster.net/groups/DevtLTz6TGeBFy8QOFQh8A

All of them have some task that has failed with "Cannot allocate memory" in the log.
<details>
<summary>Example from the first above:</summary>

```
Traceback (most recent call last):
  File "./wpt", line 5, in <module>
    wpt.main()
  File "/home/test/web-platform-tests/tools/wpt/wpt.py", line 129, in main
    rv = script(*args, **kwargs)
  File "/home/test/web-platform-tests/tools/wpt/run.py", line 540, in run
    rv = run_single(venv, **kwargs) > 0
  File "/home/test/web-platform-tests/tools/wpt/run.py", line 547, in run_single
    return wptrunner.start(**kwargs)
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/wptrunner.py", line 331, in start
    return not run_tests(**kwargs)
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/wptrunner.py", line 188, in run_tests
    env_extras) as test_environment:
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/environment.py", line 61, in __init__
    self.cache_manager = multiprocessing.Manager()
  File "/usr/lib/python2.7/multiprocessing/__init__.py", line 99, in Manager
    m.start()
  File "/usr/lib/python2.7/multiprocessing/managers.py", line 524, in start
    self._process.start()
[...]

PR #14208, opened 2018-11-23
Revert "Find manifest for download by tags instead of commits"

foolip:
Possible but unlikely cause of https://github.com/web-platform-tests/wpt/issues/14207.

This reverts commit 3657828ff4243fcc2f3b3fa4e7bed34d646ad565.


foolip:
Using admin powers to merge immediately.

PR #14209, opened 2018-11-23
Reland "Find manifest for download by tags instead of commits"

foolip:
Reverted in https://github.com/web-platform-tests/wpt/pull/14208
as possible cause of https://github.com/web-platform-tests/wpt/issues/14207.

This reverts commit 5438f72ea0bc2abb3754b2b906a34d497bd8457c.

### DIFF
--- a/tools/manifest/download.py
+++ b/tools/manifest/download.py
@@ -33,46 +33,53 @@ def should_download(manifest_path, rebuild_time=timedelta(days=5)):
     return False
 
 
-def git_commits(repo_root):
+def merge_pr_tags(repo_root, max_count=50):
     git = Git.get_func(repo_root)
-    return [item for item in git("log", "--format=%H", "-n50").split("\n") if item]
+    tags = []
+    for line in git("log", "--format=%D", "--max-count=%s" % max_count).split("\n"):
+        for ref in line.split(", "):
+            if ref.startswith("tag: merge_pr_"):
+                tags.append(ref[5:])
+    return tags
 
 
-def github_url(commits):
-    try:
-        resp = urlopen("https://api.github.com/repos/web-platform-tests/wpt/releases")
-    except Exception:
-        return None
+def github_url(tags):
+    for tag in tags:
+        url = "https://api.github.com/repos/web-platform-tests/wpt/releases/tags/%s" % tag
+        try:
+            resp = urlopen(url)
+        except Exception:
+            logger.warning("Fetching %s failed" % url)
+            continue
 
-    if resp.code != 200:
-        return None
+        if resp.code != 200:
+            logger.warning("Fetching %s failed; got HTTP status %d" % (url, resp.code))
+            continue
 
-    try:
-        releases = json.load(resp.fp)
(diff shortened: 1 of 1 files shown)


foolip:
Putting up this PR so that I don't forget it later. Should wait until it's clear if the revert fixed the problem or not before relanding.

tools/manifest/download.py
```diff
@@ -33,46 +33,53 @@ def should_download(manifest_path, rebuild_time=timedelta(days=5)):
     return False
 
 
-def git_commits(repo_root):
+def merge_pr_tags(repo_root, max_count=50):
```
foolip: This can't be relanded as-is even if it wasn't the cause of the failures, because tags aren't fetched in the Taskcluster setup: https://github.com/web-platform-tests/wpt/issues/14207#issuecomment-441189804

foolip:
@jgraham @lukebjerring the original PR is a case where a full Taskcluster run that mimics as closely as possible the master-triggered runs would have been very valuable. What are the steps required to make that happen? Seems like something we should capture in [WPT 2019 priorities](https://docs.google.com/document/d/1UE2KB7gvaEw5gvp4aAQNS9TrFYmnzgDzh5Sr4LQqaQo/edit?usp=sharing).

lukebjerring:
Basically, needs to run on a fork with Taskcluster enabled (always instead of current master-only). I'll do that for you.
I'll also whitelist you for the wpt.fyi checks, which won't be full runs, but still helpful.

Note that the results are already being scraped in spite of it not being surfaced, e.g. https://staging.wpt.fyi/?sha=79cdb98dc9f9566cb10165870a5afe07a5ccd9df

foolip:
If we wanted a mechanism to request a full run from a PR in this repo, what are the constraints? I guess adding/removing labels wouldn't work, only updating branches would?

If Taskcluster had deep integration with the new checks API, then an action (button) that says "request full run" would be neat, but I doubt it'd be high on the Taskcluster team's priorities.

lukebjerring:
I'm not familiar with TaskCluster's capabilities, but if we can branch based on a PR's labels, that would be nice and clean - we add "if the PR is labelled `run-me`" to the set of conditions.
If not, and until TaskCluster supports ad-hoc running, we could consider having some kind of "full-taskcluster-runs/wpt" repo fork and creating an equivalent PR with the hack, using a checks action.

lukebjerring:
You can see the results of the full runs here:
https://github.com/web-platform-tests/wpt/pull/14212/checks?check_run_id=34456218

foolip:
Thanks! Clearly relanding this would break things again, so definitely I shouldn't do that yet :)

gsnedders:
@foolip we have #13263 filed for that

jgraham:
We now fetch tags in taskcluster since #14315 landed, so I think this can land.

foolip:
Yay! @jgraham, feel free to approve and merge, at a time when you can keep on eye on master to see if it broke everything again.

jgraham (approved):

PR #14211, opened 2018-11-23
SVG Painting: computed property values

ewilligers:
Keyword values are as specified.
Lengths are converted to absolute lengths.
Urls are converted to absolute paths.

Found bug:
https://bugs.chromium.org/p/chromium/issues/detail?id=908058
*-opacity should be clamped to [0.1]

### DIFF
--- a/svg/painting/parsing/fill-computed.svg
+++ b/svg/painting/parsing/fill-computed.svg
@@ -2,7 +2,7 @@
 <svg xmlns="http://www.w3.org/2000/svg"
      xmlns:h="http://www.w3.org/1999/xhtml"
      width="800px" height="800px">
-  <title>SVG Geometry Properties: getComputedValue().fill</title>
+  <title>SVG Painting: getComputedValue().fill</title>
   <metadata>
     <h:link rel="help" href="https://svgwg.org/svg2-draft/painting.html#FillProperty"/>
     <h:meta name="assert" content="fill computed value is as specified, with url values absolute."/>
@@ -26,7 +26,8 @@ test(() => {
   const target = document.getElementById('target');
   target.style['fill'] = 'url("a.b#c")';
   const result = getComputedStyle(target)['fill'];
-  assert_true(result.includes('/a.b#c'), result + ' is absolute');
+  const resolved = new URL("a.b#c", document.URL).href;
+  assert_equals(result, 'url("' + resolved + '")');
 }, 'url values are made absolute');
 
   ]]></script>
--- a/svg/painting/parsing/fill-computed.svg
+++ b/svg/painting/parsing/fill-computed.svg
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg"
+     xmlns:h="http://www.w3.org/1999/xhtml"
+     width="800px" height="800px">
+  <title>SVG Geometry Properties: getComputedValue().fill</title>
+  <metadata>
+    <h:link rel="help" href="https://svgwg.org/svg2-draft/painting.html#FillProperty"/>
+    <h:meta name="assert" content="fill computed value is as specified, with url values absolute."/>
+  </metadata>
+  <g id="target"></g>
+  <h:script src="/resources/testharness.js"/>
+  <h:script src="/resources/testharnessreport.js"/>
+  <h:script src="/css/support/computed-testcommon.js"/>
+  <script><![CDATA[
+
+test_computed_value("fill", "none");
+test_computed_value("fill", "rgb(12, 34, 56)");
+
+test_computed_value("fill", 'url("https://example.com/")');
+test_computed_value("fill", 'url("https://example.com/") none');
+test_computed_value("fill", 'url("https://example.com/") rgb(12, 34, 56)');
+
+// context-fill and context-stroke are not yet supported by browsers.
+
+test(() => {
+  const target = document.getElementById('target');
+  target.style['fill'] = 'url("a.b#c")';
+  const result = getComputedStyle(target)['fill'];
+  assert_true(result.includes('/a.b#c'), result + ' is absolute');
+}, 'url values are made absolute');
+
+  ]]></script>
+</svg>
--- a/svg/painting/parsing/stroke-computed.svg
+++ b/svg/painting/parsing/stroke-computed.svg
@@ -2,7 +2,7 @@
 <svg xmlns="http://www.w3.org/2000/svg"
      xmlns:h="http://www.w3.org/1999/xhtml"
      width="800px" height="800px">
-  <title>SVG Geometry Properties: getComputedValue().stroke</title>
+  <title>SVG Painting: getComputedValue().stroke</title>
   <metadata>
     <h:link rel="help" href="https://svgwg.org/svg2-draft/painting.html#StrokeProperty"/>
     <h:meta name="assert" content="stroke computed value is as specified, with url values absolute."/>
@@ -26,7 +26,8 @@ test(() => {
   const target = document.getElementById('target');
   target.style['stroke'] = 'url("a.b#c")';
   const result = getComputedStyle(target)['stroke'];
-  assert_true(result.includes('/a.b#c'), result + ' is absolute');
+  const resolved = new URL("a.b#c", document.URL).href;
+  assert_equals(result, 'url("' + resolved + '")');
 }, 'url values are made absolute');
 
   ]]></script>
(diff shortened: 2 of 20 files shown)


svg/painting/parsing/fill-computed.svg
```diff
@@ -0,0 +1,33 @@
...
+test_computed_value("fill", 'url("https://example.com/") none');
+test_computed_value("fill", 'url("https://example.com/") rgb(12, 34, 56)');
+
+// context-fill and context-stroke are not yet supported by browsers.
+
+test(() => {
+  const target = document.getElementById('target');
+  target.style['fill'] = 'url("a.b#c")';
+  const result = getComputedStyle(target)['fill'];
+  assert_true(result.includes('/a.b#c'), result + ' is absolute');
```
AmeliaBR: Is there a way to test this more precisely?

svg/painting/parsing/stroke-computed.svg
```diff
@@ -0,0 +1,33 @@
...
+  <h:script src="/resources/testharnessreport.js"/>
+  <h:script src="/css/support/computed-testcommon.js"/>
+  <script><![CDATA[
+
+test_computed_value("stroke", "none");
+test_computed_value("stroke", "rgb(12, 34, 56)");
+
+test_computed_value("stroke", 'url("https://example.com/")');
+test_computed_value("stroke", 'url("https://example.com/") none');
+test_computed_value("stroke", 'url("https://example.com/") rgb(12, 34, 56)');
```
AmeliaBR: Might want to test (here and for `fill`) that the `color url()` values get swapped into a canonical order.

svg/painting/parsing/fill-computed.svg
AmeliaBR: Also: there _should_ be a test of bare hash urls (`fill="url(#x)"`). 

Advance warning: that's not currently consistently handled by browsers.

svg/painting/parsing/stroke-dasharray-computed.svg
```diff
@@ -0,0 +1,33 @@
...
+test_computed_value("stroke-dasharray", "none");
+
+test_computed_value("stroke-dasharray", "10", "10px");
+test_computed_value("stroke-dasharray", "calc(10px + 0.5em)", "30px");
+test_computed_value("stroke-dasharray", "calc(10px - 0.5em)", "0px");
+test_computed_value("stroke-dasharray", "40%");
+test_computed_value("stroke-dasharray", "calc(50% + 60px)");
+
+test_computed_value("stroke-dasharray", "10px, 20%, 30px");
+test_computed_value("stroke-dasharray", "0, 5", "0px, 5px");
```
AmeliaBR: Do we have a canonical rule about using comma separators vs space separators?

(I would have expected space to be the canonical version, myself.) Either way, we should test both.

svg/painting/parsing/stroke-miterlimit-computed.svg
```diff
@@ -0,0 +1,22 @@
...
+  <g id="target"></g>
+  <h:script src="/resources/testharness.js"/>
+  <h:script src="/resources/testharnessreport.js"/>
+  <h:script src="/css/support/computed-testcommon.js"/>
+  <script><![CDATA[
+
+test_computed_value("stroke-miterlimit", "0");
+test_computed_value("stroke-miterlimit", "0.5");
+test_computed_value("stroke-miterlimit", "1");
+test_computed_value("stroke-miterlimit", "7.5");
```
AmeliaBR: Can this particular testing framework test correct handling of invalid values? E.g., a negative value in this case should be rejected by the parser.

AmeliaBR:
Another general comment:

The testing of length values should include more units, including CSS3 units.

Of course, confirming that the CSS systems correctly handle those units doesn't mean that the rendering system will handle them correctly (test what some of the `calc()` values render as, to find out what I mean). But it's a start.

svg/painting/parsing/fill-computed.svg
fsoder: >let resolved = new URL("a.b#c", document.URL).href;
>assert_equals('url("' + resolved + '")', result);

or so ought to work. (Or use HTMLAnchorElement.href, but URL is probably supported widely enough by now.)

ewilligers:
> Do all browsers downcase keyword values

Firefox and Chrome Canary downcase, consistent with the spec [Serializing CSS Values](https://drafts.csswg.org/cssom/#serializing-css-values)
> To serialize a CSS component value depends on the component, as follows:
> keyword
> - The keyword converted to ASCII lowercase.

There are open bugs for [WebKit](https://bugs.webkit.org/show_bug.cgi?id=190685) and [Edge](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/19328562/).


> I'd expect to use the canonical spelling in the spec.

I'd also be happy for the spec to use lower case.

We have free choice: the Values and Units spec has [Hz, kHz](https://www.w3.org/TR/css-values-4/#frequency) and [Q](https://www.w3.org/TR/css-values-4/#absolute-lengths), and all of the following CSS spec files contain both `currentColor` and `currentcolor`:
* css-backgrounds-3/Overview.bs
* css-color-3/Overview.html
* css-color-3/Overview.src.html
* css-color-4/Overview.bs
* css-pseudo-4/Overview.bs
* css-text-decor-3/Overview.bs
* css-ui-4/Overview.bs
* css-values-4/Overview.bs

svg/painting/parsing/stroke-miterlimit-computed.svg
ewilligers: Invalid values are tested in `*/parsing/*-invalid.svg`; these found browser bugs relating to [image-rendering](https://bugs.chromium.org/p/chromium/issues/detail?id=901669),  [stroke-opacity](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/19656838/), [stroke-dasharray](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/19656979/), [stroke-width](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/19656907/) and more.

[stroke-miterlimit-invalid.svg](https://wpt.fyi/results/svg/painting/parsing/stroke-miterlimit-invalid.svg?label=stable&aligned) tests that negative values and misformed numbers are rejected.

svg/painting/parsing/stroke-dasharray-computed.svg
ewilligers: https://drafts.csswg.org/css-values-4/#component-multipliers
> A hash mark (#) indicates that the preceding type, word, or group occurs one or more times, separated by comma tokens 

https://drafts.csswg.org/cssom/#common-serializing-idioms
> To serialize a comma-separated list concatenate all items of the list in list order while separating them by ", ", i.e., COMMA (U+002C) followed by a single SPACE (U+0020).

I have added a test to stroke-dasharray-computed.svg (this PR) and stroke-dasharray-valid.svg (https://github.com/web-platform-tests/wpt/pull/14367)

svg/painting/parsing/stroke-computed.svg
ewilligers: I think values may only occur in one order - there is no `||` in the `<paint>` [grammar](https://svgwg.org/svg2-draft/painting.html#SpecifyingPaint).

The [stroke-invalid.svg](https://wpt.fyi/results/svg/painting/parsing/stroke-invalid.svg?label=stable&aligned) and [fill-invalid.svg](https://wpt.fyi/results/svg/painting/parsing/fill-invalid.svg?label=stable&aligned) tests reject invalid orders.

svg/painting/parsing/fill-computed.svg
ewilligers: Changed to use URL

ewilligers:
I think all comments have been addressed now, in either this PR or https://github.com/web-platform-tests/wpt/pull/14367.

svg/painting/parsing/stroke-computed.svg
AmeliaBR: Interesting. I'd never tested it. Another case where SVG rules aren't following modern CSS practices. But I'm not going to argue for changing it at this point.

svg/painting/parsing/stroke-dasharray-computed.svg
AmeliaBR: It's a bit more complicated than that with SVG, because most properties allow either comma or space separation, with no distinction in meaning.  SVG 2 uses the `<value>#*` combination list operator to describe it (a space-separated list of comma-separated lists), but that is slightly abusing the CSS syntax.

In this particular case, comma separation makes sense since it's an arbitrary-length list of similar values. And all browsers I tested did it that way, anyway. If we ever make `viewBox` or `points` or other attributes into CSS properties, then we'll have to deal with whether we want to enforce logic on the list separators!

Thanks for adding the extra test for space-separated values.

AmeliaBR (approved):

PR #14213, opened 2018-11-23
Rename "safari_webdriver" to "safari", get rid of "safari"

gsnedders:
@jugglinmike you don't use `safari_webdriver` do you? therefore this should be landable without breaking anything


jgraham (approved):

gsnedders:
To be clear: the "do not merge yet" label is pending @jugglinmike responding to the above; Mike: feel free to remove it when you do.

jugglinmike:
I wasn't aware that `safari_webdriver` existed. Thanks for the heads-up, Geoffrey!

jugglinmike (approved):

foolip:
Yay! Let's hope that the Safari runs from @jugglinmike's setup still finish, if they don't we should immediately revert this.

gsnedders:
Most of the apparent regressions are things going from ERROR -> OK at a harness level and getting new failures like that. Looking through it relatively quickly I can't see anything that looks actively bad?

PR #14217, opened 2018-11-23
Set default for manifest_update flag appropriately.

jdm:
This is required as part of https://github.com/servo/servo/pull/22258/ in order to unbreak Servo's nightly sync process.


lukebjerring (approved):
Rubber stamp ™

PR #14227, opened 2018-11-26
Remove the timeout in async_test for missing tests

qiuzhong:
Some async_test tests in webrtc and xhr with timeout are missed during
removing all the timeout in async_test. As part of goal #11120, this tests
should be updated too.

### DIFF
--- a/webrtc/simplecall.https.html
+++ b/webrtc/simplecall.https.html
@@ -17,7 +17,7 @@
   <script src="/resources/testharness.js"></script>
   <script src="/resources/testharnessreport.js"></script>
   <script type="text/javascript">
-  var test = async_test('Can set up a basic WebRTC call.', {timeout: 5000});
+  var test = async_test('Can set up a basic WebRTC call.');
 
   var gFirstConnection = null;
   var gSecondConnection = null;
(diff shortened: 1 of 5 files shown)


xhr/abort-during-upload.htm
```diff
@@ -11,7 +11,7 @@
   <body>
     <div id="log"></div>
     <script>
-      var test = async_test(document.title, {timeout:1100})
+      var test = async_test(document.title);
```
foolip:
```suggestion
      var test = async_test()
```

foolip: `document.title` is the default test name if none is given, so this can be removed.

webrtc/simplecall.https.html
```diff
@@ -17,7 +17,7 @@
   <script src="/resources/testharness.js"></script>
   <script src="/resources/testharnessreport.js"></script>
   <script type="text/javascript">
-  var test = async_test('Can set up a basic WebRTC call.', {timeout: 5000});
+  var test = async_test('Can set up a basic WebRTC call.');
```
foolip: Taskcluster is failing on Firefox because this test crashed 1/10 times. That's hardly because of these changes though.

foolip (approved):
LGTM % nit with suggested fix

Ms2ger (approved):

Issue #14003, opened 2018-11-09
css/selectors/attribute-selectors/attribute-case/syntax.html test has inconsistent names of subtests


foolip:
https://wpt.fyi/results/css/selectors/attribute-selectors/attribute-case/syntax.html?sha=71fe0a6c3d

In Edge, in addition to the failures, the subtests names are different, some are missing and there's an extra "[foo='bar' i" test at the end which I cannot find in the source.

This seems likely to be a test bug.

@jugglinmike is this a sort of anomaly that shows up in any report we have?

gsnedders:
I don't really understand how this could be a test bug? It's a pretty simple sequence of sync tests…

jugglinmike:
I'm on the case!

jugglinmike:
This is the result of a deficiency in Edge/Edgedriver: string literals transmitted via the "Execute Script" and "Execute Async Scripts" commands are truncated at the first null byte.

[I've submitted a WebDriver conformance test to alert the Edge maintainers](https://github.com/web-platform-tests/wpt/pull/13880#issuecomment-441341286), and I've suggested a more immediate solution for WPT via gh-14245.

jugglinmike:
@foolip We never did come to a conclusion for unprintable characters in test names, but now that Edge has a new WebDriver implementation which tolerates that case ([as evidenced by recent test
results](https://wpt.fyi/results/css/selectors/attribute-selectors/attribute-case/syntax.html?label=experimental&label=master&aligned)), the issue is less relevant. I'm going to un-assign myself and recommend my fix, unpopular though it may be--gh-14245.

foolip:
Alright, I'll just go ahead and close this because without something to surface the problem it won't ever be a priority for anyone.

PR #14245, opened 2018-11-27
[testharness.js] Disallow use of null byte in name

jugglinmike:
Because the null byte is a non-printable character, its use in test
names can be difficult for humans to interpret. Further, the Edge
browser is currently incapable of reporting data that includes the null
byte.

Rename all tests which previously included the null byte to instead
reference it via its JavaScript escape sequence. Update testharness.js
to disallow the character in test names.

---

Although this change is motivated by [a deficiency in Edge](https://github.com/web-platform-tests/wpt/issues/14003), I think there's good reason to enforce it as a constraint on test authors. Test names are intended for human use, so non-printable characters are generally inappropriate.

This could be addressed completely within testharness.js by silently sanitizing the test titles prior to reporting results. In some ways, that would be preferable because we could translate *all* non-printable characters (not just the problematic null byte) without modifying a single test. I'm suggesting that we implement this as a constraint instead because it seems important to avoid any changes between what test authors write and what they see reported.

To validate this patch, I [inserted a null byte in an existing test name and logged all infractions to disk (`suspicious.txt`)](https://github.com/bocoup/wpt/commit/35c7ac8f2719f298b513fce3ea6642a8de7af8bd). I added that commit to [a new branch on Bocoup's fork of WPT](https://github.com/bocoup/wpt/commits/testharness-disallow-null-byte-verify), triggering [a complete trial in Chrome and Firefox](https://tools.taskcluster.net/groups/P0PCeY7MRcywRe4dT-Dx6g). When it was through, I collected the contents of the log files and verified that only the "control" null byte was reported:

    $ ./wpt tc-download --ref testharness-disallow-null-byte-verify --artifact-name suspicious.txt --repo-name bocoup/wpt --out-dir edge-2        
    INFO:tc-download:edge-2/wpt-chrome-dev-testharness-4-ANkbKtjhRXCDecRxGAUIFQ-suspicious.txt
    INFO:tc-download:edge-2/wpt-firefox-nightly-testharness-4-GAj8QqurSLeHO_bBa8Y9JQ-suspicious.txt
    $ cat edge-2/*
    http://web-platform.test:8000/dom/events/CustomEvent.html - "CustomEvent [\u0000] dispatching."
    http://web-platform.test:8000/dom/events/CustomEvent.html - "CustomEvent [\u0000] dispatching."

### DIFF
--- a/css/selectors/attribute-selectors/attribute-case/semantics.html
+++ b/css/selectors/attribute-selectors/attribute-case/semantics.html
@@ -75,6 +75,9 @@ var nomatch = [
   ["[foo='bar' i]", ["", "baz", "BAR"]],
 ];
 var mode = "standards mode";
+function replace_null(str) {
+  return str.replace(/\u0000/g, "\\u0000");
+}
 function format_attrs(attrs) {
   var rv = [];
   attrs.forEach(function(attr) {
@@ -87,7 +90,7 @@ function format_attrs(attrs) {
     str += name + "=\"" + value + "\"";
     rv.push(str);
   });
-  return rv.join(" ");
+  return replace_null(rv.join(" "));
 }
 onload = function() {
   var quirks = document.getElementById('quirks').contentWindow;
@@ -117,11 +120,11 @@ onload = function() {
         style.textContent = s + ' { visibility:hidden }';
         assert_equals(style.sheet.cssRules.length, (ns_decl ? 2 : 1), 'rule didn\'t parse into CSSOM');
         assert_equals(global.getComputedStyle(elm).visibility, 'hidden', 'selector didn\'t match');
-      }, s + ' <div ' + format_attrs(attrs) + '> in ' + global.mode);
+      }, replace_null(s) + ' <div ' + format_attrs(attrs) + '> in ' + global.mode);
       if (!ns_decl) {
         test(function() {
           assert_equals(global.document.querySelector(s), elm, 'selector didn\'t match');
-        }, s + ' <div ' + format_attrs(attrs) + '> with querySelector in ' + global.mode);
+        }, replace_null(s) + ' <div ' + format_attrs(attrs) + '> with querySelector in ' + global.mode);
       }
     });
     nomatch.forEach(function(arr) {
@@ -134,11 +137,11 @@ onload = function() {
         style.textContent = s + ' { visibility:hidden }';
         assert_equals(style.sheet.cssRules.length, (ns_decl ? 2 : 1), 'rule didn\'t parse into CSSOM');
         assert_equals(global.getComputedStyle(elm).visibility, 'visible', 'selector matched');
-      }, s + ' <div ' + format_attrs(attrs) + '> in ' + global.mode);
+      }, replace_null(s) + ' <div ' + format_attrs(attrs) + '> in ' + global.mode);
       if (!ns_decl) {
         test(function() {
           assert_equals(global.document.querySelector(s), null, 'selector matched');
-        }, s + ' <div ' + format_attrs(attrs) + '> with querySelector in ' + global.mode);
+        }, replace_null(s) + ' <div ' + format_attrs(attrs) + '> with querySelector in ' + global.mode);
       }
     });
   });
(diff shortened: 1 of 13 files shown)


annevk (commented):
I don't like this approach. If we really need to do this, why not adjust the title in the harness? (I'm also not a fan of not addressing the root issue, Edge, and instead letting everyone bear the cost.)

annevk (changes requested):
Explicitly marking as requesting changes as I'm not comfortable with the changes to tests I help maintain.

domenic:
I agree that having testharness make these changes would be ideal. Perhaps by replacing the bytes with their corresponding [control character pictures](https://en.wikipedia.org/wiki/Unicode_control_characters#Control_pictures)

jgraham:
Yeah, I think having testharness deal with null and lone surrogates would be good (I already had to change wptreport to replace lone surrogates with U+XXXX escapes since they are not trivially supported in languages with UTF-enforcing String types, and treating all the names as bytes is surprisingly difficult).

foolip:
Where does that replacing happen? I guess strings get converted to UTF-8 first when they go in WebDriver response, so in testharness.js itself? More escaping there seems reasonable.

jgraham:
Heh, such optimisim.

The WebDriver response of course has to be UTF-8. But JSON escapes are based on UCS2 for (reasons not known to me) and so you can smuggle surrogate pairs pairs through as `\ud801` or whatever. The Python JSON library is happy with this because Python 2 strings are UCS2/4 internally. On the other hand in e.g. Rust where the strings are UTF-8 you can't get away with that (even if you choose some bytes type for the name field; the library point blank refuses to deserialise escaped unpaird surrogates). So since wptreport.json is the main interchange format we started stripping out the escapes there.

Doing it all in testharness seems sensible, but you could still have a case where someone put something bad into the title or message of another test type e.g. wdspec tests (reftests are a little harder because filesystems, but certainly not impossible).

lukebjerring:
@foolip re: wpt.fyi: having the UI reflect non-printed chars seems reasonable. File and issue, with a specific example attached, and I'll have someone (probably me) take a look.

jugglinmike:
> If we really need to do this, why not adjust the title in the harness?

It seems important to avoid any changes between what test authors write and what they see reported. Do you think there's value in that?

> I'm also not a fan of not addressing the root issue, Edge, and instead > letting everyone bear the cost.

I agree that the bug should be fixed in Edge. That's why [I also submitted WebDriver tests for the expected behavior](https://github.com/web-platform-tests/wpt/pull/13880#issuecomment-441341286).  We can't know when/if the bug will be fixed, but in the mean time, many other people are already bearing the cost. The visitors to wpt.fyi see confusing results for tests which have nothing to do with WebDriver (see, e.g. gh-14003).  Even if it was reasonable to expect them to understand what was going wrong, they have no agency over the issue. This patch alerts folks who *can* understand and avoid the problem.

> What is the point where Edge truncates strings at U+0000? I tried > `JSON.stringify"A\u0000B")` and that was fine. Is it with `postMessage`?

It's in WebDriver--see gh-13880 for a description of the issue.

jgraham:
> > If we really need to do this, why not adjust the title in the harness?
> 
> It seems important to avoid any changes between what test authors write and what they see reported. Do you think there's value in that?

I think "is there any value in that" is the wrong framing. Clearly it's possible to imagine cases where there is value in the title being byte-for-byte identical to what the author put in the js file. But on the other hand
* I suspect a lot of cases where there are unusual codepoints, the origin is a list rather than something else
* The title shouldn't be affecting the test result, so changing test titles theoretically doesn't matter much, as long as you don't end up causing clashes between tests.

jugglinmike:
> I think "is there any value in that" is the wrong framing.

The addition of the word "any" makes my comment seem more like a straw man, but it was an honest question. There's a subjective component to the decision, and I'm open to alternatives. I'd like to understand the trade offs and have them documented here.

> The title shouldn't be affecting the test result

Could you expand on why that's important?

annevk:
The title doesn't really matter, so putting constraints on test authors (who often generate titles and would have to write more complicated code to do so) is worse than doing some general escaping in the library.

foolip:
I've closed https://github.com/web-platform-tests/wpt/issues/14003 which this is a fix for, and would recommend closing this as well, since the issue no longer affects any configuration we're still running regularly.

jugglinmike:
Yeah, I think there was a more fundamental disagreement about the purpose of test titles.

PR #14260, opened 2018-11-27
Origin outside of CORS and interaction with the Referrer Policy

annevk:


annevk:
Please ignore the lack of DRY for now, that will be cleaned up.

I'm curious about three things myself:

* Why do browsers have the results they do? (Note that the expectations are written to follow what Chrome does.)
* What scenarios do we need to cover? (HTTPS to HTTP isn't covered, affect of referrer policy, if any, isn't covered. Same-origin isn't covered.)
* What changes are needed to step 5.10 of https://fetch.spec.whatwg.org/#concept-http-network-or-cache-fetch to account for the way browsers generate `Origin`.

@mikewest @youennf @yutakahirano I hope you'll be able to help here.

fetch/origin/no-cors.any.js
```diff
@@ -18,3 +18,59 @@ promise_test(async function() {
...
+    frame.onload = () => {
+      resolve();
+      frame.remove();
+    }
+    document.body.appendChild(frame);
+  });
+
+  const json = await (await fetch(redirectPath + "?dump&stash=" + stash)).json();
+
+  assert_equals(json[0], "no Origin header", "first origin should equal this origin");
```
JuniorHsu: the error description doesn't not match the assertion

fetch/origin/no-cors.any.js
```diff
@@ -18,3 +18,59 @@ promise_test(async function() {
...
+          submit = form.appendChild(doc.createElement("input"));
+    form.action = url;
+    form.method = "POST";
+    submit.type = "submit";
+    submit.click();
+  });
+
+  const json = await (await fetch(redirectPath + "?dump&stash=" + stash)).json();
+
+  assert_equals(json[0], "null", "first origin should equal this origin");
```
JuniorHsu: same here

fetch/origin/no-cors.any.js
```diff
@@ -18,3 +18,59 @@ promise_test(async function() {
...
+      }
+    }, { once: true });
+    document.body.appendChild(frame);
+    const doc = frame.contentDocument,
+          form = doc.body.appendChild(doc.createElement("form")),
+          submit = form.appendChild(doc.createElement("input"));
+    form.action = url;
+    form.method = "POST";
+    submit.type = "submit";
+    submit.click();
```
JuniorHsu: timeout in firefox is because the form is not in the iframe dom tree after appendChild. Therefore, the |click()| doesn't work, neither POST navigation. Not sure if it's a DOM bug.
I can work around with form.submit(), but that's not a really POST navigation.

annevk:
Hey @JuniorHsu, thanks for taking a look. I removed the detailed descriptions for now as they're mostly wrong and I guess we haven't quite decided what behavior we want anyway. I also found a way to make final test consistently run across browsers (there's a difference between how browsers deal with newly appended frames unfortunately).

Are you still planning on writing more tests around this to figure out what kind of `Origin` header we want for the non-CORS scenarios?

annevk:
This PR now includes all @JuniorHsu's great work. Next step is figuring out the spec and then making sure this PR is aligned with what we agree on there and then filing the various implementation bugs required.

If folks want to review the additions meanwhile that'd be great!

annevk:
@JuniorHsu this now includes your `fetch()` test and I rebased on master so that the latest tooling is used when building.

Did you get closer with the intermittent issue in Firefox? Will you work on splitting this test?

JuniorHsu:
> @JuniorHsu this now includes your `fetch()` test and I rebased on master so that the latest tooling is used when building.
> 
> Did you get closer with the intermittent issue in Firefox? Will you work on splitting this test?

I send another PR
https://github.com/web-platform-tests/wpt/pull/17407

sideshowbarker (approved):

PR #14266, opened 2018-11-27
WebKit export of https://bugs.webkit.org/show_bug.cgi?id=191975

alijuma:
* WebKit export of https://bugs.webkit.org/show_bug.cgi?id=191975
* Intersection Observer: rootMargin: '' gives weird results


alijuma:
Patch has been reviewed downstream.

fred-wang (approved):
LGTM, maybe we would need a reference to the spec to make it clear.

From https://w3c.github.io/IntersectionObserver/#dom-intersectionobserverinit-rootmargin an empty string is invalid so I guess the default "0px" should be used but maybe the spec should be more explicit about how to handle invalid values...

PR #14270, opened 2018-11-28
[resource-timing] Test behavior with cached assets

jugglinmike:
Add a note to the referenced test file to avoid accidental errors in
future renaming/deletion operations.

---

This is on the edge of "testable from WPT." It fails consistently on my machine, but because we don't have explicit control over the browser cache, it might actually be flaky in some configurations. Though Chromium is currently flaky when running the referenced test in a batch (since it may cache the resources while executing a prior test).

[I've filed a bug against the Chromium project to report the behavior](https://bugs.chromium.org/p/chromium/issues/detail?id=909094) in case we decide WPT isn't the right place to be testing this particular behavior.

### DIFF
--- a/resource-timing/single-entry-per-resource-cached.html
+++ b/resource-timing/single-entry-per-resource-cached.html
@@ -1,0 +1,20 @@
+<!DOCTYPE HTML>
+<meta charset=utf-8>
+<meta name="timeout" content="long">
+<title>One resource when reusing data (cached resources)</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<body>
+<p>
+Chrome 70 emits delayed performance events for &lt;script&gt; elements in the
+document head which reference chached resources. This test attempts to prompt
+that condition by first loading <code>testharness.js</code> and then loading
+the relevant web-platform-test in an iframe. The nested test's reference to
+<code>testharness.js</code> is expected to cause a cache hit, prompting Chrome
+to emit the unexpected performance event.
+</p>
+<iframe id="subject" src="single-entry-per-resource.html"></iframe>
+<script>
+fetch_tests_from_window(document.getElementById('subject').contentWindow);
+</script>
+</body>
--- a/resource-timing/single-entry-per-resource.html
+++ b/resource-timing/single-entry-per-resource.html
@@ -6,6 +6,7 @@
 <script src="/resources/testharnessreport.js"></script>
 <body>
 <script>
+  // This test is referenced by `single-entry-per-resource-cached.html`.
   var img_entries = 0;
   var observed = 0;
   var img_url = "resources/blue.png";


yoavweiss:
Thanks for adding RT WPTs! :)

At the same time, at least from the description, what you're describing is WAI - cached resources should indeed include resources that are fetched from the cache (although with a `transferSize` of 0).

jugglinmike:
Aha, [I see that now](https://w3c.github.io/resource-timing/#processing-model). That makes me think that the test (and not Chrome) may be flaky. The harness files will create PerformanceResourceTiming objects, but it seems as though we can't guarantee that those will been queued and dispatched prior to the registration of this test's observer.

If we registered the observer in the `<head>`, before the test harness `<script>` tags, we could assert the presence of timing objects for those and be confident in their presence regardless of implementation-defined behavior:

```diff
--- a/resource-timing/single-entry-per-resource.html
+++ b/resource-timing/single-entry-per-resource.html
@@ -2,9 +2,6 @@
 <meta charset=utf-8>
 <meta name="timeout" content="long">
 <title>One resource when reusing data</title>
-<script src="/resources/testharness.js"></script>
-<script src="/resources/testharnessreport.js"></script>
-<body>
 <script>
   var img_entries = 0;
   var observed = 0;
@@ -19,11 +16,16 @@
       }
     });
   observer.observe({entryTypes: ["resource"]});
+</script>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<body>
+<script>
   window.onload = function() {
     // A timeout is needed as PerformanceObserver is not guaranteed to run before onload triggered.
     setTimeout(function() {
       test(function(){
-        assert_equals(observed, 1);
+        assert_equals(observed, 3);
         assert_equals(img_entries, 1);
       }, "Only one resource entry per resource");
     },0)
```

Then again, that increases the amount of behavior under test. Maybe instead, we should just ignore  entries for the harness files if we happen to catch them and leave asserting that behavior to a dedicated test:

```diff
--- a/resource-timing/single-entry-per-resource.html
+++ b/resource-timing/single-entry-per-resource.html
@@ -13,6 +13,9 @@
     function (entryList, obs) {
       var entries = entryList.getEntriesByType("resource");
       for (var i = 0; i < entries.length; ++i) {
+        if (/testharness(report)?\.js/.test(entries[i].name)) {
+          continue;
+        }
         ++observed;
         if (entries[i].name.indexOf(img_url) != -1)
             ++img_entries;
```

What do you think?

yoavweiss:
@jugglinmike - apologies for lack of responsiveness. Hectic few weeks...
Ignoring the harness files SGTM. It's true that we cannot guarantee that they would be dispatched before, although we could schedule the PO registration to happen in a next task (using a `setTimeout`).
At the same time, simply ignoring their entries would work as well.
Care to send a PR to that effect?

jugglinmike:
No worries about the delay; I appreciate the feedback. Since the correct solution is so dissimilar from the change I've proposed here (and requires a new commit message), [I've opened a new pull request to supersede this one](https://github.com/web-platform-tests/wpt/pull/14487).

PR #14273, opened 2018-11-28
Remove timeout from async_test for argument form.

qiuzhong:
There're cases that the timeout is passing to async_test in the tests.
They should be removed as well.
Related: #11120


qiuzhong:
@foolip , the Taskcluster CI fails sometimes and maybe get through with a commit even there's no reasonable fix. Do you know if there is a way to trigger it to rebuild again if I have no further commit for a PR?

lukebjerring:
Cloned here; https://github.com/lukebjerring/wpt/pull/12/checks

foolip:
Thanks! It looks like the results in css/css-transitions are unchanged, as expected. And the other changes appear to be for manual tests and stuff that for one reason or another isn't run.

foolip (approved):

PR #14288, opened 2018-11-28
Test when a SW replies with a fetch from a different URL

jyasskin:
The resulting document's location should be the originally requested
URL.

This goes with https://github.com/whatwg/html/pull/4205. Please let me know if there's a better way to approach this.

### DIFF
--- a/service-workers/service-worker/navigation-redirect.https.html
+++ b/service-workers/service-worker/navigation-redirect.https.html
@@ -668,6 +668,24 @@ redirect_test(
     'SW-fetched redirect to other-origin in-scope.');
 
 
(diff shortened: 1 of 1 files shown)


mfalken (commented):
Thanks! I'm somewhat surprised (but not very surprised) that we didn't' have test coverage for this, but I indeed can't find an existing test.

Maybe update the commit description first sentence to mention this is a test for navigations.

It'd also be worth having tests that combine redirects + fetch(other-url). E.g., url1 redirects to url2 and SW responds with fetch to url3... the location should be url2. But it doesn't have to be in this CL.

mfalken (approved):

PR #14290, opened 2018-11-28
Attempt to monitor memory from userspace

Hexcles:
It is suspected that Taskcluster jobs sometimes fail due to OOM. It is
hard to know what exactly happens if the kernel OOM killer kicks in.
This change attempts to use a user-space solution: earlyoom.

There are two benefits of earlyoom:
1. Logging: we can log the memory usage periodically and the actions
   taken by earlyoom when the system is low on memory.
2. Recovery: we can teach earlyoom to prefer to kill browser processes
   and avoid python processes (wpt) so that the runner has a chance to
   recover (marking the test as CRASH and carrying on).

This is an effort to investigate (and hopefully alleviate) #13989.

### DIFF
--- a/tools/docker/start.sh
+++ b/tools/docker/start.sh
@@ -17,6 +17,10 @@ REVISION=${3:-FETCH_HEAD}
 BROWSER=${4:-all}
 CHANNEL=${5:-nightly}
 
+# Start userspace OOM killer: https://github.com/rfjakob/earlyoom
+# It will report memory usage every minute and prefer to kill browsers.
+sudo earlyoom -p -r 60 --prefer '(chrome|firefox)' --avoid 'python' &
+
 cd ~
 
 mkdir web-platform-tests
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -61,7 +61,7 @@ tasks:
               owner: ${event.pusher.email}
               source: ${event.repository.url}
             payload:
-              image: harjgam/web-platform-tests:0.22
+              image: hexcles/web-platform-tests:0.23
               maxRunTime: 7200
               artifacts:
                 public/results:
@@ -125,7 +125,7 @@ tasks:
                 owner: ${event.pull_request.user.login}@users.noreply.github.com
                 source: ${event.repository.url}
               payload:
-                image: harjgam/web-platform-tests:0.22
+                image: hexcles/web-platform-tests:0.23
                 maxRunTime: 7200
                 artifacts:
                   public/results:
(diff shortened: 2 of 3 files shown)


Hexcles:
cc @foolip

jugglinmike:
This is an interesting idea!

The documentation suggests that information is expected to be output on standard error, but it doesn't define the frequency. The Taskcluster logs include some initial data:

    ++ sudo earlyoom -p -r 60 --prefer '(chrome|firefox)' --avoid python
    Initialized empty Git repository in /home/test/web-platform-tests/.git/
    ++ git remote add origin https://github.com/web-platform-tests/wpt.git
    ++ git fetch --quiet --depth=50 origin refs/pull/14290/merge
    earlyoom v1.0
    mem total: 15038 MiB, min: 1503 MiB (10 %)
    swap total: 0 MiB, min: 0 MiB (10 %)
    Could not lock memory - continuing anyway: Cannot allocate memory
    Could not set priority - continuing anyway: Permission denied

...but nothing following that. Those errors are being ignored; do you know if the process is still running?

Hexcles:
@jugglinmike good point. I used `-r 60` to ask it to report every 60 seconds. I've pushed another test commit which should trigger some much longer tasks on Taskcluster and we'll see.

Hexcles:
(Apparently modifying `testharness.js` would not trigger "affected tests", of course... Let me change something else.)

Hexcles:
Just confirmed that memory reporting is working correctly: https://tools.taskcluster.net/groups/e4VuezM_RxKW7J82c569wA/tasks/Fif-hWo-SH6FK3tR07dIcA/runs/0/logs/public%2Flogs%2Flive.log#L550

The second error seems rather harmless (we probably can't set niceness to -20 inside a non-privileged Docker container), but I'm less certain about the first error: the killer part of `earlyoom` might not function properly (though looking at the source code, I think it should as it basically uses `kill -9`). Regardless, I think the logging would be helpful. One question is: is 1 stat per minute the right frequency?

tools/docker/start.sh
```diff
@@ -17,6 +17,10 @@ REVISION=${3:-FETCH_HEAD}
 BROWSER=${4:-all}
 CHANNEL=${5:-nightly}
 
+# Start userspace OOM killer: https://github.com/rfjakob/earlyoom
+# It will report memory usage every minute and prefer to kill browsers.
+sudo earlyoom -p -r 60 --prefer '(chrome|firefox)' --avoid 'python' &
```
foolip: Very nice!

foolip (approved):
LGTM % DO NOT SUBMIT.

Please land this at a time when you can monitor the results for several hours afterwards, to revert if we get results regressions or perhaps totally broken runs because of killed browsers.

PR #14291, opened 2018-11-28
[WebCryptoAPI] Remove race condition

jugglinmike:
Configure testharness.js to wait until the global `done` function is
invoked before reporting results. This ensures that all
asynchronously-declared tests are defined and subsequently executed
regardless of variations in timing.

---

The instability is apparent when comparing the test results between Chrome, Firefox, and Safari, e.g. https://wpt.fyi/results/WebCryptoAPI/sign_verify?sha=58523789a0

2018-11-28-crypto-race

Firefox and Safari are missing results from tests that depend on `WebCryptoAPI/sign_verify/ecdsa.js` due to the race condition which this patch corrects.

### DIFF
--- a/WebCryptoAPI/sign_verify/ecdsa.js
+++ b/WebCryptoAPI/sign_verify/ecdsa.js
@@ -1,5 +1,7 @@
 
 function run_test() {
+    setup({explicit_done: true});
+
     var subtle = self.crypto.subtle; // Change to test prefixed implementations
 
     // When are all these tests really done? When all the promises they use have resolved.
--- a/WebCryptoAPI/sign_verify/hmac.js
+++ b/WebCryptoAPI/sign_verify/hmac.js
@@ -1,5 +1,7 @@
 
 function run_test() {
+    setup({explicit_done: true});
+
     var subtle = self.crypto.subtle; // Change to test prefixed implementations
 
     // When are all these tests really done? When all the promises they use have resolved.
--- a/WebCryptoAPI/sign_verify/rsa.js
+++ b/WebCryptoAPI/sign_verify/rsa.js
@@ -1,5 +1,7 @@
 
 function run_test() {
+    setup({explicit_done: true});
+
     var subtle = self.crypto.subtle; // Change to test prefixed implementations
 
     // When are all these tests really done? When all the promises they use have resolved.


jimsch:
I would just like to verify that it does not matter when the call to setup is made.  Specifically that it does not matter if it is not called from all of the run_test functions.

jugglinmike:
`setup` should be called before the `load` event on the document has fired.  In worker contexts (where there is no document), `explicit_done` is automatically set to `true`. We could move the setting to the HTML file, but I feel that it's better to have it in-line with the code that requires it. There's more detail in the framework's documentation: https://web-platform-tests.org/writing-tests/testharness-api.html

I'm glad you mentioned the other `run_test` functions. I was focused on the instability we're experiencing today, which happens to be limited to the tests for ECDSA. However, `rsa.js` and `hmac.js` use the same pattern and could exhibit the same problem, so I've updated those as well.

Ms2ger (approved):
This is good. However, the tests are written as

```js
var promise = importVectorKeys(vector, ["verify"], ["sign"]).then(function(vectors) {
  promise_test(function(test) { /* test stuff */ });
}, function(err) {
  promise_test(function(test) { assert_unreached(/* ... */) });
});
```

… while I would expect

```js
promise_test(function(test) {
  return importVectorKeys(vector, ["verify"], ["sign"]).then(function(vectors) {
    /* test stuff */
  });
});
```

Why is that?

jimsch:
The reason is that not all algorithms and not all sizes may be supported.  This provides a uniform error in those cases.

jugglinmike:
@jimsch Thanks for the review. Would you mind merging on my behalf?

jimsch:
I am used to this happening automatically, but two of the tests never finished

jugglinmike:
Those "checks" are intended to produce test results to assist in review. They're still under development, though, which is why they are not required for merging.

@lukebjerring Would you like me to let you know when/if I see issues with those checks?

lukebjerring:
Yep; this bad state shouldn't occur anymore, and I've also updated the body of the check summary to include a link for reporting issues, e.g. https://github.com/web-platform-tests/wpt/pull/14291/checks?check_run_id=39572969

> Note: wpt.fyi checks are still in beta!
> See something wrong? Please [file an issue](https://github.com/web-platform-tests/wpt.fyi/issues/new?labels=bug&projects=web-platform-tests%2Fwpt.fyi%2F6&template=checks.md&title=Regression+checks+issue)!

jugglinmike:
Very nice! Thanks, Luke

PR #14294, opened 2018-11-29
[html] Add more protocol handlers tests

Honry:
- Imported and merged protocol handler tests from chromium(https://chromium.googlesource.com/chromium/src/+/HEAD/third_party/blink/web_tests/http/tests/navigatorcontentutils)
  - Fixed bug that it should throw SecurityError with invaild schemes
  - Did some code optimization

### DIFF
--- a/html/webappapis/system-state-and-capabilities/the-navigator-object/protocol.https.html
+++ b/html/webappapis/system-state-and-capabilities/the-navigator-object/protocol.https.html
@@ -138,8 +138,9 @@ for (const scheme of blacklist) {
  * We're just testing that the call goes through here. Whether or not they
  * actually work as handlers is covered by the interactive tests. */
 
-const whitelist = [
-  /*safelisted schemes*/
+const safelist = [
+  /* safelisted schemes listed in
+   * https://html.spec.whatwg.org/multipage/system-state.html#safelisted-scheme */
   'bitcoin',
   'geo',
   'im',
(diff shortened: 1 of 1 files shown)


foolip:
@Honry, are you able to send CLs to Chromium directly? If so, could you make this change in Chromium instead so that the existing tests are also deleted at the same time? Otherwise, we'll end up with both sets of tests and probably forget to ever deduplicate.

Honry:
> @Honry, are you able to send CLs to Chromium directly?

Not at present. But I'm processing the legal stuff. :)

Honry:
@foolip, a CL at https://chromium-review.googlesource.com/c/chromium/src/+/1365250 is ready, would you like to help review it?

foolip (approved):
Sorry for the extra roundtrips, in the end it's just as well we land this PR and let the changes be imported into Chromium.

html/webappapis/system-state-and-capabilities/the-navigator-object/protocol.https.html
```diff
@@ -170,45 +109,75 @@
...
+    assert_throws('SECURITY_ERR', () => { navigator.unregisterProtocolHandler(scheme, location.href + '/%s', 'foo'); });
+  }, 'unregisterProtocolHandler: Attempting to override the "' + scheme + '" protocol should throw SECURITY_ERR.');
 }
 
 /* The following protocols must be possible to override.
  * We're just testing that the call goes through here. Whether or not they
  * actually work as handlers is covered by the interactive tests. */
-var whitelist = new Array(
+
+const whitelist = [
```
annevk: Can we rename the variable to safelist while doing this refactoring as well?

(Looks great overall btw, thanks for simplifying this!)

Honry: Sure, done.

PR #14295, opened 2018-11-29
Move back resources/custom-elements-helpers.js into custom-elements/resources/custom-elements-helpers.js after darsh7807

rniwa:
This file has no business being in the top-level resources directory.


rniwa:
Note that the move also broke all other exiting tests.

fred-wang (approved):

sideshowbarker (approved):

PR #14297, opened 2018-11-29
Test payment request is showing boolean

marcoscaceres:
Tests for https://github.com/w3c/payment-request/pull/811 and https://github.com/w3c/payment-request/pull/821

### DIFF
--- a/payment-request/payment-is-showing.https.html
+++ b/payment-request/payment-is-showing.https.html
@@ -1,0 +1,135 @@
+<!DOCTYPE html> <meta charset="utf-8" />
+<title>Test for PaymentRequest.show(optional promise) method</title>
+<link
+  rel="help"
+  href="https://w3c.github.io/browser-payment-api/#dfn-payment-request-is-showing"
+/>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/resources/testdriver-vendor.js"></script>
+<script src="/resources/testdriver.js"></script>
+<script>
+  "use strict";
+  setup({ explicit_done: true });
+  const basicCard = { supportedMethods: "basic-card" };
+  const methods = [basicCard];
+  const details = {
+    total: {
+      label: "Total",
+      amount: {
+        currency: "USD",
+        value: "1.00",
+      },
+    },
+  };
+
+  promise_test(async t => {
+    const showPromise = test_driver.bless("show payment request", () => {
+      // Set payment request is showing boolean to true
+      return new PaymentRequest(methods, details).show();
+    });
+    await promise_rejects(
+      new PaymentRequest(methods, details).show(),
+      "AbortError",
+      "Expected to reject with an AbortError DOMException"
+    );
+    showPromise.abort();
+  }, "Simple payment-relevant browsing context's payment request is showing boolean is true");
+
+  promise_test(async t => {
+    // Load up an iframe, so it can set the payment is showing boolean
+    const iframe = document.createElement("iframe");
+    iframe.src = "blank.html";
+    document.body.appendChild(iframe);
+    await new Promise(resolve => (iframe.onload = resolve));
+
+    const iframeRequest = new iframe.contentWindow.PaymentRequest(
+      methods,
+      details
+    );
+
+    // Let's get some blessed showPromises
+    const [iframeShowPromise, showPromise] = await test_driver.bless(
+      "show payment request",
+      () => {
+        // Set top-level is showing boolean to true
+        const iframeShowPromise = iframeRequest.show();
+        // The window's showPromise should be rejected, because iframe is showing
+        const showPromise = window.PaymentRequest(methods, details).show();
+        return [iframeShowPromise, showPromise];
+      }
+    );
+
+    await promise_rejects(
+      showPromise,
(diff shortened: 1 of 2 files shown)


rsolomakhin (approved):

payment-request/payment-is-showing.https.html
```diff
@@ -0,0 +1,135 @@
...
+    );
+
+    // Let's get some blessed showPromises
+    const [iframeShowPromise, showPromise] = await test_driver.bless(
+      "show payment request",
+      () => {
+        // Set top-level is showing boolean to true
+        const iframeShowPromise = iframeRequest.show();
+        // The window's showPromise should be rejected, because iframe is showing
+        const showPromise = window.PaymentRequest(methods, details).show();
```
domenic: Missing `new`

payment-request/payment-is-showing.https.html
```diff
@@ -0,0 +1,135 @@
...
+    await promise_rejects(
+      showPromise,
+      "AbortError",
+      "Expected to reject with an AbortError DOMException, iframe is showing a payment request"
+    );
+
+    // Bring down the iframe's payment sheet
+    await iframeRequest.abort();
+
+    // We should now be able to show again with top level browsing context,
```
domenic: This may be clearer as two separate tests

domenic (commented):
Looks good overall (once you fix things like missing `new`, by running the tests to confirm that they work) but a popup/cross-global test would exercise the truly tricky things.

marcoscaceres:
@domenic, I refactored these quite a bit and added popup tests based on your feedback. Also added tests that mix frames, windows, and popups  - sorry about the length of the file. I'm trying to be as thorough as possible, while also keeping each test as legible as possible as these tests could be tricky/annoying to debug. 

Firefox passes all but the last two: we don't reject the `showPromise` if the browsing context that created the request is navigated - oops!

@rsolomakhin, I tried running these in Chrome using:

```
./wpt run chrome payment-request/payment-is-showing.https.html
```

But it keeps (again) complaining that "basic-card" is not supported by Chrome?

> "NotSupportedError: The payment method "basic-card" is not supported".

I'm not sure what I'm doing wrong 😢 Would really appreciate your help. If you can run the locally, that would be greatly appreciated.

rsolomakhin:
> `"NotSupportedError: The payment method "basic-card" is not supported".`

That happens if your HTTPS icon is not green. Are you using a self-signed HTTPS cert? That would do it. Use localhost, local file, or a valid HTTPS cert instead. For example, press Ctrl-O, navigate to `payment-is-showing.https.html` and open it. Is it doing anything special that requires the `./wpt` tool to make it work?

marcoscaceres:
@rsolomakhin, thanks for the clarification! The problem is clearly around the certs - because you are perfectly describing my issue :) However, the [Trusting Root CA](https://github.com/web-platform-tests/wpt#trusting-root-ca) section of the docs state: 

> To avoid this problem when running tests in Chrome or Firefox use `wpt run`, which disables certificate checks and therefore doesn't require the root CA to be trusted.

I guess the above doesn't hold for Payment Request API in Chrome (the API is there, but "basic-card" is not available)? @foolip just wanted to bring this problem to your attention too (probably not something you can help me with... I guess just FYI).

foolip:
@marcoscaceres thanks for pointing this out. This is the first case I've been made aware of where there's a behavioral difference depending on the "level" of the secure connection. I think we want the HTTPS setup used to be trusted enough to get access to all APIs that a real site could. I'm not sure I understand the criteria used exactly, could you please file an issue? Perhaps it's unfixable without doing things that security folks would object to, but who knows.

rsolomakhin:
@foolip : Oh my, I may have violated the age-old maxim, "Don't reinvent your own security." The behavior you're seeing with `basic-card` being unsupported with certificate errors are due to my design decision early on in Payment Request development process. Although the spec says the API should be available in secure context, I nerfed the capabilities if the certificate has any kind of errors. This prevents the user from bypassing a security interstitial and then sending their payment info to a sketchy website. The result is the behavior that @marcoscaceres has observed. Although we typically don't do this, would the security team allow the special case tightening in the context of payments?

foolip:
@rsolomakhin, I suspect that this isn't the only behavior that depends on something about the cert, but I'm no expert and don't *know* of any other cases. I'd ask @mikewest, and now I did :)

If you do think the wpt setup should change to make this testable, please file an issue. I know we do things slightly different for content_shell, passing the digest of a cert to trust, so it's possible that the same could be done here in upstream wpt. @Hexcles please correct me if I'm wrong.

rsolomakhin:
@foolip : No need to change `wpt`. I have a bug filed against me to fix PaymentRequest ignoring the command-line flag to relax certificate handling.

foolip:
Excellent, thanks @rsolomakhin!

marcoscaceres:
@rsolomakhin, thanks for looking deeper into the security issue. In the interest of moving this forward, would you mind giving this PR one more look (and re-approval🤞)? It's changed quite a bit since you last approved it, so would feel more comfortable merging only once it's had another set of eyes on it.

rsolomakhin (approved):
You're a machine! So many test cases....

PR #14300, opened 2018-11-29
[Azure Pipelines] Run infrastructure/ tests on Windows 10

foolip:


foolip:
Getting Python installed is the first challenge. I've tried `msiexec /a python-2.7.15.amd64.msi` but it fails with "The Windows Installer Service could not be accessed. This can occur if the Windows Installer is not correctly installed. Contact your support personnel for assistance." @thejohnjansen do you know the best way to install stuff? Does the context where the Azure Pipelines steps are run have administrator rights, or some equivalent of sudo rights?

foolip:
@thejohnjansen this is looking quite promising, I got Python installed and got the tools/ unittests running, although as expected with failures.

Trying to run the infrastructure/ tests with Edge I got stuck and not being able to update System32\drivers\etc\hosts. I see these options:
- Set up everything that requires administrator access as part of the VM setup
- Make it possible to run steps as adminstrator in Azure Pipelines. It looks like that currently isn't possible: https://github.com/Microsoft/azure-pipelines-tasks/issues/5073

We could make the former work, but it seems more fragile than having full control over the VM in `.azure-pipelines.yml` and rolling back to a clean state between every run.

foolip:
Hmm, this will be tricky. We have to run as administrator by default to do a lot of the setup, but can't run `MicrosoftWebDriver`as adminstrator. But in order to do `./wpt run` as non-admin, the whole checkout will also have to be done as the user, and any file that the non-admin user writes may have to be copied by the admin user to the artifacts directory. It's all probably doable, but pretty intricate.

foolip:
https://github.com/Microsoft/azure-pipelines-tasks/issues/5073 would help.

foolip:
Oh no. https://serverfault.com/questions/773038/run-as-a-different-user-without-a-password-from-an-elevated-prompt-in-windows looks like it's going to be a problem. We need either password-less `sudo` or password-less "anti-sudo", but right now it doesn't seem like either exist :(

.azure-pipelines.yml
```diff
@@ -1,122 +1,55 @@
...
 # https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema
+# https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/v2-windows
 # https://docs.microsoft.com/en-us/azure/devops/pipelines/process/multiple-phases
 # https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates
 # https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables
 # https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/index
 #
+# Self-hosted agents for Windows 10 and Insider Preview are used:
+# - 'Hosted Windows Client' is the latest Windows 10
+# - 'Hosted Windows Client Next' is 10 Windows Insider Preview
```
gsnedders: `s/10 Windows/Windows 10/`

.azure-pipelines.yml
```diff
@@ -1,122 +1,55 @@
...
 # In addition to this configuration file, the "Build pull requests from forks
 # of this repository" setting must also be enabled in the Azure DevOps project:
 # https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github#validate-contributions-from-forks
 
 trigger: none # disable builds for branches
 
 jobs:
-# The affected tests jobs are unconditional for speed, as most PRs have one or
-# more affected tests: https://github.com/web-platform-tests/wpt/issues/13936.
-- job: affected_macOS
```
gsnedders: Why are we getting rid of the macOS jobs here?

.azure-pipelines.yml
```diff
@@ -1,122 +1,55 @@
...
-  - script: no_proxy='*' ./wpt run --yes --no-manifest-update --manifest MANIFEST.json --metadata infrastructure/metadata/ --log-tbpl $(Build.ArtifactStagingDirectory)/firefox.tbpl.log --log-tbpl-level info --channel nightly firefox infrastructure/
-    displayName: 'Run tests (Firefox Nightly)'
-  - script: no_proxy='*' ./wpt run --yes --no-manifest-update --manifest MANIFEST.json --metadata infrastructure/metadata/ --log-tbpl $(Build.ArtifactStagingDirectory)/safari.tbpl.log --log-tbpl-level info --channel preview safari infrastructure/
-    displayName: 'Run tests (Safari Technology Preview)'
+  - script: python ./wpt run --yes --no-manifest-update --install-fonts --manifest MANIFEST.json --metadata infrastructure/metadata/ --log-tbpl $(Build.ArtifactStagingDirectory)/edge.tbpl.log --log-tbpl-level info edge_webdriver infrastructure/
+    displayName: 'Run tests (Edge)'
+  - script: python ./wpt run --yes --no-manifest-update --install-browser --install-fonts --manifest MANIFEST.json --metadata infrastructure/metadata/ --log-tbpl $(Build.ArtifactStagingDirectory)/firefox.tbpl.log --log-tbpl-level info --channel nightly firefox infrastructure/
```
gsnedders: We don't install Firefox above?

foolip:
This is all still an experiment. I've self-assigned, but if I remove reviewers I suspect wpt-pr-bot will just add them back. Sorry :/

gsnedders:
@foolip I was mostly having a quick look with regards to running the unit tests on Win Server!

.azure-pipelines.yml
foolip: Just to not trigger them for every test commit and waste resources.

.azure-pipelines.yml
foolip: We use `--install-browser` here.

.azure-pipelines.yml
```diff
@@ -120,3 +125,36 @@ jobs:
   - template: tools/ci/azure/tox_pytest.yml
     parameters:
       directory: tools/wpt/
+
+- job: infrastructure_win10_next
+  displayName: 'infrastructure/ tests (Windows 10 Insider Preview)'
+  #dependsOn: decision
```
foolip: This is the last thing that should be uncommented before merging this.

foolip:
This is ready for real review now. @gsnedders?

foolip:
After discussion with @thejohnjansen, renaming to focus on Windows 10 stable and not Insider Preview.

.azure-pipelines.yml
```diff
@@ -127,6 +134,39 @@ jobs:
...
+  - template: tools/ci/azure/update_hosts.yml
+  - template: tools/ci/azure/update_manifest.yml
+  - script: python ./wpt run --yes --no-manifest-update --install-fonts --manifest MANIFEST.json --metadata infrastructure/metadata/ --log-tbpl $(Build.ArtifactStagingDirectory)/edge.tbpl.log --log-tbpl-level info edge_webdriver infrastructure/
+    displayName: 'Run tests (Edge)'
+  - task: PublishBuildArtifacts@1
+    displayName: 'Publish results'
+    inputs:
+      artifactName: 'infrastructure'
+    condition: succeededOrFailed()
+  - powershell: |
```
gsnedders: Why is this not in a YML file like the hosts update?

foolip: I don't like this bit, hoping to not need it when this is merged. That's blocked on the agent pool being reliably populated with responsive agents that don't get their state messed up so easily.

gsnedders: What else is this waiting on now? You've got the review you asked for now. :)

foolip: I'll write email about the agent pool.

foolip:
@gsnedders this is now failing only because /infrastructure/reftest/size.html fails. I'll file a bug and mark that as an expected failure.

foolip:
@gsnedders this is ready for final review or even merging (squashing!) now.

Ms2ger (approved):
Please squash!

foolip:
Squahed! Thanks @Ms2ger!

PR #14306, opened 2018-11-30
[css-text] Move 'white-space-collapsing' tests to Level 4 with renaming

MurakamiShinyu:
**Property name change:** The property 'white-space-collapsing' existed in old CSS3 Text drats https://www.w3.org/TR/2011/WD-css3-text-20110215/#white-space-collapsing, after that the name changed to 'text-space-collapse' and 'text-space-trim' and moved to [Level 4](https://drafts.csswg.org/css-text-4/).

**Fix trim-inner test**: The "trim-inner" test was wrong and all browsers passed wrongly. I added brackets enclosing the span elements to prevent the normal white space collapsing between inner and around the span elements and make clear the effect of "trim-inner".


fantasai (approved):
rs=fantasai

PR #14309, opened 2018-11-30
Remove support for the timeout from testharness.js

qiuzhong:
As part of goal #11120, timeout should be removed from testharness.js
at the test-level.

### DIFF
--- a/resources/testharness.js
+++ b/resources/testharness.js
@@ -1476,9 +1476,6 @@ policies and contribution forms [3].
         if (tests.file_is_test && tests.tests.length) {
             throw new Error("Tried to create a test with file_is_test");
         }
-        if (properties.timeout) {
-            throw new Error("Tried to create a test with timeout");
-        }
         this.name = name;
 
         this.phase = tests.is_aborted ?
@@ -1489,10 +1486,9 @@ policies and contribution forms [3].
         this.index = null;
 
         this.properties = properties;
-        if (settings.test_timeout !== null) {
-            this.timeout_length = settings.test_timeout * tests.timeout_multiplier;
-        } else {
-            this.timeout_length = null;
+        this.timeout_length = settings.test_timeout;
+        if (this.timeout_length !== null) {
+            this.timeout_length *= tests.timeout_multiplier;
         }
 
         this.message = null;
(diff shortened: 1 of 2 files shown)


resources/testharness.js
```diff
@@ -1476,6 +1476,9 @@ policies and contribution forms [3].
         if (tests.file_is_test && tests.tests.length) {
             throw new Error("Tried to create a test with file_is_test");
         }
+        if (properties.timeout) {
```
foolip: I had this in my patch to look for tests that need updating, but I think we should just ignore `properties.timeout` when this change lands. In other words, I suggest removing this block of code.

Some time after `properties.assert` has also been removed, I could do another run to see if any tests pass properties at all, and then finally we'll just ignore extra arguments if they are given.

resources/testharness.js
```diff
@@ -1486,9 +1489,8 @@ policies and contribution forms [3].
         this.index = null;
 
         this.properties = properties;
-        var timeout = properties.timeout ? properties.timeout : settings.test_timeout;
-        if (timeout !== null) {
-            this.timeout_length = timeout * tests.timeout_multiplier;
+        if (settings.test_timeout !== null) {
+            this.timeout_length = settings.test_timeout * tests.timeout_multiplier;
```
foolip: I suspect that this can be collapsed a bit more, but I'm not sure. @jgraham?

foolip (approved):
LGTM % nit and would also like @jgraham to review.

resources/testharness.js
jgraham: You mean like
```js
this.timeout_length = settings.test_timeout !== null ? settings.test_timeout * tests.timeout_multiplier : null;
```
or
```js
this.timeout_length = settings.test_timeout;
if (this.timeout_length !== null) {
  this.timeout_length *= tests.timeout_multiplier;
}
```

resources/testharness.js
qiuzhong: Removed.

resources/testharness.js
qiuzhong: Updated.

foolip: I was wondering if more layers couldn't be peeled away, having `timeout_length` on `Test` instances seems like it should be unnecessary. With no test-level timeouts, when the harness times out any async tests that are still running have timed out.

But even after going through all the matches for "timeout" over and over I'm not too confident I understand the setup, so this simplification is fine with me.

foolip (approved):
@jgraham is busy this week. I've reviewed this, going ahead and merging.

PR #14310, opened 2018-11-30
Convert SVG 1.1 marker tests to reftests.

Ms2ger:
Correct result of marker tests because SVG 2 changes marker overflow auto to be visible

Update to resolve anti-aliasing differences.

Also addressed some other minor comments made by AmeliaBR.

Fixes #4015.


svgeesus (approved):
Cheers, @Ms2ger  !

PR #14313, opened 2018-11-30
Fix path argument to wpt lint.

Ms2ger:


jgraham (approved):

Ms2ger:
The tests found the undocumented requirement that `all_filesystem_paths` should only yield relative paths.

gsnedders:
That would be my fault… IIRC, we had some other bug involving absolute paths that led to this which you can likely find through blame?

jgraham (approved):

PR #14315, opened 2018-11-30
 Fetch tags on initial clone 

jgraham:


tools/ci/start.sh
```diff
@@ -0,0 +1,39 @@
...
+sudo earlyoom -p -r 60 --prefer '(chrome|firefox)' --avoid 'python' &
+
+if [[ ! `git rev-parse --verify -q ${REVISION}` ]];
+then
+    # But if for some reason the commit under test isn't in that range, we give in and
+    # fetch everything
+    git fetch -q --unshallow ${REMOTE}
+    git rev-parse --verify ${REVISION}
+fi
+git checkout -b build ${REVISION}
```
jugglinmike: If we keep this part in the Docket image, we'll be able to land changes to the new script without fear of affecting pull requests that are based on some commit behind the HEAD of `master`. It will also mean pull requests which modify the new file will be run with the modifications.

jgraham: This was just wrong. Good catch!

gsnedders (approved):

gsnedders:
@jugglinmike do you want to have another look at this before it lands?

jugglinmike (approved):
Thanks, @gsnedders. This looks good to me

PR #14334, opened 2018-12-03
[css-text] Update 'line-break' tests

MurakamiShinyu:
### Fix test file name numbering

The problem was that `line-break-normal-*` and `line-break-strict-*` had diffrent numbering for same ranges, e.g. `line-break-normal-021` and `line-break-strict-014` for same iteration marks. It was confusing, so I changed the numbering for 'normal' to be the same as for 'strict'.

- line-break-normal-021  → line-break-normal-014
- line-break-normal-022  → line-break-normal-015
- line-break-normal-023a → line-break-normal-016a
- line-break-normal-023b → line-break-normal-016b
- line-break-normal-024a → line-break-normal-017a
- line-break-normal-024b → line-break-normal-017b

### Fix tests for Inseparatable characters

The tests for Inseparatable characters `line-break-strict-015` and `line-break-normal-015` (was `line-break-normal-022`) had wrong assertion "does not allow line breaking **before** inseparable characters". This should be "does not allow line breaking **between** inseparable characters".

I changed the tests to test line breaking **between** (not **before**) two inseparable characters.

### Fix tests for Centered punctuation marks

The tests for Centered punctuation marks `line-break-strict-016a` and `line-break-normal-016a` (was `line-break-normal-023a`) contained tests for EAW=Narrow characters (U+003A, U+003B, U+0021, U+003F) that are excluded in the current 'line-break' spec.

### Fix tests for Postfixes

The tests for Postfixes `line-break-strict-017a` and `line-break-normal-017a` (was `line-break-normal-024a`) contained tests for EAW=Narrow characters (U+0025 and U+00A2) that are excluded in the current 'line-break' spec.

### Add tests for Prefixes

Added the tests for line breaking after Prefixes `line-break-*-018`.

### Add 'loose' and 'normal' values tests that were missing

- line-break-loose-011 .. line-break-loose-017b
- line-break-normal-011 .. line-break-normal-013

### Adjust width of the test block

I changed the width of the test block `p.test, p.control {...}` that was `width: 10em` to `width: 10.2em`. This extra (.2em) width is needed to prevent unexpected line breaking caused by some symbol characters wider than 1em.

I found this problem with `line-break-loose-018`. The width of U+2116 "№" is slightly wider than 1em and if the block width is 10em the line breaking occurs before that character, and the test always fails.

### DIFF
--- a/css/css-text/line-break/line-break-loose-011.xht
+++ b/css/css-text/line-break/line-break-loose-011.xht
@@ -2,14 +2,14 @@
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">
 <html xmlns="http://www.w3.org/1999/xhtml">
 	<head>
-		<title>CSS Text Test: line-break - strict and iteration marks</title>
-		<!-- iteration marks -->
+		<title>CSS Text Test: line-break - loose and Japanese small kana</title>
+		<!-- Japanese small kana -->
 		<link rel="author" title="Taka Oshiyama" href="mailto:takaoshiyama@gmail.com" />
 		<link rel="author" title="Shinyu Murakami" href="mailto:murakami@vivliostyle.org" />
 		<link rel="help" title="5.3. Line Breaking Rules: the 'line-break' property" href="https://drafts.csswg.org/css-text-3/#line-break-property" />
-		<link rel="match" href="reference/line-break-strict-014-ref.xht" />
+		<link rel="match" href="reference/line-break-loose-011-ref.xht" />
 		<meta http-equiv="content-language" content="en, ja" />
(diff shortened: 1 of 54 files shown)


r12a (commented):
I don't currently have time to review these tests properly, but i'll just note that i have similar-looking tests at https://w3c.github.io/i18n-tests/results/line-breaks-jazh (click on the link in the left column to run the tests).  I note that because my tests produce js-based results - you may want to consider doing something similar, rather than making these manual tests (iiuc).

MurakamiShinyu:
This pull request is not for adding new tests, but for updating existing css/css-text/line-break/ tests that were made in early 2011. The css-text-3 'line-break' spec has been changed since then but these tests had not been updated. In this PR, I updated the outdated tests and also added some missing tests.

@r12a Yes, these tests are similar to your https://w3c.github.io/i18n-tests/results/line-breaks-jazh (thanks!), but non-JavaScript and can be used to test layout engines without JavaScript.

fantasai (approved):
rs=fantasai

PR #14345, opened 2018-12-03
Make sure that Actions.perform is always passed a list

### LINKED ISSUE #14343: Actions.perform function should only accept a list
In Actions.perform function, it says that ":param actions: List of input source action sequences." But in ActionSequenceAction class, self.protocol.action_sequence.send_actions({"actions": actions}), it sends the actions as a dictionary, which should be self.protocol.action_sequence.send_actions(actions), "actions" is a list.

https://github.com/web-platform-tests/wpt/blob/e94a910ee358c57dc18cbae4b4874d69ef26e569/tools/webdriver/webdriver/client.py#L213
https://github.com/web-platform-tests/wpt/blob/31664a8d285f2df3c86f7051c8e2a45a75bbd339/tools/wptrunner/wptrunner/executors/base.py#L591

LanWei22:
Fixes #14343 
When calling action_sequence.send_actions, we should pass a list not dictionary, so Actions.perform will get a list.


NavidZ:
You need to move base.py to the right address (tools/wptrunner/wptrunner/executors/base.py) to show only the right changes.

JohnChen0:
While this change fixes the test on Chrome and Safari, it breaks the test on Firefox. The reason is Firefox uses a different executor, tools/wptrunner/wptrunner/executors/executormarionette.py, which expects the actions to be specified in the previous format.

To fix the test on Chrome without breaking Firefox, I think the right place is in tools/wptrunner/wptrunner/executors/executorwebdriver.py:

```python
class WebDriverActionSequenceProtocolPart(ActionSequenceProtocolPart):
    ...
    def send_actions(self, actions):
        self.webdriver.actions.perform(actions['actions'])
```

jgraham (changes requested):
Per @JohnChen0 's comment

SummerLW:
Thank @JohnChen0 for your comment, I have made the change.

LanWei22 (commented):
I have submitted the change, please take a look again, thank you.

SummerLW:
@jgraham I fixed the check failure, could you please take a look, thank you?

jgraham (approved):

PR #14353, opened 2018-12-04
[css-text] Some fixes on 'word-break-break-all-*' tests

MurakamiShinyu:
### word-break-break-all-010 and word-break-break-all-011

- `line-height: 1em` was specified on the `.test {...}` but not on the `.red {...}`. As a result, the line-height of the `.red` block depended on the browser, and the red color appeared on Firefox.

### word-break-break-all-020

Fix erros (typo?)
- `div::nth-of-type` → `div:nth-of-type`
- `break-word` → `break-all`

### DIFF
--- a/css/css-text/word-break/reference/word-break-break-all-020-ref.html
+++ b/css/css-text/word-break/reference/word-break-break-all-020-ref.html
@@ -8,9 +8,9 @@ div {
   width: 2em;
 }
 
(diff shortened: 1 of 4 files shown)


fantasai (commented):
rs=fantasai

fantasai (approved):

PR #14356, opened 2018-12-04
[wptrunner] Support any Sauce Connect argument

jugglinmike:
Implement the command-line argument `--sauce-connect-arg` as a
pass-through to the Sauce Connect binary. This gives users more control
over the way the binary is configured which is particularly important in
combination with the existing `--sauce-connect-binary` argument (since
the list of available arguments is subject to change between releases).


jgraham (approved):

jugglinmike:
Thanks James and Geoffrey!

Issue #13127, opened 2018-09-20
[cookie-store] Integrate with async cleanup


jugglinmike:
Many tests for async cookies need to perform asynchronous "test cleanup" logic.  Previously, testharness.js did not offer support for this use case, so various workarounds were implemented here, e.g.

>     // Workaround because add_cleanup doesn't support async functions yet.
>     // See https://github.com/web-platform-tests/wpt/issues/6075
>     async function async_cleanup(cleanup_function) {
>       try {
>         await cleanup_function();
>       } catch (e) {
>         // Errors in cleanup functions shouldn't result in test failures.
>       }
>     }

gh-6075 was recently resolved, meaning testharness.js now honors promise values returned by `add_cleanup` functions. The tests should be refactored to use this API. One of the design decisions for the new feature is particularly relevant for this test suite:

> - Rejected promises should cause test failures, but if the test body itself has failed, the test error should take precedence.

That conflicts with the expectation expressed in the code sample above. If the maintainers wish to be permissive of asynchronous errors, they will have to persist the logic to provide that.

/cc @pwnall

pwnall:
I don't care enough about this to maintain custom logic. Happy to have cleanup errors cause the test to fail. We'll be switching to `add_cleanup` as soon as someone finds some time to do that.

Thank you very much for your awesome work, @jugglinmike!

jarson17:
> Many tests for async cookies need to perform asynchronous "test cleanup" logic. Previously, testharness.js did not offer support for this use case, so various workarounds were implemented here, e.g.
> 
> > ```
> > // Workaround because add_cleanup doesn't support async functions yet.
> > // See https://github.com/web-platform-tests/wpt/issues/6075
> > async function async_cleanup(cleanup_function) {
> >   try {
> >     await cleanup_function();
> >   } catch (e) {
> >     // Errors in cleanup functions shouldn't result in test failures.
> >   }
> > }
> > ```
> 
> [gh-6075](https://github.com/web-platform-tests/wpt/issues/6075) was recently resolved, meaning testharness.js now honors promise values returned by `add_cleanup` functions. The tests should be refactored to use this API. One of the design decisions for the new feature is particularly relevant for this test suite:
> 
> > * Rejected promises should cause test failures, but if the test body itself has failed, the test error should take precedence.
> 
> That conflicts with the expectation expressed in the code sample above. If the maintainers wish to be permissive of asynchronous errors, they will have to persist the logic to provide that.
> 
[...]

PR #14364, opened 2018-12-05
[cookie-store] Improve cleanup logic

jugglinmike:
While researching for a WPT infrastructure project, I found these tests to be behaving erratically. In a handful of cases, the cleanup invocation was not properly awaited. Since [we've been meaning to take advantage of the new "async cleanup" functionality](https://github.com/web-platform-tests/wpt/issues/13127), I took the opportunity to use that API. I'm submitting this in two commits to make the bug fix easier to identify.

To validate this change set, I ran the tests in Chromium (70.0.3538.77) on `master` (3960fff73862bd9745f8b0add9358e923f916dcb) and on this branch via the command

     $ xvfb-run --auto-servernum ./wpt run --no-pause --log-mach - --log-mach-level DEBUG --channel dev --no-manifest chrome cookie-store/

The results were identical:

On master:

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 338 checks (35 tests, 303 subtests)
    Expected results: 328
    Unexpected results: 10
      subtest: 10 (10 fail)

With patch applied:

    web-platform-test
    ~~~~~~~~~~~~~~~~~
    Ran 338 checks (35 tests, 303 subtests)
    Expected results: 328
    Unexpected results: 10
      subtest: 10 (10 fail)

### DIFF
--- a/cookie-store/cookieListItem_attributes.tentative.https.window.js
+++ b/cookie-store/cookieListItem_attributes.tentative.https.window.js
@@ -1,15 +1,5 @@
 'use strict';
 
-// Workaround because add_cleanup doesn't support async functions yet.
-// See https://github.com/web-platform-tests/wpt/issues/6075
-async function async_cleanup(cleanup_function) {
-  try {
-    await cleanup_function();
-  } catch (e) {
-    // Errors in cleanup functions shouldn't result in test failures.
-  }
-}
-
 const kCurrentHostname = (new URL(self.location.href)).hostname;
 
 const kOneDay = 24 * 60 * 60 * 1000;
@@ -23,6 +13,7 @@ promise_test(async testCase => {
   await cookieStore.delete('cookie-name');
 
   await cookieStore.set('cookie-name', 'cookie-value');
+  testCase.add_cleanup(() => cookieStore.delete('cookie-name'));
 
   const cookie = await cookieStore.get('cookie-name');
   assert_equals(cookie.name, 'cookie-name');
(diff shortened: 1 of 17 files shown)


jugglinmike:
It looks like these tests are still unstable even with my patch applied. I tried to reproduce the instability locally using the following command:

    ./wpt run --no-pause --log-mach - --channel dev --verify chrome cookie-store

...but that didn't catch anything on my local system (Chromium 70.0.3538.77) or in the Docker image used by Taskcluster (Google Chrome 72.0.3622.0 dev). Even the exact command reported in the Taskcluster logs (too lengthy to post here) didn't demonstrate the instability. Those `CRASH`es are suspicious, though. I don't think it's a memory limitation because the utilization is documented in the logs:

    2018-12-05 01:37:06,210 - tc-run - INFO - Identifying tests affected in range 'HEAD^'...
    mem avail: 29462 MiB (96 %), swap free: 0 MiB (0 %)
    2018-12-05 01:37:34,339 - tc-run - INFO - Identified 17 affected tests
    
    [...]
    
    PID 1140 | Only local connections are allowed.
    mem avail: 27550 MiB (89 %), swap free: 0 MiB (0 %)
    PID 1452 | Starting ChromeDriver 2.44.609551 (5d576e9a44fe4c5b6a07e568f1ebc753f1214634) on port 4444
    
    [...]
    
    PID 10044 | Only local connections are allowed.
    mem avail: 28787 MiB (93 %), swap free: 0 MiB (0 %)
    PID 10320 | Starting ChromeDriver 2.44.609551 (5d576e9a44fe4c5b6a07e568f1ebc753f1214634) on port 4444

It's strange that the container has no swap memory, but I'm not sure that's a problem. It doesn't seem as though we'd need to do any paging with 28 gigabytes of free memory. Unfortunately, I can't replicate that condition via Docker running on my system due to kernel limitations.

@Hexcles do you have thoughts on any of this?

Hexcles:
Right, our Docker containers don't have swap, which shouldn't have any negative affects. We should never use anything even close to the total available memory (guaranteed to be at least 8GB).

pwnall (approved):
LGTM.

Thank you very much for the cleanup! I don't know what's up with the bots, but this looks like a step forward for the tests.

jugglinmike:
This was addressed in Chromium directly and merged to WPT via gh-17949. I've verified that patch addressed all of the instances of asynchronous cleanup that we identified here. It's good that the change eventually made it through stability checks!

PR #14374, opened 2018-12-05
Move overscrollBehavior-manual.html into css/css-overscroll-behavior

fred-wang:
This property does not belong to the CSSOM View spec.


smfr (approved):

fred-wang:
@smfr Thanks. It seems you don't have write access :-/

Ms2ger (approved):

PR #14382, opened 2018-12-05
Run affected tests on Taskcluster without the patch

Hexcles:
This PR adds one task to each browser on Taskcluster for PRs to run the affected tests *without* the patch by running the tests on the first parent of the GitHub merge commit (which is the latest revision of the target branch of the PR when the merge commit is generated by GitHub, shortly after the PR is updated).

This is crucial for reliable regression detection, because without it, we have to use heuristics to compare the PR results against some revision on master.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -98,14 +98,25 @@ tasks:
           each(browser):
             $map:
               - name: wpt-${browser.name}-${browser.channel}-stability
+                checkout: FETCH_HEAD
+                diff_range: HEAD^
                 description: >-
                   Verify that all tests affected by a pull request are stable
                   when executed in ${browser.name}.
                 extra_args: '--verify'
               - name: wpt-${browser.name}-${browser.channel}-results
+                checkout: FETCH_HEAD
+                diff_range: HEAD^
                 description: >-
                   Collect results for all tests affected by a pull request in
                   ${browser.name}.
+                extra_args: '--no-fail-on-unexpected --log-wptreport=../artifacts/wpt_report.json'
+              - name: wpt-${browser.name}-${browser.channel}-results-without-changes
+                checkout: FETCH_HEAD^
+                diff_range: FETCH_HEAD
+                description: >-
+                  Collect results for all tests affected by a pull request in
+                  ${browser.name} but without the changes in the PR.
                 extra_args: '--no-fail-on-unexpected --log-wptreport=../artifacts/wpt_report.json'
             each(operation):
               taskId: {$eval: 'as_slugid(operation.name)'}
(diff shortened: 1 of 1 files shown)


Hexcles:
I realize this PR doesn't really depend on #14354 , so I got rid of the irrelevant commits. @jugglinmike , could you take a look?

Example: https://tools.taskcluster.net/task-group-inspector/#/ElImMxb-Rt-NsiNm7t23eg

.taskcluster.yml
```diff
@@ -98,15 +98,26 @@ tasks:
...
                   when executed in ${browser.name}.
                 extra_args: '--verify'
               - name: wpt-${browser.name}-${browser.channel}-results
+                checkout: FETCH_HEAD
+                diff_range: HEAD^
                 description: >-
                   Collect results for all tests affected by a pull request in
                   ${browser.name}.
                 extra_args: '--no-fail-on-unexpected --log-wptreport=../artifacts/wpt_report.json'
+              - name: wpt-${browser.name}-${browser.channel}-results-without-patch
```
lukebjerring: Maybe ...-master-results for the name?

Hexcles: I don't think we can say the target branch is always master (one can open a PR against any branch, and `FETCH_HEAD^` will always be that target branch). Besides, "without patch" is the canonical suffix in Chromium CI for tasks like this (but I'm not sure whether non-Chromies would be able to immediately tell the meaning of it).

.taskcluster.yml
```diff
@@ -125,7 +136,7 @@ tasks:
                 owner: ${event.pull_request.user.login}@users.noreply.github.com
                 source: ${event.repository.url}
               payload:
-                image: hexcles/web-platform-tests:0.23
+                image: harjgam/web-platform-tests:0.25
```
Hexcles: Note: this was an oversight when the Docker image was upgraded last time.

interfaces/beacon.idl
```diff
@@ -4,5 +4,5 @@
 // Source: Beacon (https://w3c.github.io/beacon/)
 
 partial interface Navigator {
-    boolean sendBeacon(USVString url, optional BodyInit? data = null);
+    boolean sendBacon(USVString url, optional BodyInit? data = null);
```
foolip: Revert this lovely bacon :)

foolip (approved):

jgraham (approved):
Revert test change ofc.

.taskcluster.yml
foolip: I called it `sans_changes` and "without changes" in https://github.com/web-platform-tests/wpt/pull/14156, so `-without-changes` here would work, I can align with that.

Hexcles: Alright, I'll go with "without changes".

PR #14383, opened 2018-12-06
[eme] Check if encryption scheme is supported

jrummell-chromium:
In [1] there is a proposal to add the ability for querying the encryption
schemes supported by the CDM to MediaKeySystemConfiguration. Adding a test
to check this.

[1] https://github.com/WICG/encrypted-media-encryption-scheme/blob/master/explainer.md

### DIFF
--- a/encrypted-media/scripts/check-encryption-scheme.js
+++ b/encrypted-media/scripts/check-encryption-scheme.js
@@ -1,0 +1,42 @@
+function runTest(config, qualifier)
+{
+  function checkEncryptionScheme(encryptionScheme)
+  {
+    var simpleConfig = getSimpleConfiguration();
+    assert_greater_than(simpleConfig[0].audioCapabilities.length, 0);
+    simpleConfig[0].audioCapabilities.forEach(function(capability) {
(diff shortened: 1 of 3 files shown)


encrypted-media/scripts/check-encryption-scheme.js
```diff
@@ -0,0 +1,42 @@
...
+                }
+
+                // If "encryptionScheme" is supported, it should be returned.
+                assert_equals(capability.encryptionScheme, encryptionScheme);
+              }
+              return Promise.resolve('Supported');
+            },
+            function error() {
+              // CDM does not support "encryptionScheme". Test should
+              // still pass.
```
xhwang-chromium: nit: does this fit in one line?

xhwang-chromium (approved):
lgtm % nit

encrypted-media/scripts/check-encryption-scheme.js
jrummell-chromium: This was done by the Chromium formatter. Fixed.

foolip (approved):

PR #14384, opened 2018-12-06
[eme] Check if HDCP policy is supported

jrummell-chromium:
In [1] there is a proposal to add the ability to know before fetching content
if HDCP (and what version) can be enforced. This adds a test to check if it
is supported.

[1] https://github.com/WICG/hdcp-detection/blob/master/explainer.md


encrypted-media/scripts/check-hdcp-policy.js
```diff
@@ -0,0 +1,26 @@
...
+    return navigator.requestMediaKeySystemAccess(config.keysystem, getSimpleConfiguration())
+        .then(function(access) {
+          return access.createMediaKeys();
+        })
+        .then(function(mediaKeys) {
+          // As HDCP policy depends on the hardware running this test,
+          // don't bother checking the result returned as it may or
+          // may not be supported. This simply verifies that
+          // getStatusForPolicy() exists and doesn't blow up.
+          return mediaKeys.getStatusForPolicy({minHdcpVersion: hdcpPolicy});
```
xhwang-chromium: This might return a rejected promise, e.g. for clear key. Will this cause the test to fail? For the context, see https://github.com/WICG/hdcp-detection/issues/2

encrypted-media/scripts/check-hdcp-policy.js
```diff
@@ -0,0 +1,26 @@
...
+        });
+  }
+
+  promise_test(
+      () => checkHdcpPolicy(''),
+      testnamePrefix(qualifier, config.keysystem) +
+          ' support for empty HDCP policy.');
```
xhwang-chromium: s/policy/version

encrypted-media/scripts/check-hdcp-policy.js
```diff
@@ -0,0 +1,26 @@
+function runTest(config, qualifier)
+{
+  function checkHdcpPolicy(hdcpPolicy)
```
xhwang-chromium: This should be something like checkStatusForMinHdcpVersionPolicy(hdcpVersion)

encrypted-media/scripts/check-hdcp-policy.js
xhwang-chromium: see above. s/hdcpPolicy/hdcpVersion

encrypted-media/clearkey-check-hdcp-policy.https.html
```diff
@@ -0,0 +1,32 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset=utf-8>
+    <title>Encrypted Media Extensions: Check HDCP policy with Clear Key</title>
```
xhwang-chromium: Here and below and in the file names, to be accurate, this should be something like "Check status for minHdcpVersion policy"

jrummell-chromium: Done. Kept the files names simple with check-hdcp-version.

encrypted-media/scripts/check-hdcp-policy.js
jrummell-chromium: Done.

encrypted-media/scripts/check-hdcp-policy.js
jrummell-chromium: The test will fail, but if you look at the results you will see the failure due to error NotSupportedException. I think this is better for developers as they will see that it is not supported on the platform (by clear key, for example) and recognize they need to handle the failure. This is similar to the persistent-session tests (that are optional, but it is useful to know that persistent sessions aren't supported).

jrummell-chromium: Done.

encrypted-media/scripts/check-hdcp-policy.js
jrummell-chromium: Done.

foolip (approved):

PR #14392, opened 2018-12-06
Try Safari Technology Preview 71

foolip:


foolip:
Testing if everything will blow up when https://github.com/Homebrew/homebrew-cask-versions/pull/6665 is merged. Nothing would cause infrastructure/ tests to be run again when it is, so breakage would show up first on some unlucky PR.

foolip:
infrastructure/ tests still pass!

PR #14395, opened 2018-12-06
Remove stray property from memory/constructor.any.js.

Ms2ger:
It was copied from the WebAssembly.Table tests.


Ms2ger:
From https://github.com/WebAssembly/spec/pull/916. CC @littledan

littledan (approved):

PR #14400, opened 2018-12-06
Made the trusted-types tests run correctly in a vanilla wptrunner.

koto:
Tests that set up trusted-types-enforcing CSP now execute in a separate iframe, not to clash with the harness using the DOM directly.

### DIFF
--- a/trusted-types/TrustedTypePolicyFactory-createPolicy-createXYZTests.tentative.html
+++ b/trusted-types/TrustedTypePolicyFactory-createPolicy-createXYZTests.tentative.html
@@ -1,7 +1,9 @@
 <!DOCTYPE html>
+<head>
 <script src="/resources/testharness.js" ></script>
 <script src="/resources/testharnessreport.js"></script>
 <script src="support/helper.sub.js"></script>
+</head>
 <body>
 <script>
   //HTML tests
--- a/trusted-types/TrustedTypePolicyFactory-createPolicy-cspTests-noNamesGiven.tentative.html
+++ b/trusted-types/TrustedTypePolicyFactory-createPolicy-cspTests-noNamesGiven.tentative.html
@@ -1,15 +1,20 @@
 <!DOCTYPE html>
+<head>
 <script src="/resources/testharness.js" ></script>
 <script src="/resources/testharnessreport.js"></script>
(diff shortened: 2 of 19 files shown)


mikewest:
@otherdaniel, mind taking a look?

mikewest:
(We talked about this briefly yesterday, and @otherdaniel had interesting ideas about modifying `testharness.js` instead of all the tests.)

jgraham:
OOI what are the requirements here?

mikewest:
@jgraham: Trusted Types gives developers the ability to change the requirements for things like `innerHTML`. The test harness uses `innerHTML` to dump out a table of results. This explodes in tests that disable string-based `innerHTML` rendering. @otherdaniel's approach would be to teach `testharness.js` to use trusted types if the browser supports them, just as any other library would want to do.

koto:
I went down that path initially a couple of months ago (the patch to
harness is in the polyfill repository under tests/).

However, the patch was quickly broken by new changes to testharness, and
surfaced another errors (e.g. chromedriver likes to use innerhtml as well,
and its code is outside of the wpt repository). I could not even debug some
of the errors, as they were not throwing in the console.

For practical reasons, it might be easier to run the CSP requiring tests in
a separate iframe, which is what I did here, which feels less invasive to
the harness infrastructure.

otherdaniel:
My version of solving this in testharness ends up looking almost exactly like koto's, with likely the same drawbacks. (https://chromium-review.googlesource.com/c/chromium/src/+/1371874) The main plus would be that I've modified all the tests to allow for a 'testharness' policy, so that all WPT users can be sure that a policy with that name can be built for testharness' own use.

On second thought, I'm no longer sure that's the better path than what koto suggests here.

koto:
Friendly ping on this one.

fred-wang:
Do we still need this?

Isn't the issue addressed by https://github.com/web-platform-tests/wpt/pull/44561 ?

otherdaniel:
> Do we still need this?
> Isn't the issue addressed by #44561 ?

I think this has indeed be handled by #44561, and I think this PR is now obsolete.

fred-wang:
Closing per previous comment.

PR #14402, opened 2018-12-06
Improve support and add infrastructure test for file upload

jgraham:


tools/wptrunner/wptrunner/browsers/firefox.py
```diff
@@ -248,6 +248,7 @@ def start(self, group_metadata=None, **kwargs):
             "network.proxy.type": 0,
             "places.history.enabled": False,
             "network.preload": True,
+            "dom.file.createInChild": True
```
andreastt: I’m surprised this isn’t set automatically by Marionette:
https://searchfox.org/mozilla-central/source/testing/marionette/components/marionette.js#179

Are you overwriting it somewhere else?

jgraham: We don't use GeckoInstance.

andreastt: That’s not GeckoInstance. We set it as a recommended preference when Firefox starts.

jugglinmike (commented):
Can you add a few tests?

tools/wptrunner/wptrunner/browsers/firefox.py
jgraham: Oh, right, sorry but we don't get marionette to start firefox either.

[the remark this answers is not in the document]
jgraham: What mozbase patch? I'm confused :)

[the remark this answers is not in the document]
jugglinmike: A comment that describes these properties (but is cropped from the patch summary) references a Bugzilla ticket: https://github.com/web-platform-tests/wpt/blob/fc87d9b363525fdcc5b5053a7c9624614c3fd730/tools/wptrunner/wptrunner/browsers/firefox.py#L247

```python
            # TODO: Remove preferences once Firefox 64 is stable (Bug 905404)
            "network.proxy.type": 0,
            "places.history.enabled": False,
            "network.preload": True,
```

That includes a patch which introduces the existing preferences. If I'm reading this correctly, then we *won't* want to remove this new preference "once Firefox 64 is stable," so it should be declared elsewhere.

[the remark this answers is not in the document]
jgraham: Oh, I see! You're quite right.

Ultimately I should put this in one of the prefs files in m-c, but this is a little easier for now and I'm going to do a big clean up there later anyway.

jgraham:
Added some tests and updated.

tools/wptrunner/wptrunner/browsers/firefox.py
andreastt: Right, but you start Firefox with `--marionette` right? That should initialise the server and set recommended prefs unless `marionette.prefs.recommended` is set to false.

I’m not saying adding this pref is wrong, but I don’t understand that you need it.

tools/wptserve/wptserve/pipes.py
```diff
@@ -354,6 +354,8 @@ def sub(request, response, escape_type="html"):
       sha224, sha256, sha384, and sha512. For example:
 
         {{file_hash(md5, dom/interfaces.html)}}
+    path(filepath)
```
jugglinmike:
```suggestion
    fs_path(filepath)
```

tools/wptrunner/wptrunner/browsers/firefox.py
jgraham: Ah, the unittest pref set which we apply set that to false: https://searchfox.org/mozilla-central/source/testing/profiles/unittest/user.js#193. Which makes some sense since it makes things a bit less explicit in terms of what's enabled.

andreastt: OK, then I think it makes sense to re-enable it here rather than changing the underlying file.

jugglinmike (approved):
Nicely done, @jgraham

jugglinmike:
Since you've squashed changes for earlier review feedback, I take it that you'd like this work landed as distinct commits. The most recent commit from me probably shouldn't be merged on its own, so I've taken the liberty to squash it. Hope that's okay!

The original version of this branch is available here:

https://github.com/web-platform-tests/wpt/compare/master...bocoup:file_upload-orig

PR #14415, opened 2018-12-07
webdriver: fix expected key code for "Unidentified" keys

andreastt:
Per the UI Events specification
(https://w3c.github.io/uievents-code/#key-legacy), keys such as
\u00e0 (i.e., à) should have a code of "Unidentified".  Some tests
in key.py expect an empty value, which seems to be an oversight.

This PR supersedes https://github.com/web-platform-tests/wpt/pull/12013.


andreastt:
Yes, probably if it’s not there already. This is really a question for @burg, since I just rebased his original PR.

whimboo:
@jgraham would you mind having a look at this PR? In our implementation for Firefox we also still fallback to an empty string.

whimboo:
@jgraham all the webdriver tests are passing which is interesting. Maybe I miss something?

PR #14423, opened 2018-12-07
Add some tests that check SDP after transceiver stop

youennf:


webrtc/RTCRtpTransceiver-stop.html
```diff
@@ -0,0 +1,60 @@
...
+    answerer.setRemoteDescription(offer)
+  ]);
+  const answer = await answerer.createAnswer();
+  await Promise.all([
+    answerer.setLocalDescription(answer),
+    offerer.setRemoteDescription(answer)
+  ]);
+}
+
+promise_test(async ()=> {
```
henbos: nit: double indent

webrtc/RTCRtpTransceiver-stop.html
```diff
@@ -0,0 +1,60 @@
...
+    const pc1 = new RTCPeerConnection();
+
+    pc1.addTransceiver("audio", { direction: "sendonly" });
+    pc1.addTransceiver("video");
+    pc1.getTransceivers()[0].stop();
+
+    const offer = await pc1.createOffer();
+    assert_false(offer.sdp.includes("m=audio"), "offer should not contain an audio m section");
+}, "An offer should be generated even if there is no valid transceiver");
```
alvestrand: This description confuses me, since there is a valid video transceiver.

henbos: Incorrect test name

youennf: Title should be changed

webrtc/RTCRtpTransceiver-stop.html
```diff
@@ -0,0 +1,60 @@
...
+    pc1.addTransceiver("audio");
+
+    await negotiate(pc1, pc2);
+
+    pc1.getTransceivers()[0].direction = "sendonly";
+    pc1.getTransceivers()[0].stop();
+
+    const offer = await pc1.createOffer();
+    assert_true(offer.sdp.includes("a=sendonly"), "m section should be sendonly");
+}, "A stopped sendonly transceiver should generate an inactive m section in the offer");
```
alvestrand: This is testing that a stopped transceiver is present. That seems odd; you should end up with no m=audio section in the SDP.

webrtc/RTCRtpTransceiver-stop.html
henbos: Question: does a stopped m= section not show up in the offer?

webrtc/RTCRtpTransceiver-stop.html
alvestrand: I see now - this test does negotiation, so it's already generated the section.
This test is not testing for port=0, which is the definition of stopped, though.

alvestrand:
Needs more review, but very good to have.

webrtc/RTCRtpTransceiver-stop.html
```diff
@@ -0,0 +1,60 @@
...
+  ]);
+  const answer = await answerer.createAnswer();
+  await Promise.all([
+    answerer.setLocalDescription(answer),
+    offerer.setRemoteDescription(answer)
+  ]);
+}
+
+promise_test(async ()=> {
+    const pc1 = new RTCPeerConnection();
```
youennf: add clean up

webrtc/RTCRtpTransceiver-stop.html
youennf: Added an assert for port=0

webrtc/RTCRtpTransceiver-stop.html
youennf: Not as per my reading of https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-24#section-5.2
Firefox and Safari behaves accordingly there

youennf:
Updated the tests according review and added some more checks.
As per my reading of https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-24#section-5.2, the first test is correct.
Firefox and Safari behaves the same for all tests except the 3rd one where Firefox always set the m section as inactive.

henbos (approved):
LGTM

PR #14426, opened 2018-12-09
Add tests for clicking the child of a button element

Zirro:
Three tests for https://github.com/jsdom/jsdom/pull/2450.

The last test passes in Firefox but fails in Chrome and Safari. I think the test is correct based on the [Dispatch an event](https://dom.spec.whatwg.org/#dispatching-events) algorithm, in that non-bubbling events should not cause the activation behavior of parents to run.

### DIFF
--- a/html/semantics/forms/the-button-element/button-click-submits.html
+++ b/html/semantics/forms/the-button-element/button-click-submits.html
@@ -146,4 +146,65 @@ test(t => {
 
 }, "clicking a button inside a disabled fieldset's legend *should* trigger submit");
 
+async_test(t => {
+
+  const form = document.createElement("form");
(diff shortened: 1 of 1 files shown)


tkent-google (approved):

annevk:
I think this is wrong. The way "dispatch an event" does this is by setting _activationTarget_ while building the event path. That doesn't care about bubbling, right?

Can we invert the pass condition here and file an issue against Firefox?

Zirro:
@annevk Okay, I've filed the following bugs:
https://bugs.webkit.org/show_bug.cgi?id=192741
https://bugs.chromium.org/p/chromium/issues/detail?id=915476

PR #14442, opened 2018-12-10
[selectors] selector list with invalid selector

ewilligers:
If a selector is invalid, the entire selector list is invalid.
https://drafts.csswg.org/selectors-4/#grouping

### DIFF
--- a/css/selectors/contains-invalid.html
+++ b/css/selectors/contains-invalid.html
@@ -1,0 +1,64 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>CSS Selectors: selector list containing invalid selector</title>
+    <link rel="help" href="https://drafts.csswg.org/selectors-4/#grouping">
+    <meta name="assert" content="If a selector is invalid, the entire selector list is invalid.">
+    <meta name="assert" content="Pseudo-elements are not valid within :is :not :where.">
+    <script src="/resources/testharness.js"></script>
+    <script src="/resources/testharnessreport.js"></script>
+    <style>
+      .a, .b, .c {
+        color: lime;
+      }
+
+      .a, :nonsense {
+        color: red;
+      }
+
+      /* Pseudo-elements cannot be represented by the matches-any pseudo-class; they are not valid within :is(). */
+      .b, :is(::before) {
+        color: red;
+      }
+      .b, :not(::before) {
+        color: red;
+      }
+      .b, :where(::before) {
+        color: red;
+      }
+
+      .c:is(*, :nonsense) {
+        color: red;
+      }
+      .c:not(*, :nonsense) {
+        color: red;
+      }
+      .c:where(*, :nonsense) {
+        color: red;
+      }
+    </style>
+  </head>
+  <body>
+    <div id="a" class="a">
+    </div>
+    <div id="b" class="b">
+    </div>
+    <div id="c" class="c">
+    </div>
+    <script>
+      const lime = "rgb(0, 255, 0)";
+
+      test(() => {
+        assert_equals(getComputedStyle(a).color, lime);
+      }, "If a selector is invalid, the entire selector list is invalid.");
+
+      test(() => {
+        assert_equals(getComputedStyle(b).color, lime);
+      }, "Pseudo-elements are not valid within :is :not :where.");
+
+      test(() => {
+        assert_equals(getComputedStyle(c).color, lime);
+      }, "If a selector in :is :not :where is invalid, the entire selector is invalid.");
+    </script>
+  </body>
+</html>


css/selectors/contains-invalid.html
```diff
@@ -0,0 +1,64 @@
...
+        color: red;
+      }
+      .b, :not(::before) {
+        color: red;
+      }
+      .b, :where(::before) {
+        color: red;
+      }
+
+      .c:is(*, :nonsense) {
```
frivoal: This has merely been resolved and not edited into the spec yet, but based on https://github.com/w3c/csswg-drafts/issues/3264#issuecomment-440752606,
this one should actually match.

css/selectors/contains-invalid.html
```diff
@@ -0,0 +1,64 @@
...
+        color: red;
+      }
+      .c:not(*, :nonsense) {
+        color: red;
+      }
+      .c:where(*, :nonsense) {
```
frivoal: As for `:is`, based on https://github.com/w3c/csswg-drafts/issues/3264#issuecomment-440752606, this one should match.

css/selectors/contains-invalid.html
```diff
@@ -0,0 +1,64 @@
...
+        color: red;
+      }
+      .b, :where(::before) {
+        color: red;
+      }
+
+      .c:is(*, :nonsense) {
+        color: red;
+      }
+      .c:not(*, :nonsense) {
```
frivoal: There's remaining discussion in https://github.com/w3c/csswg-drafts/issues/3264#issuecomment-440752606 about what should happen to `:not()` with invalid arguments. As currently resolved, it should match (so the test would be wrong), but maybe we'll change that, so I would wait for the dust to settle before landing this PR, as this may or may not need fixing.

css/selectors/contains-invalid.html
```diff
@@ -0,0 +1,64 @@
...
+      .a, .b, .c {
+        color: lime;
+      }
+
+      .a, :nonsense {
+        color: red;
+      }
+
+      /* Pseudo-elements cannot be represented by the matches-any pseudo-class; they are not valid within :is(). */
+      .b, :is(::before) {
```
frivoal: As the edits for https://github.com/w3c/csswg-drafts/issues/3264#issuecomment-440752606 haven't been done yet and the discussion is still ongoing, it's not entirely clear what should happen to `:is()` with a single invalid argument. I would expect that the new invalidation rules would mean that since `::before` is not allowed in `:is()` it gets discarded, but that this merely cause that argument of `:is()` (in this case its only argument) to be discarded, and the `:is()` construct as a whole remains valid, even if it matches nothing. So the whole rule is not treated as a syntax error, `.b` matches, and the element turns red. Which makes the test wrong.

css/selectors/contains-invalid.html
```diff
@@ -0,0 +1,64 @@
...
+        color: red;
+      }
+      .b, :not(::before) {
```
frivoal: Sort of the same comment as https://github.com/web-platform-tests/wpt/pull/14442/files#r240427239, except that there's still ongoing discussion in that issue about how invalidation of `:not()` arguments should work. We should wait until the dust settles on that.

frivoal (changes requested):
There's no inherent dependency on Javascript, so I'd prefer if this was a (series of?) reftests, but that's just a preference, not a show stopper.

Other than that, it seems to me this test is largely incorrect, due to https://github.com/w3c/csswg-drafts/issues/3264#issuecomment-440752606. See individual comments.

foolip:

foolip:
> There's no inherent dependency on Javascript, so I'd prefer if this was a (series of?) reftests, but that's just a preference, not a show stopper.

@frivoal, I just came here to clean up my Travis mess, but am curious about this. Is it not preferable to use CSSOM and things like `element.matches` to test selectors and style calculation, with reftests used only when the testharness.js variant would be impossible to write, or would effectively be describing "there should be a green box here"? Or are testharness.js tests harder to debug when they fail?

ewilligers:
I was planning to update this test as part of the Blink CL implementing error recovery when the spec updates. I can instead implement that now (without merging into Blink).

I prefer having a dozen asserts to having a dozen reftests or a single reftest with a dozen diffs.
My use of assertions is consistent with https://web-platform-tests.org/writing-tests/index.html
> Reftests should be used to test rendering and layout.
> testharness.js tests should be used (where possible!) for testing everything else.

foolip:
@ewilligers that makes sense to me.

frivoal:
> Is it not preferable to use CSSOM and things like element.matches to test selectors and style calculation, with reftests used only when the testharness.js variant would be impossible to write, or would effectively be describing "there should be a green box here"?

@foolip Answering this question way late, but given that there exists CSS implementations that do not support JS, it seems better when features that could either be tested in a reftest or in a testharness test are tested in a reftest, without depending on JS. For browsers it doesn't make a difference, but for other implementations it does, and the more common tests we have, the better the interop and the lower the risk of an accidental fork in the platform.

frivoal:
> Other than that, it seems to me this test is largely incorrect, due to w3c/csswg-drafts#3264 (comment). See individual comments.

@ewilligers Any plan on fixing this (or telling me why I'm wrong), or should we close this?

PR #14455, opened 2018-12-11
Encoding: make it clear sniffing for UTF-8 is not acceptable

annevk:
See https://github.com/whatwg/encoding/issues/68 for background.


inexorabletash (approved):
Only minor tidbit might be that seeing the content in the results looks ugly - maybe 'display: none' the content.

annevk:
I did consider that, but it seems that would make debugging harder than it needs to be and mostly bots will see these results anyway.

PR #14460, opened 2018-12-11
HTML: test that non-synthetic errors get muted too

annevk:
Per request from @bzbarsky in https://bugzilla.mozilla.org/show_bug.cgi?id=1512965#c5.


foolip (approved):

bzbarsky:
Actually, what I was requesting was that we have a test for the case of a syntax error in the script (e.g. if non-script data is being linked to, to make sure we don't expose the cross-site non-script data in the error handler).

annevk:
Created https://github.com/web-platform-tests/wpt/pull/14461, but I'm not entirely convinced this is only useful for syntax errors given the other information on Error objects.

PR #14484, opened 2018-12-12
Normalize version strings for Safari Tech Preview

### LINKED ISSUE #14465: Safari Tech Preview version string inconsistent with other products
Noticed in https://github.com/lukebjerring/wpt/pull/8/checks?check_run_id=39929938

> safari-Technology Preview (Release 70, 13607.1.13)-mac-10.13

Hexcles:
Fixes #14465 .

### DIFF
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -636,6 +636,9 @@ class Safari(Browser):
         # "Included with Safari 12.1 (14607.1.11)"
         # "Included with Safari Technology Preview (Release 67, 13607.1.9.0.1)"
         # The `--version` flag was added in STP 67, so allow the call to fail.
+        # For the two examples, this function will return, respectively:
+        # "12.1 (14607.1.11)"
+        # "67 (13607.1.9.0.1)"
         try:
             version_string = call(webdriver_binary, "--version").strip()
         except subprocess.CalledProcessError:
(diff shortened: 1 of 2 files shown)


lukebjerring (changes requested):
Add a test!

lukebjerring (approved):

tools/wpt/browser.py
```diff
@@ -636,6 +636,9 @@ def version(self, binary=None, webdriver_binary=None):
         # "Included with Safari 12.1 (14607.1.11)"
         # "Included with Safari Technology Preview (Release 67, 13607.1.9.0.1)"
         # The `--version` flag was added in STP 67, so allow the call to fail.
+        # For the two examples, this function will return, respectively:
+        # "12.1 (14607.1.11)"
+        # "67 (13607.1.9.0.1)"
```
foolip: Why delete the channel information? We keep it for Chrome, and it's represent as an "a" in the Firefox version.

Hexcles: Yeah we can have the channel at the end. What about "67 (13607.1.9.0.1) Preview"? (Or perhaps the full "Technology Preview"?)

foolip: What is the code that will end up parsing this string for wpt.fyi? I feel like this is to avoid changing that but don't understand why. It seems very likely there will be occasional changes to the version string formats that existing code handles incorrectly, and we'll have to fix it after the fact. Doing too much parsing here might mangle it in the interim.

Hexcles: https://github.com/web-platform-tests/wpt/issues/14465#issuecomment-446773143

Hexcles:
Abandoned in favour of fixing from wpt.fyi

Issue #14433, opened 2018-12-10
Many Resource Timing tests are flaky


foolip:
https://github.com/web-platform-tests/wpt/pull/14409#issuecomment-445753408 lists a number of tests in resource-timing/ that were flaky and not plausibly caused by the test changes.

Suggested method to investigate:
1. Add a trailing whitespace or other no-op to all tests in resource-timing/ in a test PR
1. Check which were flaky on Taskcluster
1. Fix them :)

foolip:
Notifying suggested reviewers for resource-timing: @haoxli @plehegar @zqzhang @igrigorik @toddreifsteck @yoavweiss

PR #14487, opened 2018-12-12
[resource-timing] Avoid race condition

jugglinmike:
Depending on the scheduling of resource retrieval, performance entries
for the test harness files may be observable by this test. That behavior
is beyond the focus of this test, so such entries should be ignored if
present.


yoavweiss (approved):
LGTM!

yoavweiss:
Thanks for fixing flakiness! :)

We've had [reports](https://github.com/web-platform-tests/wpt/issues/14433) that some of the other RT tests are flaky. Could it be related to the same issue?

PR #14501, opened 2018-12-13
Throw rejection reason in fullscreen test to improve error message

foolip:


upsuper:
It's not going to make any difference, is it? I don't think our implementation includes any information in the error, the spec doesn't say so, and there may be a good reason not to do that.

foolip:
Oh, I assumed there was more information and I'd just mangled it.

upsuper:
Does Blink expose more information on the error object? Is there privacy concern over exposing them?

foolip:
Thanks to @lukebjerring's magic we can see the effect that this had:
https://staging.wpt.fyi/results/fullscreen/api/element-request-fullscreen.html?diff&run_id=5042056388935680&run_id=6318713171083264

So no more information, actually less, it's harder to tell where the problem came from.

> Does Blink expose more information on the error object? Is there privacy concern over exposing them?

The fullscreen errors don't have very interesting messages. I've considered saying what part of the fullscreen element ready check failed, which would be useful, but I haven't thought of whether that could expose information about the ancestor iframes which otherwise aren't available.

foolip:
Closing this PR, it doesn't help.

Issue #14394, opened 2018-12-06
Remove testharness.js test-level asserts


foolip:
After https://github.com/web-platform-tests/wpt/issues/11120, test-level asserts is the only remaining use of is for test-level assert metadata. Example:
https://github.com/web-platform-tests/wpt/blob/8bb0b769af8d7a6d245db31c0b6029eb3b7da861/hr-time/monotonic-clock.any.js#L1-L3

https://chromium-review.googlesource.com/c/chromium/src/+/1070155 identified tests that use this and will need updating before the feature is removed:
```
/css/css-transitions/currentcolor-animation-001.html
/css/css-values/calc-unit-analysis.html
/css/css-variables/test_variable_legal_values.html
/css/cssom/computed-style-001.html
/css/cssom/computed-style-005.html
/css/cssom/inline-style-001.html
/css/cssom/medialist-interfaces-001.html
/css/cssom/medialist-interfaces-002.html
/css/cssom/medialist-interfaces-003.html
/css/cssom/medialist-interfaces-004.html
/css/cssom/style-sheet-interfaces-001.html
/css/cssom/style-sheet-interfaces-002.html
/css/cssom-view/offsetParent_element_test.html
/css/mediaqueries/test_media_queries.html
/hr-time/monotonic-clock.any.js
/navigation-timing/test_document_open.html
/navigation-timing/test_navigate_within_document.html
/navigation-timing/test_navigation_attributes_exist.html
/navigation-timing/test_navigation_redirectCount_none.html
/navigation-timing/test_navigation_type_backforward.html
/navigation-timing/test_navigation_type_enums.html
/navigation-timing/test_navigation_type_reload.html
/navigation-timing/test_no_previous_document.html
/navigation-timing/test_performance_attributes_exist.html
/navigation-timing/test_performance_attributes_exist_in_object.html
/navigation-timing/test_readwrite.html
/navigation-timing/test_timing_attributes_exist.html
/navigation-timing/test_timing_attributes_order.html
/navigation-timing/test_timing_client_redirect.html
/navigation-timing/test_timing_reload.html
/navigation-timing/test_timing_server_redirect.html
/navigation-timing/test_timing_xserver_redirect.html
/navigation-timing/test_unique_performance_objects.html
/requestidlecallback/basic.html
/resource-timing/resource_TAO_cross_origin_redirect_chain.html
/resource-timing/resource_TAO_match_origin.htm
/resource-timing/resource_TAO_match_wildcard.htm
[...]

Issue #14499, opened 2018-12-13
Migrate from travis-ci.org to travis-ci.com


foolip:
We now have the ability to migrate individual repos in the web-platform-tests org from travis-ci.org to travis-ci.com, and we have already migrated wpt.fyi as a trial. Old links and build history are preserved.

There isn't a strong reason to migrate *now*, but we eventually have to. One upside is that we get the new [Checks API](https://developer.github.com/v3/checks/) integration which is also what Azure Pipelines uses. The Travis integration isn't very fancy, but it's a small improvement. Example from wpt.fyi:
https://github.com/web-platform-tests/wpt.fyi/pull/910/checks?check_run_id=40636708

@web-platform-tests/admins I would like to do this migration org-wide. If I do not hear concerns within a week I'll go ahead.

Hexcles:
Just a reminder that you'd need to change the "required checks" for protected branches after migration.

foolip:
Yep. There are also details about cron jobs that I'll take care of.

foolip:
I will do this now, one day earlier than planned, to give a few more days to clean up the mess if I break stuff.

foolip:
The most recent PR is now https://github.com/web-platform-tests/wpt/pull/14600, any PRs created after then but before the transition is complete need to be inspected so they're not blocked on checks that will never run.

foolip:
The app is now installed org-wide instead of just for wpt.fyi. Following the documentation in https://docs.travis-ci.com/user/open-source-repository-migration/ to migrate each individual repo.

foolip:
https://github.com/web-platform-tests/wpt/pull/13352 has finished running on travis-ci.com, and I sent https://github.com/web-platform-tests/wpt/pull/14603 to test the full set of tests.

I've changed the required check from "continuous-integration/travis-ci" to "Travis CI - Pull Request", but unfortunately that makes it show up as required on all existing PRs as well. https://github.com/web-platform-tests/wpt/pull/14600 is an example of this, now in the state "Required statuses must pass before merging". I'll see if there's a way to trigger the runs without closing and reopening all PRs.

foolip:
For master, https://github.com/web-platform-tests/wpt/runs/42781561 was the first job to run after the change, followed by https://github.com/web-platform-tests/wpt/runs/42784876. The worked as intended:
https://github.com/web-platform-tests/wpt/releases/tag/merge_pr_14570
[...]

PR #14519, opened 2018-12-14
Remove some useless test-level properties

qiuzhong:
Remove useless test-level properties, including:
* Incorrect type of the properties parameter for `promise_test()`
* `flags` property passed to `test()`, which is unused in testharness.js
* Incorrect type of the properties `generate_tests`: moved to comments
* `help` property: moved the head of the test files
* Default empty properties dictionary
Related: #14394


html/infrastructure/urls/resolving-urls/query-encoding/resources/resolve-url.js
```diff
@@ -1,6 +1,34 @@
 // NOTE: this file needs to be split up rather than expanded. See ../location.sub.html for some
 // extracted tests. Tracked by https://github.com/web-platform-tests/wpt/issues/4934.
 
+/*
+* help:
```
foolip: @cvrebert @zcorpan have you been using the test-level help metadata for these tests in any way? Ideas about how it could be used? I'm not sure if it's worth preserving or not as machine-readable metadata.

html/semantics/forms/the-input-element/datetime.html
```diff
@@ -6,6 +6,13 @@
     <link rel="author" title="Fabrice Clari" href="mailto:f.clari@inno-group.com">
     <link rel="author" title="Dimitri Bocquet" href="mailto:Dimitri.Bocquet@mosquito-fp7.eu">
     <link rel="help" href="https://html.spec.whatwg.org/multipage/#the-input-element">
+    <link rel="help" href="https://html.spec.whatwg.org/multipage/#dom-input-type">
```
foolip: @Ms2ger you added/edited the spec links in this test. Do you think we should keep them?

media-source/mediasource-util.js
```diff
@@ -330,16 +330,15 @@
     };
 
     window['MediaSourceUtil'] = MediaSourceUtil;
-    window['media_test'] = function(testFunction, description, options)
+    window['media_test'] = function(testFunction, description)
```
foolip: @qiuzhong can you separate out just this change? This is one that we definitely should do, since the options aren't used anywhere.

qiuzhong: I reverted the `media_test` function.

foolip: I mean that these are changes that we should land whether or not we remove properties entirely, so submitting those as a separate PR would be great.

qiuzhong: Sorry, I misunderstood your meaning. This time I reverted the whole test file and  I'll file a new PR for it.

foolip:

html/infrastructure/urls/resolving-urls/query-encoding/resources/resolve-url.js
zcorpan: I've written tests with `help` but it's likely less useful than a js line comment with a spec link (the only use I imagine is for someone debugging a test).

dbaron:
Behind the CSS test changes here is, I think, a question for @fantasai or @plinss:  does the CSS WG intend to do something with per-test data for `testharness.js`-based tests of whether the test is testing invalid syntax?

fantasai:
@dbaron AFAIK, no. I think the main purpose of flagging 'invalid' was to let people use validator tools to help check syntax on their tests; flagging tests with  deliberately invalid syntax would let the tools skip those tests. I don't think this logic applies to testharness.js tests, and I'm not aware of any other purpose for this metadata. Maybe @plinss knows more, though.

plinss:
There were plans to use the per-test metadata in testharness.js tests a long time ago, but those have never developed and are unlikely to at this point.

zcorpan (approved):
As far as I can tell, this looks good. The PR needs to be rebased, though.

dbaron:
This can't currently be merged because the wpt-firefox-nightly-stability test is showing [unstable results](https://github.com/web-platform-tests/wpt/pull/14519/checks?check_run_id=1039630341) (timing out about 1/3 of the time) on various `/html/infrastructure/urls/resolving-urls/query-encoding-*` tests.

That seems likely to be an existing problem... I think?

zcorpan:
See
https://bugs.chromium.org/p/chromium/issues/detail?id=930297
https://bugzilla.mozilla.org/show_bug.cgi?id=1034063 
https://github.com/web-platform-tests/wpt/issues/4934

Yes, these tests are known unstable.

I can admin merge this. Thanks @dbaron !

PR #14532, opened 2018-12-14
Port filter subregion test from example in spec

svgeesus:
This is simply the [SVG example](https://drafts.fxtf.org/filter-effects/examples/filtersubregion00.svg) from the [relevant section of filters-1](https://drafts.fxtf.org/filter-effects/#FilterPrimitiveSubRegion) and the reference is the [image for that example](https://drafts.fxtf.org/filter-effects/examples/filtersubregion00.png).


svgeesus:
This is upstream reviewed because this is a spec example (and was previously an example in the SVG spec). But additional review is welcome.

foolip (approved):

foolip:
(Changed the title of PR and commit because I read "filter" as the verb, but mostly because commit and PR titles were different.)

PR #14004, opened 2018-11-09
Add unit tests for wrapper handlers in tools/serve

qiuzhong:
Add unit tests for all kinds of wrapper handlers in tools/serve/serve.py.
See #12452

PR #14537, opened 2018-12-14
Cache manifest everywhere

### LINKED ISSUE #14421: Make it possible to thread the manifest through the code when it's used multiple times
We have some cases where we end up reloading the manifest multiple times during a single wpt command because each part needs the data and there isn't any way to pass it between the different subsystems. We should arrange it so that anything that needs to load the manifest can also have an existing Manifest object passed in, and avoid all the duplicate work.

Hexcles:
Move the cache layer of manifest objects from wpt.testfiles to
manifest.manifest so that all users of the manifest module can benefit.

This prevents us from having two in-memory copies of the manifest when
running affected tests (one loaded by wpt.testfiles and the other loaded
by wptrunner.testloader). Partly fixes #14421 (the memory usage part of
it, but not the code health part).

### DIFF
--- a/tools/manifest/manifest.py
+++ b/tools/manifest/manifest.py
@@ -400,8 +400,16 @@ def load(tests_root, manifest, types=None, meta_filters=None):
     return _load(logger, tests_root, manifest, types, meta_filters)
 
 
+__load_cache = {}
+
+
 def _load(logger, tests_root, manifest, types=None, meta_filters=None):
     # "manifest" is a path or file-like object.
+    manifest_path = (manifest if isinstance(manifest, string_types)
+                     else manifest.name)
+    if manifest_path in __load_cache:
+        return __load_cache[manifest_path]
+
     if isinstance(manifest, string_types):
         if os.path.exists(manifest):
             logger.debug("Opening manifest at %s" % manifest)
@@ -418,12 +426,14 @@ def _load(logger, tests_root, manifest, types=None, meta_filters=None):
         except ValueError:
             logger.warning("%r may be corrupted", manifest)
             return None
-        return rv
+    else:
+        rv = Manifest.from_json(tests_root,
+                                json.load(manifest),
+                                types=types,
+                                meta_filters=meta_filters)
 
-    return Manifest.from_json(tests_root,
-                              json.load(manifest),
-                              types=types,
-                              meta_filters=meta_filters)
+    __load_cache[manifest_path] = rv
+    return rv
 
 
 def load_and_update(tests_root,
--- a/tools/wpt/testfiles.py
+++ b/tools/wpt/testfiles.py
@@ -8,8 +8,6 @@ import sys
 from collections import OrderedDict
 from six import iteritems
 
-from ..manifest import manifest
-
 here = os.path.dirname(__file__)
 wpt_root = os.path.abspath(os.path.join(here, os.pardir, os.pardir))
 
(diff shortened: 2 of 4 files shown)


tools/wpt/testfiles.py
```diff
@@ -183,24 +181,12 @@ def _in_repo_root(full_path):
...
-        wpt_manifest = manifest.load_and_update(wpt_root, manifest_path, "/",
-                                                update=manifest_update)
-        c[manifest_path] = wpt_manifest
-        return c[manifest_path]
-    return load
-
-
-load_manifest = _init_manifest_cache()
+def load_manifest(manifest_path=None, manifest_update=True):
+    from manifest import manifest
```
Hexcles: Switch to a delayed import here because otherwise Python will think there are two different `manifest` modules (because of different import paths) and we don't get the singleton.

Hexcles:
On my machine, this reduces 2GB of memory usage (peak memory 5GB -> 3GB) when running

```bash
./wpt run --affected HEAD^1 --log-mach - --binary ~/tools/firefox-nightly/firefox firefox
```
(with a dummy local commit)

tools/wpt/testfiles.py
jugglinmike: What do you say we document your rationale with a comment? I'm afraid of your
optimization being accidentally subverted by some future refactoring (the
corresponding performance degradation might not be noticed immediately).

This is beyond my experience with Python, though. Can you advise on any
longer-term goals that would obviate this kind of consideration?

tools/manifest/manifest.py
```diff
@@ -400,8 +400,18 @@ def load(tests_root, manifest, types=None, meta_filters=None):
     return _load(logger, tests_root, manifest, types, meta_filters)
 
 
+__load_cache = {}
+
+
 def _load(logger, tests_root, manifest, types=None, meta_filters=None):
     # "manifest" is a path or file-like object.
+    manifest_path = manifest if isinstance(manifest, string_types) \
```
jugglinmike: I share @jgraham's [mistrust of line continuations](https://github.com/web-platform-tests/wpt/pull/14004#discussion_r232222841), but [it doesn't look like we can disallow them with linting](https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes). Without that, I don't know if we can fairly reject the use of that syntax...

tools/manifest/manifest.py
```diff
@@ -400,8 +400,18 @@ def load(tests_root, manifest, types=None, meta_filters=None):
...
+        else manifest.name
+    if manifest_path in __load_cache:
+        return __load_cache[manifest_path]
+    # Cache at most one manifest.
+    __load_cache.clear()
```
jgraham: Gecko uses two and servo uses three. I think we should allow one per url-base rather than one total.

tools/manifest/manifest.py
jgraham: Yes we can :p Please don't use continuation lines, use additional parens instead.

jgraham (changes requested):

Hexcles:
I addressed the review comments, but am having trouble debugging the `tools/wpt` failure on macOS -- I can't reproduce the error locally on my Macbook.

I'll try to dig deeper.

tools/wpt/testfiles.py
Hexcles: I added a comment here, but I don't really have great ideas how to make sure we always get a singleton.

Hexcles:
So my working theory is that `test_wpt.py` had a module-level fixture to create a temporary manifest, which means the manifest path was the same for all test cases, but some test cases need to modify the manifest, so caching by path failed. The workaround is to use a different temporary manifest for each test case (and the temporary manifest is copied from a "persistent" one, which is initialized in a module-level setup function to avoid having to generate the manifest every time). See my second commit. This seems to work. However, I don't really understand why the problem only demonstrated on macOS.

@jgraham PTAL again.

foolip:
FYI, I've rebased this branch to trigger "Travis CI - Pull Request", manually setting things right after the travis-ci.com transition in https://github.com/web-platform-tests/wpt/issues/14499 changed the name of the required check, which applies retroactively.

tools/wpt/testfiles.py
jgraham: Did you verify that the previous code was broken? I know that `foo.bar` and `bar` are different even if they're the same `bar.py` file, but I'm slightly surprised if that also affects relative imports.

jgraham (approved):

PR #14545, opened 2018-12-16
Update <button type="menu"> test to reflect its removal

Zirro:
`<button type="menu">` got removed in https://github.com/whatwg/html/pull/2342 and this test was [failing](https://wpt.fyi/results/html/semantics/forms/the-button-element/button-validation.html?label=stable&aligned&q=the-button-element) in all browsers.


domenic (approved):
Nice catch

PR #14551, opened 2018-12-17
[css-text] Correct test metadata

frivoal:
The section title is out of date and unneeded. Deleting it.

The link is broken. Fixing it.


frivoal:
I don't understand why taskcluster is repporting an error here. It claims instability about css/css-text/overflow-wrap/overflow-wrap-002.html, but I cannot reproduce locally, and it seems unlikely to be true on this kind of reftest. I suspect a failure in the tool is more likely

Not sure how to continue from here. cc @gsnedders @foolip

PR #14558, opened 2018-12-17
SVG Geometry: computed width and height

ewilligers:
The properties width and height accept percentages and units.

https://drafts.csswg.org/css-sizing-3/#preferred-size-properties
https://svgwg.org/svg2-draft/geometry.html#Sizing
https://drafts.csswg.org/cssom/#resolved-values
> If the property applies to the element or pseudo-element and the resolved value of the display property is not none or contents, then the resolved value is the used value. Otherwise the resolved value is the computed value.

Each test passes with Blink.

Firefox and Safari give percentages and calc expressions.

With Edge, all results are 0px.

### DIFF
--- a/svg/geometry/parsing/sizing-properties-computed.svg
+++ b/svg/geometry/parsing/sizing-properties-computed.svg
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg"
+     xmlns:h="http://www.w3.org/1999/xhtml"
+     width="800px" height="800px">
+  <title>SVG Geometry Properties: getComputedValue().width</title>
+  <metadata>
+    <h:link rel="help" href="https://svgwg.org/svg2-draft/geometry.html#Sizing"/>
+    <h:link rel="help" href="https://drafts.csswg.org/css-sizing-3/#preferred-size-properties"/>
+    <h:link rel="help" href="https://drafts.csswg.org/cssom/#resolved-values"/>
+  </metadata>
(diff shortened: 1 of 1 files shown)


AmeliaBR:
The test is incorrect. As discussed in https://github.com/w3c/svgwg/issues/349, `getComputedStyle` should always return the *used value* (i.e., px length)  for `width` and `height` on an element where the geometry properties apply.  Percentages and calc expressions should only be preserved if the property _doesn't_ apply, e.g. on a `<circle>`.

ewilligers:
@emilio @fantasai @tabatkins @upsuper - this came up in an [SVG call](https://www.w3.org/2018/12/10-svg-minutes.html#item05). Tests show that for SVG elements (including those with width/height), browsers currently return a `<length-percentage>` from getComputedStyle().width/height instead of always returning a `<length>` as described in CSSOM

https://drafts.csswg.org/cssom-1/#resolved-value
> If the property applies to the element or pseudo-element and the resolved value of the display property is not none or contents, then the resolved value is the used value.

https://drafts.csswg.org/css-cascade-4/#used-value
> The used value is the result of taking the computed value and completing any remaining calculations to make it the absolute theoretical value used in the layout of the document.

tabatkins:
...huh. Seems we should treat that as a bug against all browsers? It would be *really* nasty to have properties return *either* computed or used values, depending on what type of element you call gCS() on.

ewilligers:
Lots of thanks. Test updated to match spec and latest Blink.

dirkschulze (approved):

PR #14573, opened 2018-12-17
wptserve: Add a header_or_default function which works when the request header is absent.

### LINKED ISSUE #14572: wptserve sub filter: cannot substitute headers which were not sent
e.g. writing this: `{{headers[Referer]}}`

works if there is a Referer header, but raises an exception if the header was not sent. In my use case I'd like to still be able to proceed in this case.

The simplest solution would seem to be to modify the way the templates treat indexing with square brackets to catch KeyError and produce a blank value such as `''` instead (and I have a commit that would make that change).

It would be possible to do something more sophisticated, like add some other syntax for reading a header or a default value, perhaps something like `{{header_or_default(Referer, none)}}`, but this seems like it might be overkill.

The major downside would be that typos in header names in tests might be more difficult to catch, since they would yield a blank result instead of a noisy exception.

WDYT?

jeremyroman:
wptserve: Produce an empty string if indexed access fails during sub template substitution.

Fixes #14572.

### DIFF
--- a/tools/wptserve/tests/functional/test_pipes.py
+++ b/tools/wptserve/tests/functional/test_pipes.py
@@ -126,6 +126,11 @@ server: http://localhost:{0}""".format(self.server.port).encode("ascii")
 """ % {"root": root, "sep": os.path.sep}
         self.assertEqual(resp.read(), expected.encode("utf8"))
 
+    def test_sub_header_or_default(self):
+        resp = self.request("/sub_header_or_default.sub.txt", headers={"X-Present": "OK"})
+        expected = b"OK\nabsent-default"
+        self.assertEqual(resp.read().rstrip(), expected)
+
 class TestTrickle(TestUsingServer):
     def test_trickle(self):
         #Actually testing that the response trickles in is not that easy
--- a/tools/wptserve/docs/pipes.rst
+++ b/tools/wptserve/docs/pipes.rst
@@ -52,6 +52,10 @@ and `}}`. Inside the block the following variables are available:
     The HTTP headers in the request
       e.g. `{{headers[X-Test]}}` for a hypothetical `X-Test` header.
 
+  `{{header_or_default(header, default)}}`
+    The value of an HTTP header, or a default value if it is absent.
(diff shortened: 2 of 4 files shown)


jeremyroman:
PTAL. A more complicated solution would be possible; starting with the simplest workable thing seemed reasonable.

gsnedders (approved):
I guess this makes sense; in one way it seems like it might be more useful to have some more powerful templating with conditionals and the like, but that quickly becomes something far more complex.

@jgraham?

gsnedders (dismissed):
Actually, can you add a test for this in `tools/wptserve/tests/functional/test_pipes.py`?

jgraham:
Hmm, yeah I don't love the idea of silent failure here; we already see a lot of bugs in tests that go unnoriced even with logging; this seems like it would make that much worse. We support multi-argument functions in the file path already so something like `header_default(name, default)` seems like it should be possible to implement without building new infrastructure and much more reliable.

jeremyroman:
Updated to create a `header_or_default` function, since there was some concern about tweaking the existing behavior. Test and docs included.

tools/wptserve/tests/functional/test_pipes.py
```diff
@@ -126,6 +126,11 @@ def test_sub_fs_path(self):
 """ % {"root": root, "sep": os.path.sep}
         self.assertEqual(resp.read(), expected.encode("utf8"))
 
+    def test_sub_headers(self):
```
gsnedders: ./wptserve/tests/functional/test_pipes.py:129:5: F811 redefinition of unused 'test_sub_headers' from line 81

Probably call this `test_sub_header_or_default`?

jeremyroman:
Okay, fixed that, though taskcluster is apparently failing on chrome and firefox without (these?) changes, and that is blocking this PR?

gsnedders:
Looks like some weird network issue? It's failing to connect over the network to fetch the repo, seemingly.

jgraham (approved):

PR #14598, opened 2018-12-19
Add a way for single layout test to report all assertions and continue execution after the first failing assertion

youennf:
This PR records all executed assertions for single test files.
These assertions are passed to the completion callback so that results can be displayed in various ways.

This PR adds a method to enable continue executing the test even if an assertion is failing.
A test opts in by calling a specific method.

A test using this new mode is added to validate the mechanism.
This test is a testharness-based version of https://github.com/WebKit/webkit/blob/master/LayoutTests/fast/frames/detached-frame-property.html

### DIFF
--- a/html/browsers/the-window-object/detached-frame-property.html
+++ b/html/browsers/the-window-object/detached-frame-property.html
@@ -1,0 +1,38 @@
+<title>Checks the value of detached subframe properties</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+execute_single_page_test_after_failing_assertion();
+
+onload = function()
+{
+    detachedWindow = frames[0];
+    document.body.removeChild(document.getElementsByTagName("iframe")[0]);
+
+    assert_true(!!detachedWindow.postMessage, "detachedWindow.postMessage is defined");
+    assert_true(!!detachedWindow.close, "detachedWindow.close is defined");
+    assert_true(!!detachedWindow.locationbar, "detachedWindow.locationbar is defined");
+    assert_true(!!detachedWindow.history, "detachedWindow.history is defined");
+    assert_true(!!detachedWindow.screen, "detachedWindow.screen is defined");
+    assert_true(!!detachedWindow.location, "detachedWindon.location is defined");
+    assert_true(detachedWindow.closed, "detachedWindow.closed is true");
+    assert_equals(detachedWindow.top, null, "detachedWindow.top is null");
+    assert_equals(detachedWindow.opener, null, "detachedWindow.opener is null");
+    assert_equals(detachedWindow.parent, null, "detachedWindow.parent is null");
+    assert_equals(detachedWindow.frameElement, null, "detachedWindow.frameElement is null");
+    assert_equals(detachedWindow.window, null, "detachedWindow.window is null");
+    assert_equals(detachedWindow.frames, null, "detachedWindow.frames is null");
+    assert_equals(detachedWindow.self, null, "detachedWindow.self is null");
+
+    assert_true(!detachedWindow.localStorage, "detachedWindow.localStorage is undefined");
+    assert_true(!!detachedWindow.document, "detachedWindow.document is defined");
+    assert_true(!!detachedWindow.XMLHttpRequest, "detachedWindow.XMLHttpRequest is defined");
+    assert_true(!!detachedWindow.getComputedStyle, "detachedWindow.getComputedStyle is defined");
+
+    assert_equals(detachedWindow.innerHeight, 0, "detachedWindow.innerHeight is 0");
+    assert_equals(detachedWindow.innerWidth, 0, "detachedWindow.innerWidth is 0");
+
+    done();
+}
+</script>
+<iframe src="about:blank"></iframe>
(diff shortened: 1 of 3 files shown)


foolip:

resources/testharness.js
```diff
@@ -2979,11 +2985,30 @@ policies and contribution forms [3].
...
         if (tests.tests.length === 0) {
             tests.set_file_is_test();
         }
+        if (tests.file_is_test) {
+            let test = tests.tests[0];
+            let message = expected_true ? description : make_message(function_name, description, error, substitutions);
+            test.add_assertion({passed: !!expected_true, message: message});
+            if (continue_execution_after_failing_assertion) {
+                if (!expected_true && test.status !== test.FAIL) {
+                    test.set_status(test.FAIL, "At least one assertion failed: " + message);
```
foolip: Looks like this will list the first failure. Is the idea that the treatment of `assertions` is entirely vendor-specific, or how will one make use of the list of assertions?

For wpt.fyi, how should this feature change what is shown? Since this is limited to single-page tests we *could* try to treat the assertions as tests in themselves, but then we'll also have to require that each assertion's message is unique, or possibly make it unique automatically if it isn't. The only other option I see is to require the assertion order to be stable across runs or between browsers, as otherwise there's no way to compare results.

@lukebjerring for more ideas.

resources/testharnessreport.js
```diff
@@ -28,6 +28,15 @@ function dump_test_results(tests, status) {
                 stack: status.stack};
     results_element.textContent = JSON.stringify(data);
 
+    if (tests.length === 1) {
```
foolip: Is this entirely for debugging and intended to be reverted, or is it an important part of this patch?

resources/testharness.js
lukebjerring: My 5c: subtests are different that assertions, which are more like a precondition (such as asserting that something is non-null, before asserting the correct value of a property on it).
If you want to have separate failures for multiple things, that don't cascade, then we should be creating subtests. 
Something like https://github.com/lukebjerring/wpt/pull/14 is what I mean in this specific case.

html/browsers/the-window-object/detached-frame-property.html
```diff
@@ -0,0 +1,38 @@
+<title>Checks the value of detached subframe properties</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+execute_single_page_test_after_failing_assertion();
```
jgraham: If there has to be a big global switch (which I don't love) it should be a property passed to `setup()`

resources/testharness.js
jgraham: I agree with @lukebjerring here; this model looks like it should work by wrapping each `assert()` in an implicit `test()`;

jgraham (changes requested):

resources/testharness.js
youennf: > Looks like this will list the first failure. Is the idea that the treatment of `assertions` is entirely vendor-specific, or how will one make use of the list of assertions?

Yes, this would be vendor specific.
WebKittens want more reporting in what is produced by WebKitTestRunner.
Ideally, for every failed test (single file test or not), it would be good to have the list of assertions that passed for the given test.
testharnessreport.js could be used to display that information in whatever form.
For instance, it could be dumped in stderr for debugging purposes by WebKitTestRunner and on the inspector console on a browser.

The current PR provides the first error message in the test failing status.
To decrease flakiness, it might just be better to not report it and let testharnessreport.js actually decide how it wants to output the information.

Implicit test() is another valid approach.
My main concern is how much flakiness this actually ends up creating in the output.
Test writers will need to ensure that assertions are called the exact same amount of time in the exact same order. I am not sure though how practical this is to require in WPT.
WebKit js-test actually requires that so I am fine trying it here as well.

resources/testharnessreport.js
youennf: I did it to ensure this was working.
I do not strongly care about this particular change but think that this might be convenient for people running tests in a browser.
If we go with implicit test(), this change is meaningless.
Otherwise, providing additional information to people running tests might be useful.

html/browsers/the-window-object/detached-frame-property.html
youennf: ok

resources/testharness.js
jgraham: The order shouldn't be important; it's not part of the API. You may need to find a way to create a unique identity for the alert to ensure the order doesn't leak.

foolip:
Please take a look at https://github.com/web-platform-tests/wpt/pull/14616#pullrequestreview-187171969, apparently some tests have already added their own helpers for something very similar. There the wrappers are being removed though, because they were implicated in flakiness.

youennf:
Maybe I should split the PR in two parts, one for being able to store assertions and one to have a mode to continue running single file test after an assertion failure.

If we go with test-wrapping asserts, I think it should be done for all single file tests, not just the ones that opt-in in the new mode.

Similarly, we should decide whether the order of running assertions is important or not, independently of the actual single test file mode. If we test-wrap asserts, the execution order becomes meaningful.
For instance, the number of wrapped asserts might change run after run.

Adding more determinism to tests is usually a good thing.
I can have a look on how existing single test files would be ok with this new constraint.
I would prefer reaching consensus on this approach before spending some time there.

PR #14604, opened 2018-12-19
Fix #13073: make CRTICIAL errors cause non-zero exit

### LINKED ISSUE #13073: Critical TestManager errors should cause non-zero exit
e.g. https://github.com/web-platform-tests/wpt/issues/12874 which exceeds max restarts should be a critical error which should then cause a non-zero exit (and a job failure on TC etc.), given we haven't actually run all tests requested.

gsnedders:
Fixes #13073.

### DIFF
--- a/tools/wpt/run.py
+++ b/tools/wpt/run.py
@@ -581,17 +581,6 @@ def setup_logging(kwargs, default_config=None):
 
     global logger
 
-    class LoggedAboveLevelHandler(object):
-        def __init__(self, min_level):
-            self.min_level = mozlog.structuredlog.log_levels[min_level.upper()]
-            self.has_log = False
-
-        def __call__(self, data):
-            if (data["action"] == "log" and
-                not self.has_log and
-                mozlog.structuredlog.log_levels[data["level"]] <= self.min_level):
-                self.has_log = True
-
     # Use the grouped formatter by default where mozlog 3.9+ is installed
     if default_config is None:
         if hasattr(mozlog.formatters, "GroupingFormatter"):
@@ -601,9 +590,7 @@ def setup_logging(kwargs, default_config=None):
         default_config = {default_formatter: sys.stdout}
     wptrunner.setup_logging(kwargs, default_config)
     logger = wptrunner.logger
-    has_log = LoggedAboveLevelHandler("CRITICAL")
-    logger.add_handler(mozlog.handlers.LogLevelFilter(has_log, "CRITICAL"))
-    return logger, has_log
+    return logger
 
 
 def setup_wptrunner(venv, prompt=True, install_browser=False, **kwargs):
@@ -684,7 +671,7 @@ def setup_wptrunner(venv, prompt=True, install_browser=False, **kwargs):
 
 
 def run(venv, **kwargs):
-    logger, has_log = setup_logging(kwargs)
+    setup_logging(kwargs)
 
     # Remove arguments that aren't passed to wptrunner
     prompt = kwargs.pop("prompt", True)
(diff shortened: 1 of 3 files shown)


tools/wpt/run.py
```diff
@@ -465,6 +465,18 @@ def setup_logging(kwargs, default_config=None):
 
     global logger
 
+    class HasLogHandler(mozlog.handlers.BaseHandler):
+        def __init__(self):
+            self.inner = None
+            mozlog.handlers.BaseHandler.__init__(self, self.inner)
+            self.has_log = False
+
+            # XXX: I have no idea what this second arg does but it's required
```
gsnedders: I filed [bugzilla #1515327](https://bugzilla.mozilla.org/show_bug.cgi?id=1515327) for this, given even reading the source I'm not clear what's the right thing to do here.

tools/wpt/run.py
```diff
@@ -465,6 +465,18 @@ def setup_logging(kwargs, default_config=None):
 
     global logger
 
+    class HasLogHandler(mozlog.handlers.BaseHandler):
+        def __init__(self):
+            self.inner = None
```
gsnedders: This feels weird, given we don't really want an inner logger at all, but it's not at all clear what the intended way to write a new handler is ([bugzilla 1515330](https://bugzilla.mozilla.org/show_bug.cgi?id=1515330)).

tools/wpt/run.py
```diff
@@ -559,7 +574,7 @@ def run(venv, **kwargs):
                              install_browser=install_browser,
                              **kwargs)
 
-    rv = run_single(venv, **kwargs) > 0
+    rv = run_single(venv, **kwargs) > 0 or has_log.has_log
```
gsnedders: Perhaps stupidly, this does actually run everything (esp. if the CRITICAL error is in setup) and then exit non-zero. But I don't really see how to make a logger call cause a quick exit? Maybe we should just go through every CRITICAL case in the codebase and make it throw/return after?

jgraham (changes requested):

tools/wpt/run.py
```diff
@@ -474,6 +486,9 @@ def setup_logging(kwargs, default_config=None):
         default_config = {default_formatter: sys.stdout}
     wptrunner.setup_logging(kwargs, default_config)
     logger = wptrunner.logger
+    has_log = HasLogHandler()
+    logger.add_handler(mozlog.handlers.LogLevelFilter(has_log, "CRITICAL"))
+    return has_log
```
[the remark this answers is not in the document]
gsnedders: @jgraham in what other way would you like to get the reference to the handler?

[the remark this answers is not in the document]
jgraham: I think all this code should be moved into wptrunner itself, since run.py is a frontend and isn't used in all browser integrations that are using wptrunner (e.g. gecko only uses it for non-Firefox runs).

tools/wptrunner/wptrunner/wptrunner.py
```diff
@@ -332,16 +332,25 @@ def check_stability(**kwargs):
...
-        list_disabled(**kwargs)
-    elif kwargs["list_tests"]:
-        list_tests(**kwargs)
-    elif kwargs["verify"] or kwargs["stability"]:
-        return check_stability(**kwargs)
-    else:
-        return not run_tests(**kwargs)
+    assert logger is not None
+
+    logged_above = wptlogging.LoggedAboveLevelHandler("CRITICAL")
```
jgraham: Can we call this `logged_critical` or so?

jgraham (changes requested):

jgraham (approved):

foolip:
Yay! @gsnedders, can you check how things behave if we remove the `--exclude` args from Safari, do they fail as they should?

gsnedders:
@foolip [Safari](https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=30377) and [STP](https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=30378) Azure builds underway!

PR #14608, opened 2018-12-19
Remove trailing spaces from MANIFEST.json

Hexcles:
https://github.com/web-platform-tests/wpt/commit/1ca46e0bec3da78ea11a824f7e6d07755e1f8399#diff-2a239eec150d25cf7486394513fb1e63L255
removed the `separators=(',', ': ')` argument to `json.dump`, which
caused `json.dump` to use the default ", " separator that will add a
trailing space after each item.

Doc: https://docs.python.org/2/library/json.html#json.dump

### DIFF
--- a/tools/manifest/manifest.py
+++ b/tools/manifest/manifest.py
@@ -11,8 +11,10 @@ from .utils import from_os_path, to_os_path
 
 try:
     import ujson as json
+    JSON_LIBRARY = 'ujson'
 except ImportError:
     import json
+    JSON_LIBRARY = 'json'
 
 CURRENT_VERSION = 5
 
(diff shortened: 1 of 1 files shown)


gsnedders (approved):

gsnedders (dismissed):
```
17:27 <@jgraham> Hexcles: Hmm, I think that change will break if you happen to have ujson 
                 installed
17:28 <@jgraham> Which gecko does for doing metadata updates because it's a noticable perf win
17:29 < gsnedders> jgraham: why would that break?
17:29 < gsnedders> jgraham: oh does ujson not support it?
17:29 < gsnedders> jgraham: why do we not test this code with ujson installed if Gecko relies on 
                   it?
17:29 <@jgraham> gsnedders: Yeah :/
```

jugglinmike:
In what context is the presence of a trailing space relevant?

jgraham:
It's not, it's just that the Chromium code review system doesn't like the diff this change creates.

jgraham (approved):
This is OK for now since you're blocked, but I wonder if you should just add a step in your import script that allows you to control the precise serialization of the committed manifest so that we can treat it as undefined in the rest of the project?

PR #14612, opened 2018-12-20
Rename pointerevent_pointerout_pen-manual.html

LanWei22:
Since we are adding actions into pointerevent_pointerout_pen-manual.html file to simulate pen inputs, we rename it to pointerevent_pointerout_pen.html.


NavidZ:
If you are removing -manual here can you also add the test driver stuff in this very change? Because otherwise the test is going to time out until you land your chromium change and then it syncs to wpt here.

zcorpan:
It appears this file has been renamed (maybe through a chromium export?). Closing. Please reopen if I missed something!

PR #14614, opened 2018-12-20
Fix shadow-dom/scroll-to-the-fragment-in-shadow-tree.html to work in iOS

rniwa:
Because all scrolling happens asynchronously in iOS, we can't rely on 0 second timer to detect whether the fragment navigation had happened or not.

Instead, test whether the fragment navigation had moved the focus or not, which always happens asynchronously even in iOS.

### DIFF
--- a/shadow-dom/scroll-to-the-fragment-in-shadow-tree.html
+++ b/shadow-dom/scroll-to-the-fragment-in-shadow-tree.html
@@ -65,7 +65,7 @@ function tallElementMarkup()
 
 function targetMarkup(elementType)
 {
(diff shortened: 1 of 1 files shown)


rniwa:
Ping reviewers.

hayatoito:
Sorry for the delay. LGTM.

kojiishi (approved):
lgtm

hayatoito (approved):

rniwa:
Thanks. Re-triggering the Travis jobs...

rniwa:
Hm... somehow the test is failing on wpt-chrome-dev-stability...

zcorpan:
@rniwa can you rebase and see if the checks pass on a new run?

PR #14616, opened 2018-12-20
[Resource-Timing] Reduce test flakiness

yoavweiss:
Closes https://github.com/w3c/resource-timing/issues/182 and https://github.com/web-platform-tests/wpt/issues/14433

Reducing RT test flakiness by:
* Adding an extra task before actual testing in buffer-full-add-after-full-event.html, as the timing of when entries are added to the buffer is not really well-defined at the moment.
* Reduce the use of `test_equals` and replace it with `assert_equals`, as test_equals seems to flakily fail.
* Also renamed a bunch of files to have nicer names while I was there.


yoavweiss:
Filed related Firefox issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1515696

resource-timing/clear_resource_timing_functionality.html
```diff
@@ -14,9 +14,9 @@
     const context = new PerformanceContext(performance);
     function onload_test()
     {
-        test_equals(context.getEntriesByType('resource').length, 4, 4 + ' resource timing entries should be stored in this page.');
+        assert_equals(context.getEntriesByType('resource').length, 4, 4 + ' resource timing entries should be stored in this page.');
```
foolip: I'd expect this test no longer has any subtests. Doesn't this result in a single test with a less useful name? Can you get review from someone who was involved with adding all of this test_* stuff?

resource-timing/resources/webperftestharness.js
```diff
@@ -107,7 +107,7 @@ function test_timing_greater_than(attribute_name, greater_than, properties)
 function test_timing_equals(attribute_name, equals, msg, properties)
 {
     var test_msg = msg || "window.performance.timing." + attribute_name + " == " + equals;
-    test_equals(performanceNamespace.timing[attribute_name], equals, test_msg, properties);
+    assert_equals(performanceNamespace.timing[attribute_name], equals, test_msg, properties);
```
foolip: This assert might not be inside any test, doesn't that break any test when the assert fails?

yoavweiss:
Abandoning this as https://github.com/web-platform-tests/wpt/pull/14643 seems like a better way to go about this

PR #14619, opened 2018-12-20
Make `--verbose` as an alias of `--log-mach-level debug`

### LINKED ISSUE #7846: Add ./wpt run --verbose as an alias of --log-mach-level debug
It's a pain that we use something non-standard here and not that memorable.

qiuzhong:
Add a new option `--verbose` to `./wpt run` command line as an alias
of `--log-mach-level debug`.
Fix: #7846

### DIFF
--- a/tools/wpt/run.py
+++ b/tools/wpt/run.py
@@ -473,7 +473,7 @@ def setup_logging(kwargs, default_config=None):
         else:
             default_formatter = "mach"
         default_config = {default_formatter: sys.stdout}
-    wptcommandline.check_log_verbose(kwargs)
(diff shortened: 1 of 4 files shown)


jgraham (changes requested):
This only works if we are using the mach logger. In case we aren't it will cause an error, so something else is needed.

gsnedders:
See also the comments earlier in  #7846

qiuzhong:
@jgraham , @gsnedders , please take a look at the new commits for issue #7846. `--verbose` option should be able to support all types of loggers now, not just for the mach logger.

PR #14624, opened 2018-12-20
Don't create a logging output element until we have DOMContentLoaded

jgraham:
Otherwise running with logs enabled can implicitly create a
document.body that can change the results of tests.

### DIFF
--- a/resources/test/tests/functional/log-insertion.html
+++ b/resources/test/tests/functional/log-insertion.html
@@ -6,6 +6,12 @@
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <script>
+test(function(t) {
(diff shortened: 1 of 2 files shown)


gsnedders (approved):

resources/test/tests/functional/log-insertion.html
```diff
@@ -6,6 +6,9 @@
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <script>
+test(function(t) {
+  assert_equals(document.body, null);
+}, "Log insertion before load");
```
gsnedders: I wonder if it's worthwhile having a second test to make sure completion of the first test doesn't trigger a body to be inserted?

gsnedders (approved):

Issue #14471, opened 2018-12-11
wpt run produces an empty file for a report when no tests are executed


lukebjerring:
Relates to https://github.com/web-platform-tests/wpt/pull/14156

#14354 collapsed `tests-affected` into a flag for `wpt run`, which in the case of no affected tests is causing Taskcluster to produce an empty report file. In turn, we create a check for wpt.fyi that never resolves (since a run never gets created).

e.g. https://github.com/web-platform-tests/wpt/pull/14454

foolip:
Whatever the solution is, the should be possible to tell the difference between no tests being affected, and no results being produced due to an infra error.

lukebjerring:
@Hexcles and I were discussing this. It seems to me that it would be highly preferable to still produce a valid JSON blob in the report file, for which the results field is empty, rather than a fully empty file.

@jgraham - I'm under the impression this isn't as trivial as described. One suggestion I made was that we execute a canonical "strawman" placeholder test that produces a "no tests found" message, in the case that none of the requested tests are in the manifest, which would cause the remainder of the mozlog behaivours to produce content. On a scale of `1` to `Very bad idea`, where does this sit?

lukebjerring:
On further inspection, we don't need to have any tests actually run; we just need to coax the command line into thinking it should still try when `--affected` is passed.

Hexcles:
The approach above (as implemented in either #14630 or #14649) solves the common case of this issue: `wpt run --affected` with no tests affected. We still don't have easy ways to solve an edge (but not extremely rare) case (I think originally reported by @jugglinmike ): if something goes wrong before the suite starts (like uncaught exceptions or errors during setup), we would still produce an empty (and hence invalid) JSON file. This edge case is hard to fix and presumably requires some fiddling with `mozlog`.

PR #14630, opened 2018-12-20
Plumb empty test_list when using --affected

lukebjerring:
This changes the filtering setup for `TestLoader` to treat an empty `include` list differently than a null one. Also changes the plumbing for wptcommandline to forward an empty `test_list` when `--affected` was specified, and to not pop `--affected` from the args to be able to make this check later.
The error-code of the run is also affected, with a value of `0` for no executed tests.
Executing without specifying a whitelist of test paths has the same behaviour as before (runs the full test suite as per the manifest.)
The effect on the output is that we still produce a report, which has `results: []`, rather than no report file, for the case that `--affected` produces no test results.

Fixes https://github.com/web-platform-tests/wpt/issues/14471

### DIFF
--- a/tools/wpt/run.py
+++ b/tools/wpt/run.py
@@ -494,12 +494,11 @@ def setup_wptrunner(venv, prompt=True, install_browser=False, **kwargs):
     setup_cls = product_setup[kwargs["product"]](venv, prompt, sub_product)
     setup_cls.install_requirements()
 
-    affected_revish = kwargs.pop("affected", None)
-    if affected_revish is not None:
+    if "affected" in kwargs and kwargs["affected"] is not None:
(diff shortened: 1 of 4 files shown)


Hexcles:
Sorry I didn't see this one before sending out my #14649.

This fix is a bit problematic because of a few reasons:
1. We are passing an extra arg via `kwargs` to `wptrunner`, which isn't defined in `wptrunner`'s argparser. It'd be very difficult later on to figure out where this field comes from and what it means. Adding custom fields to the dict populated by `argparser` is surprising.
2. I don't like relying on `[]` vs `None`. Yes, they are different and can be used this way, but I usually find it hard to maintain and error-prone in the long run.

Basically, I'd like to make things more explicit as in #14649 .

PR #14635, opened 2018-12-20
Fix WebRTC test use of Resolver after PR 14417

youennf:


alvestrand:
Exactly the same fixes should have been uplifted from Chrome, so @youennf needs to check whether he needs to rebase. @henbos to help.

henbos:
This is what I fixed: https://chromium-review.googlesource.com/c/chromium/src/+/1396140

youennf:
One change was not made in the Chromium patch.
Rebased patch accordingly, @henbos, could you take a look?

henbos (approved):
LGTM

PR #14639, opened 2018-12-21
Remove options parameter from mediasource.

qiuzhong:
Part of removing the support of test-level properties from testharness.js

### DIFF
--- a/media-source/mediasource-util.js
+++ b/media-source/mediasource-util.js
@@ -330,16 +330,15 @@
     };
 
     window['MediaSourceUtil'] = MediaSourceUtil;
-    window['media_test'] = function(testFunction, description, options)
+    window['media_test'] = function(testFunction, description)
     {
-        options = options || {};
         return async_test(function(test)
         {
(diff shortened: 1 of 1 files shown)


qiuzhong:
Please don't merge this before #14519 is landed.

qiuzhong:
> Is there an alternative to the old " { timeout: timeout_in_milliseconds }" option style that I may use downstream, or is such alternative also no longer necessary for some other reason?

Please refer to #11120 , test-level `timeout` property was removed from tests and testharness.js will support it any more. So you'd better not use them it in the tests. Generally, testharness.js will set a default [timeout value](https://github.com/web-platform-tests/wpt/blob/master/resources/testharness.js#L23) for each test: 10 seconds. If the tests are slow, you can add `<meta name=timeout content=long>` in the head of the test html files, that will set the timeout to be 60 seconds.

zcorpan:
#14519 has not landed yet, adding label to indicate the dependency.

PR #14641, opened 2018-12-21
Log where we are downloading Firefox from

jgraham:
To make logging in the browser class work better we explicitly pass in a logger,
rather than depending on a global variable.


foolip (approved):
LGTM % test failures

PR #14642, opened 2018-12-21
Add firefox buildid and source changeset to wptreport.json

jgraham:
This adds two keys to the run_info dict for Firefox; browser_build_id containing a build id
and browser_changeset containing the SHA1 of the source commit from which the browser was built.


foolip (approved):
Yay, thanks!

PR #14646, opened 2018-12-21
Use get-host-info to run fetch/abort test in CI

youennf:
Use method.py to get a response without any body

### DIFF
--- a/fetch/api/abort/general.any.js
+++ b/fetch/api/abort/general.any.js
@@ -1,5 +1,6 @@
 // META: global=window,worker
 // META: script=/common/utils.js
+// META: script=/common/get-host-info.sub.js
 // META: script=../request/request-error.js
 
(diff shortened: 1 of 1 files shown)


jdm (approved):

annevk:
How does https://github.com/youennf/web-platform-tests/blob/master/fetch/api/resources/empty.txt contain a body?

jdm:
@annevk I'm not sure I understand the question or why it's relevant.

jdm:
Oh, the implication being that empty.txt used to somehow provide a body and was changed to method.py for this reason.

annevk:
Yeah, I don't understand that change. And if it's needed presumably other files referencing empty.txt need to be modified too (or empty.txt can be removed).

youennf:
Right, this is a change specific for WebKit as we are missing the empty.txt file.
I'll fix it on our side.

youennf:
PR https://github.com/web-platform-tests/wpt/pull/14778 is reverting empty.txt change.

PR #14651, opened 2018-12-21
[Azure Pipelines] Use raw URL with `brew cask install`

foolip:
Due to caching at some level, Safari TP on Azure Pipelines is still 
release 70, although it should now be 72:
https://github.com/Homebrew/homebrew-cask-versions/pull/6747


Hexcles (approved):
LGTM

foolip:
Confirmed in the logs the newest version was now installed. It's possible that some other caching is in place which will break this again later, but we'll see.

Also have to make the same change for Chrome and Firefox.

PR #14669, opened 2018-12-26
[css-text-3] tests for shaping at wrapping opportunities

frivoal:


frivoal:
@fantasai & @r12a  I'm reasonably confident these tests are right, but I wouldn't mind you double checking.

(Note: passes in Firefox, fails in the webkit/blink family of browsers)

fantasai (dismissed):
So, actually, the Uyghur hyphen (the one used for hyphenated breaking, rather than hyphenated words) doesn't look like either of those characters. See https://github.com/alif-type/amiri/issues/100 and https://lists.w3.org/Archives/Public/www-international/2014AprJun/0162.html and  https://github.com/w3c/alreq/issues/108

I think the best thing to do here would be to mask out the hyphen, by making it transparent or something. Then you can just check that the shaping is as expected.

You'll also need the test phrase duplicated without the masking so it's visible and have a mismatch reference breaking using ZWJ without any hyphen to prove that some kind of hyphenation character does appear there.

frivoal:
@fantasai updated based on your feedback. Re-review appreciated.

r12a:
Sorry, i seem to have missed this. I have a worry about these tests. Take hyphens-shaping-001.html. The test has a span around the shy, ie.
```
 <div dir=rtl lang=ug id=test>دامي<span>&shy;</span>دى</div> 
```
but we know that span markup can break cursive connections in Chrome & Safari, so the fact that those browsers fail this test may be down to the use of the span, rather than what the test is supposed to be testing.

frivoal:
That behavior is in itself a bug, and ought to be corrected too. Tests are there to exercise the features to get them a chance to fail, so that we can fix them if they do. With a sufficiently large body of tests, we can get confident that all features work. You're right that there are things other than the main purpose of this test that might cause it to fail, but they should get fixed too.

Tests should still be relatively focused, because if they exercise dozens of features at once, it will be hard to find out what is possibly failing, but that isn't the case here, and the other thing that can fail here is arguably more basic and fundamental than the expressed goal of the test.

It'd would be different if I were relying on something completely unrelated (like animations, or filters, or variable fonts or what have you), but here I'm not.

r12a:
Sorry to disagree, but the assertion "_Shaping characters are still shaped as if the word were not broken when the word is hyphenated._" quite rightly says nothing about spans. Afaict, you only added the span there for cosmetic reasons related to viewing the test – it wouldn't be there normally. It's quite possible that the assertion actually holds for Chrome and Safari, but you aren't able to tell that from this test as it is. (As it happens, they don't.) But neither can you say that they don't from the results of this test. You could, though, if the test were designed differently.

frivoal:
> Sorry to disagree, but the assertion "Shaping characters are still shaped as if the word were not broken when the word is hyphenated." quite rightly says nothing about spans. 

Indeed, but that doesn't change the fact that styles having no effect on hyphenation is a spec requirement as well:

> However, inline element boundaries must be ignored when determining word boundaries. 

and that this is [covered in the test suite](https://github.com/web-platform-tests/wpt/blob/master/css/css-text/hyphens/hyphens-span-001.html)

I also depend on the fact that the hyphens property works (or that divs are block elements, or that the font-family or width properties work...). If basic things misbehave, the test will fail for reasons unrelated to shaping.

So the test will fail on safari and edge (`hyphens` is only supported through a prefix), as well as in Chrome due to the bug you are referring to.

But as long as the things I am depending on:
* Are spec requirements as well
* Have tests of their own
* Are in the same spec or more mature specs
* don't make the test needlessly complicated

I don't think it is blocking.

> Afaict, you only added the span there for cosmetic reasons related to viewing the test

Not cosmetic reasons: in order to make the test automated. The spec doesn't mandate what character must be used for hyphenation, so If it is going to be visible, I cannot make a ref for it.

> You could, though, if the test were designed differently.

If you have a different automated design, I'm interested. I think the problem you're describing is not a fatal one, but if we can simplify the test further and keep it automated, that's even better. If it comes at the cost of automation, or of making the test significantly more complicated, then I'd continue to think this is an acceptable compromise.

r12a:
Fwiw, i created a set of tests for the i18n test suite that can be accessed from this results page: https://w3c.github.io/i18n-tests/results/word-break-shaping

They aer similar, but have a different design.

PR #14670, opened 2018-12-26
[css-text] Test that overflow-wrap doesn't break grapheme clusters

frivoal:


frivoal:
@fantasai & @r12a, I'm reasonably confident these tests are right, but I wouldn't mind you double-checking.

fantasai:
I'll have to defer to @r12a on the grapheme cluster questions, but I'd include enough text to make sure that `overflow-wrap` is taking effect, to prove that it allows breaks between grapheme clusters but not within them.

PR #14671, opened 2018-12-26
[css-text] Test for relaxation of word-break:keep-all when overflowing

frivoal:


css/css-text/overflow-wrap/overflow-wrap-normal-keep-all-001.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Text level 3 Test: word-break-keep-all and overflow-wrap:normal</title>
+<link rel="author" title="Florian Rivoal" href="https://florian.rivoal.net/">
+<link rel="help" href="https://www.w3.org/TR/css-text-3/#overflow-wrap-property">
+<link rel="match" href="reference/overflow-wrap-normal-keep-all-001-ref.html">
+<meta name="flags" content="may"
```
fantasai: Syntax error

fantasai:
Looks good. You have a syntax error in the test, though: missing > on the meta tag. Fix that and merge.

PR #14673, opened 2018-12-26
[css-text] Arabic cursive joining and ZWJ

frivoal:
This is a tentative test. Review from @r12a, @fantasai or @behdad very much appreciated.

As far as I understand:

1. putting a U+200D ZERO WIDTH JOINER after an arabic letter that is otherwise not followed by more arabic letters can be used force it to take it's medial form instead of its final form
2. putting a U+200D ZERO WIDTH JOINER before an arabic letter than is otherwise not preceeded by more arabic letters can be used to force it to take its media form instead of its initial form.

As was working on other tests, It looked like all browsers were correctly doing 1, but Chrome and Safari were not doing 2.

So I though I'd write a test for that, and that's what this PR is doing.

However:
1. Is this actually mandated by a spec? If yes, which? I couldn't identify the link I should put in the `<link rel="help" href="?????">`
2. This test fails in Chrome and Safari due to not doing 2, which is what I expected. See this screenshot: 
screenshot 2018-12-26 23 44 42

3. It also fails in Firefox for reasons that are less clear to me, since I don't actually read any arabic-script language. As far as I can tell, Firefox is doing the right thin in terms of shaping, but uses slightly different gliphs for `U+FE98 : ARABIC LETTER TEH MEDIAL FORM / U+FEE6 : ARABIC LETTER NOON FINAL FORM` vs `U+062A : ARABIC LETTER TEH / U+0646 : ARABIC LETTER NOON`. Presumably that may be coming from the font, which would make it conforming even if annoying, but that's a bit surprising since the same font (Geeza Pro) doesn't do that in in Chrome. Is that what's going on? If it's not what's going on, what could be? Does this look like a bug in Firefox, or is my test relying on something that cannot be relied on?
screenshot 2018-12-26 23 54 32


fantasai:
It looks like what's happening is that Firefox is running the text through a more sophisticated font-rendering path, which is giving you contextual glyphs for the teh-noon combination, and there are no equivalent glyphs for the pre-shaped codepoints. To fix the test, you need to use a dumber font which also covers all the relevant code points.

I would also update the test to be more comprehensive: check that initial, medial, and final forms are correctly triggered by ZWJ by inserting one between teh and noon. (And I suppose we should also have equivalent tests for ZWNJ and ZWSP.)

As for what spec to tag.. I'd say https://www.unicode.org/versions/Unicode11.0.0/ch23.pdf is a good landing point, with https://www.unicode.org/versions/Unicode11.0.0/ch09.pdf as a secondary reference. As for pulling into the CSS specs, maybe we should normatively reference Unicode somehow from CSS Text.

frivoal:
@fantasai Alright. Updated the tests, to cover:
* zwj
* zwj without language declared ('cause it shouldn't make a difference, but it does)
* initial / medial / final positions of letters that have all 3
* initial / medial / final positions of letters that don't have all 3, for sanity check
* using tatweel instead of zwj, for sanity check
* Disjoining Enforcement with zwnj
* Disjoining Enforcement with zero width space

The good news is that all parts of all tests are passed by at least one browser when the language is tagged. The bad news is that no browser passes everything (even discounting the case where the language isn't tagged).

r12a:
I have some concerns about these tests. Since no font is applied, the test will use whatever system font is available in the absence of a lang attribute, or in some browsers will use a font set by the browser.

In the test without lang attribute, my computer shows a final glyph that could also be an isolated glyph (no lead in from the right - common in simpler fonts).  The test that does use lang uses a completely different system font.

Furthermore, the font chosen by the system may or may not have glyphs for the presentation forms used as refs. In order to see the result more clearly i applied my preferred Scheherazade font, and found that the shapes are very different in the left and right columns because of this.

I think you need to set a font, and i suggest using the Noto Naskh Arabic font, which should be available in the WPT font directory. It has distinctive shapes for the joining forms, and uses the same shapes for the presentation forms.

r12a:
Just to be clear: Some browsers apply a font if there is a lang tag, in the same way as they apply Japanese/Chinese fonts to han characters depending on the language. We have a test for this at https://w3c.github.io/i18n-tests/results/default-font

r12a:
Also, i don't think you should add these tests to the i18n directory. I think you should add it to a directory that reflects the structure of the CSS Text spec, ie. text-encoding.

As i mentioned in a previous (offline) discussion, i have a set of replacement tests lined up for those currently in the i18n directory.  I would like to add those to WPT in several directories referring to the relevant sections and remove the i18n directory completely.

r12a (dismissed):
See my comments in the main dialogue.

frivoal:
@r12a I've moved the files as you suggested.

As for the font, I'm confused. Adding a (good) font to make sure the test work on system which don't have coverage for the relevant glyphs makes plenty of sense to me. And if it had no other effect, I'd be happy about that.

So I tried adding Noto Naskh Arabic, but it makes browsers pass many more cases than they did without specifying the font (both for when the language is specified and when it is not). Now, as a web designer, great, I'll just use a good font and get better results. But from the point of view of testing, isn't Arabic shaping supposed to work in any font, including system fonts (assuming the font has glyphs for Arabic of course)?

Lots of sites don't specify a font, or specify one but then it fails to load. It seems to me that sort of stuff still ought to be working, and therefore that having tests that don't rely on a specific font is a good thing if it exposes a problem.

Beyond having a font that supports Arabic at all, why is this dependent on the font?

Or should I have a double set of tests, one with the font specified and one not, to check that the browsers do it right when stars align **and** in the more general case?

r12a:
It's difficult for me to know where the problem lies, given that your system is different from mine, unless i can see, for the cases that fail for you, what font is actually being applied to what characters.

Essentially, i think the problem may be with the font, rather than with the browser.  For me, all the tests work fine, on Firefox, Chrome, Safari, Opera, Vivaldi, iPad, and Edge, with the exception of the following on Webkit/Blink engines:

screen shot 2019-01-11 at 10 57 30

The inspector tells me that the zwsp glyphs come from the Geeza Pro font, but the BEH glyph comes from Helvetica.  This might explain why, if you try the same thing at https://r12a.github.io/pickers/arabic/?text=%20%E2%80%8D%D8%A8%E2%80%8D%09%EF%BA%92 and apply a Geeza Pro font to the element, the result is what you'd expect.

Btw, if you use BEH for the test you have to squint a bit sometimes to see differences, not to mention special joining forms.  A better choice imo would be AIN, which has very different shapes for each position, ie. ععع ع.


There are at least three possible factors that are obscuring the results:
A. a particular font may not support the link between zwj and a character, or font fallback may be affecting your result.
B. the characters may look different or wrong, but may actually be right, eg. no difference between initial and medial forms of TA in simple fonts
C. the font may not have a repertoire that covers the (deprecated) presentation forms.

I suspect (A) may be a factor here. Not all fonts on my system produce joining forms from zwj, but most do.  (I'll put up an exploratory test for that shortly.)

Wrt (B), looking at the images of words [above](https://github.com/web-platform-tests/wpt/pull/14673#issue-240992781): The image under (2) may be ok, because some fonts don't show a lead-in stroke.  If you are using similar but actually different fonts for certain characters, there may not be an issue.  Hard to tell without being able to see the characters used and which font was applied. The image under (3) passes the actual test perfectly well - the higher join between TA and N is a stylistic thing that happens in more traditional fonts.  The fact that they look different is not relevant as to whether the test passed or not. It does of course spoil the ref test if the fonts are different. Again, i'd need to see what characters are used, where to understand better.

r12a:
cc. @xfq @himono for the educational value of this discussion about test design.

r12a:
I wonder whether this test is getting closer to what you originally set out to do:
https://w3c.github.io/i18n-tests/incubator/zwj/zwj_shaping.html

It shows that zwj produces the expected shaping behaviour reliably only when dir=rtl.

If the base direction is ltr, Webkit/Blink fail when there are LRM characters alongside, and Firefox fails in what seems to be any situation where bidi isolation is applied around the text item (eg. bdi, table cell, flexbox item), too.

r12a:
I ran through the fonts on my system in the pulldown list at https://r12a.github.io/pickers/arabic/?text=%20%D8%B9%20%D8%B9%E2%80%8D%20%E2%80%8D%D8%B9%E2%80%8D%20%E2%80%8D%D8%B9 (I have a lot of them).  All the Microsoft fonts seemed fine on Edge+W10, but 3 fonts on FF/Chrome/Safari+Mac didn't acknowledge the presence of zwj: Al Bayan, KufiStandardGK, Decotype Naskh, and Nadeem.  I'm guessing this is because of the design of those fonts(?).

cc @behnam @shervinafshar

frivoal:
I guess I am still confused: on the one hand, of course, it is possible to run more reliable experiments by picking the font. But our goal isn't generally to make tests that pass reliably, but rather to make tests that can pass but that do fail if the browser is doing something wrong. I am not trying to find out if Arabic can work, but to find if it reliably works in all the cases where it is supposed to.

As far as I can tell, as long as an Arabic font is present at all, it should always be possible for the browser to show the correct positional variant (whether by passing `&zwj;` and &zwnj;` to the font subsystem, or by picking the positional code points).

Therefore, the fact that it only works in some cases with some fonts seems to me to be a bug, and I don't want to hide the bug by picking a combination of settings that happens not to trigger it.

r12a:
@frivoal a fundamental concern for me is that i was unable to replicate the results you have obtained, and because i don't know how your test platform is set up i'm still not exactly clear what caused your issues. 

Reliability or replication being a basic tenet of experimental design, if some people can run a set of tests and find that they all pass, while others find tests that don't pass, then i think this indicates a problematic test. To make the tests reliable, i think we need to explore more deeply what is causing the failures, and then design tests with additional controls to isolate the relevant factors.  

For example, the tests that i created at https://w3c.github.io/i18n-tests/incubator/zwj/zwj_shaping.html indicated that base direction has an important impact on the result, independent of any font choice. This is quite possibly a browser bug.

On the other hand, I also did tests with various fonts that indicate that if you keep the other factors the same, the particular font used can create a difference in the results. This is why i'm concerned that your tests don't control the fallback font used, since one person's fallback font will produce different results from another's when the test is run, ie. some people may think all your tests pass. 

> As far as I can tell, as long as an Arabic font is present at all, it should always be possible for the browser to show the correct positional variant

What i discovered is that in fact that's not true – the font chosen can affect the behaviour, and cause failures that have nothing to do with the browser at all. 

> or by picking the positional code points

If you by that mean using compatibility characters, you also introduce variability into the results based on whether a tester's fallback font has glyphs for those code points or not.  Some will find that the tests pass, others will not.

Is that any clearer?

r12a:
Some examples of problems using ref tests that rely on compatibility characters, using very common fonts.  The sequence of characters is:
```
 U+0020 SPACE
‍ U+200D ZERO WIDTH JOINER
ب U+0628 ARABIC LETTER BEH
‍ U+200D ZERO WIDTH JOINER
U+0009 [control]
ﺒ U+FE92 ARABIC LETTER BEH MEDIAL FORM
```
Only the font was changed.

Geeza Pro (basic Mac font): appears to match.

Screenshot 2019-04-17 at 11 16 12

Al Bayan (basic Mac font): fails to render the zwj properly, so fails to match the compatibility character. Seems a reasonable result.

Screenshot 2019-04-17 at 11 09 55

Scheherazade: has no glyphs for the compatibility characters, so ref doesn't match. Indicates a problem with the test, rather than a failure related to zwj.

Screenshot 2019-04-17 at 11 09 28

Noto Naskh Arabic: Seems to use different glyphs for the compatibility characters and ordinary medial characters. 

Screenshot 2019-04-17 at 11 09 40

The test would indicate a failure but actually it's not entirely clear whether this is a failure or whether the glyph used with zwj is appropriate for the given context (simple fonts may not have a lead in for the medial form). See, for example, the position of the cursor if i add another glyph immediately before the BEH:

Screenshot 2019-04-17 at 11 22 12

frivoal:
Eureka!

The problem isn't about which font is used for arabic, but whether the zwj (and friends) come from the same font as the letters or not, and if it's a different font, why (font fallback vs setting it explicitly).

All browsers do the right thing when everything is from the same font (except for zero width space, which only firefox deals with correctly). 

If the zwj (and friends) come from a different from because of font fallback, firefox stops doing the shaping, while chrome and safari continue to do it right.

If the zwj (and friends) come from a different from because of specifying that font expicitely, then only chrome canary (v75) gets it right (current chrome v73 fails).

Updated the tests to take all that into account, now with explict `@font-face` for everything, for cross platform reproducibility.

behdad:
> All browsers do the right thing when everything is from the same font (except for zero width space, which only firefox deals with correctly).

What's the problem with Chrome? cc @drott

drott:
> What's the problem with Chrome? cc @drott

Is there still one? Looks like frivoal@ found out a working solution in https://github.com/web-platform-tests/wpt/pull/14673#issuecomment-484066336 - 

Indeed Chrome generally tries to render as many Unicode grapheme clusters with the primary font, then with the secondary, etc. until performing per-character fallback. In that sense, shaping performs better when the primary font has full coverage, or at least, when it does not contain any partial coverage, which would then cause breaking up runs and breaking subsequent shaping.

frivoal:
@behdad
> What's the problem with Chrome?

I'm seeing greatly improved behavior between 73 and 75 (or maybe I have some experimental flags on in Canary, not sure, either way, newer is better) and almost everything now works fine, but one thing that still isn't right is that `zero width space` is supposed to break shaping like `zero width non joiner` ([see table 9.3 in unicode chapter 9](https://www.unicode.org/versions/Unicode11.0.0/ch09.pdf)), but it doesn't.

behdad:
> @behdad
> 
> > What's the problem with Chrome?
> 
> I'm seeing greatly improved behavior between 73 and 75 (or maybe I have some experimental flags on in Canary, not sure, either way, newer is better) and almost everything now works fine, but one thing that still isn't right is that `zero width space` is supposed to break shaping like `zero width non joiner` ([see table 9.3 in unicode chapter 9](https://www.unicode.org/versions/Unicode11.0.0/ch09.pdf)), but it doesn't.

I agree with you that ZWSP breaking shaping sounds correct to me.  However, that is NOT what Unicode specifies.  The table you refer to talks about "all spacing characters", which ZWSP is NOT.  It does NOT say "all space characters".

The source of the truth though, as far as HarfBuzz is concerned, is this file:

https://www.unicode.org/Public/UCD/latest/ucd/ArabicShaping.txt

Since U+200B is NOT listed there, and if you read the header:
```

# Note: Code points that are not explicitly listed in this file are
# either of joining type T or U:
#
# - Those that are not explicitly listed and that are of General Category Mn, Me, or Cf
#   have joining type T.
# - All others not explicitly listed have joining type U.
```
and given that U+200B is general-category=Cf, then it has deduced type T and as such should NOT break shaping.

I agree this is not intuitive nor what I expect.  Maybe @roozbehp can shed some light.  If you have strong opinion about this, I encourage you to report it to Unicode: https://unicode.org/reporting.html

In the mean time, Chrome behavior is correct according to Unicode.  I also expect Firefox to behave the same BTW.

roozbehp:
U+200B is supposed to be transparent with regard to joining, as @behdad pointed out. If you look at Unicode Character Database's XML files, you also see that its joining type is listed as Transparent.

This is intentional. See Section 5.21 of Unicode Core Specification 12, page 249:

"***Characters Ignored in Cursive Joining***

U+200C ZERO WIDTH NON-JOINER and U+200D ZERO WIDTH JOINER are format controls
specifically intended to influence cursive joining. However, there are other format controls
which are explicitly ignored when processing text for cursive joining. In particular, U+2060
WORD JOINER, U+FEFF ZERO WIDTH NO-BRAK SPACE, and U+200B ZERO WIDTH SPACE
influence text segmentation and line breaking, but should be ignored for cursive joining.
U+034F COMBINING GRAPHEME JOINER is also ignored for cursive joining.

More generally, there is a broad class of characters whose occurrence in a string should be
ignored when calculating cursive connections between adjacent letters subject to cursive
joining. This class is defined by the property value, Joining_Type = Transparent, and
includes all nonspacing marks and most format characters other than ZWNJ and ZWJ. See
the detailed discussion of cursive joining in *Section 23.2, Layout Controls*."

frivoal:
> The table you refer to talks about "all spacing characters", which ZWSP is NOT. It does NOT say "all space characters".

Indeed, that's where I went wrong. Maybe Unicode should switch to bikeshed and link from the terms used to their definition more pervasively :) 

I'll update my tests accordingly.

> If you have strong opinion about this, I encourage you to report it to Unicode: https://unicode.org/reporting.html

I don't. It would have made sense to me that it'd break shaping, as it is a word divider, but I am not an expert on Arabic, so I'm OK with going with the flow.  However, if you think it's wrong, maybe we should look deeper into it.

> I also expect Firefox to behave the same BTW.

According to my tests, Firefox does treat ZWSP as breaking shaping.

r12a:
@roozbehp this is slightly veering off course from this thread, but in case there's a simple answer.. 

I'm actually surprised about ZWSP not breaking cursive joining. My understanding is that when using nastaliq script, esp. in Urdu, inter-word spaces are often not applied, because words are separated enough by the arrangement of glyphs along the sloping baselines.  If you do, however, want to indicate word boundaries in those situations without unsightly spacing, using ZWSP seems to be an obvious way of doing so.  What's the rationale behind ZWSP not breaking cursiveness?

frivoal:
I suggest:
* putting the ZWSP tests in a separate PR, not to be merged yet
* Merging the rest
* Opening an issue on the CSSWG to follow up on ZWSP and arabic shaping. If we want to change it we'll have to follow up with Unicode, but at least that makes sure we have a place to discuss, and that it doesn't get lost.

frivoal:
Opened https://github.com/w3c/csswg-drafts/issues/3861 on the csswg to follow up

frivoal:
> putting the ZWSP tests in a separate PR, not to be merged yet

here: https://github.com/web-platform-tests/wpt/pull/16427

khaledhosny:
> @roozbehp this is slightly veering off course from this thread, but in case there's a simple answer..
> 
> I'm actually surprised about ZWSP not breaking cursive joining. My understanding is that when using nastaliq script, esp. in Urdu, inter-word spaces are often not applied, because words are separated enough by the arrangement of glyphs along the sloping baselines. If you do, however, want to indicate word boundaries in those situations without unsightly spacing, using ZWSP seems to be an obvious way of doing so. What's the rationale behind ZWSP not breaking cursiveness?

This not unique no Nastaliq, almost all Arabic calligraphy hands make no distinction between inter-word and intra-word spaces, they are treated the same. So it makes no much sense for Nastaliq to use a different space charter, it should be up to the font to handle this.

https://en.wikipedia.org/wiki/Kufic#/media/File:Folio_from_a_Koran_(8th-9th_century).jpg
image

https://en.wikipedia.org/wiki/Islamic_calligraphy#/media/File:Qur'anic_Manuscript_-_Mid_to_Late_15th_Century,_Turkey.jpg
image

https://commons.wikimedia.org/wiki/Category:Islamic_calligraphy#/media/File:Containing_Haf%C4%B1z_Osman's_calligraphies_-_Murakka_(calligraphic_album)_-_Google_Art_Project.jpg
image

PR #14678, opened 2018-12-27
Fix key code for NUMPAD4 and NUMPAD5

JohnChen0:


JohnChen0:
Right now, webdriver/tests/perform_actions/support/keys.py says the key code for NUMPAD4 key is PageDown, and the code for NUMPAD5 is PageUp. I assume this must be a typo?

andreastt (approved):

andreastt:
Yes 😄  Good catch!

PR #14679, opened 2018-12-28
[css-text] Check interraction of tab-size, letter-spacing, word-spacing

frivoal:


frivoal:
@fantasai @kojiishi I think I got this right, but it's failing (differently) in all browsers, so I wouldn't mind a quick sanity-check review.

fantasai:
Looks correct to me. However, I would not use the overlapping technique for this one if there's a problem with antialiasing. I'd just stack the references and say something like “There should be a blue square with perfectly smooth sides below.” or something, and make each one a 2:1 rectangle.

PR #14702, opened 2019-01-03
Fix Mojo detection in bluetooth-helpers.js

Hexcles:
Avoid throwing a reference error when Mojo is not defined.

### DIFF
--- a/bluetooth/resources/bluetooth-helpers.js
+++ b/bluetooth/resources/bluetooth-helpers.js
@@ -18,8 +18,8 @@ function loadScripts(paths) {
 }
 
 function performChromiumSetup() {
-  // Make sure we are actually on Chromium.
-  if (!Mojo) {
+  // Make sure we are actually on Chromium with Mojo enabled.
+  if (typeof Mojo === 'undefined') {
     return;
   }
 


jgraham (approved):

jyasskin (approved):
This seems right to me. @cco3 FYI.

jyasskin:
I'm not sure what best practice is here. The file is intended to test the behavior of things inside `navigator.bluetooth`, so if that doesn't exist, all the tests should fail. It'd be fine to fail them all with a more concise message, if there's a way to do that, but they shouldn't pass on browsers that haven't implemented the API.

reillyeon:
Is there a general recommendation for _how_ tests should fail? I agree that triggering immediate failure in bluetooth_test() if navigator.bluetooth is undefined might clean things up but is it necessary?

reillyeon (approved):

Hexcles:
Since this PR itself is non-controversial, let's take this small but concrete improvement first and we can discuss early failures later.

Taskcluster failed because too many tests were affected (this helper script is used by almost all bluetooth tests) and the task timed out. In addition, within the iterations Taskclsuter did complete, there was no flaky result.

PR #14713, opened 2019-01-04
Fix resource-timing.html document.domain test - take 2

yoavweiss:
Replaces https://github.com/web-platform-tests/wpt/pull/14636
Because [we can't make sure a test loads onto the right domain](https://github.com/web-platform-tests/wpt/pull/14636#discussion_r243443726), I extracted the document.domain tests from the larger file and only loaded that test on the right domain, as an external document.

### DIFF
--- a/common/sleep.py
+++ b/common/sleep.py
@@ -1,0 +1,8 @@
+import time
+
+# sleep can be lower than requested value in some platforms: https://bugs.python.org/issue31539
+# We add padding here to compensate for that.
+sleep_padding = 15.0
+
+def sleep_at_least(sleep_in_ms):
+    time.sleep((sleep_in_ms + sleep_padding) / 1E3);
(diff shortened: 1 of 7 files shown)


Hexcles:
Just to double check the domain setup here: we have a test page that loads `http://www.web-platform.test:PORT/...` in a new window, which then loads another `http://www.web-platform.test:PORT/...` in an iframe, and finally the iframe sets its `document.domain` to `web-platform.test` (without `www`). Is this what you want?

yoavweiss:
> Just to double check the domain setup here: we have a test page that loads `http://www.web-platform.test:PORT/...` in a new window, which then loads another `http://www.web-platform.test:PORT/...` in an iframe, and finally the iframe sets its `document.domain` to `web-platform.test` (without `www`). Is this what you want?

Indeed. The test makes sure that a same domain resource without TAO remains visible even after it changed its document.domain to a different origin.

Hexcles (approved):
Alright, then this fix LGTM.

(I was going to save my approval because I don't know about resource timing, but after a closer look, the PR mostly just restructured the test to achieve the desired domain setup.)

yoavweiss:
Thanks for reviewing! :)

PR #14715, opened 2019-01-04
Prevent webdriver client Session.close() exception

### LINKED ISSUE #14714: TypeError: object of type 'NoneType' has no len() exception from webdriver client Session.close()
Trying `./wpt run` from a fresh clone results in many tests not being run at all due to an internal wpt failure in the webdriver client `Session.close()` code.

Here’s the full error message/trace:

```
  File "/private/tmp/wpt/tools/wptrunner/wptrunner/executors/executorwebdriver.py", line 274, in _run
  File "/private/tmp/wpt/tools/wptrunner/wptrunner/executors/executorwebdriver.py", line 335, in do_testharness
    parent_window = protocol.testharness.close_old_windows()
  File "/private/tmp/wpt/tools/wptrunner/wptrunner/executors/executorwebdriver.py", line 92, in close_old_windows
    self.webdriver.close()
  File "/private/tmp/wpt/tools/webdriver/webdriver/client.py", line 20, in inner
    return func(self, *args, **kwargs)
  File "/private/tmp/wpt/tools/webdriver/webdriver/client.py", line 569, in close
    if len(handles) == 0:
TypeError: object of type 'NoneType' has no len()
```

sideshowbarker:
This change adds handling for the case when the call to `self.send_session_command("DELETE", "window")` in `Session.close()` in `tools/webdriver/webdriver/client.py` returns None.

Otherwise without this change, the code ends up checking `len()` on the return value of `self.send_session_command("DELETE", "window")`, which can result in a _"TypeError: object of type 'NoneType' has no len()"_ exception.

Fixes https://github.com/web-platform-tests/wpt/issues/14714

---

If the code in this patch isn’t the right/best way (or the most idiomatic way in Python) to deal with the problem in #14714, then let’s use this PR branch to update the patch to fix it the right way.

### DIFF
--- a/tools/webdriver/webdriver/client.py
+++ b/tools/webdriver/webdriver/client.py
@@ -566,7 +566,7 @@ class Session(object):
     @command
     def close(self):
         handles = self.send_session_command("DELETE", "window")
-        if len(handles) == 0:
+        if handles is not None and len(handles) == 0:
             # With no more open top-level browsing contexts, the session is closed.
             self.session_id = None
 


andreastt (approved):

gsnedders:
Do we still have any test for the expected return value here?

PR #14719, opened 2019-01-04
[resource-timing] test_resource_timing.html to pass also when run in HTTPS

yoavweiss:
I spotted a recent [regression](https://staging.wpt.fyi/results/resource-timing/test_resource_timing.html?sha=cf8b055289) on wpt.fyi where test_resource_timing.html fails in experimental. I suspected it may be related to this test running in HTTPS, but I'm less sure that's the case.

Regardless, when I looked into it, it became clear that this test was never meant to pass under HTTPS, so fixing that here.

### DIFF
--- a/resource-timing/test_resource_timing.https.html
+++ b/resource-timing/test_resource_timing.https.html
@@ -1,0 +1,24 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="UTF-8" />
+        <title>window.performance Resource Timing Entries exist</title>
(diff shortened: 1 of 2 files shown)


nicjansma:
👍

yoavweiss:
The issue that's causing flakiness on the chrome dev taskcluster run is now fixed in https://chromium-review.googlesource.com/c/chromium/src/+/1396129

nicjansma (approved):

Hexcles:
Admin-merge as per the comment above.

PR #14728, opened 2019-01-04
[manifest] Make sure deleted tests are removed from manifest

### LINKED ISSUE #14568: Deleted tests are ERROR instead of MISSING for Taskcluster results
From https://github.com/web-platform-tests/wpt/commit/79001b1d0d4a111b7f44608404d7b5504d755a1c

[Chrome diff](https://staging.wpt.fyi/results/resource-timing?diff&run_id=6508819563675648&run_id=6003242655809536)
[Firefox diff](https://fyi-dot-wptdashboard-staging.appspot.com/results/resource-timing?diff&run_id=5481842887098368&run_id=5893510804799488)

Both [results for 79001b1d0d4a111b7f44608404d7b5504d755a1c](https://staging.wpt.fyi/results/resource-timing?sha=79001b1d0d&q=resource_timing_) include attempting to run the deleted files:
- /resource-timing/resource_timing_buffer_full_when_populate_entries.html
- /resource-timing/resource_timing_store_and_clear_during_callback.html

Hexcles:
This fix is developed in a TDD style. The first commit tweaks two existing test cases to make them more meaningful and cover case 1, and adds a new test case for case 2. The second commit fixes the two edge cases and passes the tests.

Fixes #14568 .

### DIFF
--- a/tools/manifest/manifest.py
+++ b/tools/manifest/manifest.py
@@ -79,6 +79,8 @@ class TypeData(object):
         return rv
 
     def __delitem__(self, key):
+        # First access the index to ensure it is loaded.
+        self[key]
         del self.data[key]
 
     def __setitem__(self, key, value):
@@ -279,6 +281,10 @@ class Manifest(object):
                     if old_hash != file_hash:
                         new_type, manifest_items = source_file.manifest_items()
                         hash_changed = True
+                        try:
+                            del self._data[old_type][rel_path]
+                        except KeyError:
+                            pass
                     else:
                         new_type, manifest_items = old_type, self._data[old_type][rel_path]
                     if old_type in reftest_types and new_type != old_type:
@@ -306,10 +312,7 @@ class Manifest(object):
                     _, old_type = self._path_hash[rel_path]
                     if old_type in reftest_types:
                         reftest_changes = True
-                    try:
-                        del self._path_hash[rel_path]
-                    except KeyError:
-                        pass
+                    del self._path_hash[rel_path]
                     try:
                         del self._data[old_type][rel_path]
                     except KeyError:
--- a/tools/manifest/manifest.py
+++ b/tools/manifest/manifest.py
@@ -79,9 +79,12 @@ class TypeData(object):
         return rv
 
     def __delitem__(self, key):
-        # First access the index to ensure it is loaded.
-        self[key]
-        del self.data[key]
+        if key in self.data:
+            del self.data[key]
+        elif self.json_data is not None:
+            del self.json_data[from_os_path(key)]
+        else:
+            raise KeyError
 
     def __setitem__(self, key, value):
         self.data[key] = value
--- a/tools/manifest/manifest.py
+++ b/tools/manifest/manifest.py
@@ -284,10 +284,11 @@ class Manifest(object):
                     if old_hash != file_hash:
                         new_type, manifest_items = source_file.manifest_items()
                         hash_changed = True
-                        try:
-                            del self._data[old_type][rel_path]
-                        except KeyError:
-                            pass
+                        if new_type != old_type:
+                            try:
+                                del self._data[old_type][rel_path]
+                            except KeyError:
+                                pass
                     else:
                         new_type, manifest_items = old_type, self._data[old_type][rel_path]
                     if old_type in reftest_types and new_type != old_type:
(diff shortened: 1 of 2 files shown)


gsnedders:
@jgraham any idea why Travis is failing here? given the changes here, it's surprising it would fail here but not on master?

jgraham:
I think that travis test is super-flaky.

tools/manifest/manifest.py
```diff
@@ -279,6 +281,10 @@ def update(self, tree):
                     if old_hash != file_hash:
                         new_type, manifest_items = source_file.manifest_items()
                         hash_changed = True
+                        try:
```
gsnedders: Can we not put this in a `if new_type != old_type` block? Why do we need to handle the exception here?

tools/manifest/manifest.py
```diff
@@ -306,10 +312,7 @@ def update(self, tree):
                     _, old_type = self._path_hash[rel_path]
                     if old_type in reftest_types:
                         reftest_changes = True
-                    try:
-                        del self._path_hash[rel_path]
-                    except KeyError:
-                        pass
+                    del self._path_hash[rel_path]
```
gsnedders: IIRC, this handles the `KeyError` for when we have an old manifest without the hash data, but maybe we just outright fail without that nowadays.

tools/manifest/tests/test_manifest.py
```diff
@@ -353,13 +353,13 @@ def test_no_update():
     assert list(m) == [("testharness", test1.path, {test1}),
                        ("testharness", test2.path, {test2})]
 
-    s1_1 = SourceFileWithTest("test1", "1"*40, item.TestharnessTest)
+    s1_1 = SourceFileWithTest("test1", "1"*40, item.ManualTest)
```
gsnedders: Let's rename the test, given this is now a fundamentally different test (and I think both are worth having, on the whole); this test was "no update" from the point of view of the manifest items, but now it has a notable change.

tools/manifest/manifest.py
```diff
@@ -79,6 +79,8 @@ def __len__(self):
         return rv
 
     def __delitem__(self, key):
+        # First access the index to ensure it is loaded.
+        self[key]
```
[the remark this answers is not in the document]
Hexcles: We do. The new `test_update_from_json` would fail without this (though I'd admit it's not obvious at all).

tools/manifest/manifest.py
Hexcles: We already accessed `self._path_hash[rel_path]` a few lines above without any exception handling, so if it doesn't exist, we wouldn't even get to this point.

tools/manifest/manifest.py
Hexcles: Is it necessary? We unconditionally set `self._data[new_type][rel_path]` later. I'd prefer having fewer conditional switches.

tools/manifest/tests/test_manifest.py
Hexcles: I guess I simply didn't understand the original test. What was it testing and what did "no update" mean?

The most interesting part is the next: it used to be just `m.update([(s1, True), (s2.rel_path, False)])` and the newly created `s1_1` wasn't used. It seems like it was simply testing the handling of the (rel_path, False) type argument.

tools/manifest/manifest.py
jgraham: I had the same comment. I think that although it increases the cyclomatic complexity of the function it also serves as documentation as to why we're doing it.

tools/manifest/manifest.py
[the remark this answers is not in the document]
jgraham: Yeah maybe. I think the patch is fairly simple; something like

```
if key in self.data:
    del self.data[key]
else:
    path = from_os_path(key)
    if self.json_data and path in self.json_data:
       del self.json_data[path]
```

but I don't insist that you make that change in order to land this.

tools/manifest/manifest.py
Hexcles: Done.

Hexcles:
Re Travis errors: the tools/wpt suite has a bunch of end-to-end integration tests that start the full browser and are hence quite memory intensive. Travis occasionally failed because "unable to allocate memory". I don't see how this PR affects memory usage so I assume the flakiness already exists.

jgraham (approved):

tools/manifest/tests/test_manifest.py
gsnedders: So the original goal of the test was that it produced an identical set of tests before/after the update. I could've screwed up when I originally wrote it, though. 🙂

PR #14752, opened 2019-01-08
[Azure Pipelines] Run full daily runs of Safari Technology Preview

foolip:


foolip:
https://dev.azure.com/foolip/wpt/_build/results?buildId=342 is a run that actually matches this branch. Shows that I failed to properly switch back to Safari TP after trying stable.

foolip:
> How're you planning to get these uploaded? We currently respond to Azure CheckRun events, which I don't think will be triggered by a daily task?

I suspect that the push runs also create checks as opposed to statuses, but it's probably easiest to verify when this has been merged as opposed to on my fork.

> Do we have any idea what effect this will have on the overall capacity.

We have 10 parallel builds, so once a day we'll be down to 6 for <2h. I mean to discuss raising our capacity, so picked a number that wouldn't be a huge problem. 8 is the number I think we should use eventually.

.azure-pipelines.yml
```diff
@@ -121,3 +122,29 @@ jobs:
   - template: tools/ci/azure/tox_pytest.yml
     parameters:
       directory: tools/wpt/
+
+- job: all_macOS
+  displayName: 'all tests (Safari Technology Preview)'
+  condition: eq(variables['Build.SourceBranch'], 'refs/heads/azure-daily-safari')
```
lukebjerring: Looks like (capacity allowing,) this would be easy to make run on all master commits by tweaking this condition.

foolip: Indeed!

foolip:
https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=2970 has finished, and the published results.zip has 4 files, wpt_report_1.json through wpt_report_4.json.  But the slowest shard took 2h19m. I'll just change the comment to reflect that, I don't want to use more parallel capacity yet.

.azure-pipelines.yml
```diff
@@ -121,3 +122,29 @@ jobs:
...
+  - template: tools/ci/azure/install_safari.yml
+  - template: tools/ci/azure/update_hosts.yml
+  - template: tools/ci/azure/update_manifest.yml
+  - script: no_proxy='*' ./wpt run --no-manifest-update --no-restart-on-unexpected --no-fail-on-unexpected --this-chunk=$(System.JobPositionInPhase) --total-chunks=$(System.TotalJobsInPhase) --chunk-type hash --log-wptreport $(Build.ArtifactStagingDirectory)/wpt_report_$(System.JobPositionInPhase).json --channel preview safari
+    displayName: 'Run tests'
+  - task: PublishBuildArtifacts@1
+    displayName: 'Publish results'
+    inputs:
+      artifactName: 'results'
+    condition: succeededOrFailed()
```
foolip: @lukebjerring this condition means we'd need to be careful not to use results.zip if the overall check didn't pass. But I thought publishing it even on failure might prove useful for debugging.

.azure-pipelines.yml
```diff
@@ -121,3 +122,29 @@ jobs:
...
+  - template: tools/ci/azure/checkout.yml
+  - template: tools/ci/azure/pip_install.yml
+    parameters:
+      packages: virtualenv
+  - template: tools/ci/azure/install_fonts.yml
+  - template: tools/ci/azure/install_certs.yml
+  - template: tools/ci/azure/install_safari.yml
+  - template: tools/ci/azure/update_hosts.yml
+  - template: tools/ci/azure/update_manifest.yml
+  - script: no_proxy='*' ./wpt run --no-manifest-update --no-restart-on-unexpected --no-fail-on-unexpected --this-chunk=$(System.JobPositionInPhase) --total-chunks=$(System.TotalJobsInPhase) --chunk-type hash --log-wptreport $(Build.ArtifactStagingDirectory)/wpt_report_$(System.JobPositionInPhase).json --channel preview safari
```
foolip: Note that I'm using `--chunk-type hash` here for a more even time between chunks. However, that does split directories across chunks. I'd like to see if there are differences to our current results and perhaps this will turn out to be a mistake.

lukebjerring (approved):

.azure-pipelines.yml
foolip: I was intending on keeping it, it's not just for initial debugging, so I'm not sure what a TODO or comment would say :)

Hexcles: IIRC, wptreport.json is only flushed to disk at the *end* of the test suite, so if anything goes wrong in the middle (e.g. `wptrunner` throws an uncaught exception), we'll have a empty file (not as in zero results, but an empty file, which isn't valid JSON). Hence, we probably won't get much value from archiving the output files when the process exits abnormally.

.azure-pipelines.yml
foolip: It was quite uneven in my testing with 8 chunks, maybe the slowest is 2x the fastest. I also expect to see discrepancies, but am curious to see how many they are. We'll probably remove this bit unless it's not as bad as suspected.

foolip:
Hmm, it didn't work. epochs/daily was updated around UTC midnight (40 minutes ago) and https://dev.azure.com/web-platform-tests/wpt/_build?definitionId=1 should lists all builds.

https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=3108 is the "CI build" for commit e2ac2e7d7539c4eb9963b7f23dbb3f7692fce537, but judging by the timing it was triggered by the master push. I see that all macOS agents are idle too, and the jobs would still be running if they'd been started.

I suspect that this is in some way a consequence of checks being defined not for a symbolic ref (branch/tag) but for a commit sha. Or possibly Azure Pipelines treats PRs and pushes separately even if for the same commit, but has only one bucket for pushes, so that one branch trailing another doesn't work. @thejohnjansen @mustjab can you help us ask around if it's possible to achieve what we're trying to do with Azure Pipelines?

foolip:
Investigation continues in https://github.com/web-platform-tests/wpt/issues/14836.

PR #14766, opened 2019-01-09
Import wasm/jsapi/limits.any.js.

Ms2ger:
From https://github.com/WebAssembly/spec/blob/b42efa9b07c5544079c31f6088a66bead617559c/test/js-api/limits.any.js


jgraham (approved):

Ms2ger:
Still some timeouts with the latest changes, unfortunately.

CC @xtuc

xtuc:
I would like to increase the timeout of these tests, I'm wondering why `setup({timeout_multiplier: 10 * 60 * 1000});` at the begining doesn't work?

The `timeout=long` is not sufficient.

PR #14775, opened 2019-01-09
Encoding: align encodeInto() test with reality

annevk:
While IDL said this should throw, that was in error: https://github.com/heycam/webidl/pull/605.

### DIFF
--- a/encoding/encodeInto.any.js
+++ b/encoding/encodeInto.any.js
@@ -142,8 +142,8 @@ test(() => {
   new MessageChannel().port1.postMessage(buffer, [buffer]);
   ({ read, written } = new TextEncoder().encodeInto("", view));
   assert_equals(read, 0);
-  assert_equalss(written, 0);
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):

ricea:
Interesting. My slow and dumb polyfill didn't pass this test, but I assumed it was just because I wasn't trying hard enough.

I will retest once you've fixed @Honry's comments.

annevk:
Thanks @Honry! Out-of-band Henri told me that it didn't work in workers, so I've fixed that too. And Domenic helped me figure out a way to make object destructuring work without a leading let/const. 😊

hsivonen:
With the latest changes changes to the test, my implementation of the feature passes the test, so the detachment part is now reality-aligned at least for Gecko purposes.

ricea:
Retested. All tests pass now, including in a worker.

PR #14786, opened 2019-01-10
[css-scroll-snap] Parsing scroll-margin*

ewilligers:
Test parsing and serialization of the properties
scroll-margin* and scroll-snap*

https://drafts.csswg.org/css-scroll-snap-1/#property-index


fantasai (changes requested):
Overall looks good, however

- You should include misordered values in the scroll-snap-type invalid tests, e.g. 'proximity y
- For the “invalid values” test, use a positive percentage rather than a negative one. That way we know it's being thrown out because it's a percentage, and not because it's negative.'.
- I would recommend merging the valid and invalid parsing tests into one file. It's faster to run, and equally understandable what's happening. (We want to optimize first for understandability and maintainability when deciding how to split tests; but all things being equal we might as well make it more performant.)
- As with the padding tests, maybe think about merging each set of longhands into one file.

r+ with the first two issues fixed, but think about the last two. :)

ewilligers:
I have added positive percentage tests and `proximity y`.

For all properties, I have kept separate the `parsing/*-valid.html` and `parsing/*-invalid.html` 
- for consistency between properties (across all of CSS)
- when properties are only supported by some browsers, the failures are easier to review
- some properties have many test cases, e.g. border-image, box-shadow, clip-path, filter, grid-area, transform

fantasai (approved):

PR #13688, opened 2018-10-23
[css-backgrounds] Initial values and inheritance

ewilligers:
Test that CSS Backgrounds and Borders properties have the initial
values given in the spec. None of the properties inherit.

https://drafts.csswg.org/css-backgrounds/

Identified bugs:

repeated keywords in initial value
~https://bugzilla.mozilla.org/show_bug.cgi?id=1501261~

border-image initial values
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/19381805/

box-shadow serialization order
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/19381950/

border-image-outset initial value
https://bugs.chromium.org/p/chromium/issues/detail?id=898203

PR #14787, opened 2019-01-10
[css-scroll-snap] Parsing some scroll-padding*

ewilligers:
Test that scroll-padding-* accept auto
and accept postive length-percentage.

Not yet tested: negative values, invalid values.

https://drafts.csswg.org/css-scroll-snap-1/#property-index


fantasai (commented):
Some comments:

- calc() values for padding should include a percentage, to ensure that length/percentage combinations are parsed
- Might consider pulling in the full set of tests for each set of longhands (physical, logical) into one file: it's still understandable, but also less work for the test runners to execute.
- When you get around to invalid tests, include a calc() with 'auto' inside it.

Otherwise looks good!

ewilligers:
Comments addressed. 

https://github.com/web-platform-tests/wpt/pull/13688 and https://github.com/web-platform-tests/wpt/pull/13572 also need review :-)

fantasai (approved):

PR #14797, opened 2019-01-10
Use get_host_info() for WebKit

cvazac:
The template `{{domains[www1]}}` doesn't work when running in the WebKit CLI because `www1.localhost` doesn't resolve. 

cc @youennf


yoavweiss (approved):
LGTM, although it'd be good if the WebKit infra wouldn't require jumping through these extra hoops.

PR #14814, opened 2019-01-11
xhr: Check if entry lists associated to new FormData instance and formData attribute are not identical.

tkent-google:
This is a test for https://github.com/whatwg/html/pull/4283


annevk (approved):
Ideally we'd also have a test that checks submission isn't affected.

Issue #11231, opened 2018-05-29
/html/semantics/selectors/pseudo-classes/link.html is broken


Zirro:
The test at [html/semantics/selectors/pseudo-classes/link.html](https://github.com/web-platform-tests/wpt/blob/master/html/semantics/selectors/pseudo-classes/link.html) matches elements against the `:link` pseudo-class.

Because it depends in part on elements which are shared across tests, it [was broken](https://wpt.fyi/results/html/semantics/selectors/pseudo-classes/link.html) by this mass-commit: https://github.com/web-platform-tests/wpt/commit/5cdc807b893fb2e033f8a99a4bf95a9dcce14f08#diff-e1ed1563923a0205d0ba3f1775edabe6

While the test could be fixed by merely adjusting for the removed element, I wonder whether relying on the `<link>`'s that are still present is considered acceptable.

kaixinjxq:
>BTW, test still fails on Chrome seems there's a bug in Chromium that it could not match ":link" to link elements, I will fire one later on.

The new bug link: https://bugs.chromium.org/p/chromium/issues/detail?id=921880

PR #14818, opened 2019-01-11
Delete non-existent id

xiuqijix:
- Fixed issue #11231.


zcorpan:
Isn't a better fix to put back a stylesheet link (but not point to testharness.css)?

zcorpan (approved):

PR #14819, opened 2019-01-11
Add a test for deleting HTMLCollection properties.

Ms2ger:


annevk (commented):
Each test seems to mirror c with its own c. Reading the comment it seems you forgot to remove those c's in the tests or am I missing something?

Ms2ger:
You were correct; fixed.

annevk (approved):

PR #14824, opened 2019-01-11
HTML: Tests for button layout

zcorpan:
Follows https://github.com/whatwg/html/pull/4143.

### DIFF
--- a/html/rendering/non-replaced-elements/form-controls/button-style.html
+++ b/html/rendering/non-replaced-elements/form-controls/button-style.html
@@ -1,0 +1,20 @@
+<!doctype html>
+<title>default style on buttons</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div class="tests">
+ <input type="reset">
+ <input type="button">
+ <input type="submit">
+ <input type="color">
+ <button></button>
+</div>
+<script>
+[].slice.call(document.querySelectorAll('.tests > *')).forEach(el => {
+  test(() => {
+    assert_equals(getComputedStyle(el).display, 'inline-block');
+  }, `default display on ${el.outerHTML}`);
+  test(() => {
+    assert_equals(getComputedStyle(el).boxSizing, 'border-box');
+  }, `default box-sizing on ${el.outerHTML}`);})
+</script>
--- a/html/rendering/widgets/button-layout/abspos.html
+++ b/html/rendering/widgets/button-layout/abspos.html
@@ -1,0 +1,31 @@
+<!doctype html>
+<title>abspos button with auto width, non-auto left/right</title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<link rel=help href=https://html.spec.whatwg.org/multipage/rendering.html#button-layout>
+<!--
+If the element is absolutely positioned, then for the purpose of the CSS visual formatting model, act as if the element is a replaced element.  [CSS]
+-->
+<link rel=help href=https://drafts.csswg.org/css2/visudet.html#abs-replaced-width>
+<!--
+If at this point the values are over-constrained, ignore the value for either 'left' (in case the 'direction' property of the containing block is 'rtl') or 'right' (in case 'direction' is 'ltr') and solve for that value.
+-->
+<style>
+#overconstrained { position: absolute; left: 100px; right: 100px; }
+#ltr { position: absolute; left: 100px; margin-top: 2em }
+#rtl { position: absolute; right: 100px; margin-top: 2em }
+</style>
+<button id=overconstrained>test</button>
+<button id=ltr>test</button>
+<button id=rtl>test</button>
+<script>
+for (const dir of ['rtl', 'ltr']) {
+  test(() => {
+    document.documentElement.dir = dir;
+    const overconstrained = document.getElementById('overconstrained');
+    const ref = document.getElementById(dir);
+    assert_equals(overconstrained.offsetLeft, ref.offsetLeft, 'offsetLeft');
+    assert_equals(overconstrained.clientWidth, ref.clientWidth, 'clientWidth');
+  }, `${document.title} (${dir})`);
+}
+</script>
--- a/html/rendering/widgets/button-layout/computed-style.html
+++ b/html/rendering/widgets/button-layout/computed-style.html
@@ -1,0 +1,31 @@
+<!doctype html>
+<title>computed style on buttons</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div class="tests">
+ <input type="reset">
+ <input type="button">
+ <input type="submit">
+ <input type="color">
+ <button></button>
+</div>
+<script>
+// "behave as" doesn't change computed value.
+const displayValues = ['inline', 'block', 'list-item', 'inline-block', 'table', 'inline-table', 'table-row-group', 'table-header-group', 'table-footer-group', 'table-row', 'table-column-group', 'table-column', 'table-cell', 'table-caption', 'none', 'contents', 'flow', 'flow-root', 'flex', 'grid', 'ruby', 'ruby-base', 'ruby-text', 'ruby-base-container', 'ruby-text-container', 'inline-flex', 'inline-grid'];
+for (const val of displayValues) {
+  [].slice.call(document.querySelectorAll('.tests > *')).forEach(el => {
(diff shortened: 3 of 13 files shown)


zcorpan:
Chromium bug for wrong baseline for inline-block buttons: https://bugs.chromium.org/p/chromium/issues/detail?id=894849

zcorpan:
Chromium bug for not wrapping `input type=button`:
https://bugs.chromium.org/p/chromium/issues/detail?id=922011

zcorpan:
Bugs for abspos.html test:
https://bugs.webkit.org/show_bug.cgi?id=193446
https://bugs.chromium.org/p/chromium/issues/detail?id=922054

zcorpan:
I think this is ready for review now.

zcorpan:
A thing I couldn't figure out how to test is the effect of `align-self: normal` on replaced vs non-replaced elements. Any advice?

zcorpan:
@MatsPalmgren or @emilio, are you able to review this?

emilio:
@MatsPalmgren is a better reviewer for this than me I think. @dholbert may be able to review this more easily too? But otherwise sure, I'll try to get to this over the week.

zcorpan:
Friendly ping :)

html/rendering/non-replaced-elements/form-controls/button-style.html
```diff
@@ -0,0 +1,20 @@
+<!doctype html>
+<title>default style on buttons</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
```
emilio: Could you link to the HTML spec PR from here? Or just adding links when it gets merged into the spec I guess.

html/rendering/non-replaced-elements/form-controls/button-style.html
```diff
@@ -0,0 +1,20 @@
...
+<div class="tests">
+ <input type="reset">
+ <input type="button">
+ <input type="submit">
+ <input type="color">
+ <button></button>
+</div>
+<script>
+[].slice.call(document.querySelectorAll('.tests > *')).forEach(el => {
```
emilio: nit: querySelectoAll should be iterable, so you can call forEach on it.

html/rendering/non-replaced-elements/form-controls/button-style.html
```diff
@@ -0,0 +1,20 @@
...
+  test(() => {
+    assert_equals(getComputedStyle(el).display, 'inline-block');
+  }, `default display on ${el.outerHTML}`);
+  test(() => {
+    assert_equals(getComputedStyle(el).boxSizing, 'border-box');
+  }, `default box-sizing on ${el.outerHTML}`);})
```
emilio: nit: I'd move the `})` to the next line.

html/rendering/widgets/button-layout/abspos.html
```diff
@@ -0,0 +1,31 @@
...
+-->
+<style>
+#overconstrained { position: absolute; left: 100px; right: 100px; }
+#ltr { position: absolute; left: 100px; margin-top: 2em }
+#rtl { position: absolute; right: 100px; margin-top: 2em }
+</style>
+<button id=overconstrained>test</button>
+<button id=ltr>test</button>
+<button id=rtl>test</button>
+<script>
```
emilio: just double-checking, is there a WebKit / Blink bug on this test failing for rtl?

html/rendering/widgets/button-layout/computed-style.html
```diff
@@ -0,0 +1,31 @@
...
+// "behave as" doesn't change computed value.
+const displayValues = ['inline', 'block', 'list-item', 'inline-block', 'table', 'inline-table', 'table-row-group', 'table-header-group', 'table-footer-group', 'table-row', 'table-column-group', 'table-column', 'table-cell', 'table-caption', 'none', 'contents', 'flow', 'flow-root', 'flex', 'grid', 'ruby', 'ruby-base', 'ruby-text', 'ruby-base-container', 'ruby-text-container', 'inline-flex', 'inline-grid'];
+for (const val of displayValues) {
+  [].slice.call(document.querySelectorAll('.tests > *')).forEach(el => {
+    el.style.display = ''
+    el.style.display = val;
+    const attrs = el.type ? ` type=${el.type}` : '';
+    const tag = `<${el.localName}${attrs}>`;
+    test(() => {
+     assert_not_equals(el.style.display, '', `display: ${val} is not supported`)
```
emilio: Maybe we should skip the test if the display value is not supported?

html/rendering/widgets/button-layout/display-other.html
```diff
@@ -0,0 +1,52 @@
...
+<script src=/resources/testharnessreport.js></script>
+<style>
+body { margin: 0 }
+.float { width: 100px; height: 100px; float: left; background: blue; margin: 10px }
+</style>
+<div class=float></div>
+<button id=button style="display: block;"><div class=float></div></button><span id=after>after</span>
+<script>
+// These should all behave as flow-root.
+const displayValues = ['run-in', 'flow', 'flow-root', 'table', 'list-item',
```
emilio: nobody implements run-in, so maybe remove it from the list? Though not a huge deal otherwise I guess?

html/rendering/widgets/button-layout/display-other.html
```diff
@@ -0,0 +1,52 @@
...
+                       'table-row-group', 'table-header-group', 'table-footer-group',
+                       'table-row', 'table-cell', 'table-column-group', 'table-column',
```
emilio: Also nobody seems to pass the table parts tests re. offsetLeft / offsetRight, are we sure we really want to treat them as flow-root?

emilio (approved):
Looks good to me with the gotcha of the table-pats bits. Do we understand why they don't pass?

html/rendering/non-replaced-elements/form-controls/button-style.html
zcorpan: The relevant section of the spec is supposed to be able to get from the last dir name for html, I think. Is that enough?

html/rendering/non-replaced-elements/form-controls/button-style.html
zcorpan: Yeah but it hasn't been for long (not sure it works cross-browser yet?), I didn't want to rely on newish features here.

html/rendering/widgets/button-layout/computed-style.html
zcorpan: No, that makes it hard to reason about test results between browsers and over time.

html/rendering/widgets/button-layout/display-other.html
zcorpan: Can we remove it from the spec? Cc @TabAtkins

html/rendering/widgets/button-layout/display-other.html
zcorpan: The table values also don't "work"... Not sure what actually happens but I figured this was close enough for web compat and still simple.

Or do offsetLeft/Top check computed value of display and have different behavior for table values?

html/rendering/widgets/button-layout/computed-style.html
emilio: Fair enough.

html/rendering/non-replaced-elements/form-controls/button-style.html
emilio: Fair :)

html/rendering/non-replaced-elements/form-controls/button-style.html
emilio: Ok, that works, thanks :)

html/rendering/widgets/button-layout/display-other.html
tabatkins: run-in is still a useful tool (the use-cases from 20 years ago are just as valid today, and just as annoying to work around), and we editors believe it's specified in an implementable, useful way. I'd prefer not to let chicken-and-egg issues ruin this.

gsnedders:
Is this waiting for anything or should this just land?

zcorpan:
I think this and the spec change should be ready to merge. They have both been reviewed. I'll merge both together, but will check with the HTML editors, first.

zcorpan:
Rebased on master. When checks are green, it's OK to squash and merge with this commit message

```
HTML: Tests for button layout

Follows https://github.com/whatwg/html/pull/4143.
```

PR #14825, opened 2019-01-11
HTML: Tests for button layout (WIP)

zcorpan:
Follows https://github.com/whatwg/html/pull/4143.

(More tests will be added to this PR.)

### DIFF
--- a/html/rendering/non-replaced-elements/form-controls/button-style.html
+++ b/html/rendering/non-replaced-elements/form-controls/button-style.html
@@ -1,0 +1,20 @@
+<!doctype html>
+<title>default style on buttons</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div class="tests">
+ <input type="reset">
+ <input type="button">
+ <input type="submit">
+ <input type="color">
+ <button></button>
+</div>
+<script>
+[].slice.call(document.querySelectorAll('.tests > *')).forEach(el => {
+  test(() => {
+    assert_equals(getComputedStyle(el).display, 'inline-block');
+  }, `default display on ${el.outerHTML}`);
+  test(() => {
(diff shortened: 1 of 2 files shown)


zcorpan:
I'm not sure what happened here. This and #14824 both result in 404 for me when trying to view them. (They loaded when I submitted the PR but then I couldn't add labels, and reloading the page gave 404.)

html/rendering/non-replaced-elements/form-controls/button-style.html
```diff
@@ -0,0 +1,20 @@
+<!doctype html>
+<title>default style on buttons</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
```
javifernandez: You should add a comment, using the 'meta' tag and  'assert' id,  to explain briefly what the test tries to verify.

html/rendering/widgets/button-layout/computed-style.html
```diff
@@ -0,0 +1,31 @@
+<!doctype html>
+<title>computed style on buttons</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
```
javifernandez: Ditto

jgraham:
Adding `<meta name=assert>` is explictly not required in web-platform-tests (although authors are of course free to do that if they like). Some specs may try to impose stronger requirements here, but for HTML there isn't any such rule.

html/rendering/non-replaced-elements/form-controls/button-style.html
zcorpan: Isn't it clear from the title and the test names?

zcorpan:
Please continue discussion / review in https://github.com/web-platform-tests/wpt/pull/14824

PR #14834, opened 2019-01-14
Unify the log level for wpt server

### LINKED ISSUE #10512: wpt run --log-*-level doesn't get passed through to wptserve
e.g., `./wpt run --log-mach-level debug firefox websockets/binary/001.html`

server.py includes:

```
            self.logger.debug("%i %s %s (%s) %i" % (response.status[0],
                                                    request.method,
                                                    request.request_path,
                                                    request.headers.get('Referer'),
                                                    request.raw_input.length))
```

As such, we should have log output for every HTTP request made to the server.

qiuzhong:
When `--log-*-level` option is used in `./wpt run` command line to set
the log level, it should be applied to the wpt server as well.
Fixed: #10512

### DIFF
--- a/tools/wptrunner/wptrunner/environment.py
+++ b/tools/wptrunner/wptrunner/environment.py
@@ -149,10 +149,6 @@ class TestEnvironment(object):
     def setup_server_logging(self):
         server_logger = get_default_logger(component="wptserve")
         assert server_logger is not None
-        log_filter = handlers.LogLevelFilter(lambda x:x, "info")
-        # Downgrade errors to warnings for the server
-        log_filter = LogLevelRewriter(log_filter, ["error"], "warning")
-        server_logger.component_filter = log_filter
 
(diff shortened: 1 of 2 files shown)


jgraham (changes requested):
That code can't just be deleted. Not treating server errors as a level higher than warning is important for gecko infra because we occasionally get tests that create errors in the handler script which would, if allowed to log at error level, cause jobs to fail.

foolip:
@qiuzhong Travis CI is green now, looks you rebased to trigger it again. Is this now waiting for re-review by @jgraham or what remains to be done?

qiuzhong:
@foolip , I fixed some the CI errors. Yes, it is now waiting for re-reviewing by the reviewers.

qiuzhong:
@jgraham, @gsnedders , @jugglinmike , would you please take a look at this?

Issue #14179, opened 2018-11-21
[wpt] `test_wpt.py` is flaky


jugglinmike:
The test titled `test_run_firefox` in `tools/wpt/tests/test_wpt.py` intermittently fails. I can't reproduce the failure locally, but we witnessed it in TravisCI during the review of gh-14178.

<details>
  <summary>Output of failing test</summary>

    tests/test_browser.py ..                                                 [  5%]
    tests/test_run.py .................                                      [ 48%]
    tests/test_wpt.py .....F.sss..........                                   [100%]
    =================================== FAILURES ===================================
    _______________________________ test_run_firefox _______________________________
    manifest_dir = '/tmp/tmpR8JUXI'
        @pytest.mark.slow
        @pytest.mark.remote_network
        @pytest.mark.xfail(sys.platform == "win32",
                           reason="Tests currently don't work on Windows for path reasons")
        def test_run_firefox(manifest_dir):
            # TODO: It seems like there's a bug in argparse that makes this argument order required
            # should try to work around that
            if is_port_8000_in_use():
                pytest.skip("port 8000 already in use")
        
            if sys.platform == "darwin":
                fx_path = os.path.join(wpt.localpaths.repo_root, "_venv", "browsers", "nightly", "Firefox Nightly.app")
            else:
                fx_path = os.path.join(wpt.localpaths.repo_root, "_venv", "browsers", "nightly", "firefox")
            if os.path.exists(fx_path):
                shutil.rmtree(fx_path)
            with pytest.raises(SystemExit) as excinfo:
                wpt.main(argv=["run", "--no-pause", "--install-browser", "--yes",
                               # The use of `--binary-args` is intentional: it
                               # demonstrates that internally-managed command-line
                               # arguments are properly merged with those specified by
                               # the user. See
                               # https://github.com/web-platform-tests/wpt/pull/13154
                               "--binary-arg=-headless",
                               "--metadata", manifest_dir,
                               "firefox", "/dom/nodes/Element-tagName.html"])
            assert os.path.exists(fx_path)
            shutil.rmtree(fx_path)
    >       assert excinfo.value.code == 0
    E       assert 1 == 0
    E        +  where 1 = SystemExit(1,).code
    E        +    where SystemExit(1,) = <ExceptionInfo SystemExit tblen=2>.value
    tests/test_wpt.py:190: AssertionError
    ----------------------------- Captured stdout call -----------------------------
    Running 1 tests in web-platform-tests

</details>

[...]

PR #14852, opened 2019-01-14
Add support for Epiphany web browser

mcatanzaro:
It's similar to the existing WebKit support, but a bit easier to use as
you can simply ./wpt run epiphany

Opportunistically improve a Servo error message, since we are adding
similar error messages below and consistency is nice.

Next step is to figure out how to run the tests under Weston. Currently they run in my normal desktop session, stealing input and guaranteeing that the automation tests all take forever and fail when I manually grab input back to whatever I was doing. That's why I don't have test results yet.

### DIFF
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -758,3 +758,32 @@ class WebKit(Browser):
 
     def version(self, binary=None, webdriver_binary=None):
         return None
+
+
+class Epiphany(Browser):
+    """Epiphany-specific interface."""
+
+    product = "epiphany"
+    requirements = "requirements_epiphany.txt"
+
+    def install(self, dest=None, channel=None):
+        raise NotImplementedError
+
+    def find_binary(self, venv_path=None, channel=None):
+        return find_executable("epiphany")
+
+    def find_webdriver(self, channel=None):
+        return find_executable("WebKitWebDriver")
+
+    def install_webdriver(self, dest=None, channel=None):
+        raise NotImplementedError
+
+    def version(self, binary=None, webdriver_binary=None):
+        if binary is None:
+            return None
+        output = call(binary, "--version")
+        if output:
+            # Stable release output looks like: "Web 3.30.2"
+            # Tech Preview output looks like "Web 3.31.3-88-g97db4f40f"
+            return output.split()[1]
+        return None
--- a/tools/wpt/run.py
+++ b/tools/wpt/run.py
@@ -422,7 +422,7 @@ class Servo(BrowserSetup):
             binary = self.browser.find_binary(self.venv.path, None)
 
             if binary is None:
-                raise WptrunError("Unable to find servo binary on the PATH")
+                raise WptrunError("Unable to find servo binary in PATH")
             kwargs["binary"] = binary
 
 
@@ -442,6 +442,29 @@ class WebKit(BrowserSetup):
         pass
 
 
+class Epiphany(BrowserSetup):
+    name = "epiphany"
+    browser_cls = browser.Epiphany
+
+    def install(self, venv, channel=None):
+        raise NotImplementedError
+
+    def setup_kwargs(self, kwargs):
+        if kwargs["binary"] is None:
+            binary = self.browser.find_binary()
+
+            if binary is None:
+                raise WptrunError("Unable to find epiphany in PATH")
+            kwargs["binary"] = binary
+
+        if kwargs["webdriver_binary"] is None:
+            webdriver_binary = self.browser.find_webdriver()
+
+            if webdriver_binary is None:
+                raise WptrunError("Unable to find WebKitWebDriver in PATH")
+            kwargs["webdriver_binary"] = webdriver_binary
+
+
 product_setup = {
     "fennec": Fennec,
     "firefox": Firefox,
@@ -456,6 +479,7 @@ product_setup = {
     "sauce": Sauce,
     "opera": Opera,
     "webkit": WebKit,
+    "epiphany": Epiphany,
 }
 
 
--- a/tools/wptrunner/requirements_epiphany.txt
+++ b/tools/wptrunner/requirements_epiphany.txt
@@ -1,0 +1,1 @@
+mozprocess == 0.26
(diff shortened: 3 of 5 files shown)


mcatanzaro:
The remaining CI failure at least seems to be unrelated:

```
manifest/tests/test_manifest.py:70: in <module>
    unique_by=lambda x: x.rel_path))
E   TypeError: lists() got an unexpected keyword argument 'average_size'
```

I found the documentation of the python-hypothesis library, [and the `average_size` parameter indeed doesn't exist](https://hypothesis.readthedocs.io/en/latest/data.html?highlight=lists#hypothesis.strategies.lists). The broken code hasn't changed since https://github.com/web-platform-tests/wpt/commit/8777171a5568e9fa967ef7a807dc400d89a07637 so not sure what could have changed.

tools/wptrunner/requirements_epiphany.txt
```diff
@@ -0,0 +1,2 @@
+mozprocess == 0.26
+selenium == 3.9.0
```
jgraham: I'm not sure that selenium is required since `ExecutorWebDriver` doesn't use it (some existing files are wrong here). We could probably add `mozprocess` to `requirements.txt` and eliminate a bunch of `requirements_*` files.

tools/wpt/run.py
```diff
@@ -422,7 +422,7 @@ def setup_kwargs(self, kwargs):
             binary = self.browser.find_binary(self.venv.path, None)
 
             if binary is None:
-                raise WptrunError("Unable to find servo binary on the PATH")
+                raise WptrunError("Unable to find servo binary in PATH")
```
jgraham: Not sure that this was a change important enough to make here.

jgraham (approved):
This looks good, but there are a couple of changes I would appreciate before landing.

tools/wpt/browser.py
```diff
@@ -758,3 +758,33 @@ def install_webdriver(self, dest=None, channel=None):
...
+        raise NotImplementedError
+
+    def version(self, binary=None, webdriver_binary=None):
+        if binary is None:
+            return None
+        output = call(binary, "--version")
+        if output:
+            # Stable release output looks like: "Web 3.30.2"
+            # Tech Preview output looks like "Web 3.31.3-88-g97db4f40f"
+            # So split off the first word, then truncate after the first hyphen.
```
foolip: Why not include the whole 3.31.3-88-g97db4f40f? Code that processes the version string (in wpt.fyi) could make it prettier if needed, but having the full string seems potentially useful if there can be updates which don't change the leading part. @lukebjerring opinions?

tools/wptrunner/wptrunner/browsers/epiphany.py
```diff
@@ -0,0 +1,75 @@
+from .base import Browser, ExecutorBrowser
+from .base import get_timeout_multiplier   # noqa: F401
+from .webkit import WebKitBrowser
+from ..executors import executor_kwargs as base_executor_kwargs
+from ..executors.executorwebdriver import (WebDriverTestharnessExecutor,  # noqa: F401
```
foolip: Nice to see ExecutorWebDriver being reused :D

foolip (approved):

tools/wptrunner/requirements_epiphany.txt
mcatanzaro: OK, I was confused by that too, but just copy/pasted from requirements_webkit.txt.

(What is `mozprocess` for?)

tools/wpt/run.py
mcatanzaro: Er, I guess I'll leave it then? Of course it's very easy to remove if desired.

(Trivia: it most likely comes from Spanish, which uses "en" for both "in" and "on," making it rather difficult to pick the right one when writing in English.)

tools/wpt/browser.py
lukebjerring: Can keep it + adapt to this format - fyi @Hexcles - this may need to be handled in our version-string special cases.

Hexcles: Is "Web" the name of the application? If so, the convention is to strip it, but all the remaining stuff can be kept, e.g. `3.30.2` or `3.31.3-88-g97db4f40f`.

mcatanzaro: The codename of the project is "Epiphany" and that's probably what we'd want to display on wpt.fyi, but the actual user-visible name is just "Web". (We can also write "GNOME Web" when needed to disambiguate.)

mcatanzaro:
OK, addressed all comments here. I think the only remaining problem is:

> The remaining CI failure at least seems to be unrelated:
> 
> ```
> manifest/tests/test_manifest.py:70: in <module>
>     unique_by=lambda x: x.rel_path))
> E   TypeError: lists() got an unexpected keyword argument 'average_size'
> ```
> I found the documentation of the python-hypothesis library, [and the `average_size` parameter indeed doesn't exist](https://hypothesis.readthedocs.io/en/latest/data.html?highlight=lists#hypothesis.strategies.lists). The broken code hasn't changed since [8777171](https://github.com/web-platform-tests/wpt/commit/8777171a5568e9fa967ef7a807dc400d89a07637) so not sure what could have changed.

which has me stumped, since other pull requests don't seem to be hitting this failure.

foolip:
@mcatanzaro, some tests only run based on what has changed in a PR, so it's possible it's already broken but just not showing up. I'll send a PR to probe for currently failing tests.

foolip:
Yep, the failure was already present: https://travis-ci.com/web-platform-tests/wpt/jobs/170769863

@jgraham I think we should admin merge this, but do you have final thoughts?

jgraham:
There are flake8 errors specific to this PR. Please fix those before merging.

foolip:
https://github.com/web-platform-tests/wpt/pull/14890 should fix the existing failure.

foolip:
But more failures will remain :(

mcatanzaro:
> There are flake8 errors specific to this PR. Please fix those before merging.

Hi @jgraham, where can I see these failures?

foolip:
@mcatanzaro flak8 is run as part of "tools/ unittests (Python 2)" and "tools/ unittests (Python 3)" on Travis. I've filed https://github.com/web-platform-tests/wpt/issues/14893 about this not being ideal.

mcatanzaro:
I think flake8 is fixed. There are other failures on travis, but again they look unrelated.

foolip:
Yep, what remains is because of https://github.com/web-platform-tests/wpt/issues/14179 and https://github.com/web-platform-tests/wpt/issues/14894, so I'll merge this now.

PR #14855, opened 2019-01-15
Move allowpaymentrequest attribute tests to HTML

marcoscaceres:
@zcorpan, if you recall, we discussed moving these tests during TPAC. IIRC, we had concluded that the test were not 100% right, because of some under-defined things in HTML. 

Step 1 might just be to merge these, then we can look at fixing them up.


zcorpan (approved):

zcorpan:
I think the "needs changes" thing we discussed was this:

The tests expect the allow* flag to be set once when an iframe with a src attribute is added to the document. But the spec probably sets it twice -- when the browsing context is created, and then when it is navigated. Either the tests or the spec need changes.

zcorpan:
cc @bzbarsky

annevk:
It might make sense to move these to Feature Policy instead since that effectively defines this attribute these days in https://w3c.github.io/webappsec-feature-policy/#iframe-allowpaymentrequest-attribute.

marcoscaceres:
Good suggestion @annevk. I was a bit confused about the redefinition of `allowpaymentrequest` in Feature Policy, as I thought it was more formally defined in HTML (I guess it still is?). 

The other nice thing is that Feature Policy defines how the "payment" feature and the `allowpaymentrequest` attribute rules are supposed to play together.

annevk:
HTML defers to Feature Policy for some months now I think.

marcoscaceres:
Ok, I see. I made the following changes to Payment Request API: https://github.com/w3c/payment-request/pull/822 . Referencing Feature Policy seems most accurate now, as it defines the behavior. However, happy to make any suggested changes.

marcoscaceres:
`allowpaymentrequest` attribute is no longer a thing.

Issue #14836, opened 2019-01-14
Daily runs of Safari Technology Preview on Azure Pipelines


foolip:
In https://github.com/web-platform-tests/wpt/pull/14752 I attempted to enable full runs of Safari Technology Preview on Azure Pipelines, triggered by the epochs/daily branch.

As described in https://github.com/web-platform-tests/wpt/pull/14752#issuecomment-452928106 it didn't work, so this issue is to continue investigating until fixed.

foolip:
On my own fork (foolip/wpt) I tried pushing the current epochs/daily branch (commit 9d21b523b4959f402769d58248e4587fda7f39ff) and that did correctly trigger a build:
https://dev.azure.com/foolip/wpt/_build/results?buildId=343

I then pushed the same commit to the master branch. That *doesn't* cause another Azure Pipelines "CI build" to start, which is consistent with https://github.com/web-platform-tests/wpt/pull/14752#issuecomment-452928106 and suggests that the order of branches updating is the key.

I do see webhooks for both push events being delivered to https://dev.azure.com/foolip/_apis/public/hooks/externalEvents, but these aren't the webhooks that should be used, and I don't see the same installed for this repo.

@lukebjerring do you know of a way to check what webhooks for creating check suites were delivered to a GitHub app?

foolip:
I found it: https://github.com/organizations/web-platform-tests/settings/apps/wpt-fyi-status-check/advanced

But I don't think I can see the same for the Azure Pipelines app, since it isn't our own. I can't tell from the webhook delivery logs if two checks are created when one branch follow another, because there are too many. @lukebjerring do you know, or can you tell from logs?

lukebjerring:
So, as discuss offline: exactly one check suite gets created _when the commit is pushed_. Updating a ref (or a branch, or anything else) to include/set HEAD to the commit doesn't trigger a new check suite.

I think we need to move this pipeline into a daily-triggered pipeline, and tweak it to sync to the SHA provided by the announcer (https://wpt.fyi/api/revisions/epochs).

foolip:
Azure Pipelines does support cron jobs, I could try that.

foolip:
I added a cron job for 14:45 UTC which was triggered:
https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=3626

Unfortunately it didn't work because of https://github.com/web-platform-tests/wpt/issues/14860.

[...]

PR #14859, opened 2019-01-15
Use the checks API with taskcluster jobs

jgraham:


foolip:
@lukebjerring this will require rejigging the results receiver, right?

@jgraham do you know what the triggers for Taskcluster are with this integration, i.e. if this will cause epochs/* branches to stop working?

jgraham:
AFAIK it doesn't change anything about the integration other than where the results go. But @owlishDeveloper is the expert here and should confirm.

foolip:
If it doesn't change anything, it suggests that Taskcluster is using the Checks API in an unconventional way. At least when GitHub creates the check suites (this is configurable) GitHub will send a webhook notifying about this and I think this event is what GitHub intends to be used to trigger check runs. If push events (or something) is also used, then epochs/* will continue to work, but check runs will presumably be overwritten.

Not sure if any of this is a problem, just want to make sure we're not taken by surprise if something like https://github.com/web-platform-tests/wpt/issues/14836#issuecomment-453953098 happens.

jgraham:
Per @owlishDeveloper on irc, enabling checks on GitHub doesn't affect what events we respond to (which makes sense because that's exposed directly in the `.taskcluster.yml` file). So I don't think anything will break by enabling this. Presumably tc looks up the check for the commit later on. Which seems at least as reasonable as looking for the `check_suite` event and seems to be recommended in the docs for cases like ours.

lukebjerring:
Shouldn't be difficult at all, we already have working code for the statuses API and example code for azure.

foolip:
@lukebjerring should I file an issue for that? This PR ought not be filed until that's done, but making the switch is also not particularly urgent I think.

lukebjerring:
Sure, Feel free.

lukebjerring:
Sure, Feel free.

Hexcles:
FWIW, the runs did end up on staging: https://staging.wpt.fyi/results/?sha=f7794b4158fe42dde511108112c4df4cae0ba0fd

so it seems like we don't even need to change the event our webhook is listening to ?

foolip:
@Hexcles this PR hasn't been merged :)

Hexcles:
@foolip the runs I pointed at came from the Taskcluster runs of this exact PR, so I assumed it'd use the Checks API.

foolip:
Ah, that makes sense :)

foolip (approved):
Let's try it!

foolip:
@jgraham do avoid blocking all new PRs I made the "Taskcluster (pull_request)" check not required. wpt-firefox-nightly-stability and wpt-chrome-dev-stability will need to be made blocking instead, but doing it right away will block all *old* PRs instead. Is a gap of testing for a week or so acceptable, or should we revert this PR and reconsider?

foolip:
Relanding this is blocked on https://github.com/web-platform-tests/wpt.fyi/issues/1182.

@jgraham do you think we can request check names to be configurable? Changing which checks are required results in a painful transition, so getting them right the first time would be very good.

jgraham:
The names are configurable I think? They are coming from the `.taskcluster.yml` file.

foolip:
Oh, I assumed that name had to be identifier-like, could we make it something like "Firefox Nightly stability" then? @lukebjerring @Hexcles that would mess with logic used to get results into the results receiver I guess?

@lukebjerring note also Taskcluster splitting tasks into individual checks, which you thought was a bad idea for Azure Pipelines, in case you think it'll cause trouble here as well.

lukebjerring:
Well, revised opinion: Each task being its own check run actually is an advantage (can react to individual results faster - see https://github.com/web-platform-tests/wpt.fyi/issues/730).

https://github.com/web-platform-tests/wpt.fyi/pull/1183 covers reacting.

foolip:
Cool :)

Are you having to bake the names of the checks into wpt.fyi? What do you think about the naming: human readable or not, prefixed with "Taskcluter -" or not?

lukebjerring:
We rely on _specific_ names, but those specific names can be changed (as long as they are deterministic). The GitHub UI groups by CheckRun name, so not-prefixed is fine.

foolip:
Would "Sta

foolip:
Right now everything starts with "wpt-" so it's grouped together. The names I was thinking of were "stability - Chrome" or something like that, grouping things that are likely fail together. Sound OK?

foolip:
And doing that wouldn't group all Taskcluster things together.

PR #14861, opened 2019-01-15
Update MathML spec link

foolip:
https://w3c.github.io/mathml/ is 404.


fred-wang (approved):
Thanks. Hopefully we will have an official document on w3c.github.io soon

Issue #14860, opened 2019-01-15
Azure Pipelines affected tests jobs are being canceled


zcorpan:
In https://github.com/web-platform-tests/wpt/pull/14824 I made a [commit](https://github.com/web-platform-tests/wpt/pull/14824/commits/af40482c9f2c1f6105658a3001bc523f8a05bc87) that failed [the Azure Pipelines build](https://github.com/web-platform-tests/wpt/runs/50120087).

I asked @foolip for advice on irc:

> foolip zcorpan: I've never seen that before, seems like agents can get restarted while running a job. Can you point John Jansen at it to ask?

@thejohnjansen can you take a look? I suppose restarting the job could work but I wanted to make sure this gets due attention if something needs fixing.

foolip:
I've found this is happening more widely and wasn't just an infra restart or similar. https://github.com/web-platform-tests/wpt/pull/14850 is the earliest PR (in creation order) that I can see is affected, but not every PR since then is affected. https://github.com/web-platform-tests/wpt/pull/14863 is the most recent.

When it fails, it's in the "Update manifest" step. https://gist.github.com/foolip/edb0f8dd21738b0d890a15a5e6afc1d0 has the log for one failure.

I have seen these manifest failures on Azure Pipelines before, but then it's fallen back to generating the manifest. I suspect that the machines are exceeding the GitHub API quote for unauthenticated requests. How that can lead to the job being canceled I don't know, perhaps there is an exit code that Azure Pipelines treats as a request to cancel?

@thejohnjansen @mustjab any pointers appreciated.

foolip:
@jgraham have you ever observed a GitHub API quota issues with this code on Taskcluster, either for upstream wpt or Mozilla's CI?

foolip:
I've sent https://github.com/web-platform-tests/wpt/pull/14865 which should avoid the issue at the cost of making jobs slower.

foolip:
A problem here is that if https://github.com/web-platform-tests/wpt/pull/14865 is merged and it takes a certain amount of jobs to exceed the quota, then it would be very hard to know when it's safe to revert it.

foolip:
[...]

PR #14865, opened 2019-01-15
Use `./wpt manifest --no-download` to sidestep issue #14860

foolip:


jgraham:
I don't see any evidence that the problem comes from downloading the manifest from GitHub; in at least one of the previous jobs the manifest download seemed to complete, and this PR fails in the same way.

foolip:
It did happen in the logs (https://github.com/web-platform-tests/wpt/issues/14860#issuecomment-454405334) after the manifest download failed, but given that it still fails I suspect it's really an OOM issue.

Issue #14645, opened 2018-12-21
`./wpt run --verify` claims flakiness where there are duplicate test names


foolip:
https://wpt.fyi/results/css/geometry/interfaces.worker.html?label=stable&aligned shows that the test has a duplicate test name "Test default toJSON operation of toJSON object" in Edge and Firefox. That it is implementation-dependent is surprising and wrong, but all that matters for this issue is that there are duplicate test names.

https://taskcluster-artifacts.net/BX_K1_06R7OU-LiJjKyYmQ/0/public/logs/live_backing.log for https://github.com/web-platform-tests/wpt/pull/14629 claimed that the test was flaky:

|                    Test                   |                                                     Subtest                                                      |            Results            |                                                              Messages                                                              |
|-------------------------------------------|------------------------------------------------------------------------------------------------------------------|-------------------------------|------------------------------------------------------------------------------------------------------------------------------------|
| `/css/geometry/interfaces.worker.html`    | `Test default toJSON operation of toJSON object`                                                                 | **FAIL: 60/10**               | `memberHolderObject is undefined`                                                                                                  |

However, FAIL is the only result, and as indicated by "60/10" there's probably a bug in the `./wpt run --verify` code here, where the tests are overcounted and anything other than "10/10" is treated as flakiness.

jgraham:
Having duplicate test names isn't allowed, so in the absence of anything else catching this, the flakiness check is the best place (it's a form of flakiness because infrastructure will confuse the results of multiple tests, so even if it's not flaky now it's likely to become so in the future and wrong in any case).

foolip:
Alright, let's make this issue about two things:
- some else blocking PRs on harness errors, probably after first taking an inventory of harness errors
- requiring more than one unique result to consider something flaky in `./wpt run --verify`

Letting the first block the second seems fine.

foolip:
https://github.com/web-platform-tests/wpt/pull/14841 was affected by this.

foolip:
@Ms2ger I guess https://github.com/web-platform-tests/wpt/pull/14866 resolved this in the only case we knew it happened, or do you know of others?

Ms2ger:
I think this issue isn't about fixing the existing duplicated test names, per https://github.com/web-platform-tests/wpt/issues/14645#issuecomment-449445728?

foolip:
@tabatkins ran into this as well and asked me about it.

[...]

Issue #14254, opened 2018-11-27
Intermittent "Unrecognized type EventHandlerNonNull" in html/dom/interfaces.https.html


Ms2ger:
I've seen this in https://github.com/web-platform-tests/wpt/pull/14252, but also before that. I tried to reproduce at the time, but failed.

foolip:
This happened in https://github.com/web-platform-tests/wpt/pull/14954. From [the logs](https://taskcluster-artifacts.net/Nr8maQq6QIKpx_Au_ImTdQ/0/public/logs/live_backing.log):

|                             Test                             |                                     Subtest                                      |          Results           |                 Messages                |
|--------------------------------------------------------------|----------------------------------------------------------------------------------|----------------------------|-----------------------------------------|
| `/html/dom/interfaces.https.html?include=(Document\|Window)` | `Window interface: window must inherit property "onunload" with the proper type` | **FAIL: 9/10, PASS: 1/10** | `Unrecognized type EventHandlerNonNull` |

foolip:
This happened in https://github.com/web-platform-tests/wpt/pull/15123 too.

foolip:
@Ms2ger will https://github.com/web-platform-tests/wpt/pull/14866 fix this, or only partly?

foolip:
Oh, disregard, you were saying this was the remaining Taskcluster issue. I wanted to poke at https://github.com/web-platform-tests/wpt/issues/14645.

foolip:
https://github.com/web-platform-tests/wpt/pull/12804 was affected by this.

foolip:
https://github.com/web-platform-tests/wpt/pull/15896 too.

foolip:
This happens when SVG.idl is updated as well: https://github.com/web-platform-tests/wpt/pull/16765

foolip:
@jgraham @lukebjerring this is the issue we mentioned in the meeting just now.

lukebjerring:
This seems to be flakiness in the order in which the dependencies are resolved during `add_dependency_idls`. 

[...]

PR #14866, opened 2019-01-15
Improve the test names of tests for toJSON behavior.

Ms2ger:
In particular, this avoids duplicate test names when testing with multiple objects (ref #14645).

### DIFF
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -2309,7 +2309,7 @@ IdlInterface.prototype.do_member_operation_asserts = function(memberHolderObject
     }
 }
 
-IdlInterface.prototype.test_to_json_operation = function(memberHolderObject, member) {
+IdlInterface.prototype.test_to_json_operation = function(desc, memberHolderObject, member) {
     var instanceName = memberHolderObject && memberHolderObject.constructor.name
         || member.name + " object";
     if (member.has_extended_attribute("Default")) {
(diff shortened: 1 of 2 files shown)


Ms2ger:
Only issue os https://github.com/web-platform-tests/wpt/issues/14254

resources/idlharness.js
```diff
@@ -2325,12 +2325,12 @@ IdlInterface.prototype.test_to_json_operation = function(memberHolderObject, mem
                 this.array.assert_type_is(json[k], type);
                 delete json[k];
             }, this);
-        }.bind(this), "Test default toJSON operation of " + instanceName);
+        }.bind(this), this.name + " interface: Test default toJSON operation on " + desc);
```
foolip:
```suggestion
        }.bind(this), this.name + " interface: default toJSON operation on " + desc);
```

resources/idlharness.js
```diff
@@ -2325,12 +2325,12 @@ IdlInterface.prototype.test_to_json_operation = function(memberHolderObject, mem
...
     } else {
         subsetTestByKey(this.name, test, function() {
             assert_true(this.array.is_json_type(member.idlType), JSON.stringify(member.idlType) + " is not an appropriate return value for the toJSON operation of " + instanceName);
             this.array.assert_type_is(memberHolderObject.toJSON(), member.idlType);
-        }.bind(this), "Test toJSON operation of " + instanceName);
+        }.bind(this), this.name + " interface: Test toJSON operation on " + desc);
```
foolip:
```suggestion
        }.bind(this), this.name + " interface: toJSON operation on " + desc);
```

foolip (approved):
Suggest also removing "Test" from the names, this appears to have been the only subtests with that in the name.

foolip (approved):

PR #14867, opened 2019-01-15
Ensure that all PerformanceObserver.supportedEntryTypes are observable

cvazac:
From what I can tell, this PR reveals some bugs in the Chrome impl. 

From a worker context, `navigation` and `paint` types should not be observable. 

Also, because `longtask`s [aren't exposed on Worker](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/timing/performance_long_task_timing.idl?l=6), they shouldn't be observable from that context either.

cc @npm1 @yoavweiss


performance-timeline/supportedEntryTypes.any.js
```diff
@@ -1,12 +1,81 @@
-test(() => {
+function forcePerformanceEntry(entryType) {
+  switch (entryType) {
+    case 'resource':
+      // Use `self` for Workers
+      fetch(self.location.href + "?" + Math.random());
+      break;
+
+    case 'longtask':
```
npm1: Longtasks are not observable in workers

npm1 (commented):
Looks good! Does anyone have thoughts on whether the specific entry types should be tested on a performance-timeline test or if they should be tested separately (paint-timing folder for one, etc.)?

cvazac:
> Looks good! Does anyone have thoughts on whether the specific entry types should be tested on a performance-timeline test or if they should be tested separately (paint-timing folder for one, etc.)?

I could go either way, but one of the benefits of baking the tests into performance-timeline: if a browsers adds a new `supportedEntryType`, this test will fail until [this switch](https://github.com/web-platform-tests/wpt/pull/14867/files#diff-88397db494b2e5276a13e2fa289ac429R2) is updated.

igrigorik:
> > Looks good! Does anyone have thoughts on whether the specific entry types should be tested on a performance-timeline test or if they should be tested separately (paint-timing folder for one, etc.)?
> 
> I could go either way, but one of the benefits of baking the tests into performance-timeline: if a browsers adds a new `supportedEntryType`, this test will fail until [this switch](https://github.com/web-platform-tests/wpt/pull/14867/files#diff-88397db494b2e5276a13e2fa289ac429R2) is updated.

FWIW, I think these tests should live in separate test files associated with each spec.

- We should check that `supportedEntry` exists / is exposed in the tests associated with PerfTimeline: that's the only requirement that PerfTimeline enforces in the spec.
- We should check that supportedEntry exposes relevant name in the tests associated with each of the specs: if the browser supports that spec, it must also expose it via supportedEntry.

This aligns with how we specced entryTypes.. We don't enumerate all the types in PerfTimeline spec.

/cc @yoavweiss @toddreifsteck for sanity check

yoavweiss:
What @igrigorik said.
We don't want PerfTimeline tests to fail because browsers haven't implemented one of the specs that uses it. Therefore, it's better to have each "type"'s test to live in the related spec's test directory.

cvazac:
> What @igrigorik said.
> We don't want PerfTimeline tests to fail because browsers haven't implemented one of the specs that uses it. Therefore, it's better to have each "type"'s test to live in the related spec's test directory.

Making the change now, but FWIW, this performance-timeline test won't fail under that case. The logic I've added equates to: verify that every `supportedEntryType` that `PerformanceObserver` claims to support is actually supported.

npm1:
Oh maybe we should not be including things like 'longtask' in supportedEntryTypes for worker, good point. Should we update tests to not use .any.js for cases where it should only be observable on window? I'll update the Chrome implementation.

cvazac:
> Oh maybe we should not be including things like 'longtask' in supportedEntryTypes for worker, good point. Should we update tests to not use .any.js for cases where it should only be observable on window? I'll update the Chrome implementation.

Are you suggesting an exclude-list for workers to ensure that no impl. exposes an entryType that we don't think is possible from workers? `navigation` and `paint` would make sense to exclude, but `longtask` _could_ be exposed on workers in the future.

npm1:
> > Oh maybe we should not be including things like 'longtask' in supportedEntryTypes for worker, good point. Should we update tests to not use .any.js for cases where it should only be observable on window? I'll update the Chrome implementation.
> 
> Are you suggesting an exclude-list for workers to ensure that no impl. exposes an entryType that we don't think is possible from workers? `navigation` and `paint` would make sense to exclude, but `longtask` _could_ be exposed on workers in the future.

Well I was thinking the criteria would be: is it possible to obtain entries for that entry type? In particular, while in the future we might expose longtasks in workers, they are not supported at the moment.

yoavweiss (approved):
LGTM
Maybe there's a way to extract the identical parts of those tests into an external file, but we can look at that as a followup

PR #14873, opened 2019-01-15
Add tentative WPT to assert download has / has not happened in sandbox

xyaoinum:
The plan is to implement "blocking download in sandbox without user gesture", unless user opt-in.
 
The issue was discussed https://github.com/whatwg/html/issues/3236
Pull request for the spec: https://github.com/whatwg/html/pull/4293

General testing idea:
For a network request, we store a token in stash. And after a fixed period of time sufficient for the download to finish, if there are any, we validate the token in the stash to see if the network request was issued and finished successfully.
- In the case of `<a download>` where the decision of download can be made before resource fetching, the server sets the token immediately.
- In the case of navigation to a download, the server will stream a response over 1 seconds and set the token at the end only when the socket has been connected. So it is able to detect any download cancellation while fetching the resource.


xyaoinum:
Decided to use the chromium review process. Close this pull request.

PR #14876, opened 2019-01-16
SVG Interactivity: onwheel attribute

ewilligers:
Test that script specified by the the onwheel attribute is run
in response to mouse wheel events.

https://github.com/w3c/svgwg/issues/627


AmeliaBR:
The test looks fine. My only question is where to put it.

Would it be better to create a sub-folder for all manual tests? (In a full test suite, we probably will have various manual tests for event handlers).

So the new file would be `svg/interact/manual/event-attribute-001-manual.svg` or something similar.

cc @gsnedders

ewilligers:
I have moved the test.

AmeliaBR (approved):

PR #14879, opened 2019-01-16
test POST navigation with Referrer-Policy

JuniorHsu:
test for POST navigation with Referrer-Policy


fetch/origin/no-cors.window.js
```diff
@@ -76,3 +76,65 @@ promise_test(async function () {
...
+    assert_equals(json[1], expectedOrigin);
+  };
+}
+
+function referrerPolicyTestString(referrerPolicy, isCors) {
+  return "Origin header and POST" + (isCors ? "Cross-origin" : "same-origin") +
+         " navigation with Referrer-Policy " + referrerPolicy;
+}
+
+promise_test(referrerPolicyFucnctor("no-referrer", false, "null"),
```
domfarolino: Fucnctor => Functor

fetch/origin/resources/redirect-and-stash.py
```diff
@@ -20,6 +20,13 @@ def main(request, response):
 
     request.server.stash.put(key, origin_list)
 
+    if "referrerPolicy" in request.GET:
+        response.status = 200
+        response.headers.set("content-Type", "text/html")
+        response.headers.set("Referrer-Policy",
+                             request.GET.first("referrerPolicy"))
+        return
```
annevk: This block could be much higher up, right? It might even be good to have this in a separate file to indicate it doesn't really affect all the stash logic.

(I guess with this setup you're implicitly testing the GET scenario, but that really seems too subtle. I'd much rather we test that explicitly as a distinct set of tests from the POST tests or as some kind of permutation upon them.)

fetch/origin/no-cors.window.js
```diff
@@ -76,3 +76,65 @@ promise_test(async function () {
...
+      const rp = "no-referrer";
+      frame.src = origins.HTTP_ORIGIN + redirectPath +
+                  "?referrerPolicy=" + referrerPolicy + "&stash=" + stash;
+      self.addEventListener("message", e => {
+        if (e.data === "loaded") {
+          resolve();
+          frame.remove();
+        }
+      }, { once: true });
+      frame.onload = () => {
```
annevk: There's an interesting thing here in that in Firefox the above message is delivered and the frame element is removed by the time another load event dispatches, but in Chrome/Safari this is not the case and therefore accessing doc.body will fail for cross-origin loads. Setting frame.onload to null addresses this, but please file an issue against whatwg/html or some such on exploring that further as that doesn't seem entirely right.

fetch/origin/no-cors.window.js
```diff
@@ -76,3 +76,65 @@ promise_test(async function () {
   assert_equals(json[0], "null");
   assert_equals(json[1], "null");
 }, "Origin header and POST navigation");
+
+function referrerPolicyFucnctor(referrerPolicy, isCors, expectedOrigin) {
```
annevk: Instead of "isCors", let's make this "crossOrigin" or "isCrossOrigin" or some such as it doesn't really relate to CORS.

fetch/origin/no-cors.window.js
```diff
@@ -76,3 +76,65 @@ promise_test(async function () {
...
+  };
+}
+
+function referrerPolicyTestString(referrerPolicy, isCors) {
+  return "Origin header and POST" + (isCors ? "Cross-origin" : "same-origin") +
+         " navigation with Referrer-Policy " + referrerPolicy;
+}
+
+promise_test(referrerPolicyFucnctor("no-referrer", false, "null"),
+             referrerPolicyTestString("no-referrer", false));
```
annevk: I would consider refactoring this as a list of objects upon which you iterate and then call promise_test to avoid some duplication of data. E.g.,
```js
[
  {
    "policy": "no-referrer",
    "crossOrigin": false,
    "expectedOrigin": "null"
  },
  ...
].forEach(testObj => {
  promise_test(..., ...);
});
```

fetch/origin/no-cors.window.js
```diff
@@ -76,3 +76,65 @@ promise_test(async function () {
...
+
+promise_test(referrerPolicyFucnctor("no-referrer", true, "null"),
+             referrerPolicyTestString("no-referrer", true));
+
+promise_test(referrerPolicyFucnctor("same-origin", false, "http://web-platform.test:8000"),
```
annevk: You cannot hardcode the `http://web-platform.test:8000` value (these are technically variables that can be configured by whoever runs these tests). You could use a string such as "same" and make that to mean `origins.HTTP_ORIGIN` in the test code.

fetch/origin/no-cors.window.js
```diff
@@ -76,3 +76,65 @@ promise_test(async function () {
...
+    const stash = token(),
+          origins = get_host_info(),
+          redirectPath = "/fetch/origin/resources/redirect-and-stash.py";
+
+    let url = (isCors ? origins.HTTP_REMOTE_ORIGIN : origins.HTTP_ORIGIN) +
+              redirectPath + "?stash=" + stash;
+
+    await new Promise(resolve => {
+      const frame = document.createElement("iframe");
+      const rp = "no-referrer";
```
annevk: This constant does not appear to be used.

annevk (commented):
Thanks, this looks pretty good.

I think we should test the remaining referrer policies as well (e.g., `unsafe-url`).

It might also be worth it to include an explicit HTTPS to HTTP test.

It might also be worth testing that if a redirect doesn't cross the origin boundary, the `Origin` header doesn't get nullified.

So I guess what Chrome is doing is that it bases the `Origin` header on the `Referrer` header for these non-CORS cases. In particular that would explain how for an `about:blank` `<iframe>` it ends up without an origin rather than the same origin as the parent. I'm not sure I like that behavior as it doesn't match how we determine the `Origin` header value normally. I guess we could chose to define it that way, but it's rather weird to have that kind of branch. On the other hand, we have that branch already due to taking referrer policy into account, so maybe it's okay.

cc @mikewest

mikewest:
Pulling in @jeisinger and @estark37, who I think have thought about Referrer Policy / `Origin` interactions.

jeisinger:
I read https://fetch.spec.whatwg.org/#http-network-or-cache-fetch as stating that Origin is created from the client's origin, and not from the referrer (which I think makes sense), and I would consider behavior in chromium where we set the origin header from the referrer a bug.

JuniorHsu:
> I read https://fetch.spec.whatwg.org/#http-network-or-cache-fetch as stating that Origin is created from the client's origin, and not from the referrer (which I think makes sense), and I would consider behavior in chromium where we set the origin header from the referrer a bug.

@jeisinger Is that the case for about:blank? i.e., the iframe from about:blank to https://foo.com should result in Origin: https://foo.com

jeisinger:
> > I read https://fetch.spec.whatwg.org/#http-network-or-cache-fetch as stating that Origin is created from the client's origin, and not from the referrer (which I think makes sense), and I would consider behavior in chromium where we set the origin header from the referrer a bug.
> 
> @jeisinger Is that the case for about:blank? i.e., the iframe from about:blank to https://foo.com should result in Origin: https://foo.com

That's my understanding, yes, but I'm a referrer person, not an origin or fetch person :)

jdm:
Continued in #15937.

annevk:
I contacted GitHub about not being able to close this PR.

PR #14887, opened 2019-01-16
DO NOT MERGE: run all CI test jobs

foolip:


foolip:
The failure is now with test_manifest_idempotent. This could be a real bug, and it seems plausible the reason it's happening now is because of changes to the hypothesis package.

foolip:
Reopening to see if there's still more failures.

foolip:
Cool, no failures :)

Issue #14889, opened 2019-01-16
manifest/tests/test_manifest.py is failing on master


foolip:
See https://github.com/web-platform-tests/wpt/pull/14887, it happens both in Travis and Azure Pipelines.

This is probably due to an update of the hypothesis library, which we do not lock down using a requirements.txt file.

foolip:
The version of hypothesis I currently get locally is 4.0.1. https://github.com/HypothesisWorks/hypothesis is the GitHub repo for the package, looking at history now.

PR #14890, opened 2019-01-16
Drop the use of `average_size` with hypothesis.strategies.lists

foolip:
It has been removed and now causes an error:
https://hypothesis.readthedocs.io/en/latest/changes.html#v4-0-0

Fixes https://github.com/web-platform-tests/wpt/issues/14889.

### DIFF
--- a/tools/manifest/tests/test_manifest.py
+++ b/tools/manifest/tests/test_manifest.py
@@ -66,8 +66,7 @@ def sourcefile_strategy(draw):
 
 
 @h.given(hs.lists(sourcefile_strategy(),
-                  min_size=1, average_size=10, max_size=1000,
-                  unique_by=lambda x: x.rel_path))
+                  min_size=1, max_size=1000, unique_by=lambda x: x.rel_path))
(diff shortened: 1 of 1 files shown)


foolip:
It was deprecated in https://hypothesis.readthedocs.io/en/latest/changes.html#v3-51-0. The release notes there also suggest this naive fix is the right one:
> You should simply delete it wherever it was used in your tests, as it no longer has any effect.
>
> In early versions of Hypothesis, the `average_size` argument was treated as a hint about the distribution of examples from a strategy. Subsequent improvements to the conceptual model and the engine for generating and shrinking examples mean it is more effective to simply describe what constitutes a valid example, and let our internals handle the distribution.

Ms2ger (approved):

foolip:
There are remaining failures that also show up in https://github.com/web-platform-tests/wpt/pull/14887 that I didn't pay attention to originally. The early failure due to `average_size` may also have been masking a later failure.

I'm pretty confident this is not introducing new failures and is fixing an existing one, so I'll admin merge this and keep filing issues for what remains.

PR #14907, opened 2019-01-16
Split fetch CORP tests as HTTP and HTTPS.

youennf:
Add fetch.https.any.js in addition to fetch.any.js.
Make fetch.any.js not run in service worker.
Make fetch.https.any.js run also in service worker.
Update a test in fetch.any.js for the "same-site HTTP -> HTTPS" case.


anforowicz:
Drive-by comment... feel free to ignore :-)

Thank you very much for putting together the PR.  I can confirm that the tests pass against my Chromium WIP CL @ https://chromium-review.googlesource.com/c/chromium/src/+/1416813.

And separate thanks for having the luxury of implementing the feature with already existing test coverage .  One benefit is that it prevent me from forgetting to support redirects (the last testcase in fetch.any.js fails).  Thanks!

fetch/cross-origin-resource-policy/fetch.any.js
```diff
@@ -40,7 +40,7 @@ promise_test((test) => {
 
 promise_test((test) => {
     const remoteURL = httpsBaseURL + "resources/hello.py?corp=same-site";
-    return fetch(remoteURL, { mode: "no-cors" });
+    return promise_rejects(test, new TypeError, fetch(remoteURL, { mode: "no-cors" }));
 }, "Cross-origin no-cors fetch to a same-site URL with a 'Cross-Origin-Resource-Policy: same-site' response header.");
```
annevk: Please make this say "cross-scheme same-site" or some such. I forgot to read the PR description and it took me a long time to figure out what was going on.

fetch/cross-origin-resource-policy/fetch.https.any.js
```diff
@@ -0,0 +1,56 @@
+// META: global=window,worker
+// META: script=/common/get-host-info.sub.js
+
+const host = get_host_info();
+const path = "/fetch/cross-origin-resource-policy/";
+const localBaseURL = host.HTTPS_ORIGIN + path;
+const notSameSiteBaseURL = host.HTTPS_NOTSAMESITE_ORIGIN + path;
+const httpBaseURL = host.HTTP_ORIGIN + path;
```
annevk: This appears unused, remove it?

annevk (approved):
Looks great, two minor nits.

fetch/cross-origin-resource-policy/fetch.any.js
youennf: OK, added Cross-scheme (HTTP->HTTPS)

youennf:
Taskcluster is failing for both Firefox and Chrome.
Looking at the logs, there does not seem to be anything suspicious related to the PR.
@foolip, any idea what this might be?

annevk:
I think the problem is the change to `common/get-host-info.sub.js` which presumably triggers runs of lots of tests, causing the whole thing to time out.

PR #14924, opened 2019-01-17
Bug #14179: Disable test_run_firefox for flakiness

gsnedders:
#14179


jgraham (approved):

jgraham (changes requested):
New plan: delete this test and the corresponding Chrome test since they are entirely covered by the wptrunner_infrastructure tests.

jgraham (approved):

Hexcles (approved):
Glad to see they are gone. I also ran into flakiness caused by these tests quite a few times.

PR #14936, opened 2019-01-18
Add some tests for the named property object.

Ms2ger:


Ms2ger:
https://github.com/heycam/webidl/pull/626

html/browsers/the-window-object/named-access-on-the-window-object/cross-global-npo.html
```diff
@@ -0,0 +1,37 @@
...
+      // [value, is own property]
+      [element, false],
+      [element, false],
+      [element, true],
+      [undefined, false],
+      [undefined, false],
+    ];
+    for (var object = win; object; object = Object.getPrototypeOf(object)) {
+      var expected = expectedValues.shift();
+      assert_equals(object[name], expected[0]);
```
foolip: Can you add descriptions to the `assert_equals` so that upon failure it's clear which failed?

foolip (approved):
LGTM with nit

foolip:
I missed that my comments had been resolved until now. Oops.

PR #14940, opened 2019-01-18
WebKit export of https://bugs.webkit.org/show_bug.cgi?id=193553

alijuma:
* FetchResponse::url should return the empty string for tainted responses


youennf (approved):
This test should probably use get-host-info.sub.js instead of using sub itself

youennf:
> This test should probably use get-host-info.sub.js instead of using sub itself

Nevermind, not needed

PR #14948, opened 2019-01-19
[css-logical] Inheritance, initial values

ewilligers:
Test that CSS Logical Properties and Values properties do not inherit.

https://www.w3.org/TR/css-logical-1/#property-index

### DIFF
--- a/css/css-logical/inheritance.html
+++ b/css/css-logical/inheritance.html
@@ -1,0 +1,95 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Inheritance of CSS Logical Properties and Values properties</title>
+<link rel="help" href="https://www.w3.org/TR/css-logical-1/#property-index">
+<meta name="assert" content="Properties do not inherit.">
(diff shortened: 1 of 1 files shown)


ewilligers:
Tests pass on Blink/Edge/Safari.

Firefox fails 4 tests:-

> Fail	Property border-block-end-width has initial value 2.5px	assert_equals: expected "2.5px" but got "3px"
> Fail	Property border-block-start-width has initial value 2.5px	assert_equals: expected "2.5px" but got "3px"
> Fail	Property border-inline-end-width has initial value 2.5px	assert_equals: expected "2.5px" but got "3px"
> Fail	Property border-inline-start-width has initial value 2.5px	assert_equals: expected "2.5px" but got "3px"

I don't understand why the inline-size test passes when these tests fail.

ewilligers:
@fantasai please review

ewilligers:
@atanassov please review

ewilligers:
@tabatkins Would you be happy to be a [test reviewer](https://github.com/web-platform-tests/wpt/blob/master/css/css-logical/META.yml) for css-logical?

tabatkins (approved):

PR #14979, opened 2019-01-21
[css-ui] Test for appearance/-webkit-appearance and initial value

zcorpan:


jugglinmike (approved):
I'd rather see this expressed without a loop, but it's not worth blocking on. I'm a newcomer to this specification, so I'll defer to you on when this can be merged.

Issue #14938, opened 2019-01-18
Use a window manager when running test jobs via xvfb 


whimboo:
I investigated some WebDriver test failures for Firefox as shown on [wpt.fyi](https://wpt.fyi/results/webdriver/tests?label=master&label=experimental) the last days, and noticed that the window manipulation tests (fullscreen, maximize, minimize, set window rect) mostly fail for me also locally when I use `xvfb` and having no window manager installed. By that config at least Firefox cannot correctly set those modes. Related here was also an earlier discussion at https://github.com/web-platform-tests/wpt/issues/13005#issuecomment-438199267. 

Installing and using a window manager like `fluxbox` in combination with `xvfb` made it work for me.

As @jgraham pointed me out now the `wpt` setup doesn't make use of a window manager: https://github.com/web-platform-tests/wpt/blob/master/tools/ci/start.sh#L30

So it would be great to get this command updated to also include the usage of a simple window manager like fluxbox.

jgraham:
I think there are two parts here

1) Check what's installed in the docker image. If there isn't a WM installed we will need to install one, and rebuild the docker file, so hopefully there is?
2) Write the code in `tools/ci/start.sh` to actually start the WM. https://stackoverflow.com/questions/41298489/xvfb-start-or-attach-a-window-manager-to-xvfb might be useful for the latter.

foolip:
@mcatanzaro this seems like it could be relevant for your work on Epiphany. Will you need a real window manager there as well?

foolip:
@whimboo since you've started work on this I marked it as priority:roadmap and assigned to you. Hope that's OK. It means that it someone will ask about progress if it's inactive for 60 days.

whimboo:
Sure, once we agreed on a window manager we can continue here. I only would need someone to build and upload the docker image because I don't have permissions.

mcatanzaro:
> @mcatanzaro this seems like it could be relevant for your work on Epiphany. Will you need a real window manager there as well?

I just haven't gotten around to trying to make the tests work headless yet. It's high on my TODO. I'm just always a little swamped.

I would expect that using a real window manager might lead to more accurate test results in all browsers, so it seems desirable, but when running tests locally I'd want them to run headless to avoid disruption. Both ways seem valuable.

whimboo:
> I just haven't gotten around to trying to make the tests work headless yet. It's high on my TODO. I'm just always a little swamped.

[...]

PR #14981, opened 2019-01-21
Use fluxbox as window manager for xvfb in CI (14938)

whimboo:
This is an attempt to fix issue #14938 by trying to use fluxbox as window manager. As is it might fail because the package is most likely not installed in the docker image by default.

### DIFF
--- a/tools/ci/start.sh
+++ b/tools/ci/start.sh
@@ -28,3 +28,4 @@ then
 fi
 
 sudo Xvfb $DISPLAY -screen 0 ${SCREEN_WIDTH}x${SCREEN_HEIGHT}x${SCREEN_DEPTH} &
+sudo fluxbox -display $DISPLAY &
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -61,7 +61,7 @@ tasks:
               owner: ${event.pusher.email}
               source: ${event.repository.url}
             payload:
-              image: harjgam/web-platform-tests:0.25
+              image: harjgam/web-platform-tests:0.26
               maxRunTime: 7200
               artifacts:
                 public/results:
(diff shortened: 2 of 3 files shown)


whimboo:
Here the expected failure: `tools/ci/start.sh: line 30: fluxbox: command not found`

@jgraham I just pushed an update which includes the installation of `fluxbox` in the docker image. Would you mind checking that, and build/upload it if it looks fine to you? Thanks.

jugglinmike:
> I wonder if anyone has expertise in which lightweight window manager would be good to use. When searching in combination with `xvfb` I usually find `fluxbox` as the suggested one.

@paultag any chance you have a minute to advise?

gsnedders:
TC seems to have failed to fetch the git repo :(

whimboo:
The second last push seems to work now. I pushed another debug commit with changes to wdspec tests which need a window manager. If that works we will be good.

whimboo:
Test results for wdspec tests with Firefox look amazing now. There is only a single failure for stability tests: https://tools.taskcluster.net/groups/V1tYKZZZSMW-aXT9qbd8WQ/tasks/ETc9G-HqS56GgCZh1hHJSQ/runs/0/logs/public%2Flogs%2Flive.log#L2532

I'm going to refresh the PR now to make it ready for review.

tools/ci/start.sh
```diff
@@ -28,3 +28,4 @@ then
 fi
 
 sudo Xvfb $DISPLAY -screen 0 ${SCREEN_WIDTH}x${SCREEN_HEIGHT}x${SCREEN_DEPTH} &
+sudo fluxbox &
```
jgraham: I think this should be `fluxbox -display $DISPLAY` although it's possible that with one display only running we happen to get the right one.

whimboo: I had this before and reverted it. If you want I can definitely add it back.

jgraham (approved):

PR #14993, opened 2019-01-22
[css-ui] Test appearance: button on drop-down select

zcorpan:
Follows https://github.com/w3c/csswg-drafts/pull/3543


css/css-ui/appearance-button-drop-down-select-001.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Basic User Interface Test: appearance: button on drop-down select</title>
+<link rel="help" href="https://drafts.csswg.org/css-ui-4/#appearance-switching">
+<meta name="assert" content="The button value works on a drop-down select.">
+<link rel="mismatch" href="drop-down-select-ref.html">
```
frivoal: There's nothing that requires that the look and feel of <select> and the look and feel of <button> be the same, so the reference file should contain an actual button element, otherwise the test is not doing what it's claiming.

css/css-ui/appearance-button-drop-down-select-001.html
```diff
@@ -0,0 +1,11 @@
...
+<link rel="mismatch" href="drop-down-select-appearance-none-ref.html">
+<style>
+ select { -webkit-appearance: button; }
```
frivoal: I am of two minds about this. The spec does mandate that the -webkit- prefix be supported, so it's not wrong to have tests with it, but we should test the unprefixed name as well. One approach would be to duplicate all the tests, so that we have a prefixed and an unprefixed variant of each. We could also have a lot of tests covering one of the names and a few checking that the other name works as well, but in that case we usually go with the unprefixed name for most tests.

Either way, I'm OK with using testing the prefixed name, as long as we're not **only** testing the prefixed name.

frivoal: (this comment applies to all tests in this PR)

css/css-ui/appearance-button-list-box-select-001.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Basic User Interface Test: appearance: button on list box select (multiple)</title>
+<link rel="help" href="https://drafts.csswg.org/css-ui-4/#appearance-switching">
+<meta name="assert" content="The button value acts like auto on a list box select.">
+<link rel="match" href="list-box-select-ref.html">
```
frivoal: That reference does not contain `appearance: auto`, and instead relies on the UA stylesheet having that value. Which in practice will be the case, but isn't yet (as far as I know) in any spec, and is a separate thing to test anyway. I suggest changing to a reference file that explicitly uses `appearance: auto`

css/css-ui/appearance-button-list-box-select-002.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Basic User Interface Test: appearance: button on list box select (size)</title>
+<link rel="help" href="https://drafts.csswg.org/css-ui-4/#appearance-switching">
+<meta name="assert" content="The button value acts like auto on a list box select.">
+<link rel="match" href="list-box-select-ref.html">
```
frivoal: Same comment as https://github.com/web-platform-tests/wpt/pull/14993/files#r250017177

css/css-ui/appearance-button-not-form-controls-001-ref.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Reference for CSS Basic User Interface Test: appearance: button on elements that are not form controls</title>
+<style>
+ #container { width: 500px; }
+</style>
+<div id="container">
+	<a>a</a>
```
frivoal: All elements in this container should be replaced with <button>, otherwise this test is a no-op.

css/css-ui/appearance-button-other-form-controls-001-ref.html
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Reference for CSS Basic User Interface Test: appearance: button on form controls other than select</title>
+<style>
```
frivoal: As per https://github.com/web-platform-tests/wpt/pull/14993/files#r250017177, please explicitly add `appearance: auto` to this ref file, since it isn't the initial value of the property. Otherwise, we're testing the content of the UA stylesheet, which is a valid, but separate, thing to do.

frivoal (changes requested):
The intent of these tests is good, but there are a few points that need to be fixed before merging. See individual comments.

css/css-ui/appearance-button-drop-down-select-001.html
zcorpan: Note that the reference is "mismatch", i.e. they should not match. An actual button might have different size than a select with appearance: button (e.g. due to different default border width), so it doesn't really work to have a "match" reference here I think. (Setting the border with author CSS disables the appearance, so we can't do that, either.)

css/css-ui/appearance-button-drop-down-select-001.html
zcorpan: A few tests test both (like JS tests where that is easy). If we want to duplicate the tests I think I would write a build script to generate them, though I wanted to keep things simple. :)

css/css-ui/appearance-button-not-form-controls-001.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Basic User Interface Test: appearance: button on elements that are not form controls</title>
+<link rel="help" href="https://drafts.csswg.org/css-ui-4/#appearance-switching">
+<meta name="assert" content="button should work on elements in general.">
+<link rel="match" href="appearance-button-other-form-controls-001-ref.html">
```
zcorpan: This should be "mismatch"

css/css-ui/appearance-button-not-form-controls-001-ref.html
zcorpan: I intended for the test to use "mismatch". Using a button element will have different sizing so doesn't work.

css/css-ui/appearance-button-drop-down-select-001.html
zcorpan: I wrote the script in https://github.com/web-platform-tests/wpt/pull/14996 , when that is merged I can rebase the remaining PRs and fix as appropriate.

zcorpan: Actually I fixed this PR without rebasing (I just ran the script without committing the script file itself here).

zcorpan:
Addressed comments, thanks

zcorpan:
Taskcluster [`wpt-chrome-dev-results`](https://tools.taskcluster.net/groups/VsSGc0aFQla_JfXmtwBspg/tasks/WwKkEMfLRuSs9ceb_9_g7w/details) and [`wpt-firefox-nightly-results`](https://tools.taskcluster.net/groups/VsSGc0aFQla_JfXmtwBspg/tasks/D1MxcRiNQqWuWLhucfyRoQ/details) failed. Not sure why. For now I'll close and reopen to rerun bots.

zcorpan:
@frivoal can you take another look? Thanks!

css/css-ui/appearance-button-not-form-controls-001-ref.html
frivoal: Ah, right, I see. But then the test is checking that at least one of these elements can be made to look like a button, not that all of them can. If you want to do this with a mismatch (which makes sense, given your comment above), you need to break this test case (and its ref) into a separate file for each element.

frivoal:
@zcorpan It's better now, but there are still a couple of things to fix. See
https://github.com/web-platform-tests/wpt/pull/14993#discussion_r250018367 (though that's not critical) and more importantly https://github.com/web-platform-tests/wpt/pull/14993#discussion_r361847836

zcorpan:
This doesn't match what's in the spec now. `appearance: button` on drop-down select is covered in /css/css-ui/appearance-button-001.html

PR #14994, opened 2019-01-22
[css-ui] Test appearance: textfield

zcorpan:
Follows https://github.com/w3c/csswg-drafts/pull/3544


jugglinmike:
If I understand your use of `appearance: none` correctly, it seems like a neat pattern. It took me a few minutes to work out, though, which makes me think it might be susceptible to removal by future maintainers. Is it a common pattern in CSS testing?

If not, a comment would go a long way. Something like, "For most elements, the expected behavior identical to the initial value. Initially set the property value to `none` to verify that this is value is being interpreted (instead of simply ignored)."

zcorpan:
I don't know if it's common in the css testsuite to rely on this particular behavior, but regardless, a comment seems like a good idea.

I think it could be stated more simply

```
/* If the value being tested is not supported, then 'none' is used instead, which is intended to fail the test. */
```

Also I think this test needs

* rebasing
* change `-webkit-appearance` to `appearance`
* run the build script to generate the `-webkit-appearance` variant of this test

jugglinmike:
I followed up with @zcorpan's suggestions. That makes me a co-author, so I'd like further review before merging. @frivoal @mrego @plinss @tantek would one of you mind giving this a look?

([The prior version of this branch is available here.](https://github.com/web-platform-tests/wpt/compare/zcorpan/appearance-textfield-orig))

zcorpan:
The change LGTM, @jugglinmike . I can't mark as reviewed, but I can admin-merge.

jugglinmike:
Thanks!

PR #14996, opened 2019-01-22
[css-ui] Test appearance <compat> values and 'auto'

zcorpan:
Follows
https://github.com/w3c/csswg-drafts/pull/3545
https://github.com/w3c/csswg-drafts/pull/3546
https://github.com/w3c/csswg-drafts/pull/3574


css/css-ui/appearance-auto-001.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Basic User Interface Test: appearance: auto</title>
+<link rel="help" href="https://drafts.csswg.org/css-ui-4/#appearance-switching">
+<meta name="assert" content="auto is supported.">
+<link rel="match" href="appearance-auto-ref.html">
+<style>
+ #container { width: 500px; }
+ #container > * { -webkit-appearance: none; -webkit-appearance: auto; }
```
frivoal: Same comment as I made in [an other pull request](https://github.com/web-platform-tests/wpt/pull/14993#discussion_r250016683): the prefixed version is mandated by spec, so it's ok to test with it, but we should not **only** test with it.

css/css-ui/appearance-auto-ref.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Reference: appearance: auto</title>
+<style>
```
frivoal: That reference does not contain `appearance: auto`, and instead relies on the UA stylesheet having that value (or some other value with the same effect). Which in practice will be the case, but isn't yet (as far as I know) in any spec, and is a separate thing to test anyway. Please explicitly add `appearance: auto` to this ref file, since it isn't the initial value of the property. Otherwise, we're testing the content of the UA stylesheet, which is a valid, but separate, thing to do.

frivoal (dismissed):
Mostly good, just a couple of tweaks. See individual comments.

css/css-ui/appearance-auto-001.html
zcorpan: I wrote a python script to generate the -webkit- prefixed reftests.

zcorpan:
Thanks, addressed your comments.

css/css-ui/tools/appearance-build-webkit-reftests.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python3
```
foolip: Can you run this script in https://github.com/web-platform-tests/wpt/blob/master/update-built-tests.sh to make sure committed output doesn't diverge from script output?

css/css-ui/tools/appearance-build-webkit-reftests.py
```diff
@@ -0,0 +1,29 @@
...
+re_testname = re.compile(r"^appearance-.+\d\d\d\.html$")
+re_link_match = re.compile(r'<link rel="(mis)?match"')
+
+files = {}
+
+parentdir = os.path.abspath(os.pardir)
+assert parentdir.endswith('/css-ui')
+
+for filename in os.listdir(parentdir):
+    path = parentdir + '/' + filename
```
foolip:
```suggestion
    path = os.path.join(parentdir, filename)
```

css/css-ui/tools/appearance-build-webkit-reftests.py
```diff
@@ -0,0 +1,29 @@
...
+    if os.path.isfile(path) and re_testname.search(filename) and not filename in files:
+        with open(path, "r") as file:
+            files[filename] = file.read()
+
+warning = """<!-- DO NOT EDIT THIS FILE.
+Edit the appearance-* file instead and then run:
+    cd tools
+    python3 appearance-build-webkit-reftests.py
```
foolip: Why not `./appearance-build-webkit-reftests.py`? The script does not need the `#!` otherwise.

css/css-ui/tools/appearance-build-webkit-reftests.py
```diff
@@ -0,0 +1,29 @@
...
+-->
+"""
+
+for filename, text in files.items():
+    if re_link_match.search(text):
+        with open(parentdir + "/" + filename.replace("appearance-", "webkit-appearance-"), "w") as outfile:
```
foolip:
```suggestion
        with open(os.path.join(parentdir, filename.replace("appearance-", "webkit-appearance-")), "w") as outfile:
```

foolip (approved):
LGTM % minor nits

css/css-ui/tools/appearance-build-webkit-reftests.py
zcorpan: Fixed

css/css-ui/tools/appearance-build-webkit-reftests.py
zcorpan: Fixed

PR #15041, opened 2019-01-24
Change the condition for running all tests for branch to reason

foolip:
This is slightly cleaner in that the branch name is only used in one
place (the scheduled trigger) but more importantly makes it possible
to debug this by scheduling jobs for other branches at any time.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -10,11 +10,9 @@
 # project is required:
 #  - The "Build pull requests from forks of this repository" setting must be
 #    enabled: https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github#validate-contributions-from-forks
-#  - A scheduled build for the epochs/daily branch needs to be set up, the push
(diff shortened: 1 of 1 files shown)


.azure-pipelines.yml
```diff
@@ -10,11 +10,9 @@
 # project is required:
 #  - The "Build pull requests from forks of this repository" setting must be
 #    enabled: https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github#validate-contributions-from-forks
-#  - A scheduled build for the epochs/daily branch needs to be set up, the push
-#    to the branch alone is not enough: https://github.community/t5/GitHub-API-Development-and/Can-the-Checks-API-be-used-to-trigger-two-runs-when-two-pushed/m-p/17777
-#  - Because scheduled builds don't show up as check runs on GitHub, the wpt.fyi
-#    integration is different; a service connection named staging.wpt.fyi with
-#    URL https://staging.wpt.fyi needs to be created.
+#  - A scheduled builds need to be set up for the the epochs/daily branch.
```
gsnedders: "A scheduled build" or "Scheduled builds"

foolip: Just the one.

gsnedders (approved):

PR #15043, opened 2019-01-24
Updating WebDriver new window command tests to switch window

jimevans:
As originally implemented, the tests set the `handle` property of
the session instead of the `window_handle` property. Moreover, the
tests did not switch to the new window context before checking the
url, opener, and name. This commit fixes both issues.

### DIFF
--- a/webdriver/tests/new_window/new_tab.py
+++ b/webdriver/tests/new_window/new_tab.py
@@ -26,11 +26,8 @@ def test_new_tab_opens_about_blank(session):
     value = assert_success(response)
     assert value["type"] == "tab"
 
-    original_handle = session.window_handle
     session.window_handle = value["handle"]
-    new_window_url = session.url
-    session.window_handle = original_handle
-    assert new_window_url == "about:blank"
+    assert session.url == "about:blank"
 
 
 def test_new_tab_sets_no_window_name(session):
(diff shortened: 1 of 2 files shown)


webdriver/tests/new_window/new_tab.py
```diff
@@ -26,23 +26,32 @@ def test_new_tab_opens_about_blank(session):
     value = assert_success(response)
     assert value["type"] == "tab"
 
-    session.handle = value["handle"]
-    assert session.url == "about:blank"
+    original_handle = session.window_handle
+    session.window_handle = value["handle"]
+    new_window_url = session.url
+    session.window_handle = original_handle
```
whimboo: Why are you switching back to the original handle? I don't see that this is necessary.

whimboo (changes requested):
Ouch. Good find. Strange that this slipped through, but I hate the difference between `window_handle` and `handles`. Maybe we should rename the latter to be in sync with naming.

webdriver/tests/new_window/new_tab.py
jimevans: Does the framework do so automatically? I didn’t pay close enough attention to what browsing contexts got closed at the end of each subtest. If it’s unneeded, I can easily adjust.

whimboo: Yes, we run the `cleanup_session()` method from fixtures.py after each test, which cleans everything up.

jimevans: Alright then, I’ll make the changes to revert back to the earlier assert location and behavior.

jimevans: Fixed now.

whimboo (approved):

PR #15052, opened 2019-01-24
[Azure Pipelines] Include wpt.fyi prod hook for builds

foolip:


foolip:
@lukebjerring please merge this when you think the time is right :)

lukebjerring (approved):

PR #15062, opened 2019-01-25
Change error from UnknownError to InvalidModificationError when a track is added/removed from a recorded stream

youennf:
Tests for issue https://github.com/w3c/mediacapture-record/issues/151

### DIFF
--- a/mediacapture-record/MediaRecorder-error.html
+++ b/mediacapture-record/MediaRecorder-error.html
@@ -39,7 +39,7 @@
         recorder.onerror = t.step_func(mediaRecorderErrorEvent => {
             assert_true(mediaRecorderErrorEvent instanceof MediaRecorderErrorEvent, 'the type of event should be MediaRecorderErrorEvent');
             assert_equals(mediaRecorderErrorEvent.error.name, 'InvalidModificationError', 'the type of error should be InvalidModificationError when track has been added or removed');
-            assert_true(mediaRecorderErrorEvent.isTrusted, 'isTrusted should be true when the event is created by C++');
+            assert_true(mediaRecorderErrorEvent.isTrusted, 'isTrusted should be true when the event comes from the UA operations');
             assert_equals(recorder.state, "inactive", "MediaRecorder has been stopped after adding a track to stream");
             t.done();
         });
--- a/mediacapture-record/MediaRecorder-error.html
+++ b/mediacapture-record/MediaRecorder-error.html
@@ -38,7 +38,7 @@
         let recorder = new MediaRecorder(video);
         recorder.onerror = t.step_func(mediaRecorderErrorEvent => {
             assert_true(mediaRecorderErrorEvent instanceof MediaRecorderErrorEvent, 'the type of event should be MediaRecorderErrorEvent');
-            assert_equals(mediaRecorderErrorEvent.error.name, 'UnknownError', 'the type of error should be UnknownError when track has been added or removed');
+            assert_equals(mediaRecorderErrorEvent.error.name, 'InvalidModificationError', 'the type of error should be InvalidModificationError when track has been added or removed');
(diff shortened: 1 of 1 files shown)


mediacapture-record/MediaRecorder-error.html
```diff
@@ -57,7 +57,7 @@
         let recorder = new MediaRecorder(video);
         recorder.onerror = t.step_func(mediaRecorderErrorEvent => {
             assert_true(mediaRecorderErrorEvent instanceof MediaRecorderErrorEvent, 'the type of event should be MediaRecorderErrorEvent');
-            assert_equals(mediaRecorderErrorEvent.error.name, 'UnknownError', 'the type of error should be UnknownError when track has been added or removed');
+            assert_equals(mediaRecorderErrorEvent.error.name, 'InvalidModificationError', 'the type of error should be UnknownError when track has been added or removed');
```
Pehrsons:
```suggestion
            assert_equals(mediaRecorderErrorEvent.error.name, 'InvalidModificationError', 'the type of error should be InvalidModificationError when track has been added or removed');
```

yellowdoge:
Changes LGTM with a couple of notes:
- There's still a discussion in https://github.com/w3c/mediacapture-record/issues/4 as to if it's allowed or not for a `MediaStream` feeding a `MediaRecorder` to change its tracks or not; this change assumes this behaviour would not be allowed (which I advocate for).
- Can you please create a bug in Chromium and/or Firefox (if needed) and reference this change? (I know Chromium doesn't return `InvalidModificationError`, don't know about FF).

guest271314:
@yellowdoge 

> There's still a discussion in w3c/mediacapture-record#4 as to if it's allowed or not for a MediaStream feeding a MediaRecorder to change its tracks or not; this change assumes this behaviour would not be allowed (which I advocate for).

Where does the specification clearly state that tracks cannot be changed? See https://github.com/w3c/mediacapture-record/issues/147#issuecomment-480580164. Cobbled together code which _replaces_ tracks at Chromium 73 during the recording. Propose  that `MediaRecorder` have a `.replaceTrack()` added method similar to or exceeding the functionality of 

> [5.2 RTCRtpSender Interface](http://w3c.github.io/webrtc-pc/#rtcrtpsender-interface)
> 
> replaceTrack
> 
> 6.4.3. If sending is true, and withTrack is not null, have the sender **switch seamlessly to transmitting withTrack instead of the sender's existing track.** (emphasis added). 

> Can you please create a bug in Chromium and/or Firefox (if needed) and reference this change? (I know Chromium doesn't return InvalidModificationError, don't know about FF).


Firefox bug report https://bugzilla.mozilla.org/show_bug.cgi?id=1542616.

dontcallmedom (approved):

mediacapture-record/MediaRecorder-error.html
```diff
@@ -38,7 +38,7 @@
         let recorder = new MediaRecorder(video);
         recorder.onerror = t.step_func(mediaRecorderErrorEvent => {
             assert_true(mediaRecorderErrorEvent instanceof MediaRecorderErrorEvent, 'the type of event should be MediaRecorderErrorEvent');
-            assert_equals(mediaRecorderErrorEvent.error.name, 'UnknownError', 'the type of error should be UnknownError when track has been added or removed');
+            assert_equals(mediaRecorderErrorEvent.error.name, 'InvalidModificationError', 'the type of error should be InvalidModificationError when track has been added or removed');
             assert_true(mediaRecorderErrorEvent.isTrusted, 'isTrusted should be true when the event is created by C++');
```
dontcallmedom, line 42:
```suggestion
            assert_true(mediaRecorderErrorEvent.isTrusted, 'isTrusted should be true when the event comes from the UA operations');
```

PR #15068, opened 2019-01-25
Add Noto Naskh Arabic regular woff2 font

himorin:
file original from resources.whatwg.org/fonts/NotoNaskhArabic-Regular.woff2

We are planning to add some tests for css-text shaping using Arabic font, so we want to add Noto Naskh Arabic regular woff2 font file to wpt.

cc @fantasai @xfq @r12a


r12a:
cc @frivoal since this may be relevant for the zwj tests you were working on at https://github.com/web-platform-tests/wpt/pull/14673

kojiishi (approved):
lgtm

frivoal (approved):

jgraham:
I assume the intent is that this is only usable as a webfont?

frivoal:
what else do you have in mind?

gsnedders:
@frivoal we install Ahem at a system level, but… that's painful and somewhat fragile.

frivoal:
Right. No need for that. These fonts can be called in explicitly by tests that need them.

PR #15069, opened 2019-01-25
Check if the test verification results is empty

qiuzhong:
* Check the test verification results: if it's empty then the step
results should be set to "FAIL".
* Set the summary results based on the step results list.

Fix: #15020

### DIFF
--- a/tools/wptrunner/wptrunner/stability.py
+++ b/tools/wptrunner/wptrunner/stability.py
@@ -343,6 +343,14 @@ def check_stability(logger, repeat_loop=10, repeat_restart=5, chaos_mode=True, m
             write_summary(logger, step_results, "FAIL")
             return 1
 
-        step_results.append((desc, "PASS"))
+        step_result = "PASS"
+        if not results:
+            step_result = "FAIL"
+        step_results.append((desc, step_result))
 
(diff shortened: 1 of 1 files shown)


foolip:
@jgraham would you be able to review this? I could do it, but haven't looked at this code before.

tools/wptrunner/wptrunner/stability.py
```diff
@@ -343,6 +343,14 @@ def check_stability(logger, repeat_loop=10, repeat_restart=5, chaos_mode=True, m
             write_summary(logger, step_results, "FAIL")
             return 1
 
-        step_results.append((desc, "PASS"))
-
-    write_summary(logger, step_results, "PASS")
+        step_result = "PASS"
+        if not results:
```
gsnedders: This can only happen when there are no results, right @jgraham, and not because it's the wrong test type or something? I forget what exactly the steps are, and it's not obvious looking from the code.

jgraham: I'm not sure we want to fail in this case TBH. I'm insufficeintly convinced that we never schedule this job with either no test paths or bogus test paths.

Issue #12515, opened 2018-08-16
CSSOMString needs to be defined but is gone from cssom.idl in reffy-reports


foolip:
In https://github.com/w3c/csswg-drafts/commit/08860d449be888547017dcd88531fc68f6faf882 @tabatkins changed CSSOM so that both of the `CSSOMString` definitions in https://drafts.csswg.org/cssom/#cssomstring-type are on equal footing, and as a result neither of them gets treated by Reffy as an IDL block. This caused `typedef USVString CSSOMString` to disappear from cssom.idl in https://github.com/tidoust/reffy-reports/commit/bbbcccbfc13d9a628dee77a299f2a0632993d963.

When this change is brought in, every test that depends on cssom.idl will need to define CSSOMString in some other way.

Alternatively, we could add it as a known string type in webidl2.js itself, here:
https://github.com/w3c/webidl2.js/blob/91bba911e5a316b96a1b8d2cd8475ec0ccd703b5/lib/webidl2.js#L18

@saschanaz, might you have thoughts on the best way to proceed here?

saschanaz:
The string type list in webidl2.js is [to process relavant Web IDL syntax](https://heycam.github.io/webidl/#prod-StringType), so probably it shouldn't be there.

```
StringType ::
    ByteString
    DOMString
    USVString
```

foolip:
Unfortunately I'll have to agree that makes sense, so we can't push the problem down to webidl2.js.

I think what we should do is include `typedef DOMString CSSOMString` in cssom.idl manually somehow, because a separate cssom-CSSOMString.idl for just that would be annoying. I think we'll do it as part of https://github.com/foolip/wpt-idl-importer then.

foolip:
Fixing this in the import step now: https://github.com/tidoust/reffy-reports/pull/29

foolip:
Fixed now.

PR #15071, opened 2019-01-25
Convert css/ idlharness.js tests to use idl_test

foolip:
This is to make all loads of cssom.idl go through the same code
path which will make it possible to add a workaround for
https://github.com/web-platform-tests/wpt/issues/12515 in idl_test.

### DIFF
--- a/css/css-conditional/idlharness.html
+++ b/css/css-conditional/idlharness.html
@@ -9,6 +9,16 @@
   <script src="/resources/testharnessreport.js"></script>
   <script src="/resources/WebIDLParser.js"></script>
   <script src="/resources/idlharness.js"></script>
+  <!-- used to provide objects -->
+  <style>
+    div { display: block; }
+  </style>
+  <style>
+    @media print { }
+  </style>
+  <style>
+    @supports (display: block) { }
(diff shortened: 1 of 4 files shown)


css/css-conditional/idlharness.html
```diff
@@ -16,16 +16,13 @@
...
-      idl_array.add_dependency_idls(cssom);
-      idl_array.add_dependency_idls(dom);
-      idl_array.test();
-    }, 'Test css-conditional IDL implementation');
+    idl_test(
+      ['css-conditional'],
+      ['cssom', 'dom'],
+      idl_array => {
+        // No objects
+      }
```
Ms2ger: Nit: consider dropping the callback argument entirely.

Ms2ger (approved):

css/css-conditional/idlharness.html
foolip: They were sort of a placeholder for objects to be added, so I've pushed a commit to do just that.

lukebjerring:
Ignoring "regressions" which are actually new tests, which are expectedly failing.

PR #15072, opened 2019-01-25
Update cssom.idl and work around removal of CSSOMString

foolip:
Fixes https://github.com/web-platform-tests/wpt/issues/12515.

### DIFF
--- a/resources/idlharness.js
+++ b/resources/idlharness.js
@@ -3152,11 +3152,25 @@ function idl_test(srcs, deps, idl_setup_func) {
  */
 function fetch_spec(spec) {
     var url = '/interfaces/' + spec + '.idl';
-    return fetch(url).then(function (r) {
+    return fetch(url).then(function(r) {
         if (!r.ok) {
             throw new IdlHarnessError("Error fetching " + url + ".");
         }
         return r.text();
+    }).then(function(idl) {
+        if (spec == 'cssom') {
+            // CSSOMString is an implementation-defined typedef of either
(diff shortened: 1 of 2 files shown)


foolip:
https://github.com/web-platform-tests/wpt/pull/15071 needs to be merged first and the tests for this then run again. Should have done a single PR :)

resources/idlharness.js
```diff
@@ -3152,11 +3152,25 @@ function idl_test(srcs, deps, idl_setup_func) {
...
+            // CSSOMString is an implementation-defined typedef of either
+            // DOMString or USVString in CSSOM:
+            // https://drafts.csswg.org/cssom/#cssomstring-type
+            // Because of this, neither typedef is included in cssom.idl.
+            // We need some definition, and use DOMString because USVString
+            // has additional requirements in type conversion and could result
+            // in spurious failures for implementations that use DOMString.
+            // Note: No newline is added to preserve line numbers.
+            return 'typedef DOMString CSSOMString;' + idl;
+        }
```
Ms2ger: This is an odd place to put it; can we just always add it in the `IdlArray` constructor?

lukebjerring: I don't think special-casing it in the testing framework is the right approach. The fact that it's defined somewhere in the bikeshed source (differently than an IDL snippet) should be handled by the scraper, or importer, but not by the idlharness or test layer.

foolip:
New deal: https://github.com/tidoust/reffy-reports/pull/29

PR #15078, opened 2019-01-25
HTML: window.open() can set opener of a nested browsing context

annevk:
Needed for https://github.com/whatwg/html/pull/4284.

### DIFF
--- a/html/browsers/windows/embedded-opener-a-form.html
+++ b/html/browsers/windows/embedded-opener-a-form.html
@@ -1,0 +1,30 @@
+<!doctype html>
+<title>opener and embedded documents; using a and form</title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<div id=log></div>
+<iframe name=matchesastring></iframe>
+<a href=/common/blank.html target=matchesastring>&lt;a></a>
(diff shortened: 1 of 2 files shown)


html/browsers/windows/embedded-opener-a-form.html
```diff
@@ -0,0 +1,30 @@
+<!doctype html>
+<title>opener and embedded documents; using a and form</title>
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<div id=log></div>
+<iframe name=matchesastring></iframe>
+<a href=/common/blank.html target=matchesastring>&lt;a></a>
+<form action=/common/blank.html target=matchesastring><input type=submit value="<form>"></form>
```
zcorpan: The form part is not tested?

annevk: `document.querySelector("input").click();` tests it, no?

annevk:
@foolip @domenic can you look at this?

domenic (approved):
These tests both seem reasonable, although I don't understand the relationship between them.

annevk:
They're both named targeting, but one of them ends up setting opener and the other doesn't. This was one of the changes with browsing context refactoring, that `window.open()` now directly sets opener sometimes.

PR #15092, opened 2019-01-26
Escaped EOF

tabatkins:
Tests <https://github.com/w3c/csswg-drafts/issues/3182> and <https://github.com/w3c/csswg-drafts/issues/1821>

### DIFF
--- a/css/css-syntax/escaped-eof.html
+++ b/css/css-syntax/escaped-eof.html
@@ -1,0 +1,53 @@
+<!doctype html>
+<title>Escaped EOF</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<meta name="author" title="Tab Atkins-Bittner">
+<link rel=help href="https://drafts.csswg.org/css-syntax/#consume-escaped-code-point">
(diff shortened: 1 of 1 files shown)


css/css-syntax/escaped-eof.html
```diff
@@ -0,0 +1,53 @@
...
+<script>
+
+function threw(fn) {
+    try{fn();}
+    catch(e) { return true; }
+    return false;
+}
+
+test(()=>{
+    assert_true(threw(()=>{document.querySelector("#123");}), "numeric hash token is invalid in a selector");
```
domenic: Use assert_throws, so you can test the exception type too?

css/css-syntax/escaped-eof.html
```diff
@@ -0,0 +1,53 @@
...
+    assert_false(threw(()=>{document.querySelector("#foo\\");}), "escaped-EOF in a hash token is valid in a selector");
```
domenic: If you just perform the querySelector, any exceptions will cause automatic test failures, with better failure messages than "got true, expected false".

css/css-syntax/escaped-eof.html
tabatkins: Was using it originally, but didn't like the asymmetry in my second condition. That said... (see next response).

PR #954, opened 2014-04-30
Unit tests for SourceBuffer#appendStream function

shishimaru:
I added unit tests for SourceBuffer#appensStream function.
Please review the codes.

Hitoshi Uchida uchida@w3.org

PR #15097, opened 2019-01-26
ReadableStream @@asyncIterator

MattiasBuelens:
Add tests for async-iterating a `ReadableStream`. See whatwg/streams#980 for the accompanying spec change.

This PR continues from @devsnek's work in #13362.

The tests are based on [Domenic's comment](https://github.com/whatwg/streams/pull/954#issuecomment-421107013) in whatwg/streams#954. So far, the following tests have been implemented:

* [x] Basic test of a stream with a few chunks that closes
  * [x] Using a "push" underlying source (calls `c.enqueue()` in `start`) that enqueues all at once
  * [x] Using a "push" underlying source  where you call `c.enqueue()` "just in time" (i.e. inside the loop body)
  * [x] Using a "pull" underlying source (calls `c.enqueue()` in `pull`) using recordingReadableStream to verify the correct underlying source method calls
* [x] Degenerate streams
  * [x] Async-iterating an errored stream throws
  * [x] Async-iterating a closed stream never executes the loop body, but works fine
  * [x] Async-iterating an empty but not closed/errored stream never executes the loop body and stalls the async function (end the test if 50 ms pass without the promise resolving)
* [x] `@@asyncIterator()` method is `===` to `getIterator()` method
* [x] Cancellation behavior (use recordingReadableStream)
  * [x] Manually calling `return()` causes a cancel
  * [x] `throw`ing inside the loop body causes a cancel
  * [x] `break`ing inside the loop body causes a cancel
  * [x] `return`ing inside the loop body causes a cancel
  * [x] All of the above, but with `preventCancel: true` and the pass conditions reversed
* [x] Manual manipulation
  * [x] double-return rejects ([#954 (comment)](https://github.com/whatwg/streams/pull/954#discussion_r216237031))
  * [x] next()'s fulfillment values have exactly the right shape (Object.prototype, only two properties, correct property descriptors)
  * [x] throw method does not exist
  * [x] Calling next() or return() on non-ReadableStreamDefaultReaderAsyncIterator instances rejects with TypeError (there are existing brand check tests you can add to)
  * [x] Calling return() while there are pending reads (i.e. unsettled promises returned by next()) rejects the return Promise ([#950 (comment)](https://github.com/whatwg/streams/pull/950#issuecomment-417393457))
* [x] Interaction with existing infrastructure
  * [x] getIterator/@@asyncIterator throw if there's already a lock
  * [x] Monkey-patch getReader and default reader's prototype methods and ensure this does not interfere with async iteration (e.g. using one of the basic test cases)
  * [x] Basic test with a few chunks but you've already consumed one or two via a normal reader (which you then released)
* [x] Auto-release
  * [x] You can still acquire a reader and successfully use its `.closed` promise after exhausting the async iterator via for-await-of
  * [x] You can still acquire a reader and successfully use its `.closed` promise after `return()`ing from the async iterator (either manually or via `break`; take your pick)

### DIFF
--- a/streams/readable-streams/async-iterator.any.js
+++ b/streams/readable-streams/async-iterator.any.js
@@ -113,34 +113,34 @@ promise_test(async () => {
 for (const type of ['throw', 'break', 'return']) {
   for (const preventCancel of [false, true]) {
     promise_test(async () => {
-    const s = recordingReadableStream({
-      start(c) {
-        c.enqueue(0);
-      }
-    });
-
-    // use a separate function for the loop body so return does not stop the test
-    const loop = async () => {
-      for await (const c of s.getIterator({ preventCancel })) {
-        if (type === 'throw') {
-          throw new Error();
-        } else if (type === 'break') {
-          break;
-        } else if (type === 'return') {
-          return;
+      const s = recordingReadableStream({
+        start(c) {
+          c.enqueue(0);
         }
+      });
+
+      // use a separate function for the loop body so return does not stop the test
+      const loop = async () => {
+        for await (const c of s.getIterator({ preventCancel })) {
+          if (type === 'throw') {
+            throw new Error();
+          } else if (type === 'break') {
+            break;
+          } else if (type === 'return') {
+            return;
+          }
+        }
+      };
+
+      try {
+        await loop();
+      } catch (e) {}
+
+      if (preventCancel) {
+        assert_array_equals(s.events, ['pull'], `cancel() should not be called when type = '${type}' and preventCancel is true`);
+      } else {
+        assert_array_equals(s.events, ['pull', 'cancel', undefined], `cancel() should be called when type = '${type}' and preventCancel is false`);
       }
-    };
-
-    try {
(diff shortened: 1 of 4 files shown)


ricea (commented):
Looks good.

ricea (approved):
lgtm

We should land this at the same time as the standard changes.

MattiasBuelens:
I still want to add two more tests that were originally suggested:
* Using a "pull" underlying source (calls `c.enqueue()` in `pull`) using recordingReadableStream to verify the correct underlying source method calls
* throw method does not exist

I'll look into adding them.

MattiasBuelens:
Last few tests added. Now it's just a matter of fixing the reference implementation to pass them...

ricea:
Still lgtm with latest changes.

guest271314:
Is this behind a flag at Chromium?

ricea:
> Is this behind a flag at Chromium?

No. See https://bugs.chromium.org/p/chromium/issues/detail?id=929585.

PR #15104, opened 2019-01-27
added 2 tests coming from chromium project

TalbotG:
These 2 tests are adapted and separated parts of the original test at chromium project which is now:

https://chromium.googlesource.com/chromium/src/+/c825d655f6aaf73484f9d56e9012793f5b9668cc/third_party/WebKit/LayoutTests/css3/calc/calc-with-time-angle-and-frequency-values.html

### DIFF
--- a/css/css-values/calc-angle-values.html
+++ b/css/css-values/calc-angle-values.html
@@ -1,0 +1,170 @@
+<!DOCTYPE html>
+
+  <meta charset="UTF-8">
+
+  <title>CSS Values and Units Test: calc() function with angle values</title>
+
+  <!--
+
+  Original test is:
+
+  https://chromium.googlesource.com/chromium/src/+/c825d655f6aaf73484f9d56e9012793f5b9668cc/third_party/WebKit/LayoutTests/css3/calc/calc-with-time-angle-and-frequency-values.html
+
+  Issue 917718: [css-values] calc-with-time-angle-and-frequency-values
+  test is highly unreliable, transition-delay testing causes side effects
+  https://bugs.chromium.org/p/chromium/issues/detail?id=917718
+
+  -->
+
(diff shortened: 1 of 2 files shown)


css/css-values/calc-angle-values.html
```diff
@@ -0,0 +1,170 @@
...
+       Turns. There is 1 turn in a full circle.
+
+   */
+
+
+ /* compareValue(property_name, calcValue, expectedValue, description)  */
+
+    compareValue("transform", "rotate(calc(45deg + 45deg))", "rotate(90deg)", "addition of 2 angle units: deg plus deg");
+
+    compareValue("transform", "rotate(calc(45deg + 1rad))", "rotate(102.3deg)", "addition of 2 angle units: deg plus rad");
```
frivoal: Please use something like shouldBeCloseTo()

frivoal (changes requested):

css/css-values/calc-angle-values.html
TalbotG: Inserting Math.PI into the functions works as expected. Eg

http://www.gtalbot.org/BrowserBugsSection/CSS3Values/calc-Math.PI-rad-and-grad.html

I will adapt+adjust calc-angle-values.html test to use Math.PI

TalbotG: I have create compareValuesCloseTo which defines a tolerance precision (epsilon):

http://www.gtalbot.org/BrowserBugsSection/CSS3Values/calc-angle-values-latest.html

I am closing this Pull Request since I have submitted a new one 
https://github.com/web-platform-tests/wpt/pull/15342
with all of the requested improvements in here.

TalbotG:
Please visit pull request 15342
https://github.com/web-platform-tests/wpt/pull/15342

PR #15110, opened 2019-01-28
[Azure Pipelines] Rename all/affected Safari job/artifact names

foolip:
This is to allow for adding other browsers than Safari Technology
Preview without putting all the results in a single artifact, which
might be inconveniently large if it holds 3 full runs.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -22,7 +22,7 @@
 jobs:
 # The affected tests jobs are unconditional for speed, as most PRs have one or
 # more affected tests: https://github.com/web-platform-tests/wpt/issues/13936.
-- job: affected_macOS
+- job: affected_safari_preview
   displayName: 'affected tests (Safari Technology Preview)'
   condition: eq(variables['Build.Reason'], 'PullRequest')
   pool:
@@ -30,9 +30,9 @@ jobs:
   steps:
   - template: tools/ci/azure/affected_tests.yml
     parameters:
-      artifactName: 'affected-tests'
+      artifactName: 'safari-preview-affected-tests'
 
-- job: affected_without_changes_macOS
+- job: affected_without_changes_safari_preview
   displayName: 'affected tests without changes (Safari Technology Preview)'
   condition: eq(variables['Build.Reason'], 'PullRequest')
   pool:
@@ -42,7 +42,7 @@ jobs:
     parameters:
       checkoutCommit: 'HEAD^1'
       affectedRange: 'HEAD@{1}'
-      artifactName: 'affected-tests-without-changes'
+      artifactName: 'safari-preview-affected-tests-without-changes'
 
 # The decision jobs runs `./wpt test-jobs` to determine which jobs to run,
 # and all following jobs wait for it to finish and depend on its output.
@@ -157,7 +157,7 @@ jobs:
     condition: succeededOrFailed()
   - template: tools/ci/azure/cleanup_win10.yml
 
-- job: all_edge
+- job: results_edge
   displayName: 'all tests (Edge)'
   # This job is only triggered manually until it has been shown to be robust.
   condition: and(eq(variables['Build.Reason'], 'Manual'), variables['run_all_edge'])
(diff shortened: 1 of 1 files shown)


foolip:
@lukebjerring I presume this will require changes in wpt.fyi. I'm assuming here that it's possible to enumerate the artifacts and pattern match the names, but if that's not possible this won't work.

lukebjerring:
What's the need for the rename? The impact on wpt.fyi is that it won't label things correctly.

foolip:
If a single build runs multiple full runs with different browsers, the choices are a single  (potentially quite large) artifact with a file naming pattern inside the zip file, or multiple artifact, one per run. The latter seemed better, but either would require wpt.fyi to adapt I assume?

lukebjerring:
I agree that split is better. We enumerate all of the artifacts in either case, so tweaking the name to label map is easier than grouping the internals of the artifacts contents.

.azure-pipelines.yml
```diff
@@ -150,13 +152,13 @@ jobs:
   - task: PublishBuildArtifacts@1
     displayName: 'Publish results'
     inputs:
-      artifactName: 'results'
+      artifactName: 'all-safari-preview'
```
foolip: @lukebjerring would results-safari-preview be a better name? Not sure what to make this consistent with. All artifacts have results of some kind. But this also isn't *all* tests, just the ones from `./wpt run`. Since every rename here will require changes to the receiver, I'll use whatever name you say and then we won't change it again.

lukebjerring: Let's stick with results-[product] then.

foolip: Done!

.azure-pipelines.yml
```diff
@@ -17,17 +17,17 @@
...
+- job: affected_safari_preview
   displayName: 'affected tests (Safari Technology Preview)'
   condition: eq(variables['Build.Reason'], 'PullRequest')
   pool:
     vmImage: 'macOS-10.13'
   steps:
   - template: tools/ci/azure/affected_tests.yml
     parameters:
-      artifactName: 'affected-tests'
+      artifactName: 'affected-safari-preview'
```
lukebjerring:
```suggestion
      artifactName: 'safari-preview-affected-tests'
```

.azure-pipelines.yml
```diff
@@ -37,7 +37,7 @@ jobs:
     parameters:
       checkoutCommit: 'HEAD^1'
       affectedRange: 'HEAD@{1}'
-      artifactName: 'affected-tests-without-changes'
+      artifactName: 'affected-without-changes-safari-preview'
```
lukebjerring:
```suggestion
      artifactName: 'safari-preview-affected-tests-without-changes'
```

.azure-pipelines.yml
```diff
@@ -150,13 +152,13 @@ jobs:
   - task: PublishBuildArtifacts@1
     displayName: 'Publish results'
     inputs:
-      artifactName: 'results'
+      artifactName: 'results-safari-preview'
```
lukebjerring:
```suggestion
      artifactName: 'safari-preview-results'
```

lukebjerring (commented):
Tweaked to more closely match taskcluster, without the deliberate test-type sharding (reftest, etc) and the `wpt-` prefix.

.azure-pipelines.yml
foolip: Hmm, don't remember if we said something out of band, but https://github.com/web-platform-tests/wpt.fyi/pull/1107 expects safari-results now.

foolip:
https://wpt.fyi/about says that https://github.com/web-platform-tests/wpt.fyi/commit/980ee163000f74f18435157a9015946144bf65cb is deployed, which includes https://github.com/web-platform-tests/wpt.fyi/pull/1107, so this change should be safe to make now.

@lukebjerring I've rebased to resolve conflicts, can you review?

lukebjerring (approved):

Issue #11726, opened 2018-06-29
assert_throws is too loose, misses browser bugs


bzbarsky:
I discovered today that some browsers throw a `DOMException` with `name == "TypeError"` in some cases when specs say to throw a `TypeError`.

Unfortunately, wpt's `assert_throws(new TypeError(), stuff)` doesn't catch this, because it just compares `.name` values.

I was going to suggest comparing `.constructor.name` values as well, but then I discovered https://github.com/web-platform-tests/wpt/pull/11643 and the issues mentioned in the review comments there.

It seems like we really want the following primitives for testing throwing behavior:

1. Test that a DOMException is thrown with the given name/code and corresponding code/name.  Right now you can sort of get that by passing a string name or integer code as the first arg, instead of an object, though even then `assert_throws` doesn't check that the result is a `DOMException` in any meaningful way (e.g. by checking `constructor.name` or trying to use a `DOMException.prototype` getter on it or something).

2. Test that an exception is thrown with the given `.constructor.name`, or whatever other way we can think of to detect a "real TypeError".

3.  Test whatever https://github.com/web-platform-tests/wpt/pull/11643 is trying to test.

The current `assert_throws` does a mix of these, without getting any of them right.

Are there other things one might want to test?  Should we even be trying to shoehorn all of these into a single `assert_throws`?

//cc @zcorpan @saschanaz @gsnedders @jgraham @annevk @domenic

saschanaz:
`assert_throws` currently looks too long, how about `assert_throws_domexception` for DOMException cases?

domenic:
> Test that an exception is thrown with the given .constructor.name, or whatever other way we can think of to detect a "real TypeError".

It may be a good idea to test if .constructor is ===, to catch wrong-realm issues. That may be harder to mass-convert though

Here is my ideal world:

- `assert_throws(TypeError, stuff)`: checks `ex.constructor === TypeError`
- `assert_throws_domexception("StringName", stuff)`: checks `ex.constructor.name === "DOMException"`, and `ex.name === "StringName"`
  - Get rid of the integer variant; it just leads to confusion as to which one to use
  - Maybe extend this with an overload that also checks the realm? I dunno.
- `assert_throws_exactly(value, stuff)`: checks `Object.is(ex, value)`.

Whether anyone wants to invest all the time necessary to fix the thousands of callers... seems unlikely.

[...]

PR #15116, opened 2019-01-28
Clarify that object arg to assert_throws is for non-DOMExceptions

tabatkins:
Previous text made it seem like passing any object with the correct "name" property would be fine, but in reality, if you're specifically expecting a DOMException, you should be passing a string. (per @Ms2ger feedback)

Preserved the original wording as much as possible otherwise.

(Probably `assert_throws` itself should check the error type if called with an object, and verify that it's *not* a DOMException. But that's a separate cleanup project.)


Ms2ger:
Ref: #11726

Ms2ger (approved):

PR #15105, opened 2019-01-27
Improve `EventListener::handleEvent()` coverage

shvaikalesh:
Came to my attention because of [this tweet](https://twitter.com/roman01la/status/1088845282757001219): Chrome and Safari seem to ignore non-callable `handleEvent` method, while Firefox follows [the spec](https://heycam.github.io/webidl/#call-a-user-objects-operation) (step 10.4).

PR #15122, opened 2019-01-29
Align legacy callback interface tests

shvaikalesh:
Follow-up of #15105. Safari seem to swallow an exception from `acceptNode` getter and throw `TypeError` instead. I will make bug reports shortly.

### DIFF
--- a/dom/events/EventListener-handleEvent.html
+++ b/dom/events/EventListener-handleEvent.html
@@ -69,6 +69,19 @@ test(function(t) {
 test(function(t) {
     var type = "foo";
     var target = document.getElementById("target");
+
+    var eventListener = function() {};
+    eventListener.handleEvent = function() {
+        assert_unreached("`handleEvent` method should not be called on functions");
+    };
+
+    target.addEventListener(type, eventListener);
+    target.dispatchEvent(new Event(type));
+}, "doesn't call `handleEvent` method on callable `EventListener`");
+
+test(function(t) {
+    var type = "foo";
+    var target = document.getElementById("target");
     var uncaughtError;
 
     window.addEventListener("error", function(event) {
--- a/dom/events/EventListener-handleEvent.html
+++ b/dom/events/EventListener-handleEvent.html
@@ -73,12 +73,13 @@ test(function(t) {
 test(function(t) {
     var type = "foo";
     var target = document.getElementById("target");
-
-    var eventListener = function() {};
+    var calls = 0;
+    var eventListener = function() { calls++; };
     eventListener.handleEvent = t.unreached_func("`handleEvent` method should not be called on functions");
 
     target.addEventListener(type, eventListener);
     target.dispatchEvent(new Event(type));
+    assert_equals(calls, 1);
 }, "doesn't call `handleEvent` method on callable `EventListener`");
 
 test(function(t) {
--- a/dom/events/EventListener-handleEvent.html
+++ b/dom/events/EventListener-handleEvent.html
@@ -5,13 +5,12 @@
 <script src="/resources/testharnessreport.js"></script>
 <link rel="help" href="https://dom.spec.whatwg.org/#callbackdef-eventlistener">
 <div id=log></div>
-<div id=target></div>
 <script>
 setup({ allow_uncaught_exception: true });
 
 test(function(t) {
     var type = "foo";
-    var target = document.getElementById("target");
+    var target = document.createElement("div");
     var eventListener = {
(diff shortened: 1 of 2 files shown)


dom/events/EventListener-handleEvent.html
```diff
@@ -66,6 +66,19 @@
     assert_equals(calls, 2);
 }, "performs `Get` every time event is dispatched");
 
+test(function(t) {
+    var type = "foo";
+    var target = document.getElementById("target");
+
+    var eventListener = function() {};
```
annevk: Add something here that proves the function itself is called?

dom/events/EventListener-handleEvent.html
```diff
@@ -81,5 +94,22 @@
...
+}, "throws if `handleEvent` is falsy and not callable");
+
+test(function(t) {
+    var type = "foo";
+    var target = document.getElementById("target");
+    var uncaughtError;
+
+    window.addEventListener("error", function(event) {
+        uncaughtError = event.error;
+    });
```
annevk: We should maybe add `{ once: true }` to this event listener (and the others in the file) as currently all three of them will run by the time you get to this test. It doesn't matter much as they do the same thing, but it still seems better to account for future changes now and not let someone else try to debug that.

annevk (commented):
Thanks a lot for writing this. I only have a couple nits.

By the way, you should feel free to use `let` and `const` in new tests, if you're comfortable with those.

cc @Ms2ger

dom/events/EventListener-handleEvent.html
```diff
@@ -66,6 +66,19 @@
     assert_equals(calls, 2);
 }, "performs `Get` every time event is dispatched");
 
+test(function(t) {
+    var type = "foo";
+    var target = document.getElementById("target");
+
+    var eventListener = function() {};
+    eventListener.handleEvent = function() {
+        assert_unreached("`handleEvent` method should not be called on functions");
```
Ms2ger: This should be wrapped in `this.step_func()` in general, but for this case, there's a shorter approach with `this.unreached_func()`.

dom/traversal/TreeWalker-acceptNode-filter.html
```diff
@@ -135,6 +124,38 @@
     assert_node(walker.currentNode, { type: Element, id: 'root' });
 }, 'Testing with filter function that throws');
 
+test(function() {
+    var testError = new Error();
```
Ms2ger: I would typically use `{ "name": "test" }` here.

Ms2ger (commented):
Great tests!

dom/events/EventListener-handleEvent.html
shvaikalesh: d2a45d1

dom/traversal/TreeWalker-acceptNode-filter.html
shvaikalesh: b43a121

dom/events/EventListener-handleEvent.html
shvaikalesh: 4b79865

dom/events/EventListener-handleEvent.html
shvaikalesh: Even with `once: true` we risk tests affecting each other in future: non-conformant implementation do not throw `"error"` event.  With current code that would not be a problem, but to make the tests 100% pure I've used `add_cleanup` helper to manually remove `"error"` handlers in a6d0d08.

shvaikalesh:
Thanks for quick & thorough reviews.

After making requested changes I've discovered that "rethrow in getter" test affects later "non-callable `handleEvent`" tests, so I've switched to creating `EventTarget` instance in each test (1a849b6).

shvaikalesh:
Taskcluster fails on:

```
Switched to a new branch 'build'
++ source tools/ci/start.sh
/home/test/start.sh: line 40: tools/ci/start.sh: No such file or directory
```

@annevk, is there anything I could do to get this merged?

annevk:
Apparently changes have been made so I rebased for you.

annevk (approved):
This looks good to me. Could you link the tests you filed? @Ms2ger want to have a final look?

Ms2ger (approved):
Thanks!

shvaikalesh:
WebKit issue: [Error thrown during "acceptNode" lookup is overridden](https://bugs.webkit.org/show_bug.cgi?id=200735).

PR #15131, opened 2019-01-29
Add some tests for Symbol properties on Storage objects.

Ms2ger:


inexorabletash (approved):
lgtm, thanks!

Hopefully this will some day prevent some bug somewhere.

Ms2ger:
One double failure in Chrome, four in Gecko, and a crash in Servo. I'd call that success :)

bzbarsky:
@Ms2ger I think the "double failure in Chrome" is a test bug; I get the same failures in Gecko when I fix https://bugzilla.mozilla.org/show_bug.cgi?id=1520831 per what I understand the spec to say.

The test is asserting that after doing:
```
  var key = Symbol();
  Object.defineProperty(storage, key, { "value": "test", "configurable": false });
  var desc = Object.getOwnPropertyDescriptor(storage, key);
```
the value of `desc.configurable` is true.  But I don't see why that should be the case.  https://heycam.github.io/webidl/#legacy-platform-object-defineownproperty does not change any properties of Desc and just calls OrdinaryDefineOwnProperty, which will define it as a non-configurable property.  Then https://heycam.github.io/webidl/#legacy-platform-object-getownproperty calls https://heycam.github.io/webidl/#LegacyPlatformObjectGetOwnProperty which for the case of a symbol-named property will return `OrdinaryGetOwnProperty(O, P)`, which will return a descriptor with `configurable` set to false as far as I can tell.

Am I just missing something?

Ms2ger:
I think you're missing step 3 in <https://heycam.github.io/webidl/#legacy-platform-object-defineownproperty>? (Then again, I knew it was there and it still took me three tries to find it.)

bzbarsky:
Hmm.  I did miss that, and I was explicitly looking for it!

I think that step is bogus, fwiw.  Filed https://github.com/heycam/webidl/issues/669 to get that sorted out.

PR #15135, opened 2019-01-29
Fix useless test in dataset-delete.html.

Ms2ger:
This is not currently testing anything useful, and hasn't been since I
introduced this subtest in d5a8cbe61b61c7fa9891f380bdfe1e0c142fbc84 back in
2012.

However, the test I meant to write was not correct either, since the steps to
delete an existing named property in HTML are only run if the property is
exposed on the object, which isn't the case here.


annevk (commented):
Also remove the testDeleteNoAdd function then?

Ms2ger:
It's still used lower down.

annevk (approved):
Sorry!

PR #15141, opened 2019-01-29
Use print() function in both Python 2 and Python 3

cclauss:
Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.

### DIFF
--- a/css/tools/apiclient/test.py
+++ b/css/tools/apiclient/test.py
@@ -11,6 +11,7 @@
 #  [1] http://www.w3.org/Consortium/Legal/2002/copyright-software-20021231
 #
 
+from __future__ import print_function
 import sys
 import os
 import glob
@@ -24,11 +25,11 @@ from apiclient import apiclient
 
 def runTests(testFileSearch):
     for testFilePath in glob.glob(testFileSearch):
-        print 'Running tests from: ' + testFilePath
+        print('Running tests from: ' + testFilePath)
         with open(testFilePath) as testFile:
             testData = json.load(testFile, object_pairs_hook = collections.OrderedDict)
             for testSetName in testData:
-                print testSetName + ':'
+                print(testSetName + ':')
                 testSet = testData[testSetName]
                 vars = testSet['variables']
                 for test in testSet['testcases']:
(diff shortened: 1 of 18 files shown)


css/tools/apiclient/test.py
```diff
@@ -11,6 +11,7 @@
 #  [1] http://www.w3.org/Consortium/Legal/2002/copyright-software-20021231
```
Ms2ger: It's not clear to me if we want to keep this in sync with upstream; CC @gsnedders @plinss

tools/pywebsocket/example/echo_client.py
```diff
@@ -51,6 +51,7 @@
 # run echo client to test IETF HyBi 00 protocol
  run with --protocol-version=hybi00
 """
+from __future__ import print_function
```
Ms2ger: Same question @gsnedders , but at least here upstream seems dead.

tools/third_party/py/bench/localpath.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
```
Ms2ger: https://github.com/pytest-dev/pytest/ seems alive still.

tools/wptrunner/setup.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
 # This Source Code Form is subject to the terms of the Mozilla Public
```
Ms2ger: We tend to put the import after the license block.

tools/wptrunner/wptrunner/wptrunner.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import unicode_literals
+from __future__ import print_function
```
Ms2ger: Merge the imports

Ms2ger (dismissed):
I'd be happy to take the changes to our own code, but we need to discuss the changes to upstream code.

css/tools/apiclient/test.py
gsnedders: Yeah, let's not fork this.

tools/pywebsocket/example/echo_client.py
gsnedders: This is in `tools/` and not `tools/third_party/` because we've already forked it, so this is fine from my POV.

tools/third_party/py/bench/localpath.py
gsnedders: Yeah, nothing in `tools/third_party` should be touched.

gsnedders:
The latest CI failures here are #15232.

css/tools/apiclient/test.py
cclauss: What is the repo for upstream.  These are _syntax errors_ that crash will the runtime.

gsnedders: Given the CSS test suite build system has been essentially unmaintained for years and it seems unlikely anyone is going to put in the work to get it running on Py3 (and the single digit number of users will likely just rely on whatever RH fork of Py2 exists in the near-term), I'd just not bother. There's much more than just syntax errors in that codebase and having it fail fast at least makes it clear it's unsupported.

You can try submitting a PR to https://github.com/w3c/csswg-apiclient and try and get @plinss to merge it upstream, but I'd suggest it's not worth the effort.

cclauss: https://github.com/w3c/csswg-apiclient/pull/2

cclauss:
@Ms2ger @gsnedders Your reviews please on what remains in this PR?

css/tools/w3ctestlib/Sources.py
```diff
@@ -3,6 +3,7 @@
 # Initial code by fantasai, joint copyright 2010 W3C and Microsoft
 # Licensed under BSD 3-Clause: <http://www.w3.org/Consortium/Legal/2008/03-bsd-license>
 
+from __future__ import print_function
```
Ms2ger: Defer to @gsnedders if this needs to land in https://github.com/gsnedders/w3ctestlib-git or somewhere else.

Ms2ger (commented):
LGTM besides one comment

css/tools/w3ctestlib/Sources.py
gsnedders: Doesn't need to land anywhere else.

gsnedders (approved):

PR #15149, opened 2019-01-30
Assert `selectedOptions` is `HTMLCollection` instance

shvaikalesh:


html/semantics/forms/the-select-element/select-selectedOptions.html
```diff
@@ -103,6 +109,13 @@
 }, ".selectedOptions without the 'multiple' attribute but " +
    "more than one selected option should return the last one");
 
+test(() => {
+  const select = document.getElementById("select-named-selected");
+
+  assert_equals(select.selectedOptions.constructor, HTMLCollection);
```
shvaikalesh: I don't use `instanceof` here to make sure `selectedOptions` is exactly `HTMLCollection` instance, not a subclass like `HTMLOptionsCollection`.

domenic (approved):

Issue #14862, opened 2019-01-15
/css/css-font-loading/idlharness.https.html is flaky in Chrome


foolip:
Discovered in https://github.com/web-platform-tests/wpt/pull/14629 and https://github.com/web-platform-tests/wpt/pull/14657, /css/css-font-loading/idlharness.https.html is flaky in Chrome:

|                                                            Subtest                                                             |         Results         |                                                   Messages                                                   |
|--------------------------------------------------------------------------------------------------------------------------------|-------------------------|--------------------------------------------------------------------------------------------------------------|
|                                                                                                                                | **ERROR: 3/5, OK: 2/5** |                                                                                                              |
| `idl_test setup`                                                                                                               | PASS                    |                                                                                                              |
| `FontFace interface: existence and properties of interface object`                                                             | PASS                    |                                                                                                              |
| `FontFace interface object length`                                                                                             | PASS                    |                                                                                                              |
| `FontFace interface object name`                                                                                               | PASS                    |                                                                                                              |
| `FontFace interface: existence and properties of interface prototype object`                                                   | PASS                    |                                                                                                              |
| `FontFace interface: existence and properties of interface prototype object's "constructor" property`                          | PASS                    |                                                                                                              |
| `FontFace interface: existence and properties of interface prototype object's @@unscopables property`                          | PASS                    |                                                                                                              |
| `FontFace interface: attribute family`                                                                                         | PASS                    |                                                                                                              |
| `FontFace interface: attribute style`                                                                                          | PASS                    |                                                                                                              |
| `FontFace interface: attribute weight`                                                                                         | PASS                    |                                                                                                              |
| `FontFace interface: attribute stretch`                                                                                        | PASS                    |                                                                                                              |
| `FontFace interface: attribute unicodeRange`                                                                                   | PASS                    |                                                                                                              |
| `FontFace interface: attribute variant`                                                                                        | PASS                    |                                                                                                              |
| `FontFace interface: attribute featureSettings`                                                                                | PASS                    |                                                                                                              |
| `FontFace interface: attribute variationSettings`                                                                              | FAIL                    | `assert_true: The prototype object must have a property "variationSettings" expected true got false`         |
| `FontFace interface: attribute display`                                                                                        | PASS                    |                                                                                                              |
| `FontFace interface: attribute status`                                                                                         | PASS                    |                                                                                                              |
| `FontFace interface: operation load()`                                                                                         | PASS                    |                                                                                                              |
[...]

PR #15170, opened 2019-01-30
[css-font-loading] Handle a fontFace.loaded promise rejection in idlharness test

foolip:
Fixes https://github.com/web-platform-tests/wpt/issues/14862.

### DIFF
--- a/css/css-font-loading/idlharness.https.html
+++ b/css/css-font-loading/idlharness.https.html
@@ -14,10 +14,15 @@ idl_test(
   idl_array => {
     idl_array.add_objects({
       Document: ['document'],
-      FontFace: ['new FontFace("family", "src")'],
-      FontFaceSetLoadEvent: ['new FontFaceSetLoadEvent("type")'],
+      FontFace: ['fontFace'],
+      FontFaceSetLoadEvent: ['fontFaceSetLoadEvent'],
       FontFaceSet: ['document.fonts'],
     });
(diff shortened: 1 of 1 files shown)


css/css-font-loading/idlharness.https.html
```diff
@@ -14,10 +14,15 @@
...
-      FontFaceSetLoadEvent: ['new FontFaceSetLoadEvent("type")'],
+      FontFace: ['fontFace'],
+      FontFaceSetLoadEvent: ['fontFaceSetLoadEvent'],
       FontFaceSet: ['document.fonts'],
     });
+    self.fontFace = new FontFace("family", "src");
+    // The `fontFace.loaded` promise will be rejected, so handle that to
+    // avoid an unhandled promise rejection manifesting as a harness error.
+    self.fontFace.loaded.catch(() => {});
+    self.fontFaceSetLoadEvent = new FontFaceSetLoadEvent("type");
```
domenic: Why move this one?

domenic (approved):

css/css-font-loading/idlharness.https.html
foolip: Just for consistency, since the variable names are what will show up in the test names.

It does mean that the order matters, but I guessed that no implementation would support `FontFaceSetLoadEvent` without first supporting `FontFace`, so if the above throws it wouldn't result in less coverage in practice.

PR #15172, opened 2019-01-30
Get Edge version using PowerShell command

foolip:
Adapted from https://stackoverflow.com/a/52056711/191722.

With the Windows 10 VMs currently in the self-hosted agent pool for
Azure Pipelines, it returns "44.17763.1.0".

### DIFF
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -579,8 +579,12 @@ class Edge(Browser):
         raise NotImplementedError
 
     def version(self, binary=None, webdriver_binary=None):
-        return None
-
+        command = "(Get-AppxPackage Microsoft.MicrosoftEdge).Version"
+        try:
+            return call("powershell.exe", command).strip()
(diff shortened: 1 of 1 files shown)


tools/wpt/browser.py
```diff
@@ -579,8 +579,12 @@ def install_webdriver(self, dest=None, channel=None):
         raise NotImplementedError
 
     def version(self, binary=None, webdriver_binary=None):
-        return None
-
+        command = "(Get-AppxPackage Microsoft.MicrosoftEdge).Version"
+        try:
+            return call("powershell.exe", command).strip()
+        except subprocess.CalledProcessError:
```
foolip: From the failures it appears that OSError is thrown when powershell.exe doesn't even exist. That hasn't happened for safari because of the `webdriver_binary` check in place.

foolip: Catching `OSError` too worked.

gsnedders (approved):

PR #15191, opened 2019-01-31
Use get_host_info() for WebKit

cvazac:
The template `{{hosts[][www]}}` doesn't work when running in the WebKit CLI because www.localhost doesn't resolve.

cc @youennf

### DIFF
--- a/resource-timing/redirects.sub.html
+++ b/resource-timing/redirects.sub.html
@@ -9,11 +9,14 @@
 <script src="/resources/testharnessreport.js"></script>
 <script src="resources/webperftestharness.js"></script>
 <script src="resources/webperftestharnessextension.js"></script>
+<script src="/common/get-host-info.sub.js"></script>
 <script>
+const host = get_host_info();
+
 let iframe;
 const redirect_url = 'common/redirect.py';
 const url_prefix = redirect_url + '?location=/resource-timing/resources/';
-const https_url_prefix = redirect_url + '?location=https://{{hosts[][www]}}:{{ports[https][0]}}/resource-timing/resources/';
+const https_url_prefix = redirect_url + '?location=' + host.HTTPS_REMOTE_ORIGIN + '/resource-timing/resources/';
 function setup_iframe() {
     const iframe_content =
         '<link rel="stylesheet" href="/' + url_prefix + 'resource_timing_test0.css"></link>' +
--- a/resource-timing/resource_connection_reuse_mixed_content.html
+++ b/resource-timing/resource_connection_reuse_mixed_content.html
@@ -19,6 +19,10 @@ test_namespace('getEntriesByType');
 
 function setup_iframe() {
     iframe = document.getElementById('frameContext');
+
+    // Removes any entries from helper files (get-host-info.sub.js).
+    iframe.contentWindow.performance.clearResourceTimings()
+
     d = iframe.contentWindow.document;
     iframe.addEventListener('load', onload_test, false);
 }
--- a/resource-timing/resource_connection_reuse_mixed_content_redirect.html
+++ b/resource-timing/resource_connection_reuse_mixed_content_redirect.html
@@ -19,6 +19,10 @@ test_namespace('getEntriesByType');
 
 function setup_iframe() {
     iframe = document.getElementById('frameContext');
+
+    // Removes any entries from helper files (get-host-info.sub.js).
+    iframe.contentWindow.performance.clearResourceTimings()
+
     d = iframe.contentWindow.document;
     iframe.addEventListener('load', onload_test, false);
 }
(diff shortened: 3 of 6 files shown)


yoavweiss (approved):
LGTM, although it'd be good if WebKit fixed their infra to make this unnecessary

cvazac:
This will fix the following failures in webkit:
LayoutTests/imported/w3c/web-platform-tests/resource-timing/redirects.sub.html
LayoutTests/imported/w3c/web-platform-tests/resource-timing/resource_connection_reuse_mixed_content.html
LayoutTests/imported/w3c/web-platform-tests/resource-timing/resource_connection_reuse_mixed_content_redirect.html

jgraham (changes requested):
This is the third PR that changes syntax that's expected to work in order to handle the broken setup in the WebKit repo. I don't think this is a sustainable solution to the problem since there are hundreds of tests depending on the hostname substiution, and increasingly features that special-case `localhost`. 

I object to merging this until there's some kind of plan to fix WebKit rather than trying to covert the repository one test at a time 

cc @youennf

youennf:
I do not think that there is any short term plan to support [www] in WebKit. Requiring developers to change etc/hosts has a cost which is seen as greater than the benefit of running the few WPT tests that really require [www] or other names.

Some ideas to move this forward:
1. [www] use should be reserved to tests that require hierarchical domain names. This should make its use in WPT rather limited.
2. It would be nice to make WPT reviewers and test authors aware of that rule. It would also be nice to have linter rules. I would think that step 1 will improve the situation since authors might often look at other tests
3. Use of [port], http/https or get-host-info.sub.js is sufficient in most cases and should be encouraged.
4. hosts[www] is used in some tests as a replacement to ports[https][1]. Maybe we should beef up WPT infra to allow ports[htts][1].
5. The number of test files using 'www' (25 doing a grep plus a few plus a few test files using 'www' concatenation) is not large. WebKit contributors could envision to do step 1 in a burst instead of progressively.

@jgraham, I must say I do not precisely understand the rationale behind your objection.

I agree that this creates a burden.The cost is mostly on WebKit contributors anyway, not on other WPT contributors though I might underestimate it. Could you elaborate on the cost for non WebKit WPT contributors?

jgraham:
Telling all developers that they have to use a totally different syntax depending on whether they "really need" subdomains or not is offloading complexity onto developers in order to fix a WebKit-specific problem. That's not a good tradeoff; other browsers have mechanisms for dealing with this that don't involve all developers editing `/etc/hosts` e.g. Firefox and Servo have pref settings that act like a hostsfile for the case where the server is on the local domain. It might even be possible to set this up using a proxy autoconfig file to have the right behaviour.

In general we have been adding more features to allow testing across multiple subdomains because the interaction of features with the origin policy is a common source of serious interop bugs both in new features (e.g. related to site isolation, cookies, etc.) or legacy features (e.g. document.domain). Making it difficult/impossible for WebKit developers to run these tests is in itself bad, even before we consider the cost being externalised onto every other contributor by adding additional requirements to support WebKit not fixing their infrastructure here.

youennf:
> Making it difficult/impossible for WebKit developers to run these tests is in itself bad

Agreed. At some point, WebKit might indeed fix that issue as it becomes more and more costly to do that.
In the meantime, either WebKit will not run these tests at all (which seems to be what you are suggesting), or WebKit will spend time on updating them.

Again, understanding the exact cost externalized to non WebKit WPT contributors will help.
If the issue is that it eats review time, I think we can find solutions.
If the issue is that it makes life of original test authors/maintainers more difficult, we can try to make the changes smaller, see below.

In general, making the tests as atomic as possible is good.
Using [www] where it does matter makes it easier for browser developers to understand the exact purpose of the test (oh yes, this is about hierarchical URLs/ETLD+1 vs. this is HTTP/HTTPS downgrade vs. this is about cross-origin loads...).

If you look at the proposed changes, host.HTTPS_REMOTE_ORIGIN  is easier to understand than https://{{hosts[][www]}}:{{ports[https][0]}}.
The WPT infra could provide easier to use macros like {{https-remote-origin}} instead that would be computed by the WPT server based on some config files.
Or we could have {{hosts[][other]}} for instance.

rniwa:
> WebKit not fixing their infrastructure here.

This isn't so much about WebKit not fixing the infrastructure as much as the underlying system not having such a capability / facility that's available to users. Because we rely on CFNetwork to provide the networking layer on iOS and macOS, we're at the mercy of their capability to configure things. We don't have own DNS stack, etc... and we don't want to maintain one.

Also, there is definitely no way for users to override `/etc/hosts` in iOS yet it seems prudent that we support running WPT in iOS given iOS Safari has the highest market share of any browser engine after Chrome according to various sources.

We might be able to come up with some mechanism to overcome such challenges internally at Apple, but then WPT will be only runnable by Apple employees on iOS, which won't be great either.

youennf:
Closing this PR in favor of https://github.com/web-platform-tests/wpt/pull/16567 which is more lightweight.

PR #15198, opened 2019-02-01
Enabling older browsers values for KeyboardEvent.key in WebDriver tests

jimevans:
When propagating keyboard events, older browsers use different strings
for the "key" property of the KeyboardEvent object. This commit enables
the ability of the WebDriver tests for keyboard events to validate the
comparison to these older values in addition to the values used by
modern browsers. See the footnotes in the sections of the following:

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values


jimevans:
This PR isn't ready. I'll recreate it when I've gotten it right.

PR #15204, opened 2019-02-01
Enabling older browsers values for KeyboardEvent.key in WebDriver tests

jimevans:
When propagating keyboard events, older browsers use different strings
for the "key" property of the KeyboardEvent object. This commit enables
the ability of the WebDriver tests for keyboard events to validate the
comparison to these older values in addition to the values used by
modern browsers. See the footnotes in the sections of the following:

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values

### DIFF
--- a/webdriver/tests/perform_actions/key_events.py
+++ b/webdriver/tests/perform_actions/key_events.py
@@ -1,8 +1,9 @@
 # META: timeout=long
 
 import pytest
+import copy
 
-from tests.perform_actions.support.keys import ALL_EVENTS, Keys
+from tests.perform_actions.support.keys import ALL_EVENTS, Keys, ALTERNATIVE_KEY_NAMES
 from tests.perform_actions.support.refine import filter_dict, get_events, get_keys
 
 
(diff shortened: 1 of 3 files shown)


whimboo:
Please note that Chrome tests are failing. Please have a look at those results:
https://tools.taskcluster.net/groups/WC_kdXw3RSyOhxUdOfK9qA/tasks/axZqmRlzRs2fdtU0NOeEOQ/runs/0/logs/public%2Flogs%2Flive.log#L1278

jimevans:
@whimboo I’m having a challenge seeing how this change caused those failures. The logs indicate an issue around the Meta modifier key, which I never referenced nor modified in my change. I’d really like to understand what happened here, since I have no control over the fact that Chrome seems to be reporting `OSLeft` and `OSRight` as the key values instead of the expected `MetaLeft` and `MetaRight`.

andreastt:
AIUI these are stability tests so it means this change has made Chrome _unstable_, not that Chrome necessarily does the right thing. I don’t entirely understand _what_ part of the change would make Chrome unstable yet…

gsnedders:
The only failure I see there is:

## Slow tests ##
|                      Test                     | Result | Longest duration (ms) | Timeout (ms) |
|-----------------------------------------------|--------|-----------------------|--------------|
| `/webdriver/tests/perform_actions/pointer.py` | `OK`   | `20489`               | `25000`      |

i.e., `/webdriver/tests/perform_actions/pointer.py` is a bit too close to the timeout for comfort. Just bump up the timeout on that?

andreastt (approved):
Thanks @gsnedders! It sounds like we should increase the timeout of that test to prevent this from cropping up again.

Otherwise, this patch looks good, so: r+wc

Issue #15174, opened 2019-01-30
Daily runs of Edge on Azure Pipelines


foolip:
This is like https://github.com/web-platform-tests/wpt/issues/14836, but for Edge on Windows 10.

@mustjab and @thejohnjansen have provided a self-hosted agent pool with 5 Window 10 VMs to make this possible.

foolip:
https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=5330 was the first full build, but it failed to get into wpt.fyi, seemingly because of `browser_version` missing from the JSON reports.

https://github.com/web-platform-tests/wpt/pull/15172 fixes that and https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=5461 is in progress with that fix. We'll see what the next error is :)

skeets1967:
Thanks for your time and help in the way you have

skeets1967:
I'm New to all of this

foolip:
https://staging.wpt.fyi/results/?sha=d11be4eae8 is the first sign of this maybe working. I uploaded wpt_report_10.json.gz from https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=5461 at https://staging.wpt.fyi/api/results/upload and changed the browser name to "edge" to get anything to show up.

@lukebjerring @Hexcles what to do about "edge_webdriver" as the product name?

skeets1967:
Sound like heaven. Man I can not enoughThank you
Sent from Yahoo Mail on Android 
 
  On Wed, Jan 30, 2019 at 6:20 PM, Philip Jägenstedt<notifications@github.com> wrote:   
https://staging.wpt.fyi/results/?sha=d11be4eae8 is the first sign of this maybe working. I uploaded wpt_report_10.json.gz from https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=5461 at https://staging.wpt.fyi/api/results/upload and changed the browser name to "edge" to get anything to show up.

@lukebjerring @Hexcles what to do about "edge_webdriver" as the product name?

lukebjerring:
I'm not sure why the distinction propagates into the product name, but is this something that we should be handling in general? Do we have an exhaustive list of these cases?

`edge_webdriver` can be cleanly mapped to an `edge` run with a `webdriver` label.

foolip:
[...]

PR #15207, opened 2019-02-01
Strip a "_webdriver" suffix from product name in RunInfo

foolip:
Context: https://github.com/web-platform-tests/wpt/issues/15174#issuecomment-459311676

### DIFF
--- a/tools/wptrunner/wptrunner/wpttest.py
+++ b/tools/wptrunner/wptrunner/wpttest.py
@@ -85,7 +85,13 @@ class RunInfo(dict):
         if rev:
             self["revision"] = rev
 
-        self["product"] = product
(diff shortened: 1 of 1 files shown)


jgraham (changes requested):
Apart from being a gross hack, this doesn't allow us to do things like change the expected results of infrastructure tests depending on which backend we are using, which is something we really want to be able do.

foolip:
@jgraham I tried to work out if this RunInfo is used at runtime by just following code around but couldn't connect all the dots.

I just ran https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=5678 to run the infrastructure/ tests which have edge_webdriver in the expectations metadata. And it failed, so contrary to what I hoped tweaking `RunInfo` affects more stuff than what goes into the reports.

PR #15208, opened 2019-02-01
Replicating "Alt" key workaround for IE WebDriver tests

jimevans:
When the Alt key is pressed on Windows, it activates the menu bar for
the focused application. This means that subsequent operations must
first dismiss the menu before being able to carry on. This commit
for Internet Explorer duplicates the workaround for Alt keys from
the test_special_key_sends_keydown subtest in key_events.py to the
test_modifier_key_sends_correct_events subtest.

### DIFF
--- a/webdriver/tests/perform_actions/key_events.py
+++ b/webdriver/tests/perform_actions/key_events.py
@@ -31,7 +31,7 @@ def test_modifier_key_sends_correct_events(session, key_reporter, key_chain, key
     code = ALL_EVENTS[event]["code"]
     value = ALL_EVENTS[event]["key"]
 
-    if (session.capabilities["browserName"] == 'internet explorer'):
+    if session.capabilities["browserName"] == "internet explorer":
         key_reporter.click()
         session.execute_script("resetEvents();")
     key_chain \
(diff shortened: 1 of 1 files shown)


webdriver/tests/perform_actions/key_events.py
```diff
@@ -31,6 +31,9 @@ def test_modifier_key_sends_correct_events(session, key_reporter, key_chain, key
     code = ALL_EVENTS[event]["code"]
     value = ALL_EVENTS[event]["key"]
 
+    if (session.capabilities["browserName"] == 'internet explorer'):
```
andreastt: Style nits (sorry, but it will break the linter):

- Drop `(` and `)` for grouping.
- Use double quotes for the second string.

andreastt (approved):
r+wc

Looks good, feel free to merge when fixed up.

webdriver/tests/perform_actions/key_events.py
jimevans: Fixed in next commit. Also fixed it in the other place where the same construct was used.

PR #15210, opened 2019-02-01
[css-grid] Adds test for checking if a button has grid context applied properly

brob:
Test currently passes in Firefox, but fails in chromium browsers. 

This is in response to this twitter thread about grid in buttons with interop issues: https://twitter.com/chrisdhanaraj/status/1091425085305454592

CC: @gregwhitworth @argyleink

### DIFF
--- a/css/css-grid/grid-model/grid-button-001.html
+++ b/css/css-grid/grid-model/grid-button-001.html
@@ -5,7 +5,7 @@
 <link rel="help" href="https://drafts.csswg.org/css-grid-1/#grid-containers">
 <style>
   .grid { display: grid; grid-template-columns: 100px 200px; border: 2px solid purple; box-sizing: border-box; }
-  .span { border: 1px dashed green; box-sizing: border-box; }
+  span { border: 1px dashed green; box-sizing: border-box; }
 </style>
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
--- a/css/css-grid/grid-model/grid-button-001.html
+++ b/css/css-grid/grid-model/grid-button-001.html
@@ -4,8 +4,8 @@
 <link rel="author" title="Bryan Robinson" href="bryanlrobinson@gmail.com">
 <link rel="help" href="https://drafts.csswg.org/css-grid-1/#grid-containers">
 <style>
-  .grid { display: grid; grid-template-columns: 100px 200px; border: 2px solid red; box-sizing: border-box; /* for visual */}
-  span { border: 1px solid green; box-sizing: border-box; /* for visual */ }
+  .grid { display: grid; grid-template-columns: 100px 200px; border: 2px solid purple; box-sizing: border-box; }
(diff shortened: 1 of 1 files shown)


argyleink (approved):
💀🤘

css/css-grid/grid-model/grid-button-001.html
```diff
@@ -0,0 +1,26 @@
...
+<link rel="author" title="Bryan Robinson" href="bryanlrobinson@gmail.com">
+<link rel="help" href="https://drafts.csswg.org/css-grid-1/#grid-containers">
+<style>
+  .grid { display: grid; grid-template-columns: 100px 200px; border: 2px solid red; box-sizing: border-box; /* for visual */}
+  span { border: 1px solid green; box-sizing: border-box; /* for visual */ }
+</style>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<button class="grid"><span class="item1">item 1</span> <span class="item2">item 2</span></button>
```
mrego: You can use `check-layout-th.js` and things like `data-expected-with` and even `data-offset-x` to be sure things are working as expected.

css/css-grid/grid-model/grid-button-001.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
+<title>Grid items work inside a button</title>
+<meta name="assert" content="When a button is set to display: grid, its children should flow into its grid cells">
+<link rel="author" title="Bryan Robinson" href="bryanlrobinson@gmail.com">
+<link rel="help" href="https://drafts.csswg.org/css-grid-1/#grid-containers">
+<style>
+  .grid { display: grid; grid-template-columns: 100px 200px; border: 2px solid red; box-sizing: border-box; /* for visual */}
```
mrego: Please avoid using `red` if it's visible when the test runs properly (it can be confusing). Better use `black` or any other color here.
Nit: Also the comments `/* for visual */` are not really needed, you can remove them.

mrego (commented):
Thanks for the test, some inline comments to make it better. :wink:

css/css-grid/grid-model/grid-button-001.html
brob: Oh, that's handy! One question that I couldn't quite find in the source myself: Is there a way to do custom messages in the test using checkLayout() or is the default messaging good enough? Either way, it's super nice not to have to write all that myself (update incoming)

css/css-grid/grid-model/grid-button-001.html
brob: Makes sense! I wasn't thinking

css/css-grid/grid-model/grid-button-001.html
mrego: Mmm, I'm not sure if there are anything like that. Usually the default message is clear enough.

mrego (approved):
Thanks for the test.

Issue #15159, opened 2019-01-30
infrastructure/reftest/size-ref.html flaky on Azure/Edge


foolip:
In https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=5357 (before https://github.com/web-platform-tests/wpt/pull/14300 was landed) /infrastructure/reftest/size-ref.html failed.

https://gist.github.com/foolip/d4b8dc5e236ebcf0cf10b28bf50684da is the edge.tbpl.log.

Pasting that into https://hg.mozilla.org/mozilla-central/raw-file/tip/layout/tools/reftest/reftest-analyzer.xhtml reveals that the output was "innerWidth x innerHeight: 600x601" instead of the expected "innerWidth x innerHeight: 600x600":
![download](https://user-images.githubusercontent.com/498917/51982795-418e0980-2497-11e9-9b11-92f0fd1000b0.png)

The size of this image is also 600x601.

foolip:
Interesting, this might be flaky.

In https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=5605 for https://github.com/web-platform-tests/wpt/pull/15196 the test unexpectedly passed.

gsnedders:
It's not that surprising, resizing is a bit weird and complicated and OS dependent to start with, especially if we're dealing with anything where there isn't an integral scaling factor (though I'd be surprised if this were the case for what we're running on Azure?).

gsnedders:
One thing that might be worth trying is moving the window to some constant location before we start trying to resize it?

gsnedders:
@foolip is there any way to just trigger a large number of runs now and see if it's at all stable?

foolip:
Sure, you can push a branch that uses parallel jobs like all_edge but each does the same steps.

gsnedders:
@foolip do you know if this flakiness went away after the above PR?

foolip:
@gsnedders no, I haven't checked.

foolip:
We're no longer running this configuration.

PR #15222, opened 2019-02-02
Try and improve #15159 by moving the window to a deterministic location

gsnedders:
Especially on systems where we don't have an integral scaling factor between CSS pixels and device pixels we can end up with our ability to resize the window depending on the location of the window.

See #15159.


gsnedders:
@foolip how do I set `run_infrastructure_win10`?

foolip:
@gsnedders see screenshot in https://github.com/web-platform-tests/wpt/pull/15201

gsnedders:
Bizarrely this seems to make Chrome fail/error a bunch of other tests.

foolip:
Maybe they were already failing on master?

gsnedders:
CI failure due to #15232.

jgraham (approved):

PR #15244, opened 2019-02-04
[Azure Pipelines] Enable much more verbose TBPL logs for full Edge runs

foolip:
Builds with this setup have been fairly unreliable:
https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=5766
https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=5774
https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=5841

In these builds, a job took 6 hours and was then canceled, causing the
overall builds to fail. The logs don't have enough clues about why test
execution might have stalled.


foolip (commented):
I'll go ahead and self-review and merge this so that it's on master for the next daily run. It'll help debug if that fails, as it has fairly regularly so far.

Issue #15183, opened 2019-01-31
Tests for dynamic import in HTML event handlers are not consistent with latest spec


jonco3:
The HTML spec is changing so that HTML event handler functions are given a null [[ScriptOrModule]] and this changes the behaviour of import():

https://github.com/whatwg/html/pull/4302

The tests need to be updated to follow this change.

PR #15251, opened 2019-02-05
Update dynamic import tests now HTML event handler functions do not have [[ScriptOrModule]] set

jonco3:
Fixes https://github.com/web-platform-tests/wpt/issues/15183

This updates paths/expected nonce behaviour now some methods of loading are not based off the current script.

cc @domenic


domenic (approved):
Great work, thank you!!

PR #15254, opened 2019-02-05
Set a viewport meta tag to a intersection observer test

hiikezoe:
To avoid the content is zoomed out and to avoid expanding the layout viewport
to the minimum-scale size.

On mobile without specifying the viewport meta tag, the target element,
positioned at the center of the contents, is initially visible because the
content is zoomed out to some extent.  But the test supposes that the target
element is scrolled out of view in the first place and is moved into the
root margin after setting scrollLeft to 100 in step0().  Thus all subsequent
test cases don't work as expected.

### DIFF
--- a/intersection-observer/root-margin.html
+++ b/intersection-observer/root-margin.html
@@ -1,4 +1,5 @@
 <!DOCTYPE html>
+<meta name="viewport" content="width=device-width,initial-scale=1,minimum-scale=1">
 <script src="/resources/testharness.js"></script>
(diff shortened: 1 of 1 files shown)


hiikezoe:
This test  fail on [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1524220) on Android.
I wonder whether Chrome team does run wpt on their CI servers.

@szager-chromium would you mind reviewing this?  Unfortunately there is no reasonable person to review this in Mozilla at this moment (the change itself is trivial though).

PR #15264, opened 2019-02-06
Add insertion tests for the new changes with the "deferred steps queue"

nox:
Those tests expose differences between Safari and the spec on one side, and the rest of the world on the other side.

\[These are for https://github.com/whatwg/dom/pull/732 and https://github.com/whatwg/html/pull/4354. - Anne]

### DIFF
--- a/dom/nodes/Node-appendChild-script-and-button-from-div.html
+++ b/dom/nodes/Node-appendChild-script-and-button-from-div.html
@@ -1,9 +1,6 @@
 <!doctype html>
 <meta charset=utf-8>
 <title>Node.appendChild: inserting script and button from a div</title>
-<link rel=help href="https://dom.spec.whatwg.org/#dom-node-appendChild">
-<link rel=help href="https://dom.spec.whatwg.org/#concept-node-insert">
-<link rel=help href="https://github.com/whatwg/dom/issues/575">
 <script src=/resources/testharness.js></script>
 <script src=/resources/testharnessreport.js></script>
 <div id="log"></div>
@@ -11,24 +8,20 @@
 <body>
 <form id="form"></form>
 <script>
-var form = document.getElementById("form");
-
-var script = document.createElement("script");
-var scriptRan = false;
-script.textContent = `
-    assert_equals(button.form, form, "button does not have a form owner");
-    scriptRan = true;
+let button = null;
+let buttonForm = null;
+test(() => {
+  const form = document.getElementById("form");
+  const script = document.createElement("script");
+  script.textContent = `
+    buttonForm = button.form;
 `;
-
-var button = document.createElement("button");
-
-var div = document.createElement("div");
-div.appendChild(script);
-div.appendChild(button);
-
-
-assert_false(scriptRan, "script ran");
-form.appendChild(div);
-assert_true(scriptRan, "script has not run");
-done();
+  button = document.createElement("button");
+  const div = document.createElement("div");
+  div.appendChild(script);
+  div.appendChild(button);
+  assert_equals(buttonForm, null);
+  form.appendChild(div);
+  assert_equals(buttonForm, form);
+});
 </script>
--- a/dom/nodes/Node-appendChild-script-and-button-from-div.html
+++ b/dom/nodes/Node-appendChild-script-and-button-from-div.html
@@ -1,0 +1,34 @@
+<!doctype html>
+<meta charset=utf-8>
+<title>Node.appendChild: inserting script and button from a div</title>
+<link rel=help href="https://dom.spec.whatwg.org/#dom-node-appendChild">
+<link rel=help href="https://dom.spec.whatwg.org/#concept-node-insert">
+<link rel=help href="https://github.com/whatwg/dom/issues/575">
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<div id="log"></div>
+</head>
+<body>
+<form id="form"></form>
+<script>
+var form = document.getElementById("form");
+
+var script = document.createElement("script");
+var scriptRan = false;
+script.textContent = `
+    assert_equals(button.form, form, "button does not have a form owner");
+    scriptRan = true;
+`;
+
+var button = document.createElement("button");
+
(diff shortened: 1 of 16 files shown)


nox:
The most interesting one is the button one because it doesn't even rely on a document fragment.

dom/nodes/Node-appendChild-script-and-button-from-div.html
```diff
@@ -0,0 +1,33 @@
...
+</head>
+<body>
+<form id="form"></form>
+<script>
+var form = document.getElementById("form");
+
+var script = document.createElement("script");
+var scriptRan = false;
+script.textContent = `
+    assert_equals(button.form, null, "button already has a form owner");
```
annevk: So this is basically assuming the maybe-WebKit model whereby the script is executed as part of the insertion steps rather than after as part of its own loop?

dom/nodes/Node-appendChild-script-and-div-from-fragment.html
```diff
@@ -0,0 +1,27 @@
...
+<script src=/resources/testharness.js></script>
+<script src=/resources/testharnessreport.js></script>
+<div id="log"></div>
+<script>
+var script = document.createElement("script");
+var div = document.createElement("div");
+
+var scriptRan = false;
+script.textContent = `
+    assert_equals(div.parentNode, null, "div already has a parent");
```
annevk: If the script executes as you go, why would the parentNode not be the DocumentFragment? Do we remove all before insertion somehow?

annevk (commented):
Thank you for writing these! I have some questions and I'm also not sure we should merge these given we're unclear on the direction of the specification still. I'd be somewhat okay if we put .tentative. in the name, but it's not even clear these are the results we want...

dom/nodes/Node-appendChild-script-and-button-from-div.html
nox: Yes, script preparation can happen when an element [becomes connected](https://html.spec.whatwg.org/multipage/scripting.html#script-processing-model:becomes-connected).

dom/nodes/Node-appendChild-script-and-div-from-fragment.html
nox: Yes, in step 4 of the [insert algorithm](https://dom.spec.whatwg.org/#concept-node-insert).

dom/nodes/Node-appendChild-script-and-button-from-div.html
```diff
@@ -0,0 +1,32 @@
...
+var scriptRan = false;
+script.textContent = `
+    assert_equals(button.form, form, "button does not have a form owner");
+    scriptRan = true;
+`;
+
+var button = document.createElement("button");
+
+var div = document.createElement("div");
+div.appendChild(script);
```
annevk: Maybe assert_false on _scriptRan_ here?

dom/nodes/Node-appendChild-script-and-custom-from-fragment.html
```diff
@@ -0,0 +1,35 @@
...
+var customConstructed = false;
+class CustomElement extends HTMLElement {
+    constructor() {
+        super();
+        customConstructed = true;
+    }
+}
+
+var df = document.createDocumentFragment();
+df.appendChild(script);
```
annevk: Same, assert_false on scriptRan?

dom/nodes/Node-appendChild-script-and-div-from-fragment.html
```diff
@@ -0,0 +1,27 @@
...
+var div = document.createElement("div");
+
+var scriptRan = false;
+script.textContent = `
+    assert_equals(div.parentNode, script.parentNode, "div does not have same parent as script");
+    scriptRan = true;
+`;
+
+var df = document.createDocumentFragment();
+df.appendChild(script);
```
annevk: And again...

annevk (commented):
I'd like to see a test where multiple scripts are inserted at once through a fragment and one of the scripts that executes earlier modifies another script. It'd be good to know that the modification takes effect, as per the current proposed text.

And a similar test where the subsequent script is removed from the tree to see it gets properly aborted.

dom/nodes/Node-appendChild-script-and-button-from-div.html
nox: Added an assertion.

dom/nodes/Node-appendChild-script-and-div-from-fragment.html
nox: Added an assertion here and in other tests too.

annevk:
FYI: I'm in the process of rewriting these tests so they work with newish wpt requirements.

dom/nodes/Node-appendChild-text-in-script.html
```diff
@@ -10,15 +10,19 @@
...
+const happened = [];
+test(() => {
+  const script = document.getElementById("script");
+  const df = document.createDocumentFragment();
+  df.appendChild(new Text("happened.push('t1');"));
+  df.appendChild(new Text("happened.push('t2');"));
+  assert_array_equals(happened, []);
+  script.appendChild(df);
+  assert_array_equals(happened, ["t1", "t2"]);
+  // At this point it's already executed so further motifications are a no-op
```
guest271314: "motifications" shuld be "notifications". 

These tests do apply to `<script type="module">`, where the exported value will always be the same no matter how many different scipts import the value, correct?

annevk:
These tests are cleaned up now, but pass conditions probably still need adjusting.

annevk:
@domfarolino you might want to repurpose these for #44308. There's a couple other interesting cases these tests touch.

domfarolino:
Closing this out in favor of https://github.com/web-platform-tests/wpt/pull/44658 which is [being reviewed now](https://github.com/web-platform-tests/wpt/pull/44658#pullrequestreview-1896207153). We'd like to land it soon, with Git co-author credit to @nox for working on this suite of tests initially.

Thanks a lot!

PR #13873, opened 2018-11-02
[css-overscroll-behavior] Parsing, serialization

ewilligers:
Spec:
https://drafts.csswg.org/css-overscroll-behavior/#property-index

Blink currently fails the proposed serialization
tests for the overscroll-behavior shorthand, as it
is not following the following guideline:

https://drafts.csswg.org/cssom/#serializing-css-values
If component values can be omitted or replaced with a shorter
representation without changing the meaning of the value,
omit/replace them.

PR #15287, opened 2019-02-08
[css-overscroll-behavior] Computed values

ewilligers:
The computed values of the overscroll-behavior-x and overscroll-behavior-y
properties are the keywords as specified.
https://drafts.csswg.org/css-overscroll-behavior/#property-index


ewilligers:
I consolidated the -x and -y tests into a common file, as suggested in https://github.com/web-platform-tests/wpt/pull/13873

theres-waldo (approved):

PR #15289, opened 2019-02-08
[css-images] Computed values are as specified

ewilligers:
For keyword properties such as object-fit,
computed value is specified keyword.
https://drafts.csswg.org/css-images-4/


ewilligers:
@plinss please review

tabatkins (approved):

PR #15293, opened 2019-02-08
[css-flexbox] Inheritance, initial values

ewilligers:
Test that properties do not inherit.
Test that they have the expected initial values.
https://drafts.csswg.org/css-flexbox-1/#property-index

### DIFF
--- a/css/css-flexbox/inheritance.html
+++ b/css/css-flexbox/inheritance.html
@@ -4,31 +4,33 @@
 <meta charset="utf-8">
 <title>Inheritance of CSS Flexible Box Layout properties</title>
 <link rel="help" href="https://drafts.csswg.org/css-flexbox-1/#property-index">
-<meta name="assert" content="Properties inherit or not according to the spec.">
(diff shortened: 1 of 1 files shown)


ewilligers:
Note that 3 initial value tests fail as the spec is inconsistent with browsers and CSS Box Alignment.
https://github.com/w3c/csswg-drafts/issues/3636

ewilligers:
I have update the initial value tests so they pass.

@fantasai please review.

ewilligers:
@mrego  Please review.

css/css-flexbox/inheritance.html
```diff
@@ -0,0 +1,37 @@
...
+</head>
+<body>
+<div id="container">
+  <div id="target"></div>
+</div>
+<style>
+  #container, #target {
+    display: flex;
+  }
+</style>
```
mrego: Nit: Kind of weird place for the `<style>` tag, I'd put it in the `<head>`.

mrego (approved):
LGTM.

PR #15297, opened 2019-02-08
Use CSS style to set color of text and background.

fred-wang:
Currently, many tests unnecessarily rely on the MathML mathcolor/mathbackground
attributes to be implemented. This commit use the equivalent CSS color and background properties instead, except for the test that actually checks support for the MathML color attributes (relations/html5-tree/color-attributes-1.html).


Ms2ger (approved):

rwlbuis:
LGTM as well.

fred-wang:
Thanks.

PR #15301, opened 2019-02-08
Import WPT fraction tests from WebKit

fred-wang:
See https://bugs.webkit.org/show_bug.cgi?id=194436
Tests have been reformated and rewritten.

The following are excluded:
- Tests have equivalent (and more reliable testing) in WPT.
- Tests that verify crashs/asserts, hangs or invalid markup.


mathml/presentation-markup/fractions/frac-color-001-ref.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
```
Ms2ger: I'd call this `*-notref.html`

mathml/presentation-markup/fractions/frac-linethickness-003-ref.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
```
Ms2ger: Ditto

rwlbuis (approved):

Ms2ger (approved):

Issue #14564, opened 2018-12-17
Many PRs blocked due to failing Taskcluster stability checks


foolip:
With https://github.com/web-platform-tests/wpt/issues/14165 resolved, Taskcluster was made a required check again on Dec 12. Since then, many PRs have become blocked on a Taskcluster failure:

wpt-chrome-dev-stability failures:
- https://github.com/web-platform-tests/wpt/pull/14400
- https://github.com/web-platform-tests/wpt/pull/14531

wpt-firefox-nightly-stability failures:
- https://github.com/web-platform-tests/wpt/pull/14420
- https://github.com/web-platform-tests/wpt/pull/14469
- https://github.com/web-platform-tests/wpt/pull/14536

wpt-chrome-dev-stability + wpt-firefox-nightly-stability failures:
- https://github.com/web-platform-tests/wpt/pull/13906
- https://github.com/web-platform-tests/wpt/pull/14417
- https://github.com/web-platform-tests/wpt/pull/14519
- https://github.com/web-platform-tests/wpt/pull/14551

Some issues filed:
- https://github.com/web-platform-tests/wpt/issues/14433
- https://github.com/web-platform-tests/wpt/issues/14434

Authors of those PRs, please comment here to say whether the failure was related to your changes. If we have too many PRs blocked for unrelated reasons, we'll have to reconsider.

jgraham:
All but one seems to be an actual stability check failure, so if there's a problem here it isn't with Taskcluster as such, but with the stability checking.

foolip:
The trouble would be if we have so much pre-existing flakiness that an unreasonable number of PRs would be blocked on pre-existing flakiness. I don't know where to draw the line, filed this issue to start collecting cases of it happening.

If it's too bad, I think we'd have to get our hands on data about which tests are most frequently affected and which of those are most flaky, and fix those.

jgraham:
Fixing flaky tests seems like a good idea for many reasons.

frivoal:
For #14551, I've restarted it a couple of times by pushing a rebased commit, and I believe that it was once reported as firefox flakiness, and twice as Chrome's. But I've never been able to reproduce locally, and while instability in a basic non animated layout reftest isn't totally impossible if the browser engine happens to have some kind of non deterministic bug, it is pretty unlikely. So I don't know what's causing this, but I'm pretty sure it isn't the test itself.

foolip:
[...]

PR #15305, opened 2019-02-08
Add testdriver support for generate_test_report WebDriver command.

paulmeyer90:
Also adds Reporting API tests which utilize this command.

### DIFF
--- a/reporting/bufferSize.html
+++ b/reporting/bufferSize.html
@@ -5,6 +5,7 @@
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <script src="/resources/testdriver.js"></script>
+<script src="/resources/testdriver-vendor.js"></script>
 <script>
   async_test(async function(test) {
     for (i = 0; i != 110; ++i)
--- a/reporting/bufferSize.html
+++ b/reporting/bufferSize.html
@@ -1,0 +1,25 @@
+<!DOCTYPE HTML>
+<meta charset=utf-8>
+<title>Reporting: Buffer size</title>
+<link rel="author" title="Paul Meyer" href="paulmeyer@chromium.org">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/resources/testdriver.js"></script>
+<script>
+  async_test(async function(test) {
+    for (i = 0; i != 110; ++i)
+      await test_driver.generate_test_report("" + i);
+
+    var observer = new ReportingObserver(function(reports) {
+      test.step(function() {
+        assert_equals(reports.length, 100);
+        for(i = 0; i != 100; ++i) {
+          assert_equals(reports[i].body.message, "" + (i + 10));
+        }
+      });
+
+      test.done();
+    }, {buffered: true});
+    observer.observe();
+  }, "Buffer size");
+</script>
--- a/reporting/bufferSize.html
+++ b/reporting/bufferSize.html
@@ -7,12 +7,15 @@
 <script src="/resources/testdriver.js"></script>
 <script src="/resources/testdriver-vendor.js"></script>
 <script>
+  // Test the buffer size (100) of ReportingObserver.
   async_test(async function(test) {
     for (i = 0; i != 110; ++i)
       await test_driver.generate_test_report("" + i);
 
     var observer = new ReportingObserver(function(reports) {
       test.step(function() {
+        // Only (the most recent) 100 reports should be observed, even though
+        // 110 were buffered.
         assert_equals(reports.length, 100);
         for(i = 0; i != 100; ++i) {
           assert_equals(reports[i].body.message, "" + (i + 10));
--- a/reporting/disconnect.html
+++ b/reporting/disconnect.html
@@ -1,0 +1,24 @@
+<!DOCTYPE HTML>
+<meta charset=utf-8>
+<title>Reporting: Disconnect</title>
+<link rel="author" title="Paul Meyer" href="paulmeyer@chromium.org">
+<script src="/resources/testharness.js"></script>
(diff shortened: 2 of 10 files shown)


jgraham (dismissed):
TestDriver additions now require an RFC in https://github.com/web-platform-tests/rfcs/ to allow fedback from across the project.

In this case it's unclear what specifies the underlying WebDriver API since it isn't part of the WebDriver spec. So your RFC should link to the relevant spec text.

paulmeyer90:
I have opened an RFC issue here: https://github.com/web-platform-tests/rfcs/issues/7

paulmeyer90:
> TestDriver additions now require an RFC in https://github.com/web-platform-tests/rfcs/ to allow fedback from across the project.
> 
> In this case it's unclear what specifies the underlying WebDriver API since it isn't part of the WebDriver spec. So your RFC should link to the relevant spec text.

RFC has passed

paulmeyer90:
@jgraham are you satisfied with this change now?

jgraham:
I think the testdriver changes are OK, but I shouldn't review the actual tests :)

reporting/bufferSize.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE HTML>
+<meta charset=utf-8>
+<title>Reporting: Buffer size</title>
+<link rel="author" title="Paul Meyer" href="paulmeyer@chromium.org">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/resources/testdriver.js"></script>
+<script src="/resources/testdriver-vendor.js"></script>
+<script>
+  async_test(async function(test) {
```
LukeZielinski: Could you add a brief comment describing each test (eg: kind of like the ones in the Disconnect and Order tests)

paulmeyer90: Done!

reporting/bufferSize.html
```diff
@@ -0,0 +1,29 @@
...
+<script>
+  // Test the buffer size (100) of ReportingObserver.
+  async_test(async function(test) {
+    for (i = 0; i != 110; ++i)
+      await test_driver.generate_test_report("" + i);
+
+    var observer = new ReportingObserver(function(reports) {
+      test.step(function() {
+        // Only (the most recent) 100 reports should be observed, even though
+        // 110 were buffered.
```
LukeZielinski: nit: so we're expecting reports 10-110? Can the loop go through that range instead of adding "10" to each expected report? I think that might be a bit clearer.

LukeZielinski (approved):
Lgtm, thanks Paul.

paulmeyer90:
@dcreager could you please take a look at the specific tests I have here?

reporting/bufferSize.html
paulmeyer90: I could, but then I would have to decrement 10 from the index each time instead. I don't think I agree that that would be clearer.

LukeZielinski: Yep, my fault, I missed that detail. Thanks.

foolip:
The failing Taskcluster run on this PR is because testdriver.js was modified and tests including it were run. If they weren't stable, well then the job fails. Here are the Chrome failures:

|                            Test                           |                          Subtest                          |            Results            |                                 Messages                                 |
|-----------------------------------------------------------|-----------------------------------------------------------|-------------------------------|--------------------------------------------------------------------------|
| `/event-timing/event-timing-timingconditions.html`        | `Event Timing only times certain types of trusted event.` | **FAIL: 2/10, PASS: 8/10**    | `assert_equals: Should only observe one entry: []. expected 1 but got 0` |
| `/pointerevents/pointerevent_pointerout_pen.html`         |                                                           | **ERROR: 1/10, OK: 9/10**     |                                                                          |
| `/pointerevents/pointerevent_pointerout_pen.html`         | `pointerout event received`                               | **PASS: 9/10, MISSING: 1/10** |                                                                          |
| `/speech-api/SpeechSynthesis-pause-resume.tentative.html` |                                                           | **OK: 1/10, TIMEOUT: 9/10**   |                                                                          |
| `/speech-api/SpeechSynthesis-pause-resume.tentative.html` | `speechSynthesis.pause() and resume() state and events`   | **FAIL: 1/10, TIMEOUT: 9/10** | `assert_unreached: error event Reached unreachable code;Test timed out`  |

And Firefox:

|                                    Test                                    |                   Subtest                   |            Results            |                 Messages                |
|----------------------------------------------------------------------------|---------------------------------------------|-------------------------------|-----------------------------------------|
| `/pointerevents/pointerevent_setpointercapture_inactive_button_mouse.html` |                                             | **OK: 8/10, TIMEOUT: 2/10**   |                                         |
| `/pointerevents/pointerevent_setpointercapture_inactive_button_mouse.html` | `setPointerCapture + inactive button state` | **FAIL: 8/10, MISSING: 2/10** | `InvalidPointerId: Invalid pointer id.` |

@paulmeyer90 if none of these are due to changes that you made, which it doesn't seem like, then just let me know when this is ready to merge. I'll file bugs about the remaining failures and admin merge this. Sorry about the inconveniece, this is tracked by https://github.com/web-platform-tests/wpt/issues/14564 but I don't have a good fix for it.

paulmeyer90:
@foolip None of those appear to be related to my change, since all the tests added by my change pass, and none of those failing tests use the new API that I added (and those are the only changes in this PR).

Hexcles:
Force-merged as per the comment above.

foolip:
@Hexcles can you file the bugs as well?

Hexcles:
Filed #15486 for pointer events, as well as #15485 for the more general question of skipping `testdriver.js` altogether (same as `testharness.js`).

I'm looking into the event-timing and speech-api flakes. Those don't seem to be flaky on master.

Issue #15137, opened 2019-01-29
Make ./wpt run [single test] quicker


gsnedders:
The goal here is to make `time ./wpt run --headless --no-pause chrome infrastructure/assumptions/html-elements.html` take under 4s on my local laptop (MacBook Pro (Retina, 15-inch, Mid 2014), i7-4870HQ, 16GB RAM, macOS 10.14).

gsnedders:
To give some rough ideas about perf:

cProfile claims that `testrunner.run` is only 28% of a `--no-manifest` run. Other significant things are setting up the venv (12%) and requirements (15%), `wptrunner.get_loader` (34%).

Also looking at VMProf (which seems not to cope with threading locally?! so percentages are in no way comparable), it appears that the more expensive part of `wptrunner.get_loader` is not the JSON parsing, (`manifest.load_and_update` is 23%) but constructing all the items (`manifest.TypeData.load_all` is 55%).

So it seems our biggest enemy is constructing all the manifest item instances, assuming the manifest update takes no time (which is, of course, unrealistic, but it can be much closer to zero, and if we need to redesign the manifest format again we need to know where the bottlenecks are after update).

jgraham:
There exists a (some what fragile) optimistation to only construct instances on demand, but we are presumably not using it in this case.

jgraham:
Right, if we insist on iterating all tests then obviously we don't use the optimisation. But the <code>[TypeData](https://searchfox.org/mozilla-central/source/testing/web-platform/tests/tools/manifest/manifest.py#46)</code> class has the ability to only convert from the JSON representation to the object representation on demand. I imagine in this case we aren't using it because the test matching assumes that we have access to the full list of tests and paths which currently requires us to actually construct all the objects.

gsnedders:
Vague to-do from the various things I've looked into in the past week:

 * Make sure the lazy item object creation optimisation applies in the jsshell test where it should already apply (cc/ @Ms2ger)
 * Apply the lazy item object creation optimisation when we know what tests we're running (e.g., the single test case)
 * Ensure we don't create objects if we don't need to for a no-op manifest update (this probably means extending the mtime cache)
 * Make it possible to do partial manifest updates (this is currently hard in the reftest case, where paths can cease being a test and become a node in the graph)
[...]

Issue #12874, opened 2018-09-06
Tasks going OOM on Taskcluster


gsnedders:
e.g., https://tools.taskcluster.net/groups/f7409aGOTa6OJLwvyw7igw/tasks/N0HriXt3TH62lxQhwftRlg/details went OOM (in the job for https://github.com/web-platform-tests/wpt/pull/12380)

The suite ends with:

```
[2m[34m 9:39.10(B[m [33mTEST_START(B[m: /html/infrastructure/urls/resolving-urls/query-encoding/location.sub.html?encoding=utf8
[2m[34m 9:59.22(B[m [31mTEST_END(B[m: TIMEOUT, expected OK
[2m[34m 9:59.55(B[m [33mWARNING(B[m u'runner_teardown': ()
[2m[34m 9:59.57(B[m [34mINFO(B[m STDERR: ['/home/test/web-platform-tests/_venv/bin/chromedriver', '--port=4444', '--url-base=/']
[2m[34m 9:59.60(B[m [33mWARNING(B[m Failure during init Traceback (most recent call last):
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/testrunner.py", line 195, in init
    self.browser.start(group_metadata=group_metadata, **self.browser_settings)
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/browsers/chrome.py", line 83, in start
    self.server.start(block=False)
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/webdriver_server.py", line 49, in start
    self._run(block)
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/webdriver_server.py", line 63, in _run
    self._proc.run()
  File "/home/test/web-platform-tests/_venv/lib/python2.7/site-packages/mozprocess/processhandler.py", line 760, in run
    self.proc = self.Process([self.cmd] + self.args, **args)
  File "/home/test/web-platform-tests/_venv/lib/python2.7/site-packages/mozprocess/processhandler.py", line 114, in __init__
    universal_newlines, startupinfo, creationflags)
  File "/usr/lib/python2.7/subprocess.py", line 711, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1235, in _execute_child
    self.pid = os.fork()
OSError: [Errno 12] Cannot allocate memory

[2m[34m 9:59.60(B[m [31mERROR(B[m Traceback (most recent call last):
[...]

PR #15308, opened 2019-02-09
Refactor code to simplify manifest items and their creation

gsnedders:


tools/manifest/item.py
```diff
@@ -108,6 +88,10 @@ def id(self):
     def meta_flags(self):
         return set(urlparse(self.url).path.rsplit("/", 1)[1].split(".")[1:-1])
 
+    @property
+    def url_base(self):
+        return "/"
```
gsnedders: This is blatantly wrong. But also currently doesn't break anything (uh, we should have tests!).

tools/manifest/item.py
```diff
@@ -52,12 +37,7 @@ def id(self):
 
     @property
     def meta_flags(self):
-        return set(self.source_file.meta_flags)
-
-    @property
-    def path(self):
-        """The test path relative to the test_root"""
-        return self.source_file.rel_path
+        return set()
```
jgraham: Is the belief that this isn't used anywhere? If so can we add a warning or something if we don't want to remove it?

gsnedders: Yes. AFAICT, it is literally only used by `ManifestItem.https` and nowhere else, and we don't need to deal with that for cases where we don't have a URL. I guess `https` should really move to `URLManifestItem`, and `meta_flags` made private?

tools/wptrunner/wptrunner/wpttest.py
```diff
@@ -149,12 +149,12 @@ def update_metadata(self, metadata=None):
...
         timeout = cls.long_timeout if manifest_item.timeout == "long" else cls.default_timeout
         protocol = "https" if hasattr(manifest_item, "https") and manifest_item.https else "http"
-        return cls(manifest_item.source_file.tests_root,
+        return cls(manifest_file.tests_root,
                    manifest_item.url,
                    inherit_metadata,
                    test_metadata,
                    timeout=timeout,
-                   path=manifest_item.source_file.path,
+                   path=os.path.join(manifest_file.tests_root, manifest_item.path),
```
jgraham: You could have a method on the manifest to return this

gsnedders: That seems a bit weird to me, at least.

tools/manifest/item.py
gsnedders: Hah, seemingly this is actually a constant! (We always set it to `/` and `SourceFile` expands the full URL.) And only used by `.url`, so we can just drop this.

gsnedders:
So, despite this mostly being done as refactoring to do other work for #15137, this has made the single-test case (`time ./wpt run --binary=`which chromium` --headless --no-pause chrome infrastructure/assumptions/html-elements.html`) 20% faster on my desktop computer.

tools/wptrunner/wptrunner/testloader.py
```diff
@@ -504,7 +504,7 @@ def iter_tests(self):
             manifest_items = self.chunker(manifest_items)
 
         for test_type, test_path, tests in manifest_items:
-            manifest_file = manifests_by_url_base[iter(tests).next().url_base]
+            manifest_file = manifests_by_url_base["/"]
```
gsnedders: This is obviously wrong. However, given `item.url_base` was previously just the constant `"/"`, this isn't a regression. I presume @jgraham will say I need to fix this code though. )

tools/manifest/item.py
jgraham: Note that it's set by the manifest and doenstream projects (notably gecko and servo) have manifests which set values other than `/` e.g. https://searchfox.org/mozilla-central/source/testing/web-platform/mozilla/meta/MANIFEST.json#1768

jgraham: (That might not mean this code wasn't useless, just checking you're aware of where the url base is used).

tools/wptrunner/wptrunner/testloader.py
jgraham: You need to fix this code. Or change the iterator to produce something different like `(manifest, source_file, tests)`.

tools/wptrunner/wptrunner/wpttest.py
```diff
@@ -313,32 +313,28 @@ class TestharnessTest(Test):
...
     def __init__(self, tests_root, url, inherit_metadata, test_metadata,
                  timeout=None, path=None, protocol="http", testdriver=False,
-                 jsshell=False, scripts=None):
+                 jsshell=False):
         Test.__init__(self, tests_root, url, inherit_metadata, test_metadata, timeout,
                       path, protocol)
 
         self.testdriver = testdriver
         self.jsshell = jsshell
-        self.scripts = scripts or []
```
Ms2ger: I do need *some* way to get these; they're used in https://searchfox.org/mozilla-central/rev/cb7faaf6b4ad2528390186f1ce64618dea71031e/js/src/tests/jstests.py#404

tools/wptrunner/wptrunner/testloader.py
gsnedders: So, we only get this right when loading the JSON, not when actually creating the manifest.

tools/wptrunner/wptrunner/wpttest.py
gsnedders: @jgraham told me it was unused! :)

Guess we need to add it to the JSON, then.

jgraham: Well it's not that expensive to read from the file. But if the goal is that we don't read any test file data except metadata when the test file hasn't been updated, then yes.

gsnedders: > But if the goal is that we don't read any test file data except metadata when the test file hasn't been updated, then yes.

We shouldn't need to read any test file data to just run the test.

gsnedders: So I've fixed this by putting it in the manifest… but I fear this means we're now gonna be broken until the MANIFEST is regenerated from scratch given it won't contain this until it's regenerated…

jgraham: We can bump the manifest version number to force it to be regenerated. We should probably also make sure that all the manifest upload things regenerate from scratch every time.

gsnedders:
Can't figure out why the `test_wpttest.py` tests are failing; will need to dig in deeper next week.

tools/wptrunner/wptrunner/wpttest.py
gsnedders: Following on from discussion on IRC, I've re-added `tests_root` to items (yay memory consumption!) and made it possible for them to create `SourceFile` objects (albeit behind a private property) to fetch the `scripts` from there if the `_extras` doesn't already contain it.

gsnedders: And, to be clear, when we next bump the MANIFEST version we should eliminate all of that.

gsnedders:
The one remaining failure is in `test_list_tests_missing_manifest`. This appears to be caused by us passing `--tests tools/wpt/tests` into the `wpt run` call in that test, and this hella confuses things. We end up generating a manifest for the git repo root but then plenty of other places aren't aware we're doing this.

gsnedders:
Given @Hexcles mentioned memory consumption of manifest generation in https://github.com/web-platform-tests/wpt/issues/12874#issuecomment-465794369:

On Linux/x86_64, on this branch, manifest generation from scratch consumes max 487MB, clean regeneration 202MB.

`master`, by way of comparison requires max 1408MB and 290MB respectively.

Hexcles:
That is definitely promising! I also tested the memory footprint of manifest before (~1.5G from scratch) and it is indeed a problem for folks running tests on VMs (or Taskcluster workers, from forks of this repo) with <=4G memory.

We should also try to free the memory ASAP. How much memory does the manifest take when we are running the tests (without manifest generation)? I assume it'd take take less than generating the manifest, and if so, it'd be great to make sure the delta is freed when the generation is completed.

gsnedders:
And the commit I've just pushed here makes it go down to max 165MB and 187MB respectively (the delta I presume is caused by it loading the old manifest and therefore having that in memory too).

But this does now mean we aren't holding on to any file beyond the point at which we're directly processing that file.

jgraham (commented):
THis looks great. Let me double check it doesn't break gecko before approving.

jgraham:
It seems like this broke gecko by calling git unconditionally https://treeherder.mozilla.org/#/jobs?repo=try&author=james%40hoppipolla.co.uk&collapsedPushes=420652&fromchange=551ebec95f9b6ad9b5ac3d8812ee04b61d9024c3&selectedJob=230362281

jgraham (changes requested):
Need to ensure that `git` isn't required when updating with `--work`.

gsnedders:
> Need to ensure that `git` isn't required when updating with `--work`.

That's fixed now.

tools/manifest/item.py
```diff
@@ -38,39 +22,26 @@ def __new__(cls, name, bases, attrs, **kwargs):
...
-    @property
-    def https(self):
-        flags = self.meta_flags
-        return ("https" in flags or "serviceworker" in flags)
-
     def key(self):
         """A unique identifier for the test"""
         return (self.item_type, self.id)
 
     def meta_key(self):
```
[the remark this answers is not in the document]
gsnedders: I… don't think so? But I don't really have the confidence here. We represent items as sets in various places which means we rely on the hashes to some extent, at least.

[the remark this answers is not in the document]
jgraham: Sure, but unless we are (still) using the hash to determine whether a test updated or not then we can probably remove this method. Not for this PR anyway.

jgraham (approved):

foolip:
@gsnedders I've been running a script to regenerate manifests for all tagged commits locally and compare with what can be downloaded from GitHub releases. Starting with merge_pr_15308 there are a lot that have an added slash at the beginning of paths. I'm not sure why that happened, was that ever the intended behavior before, during or after this change?

foolip:
@gsnedders using "dom/nodes/Document-doctype.html" as a test path that's been around since 2014, it looks like the leading slash was originally not there in the manifest and at some point was added. From which commit should it be there, when doing a full rebuild of the manifest?

gsnedders:
@foolip not sure; it shouldn't be there (and isn't today).

foolip:
@gsnedders alright, I'll work on the assumption that it should never have been there. So far it looks like it's there for some manifests created using an incremental update, but not if you recreate the manifest from scratch on the same commit.

PR #15311, opened 2019-02-10
[css-scroll-snap] scroll-snap-type computed value

ewilligers:
Spec: https://drafts.csswg.org/css-scroll-snap-1/

Blink/Safari fail by reporting the scroll-snap-type computed value
as "x proximity" instead of "x".

### DIFF
--- a/css/css-scroll-snap/parsing/scroll-snap-align-computed.html
+++ b/css/css-scroll-snap/parsing/scroll-snap-align-computed.html
@@ -1,0 +1,25 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>CSS Scroll Snap: getComputedValue().scrollSnapAlign</title>
+<link rel="help" href="https://drafts.csswg.org/css-scroll-snap-1/#propdef-scroll-snap-align">
+<meta name="assert" content="scroll-snap-align computed value is as specified.">
+
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script src="/css/support/computed-testcommon.js"></script>
+</head>
+<body>
+<div id="target"></div>
+<script>
+test_computed_value("scroll-snap-align", "none");
+test_computed_value("scroll-snap-align", "start");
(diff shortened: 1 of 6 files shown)


ewilligers:
Note the following two principles in [CSSOM](https://drafts.csswg.org/cssom/#serializing-css-values):

> If component values can be omitted or replaced with a shorter representation without changing the meaning of the value, omit/replace them.

> If either of the above syntactic translations would be less backwards-compatible, do not perform them.
 
Blink and Safari are the only browsers that support scroll-snap-type values like "x" and "x proximity", and they give "x proximity" as the computed value, so expecting "x proximity" would be more backwards-compatible.

ewilligers:
@tabatkins please review

css/css-scroll-snap/parsing/scroll-snap-align-computed.html
```diff
@@ -0,0 +1,25 @@
...
+<body>
+<div id="target"></div>
+<script>
+test_computed_value("scroll-snap-align", "none");
+test_computed_value("scroll-snap-align", "start");
+test_computed_value("scroll-snap-align", "end");
+test_computed_value("scroll-snap-align", "center");
+
+test_computed_value("scroll-snap-align", "start none");
+test_computed_value("scroll-snap-align", "center end");
```
tabatkins:
```suggestion
test_computed_value("scroll-snap-align", "center end");
test_computed_value("scroll-snap-align", "start start", "start");
```

tabatkins (changes requested):

tabatkins:
> Blink and Safari are the only browsers that support scroll-snap-type values like "x" and "x proximity", and they give "x proximity" as the computed value, so expecting "x proximity" would be more backwards-compatible.

Back-compat is about pages depending on the behavior, not browsers happening to have done it a particular way in the past. (That's colloquially called "bugwards compat". ^_^) This is new enough that I doubt anyone's depending on our too-long serializations, so we should continue to follow shortest-serialization principle until shown evidence otherwise.

tabatkins (approved):

PR #15313, opened 2019-02-10
[css-ui] Parsing of user-select

ewilligers:
https://drafts.csswg.org/css-ui-4/#content-selection
user-select supports keywords: auto | text | none | contain | all


svgeesus (approved):
Yes, simple, straightforward

frivoal:
Well, except that's not what the spec says. The computed value of auto is not supposed to be auto.

frivoal:
Opened https://github.com/web-platform-tests/wpt/issues/15456 to follow up.

PR #15083, opened 2019-01-25
added calc-min-height.html reference/calc-min-height-ref.html calc-ne…

TalbotG:
…sting.html calc-nesting-002.html

Those 3 tests and reference come from 
chromium.googlesource.com/chromium/src/+/c825d655f6aaf73484f9d56e9012793f5b9668cc/third_party/WebKit/LayoutTests/css3/calc/
and are unique in the sense that they were not shared or coming from webkit repository

PR #14987, opened 2019-01-22
Added 2 new CSS values tests

TalbotG:

PR #15320, opened 2019-02-10
Values3 gt pr16

TalbotG:
This pull request adopts the corrections requested in pull request 15083
https://github.com/web-platform-tests/wpt/pull/15083


frivoal (commented):
Reviewed, and these tests seem correct according [to the spec](https://www.w3.org/TR/css-values-4/#calc-serialize), but that same spec says:

> This section is still under discussion.

Editors (@tabatkins @fantasai) should we hold off having tests for this until the discussion concludes, or is having test helpful regardless?

(same question as in https://github.com/web-platform-tests/wpt/pull/14987)

frivoal (approved):
Checked with @tabatkins and he confirmed it's worth having tests for that part of the specification despite the warning.

PR #15326, opened 2019-02-11
Use IDL mixin syntax for pointerevents GlobalEventHandlers

plehegar:
This matches
  https://github.com/w3c/pointerevents/pull/275


lukebjerring:
FYI - While there's nothing wrong with fast-tracking this change manually via this PR, once the spec change is live, it will automatically be scraped by tidoust/reffy-reports

plehegar:
oh. Feel free to close it if that's better then. Thank you for letting me know.

PR #15328, opened 2019-02-11
HTML: noopener keyword is tokenized

annevk:
See https://github.com/whatwg/html/pull/4079.


annevk:
It seems this test was missed when the standard got updated.

domenic (approved):

PR #15329, opened 2019-02-11
Don't use mozdownload to download firefox

jgraham:
Instead use the same URL that's used by mozilla.org. This hopefully reduces the
problem where mozdownload tries to download a file that doesn't exist yet.
Note that we are still using mozdownload for geckodriver

### DIFF
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -5,9 +5,12 @@ import shutil
 import stat
 import subprocess
 import tempfile
+import urlparse
 from abc import ABCMeta, abstractmethod
 from datetime import datetime, timedelta
 from distutils.spawn import find_executable
+
+import requests
 
 from utils import call, get, untar, unzip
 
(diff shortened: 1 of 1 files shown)


jgraham:
https://github.com/web-platform-tests/wpt/issues/13274

Ms2ger (approved):

tools/wpt/browser.py
```diff
@@ -93,48 +96,64 @@ def platform_string_geckodriver(self):
...
-        scraper = {
-            "nightly": "daily",
-            "beta": "release",
-            "stable": "release"
+        import mozinstall
+
+        product = {
+            "nightly": "firefox-nightly-latest-ssl",
+            "beta": "firefox-beta-latest-ssl",
+            "stable": "firefox-beta-latest-ssl"
```
whimboo: @jgraham I think what you really want here for stable is https://download.mozilla.org/?product=firefox-latest-ssl&os=osx&lang=en-US. Otherwise you aren't using the release but beta version on stable.

tools/wpt/browser.py
```diff
@@ -93,48 +96,64 @@ def platform_string_geckodriver(self):
...
+        if content_disposition:
+            filenames = re.findall("filename=(.+)", content_disposition)
+            if filenames:
+                filename = filenames[0]
+
+        if not filename:
+            filename = urlparse.urlsplit(resp.url).path.rsplit("/", 1)[1]
+
+        if not filename:
+            filename = "firefox.tar.bz2"
```
whimboo: So we don't run those tests on other platforms than Linux?

PR #15334, opened 2019-02-11
Fix shebangs and executable bits for a bunch of files

LukeZielinski:
Mostly adding missing executable bits or removing unnecessary ones. In a few cases, missing shebangs were added.

The motivation for this is to be able to cleanly import these files into the Chromium repo (where presubmit checks exist to ensure that shebangs and executable bits line up correctly).


LukeZielinski:
Let me know if there are objections to this. The other solution is to avoid importing some of these files, or adding them to internal whitelists, but this seems cleaner.

Thanks!

Hexcles:
This looks good except the `third_party` bits. Files in that directory are directly imported from third parties, and changes will be overwritten next time when we upgrade these deps.

In particular, a lot of inconsistencies are in hyper/h2. Could you send a PR to https://github.com/python-hyper to fix those? Also cc @dhdavvie who works/worked on H2 stuff.

jugglinmike:
I'm surprised that the linter is aware of those files at all. @gsnedders you you think we should skip `tools/third_party/`?

Hexcles:
@jugglinmike it's the Chromium linter that complains about inconsistent permissions & shebangs.

gsnedders (changes requested):
Don't change `/tools/third_party`, per above comments.

jgraham (approved):

gsnedders (approved):

PR #15350, opened 2019-02-12
Fetch: use `t.step_timeout()` instead of bare `step_timeout()`

foolip:
This avoid a harness error in the case of failure:
https://wpt.fyi/results/fetch/stale-while-revalidate?run_id=6045131320852480&run_id=4797529354928128&run_id=6524192509919232&run_id=6492651545165824

### DIFF
--- a/fetch/stale-while-revalidate/stale-css.tentative.html
+++ b/fetch/stale-while-revalidate/stale-css.tentative.html
@@ -12,7 +12,7 @@ https://github.com/whatwg/fetch/pull/853
 
 async_test(t => {
   window.onload = t.step_func(() => {
-    step_timeout(() => {
+    t.step_timeout(() => {
       assert_equals(window.getComputedStyle(document.body).getPropertyValue('background-color'), "rgb(0, 128, 0)");
       var link2 = document.createElement("link");
       link2.onload = t.step_func(() => {
--- a/fetch/stale-while-revalidate/stale-image.tentative.html
+++ b/fetch/stale-while-revalidate/stale-image.tentative.html
@@ -18,7 +18,7 @@ See: https://html.spec.whatwg.org/#the-list-of-available-images
 
 async_test(t => {
   window.onload = t.step_func(() => {
-    step_timeout(() => {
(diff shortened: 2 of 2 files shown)


foolip:
From https://github.com/web-platform-tests/wpt/pull/15047

foolip:
Flakiness from Taskcluster logs:

## Chrome

### /fetch/stale-while-revalidate/stale-css.tentative.html ###
|            Subtest             |          Results           |                               Messages                              |
|--------------------------------|----------------------------|---------------------------------------------------------------------|
|                                | OK                         |                                                                     |
| `Cache returns stale resource` | **FAIL: 9/10, PASS: 1/10** | `assert_equals: expected "rgb(0, 128, 0)" but got "rgb(255, 0, 0)"` |

## Firefox

### /fetch/stale-while-revalidate/stale-css.tentative.html ###
|            Subtest             |            Results            |                                      Messages                                      |
|--------------------------------|-------------------------------|------------------------------------------------------------------------------------|
|                                | **OK: 9/10, TIMEOUT: 1/10**   |                                                                                    |
| `Cache returns stale resource` | **FAIL: 9/10, TIMEOUT: 1/10** | `assert_equals: expected "rgb(0, 128, 0)" but got "rgb(255, 0, 0)";Test timed out` |

The similarity suggests an issue with the test rather than Chrome or Firefox. @dtapuska I guess the test assumes that it will only be run once, because after that the thing is in cache again? Would adding a random token somewhere fix the problem?

youennf (approved):
I guess we should we recommend Test.step_timeout whenever feasible.

foolip:
Yes, but it's not clear what a lint could look like, since when a bare `step_timeout` is used the callback could actually be correctly wrapped directly or indirectly, or the callback could be so trivial that it won't ever throw. At least that's what I found when grepping a bit.

@lukebjerring, what remains to make harness errors require human confirmation to land? That would have avoided this case.

foolip:
Ignoring the flakiness as pre-existing and merging.

PR #15352, opened 2019-02-12
HTML: window.open("", "", "noreferrer")

annevk:
For https://github.com/whatwg/html/pull/4331.


html/browsers/the-window-object/apis-for-creating-and-navigating-browsing-contexts-by-name/resources/open-features-tokenization-boolean.js
```diff
@@ -0,0 +1,152 @@
...
+      `=${feature}`,
+      `,,${feature}`,
+      `,=, ${feature}`,
+      `\n=${feature}=`,
+      `\t${feature}`,
+      `\r,,,=${feature}`,
+      `\u000C${feature}`
+    ].forEach(variant => {
+      const win = window.open(windowURL, "", variant);
+      assert_equals(win, null, `"${variant}" should activate feature "${feature}"`);
```
zcorpan: This assertion (and the `assert_not_equals` later) is specific to the `noopener` feature. Although `noreferrer` means the same thing, the file name/function name suggests this is more generic than it is.

annevk: That's a good point, I'll rename.

zcorpan (approved):
LGTM % nit

cdumez:
BroadcastChannel :( I had to rewrite the test using localStorage instead so it would run in WebKit.

annevk:
@cdumez what is an alternative strategy? Also, any chance WebKit will add BroadcastChannel?

cdumez:
As I said, I used localStorage as an alternative. I guess cookies could work too. I am not aware of any plan to adopt BroadcastChannel.

annevk:
@cdumez ah right, sorry. Wouldn't you rather folks use `BroadcastChannel` than abuse `localStorage` for messaging between disconnected top-level browsing contexts?

cdumez:
I do not think people should do this in shipping code, and localStorage is good enough for testing purposes.

PR #15356, opened 2019-02-12
HTML: <form rel>

annevk:
For https://github.com/whatwg/html/pull/4332.


annevk:
I think this is good enough now. Implementers could add even more tests.

foolip (approved):

PR #15362, opened 2019-02-12
Fix flakiness with stale-while-revalidate tests.

dtapuska:
Using cookies didn't seem reliable in the wpt test runner when it was
re-run. Stop using cookies and use the server side stash instead
like the rest of the fetch tests do.

Increase timeout on staleness.


dtapuska:
@foolip PTAL. This fixes the tests encountered on the chrome bot.

dtapuska:
@Hexcles PTAL

Hexcles (approved):
LGTM (assuming you've done some sanity checks by of the test results)

FWIW, I think the cookie approach might work if you added a cleanup hook to clear the cookies, but I'd prefer using stash, too.

foolip (approved):
Thanks @dtapuska!

PR #15366, opened 2019-02-12
fix align start and end refs

gkatsev:
align start and align end are equivalent to align left and align right in LTR writing mode, which is the default. The size of the cue box doesn't change. This means that align:start should be on the left side of the video viewport and align:end should be on the right side of the video viewport. Browsers have implemented this correctly but the reference set a width which cause a mismatch between the two.

This was updated in this PR https://github.com/w3c/webvtt/pull/421 which was likely done after these tests were first written.


silviapfeiffer (approved):
Thanks, that finally brings the tests in line with the spec on this issue.

PR #15392, opened 2019-02-14
HTML: nested browsing contexts name attribute

annevk:


annevk:
It might be somewhat nice to extend this with cross-origin support. That'd require the frame to post the name to the parent and a state switcher in the listener that then does the various aspects of the test as you cannot get `name` cross-origin from a `WindowProxy` object. Does anyone think it'll matter if we already have same-origin coverage?

domenic:
Still working through my queue of (really cool) PRs from you this morning, but I'll chime in quickly to say that I'm afraid cross-origin vs. same-origin goes through rather different code paths, at least in Chrome with site-isolation. So, it would be good to add.

I certainly wouldn't require you to add that before merging any spec or test PRs. But, if you don't want to do the extra work yourself, it'd be good to file a follow-up issue when we do merge.

annevk:
That's a good point, let me try to revive what I had earlier and make it work, but might have to wait for tomorrow if I'm not done quickly.

html/browsers/windows/nested-browsing-contexts/resources/post-to-parent.html
```diff
@@ -0,0 +1,6 @@
+<script>
+const ident = new URL(location).searchParams.get("ident"),
+      post = () => parent.postMessage({ name, ident }, "*");
```
domenic: It's a little unclear that this is equivalent to `name: window.name` so I'd suggest changing it to make it explicit.

domenic (approved):

PR #15393, opened 2019-02-14
[Screen Orientation] Add locking not supported test

Johanna-hub:
For feedback

Test for first step of `Locking orientation algorithm`
          <li>If the <a>user agent</a> does not support locking the screen
          orientation, return a promise rejected with a
          <code>DOMException</code> whose name is
          <code>NotSupportedError</code> and abort these steps.


marcoscaceres (commented):
Sadly, I don't think we can test for this because we can't assure that one of the types is not supported:(if an implementation supports all the orientation types, this will always fail :( ).

marcoscaceres:
To be clear, this will only pass on (for example) desktop browsers.... but not a mobile browser.

PR #15422, opened 2019-02-15
Do not throw from taskcluster-run.py if wpt fails

Hexcles:
Previously when `wpt run` stability check fails, the wrapper script
taskcluster-run.py would throw an uncaught exception, which might be
confused with an infra error.

This change makes taskcluster-run.py return a non-zero exit code without
printing the traceback.


Hexcles:
This was an ask from @KyleJu and I think it's very reasonable.

jugglinmike (approved):

PR #15423, opened 2019-02-15
Added calc-serialization-002  getComputedStyle-border-radius-002 tests

TalbotG:
This replaces pull request 14987
https://github.com/web-platform-tests/wpt/pull/14987
which corrects and improves the mentioned tests.

### DIFF
--- a/css/css-values/calc-serialization-002.html
+++ b/css/css-values/calc-serialization-002.html
@@ -1,0 +1,95 @@
+<!DOCTYPE html>
+
+  <meta charset="UTF-8">
+
+  <title>CSS Values and Units: calc() serialization of a summation (complex)</title>
+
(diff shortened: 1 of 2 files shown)


frivoal (approved):

TalbotG:
I had a quick phone call with @fantasai friday Feb. 15th. From that short conversation, I am pretty sure that the spec needs to be clearer otherwise clarified. Also from that short conversation, I believe the original intent was that **serialization of specified value should not convert absolute length units (cm, in, pc, q, pt) into px unit**. Because the web author would certainly want to see his original CSS code remain similar, using the same absolute length units. 

Anyway, this is my understanding and interpretation. The spec editors have the final word on all this.

PR #15429, opened 2019-02-16
[Screen Orientation] Add orientation and associated angle test

Johanna-hub:
For feedback

Test to check that orientation and angle are assigned from the options in the screen orientation values table

### DIFF
--- a/screen-orientation/orientation-reading.html
+++ b/screen-orientation/orientation-reading.html
@@ -1,6 +1,8 @@
 <!DOCTYPE html>
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
+<script src="/resources/testdriver.js"></script>
+<script src="/resources/testdriver-vendor.js"></script>
 <script>
 test(() => {
   assert_true('type' in screen.orientation);
@@ -19,6 +21,39 @@ test(() => {
 
   assert_true(angle == 0 || angle == 90 || angle == 180 || angle == 270);
 }, "Test screen.orientation default values.");
+
+promise_test(async t => {
+  await document.documentElement.requestFullscreen();
+  let primaryOrientation;
+  await screen.orientation.lock("portrait-primary");
+  if (screen.orientation.angle === 0) {
+    primaryOrientation = "portrait-primary";
+  } else {
+    primaryOrientation = "landscape-primary";
+  }
+  if (primaryOrientation === "portrait-primary") {
+    await screen.orientation.lock("portrait-secondary");
+    assert_true(screen.orientation.angle == 180);
+    await screen.orientation.lock("landscape-primary");
+    const primary2angle = screen.orientation.angle;
+    assert_true(primary2angle == 90 || primary2angle == 270);
+    const secondary2angle = primary2angle === 90 ? 270 : 90;
+    await screen.orientation.lock("landscape-secondary");
+    assert_true(screen.orientation.angle == secondary2angle);
+  }
+  if (primaryOrientation === "landscape-primary") {
+    await screen.orientation.lock("landscape-secondary");
+    assert_true(screen.orientation.angle == 180);
+    await screen.orientation.lock("portrait-primary");
+    const primary2angle = screen.orientation.angle;
+    assert_true(primary2angle == 90 || primary2angle == 270);
+    const secondary2angle = primary2angle === 90 ? 270 : 90;
+    await screen.orientation.lock("portrait-secondary");
+    assert_true(screen.orientation.angle == secondary2angle);
+  }
+  screen.orientation.unlock();
+  return document.exitFullscreen();
+}, "Test the orientation and associated angle");
 
 test(() => {
   const type = screen.orientation.type;
--- a/screen-orientation/orientation-reading.html
+++ b/screen-orientation/orientation-reading.html
@@ -40,21 +40,21 @@ promise_test(async t => {
           secondaryOrientation2: "portrait-secondary"
         };
   await screen.orientation.lock(orientations.secondaryOrientation1);
-  assert_true(
+  assert_equals(
     screen.orientation.angle == 180,
-    "Secondary orientation 1 angle should be 180"
+    "Secondary orientation 1 angle must be 180"
   );
   await screen.orientation.lock(orientations.primaryOrientation2);
   assert_true(
     screen.orientation.angle == 90 || screen.orientation.angle == 270,
-    "Primary orientation 2 angle should be either 90 or 270"
+    "Primary orientation 2 angle must be either 90 or 270"
   );
   const primaryOrientation2Angle = screen.orientation.angle;
   const secondaryOrientation2Angle = primaryOrientation2Angle === 90 ? 270 : 90;
   await screen.orientation.lock(orientations.secondaryOrientation2);
-  assert_true(
+  assert_equals(
     screen.orientation.angle == secondaryOrientation2Angle,
-    "Secondary orientation 2 angle should be the one of 90 or 270 not set to primary orientation 2"
+    "Secondary orientation 2 angle must be the one of 90 or 270 not set to primary orientation 2"
   );
   screen.orientation.unlock();
   return document.exitFullscreen();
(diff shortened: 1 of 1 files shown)


screen-orientation/orientation-reading.html
```diff
@@ -20,6 +22,41 @@
   assert_true(angle == 0 || angle == 90 || angle == 180 || angle == 270);
 }, "Test screen.orientation default values.");
 
+promise_test(async t => {
+  await test_driver.bless("request full screen", () => {
+    return document.documentElement.requestFullscreen();
+  });
+  let primaryOrientation;
```
marcoscaceres: Could maybe instead do:

```JS
  await screen.orientation.lock("portrait-primary");
  const primaryOrientation = screen.orientation.angle === 0 
    ? "portrait-primary"
    : "landscape-primary";
```

But I'm a bit confused... if you do:

```
await screen.orientation.lock("portrait-primary");
```

Won't that assure the orientation is "portrait-primary"? 

And even if not assured, it's probably better to the value straight from "the horse's mouth"... i.e., `primaryOrientation` should be `screen.orientation.type`, right? Or am I missing something?

screen-orientation/orientation-reading.html
```diff
@@ -20,6 +22,41 @@
...
+  await screen.orientation.lock("portrait-primary");
+  if (screen.orientation.angle === 0) {
+    primaryOrientation = "portrait-primary";
+  } else {
+    primaryOrientation = "landscape-primary";
+  }
+  if (primaryOrientation === "portrait-primary") {
```
marcoscaceres: Note the code repetition here... and below from line 46... if we look at these two, what we actually are testing is "expected" angles.  

I would test this incrementally and get rid of the if statements - you can then split this into 4 tests. 

```JS
await screen.orientation.lock("portrait-primary");
const {angle, type} = screen.orientation;
assert_equals(type, "portrait-primary");
assert_true(angle === 90 || angle === 270);
```

And so on...

marcoscaceres (changes requested):
It's great how you've captured all the angles, but I think we simplify this a bit. The test is doing a lot of work - so if we split the test up, we can check the angles independently.

screen-orientation/orientation-reading.html
Johanna-hub: I wrote this to try to find out what `primaryOrientation` is - what we call in the screen orientation values table `primary orientation 1 `- and using this work out all the other angles as stated in the table. https://w3c.github.io/screen-orientation/#reading-the-screen-orientation
If when the screen is locked to `portrait-primary` the angle is 0 - then we know that this is `primary orientation 1`, so `secondary orientation 1` would then be `portrait-secondary` at 180 degrees.  `Primary orientation 2` would be `landscape primary `at either 90 or 270 and `secondary orientation 2` would be the remaining angle allocated to `landscape-secondary.`
If the screen is locked to `portrait-primary` and the angle is not 0, (it should either be 90 or 270) then `landscape-primary` would be `primary orientation 1 `at 0.  `Landscape-secondary` would be `secondary orientation 1` at 180.  `Portrait-primary` at 90 or 270 would be `primary-orientation 2` and so on.
This enables a more thorough check of the relationship of orientation to angle.

marcoscaceres: Ok, yes, I get the logic and rationale, which is perfect. 

But I'm still a bit confused (heh, I should read the spec!) about the following... if we do:

```JS
await screen.orientation.lock("portrait-primary");
```

Won't the `screen.orientation.type` be "portrait-primary"? 

If yes, then it follows that `primaryOrientation` will just be  "portrait-primary".

Johanna-hub: Yes, but we're testing for the angle not the type to decide primaryOrientation.  So once we are locked to portrait-primary - if the angle is 0 - then the primary orientation is "portrait-primary".

If the angle is 90 or 270, we know that the primary orientation is "landscape-primary" and portrait-primary is a rotation from that.

Am I confused though??

screen-orientation/orientation-reading.html
```diff
@@ -26,36 +26,39 @@
...
+  await screen.orientation.lock(orientations.primaryOrientation2);
+  assert_true(
+    screen.orientation.angle == 90 || screen.orientation.angle == 270,
+    "Primary orientation 2 angle should be either 90 or 270"
+  );
+  const primaryOrientation2Angle = screen.orientation.angle;
+  const secondaryOrientation2Angle = primaryOrientation2Angle === 90 ? 270 : 90;
+  await screen.orientation.lock(orientations.secondaryOrientation2);
+  assert_true(
+    screen.orientation.angle == secondaryOrientation2Angle,
```
marcoscaceres: assert_equals here

screen-orientation/orientation-reading.html
```diff
@@ -20,6 +22,44 @@
...
+          primaryOrientation2: "landscape-primary",
+          secondaryOrientation2: "landscape-secondary"
+        }
+      : {
+          secondaryOrientation1: "landscape-secondary",
+          primaryOrientation2: "portrait-primary",
+          secondaryOrientation2: "portrait-secondary"
+        };
+  await screen.orientation.lock(orientations.secondaryOrientation1);
+  assert_true(
```
marcoscaceres: assert_equals here too

marcoscaceres (changes requested):
Looking really good. Two little things to fix.

screen-orientation/orientation-reading.html
```diff
@@ -20,6 +22,44 @@
...
+    screen.orientation.angle == 180,
+    "Secondary orientation 1 angle should be 180"
+  );
+  await screen.orientation.lock(orientations.primaryOrientation2);
+  assert_true(
+    screen.orientation.angle == 90 || screen.orientation.angle == 270,
+    "Primary orientation 2 angle should be either 90 or 270"
```
marcoscaceres:
```suggestion
    "Primary orientation 2 angle must be either 90 or 270"
```

marcoscaceres: The other “should” should be “must”, per the spec.

screen-orientation/orientation-reading.html
```diff
@@ -40,21 +40,21 @@
           secondaryOrientation2: "portrait-secondary"
         };
   await screen.orientation.lock(orientations.secondaryOrientation1);
-  assert_true(
+  assert_equals(
```
marcoscaceres: this should be:
```
assert_equals(screen.orientation.angle, 180, "Secondary orientation 1 angle must be 180") 
```
i.e., assert_equals takes 3 arguments.

marcoscaceres (commented):
assert_equals need updating :)

screen-orientation/orientation-reading.html
Johanna-hub: Sorry!! Wasn't thinking properly!

marcoscaceres: No problem. What reviews are for 🥳

screen-orientation/orientation-reading.html
```diff
@@ -53,7 +54,8 @@
   const secondaryOrientation2Angle = primaryOrientation2Angle === 90 ? 270 : 90;
   await screen.orientation.lock(orientations.secondaryOrientation2);
   assert_equals(
-    screen.orientation.angle == secondaryOrientation2Angle,
+    screen.orientation.angle,
+    secondaryOrientation2Angle,
     "Secondary orientation 2 angle must be the one of 90 or 270 not set to primary orientation 2"
   );
```
marcoscaceres: The description here doesn’t seem to match the test very well... something doesn’t read well about it. Could you please rephrase it?

marcoscaceres (changes requested):
One little thing

screen-orientation/orientation-reading.html
Johanna-hub: I've given it a go, see what you think!

marcoscaceres: Ah! much better!

marcoscaceres (approved):

marcoscaceres (approved):

PR #15455, opened 2019-02-19
[html] Add test that cancelAnimationFrame cancels a pending animation frame callback

birtles:
See https://github.com/whatwg/html/issues/4359


birtles:
I confirmed that this test fails in Firefox and passes in Chrome.

jakearchibald (approved):
lgtm

PR #15460, opened 2019-02-19
HTML: window.focus() sans browsing context

annevk:


foolip (approved):

zcorpan:
I think it would be good to add a test for `blur()` also (to assert that it doesn't throw or crash the browser I guess).

annevk:
You mean add a line somewhere that invokes it? Where? I guess that would also need a comment to indicate why it's there...

zcorpan:
How about a separate test for `blur()` to assert that it does nothing, with or without a browsing context. When I write this I think it makes sense to have the spec just say that it does nothing, also. At least in Chromium it seems it does nothing.

annevk:
Dealing with `blur()` is https://github.com/whatwg/html/issues/4381 now.

PR #15463, opened 2019-02-19
Add missing tests for ReadableStream

ricea:
* Test the specific rejection returned by calling it.return() twice (to avoid
  being confused by assertions).
* Verify that it.next() after it.return() fails.
* Verify that it.return() unlocks the stream synchronously.
* Add brand checks for getIterator(), locked and desiredSize.
* Verify that canceling tee branches in reverse order works.
* Test that erroring a teed stream errors both branches, even when they
  are not waiting for chunks.
* Verify that controller.close() clears the list of pending reads

### DIFF
--- a/streams/readable-streams/async-iterator.any.js
+++ b/streams/readable-streams/async-iterator.any.js
@@ -342,3 +342,12 @@ promise_test(async t => {
   await it.return();
   return promise_rejects(t, new TypeError(), it.next(), 'next() should reject');
 }, 'calling next() after return() should reject');
+
+for (const preventCancel in [false, true]) {
+  test(t => {
+    const rs = new ReadableStream();
+    rs.getIterator({ preventCancel }).return();
+    // The test passes if this line doesn't throw.
+    rs.getReader();
+  }, `return() should unlock the stream synchronously when preventCancel = ${preventCancel}`);
+}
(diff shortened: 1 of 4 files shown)


streams/readable-streams/async-iterator.any.js
```diff
@@ -338,3 +335,10 @@ promise_test(async () => {
   assert_equals(readResult.value, 3, 'should read remaining chunk');
   await reader.closed;
 }, 'Acquiring a reader and reading the remaining chunks after partially async-iterating a stream with preventCancel = true');
+
+promise_test(async t => {
+  const rs = new ReadableStream();
+  const it = rs.getIterator();
+  await it.return();
+  return promise_rejects(t, new TypeError(), it.next(), 'next() should reject');
+}, 'calling next() after return() should reject');
```
domenic: Should we also test what happens if you do these two tests (return-return and return-next) without awaiting the return? I think that is also determined, right?

streams/readable-streams/tee.any.js
```diff
@@ -184,6 +185,32 @@ promise_test(() => {
 
 }, 'ReadableStream teeing: canceling both branches should aggregate the cancel reasons into an array');
 
+promise_test(() => {
+
+  const reason1 = new Error('This little one\'s not worth the effort.');
+  const reason2 = new Error('Come, let me get you something.');
```
domenic: 💯

domenic (approved):
LGTM with potential addition

streams/readable-streams/async-iterator.any.js
ricea: Instead I added a test that the lock is released synchronously by return(), which is roughly equivalent and slightly improves coverage.

PR #15465, opened 2019-02-19
Move to running unittests on Py3.7 rather than 3.6

gsnedders:
This requires running on Xenial on Travis CI, as 3.7 isn't available Trusty.


jgraham:
If you want to change this we need to do it in TaskCluster rather than Travis, which is effectively a whole new PR.

PR #15477, opened 2019-02-20
[Azure Pipelines] Remove the wpt.fyi hook (no longer needed)

foolip:
As found in https://github.com/web-platform-tests/wpt.fyi/issues/1130,
this hook is no longer needed, check runs are created also for
scheduled builds on Azure Pipelines.


gsnedders (approved):

lukebjerring:
@foolip - we'll need to reinstate this, since there's no reliable way to tell whether "all the shards" succeeded from wpt.fyi. It should also be added to the affected-tests jobs.

PR #15482, opened 2019-02-20
Add some tests for the realms involved in creating new platform objects.

Ms2ger:


Ms2ger:
Depends on https://github.com/heycam/webidl/pull/635

WebIDL/ecmascript-binding/constructors.html
```diff
@@ -0,0 +1,93 @@
+<!DOCTYPE html>
+<meta charset=utf-8>
+<title>Realm for constructed objects</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div id=log></div>
+<script>
+function object_realm(dp) {
+  var url = DOMParser.prototype.parseFromString.call(dp, "x", "text/html").documentURI;
+  var file = url.slice(url.lastIndexOf("/") + 1);
```
domenic: It's really unclear why the documentURI would be derived from the this value, looking at the DOMParser spec :(. Maybe add a comment pointing to the spec, or if this is underspecified, open an issue and then add a comment pointing to the issue?

WebIDL/ecmascript-binding/constructors.html
```diff
@@ -0,0 +1,93 @@
...
+    }, "Subclass constructor in parent window");
+
+    test(function() {
+      var dp = new child.MyDOMParser();
+      assert_equals(Object.getPrototypeOf(dp), child.MyDOMParser.prototype);
+      assert_equals(object_realm(dp), "child window");
+    }, "Subclass constructor in child window");
+
+    test(function() {
+      class ForeignDOMParser extends child.DOMParser {}
```
domenic: Maybe "MyForeignDOMParser"

WebIDL/ecmascript-binding/constructors-support.html
```diff
@@ -0,0 +1,8 @@
+<!DOCTYPE html>
```
domenic: If you made this a .js file you could include it in both places and thus avoid some of the duplication of these declarations inside the test file.

WebIDL/ecmascript-binding/constructors-support.html
```diff
@@ -0,0 +1,8 @@
...
+<script>
+window.cons = function() {};
```
domenic: I think "badNewTarget" would be a clearer variable name here.

WebIDL/ecmascript-binding/constructors-support.html
```diff
@@ -0,0 +1,8 @@
...
+cons.prototype = 8;
+
+window.MyDOMParser = class extends DOMParser {}
+window.ForeignDOMParser = class extends parent.DOMParser {}
```
domenic: Clearer variable names might be "DOMParserSubclass" and "ForeignDOMParserSubclass"

domenic (approved):
LGTM with nits

WebIDL/ecmascript-binding/constructors-support.html
Ms2ger: Probably not worth it, because `ForeignDOMParserSubclass` needs to inherit from parent/child depending on context.

PR #15487, opened 2019-02-20
Test presence and content of customized template custom element content property

tomalec:
in constructor and connectedCallback

as there is an interoperability issue between Chromium and Firefox

### DIFF
--- a/custom-elements/builtin-coverage/template.html
+++ b/custom-elements/builtin-coverage/template.html
@@ -8,20 +8,19 @@
 <script>
   let contentInConstructor = undefined;
   let numberOfContentChildNodesInConnectedCallback = 0;
-  let MyTemplate = class extends HTMLTemplateElement {
+  const MyTemplate = class extends HTMLTemplateElement {
     constructor(...args) {
       super(...args);
       contentInConstructor = this.content;
     }
     connectedCallback(...args) {
-      super.connectedCallback && super.connectedCallback();
       numberOfContentChildNodesInConnectedCallback = this.content.childNodes.length;
     }
   }
   customElements.define('my-template', MyTemplate, {extends: 'template'});
 </script>
 </head>
-<body is="my-body">
+<body>
 <div id="container"><template is="my-template">foo</template></div>
 <script>
 test(() => {
--- a/custom-elements/builtin-coverage/template.html
+++ b/custom-elements/builtin-coverage/template.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html is="my-html">
+<html>
 <head>
 <meta charset="utf-8">
 <meta name="help" content="https://html.spec.whatwg.org/multipage/custom-elements.html#element-definition">
--- a/custom-elements/builtin-coverage/template.html
+++ b/custom-elements/builtin-coverage/template.html
@@ -7,14 +7,10 @@
 <script src="/resources/testharnessreport.js"></script>
 <script>
   let contentInConstructor = undefined;
-  let numberOfContentChildNodesInConnectedCallback = 0;
   const MyTemplate = class extends HTMLTemplateElement {
     constructor(...args) {
       super(...args);
       contentInConstructor = this.content;
-    }
-    connectedCallback(...args) {
-      numberOfContentChildNodesInConnectedCallback = this.content.childNodes.length;
     }
   }
   customElements.define('my-template', MyTemplate, {extends: 'template'});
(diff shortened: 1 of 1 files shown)


custom-elements/builtin-coverage/template.html
```diff
@@ -0,0 +1,36 @@
...
+<html is="my-html">
+<head>
+<meta charset="utf-8">
+<meta name="help" content="https://html.spec.whatwg.org/multipage/custom-elements.html#element-definition">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<script>
+  let contentInConstructor = undefined;
+  let numberOfContentChildNodesInConnectedCallback = 0;
+  let MyTemplate = class extends HTMLTemplateElement {
```
rniwa: Use const?

custom-elements/builtin-coverage/template.html
```diff
@@ -0,0 +1,36 @@
...
+    constructor(...args) {
+      super(...args);
+      contentInConstructor = this.content;
+    }
+    connectedCallback(...args) {
+      super.connectedCallback && super.connectedCallback();
```
rniwa: This is dead code because HTMLTemplateElement will never have connectedCallback. Remove?

custom-elements/builtin-coverage/template.html
```diff
@@ -0,0 +1,36 @@
...
+<body is="my-body">
+<div id="container"><template is="my-template">foo</template></div>
+<script>
+test(() => {
+  assert_true(contentInConstructor instanceof DocumentFragment);
+}, `Parsed customized template: Property '.content' should contain 'DocumentFragment' on constructor`);
+
+test(() => {
+  assert_equals(numberOfContentChildNodesInConnectedCallback, 1);
+}, `Parsed customized template: Property '.content' should contain child nodes on connectedCallback`);
```
rniwa: This behavior is implementation dependent. The browser engines are free to pause the HTML immediately after it had processed the start of a template element but before "foo" has been encountered.

Imagine the case where the network packet was immediately terminated between `<template is="my-template">` and `foo`. The HTML parser would parse the template element and happily insert it into the document and invoke the connected callback before the text node for "foo" is created.

rniwa (changes requested):

custom-elements/builtin-coverage/template.html
rniwa: This test shouldn't be merged until this implementation dependent test case is fixed.

custom-elements/builtin-coverage/template.html
```diff
@@ -0,0 +1,36 @@
+<!DOCTYPE html>
+<html is="my-html">
```
domenic:
```suggestion
<html>
```

domenic (changes requested):
+1 to @rniwa's point about not testing network-dependent behavior.

custom-elements/builtin-coverage/template.html
tomalec: Thanks for explaining this behavior. 

Does it mean that the only way for authors to observe when the `content` is parsed is `MutationObserver` on `this.content`? Also, on `connectedCallback` there is no way to guess whether `this.content` is empty because it is empty, or it was not yet parsed?

What is then the reason to attach empty `DocumentFragment` on constructor, wouldn't it be more intuitive if it would be assigned once parsed? Then authors could use `Proxy` to observe `content` being set, if such native set would trigger the trap. (The algorithm to set node document and host, or even creation of new fragment could still happen on constructor)

domenic: Yes, you can only observe when your children are mutated by using a MutationObserver.

You can't use proxies to observe changes to DOM properties. That's why things like MutationObserver exist.

rniwa: Well, the `DocumentFragment` associated with a template element gets elements streamed. So it might be initially empty, and as the parser receives more data, it gets more & more elements until the template element's end tag eventually appears.

In theory, we could queue up those elements somewhere else and add them all at once but that's not what the current HTML5 spec says. For the template element, that might make sense because the reason we don't do this elsewhere is so that the user can see the content progressively as the more content loads over the network but the template element never shows its content so it's irrelevant here.

tomalec: > Well, the DocumentFragment associated with a template element gets elements streamed. So it might be initially empty, and as the parser receives more data, it gets more & more elements until the template element's end tag eventually appears.

Does the above mean, that MutationObserver could be notified multiple times when the element is parsed? So, that an author of customized template element, or autonomous custom element that expects template element inside (like `<shadow-root><template>foo</template></shadow-root>`) cannot get the time when `template`'s content is fully parsed and ready?

tomalec:
I removed the implementation-specific test. Is that enough to merge the test for availability of `.content` in `constructor`?
The discussion regarding availability of content's child nodes could be moved elsewhere.  https://github.com/whatwg/html ?

domenic:
Thanks. It's hard to imagine when or how this test would fail, but more coverage doesn't hurt.

Can you remove the is="my-html" too?

The discussion regarding availability of content's child nodes could be moved elsewhere. https://github.com/whatwg/html ?

That's a reasonable place, although maybe StackOverflow is more appropriate.

domenic (approved):

PR #15488, opened 2019-02-20
[wptrunner] Reset internal state during "rerun"

jugglinmike:
The "reftest" implementation uses an internal cache for screenshots as
an optimization for running similar tests. That optimization is
inappropriate for the CLI's "rerun" feature since in that context,
repeatedly running the same tests is an explicit goal.

Introduce a generic "reset" message that is emitted by the
TestRunnerManager during "rerun", and extend the RefTestExecutor to
handle this message by emptying its internal cache.

### DIFF
--- a/tools/wptrunner/wptrunner/executors/executormarionette.py
+++ b/tools/wptrunner/wptrunner/executors/executormarionette.py
@@ -780,6 +780,9 @@ class MarionetteRefTestExecutor(RefTestExecutor):
             self.logger.warning("Exception during reftest teardown:\n%s" %
                                 traceback.format_exc(e))
 
+    def reset(self):
+        self.implementation.reset(**self.implementation_kwargs)
+
     def is_alive(self):
         return self.protocol.is_alive
 
@@ -860,6 +863,10 @@ class InternalRefTestImplementation(object):
                                 if value > 1}
         self.executor.protocol.marionette.set_context(self.executor.protocol.marionette.CONTEXT_CHROME)
         self.executor.protocol.marionette._send_message("reftest:setup", data)
+
+    def reset(self, screenshot=None):
+        self.teardown()
+        self.setup(screenshot)
 
     def run_test(self, test):
         references = self.get_references(test)
--- a/tools/wptrunner/wptrunner/executors/base.py
+++ b/tools/wptrunner/wptrunner/executors/base.py
@@ -152,6 +152,11 @@ class TestExecutor(object):
         if self.protocol is not None:
             self.protocol.teardown()
 
+    def reset(self):
+        """Re-initialize internal state to facilitate repeated test execution
+        as implemented by the `--rerun` command-line argument."""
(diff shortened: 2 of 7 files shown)


gsnedders:
@jgraham how do we want this to interact with the internal reftest runner for Marionette? we drop the canvas cache on `reftest:teardown` if I'm not sorely mistaken.

@jugglinmike does this not require changes in every executor we want this to work with? in most cases just calling `self.implementation.reset()`? can you do what you've done for `executorwebdriver` also for:

* tools/wptrunner/wptrunner/executors/executormarionette.py
* tools/wptrunner/wptrunner/executors/executorselenium.py
* tools/wptrunner/wptrunner/executors/executorservo.py
* tools/wptrunner/wptrunner/executors/executorservodriver.py

For `executormarionette.InternalRefTestImplementation` you'll need to add a new `reset` function too, but that can just call teardown then setup I'm pretty sure?

jugglinmike:
Thanks for the tip, @gsnedders! I've extended the fix to the other executors, but I don't know how to test this code, even manually. I won't be able to test Safari, but I just installed Servo, so I may be able to verify the others. Could you advise?

tools/wptrunner/wptrunner/executors/executormarionette.py
```diff
@@ -780,6 +780,9 @@ def teardown(self):
             self.logger.warning("Exception during reftest teardown:\n%s" %
                                 traceback.format_exc(e))
 
+    def reset(self):
+        self.implementation.reset(**self.implementation_kwargs)
```
gsnedders: This will throw with the external reftest runner, given it expects 0 args (well, 1: `self`).

jugglinmike: I modeled this after the `setup` method which invokes `self.implementation.setup` in a similar fashion. In the class's constructor, the value is defined as an empty dictionary when the external reftest runner is in use.

gsnedders: Oh, right. I can't read. :)

gsnedders:
There isn't really a great way to test this, I don't think. :/

Need to construct something that renders differently on sequential loads (set a cookie, output a number?), but we also know exactly how the first load renders (so we can have the ref pass once and fail the next, which cookies with an incrementing number should suffice for?). And hope we don't clear too much state before starting the second run…

Honestly, given how most of this code is done, I'm mostly tempted to take it just on trust, even though it is far from ideal (I'd really rather we had a better way to test executors, especially around edge-case behaviour, but they're sufficiently coupled to browsers we can't really do anything that great).

gsnedders (approved):

jugglinmike:
Thanks, Sam! Is this still pending input from @jgraham?

gsnedders:
Let's just land this rather than trying to figure out some way to test this. :)

PR #15491, opened 2019-02-20
Remove new lines in the reference for 'white-space: pre' tests

gkatsev:
After spending some time thinking about what the spec says and what `white-space: pre` is supposed to do. The references for these tests is incorrect. From my reading of the spec and the white-space property behavior, the CSS should take precedence over the regular behavior of cues (which is to wrap) and `white-space: pre` is specifically a no-wrap property except for explicit new-lines.

### DIFF
--- a/webvtt/rendering/cues-with-video/processing-model/selectors/cue/white-space_pre_wrapped-ref.html
+++ b/webvtt/rendering/cues-with-video/processing-model/selectors/cue/white-space_pre_wrapped-ref.html
@@ -25,4 +25,4 @@ body { margin:0 }
     white-space: pre
 }
 </style>
-<div class="video"><span class="cue"><span>A A  A   A	A	<br>	A		<br>A	A   A  A A</span></span></div>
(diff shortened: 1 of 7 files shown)


silviapfeiffer:
I agree. I had to track down all the original vtt files for these tests and none of them have newlines in them, so after applying white-space: pre; none of them should have created newlines. That agrees to removing the &le;br/> elements.
Also, white-space:pre; does not actually wrap when it goes out of the box (we need white-space: pre-wrap; for that). Therefore, the "overflow: hidden;" makes sense because we have to stay within the rendering box.

Do we need to add a "white-space: pre-wrap;" to the spec?

silviapfeiffer (approved):

gkatsev:
I think pre-wrap is already covered by some other tests as well. For example, http://w3c-test.org/webvtt/rendering/cues-with-video/processing-model/selectors/cue_function/bold_object/bold_white-space_pre-wrap_wrapped.html

PR #15493, opened 2019-02-21
Update algorithm to find ChromeDriver to download

JohnChen0:
Find ChromeDriver to download, based on recommendation from
https://sites.google.com/a/chromium.org/chromedriver/downloads/version-selection

### DIFF
--- a/tools/wpt/browser.py
+++ b/tools/wpt/browser.py
@@ -440,6 +440,21 @@ class Chrome(Browser):
 
         return "%s%s" % (platform, bits)
 
+    def chromium_platform_string(self):
+        platform = {
+            "Linux": "Linux",
+            "Windows": "Win",
+            "Darwin": "Mac"
+        }.get(uname[0])
+
+        if platform is None:
+            raise ValueError("Unable to construct a valid Chromium package name for current platform")
+
+        if (platform == "Linux" or platform == "Win") and uname[4] == "x86_64":
+            platform += "_x64"
+
+        return platform
+
     def find_binary(self, venv_path=None, channel=None):
         raise NotImplementedError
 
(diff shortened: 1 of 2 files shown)


tools/wpt/browser.py
```diff
@@ -446,10 +446,26 @@ def find_binary(self, venv_path=None, channel=None):
...
+    def install_webdriver(self, dest=None, channel=None, browser_binary=None):
         if dest is None:
             dest = os.pwd
-        latest = get("http://chromedriver.storage.googleapis.com/LATEST_RELEASE").text.strip()
+        latest = None
+        chrome_version = self.version(browser_binary)
+        if chrome_version is not None:
+            parts = chrome_version.split(".")
+            if len(parts) == 4:
+                latest_url = "http://chromedriver.storage.googleapis.com/LATEST_RELEASE_%s.%s.%s" % (
```
Hexcles:
```suggestion
                latest_url = "https://chromedriver.storage.googleapis.com/LATEST_RELEASE_%s.%s.%s" % (
```

tools/wpt/browser.py
```diff
@@ -446,10 +446,26 @@ def find_binary(self, venv_path=None, channel=None):
...
+                    parts[0], parts[1], parts[2])
+                try:
+                    latest = get(latest_url).text.strip()
+                except:
+                    latest_url = "http://chromedriver.storage.googleapis.com/LATEST_RELEASE_%s" % parts[0]
```
Hexcles:
```suggestion
                    latest_url = "https://chromedriver.storage.googleapis.com/LATEST_RELEASE_%s" % parts[0]
```

tools/wpt/browser.py
```diff
@@ -446,10 +446,26 @@ def find_binary(self, venv_path=None, channel=None):
...
+                    try:
+                        latest = get(latest_url).text.strip()
+                    except:
+                        pass
+        if latest is None:
+            latest = get("http://chromedriver.storage.googleapis.com/LATEST_RELEASE").text.strip()
```
Hexcles:
```suggestion
            latest = get("https://chromedriver.storage.googleapis.com/LATEST_RELEASE").text.strip()
```

tools/wpt/browser.py
```diff
@@ -446,10 +446,26 @@ def find_binary(self, venv_path=None, channel=None):
...
         url = "http://chromedriver.storage.googleapis.com/%s/chromedriver_%s.zip" % (latest,
```
Hexcles:
```suggestion
        url = "https://chromedriver.storage.googleapis.com/%s/chromedriver_%s.zip" % (latest,
```

Hexcles (commented):
Thanks for the patch, John!

Again regarding the dev channel (which is heavily used in WPT), IIUC, this code will eventually fall back to download `https://chromedriver.storage.googleapis.com/LATEST_RELEASE`, which is currently 2.46 and seems to be older than ChromeDriver 73.0.3683.20 (and probably doesn't work with Chrome Dev 74). I left some thoughts regarding the dev channel in https://crbug.com/chromedriver/2656#c13 . WDYT?

JohnChen0:
> Again regarding the dev channel (which is heavily used in WPT), IIUC, this code will eventually fall back to download `https://chromedriver.storage.googleapis.com/LATEST_RELEASE`, which is currently 2.46 and seems to be older than ChromeDriver 73.0.3683.20 (and probably doesn't work with Chrome Dev 74). I left some thoughts regarding the dev channel in https://crbug.com/chromedriver/2656#c13 . WDYT?

ChromeDriver 2.46 actually contains slightly more features than ChromeDriver 73.0.3683.20, even though it was built earlier, since 2.46 was built from master branch, while 73.0.3683.20 was built from a release branch that was created earlier. That's the reason the non-versioned LATEST_RELEASE still points to 2.46.

I have implemented the suggestion in https://crbug.com/chromedriver/2656#c13. Does WPT routinely run tests against the latest Dev release? I think the right thing to do is update this PR to fallback to latest ChromeDriver at https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html. WDYT?

gsnedders:
> Does WPT routinely run tests against the latest Dev release?

More often than stable.

Hexcles:
> I think the right thing to do is update this PR to fallback to latest ChromeDriver at https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html. WDYT?

Yes! As the last fallback, instead of using the non-versioned `LATEST_RELEASE`, let's grab the chromedriver build archive from https://commondatastorage.googleapis.com/chromium-browser-snapshots/{PLATFORM}/LAST_CHANGE , because in this code path we are most likely running the latest dev version of Chrome and the ToT build is more likely to work than the latest release.

Hexcles:
I've directly modified the PR to use the tip-of-tree ChromeDriver from Chromium build archive as the last fallback. This should PR should also fix the massive failures of ChromeDriver in `webdriver/`, as the necessary patch is included in the latest ChromeDriver.

Since I'm now a co-author of the PR, I won't review it myself. @gsnedders can you take a look?

foolip:
Ping @gsnedders for review. Or perhaps @LukeZielinski can take a look?

@Hexcles can you fix the CI failures?

jgraham (approved):

Hexcles:
Thank you to @JohnChen0 for the original patchset and to @jgraham for reviewing. This now merged and we should have much better/accurate Chrome results now.

PR #15498, opened 2019-02-21
[Azure Pipelines] Add wpt.fyi hooks for each individual job/run

foolip:


foolip:
It looks like this will work. But because every job on Azure Pipelines becomes a check run on GitHub, the hooks themselves are showing up here. The number of checks from Azure Pipelines is now 9 :(

lukebjerring (approved):

PR #15510, opened 2019-02-21
Add one test for "Update the Rendering" section of HTML.

dbaron:
I'm hoping to write a bunch of tests for this section, as part of the
investigation of w3ctag/design-reviews#72, and this is the first one.
It tests that request animation frame callbacks happen in the correct
order across multiple documents.

Gecko, Chromium, and WebKit all fail in different (and nondeterministic)
ways, although I've seen occasional passes.

### DIFF
--- a/html/webappapis/update-rendering/child-document-raf-order.html
+++ b/html/webappapis/update-rendering/child-document-raf-order.html
@@ -1,0 +1,118 @@
+<!DOCTYPE HTML>
+<meta charset=UTF-8>
+<title>Ordering of steps in "Update the Rendering" - child document requestAnimationFrame order</title>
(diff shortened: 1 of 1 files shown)


foolip:
Yep, everyone fails:
https://wpt.fyi/results/html/webappapis/update-rendering/child-document-raf-order.html?diff&filter=ADC&run_id=5383254781722624&run_id=5926943786532864
https://wpt.fyi/results/html/webappapis/update-rendering/child-document-raf-order.html?diff&filter=ADC&run_id=5455528981430272&run_id=5224365150961664
https://wpt.fyi/results/html/webappapis/update-rendering/child-document-raf-order.html?diff&filter=ADC&run_id=4672896517537792&run_id=5883670011314176

foolip (approved):

PR #15514, opened 2019-02-22
[css-syntax] Add character encoding tests

xfq:
This PR ports https://www.w3.org/International/tests/repo/results/css-input-byte-stream to WPT.

It contains tests for basic functionalities, unknown `@charset` value, `@charset` typos, and the precedence of HTML/HTTP/BOM/`@charset`.

/cc @r12a

### DIFF
--- a/css/css-syntax/tokenizing-and-parsing/input-byte-stream/i18n-css-syntax-bytes-001.html
+++ b/css/css-syntax/tokenizing-and-parsing/input-byte-stream/i18n-css-syntax-bytes-001.html
@@ -1,0 +1,46 @@
+<!DOCTYPE html>
+<html  lang="en" >
+<head>
+<meta charset="ISO-8859-15"/>
+<title>CSS3 Syntax, input byte stream: CSS with UTF-signature</title>
+<link rel='author' title='Richard Ishida' href='mailto:ishida@w3.org'>
+<link rel='help' href='http://www.w3.org/TR/css3-syntax/#input-byte-stream'>
+<script src="../../../resources/testharness.js"></script>
+<script src="../../../resources/testharnessreport.js"></script>
+<meta name='flags' content='http dom'>
+<meta name="assert" content="A UTF-8 signature at the beginning of a css style sheet sets the encoding for the style sheet to UTF-8.">
+<style type='text/css'>
+.test div { width:50px; }
+
+p { color: gray; margin-top: 3em; font-size: 80%; }
+</style>
+<link rel="stylesheet" type="text/css" href="support/bom.css">
+</head>
+<body>
+
+
+
+<div class="test"><div id="box" class="ýäè">&#xA0;</div></div>
+
+
+<!--Notes:
+In the style sheet, the CSS selector identifies a class name with the following byte sequence: c3 bd c3 a4 c3 a8.  Read as UTF-8 this gives the characters ýäè.  Read as ISO 8850-15, this gives ÃœÃ€Ãš.
+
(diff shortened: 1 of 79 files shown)


xfq:
For those who may want to edit my branch for whatever reason, please be very careful so that the editor doesn't change the encoding to UTF-8 when saving for those files that should be in a different encoding.

gsnedders:
Do you know how these compare with/overlap with those in `css/CSS2/i18n/syndata` and `css/css-syntax/charset`?

gsnedders:
And I think there's even more in `css/CSS2/syntax/` too. We have too many near-identical tests scattered over too many directories… 🙁

r12a:
Ah, i suspected these tests were hanging around somewhere, though i didn't find them when i checked.  @xfq could you take a look?  I also suspect that the tests @gsnedders is pointing to may not be exactly the same or as good as the ones in this PR.(?)

xfq:
I saw some similar (but not same) tests in the CSS2/* directory. Should I move these tests into one directory and remove redundant tests that do the same thing?

r12a:
I would assume that we need to ensure that these tests (being the latest versions) are stored under the css-syntax hierarchy.  Whether we should keep the existing tests under CSS2, or replace them with this version, i don't know. @gsnedders any suggestion?

gsnedders:
@r12a Having duplicates kept around isn't the end of the world; I'm much more concerned about the potential of having contradictory tests (where CSS2 has errata, and the tests don't match CSS2+errata & current CSS Syntax).

A lot of this depends on how hard it'd be to find which tests are duplicates (and having purely numbered tests doesn't help with this, as it makes it virtually impossible to do without reading each test!).

tabatkins:
So, these tests looks pretty reasonable. Should I approve+merge, or wait for some sort of deduping, or...?

gsnedders:
The lint failures from CI need dealt with if nothing else. Not so fussed about deduplicating.

xfq:
OK. I'll deal with the lint failures.

css/css-syntax/tokenizing-and-parsing/input-byte-stream/i18n-css-syntax-bytes-001.html
```diff
@@ -20,13 +20,13 @@
 
 
 
-<div class="test"><div id="box" class="���">&#xA0;</div></div>
+<div class="test"><div id="box" class="???">&#xA0;</div></div>
```
r12a: @xfq  Here and in other places, it looks like the the characters have been corrupted.

xfq:
I think I opened the files in ISO 8859-1 and 8859-15 for the files in these encodings and saved them as such, but I'll check them again.

r12a:
Looks like it's better, although i'd advise to run the test just to be sure.  Note that there are other files where the sequence was corrupted too.  (I really hated working on these tests, precisely because of the difficulty in getting this right, and the amount of thought required to ensure i didn't mess it up.)

xfq:
I have fixed the lint failures and run the tests (all tests passed).

gsnedders:
@tabatkins want to review/merge these now?

r12a:
Ping?  Seems like these have been ready to go for the past 2 and half years.  I'd like to get them off my TTD list, one way or another. Thanks.

tabatkins (approved):

tabatkins:
Thanks for the ping. Approving based on my previous approval and the assertion that the test corruptions have been fixed, but not personally re-verifying that.

r12a:
@gregwhitworth are you ok to merge this?  You're the pending reviewer.

PR #15517, opened 2019-02-22
WebDriver: Add tests to check script abortion due to location change

christian-bromann:
Ref: https://github.com/w3c/webdriver/issues/1400

Added test to ensure that script is aborted when location or context is being changed.

Note: This is work in progress until consensus is found for this case.


andreastt:
@whimboo You did some work in this area recently.

whimboo:
As mentioned this is work in progress, and we have to reach an agreement first via the linked issue in the first comment.

PR #15518, opened 2019-02-22
HTML: opener and discarded auxiliary browsing context

annevk:
For https://github.com/whatwg/html/pull/4379.

### DIFF
--- a/html/browsers/windows/embedded-opener-remove-frame.html
+++ b/html/browsers/windows/embedded-opener-remove-frame.html
@@ -1,10 +1,36 @@
 <!doctype html>
-<title>opener and "removed" embedded documents</title>
+<title>opener and discarded browsing contexts</title>
 <script src=/resources/testharness.js></script>
 <script src=/resources/testharnessreport.js></script>
 <div id=log></div>
 <iframe name=matchesastring></iframe>
 <script>
+function testOpener(otherW, thisW, discardOtherBC) {
+  assert_equals(otherW.opener, thisW, "opener before removal");
+
+  const openerDesc = Object.getOwnPropertyDescriptor(otherW, "opener"),
+        openerGet = openerDesc.get;
+
+  assert_equals(openerGet(), thisW, "opener before removal via directly invoking the getter");
+  discardOtherBC();
+  assert_equals(otherW.opener, null, "opener after removal");
+  assert_equals(openerGet(), null, "opener after removal via directly invoking the getter");
+
+  otherW.opener = null;
+  assert_equals(openerGet(), null, "opener after setting it null via directly invoking the getter");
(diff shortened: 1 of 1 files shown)


html/browsers/windows/embedded-opener-remove-frame.html
```diff
@@ -15,32 +41,17 @@
...
-    frameW.opener = "immaterial";
-    assert_equals(openerGet(), null, "opener after setting it \"immaterial\" via directly invoking the getter");
-    const openerDescImmaterial = Object.getOwnPropertyDescriptor(frameW, "opener");
-    assert_equals(openerDescImmaterial.value, "immaterial");
-    assert_true(openerDescImmaterial.writable);
-    assert_true(openerDescImmaterial.enumerable);
-    assert_true(openerDescImmaterial.configurable);
+    testOpener(frameW, window, () => frame.remove());
 
     t.done();
```
domenic: Can use t.step_func_done for this one too, right?

domenic (approved):

html/browsers/windows/embedded-opener-remove-frame.html
annevk: No, due to different load event behavior.

annevk:
This probably should not be merged until https://github.com/whatwg/html/pull/4379#issuecomment-466351278 is resolved somehow.

html/browsers/windows/embedded-opener-remove-frame.html
```diff
@@ -1,10 +1,36 @@
...
+
+  const openerDesc = Object.getOwnPropertyDescriptor(otherW, "opener"),
+        openerGet = openerDesc.get;
+
+  assert_equals(openerGet(), thisW, "opener before removal via directly invoking the getter");
+  discardOtherBC();
+  assert_equals(otherW.opener, null, "opener after removal");
+  assert_equals(openerGet(), null, "opener after removal via directly invoking the getter");
+
+  otherW.opener = null;
```
bzbarsky: It's a little weird to do this after the opener is already asserted to be null, but I guess we were already testing that.

bzbarsky (approved):
This seems fine, modulo sorting out the `close()` behavior.  I commented over in https://github.com/whatwg/html/pull/4379 about what's going on there in Firefox...

html/browsers/windows/embedded-opener-remove-frame.html
annevk: Mainly to ensure the setter doesn't do anything weird. I could remove this if desired...

domenic (commented):
New version LGTM again. It assumes that the spec will queue a task for discarding, which per IRC conversations sounds like it's a good assumption.

PR #15521, opened 2019-02-22
Verify that tee doesn't pull more chunks than the branch HWM

ricea:
ReadableStreamTee should only pull enough chunks to fill the HWM of the
branches (which is 1), and no more.

When pulling from the branches, verify that pull() on the original
stream still is only called enough to fill the emptiest branch.

Corresponding standard change is https://github.com/whatwg/streams/pull/997.


domenic (approved):

ricea:
Superseded by #16167.

PR #15519, opened 2019-02-22
[Azure Pipelines] POST to the /api/checks/azure/1234 endpoint

foolip:
This regressed in https://github.com/web-platform-tests/wpt/pull/15384.

PR #15522, opened 2019-02-22
[Azure Pipelines] Put artifact name in urlencoded POST body

foolip:
... instead of in the query string. This will probably resolve the
"POST requests require a Content-length header" problem [1] assuming
that the cause is a version of curl that doesn't send the header if
the request has no body.

[1] https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=7671

### DIFF
--- a/tools/ci/azure/fyi_hook.yml
+++ b/tools/ci/azure/fyi_hook.yml
@@ -12,7 +12,7 @@ jobs:
   pool:
     vmImage: 'ubuntu-16.04'
   steps:
-  - script: curl -s -S -X POST https://wpt.fyi/api/checks/azure/$(Build.BuildId)?artifact=${{ parameters.artifactName }}
+  - script: curl -s -S -d "artifact=${{ parameters.artifactName }}" -X POST https://wpt.fyi/api/checks/azure/$(Build.BuildId)
     displayName: 'Invoke wpt.fyi hook'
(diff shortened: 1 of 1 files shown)


lukebjerring (approved):
Assume you tried this command locally, with success?

foolip:
> Assume you tried this command locally, with success?

I didn't have a build I could try it on without creating a dupe, but I've just tried it `curl -s -S -d "artifact=safari-preview-results" -X POST https://staging.wpt.fyi/api/checks/azure/7671` with "Azure build artifacts retrieved successfully" and will just accept the dupe.

lukebjerring:
That won't duplicate. The results-processor takes a hash of the full report, so two attempts to process _exactly_ the same reports won't duplicate :)

foolip:
> That won't duplicate

Ah, then I dare also try `curl -s -S -d "artifact=safari-preview-results" -X POST https://wpt.fyi/api/checks/azure/7671`. "Azure build artifacts retrieved successfully". So this might just work, although the `artifacts` filtering won't until the next prod release.

foolip:
https://travis-ci.com/web-platform-tests/wpt/jobs/179836991 is the same failure as in https://github.com/web-platform-tests/wpt/pull/15519 and that went away on retry, so it's flaky.

Issue #5492, opened 2017-04-07
Get rid of cyclic tests (border-radius tests)


wpt-issue-mover:
Originally posted as https://github.com/w3c/csswg-test/issues/1127 by @gsnedders on 16 Sep 2016, 15:51 UTC:

> Currently, css-backgrounds-3/border-{bottom,top}-{left,right}-radius-{003/004,005/006,007/009}.xht form simple cycles in the reftest graph with nothing else in them. As such, e.g., css-backgrounds-3/border-bottom-left-radius-003.xht and css-backgrounds-3/border-bottom-left-radius-004.xht are entirely equivalent tests despite their different assertions.
> 
> We probably want to write refs (SVGs?) that match the expected border, and have both tests point to that.

csswg-test-issue-mover:
Originally posted as https://github.com/w3c/csswg-test/issues/1127#issuecomment-250468413 by @gsnedders on 29 Sep 2016, 13:39 UTC:

> Unless I'm being stupid, the following should match (top-left):
> 
> ```
> <svg width="200" height="100"
>      xmlns="http://www.w3.org/2000/svg">
>  <path d="M25,0 a25,25 0 0,0 -25,25 V100 H200 V0 z"
>         fill="#dddddd" stroke="#a1a1a1" stroke-width="2" />
> </svg>
> ```

csswg-test-issue-mover:
Originally posted as https://github.com/w3c/csswg-test/issues/1127#issuecomment-250478350 by @gsnedders on 29 Sep 2016, 14:15 UTC:

> Oh, using SVGs for refs for border-radius falls into the trap of SVG strokes being centred on the edge of the path and CSS borders being outside the path, which also makes drawing the border arc impossible, given in CSS the arc radii are those on the outside edge of the border.
> 
> Any ideas, @tabatkins?

csswg-test-issue-mover:
Originally posted as https://github.com/w3c/csswg-test/issues/1127#issuecomment-250478698 by @gsnedders on 29 Sep 2016, 14:16 UTC:

> I mean I guess we can test it if we have fill/stroke & background/border the same colour, but that isn't great. Maybe the best we can do with references, though?

csswg-test-issue-mover:
Originally posted as https://github.com/w3c/csswg-test/issues/1127#issuecomment-250556689 by @tabatkins on 29 Sep 2016, 18:48 UTC:

[...]

PR #15523, opened 2019-02-22
Reftest cleanup

gsnedders:
This:

 * moves references out of support directories outside of `css/vendor-imports/mozilla` (see #5843)
 * standardises the directory in which references are contained as `reference` (v. a mixture of that and `reftest`)
 * deals with all the cyclic border-radius tests, fixing #5492 (this just ends up with us testing percentages are resolved correctly, but it's better than the status quo which is pretty meaningless)
 * gets rid of some cases of using images as references (which is just generally broken)

The start of the branch is split out from #5844, but we should definitely land most of this.


Ms2ger (approved):

dbaron:
The `text-indent-wrap-001` changes look good to me.  (Didn't look too closely at the rest.)

PR #15537, opened 2019-02-23
MathML: Update meta help to point to the MathML Core spec.

fred-wang:


davidcarlisle (approved):
new location for the spec and new stable non-numeric ids

PR #15548, opened 2019-02-25
Add a test for the URL used in DOMParser.

Ms2ger:


Ms2ger:
Note that this matches browsers, but not the current spec: https://github.com/w3c/DOM-Parsing/issues/46.

CYBAI:
I saw this is mentioned in web-platform-tests/wpt#21041. As that one is merged, should this one be closed 👀?

domenic (approved):
This looks good but I will push a commit that renames it to `DOMParser-parseFromString-url-more-tests.html` so it is more clearly connected to the tests merged in #21041.

PR #15556, opened 2019-02-26
[wptrunner] Do not use undefined method

jugglinmike:
The Selenium wire protocol does not define a "set window rect" method
[1]. Implement the desired behavior by composing the "set window
position" and "set window size" methods.

[1] https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol


jugglinmike:
For context: [this is interfering with results collection when using Edge via Sauce Labs](https://github.com/web-platform-tests/results-collection/issues/653).

jgraham (approved):

PR #15558, opened 2019-02-26
Events from GlobalEventHandlers must dispatch on disabled form elements

marcoscaceres:
Web Compatibility bugs bonanza... weird failures across all UAs. 

I'm thinking I might split this into multiple test based on each interface type.

### DIFF
--- a/dom/events/Event-dispatch-on-disabled-elements.html
+++ b/dom/events/Event-dispatch-on-disabled-elements.html
@@ -99,6 +99,123 @@ test(() => {
 }, "Calling click() on disabled elements must not dispatch events.");
 
 promise_test(async () => {
+  // Based on GlobalEventHandlers and mixins
+  // https://html.spec.whatwg.org/multipage/webappapis.html#globaleventhandlers
+  // https://drafts.csswg.org/css-transitions/#interface-globaleventhandlers-idl
+  const globalEventHandlers = new Map([
+    [
+      DragEvent,
+      [
+        "drag",
+        "dragend",
+        "dragenter",
+        "dragexit",
+        "dragleave",
+        "dragover",
+        "dragstart",
+        "drop",
+      ],
+    ],
+    [
+      Event,
+      [
+        "abort",
+        "cancel",
+        "canplay",
+        "canplaythrough",
+        "change",
+        "close",
+        "cuechange",
+        "durationchange",
+        "emptied",
+        "ended",
+        "input",
+        "invalid",
+        "loadeddata",
+        "loadedmetadata",
+        "loadend",
+        "pause",
+        "play",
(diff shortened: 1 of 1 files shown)


dom/events/Event-dispatch-on-disabled-elements.html
```diff
@@ -98,6 +98,123 @@
...
+        "timeupdate",
+        "toggle",
+        "volumechange",
+        "waiting",
+      ],
+    ],
+    [FormDataEvent, ["formdata"]],
+    [KeyboardEvent, ["keydown", "keypress", "keyup"]],
+    [
+      MouseEvent,
```
marcoscaceres: Actually, some of these mouse ones shouldn't be here... the ones that map to a "click".

annevk:
This is great, but it seems to me that you can basically use any interface with any event name (including custom names) and it should work, right? The only special case in the specification is activation behavior (tied to click), but also has some bugs.

And all implementations maintain various event lists that they treat in special ways, which doesn't really match anything.

https://github.com/whatwg/html/issues/2368 tracks this standards-wise. And https://github.com/whatwg/html/issues/4328 is related.

marcoscaceres:
> This is great, but it seems to me that you can basically use any interface with any event name (including custom names) and it should work, right?

You are absolutely right. Part of it was simply to figure out what interfaces are not implemented (how I found Firefox was missing `FormDataEvent`, for instance... maybe it's too indirect?). 

Nonetheless, I'm seeing some weird/interesting things... like in Firefox, "cancel", "securitypolicyviolation", and "cuechange" don't fire (will file a bug... trying to figure out why). 

Chrome fails on a bunch of drag events, and a few other things. Same with Safari... like you said, it may be complete unrelated to using any interface.   

> The only special case in the specification is activation behavior (tied to click), but also has some bugs.

I'm currently working on that code, so any pointers to those bugs would be helpful. Our current code has a whitelist of events the can be triggered when a form element is disabled... which is not great, because we keep missing things (e.g., [CSS transitions](https://phabricator.services.mozilla.com/D21166)).  @smaug---- suggested we maybe just blacklist `click()` and let other things through.

annevk:
Yeah, only handling `click` in a special way is what the DOM dispatch algorithm does. If that's web-compatible that'd be great. https://bugzilla.mozilla.org/show_bug.cgi?id=1520869 is the bug where Alphan and I investigated this a bit. Corresponding standards issues are in my previous comment.

marcoscaceres:
Thanks for the pointers. I subscribed to the spec bugs to follow along.

smaug----:
At some point Chrome folks were discussing about blocking some more events, was it *down/*up events. Can't recall the details.

PR #15562, opened 2019-02-26
[Azure Pipelines] Drop --no-restart-on-unexpected for Edge tests

foolip:
Jobs often fail or time out, try restarting to see if it finishes more reliable.

This will likely make runs much slower.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -183,7 +183,7 @@ jobs:
   - template: tools/ci/azure/install_certs.yml
   - template: tools/ci/azure/update_hosts.yml
   - template: tools/ci/azure/update_manifest.yml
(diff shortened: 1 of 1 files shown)


foolip:
I started https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=8111 which is also showing up as checks on this PR. Let's see how that goes.

@lukebjerring if it succeeds that run would be labeled as master because of the artifact name. What would be your preference for preventing this, both in Azure Pipelines and with one-off Taskcluster runs? My only idea would be to send the branch name to the results receiver, are there other better ways?

lukebjerring:
I guess the branch name could be considered source-of-truth on whether the run is on master. I believe that's what we already do in Taskcluster:
https://github.com/web-platform-tests/wpt.fyi/blob/ec9455cc5cf4f701663e16f2ddefde919092df64/api/taskcluster/webhook.go#L158

lukebjerring:
https://github.com/web-platform-tests/wpt.fyi/pull/1180

foolip:
Sweet, thanks Luke!

foolip:
With this change the run finished, but it isn't very complete:
https://wpt.fyi/results/?run_id=5434547495763968

I'll rename the build to hide it, and I think https://github.com/web-platform-tests/wpt/issues/13073 is the cause.

@mustjab have you been able to diagnose what's causing some of the Edge failures? The success rate has been very low in past few days of attempts.

Ms2ger (approved):
Ok, I guess.

PR #15568, opened 2019-02-26
[css-grid] Import grid-item-percentage-sizes tests

Loirooriol:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1526567


mrego (approved):
I guess this is the best way we have to import Mozilla Grid Layout tests in WPT (so we can use them in Chromium/WebKit).

Just double-checking with @emilio that we're doing the right thing.

mrego (changes requested):
So it seems this should be exported automatically from Mozilla, and something failed in the process. So let's close this.

PR #15570, opened 2019-02-27
[wptrunner] Add screenshot hashes to wptreport JSON along with a new formatter for screenshot data URIs

Hexcles:
Data flow:
1. In executors, if external reftest runner is used, we directly add hashes to the `reftest_screenshots` field of the `extra` log data in [`RefTestImplementation`](https://github.com/web-platform-tests/wpt/pull/15570/files#diff-2358aba85eb4ec818ff48199ced359dcR345).
2. Marionette internal reftests does not use `RefTestImplementation` and produces log data from within Marionette, so we use [`reftest_result_converter`](https://github.com/web-platform-tests/wpt/pull/15570/files#diff-2358aba85eb4ec818ff48199ced359dcR103) to check & compute the hash if it does not exist. This introduces some additional computation inadvertently.
3. When log data is sent to the formatter, [`WptreportFormatter`](https://github.com/web-platform-tests/wpt/pull/15570/files#diff-3d4aa557b26cec06dbd54ecd31ead833R119) extracts test paths and hashes from `reftest_screenshots` into `wptreport.json` in the format expected by wpt.fyi; [`WptscreenshotFormatter`](https://github.com/web-platform-tests/wpt/pull/15570/files#diff-3d4aa557b26cec06dbd54ecd31ead833R164), a new formatter, extracts only the screenshots and dumps them into another file as data URIs.

The implementation should be compatible with all existing log formatters and other consumers of logs, since we only add a new field without modifying existing fields. I've checked `--log-mach` and `--log-tbpl` formatters, as well as `reftest-analyzer-structured` (which takes raw logs produced by `--log-raw`).

See individual commit messages for more detailed explanation.

Working towards [web-platform-tests/wpt.fyi#57](https://github.com/web-platform-tests/wpt.fyi/issues/57)

### DIFF
--- a/tools/wptrunner/wptrunner/executors/base.py
+++ b/tools/wptrunner/wptrunner/executors/base.py
@@ -77,11 +77,34 @@ class TestharnessResultConverter(object):
 testharness_result_converter = TestharnessResultConverter()
 
 
+def hash_screenshot(data):
+    """Computes the sha1 checksum of a base64-encoded screenshot."""
+    return hashlib.sha1(base64.b64decode(data)).hexdigest()
+
+
+def _ensure_hash_in_reftest_screenshots(extra):
+    """Make sure reftest_screenshots have hashes.
+
+    Marionette internal reftest runner does not produce hashes.
+    """
+    log_data = extra.get("reftest_screenshots")
+    if not log_data:
+        return
+    for item in log_data:
+        if type(item) != dict:
+            # Skip relation strings.
+            continue
+        if "hash" not in item:
+            item["hash"] = hash_screenshot(item["screenshot"])
+
+
 def reftest_result_converter(self, test, result):
+    extra = result.get("extra", {})
+    _ensure_hash_in_reftest_screenshots(extra)
     return (test.result_cls(
         result["status"],
         result["message"],
-        extra=result.get("extra", {}),
+        extra=extra,
         stack=result.get("stack")), [])
 
 
@@ -262,7 +285,7 @@ class RefTestImplementation(object):
                 return False, data
 
             screenshot = data
-            hash_value = hashlib.sha1(base64.b64decode(screenshot)).hexdigest()
+            hash_value = hash_screenshot(data)
 
             self.screenshot_cache[key] = (hash_value, None)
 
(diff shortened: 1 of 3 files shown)


tools/wptrunner/wptrunner/executors/base.py
```diff
@@ -318,12 +319,19 @@ def run_test(self, test):
                 if success:
                     screenshots[i] = screenshot
 
+        # For Mozilla reftest analyzer (included in tbpl):
         log_data = [{"url": nodes[0].url, "screenshot": screenshots[0]}, relation,
                     {"url": nodes[1].url, "screenshot": screenshots[1]}]
+        # For wpt.fyi (included in wptreport.json):
+        new_log_data = {nodes[0].url: (hashes[0], screenshots[0]),
```
jgraham: Instead of passing multiple formats of the same data around, couldn't we just add the hashes key to the existing data, which should be backwards-compatible and rearrange it in the formatter?

Hexcles: Right, this is the main question I'd like to ask you: it's currently done this way because of Marionette. If we want to modify the existing data for `reftests_screenshots`, we'd need to modify Marionette as well, as this code path isn't used by Firefox by default. If you'd prefer modifying Marionette, could you give me a pointer to the code?

jgraham: I"m not sure why we need to modify marionette (but sure I"m happy for you to do that). Why can't we just use the same data format and in the formatter (or somewhere earlier like in wpttest.py) compute the hashes if they're missing? That seems like the simplest change to me. What am I not seeing?

tools/wptrunner/wptrunner/formatters.py
```diff
@@ -50,6 +53,37 @@ def replace_lone_surrogate(data):
     return LONE_SURROGATE_RE.subn(surrogate_replacement, data)[0]
 
 
+def convert_reftest_screenshots(extra):
```
jgraham: Can this go in `executors/base.py` as part of `convert_reftest_result`? That way we only do it once per result, not once per formatter.

tools/wptrunner/wptrunner/formatters.py
```diff
@@ -50,6 +53,37 @@ def replace_lone_surrogate(data):
...
+    """Converts reftest_screenshots filed into the following format:
+    {"path": {"hash": "sha1:...", "screenshot": "base64-encoded data"}, ...}
+    """
+    def _hash(data):
+        return hashlib.sha1(base64.b64decode(data)).hexdigest()
+
+    def _strip_server(url):
```
jgraham: This function is already defined in `executors/base.py`

jgraham (changes requested):

tools/wptrunner/wptrunner/formatters.py
```diff
@@ -50,6 +53,37 @@ def replace_lone_surrogate(data):
...
+    log_data = extra.get("reftest_screenshots")
+    if not log_data:
+        return None
+    screenshots = {}
+    for item in log_data:
+        if type(item) != dict:
+            # Skip relation strings.
+            continue
+        # Marionette internal reftest runner does not produce the hash field.
+        hash_val = item.get("hash", _hash(item["screenshot"]))
```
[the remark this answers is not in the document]
Hexcles: Oops somehow that didn't even occur to me... Thanks for spotting this!

tools/wptrunner/wptrunner/formatters.py
Hexcles: Yeah, I've moved it into `executors/base.py` as part of `reftest_result_converter` which will be called regardless of whether Marionette internal reftest runner is used (originally I didn't think it was possible because I was looking at `RefTestImplementation`, which isn't used by the internal reftest runner).

tools/wptrunner/wptrunner/executors/base.py
```diff
@@ -76,11 +77,45 @@ def __call__(self, test, result, extra=None):
...
+            "screenshot": item["screenshot"],
+        }
+    return screenshots
+
+
 def reftest_result_converter(self, test, result):
+    extra = result.get("extra", {})
+    screenshots = _convert_reftest_screenshots(extra)
+    if screenshots:
+        extra["screenshots"] = screenshots
```
Hexcles: Note that I think this will modify `result["extra"]` in place. I don't think it'd cause problems in this case, but it could be dangerous.

Hexcles:
@jgraham sorry for the confusion. Please take a look again. (I've also updated the PR description.)

jgraham (approved):

gsnedders (approved):

PR #14798, opened 2019-01-10
Remove inactive reviewers

gsnedders:
This removes reviewers who have had more than three review requests in the past year but who haven't reviewed anything. Fixes #13702.

Places where this leaves us with no reviewers:

* css/css-display/run-in (but parent has reviewers, so let's just kill that)
* ~css/css-position (very few PRs; @atanassov is the current reviewer)~
* ~css/css-scrollbars (very few PRs; @tantek is current? maybe +@upsuper?)~
* encrypted-media (very few PRs; @ddorwin is current)
* ~feature-policy (very few PRs; @clelland is current)~
* ~gamepad (very few PRs; @luser is current)~

I believe everyone removed here was pinged in https://github.com/web-platform-tests/wpt/issues/13702#issuecomment-433376601, so I don't think we need to check about anyone except for the above. If the people mentioned above want to remain reviewers, please speak now.

PR #15577, opened 2019-02-27
Wrap Feature Policy common code in setup()

foolip:
That ReportingObserver isn't available in Safari causes the test to be
treated as a single page test with a single failing subtest:
https://wpt.fyi/results/feature-policy/experimental-features/layout-animations-disabled-violation-report-js-tentative.html?run_id=5374658471788544&run_id=5135755412242432&run_id=5719614843518976

Wrapping it will cause there to be a harness error instead.

This will help with https://bugzilla.mozilla.org/show_bug.cgi?id=1530648.

### DIFF
--- a/feature-policy/experimental-features/resources/common.js
+++ b/feature-policy/experimental-features/resources/common.js
@@ -66,13 +66,15 @@ function wait_for_load(e) {
   });
 }
 
-window.reporting_observer_instance = new ReportingObserver((reports, observer) => {
-  if (window.reporting_observer_callback) {
-    reports.forEach(window.reporting_observer_callback);
-  }
-}, {types: ["feature-policy-violation"]});
-window.reporting_observer_instance.observe();
-window.reporting_observer_callback = null;
+setup(() => {
+  window.reporting_observer_instance = new ReportingObserver((reports, observer) => {
+    if (window.reporting_observer_callback) {
+      reports.forEach(window.reporting_observer_callback);
+    }
+  }, {types: ["feature-policy-violation"]});
+  window.reporting_observer_instance.observe();
+  window.reporting_observer_callback = null;
+});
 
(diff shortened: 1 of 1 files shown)


marcoscaceres (commented):
I don't know much about the ReportingObserver, but the code looks fine to me. Do you want me to confirm it's addressing the payment request API failures in Gecko?

foolip:
@marcoscaceres this wasn't intended to change the results in Firefox, and it looks like results are unaffected:
https://wpt.fyi/results/feature-policy/experimental-features?diff&filter=ADC&run_id=5744326038519808&run_id=5676496995745792

What it should do is change the failure mode for Safari, so that there's a harness error and it doesn't look like Firefox has a lone failure there.

marcoscaceres:
Ah, apologies, I'd misunderstood.

foolip:
https://github.com/web-platform-tests/wpt.fyi/issues/1160 is the bug about Safari results not appearing. The results were uploaded though, and I was able to find the two runs to compare:
https://wpt.fyi/results/feature-policy/experimental-features?diff&filter=ADC&run_id=5080330100801536&run_id=5724651531534336

This had the intended effect, and should remove some things from a list of Firefox-specific failures, and maybe something from a list of Safarai-specific failures too.

foolip:
I just rebased to runs checks again before pinging review of this. I see that @bakulf is the only suggested reviewer, but not added because of not being an org member. @bakulf, I've sent an org invite to sort that out, please accept if you're willing to do reviews here.

@clelland, looks like you were removed in #14798, sent https://github.com/web-platform-tests/wpt/pull/18537 to fix that.

clelland (approved):
This looks good, and I'm happy to see the more useful results from the Safari bot

PR #15580, opened 2019-02-27
Add Dockerfile for Epiphany testing

foolip:


foolip:
@mcatanzaro I tweaked some things, but I'm still not able to get this working locally, with this command:
```
docker run -it e024b3f58ee3 /bin/bash --login -c "set -ex; ~/start.sh https://github.com/web-platform-tests/wpt master FETCH_HEAD epiphany; cd ~/web-platform-tests; NO_AT_BRIDGE=1 dbus-run-session ./tools/ci/taskcluster-run.py epiphany"
```

(e024b3f58ee3 is just the sha my image got, you'll see something different.)

It fails like this:
```
Using 1 client processes
Starting http server on web-platform.test:8000
Starting http server on web-platform.test:8001
Starting https server on web-platform.test:8443
SUITE-START | Running 32032 tests
Running reftest tests
PID 183 | Failed to start HTTP server at port 4444: Could not listen on address ::1, port 4444: Error binding to address: Cannot assign requested address
WebDriver was not accessible within the timeout:
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/webdriver_server.py", line 73, in _run
    wait_for_service((self.host, self.port))
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/webdriver_server.py", line 265, in wait_for_service
    raise socket.error("Service is unavailable: %s:%i" % addr)
error: Service is unavailable: 127.0.0.1:4444
```

Is this because I used `FROM fedora:rawhide`? I couldn't get your original Dockerfile to build, it failed on GPG errors.

jugglinmike:
WPT has consistently used Ubuntu when a GNU/Linux distribution has been required. Using Fedora for Epiphany could reduce maintainability, so I'd like to have the motivation documented. Could you say a little to that effect?

mcatanzaro:
> WPT has consistently used Ubuntu when a GNU/Linux distribution has been required. Using Fedora for Epiphany could reduce maintainability, so I'd like to have the motivation documented. Could you say a little to that effect?

Automation support is new in Epiphany 3.32, which is only available in Ubuntu 19.04:

https://launchpad.net/ubuntu/+source/epiphany-browser

We can absolutely use Ubuntu if there is an appropriate 19.04 prerelease container image available.

Eventually, we probably want to try building Epiphany ourselves -- it's just a matter of infrastructure work -- but using distro packages seems like by far the easiest way to start.

mcatanzaro:
> Is this because I used `FROM fedora:rawhide`? I couldn't get your original Dockerfile to build, it failed on GPG errors.

I'm not sure, but maybe it only works if built on Fedora hosts. Fedora GPG keys are preinstalled for package verification; maybe they're used for Docker images too.

mcatanzaro:
Note that, besides package names, this line in the Dockerfile would need to be changed if using Ubuntu since Ubuntu has a custom admin group:

````
&& usermod -a -G wheel test \
```

wheel -> sudo or admin (I can't ever remember which it is)

mcatanzaro:
I should mention: Epiphany rarely receives updates in Ubuntu (e.g. the version in 18.04 is really broken) so if we use Ubuntu, then in the long-term we really need to build Epiphany ourselves, and probably also WebKit and the entire network and multimedia stacks (at least).

foolip:
I'd like us to use Fedora, because that is the OS where Epiphany (GNOME Web) is "native", and consider this setup more like the browser-is-part-of-OS like Edge and Safari, than our setup for Chrome and Firefox.

It won't be used for any testing on PRs, just to run it daily on Taskcluster.

foolip:
@mcatanzaro will you be able to take a look at the Dockerfile and get it working again?

mcatanzaro:
Sometime soon, yes.

mcatanzaro:
I was able to reproduce the GPG errors today by rebuilding the Dockerfile with `$ sudo docker build --tag=ephywpt --no-cache .`. Seems rawhide is just broken right now. Not a problem, because Fedora 30 has been branched, so I switched my example Dockerfile https://gitlab.gnome.org/GNOME/epiphany/snippets/616 to use Fedora 30 instead of rawhide. That will be a much safer base anyway. This builds fine for me, without the GPG errors.

Now, regarding `error: Service is unavailable: 127.0.0.1:4444`, I was not able to reproduce this locally, but one of my colleagues discovered the same error occurs on one of Igalia's build machines. We don't know why yet.

One possible difference is different docker versions. On Fedora 30, I have Docker version 1.13.1, build 1185cfd/1.13.1, which is what I'm testing with. The build machine where this error occurs is running Docker version 17.09.1-ce, build 19e2cf6. The Fedora versioning looks really confusing and I don't understand that fully, but it seems to be git snapshots for some reason.

foolip:
@mcatanzaro can you suggest a patch against this branch? Is it important that we use  `FROM registry.fedoraproject.org/fedora:31` and not just `FROM fedora:31`? Also, isn't 31 and rawhide the same thing? It looks like it in https://hub.docker.com/_/fedora.

If we can get past the GPG errors, maybe there will be a new error to deal with :)

mcatanzaro:
> @mcatanzaro can you suggest a patch against this branch?

Well no, because we don't know how to solve the `Service is unavailable` problem....

>Is it important that we use `FROM registry.fedoraproject.org/fedora:31` and not just `FROM fedora:31`? 

I don't know. Probably they are the same?

> Also, isn't 31 and rawhide the same thing? It looks like it in https://hub.docker.com/_/fedora.

Um, yes, that's my mistake. I was meaning to try fedora:30 instead. We should use fedora:30. I'll update my example GitLab snippet and test to make sure it works as expected.

I hit the GPG error with fedora:rawhide but not after switching from fedora:rawhide to fedora:31, so that tricked me into thinking I had actually made a useful change. Maybe that's just coincidence that I didn't get the error on my second attempt, or bad mirrors, or something. I'm pretty sure there should be no GPG errors so long as we stick with fedora:30.

> If we can get past the GPG errors, maybe there will be a new error to deal with :)

I'm not worried about the GPG errors, I'm worried about:

```
PID 183 | Failed to start HTTP server at port 4444: Could not listen on address ::1, port 4444: Error binding to address: Cannot assign requested address
WebDriver was not accessible within the timeout:
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/webdriver_server.py", line 73, in _run
    wait_for_service((self.host, self.port))
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/webdriver_server.py", line 265, in wait_for_service
    raise socket.error("Service is unavailable: %s:%i" % addr)
error: Service is unavailable: 127.0.0.1:4444
```

I'm pretty discouraged that this error occurs on certain hosts and not others. The whole point of containerization is to avoid host-specific problems. :(

mcatanzaro:
> > Is it important that we use `FROM registry.fedoraproject.org/fedora:31` and not just `FROM fedora:31`?
> 
> I don't know. Probably they are the same?

My example now uses the shorter version: `FROM fedora:30`. It works fine.

mcatanzaro:
OK, progress. On the one system where we are able to reproduce this error, the problem is due to `WebKitWebDriver` failing to listen():

```
$ /usr/bin/WebKitWebDriver --port=4444
Failed to start HTTP server at port 4444: Could not listen on address ::1, port 4444: Error binding to address: Cannot assign requested address
```

foolip, that error looks to be the same one that you were blocked on up above.

Now, this is the very first command I run after entering the Docker container with `sudo docker run --rm -it epiphany-wpt /bin/bash` so there is no way any other program could have possibly listened on 4444 prior to WebKitWebDriver. This works perfectly fine with the identical Dockerfile on my computer, so somehow the host system matters here. The system where this error occurs is itself an LXC container, so it's Docker running inside LXC.

Note also that this error is not EADDRINUSE, as I would expect for a listen() failure. It is EADDRNOTAVAIL. Seems like nothing at all is allowed to listen on that address.

Eventually I thought to try:

```
$ ping -c3 ::1
ping: connect: Cannot assign requested address
```

Problem is WebKitWebDriver is trying to listen on ::1, but the system doesn't support IPv6!

mcatanzaro:
WebKitWebDriver is using `soup_server_listen_local()` and passing neither `SOUP_SERVER_LISTEN_IPV4_ONLY` nor `SOUP_SERVER_LISTEN_IPV6_ONLY`. It's _supposed_ to work using IPv4 if only IPv4 is available, but looks like that's not happening. I don't really want to debug it further because containers aren't fun to debug inside. Anyway, if we invoke it like so:

```
$ /usr/bin/WebKitWebDriver --host="127.0.0.1" --port=4444
```

Then we're in business. Now I'll try to change the WPT scripts accordingly....

mcatanzaro:
[Updated the basic dockerfile to work on the environment without IPv6](https://gitlab.gnome.org/GNOME/epiphany/snippets/616). The only change was adding `--webdriver-arg="--host=127.0.0.1"`.

Next up: I'll try foolip's Dockerfile so we can progress with the WPT infrastructure integration.

mcatanzaro:
@foolip I made this change to your Dockerfile.epiphany:

```
$ git diff
diff --git a/tools/docker/Dockerfile.epiphany b/tools/docker/Dockerfile.epiphany
index ef62b05773..f06c64091d 100644
--- a/tools/docker/Dockerfile.epiphany
+++ b/tools/docker/Dockerfile.epiphany
@@ -1,7 +1,7 @@
-FROM fedora:rawhide
+FROM fedora:30
 
-# It might be nice to eventually use flatpak instead of rawhide
-# packages, since rawhide is fragile and prone to breakage.
+# It might be nice to eventually use flatpak instead of Fedora
+# packages, but this probably requires 'docker run --privileged'
 #
 # RUN flatpak remote-add --if-not-exists gnome-nightly https://sdk.gnome.org/gnome-nightly.flatpakrepo && \
 #     flatpak remote-add --if-not-exists gnome-apps-nightly --from https://sdk.gnome.org/gnome-apps-nightly.flatpakrepo
```

Then this beautiful command works for me (remember to update the hash):

```
$ sudo docker run --rm -it 3ed2bd93ba5a /bin/bash --login -c "set -ex; ~/start.sh https://github.com/web-platform-tests/wpt master FETCH_HEAD epiphany; cd ~/web-platform-tests; sudo sh -c './wpt make-hosts-file >> /etc/hosts'; NO_AT_BRIDGE=1 xvfb-run -a -s '-screen 0 1024x768x24' dbus-run-session ./tools/ci/taskcluster-run.py epiphany -- --webdriver-arg='--host=127.0.0.1'"
```

mcatanzaro:
@foolip afaik the only remaining task here is to ensure the CI runs the command in my previous post. In particular, it needs to use `-- --webdriver-arg='--host=127.0.0.1'"` to avoid [the error that you were hitting before](https://github.com/web-platform-tests/wpt/pull/15580#issuecomment-467929857).

The diff in my previous post also switches the CI from rawhide to F30, which would be advisable for now. (That would get you a stable build rather than a devel build, but that seems like a good first step. Ideally we would eventually test both.)

lukebjerring:
@foolip, what's the status of this? @clopez was wondering where we got to with it.

foolip:
The status of this when I left it was that it didn't quite work, but my expectation was that it was just a matter of tackling a series of small problems and that running Epiphany in Taskcluster is fundamentally not a hard thing. To pick this up, here's what I think I'd do:
- Try getting the Dockerfile working to run the /infrastructure/ tests on Fedora 30 or 31
  - If hard try it on a real install of Fedora first, with a window manager and all
- Try again to get Epiphany nightly working with Flatpak locally on any distro
- Evaluate which approach will be most robust for ongoing runs using as recent WebKit code as possible. Anything that rolls out slower than Safari Technology Preview is too slow, I'd say.
- Get it into Taskcluster daily or 6-hourly :)

clopez:
> The status of this when I left it was that it didn't quite work, but my expectation was that it was just a matter of tackling a series of small problems and that running Epiphany in Taskcluster is fundamentally not a hard thing. To pick this up, here's what I think I'd do:
> 
>     * Try getting the Dockerfile working to run the /infrastructure/ tests on Fedora 30 or 31
>    
>   
@foolip Running means that it runs without hanging the /infrastructure/ tests or that it has to pass without failures all those tests?.
Currently I'm getting
```
Ran 63 tests finished in 143.4 seconds.
  • 36 ran as expected. 0 tests skipped.
  • 5 tests had errors unexpectedly
  • 7 tests failed unexpectedly
  • 3 tests timed out unexpectedly
  • 13 tests had unexpected subtest results
```
Is that ok or not?

jgraham:
Some of the infrastructure tests are expected to fail in various ways; you want to run with `--metadata` set to `infrastructure/meta` in order to get the correct expectations for all of these.

foolip:
@clopez some infrastructure tests are explicitly expected to fail, so no, getting them all to pass isn't the goal. Rather, you'll want to understand the failures, can you start by just listing them and checking how other browsers do on the same tests in https://wpt.fyi/results/infrastructure?

If we want to run these tests as part of CI in the future, we'd have to add expectations in infrastructure/metadata for those. https://github.com/web-platform-tests/wpt/blob/936827a6527f1c53051d3bc5bc79304c88c0737f/.azure-pipelines.yml#L96 shows the command line for using the metadata. However, we probably don't need that to begin with.

foolip:
I also have another WIP branch where I made some Taskcluster config changes:
https://github.com/web-platform-tests/wpt/compare/master...foolip:taskcluster-epiphany?expand=1

I'll just leave that here in case it might come to use, it's not actually working either.

clopez:
> @clopez some infrastructure tests are explicitly expected to fail, so no, getting them all to pass isn't the goal. Rather, you'll want to understand the failures, can you start by just listing them and checking how other browsers do on the same tests in https://wpt.fyi/results/infrastructure?

I have added some initial expectations for epiphany and webkitgtk on my wpt clone, here is the commit adding them: https://github.com/clopez/wpt/commit/4e40311dd02aee62a9e99e0e924fc2b4b88cac00

Those expectations make all /infrastructure test pass but /infrastructure/assumptions/ahem.html (which seems to fail for all browsers).

The main issue I find is that on Fedora 31 (rawhide) I get extra failures with the same expectations and version of epiphany/webkitgtk than on Debian experimental or with the Epiphany Tech preview flatpak.

The extra failures I get on Fedora happen on this tests:
/infrastructure/testdriver/actions/elementPosition.html
/infrastructure/testdriver/actions/elementTiming.html
/infrastructure/testdriver/bless.html
/infrastructure/testdriver/click.html
/infrastructure/testdriver/file_upload.sub.html
/infrastructure/testdriver/send_keys.html
And all with the same error:
```
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/executors/executorwebdriver.py", line 286, in _run
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/executors/executorwebdriver.py", line 377, in do_testharness
    done, rv = handler(result)
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/executors/base.py", line 604, in __call__
    return callback(url, payload)
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/executors/base.py", line 618, in process_action
    action_handler(payload)
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/executors/base.py", line 654, in __call__
    element = self.protocol.select.element_by_selector(selector)
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/executors/protocol.py", line 239, in element_by_selector
    elements = self.elements_by_selector_and_frame(element_selector, frame)
  File "/home/test/web-platform-tests/tools/wptrunner/wptrunner/executors/executorwebdriver.py", line 147, in elements_by_selector_and_frame
    return self.webdriver.find.css(element_selector, frame=frame)
  File "/home/test/web-platform-tests/tools/webdriver/webdriver/client.py", line 20, in inner
    return func(self, *args, **kwargs)
  File "/home/test/web-platform-tests/tools/webdriver/webdriver/client.py", line 319, in css
    elements = self._find_element("css selector", element_selector, all)
  File "/home/test/web-platform-tests/tools/webdriver/webdriver/client.py", line 326, in _find_element
    return self.session.send_session_command("POST", route, body)
  File "/home/test/web-platform-tests/tools/webdriver/webdriver/client.py", line 516, in send_session_command
    return self.send_command(method, url, body)
  File "/home/test/web-platform-tests/tools/webdriver/webdriver/client.py", line 480, in send_command
    raise err
JavascriptErrorException: javascript error (500): Unexpected string literal '/'. Expected ')' to end a compound expression.
```

Not sure what is causing this extra failures, but only is reproducible to me on Fedora.
I will try to hunt it down.

clopez:
> Not sure what is causing this extra failures, but only is reproducible to me on Fedora.
> I will try to hunt it down.

I still don't know what causes the issue, but I least now I have a reduced test-case. I suspect it may be related to some compiler bug or compiler flag. I have reported it here: https://bugs.webkit.org/show_bug.cgi?id=200894

clopez:
> I still don't know what causes the issue, but at least now I have a reduced test-case. I suspect it may be related to some compiler bug or compiler flag. I have reported it here: https://bugs.webkit.org/show_bug.cgi?id=200894

The issue has been found and it affects also Fedora stable (30). Some of the WebKitGTK+ internal JavaScript code is broken when WebKitGTK+ is built with python3 instead of python2. This causes (among other things) internal errors in the WebKitWebDriver code.

A fix is being worked on, but it will take some time until it reaches Fedora.

In the meantime, to not block here (and if you agree) I will try to continue with a dockerfile for Debian unstable. Debian WebKitGTK+ builds are done with python2, so they are not affected by this.

lukebjerring:
Sounds like the faster route to producing results :)

clopez:
I have opened PR #18595 with a dockerfile based on Debian

foolip:
Closing in favor of https://github.com/web-platform-tests/wpt/pull/18595, thanks @clopez!

PR #15582, opened 2019-02-27
Fix tabs and trailing whitespace in a couple of files.

LukeZielinski:
In preparation for importing the tools directory into Chromium repo, we
want to avoid running into lint warnings for these files.

### DIFF
--- a/tools/scripts/manifest.js
+++ b/tools/scripts/manifest.js
@@ -2,119 +2,119 @@
 function __result_handler() {
 
     function __get_metadata() {
-	var obj = new Object();
-	var author = [];
-	var assert = [];
-	var help = [];
-	var match = [];
-	var mismatch = [];
-	var flags = [];
-	var nodes;
+        var obj = new Object();
+        var author = [];
+        var assert = [];
+        var help = [];
+        var match = [];
+        var mismatch = [];
+        var flags = [];
+        var nodes;
 
-	nodes = document.querySelectorAll('link[rel="author"]');
-	for (var i = 0; i < nodes.length; i++) {
-	    var href = nodes[i].getAttribute("href");
-	    var title = nodes[i].getAttribute("title");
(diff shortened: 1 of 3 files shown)


jugglinmike:
If WPT isn't linting for this, will there be a risk of regressing in Chromium?

LukeZielinski:
Yeah there is a risk. The effect will be that our importer will choke on any PR that introduces new lint errors, and I'll presumably have to fix it in order to stay in sync. I'm actually not sure how much of a burden this will be.

Note that WPT does lint for these types of things, but we unfortunately whitelist the entire tools directory (with the reason being "third party code"). This seems wrong - wouldn't we want to lint the tools directory and only whitelist tools/third_party/*? There is definitely a massive lint debt in tools right now (almost 30k errors), but is there an appetite for whittling that down? @foolip @jgraham for opinions?

tools/scripts/manifest.js
```diff
@@ -2,139 +2,139 @@
 function __result_handler() {
```
jgraham: This file can/should just be deleted.

jgraham (approved):

jgraham:
A bunch of the tools stuff is linted by flake8, so the problem isn't as bad as you make out since all flake8-covered code, which is most of it, is actually checked for tabs and training spaces. We could turn the whitespace lints on in wpt lint for the whole directory if that would help you.

foolip:
I'm not too concerned about import being broken by new whitespace errors being introduced in tools/. It probably will happen, and we could replace `*: tools/*` in lint.whitelist with something more selective, but any mishap is easily fixed, so if that turns messy I would wait for the first time it goes wrong to fix it.

LukeZielinski:
This sounds like a plan. For now, fix lint errors as we go and if it gets to be too much work fix lint.whitelist to check more of tools.

PR #15587, opened 2019-02-27
Update easing-tests.js from m-c

jgraham:
It seems like a merge error resulted in this file being incorrectly synced with gecko.
Revert to the gecko version.

### DIFF
--- a/css/css-typed-om/the-stylepropertymap/properties/animation-timing-function.html
+++ b/css/css-typed-om/the-stylepropertymap/properties/animation-timing-function.html
@@ -24,7 +24,7 @@ runListValuedPropertyTests('animation-timing-function', [
 ]);
 
 runUnsupportedPropertyTests('animation-timing-function', [
-  'cubic-bezier(0.1, 0.7, 1.0, 0.1)', 'steps(4, end)', 'frames(10)'
+  'cubic-bezier(0.1, 0.7, 1.0, 0.1)', 'steps(4, end)'
 ]);
 
 </script>
--- a/css/css-typed-om/the-stylepropertymap/properties/transition-timing-function.html
+++ b/css/css-typed-om/the-stylepropertymap/properties/transition-timing-function.html
@@ -24,7 +24,7 @@ runListValuedPropertyTests('transition-timing-function', [
 ]);
 
 runUnsupportedPropertyTests('transition-timing-function', [
-  'cubic-bezier(0.1, 0.7, 1.0, 0.1)', 'steps(4, end)', 'frames(10)'
+  'cubic-bezier(0.1, 0.7, 1.0, 0.1)', 'steps(4, end)'
 ]);
 
 </script>
--- a/web-animations/resources/easing-tests.js
+++ b/web-animations/resources/easing-tests.js
@@ -41,11 +41,6 @@ const gEasingTests = [
     serialization: 'steps(2)'
   },
   {
-    desc: 'frames function',
-    easing: 'frames(5)',
-    easingFunction: framesTiming(5)
-  },
-  {
     desc: 'linear function',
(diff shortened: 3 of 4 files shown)


stephenmcgruer (approved):

birtles:
@jgraham We still appear to be missing the changes to the following files from [same patch](https://hg.mozilla.org/mozilla-central/rev/b46aa2883e42):

* web-animations/testcommon.js
* css/css-typed-om/the-stylepropertymap/properties/animation-timing-function.html
* css/css-typed-om/the-stylepropertymap/properties/transition-timing-function.html

birtles (changes requested):
We still appear to be missing the changes to the following files from [same patch](https://hg.mozilla.org/mozilla-central/rev/b46aa2883e42):
* web-animations/testcommon.js
* css/css-typed-om/the-stylepropertymap/properties/animation-timing-function.html
* css/css-typed-om/the-stylepropertymap/properties/transition-timing-function.html

jgraham:
I added the other changes from the files mentioned.

gsnedders:
TC stability failure partly due to https://github.com/web-platform-tests/wpt/issues/15596 (other failures look genuine?).

birtles (approved):

birtles:
Thanks! Looks good to me.

gsnedders:
@stephenmcgruer given the changes and the flakiness described at the end of the [wpt-chrome-stability](https://tools.taskcluster.net/groups/bEe7MFRdShqakLj-QpHp4g/tasks/IyCRct5aTr2Y1j6k0-1O8w/runs/0/logs/public%2Flogs%2Flive.log) job, I presume we should treat this as pre-existing flakiness?

stephenmcgruer:
If I understand that job correctly, all the unstable results are AnimationWorklet (why were they run for this change? Is their flakiness blocking changes across all of web-platform-tests?). If so then yes please treat as existing flakiness. @majido owns making the AW tests not flaky.

gsnedders:
@stephenmcgruer because they rely on `/web-animations/testcommon.js`

majido:
I am working on reducing flakiness on these tests but I suggest you disable the flaky ones (here is the [list](https://cs.chromium.org/chromium/src/third_party/blink/web_tests/TestExpectations?g=0&l=5949) in Chromium) until they are fixed.

### Context
AFAICT the standard setup in Chromium is to run wpt tests without threaded compositing. This has masked the flakiness (caused by threaded animation) in some of these tests. I just enabled running these with [threaded compositing](https://chromium-review.googlesource.com/c/chromium/src/+/1492436) which is closer to wpt.fyi configuration (and I assume your bots) which does not use content_shell . This has exposed the flakiness and I am gradually fixing [these](https://chromium-review.googlesource.com/c/chromium/src/+/1495227).

As a side note, I feel we should have at least a FYI bot in chromium that runs the same configuration (or as close as possible) to upstream bots so that we catch these type of issues earlier. @foolip  do you have any ideas if this is possible?

majido:
@gsnedders so the patch changes the common file and bots run all test that include that file which involves these flaky tests.

Removing usage of `/web-animations/testcommon.js` in these tests is not to difficult (just duplicating a few functions) but I think the right solution is to make them as flaky. wdyt?

foolip:
> As a side note, I feel we should have at least a FYI bot in chromium that runs the same configuration (or as close as possible) to upstream bots so that we catch these type of issues earlier. @foolip do you have any ideas if this is possible?

@LukeZielinski is working on using wptrunner in Chromium. It's not clear yet if it'll use the chrome or content_shell binary however.

foolip:
@Hexcles what do you think about running stability checks for new tests in Chromium CQ? The other alternative we've discussed is reflecting back the status from Taskcluster into Gerrit code reviews, which would also show flakiness in other browsers.

PR #15591, opened 2019-02-28
Fix updates from Bug 1513959

jgraham:
Seems we failed to handle the backout correctly and landed the initial patch rather than
the different one that finally stuck


gsnedders:
TC stability failure due to https://github.com/web-platform-tests/wpt/issues/15596.

gsnedders:
This looks like there might genuinely be flakiness now?

gsnedders:
`/css/css-sizing/range-percent-intrinsic-size-1.html` seems to have come back with a totally blank screenshot once from Firefox. That's very weird.

PR #15592, opened 2019-02-28
Screen Orientation: Browsing context pending promise rejection test

Johanna-hub:
A first attempt at testing the following part of the `Applying an orientation lock algorithm:
`
> 2. The following sub-steps MAY be run asynchronously for performance reasons, for example, if the user agent has browsing contexts living in different processes:
>     1. Let browsing contexts be the list of the descendant browsing contexts of the top-level browsing  context's document.
>     2. If one of the browsing contexts's document's [[orientationPendingPromise]] is not null:
>        1. Let doc be the document which has a not null [[orientationPendingPromise]].
>        2. Reject doc's [[orientationPendingPromise]] with DOMException whose name is AbortError.
>        3. Set doc's [[orientationPendingPromise]] to null .

In browsing-contexts.html - I am just attempting to lock the screen on load.  In iframe-lock, I am trying to also lock and hopefully that promise will be rejected.  I have put the iframe code in resources, but that is actually where the test is so not sure if that is right?


Johanna-hub:
@marcoscaceres I now have empty iframe file and moved code to the test file.  I had to make a few changes to your suggestions to make it work, so see what you think.

After testing out only the iframe code:
on Chrome mobile: goes full screen and locks to landscape completely fine
on Chrome desktop: iframe goes full screen and then get the usual operation not supported error
on Firefox desktop: iframe goes full screen fine and then get the usual operation not supported error
on Firefox mobile: goes full screen and then `SecurityError: The operation is insecure`
Not sure if you have any suggestions about what to do for Firefox mobile?

marcoscaceres:
That’s excellent. Congrats on finding yet another web compat bug! That’s a bad one too. Same origin iframe should be able to switch orientation.

marcoscaceres:
Ok, in a new test, you need to add this to the iframe:

`<iframe sandbox="allow-orientation-lock" src=....`

You shouldn’t need that normally for same origin iframes, so that’s definitely a Firefox bug.

Johanna-hub:
Doing that seems to stop the iframe from locking in Chrome and it still doesn't work in Firefox...

Johanna-hub:
I've refactored the test to see whether iframe or outer window locks, see what you think!

I've also added the test for locked iframe navigating to new top-level browsing context.  I've run into the same issue that i did here #15705 - when I test on my phone it's easy to see if the screen is now unlocked as I can just turn it and screen will rotate.  Unsure how to test this here.  So at the moment just seeing that orientation is staying the same (since it is just unlocked not locked to new orientation) until we figure it out!

Johanna-hub:
Added pop-up window test

screen-orientation/browsing-contexts.html
```diff
@@ -0,0 +1,93 @@
...
+    await test_driver.bless("request full screen", () => {
+      return document.documentElement.requestFullscreen();
+    });
+    const { type: currentOrientation } = screen.orientation.type;
+    const isPortrait = currentOrientation.includes("portrait");
+    const newOrientation = `${isPortrait ? "landscape" : "portrait"}`;
+    const promiseToLockIFrame = iframe.contentWindow.screen.orientation.lock(
+      `${newOrientation}-secondary`
+    );
+    await screen.orientation.lock(`${newOrientation}-primary`);
```
marcoscaceres: here we want to again capture the promise `const p = screen.orientation.lock(`${newOrientation}-primary`);`...

The `await p` before calling `unlock()`...

screen-orientation/browsing-contexts.html
```diff
@@ -0,0 +1,93 @@
...
+
+  promise_test(async t => {
+    const iframe = await loadIframe();
+    const promiseToChange = new Promise(resolve => {
+      screen.orientation.addEventListener("change", resolve);
+    });
+    await test_driver.bless("request full screen", () => {
+      return document.documentElement.requestFullscreen();
+    });
+    const { type: originalOrientation } = screen.orientation.type;
```
marcoscaceres:
```suggestion
    const { type: originalOrientation } = screen.orientation;
```

screen-orientation/browsing-contexts.html
```diff
@@ -0,0 +1,93 @@
...
+      iframe.onload = () => r(iframe);
+    });
+  }
+
+  promise_test(async t => {
+    const iframe = await loadIframe();
+    await test_driver.bless("request full screen", () => {
+      return document.documentElement.requestFullscreen();
+    });
+    const { type: currentOrientation } = screen.orientation.type;
```
marcoscaceres:
```suggestion
    const { type: currentOrientation } = screen.orientation;
```

screen-orientation/browsing-contexts.html
```diff
@@ -0,0 +1,93 @@
...
+    await promiseToChange;
+    assert_equals(
+      screen.orientation.type,
+      originalOrientation,
+      "Navigating from iframe to new top-level browsing context unlocks orientation"
+    );
+    screen.orientation.unlock();
+    await document.exitFullscreen();
+    iframe.remove();
+  }, "Iframe navigation test");
```
marcoscaceres:
```suggestion
  }, "Navigating a nested browsing context causes the orientation to unlock()");
```

screen-orientation/browsing-contexts.html
```diff
@@ -0,0 +1,93 @@
...
+
+  promise_test(async t => {
+    const promiseToChange = new Promise(resolve => {
+      screen.orientation.addEventListener("change", resolve);
+    });
+    await test_driver.bless("request full screen", () => {
+      return document.documentElement.requestFullscreen();
+    });
+    const { type: originalOrientation } = screen.orientation.type;
```
marcoscaceres:
```suggestion
    const { type: originalOrientation } = screen.orientation;
```

screen-orientation/browsing-contexts.html
```diff
@@ -0,0 +1,93 @@
...
+    const isPortrait = originalOrientation.startsWith("portrait");
+    const newOrientation = `${isPortrait ? "landscape" : "portrait"}-primary`;
+    await screen.orientation.lock("newOrientation");
```
marcoscaceres:
```suggestion
    await screen.orientation.lock(newOrientation);
```

marcoscaceres (changes requested):
Couple of bugs...

screen-orientation/browsing-contexts.html
```diff
@@ -63,18 +62,25 @@
...
-    await test_driver.bless("request full screen", () => {
-      return document.documentElement.requestFullscreen();
+    const loadPromise = new Promise(r => {
+      if (popWindow.document.readyState === "complete") {
+        r();
+      }
+      popWindow.onload = () => r();
     });
+    await loadPromise;
     popWindow.onload = function navigation() {
```
marcoscaceres: This is now redundant, I guess. Eventually, after orientation change, you want to navigate to: 
```
popWindow.location.href = "https://not-web-platform.test:8443";
```

Which should drop popup out of fullscreen, and change presumedly the orientation.

screen-orientation/browsing-contexts.html
```diff
@@ -5,8 +5,8 @@
 <script src="/resources/testdriver-vendor.js"></script>
 
 <script>
-  import { loadIframe } from "./resources/utils.js";
-  import { getOppositeOrientation } from "./resources/utils.js";
+  import { loadIframe } from "/screen-orientation/resources/orientation-utils.js";
```
marcoscaceres: nit, combine these two:
```
import {
  getOppositeOrientation,
  loadIframe,
} from "/screen-orientation/resources/orientation-utils.js";
```

marcoscaceres:
@haoxli, we could really use some help running these tests... and the rest of the screen orientation ones. Is that something you can do on Chromium infrastructure?

PR #15593, opened 2019-02-28
Add support for verifying taskcluster config

jgraham:
Adds as wpt tc-verify command that verifies that the TaskCluster
config is a valid yaml file and computes the tasks that will run on a
PR synchronize event. This can be expanded to more events and pushes
in the future.


tools/taskcluster/verify.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def create_parser():
+    return argparse.ArgumentParser()
+
+
+def run(venv, **kwargs):
+    with open(os.path.join(root, ".taskcluster.yml")) as f:
+        template = yaml.safe_load(f)
+    with open(os.path.join(here, "testdata/pr.json")) as f:
```
gsnedders: Can we call this `pr_event.json` so it's clearer what it is?

gsnedders (approved):

PR #15597, opened 2019-02-28
Fix #15596: change window after reftest:teardown

### LINKED ISSUE #15596: Firefox stability jobs failing when running reftests
appears to be regression from https://github.com/web-platform-tests/wpt/pull/15488 (cc @jugglinmike); @jgraham suspects Marionette's `reftest:teardown` doesn't actually work.

gsnedders:
Fixes #15596.

### DIFF
--- a/tools/wptrunner/wptrunner/executors/executormarionette.py
+++ b/tools/wptrunner/wptrunner/executors/executormarionette.py
@@ -894,6 +894,10 @@ class InternalRefTestImplementation(object):
         except Exception as e:
             # Ignore errors during teardown
             self.logger.warning(traceback.format_exc(e))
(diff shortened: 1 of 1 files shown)


jgraham (changes requested):
This should go inside the exception handler and the if block.

gsnedders:
@jgraham do you know why we catch *all* exceptions there? there's definitely plenty of cases where we want to treat teardown as having failed if an exception is thrown, no?

jgraham:
I don't think we ever want to fail the job for an exception during teardown unless it causes some other problem. Your code is broken if we get an timeout during startup, for example, because in that case there are no windows.

jgraham (approved):

PR #15600, opened 2019-02-28
Add tests for deduplicating unforgeable member functions.

Ms2ger:
Spec change in <https://github.com/heycam/webidl/pull/668>.


annevk (approved):
Thanks! For future tests, please consider adopting const/let for extra clarity.

Issue #14389, opened 2018-12-06
[web-nfc] Update tests to use new NFCReader and NFCWriter object


Honry:
Reflects to latest spec change at https://github.com/w3c/web-nfc/pull/177 and https://github.com/w3c/web-nfc/pull/179

Honry:
Four more spec changes that require updating tests: 
- Now that we use event listeners for watches, remove AbortSignal https://github.com/w3c/web-nfc/pull/181
- Move watch options to NFCReader ctor https://github.com/w3c/web-nfc/pull/180
- As event listeners cannot have side-effects, intro start/stop https://github.com/w3c/web-nfc/pull/183
- Introduce error event handler and make start not return a promise  https://github.com/w3c/web-nfc/pull/184

I will start this task once spec becomes a bit more stable.

Honry:
Fixed by #15616, #15646, #15647, #15648, #15666.

PR #15616, opened 2019-03-01
Update web-nfc recordType tests

xiuqijix:
part of #14389

### DIFF
--- a/web-nfc/nfc_recordType_empty-manual.https.html
+++ b/web-nfc/nfc_recordType_empty-manual.https.html
@@ -22,13 +22,13 @@
 setup({ explicit_timeout: true });
 
 const desc = "Test that write and read data succeed when recordType is set to 'empty'.";
-const watchOptions = {
+const readOptions = {
   recordType: "empty",
   mediaType: "",
   url: ""
 }
 const message = createMessage([createRecord('empty', '')]);
 
-testNDEFMessage(message, watchOptions, desc);
+testNDEFMessage(message, readOptions, desc);
 
 </script>
--- a/web-nfc/nfc_recordType_empty-manual.https.html
+++ b/web-nfc/nfc_recordType_empty-manual.https.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <meta charset=utf-8>
-<title>Web NFC Test: push and read empty records</title>
+<title>Web NFC Test: write and read empty records</title>
 <link rel="author" title="Intel" href="http://www.intel.com"/>
 <link rel="help" href="https://w3c.github.io/web-nfc/"/>
 <meta name="timeout" content="long">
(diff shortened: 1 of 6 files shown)


web-nfc/resources/nfc_help.js
```diff
@@ -68,14 +68,13 @@ function assertWebNDEFMessagesEqual(a, b) {
...
-    return navigator.nfc.push(pushedMessage)
-      .then(() => {
-        return new Promise(resolve => {
-          navigator.nfc.watch((message) => resolve(message), watchOptions);
-        }).then((message) => {
-          assertWebNDEFMessagesEqual(message, pushedMessage);
-        });
-      });
+  promise_test(async t => {
+    let writer = new NFCWriter();
```
Honry: Nit: s/let/const

web-nfc/nfc_recordType_empty-manual.https.html
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <meta charset=utf-8>
-<title>Web NFC Test: push and watch empty records</title>
+<title>Web NFC Test: push and read empty records</title>
```
Honry: Nit: s/push/write

kaixinjxq: @Honry, we need replace `push` with `write` in all descriptions, right?

web-nfc/resources/nfc_help.js
kaixinjxq: The same as above, we need replace all `let` with `const`, or only this one?

web-nfc/nfc_recordType_empty-manual.https.html
Honry: Right.

web-nfc/resources/nfc_help.js
Honry: All `let` could be replaced with `const`.

kaixinjxq:
@Honry, many thanks for your review! I just submit a new commit to address your comments, PTAL.

web-nfc/nfc_recordType_opaque-manual.https.html
```diff
@@ -21,13 +21,13 @@
 
 setup({ explicit_timeout: true });
 
-let desc = "Test that nfc.push and nfc.watch succeed when recordType is set to 'opaque'."
-let watchOptions = {
+const desc = "Test that write and read data succeed when recordType is set to 'opaque'."
+const watchOptions = {
```
Honry: Nit: s/watchOptions/readOptions
Please update the rest with same name. :)

kaixinjxq: Updated

Honry (approved):
Thanks, LGTM

PR #15627, opened 2019-03-01
Unset the executable bit of non-executable files

Hexcles:
Using this command:
```bash
find . -perm 755 -type f -not -name "*.py" -not -wholename "./tools/*" \
-not -wholename "./.git/*" -not -wholename "./_venv/*" \
-not -name "wpt" -not -name "*.sh" -not -name "regenerate" \
-not -name "*.py" -not -name "*.rb" | xargs chmod -x
```

Note that the excluded list is meant to be conservative (e.g. some .py
or .rb files probably shouldn't have +x as well).

We should add a lint rule for this later.


inexorabletash (approved):
lgtm!

Hexcles:
@jgraham @gsnedders do we have a lint rule for file permissions? I vaguely remember so but can't find it.

jgraham (approved):

gsnedders:
@Hexcles no we don't

PR #15629, opened 2019-03-01
fix(webvtt): one-line cue and wrapped-cue example

gkatsev:
In my reading of the spec, cues are rendered one at a time and once a
cue has been rendered, it isn't allowed to move. This means, that two
cues that happen at the same time would be rendered starting with the
first cue at the bottom and rendering up subsequent cues.


silviapfeiffer:
Yes, there is no cue-moving without regions.

silviapfeiffer (approved):

PR #15631, opened 2019-03-01
[csp] Wait for all tests to be defined

jugglinmike:
Configure the test harness to delay completion until all
asynchronously-defined tests are available.

### DIFF
--- a/content-security-policy/generic/generic-0_2.html
+++ b/content-security-policy/generic/generic-0_2.html
@@ -7,12 +7,19 @@
     <script src='/resources/testharnessreport.js'></script>
     <script nonce='abc'>
       var t_spv = async_test("Should fire violation events for every failed violation");
-      window.addEventListener("securitypolicyviolation", t_spv.step_func_done(function(e) {
-          assert_equals(e.violatedDirective, "script-src");
-      }));      
+      window.addEventListener(
+        "securitypolicyviolation", t_spv.unreached_func("securitypolicyviolation should not be emitted"));
+
+      window.addEventListener("load", function() {
+        t_spv.done();
+      });
     </script>
     <script src='positiveTest.js'></script>
-    <script nonce='abc'>t_spv.done();</script>
+    <script nonce='abc'>
+      test(function() {
+        assert_true(window.cspPositiveTest);
+      }, "Allows scripts from the same host.");
+    </script>
 </head>
 <body>
     <h1>'self' keyword positive test</h1>
--- a/content-security-policy/reporting/report-only-in-meta.sub.html
+++ b/content-security-policy/reporting/report-only-in-meta.sub.html
@@ -33,6 +33,12 @@
       script.defer = true;
       script.src = '../support/checkReport.sub.js?reportExists=false'
       document.body.appendChild(script);
+
+      // Immediately declare a test so that the harness does not infer
+      // completion if the image loads before the script.
+      var checkReportTest = async_test("checkReport tests loaded");
+      script.onload = checkReportTest.step_func_done();
+      script.onerror = checkReportTest.unreached_func();
     });
   </script>
 </body>
--- a/content-security-policy/reporting/report-uri-from-child-frame.html
+++ b/content-security-policy/reporting/report-uri-from-child-frame.html
@@ -8,8 +8,6 @@
 </head>
 <body>
     <script nonce="abc">
-      var t1 = async_test("Check that we received a message from the child frame");
-
       window.onmessage = function(e) {
         if (e.data == 'cookie set') {
           var s = document.createElement('script');
@@ -17,8 +15,6 @@
           s.defer = true;
           s.src = '../support/checkReport.sub.js?reportField=violated-directive&reportValue=script-src%20%27self%27%20%27nonce-abc%27&reportCookieName=generate-csp-report';
           document.body.appendChild(s);
-
-          t1.done();
         }
       }
     </script>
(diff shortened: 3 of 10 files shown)


content-security-policy/generic/generic-0_2.html
```diff
@@ -7,12 +7,19 @@
     <script src='/resources/testharnessreport.js'></script>
     <script nonce='abc'>
       var t_spv = async_test("Should fire violation events for every failed violation");
-      window.addEventListener("securitypolicyviolation", t_spv.step_func_done(function(e) {
-          assert_equals(e.violatedDirective, "script-src");
-      }));      
+      window.addEventListener(
+        "securitypolicyviolation", t_spv.unreached_func("securitypolicyviolation should not be emitted"));
```
andypaicu: This expectation seems to have flipped, was this test incorrect to begin with?

content-security-policy/reporting/report-uri-from-child-frame.html
```diff
@@ -8,17 +8,13 @@
 </head>
 <body>
     <script nonce="abc">
-      var t1 = async_test("Check that we received a message from the child frame");
```
andypaicu: I believe this was here precisely to have a declared test so the testing harness won't immediately conclude the test when the page loads. I think this test now will succeed immediately without waiting for a message.

content-security-policy/reporting/report-only-in-meta.sub.html
```diff
@@ -33,6 +33,12 @@
       script.defer = true;
       script.src = '../support/checkReport.sub.js?reportExists=false'
       document.body.appendChild(script);
+
+      // Immediately declare a test so that the harness does not infer
+      // completion if the image loads before the script.
+      var checkReportTest = async_test("checkReport tests loaded");
+      script.onload = checkReportTest.step_func_done();
```
andypaicu: Is there a chance of the script element onload event being fired before we set the handler? Could we move the document.body.appendChild(script); line after this?

jugglinmike: DOM event handlers do not run in parallel, but I'm happy to relocate if you feel it improves readability.

content-security-policy/reporting/report-uri-from-child-frame.html
jugglinmike: In the absence of any tests, testharness.js will not react to the `load` event. Instead, it will wait for the global `done` function to be invoked (as an indication of a single-page test) or for a sub-test to be defined. This test defines one sub-test via `checkReport.sub.js`, and since that is the only test it defines, there's no need to explicitly manage scheduling.

More than being superfluous, the removed sub-test was a source of instability. It completed before the meaningful sub-test was defined, so there was a race condition regarding when testharness.js determined completion.

content-security-policy/generic/generic-0_2.html
jugglinmike: Yes, it was incorrect. There are no CSP violations, so the event should not have been emitted at all (regardless of the value of the `violatedDirective` property). Also, the sub-test was written to complete in two different ways: via a subsequent inline script and via the `step_func_done` method.

content-security-policy/reporting/report-uri-from-child-frame.html
andypaicu: Ah ok, I did not know it behaves like that when there are no tests.

content-security-policy/reporting/report-only-in-meta.sub.html
andypaicu: I think it's nice to set all the properties on the element before adding it to the DOM but it does not really matter.

andypaicu (approved):

content-security-policy/generic/generic-0_2.html
andypaicu: Ok

jugglinmike:
Thanks for the review, @andypaicu! Would you mind merging this?

jugglinmike:
@andypaicu seems like this may have slipped off your radar. Would you mind merging it?

foolip:
@jugglinmike, if approved please do merge yourself, it reduces round-trips and would have avoided the possibility of me trying to fix the same bug.

jugglinmike:
@foolip I interpret an approval that isn't immediately followed by merging to mean that further review is expected. If there is substantive feedback or merge conflicts, it's more clear that the ball's in my court. That's not the case here, though. This isn't the first time I've seen this in WPT, so I may be misunderstanding something. How do you interpret this situation?

foolip:
@jugglinmike this is a matter of convention and not written down anywhere, but it's very common for people to approve with no comment without merging. I see @Ms2ger do it, and I do it. IMHO it's a good idea to do so if you're not sure if the author might have hoped for someone else to review. Approving with nits and trusting the author to fix those and then merge is also common, typically as "LGTM % nits".

Basically if merging is technically possible it's OK. Reviewers will learn to not approve if that's not what they mean :)

jugglinmike:
Got it. I'll miss having the approval memorialized in the git history, but I'm probably overthinking things. Thanks for the explanation!

foolip:
I love a good git history, but for this project I don't have high hopes... We have loads of automatic exports with commit messages that are nonsensical taken out of the original context. And people squash-merge PRs without cleaning up the commit messages, making them just a bullet list of somewhat random phrases. Often the useful explanation is in PRs and not in the commit messages.

The wrong committer won't stand out much against this background :D

PR #15633, opened 2019-03-01
[visual viewport] Wait for all tests to be defined

jugglinmike:
Configure the test harness to delay completion until all
asynchronously-defined tests are available.

Issue #14533, opened 2018-12-14
Multiple same-name subtests not always triggering a harness error


mdittmer:
E.g., http://w3c-test.org/css/css-values/calc-unit-analysis.html

mdittmer:
Further context:

The link claims 18 tests, but only 9 results inside the results script tag. When these tests are run and the results are sent to a JSON report, 18 subtest entries (not 9) appear, along with test status of `OK`. [Example run report](https://storage.googleapis.com/wptd-results-staging/48d5d5033d2f634125d0ae2dd494d68f8d01de79/chrome-72.0.3626.14_dev-linux-18.04-f7742a2216/report.json).

jugglinmike:
Taking a look now!

lukebjerring:
@jugglinmike - did you end up looking into this?

jugglinmike:
I did! I'll follow up with a few patches shortly

jugglinmike:
This is a result of tests which erroneously declare sub-tests following harness-level completion. In the interest of speed and stability, I think the best we can do (beyond being more careful when writing tests) is silently ignore those sub-tests. That's what I'm proposing in gh-15634, anyway.

foolip:
I've reviewed/merged https://github.com/web-platform-tests/wpt/pull/15634, closing this.

PR #15634, opened 2019-03-01
[testharness.js] Tolerate late tests

jugglinmike:
We might try to signal a harness error whenever this occurs, but a few details eventually convinced me to silently ignore it:

First, correctly reporting this condition would require a large amount of structural change. testharness.js was not designed for a "COMPLETE to COMPLETE" transition, so things like updating the DOM will take some doing. In gh-14533, @mdittmer noted a discrepancy between the reported data and the JSON written to the document--that's a side effect of the harness emitting an array which it later extends.

It's also not clear how this would effect `test_done_callback`. It would be tricky to name/document if we allowed for functions registered with that API to be invoked more than once. Alternatively, we could implement some kind of "all clear" period, where the harness sits idle waiting for late tests before reporting completion. I don't know how to balance the desire to guard against this edge case against our interest to minimize the duration of test execution.

Finally, we can't detect this situation deterministically. We currently ignore late tests--just in an inconsistent way. If we tried to disallow them, then we might prevent some test bugs, but we might also allow test bugs to detract from stability.

There are a handful of tests which exhibit this behavior today. I've submitted corrections for those in separate patches:

- https://github.com/web-platform-tests/wpt/pull/15631
- https://github.com/web-platform-tests/wpt/pull/15632
- https://github.com/web-platform-tests/wpt/pull/15633

### DIFF
--- a/resources/test/tests/unit/late-test.html
+++ b/resources/test/tests/unit/late-test.html
@@ -1,0 +1,54 @@
+<!DOCTYPE HTML>
+<html>
+<head>
+<title>Test declared after harness completion</title>
+</head>
+<body>
+<div id="log"></div>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<p>This test simulates an automated test running scenario, where the test
+results emitted by testharness.js may be interpreted after some delay. It is
+intended to demonstrate that in such cases, any additional tests which are
+executed during that delay are <em>not</em> included in the dataset.</p>
+
+<p>Although these "late" tests are likely an indication of a mistake in test
+design, they cannot be detected deterministically, so in the interest of
+stability, they should be silently tolerated.</p>
+<script>
+async_test(function(t) {
+    var source = [
+        "<div id='log'></div>",
+        "<script src='/resources/testharness.js'></" + "script>",
+        "<script src='/resources/testharnessreport.js'></" + "script>",
+        "<script>",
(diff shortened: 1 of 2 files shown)


gsnedders:
> We currently ignore late tests--just in an inconsistent way.

How are we inconsistent about it?

jugglinmike:
Sub-tests defined after `onload` (or after `done` for tests configured with `explicit_done`) are recognized if they are defined while some asynchronous sub-test is running.

In general, it's hard to say if the timing of such cases is deterministic. My third point above is if we trigger a harness failure when we see a sub-test defined at that moment, we might make the suite more flaky overall.

Then again, it will still be flaky--just in a number of sub-tests. That will be more of an issue in cases where the late sub-tests fail.

If this were the only argument against rejecting late sub-tests, I think I could go either way. I submitted this patch because there are some additional considerations that favor it (see above).

gsnedders (approved):
LGTM. Would like comments from others, though, as to whether they agree this is a good idea.

jgraham:
Defining subtests after `onload` but when another test is still running is a pretty common pattern I think; one way of handling a sequence of tests is to have the current test schedule the next test just before it completes. There are obviously problems with this approach, but I think that breaking it would be bad.

As far as I can tell this patch doesn't do that and it's just a bugfix to the existing behaviour which is intended to be that once you're in `COMPLETE` state nothing else can cause a test to run.

foolip (approved):

PR #15649, opened 2019-03-04
HTML: test queueing behavior of window.close()

annevk:
The reason the second test fails in Firefox is due to `window.open()` using a nested event loop. I could make Firefox pass the test by removing the `openee.opener` assert, but seems bad?

### DIFF
--- a/html/browsers/the-window-object/close-method.window.js
+++ b/html/browsers/the-window-object/close-method.window.js
@@ -1,0 +1,30 @@
+async_test(t => {
+  const openee = window.open();
+  assert_false(openee.closed);
+  openee.close();
+  assert_true(openee.closed);
+  assert_equals(openee.opener, self);
+  t.step_timeout(() => {
+    assert_equals(openee.opener, null);
+    t.done();
+  }, 100);
+}, "window.close() queues a task to discard, but window.closed knows immediately");
+
+async_test(t => {
+  const openee = window.open("", "greatname");
+  assert_false(openee.closed);
+  openee.close();
+  assert_true(openee.closed);
+  const openee2 = window.open("", "greatname");
+  assert_not_equals(openee, openee2);
+
+  assert_equals(openee.opener, self);
+  openee2.close();
+  assert_equals(openee2.opener, self);
+
+  t.step_timeout(() => {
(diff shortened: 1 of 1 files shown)


annevk:
@foolip assigning this to you due to unclear Azure failure.

annevk:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1533416 on Firefox.

html/browsers/the-window-object/close-method.window.js
```diff
@@ -0,0 +1,30 @@
+async_test(t => {
+  const openee = window.open();
+  assert_false(openee.closed);
```
domenic: Since you're doing sanity checks anyway might as well check .opener too.

html/browsers/the-window-object/close-method.window.js
```diff
@@ -0,0 +1,30 @@
...
+  openee.close();
+  assert_true(openee.closed);
+  assert_equals(openee.opener, self);
+  t.step_timeout(() => {
+    assert_equals(openee.opener, null);
+    t.done();
+  }, 250);
```
domenic: Just realized: could you make these tests reliable by using the pagehide or unload event, plus a setTimeout(,0)? Seems like it'd be a tighter test too.

html/browsers/the-window-object/close-method.window.js
```diff
@@ -0,0 +1,30 @@
+async_test(t => {
+  const openee = window.open();
+  assert_false(openee.closed);
+  openee.close();
+  assert_true(openee.closed);
+  assert_equals(openee.opener, self);
+  t.step_timeout(() => {
+    assert_equals(openee.opener, null);
```
domenic: Again doesn't hurt to also check .closed here. Basically always check them together.

domenic (commented):
I think it's worth trying to use events instead of timeouts, if we can.

annevk:
Thanks, events do indeed work due to this being same-origin.

annevk:
Safari bug is referenced from https://github.com/whatwg/html/pull/4402.

domenic (approved):

PR #15650, opened 2019-03-04
Call pkg_resources.require rather than always calling out to pip

gsnedders:
This makes a decent perf gain in the case when all the requirements are already met.


jgraham (approved):

gsnedders:
Ah, we need something like https://github.com/pypa/pip/blob/04f1e70427fd392e1a68c13174816b1b21f48baa/src/pip/_internal/req/req_install.py#L400 to deal with the case of having an incompatible version in the `site-packages`.

jgraham (approved):

PR #15655, opened 2019-03-04
Split wptrunner formatters into separate files in their own directory

LukeZielinski:
We have 2 formatters (wptreport, wptscreenshot) in the same file (formatters.py) right now. A third formatter is being created and, instead of adding it to formatters.py, I propose creating a new `formatters` directory and splitting each formatter into its own file.


Hexcles (approved):
LGTM.

A bit more context for others: Luke wants to add yet another formatter (for better integration with Chromium infra), so we figure it'd be better to create a directory for formatters for better code health.

Hexcles (changes requested):
Luke, I think you accidentally changed the indentation from 4 spaces to 2 spaces when copy pasting.

Also, there are some trailing empty lines in `wptreport.py` that need to be removed.

Hexcles (approved):
LGTM

PR #15656, opened 2019-03-04
Improve `wpt manifest` update performance

gsnedders:
This:

 * moves one `if` out of the every existing file case, avoiding the condition needing to be evaluated the majority of the time,

 * avoids reading the manifest items if they're unchanged except for in the reftest case.

### DIFF
--- a/tools/manifest/manifest.py
+++ b/tools/manifest/manifest.py
@@ -281,18 +281,20 @@ class Manifest(object):
                         hash_changed = True
                         if new_type != old_type:
                             del self._data[old_type][rel_path]
+                            if old_type in reftest_types:
+                                reftest_changes = True
                     else:
-                        new_type, manifest_items = old_type, self._data[old_type][rel_path]
-                    if old_type in reftest_types and new_type != old_type:
-                        reftest_changes = True
+                        new_type = old_type
+                        if old_type in reftest_types:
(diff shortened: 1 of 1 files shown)


tools/manifest/manifest.py
```diff
@@ -281,18 +281,20 @@ def update(self, tree):
...
                             del self._data[old_type][rel_path]
+                            if old_type in reftest_types:
+                                reftest_changes = True
                     else:
-                        new_type, manifest_items = old_type, self._data[old_type][rel_path]
-                    if old_type in reftest_types and new_type != old_type:
-                        reftest_changes = True
+                        new_type = old_type
+                        if old_type in reftest_types:
+                            manifest_items = self._data[old_type][rel_path]
```
jgraham: To be clear I assume it's OK to not set `reftest_changes` here since the check below under `new_type in reftest_types` covers all the relevant changes?

jgraham (approved):

tools/manifest/manifest.py
gsnedders: Dunno. But we didn't do it before this patch either (given in this case `new_type == old_type`).

gsnedders: Oh, also this is the "no changes" case, so there aren't any `reftest_changes` here.

PR #15658, opened 2019-03-04
Make manifest items lists instead of sets

gsnedders:
This saves ~0.4s locally when running ./wpt run. Semantically, this is wrong and could potentially cause issue in tests when comparing when items we get, but it does provide a reasonable performance gain through avoiding hashing the items.

Pushing this PR up even though I fundamentally dislike it to at least show what the changes are.

The perf gain comes from two places:

 * updating the manifest at the start of the run (and we create all the reftest items then, which we won't have to do assuming https://github.com/web-platform-tests/rfcs/pull/15 passes),
 * when loading all manifest items prior to filtering tests to run (which we won't have to do if we're more intelligent about the filtering, which we need to be to get larger perf gains in the specified-test case).

Personally, I'd rather wait for the other two things to fix rather than using the wrong data structure here, despite the perf benefit.

### DIFF
--- a/tools/manifest/manifest.py
+++ b/tools/manifest/manifest.py
@@ -129,11 +129,11 @@ class TypeData(object):
     def load(self, key):
         """Load a specific Item given a path"""
         if self.json_data is not None:
-            data = set()
+            data = []
             path = from_os_path(key)
             for test in iterfilter(self.meta_filters, self.json_data.get(path, [])):
                 manifest_item = self.type_cls.from_json(self.manifest, path, test)
-                data.add(manifest_item)
+                data.append(manifest_item)
             try:
                 del self.json_data[path]
             except KeyError:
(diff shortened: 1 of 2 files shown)


Hexcles:
Is the "~0.4s" the total overhead for each `wpt run` invocation regardless of the number of tests it runs?

gsnedders:
> Is the "~0.4s" the total overhead for each `wpt run` invocation regardless of the number of tests it runs?

Currently, yes.

The latter of the two causes should reduce the overhead to just the tests run, which will make it negligible when we're running few tests. (And the former will go down to 0.)

jgraham:
Are you planning to ressurect this or should it be closed?

gsnedders:
With `meta_keys` now gone I think the hashing is quicker, so let's just kill this.

PR #15659, opened 2019-03-04
Fix #15625: deal with Safari throwing when moving the window to (0,0)

### LINKED ISSUE #15625: Unable to run reftests in stable release of Safari
[A recent change to the WebDriver reftest runner](https://github.com/web-platform-tests/wpt/pull/15222) introduced an invocation of the "set window rect" method with arguments "x" and "y" both set to the integer value `0`.

This command is rejected by the WebDriver implementation for the stable release of Safari ("safaridriver"), so all reftests are now failing for that browser. Here's an example of the error from [a recent test on wpt.fyi](https://wpt.fyi/results/infrastructure/reftest-wait.html?sha=44e2efb923):

    ERROR message: Parameter 'x' has a boolean value, which is not allowed.
      File "/Users/kazooie/worker/macOS_Chunked_Runner/build/tools/wptrunner/wptrunner/executors/base.py", line 167, in run_test result = self.do_test(test)
      File "/Users/kazooie/worker/macOS_Chunked_Runner/build/tools/wptrunner/wptrunner/executors/executorwebdriver.py", line 419, in do_test self.protocol.webdriver.window.position = (0, 0)
      File "/Users/kazooie/worker/macOS_Chunked_Runner/build/tools/webdriver/webdriver/client.py", line 20, in inner return func(self, *args, **kwargs)
      File "/Users/kazooie/worker/macOS_Chunked_Runner/build/tools/webdriver/webdriver/client.py", line 274, in position self.session.send_session_command("POST", "window/rect", body)
      File "/Users/kazooie/worker/macOS_Chunked_Runner/build/tools/webdriver/webdriver/client.py", line 508, in send_session_command return self.send_command(method, url, body)
      File "/Users/kazooie/worker/macOS_Chunked_Runner/build/tools/webdriver/webdriver/client.py", line 472, in send_command raise err InvalidArgumentException: invalid argument (400): Parameter 'x' has a boolean value, which is not allowed.

Firefox, Chrome, and Safari Technology Preview all pass that test.

My initial guess is that this is the result of erroneous type coercion in safaridriver, though it seems unlikely that we're the first to try to move a Safari window to (0, 0).

@burg do you have any ideas about this?

gsnedders:
This is fixed in STP and likely will not persist beyond Safari 12. Fixes #15625.

### DIFF
--- a/tools/wptrunner/wptrunner/executors/executorwebdriver.py
+++ b/tools/wptrunner/wptrunner/executors/executorwebdriver.py
@@ -416,7 +416,11 @@ class WebDriverRefTestExecutor(RefTestExecutor):
             """return [window.outerWidth - window.innerWidth,
                        window.outerHeight - window.innerHeight];"""
         )
-        self.protocol.webdriver.window.position = (0, 0)
+        try:
+            self.protocol.webdriver.window.position = (0, 0)
+        except client.InvalidArgumentException:
+            # Safari 12 throws with 0 or 1, treating them as bools; fixed in STP
+            self.protocol.webdriver.window.position = (2, 2)
         self.protocol.webdriver.window.size = (800 + width_offset, 600 + height_offset)
 
         result = self.implementation.run_test(test)


burg (approved):
How long will workarounds like this need to be in trunk? I would expect that we want Safari 12 test results to continue working for some time. Do we expect to run newer tests against older-than-GM Safari releases?

The change looks good. I am just worried about cruft like this accumulating without some obvious way to know when it's okay to jettison.

jgraham (approved):
Hah. wrt @burg 's concerns, it seems that at least a comment about when to remove it would be nice. I don't think we care about testing legacy Safari.

gsnedders:
> How long will workarounds like this need to be in trunk? I would expect that we want Safari 12 test results to continue working for some time. Do we expect to run newer tests against older-than-GM Safari releases?

In general we support GM and newer, nothing older. (And I don't think we support older in the Safari case now we unconditionally use W3C WebDriver.)

> it seems that at least a comment about when to remove it would be nice

Does the current comment not suffice for that?

jugglinmike:
Maybe adding `TODO` will make our intention to remove the code more obvious (and discoverable to others).

Separately, what do you thinkg about using `(2, 2)` unconditionally? I'm mostly asking in the interest of consistency, though there's also a chance that avoiding known-bad requests could be helpful (i.e. for consumers running large numbers of reftests).

gsnedders:
@jugglinmike the reason to be moving it at all is to try and ensure consistency on a given setup with regards to CSS pixel/display pixel alignment, and (0,0) is probably the most reliable place to move the window from that point of view. that said, we should probably look harder and see if it actually improved anything and otherwise drop it.

jugglinmike:
I'm pessimistic about that because the coordinates are interpreted of the OS window. The browser chrome will influence the screen coordinates of the document itself, so it doesn't seem like we can expect any particular value to promote clean pixel alignment.

gsnedders:
Yeah, it's quite likely not really helping. The hope was at the very least it provided constant behaviour.

jugglinmike:
@gsnedders [I ran a couple trials to get that data.](https://github.com/web-platform-tests/wpt/compare/gh-15626-validate-window-positioning)

- Comparing results when browsers are positioned to (0, 0) versus when position is not explicitly set
  - [Firefox](https://staging.wpt.fyi/results/?sha=3f73c29e93&sha=0e897ac&label=jugglinmike&label=firefox&diff&filter=ADC) (the large disparity in `html/` is a result of [a known memory management problem](https://github.com/web-platform-tests/results-collection/issues/616))
  - [Chrome](https://staging.wpt.fyi/results/?sha=3f73c29e93&sha=0e897ac&label=jugglinmike&label=chrome&diff&filter=ADC)
- Control for flakiness: comparing repeated collections when browsers are positioned to (0, 0)
  - [Firefox](https://staging.wpt.fyi/results/?sha=b3ebabb&sha=0e897ac&label=taskcluster&label=firefox&diff&filter=ADC)
  - [Chrome](https://staging.wpt.fyi/results/?sha=b3ebabb&sha=0e897ac&label=taskcluster&label=chrome&diff&filter=ADC)

The discrepancies between each browser seem roughly equivalent, suggesting that setting the position has no effect. What do you think?

gsnedders:
@jugglinmike Note that this originally comes from problems on Azure with Edge/Win10 (#15159); I've landed this because it doesn't make anything worse and gets us working on Safari again. I think we need to evaluate how things are on Edge/Win10 and then go from there when it comes to  deciding whether to drop this or not.

PR #15664, opened 2019-03-05
[csp] Enable string substitution behavior

jugglinmike:
Two content-security policy tests were authored to use the wptserver's
substitution feature [1], but the files were not named in a way to
enable the desired functionality.

Rename the tests to enable substitution and rewrite the references to
reflect their intended purpose.

[1] https://wptserve.readthedocs.io/en/latest/pipes.html

### DIFF
--- a/content-security-policy/generic/generic-0_1-script-src.sub.html
+++ b/content-security-policy/generic/generic-0_1-script-src.sub.html
@@ -2,7 +2,7 @@
 <html>
 <head>
     <title>default-src should cascade to script-src directive</title>
-    <meta http-equiv="Content-Security-Policy" content="default-src 'self' 'unsafe-inline'; report-uri  ../support/report.py?op=put&reportID={{$id}}">
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self' 'unsafe-inline'; report-uri  ../support/report.py?op=put&reportID={{uuid()}}">
     <script src='/resources/testharness.js'></script>
     <script src='/resources/testharnessreport.js'></script>
     <script src='../support/siblingPath.js'></script>
--- a/content-security-policy/generic/generic-0_10.sub.html
+++ b/content-security-policy/generic/generic-0_10.sub.html
@@ -2,7 +2,7 @@
 <html>
 <head>
     <title>test implicit port number matching (requires port 80)</title>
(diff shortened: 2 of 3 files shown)


content-security-policy/generic/generic-0_1-script-src.sub.html
```diff
@@ -2,7 +2,7 @@
 <html>
 <head>
     <title>default-src should cascade to script-src directive</title>
-    <meta http-equiv="Content-Security-Policy" content="default-src 'self' 'unsafe-inline'; report-uri  ../support/report.py?op=put&reportID={{$id}}">
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self' 'unsafe-inline'; report-uri  ../support/report.py?op=put&reportID={{uuid()}}">
```
andypaicu: Actually this test doesn't seem to actually check whether the report is sent or not and instead relies on the securitypolicyviolation event. I would just remove the entire report-uri directive.

content-security-policy/generic/generic-0_10.sub.html
```diff
@@ -2,7 +2,7 @@
 <html>
 <head>
     <title>test implicit port number matching (requires port 80)</title>
-    <meta http-equiv="Content-Security-Policy content="script-src 'self' www.{{host}} 'unsafe-inline';">
+    <meta http-equiv="Content-Security-Policy content="script-src 'self' {{domains[www]}} 'unsafe-inline';">
```
andypaicu: Could you modify the line below that build the www url to use the correct template?

http://www." + location.hostname => http://{{domains[www]}}

andypaicu: The fact that this test passed before makes me suspect it's flawed entirely but I don't really have the cycles to investigate.

content-security-policy/generic/generic-0_1-script-src.sub.html
jugglinmike: Got it

content-security-policy/generic/generic-0_10.sub.html
jugglinmike: Sure thing

jugglinmike:
Thanks for the review, @andypaicu!

andypaicu (approved):
LGTM

jugglinmike:
Thanks for the review! I'll merge this, now

PR #15670, opened 2019-03-05
Fix background-root-018-ref.xht to use same font size as test

gsnedders:


kojiishi (approved):
lgtm, thank you for catching this!

PR #15671, opened 2019-03-05
Fix a bunch of universal failures in css/CSS2

gsnedders:
Quickly looking through what fails everywhere in css/CSS2, these are some obvious, simple fixes to test bugs. After this and #15670 we're down to 57 universal failures.


svgeesus (approved):
Simple, obvious fixes which help automated comparison

PR #15672, opened 2019-03-05
Remove Chrome's unified plan flag (now on by default)

foolip:


foolip:
@henbos can you confirm that this is no longer needed with Chrome 72?

foolip:
Added in https://github.com/web-platform-tests/wpt/pull/13795.

henbos:
This did indeed ship in M72, but it shipped through an experiment making it 100% enabled (otherwise off by default). I don't know how the bots run Chrome, but a fresh install might need to be restarted before experiments are run, otherwise wait for M73 stable? Though if you are running Beta then you're good to go already.

henbos:
Wait you're running Dev? Go ahead and remove the special flag.

foolip:
@henbos please approve or I'd have to admin merge :)

gsnedders (approved):
r+ based on @henbos's comment.

PR #15673, opened 2019-03-05
Add script tags for test_driver into screen-orientation/onchange-event.html

sttk:
This html file needs a script tag for `testdriver.js` because of using `test_driver`. (A script tag for `testdriver-vendor.js` is declared to adjust to other htmls.


marcoscaceres (approved):
Good catch! thanks!

Issue #8423, opened 2017-11-24
Acid2 fails everywhere


gsnedders:
http://w3c-test.org/acid/acid2/test.html#top v. http://w3c-test.org/acid/acid2/reference.html

The test has the nose 1px higher than the reference in all browsers. Note the reference is a PNG provided with the test, and I haven't investigated to figure out where the difference comes from.

foolip:
The anti-aliasing is different in Chrome as well.

PR #15677, opened 2019-03-05
Fix Acid2 to take the screenshot after it loads

gsnedders:


gsnedders:
(Note the reftest still fails everywhere due to #8423.)

gsnedders:
Hopefully this is enough for it to pass everywhere on wpt.fyi.

jgraham:
Isn't taking the screenshot after load the default behviour? Why is something special needed here?

gsnedders:
> Isn't taking the screenshot after load the default behviour? Why is something special needed here?

Yeah, that's true. Ergh. I have no idea what's up. :/

gsnedders:
Oh, maybe it's the fact it has to scroll the test page before the screenshot can happen?

gsnedders:
Okay, so synchronously scrolling seems to work.

gsnedders:
Okay something weird is up, three Travis PR jobs?!

Ms2ger (approved):
If it passes, who am I to complain?

PR #15679, opened 2019-03-05
Add a basic log formatter for Chromium JSON format.

LukeZielinski:
This is in preparation for integrating WPT into Chromium CI. The output format of build steps in Chromium is defined by the [JSON Test Result format](https://chromium.googlesource.com/chromium/src/+/master/docs/testing/json_test_results_format.md).

This change adds the basics of the formatter, covering the required fields. Future changes may add additional output. Also added unit tests for this formatter.

### DIFF
--- a/tools/wptrunner/wptrunner/formatters/chromium.py
+++ b/tools/wptrunner/wptrunner/formatters/chromium.py
@@ -1,0 +1,97 @@
+import json
+import time
+
+from mozlog.formatters import base
+
+
+class ChromiumFormatter(base.BaseFormatter):
+    """Formatter to produce results matching the Chromium JSON Test Results format.
+    https://chromium.googlesource.com/chromium/src/+/master/docs/testing/json_test_results_format.md
+    """
+
+    def __init__(self):
+        # Whether the run was interrupted, either by the test runner or user.
+        self.interrupted = False
+
+        # A map of test status to the number of tests that had that status.
+        self.num_failures_by_status = {}
+
+        # Start time, expressed as offset since UNIX epoch in seconds.
+        self.start_timestamp_seconds = None
+
+        # Trie of test results. Each directory in the test name is a node in
+        # the trie and the leaf contains the dict of per-test data.
+        self.tests = {}
+
+    def _store_test_result(self, name, actual, expected):
+        """
+        Stores the result of a single test in |self.tests|
+        :param str name: name of the test.
+        :param str actual: actual status of the test.
+        :param str expected: expected status of the test.
+        """
+        name_parts = filter(None, name.split("/"))
+        cur_dict = self.tests
+        for name_part in name_parts:
+            if name_part not in cur_dict:
+                cur_dict[name_part] = {}
+            cur_dict = cur_dict[name_part]
+        cur_dict["actual"] = actual
+        cur_dict["expected"] = expected
+
+    def _map_status_name(self, status):
+        """
+        Maps a WPT status to a Chromium status.
+
+        Chromium has five main statuses that we have to map to:
+        CRASH: the test harness crashed
+        FAIL: the test did not run as expected
+        PASS: the test ran as expected
+        SKIP: the test was not run
+        TIMEOUT: the did not finish in time and was aborted
+
+        :param str status: the string status of a test from WPT
+        :return: a corresponding string status for Chromium
+        """
+        if status == "OK":
+            return "PASS"
+        if status == "NOTRUN":
+            return "SKIP"
+        if status == "EXTERNAL-TIMEOUT":
+            return "TIMEOUT"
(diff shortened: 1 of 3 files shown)


tools/wptrunner/wptrunner/formatters/tests/test_chromium.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # Set up the handler.
+    output = StringIO()
+    logger = structuredlog.StructuredLogger("test_a")
+    logger.add_handler(handlers.StreamHandler(output, chromium.ChromiumFormatter()))
+
+    # output a bunch of stuff
+    logger.suite_start(["test-id-1"], run_info={}, time=123)
+    logger.test_start("test-id-1")
+    time.sleep(0.125)
```
gsnedders: Is there any need for this at all?

LukeZielinski: Oops, lazy c/p by me. Removed.

tools/wptrunner/wptrunner/formatters/chromium.py
```diff
@@ -0,0 +1,97 @@
...
+        name_parts = filter(None, name.split("/"))
+        cur_dict = self.tests
+        for name_part in name_parts:
+            if name_part not in cur_dict:
+                cur_dict[name_part] = {}
+            cur_dict = cur_dict[name_part]
+        cur_dict["actual"] = actual
+        cur_dict["expected"] = expected
+
+    def _map_status_name(self, status):
```
gsnedders: Is there any reason why this is a function and not just a dict? I don't have an overly strong preference here.

If you keep it as a function can you add a failure case that throws if the status is unknown?

tools/wptrunner/wptrunner/formatters/chromium.py
```diff
@@ -0,0 +1,97 @@
...
+        self.tests = {}
+
+    def _store_test_result(self, name, actual, expected):
+        """
+        Stores the result of a single test in |self.tests|
+        :param str name: name of the test.
+        :param str actual: actual status of the test.
+        :param str expected: expected status of the test.
+        """
+        name_parts = filter(None, name.split("/"))
```
gsnedders: Can you add a comment saying what `None` does as the first arg? It's a bit non-obvious!

Also: why is this needed? Is there any case when we're returning something with `//` in it?

tools/wptrunner/wptrunner/formatters/chromium.py
```diff
@@ -0,0 +1,97 @@
...
+        cur_dict = self.tests
+        for name_part in name_parts:
+            if name_part not in cur_dict:
+                cur_dict[name_part] = {}
+            cur_dict = cur_dict[name_part]
```
gsnedders: This whole body of the `for` is equivalent to `cur_dict = cur_dict.setdefault(name_part, {})`.

tools/wptrunner/wptrunner/formatters/chromium.py
```diff
@@ -0,0 +1,97 @@
...
+        self.start_timestamp_seconds = data["time"] if "time" in data else time.time()
+
+    def test_end(self, data):
+        actual_status = self._map_status_name(data["status"])
+        expected_status = data["expected"] if "expected" in data else "PASS"
+        self._store_test_result(data["test"], actual_status, expected_status)
+
+        # Update the count of how many tests ran with each status.
+        try:
+            self.num_failures_by_status[actual_status] += 1
```
gsnedders: Can we just use a `collections.defaultdict` here instead?

tools/wptrunner/wptrunner/formatters/chromium.py
LukeZielinski: Done - it's to deal with test names having a leading / which produces an empty str in the list from `split`

tools/wptrunner/wptrunner/formatters/chromium.py
LukeZielinski: Added a default result, which I chose to just return the status as-is. Not sure which would be better quite yet, this may change after this logger actually gets used for a bit.

tools/wptrunner/wptrunner/formatters/chromium.py
LukeZielinski: Done, thanks!

tools/wptrunner/wptrunner/formatters/chromium.py
LukeZielinski: Yep, done.

LukeZielinski (commented):
Thanks Sam, appreciate the py wisdom.

gsnedders (approved):

PR #15701, opened 2019-03-06
Add missing manifest item class.

jdm:
fc4b2a2ca209a1987198a1d35a1d66d8b13837e6 added new manifest item classes but did not update the mapping from name to class. This broke Servo's code that slurps a WPT log and updates the expected test results based on the contents.


jdm:
Closed in favour of #15704.

PR #15704, opened 2019-03-06
Add missing manifest item class

jdm:
fc4b2a2 added new manifest item classes but did not update the mapping from name to class. This broke Servo's code that slurps a WPT log and updates the expected test results based on the contents.

### DIFF
--- a/tools/manifest/manifest.py
+++ b/tools/manifest/manifest.py
@@ -4,7 +4,7 @@ from collections import defaultdict
 from six import iteritems, iterkeys, itervalues, string_types
 
 from . import vcs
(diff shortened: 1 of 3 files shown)


jgraham (approved):

gsnedders:
We shouldn't ever end up with a `reftest_base` anyway? :/

jdm:
Possibly the correct choice would be adding reftest_base to the list of ignored types along with support, manual, etc.

gsnedders:
@jdm oh there, in the update code; yeah, that would definitely be a better fix, if you want to file a PR with that.

PR #15719, opened 2019-03-07
Change each sleep to be shorter when waiting for servers to come up

gsnedders:
This makes the average total sleep locally go from 0.5s to 0.04s.


jgraham (approved):
10ms seems pretty short, but sure if this provides the best perf. How did you arrive at the number?

gsnedders:
@jgraham I think I started at 1ms, then changed it to 10ms when it became clear it basically always took tens of ms to succeed. Focusing on order of magnitude, not the exact number.

PR #15720, opened 2019-03-07
HTML: test WindowProxy self references

annevk:
For https://github.com/whatwg/html/pull/4410.

### DIFF
--- a/html/browsers/the-window-object/self-et-al.window.js
+++ b/html/browsers/the-window-object/self-et-al.window.js
@@ -21,6 +21,6 @@
       assert_equals(otherW.opener, null); // Ensure browsing context is discarded
       assert_equals(otherW[windowProxySelfReference], otherW);
       t.done();
-    }, 250);
+    }, 500);
   }, `popupWindow.${windowProxySelfReference} before, after closing, and after discarding`)
 });
--- a/html/browsers/the-window-object/self-et-al.window.js
+++ b/html/browsers/the-window-object/self-et-al.window.js
@@ -1,0 +1,26 @@
+[
+  "frames",
+  "globalThis",
+  "self",
+  "window"
+].forEach(windowProxySelfReference => {
+  test(() => {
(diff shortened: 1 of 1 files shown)


html/browsers/the-window-object/self-et-al.window.js
```diff
@@ -0,0 +1,26 @@
...
+    frame.remove();
+    assert_equals(otherW[windowProxySelfReference], otherW);
+  }, `iframeWindow.${windowProxySelfReference} before and after removal`);
+
+  async_test(t => {
+    const otherW = window.open();
+    assert_equals(otherW[windowProxySelfReference], otherW);
+    otherW.close();
+    assert_equals(otherW[windowProxySelfReference], otherW);
+    t.step_timeout(() => {
```
domenic: Use events to decrease potential flakiness?

html/browsers/the-window-object/self-et-al.window.js
```diff
@@ -0,0 +1,26 @@
...
+  "globalThis",
+  "self",
+  "window"
+].forEach(windowProxySelfReference => {
+  test(() => {
+    const frame = document.body.appendChild(document.createElement("iframe")),
+          otherW = frame.contentWindow;
+    assert_equals(otherW[windowProxySelfReference], otherW);
+    frame.remove();
+    assert_equals(otherW[windowProxySelfReference], otherW);
```
domenic: Adding assertion text to these asserts would help, e.g. right now https://wpt.fyi/results/html/browsers/the-window-object/self-et-al.window.html?diff&filter=ADC&run_id=6201193340600320&run_id=5327201419395072 is not very informative.

domenic: Would it be interesting or worthwhile to add a test after the pagehide/unload event, just in case some browser has the bright idea to null these things out from a queued task or something?

annevk:
Thanks @domenic, should be good now!

domenic (approved):

PR #15721, opened 2019-03-07
Improve perf when creating Item URLs

gsnedders:
Also fix JSON serialisation of reftests, which currently get serialised with their `url_base` (this bug goes back to the introduction of manifestv4).

### DIFF
--- a/tools/manifest/item.py
+++ b/tools/manifest/item.py
@@ -84,6 +84,14 @@ class URLManifestItem(ManifestItem):
 
     @property
     def url(self):
+        # we can outperform urljoin, because we know we just have path relative URLs
+        if self._url[0] == "/":
+            # TODO: MANIFEST6
+            # this is actually a bug in older generated manifests, _url shouldn't
+            # be an absolute path
+            return self._url
+        if self.url_base == "/":
+            return "/" + self._url
         return urljoin(self.url_base, self._url)
 
     @property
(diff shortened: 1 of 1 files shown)


tools/manifest/item.py
```diff
@@ -84,7 +84,15 @@ def id(self):
 
     @property
     def url(self):
-        return urljoin(self.url_base, self._url)
+        # we can outperform urljoin, because we know we just have path relative URLs
+        if self._url.startswith("/"):
```
jgraham: Is `self.url[0]` not going to be faster?

tools/manifest/item.py
```diff
@@ -84,7 +84,15 @@ def id(self):
...
+            # TODO: MANIFEST6
+            # this is actually a bug in older generated manifests
+            return self._url
```
jgraham: Also is this actually correct if the `url_base` isn't `/`?

gsnedders: Yes, given resolving `/foo/bar` against `/_mozilla` is `/foo/bar`. Really the problem is the old manifests are broken…

tools/manifest/item.py
gsnedders: oh, probably, hopefully we never have empty strings!

tools/manifest/item.py
```diff
@@ -84,7 +84,15 @@ def id(self):
...
+        elif self.url_base == "/":
+            return "/" + self._url
+        else:
+            return urljoin(self.url_base, self._url)
```
Ms2ger: Nit: no else-after-return, please.

PR #15722, opened 2019-03-07
Add test to check rendering of children of mspace

rwlbuis:
Test taken from WebKit and made into a reftest.


fred-wang (approved):
This LGTM

I wonder if not rendering the children is what we want per discussion on https://github.com/mathml-refresh/mathml/issues/15 ; how do we hand empty elements like `<br>` in HTML?

fred-wang:
cc @bfgeek

PR #15723, opened 2019-03-07
Avoid needless allocs in item with empty data structures

gsnedders:
`or` is concise but a perf nightmare.


jgraham (approved):
I am surprised that `a or b` constructs b here. Maybe I shouldn't be.

gsnedders:
@jgraham the problem is when we end up with `(extras={}) or {}` it constructs the RHS object, because the LHS is false, and in both places this is the common case.

PR #15724, opened 2019-03-07
Move more Travis jobs to TaskCluster

jgraham:
tc-jobs: all


tools/ci/start.sh
```diff
@@ -4,6 +4,8 @@
 # It will report memory usage every minute and prefer to kill browsers.
 sudo earlyoom -p -r 60 --prefer '(chrome|firefox)' --avoid 'python' &
 
+git branch --all
```
gsnedders: Why's this here?

tools/ci/run_tc.py
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env python
```
gsnedders: Can you add some docstring documenting what this does at a high level?

tools/ci/run_tc.py
```diff
@@ -0,0 +1,122 @@
...
+    p.add_argument("script", help="Script to run for the job")
+    return p
+
+
+def get_extra_jobs(event):
+    body = None
+    jobs = set()
+    if "commits" in event:
+        body = event["commits"][0]["message"]
+    elif "pull_request" in event:
```
gsnedders: Can we not make this an if?

Can we make it support both commit message and PR body? I think it ignoring commit message is surprising in the PR case?

tools/ci/run_tc.py
```diff
@@ -0,0 +1,122 @@
...
+            for item in items.split(","):
+                jobs.add(item.strip())
+            break
+    return jobs
+
+
+def set_variables(event):
+    # Set some variables that we use to get the commits on the current branch
+    ref_prefix = "refs/heads/"
+    pull_request = "false"
```
gsnedders: `"false"` seems potentially more confusing than something else here: an empty string is falsy in most languages, unlike `"false"`, or just not setting the env variable?

tools/ci/run_tc.py
jgraham: It typically works OK anyway because the default PR body is the commit message when there's only one push. In any case just making this an if doesn't work because the PR event doesn't contain commits so we'd have to actually check with git, which I would at least like to avoid at first.

tools/ci/run_tc.py
jgraham: This is purely because that's what travis did.

tools/ci/start.sh
jgraham: Leftover debugging.

tools/ci/run_tc.py
gsnedders: Ah, OK. :/

tools/ci/run_tc.py
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env python
+
+"""Wrapper script for running jobs in TaskCluster
+
+This is intended for running test jobs in TaskCluster. The script takes
+a two arguments which are the name of the test job and the script to actually
+run.
+
+The name of the test job is used to determine whether the script should be run
+for this push (this is in lieu of having a proper decision task. There are
```
gsnedders:
```suggestion
for this push (this is in lieu of having a proper decision task). There are
```

tools/ci/run_tc.py
```diff
@@ -0,0 +1,150 @@
...
+    jobs_str = run([os.path.join(root, "wpt"),
+                    "test-jobs"], return_stdout=True)
+    print(jobs_str)
+    return job in set(jobs_str.splitlines())
+
+
+def main():
+    args = get_parser().parse_args()
+    try:
+        event = json.loads(os.environ["TASK_EVENT"])
```
jugglinmike: In what cases will this environment variable not be available?

tools/ci/run_tc.py
jugglinmike: Avoiding confusion seems more valuable than mimicking TravisCI

tools/ci/tests/test_run_tc.py
```diff
@@ -0,0 +1,30 @@
...
+    ("Some initial line\n\ntc-jobs:foo, bar", set(["foo", "bar"])),
+    ("tc-jobs:foo, bar   \nbaz", set(["foo", "bar"])),
+    ("tc-jobs:all", set(["all"])),
+    ("", set())])
+@pytest.mark.parametrize("event", [
+    {"commits": [{"message": "<message>"}]},
+    {"pull_request": {"body": "<message>"}}
+])
+def test_extra_jobs_pr(msg, expected, event):
+    def sub(obj):
```
jugglinmike: This took me a while to suss out. Would you mind just duplicating the first parameter set instead?

tools/ci/tests/test_run_tc.py
```diff
@@ -0,0 +1,30 @@
...
+
+from six import iteritems
+
+from tools.ci import run_tc
+
+
+@pytest.mark.parametrize("msg,expected", [
+    ("Some initial line\n\ntc-jobs:foo,bar", set(["foo", "bar"])),
+    ("Some initial line\n\ntc-jobs:foo, bar", set(["foo", "bar"])),
+    ("tc-jobs:foo, bar   \nbaz", set(["foo", "bar"])),
```
jugglinmike: How about a test for multiple occurrences of `tc-jobs:`? (Currently, the script ignores those)

tools/ci/run_tc.py
```diff
@@ -0,0 +1,150 @@
...
+A job declaration is a line appearing in the pull request body (for
+pull requests) or first commit message (for pushes) of the form:
+
+tc-jobs: job1,job2,[...]
+
+In addition to scheduling the event, the script sets two environment variables;
+GITHUB_BRANCH which is the branch that the commits will merge into (if it's a PR)
+or the branch that the commits are on (if it's a push), and GITHUB_PULL_REQUEST
+which is the string "false" if the event triggering this job wasn't a pull request
+or the pull request number if it was. The semantics of these varaibles are chosen
```
jugglinmike: "varaibles" -> "variables"

tools/wpt/testfiles.py
```diff
@@ -34,14 +34,14 @@ def display_branch_point():
...
+        os.environ.get("GITHUB_BRANCH") == "master"):
         # For builds on the master branch just return the HEAD commit
         return git("rev-parse", "HEAD")
-    elif os.environ.get("TRAVIS_PULL_REQUEST", "false") != "false":
-        # This is a PR, so the base branch is in TRAVIS_BRANCH
-        travis_branch = os.environ.get("TRAVIS_BRANCH")
-        assert travis_branch, "TRAVIS_BRANCH environment variable is defined"
+    elif os.environ.get("GITHUB_PULL_REQUEST", "false") != "false":
+        # This is a PR, so the base branch is in GITHUB_BRANCH
+        travis_branch = os.environ.get("GITHUB_BRANCH")
```
jugglinmike: The variable name should be updated as well

tools/ci/run_tc.py
```diff
@@ -0,0 +1,150 @@
...
+    args = get_parser().parse_args()
+    try:
+        event = json.loads(os.environ["TASK_EVENT"])
+    except ValueError:
+        event = {}
+
+    if event:
+        set_variables(event)
+
+    if os.environ.get("GITHUB_BRANCH"):
```
jugglinmike: `GITHUB_BRANCH` is both input and output, and I find that a bit confusing. Documenting it as such might help, but so would moving the git invocation as suggested below (since then we wouldn't have to use it here). Is the "TODO" something you are planning to do prior to merging this patch?

tools/ci/run_tc.py
jgraham: If you are running locally.

tools/ci/tests/test_run_tc.py
jgraham: That seems needlessly repetative and hard to maintain. I'll add a comment instead.

tools/ci/run_tc.py
jgraham: It's not really an input and an output? We set it and then use it later in the script and also inherit it into the subprocess. I could duplicate it in two different places or something? A better patch might make `wpt testfiles` not depend on environment varaibles, but instead use command line arguments. But that seems like followup work.

tools/ci/tests/test_run_tc.py
jugglinmike: I find recursive functions harder to maintain than lists of values, but to each his own.

tools/ci/run_tc.py
jugglinmike: Agreed--moving from an environment variable to a command-line argument would be better still!

tools/ci/run_tc.py
```diff
@@ -0,0 +1,257 @@
...
+3. The string "all" is included in the job declaration
+4. The job name is set to "all"
+
+A job declaration is a line appearing in the pull request body (for
+pull requests) or first commit message (for pushes) of the form:
+
+tc-jobs: job1,job2,[...]
+
+In addition, there are a number of keyword arguments used to set options for the
+environment in which the jobs run. Documentation for these in in the command help.
```
jugglinmike: Typo: duplicate "in"

tools/ci/run_tc.py
jugglinmike: The script's name and documentation only describe usage in Taskcluster. That will make it hard for others to maintain support. Could you note somewhere that this script needs to work outside of Taskcluster?

tools/ci/run_tc.py
gsnedders: @jugglinmike can you file a bug to change `wpt testfiles`?

gsnedders (approved):
I have one or two concerns related to how familiar people are with TC's UI, and it being non-obvious in places, versus familiarity with Travis, but in almost every other way this seems like a great move!

jgraham:
I think the TC UI problems will be largely fixed when we are able to turn on the checks api. As far as I know that's still blocked on wpt.fyi, but @Hexcles may know more. Also they are redesigning the UI and I opened an issue about making the logs more prominent.

@foolip Any concerns about merging this?

jgraham:
I intend to merge this tomorrow in the absence of feedback to the contrary.

foolip:
That's OK with me!

foolip:
@jgraham is there an open issue or PR about enabling the checks integration for Taskcluster again? https://github.com/web-platform-tests/wpt.fyi/issues/1182 is resolved now. @Hexcles, has that been deployed to prod as well? Can we confirm it working in prod before flipping the switch again?

jgraham:
https://github.com/web-platform-tests/wpt/pull/15946 to reenable the checks API.

Hexcles:
Affirmative. And according to https://github.com/web-platform-tests/wpt.fyi/pull/1183#issuecomment-468309025 , it's working.

tools/ci/run_tc.py
```diff
@@ -0,0 +1,261 @@
...
+    p.add_argument("script_args",
+                   nargs=argparse.REMAINDER,
+                   help="Additional arguments to pass to the script")
+    return p
+
+
+def start_userspace_oom_killer():
+    # Start userspace OOM killer: https://github.com/rfjakob/earlyoom
+    # It will report memory usage every minute and prefer to kill browsers.
+    start(["sudo", "earlyoom", "-p", "-r", "60" "--prefer=(chrome|firefox)", "--avoid=python"])
```
foolip: There's a missing comma here, sent https://github.com/web-platform-tests/wpt/pull/18630 to fix that.

PR #15727, opened 2019-03-07
Add a test for iframes with an id in the window named property getter.

Ms2ger:


Ms2ger:
This catches a bug in WebKit.

html/browsers/the-window-object/named-access-on-the-window-object/window-named-properties.html
```diff
@@ -25,6 +26,13 @@
   assert_equals(window["bar"],
                 document.getElementsByTagName("iframe")[0].contentWindow);
 }, "Static name");
+
+test(function() {
+  assert_true("quux" in window, "quux not in window");
+  assert_equals(window["quux"],
+                document.getElementsByTagName("iframe")[4]);
```
annevk: Doesn't the spec require to return the nested browsing context's WindowProxy object, if there is one?

Ms2ger: There is one, but it doesn't have a name per <https://html.spec.whatwg.org/multipage/browsers.html#creating-a-new-nested-browsing-context>. (Note that the iframe has an `id` attribute.)

annevk (approved):

Issue #15728, opened 2019-03-07
infrastructure Firefox tests failing on Travis


gsnedders:
#15721 ([Travis job](https://travis-ci.com/web-platform-tests/wpt/jobs/183097178)), #15723 ([Travis job](https://travis-ci.com/web-platform-tests/wpt/jobs/183119872)), and #15726 ([Travis job](https://travis-ci.com/web-platform-tests/wpt/jobs/183122570)) have all failed on the infrastructure tests on Travis CI in Firefox, all getting one unexpected error:

```
TEST-UNEXPECTED-CRASH | /infrastructure/assumptions/cookie.html | expected OK
```

gsnedders:
Awkwardly, I can't manage to reproduce this locally.

This seems to have started earlier today though…

Hexcles:
Also #15730 ([job](https://travis-ci.com/web-platform-tests/wpt/jobs/183227311#L790))

alcooper91:
I'm hitting this with #15735 ([Job](https://travis-ci.com/web-platform-tests/wpt/jobs/183509540)) as well.

gsnedders:
filed https://bugzilla.mozilla.org/show_bug.cgi?id=1534765 for this, no idea how to debug this :/

gsnedders:
https://bugzilla.mozilla.org/show_bug.cgi?id=1533539 makes it sound like it's a Firefox without Pulseaudio issue. So installing Pulseaudio should fix this?

PR #15742, opened 2019-03-08
Move Travis CI to Xenial (16.04 LTS)

gsnedders:


gsnedders:
Largely trying this to see if it fixes #15728, given nobody can repro it locally and it doesn't happen on TC (running 18.04 LTS).

gsnedders:
see https://github.com/travis-ci/docs-travis-ci-com/issues/2233 for the Xvfb changes.

gsnedders:
still hitting #15728 sadly, so this doesn't fix that :/

jgraham (changes requested):
I think we should drop this and move off travis instead.

PR #15744, opened 2019-03-08
Add test to check rendering of children of mspace

rwlbuis:
Test adopted from WebKit.

### DIFF
--- a/mathml/presentation-markup/spaces/mspace-children-ref.html
+++ b/mathml/presentation-markup/spaces/mspace-children-ref.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>space (reference)</title>
+</head>
+<body>
+  <p>Test passes if you see a green square and no text.</p>
+  <math><mspace width="200px" height="200px" style="background: green"></mspace></math>
+</body>
+</html>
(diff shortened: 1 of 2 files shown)


fred-wang (approved):
LGTML This is how all web engines handle children in mspace. For the record, my original review was https://github.com/web-platform-tests/wpt/pull/15722#pullrequestreview-211885618

rwlbuis:
@foolip @Ms2ger any idea what could be wrong with Taskcluster/Firefox check? I think the test is valid MathML.

foolip:
@rwlbuis at the end of the logs you can find this:

## All results ##

### /mathml/presentation-markup/spaces/mspace-children.html ###
| Subtest |         Results          | Messages |
|---------|--------------------------|----------|
|         | **FAIL: 1/5, PASS: 4/5** |          |

## Unstable results ##

|                            Test                           | Subtest |         Results          | Messages |
|-----------------------------------------------------------|---------|--------------------------|----------|
| `/mathml/presentation-markup/spaces/mspace-children.html` |         | **FAIL: 1/5, PASS: 4/5** |          |

So it looks like the test is flaky in Firefox. Is that a problem with the test, or a problem with Firefox?

fred-wang:
> So it looks like the test is flaky in Firefox. Is that a problem with the test, or a problem with Firefox?

Probably a firefox bug. I believe it should either always fail (if Firefox accepts children in mspace) or always pass (if it does not accept children).

foolip:
@jgraham do you need to do something to prepare for an incoming flaky test? Can you admin merge this?

rwlbuis:
This landed in pr 15947.

PR #15755, opened 2019-03-11
Set default log formatter when `--verify` is set

### LINKED ISSUE #15013: Document how to use `./wpt run --verify` locally
I'm finding that when running `./wpt run --verify --channel dev --binary `which google-chrome-unstable` chrome /web-locks/signal.tentative.https.any.serviceworker.html` locally (for https://github.com/web-platform-tests/wpt/pull/15012) all I get is "::: Test verification FAIL".

I couldn't guess if this meant the test is flaky or that something went wrong internally.

Looking at taskcluster-run.py I found that passing `--log-mach -` might fix the problem, and it does cause the extra stuff you see in Taskcluster to be printed.

Without that extra output `./wpt run --verify` is not very useful. Might also consider changing the defaults.

qiuzhong:
When no log formatter is set, `./wpt run --verify` is not useful as
no output is printed. This patch tries to set the default log formatter
to `log_mach` with the value of sys.stdout.
Fixed: #15013

### DIFF
--- a/tools/wptrunner/wptrunner/wptlogging.py
+++ b/tools/wptrunner/wptrunner/wptlogging.py
@@ -8,6 +8,7 @@ from mozlog import commandline, stdadapter, set_default_logger
 from mozlog.structuredlog import StructuredLogger
 
 def setup(args, defaults):
+    set_default_log_formatter(args)
     logger = args.pop('log', None)
     if logger:
         set_default_logger(logger)
@@ -26,6 +27,22 @@ def setup(args, defaults):
 def setup_stdlib_logger():
     logging.root.handlers = []
     logging.root = stdadapter.std_logging_adapter(logging.root)
+
+
+def set_default_log_formatter(kwargs):
+    """When --verify is set and no default log formatter was set,
+    set a default log formatter to mach and its value is sys.stdout.
+    This is useful for ./wpt run --verify."""
+    log_formatters = ("raw", "unittest", "xunit", "html", "mach", "tbpl",
+                      "grouped", "errorsummary")
+    is_formatter_set = False
+    if kwargs.get("verify"):
+        for log_formatter in log_formatters:
+            if kwargs.get("log_" + log_formatter):
+                is_formatter_set = True
+                return
+        if not is_formatter_set:
+            kwargs["log_mach"] = "-"
 
 
 class LogLevelRewriter(object):
--- a/tools/wptrunner/wptrunner/wptlogging.py
+++ b/tools/wptrunner/wptrunner/wptlogging.py
@@ -33,13 +33,10 @@ def set_default_log_formatter(kwargs):
     """When --verify is set and no default log formatter was set,
     set a default log formatter to mach and its value is sys.stdout.
     This is useful for ./wpt run --verify."""
-    log_formatters = ("raw", "unittest", "xunit", "html", "mach", "tbpl",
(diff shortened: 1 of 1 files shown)


tools/wptrunner/wptrunner/wptlogging.py
```diff
@@ -28,6 +29,22 @@ def setup_stdlib_logger():
     logging.root = stdadapter.std_logging_adapter(logging.root)
 
 
+def set_default_log_formatter(kwargs):
+    """When --verify is set and no default log formatter was set,
+    set a default log formatter to mach and its value is sys.stdout.
+    This is useful for ./wpt run --verify."""
+    log_formatters = ("raw", "unittest", "xunit", "html", "mach", "tbpl",
+                      "grouped", "errorsummary")
```
jugglinmike: [Also note `--log-wptreport` and `--log-wptscreenshot`.](https://github.com/web-platform-tests/wpt/blob/26c03768165078f56415db91c78ff64bcf9cc142/tools/wptrunner/wptrunner/wptcommandline.py#L329-L330) Maintained here, a list of formatters here is bound to go stale; do you think you could avoid it by inspecting the `commandline` module of the `mozlog` package? It's a singleton, so I'm not sure where would be the most appropriate place to import it.

tools/wptrunner/wptrunner/wptlogging.py
```diff
@@ -28,6 +29,22 @@ def setup_stdlib_logger():
...
+    is_formatter_set = False
+    if kwargs.get("verify"):
+        for log_formatter in log_formatters:
+            if kwargs.get("log_" + log_formatter):
+                is_formatter_set = True
+                return
```
jugglinmike: The early `return` makes the flag unnecessary

jugglinmike (commented):
This seems more like a feature of the stability checker than of the logger. Relocating it in the stability checking code (which [already contains similar command-line argument manipulation](https://github.com/web-platform-tests/wpt/blob/26c03768165078f56415db91c78ff64bcf9cc142/tools/wptrunner/wptrunner/wptrunner.py#L313-L329)) will avoid adding complexity to this more widely-used codepath.

qiuzhong:
@jugglinmike , the original issue is that there is no extra output for`./wpt run --verify` command. I guess it's related to the logger configuration. `check_stability` function actually runs the test for several times and checks the consistency. Meanwhile, it's executed after the logger has been set up. I thought I should put it before `wptlogging.setup()`.

tools/wptrunner/wptrunner/wptlogging.py
qiuzhong: I didn't notice the `--log-wptreport` and `--log-wptscreenshot`. The `log_formatters` is from the `commandline` module in `mozlog` and this is the initial value. Even so, I realize it's not a good idea to hard-code here as it might be updated at other places. I'll avoid by importing `commandline` from `mozlog` package and using `commandline.log_formatters`.

tools/wptrunner/wptrunner/wptlogging.py
qiuzhong: I'll remove the unnecessary line.

jugglinmike:
@jgraham you're much more familiar with the organization of wptrunner than I. Does this look like the right place to implement the desired functionality?

PR #15766, opened 2019-03-11
Don't call str.replace if it's a no-op

gsnedders:


Ms2ger (approved):

gsnedders:
Ignoring failure due to #15728.

PR #15769, opened 2019-03-11
HTML: named objects of navigated documents

annevk:
For https://github.com/whatwg/html/pull/4404.

### DIFF
--- a/html/browsers/the-window-object/named-access-on-the-window-object/navigated-named-objects.window.js
+++ b/html/browsers/the-window-object/named-access-on-the-window-object/navigated-named-objects.window.js
@@ -1,49 +1,67 @@
-[
-  {
-    "namedObject": "<div id=abc></div>",
-    "namedObjectLocalName": "div"
-  },
-  {
-    "namedObject": "<object name=abc></object>",
-    "namedObjectLocalName": "object"
-  },
-  {
-    "namedObject": "<iframe id=abc></iframe>",
-    "namedObjectLocalName": "iframe"
-  }
-].forEach(testData => {
-async_test(t => {
-  const frame = document.createElement("iframe");
-  t.add_cleanup(() => frame.remove());
-  frame.srcdoc = `<script>function f() { return abc }</script>${testData.namedObject}`;
-  let f, associatedAbc;
-  frame.onload = t.step_func(() => {
-    f = frame.contentWindow.f;
-    associatedAbc = f();
-    frame.onload = t.step_func_done(() => {
-      assert_equals(f(), associatedAbc);
-      assert_equals(associatedAbc.localName, testData.namedObjectLocalName);
-    });
-    frame.srcdoc = "<span id=abc></span>";
-  });
-  document.body.append(frame);
-  }, `Window's associated Document object is used for finding named objects (<${testData.namedObjectLocalName}>)`);
-});
+// META: script=/common/get-host-info.sub.js
 
-async_test(t => {
-  const frame = document.createElement("iframe");
-  t.add_cleanup(() => frame.remove());
-  frame.srcdoc = "<script>function f() { return abc }</script><object name=abc data='about:blank'></object>";
-  let f, associatedAbc, associatedAbcContainer;
-  frame.onload = t.step_func(() => {
-    f = frame.contentWindow.f;
-    associatedAbc = f();
-    associatedAbcContainer = associatedAbc.frameElement;
-    frame.onload = t.step_func_done(() => {
-      assert_equals(f(), associatedAbcContainer);
-      assert_equals(associatedAbcContainer.contentWindow, null);
-    });
-    frame.srcdoc = "<span id=abc></span>";
+function echoURL(content) {
+  return `/common/echo.py?content=${encodeURIComponent(content)}`;
+}
+
+function setSrc(frame, type, content) {
+  if (type === "same-origin") {
+    frame.src = echoURL(content);
+  } else if (type === "cross-site") {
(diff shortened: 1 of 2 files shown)


annevk:
Firefox failure: https://bugzilla.mozilla.org/show_bug.cgi?id=1534325.

annevk:
Safari: https://bugs.webkit.org/show_bug.cgi?id=195562.

Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=940509.

annevk:
@domenic would be good to land these now the specification change has landed.

html/browsers/the-window-object/named-access-on-the-window-object/navigated-named-objects.window.js
```diff
@@ -0,0 +1,49 @@
...
+  {
+    "namedObject": "<iframe id=abc></iframe>",
+    "namedObjectLocalName": "iframe"
+  }
+].forEach(testData => {
+async_test(t => {
+  const frame = document.createElement("iframe");
+  t.add_cleanup(() => frame.remove());
+  frame.srcdoc = `<script>function f() { return abc }</script>${testData.namedObject}`;
+  let f, associatedAbc;
```
domenic: Why declare these here instead of inside the onload?

html/browsers/the-window-object/named-access-on-the-window-object/navigated-named-objects.window.js
```diff
@@ -0,0 +1,49 @@
...
+  frame.onload = t.step_func(() => {
+    f = frame.contentWindow.f;
+    associatedAbc = f();
+    frame.onload = t.step_func_done(() => {
+      assert_equals(f(), associatedAbc);
+      assert_equals(associatedAbc.localName, testData.namedObjectLocalName);
+    });
+    frame.srcdoc = "<span id=abc></span>";
```
domenic: Using srcdoc seems kind of like asking for extra complexity. Maybe duplicate this test with a normal `.src` setting?

domenic (commented):
Given how cross-origin goes down a completely different codepath, both in spec and implementations, it would be really good to test those cases too.

html/browsers/the-window-object/named-access-on-the-window-object/navigated-named-objects.window.js
annevk: I'm not sure what you mean?

html/browsers/the-window-object/named-access-on-the-window-object/navigated-named-objects.window.js
```diff
@@ -0,0 +1,67 @@
...
+      "namedObject": "<iframe id=abc></iframe>",
+      "namedObjectLocalName": "iframe"
+    }
+  ].forEach(testData => {
+    async_test(t => {
+      const frame = document.createElement("iframe");
+      t.add_cleanup(() => frame.remove());
+      setSrc(frame, initialType, `<script>function f() { return abc }</script>${testData.namedObject}`);
+      frame.onload = t.step_func(() => {
+        const f = frame.contentWindow.f,
```
domenic: It is very surprising that Chrome/Safari fail with "f is not a function". f should always be a function... Maybe this is one of those cases where we get multiple onloads, and you need to skip the about:blank?

domenic (commented):
Code looks really good, but I am worried that due to the many Safari/Chrome failures, we are not getting appropriate test coverage. I suspect a load event problem.

html/browsers/the-window-object/named-access-on-the-window-object/navigated-named-objects.window.js
annevk: I thought about that too and alerted window.location.href at this spot and it didn't give me about:blank.

annevk: Hah, I think I figured it out, I'm hitting some kind of non-standard XSS prevention thing.

annevk:
So with that it's really only Chrome that's behind the times. The only failure Safari has is not supporting `contentWindow` on `<object>`. And Firefox has a weird timing issue with discarding the nested browsing context.

html/browsers/the-window-object/named-access-on-the-window-object/navigated-named-objects.window.js
```diff
@@ -0,0 +1,67 @@
...
+            associatedAbc = f(),
+            associatedAbcContainer = associatedAbc.frameElement;
+      frame.onload = t.step_func_done(() => {
+        assert_equals(f(), associatedAbcContainer);
+        assert_equals(associatedAbcContainer.contentWindow, null);
+      });
+      setSrc(frame, type, "<span id=abc></span>");
+    });
+    document.body.append(frame);
+  }, `Window's associated Document object is used for finding named objects (<object> with browsing ccontext via ${type} <iframe)>`);
```
domenic:
```suggestion
  }, `Window's associated Document object is used for finding named objects (<object> with browsing context via ${type} <iframe)>`);
```

domenic (approved):
After much local tinkering, I've determined the Chrome problem is due to something subtle and weird, probably around bare name resolution. It is not a simple onload problem or similar.

So, let's merge this. But I am less hopeful about getting any underlying issues fixed or identified in Chrome here, since it seems likely this will be blocked on some other big interop issue around global scope references. I'll try to follow along and chime in on the bug you filed.

domenic:
Of course, I now see that you already discovered this, when you merged https://bugs.chromium.org/p/chromium/issues/detail?id=940509 into https://bugs.chromium.org/p/chromium/issues/detail?id=942075 . In which case, I'm just sorry for the delay.

PR #15771, opened 2019-03-11
Send screenshots from Taskcluster to wpt.fyi

Hexcles:
Part of web-platform-tests/wpt.fyi#57

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -84,6 +84,7 @@ tasks:
                     --
                     --channel=${browser.channel}
                     --log-wptreport=../artifacts/wpt_report.json
+                    --log-wptscreenshot=../artifacts/wpt_screenshot.txt
                     --no-fail-on-unexpected
                     --test-type=${chunk[0]}
                     --this-chunk=${chunk[1]}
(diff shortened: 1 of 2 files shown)


.taskcluster.yml
```diff
@@ -84,6 +84,7 @@ tasks:
                     --
                     --channel=${browser.channel}
                     --log-wptreport=../artifacts/wpt_report.json
+                    --log-wptscreenshot=../artifacts/wpt_screenshot.db
```
jgraham: I don't really like the `.db` extension, but I guess you've hardcoded this elsewhere now so meh.

jgraham (approved):

.taskcluster.yml
Hexcles: @jgraham not really. Any suggestions?

gsnedders: From IRC:

15:03 < jgraham> Hexcles: Honestly if it's a made-up format I'd be pretty happy with no extension
15:07 < Hexcles> jgraham: hmm what about txt?
15:09 < jgraham> Hexcles: txt is fine […]

Hexcles:
Force-merged because GitHub APIs (and therefore, CIs) were totally busted.

PR #15781, opened 2019-03-12
Use GitHub actions for building manifest

jgraham:


jgraham:
@foolip

tools/ci/action_manifest_build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
gsnedders: We probably want `set -e` (and maybe `-x` too) as we use elsewhere.

tools/ci/action_manifest_build.sh
```diff
@@ -0,0 +1,7 @@
...
+mkdir -p ~/meta
+
+WPT_MANIFEST_FILE=~/meta/MANIFEST.json
+
+./wpt manifest -p $WPT_MANIFEST_FILE
+gzip -f $WPT_MANIFEST_FILE
```
gsnedders: I wonder if we should use `--best` (i.e., `-9)` here, given we don't generate it that often (and we have the cost of the Docker image and everything anyway, so it's a negligible time wise). Saves about 3%?

There again, we should probably support bzip2 here—that saves 21%. (It also exists in the stdlib so it's not a problem on the Python side to decode, unlike other things that might save more.) But that requires changes to elsewhere in our codebase, so that's probably not something for this PR.

.github/main.workflow
```diff
@@ -0,0 +1,16 @@
+workflow "Build & Release Manifest" {
+  on = "push"
```
gsnedders: Do we want to limit this just to master? Is there any benefit of build the manifest on anything else?

tools/ci/action_manifest_build.sh
jgraham: I think bz2 is not clearly better because the decompression speed is so poor; it would need a benchmark to convince me. I would be in favour of publishing zstandard or similar alongside the gzip files and making that an optional dependency for the Python code. But not in this PR.

.github/main.workflow
jgraham: I don't think the actions definitions have any way to limit by branch. We could limit to master in the action itself, but it's not totally straightforward. Happy to look at that in a followup.

foolip:
For reference, https://github.com/foolip/wpt/commits/github-actions was my experiment with the same.

tools/ci/action_manifest_build.sh
gsnedders: Unconvinced bz2 is so bad: on my desktop, you need to be downloading the file at over 2MB/s (so 16Mbps) for it to slower.

gsnedders: Also you said you were going to change it to using `-9` here, which you haven't.

.github/main.workflow
gsnedders: Yeah, seemingly not. :(

tools/ci/action_manifest_build.sh
jgraham: I have now; I pushed to the wrong repo. I think we should investigate bz2 and other options in a seperate issue.

gsnedders (approved):

PR #15788, opened 2019-03-12
HTML: window.document and discarded browsing contexts

annevk:


domenic (approved):
LGTM but as always it'd also be interesting to test window.close()-based discarding

annevk:
I might look into creating an abstraction later I suppose.

PR #15793, opened 2019-03-12
Fix expectation in attributes-and-toJSON-method-manual.https.html

danyao:
This test currently fails when the PaymentAddress returned by the user agent contains a `region` field. Since `region` is part of the PaymentAddress interface [1], I think we should add `region` to the expected output to fix the test.

[1] https://w3c.github.io/payment-request/#dom-paymentaddress


marcoscaceres (approved):

marcoscaceres:
Good catch.

PR #15798, opened 2019-03-13
Correcting use of pytest.approx for approximate equality checking

jimevans:
For the pointer action tests, this now tests that the pointer
coordinates are within +/- 1 pixel. If this tolerance is too high,
it can be adjusted in the PR.

### DIFF
--- a/webdriver/tests/perform_actions/pointer.py
+++ b/webdriver/tests/perform_actions/pointer.py
@@ -85,8 +85,8 @@ def test_click_element_center(session, test_actions_page, mouse_chain):
     assert ["mousemove", "mousedown", "mouseup", "click"] == event_types
     for e in events:
         if e["type"] != "mousemove":
-            assert pytest.approx(e["pageX"], center["x"])
-            assert pytest.approx(e["pageY"], center["y"])
+            assert e["pageX"] == pytest.approx(center["x"], abs = 1.0)
+            assert e["pageY"] == pytest.approx(center["y"], abs = 1.0)
             assert e["target"] == "outer"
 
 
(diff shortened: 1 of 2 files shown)


whimboo:
@jimevans why exactly do we need this? Which situations fail for you?

jimevans:
It’s more the point that the tests as written will never fail, under any circumstances, unless I am misunderstanding what `pytest.approx` does. If I’m reading the docs right, `approx(x, y)` doesn’t compare x to y within some tolerance. Instead, it returns s value to be compared to (something like “x +/- y”). Unless that value is zero, assert will always see it as `True`, meaning any test using approx the way we are here will always pass, as if the test simply said `assert True`. This strikes me as counterproductive, and probably not the original intent.

whimboo:
I would like other to speak up too, but I think that eg if you have a box of 100x100 and you want to move the pointer to (30/30) it should exactly be at this position. In the worst case there is a 1x1 pixel area you have to interact with. IMHO we should think about an approximation when there is a need for and real tests are failing, whereby this might also indicate a real problem with the driver implementation. See bugs like https://bugzilla.mozilla.org/show_bug.cgi?id=1499360.

jimevans:
Sure, people should feel free to speak up. The larger point is that these tests need to change, because using `pytest.approx` as they do currently masks real issues in implementations. I’ve already uncovered one such issue in IE.

As an aside, my motivation for looking into this was that I wanted to add a new test that I’m 100% sure fails in the current chromedriver case (where “current” == “pulled directly from CI”), but the test as written passes when it shouldn’t because of the improper use of the approx() method.

burg:
I see tests fail with 1px difference in window size for safaridriver. Adding tolerance explicitly would be a good improvement.

Ms2ger (approved):

PR #15803, opened 2019-03-13
[wake-lock] Adapted the rest tests to latest spec

Honry:
- Remove wakelock-api.https.html as I believe they should be
  covered by IDL tests
- Remove wakelock-cancel-twice.https.html, wakelockrequest-is-independent.https.html,
  wakelock-promise.https.html as those APIs had been changed, I will add
  tests for the new ones in a follow-up.


Honry:
@kenchris, @marcoscaceres, would you PTAL? thanks!

marcoscaceres (approved):
The usage of AbortSignal throughout looks good.

Honry:
Thanks @marcoscaceres, merge it as is.

PR #15809, opened 2019-03-13
HTML: initial about:blank navigation

annevk:
This fails in Chrome and Safari, against the HTML Standard.

See also https://github.com/whatwg/html/issues/3267. This bit me today.

### DIFF
--- a/html/browsers/browsing-the-web/navigating-across-documents/about-blank-iframe.window.js
+++ b/html/browsers/browsing-the-web/navigating-across-documents/about-blank-iframe.window.js
@@ -1,9 +1,11 @@
 async_test(t => {
   const frame = document.body.appendChild(document.createElement("iframe")),
-        frameW = frame.contentWindow;
+        frameW = frame.contentWindow,
+        path = "/common/blank.html";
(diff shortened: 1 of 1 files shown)


domenic (approved):
Looks correct according to the current spec, although if we could simplify this by removing the 1-window-2-document case, that'd be ideal. I'll discuss on the issue tracker.

annevk:
I guess I'll leave this open for now as it's not yet clear who should align with what. I also wonder if there's a difference when `src` is already set upon insertion.

And another thing that needs to be sorted out is when the `load` event is dispatched as that also differs across browsers for some cases.

domenic:
I think @zetafunction ended up landing a counterpart to this in https://github.com/web-platform-tests/wpt/blob/b9feb6c7a65f986cc30aa1d4a0933e3b11053c1a/html/browsers/the-window-object/window-reuse-in-nested-browsing-contexts.tentative.html#L17-L34 so I'll close this. Please reopen if I misunderstood.

annevk:
Thanks, that does look equivalent if not identical to me.

PR #15810, opened 2019-03-13
Screen Orientation: Add orientation-utils.js

Johanna-hub:
Add orientation-utils.js to resources


screen-orientation/resources/orientation-utils.js
```diff
@@ -0,0 +1,17 @@
+export async function loadIframe(src = "./resources/blank.html") {
```
marcoscaceres:
```suggestion
export async function loadIframe(src = "/screen-orientation/resources/blank.html") {
```

marcoscaceres (approved):

PR #15812, opened 2019-03-13
Selection and document.open()

annevk:
Chrome fails this, interestingly enough, but Firefox and Safari both pass.

cc @bzbarsky @TimothyGu @domenic

### DIFF
--- a/selection/Document-open.html
+++ b/selection/Document-open.html
@@ -6,39 +6,23 @@
 <script>
 "use strict";
 
-// This tests the HTML spec requirement "Replace the Document's singleton
-// objects with new instances of those objects. (This includes in particular
-// the Window, Location, History, ApplicationCache, and Navigator, objects, the
-// various BarProp objects, the two Storage objects, the various HTMLCollection
-// objects, and objects defined by other specifications, like Selection and the
-// document's UndoManager. It also includes all the Web IDL prototypes in the
-// JavaScript binding, including the Document object's prototype.)" in the
-// document.open() algorithm.
+const iframe = document.createElement("iframe");
+async_test(t => {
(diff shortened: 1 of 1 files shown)


bzbarsky:
Which test assertion does Chrome fail?

annevk:
That the object must be the same. Presumably also the final assertion as it passes the current test.

domenic (approved):
LGTM. Curious result. A Chrome bug would be great :).

annevk:
https://bugs.chromium.org/p/chromium/issues/detail?id=941913

yosinch:
As of today, Blink's Selection comes from `document.getSelection()`. `window.getSelection()` calls `document.getSelection()`. This is reason why Blink have new selection object after `document.open()`.

To fix this issue, we need to get rid `getSelection()` from `DocumentOrShadowRoot` interface.

annevk:
I don't really see why that matters here. Neither the Window nor the Document object change here. I also strongly suspect that setup is the same in all implementations as it's what https://w3c.github.io/selection-api/#dom-window-getselection requires.

TimothyGu:
Currently, Chromium has an [explicit call]( https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/dom/document.cc?l=3109&rcl=7a4a97842daad111dd16019dd37f6d6d01dc6c38) to reset the `Selection` object in `document.open()`. That call was added in chromium/chromium@bff9c922e36599cb97e47c0e8f1f190cc97e69ad for https://crbug.com/683066 in order to pass the older version of `selection/Document-open.html`, which was changed in this PR. It seems like the fix should just be to revert that commit.

PR #15815, opened 2019-03-13
HTML: BarProp objects

annevk:
For https://github.com/whatwg/html/issues/2579.


html/browsers/the-window-object/BarProp.window.js
```diff
@@ -0,0 +1,53 @@
+function assert_barProps(barPropObjects, visible) {
+  let lastBarProp = undefined;
+  for (let barProp of barPropObjects) {
+    const currentBarProp = barProp;
```
domenic: `const currentBarProp of barPropObjects`?

html/browsers/the-window-object/BarProp.window.js
```diff
@@ -0,0 +1,53 @@
...
+  assert_barProps(oldBarPropObjects, visible);
+}
+
+async_test(t => {
+  const frame = document.body.appendChild(document.createElement("iframe")),
+        frameW = frame.contentWindow,
+        barProps = ["locationbar", "menubar", "personalbar", "scrollbars", "statusbar", "toolbar"],
+        barPropObjects = barProps.map(val => frameW[val]);
+
+  assert_barProps(barPropObjects, true);
```
domenic: Honestly it's pretty surprising these count as visible for an iframe. Something for https://github.com/whatwg/html/issues/4431 I suppose.

html/browsers/the-window-object/BarProp.window.js
```diff
@@ -0,0 +1,53 @@
...
+  const openee = window.open("/common/blank.html"),
+        barProps = ["locationbar", "menubar", "personalbar", "scrollbars", "statusbar", "toolbar"],
+        barPropObjects = barProps.map(val => openee[val]);
+
+  assert_barProps(barPropObjects, true);
+  openee.onload = t.step_func(() => {
+    assert_identical_barProps(barProps, openee, barPropObjects, true);
+
+    openee.onunload = t.step_func(() => {
+      assert_barProps(barPropObjects, true);
```
domenic: Whenever you use `assert_barProps` instead of `assert_identical_barProps`, I wonder to myself: does this mean they're supposed to have changed since the last time? If so, replacing it with an assertion that they are supposed to be different would be good. If not, replacing it with `assert_identitical_barProps` would be good.

domenic (approved):

PR #15818, opened 2019-03-13
HTML: window.opener clearing

annevk:
For https://github.com/whatwg/html/issues/4429.


annevk:
Interesting that in automation Firefox sometimes passes.

PR #15822, opened 2019-03-13
IndexedDB: test a self-removing 'complete' handler.

quasicomputational:
At least one mocking implementation has a bug that this code tickles -
see dumbmatter/fakeIndexedDB#26. Since I'd already reduced it down to
a few lines of code, it might as well become a WPT.

### DIFF
--- a/IndexedDB/remove-listener-in-handler.any.js
+++ b/IndexedDB/remove-listener-in-handler.any.js
@@ -1,0 +1,49 @@
+// META: script=support-promises.js
+
+const DB_NAME = databaseName("");
+
+const STORE = "store";
+
+// In the original bug tickled by this test case, only two of the
+// queries worked and the rest stalled! Do five iterations and keep
+// count to make sure that they do all succeed. Note that we don't
+// need any assertions in this test - if all the queries complete, we
+// pass!
+//
+// See <https://github.com/dumbmatter/fakeIndexedDB/issues/26> for the
+// original bug.
+const ITERATIONS = 5;
+
+const t = async_test("removeEventListener in a handler doesn't interfere with future transactions.");
+
+t.add_cleanup(() => {
+    indexedDB.deleteDatabase(DB_NAME);
+})
+
+let count = 0;
+
+const go = () => {
+    const dbReq = indexedDB.open(DB_NAME, 1);
+    dbReq.addEventListener("upgradeneeded", t.step_func((ev) => {
+        ev.target.result.createObjectStore(STORE);
+    }));
+    dbReq.addEventListener("success", t.step_func((ev) => {
+        const db = ev.target.result
+        const tx = db.transaction(STORE, "readwrite");
+        const complete = t.step_func(() => {
+            tx.removeEventListener('complete', complete);
+        });
+        tx.addEventListener('complete', complete);
+        const store = tx.objectStore(STORE);
+        store.get(0).addEventListener("success", t.step_func(() => {
+            count += 1;
+            if (count === ITERATIONS) {
+                t.done();
+            }
+        }));
+    }));
+};
+
+for (let i = 0; i < ITERATIONS; i++) {
+    go();
+}


IndexedDB/remove-listener-in-handler.htm
```diff
@@ -0,0 +1,40 @@
+<!DOCTYPE html>
```
inexorabletash: Can you make this a `.any.js` test for more coverage and less boilerplate?

IndexedDB/remove-listener-in-handler.htm
```diff
@@ -0,0 +1,40 @@
...
+<script>
+const t = async_test();
+
+const STORE = "store";
+
+let count = 0;
+
+const go = () => {
+    const dbReq = indexedDB.open("db", 1);
+    dbReq.addEventListener("upgradeneeded", (ev) => {
```
inexorabletash: Needs `t.step_func` wrapper - this ensures that unexpected exceptions thrown during async steps property fail the test.

IndexedDB/remove-listener-in-handler.htm
```diff
@@ -0,0 +1,40 @@
...
+        ev.target.result.createObjectStore(STORE, { autoIncrement: true });
+    });
+    dbReq.addEventListener("success", (ev) => {
```
inexorabletash: Needs `t.step_func`

IndexedDB/remove-listener-in-handler.htm
```diff
@@ -0,0 +1,40 @@
...
+        const db = ev.target.result
+        const tx = db.transaction([STORE], "readwrite");
+        const complete = () => {
```
inexorabletash: Needs `t.step_func`

IndexedDB/remove-listener-in-handler.htm
```diff
@@ -0,0 +1,40 @@
...
+let count = 0;
+
+const go = () => {
+    const dbReq = indexedDB.open("db", 1);
+    dbReq.addEventListener("upgradeneeded", (ev) => {
+        ev.target.result.createObjectStore(STORE, { autoIncrement: true });
+    });
+    dbReq.addEventListener("success", (ev) => {
+        const db = ev.target.result
+        const tx = db.transaction([STORE], "readwrite");
```
inexorabletash: nit: With a single store in scope, the `[]` are not needed. Harmless, but it's extra noise.

IndexedDB/remove-listener-in-handler.htm
```diff
@@ -0,0 +1,40 @@
...
+        const complete = () => {
+            tx.removeEventListener('complete', complete);
+        };
+        tx.addEventListener('complete', complete);
+        const store = tx.objectStore(STORE);
+        store.get(0).addEventListener("success", () => {
```
inexorabletash: needs `t.step_func`

IndexedDB/remove-listener-in-handler.htm
```diff
@@ -0,0 +1,40 @@
...
+<script src="support.js"></script>
+<script>
+const t = async_test();
+
+const STORE = "store";
+
+let count = 0;
+
+const go = () => {
+    const dbReq = indexedDB.open("db", 1);
```
inexorabletash: Please include an `add_cleanup` call to delete the database, and consider using helpers from `support.js` to generate unique names and do pre/post test cleanup, so that this test doesn't start failing if run after another test (e.g. that also fails to clean up)

IndexedDB/remove-listener-in-handler.htm
```diff
@@ -0,0 +1,40 @@
...
+        store.get(0).addEventListener("success", () => {
+            count += 1;
+            if (count === 5) {
+                t.done();
+            }
+        });
+    });
+};
+
+for (let i = 0; i < 5; i++) {
```
inexorabletash: Add a comment why running this more than 2 times is needed?

IndexedDB/remove-listener-in-handler.htm
```diff
@@ -0,0 +1,40 @@
...
+        const db = ev.target.result
+        const tx = db.transaction([STORE], "readwrite");
+        const complete = () => {
+            tx.removeEventListener('complete', complete);
+        };
+        tx.addEventListener('complete', complete);
+        const store = tx.objectStore(STORE);
+        store.get(0).addEventListener("success", () => {
+            count += 1;
+            if (count === 5) {
```
inexorabletash: Make 5 a constant somewhere, instead of a magic number used twice?

inexorabletash (changes requested):
Thanks for adding the test! Turning regression tests for implementation bugs into a shared quality foundation for the platform is great.

quasicomputational:
Thanks for the review! I think I got to all of the suggestions. The one thing I didn't do was to use one of the more framework-y helpers, since this code is testing a buggy event listener implementation, but those helpers add listeners for various things themselves and I'm worried it might wind up not tickling the original bug if it's too far removed from the reproducer.

IndexedDB/remove-listener-in-handler.any.js
```diff
@@ -0,0 +1,43 @@
+// META: script=support-promises.js
+
+const DB = databaseName("");
```
inexorabletash: nit: Both `DB` and `db` are used as variable names here... maybe call this `DB_NAME` ?

IndexedDB/remove-listener-in-handler.any.js
```diff
@@ -0,0 +1,43 @@
...
+        const tx = db.transaction(STORE, "readwrite");
+        const complete = t.step_func(() => {
+            tx.removeEventListener('complete', complete);
+        });
+        tx.addEventListener('complete', complete);
+        const store = tx.objectStore(STORE);
+        store.get(0).addEventListener("success", t.step_func(() => {
+            count += 1;
+            if (count === ITERATIONS) {
+                t.done();
```
inexorabletash: Since there are no assertions, maybe add a comment e.g. "Test passes if all iterations complete, fails if it times out."

This is implied by the comment up where `ITERATIONS` is defined, but I also had to read https://github.com/dumbmatter/fakeIndexedDB/issues/26 to confirm. Links or comments in the test to help future maintainers understand what this test is verifying (i.e. that it's a regression test for a previously buggy implementation)  would be great. Otherwise, some poor maintainer will try and figure out what part of the spec this is validating.

IndexedDB/remove-listener-in-handler.any.js
```diff
@@ -0,0 +1,43 @@
...
+t.add_cleanup(() => {
+    indexedDB.deleteDatabase(DB);
+})
+
+let count = 0;
+
+const go = () => {
+    const dbReq = indexedDB.open(DB, 1);
+    dbReq.addEventListener("upgradeneeded", t.step_func((ev) => {
+        ev.target.result.createObjectStore(STORE, { autoIncrement: true });
```
inexorabletash: Is `autoIncrement` needed for the repro? If not, remove to minimize distraction. If so, comment about it?

inexorabletash (changes requested):
Few more really minor things. Thanks again!

quasicomputational:
Ping. I think I've addressed all the comments. I dunno what's up with Taskcluster being red, though.

IndexedDB/remove-listener-in-handler.any.js
quasicomputational: I think the comment at the top is good enough now - it explicitly says that passing all the iterations is good enough, and the file's short enough that it's unlikely to be missed.

quasicomputational:
Ping?

inexorabletash (approved):
Looks good; sorry about forgetting about this and thanks for pinging.

marcoscaceres:
Might need a rebase...

PR #15826, opened 2019-03-13
Make Chrome use server_host hard-coded to localhost.

LukeZielinski:
This is required to run WPT in chromium CI, where manipulating hosts is not possible.

### DIFF
--- a/tools/wpt/run.py
+++ b/tools/wpt/run.py
@@ -106,7 +106,7 @@ otherwise install OpenSSL and ensure that it's on your $PATH.""")
 
 
 def check_environ(product):
-    if product not in ("firefox", "servo"):
+    if product not in ("chrome", "firefox", "servo"):
         config_builder = serve.build_config(os.path.join(wpt_root, "config.json"))
         # Override the ports to avoid looking for free ports
         config_builder.ssl = {"type": "none"}
--- a/tools/wpt/run.py
+++ b/tools/wpt/run.py
@@ -287,6 +287,9 @@ class Chrome(BrowserSetup):
         # Shorten delay for Reporting <https://w3c.github.io/reporting/>.
         kwargs["binary_args"].append("--short-reporting-delay")
 
+        # Point all .test domains to localhost for Chrome
+        kwargs["binary_args"] += ["--host-resolver-rules=MAP nonexistent.*.test ~NOTFOUND, MAP *.test 127.0.0.1"]
+
 
 class ChromeAndroid(BrowserSetup):
     name = "chrome_android"
--- a/tools/wpt/run.py
+++ b/tools/wpt/run.py
@@ -278,18 +278,6 @@ class Chrome(BrowserSetup):
             logger.info("Automatically turning on experimental features for Chrome Dev")
             kwargs["binary_args"].append("--enable-experimental-web-platform-features")
 
-        # Allow audio autoplay without a user gesture.
-        kwargs["binary_args"].append("--autoplay-policy=no-user-gesture-required")
-
-        # Allow WebRTC tests to call getUserMedia.
-        kwargs["binary_args"] += ["--use-fake-ui-for-media-stream", "--use-fake-device-for-media-stream"]
-
-        # Shorten delay for Reporting <https://w3c.github.io/reporting/>.
-        kwargs["binary_args"].append("--short-reporting-delay")
-
-        # Point all .test domains to localhost for Chrome
-        kwargs["binary_args"] += ["--host-resolver-rules=MAP nonexistent.*.test ~NOTFOUND, MAP *.test 127.0.0.1"]
-
 
 class ChromeAndroid(BrowserSetup):
     name = "chrome_android"
--- a/tools/wptrunner/wptrunner/browsers/chrome.py
+++ b/tools/wptrunner/wptrunner/browsers/chrome.py
@@ -56,15 +56,29 @@ def executor_kwargs(test_type, server_config, cache_manager, run_info_data,
     if test_type == "testharness":
         capabilities["pageLoadStrategy"] = "none"
 
-    for (kwarg, capability) in [("binary", "binary"), ("binary_args", "args")]:
-        if kwargs[kwarg] is not None:
-            capabilities["goog:chromeOptions"][capability] = kwargs[kwarg]
+    chrome_options = capabilities["goog:chromeOptions"]
+    if kwargs["binary"] is not None:
+        chrome_options["binary"] = kwargs["binary"]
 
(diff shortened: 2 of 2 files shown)


gsnedders:
@LukeZielinski how does Chrome itself know that `web-platform.test` is `127.0.0.1` then?

LukeZielinski:
@gsnedders For that part, we pass a command-line flag to Chrome (--host-resolver-rules=MAP nonexistent.*.test ~NOTFOUND, MAP *.test 127.0.0.1) That configuration lives inside chromium.

gsnedders:
@LukeZielinski is there any reason why that configuration can't live in here? we need to make sure this keeps working even without configuration that lives within the Chromium repo.

LukeZielinski:
@gsnedders True, I added the flag to wpt. I guess the effect of this is that localhost is now the only place that wptserve is expected to be running. Any sense if that's a problem we need to address (ie: does anyone run wpt with serve on some other IP)?

jgraham:
Yes, e.g. http://w3c-test.org

LukeZielinski:
Ok, so in the interest of not breaking non-localhost usages, how about I add a flag to just make this change for chromium CI (in which case it seems like we don't need the --host-resolver-rules part in WPT since it's clear that this is only for chromium CI).

WDYT?

LukeZielinski:
@gsnedders Ping for another look? Thanks!

jgraham:
Is `--host-resolver-rules` something we can pass to released Chrome builds? If it is it seems like we should just do that, similiar to the way we treat Gecko and Servo, both of which have browser-specific host configurations that mean we aren't relying on `/etc/hosts`.

In that case I think the only necessary part of this PR might be skipping the requirement to set up the hosts on Chrome, just like we do for Firefox and Servo. I'm not entirely sure why changing the host binding stuff is necessary at all, because the setup seems like it will become very similar to that of other browsers.

LukeZielinski:
Both Servo and Gecko hard-code the server_host the same as I'm doing here. If I omit that setting, then [wptserve will fail to start and complain that /etc/hosts needs to be updated.](https://github.com/web-platform-tests/wpt/blob/master/tools/wptserve/wptserve/server.py#L672)

jgraham:
But can we use this codepath always for Chrome rather than just in CI?

LukeZielinski:
Right, so that's the question I had earlier. Maybe I misunderstood what you meant :)

> LukeZielinski commented 8 days ago
> > @gsnedders True, I added the flag to wpt. I guess the effect of this is that localhost is now the only place that wptserve is expected to be running. Any sense if that's a problem we need to address (ie: does anyone run wpt with serve on some other IP)?

> jgraham commented 8 days ago
> > Yes, e.g. http://w3c-test.org

So I agree that it'd be much cleaner if we could just hard-code server_host + pass --host-resolver-rules for Chrome always. But I'm worried about whether that would break any other users of WPT. Thoughts?

jgraham:
Oh, right I mistook your question for a more general one.

People run the wpt server with the server not on localhost. People don't run wptrunner with a desktop browser and the server not on localhost. Hardcoding this is fine (better in fact); you don't need an option.

LukeZielinski:
Ok, back to the simple version then (hard-coded server_host and added --host-resolver-rules). Thanks!

tools/wpt/run.py
```diff
@@ -287,6 +287,9 @@ def setup_kwargs(self, kwargs):
         # Shorten delay for Reporting <https://w3c.github.io/reporting/>.
         kwargs["binary_args"].append("--short-reporting-delay")
 
+        # Point all .test domains to localhost for Chrome
+        kwargs["binary_args"] += ["--host-resolver-rules=MAP nonexistent.*.test ~NOTFOUND, MAP *.test 127.0.0.1"]
```
jgraham: I think this should go in `tools/wptrunner/wptrunner/browsers/chrome.py`

jgraham (changes requested):

tools/wpt/run.py
LukeZielinski: Do you mean this one specifically, or all of the binary_args appended here?
Seems `chrome.py` doesn't do anything with binary_args right now, all of it gets updated here.

Could you point to an example?

jgraham: Apologies for using the gecko sync, but the search tools are better there ;) 

https://searchfox.org/mozilla-central/source/testing/web-platform/tests/tools/wptrunner/wptrunner/browsers/chrome.py#59-67

is where the binary kwargs are actually passed down to ChromeDriver. For `--headless` we add an argument there. In this case I'd expect that we just set the args unconditionally i.e. replace the above code with something like:

```python
    chrome_options = capabilities["goog:chromeOptions"]

    chrome_options["args"] = ["--host-resolver-rules=MAP nonexistent.*.test ~NOTFOUND, MAP *.test 127.0.0.1"]
    if kwargs["binary_args"] is not None:
        chrome_options["args"].extend(kwargs["binary_args"])
    if kwargs["headless"] and "--headless" not in kwargs["binary_args"]:
        chrome_options["args"].append("--headless")

     if kwargs["binary"] is not None:
        chrome_options["binary"] = kwargs["binary"]
 ```

Comment (LukeZielinski) [reply tools/wpt/run.py]: Thanks for the pointer, done.

I also moved the handful of other flags that are unconditionally set so they're all in the same place.

**Review by jgraham — APPROVED** [review]
Thanks!

File: tools/wptrunner/wptrunner/browsers/chrome.py [hunk 2]
```diff
@@ -56,15 +56,29 @@ def executor_kwargs(test_type, server_config, cache_manager, run_info_data,
...
+    chrome_options["args"] = []
+    # Allow audio autoplay without a user gesture.
+    chrome_options["args"].append("--autoplay-policy=no-user-gesture-required")
+    # Allow WebRTC tests to call getUserMedia.
+    chrome_options["args"].append("--use-fake-ui-for-media-stream")
+    chrome_options["args"].append("--use-fake-device-for-media-stream")
+    # Shorten delay for Reporting <https://w3c.github.io/reporting/>.
+    chrome_options["args"].append("--short-reporting-delay")
+    # Point all .test domains to localhost for Chrome
+    chrome_options["args"].append("--host-resolver-rules=MAP nonexistent.*.test ~NOTFOUND, MAP *.test 127.0.0.1")
```
gsnedders: @LukeZielinski @jgraham I'd rather we didn't hardcode `*.test` here, given we build up the right set of hosts dynamically for every other browser (from `self.config`), but I don't think it's a big deal.

PR #15827, opened 2019-03-13
Add reftest for CSS overflow with min content

jugglinmike:


gsnedders (changes requested):
I'd rather you didn't rely on the width computation being done in the ref; I'd rather just use Ahem and explicitly specify the width there.

I realise this doesn't hit the buggy codepath in Gecko in this case, but it's mildly surprising that `overflow: scroll` isn't similarly broken.

gsnedders:
(also why did this get `needs-reviewers`?)

zcorpan:
Wild guess: it could be bogus error handling if there was a network error when fetching.

MatsPalmgren:
I disagree with the general statement in this testcase.  See discussion in bug [1535195](https://bugzilla.mozilla.org/show_bug.cgi?id=1535195#c4).

jugglinmike:
@gsnedders Thanks for the review! I'm following up with @MatsPalmgren on BugZilla to see if/how this behavior might be specified. In the mean time, I'm going to hold off on implementing your feedback.

gsnedders:
I think the conclusion from @MatsPalmgren is that this test isn't backed up by any spec?

PR #15838, opened 2019-03-14
Test linethickness with unitless zero

rwlbuis:


mathml/presentation-markup/fractions/frac-linethickness-004-ref.html
```diff
@@ -0,0 +1,28 @@
...
+        /* FractionRuleThickness = 10000 * 1 / 1000 = 10px; */
+        font-family: "TestFont";
+        font-size: 1px;
+      }
+    </style>
+  </head>
+  <body>
+    <p>This test passes if you see no fraction bar.</p>
+    <math>
+      <mspace width="0px"/>
```
fred-wang: I don't think this mspace is needed for this test

mathml/presentation-markup/fractions/frac-linethickness-004-ref.html
```diff
@@ -0,0 +1,28 @@
...
+      <mfrac>
```
fred-wang: it should be linethickness="0px" right?

fred-wang (changes requested):

mathml/presentation-markup/fractions/frac-linethickness-004-ref.html
rwlbuis: Yeah, not sure what I was thinking here, fixed.

mathml/presentation-markup/fractions/frac-linethickness-004-ref.html
rwlbuis: Right, fixed.

fred-wang (approved):
LGTM

PR #15862, opened 2019-03-15
Fix git manifest generation to actually find local changes

gsnedders:
Previously we'd never consider anything changed, ever, because we took the whole short-format line to be the `rel_path` (it isn't). This has been broken since MANIFESTv4 landed in 2016(!).


gsnedders:
I broke this by relying on https://github.com/git/git/commit/e8b2dc2c2a8415bfd180ecf5cc237a54e69ac2e9 and not checking what version of `git` it required (2.21.0). Whoops.

jgraham (approved):

jgraham (approved):

PR #15863, opened 2019-03-15
Add a test for the order of operations in the WebAssembly.Global constructor.

Ms2ger:
In particular, check that ToNumber() is called after the type checks.


titzer:
lgtm

titzer (approved):

PR #15871, opened 2019-03-15
HTML: mutating the style element

annevk:
I created this for https://github.com/whatwg/dom/pull/732#discussion_r266083085 since I didn't really trust the current abstractions we have in place. And we also haven't tested them as I traced in https://freenode.logbot.info/whatwg/20190315#c2058477. So this is an initial test before I call it a week to show there's a problem with the standard (all browsers fail this test and change the style sheet upon insertion of a `Comment` node).

cc @domenic @nox

### DIFF
--- a/html/semantics/document-metadata/the-style-element/mutations.window.js
+++ b/html/semantics/document-metadata/the-style-element/mutations.window.js
@@ -32,7 +32,7 @@ test(t => {
 test(t => {
   const style = document.body.appendChild(document.createElement("style"));
   t.add_cleanup(() => style.remove());
-  const comment =style.appendChild(new Comment());
+  const comment = style.appendChild(new Comment());
   const sheet = style.sheet;
   comment.remove();
   assert_not_equals(sheet, style.sheet);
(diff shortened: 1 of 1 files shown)


html/semantics/document-metadata/the-style-element/mutations.window.js
```diff
@@ -0,0 +1,7 @@
+test(() => {
+  const style = document.body.appendChild(document.createElement("style")),
+        styleS = style.sheet;
```
foolip: Optional nit: `sheet` as the variable name might read better

foolip (approved):

guest271314:
Do `assert_equals` and `assert_not_equals` expect an `Object`?  What is the expected result of the test, and why?

html/semantics/document-metadata/the-style-element/mutations.window.js
```diff
@@ -0,0 +1,39 @@
...
+  const sheet = style.sheet;
+  style.normalize();
+  assert_equals(style.childNodes.length, 1);
+  assert_not_equals(sheet, style.sheet);
+}, "Mutating the style element: using normalize()");
+
+test(t => {
+  const style = document.body.appendChild(document.createElement("style"));
+  t.add_cleanup(() => style.remove());
+  const comment =style.appendChild(new Comment());
```
foolip: Missing space after =

foolip (approved):

PR #15874, opened 2019-03-15
PaymentDetailsModifier.supportedMethods should be a DOMString

danyao:
Fixing a bug in test configuration. This bug is causing the user agent to fail on invalid payment method identifier, instead of the later check of recursive payment data.


marcoscaceres (approved):
Oops! Good catch.

PR #15907, opened 2019-03-18
Remove domenic as a payment-request reviewer

domenic:
It's been fun, but it seems like folks have these in hand these days!


rsolomakhin (approved):
Bye, we'll miss ya!

PR #15908, opened 2019-03-18
Add --short-reporting-delay flag for Chrome

foolip:


chlily1:
Thanks so much, @foolip!

Hexcles (approved):
LGTM, but just want to double check: is this new flag `--short-reporting-delay` available and supposed to be turned on stable as well?

Hexcles:
cc @chlily1 see question above ^

chlily1:
I believe it's made it to Dev channel so far: https://chromiumdash.appspot.com/commit/c2c755de8b324198d4c5971c7c8d4498dfb2a21b
Should this wait until it hits stable?

It is not meant to be used for normal browsing, just for running tests. It lowers a 1 minute delay to 100 ms for Reporting reports to be sent within the timeframe of a test.

Hexcles:
@chlily1 so eventually it'll be available in stable, and when it does, it should be turned on. If that's the case, then this PR can be merged as is (Chrome doesn't seem to complain about unrecognized flags, so it's fine to add the flag unconditionally even if it's not yet available in stable).

chlily1:
Great, thanks so much!

Issue #15824, opened 2019-03-13
Reorganizing documentation (2019)


jugglinmike:
The project's documentation (available online at https://web-platform-tests.org) has grown somewhat disorganized over the past few years. @zcorpan and I are planning to reorganize the content to address some of [the issues that have been filed](https://github.com/web-platform-tests/wpt/issues?q=is%3Aissue+is%3Aopen+label%3Adocs) and recognize the perspective gained from [the 2018 WPT survey](https://docs.google.com/spreadsheets/d/1HFFXXxVhGy0sT4BAzvegIb-rfGgQHaT5ZjTMO_bggQE). To give a sense for what we have in mind, [here's a diagram of the current site's organization](https://user-images.githubusercontent.com/677252/54238586-41017d80-44ef-11e9-9320-b8f766a0885f.png), and [here's a version that describes some modifications](https://user-images.githubusercontent.com/677252/54238589-43fc6e00-44ef-11e9-9a53-faac75e76898.png).

We recognize a bit of tension between the different ways we could conduct this work and submit the result. The preferred git workflow includes many small, targeted patches. However, restructuring a large body of text in a holistic way will tend towards more wide-spanning change sets. Taking a cue from [similar work @gsnedders conducted back in 2017](https://github.com/gsnedders/wpt-docs/issues?utf8=%E2%9C%93&q=), we're going to collaborate on a restructuring from a fork of WPT.

The fork is: https://github.com/bocoup/wpt-docs 

And the site can be viewed at: https://bocoup.github.io/wpt-docs/ 

Feedback about the high-level approach as well as any specific issues are of course welcome.

jugglinmike:
Here's an affinity diagram which groups the 2018 survey responses according to theme:

![affinity-annotated](https://user-images.githubusercontent.com/677252/54566770-3429d180-49a8-11e9-8a73-6cf03f1daa52.png)

Each item includes a number prefixed with the letter "U". This is a user identifier and describes the row in [the survey response spreadsheet](https://docs.google.com/spreadsheets/d/1HFFXXxVhGy0sT4BAzvegIb-rfGgQHaT5ZjTMO_bggQE) from which the item was derived.

We intentionally omitted responses to questions for new ideas in order to keep the diagram focused on patterns in users' direct experience.

/cc @jorydotcom since she helped put the survey together. Thanks, Jory!

foolip:
@jugglinmike should this issue be closed now?

jugglinmike:
[...]

PR #15916, opened 2019-03-18
[docs] Introduce "search" feature for website

jugglinmike:
I've been reviewing [WPT's open "docs" issues](https://github.com/web-platform-tests/wpt/issues?q=is%3Aissue+is%3Aopen+label%3Adocs) and [the responses to the 2018 WPT survey](https://docs.google.com/spreadsheets/d/1HFFXXxVhGy0sT4BAzvegIb-rfGgQHaT5ZjTMO_bggQE), and one of the strongest themes is difficulty locating existing content. [We're hoping to improve that by restructuring the documents](https://github.com/web-platform-tests/wpt/issues/15824), and a dedicated text search is another way to help folks find what they're looking for.

This patch implements search on the client, integrating Jekyll output with [the Lunr.js JavaScript library](https://github.com/olivernn/lunr.js). As a MIT-licensed project, I don't think there are any concerns about including it in WPT. For a small client-side library, Lunr.js is surprisingly good at natural language search (e.g. the query "equals" matches the term "equality").

Building the site locally can be a bit tricky, so here are a few screenshots:

2019-03-18-search-inactive

2019-03-18-search-active

2019-03-18-search-results

---

I also experimented with [a simpler solution which relies on DuckDuckGo](https://github.com/web-platform-tests/wpt/compare/master...bocoup:docs-search-ddg). While that approach requires less code, it has some drawbacks.

- search results include matches from pulls.web-platform-tests.org, but such matches are unlikely to be desired ([we should probably publish a `robots.txt` file there](https://github.com/web-platform-tests/pulls.web-platform-tests.org/issues/65))
- DuckDuckGo has to index new content before it will be available via search

While the visualization of results in this patch could use some work, I wanted to get some feedback before spending more time on it. @gsnedders @jgraham @foolip (+anyone else): does this look promising to you? Do you think DuckDuckGo would be better? Or is there another alternative that I should look in to?


foolip:
This looks *very* promising, I was hoping we could have a client-side search just like this! I'll review.

jugglinmike:
In gh-16458, we re-implemented the build process to use Sphinx. That includes a client-side search, so this patch is no longer necessary.

PR #15931, opened 2019-03-19
Streamline executors

jugglinmike:
These are a couple simplifications I found while experimenting with an "ExecutorCDP" (a wptrunner Executor build on Chrome DevTools Protocol).

@jdm I've removed a method from the ExecutorServoDriver class, but I'm still having trouble running wpt with Servo--even on `master`. Does this look okay to you?

### DIFF
--- a/tools/wptrunner/wptrunner/executors/protocol.py
+++ b/tools/wptrunner/wptrunner/executors/protocol.py
@@ -132,13 +132,6 @@ class BaseProtocolPart(ProtocolPart):
         """Wait indefinitely for the browser to close"""
         pass
 
-    @property
-    def current_window(self):
-        """Return a handle identifying the current top level browsing context
-
-        :returns: A protocol-specific handle"""
-        pass
-
     @abstractmethod
     def set_window(self, handle):
         """Set the top level browsing context to one specified by a given handle.
--- a/tools/wptrunner/wptrunner/executors/protocol.py
+++ b/tools/wptrunner/wptrunner/executors/protocol.py
@@ -132,14 +132,6 @@ class BaseProtocolPart(ProtocolPart):
         """Wait indefinitely for the browser to close"""
         pass
 
-    @abstractmethod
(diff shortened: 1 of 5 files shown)


jdm (approved):

tools/wptrunner/wptrunner/executors/protocol.py
```diff
@@ -132,21 +132,6 @@ def wait(self):
...
 
-    @property
-    def current_window(self):
-        """Return a handle identifying the current top level browsing context
-
-        :returns: A protocol-specific handle"""
-        pass
-
-    @abstractmethod
-    def set_window(self, handle):
```
jgraham: I don't think this shuld be removed; it looks like most executors need to implement it and those that don't can explictly opt out.

jgraham (changes requested):

tools/wptrunner/wptrunner/executors/protocol.py
jugglinmike: How about moving it to a new subclass of `BaseProtocolPart`? Usually a class name is evident when an extension like this is necessary, but the best I can think of is `MutableWindowProtocolPart`. Not great.

jugglinmike: Subclasses don't necessarily need this method, so declaring an `abstractmethod` (i.e. technically requiring subclasses to implement) seems contradictory to me. Other contributors (and me in 3 months) aren't likely to understand this. They might implement the method in some other subclass (contributing dead code). They also might invoke the method from outside the subclass, generically (relying on a method with partial availability).

Adding a new subclass in the inheritance chain would allow us to declare the interface in an object-oriented way without inaccurately representing the true requirements. It was just a guess that the OO pattern is what's important here.

A different compromise would be to keep the `abstractmethod` in the base class as per your request and document the optional nature of the required method. I could use a hand writing the code comment, though, because I don't understand why it's important to declare the method in this way.

PR #15937, opened 2019-03-20
test POST navigation with Referrer-Policy

JuniorHsu:
This is a fork of https://github.com/web-platform-tests/wpt/pull/14879 since I'm not able to fix the unnecessary rebase.

This patch is testing for POST navigation with Referrer-Policy and addressing reviewer's comment. I haven't touched HTTPS->HTTP part.


fetch/origin/no-cors.window.js
```diff
@@ -76,3 +74,99 @@ promise_test(async function () {
...
+      const frame = document.createElement("iframe");
+      frame.src = origins.HTTP_ORIGIN + referrerPolicyPath +
+                  "?referrerPolicy=" + referrerPolicy;
+      self.addEventListener("message", e => {
+        if (e.data === "loaded") {
+          resolve();
+          frame.remove();
+        }
+      }, { once: true });
+      frame.onload = () => {
```
jdm: This should try to avoid the onload event that can occur from the initial about:blank.

JuniorHsu: @jdm  I try to steal some ideas in current wpt, but almost every wpt with `frame.src` set comes a `frame.onload`.

fetch/origin/no-cors.window.js
```diff
@@ -76,3 +74,99 @@ promise_test(async function () {
   assert_equals(json[0], "null");
```
annevk: I think we concluded that this should be different as about:blank does have an origin.

fetch/origin/no-cors.window.js
```diff
@@ -76,3 +74,99 @@ promise_test(async function () {
...
 }, "Origin header and POST navigation");
+
+var NavigationDestination = {
+  SameOrigin: 1,
+  CrossOrigin: 2,
+  properties: {
+    1: {name: "Same-Origin", origin: origins.HTTP_ORIGIN},
+    2: {name: "Cross-Origin", origin: origins.HTTP_REMOTE_ORIGIN},
+  }
+}
```
annevk: I wonder if instead of this setup, you could do a nested forEach whereby the second forEach iterates over `["same-origin", "cross-origin"]`. Then you could also inline referrerPolicyFunctor and referrerPolicyTestString and the whole thing would be a little easier to grasp.

fetch/origin/no-cors.window.js
annevk: Yeah, I'm not getting two load events here so I think this is fine. @jdm or are you saying there's a race condition with that? (I think there are some scenarios where it can occur, but not racily, though I cannot recall exactly how.)

annevk (commented):
Thanks for picking this up again, looks like a great start!

Aside from forms, another thing we could do here is use `fetch(..., { method:"POST", mode: "no-cors" })`. That might be good to do as it's a little different from navigation. You could basically nest the forEach further for that, though at some point I guess we'll have to worry about timeouts...

I think that's basically it apart from CORS.

fetch/origin/no-cors.window.js
jdm: IIRC different browsers suppress or do not suppress the about:blank onload event if another navigation has occurred.

fetch/origin/no-cors.window.js
JuniorHsu: Sorry I didn't review those existing tests before. Will fix it.

fetch/origin/no-cors.window.js
JuniorHsu: let's use message-based

fetch/origin/no-cors.window.js
```diff
@@ -73,6 +71,147 @@ promise_test(async function () {
...
+  ].forEach(destination => {
+    promise_test(navigationReferrerPolicy(testObj.policy,
+                                          destination.name,
+                                          destination.expectedOrigin),
+                 referrerPolicyTestString(testObj.policy,
+                                          destination.name + " navigation"));
+
+    promise_test(fetchReferrerPolicy(testObj.policy,
+                                     destination.name,
+                                     "cors",
```
annevk: When we're dealing with CORS (mode is "cors") we shouldn't take the referrer policy into account. That only applies to "no-cors" / "navigate", and "same-origin" I suppose. It's an intentional part of using that protocol that makes it okay for us to leak the origin.

JuniorHsu: Hence, spec for "Serializing a request origin" is going to be one more between 1. and 2. :
if CORS flag is unset, request’s origin is same origin with request’s current URL’s origin, and request’s referrer policy is "no-refererer", return "null"

annevk: Yeah, something like that is probably reasonable, but I mean that the test for "cors" mode does the wrong thing here.

JuniorHsu: @annevk I update the test. Only honor Referrer-Policy when no-cors/navigation and same-origin now. Please check if it's what we want. Thanks.

annevk (commented):
It took me a while to debug the testharness failure I mentioned I was seeing. It's interesting that you did not trigger it at all.

I also suspect that one of the tests I added should have its pass condition changed. A GET navigation should not include the Origin header. Similar for GET no-cors, but I don't think you're testing that at the moment.

I think a good next step in parallel with fixing these issues would be figuring out the desired patch for Fetch. That might lead to more test scenarios and some input from the other browsers on what behavior we want to have.

Hope this helps.

fetch/origin/no-cors.window.js
```diff
@@ -73,6 +71,145 @@ promise_test(async function () {
...
+    "policy": "unsafe-url",
+    "expectedOriginForSameOrigin": origins.HTTP_ORIGIN
+  },
+].forEach(testObj => {
+  [
+    {
+      "name": "same-origin",
+      "expectedOrigin": testObj.expectedOriginForSameOrigin
+    },
+    // not honor Referrer-Policy for cross-origin navigation/fetch
```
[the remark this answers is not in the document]
JuniorHsu: By [your last comment](https://github.com/web-platform-tests/wpt/pull/15937/commits/5aa109d83fe84e385ffd88c694015614b5bdf562#r270465836). Maybe I misunderstood, but it looks like "same-origin" is necessary.

fetch/origin/no-cors.window.js
```diff
@@ -73,6 +71,145 @@ promise_test(async function () {
...
+        } else if (e.data === "action") {
+          const doc = frame.contentDocument,
+                form = doc.body.appendChild(doc.createElement("form")),
+                submit = form.appendChild(doc.createElement("input"));
+          form.action = postUrl;
+          form.method = "POST";
+          submit.type = "submit";
+          submit.click();
+
+          frame.onload = null;
```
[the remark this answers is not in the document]
JuniorHsu: This is a work around for harness. See https://github.com/web-platform-tests/wpt/pull/14879#discussion_r248277925

fetch/origin/no-cors.window.js
```diff
@@ -73,6 +71,145 @@ promise_test(async function () {
...
 
-  assert_equals(json[0], "null");
+  assert_equals(json[0], origins.HTTP_ORIGIN);
   assert_equals(json[1], "null");
 }, "Origin header and POST navigation");
+
+function navigationReferrerPolicy(referrerPolicy, destination, expectedOrigin) {
+  return async function () {
+    const stash = token(),
+          referrerPolicyPath = "/fetch/origin/resources/referrer-policy.py";
```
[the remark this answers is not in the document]
JuniorHsu: Thanks for finding this (facepalm)

fetch/origin/no-cors.window.js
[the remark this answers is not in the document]
annevk: That comment was only about mode "cors", but not all cross-origin requests use mode "cors", they can use "no-cors" or "navigate" too.

fetch/origin/no-cors.window.js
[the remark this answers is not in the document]
annevk: The harness doesn't add listeners to frames. That comment was about a frame that had a listener for the load event, this frame does not.

fetch/origin/no-cors.window.js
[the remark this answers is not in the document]
JuniorHsu: Okay, I guess I got you this time.
"no-cors"/"navigate": apply referrer-policy
same origin fetch with mode "cors": apply referrer-policy
cross origin fetch with mode "cors": not apply referrer-policy

Please check my new patch to see if it's what we want.
Thanks.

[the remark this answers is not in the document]
annevk: That's still not quite it. When mode is "cors", I don't think we should apply Referrer Policy at all. Or is that what browsers do today somehow?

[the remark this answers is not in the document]
JuniorHsu: Actually webkit send Origin: anyway for fetch API ("cors" and "no-cors") I just thought too much. As you said, let's make things simple. I push another patch for this.

annevk:
This looks correct. Some questions:

* Should I merge this into my branch?
* Should we have more redirect tests?
* Should we have more non-`fetch()` tests, e.g., ``?
* The spec change still needs to be made. Will you take that on?
* Are you blocked on anything by me?

JuniorHsu:
>     * Should I merge this into my branch?
Yes since it depends on the patch you wrote on your branch
>
>     * Should we have more redirect tests?
IMO it already fine. We already have something to fix. 
>
>     * Should we have more non-`fetch()` tests, e.g., ``?
They are usually `GET` and non-CORS. I guess we're fine without those tests.
> 
>     * The spec change still needs to be made. Will you take that on?
Yes, it's my next task.
> 
>     * Are you blocked on anything by me?
Reviews are helpful. It would be good to help on the future spec review.

annevk:
This is continued in #14260.

PR #15946, opened 2019-03-20
Use the checks API with taskcluster jobs

jgraham:


Hexcles:
According to https://github.com/web-platform-tests/wpt.fyi/pull/1183#issuecomment-468309025 , wpt.fyi should now work with Taskcluster via Checks.

Hexcles (approved):

PR #15952, opened 2019-03-20
Consolidate .gitignore files under the root.

jgraham:
Multiple .gitignores aren't supported by the manifest update code when running
against the filesystem, so put all the rules in the root file instead.


.gitignore
```diff
@@ -22,11 +24,25 @@ node_modules/
...
 .vscode/
-.DS_Store
-*.rej
+\#*
+scratch
+
+# Testsuite-specific rules
+/conformance-checkers/vnu.jar
+/cors/resources/log.txt
+/css//build-temp
```
gsnedders: Probably shouldn't have `//` here?

.gitignore
```diff
@@ -22,11 +24,25 @@ node_modules/
...
+/css/dist
+/css/dist_last
+/css/tools/cache
+/webaudio/idl/*
```
gsnedders: Probably should have trailing new line. (Also, do we actually want `/*` at the end here? That's very surprising.)

gsnedders (changes requested):
If we're going to do this, we should add a lint to avoid new `.gitignore` files being added.

.gitignore
jgraham: I literally just copied the existing rule. I'm not sure what the point of it actually is.

gsnedders:
@jgraham can you rebase this so the lint unit tests run?

jgraham:
Done

gsnedders (approved):

PR #15963, opened 2019-03-20
Test HTMLMediaElement.captureStream() with src set to MediaSource Blob URL

guest271314:


guest271314:
https://github.com/web-platform-tests/wpt/pull/15963/commits/cc69d844995a973785df73ab6157253a239ac8a4 is the code which should be incorporated into `media-source` and/or `mediacapture-fromelement` directories at wpt.

guest271314:
Note, `SourceBuffer.mode` is set to `"sequence"` due to Chromium/Chrome stalling when using `"segments"` mode. The code does not output the expected result when `"sequence"` mode is used at either Firefox or Chromium/Chrome.

guest271314:
Is the crash expected due to the decision to not expose the underlying `SourceBuffer` data https://github.com/w3c/media-source/issues/209 or allow saving of `<video>` `src` where `MediaSource` `Blob URL` https://bugs.chromium.org/p/chromium/issues/detail?id=141906 ?

Issue #5381, opened 2017-04-05
Sort out -manual filename convention vs meta element "interact" flag


annevk:
Both doesn't seem like a good long term solution. The filename convention seems a little clearer.

Inspired by #5370.

cc @zcorpan @mrego @frivoal @fantasai @jgraham

gsnedders:
The meta element flag should cause a lint failure anywhere outside of `css` (or rather, a flag that causes them to become manual *without* the filename also ending in `-manual` is a lint failure).

The reason why it's in `css` is to avoid mass-renaming files in there and breaking all existing results in the CSS test harness. Potentially, we could alter the build system such that tests with `-manual` or `-visual` got that stripped and the flag added as part of the build system, but that requires someone to do that.

frivoal:
I agree that one convention would be much better than two. That said, no strong feelings from me in favor of naming vs flags. As long as it is clearly documented, and supported by the infrastructure (including the CSS test harness), I'm happy with either.

fred-wang:
Are all the manual tests supposed to use either "interact" flag or -manual filename convention?

I had to add -manual suffix for CSS animations tests ( https://github.com/web-platform-tests/wpt/pull/11215 ) ; these tests use the "animation" flag but not the "interact" one...

cc @youennf

frivoal:
as of today, there isn't a single flag for non-automatic tests. There are several ones, depending on why it cannot be automated:
* interact (the tester needs to do something)
* animated (the test is animated in final state)
* font (Requires a specific font to be installed at the OS level)
* userstyle (requires the tester to install a particular user stylesheet)

This situation is not good, but that's where we are for now.

We need to do two things:
* decide which convention we want to use going forward (-manual suffix seems nice to me)
* decide whether to convert existing tests, or to deprecate but continue to support old conventions. (convert seems ok with me, but we need to be really sure, as renaming thousands of files isn't nice on systems that have metadata linked to filenames).

annevk:
Some thoughts:

1. It seems that outside `/css` we already decided upon `-manual`.
2. Renaming lots of files has been done quite a bit (at least outside `/css`). This mostly went fine (updating links can be a pain).
3. If you want to postpone renaming you could safelist all current usage and forbid further flag usage with the lint script.

gsnedders:
> If you want to postpone renaming you could safelist all current usage and forbid further flag usage with the lint script.

[...]

Issue #7760, opened 2017-10-13
No way to test paged media


gsnedders:
CSS has quite a lot of properties to control paged media (most obviously `page-break-before`/`-after`, etc.), and we currently have no way of testing these. We should probably have some sort of reftests that test print behaviour, probably by printing to PDF and then comparing. We may want to just reuse the Mozilla `reftest-print` class for this.

JKereliuk:
@gsnedders Do you know any manual tests for these?

gsnedders:
@JKereliuk see every test with the CSS metadata flag of "paged", so currently:

```
css/CSS2/backgrounds/background-attachment-010.xht
css/CSS2/box-display/containing-block-024.xht
css/CSS2/box-display/containing-block-025.xht
css/CSS2/cascade/at-import-008.xht
css/CSS2/margin-padding-clear/margin-collapse-160.xht
css/CSS2/page-box/at-page-rule-001.xht
css/CSS2/page-box/at-page-rule-002-a.xht
css/CSS2/page-box/at-page-rule-002-b.xht
css/CSS2/page-box/at-page-rule-002-c.xht
css/CSS2/page-box/first-page-selectors-001.xht
css/CSS2/page-box/first-page-selectors-002.xht
css/CSS2/page-box/first-page-selectors-003.xht
css/CSS2/page-box/first-page-selectors-004.xht
css/CSS2/page-box/page-box-000.xht
css/CSS2/page-box/page-container-000.xht
css/CSS2/page-box/page-container-001.xht
css/CSS2/page-box/page-container-002.xht
css/CSS2/page-box/page-container-003.xht
css/CSS2/page-box/page-container-004.xht
css/CSS2/page-box/page-container-005.xht
css/CSS2/page-box/page-container-006.xht
css/CSS2/page-box/page-container-007.xht
css/CSS2/page-box/page-container-008.xht
css/CSS2/page-box/page-container-009.xht
css/CSS2/page-box/page-container-010.xht
css/CSS2/page-box/page-container-011.xht
css/CSS2/page-box/page-grammar-001.xht
css/CSS2/page-box/page-grammar-002.xht
css/CSS2/page-box/page-margin-000.xht
css/CSS2/page-box/page-margin-001.xht
css/CSS2/page-box/page-margin-002.xht
css/CSS2/page-box/page-props-100-a.xht
css/CSS2/page-box/page-props-100-b.xht
css/CSS2/page-box/page-props-101.xht
[...]

PR #15964, opened 2019-03-20
[docs] Deprecate some requirement flags

### LINKED ISSUE #5293: Deprecate some requirement flags

jugglinmike:
This is intended to resolve gh-5293.

@frivoal @gsnedders [in a subsequent issue](https://github.com/web-platform-tests/wpt/issues/5381), Sam listed a few more flags: history, paged, and speech. Should this patch deprecate those as well?

### DIFF
--- a/docs/writing-tests/css-metadata.md
+++ b/docs/writing-tests/css-metadata.md
@@ -141,7 +141,7 @@ Example 1 (one token applies):
 Example 2 (multiple tokens apply):
 
 ``` html
-<meta name="flags" content="ahem image scroll" />
+<meta name="flags" content="asis HTMLonly may" />
 ```
 
 Example 3 (no tokens apply):
--- a/docs/writing-tests/css-metadata.md
+++ b/docs/writing-tests/css-metadata.md
@@ -52,34 +52,13 @@ href="https://www.w3.org/TR/CSS21/colors.html#background-properties" />
   <th>Description</th>
 </tr>
 <tr>
-  <td>animated</td>
-  <td>Test is animated in final state. (Cannot be verified using
-    reftests/screenshots.)</td>
-</tr>
-<tr>
   <td>asis</td>
   <td>The test has particular markup formatting requirements and
     cannot be re-serialized.</td>
 </tr>
 <tr>
-  <td>font</td>
-  <td>Requires a specific font to be installed at the OS level. (A link to the
-      font to be installed must be provided; this is not needed if only web
-      fonts are used.)</td>
-</tr>
-<tr>
-  <td>history</td>
-  <td>User agent session history is required. Testing :visited is a
-    good example where this may be used.</td>
-</tr>
-<tr>
   <td>HTMLonly</td>
   <td>Test case is only valid for HTML</td>
-</tr>
-<tr>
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):

frivoal:
* `history` probably, but it might be worth checking the tests that actually use this, to see how they do so.
* `paged` I'd rather not. See https://github.com/web-platform-tests/wpt/issues/7760#issuecomment-389216902
* `speech` I don't feel qualified to say one way or another. I don't know who speech tests work, either in theory or in practice.

gsnedders:
`paged` we can't remove till #7760 is fixed.

I think what we can deprecate in favour of `-manual` is: animated, font, history, interact (because it's literally equivalent!), speech, and userstyle. (This all those that affect the manifest except for `paged`.)

The other flags fall into four categories: those that control the old CSS build system (asis, HTMLonly, nonHTML), those that might be useful when creating implementation reports (may, should), those that might be useful to non-browser implementations (scroll), and, uh, the `invalid` flag which is useful I don't know when.

@frivoal:

> * `history` probably, but it might be worth checking the tests that actually use this, to see how they do so.

None of them do anything very complicated.

> * `speech` I don't feel qualified to say one way or another. I don't know who speech tests work, either in theory or in practice.

They don't exist in practice. No one does end-to-end testing of CSS speech in any form.

jugglinmike:
I've moved the `history` and `speech` tokens to the new "deprecated" section. Thanks, you two!

sideshowbarker (approved):

jugglinmike:
Do you folks think this is ready for `master`?

sideshowbarker:
> Do you folks think this is ready for `master`?

LGTM

jugglinmike:
Rebased over `master`. Original version available here:

https://github.com/web-platform-tests/wpt/compare/master...bocoup:docs-ahem-install-orig

PR #15969, opened 2019-03-21
[wake-lock] Wake lock requests should be independent

Honry:


Honry:
Since `WakeLock.active` had been removed from latest spec, there's no good way to prove different wake lock types are independent. So close it.

PR #15977, opened 2019-03-21
Compute the virtualenv lib_path correctly

gsnedders:


gsnedders:
@jgraham I'm concerned about the TC failure here. That looks very much like a bug introduced since we landed #15724.

jgraham:
This isn't based on master, so I think there's a bug when you take the `.taskcluster.yml` from the branch and the merge commit between this head and master.

jgraham (approved):

PR #15980, opened 2019-03-21
Always use json (and never ujson) to serialize manifest

gsnedders:
ujson.dump is scarcely quicker than json.dump, but produces different output which is annoying for those who commit the generated manifest.


Hexcles (approved):
LGTM

PR #15981, opened 2019-03-21
Fix typo in a field name inside chromium log formatter.

LukeZielinski:
It should be path_delimiter not path_delimeter.


Hexcles (approved):
LGTM (I can feel your pain...)

PR #15990, opened 2019-03-21
Wheel event wpt test

robert-snakard:
Create two wheel event tests for use with this servo issue https://github.com/servo/servo/issues/22843


jdm:
@robert-snakard Servo actually has bidirectional syncing enabled, so it will be easier to add this test to the Servo PR instead.

robert-snakard:
:+1: will do

NavidZ:
One nice thing to have is to add the wheel injection automation to WebDriver and TestDriver so we can make these tests automated instead of manual. But I know that takes a little while. So feel free to land this as manual for now.

PR #15992, opened 2019-03-22
Adding OS Release and OS Build info to the run_info for Edge

stanleyhon:
Adding detailed OS Release and OS build info to the run_info for Edge runs.

### DIFF
--- a/tools/wptrunner/wptrunner/browsers/edge.py
+++ b/tools/wptrunner/wptrunner/browsers/edge.py
@@ -1,13 +1,11 @@
 from __future__ import print_function
+import subprocess
 from .base import Browser, ExecutorBrowser, require_arg
 from ..webdriver_server import EdgeDriverServer
 from ..executors import executor_kwargs as base_executor_kwargs
 from ..executors.executorselenium import (SeleniumTestharnessExecutor,  # noqa: F401
                                           SeleniumRefTestExecutor)  # noqa: F401
 from ..executors.executoredge import EdgeDriverWdspecExecutor  # noqa: F401
-
-import subprocess
-from tools.wpt.utils import call
 
 __wptrunner__ = {"product": "edge",
                  "check_args": "check_args",
@@ -102,17 +100,16 @@ class EdgeBrowser(Browser):
     def executor_browser(self):
         return ExecutorBrowser, {"webdriver_url": self.server.url}
 
+
 def run_info_extras(**kwargs):
     osReleaseCommand = "(Get-ItemProperty 'HKLM:\Software\Microsoft\Windows NT\CurrentVersion').ReleaseId"
     osBuildCommand = "(Get-ItemProperty 'HKLM:\Software\Microsoft\Windows NT\CurrentVersion').BuildLabEx"
     try:
-        os_release = call("powershell.exe", osReleaseCommand).strip()
-        os_build = call("powershell.exe", osBuildCommand).strip()
+        os_release = subprocess.check_output(["powershell.exe", osReleaseCommand]).strip()
+        os_build = subprocess.check_output(["powershell.exe", osBuildCommand]).strip()
     except (subprocess.CalledProcessError, OSError):
         return {}
 
-    rv = {
-        "os_build": os_build,
-        "os_release": os_release}
-
+    rv = {"os_build": os_build,
+          "os_release": os_release}
     return rv
--- a/tools/wptrunner/wptrunner/browsers/edge.py
+++ b/tools/wptrunner/wptrunner/browsers/edge.py
@@ -6,6 +6,9 @@ from ..executors.executorselenium import (SeleniumTestharnessExecutor,  # noqa: 
                                           SeleniumRefTestExecutor)  # noqa: F401
 from ..executors.executoredge import EdgeDriverWdspecExecutor  # noqa: F401
 
+import subprocess
+from tools.wpt.utils import call
+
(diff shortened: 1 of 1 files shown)


stanleyhon:
@foolip @mustjab @thejohnjansen

foolip:
I'll leave this review to @gsnedders.

tools/wptrunner/wptrunner/browsers/edge.py
```diff
@@ -6,6 +6,9 @@
                                           SeleniumRefTestExecutor)  # noqa: F401
 from ..executors.executoredge import EdgeDriverWdspecExecutor  # noqa: F401
 
+import subprocess
+from tools.wpt.utils import call
```
gsnedders: Let's not rely on `tools.wpt` here; just use `subprocess.check_output` without that wrapper. (I think there's still some use of wptrunner with it actually installed, and being able to import this relies on `tools` being on the path.)

tools/wptrunner/wptrunner/browsers/edge.py
```diff
@@ -6,6 +6,9 @@
                                           SeleniumRefTestExecutor)  # noqa: F401
 from ..executors.executoredge import EdgeDriverWdspecExecutor  # noqa: F401
 
+import subprocess
```
gsnedders: Minor nit: standard Python practice would be to put this standard library import before all our local imports (so after the magic `from __future__ …` but before `from .base …`).

tools/wptrunner/wptrunner/browsers/edge.py
```diff
@@ -97,3 +101,18 @@ def cleanup(self):
...
+        os_release = call("powershell.exe", osReleaseCommand).strip()
+        os_build = call("powershell.exe", osBuildCommand).strip()
+    except (subprocess.CalledProcessError, OSError):
+        return {}
+
+    rv = {
+        "os_build": os_build,
+        "os_release": os_release}
+
+    return rv
```
gsnedders: Make sure there's a trailing new line, plz.

tools/wptrunner/wptrunner/browsers/edge.py
```diff
@@ -97,3 +101,18 @@ def cleanup(self):
...
+    osBuildCommand = "(Get-ItemProperty 'HKLM:\Software\Microsoft\Windows NT\CurrentVersion').BuildLabEx"
+    try:
+        os_release = call("powershell.exe", osReleaseCommand).strip()
+        os_build = call("powershell.exe", osBuildCommand).strip()
+    except (subprocess.CalledProcessError, OSError):
+        return {}
+
+    rv = {
+        "os_build": os_build,
+        "os_release": os_release}
```
gsnedders: With the `{` on a new line, I'd rather the `}` be on one too.

gsnedders (changes requested):
This looks mostly good; one real comment (about `tools.wpt`) and a few minor nits. Thanks!

tools/wptrunner/wptrunner/browsers/edge.py
stanleyhon: I've fixed this up to be formatted the same as the firefox file (open brace on same line, in-line key values, close brace on same line)

tools/wptrunner/wptrunner/browsers/edge.py
stanleyhon: Fixed.

tools/wptrunner/wptrunner/browsers/edge.py
stanleyhon: Moved.

tools/wptrunner/wptrunner/browsers/edge.py
stanleyhon: Using subprocess.check_output.

stanleyhon:
@gsnedders is there anything else I need to do? I've addressed your previous comments.

I noticed some automated failures, but they don't seem related to my changes.

gsnedders:
@stanleyhon I think I got side-tracked into figuring out what was causing those failures, sorry!

gsnedders (approved):

tools/wptrunner/wptrunner/browsers/edge.py
```diff
@@ -97,3 +99,17 @@ def cleanup(self):
 
     def executor_browser(self):
         return ExecutorBrowser, {"webdriver_url": self.server.url}
+
+
+def run_info_extras(**kwargs):
+    osReleaseCommand = "(Get-ItemProperty 'HKLM:\Software\Microsoft\Windows NT\CurrentVersion').ReleaseId"
```
Ms2ger:
```
  /home/test/web-platform-tests/tools/wptrunner/wptrunner/browsers/edge.py:105: DeprecationWarning: invalid escape sequence \S
```

I guess this should have used double backslashes.

gsnedders: Should probably be a raw string.

stanleyhon: Do I need to make a fresh PR to fix this up?

Ms2ger: If you could, yes, please.

stanleyhon: I've put up a CR over [here ](https://github.com/web-platform-tests/wpt/pull/16234)

Issue #15994, opened 2019-03-22
Taskcluster fails when PR description contains single quotes


Hexcles:
We do not escape `even_str` properly when storing it into `TASK_EVENT` env var:
https://github.com/web-platform-tests/wpt/blob/023f87586a2f4b64b02c8a1eaada053408d109f0/.taskcluster.yml#L80

So when the string contains single quotes, we might get Bash syntax errors. Example:
https://tools.taskcluster.net/groups/KQbrBjJ6RFWoJJkQb0dtQg/tasks/Dgs3C5lyS2eobL0By4YazA/details

PR #16021, opened 2019-03-22
Put TaskCluster environment variables into the yml file

jgraham:
This avoids escaping issues when the data contains unescaped shell
special characters, notably ' but possibly also ` " $ ! and similar.

Fixes https://github.com/web-platform-tests/wpt/issues/15994

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -71,13 +71,14 @@ tasks:
                   public/results:
                     path: /home/test/artifacts
                     type: directory
+                env:
+                    TASK_EVENT: "${event_str}"
                 command:
                   - /bin/bash
                   - --login
                   - -c
                   - set -ex;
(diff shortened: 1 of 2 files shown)


Hexcles (approved):
LGTM (I didn't know we had this lever and was playing with heredocs, which isn't great...)

Hexcles:
@jgraham looking at the logs from this PR, I think this works, but it takes some deduction to come to that conclusion (based on the fact that `GITHUB_BRANCH` is correct, which is extracted from `TASK_EVENT`). Can we make the error case more obvious by adding a warning here?

https://github.com/web-platform-tests/wpt/blob/288222ed4a2ca311c6b6e2d3d9a539aaadfc5f35/tools/ci/run_tc.py#L223

PR #16022, opened 2019-03-22
Revert #15512 due to TC infrastructure/ failure

gsnedders:
Reverts #15512. This should make TC infrastructure/ pass again.

cc @foolip @Hexcles

PR #16023, opened 2019-03-22
Reland "[ChromeDriver] Set key modifiers with mouse actions"

foolip:
Landed in https://github.com/web-platform-tests/wpt/pull/15512.

Reverted in https://github.com/web-platform-tests/wpt/pull/16022.

### DIFF
--- a/infrastructure/metadata/infrastructure/testdriver/actions/actionsWithKeyPressed.html.ini
+++ b/infrastructure/metadata/infrastructure/testdriver/actions/actionsWithKeyPressed.html.ini
@@ -1,0 +1,9 @@
+[actionsWithKeyPressed.html]
+  expected:
+    if product == "safari" or product == "firefox": ERROR
+
+
+  [TestDriver actions: actions with key pressed]
+    expected:
+      if product == "chrome": FAIL
+
--- a/infrastructure/metadata/infrastructure/testdriver/actions/actionsWithKeyPressed.html.ini
+++ b/infrastructure/metadata/infrastructure/testdriver/actions/actionsWithKeyPressed.html.ini
@@ -1,9 +1,9 @@
 [actionsWithKeyPressed.html]
   expected:
-    if product == "safari" or product == "firefox": ERROR
+    if product == "safari": ERROR
 
 
   [TestDriver actions: actions with key pressed]
     expected:
-      if product == "chrome": FAIL
+      if product == "chrome" or product == "firefox": FAIL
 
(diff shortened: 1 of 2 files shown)


foolip:
@SummerLW sorry to say there was more trouble with this test. If infrastructure/ tests fail on this PR, can you check the reason and suggest the right fix?

gsnedders:
(Note that `infrastructure` currently isn't a blocking job, because otherwise we couldn't land anything from before when TC moved to using GH Checks.)

foolip:
Aha. The export bot merged https://github.com/web-platform-tests/wpt/pull/15512 without the tests even having run. Taskcluster did fail but the infrastructure tests hadn't been moved then.

foolip:
@gsnedders looks like the failure isn't due to this change, but because of `generate_test_report`. @LukeZielinski can you get that passing on master by adding the appropriate expectations?

LukeZielinski:
@foolip So the generate_test_api test already has an "ERROR" [expectation set for firefox](https://github.com/web-platform-tests/wpt/blob/master/infrastructure/metadata/infrastructure/testdriver/generate_test_report.html.ini). And this is working as expected I think.

I see in the [log](https://taskcluster-artifacts.net/LwFnpt6vS5qvV5HkAcoriw/0/public/logs/live_backing.log) that the problem is from infrastructure/testdriver/actions/actionsWithKeyPressed.html - it's expecting an ERROR but it's getting either OK or FAIL (I actually can't tell which)?

`TEST-UNEXPECTED-FAIL | /infrastructure/testdriver/actions/actionsWithKeyPressed.html | TestDriver actions: actions with key pressed - assert_array_equals: lengths differ, expected 3 got 1
@http://web-platform.test:8000/infrastructure/testdriver/actions/actionsWithKeyPressed.html:64:34
Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:1587:25
Test.prototype.step_func_done/<@http://web-platform.test:8000/resources/testharness.js:1627:32
promise callback*@http://web-platform.test:8000/infrastructure/testdriver/actions/actionsWithKeyPressed.html:64:6
Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:1587:25
async_test@http://web-platform.test:8000/resources/testharness.js:576:22
@http://web-platform.test:8000/infrastructure/testdriver/actions/actionsWithKeyPressed.html:39:1
TEST-UNEXPECTED-OK | /infrastructure/testdriver/actions/actionsWithKeyPressed.html | expected ERROR
TEST-INFO expected ERROR | took 607ms`

gsnedders:
@foolip @LukeZielinski is right, that's getting an expected ERROR.

@LukeZielinski the harness is getting OK and not ERROR; the subtest is getting FAIL and not OK, hence the two unexpected results.

foolip:
@gsnedders I tried updating expectations manually based on what you said, let's see how that goes.

foolip:
Already relanded in https://github.com/web-platform-tests/wpt/pull/17678.

PR #16025, opened 2019-03-22
Add test for w3c/DOM-Parsing#48

bwrrp:
This adds a DOM parsing test for the case where the default namespace is reset on a prefixed element. If implemented according to the spec, serialization with the `require-well-formed` flag set should throw. The error message seems to suggest that this check was intended only for prefix declarations, and indeed something like xmlns:pre="" would have been invalid. However, setting the default namespace back to null should be allowed, and from initial testing at least Firefox and Chrome seem to agree and do not throw.

### DIFF
--- a/domparsing/XMLSerializer-serializeToString.html
+++ b/domparsing/XMLSerializer-serializeToString.html
@@ -195,6 +195,15 @@ test(function() {
 }, 'Check if "ns1" is generated even if the element already has xmlns:ns1.');
 
 test(function() {
+  const xml = '<root xmlns="ns_root"><p:child xmlns:p="ns_child" xmlns=""><grandchild/></p:child></root>';
+  const root = parse(xml);
+  // Check serialization via the XMLSerializer, which does not set the require-well-formed flag
+  assert_equals(serialize(root), xml);
+  // Also check serialization via outerHTML, which sets the require-well-formed flag but should not throw in this case
+  assert_equals(root.outerHTML, xml);
+}, 'Check if the default namespace can be correctly reset on a prefixed element');
+
+test(function() {
   const root = (new Document()).createElement('root');
   root.setAttributeNS('http://www.w3.org/1999/xlink', 'href', 'v');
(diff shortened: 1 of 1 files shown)


bwrrp:
It looks like Safari fails the new test not because it throws, but because it (unnecessarily repeats the `xmlns=""` declaration on the `grandchild` element.

annevk (approved):
File a bug against Safari?

@tkent-google want to double check this?

bwrrp:
I think I tracked Safari's behavior down to [this branch](https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/editing/MarkupAccumulator.cpp#L353) in the WebKit source. Following the code in this file, it looks like WebKit will *always* add the `xmlns=""` for any element in the null namespace. Interestingly, the comment in this branch refers to the [HTML spec](http://www.whatwg.org/specs/web-apps/current-work/multipage/the-xhtml-syntax.html#xml-fragment-serialization-algorithm) to justify this behavior. Unfortunately, this exact anchor no longer exists, but it may very well have referred to the following paragraph:

> For Elements, if any of the elements in the serialization are in no namespace, the default namespace in scope for those elements must be explicitly declared as the empty string. (This doesn't apply in the Document case.) [XML] [XMLNS]

I'm wondering what the reasoning behind this is? In XML, the default namespace for elements would [always be inherited from parent elements](https://www.w3.org/TR/xml-names/#defaulting), making extra `xmlns=""` declarations on children unnecessary if the default namespace provided by the parent is already null. However, this paragraph seems to imply that there is some reason that serializations should not count on this working as specified in the Namespaces in XML spec?

annevk:
Wow, thanks for tracking that down. I ran blame and found https://github.com/whatwg/html/commit/bbe394844941e342165f4c8873b7078ed1d41f1c which doesn't give much rationale. I'm not sure how much attention we should pay to it; if implementations are mostly different and it doesn't make sense, we can adjust the text.

Probably most of that section should end up being replaced by the more exact algorithms of https://w3c.github.io/DOM-Parsing/.

tkent-google:
This change looks good.  We should merge this when the specification change is approved.

Issue #16026, opened 2019-03-22
Taskcluster isn't running any jobs on master after switching to Checks API


Hexcles:
All jobs are empty no-op because `./wpt test-job` only produces `lint` and `manifest_upload`.

Example:
https://tools.taskcluster.net/groups/US_q1KYrQwCC4iA1taLBMw/tasks/CF899lETREeQKJdH9GZQZQ/runs/0/logs/public%2Flogs%2Flive.log#L122
(from https://github.com/web-platform-tests/wpt/runs/83097404)

Hexcles:
I can also confirm this is NOT caused by #16021 . Runs before that also had the same issue.

Hexcles:
Still broken after the revert. Perhaps #15724 was the culprit? cc @gsnedders

Hexcles:
The tricky thing is: how do verity a setup would work on master? Shall we make a fork of wpt with everything (checks, apps, etc.) wired up to verify CI changes?

jgraham:
That's basically what I did to test out the migration of manifest upload to the actions API (although that case was even harder because not everywhere has actions enabled).

jgraham:
Should be fixed by #16030

gsnedders:
https://wpt.fyi/runs?label=master&max-count=100&from=2019-03-18T00%3A00 shows plenty of TC runs again, closing.

PR #16027, opened 2019-03-22
Revert "Use the checks API with taskcluster jobs"

Hexcles:
Reverts web-platform-tests/wpt#15946

Because of https://github.com/web-platform-tests/wpt.fyi/issues/1240 and #16026

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -1,5 +1,4 @@
 version: 1
-reporting: checks-v1
 policy:
   pullRequests: public
 tasks:


Hexcles:
Oops, looks like this wasn't the real culprit. https://tools.taskcluster.net/groups/MNgkLEIzTQ-r-gRmXb6fsA/tasks/WLt6nvhvQpO8oQNfZ9uOcA/runs/0/logs/public%2Flogs%2Flive.log is still no-op.

But we do still need to revert it because of web-platform-tests/wpt.fyi#1240

Issue #15840, opened 2019-03-14
Add a test for https://github.com/w3c/webrtc-pc/pull/2113


youennf:
Add a test for https://github.com/w3c/webrtc-pc/pull/2113

PR #16041, opened 2019-03-24
Add RTCCertificate setConfiguration tests

youennf:
Verify order and equality checks for setConfiguration with certificates.
Check the case of undefined/unset certificates.
This should cover https://github.com/web-platform-tests/wpt/issues/15840

### DIFF
--- a/webrtc/RTCCertificate-postMessage.html
+++ b/webrtc/RTCCertificate-postMessage.html
@@ -73,5 +73,28 @@
         assert_throws("InvalidAccessError", () => { new RTCPeerConnection({certificates: [certificate2]}) });
         iframe.remove();
     }, "Check cross-origin created RTCCertificate");
+
+    promise_test(async t => {
+        let certificate1 = await  RTCPeerConnection.generateCertificate({ name: 'ECDSA', namedCurve: 'P-256' });
+        let certificate2 = await  RTCPeerConnection.generateCertificate({ name: 'ECDSA', namedCurve: 'P-256' });
+
+        let url = "resources/RTCCertificate-postMessage-iframe.html";
+        let iframe = await with_iframe(url);
+
+        let promise = new Promise((resolve, reject) => {
+            window.onmessage = (event) => {
+                resolve(event.data);
+            };
+            t.step_timeout(() => reject("Timed out waiting for frame to send back certificate"), 5000);
+        });
+        iframe.contentWindow.postMessage(certificate1, "*");
+        let certificate3 = await promise;
+        iframe.remove();
+
+        const pc = new RTCPeerConnection({ certificates: [certificate1, certificate2] });
+        pc.setConfiguration({ certificates: [certificate1, certificate2] });
+        assert_throws("InvalidModificationError", () => { pc.setConfiguration({ certificates: [certificate3, certificate2] }) });
+        assert_throws("InvalidModificationError", () => { pc.setConfiguration({ certificates: [certificate2, certificate1] }) });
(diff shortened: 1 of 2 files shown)


webrtc/RTCCertificate-postMessage.html
```diff
@@ -73,5 +73,28 @@
...
+        let iframe = await with_iframe(url);
+
+        let promise = new Promise((resolve, reject) => {
+            window.onmessage = (event) => {
+                resolve(event.data);
+            };
+            t.step_timeout(() => reject("Timed out waiting for frame to send back certificate"), 5000);
+        });
+        iframe.contentWindow.postMessage(certificate1, "*");
+        let certificate3 = await promise;
```
alvestrand: This seems like a heavy-handed way to create a new object.
Given that RTCCertificate is Serializable, can't it be cloned in a simpler way?
(Note to self: It seems that Serializable is now removed from WebIDL. I don't know what it's called this week.)

webrtc/RTCCertificate-postMessage.html
```diff
@@ -73,5 +73,28 @@
...
+        iframe.remove();
+
+        const pc = new RTCPeerConnection({ certificates: [certificate1, certificate2] });
+        pc.setConfiguration({ certificates: [certificate1, certificate2] });
+        assert_throws("InvalidModificationError", () => { pc.setConfiguration({ certificates: [certificate3, certificate2] }) });
+        assert_throws("InvalidModificationError", () => { pc.setConfiguration({ certificates: [certificate2, certificate1] }) });
```
alvestrand: This one throws because sequence is different, right? That's also tested in the other file, and shouldn't be retested here.

webrtc/RTCCertificate-postMessage.html
```diff
@@ -73,5 +73,28 @@
...
+            };
+            t.step_timeout(() => reject("Timed out waiting for frame to send back certificate"), 5000);
+        });
+        iframe.contentWindow.postMessage(certificate1, "*");
+        let certificate3 = await promise;
+        iframe.remove();
+
+        const pc = new RTCPeerConnection({ certificates: [certificate1, certificate2] });
+        pc.setConfiguration({ certificates: [certificate1, certificate2] });
+        assert_throws("InvalidModificationError", () => { pc.setConfiguration({ certificates: [certificate3, certificate2] }) });
```
alvestrand: This one throws because they are not object-identity equal, yes?
Isn't that what you thought was a spec violation?

webrtc/RTCCertificate-postMessage.html
```diff
@@ -73,5 +73,28 @@
...
+                resolve(event.data);
+            };
+            t.step_timeout(() => reject("Timed out waiting for frame to send back certificate"), 5000);
+        });
+        iframe.contentWindow.postMessage(certificate1, "*");
+        let certificate3 = await promise;
+        iframe.remove();
+
+        const pc = new RTCPeerConnection({ certificates: [certificate1, certificate2] });
+        pc.setConfiguration({ certificates: [certificate1, certificate2] });
```
alvestrand: This part doesn't use certificate3, so it's equivalent to a test in the other file. I think you should drop it.

alvestrand (commented):
Can be improved, but no basic objections.
If the spec says that the passed certificate should be acceptable, test that it is, even though the test will fail.

alvestrand:
ping @youennf - came across this when checking TODO lists. Abandon or update?

PR #16043, opened 2019-03-24
Add RTCDataChannel-binaryType.html from PR13499 

youennf:


youennf:
@lgrahl, made some changes to the original test. PTAL.

webrtc/RTCDataChannel-binaryType.window.js
```diff
@@ -0,0 +1,29 @@
...
+const validBinaryTypes = ['blob', 'arraybuffer'];
+const invalidBinaryTypes = ['jellyfish', 'arraybuffer ', '', null, undefined];
+
+for (const binaryType of validBinaryTypes) {
+  test((t) => {
+    const pc = new RTCPeerConnection();
+    t.add_cleanup(() => pc.close());
+    const dc = pc.createDataChannel('test-binary-type');
+
+    assert_equals(dc.binaryType, "blob");
```
lgrahl: We already check that when creating a data channel. I don't think it's needed here - it distracts from what we want to test.

webrtc/RTCDataChannel-binaryType.window.js
```diff
@@ -0,0 +1,29 @@
...
+for (const binaryType of invalidBinaryTypes) {
+  test((t) => {
+    const pc = new RTCPeerConnection();
+    t.add_cleanup(() => pc.close());
+    const dc = pc.createDataChannel('test-binary-type');
+
+    assert_throws('SyntaxError', () => {
+      dc.binaryType = binaryType;
+    });
+    assert_equals(dc.binaryType, 'blob');
```
lgrahl: Same here.

lgrahl (changes requested):
LGTM apart from a nit.

youennf:
Sure, I can remove them, they kind of make sense to me though.
The test is about setting a value.

Before setting, check the value.
After setting, check the value.
After failing to set the value, check the value.

lgrahl:
Works for me then.

webrtc/RTCDataChannel-binaryType.window.js
alvestrand: We should delete this line. Youenn is filing a spec bug on changing the default, given implementation state.

lgrahl: Err... not sure I agree to be honest but we can discuss that in https://github.com/w3c/webrtc-pc/issues/2170.

dontcallmedom (approved):

alvestrand:
@youennf if this is ready and still useful, can you hit merge?

dontcallmedom:
@foolip can you help me understand what's making the CI stuck on this PR? (if not you, who should I ping?)

stephenmcgruer:
> @foolip can you help me understand what's making the CI stuck on this PR? (if not you, who should I ping?)

You can always ping @web-platform-tests/wpt-core-team for a more general set of people that can help :)

The problem here is that the PR is so old that there have been several infra shifts on the Taskcluster side that make old PRs incompatible. That said, if reviewers are happy with it and the test seems solid, I'm happy to override the CQ and admin-merge it - please let me know.

dontcallmedom:
@stephenmcgruer then please go ahead and admin-merge - thanks!

frivoal:
Force merging PRs with lint errors isn't nice, as it blocks every other PR. Now fixed by https://github.com/web-platform-tests/wpt/pull/25944

sideshowbarker:
>  Now fixed by #25944

Thanks for catching that and fixing it

> Force merging PRs with lint errors isn't nice, as it blocks every other PR.

Nobody realized it had lint errors. Those went unnoticed among the other CI errors it was triggering due its age (see previous comments here) and was force-merged because those CI errors were non-blockers for this case. Nobody would have intentionally merged it knowing it had lint errors as well.

stephenmcgruer:
> Force merging PRs with lint errors isn't nice, as it blocks every other PR. Now fixed by #25944

Echoing Mike; thanks for catching this and taking care of it! My apologies for not considering that as a possibility, we should considering taking a(n unofficial?) policy of cherry-picking changes like this one to new PRs rather than force-merging the original PR. (Or we could see if we can fix Taskcluster to not rely on the base commit for some of its mechanisms, cc @jgraham for that).

It's a pity that even when GitHub [claim its the default](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork), most PRs from forks that I come across I cannot modify in any way and instead have to rely on the original author for minor fixups or rebases.

jgraham:
FWIW I think the best option here would have been to simply rebase the branch. If we didn't have access to the underlying branch it might have indeed been good to create a new PR.

PR #16044, opened 2019-03-24
Add RTCSctpTransport-onstatechange.html test from PR13499

youennf:


youennf:
@lgrahl, made some changes to the original test. PTAL.

youennf:
Maybe we should move that test to RTCSctpTransport-events.html

youennf:
Actually, maybe this is redundant with RTCSctpTransport-events.html

lgrahl:
Agree that it should be moved. It's mostly redundant but `RTCSctpTransport-events.html` doesn't test for `connecting`.

lgrahl (changes requested):
I don't believe this test works without delaying the connection establishment (due to holding back ICE candidates).

youennf:
> Agree that it should be moved. It's mostly redundant but `RTCSctpTransport-events.html` doesn't test for `connecting`.

Well it does test for connecting in the first test and for connected in the second test.
I am closing this PR in favor of RTCSctpTransport-events.html.

> I don't believe this test works without delaying the connection establishment (due to holding back ICE candidates).

Indeed!

PR #16045, opened 2019-03-24
Add RTCPeerConnection-close.html tests from PR13499

youennf:


youennf:
PR not to be merged yet.
@lgrahl, I would like to understand why 24 channels is needed in that test, and the mix with negotiated as well.
If it is needed, I think we should add a very simple test with 1 data channel only and checking for closed state.

lgrahl:
The mix with *negotiated* makes sense because negotiated data channels that don't have a matching end on the remote side allow us to simulate a data channel that never opened (see the comment in line 67). Furthermore, negotiated data channels tend to be handled slightly different in implementations.

I guess they could be moved into separate tests.

youennf:
OK, it makes sense to me to make a very simple test checking closure of a given pc and state of its channels.

And add other tests for negotiated channels.
Maybe simplify to just 1 negotiated channel both side instead of 12?

Also, while the data channels of the pc being closed should get their state synchronously changed, this is probably asynchronously done for the channels of the other pc.
And this is probably what should be tested: negotiated channels without counterparts should get to closed readyState after closing the counterpart pc.

lgrahl:
> Maybe simplify to just 1 negotiated channel both side instead of 12?

:ok_hand: 

> Also, while the data channels of the pc being closed should get their state synchronously changed, this is probably asynchronously done for the channels of the other pc.

Good catch.

> And this is probably what should be tested: negotiated channels without counterparts should get to closed readyState after closing the counterpart pc.

Not sure I can follow you. If you mean "All local channels should be closed immediately" and this is what we should test for, then, yes, I would agree.

zcorpan:
@youennf 

> PR not to be merged yet.

Does this still apply? If yes, please add "do not merge yet" label and explain what the next step is.

Thanks!

PR #16048, opened 2019-03-25
CORS/XHR: mark Client Hints tentative

annevk:
This ensures Client Hints related tests outside the client-hints/ directory are marked as tentative (and one is effectively removed as it duplicates other tests) as there's no Fetch integration anymore.


wisniewskit (approved):
Thanks Anne!

PR #16050, opened 2019-03-25
Rewrite lengths-1.html

rwlbuis:
Unitless nonzero values were deprecated:
https://github.com/mathml-refresh/mathml-core/commit/6e821b2c9c83de22a3f734465aaf639b5d743b76


fred-wang (approved):
I guess for consistency it should be written "10px" rather than "1em" (that's the same for the font used).

Also, can you please update the "unitless" comment, mentioning it is now invalid in MathML Core and hence ignored?

PR #16063, opened 2019-03-25
Fix running tests on Taskcluster in PRs

gsnedders:
We've not been running tests on PRs properly since 53e263d0d8297a02390d56c3d8f75ee82544780d, given it added an extra fetch and then checking  out FETCH_HEAD has been wrong.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -112,14 +112,14 @@ tasks:
               $map:
                 # This is the main place to define new stability checks
                 - name: wpt-${browser.name}-${browser.channel}-stability
-                  checkout: FETCH_HEAD
+                  checkout: HEAD
                   diff_range: HEAD^
                   description: >-
                     Verify that all tests affected by a pull request are stable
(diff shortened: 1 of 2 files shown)


jgraham (approved):

jgraham (commented):
Actually it's kind of non-obvious to me why this should work. Is `HEAD` the same as the initial `FETCH_HEAD`? Presumably it depends on the thing to extend the number of commits never being called in that case? It would be kind of nice if we just made the `start.sh` script always checkout the commit corresponding to the provided ref, and then made the Python scripts set the right commit to test where required using the `build` branch for reference.

gsnedders:
Ergh, I guess it's still wrong. It's a heckuva lot more right here than before, though.

jgraham:
https://github.com/web-platform-tests/wpt/pull/16066 is a somewhat more ambitious attempt to fix this.

jgraham:
Closing in favour of #16066

PR #16066, opened 2019-03-25
Don't use HEAD or FETCH_HEAD to checkout specific revisions.

jgraham:
FETCH_HEAD is unreliable because it's a global variable that can be
accidentially clobbered by adding an additional fetch anywhere in the
pipeline. As a result running tests in CI has been broken since we
chose the wrong revisions. HEAD is more reliable but doesn't exist
until we first check something out.

Instead, do the following:

* Fetch the initial commits into a branch called task_head and check
  this out unconditionally.

* For PRs, create branches called base_head and pr_head pointing to
  the two parents of the merge commit that we test on PRs.

* Express all the other revisions in terms of task_head, pr_head and
  base_head, since they are both correct and more descriptive than
  using complex revision specifiers.

* Move as much logic as possible out of the script baked in to the
  docker image since that's hardest to update.

### DIFF
--- a/tools/ci/run_tc.py
+++ b/tools/ci/run_tc.py
@@ -117,7 +117,7 @@ def make_hosts_file():
 
 
 def checkout_revision(rev):
-    subprocess.check_call(["git", "checkout", "-q", rev])
+    subprocess.check_call(["git", "--quiet", "checkout", rev])
 
 
 def install_chrome(channel):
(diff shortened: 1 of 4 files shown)


tools/ci/run_tc.py
```diff
@@ -233,6 +233,18 @@ def main():
...
+        parents = run(["git", "show", "--format=%P", "task_head"], return_stdout=True).strip().split()
+        if len(parents) == 2:
+            base_head = parents[0]
+            pr_head = parents[1]
+
+            run(["git", "branch", "base_head", base_head])
+            run(["git", "branch", "pr_head", pr_head])
+        else:
+            print("WARNING: Pull request HEAD wasn't a 2-parent merge commit; "
+                  "expected to test the merge of PR into the base")
```
gsnedders: Can we just exit abnormally in this case?

jgraham: Sure.

gsnedders (approved):

PR #16075, opened 2019-03-26
Add tests for creating mfrac children dynamically

rwlbuis:


mathml/presentation-markup/fractions/frac-created-dynamically-2-ref.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>mfrac created dynamically</title>
+  </head>
+  <body>
+    <p>This test passes if you see a fraction.</p>
```
fred-wang: I should be render the same as an invalid fraction with 3 children

mathml/presentation-markup/fractions/frac-created-dynamically-2-ref.html
```diff
@@ -0,0 +1,17 @@
...
+    <math>
+      <mrow>
```
fred-wang: It should be mfrac

fred-wang (approved):
LGTM, but I would just use mfrac in ref files too so that we don't need to know how invalid markup is rendered.

PR #16080, opened 2019-03-26
HTML: output mutations

annevk:
For https://github.com/whatwg/html/pull/4447.

### DIFF
--- a/html/semantics/forms/the-output-element/mutations.window.js
+++ b/html/semantics/forms/the-output-element/mutations.window.js
@@ -1,0 +1,36 @@
+function assert_equal_values(output, value) {
+  assert_equals(output.value, value);
+  assert_equals(output.defaultValue, value);
+}
+function assert_values(output, value, defaultValue) {
+  assert_equals(output.value, value);
+  assert_equals(output.defaultValue, defaultValue);
+}
+
+test(() => {
+  const output = document.createElement("output"),
+        child = output.appendChild(document.createElement("span"));
+  assert_equal_values(output, "");
+  child.textContent = "x";
+  assert_equal_values(output, "x");
+  output.value = "some";
+  assert_values(output, "some", "x");
+  child.textContent = "y";
+  assert_values(output, "y", "x");
+}, "Descendant mutations and output.value and .defaultValue");
+
+test(() => {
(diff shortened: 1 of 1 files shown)


foolip (approved):
LGTM % assert messages

annevk:
(HTML PR has pointers to impl bugs.)

html/semantics/forms/the-output-element/mutations.window.js
```diff
@@ -1,36 +1,36 @@
...
+  assert_equal_values(output, "", "start");
   child.textContent = "x";
-  assert_equal_values(output, "x");
+  assert_equal_values(output, "x", "after setting textContent");
   output.value = "some";
-  assert_values(output, "some", "x");
+  assert_values(output, "some", "x", "after setting value");
   child.textContent = "y";
-  assert_values(output, "y", "x");
+  assert_values(output, "y", "x", "after setting textContent again");
```
darinadler, line 19: This expectation doesn’t makes sense given the latest version of the HTML specification. The code above sets output.value to "some", and [this includes a step that does a string replace all](https://html.spec.whatwg.org/multipage/form-elements.html#dom-output-value) on the output element. This has the effect of replacing all the children with a single text node containing the string "some". Changing the `textContent` of the child node afterward has no effect on the output element, so the value would still be "some", not "y".

I can’t tell whether it’s the specification or the test that is incorrect.

I’m trying to implement this in WebKit. Are other browsers passing this test?

domenic: Firefox is passing (you can see it at https://wpt.fyi/results/html/semantics/forms/the-output-element/mutations.window.html?label=experimental&label=master&aligned ), and Chrome is passing after https://chromium-review.googlesource.com/c/chromium/src/+/2878397 .

However, they are passing the latest version of the test, after the fix in https://github.com/web-platform-tests/wpt/commit/f850adcd8f3464f79b5252d284860ac53c81ceb0 .

darinadler: Thanks, @domenic! That is indeed the fix I was looking for. WebKit had the older version of the test in our tree. I will update to the newer one.

PR #16098, opened 2019-03-26
[testdriver.js] Improve test coverage

jugglinmike:
Geckodriver and Chromedriver both pass this test, but the executor I am working
on did not. The existing test in the modified file ostensibly tests the same
concern, but because it only generates two events, it does so with less
consistency. I'm suggesting we keep both tests because unlike this new
keyboard-based version, the existing pointerMove test verifies that the
implementation correctly maintains state for distinct input devices.

### DIFF
--- a/infrastructure/testdriver/actions/eventOrder.html
+++ b/infrastructure/testdriver/actions/eventOrder.html
@@ -9,19 +9,20 @@
 
 <button id="a">Button a</button>
 <button id="b">Button b</button>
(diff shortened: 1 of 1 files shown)


gsnedders (approved):

gsnedders:
@jugglinmike this might need rebased to pass CI

jugglinmike:
Worth a shot, @gsnedders

PR #16107, opened 2019-03-27
Run full Edge runs every 6 hours

foolip:
This will end up in the same build as the Safari TP run. The wpt.fyi
integration should ensure that one failing doesn't affect the other,
but there is some risk here and it will need to be monitored.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -165,9 +165,10 @@ jobs:
 
 - job: results_edge
   displayName: 'all tests (Edge)'
-  # This job is only triggered manually until it has been shown to be robust.
-  condition: and(eq(variables['Build.Reason'], 'Manual'), variables['run_all_edge'])
-  # There are 5 agents in the pool, but use more jobs so that each takes <1h.
+  condition: |
+    or(eq(variables['Build.Reason'], 'Schedule'),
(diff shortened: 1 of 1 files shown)


foolip:
@mustjab @thejohnjansen I'd like to try this and see what the results are like over time. The risk is that we'll have results that make Edge look worse than it really is for a time, but I hope the EdgeDriver problem can be resolved soon. Running regularly will also make it possible to compare against the results we're getting from Sauce to see if there are other large issues to resolve before turning that setup off.

@jugglinmike FYI.

jgraham (approved):

foolip:
I'll merge this now and see if the run starting in about 30 minutes is good.

foolip:
https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=11363 is progress.

foolip:
Interesting, job 3 failed in a novel way at the very end. @mustjab can you have a look at this? I've not seen this failure mode for any other platform before.

foolip:
https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=11420 is the next scheduled build, still running.

gsnedders:
> Interesting, job 3 failed in a novel way at the very end. @mustjab can you have a look at this? I've not seen this failure mode for any other platform before.

That implies it holds some resource that the kernel won't let the process end with. Dunno how to figure that out.

Issue #16111, opened 2019-03-27
Test failures on macOS due to failing to download ChromeDriver


foolip:
In https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=11342 the "infrastructure/ tests (macOS)" and "tools/wpt/ tests (macOS)" jobs failed because of this error:
> HTTPError: 404 Client Error: Not Found for url: https://storage.googleapis.com/chromium-browser-snapshots/Mac/644709/chromedriver_mac64.zip

I haven't seen this before, and it seems likely to have been caused by https://github.com/web-platform-tests/wpt/pull/15493, so poking @JohnChen0 @Hexcles @jgraham who participated in that review.

The same jobs did pass in the PR itself:
https://github.com/web-platform-tests/wpt/runs/82508511

@JohnChen0 is the new algorithm prone to some race conditions, where the URL tried might not be available yet. Or is there missing macOS-specific logic that's needed to make this work? It would be great if you can investigate this :)

JohnChen0:
This appears to be a problem in Chromium's build process. For some reason the ChromeDriver binary is not saved at the end of Mac builds. I'll need to look into our build script to see what is wrong.

foolip:
@JohnChen0 do you think this was caused by https://github.com/web-platform-tests/wpt/pull/15493, or would we have tried to fetch this URL before as well?

Short of reverting https://github.com/web-platform-tests/wpt/pull/15493, is there any quickfix for this?

Hexcles:
This is definitely caused by #15493 . I thought I verified all platforms (Linux, Mac & Win) but apparently Mac doesn't save chromedriver as an artifact. Hopefully it'd be an easy fix by tweaking the build configuration in Chromium. If John can't spot anything obvious, we can revert #15493 .

JohnChen0:
I've found the bug in the Mac build config. Will prepare a fix shortly.

JohnChen0:
The fix has been checked into Chromium source repo (https://crrev.com/645038), and we should have a Mac build with ChromeDriver within a few hours.

JohnChen0:
The first Mac build with ChromeDriver has been created, with ChromeDriver binary at https://storage.googleapis.com/chromium-browser-snapshots/Mac/645088/chromedriver_mac64.zip.

foolip:
@JohnChen0 cool, thanks! Should this fix the problem right away, or does it have to reach the dev channel somehow first?

foolip:
[...]

PR #16112, opened 2019-03-27
[Azure Pipelines] Skip checkout for wpt.fyi hook jobs

foolip:
No checkout is needed. This will save ~40s, which for PRs could matter.


foolip:
Azure Pipelines failures are because of https://github.com/web-platform-tests/wpt/issues/16111

Hexcles (approved):
LGTM

PR #16117, opened 2019-03-27
Allow run-all jobs to run on non-master branches

jgraham:


tools/ci/run_tc.py
```diff
@@ -191,9 +191,9 @@ def set_variables(event):
 
 
 def include_job(job):
-    # Special case things that unconditionally run on master
+    # Special case things that unconditionally run on pushes,
+    # assuming a higher layer is filtering the required list of branches
     if (os.environ["GITHUB_PULL_REQUEST"] == "false" and
```
Hexcles: Is this check necessary? IIUC, `.taskcluster.yml` dictates whether `run-all` is given, and we don't use `run-all` in PRs.

jgraham: It was intended to prevent people using `run-all` for tasks that also run on PRs. Maybe there's a more principled approach where we make test-jobs handle these cases correctly, but I think this is a good enough fix for now (if it works!).

Hexcles: So it is indeed not necessary but more of defensive programming. I'm fine with keeping it for now and let's keep the fix minimal.

Hexcles (approved):

PR #16118, opened 2019-03-27
Terminate Edge browser process after test completion

mustjab:
Edge driver doesn't automatically close Edge browser processes on exit and we end up with browser processes that linger after test completion and affect subsequent tests. This change will terminate any MicrosoftEdge* processes if they're still found after Edge driver process exits. 

Test results: https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=11315

### DIFF
--- a/tools/wptrunner/wptrunner/browsers/edge.py
+++ b/tools/wptrunner/wptrunner/browsers/edge.py
@@ -85,12 +85,20 @@ class EdgeBrowser(Browser):
 
     def stop(self, force=False):
         self.server.stop(force=force)
-        time.sleep(3)
-        # close Edge process if it is still running
+        # Wait for Edge browser process to exit if driver process is found
         edge_proc_name = 'MicrosoftEdge.exe'
-        procs = subprocess.check_output(['tasklist', '/fi', 'ImageName eq ' + edge_proc_name])
+        for i in range(0,5):
+            procs = subprocess.check_output(['tasklist', '/fi', 'ImageName eq ' + edge_proc_name])
+            if 'MicrosoftWebDriver.exe' not in procs:
+                # Edge driver process already exited, don't wait for browser process to exit
+                break
+            elif edge_proc_name in procs:
+                time.sleep(0.5)
+            else: break
+
         if edge_proc_name in procs:
-            subprocess.call(['taskkill.exe', '/f', '/im', 'microsoftedge*']) # close all Edge browser instances
+            # close Edge process if it is still running
+            subprocess.call(['taskkill.exe', '/f', '/im', 'microsoftedge*'])
 
     def pid(self):
         return self.server.pid
--- a/tools/wptrunner/wptrunner/browsers/edge.py
+++ b/tools/wptrunner/wptrunner/browsers/edge.py
@@ -91,7 +91,6 @@ class EdgeBrowser(Browser):
         procs = subprocess.check_output(['tasklist', '/fi', 'ImageName eq ' + edge_proc_name])
         if edge_proc_name in procs:
             subprocess.call(['taskkill.exe', '/f', '/im', 'microsoftedge*']) # close all Edge browser instances
-        subprocess.call(['taskkill.exe', '/f', '/im', 'microsoftwebdriver.exe']) # close all Edge driver instances
         self.logger.info("stopped")
 
     def pid(self):
--- a/tools/wptrunner/wptrunner/browsers/edge.py
+++ b/tools/wptrunner/wptrunner/browsers/edge.py
@@ -91,7 +91,6 @@ class EdgeBrowser(Browser):
         procs = subprocess.check_output(['tasklist', '/fi', 'ImageName eq ' + edge_proc_name])
         if edge_proc_name in procs:
             subprocess.call(['taskkill.exe', '/f', '/im', 'microsoftedge*']) # close all Edge browser instances
-        self.logger.info("stopped")
 
     def pid(self):
         return self.server.pid
--- a/tools/wptrunner/wptrunner/browsers/edge.py
+++ b/tools/wptrunner/wptrunner/browsers/edge.py
@@ -1,4 +1,5 @@
 from __future__ import print_function
+import time
 import subprocess
 from .base import Browser, ExecutorBrowser, require_arg
 from ..webdriver_server import EdgeDriverServer
--- a/tools/wptrunner/wptrunner/browsers/edge.py
+++ b/tools/wptrunner/wptrunner/browsers/edge.py
@@ -84,6 +84,14 @@ class EdgeBrowser(Browser):
 
     def stop(self, force=False):
         self.server.stop(force=force)
+        time.sleep(3)
+        # close Edge process if it is still running
+        edge_proc_name = 'MicrosoftEdge.exe'
+        procs = subprocess.check_output(['tasklist', '/fi', 'ImageName eq ' + edge_proc_name])
+        if edge_proc_name in procs:
+            subprocess.call(['taskkill.exe', '/f', '/im', 'microsoftedge*']) # close all Edge browser instances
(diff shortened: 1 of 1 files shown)


tools/wptrunner/wptrunner/browsers/edge.py
```diff
@@ -82,6 +84,13 @@ def start(self, **kwargs):
 
     def stop(self, force=False):
         self.server.stop(force=force)
+        time.sleep(3)
+        # close Edge process if it is still running
+        edge_proc_name = 'MicrosoftEdge.exe'
+        procs = subprocess.check_output(['tasklist', '/fi', 'ImageName eq ' + edge_proc_name])
+        if edge_proc_name in procs:
+            subprocess.call(['taskkill.exe', '/f', '/im', 'microsoftedge*']) # close all Edge browser instances
+        self.logger.info("stopped")
```
foolip: Did you mean to include this or was it for debugging? Can you move the logging inside the if branch and make it a bit more informative?

foolip (approved):
LGTM % nit

tools/wptrunner/wptrunner/browsers/edge.py
mustjab: This was used for debugging, removed it.

foolip (approved):

gsnedders (changes requested):
This doesn't fix the fundamental raciness we have here. (Reviewing r- to block merging while I actually write something detailed.)

tools/wptrunner/wptrunner/browsers/edge.py
```diff
@@ -84,6 +85,12 @@ def start(self, **kwargs):
 
     def stop(self, force=False):
         self.server.stop(force=force)
```
gsnedders: Things have already started to go wrong by this call to `self.server.stop`.

We call this function from: https://github.com/web-platform-tests/wpt/blob/2808e3adc6d5ee9bea8e0f1002a755d917ece2e7/tools/wptrunner/wptrunner/testrunner.py#L675

Note the line just before this: https://github.com/web-platform-tests/wpt/blob/2808e3adc6d5ee9bea8e0f1002a755d917ece2e7/tools/wptrunner/wptrunner/testrunner.py#L673

This `stop` message is sent *asynchronously* and goes through the [main testrunner loop](https://github.com/web-platform-tests/wpt/blob/2808e3adc6d5ee9bea8e0f1002a755d917ece2e7/tools/wptrunner/wptrunner/testrunner.py#L89), ending it, [making us return from running](https://github.com/web-platform-tests/wpt/blob/2808e3adc6d5ee9bea8e0f1002a755d917ece2e7/tools/wptrunner/wptrunner/testrunner.py#L151) and [exiting the context manager](https://github.com/web-platform-tests/wpt/blob/2808e3adc6d5ee9bea8e0f1002a755d917ece2e7/tools/wptrunner/wptrunner/testrunner.py#L68) where we follow our the `teardown` method including:

https://github.com/web-platform-tests/wpt/blob/2808e3adc6d5ee9bea8e0f1002a755d917ece2e7/tools/wptrunner/wptrunner/testrunner.py#L83

And that calls:

https://github.com/web-platform-tests/wpt/blob/2808e3adc6d5ee9bea8e0f1002a755d917ece2e7/tools/wptrunner/wptrunner/executors/executorwebdriver.py#L231

(Which just sends `DELETE /session/{session id}`)

Now, back on this process, note that this line (`self.server.stop`) calls [this](https://github.com/web-platform-tests/wpt/blob/2808e3adc6d5ee9bea8e0f1002a755d917ece2e7/tools/wptrunner/wptrunner/webdriver_server.py#L83), which on Windows calls `TerminateProcess` on the `MicrosoftWebDriver.exe`.

So we're racing `DELETE /session/{session id}` against outright terminating the process. This seems less than ideal.

As @mustjab said by email:

> It looks like the test rely on the driver to close sessions when it exits, but for Edge we don’t do that.

We shouldn't be relying on this. We should wait for the session being deleted (somehow!) before we terminate the WebDriver server. @jgraham do you have any suggestion as to how to implement this? We don't have access to the executor here, so we can't check if it's closed the session by running any idempotent command and seeing how if/how it fails.

tools/wptrunner/wptrunner/browsers/edge.py
```diff
@@ -84,6 +85,12 @@ def start(self, **kwargs):
...
+        time.sleep(3)
```
gsnedders: Can we avoid waiting this long by polling anything? Are we waiting for some process to terminate (the WebDriver server)? Do we need to wait after calling `taskkill.exe` below too?

mustjab: I can keep polling here(every 500 ms?) for tasklist to have no MicrosoftEdge.exe processes running. I expect that server exe exits pretty quickly and we're waiting on Edge browser windows to exit.

gsnedders: We can poll pretty quickly depending on what order of magnitude we expect it to take, really. You'd probably have to just try and see how long it generally takes for them all to exit.

mustjab: Right now I don't see them exiting at all, so we end up terminating Edge processes after every test. Once we have proper fix in place, then I'll be able to test and see how long it usually takes for process to exit.

foolip:
Is there a path forward for this PR? The most recent Edge run failed:
https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=11475

https://wpt.fyi/runs shows the most recent 4 missing, so anything to get results would be great.

tools/wptrunner/wptrunner/browsers/edge.py
jgraham: If calling `quit` in WebDriver ends up with the WebDriver process exiting then the `Browser` implementation for Edge can implement a `stop` function that initially just waits for the process, and only force-terminates it after a timeout. This is essentially what Firefox does. It's an assumption that you never call `stop` on the browser unless either you've already signalled for WebDriver to stop the process, or `force` is set to `True`, which is not ideal but works OK in practice.

gsnedders:
So I tried writing something to fix this differently, and it turns out the shutdown/restart code doesn't work as myself/@jgraham thought it did.

tools/wptrunner/wptrunner/browsers/edge.py
mustjab: @gsnedders I've updated the wait logic to reduce wait time and only wait if we see driver process still running. If we don't see driver process running, then we assume it already got closed after issuing close command to browser and we terminate any remaining browser instances.

gsnedders:
No idea why TC hasn't run here, let's close/reopen to trigger that…

tools/wptrunner/wptrunner/browsers/edge.py
```diff
@@ -84,6 +85,20 @@ def start(self, **kwargs):
...
+        # Wait for Edge browser process to exit if driver process is found
+        edge_proc_name = 'MicrosoftEdge.exe'
+        for i in range(0,5):
+            procs = subprocess.check_output(['tasklist', '/fi', 'ImageName eq ' + edge_proc_name])
+            if 'MicrosoftWebDriver.exe' not in procs:
+                # Edge driver process already exited, don't wait for browser process to exit
+                break
+            elif edge_proc_name in procs:
+                time.sleep(0.5)
+            else: break
```
gsnedders: The break should be on a new line here (this is the CI failure, flake8 complaining about two statements on one line).

gsnedders:
Aside from that, as I said in email, I think we should land this given it does improve the status, even if the shutdown/restart code is broken more generally.

tools/wptrunner/wptrunner/browsers/edge.py
foolip: I pushed https://github.com/web-platform-tests/wpt/pull/16118/commits/03a0a9fe7e0238f4df6daa2bf9c182d4c2ae69ee

gsnedders (approved):

PR #16122, opened 2019-03-27
Use step_timeout() to force spin the EventLoop.

danyao:
Test improvement for w3c/payment-request#854.

Empirically in Chrome, `await Promise.resolve()` doesn't cause what comes after it to be run in the next event loop tick, which the test requires. My best understanding of the ES6 Job Queue is that this is expected. So changing this test to use `step_timeout` which guarantees a new tick.

Also properly catch the rejection of `eventPromise` because it happens before `acceptPromise` resolves.


marcoscaceres (approved):

marcoscaceres:
TIL step_timeout() - I've not used it before! thank you.

romandev (approved):
lgtm!

marcoscaceres:
going to try re-triggering task cluster.

marcoscaceres:
hmm.... don't have rights to restart it... cheap solution might be too send an empty commit?

danyao:
Taskcluster keeps on failing, and this is always the last error:
```
/home/test/start.sh: line 40: tools/ci/start.sh: No such file or directory
```
@marcoscaceres Any idea what might be wrong?

marcoscaceres:
@jgraham sorry to bother you but we are a bit stuck with a TaskCluster failure above. Any chance you could take a look and advise us on how to proceed?

jgraham:
Rebase onto latest master. Unfortunately we read the taskcluster configuration from the PR head and try to test against the provisional merge into master and when those things mismatch we can run into problems.

danyao:
Thanks @jgraham. Somehow I completely messed up this pull request after rebasing onto upstream/master. I'll start a new PR.

PR #16124, opened 2019-03-27
Added two new test to evaluate the break-word+break-spaces combination.

javifernandez:
These 2 new tests are intended to validate the cases where the combination of break-word+break-spaces allow to break before the first white-space after a word.

### DIFF
--- a/css/css-text/overflow-wrap/overflow-wrap-break-word-008.html
+++ b/css/css-text/overflow-wrap/overflow-wrap-break-word-008.html
@@ -1,0 +1,40 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Text Test: overflow-wrap:break-word + white-space:break-spaces</title>
+<link rel="author" title="Javier Fernandez" href="mailto:jfernandez@igalia.com" />
+<link rel="help" title="3. White Space and Wrapping: the white-space property" href="https://drafts.csswg.org/css-text-3/#white-space-property">
+<link rel="help" title="5.5. Overflow Wrapping: the overflow-wrap/word-wrap property " href="https://drafts.csswg.org/css-text-3/#overflow-wrap-property">
+<link rel="help" href="https://drafts.csswg.org/css-text-3/#valdef-overflow-wrap-break-word">
+<link rel="help" href="https://drafts.csswg.org/css-text-3/#valdef-white-space-break-spaces">
+<meta name="flags" content="ahem">
+<link rel="match" href="reference/overflow-wrap-break-word-001-ref.html">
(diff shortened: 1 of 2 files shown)


css/css-text/overflow-wrap/overflow-wrap-break-word-008.html
```diff
@@ -0,0 +1,40 @@
...
+  color: red;
+  width: 4ch;
+  z-index: -1;
+
+  white-space: break-spaces;
+  overflow-wrap: break-word;
+</style>
+
+<p>Test passes if there is a <strong>filled green square</strong> and <strong>no red</strong>.</p>
+<div class="red">XXXX<br> <br>XXXX<br></div>
```
kojiishi: Are you sure the spec requires to break before the space in this case? I thought `break-spaces` allows up to one space to hang. This result doesn't seem to match to the original intention, though I might have misunderstood the original intention of `break-spaces`.

javifernandez: Well, this test is almost identical to[ overflow-wrap-break-word-003.html](https://github.com/web-platform-tests/wpt/blob/master/css/css-text/overflow-wrap/overflow-wrap-anywhere-003.html), except for the use of the Ahem font. Such test has been created by @frivoal long time ago. 

I recall some explicit statement about allowing to break before the first space after a word in some version of the spec, but it seems it's been removed. However, there is this text in the step 4 of the [Phase II: Trimming and Positioning](https://drafts.csswg.org/css-text-3/#white-space-phase-2):

> If white-space is set to break-spaces, hanging or collapsing the advance width of spaces at the end of the line is not allowed; those that overflow must wrap to the next line.

kojiishi: I see, so it overflows, and then `overflow-wrap` kicks in.

Can you then also add another case, without `break-word`, to test in that case this line overflows (am I correct)?

javifernandez: There is indeed a test that covers precisely that case, see the test [break-spaces-002](https://github.com/web-platform-tests/wpt/blob/master/css/css-text/white-space/break-spaces-002.html)

javifernandez: One additional case that could add some value would be to make visible the overflow caused by the single white-space. Unfortunately, I've been trying to figure out how to implement such test without much success, so far.

kojiishi: Thank you for the info, break-spaces-002 is what I meant.

kojiishi (approved):

PR #16135, opened 2019-03-28
Fix manifest regeneration differences

### LINKED ISSUE #16125: wpt manifest -r produces different result to updating from current download
```
rm MANIFEST.json
./wpt manifest
mv MANIFEST.json MANIFEST-from-download.json
./wpt manifest --no-download -r
diff -u MANIFEST-from-download.json MANIFEST.json
```

This should produce an empty diff, it doesn't (not by a long shot!). We have bugs.

gsnedders:
Fixes #16125.

### DIFF
--- a/tools/manifest/item.py
+++ b/tools/manifest/item.py
@@ -63,15 +63,11 @@ class URLManifestItem(ManifestItem):
 
     def __init__(self, tests_root, path, url_base, url, **extras):
         super(URLManifestItem, self).__init__(tests_root, path)
+        assert url_base[0] == "/"
         self.url_base = url_base
+        assert url[0] != "/"
         self._url = url
         self._extras = extras
-
-    @property
-    def _source_file(self):
-        """create a SourceFile for the item"""
-        from .sourcefile import SourceFile
-        return SourceFile(self._tests_root, self.path, self.url_base)
 
     @property
     def id(self):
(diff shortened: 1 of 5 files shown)


jgraham:
Do we have any way to add testing here?

gsnedders:
@jgraham this now has a test (really a whole set of unit tests for `TypeData` would be good)

jgraham (approved):

gsnedders:
@Hexcles are you okay with us just landing this, given it bumps the version? the manifest format shouldn't change, though note the `url_base` is never included in the `url` for items now (it was previously, but that was a bug)

gsnedders:
(These force-pushes just tidy up the commit messages a bit)

Hexcles:
@gsnedders IIUC, the manifest version bump will force wpt-manifest to regenerate the manifest from scratch, disregarding the existing manifest. Do you have any estimation how big the diff would be?

We'll need to manually roll this change into Chromium and regenerate the base manifest, but nothing too complicated.

gsnedders:
@Hexcles you haven't done a clean generation in a long time, I assume? every single URL will change therefore to not include the url_base. so, uh, "big".

gsnedders:
@Hexcles but we can just land this, therefore?

Hexcles:
Yes. If the diff is big we'd actually prefer a bump in version number, so this is good.

Hexcles:
@gsnedders so you said `url_base` was removed from paths, but the reference URLs still have it (e.g. `css/foo.html == /css/foo-ref.html`). Is this intentional?

gsnedders:
@Hexcles So, uh, "yes". I'm not super happy about it, but all other URLs (aside from test URLs) in the manifest appear as absolute paths. :/

Issue #16081, opened 2019-03-26
Support downloading the manifest in other compressed formats


gsnedders:
Per discussion on IRC:

11:16 < jgraham> https://github.com/web-platform-tests/wpt-actions-test/releases/tag/merge_commit_01d8001cd1172d07239760d8be2cf980ff743924 I made some compressed manifests for you, if you want to benchmark download and extract time (all made using the latest version of the relevant tool on Ubuntu using the highest compression setting but without additional tricks like prebuilt dictionaries)
11:50 < gsnedders> jgraham: I did this locally yseterday ;P
11:50 < gsnedders> (prebuilt dictionaries are unlikely to be a big gain with the size of data we have)
11:51 < jgraham> gsnedders: The point was "on GitHub" to allow a more accurate picture of download + decompress time
11:51 < gsnedders> sure, but I expect a lot of that depends on proximity to edge nodes wherever GitHub host the thing and connection speed
11:52 < gsnedders> certainly I'm concerned enough about the slow connection case that I'd e.g. say bz2 is worth it
11:52 < gsnedders> but I also work-from-train relatively frequently
11:52 < jgraham> Of course, but it at least allows you to test reasonable cases
11:53 < jgraham> And you can probably set up throttling if you want to check the work from (wherever) case
11:53 < gsnedders> (Also question of what hardware we're decompressing on; nothing I have is that recent nowadays given how slow perf gains have been with CPUs…)
11:56 < jgraham> gsnedders: At least on my system bzip2 takes 300ms to decompress compared to 90ms for gzip and maybe 35 for zstd (not sure I got that right)
11:57 < jgraham> And 50ms for brotli
11:58 < gsnedders> 280ms v. 70ms v. 30ms here
11:58 < jgraham> So yeah, I'm pretty skeptical about taking on 200-250ms of overhead on the decompress time
11:59 < gsnedders> So I think we need to always have at least gz or bz2 given we can rely on Python being able to decode them
11:59 < gsnedders> IIRC, bz2 is worth it if the download is < 16Mbps
12:02 < gsnedders> and I only just get over that on my home connection (VDSL2, sync'd at 76.6Mbps), presumably msotly due to the cost of TCP ramp-up
12:03 < gsnedders> downloading it multiple times, I often don't download speeds over 16Mbps, which implies we *do* want the cost of decompressing it
[...]

PR #16136, opened 2019-03-28
Support downloading bzip2 manifests

gsnedders:
Part of #16081

### DIFF
--- a/tools/manifest/download.py
+++ b/tools/manifest/download.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 import argparse
+import bz2
 import gzip
 import json
 import io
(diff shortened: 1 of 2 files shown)


tools/manifest/download.py
```diff
@@ -62,13 +76,15 @@ def github_url(tags):
...
-            if item["name"].startswith("MANIFEST-") and item["name"].endswith(".json.gz"):
-                return item["browser_download_url"]
-            elif item["name"] == "MANIFEST.json.gz":
-                return item["browser_download_url"]
+            score = score_name(item["name"])
+            if score is not None:
+                candidates.append((score, item["browser_download_url"]))
+
+        if candidates:
+            _, winner = sorted(candidates)[0]
```
jgraham: I think I would redesign this so we return the sorted list of all manifests, so that we can fall back to another format if we download a corrupt file or something.

jgraham (changes requested):

tools/wpt/commands.json
```diff
@@ -11,6 +11,6 @@
     "tests-affected": {"path": "testfiles.py", "script": "run_tests_affected", "parser": "get_parser_affected",
                        "help": "Get a list of tests affected by changes", "virtualenv": false},
     "install": {"path": "install.py", "script": "run", "parser": "get_parser", "help": "Install browser components",
-                "install": ["mozdownload", "mozinstall"]},
+                "install": ["mozdownload", "mozinstall", "zstandard"]},
```
gsnedders: Why's this under install?

gsnedders (commented):
Imagine I clicked approve.

jgraham (approved):

PR #16137, opened 2019-03-28
These tests should use Ahem font

javifernandez:
When using 'ch' units for the width property's value some engines experience rounding issues that lead to false failures. The use of Ahem doesn't affect the purpose of the test and gives more accuracy results.


Hexcles:
Sigh, looks like you'd need to rebase the branch to fix the Taskcluster failures.

mrego (approved):
LGTM.

Probably you could mention on the message that the browser having issues with this is Safari and link to the bug https://bugs.webkit.org/show_bug.cgi?id=196169

PR #16142, opened 2019-03-28
Use step_timeout() to force spin the EventLoop

danyao:
Test improvement for w3c/payment-request#854.

Empirically in Chrome, await Promise.resolve() doesn't cause what comes after it to be run in the next event loop tick, which the test requires. My best understanding of the ES6 Job Queue is that this is expected. So changing this test to use step_timeout which guarantees a new tick.

Also properly catch the rejection of eventPromise because it happens before acceptPromise resolves.


marcoscaceres (approved):

marcoscaceres:
🤞it works this time.

rsolomakhin (commented):
LGTM!

PR #16168, opened 2019-04-01
Modified and improved considerably calc-serialization-002.html

TalbotG:
The modifications and improvements are related to discussion we have (and are still having) in issue 3741
https://github.com/w3c/csswg-drafts/issues/3741

### DIFF
--- a/css/css-values/calc-serialization-002.html
+++ b/css/css-values/calc-serialization-002.html
@@ -4,25 +4,18 @@
 
   <title>CSS Values and Units: serialization of calc() specified values: 17 arithmetical operations (complex)</title>
 
-  <!--
-
-  Latest update: March 31st 2019
-
-  Mixed absolute length units in calc() specified values
-  should *not* be converted into px unit when serializing
-  calc() values
-
-  More info:
-  Issue 3741: [css-values-4] Should physical units in specified
-  values in calc() collapse into px unit when serialized?
-  https://github.com/w3c/csswg-drafts/issues/3741
-
-  -->
-
   <link rel="author" title="Gérard Talbot" href="http://www.gtalbot.org/BrowserBugsSection/css21testsuite/">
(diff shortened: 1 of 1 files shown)


TalbotG:
**Elika and/or Tab should be reviewing this Pull Request.**

One sub-test should trigger feedbacks. At line 87:

`  verifySerialization("calc(8pt + calc(9rem + 10px))", "calc(8pt + 10px + 9rem)", "testing calc(8pt + calc(9rem + 10px))");`

How to serialize `calc(8pt + calc(9rem + 10px))` ? I see 3 possible answers, after carefully reading the current spec:

a) `calc(8pt + calc(10px + 9rem)) `
b) `calc(8pt + (10px + 9rem))` 
c) `calc(8pt + 10px + 9rem) `

TalbotG:
> I think flattening nested calc() to parentheses seems reasonable. (...) I agree with this.

coming from
[Re: [css-values] Serialization of nested calc()](https://lists.w3.org/Archives/Public/www-style/2016Apr/0238.html)
So, possible answer a) is ruled out. There is still b) and c) left.

fantasai (approved):

fantasai:
I think this test needs to be largely rewritten after the changes to serialization rules (which, for better or worse, seems like they collapse all the absolute units at specified value time). But the changes in this PR are good, so my suggestion is to merge this and then apply the serialization rule changes in a separate PR.

TalbotG:
New calc-serialization test:

http://www.gtalbot.org/BrowserBugsSection/CSS3Values/calc-serialization-002-new.html

TalbotG:
In a few min., I will be git-submitting the new calc-serialization-002 test. It will have to be reviewed since it is an entirely updated test.

TalbotG:
I will create another pull request as I do not understand what to do and how to fix whatever needs to be fixed here.

PR #16170, opened 2019-04-01
Added parentheses to check minimum of 32 nested pairs inside calc

TalbotG:
- I added 8 pairs of nested parentheses to the original test
- I corrected title text
- I improved the text assert
- I referenced issue 3462
https://github.com/w3c/csswg-drafts/issues/3462
in a comment
@FremyCompany

### DIFF
--- a/css/css-values/calc-parenthesis-stack.html
+++ b/css/css-values/calc-parenthesis-stack.html
@@ -4,11 +4,17 @@
 	<meta charset="utf-8">
 	<title>
 		CSS Values and Units Test:
-		Calc() inside calc()
+		32 nested pairs of parentheses inside calc()
 	</title>
 	<meta name="assert" content="
-		The calc() function notation is allowed inside a calc() notation.
+		This test checks the support for the minimum required number of 32 nested pairs of parentheses inside a calc() function.
 	" />
+
+  <!--
+  More info:
+  [css-values] Limit nested pairs of parentheses in calc to 32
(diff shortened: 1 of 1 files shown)


frivoal:
Blocked on https://github.com/w3c/csswg-drafts/issues/3462

FremyCompany:
Sounds fine by me, pending resolution as mentioned by @frivoal above.

FremyCompany (commented):
We might want to add one more nesting level, if we resolve to clarify the nesting level not to include the `calc()` itself. See https://github.com/w3c/csswg-drafts/issues/3462#issuecomment-478579269

TalbotG:
@FremyCompany
François,
You want one more nesting level in the test, right? 33 pairs of parentheses. Or 32 **nested pairs** of parentheses **inside**  `calc()` , right?

FremyCompany:
@TalbotG If the tests still passes with 32 nested parentheses inside the `calc()` then yes. Could you check if all browsers (that support 31 right now) support that? If yes, I'd say we should update the test. If not, maybe let's leave it as is.

TalbotG:
Blink and WebKit engines fails with 32 nested parentheses inside the calc():

http://www.gtalbot.org/BrowserBugsSection/CSS3Values/HowManyNestedParenthesesCalc.html

FremyCompany:
Okay, then let's approve as-is. Thanks for taking a look!

FremyCompany (approved):

PR #16174, opened 2019-04-01
Run base wptrunner tests on python 3.

Ms2ger:


.azure-pipelines.yml
```diff
@@ -105,6 +105,7 @@ jobs:
     vmImage: 'macOS-10.13'
   steps:
   - template: tools/ci/azure/checkout.yml
+  - script: brew install python
```
gsnedders: @foolip would you rather we create a new file for this, or put it in `tools/ci/azure/tox_pytest.yml`? (note that `python` here, without a version specifier, is Python 3)

tools/wptrunner/tox.ini
```diff
@@ -2,7 +2,7 @@
 xfail_strict=true
 
 [tox]
-envlist = py27-{base,chrome,edge,firefox,ie,opera,safari,sauce,servo}
+envlist = py27-{base,chrome,edge,firefox,ie,opera,safari,sauce,servo},py36-base
```
gsnedders: Let's add `py37-base` here as well for those who don't don't have Python 3.6 installed (such as on Azure!).

tools/wptrunner/wptrunner/tests/test_products.py
```diff
@@ -18,7 +19,9 @@ def test_load_active_product(product):
     # test passes if it doesn't throw
 
 
-@all_products("product")
+@all_products("product", marks={
+    "firefox": pytest.mark.xfail(sys.version[0] == "3", reason="mozinfo 0.10 doesn't support py3"),
```
gsnedders: mozinfo 1.0.0 shipped six months ago, why does @pyup-bot not realise it exists?!

.azure-pipelines.yml
foolip: I think it's important that we use the system python by default, so in `tools/ci/azure/tox_pytest.yml` would work, or just at the top of the new jobs that would use py3. I'd probably use https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops to vary just the version of Python used.

You could also check if https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/use-python-version?view=azure-devops works on macOS, the "not on Linux or macOS agents" bit isn't entirely clear so it might.

tools/wptrunner/wptrunner/tests/test_products.py
gsnedders: I've emailed PyUp support about this; in the mean time, try bumping the release manually and see if we can avoid this xfail?

Ms2ger: I'd prefer doing that when we're actually running this test in CI

tools/wptrunner/tox.ini
Ms2ger: wptrunner doesn't work on 3.7 due to #16178

.azure-pipelines.yml
Ms2ger: This change wasn't needed in the end; the bot already had 3.7 installed, just not 3.6

jgraham (approved):

.azure-pipelines.yml
gsnedders: Correct.

PR #16181, opened 2019-04-01
Use a single action to generate and tag the manifest.

jgraham:
This cuts down on quota usage. Also apply filters before the manifest
generation so we cut down on unnecessary work in PRs and return the
magic status 78 in cases where the filter doesn't match so if we expand
to multiple actions in the future things work correctly.


jgraham:
Tested at https://github.com/web-platform-tests/wpt-actions-test/actions/workflow-runs/MDEwOkNoZWNrU3VpdGU5MDI0NTM5MA==

gsnedders (approved):

PR #16184, opened 2019-04-01
[testdriver.js] Improve test coverage

jugglinmike:
@gsnedders Marionette rejects the "touch" input type, and Chrome doesn't fire pointer events for it. Do you have any thoughts on making this test easier to consume?

### DIFF
--- a/infrastructure/testdriver/actions/pointers.html
+++ b/infrastructure/testdriver/actions/pointers.html
@@ -1,0 +1,93 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>TestDriver actions: element position</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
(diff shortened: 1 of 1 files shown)


gsnedders:
Defer to people who know more about pointer events than me.

jugglinmike:
Got it. Have any recommendations?

gsnedders:
@jgraham for example.

jugglinmike:
@jgraham Would you mind taking a look?

jgraham:
I wonder if this should rather be a test under `webdriver` since it seems to be testing spec conformance rather than really anything related to testdriver (I agree that there's some overlap here since we can have testdriver backends that don't directly implement WebDriver semantics).

jugglinmike:
Agreed @jgraham; we should test both. I've reported the missing WebDriver coverage via gh-16407.

Issue #16132, opened 2019-03-28
No recent stable Firefox runs on wpt.fyi


LukeZielinski:
https://wpt.fyi/runs?max-count=100&label=stable

There haven't been any Firefox runs since March 12.

LukeZielinski:
@jgraham Hi James, are you able to take a look?

jgraham:
So there are at least two problems here. One is that we accidentially broke running epoch branches for a while; that was https://github.com/web-platform-tests/wpt/pull/16117. The other issue seems to be that we're trying to download geckodriver from the wrong location on beta/stable c.f. https://tools.taskcluster.net/groups/cG42DIXlS8qiHTMF0P5jSQ/tasks/fj5cgxa4QE66yuZqgIDEeQ/runs/0/logs/public%2Flogs%2Flive.log

jgraham:
Sorry, not geckodriver, the test prefs.

foolip:
https://tools.taskcluster.net/groups/cG42DIXlS8qiHTMF0P5jSQ is the most recent run, found via https://api.github.com/repos/web-platform-tests/wpt/commits/ea8b7ff29a15265e3bf73364f49e22426a5b97d5/statuses.

Hexcles:
Any updates? Looks like we are still not getting any Firefox stable runs.

jgraham:
I landed https://github.com/web-platform-tests/wpt/pull/16176 but since we only get stable runs once a day, it's hard to be sure if it made a difference.

foolip:
We're running Edge and Safari stable every 6 hours on Azure Pipelines now, perhaps we should increase the cadence for Taskcluster as well just to iterate more quickly in cases like this? Also makes a fluke missing run less serious.

Hexcles:
Looks like we have stable runs https://wpt.fyi/results/?run_id=5818496017498112&run_id=5378137953140736&run_id=5279026683838464&run_id=5419764172193792

Thanks, @jgraham !

foolip:
Yay, thanks @jgraham!

PR #16192, opened 2019-04-01
Run Chrome/Firefox stable on Taskcluster every 6 hours

foolip:
Context: https://github.com/web-platform-tests/wpt/issues/16132#issuecomment-478710345

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -12,7 +12,7 @@ tasks:
             $flatten:
               $match: {
                 event.ref == "refs/heads/master": [{name: firefox, channel: nightly}, {name: chrome, channel: dev}],
-                event.ref == "refs/heads/epochs/daily": [{name: firefox, channel: stable}, {name: chrome, channel: stable}],
+                event.ref == "refs/heads/epochs/six_hourly": [{name: firefox, channel: stable}, {name: chrome, channel: stable}],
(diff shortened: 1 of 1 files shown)


jgraham:
There' no technical reason not to do this, but if no one realy cares about the results then it's worth wondering whether it's just a waste of resources. I don't recal encountering anyone who cares about the results for stable and when they broke it took days for anyone to notice.

For the purpose of testing, I'd much rather have a way to specify the channel(s) used for unittests much like we can now write `tc-jobs:all`, or just make some browser install tests run against all channels.

foolip:
I do compare stable to dev results occasionally to look for regressions, and compute [results and interop metrics](https://groups.google.com/a/chromium.org/d/msg/ecosystem-infra/y9QFKyAk4Ls/94SbsTEMBgAJ) using stable channel data, but daily results really are enough for this.

https://wpt.fyi/results/webvtt/rendering/cues-with-video/processing-model/align_center.html?diff&filter=ADC&run_id=5279026683838464&run_id=6388726074703872 is an example of a seeming regression that I quickly found by comparing results. @AutomatedTester are you still interesting in comparing results from stable, beta and nightly in this way to spot incoming regressions? I guess daily runs are enough for that?

jgraham:
I'm going to close this to get it off the radar until we find a use case for the more regular runs.

PR #16203, opened 2019-04-02
Add a web-nfc test about NDEFMessage url

xiuqijix:


web-nfc/nfc_url-manual.https.html
```diff
@@ -0,0 +1,46 @@
...
+  assert_equals(event1.message.records[0].recordType, "empty");
+
+  const readerWatcher2 = new EventWatcher(t, reader2, ["reading", "error"]);
+  reader2.start();
+  const event2 = await readerWatcher2.wait_for("reading");
+  const record = event2.message.records[0];
+  assert_equals(record.recordType, "text");
+  assert_equals(record.data, "Hello World");
+  assert_equals(event2.message.url, `${location.origin}/custom/path/update`);
+}, "Test that url of NFCReaderOptions is correct.");
```
Honry: I suppose you are trying to test filtering relevant data sources via URL in `ReaderOptions`. 
So the test name should be something like `NFCReader_options_url-manual.html` and the test description could be more explicit like "Test that the url of NFCReaderOptions filters relevant data sources correctly.".

kaixinjxq:
@Honry, many thanks for your review! I just submit a new commit to address your comments, PTAL.

web-nfc/NFCReader_options_url-manual.https.html
```diff
@@ -0,0 +1,44 @@
...
+setup({ explicit_timeout: true });
+
+promise_test(async t => {
+  const writer = new NFCWriter();
+  const reader = new NFCReader({url: `${location.origin}/custom/path`});
+  await writer.push({
+    url: "/custom/path/update",
+    records: [{ recordType: "text", data: "valid" }]
+  });
+  await writer.push({
```
Honry: This is wrong. You pushed messages twice to an nfc tag, but the latest message stored in the tag is the second one you pushed.

You should read the message every time you pushed and verify the one you are going to filter.

Honry (approved):
Thanks, LGTM

